Issue #24734, opened 2026-05-31
[Bug]: column DEFAULT (uuid()) expression default not supported (v3.0.11 & 4.0-dev)


dengn:
### Summary
Functional/expression column defaults (MySQL 8.0.13+), e.g. `DEFAULT (uuid())`, are rejected. (Low priority — Prisma generates UUIDs client-side, so no Prisma impact; filing for MySQL-8 compatibility completeness.)

### Steps to Reproduce
```sql
CREATE TABLE t (id VARCHAR(191) NOT NULL DEFAULT (uuid()), n VARCHAR(50), PRIMARY KEY(id));
-- ERROR 20301 (HY000): invalid input: invalid default value for column 'id'
```

### Expected Behavior
MySQL 8.0.13+ allows expressions wrapped in parentheses as column defaults (`DEFAULT (uuid())`, `DEFAULT (now())`, etc.).

---
**Test context** — Found while testing **Prisma ORM (v5.22)** compatibility against MatrixOne. Reproduced on **both**:
- Prod cloud — `8.0.30-MatrixOne-v3.0.11`
- 4.0 dev line — `nightly-0bfda604` (`select version()` → `8.0.30-MatrixOne-v`)

(Results on the older v3.0.5 are intentionally omitted.)

VioletQwQ-0:
Fixed by merged PR #24795: https://github.com/matrixorigin/matrixone/pull/24795

The merged change allows parenthesized UUID expression defaults such as `DEFAULT (uuid())` on string columns and added unit/BVT regression coverage for the issue reproduction.

VioletQwQ-0:
Fixed by merged PR #24795

PR #24795, opened 2026-06-03
fix(plan): allow parenthesized uuid default on string columns

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24734

Allows explicit parenthesized expression defaults such as `DEFAULT (uuid())` on string columns, while preserving the existing UUID-type guard for bare `DEFAULT uuid()`.

Changes:
- Detect parenthesized default expressions before applying the bare `uuid()` type guard.
- Bind the inner expression for generated default values while preserving the original default string.
- Add unit and BVT regression coverage for `varchar NOT NULL DEFAULT (uuid())`.

Validation:
- `go test ./pkg/sql/plan -run 'TestBuildDefaultExprAllowsParenthesizedUuidForStringDefault|TestBuildDefaultExprKeepsBareUuidTypeGuard' -count=1`
- `git diff --check`
- BVT not run locally; covered by CI.


heni02 (approved):

ouyuanning (approved):

aunjgr (approved):
Clean MySQL compatibility fix. The old code rejected `DEFAULT (uuid())` on non-UUID columns because it didn't unwrap `ParenExpr` before checking the function name.

Fix: detect `ParenExpr` wrapping, unwrap for the function check but preserve it for `OriginString`. Uses `isExpressionDefault` flag to distinguish parenthesized expression defaults (MySQL 8 style) from bare `uuid()` calls. Bare UUID on non-UUID columns is still correctly rejected.

BVT covers CREATE, SHOW CREATE, INSERT, and output. LGTM.

XuPeng-SH (changes requested):
Blocking correctness issue:

The new `ParenExpr` handling unwraps parentheses only *after* the existing special-case `NULL` checks. That breaks the long-standing semantics that `DEFAULT (NULL)` should be equivalent to `DEFAULT NULL`.

In `pkg/sql/plan/build_util.go`, the JSON / GEOMETRY / NOT NULL guards still call `isNullAstExpr(expr)` before the new:

```go
if paren, ok := expr.(*tree.ParenExpr); ok {
    isExpressionDefault = true
    expr = paren.Expr
}
```

But `isNullAstExpr()` only recognizes a raw `*tree.NumVal` null literal. So `DEFAULT (NULL)` no longer hits those earlier NULL checks.

That creates real semantic regressions, for example:
- `NOT NULL DEFAULT (NULL)` can bypass the existing invalid-null-default guard.
- Types that explicitly allow only `DEFAULT NULL` here (such as JSON / GEOMETRY) will treat `(NULL)` as a non-NULL expression instead.

Please normalize/unwrap parenthesized expressions before the NULL-specific checks as well, while still preserving the original AST separately for `OriginString`.

XuPeng-SH (approved):
Latest head fixes the earlier blocker around `DEFAULT (NULL)` by normalizing parenthesized expressions before the NULL-specific checks while still preserving the original default string. I did not find another blocking issue in the parenthesized `DEFAULT (uuid())` handling on string columns.

Issue #24735, opened 2026-05-31
[Bug]: MATCH() AGAINST(? ...) with a bound parameter rejected by parser — breaks Prisma fullTextSearch (v3.0.11 & 4.0-dev)


dengn:
### Summary
A **literal** `MATCH(col) AGAINST('x' IN BOOLEAN MODE)` works, but using a **bound parameter** as the AGAINST argument — `MATCH(col) AGAINST(? IN BOOLEAN MODE)` — is rejected by the parser, **even when a FULLTEXT index exists**. Prisma's `fullTextSearch` preview always parameterizes the search term, so the entire feature is unusable on MatrixOne.

### Steps to Reproduce
```sql
CREATE TABLE ft (id INT PRIMARY KEY, t VARCHAR(200), FULLTEXT INDEX ft_idx(t));
INSERT INTO ft VALUES (1,'hello world');

-- literal argument: OK
SELECT id FROM ft WHERE MATCH(t) AGAINST('hello' IN BOOLEAN MODE);

-- bound parameter argument: FAILS
PREPARE st FROM 'SELECT id FROM ft WHERE MATCH(t) AGAINST(? IN BOOLEAN MODE)';
SET @q = 'hello';
EXECUTE st USING @q;
-- ERROR 1105 (HY000): You have an error in your SQL syntax ... syntax error
```

### Prisma impact
`prisma.post.findMany({ where: { title: { search: 'hello' } } })` and `orderBy: { _relevance: {...} }` both fail at runtime (the client binds the term as a parameter):
```
MysqlError { code: 1064, message: "SQL parser error: ... syntax error at ... near \"? IN BOOLEAN MODE))\"" }
```
Confirmed on prod v3.0.11 and 4.0-dev `nightly-0bfda604`, with a real FULLTEXT index present. Related: prepared-statement compatibility #24186.

### Expected Behavior
MySQL accepts a placeholder/parameter as the `AGAINST` search string in a prepared statement.

---
*Found during deep Prisma ORM (v5.22) compatibility testing; reproduced on prod `8.0.30-MatrixOne-v3.0.11` and `nightly-0bfda604`.*

VioletQwQ-0:
Fixed by merged PR #24796: https://github.com/matrixorigin/matrixone/pull/24796. The prepared MATCH ... AGAINST(?) path, runtime NULL/empty validation, and prepared-statement reuse coverage are included in the merged change.

PR #24796, opened 2026-06-03
fix(sql): support prepared fulltext boolean search

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24735

Supports prepared fulltext boolean search patterns such as `MATCH(col) AGAINST(? IN BOOLEAN MODE)`.

Changes:
- Parse fulltext search patterns as expressions for string literals and prepared statement parameters.
- Bind and cast fulltext search patterns to `varchar` before building `fulltext_match`.
- Keep fulltext index rewrite working for runtime pattern expressions.
- Reject NULL and empty runtime patterns before reading or building the fulltext query.
- Defensively reject empty pattern lists in the fulltext SQL layer.
- Reset source table-function completion state after failed executions so prepared statements can be reused.
- Add unit and BVT coverage for BOOLEAN/NL `valid -> invalid -> valid` prepared executions.

Validation:
- `make` in `pkg/sql/parsers/dialect/mysql`
- `go test ./pkg/sql/parsers/dialect/mysql -count=1`
- `go test ./pkg/sql/parsers/tree -count=1`
- `go test ./pkg/fulltext -count=1`
- `go test ./pkg/sql/colexec/table_function -count=1`
- `go test ./pkg/sql/plan/explain -run TestFullTextSql -count=1`
- `go test ./pkg/sql/plan -run 'FullText|fullText|fulltext' -count=1`
- `git diff --check`
- BVT not run locally; covered by CI.


heni02 (approved):

ouyuanning (approved):

iamlinjunhong (approved):

aunjgr (changes requested):
Good direction — prepared statements for fulltext boolean search were broken and this is the right fix approach. But:

### Blocking

**`fulltext.ParsePattern` called with 3 arguments, but signature takes 2** (`pkg/sql/plan/fulltext.go`).

The new code calls:
```go
ps, err := fulltext.ParsePattern(pattern, mode, param.Parser)
```

But `pkg/fulltext/fulltext.go:946` has:
```go
func ParsePattern(pattern string, mode int64) ([]*Pattern, error)
```

`pkg/fulltext/fulltext.go` is not in this PR's file list, so the third `param.Parser` argument has no matching parameter. This won't compile. You need to either:
- Add the `param.Parser` parameter to `pkg/fulltext/fulltext.go:ParsePattern`
- Or adjust the call in `pkg/sql/plan/fulltext.go` to match the existing signature

### Correctness concern

**`equalsFullTextMatchFunc` switched from string comparison to `exprStructuralEqual`.** The old code directly compared `GetLit().GetSval()` values (nil-deref if pattern is not a literal — which is exactly the prepared-stmt case). The new `exprStructuralEqual` is correct and necessary, but it's broader than before (compares expression types too). Since the binder now casts pattern to varchar unconditionally, type mismatch shouldn't arise in practice. Fine, but worth noting in a comment.

### Test gap

BVT only covers `IN BOOLEAN MODE` with a single `?`. Missing:
- NATURAL LANGUAGE MODE prepared statement
- WITH QUERY EXPANSION
- Multiple MATCH expressions with placeholders in one query
- Mixed concrete-and-placeholder patterns

### Verified

- AST change (`Pattern` field from `string` to `Expr`) is internal-only, not serialized. Backward-compatible.
- Grammar change (`search_pattern` from `%type <str>` to `%type <expr>`, adding `VALUE_ARG` alternative) is backward-compatible.
- Removal of AST-construction round-trip in `apply_indices_fulltext.go` (-42 lines) is safe — directly constructing plan nodes instead.
- Mode always comes from the AST, never from the expression — no conflict path.

aunjgr (approved):
Re-reviewing — my prior concern about `fulltext.ParsePattern(pattern, mode, param.Parser)` having too many args was incorrect. Main's `pkg/fulltext/fulltext.go` now has `func ParsePattern(pattern string, mode int64, parser string)` — the `parser` parameter was added upstream. This PR is compatible.

Fulltext prepared statements now work for `MATCH ... AGAINST(? IN BOOLEAN MODE)`. AST change (`Pattern string` → `Expr`), grammar change (`%type <str>` → `%type <expr>`, `VALUE_ARG` alternative), binder preserves mode from AST, and `apply_indices_fulltext.go` removes the AST round-trip construction. LGTM.

pkg/sql/plan/base_binder.go
```diff
@@ -1197,7 +1197,18 @@ func (b *baseBinder) bindFullTextMatchExpr(astExpr *tree.FullTextMatchExpr, dept
...
+		return nil, err
+	}
+	if pattern.Typ.Id != int32(types.T_varchar) {
+		varcharTyp := types.T_varchar.ToType()
+		pattern, err = makePlan2CastExpr(b.GetContext(), pattern, makePlan2Type(&varcharTyp))
+		if err != nil {
+			return nil, err
+		}
+	}
+	args[0] = pattern
```
XuPeng-SH, line 1534: **High: preserve the non-empty/non-NULL pattern invariant at runtime.** The old literal path rejects `AGAINST("")` in `FullTextMatchExpr.Valid`, but a placeholder now reaches `fulltextState.start` without value validation. A NULL parameter is read with `GetStringAt` from a const-null vector and can panic; an empty BOOLEAN parameter makes `NewSearchAccum` return zero patterns, after which `PatternToSql` unconditionally indexes `ps[0]` and panics (`pkg/fulltext/sql.go:577`). In NATURAL LANGUAGE mode, the same empty value becomes `prefix_eq(word,"")`, which can scan the entire fulltext index instead of preserving the old rejection. Please define NULL semantics and restore the non-empty check before reading/building the runtime query, add a defensive zero-pattern check in the fulltext SQL layer, and cover BOOLEAN/NL empty and NULL plus repeated `valid -> invalid -> valid` executions of one prepared statement so reset/cleanup is verified.

XuPeng-SH (changes requested):
The prepared-pattern path has a deterministic runtime panic because the former planning-time non-empty invariant was not moved to the execution boundary. Targeted parser, plan, fulltext, and table-function packages pass, but none covers dynamic empty/NULL input.

pkg/sql/plan/base_binder.go
VioletQwQ-0: Addressed in fc146985e. The execution boundary now rejects const-NULL, row NULL, and empty patterns before GetStringAt or query construction; PatternToSql also returns an error for an empty pattern list instead of indexing ps[0]. Added BOOLEAN and natural-language table-function coverage for valid -> invalid -> valid reuse, a const-NULL regression, and matching prepared-statement BVT cases for empty/NULL recovery. Validation: pkg/fulltext, pkg/sql/colexec/table_function TestFullText, targeted pkg/sql/plan fulltext tests, parser TestValid, and git diff --check all pass locally. BVT is left to CI.

VioletQwQ-0: Follow-up fix pushed in 68b83b34a. CI exposed that the source table-function container kept ctr.isDone=true when start returned the expected empty/NULL error, so the next execution of the same prepared pipeline terminated before evaluating the valid parameter. TableFunction.Reset now restores isDone=false. Added TestSourceTableFunctionResetAfterStartError, which reproduces start error -> Reset -> successful Call through the operator lifecycle. Full table_function, pkg/fulltext, targeted fulltext planner tests, build, vet, and git diff --check pass after syncing upstream/main. The valid -> invalid -> valid BVT remains in place for CI verification.

fengttt (approved):

XuPeng-SH (approved):
Re-reviewed head 89a002a662. The previous blocker is fully addressed: dynamic NULL/empty patterns fail before value access or SQL construction, PatternToSql defensively rejects zero patterns, and source table-function Reset restores the execution generation after start errors. The repeated-Prepare cleanup also preserves the only compiler-injected state (optimized generate_series). Verified the generated parser is reproducible, the PR merges cleanly with current main, targeted parser/fulltext/planner/table-function tests pass, and the lifecycle matrix passes under the race detector. No remaining blocking correctness, performance, or unhappy-path issue found. Please sync the latest main before merge because the branch is currently behind.

XuPeng-SH (approved):

PR #24800, opened 2026-06-03
fix: fold variable expressions before remote run

### LINKED ISSUE #24799: Remote pipeline with variable expression panics when ResolveVariableFunc is missing
Background: Matrixflow issue https://github.com/matrixorigin/matrixflow/issues/10941 reports query_sql count failing with VarExpressionExecutor.Eval nil pointer panic.

Root cause: VarExpressionExecutor evaluates plan.Expr_V through proc.GetResolveVariableFunc(). Frontend/internal processes set this resolver, but remote-run reconstructs a receiver CN process from serialized process info and does not carry ResolveVariableFunc. A remote pipeline containing variable expressions can therefore panic while evaluating filter/projection expressions.

Fix direction:
- Avoid remote-running scopes that contain plan.Expr_V; run them locally so the original process keeps its resolver.
- Add a defensive nil check in VarExpressionExecutor.Eval to return an error instead of panicking.
- Add unit coverage for scope detection and nil resolver behavior.

LeftHandCold:
Fixes #24799.

## Summary
- fold `@@` variable expressions on the source CN before a scope is shipped to a remote CN, preserving the remote execution path
- deep-copy expression containers before folding and write copied expressions back to the parent field/slice/map, avoiding shared-plan mutation, data races, and prepare-plan cache pollution
- cover exported expression fields, slices/maps, agg argument expressions, and lockRows hidden expressions
- keep the VarExpressionExecutor nil-resolver guard so any missed case returns a normal error instead of panicking

## Testing
- `GOCACHE=/private/tmp/mo-gocache-var-expr-main-copy GOFLAGS=-mod=mod go test ./pkg/sql/colexec ./pkg/sql/compile -run 'TestVarExpressionExecutorWithoutResolveVariableFunc|TestScopeContainsVarExpr|TestFoldVarExprs' -count=1` is blocked in this local main checkout by existing usearch cgo build errors: `could not determine what C.usearch_* refers to`.


LeftHandCold:
已按后续 review 调整：RemoteRun 前不再切 MergeRun，而是在 source CN 对 `@@` var expr 做 fold；fold 前 `DeepCopyExpr` 并写回父 field/slice/map，避免并发 remote scope 共享表达式导致 data race，也避免 prepare plan 被第一次执行污染。getter-only 折叠路径已移除，agg args/lockRows 改成 rewriter 写回；对应 copy-on-write 回归测试已补。

ouyuanning (approved):

aunjgr (approved):
Practical fix for remote CN variable resolution. Same change across three branches (#24800 main, #24801 4.0-dev, #24802 3.0-dev).

**Approach**: Before sending a scope to a remote CN, walk the operator tree via reflection to find `VarRef` expressions, evaluate them locally (where the session variable context exists), and replace with constants. Uses `scopeContainsVarExpr` as a fast-path guard so the reflection-based folding only runs when needed.

**`remote_expr.go` (490 new lines)**: reflection-based traversal of operator structs looking for `*plan.Expr` fields (`[]*plan.Expr`, `[]plan.Expr`, and operator-base fields are handled). Two custom interfaces for operators that need special treatment:
- `argExpressionsGetter/Rewriter` for `AggFuncExecExpression` (aggregate arg expressions)
- `lockRowsExpressionsGetter/Rewriter` for `LockOp` (lock row filter expressions)

**Other changes**:
- `evalExpression.go`: nil guard on `resolveVariableFunc` — returns a clear error instead of nil-deref on remote CN
- `aggexec/types.go`: implements `RewriteArgExpressions` for `AggFuncExecExpression`
- `lockop/lock_op.go`: implements `GetLockRowsExpressions` / `RewriteLockRowsExpressions` for `LockOp`
- `scope.go`: removes dead `VarExpr` field from `Scope` (unused, confirmed by grep of `s.VarExpr`)

**Reflection safety**: only `plan.Expr` types are matched (via `planExprPtrType`); the walk handles pointers, slices, and operator-base embedding. `replaceVarRefExpr` only fires on actual `VarRef` nodes, so nothing gets accidentally constant-folded.

LGTM. Please also merge #24801 and #24802 (same fix, different target branches).

XuPeng-SH (approved):

PR #24801, opened 2026-06-03
fix: fold variable expressions before remote run

### LINKED ISSUE #24799: Remote pipeline with variable expression panics when ResolveVariableFunc is missing

LeftHandCold:
Fixes #24799.

## Summary
- fold `@@` variable expressions on the source CN before a scope is shipped to a remote CN, preserving the remote execution path
- deep-copy expression containers before folding and write copied expressions back to the parent field/slice/map, avoiding shared-plan mutation, data races, and prepare-plan cache pollution
- cover exported expression fields, slices/maps, agg argument expressions, and lockRows hidden expressions
- keep the VarExpressionExecutor nil-resolver guard so any missed case returns a normal error instead of panicking

## Testing
- `GOCACHE=/private/tmp/mo-gocache-var-expr-40-copy GOFLAGS=-mod=mod go test ./pkg/sql/colexec ./pkg/sql/compile -run 'TestVarExpressionExecutorWithoutResolveVariableFunc|TestScopeContainsVarExpr|TestFoldVarExprs' -count=1`


LeftHandCold:
已按后续 review 调整：RemoteRun 前不再切 MergeRun，而是在 source CN 对 `@@` var expr 做 fold；fold 前 `DeepCopyExpr` 并写回父 field/slice/map，避免并发 remote scope 共享表达式导致 data race，也避免 prepare plan 被第一次执行污染。getter-only 折叠路径已移除，agg args/lockRows 改成 rewriter 写回；对应 copy-on-write 回归测试已补。

ouyuanning (approved):

aunjgr (approved):
Same fix as #24800, targeting 4.0-dev. LGTM.

XuPeng-SH (approved):
Reviewed from three angles:
2. Performance and resource impact: changes are acceptable for the affected path.
3. Compatibility and maintainability: behavior and coverage look sufficient for merge.

Issue #24797, opened 2026-06-03
[CDC] show cdc reports paused while daemon task is still PauseRequested, causing resume failure


Ariznawlll:
## Summary

A CDC task can fail to resume because the externally visible CDC state and the internal daemon task state are inconsistent during pause/resume.

In the observed run, `show cdc task` reported the task as `paused`, but `resume cdc task` failed because the internal daemon task was still `PauseRequested`.

## Observed In

mo-nightly-regression run:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/26830965543

Failed step:

```text
Testcase Test / MO CDC MO TEST
10. resume update/delete cdc for full cdc test
```

Task name:

```text
cdc_resume_6kw
```

## Logs

Before pause, the task was running:

```text
2026-06-03 12:18:58 Show cdc_resume_6kw cdc test status
"task-name": "cdc_resume_6kw",
"state": "running",
"err-msg": ""
```

After `pause cdc`, `show cdc task` reported `paused`:

```text
2026-06-03 12:18:58 Pause cdc_resume_6kw cdc test
OK

2026-06-03 12:19:05 Show cdc_resume_6kw cdc test status
"task-name": "cdc_resume_6kw",
"state": "paused",
"err-msg": ""
```

Immediately after that, `resume cdc task` failed:

```text
2026-06-03 12:19:05 Resume cdc_resume_6kw cdc test
Error: invalid configuration: resume CDC task failed: Error 20101 (HY000): internal error: Task cdc_resume_6kw status can not be change, now it is PauseRequested
```

## Analysis

The issue appears to be a state consistency problem between two CDC/task states:

- `show cdc task` reports state from `mo_catalog.mo_cdc_task.state`.
- `resume cdc task` validates the internal daemon task state from `mo_task.sys_daemon_task.task_status`.

During pause, `mo_cdc_task.state` can be updated to `paused` before the daemon task has actually transitioned from `PauseRequested` to `Paused`.

As a result, users can see:

```text
show cdc task => paused
```

but `resume cdc task` still sees:

```text
sys_daemon_task.task_status => PauseRequested
```

and rejects the resume request.

This makes `show cdc task` misleading for pause/resume readiness.

## Expected Behavior

One of the following should happen:

1. `show cdc task` should not report final state `paused` until the internal daemon task is actually `Paused`.
2. Or `show cdc task` should expose an intermediate state such as `pausing` / `pause_requested`.
[...]

PR #24804, opened 2026-06-03
fix: align CDC pause visible state

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

https://github.com/matrixorigin/matrixone/issues/24797

- Backport CDC pause visible-state alignment to `3.0-dev`.
- Show `pausing` while the daemon task is still `PauseRequested`.
- Update `mo_catalog.mo_cdc_task.state` to `paused` after the CDC executor actually completes pause.

## Why
`show cdc task` could report `paused` before the daemon task reached `Paused`, so an immediate resume could fail while the daemon task was still `PauseRequested`.

## Testing
- `go test -mod=mod ./pkg/frontend -run 'Test_updateCdcTask_pause|TestCdcTask_Pause' -count=1`\n\nLocal test note: used temporary `/tmp` usearch C headers/libs matching the Go binding because local `thirdparties/install` contains an older `usearch.h`.

### DIFF
--- a/pkg/cdc/sql_builder.go
+++ b/pkg/cdc/sql_builder.go
@@ -123,6 +123,12 @@ const (
 		"WHERE " +
 		"1=1 AND account_id = %d AND task_id = '%s'"
 
+	CDCUpdateTaskStateByTaskIdAndStateSqlTemplate = "UPDATE " +
+		"`mo_catalog`.`mo_cdc_task` " +
+		"SET state = '%s' " +
+		"WHERE " +
+		"1=1 AND account_id = %d AND task_id = '%s' AND state = '%s'"
+
 	CDCGetDataKeySqlTemplate = "SELECT " +
 		"encrypted_key " +
 		"FROM mo_catalog.mo_data_key " +
@@ -262,8 +268,9 @@ const (
 	CDCOnDuplicateUpdateWatermarkErrMsgTemplate_Idx = 19
 	CDCClearTaskTableErrorsSQL_Idx                  = 20
 	CDCUpdateTaskStateByTaskIdSQL_Idx               = 21
+	CDCUpdateTaskStateByTaskIdAndStateSQL_Idx       = 22
 
-	CDCSqlTemplateCount = 22
+	CDCSqlTemplateCount = 23
 )
 
 var CDCSQLTemplates = [CDCSqlTemplateCount]struct {
@@ -313,6 +320,9 @@ var CDCSQLTemplates = [CDCSqlTemplateCount]struct {
 	},
 	CDCUpdateTaskStateByTaskIdSQL_Idx: {
 		SQL: CDCUpdateTaskStateByTaskIdSqlTemplate,
+	},
+	CDCUpdateTaskStateByTaskIdAndStateSQL_Idx: {
+		SQL: CDCUpdateTaskStateByTaskIdAndStateSqlTemplate,
 	},
 	CDCInsertWatermarkSqlTemplate_Idx: {
 		SQL: CDCInsertWatermarkSqlTemplate,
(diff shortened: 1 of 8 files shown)


XuPeng-SH (changes requested):
有一个阻塞问题：

`pkg/frontend/cdc_handle.go:376-377` 现在先把 `mo_cdc_task.state` 写成 `pausing`，然后依赖 `pkg/frontend/cdc_exector.go:581-583` 在 `CDCTaskExecutor.Pause()` 结束后再写回 `paused`。但这个最终回写只会发生在本地 active routine 真正执行到 `Pause()` 的路径上。

`pkg/taskservice/daemon_task.go:724-730` 明确说明 pause 还会处理 heartbeat timeout / 非本地 runner 的 lagged task，而且这种场景 **不会** pause active routine；同时 `pkg/taskservice/daemon_task.go:359-367` 也只有在 `t.runner.exists(tk.ID)` 时才会调用 `(*ar).Pause()`。这意味着 lagged/non-local task 走到 pause handler 后，taskservice 可以把 daemon task 推进到 `Paused`，但 `mo_cdc_task.state` 没有任何地方再从 `pausing` 改成 `paused`，外部可见状态会永久卡在 `pausing`。

建议要么对 non-local/lagged pause 保留原先直接写 `paused` 的行为，要么在 taskservice 的 pause 完成路径补一个 catalog state finalize，而不是只放在本地 executor 的 `Pause()` 里。

XuPeng-SH (changes requested):
I’m requesting changes because there is a correctness issue in the new CDC pause state flow.

`pauseTask.Handle()` still runs under a 5s timeout context (`pkg/taskservice/daemon_task.go:314`), but after persisting the daemon task as `Paused` it calls `activeRoutine.Pause()` and only then runs the new completion hook (`pkg/taskservice/daemon_task.go:347-363`). The problem is that `CDCTaskExecutor.Pause()` can legitimately exceed 5s now because it does `ForceFlush` with a 30s timeout (`pkg/frontend/cdc_exector.go:546-549`) and also waits for readers to stop.

That means the hook can run with an already-expired context, so `updateCDCTaskState(..., paused)` fails (`pkg/frontend/cdc_exector.go:818-878`) even though the daemon task status has already been stored as `Paused`. After that there is no retry path, because the scheduler only picks tasks still in `PauseRequested` (`pkg/taskservice/daemon_task.go:720-733`). In practice this can leave `mo_catalog.mo_cdc_task.state` stuck at `pausing` permanently.

Please update the final catalog-state write to use a fresh context, or otherwise guarantee retry after the daemon task has already been transitioned to `Paused`.

XuPeng-SH (changes requested):
I’m requesting changes because there is still a correctness race in the CDC pause visible-state flow.

`pauseTask.Handle()` now persists the daemon task as `Paused` before running the pause-complete hook. In that window, `RESUME CDC` / `RESTART CDC` can already move the catalog row back to `running`, but the hook then does an unconditional `UPDATE mo_catalog.mo_cdc_task SET state='paused' WHERE task_id=...`. A late pause finalize can therefore overwrite a concurrent resume/restart and leave `show cdc task` reporting `paused` while the task is already running again.

Concretely:
- `pkg/taskservice/daemon_task.go`: the daemon task is updated to `Paused` first, then `pauseTaskCompleted(...)` runs.
- `pkg/frontend/cdc_exector.go`: the hook issues an unconditional `SET state = 'paused'` update by task_id.
- `pkg/frontend/cdc_handle.go`: resume/restart still update the catalog row back to `running` without guarding against that late finalize.

Please make the final catalog-state update conditional on the catalog still being in `pausing` (or otherwise re-check daemon-task / executor state before writing `paused`).

There is also a secondary issue in the new `pauseCompletedTasks` tracking: successful entries are remembered in-memory and only cleared when the same task enters `PauseRequested` again, so tasks that are resumed/canceled/dropped without another pause leave stale IDs behind for the lifetime of the runner.

XuPeng-SH (approved):
Reviewed from three angles:
2. Performance and resource impact: changes are acceptable for the task pause path.
3. Compatibility and maintainability: state transitions and coverage look sufficient for merge.

PR #24805, opened 2026-06-03
fix: align CDC pause visible state

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

https://github.com/matrixorigin/matrixone/issues/24797

- Backport CDC pause visible-state alignment to `4.0-dev`.
- Show `pausing` while the daemon task is still `PauseRequested`.
- Update `mo_catalog.mo_cdc_task.state` to `paused` after the CDC executor actually completes pause.

## Why
`show cdc task` could report `paused` before the daemon task reached `Paused`, so an immediate resume could fail while the daemon task was still `PauseRequested`.

## Testing
- `go test -mod=mod ./pkg/frontend -run 'Test_updateCdcTask_pause|TestCdcTask_Pause' -count=1`\n- `go test -mod=mod ./pkg/taskservice -run 'TestUpdateCDCTaskBranchesInSqlMock|Test_UpdateCDCTask' -count=1`\n\nLocal test note: used temporary `/tmp` usearch C headers/libs matching the Go binding because local `thirdparties/install` contains an older `usearch.h`.


aunjgr (approved):
Clean fix for CDC pause state alignment. Two main changes:

**1. Deferred `TransitionPauseComplete` removed** (`cdc_exector.go`). The old code used `defer` to call `TransitionPauseComplete` — meaning it would fire even if the watermark `ForceFlush` failed. The state machine would say "Paused" while watermark data was inconsistent. Now the transition only happens after successful flush.

**2. `pauseTaskCompleted` tracking** (`daemon_task.go`):
- When a task is already `TaskStatus_Paused` but the pause-completed flag isn't set, the old code returned nil (idempotent but incomplete). Now it calls `pauseTaskCompleted` to finish the paused-state transition and persist.
- `clearPauseTaskCompleted` called on resume/restart to reset the flag.
- `isPauseTaskCompleted` check in pause handler: if task is already Paused AND completed, skip (true idempotent).

The tracking uses a mutex-protected map (`pauseCompletedTasks`) — thread-safe.

**Additional**:
- `handleCtx` variable rename avoids shadowing the outer `ctx` in `Handle` methods — correct but cosmetic.
- `Pause()` now checks `state == StateRunning || state == StateStarting` instead of `exec.stateMachine.IsRunning()` — more precise.
- `filterUncompletedPauseTasks` for recovery: during reconciliation, skips tasks that have already completed their pause cycle.

Tests cover pause-before-running, pause-from-running, and resume-after-pause paths. LGTM.

XuPeng-SH (approved):
Reviewed from three angles:
2. Performance and resource impact: no concerning regression found in the pause flow.
3. Compatibility and maintainability: visible-state handling and coverage look sufficient for merge.

PR #24807, opened 2026-06-03
 add UUID functions

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24486

增加uuid函数: is_uuid, uuid_short, uuid_to_bin, bin_to_uuid


heni02 (approved):

XuPeng-SH (changes requested):
I reviewed this from correctness / behavior-compatibility / coverage angles, and there is one blocking issue:

- `uuid_short()` is currently implemented as `xxhash.Sum64(uuid.NewV7())`. That is not MySQL-compatible. MySQL documents `UUID_SHORT()` as a 64-bit value built from `(server_id << 56) + (startup_time << 24) + incrementing_counter`, specifically to preserve uniqueness under documented conditions and to produce increasing values. Hashing a 128-bit UUID down to 64 bits removes both of those properties: collisions become possible by construction, and call order is no longer monotonic.

The official MySQL docs for `UUID_SHORT()` are explicit about this contract, and the current implementation does not match it. I think this needs either:
1. a real `UUID_SHORT()` implementation with server/startup/counter semantics, or
2. a different function name if the intention is to expose a hashed UUID-derived helper rather than MySQL-compatible behavior.

The current tests also only check `uuid_short() is not null` / `> 0`, so they would not catch the semantic mismatch above.

daviszhen:
> I reviewed this from correctness / behavior-compatibility / coverage angles, and there is one blocking issue:
> 
> * `uuid_short()` is currently implemented as `xxhash.Sum64(uuid.NewV7())`. That is not MySQL-compatible. MySQL documents `UUID_SHORT()` as a 64-bit value built from `(server_id << 56) + (startup_time << 24) + incrementing_counter`, specifically to preserve uniqueness under documented conditions and to produce increasing values. Hashing a 128-bit UUID down to 64 bits removes both of those properties: collisions become possible by construction, and call order is no longer monotonic.
> 
> The official MySQL docs for `UUID_SHORT()` are explicit about this contract, and the current implementation does not match it. I think this needs either:
> 
> 1. a real `UUID_SHORT()` implementation with server/startup/counter semantics, or
> 2. a different function name if the intention is to expose a hashed UUID-derived helper rather than MySQL-compatible behavior.
> 
> The current tests also only check `uuid_short() is not null` / `> 0`, so they would not catch the semantic mismatch above.

删除了uuid_short. 一是不在issue需求. 二是 mysql的实现方案 在 mo 多cn中不好做.

XuPeng-SH (changes requested):
I found one blocking compatibility issue.

`UUID_TO_BIN()` / `BIN_TO_UUID()` currently force any non-integer `swap_flag` through an `INT64` cast in `list_builtIn.go` before execution. The implementation then only checks `swapFlag != 0` in `func_builtin.go`. That is not MySQL-compatible: upstream MySQL evaluates the second argument with boolean coercion (`val_bool()`), so nonzero fractional values are true. In this patch, values such as `0.4` are narrowed to integer zero first, which means `uuid_to_bin(u, 0.4)` / `bin_to_uuid(b, 0.4)` will not swap even though MySQL does.

The current unit tests and BVT only cover integer / NULL swap flags, so this regression is not caught by coverage.

aunjgr (approved):
MySQL-compatible UUID utility functions: `IS_UUID()`, `UUID_TO_BIN()`, `BIN_TO_UUID()`. Clean implementation — swap-flag handling via `uuidToBin` matches MySQL 8.0 semantics (swaps time-low/time-mid/time-high-and-version groups). BVT covers documented UUID string forms, NULL handling, invalid UUID rejection, both swap modes. LGTM.

daviszhen:
> I found one blocking compatibility issue.
> 
> `UUID_TO_BIN()` / `BIN_TO_UUID()` currently force any non-integer `swap_flag` through an `INT64` cast in `list_builtIn.go` before execution. The implementation then only checks `swapFlag != 0` in `func_builtin.go`. That is not MySQL-compatible: upstream MySQL evaluates the second argument with boolean coercion (`val_bool()`), so nonzero fractional values are true. In this patch, values such as `0.4` are narrowed to integer zero first, which means `uuid_to_bin(u, 0.4)` / `bin_to_uuid(b, 0.4)` will not swap even though MySQL does.
> 
> The current unit tests and BVT only cover integer / NULL swap flags, so this regression is not caught by coverage.

• 本次针对 PR 24807 的 UUID comment 做了这些修改：

  1. 修复 swap_flag 类型处理
      - list_builtIn.go 不再把非整数 swap_flag 强制 cast 到 INT64。
      - 现在保留 bool / int / uint / float / decimal / MySQL string 类型。
      - 其他类型 fallback cast 到 DOUBLE，避免执行层不能读取。

  2. 修复执行语义
      - func_builtin.go 里 uuid_to_bin / bin_to_uuid 从 int64 swapFlag != 0 改成 MySQL val_bool() 风格判断。
      - 新增 makeBoolParamGetter()：
          - 整数：v != 0
          - float/double：v != 0
          - decimal：v != 0
          - string：转 float 后 != 0
          - NULL：返回 NULL

  3. 补 UT
      - TestBuiltInUUIDSwapFlagBoolCoercion
      - 覆盖 0.4、-0.4、1.2 都会触发 swap，0 不 swap，NULL 结果为 NULL。
      - Test_GetFunctionByName 增加断言：uuid_to_bin(varchar, float64) / bin_to_uuid(varbinary, float64) 不再需要 cast 到 int64。

  4. 补 BVT case
      - 在 func_string_uuid2.test 增加 0.4、-0.4、0.0 的 uuid_to_bin / bin_to_uuid case。
      - 未修改 .result 文件。

XuPeng-SH (changes requested):
I found one substantive MySQL-compatibility issue in the new `swap_flag` support for `UUID_TO_BIN()` / `BIN_TO_UUID()`.

This PR explicitly broadens the accepted `swap_flag` types to include strings (`uuidSwapFlagTypeSupported()` accepts MySQL string types), but the string branch in `makeBoolParamGetter()` uses `strconv.ParseFloat` on the entire string. That is stricter than MySQL’s coercion rules: values like `1abc` / `2xyz` are truthy in MySQL because boolean evaluation consumes the leading numeric prefix, but this implementation treats any parse failure as false and returns the unswapped result.\n\nSo right now the function signature says string swap flags are supported, but some supported string inputs do not behave like MySQL. I also do not see coverage for this path in the added SQL or unit tests — current tests cover clean numeric values like `0.4`, `-0.4`, `0.0`, but not prefix-numeric strings. I think this should either use MySQL-compatible string-to-number coercion for the boolean evaluation, or reject non-numeric strings explicitly instead of silently treating them as `0`.

daviszhen:
> I found one substantive MySQL-compatibility issue in the new `swap_flag` support for `UUID_TO_BIN()` / `BIN_TO_UUID()`.
> 
> This PR explicitly broadens the accepted `swap_flag` types to include strings (`uuidSwapFlagTypeSupported()` accepts MySQL string types), but the string branch in `makeBoolParamGetter()` uses `strconv.ParseFloat` on the entire string. That is stricter than MySQL’s coercion rules: values like `1abc` / `2xyz` are truthy in MySQL because boolean evaluation consumes the leading numeric prefix, but this implementation treats any parse failure as false and returns the unswapped result.\n\nSo right now the function signature says string swap flags are supported, but some supported string inputs do not behave like MySQL. I also do not see coverage for this path in the added SQL or unit tests — current tests cover clean numeric values like `0.4`, `-0.4`, `0.0`, but not prefix-numeric strings. I think this should either use MySQL-compatible string-to-number coercion for the boolean evaluation, or reject non-numeric strings explicitly instead of silently treating them as `0`.

mo不支持前缀数字字符串 转成 数字.  现在改为报错.

XuPeng-SH (approved):
I re-checked the latest update and this looks good to me now.

The key point I cared about before was the `swap_flag` string path. This version makes the behavior explicit and internally consistent: if a string swap flag cannot be parsed as a numeric value, it now errors instead of being silently treated as false. That is stricter than MySQLs loose numeric-prefix coercion, but as an intentional MO behavior it is coherent, well-covered by the new unit and SQL tests, and much better than the previous silent fallback.

PR #24808, opened 2026-06-03
feat: parallelize parquet load by row group

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24254 

Plan Parquet LOAD around file and row-group fanout, carry shard metadata through ExternalScan, and keep S3 prefetch behavior bounded for sharded readers.

Harden unsupported option handling, add Parquet profile stats, and cover compile/runtime/BVT regressions for schema, conversion, rollback, and parallel-load paths.


ouyuanning (approved):

heni02 (approved):

XuPeng-SH (changes requested):
I found two substantive issues in the current patch.

1. **Empty parquet files can disable row-group fanout for the remaining non-empty files.**
   The planner currently switches to row-group fanout only when `footerStats.RowGroups > len(fileList)`. That means empty parquet files still count toward `len(fileList)` even though they contribute zero row groups. So a case like `{empty.parquet, data.parquet(with 2 row groups)}` falls back to whole-file fanout instead of splitting the non-empty file by row group, which defeats the new planning logic for a very realistic edge case.

2. **Small `LOAD ... PARALLEL` files now change empty numeric fields from the normal serial behavior to `0`.**
   `ParallelLoadRequested` is latched before the planner disables parallel load for small files. Later, the external path treats `ParallelLoadRequested` the same as actual parallel execution and turns on `LoadEmptyNumericAsZero`. So the same small CSV row with an empty numeric field can now be silently loaded as `0` just because the original statement asked for `PARALLEL`, even though the actual execution path is serial. That makes the result depend on the file-size threshold, which is a real semantic regression.

I think both of these should be fixed before approval.

aunjgr (approved):

iamlinjunhong:
> I found two substantive issues in the current patch.
> 
> 1. **Empty parquet files can disable row-group fanout for the remaining non-empty files.**
>    The planner currently switches to row-group fanout only when `footerStats.RowGroups > len(fileList)`. That means empty parquet files still count toward `len(fileList)` even though they contribute zero row groups. So a case like `{empty.parquet, data.parquet(with 2 row groups)}` falls back to whole-file fanout instead of splitting the non-empty file by row group, which defeats the new planning logic for a very realistic edge case.
> 2. **Small `LOAD ... PARALLEL` files now change empty numeric fields from the normal serial behavior to `0`.**
>    `ParallelLoadRequested` is latched before the planner disables parallel load for small files. Later, the external path treats `ParallelLoadRequested` the same as actual parallel execution and turns on `LoadEmptyNumericAsZero`. So the same small CSV row with an empty numeric field can now be silently loaded as `0` just because the original statement asked for `PARALLEL`, even though the actual execution path is serial. That makes the result depend on the file-size threshold, which is a real semantic regression.
> 
> I think both of these should be fixed before approval.

fixed

XuPeng-SH (changes requested):
I re-checked the latest head. The earlier blockers around row-group planning and the small-file `ParallelLoadRequested` regression look addressed, but there is still one remaining high-severity issue in the empty-file path.

In the new row-group fanout path, empty parquet files are effectively dropped from execution:
- `readLoadParquetRowGroupMetadata()` only emits metadata rows for actual row groups
- `splitParquetRowGroupShards()` only carries files into a shard when they own at least one row group
- `compileExternScanParquetRowGroupFanout()` then executes only those sharded files

That means an empty parquet file matched by the load pattern never reaches `newParquetHandler()` once the plan has switched to row-group fanout.

But the empty-file compatibility check still lives in `pkg/sql/colexec/external/parquet.go:63-81`: that is where an empty parquet with the wrong column count is supposed to fail. So with the current fanout logic, a wildcard load that matches `{empty_bad_schema.parquet, data.parquet(with row groups)}` can silently skip the empty file and lose the existing schema/column-count validation instead of rejecting the load.

The current tests actually show this behavior clearly:
- `pkg/sql/compile/scope_test.go:885-950` asserts the fanout scopes contain only the non-empty file
- `pkg/sql/colexec/external/parquet_test.go:3055-3093` shows the mismatch is only caught when the empty file is actually opened by `newParquetHandler()`

**Concrete suggestions**

1. Validate empty parquet files during footer enumeration before discarding them from row-group metadata.
2. Or, carry matched empty files into at least one execution shard so `newParquetHandler()` still opens them and applies the existing empty-file checks.
3. Add one regression test for the mixed case explicitly: load pattern matches both
   - an empty parquet with schema/column-count mismatch, and
   - a non-empty multi-row-group parquet,
   and the whole load must still fail.

Non-blocking follow-up only: I still think an end-to-end case that actually forces the row-group-parallel execution path on a threshold-crossing parquet file would be a useful extra proof after this semantic hole is fixed.

iamlinjunhong:
> I re-checked the latest head. The earlier blockers around row-group planning and the small-file `ParallelLoadRequested` regression look addressed, but there is still one remaining high-severity issue in the empty-file path.
> 
> In the new row-group fanout path, empty parquet files are effectively dropped from execution:
> 
> * `readLoadParquetRowGroupMetadata()` only emits metadata rows for actual row groups
> * `splitParquetRowGroupShards()` only carries files into a shard when they own at least one row group
> * `compileExternScanParquetRowGroupFanout()` then executes only those sharded files
> 
> That means an empty parquet file matched by the load pattern never reaches `newParquetHandler()` once the plan has switched to row-group fanout.
> 
> But the empty-file compatibility check still lives in `pkg/sql/colexec/external/parquet.go:63-81`: that is where an empty parquet with the wrong column count is supposed to fail. So with the current fanout logic, a wildcard load that matches `{empty_bad_schema.parquet, data.parquet(with row groups)}` can silently skip the empty file and lose the existing schema/column-count validation instead of rejecting the load.
> 
> The current tests actually show this behavior clearly:
> 
> * `pkg/sql/compile/scope_test.go:885-950` asserts the fanout scopes contain only the non-empty file
> * `pkg/sql/colexec/external/parquet_test.go:3055-3093` shows the mismatch is only caught when the empty file is actually opened by `newParquetHandler()`
> 
> **Concrete suggestions**
> 
> 1. Validate empty parquet files during footer enumeration before discarding them from row-group metadata.
> 2. Or, carry matched empty files into at least one execution shard so `newParquetHandler()` still opens them and applies the existing empty-file checks.
> 3. Add one regression test for the mixed case explicitly: load pattern matches both
>    
>    * an empty parquet with schema/column-count mismatch, and
>    * a non-empty multi-row-group parquet,
>      and the whole load must still fail.
> 
> Non-blocking follow-up only: I still think an end-to-end case that actually forces the row-group-parallel execution path on a threshold-crossing parquet file would be a useful extra proof after this semantic hole is fixed.

fixed

XuPeng-SH (approved):
Re-checked the latest head. The previous blockers around row-group planning and the small-file `ParallelLoadRequested` regression are fixed, and the remaining empty-parquet semantic hole is now closed as well: empty parquet files are validated during footer enumeration before being dropped from row-group fanout, with a mixed empty+non-empty regression test to prove it. I do not see a remaining correctness issue that should keep this blocked.

PR #24809, opened 2026-06-03
fix(plan): support enum and prefix text indexes

### LINKED ISSUE #24729: [Bug]: ENUM column cannot be used in any secondary/unique/composite index (v3.0.11 & 4.0-dev)
### Summary
Creating **any** non-primary index that includes an `ENUM` column fails — secondary, unique, and composite (ENUM at any position) are all rejected. ENUM columns themselves, enum filters, and enum validation work fine; only indexing them fails.

### Steps to Reproduce
```sql
CREATE TABLE t1 (id VARCHAR(191) PRIMARY KEY, role ENUM('a','b','c'), INDEX idx_role(role));
-- ERROR 20105 (HY000): not supported: ENUM column 'role' cannot be in secondary index

CREATE TABLE t2 (id VARCHAR(191) PRIMARY KEY, role ENUM('a','b','c'), UNIQUE INDEX uq_role(role));
-- ERROR 20105 (HY000): not supported: ENUM column 'role' cannot be in unique index

CREATE TABLE t3 (id VARCHAR(191) PRIMARY KEY, name VARCHAR(191), role ENUM('a','b','c'), INDEX idx(name, role));
-- ERROR 20105 (HY000): not supported: ENUM column 'role' cannot be in secondary index
```

### Prisma impact
Any Prisma model with `@@index`/`@@unique` on an enum field fails to migrate. We had to remove 4 such indexes (e.g. `users(workspace_id, role)`, `branches(status)`) from the schema + generated SQL.

### Expected Behavior
MySQL allows ENUM columns in secondary and unique indexes.

---
**Test context** — Found while testing **Prisma ORM (v5.22)** compatibility against MatrixOne. Reproduced on **both**:
- Prod cloud — `8.0.30-MatrixOne-v3.0.11`
- 4.0 dev line — `nightly-0bfda604` (`select version()` → `8.0.30-MatrixOne-v`)

(Results on the older v3.0.5 are intentionally omitted.)

### LINKED ISSUE #24740: [Bug]: TEXT column cannot be in any index, including prefix index (v3.0.11 & 4.0-dev)
### Summary
A `TEXT` column cannot be placed in an index — **neither a prefix index `(col(N))` nor a plain index**. Both fail with `not supported: TEXT column 'x' cannot be in index`. This is the same "column-type cannot be in index" family as the ENUM restriction in #24729.

### Steps to Reproduce
```sql
-- prefix index:
CREATE TABLE tx (id INT PRIMARY KEY, t TEXT, INDEX tx_t (t(100)));
-- ERROR 20105 (HY000): not supported: TEXT column 't' cannot be in index

-- plain index:
CREATE TABLE ty (id INT PRIMARY KEY, t TEXT, INDEX ty_t (t));
-- ERROR 20105 (HY000): not supported: TEXT column 't' cannot be in index
```

### Prisma impact
Low — Prisma uses `VARCHAR(191)` for indexed `String` fields by default; only a model that places an `@@index`/`@@unique` on an `@db.Text`/`@db.LongText` field is affected. (Prefix indexes on TEXT/BLOB are the standard MySQL way to index large columns.)

### Expected Behavior
MySQL allows a prefix index on TEXT/BLOB columns, e.g. `INDEX (col(100))`.

Related: #24729 (ENUM column cannot be in index).

---
*Found during Prisma ORM (v5.22) / MySQL-compat testing; reproduced on prod `8.0.30-MatrixOne-v3.0.11` and `nightly-0bfda604`.*

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #24740
Fixes #24729

- Allows ENUM columns in secondary, unique, and composite indexes while keeping ENUM primary keys rejected.
- Allows TEXT/BLOB prefix indexes and keeps plain TEXT/BLOB indexes rejected.
- Persists index prefix lengths in index metadata and applies prefix expressions during index backfill, insert, unique dedup, and lock-key generation.
- Adds focused unit coverage and BVT cases for ENUM indexes and TEXT/BLOB prefix indexes.

Validation:

- `go test ./pkg/sql/plan -run 'TestBuildIndexAllowsEnumAndTextBlobPrefix|TestBuildIndexRejectsTextBlobPlainIndex' -count=1`
- `go test ./pkg/sql/compile -run 'TestGenInsertIndexTableSql_UsesPrefixExpression|TestBuildCreateUniqueIndexDuplicateCheckSQL' -count=1`
- `go test ./pkg/sql/colexec/preinsertunique ./pkg/sql/colexec/preinsertsecondaryindex -count=1`
- `./run.sh -p /Users/LoveYY/Develop/matrixorigin/matrixone/.worktrees/issue-24740/test/distributed/cases/ddl/create_index.sql -g -n`
- `./run.sh -p /Users/LoveYY/Develop/matrixorigin/matrixone/.worktrees/issue-24740/test/distributed/cases/dtype/enum.sql -g -n`
- `git diff --check`
- `make`
- `make static-check`


heni02 (approved):

XuPeng-SH (changes requested):
I found a substantive issue in the current prefix-index implementation.

**Prefix length is persisted in metadata, but `SHOW CREATE TABLE` still drops it, so the feature does not round-trip.**

- Prefix lengths are stored in `indexDef.IndexAlgoParams` via `prefix_lengths` (`pkg/catalog/secondary_index_utils.go:187-203`, `pkg/sql/plan/build_ddl.go:2076-2079,2456-2459`).
- But when `SHOW CREATE TABLE` renders index definitions, it only prints raw part names and then appends `catalog.IndexParamsToStringList(...)` (`pkg/sql/plan/build_show_util.go:278-303`). That renderer never reconstructs key-part suffixes like `(100)`, and `IndexParamsToStringList` does not output `prefix_lengths` either (`pkg/catalog/secondary_index_utils.go:104-160`).

So a table created as `INDEX x3(a(100))` will still be shown as if it were just `INDEX x3(a)`. That is not just cosmetic: `SHOW CREATE TABLE` no longer reproduces the original schema, and MatrixOne's snapshot/PITR/restore flows fetch DDL through `show create table` (`pkg/frontend/snapshot_restore_with_ts.go:252-268`, `pkg/frontend/pitr.go:1583-1592`, `pkg/frontend/snapshot.go:2072-2080`). In other words, a restored table can silently lose its prefix-index semantics or even fail when replaying the downgraded DDL on TEXT/BLOB keys.

I think this needs to be fixed before approval: `SHOW CREATE TABLE` / restoreDDL must reconstruct key-part lengths from `prefix_lengths` and emit `col(length)` again.

XuPeng-SH (changes requested):
Requesting changes for a prefix-index maintenance bug on the delete/update path.

The backfill / insert / unique-lock paths were updated to materialize prefix index keys with `substring(..., 1, n)`, but `appendDeleteIndexTablePlan` in `build_dml_util.go` still builds the lookup key from the full source column value (or `serial(...)` of full values).

That means for any indexed value longer than the prefix length, DELETE / UPDATE will fail to match the old hidden index row. On secondary indexes that leaves stale index rows behind; on unique prefix indexes it can later surface as false duplicate conflicts.

I think the delete-maintenance path needs to apply the same prefix expression/cast as the insert/backfill path before joining to the index table, and it would be good to add an end-to-end UPDATE/DELETE case with values longer than the prefix length.

aunjgr (commented):
XuPeng-SH's two findings are real bugs: 1) SHOW CREATE TABLE doesn't round-trip prefix lengths, breaks snapshot/PITR restore. 2) DELETE/UPDATE path uses full column value instead of prefix expression, leaving stale index rows. Both need fixes before merge.

XuPeng-SH (approved):
✅ **APPROVED after comprehensive multi-angle review**

**Analysis Summary**: 2576 lines across 25 files • 21-item systematic checklist • Zero blocking issues

**Key Strengths**:
1. ✅ **Excellent refactoring** - centralized validation in checkIndexColumnSupportability()
2. ✅ **Comprehensive test coverage** - 87 planner + 64 BVT lines covering happy/unhappy paths
3. ✅ **First-principles correctness** - substring/cast applied at all DML entry points (INSERT/UPDATE/DELETE)
4. ✅ **Robust error handling** - validates prefix lengths at parse/serialization/execution
5. ✅ **Backward compatible** - existing indexes continue working, no migration required
6. ✅ **Critical bug fix** - preinsertunique.go now uses PreInsertCtx.UkType (prevents type mismatches)

**Unhappy Path Validation**:
- ✅ Zero/negative prefix lengths → rejected at parse time
- ✅ Prefix > data length → substring() handles gracefully
- ✅ NULL values → filtered in index maintenance
- ✅ Mixed prefix/non-prefix columns → each checked independently
- ✅ Upgrade scenario → (nil, nil) return for old metadata
- ✅ GEOMETRY32 → RTREE validation preserved

**First-Principles Check**:
- Core design: PREFIX INDEX stores substring(col, 1, len) instead of full TEXT/BLOB
- Type safety: TEXT→VARCHAR(65535), BLOB→VARBINARY(65535) with proper casts
- Metadata: JSON extensible format in IndexAlgoParams
- DML completeness: 12 function usages across INSERT/DELETE/UPDATE paths

See full review document: /tmp/pr_review_24809.md

This PR is **production-ready** and represents a **significant feature addition** with **excellent quality**.

aunjgr (approved):

Issue #24743, opened 2026-06-01
[Tech Request]: GC/reclamation for pinned datalink content-addressed store (CAS)


ck89119:
### Is there an existing issue for the same tech request?

- [x] I have checked the existing issues.

### Does this tech request not affect user experience?

- [x] This tech request doesn't affect user experience.

### What would you like to be added ?

Add garbage collection / reclamation for the content-addressed store (CAS) that backs pinned datalinks.

Follow-up to #24555 / PR #24641. That PR introduced `datalink_pin()`, which freezes the bytes a datalink references into an immutable, per-account CAS keyed by content hash:

```
datalink_cas/<accountID>/<h2>/<hash>
```

The store is intentionally **write-once and immutable** for byte-reproducibility, and that PR deliberately scoped *out* reclamation. As a result, a CAS blob is currently **retained indefinitely** — nothing reclaims it after the datalink values that referenced it are gone (rows deleted, table dropped, snapshot expired, or the whole account dropped). Over time this leaks storage.

This issue tracks designing and implementing reclamation. Two candidate mechanisms (not mutually exclusive):

1. **Account-scoped cleanup on `DROP ACCOUNT`** — drop the entire `datalink_cas/<accountID>/` prefix so a removed tenant does not leak its pinned blobs. Coarse-grained, simple, and safe (the namespace is already per-account).
2. **Reference-aware sweep** — mark blobs still reachable from live `datalink` columns and live snapshots, then delete the unreferenced rest. Required for the general case because a single content hash can be shared by many rows / snapshots (per-account dedup), so a blob must not be collected while *any* reference survives.

### Why is this needed ?

Without reclamation, pinned content accumulates in the SHARED file service forever and cannot be reclaimed even when no datalink references it, which is unsustainable for long-lived deployments and for tenants that pin large volumes of content.

### Additional information

- Feature / parent: #24555 (datalink content-versioning via content-hash pinning)
- Introducing PR: #24641
- CAS implementation: `pkg/datalink/cas.go` (`CASKey` / `CASPut` / `CASExists`; note there is currently **no** `CASDelete`/GC entry point)
- Design questions to resolve: reference discovery across tables + snapshots, concurrency vs. concurrent `datalink_pin()`, and whether reclamation runs as an online sweep, on DDL events (`DROP ACCOUNT` / `DROP TABLE`), or both.

PR #24641, opened 2026-05-27
feat(datalink): add datalink_pin for content-addressed version pinning

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24555

## What this PR does / why we need it:

Today a `datalink` snapshot only versions the **reference string**, not the
**bytes of the referenced file**. After `CREATE SNAPSHOT`, if the external
object is overwritten out of band, time-travel reads return the new bytes, so
the historical state is not byte-reproducible.

This PR adds an **opt-in** way to make datalink content reproducible, without
changing the default (un-pinned) behavior.

### What it does

- **CAS (`pkg/datalink/cas.go`)** — a write-once, content-addressed store keyed
  by `sha256`, backed by the **SHARED** file service
  (`datalink_cas/<h2>/<hash>`). Identical bytes dedup naturally; objects are
  never overwritten.
- **`datalink_pin(datalink) -> datalink`** builtin (also accepts varchar/char/
  text) — reads the bytes a datalink currently references, stores them in the
  CAS, and returns a datalink carrying `?contenthash=<sha256>`. Already-pinned
  inputs are returned unchanged (idempotent); an unreadable source errors out
  (never silently un-pinned).
- **Read path (`ParseDatalink`/`NewReadCloser`)** — a `?contenthash=` value
  resolves to the immutable CAS object and is read **directly** from the SHARED
  `FileService` (not via `GetForETL`, since SHARED may be a `LocalFS` that does
  not implement `ETLFileService` in standalone). A missing CAS object errors out
  rather than falling back to the live file. Because `load_file`, wasm, fulltext
  etc. all go through `NewReadCloser`, the read path is fixed in one place.

Pin info rides along with the row value, so the existing timestamp-based
snapshot mechanism is untouched.

Example:

```sql
INSERT INTO docs VALUES (1, datalink_pin('stage://st/doc.txt'));
-- doc.txt later overwritten out of band
SELECT load_file(ref) {snapshot='s'};  -- still returns the original bytes
```

### Scope / follow-ups (not in this PR)

- Column attribute `datalink VERSIONED` + INSERT/UPDATE plan injection to
  auto-pin (this PR ships the explicit-function entry point only).
- CAS blob GC / lifecycle (blobs currently leak; tracked separately).

### Tests

[...]

PR #24835, opened 2026-06-04
feat(datalink): GC/reclamation for pinned datalink CAS

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24743

> **Depends on #24641** (introduces the datalink CAS). This PR is stacked on that branch, so the diff currently includes #24641's commits; it will be rebased onto main once #24641 merges, leaving only this issue's changes.

Adds garbage collection / reclamation for the per-account, content-addressed store (CAS) `datalink_cas/<accountID>/<hash>` that backs pinned datalinks (`datalink_pin`, #24641). Without it, pinned blobs are retained forever once their referencing datalink values are gone (rows deleted, table dropped, snapshot expired, account dropped), leaking storage in the SHARED file service.

Two mechanisms:

1. **DROP ACCOUNT cleanup** — on `DROP ACCOUNT`, the account's entire `datalink_cas/<accountID>/` prefix is reclaimed (best-effort; a failure never aborts the drop, the sweep is the backstop). `pkg/frontend/authenticate.go`, `pkg/datalink/cas.go`.

2. **Reference-aware periodic sweep** (`pkg/datalink/casgc/`) — an hourly cron task that, per account, collects the content hashes referenced by live datalink columns **and live snapshots**, then reclaims unreferenced CAS blobs. Because the file service exposes no object mtime, the grace period is a **two-pass mark-and-delete with in-process state**: a blob is deleted only after staying an orphan across passes for at least GraceWindow (default 24h, configurable; sweep interval 1h). The sweep is candidate-driven (it skips snapshot scans when nothing is collectable), and a singleton Sweeper preserves the grace-window state across cron firings.

**Snapshot safety**: CAS blobs live in the SHARED file service, not in MVCC data, so snapshots/PITR do not pin them. The sweep therefore treats hashes referenced by any live snapshot as live, preventing dangling datalink references after a restore.

### Tests
- Unit: CAS primitives, reference discovery, two-pass sweep (grace window / candidate-driven / re-reference reset / cross-account / SweepAll fault-tolerance), and a mock-based SQL reference layer. Coverage: `pkg/datalink` 76.6%, `pkg/datalink/casgc` 84.0%.
- BVT: `func_datalink_cas_gc` — pin content into a table, DROP ACCOUNT with pinned blobs present, recreate a same-named account with a clean namespace (11/11 pass).
- `make` and `make static-check` clean.


fengttt (approved):

aunjgr (commented):
There are no code changes since the earlier functional approval, but this PR remains draft and conflicting. Please rebase it on current main, resolve the conflicts, complete draft status, and request review again on the resulting head.

aunjgr (commented):
Retracted: this PR is a draft and is excluded from this review pass. No review was performed.

XuPeng-SH (changes requested):
Reviewed ec5ffea48d6d1b612a9fd485fd402ade3a2764ca. REQUEST_CHANGES: destructive CAS collection is not yet reference-safe.

### [P1] Discover schema at each retained snapshot, not only in the live catalog

SweepAccount calls refs.datalinkColumns once for the live schema, then passes that same cols list to every historical scan. Deterministic counterexample: pin a value in table t, create a snapshot, drop t (or its last datalink column), and run two sweeps separated by GraceWindow. The live cols list is empty; every snapshot scan iterates zero columns, so the blob is deleted even though the snapshot can restore its reference. No concurrent operation is required. Discover tables/columns at each snapshot's own timestamp/scope and account, including objects absent or renamed in the current catalog; protect PITR/retained restore owners or explicitly disable collection wherever their reachability cannot be proven. Add a snapshot -> drop table/column -> two sweeps -> restore/read-bytes test.

### [P1] A two-pass clock does not serialize reference publication with deletion

After the final current/snapshot reference reads and before CASDelete, another transaction may publish a reference to an old candidate hash. CASPut deduplicates FileAlreadyExists and has no shared pin/epoch/GC fence. The sweeper's mutex protects only its own pending map, not writers, snapshots or other CNs. The mature candidate can therefore be deleted underneath the newly committed row. A 24-hour grace window does not close that final read/delete interval.

Define a shared publication/reclamation protocol: pending transaction pins/reference ownership or a generation/epoch fence that makes a writer retry if GC retires its generation. Snapshot/PITR publication must participate in the same reachability boundary. Test with a barrier after the last mark read, commit a re-reference from another session/CN, then release deletion and verify the content remains readable. Keep automatic deletion disabled until this invariant is implemented; do not fix it with a longer sleep/window.

### [P2] The promised dropped-account sweep backstop never visits dropped accounts

DROP ACCOUNT cleanup is best-effort, but SweepAll enumerates only SELECT account_id FROM mo_catalog.mo_account. Once that account is removed, a failed prefix deletion is never retried by this sweeper. Retain a durable cleanup tombstone/queue or conservatively enumerate orphan CAS account namespaces against authoritative lifecycle state. Test first-delete failure followed by successful background retry.

Review scope/evidence: full change map and the new sweep, SQL reference-discovery and DROP ACCOUNT handoff paths inspected, plus CASPut's publication behavior. These are source-level counterexamples, not fresh destructive storage tests. Existing mock coverage and drop/recreate BVT do not exercise the above terminals. Preserve the dependency on #24641 and its separate outstanding review; this is not approval of that inherited implementation. Also stream/page reference reads and avoid holding the sweeper-wide mutex over remote deletes when completing the design.

PR #24839, opened 2026-06-04
compile: fix remote dispatch notify for local fallback

LeftHandCold:
## Summary

This PR fixes a chain of remote pipeline coordination failures that can hang multi-CN shuffle workloads on `3.0-dev`.

The fix covers three protocol gaps exposed by the same workload:

- dispatch prepare-done notifications must be able to reach a dispatch that executes on the coordinator;
- remote dispatch must register and later release remote receivers even if it produces no data batches;
- remote pipeline fragments for the same statement must keep sharing the same MultiCN message board until all active fragments are done.

## Problem

Large multi-CN `INSERT ... SELECT ... JOIN` workloads can hang or time out in the remote shuffle/dispatch path.

Observed failure modes included:

- `remote run pipeline in local` when the notify stream targets the local CN;
- `get dispatch process by uuid timeout` when a receiver notifies the originally planned source CN instead of the coordinator;
- `send notify msg to dispatch operator timeout` when a receiver notifies before the dispatch operator starts waiting;
- `waiting messsage timeout` for join map/runtime filter tags, with message board debug output such as `messageBoard has been reseted!` and empty boards.

## Root Cause

Remote receiver metadata is generated from planned source/target CN addresses during scope construction. Later, receivers send `PrepareDoneNotifyMessage` to `RemoteReceivRegInfos.FromAddr`, and the target CN uses the receiver uuid to find the registered dispatch process.

The workload exposes several related edge cases:

1. A planned remote source scope can fall back to local execution when it contains a dispatch or connector that sends to another local pipeline tree. The dispatch process is then registered on the coordinator, but receiver metadata could still point to the originally planned remote source CN.
2. A remote scope whose `RootOp` is a dispatch operator is a boundary case: the remote payload runs on the remote CN, but the root dispatch itself executes on the coordinator side. Receivers for that root dispatch must notify the coordinator.
3. `dispatch.prepareRemote` registered the dispatch process with an unbuffered `remoteInfo` channel. A receiver could send prepare-done before the dispatch operator produced its first output batch and entered `waitRemoteRegsReady`, causing the server-side notify send to time out.
4. If a remote dispatch source produced no data batches, `Dispatch.Call` returned `ExecStop` before calling `waitRemoteRegsReady`. Receiver notifications could already be queued in `remoteInfo`, but they were never moved into `remoteReceivers`. `Reset` therefore did not signal those receivers, so the remote server did not send `EndMessage` back to receiver streams.
5. Each remote `PipelineMessage` creates and clears its own `Compile`, while runtime filters and join maps are shared by statement id through `MessageCenter.StmtIDToBoard`. The old MultiCN `MessageBoard.Reset()` immediately removed the shared board entry. If an early remote fragment finished before later fragments for the same statement reached `SetMultiCN`, the later fragments could attach to a different or already reset board and miss queued join map/runtime filter messages.

## Changes

- Allow `pipelineClient.NewStream` to create a notify stream to the local CN address.
- Run `rewriteRemoteRunLocalFallbacks` at the beginning of `Compile.runOnce`, before root scopes are started concurrently.
- Rewrite receiver metadata for fallback-local source dispatches:
  - convert coordinator-local receivers from `RemoteRegs` to `LocalRegs` only when routing semantics allow it;
  - move the matching shuffle index from `ShuffleRegIdxRemote` to `ShuffleRegIdxLocal`;
  - remove the corresponding `RemoteReceivRegInfos` entry when the receiver is fully local;
  - downgrade `SendToAllFunc` / `SendToAnyFunc` to local variants when no remote receivers remain;
  - otherwise keep the receiver remote and rewrite `FromAddr` to the coordinator address.
- Preserve hybrid shuffle invariants:
  - for `ShuffleToAllFunc`, only `ShuffleToRegIndex` may move a receiver from `RemoteRegs` to `LocalRegs`;
  - modulo-based shuffle modes keep remote receiver entries so `len(LocalRegs)` remains stable.
- Handle root dispatch remote-run boundaries by rewriting only the dispatch operator that executes on the coordinator, not dispatch operators that stay inside the remote payload.
- Buffer `dispatch.ctr.remoteInfo` by `remoteRegsCnt`, allowing prepare-done notifications to arrive before the first data batch.
- In the remote dispatch EOF path, wait for remote receiver notifications before returning `ExecStop`, so `Reset` can notify remote receivers even when no data was produced.
- Retain MultiCN `MessageBoard` entries by statement id while remote fragments are active:
  - add a reference count for active remote compiles using the shared board;
  - after the last active reference is released, keep the board and queued messages for a bounded retention window;
  - clear stale waiters only after the board has no active references;
  - clean up retained boards and queued messages after the retention window;
  - clear stale `reset` debug state when a board is promoted or reused as MultiCN.

## Testing

Passed:

```bash
GOCACHE=/private/tmp/mo-gocache-24839-message-retain go test ./pkg/vm/message -count=1
GOCACHE=/private/tmp/mo-gocache-24839-dispatch-zero go test ./pkg/sql/colexec/dispatch -count=1
GOCACHE=/private/tmp/mo-gocache-24839-compile-rewrite go test ./pkg/sql/compile -run 'Test_checkPipelineStandaloneExecutableAtRemote|Test_rewriteRemoteRunLocalFallbacks' -count=1
git diff --check
```

Coverage added or updated:

- local notify stream self-connection;
- fallback-local dispatch metadata rewrite;
- remote root dispatch notify address rewrite;
- hybrid shuffle receiver conversion guard;
- buffered remote receiver notification delivery;
- remote dispatch shutdown when no data batches are produced;
- MultiCN message board reuse, active-reference accounting, late receiver reuse, stale reset flag clearing, and bounded cleanup.

## Manual Verification

The hang was reproduced on a 2-CN docker cluster with:

```sql
insert into c
select a.id, a.k, a.pad, b.pad
from a join b on a.k = b.k;
```

After the dispatch fixes, the remaining hang no longer showed the old uuid lookup failure. The latest logs showed `waiting messsage timeout` for join map/runtime filter tags, with empty reset MultiCN boards on the remote CN. Goroutine dumps showed probe workers waiting in `ReceiveJoinMap` / runtime filter handling while remote pipeline fragments had already reset their shared message board.

## Risk Assessment

- Local-only dispatch paths are unchanged.
- Remote dispatches that send at least one data batch keep the existing wait path; this PR adds the same wait barrier to the no-data EOF path.
- Hybrid shuffle routing semantics are preserved by avoiding receiver conversion for modulo-based shuffle modes.
- Single-CN message board reset semantics are unchanged.
- MultiCN board retention is scoped by statement id, reference-counted by active remote fragments, and bounded by a cleanup timer.

### DIFF
--- a/pkg/cnservice/cnclient/client.go
+++ b/pkg/cnservice/cnclient/client.go
@@ -15,10 +15,7 @@
 package cnclient
 
 import (
-	"fmt"
-
 	"github.com/fagongzi/goetty/v2"
-	"github.com/matrixorigin/matrixone/pkg/common/moerr"
 	"github.com/matrixorigin/matrixone/pkg/common/morpc"
 	"github.com/matrixorigin/matrixone/pkg/common/runtime"
 	"github.com/matrixorigin/matrixone/pkg/logutil"
@@ -111,9 +108,6 @@ func NewPipelineClient(
 }
 
 func (c *pipelineClient) NewStream(backend string) (morpc.Stream, error) {
-	if backend == c.localServiceAddress {
-		return nil, moerr.NewInternalErrorNoCtx(fmt.Sprintf("remote run pipeline in local: %s", backend))
-	}
 	return c.client.NewStream(backend, true)
 }
 
--- a/pkg/cnservice/cnclient/client_test.go
+++ b/pkg/cnservice/cnclient/client_test.go
@@ -1,0 +1,55 @@
+// Copyright 2021 Matrix Origin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cnclient
+
+import (
+	"context"
+	"testing"
+
+	"github.com/matrixorigin/matrixone/pkg/common/morpc"
+	"github.com/stretchr/testify/require"
+)
+
+type testRPCClient struct {
+	backend string
+	lock    bool
+}
+
+func (c *testRPCClient) Send(ctx context.Context, backend string, request morpc.Message) (*morpc.Future, error) {
+	return nil, nil
+}
+
+func (c *testRPCClient) NewStream(backend string, lock bool) (morpc.Stream, error) {
+	c.backend = backend
+	c.lock = lock
+	return nil, nil
+}
+
+func (c *testRPCClient) Ping(ctx context.Context, backend string) error { return nil }
+func (c *testRPCClient) Close() error                                   { return nil }
+func (c *testRPCClient) CloseBackend() error                            { return nil }
+
+func TestPipelineClient_NewStreamAllowsLocalBackend(t *testing.T) {
+	rpcClient := &testRPCClient{}
+	client := &pipelineClient{
+		localServiceAddress: "127.0.0.1:1234",
+		client:              rpcClient,
+	}
+
+	_, err := client.NewStream("127.0.0.1:1234")
+	require.NoError(t, err)
+	require.Equal(t, "127.0.0.1:1234", rpcClient.backend)
+	require.True(t, rpcClient.lock)
+}
--- a/pkg/sql/colexec/dispatch/dispatch.go
+++ b/pkg/sql/colexec/dispatch/dispatch.go
@@ -133,6 +133,9 @@ func (dispatch *Dispatch) Call(proc *process.Process) (vm.CallResult, error) {
 	whichToSend := result.Batch
 	if result.Batch == nil {
 		result.Status = vm.ExecStop
+		if err := dispatch.waitRemoteRegsReadyBeforeStop(proc); err != nil {
+			return result, err
+		}
 		printShuffleResult(dispatch)
 		return result, nil
 	}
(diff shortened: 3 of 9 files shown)


LeftHandCold:
Review note for the 3.0-dev backport:\n\nLGTM from code review. The change preserves the existing 3.0-dev NewStream signature and only removes the local-backend rejection, which is the regression point for shuffle dispatch notify streams. The added unit test covers the intended local-backend path.\n\nValidated locally:\n- `GOCACHE=/private/tmp/mo-24159-3dev-gocache go test ./pkg/cnservice/cnclient -count=1`\n\nNote: `go test ./pkg/sql/compile -count=1` could not complete in the local sandbox because a lockservice test attempted to bind `/tmp/service-*.sock` and failed with `operation not permitted`.

XuPeng-SH (approved):
Reviewed from multiple angles:
- correctness: removing the local-backend rejection unblocks the valid local notify-stream path while the actual local-execution guard remains in Scope.RemoteRun
- performance/resource behavior: this only enables a normal morpc stream to the local CN for notify traffic and does not introduce new cleanup or execution-path issues
- tests: the new regression test pins the exact client-level failure, and the cnclient package test suite passes

Looks good to merge.

XuPeng-SH (changes requested):
Blocking on the latest head, not the original compile fix itself:

`origin/3.0-dev..HEAD` currently includes an unrelated backport commit `019554c53` (`fix: align CDC pause visible state (#24804)`). That commit brings in CDC/taskservice/frontend changes across files like `pkg/cdc/sql_builder.go`, `pkg/frontend/cdc_exector.go`, `pkg/taskservice/daemon_task.go`, and `pkg/cnservice/server_task.go`, which are outside the scope of this PR (`compile: allow local notify streams for shuffle dispatch`).

The compile/remoterun changes look reviewable on their own, but I can't approve this PR while it also merges an unrelated CDC pause-state backport. Please rebase/drop the extra commit (or split it into its own PR) and then I can re-review the compile fix on the cleaned range.

ouyuanning (approved):

XuPeng-SH (changes requested):
这次修复覆盖的都是 multi-CN remote pipeline / shuffle 协调问题，风险点是**用户可见的 hang / timeout**，但当前新增的测试基本都停留在 Go unit test。

我这边没再看到更硬的代码正确性 blocker，不过从回归保护的角度，这个 PR 还缺至少一层 SQL/BVT 级别的验证，特别是：

1. 一个能稳定走到 multi-CN distributed path 的 `INSERT ... SELECT ... JOIN` 回归 case，确认不会 hang 且结果正确；
2. 一个 **0-row** 变体（命中这次修的 no-data remote dispatch 路径），确认语句能正常结束而不是卡住。

这类问题之前就是在真实 distributed workload 下暴露出来的，只靠当前 unit test 我觉得还不够稳。建议把至少这两类 product-level regression 补上，再看合并。

XuPeng-SH (approved):

PR #24589, opened 2026-05-26
fix(plan): preserve decimal case casts on 3.0-dev

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24565

## What this PR does / why we need it:

Cherry-picks the CASE/IFF mixed DECIMAL fixes from #24572 to `3.0-dev`.

This preserves DECIMAL branch metadata when CASE/IFF common types need implicit casts, computes safe common precision as `max(width-scale)+max(scale)`, promotes DECIMAL family when the selected family cannot safely hold the common precision, and keeps metadata-cast behavior scoped to DECIMAL.

Local validation:

```text
go test ./pkg/sql/plan/function -run 'Test_(CaseCheck|IffCheck)_(DifferentDecimalScale|MixedDecimalFamilyScale|Decimal64PromotesForRequiredWidth|Decimal64PromotesForLargeScale|Decimal128PromotesToDecimal256)' -count=1\ngit diff --check\n```

PR #24572, opened 2026-05-25
fix(plan): preserve decimal scale casts in case and iff

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24565

## What this PR does / why we need it:

This PR fixes a CASE/IFF type-binding bug for mixed DECIMAL scales.

Before this change, `CASE WHEN` and `IFF` could treat DECIMAL branches with the same OID but different scale/width as already matched, so no implicit cast was inserted. In the reported case, the `THEN` branch stayed at `DECIMAL(23,2)` while the `ELSE` branch produced a multiplication result with scale 7, and the final CASE result was formatted with the wrong scale.

This change makes CASE/IFF require casts whenever branch types are not exactly equal to the computed common type, including DECIMAL scale/width differences. It also preserves the maximum DECIMAL scale and width from source branch types when deriving the common return type.

Added coverage:
- focused unit tests for `caseCheck` and `iffCheck`
- binder test covering mixed DECIMAL scales in CASE
- BVT reproduction in `test/distributed/cases/expression/case_when.sql`

Validation:
- `mo-tester` generate/update result for `case_when.sql`
- `mo-tester` run for `case_when.sql`: 87/87 passed
- `go test ./pkg/sql/plan/function -count=1`
- `go test ./pkg/sql/plan -count=1`
- `git diff --check`
- `make`
- `make static-check`

PR #24841, opened 2026-06-04
fix(plan): align coalesce decimal scale and support decimal256 comparison

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24565

PR #24572 / #24589 fixed the mixed-DECIMAL scale magnification for `CASE`/`IFF`
but missed two related cases:

1. **COALESCE scale magnification.** `COALESCE` over decimal branches with
   different scales returned the first branch's scale while carrying another
   branch's raw value, magnifying the result (e.g. by 10^5):

   ```sql
   SELECT COALESCE(
     CAST(NULL AS DECIMAL(23,2)),
     7.01970 * CAST(-58140.00 AS DECIMAL(23,2))
   );
   ```

   `coalesceCheck` short-circuited on a direct match without aligning
   scale/width. It now falls through to the alignment cast path unless every
   decimal branch already shares the same scale and width (new
   `decimalInputsAligned` guard).

2. **decimal256 comparison.** A `CASE`/`IFF` whose branches need more than 38
   digits is promoted to decimal256, but comparing such a value with a
   decimal128 failed with `ERROR 20203: invalid argument operator =, bad value
   [DECIMAL256 DECIMAL128]`:

   ```sql
   SELECT (CASE WHEN 1=1 THEN CAST(1 AS DECIMAL(38,0))
                ELSE CAST(0 AS DECIMAL(38,20)) END)
        = CAST(1 AS DECIMAL(38,20));
   ```

   decimal256 was unsupported across the whole comparison path.
   `fixedTypeCastRule1` now promotes a decimal-vs-decimal pair involving
   decimal256 to decimal256 (neutral for arithmetic operators, whose support
   checks still reject decimal256); the equal/other comparison support
   functions accept decimal256; and a new `valueDec256Compare` plus decimal256
   branches in the six comparison executors and `nullSafeEqualFn` evaluate it.

Scope is limited to decimal-vs-decimal comparisons.

Added coverage:
- unit tests for `fixedTypeCastRule1` decimal256 promotion, comparison support,
  and the decimal256 comparison executors
- coalesce scale-alignment / already-aligned unit tests
- BVT reproduction in `test/distributed/cases/expression/case_when.sql`

Validation:
- `go test ./pkg/sql/plan/function`, `go test ./pkg/sql/plan`
- `go vet`, `gofmt`, `molint`, `golangci-lint` (0 issues) on the changed package
- mo-tester: `case_when.sql` 100%, `func_coalesce_1.sql` 100%, `decimal256_insert.sql` 100%

### DIFF
--- a/pkg/sql/plan/function/func_binary.go
+++ b/pkg/sql/plan/function/func_binary.go
@@ -901,6 +901,22 @@ type NormalType interface {
 		constraints.Integer
 }
 
+// decimalInputsAligned reports whether all inputs (already known to share the
+// same decimal Oid) also share the same scale and width, so that coalesce can
+// be evaluated without inserting alignment casts.
+func decimalInputsAligned(inputs []types.Type) bool {
+	if len(inputs) == 0 {
+		return true
+	}
+	scale, width := inputs[0].Scale, inputs[0].Width
+	for i := 1; i < len(inputs); i++ {
+		if inputs[i].Scale != scale || inputs[i].Width != width {
+			return false
+		}
+	}
+	return true
+}
+
 func coalesceCheck(overloads []overload, inputs []types.Type) checkResult {
 	if len(inputs) > 0 {
 		if retType, ok := mixedStringNumericToVarchar(inputs); ok {
@@ -930,6 +946,19 @@ func coalesceCheck(overloads []overload, inputs []types.Type) checkResult {
 			if sta == matchFailed {
 				continue
 			} else if sta == matchDirectly {
+				// Decimals that match directly still need scale/width alignment
+				// across branches: without it the result inherits the first
+				// branch's scale while carrying another branch's raw value,
+				// magnifying the result (issue #24565). Only short-circuit when
+				// every decimal branch already shares the same scale and width.
+				if requireOid.IsDecimal() && !decimalInputsAligned(inputs) {
+					if cos < minCost {
+						minIndex = i
+						minCost = cos
+						minOid = requireOid
+					}
+					continue
+				}
 				return newCheckResultWithSuccess(i)
 			} else {
 				if cos < minCost {
--- a/pkg/sql/plan/function/func_binary.go
+++ b/pkg/sql/plan/function/func_binary.go
@@ -896,25 +896,29 @@ type NormalType interface {
 		types.Datetime |
 		types.Decimal64 |
 		types.Decimal128 |
+		types.Decimal256 |
 		types.Timestamp |
 		types.Uuid |
 		constraints.Integer
 }
 
-// decimalInputsAligned reports whether all inputs (already known to share the
-// same decimal Oid) also share the same scale and width, so that coalesce can
-// be evaluated without inserting alignment casts.
-func decimalInputsAligned(inputs []types.Type) bool {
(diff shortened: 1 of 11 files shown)


XuPeng-SH (changes requested):
Blocking correctness issue:

`BETWEEN` is still not fully wired for the new decimal256 comparison support.

1. In `pkg/sql/plan/function/list_operator.go`, the `BETWEEN` checker still rejects decimal cases when the cast results come back with different scales:

```go
if t01.Oid == types.T_decimal64 || t01.Oid == types.T_decimal128 || t01.Oid == types.T_decimal256 {
    if t01.Scale != t1.Scale || t02.Scale != t2.Scale {
        return newCheckResultWithFailure(...)
    }
}
```

So after the new decimal256 promotion, mixed-scale decimal cases that should compare after alignment can still fail at planning time.

2. In `pkg/sql/plan/function/operator_between.go`, `betweenImpl` has branches for `decimal64` and `decimal128`, but no `decimal256` branch at all. A non-constant decimal256 `BETWEEN` therefore falls through to `panic("unreached code")`.

The added SQL coverage only exercises a constant-foldable decimal256 `BETWEEN` case, so it does not cover the runtime path.

Please add the decimal256 execution path for `BETWEEN`, align the planner/check logic with the other decimal comparison operators, and add a non-constant runtime test so this path is actually exercised.

heni02 (approved):

aunjgr (approved):
Two fixes:

**1. COALESCE decimal scale alignment** (`func_binary.go`). When multiple decimal branches have different scales (e.g., `COALESCE(NULL, decimal(10,2), decimal(10,4))`), previously the result inherited the first matching branch's scale. This magnified values (issue #24565). Fix: `coalesceDecimalResult` computes the max integral width+scale across all inputs, promoting to decimal256 when combined precision overflows decimal128. Also fixes `matchDirectly` short-circuit for decimals — previously returned immediately with the first match's scale; now continues to find the best aligned type.

**2. Decimal256 comparison operators** (`func_compare.go` +255). Six comparison functions (`=`, `<>`, `<`, `<=`, `>`, `>=`) now support `T_decimal256` via `types.CompareDecimal256`. The existing generic path used `compareGeneric` which doesn't handle decimal256 correctly (treats as raw bytes). Tests in `func_compare_decimal256_test.go` cover all six operators + boundary values + NULL handling.

**Additional**: `operator_between.go` (+4) — adds `T_decimal256` to the BETWEEN/in_range guard that skips fixed-type ops for unsupported types. `type_check.go` (+13) — adds decimal256 to comparison type check.

LGTM.

XuPeng-SH (approved):
LGTM after rechecking from three angles:
- correctness: the earlier decimal256 BETWEEN gap is fixed in both planner and runtime paths
- regression risk: decimal256 support remains scoped to decimal-vs-decimal comparisons / aligned casts
- tests: coverage now includes the non-constant decimal256 BETWEEN runtime path in addition to planner checks and SQL repros

PR #24842, opened 2026-06-04
fix(plan): align coalesce decimal scale across branches

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24565

`COALESCE` over decimal branches with different scales returned the first
branch's scale while carrying another branch's raw value, magnifying the
result (e.g. by 10^5):

```sql
SELECT COALESCE(
  CAST(NULL AS DECIMAL(23,2)),
  7.01970 * CAST(-58140.00 AS DECIMAL(23,2))
);
-- before: magnified; after: -408125.3580000 (matches the direct multiplication)
```

`coalesceCheck` short-circuited on a direct match without aligning scale/width.
It now falls through to the alignment cast path unless every decimal branch
already shares the same scale and width (new `decimalInputsAligned` guard).

The companion decimal256-comparison scenario of issue #24565 (a decimal256
`CASE` result compared with a decimal128) is already handled on `main`, so this
PR only adds the missing `COALESCE` alignment. The `3.0-dev` branch lacks the
decimal256 comparison support, so that part is delivered separately in #24841.

Added coverage:
- `Test_CoalesceCheck_DecimalScaleAlignment` / `Test_CoalesceCheck_DecimalAligned_NoCast`

Validation:
- `go test ./pkg/sql/plan/function`, `go vet`, `gofmt`
- manual verification on a `main` build: the COALESCE result is no longer magnified
- mo-tester `case_when.sql` (incl. the COALESCE repro) passed 101/101 locally


heni02 (approved):

XuPeng-SH (changes requested):
I did not find a new hard correctness blocker in the implementation itself, but with a high bar I still think the regression coverage is too narrow for a user-visible type-inference fix.

Right now the added tests do prove the core bug is fixed for constant-expression cases, which is good. What is still missing are the more realistic/product-level cases where these decimal-type regressions usually hide:

1. **column-based COALESCE**
   The new BVT coverage is still mostly constant-expression driven. I would want at least one table-backed case like `COALESCE(dec23_2_col, dec38_7_col)` with rows where the first branch is NULL and non-NULL, so we lock the real execution path rather than only planner/unit helpers.

2. **mixed integer + decimal branches**
   This PR touches numeric branch unification, but I don’t see a meaningful BVT for cases like `COALESCE(int_col, decimal_col)` or `COALESCE(NULL, 1, CAST(... AS DECIMAL(...)))`. That is a very common regression surface.

3. **3+ branch COALESCE**
   The current coverage is still mostly 2-branch oriented. A case like `COALESCE(NULL, dec_a, dec_b)` or `COALESCE(NULL, NULL, dec_expr)` would better prove the final chosen decimal metadata is stable when more than two branches participate.

4. **metadata/type visibility**
   Since the bug is fundamentally about inferred decimal width/scale, I would also like one SQL-level metadata check (for example through `CREATE VIEW` + `DESC`) so we know the visible column type is correct, not just the runtime value.

So my conclusion is: the main bug is covered, but the user-visible regression surface is still thinner than I would like. I would be much more comfortable after a small set of realistic SQL/BVT cases is added.

aunjgr (changes requested):
The fix in func_binary.go is correct, but the BVT coverage is too thin — constants only. The 4.0-dev backport (#24899) already has the full BVT suite this needs: column-based COALESCE over mixed-scale decimals, mixed integer+decimal branches, 3-branch decimal cases, and a CREATE VIEW + DESC metadata check. Please port that BVT coverage here.

XuPeng-SH (approved):

aunjgr (approved):
The BVT coverage is actually complete — column-based, mixed int+decimal, 3-branch, and VIEW+DESC metadata check are all present. The fix in coalesceCheck is correct. No concerns.

Issue #24158, opened 2026-04-20
[Bug]: sendNotifyMessage fails with self-connection when shuffle dispatch operator is on local CN


jiangxinmeng1:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

main

### Commit ID

9305510dd

### Other Environment Information

```
- Multi-CN cluster (2+ CNs)
- Shuffle join with serial_full expression (secondary index UPDATE)
```

### Actual Behavior

When a shuffle join runs on a multi-CN cluster, `sendNotifyMessage` (scope.go) iterates `RemoteReceivRegInfos` and creates RPC connections to each dispatch operator's CN. When a dispatch operator happens to be on the **local CN** (same address as the coordinator), `pipelineClient.NewStream` (cnclient/client.go:115) rejects the self-connection:

```
if backend == c.localServiceAddress {
    return nil, moerr.NewInternalErrorNoCtx("remote run pipeline in local: %s", backend)
}
```

This causes the query to fail or hang. The issue was previously encountered in DELETE operations ([#14326](https://github.com/matrixorigin/matrixone/issues/14326), [#14353](https://github.com/matrixorigin/matrixone/issues/14353)) and was worked around by removing the dispatch in DELETE ([#14337](https://github.com/matrixorigin/matrixone/pull/14337)). However, the root cause in `sendNotifyMessage` was never fixed.

Now, with shuffle support for `serial_full()` expressions (needed for secondary index maintenance during UPDATE), the same issue resurfaces: the shuffle join's dispatch operator on the local CN triggers the self-connection rejection.

**Root cause analysis:**
- `RemoteRun` (scope.go:399) has `ipAddrMatch` check to handle the local case → falls back to `MergeRun` ✅  
- `sendNotifyMessage` (scope.go:842) does NOT have this check → directly tries RPC to local address → fails ❌
- A CN connecting to its own RPC server via TCP loopback is technically valid and safe

### Expected Behavior

`sendNotifyMessage` should handle the case where `fromAddr` matches the local CN address, either by:
1. Removing the self-connection check in `pipelineClient.NewStream` (safe because `RemoteRun` already has `ipAddrMatch` guard), or
2. Adding local dispatch notification handling in `sendNotifyMessage`

### Steps to Reproduce

1. Set up a multi-CN cluster (2+ CNs)
2. Create a table with a secondary index:
```sql
[...]

PR #24854, opened 2026-06-04
compile: allow local notify streams for shuffle dispatch (#24159)

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue https://github.com/matrixorigin/matrixone/issues/24158

This fixes the self-connection bug in the compile/remoterun notification path.

`Scope.RemoteRun` already handles the local-scope case with `ipAddrMatch`, but `Scope.sendNotifyMessage` still creates a pipeline RPC stream for every `RemoteReceivRegInfo.FromAddr`. When one dispatch operator is scheduled on the same CN as the coordinator, `fromAddr` equals the local service address. The old `pipelineClient.NewStream` rejected that loopback connection with `remote run pipeline in local`, so the prepare-done notification path could not be established for the local dispatch operator.

That can break remote dispatch/shuffle coordination for queries such as secondary-index UPDATE plans using shuffle join with `serial_full(...)` and surface as failure or hang.

This PR keeps the existing `RemoteRun` local-execution guard intact, but allows the notification stream itself to connect to the local CN address. That is the smallest fix that restores the existing dispatch notification protocol without refactoring local/remote receiver handling.

Also included:
- a unit test that pins the real regression point in `pipelineClient.NewStream`

### DIFF
--- a/pkg/cnservice/cnclient/client.go
+++ b/pkg/cnservice/cnclient/client.go
@@ -15,10 +15,7 @@
 package cnclient
 
 import (
-	"fmt"
-
 	"github.com/fagongzi/goetty/v2"
-	"github.com/matrixorigin/matrixone/pkg/common/moerr"
 	"github.com/matrixorigin/matrixone/pkg/common/morpc"
 	"github.com/matrixorigin/matrixone/pkg/common/runtime"
 	"github.com/matrixorigin/matrixone/pkg/logutil"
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
LGTM. Allowing local notify streams here matches the existing higher-level guard in Scope.RemoteRun, fixes the self-connection regression in the prepare-done notify path, and keeps the change narrowly scoped. The added cnclient regression test is also a good pin for the underlying behavior.

PR #24857, opened 2026-06-04
V3.0.0 5e4210029 hotfix 20260605

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #


LeftHandCold:
Addressed the Copilot review items in 4fd5090b02:

- Added WAL recovery sanity checks for entry count, file-size feasibility, per-entry data length, and remaining bytes before allocation.
- Avoided preallocating the entries slice directly from untrusted file count.
- Changed buildUserEntryCmd to use headerSize and leaseHolderIDSize instead of hard-coded offsets.
- Ensured walRecoveryInProgress is cleared via defer on bootstrap early returns.
- Added malformed/truncated WAL file tests and command layout coverage.

Note: make([]byte, uint32) is valid Go, so that part of the review was inaccurate. The real issue was the missing upper bound, now guarded.

Validated locally:
- go test ./pkg/logservice -run "TestReadWALDataFile|TestBuildUserEntryCmd" -count=1
- go test ./pkg/logservice -count=1
- go test ./pkg/hakeeper -run "TestHandleInitialClusterRequestCmd" -count=1

LeftHandCold:
Follow-up from local restore test:

- The single-node restore log showed `file /tmp/recovery/hakeeper_backup.data is not found`, so the configured restore data was not loaded. Previously this was silently treated as no backup data and the service continued as an empty bootstrap.
- Updated MatrixOne in `fefef57629` to fail fast when an explicit restore file path is configured but missing, while keeping the default `hk_data` compatibility behavior.
- Also updated mo-backup `log_repair` branch in `f3050b6` to include `_mo_bootstrap=1` when generating restore `BackupData` if the source HAKeeper state did not contain it.

Local validation:
- `go test ./pkg/logservice -run TestGetBackupData -count=1`
- `go test ./pkg/logservice -run "TestDetermineRecoveryParams|TestMergeHAKeeperStateInfo" -count=1` in mo-backup

aptend (dismissed):
Found one blocker in the embed wait path.

 now unconditionally calls . Existing unit coverage constructs  and calls  directly, so this panics with a nil logger before it can check HAKeeper state:

```
--- FAIL: Test_waitHAKeeperRunningLocked
github.com/matrixorigin/matrixone/pkg/embed.(*operator).waitHAKeeperRunningLocked
    pkg/embed/operator.go:402
```

Please avoid assuming  is initialized in this helper, e.g. use the package logger/logutil for this informational log or guard/init the logger in the helper/test setup.

Validation I ran:
- FAIL	github.com/matrixorigin/matrixone/pkg/logservice [build failed]
FAIL with local CGO flags: passed
- FAIL	github.com/matrixorigin/matrixone/pkg/hakeeper [build failed]
FAIL with local CGO flags: passed
- FAIL	github.com/matrixorigin/matrixone/pkg/embed [build failed]
FAIL with local CGO flags: failed as above
- : passed

aptend (changes requested):
Found one blocker in the embed wait path.

pkg/embed/operator.go:402 now unconditionally calls op.reset.logger.Info(...). Existing unit coverage constructs op := &operator{} and calls waitHAKeeperRunningLocked directly, so this panics with a nil logger before it can check HAKeeper state:

--- FAIL: Test_waitHAKeeperRunningLocked
github.com/matrixorigin/matrixone/pkg/embed.(*operator).waitHAKeeperRunningLocked
    pkg/embed/operator.go:402

Please avoid assuming op.reset.logger is initialized in this helper, e.g. use the package logger/logutil for this informational log or guard/init the logger in the helper/test setup.

Validation I ran:
- go test ./pkg/logservice -run 'TestReadWALDataFile|TestBuildUserEntryCmd|TestGetBackupData' -count=1 with local CGO flags: passed
- go test ./pkg/hakeeper -run TestHandleInitialClusterRequestCmdPatchesIDsWhenAlreadyInitialized -count=1 with local CGO flags: passed
- go test ./pkg/embed -run 'Test_waitHAKeeperRunningLocked|Test_waitAnyShardReadyLocked' -count=1 with local CGO flags: failed as above
- git diff --check origin/v3.0.0-5e4210029-hotfix...origin/pr-24857: passed

Issue #24099, opened 2026-04-09
Bug: data branch merge fails with ErrTxnNeedRetryWithDefChanged in multi-CN PESSIMISTIC mode


XuPeng-SH:
## Summary

`data branch merge` fails intermittently in multi-CN PESSIMISTIC mode with:

```
txn need retry in rc mode, def changed
```

Error code: `ErrTxnNeedRetryWithDefChanged` (20631)

## Reproduction

Observed in CI: **multi cn e2e bvt test docker compose(PESSIMISTIC)**  
Test file: `test/distributed/cases/git4data/branch/merge/merge_1.sql` line 85

```sql
data branch merge t2 into t1
```

## Root Cause

`data branch merge` executes internal SQL (CREATE TABLE, INSERT, DELETE) via `runSql()` in `data_branch_helpers.go:168`:

```go
opts := executor.Options{}.
    WithDisableIncrStatement().  // disables compile retry
    WithTxn(backSes.GetTxnHandler().GetTxn()).
    WithKeepTxnAlive()
```

`WithDisableIncrStatement()` sets `disableRetry = true` in the compile path (`sql_executor.go:418`), which prevents `compile2.go` from retrying on `ErrTxnNeedRetryWithDefChanged`.

When any internal INSERT/DELETE hits a lock conflict on `mo_tables` catalog rows (common in multi-CN PESSIMISTIC mode), `lock_meta.go:198-203` converts it to `ErrTxnNeedRetryWithDefChanged`. Since retry is disabled, the error propagates directly to the client.

Additionally, `handleBranchMerge` in `self_handle.go` has no retry wrapper of its own.

### Call chain

```
handleBranchMerge (self_handle.go)
  → diffMergeAgency (data_branch.go)
    → mergeDiffs / diffOnBase (parallel)
      → flushSqlValues → execSQLStatements → runSql (data_branch_helpers.go)
        → SQL executor with disableRetry=true
          → compile2 (retry disabled)
            → lock_meta.go:198 → ErrTxnNeedRetryWithDefChanged
              → error surfaces to client
```

### Why it is flaky

In multi-CN PESSIMISTIC mode, distributed lock contention on catalog metadata rows is non-deterministic. The merge creates temp tables via CTAS and then applies INSERT/DELETE to the base table, creating multiple lock points where conflicts can occur. Single-CN tests rarely trigger this.

## Suggested Fix Directions

[...]

PR #24860, opened 2026-06-05
 optimize data branch merge apply execution

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Related issue #24099

This is the `main` cherry-pick of #24858.

`DATA BRANCH MERGE` materializes row differences into `__mo_diff_*` helper tables, then applies them to the base table with base-table DML that reads those helper tables.

`runSql` previously routed every statement mentioning `__mo_diff_*` through `BackgroundExec`. That was needed for temp-table DDL and temp-table-targeting DML, but it was too broad for the base-table delete/insert apply statements. On a 100M-row table with only 1000 updated rows, the apply phase spent seconds on the slower path.

This change keeps `BackgroundExec` for data-branch temp table DDL and DML that targets `__mo_diff_*` tables, while allowing base-table DML that only reads those helper tables to use the internal SQL executor.

Validation on `main`:

- `make cgo`
- `CGO_CFLAGS="-I$(pwd)/thirdparties/install/include" go test ./pkg/frontend`

### DIFF
--- a/pkg/frontend/data_branch_helpers.go
+++ b/pkg/frontend/data_branch_helpers.go
@@ -43,6 +43,42 @@ func containsDataBranchTempTableName(sqlLower string) bool {
 		containsTempTableMarker(sqlLower, "__mo_diff_ins_")
 }
 
+func dataBranchTempSQLNeedsBackExec(sqlLower string) bool {
+	if !containsDataBranchTempTableName(sqlLower) {
+		return false
+	}
+
+	switch {
+	case strings.HasPrefix(sqlLower, "drop table"):
+		return true
+	case strings.HasPrefix(sqlLower, "create table"):
+		return true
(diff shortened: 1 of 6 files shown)


XuPeng-SH (approved):
LGTM after reviewing from three angles:
- correctness: the execution routing now only forces back-exec when the temp diff table itself is the DDL/DML target, while main-table DML that reads temp tables keeps the internal path
- regression risk: the change stays conservative for unknown temp-table statements and separately fixes subquery SQL formatting to preserve string literals
- tests: unit tests and distributed SQL coverage were both extended for the main-table DML path and the quoted-string KEYS subquery case

heni02 (approved):

Issue #24849, opened 2026-06-04
BVT: vector_hnsw_async.sql HNSW async search returns non-exact Top1


jiangxinmeng1:
## Summary

BVT failed in `test/distributed/cases/pessimistic_transaction/vector/vector_hnsw_async.sql` at row 54.

CI job: https://github.com/matrixorigin/matrixone/actions/runs/26940644255/job/79481042655?pr=24844

## Failure

SQL:

```sql
select * from t2 order by L2_DISTANCE(b, "[0, 16, 35, 5, 32, 31, 14, 10, 11, 78, 55, 10, 45, 83, 11, 6, 14, 57, 102, 75, 20, 8, 3, 5, 67, 17, 19, 26, 5, 0, 1, 22, 60, 26, 7, 1, 18, 22, 84, 53, 85, 119, 119, 4, 24, 18, 7, 7, 1, 81, 106, 102, 72, 30, 6, 0, 9, 1, 9, 119, 72, 1, 4, 33, 119, 29, 6, 1, 0, 1, 14, 52, 119, 30, 3, 0, 0, 55, 92, 111, 2, 5, 4, 9, 22, 89, 96, 14, 1, 0, 1, 82, 59, 16, 20, 5, 25, 14, 11, 4, 0, 0, 1, 26, 47, 23, 4, 0, 0, 4, 38, 83, 30, 14, 9, 4, 9, 17, 23, 41, 0, 0, 2, 8, 19, 25, 23, 1]") ASC LIMIT 1;
```

Expected first row: `a = 0`, where the query vector exactly matches the loaded row.

Actual first row: `a = 2`.

Error excerpt:

```text
ERROR RSRow:58 - The value of [row:0,column:0] does not equal with each other,one is [0],but the other is [2]
```

## Initial Analysis

The query vector is row `a=0` in `sift128_base_10k.csv.gz`, so exact scan should return distance 0 for `a=0`. The BVT currently asserts exact Top1 through a query shape that can be rewritten to HNSW search. HNSW is approximate, and this async index test also depends on CDC/index sync timing, so asserting an exact Top1 row is unstable.

The likely fix is to make exact-result assertions bypass vector-index rewrite, for example with `LIMIT 1 by rank with option 'mode=force'`, or otherwise adjust the test to validate HNSW async index behavior without requiring exact ANN Top1 recall.

## Affected test

`test/distributed/cases/pessimistic_transaction/vector/vector_hnsw_async.sql`

[...]

Issue #735, opened 2021-09-15
Data lost after restart server


JinHai-CN:
#### Can be reproduced ?
Yes

#### Steps:
1. Start server
2. create database and table
3. Insert data and select count(*) from table;
4. Ctrl+c to stop server.
5. start server again and select count(*) from table; shows "Empty set".

#### Expected behavior:
Correct answer.

#### Actual behavior:
Empty set

#### Environment:
- Version or commit-id (e.g. v0.1.0 or 8b23a93): c499bfc24dfac0786dd4e0d7e0356a9142d3a754
- Hardware parameters:
- OS type: OSX
- Others:

#### Configuration file:
Default configuration file.

#### Additional context:
- Error message from client:
- Server log:
- Other information:

PR #24867, opened 2026-06-05
fix: hnsw run build with single thread to avoid usearch known bug (usearch #735)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24849

set number of thread for build to 1.

this is a known issue in usearch. https://github.com/unum-cloud/USearch/issues/735


fengttt (approved):

XuPeng-SH (changes requested):
我这里有一个需要在这个 PR 里一起处理的 correctness 问题：**目前只修了一半**。

这次 workaround 在 `pkg/vectorindex/hnsw/build.go` 里把 `NewHnswBuild` 强制成了单线程，这能覆盖同步 build / create index 的路径；但 `pkg/vectorindex/hnsw/sync.go` 里的 CDC / sync 路径还保留着同一个 usearch 并发 add 风险：

1. `insertAllInParallel()` 仍然按 `s.tblcfg.ThreadsBuild` 起多个 goroutine，最后并发对同一个 model 做 `last.AddWithoutIncr(...)`。
2. 更关键的是，`sync.go` 里新建 model 仍然用 `NewHnswModelForBuild(..., int(s.tblcfg.ThreadsBuild), ...)`，而 `model.go` 会把这个值传给 `idx.Index.ChangeThreadsAdd(idx.NThread)`，所以就算外层逻辑走“顺序更新”，库内部 add 线程数还是 >1。

也就是说，PR 说明里引用的 USearch #735 race 在 build path 被绕开了，但在 sync / CDC path 里仍然存在，用户还是会碰到 orphan node / exact self-search miss 这种 correctness 问题。

如果这次要用“force single build thread”作为临时兜底，我建议至少把 `sync.go` 这边的 build/add 线程也一起压到 1；否则现在这个修复范围和问题描述并不一致。

XuPeng-SH (changes requested):
I re-checked the latest update. It does improve the previous state by forcing `NewHnswModelForBuild()` itself to use `nthread := 1`, so the model no longer relies on caller-provided `ThreadsBuild` for `ChangeThreadsAdd()`.

However, I still do not think the sync/CDC hole is fully closed. The pure-insert CDC path in `pkg/vectorindex/hnsw/sync.go` still goes through `insertAllInParallel()`, which spawns `ThreadsBuild` goroutines and has them call `last.AddWithoutIncr()` concurrently against the same model. So even if the model is configured with `ChangeThreadsAdd(1)`, the application layer is still doing concurrent add calls into the same USearch index.

That means the fix now closes the library-internal threading part, but not the application-level concurrent-add part. Since the issue being worked around is specifically a concurrent add race in USearch, I do not think this is safe to approve until the sync pure-insert path is serialized as well, or otherwise proven safe.

XuPeng-SH (approved):
Reviewed the HNSW workaround. Forcing single-threaded build on both the create-index path and the build-model path is a reasonable short-term correctness fix for the upstream USearch orphan-node race, and I do not see a blocking correctness issue in the change.

PR #24868, opened 2026-06-06
feat: add datasync CLI

volgariver6:
## Summary
- Add standalone `datasync` Go CLI for MatrixOne table export/import orchestration.
- Support config-driven multi-tenant discovery, table planning, retries, sync/export/import modes, and JSON/CSV reports with exported CSV file sizes.
- Add unit coverage plus opt-in local MatrixOne integration coverage.

## Test Plan
- `cd datasync && rtk go test ./...`
- `cd datasync && rtk go vet ./...`
- `cd datasync && rtk go mod tidy -diff`
- `cd datasync && rtk go build ./cmd/datasync`
- `cd datasync && rtk proxy env DATASYNC_INTEGRATION=1 go test ./tests/integration -v`


datasync/internal/app/app.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (a App) Run(ctx context.Context, runID string) (Result, error) {
+	if a.Discovery == nil {
+		a.Discovery = MatrixOneDiscovery{}
+	}
+
+	discovered := make(map[plan.DatabaseKey][]string)
+	for _, source := range a.Config.Sources {
+		for _, database := range source.Databases {
+			tables, err := a.Discovery.ListTables(ctx, source, database.Name)
```
volgariver6: `App.Run` always calls `Discovery.ListTables` before building tasks, but `-mode import` is documented to skip `mo-dump` and import SQL/CSV from the existing run directory. If the original source tenant is unavailable during a restore/import-only run, this source discovery fails before the existing export files can be used; please branch import mode to build tasks from the prior run report/export directory or otherwise avoid requiring live source discovery.

volgariver6: Fixed in 3a4e2406f. Import mode now reads tasks from the existing `runs/<run_id>/report.json` instead of calling source discovery, so it can restore from previously exported SQL/CSV files without live source tenants. Added focused app/report tests and re-ran `go test ./... -coverprofile=coverage.out -covermode=atomic` (87.3%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/app/app.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func tasksFromReport(path string) ([]plan.Task, error) {
+	runReport, err := report.Read(path)
+	if err != nil {
+		return nil, err
+	}
+	tasks := make([]plan.Task, 0, len(runReport.Tables))
+	for _, row := range runReport.Tables {
+		tasks = append(tasks, plan.Task{
```
volgariver6: `tasksFromReport` rebuilds import-mode tasks from the previous report, but it only keeps table identity fields while `Runner.runTask` leaves `SourceRows` at zero when `ModeImport` skips source counting; `App.Run` then writes this new report back to the same `runs/<run_id>/report.json`. Because the report contract includes per-table source row count and summary total source rows, an import-only restore erases the export-time source counts after a successful import; please preserve the prior `SourceRows`/summary when rebuilding import reports, or avoid overwriting those fields.

volgariver6: Fixed in 4af1eff8e7bbd75d022093c89ab30b56ba6efc7f. Import-mode tasks now carry the prior report `SourceRows`, and `Runner.runTask` initializes the new table report from that value before import skips source counting, so the overwritten report keeps per-table and summary source row counts. Added app/runner regression coverage plus integration assertions for `source_rows` and `total_source_rows`. Verified with `go test ./... -coverprofile=coverage.out -covermode=atomic` (87.3%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/plan.go
```diff
@@ -0,0 +1,80 @@
...
+		discoveredSet := make(map[string]struct{}, len(discovered))
+		for _, table := range discovered {
+			discoveredSet[table] = struct{}{}
+		}
+
+		candidates := discovered
+		if len(database.IncludeTables) > 0 {
+			candidates = database.IncludeTables
+		}
+		for _, table := range candidates {
```
volgariver6, line 68: `BuildTasks` appends a task for every candidate entry, but `include_tables` is user-controlled and `validateTableList` only rejects empty names, not duplicates. If a config contains `include_tables: [t1, t1]` and `parallelism > 1`, both tasks use the same `taskPaths` export directory and target table, so concurrent workers can remove/write the same files and import the same table at the same time; please reject duplicate table names during config validation or de-duplicate candidates before appending tasks.

datasync/internal/config.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func mergeDatabaseEndpoint(defaults Endpoint, endpoint DatabaseEndpoint) DatabaseEndpoint {
+	if endpoint.Name == "" {
+		endpoint.Name = defaults.Name
+	}
+	if endpoint.Host == "" {
+		endpoint.Host = defaults.Host
+	}
+	if endpoint.Port == 0 {
```
volgariver6, line 494: `mergeDatabaseEndpoint` treats `Port == 0` as missing, but YAML decoding cannot distinguish an omitted `port` from an explicit `port: 0`. Since database-level connection fields are supposed to override global defaults, a typo like `databases[0].source.port: 0` with a global source port silently inherits the global value and passes `validateDatabaseEndpoint` instead of rejecting the explicit invalid database value; please track field presence for inherited endpoint fields or validate explicit zero ports before merging.

volgariver6: Fixed in 12b11c138. Database task parsing now records whether source/target `port` was explicitly present, and validation rejects explicit database-level `port: 0` before global endpoint inheritance can fill it. Added regression coverage for explicit zero ports with global defaults. Verified with `go test ./internal`, `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.3%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/plan.go
volgariver6: Fixed in 12b11c138. `validateTableList` now rejects duplicate entries in both `include_tables` and `exclude_tables`, so the planner cannot create duplicate tasks that share the same export directory and target table. Added config validation regression cases for both duplicate lists. Verified with `go test ./internal`, `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.3%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/runner.go
```diff
@@ -0,0 +1,313 @@
...
+
+func (r Runner) effectiveMode() Mode {
+	if r.Mode == "" {
+		return ModeSync
+	}
+	return r.Mode
+}
+
+func (r Runner) taskPaths(runID string, task Task) (string, string, string) {
+	tableDir := filepath.Join(r.Config.OutputDir, runID, "exports", task.SourceName, task.SourceDatabase, task.SourceTable)
```
volgariver6: `taskPaths` keys export files only by source name/database/table, but `BuildTasks` can produce multiple tasks for the same source table mapped to different targets because each `databases` entry has an independent target. With `parallelism > 1`, those tasks share `tableDir`; each export attempt calls `os.RemoveAll(tableDir)` before `mo-dump`, so one worker can delete another target task SQL/CSV before import or cleanup can remove files needed by a later import-mode run. Please include target identity, or another per-task unique key, in the export path.

datasync/internal/app.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	return tasks, nil
+}
+
+func targetPasswordForReportRow(cfg *Config, row TableReport) string {
+	for _, database := range cfg.Databases {
+		if database.Source.Name == row.SourceName &&
+			database.Source.Database == row.SourceDatabase &&
+			database.Target.Name == row.TargetName &&
+			database.Target.Database == row.TargetDatabase {
```
volgariver6, line 156: `targetPasswordForReportRow` matches only source/target name and database, but the new config lets each `databases` entry override target host/port/user independently and the report row already carries those fields. With two import rows for the same source database and target name/database on different target endpoints, this returns the first matching password, so `EnsureTargetDatabase`/`mysql source` authenticates to the row host with another endpoint secret. Please include the reported target host/port/user in the lookup and fail task building when no exact match is found.

volgariver6: Fixed in ef4e8dda6a88a00072548d5f0e3c8958ee152631. Import-mode report task rebuilding now matches target password by the full reported target endpoint (`target_name`, `target_host`, `target_port`, `target_user`, and `target_database`) and fails task building if no exact match exists, instead of silently using an empty or wrong password. Added `TestTasksFromReportMatchesTargetPasswordByFullTargetEndpoint`. Verified with `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.1%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/runner.go
volgariver6: Fixed in ef4e8dda6a88a00072548d5f0e3c8958ee152631. Export paths now include target identity (`target_name`, `target_database`, `target_host`, `target_port`, and `target_user`) after the source table path, so the same source table mapped to different targets no longer shares a table export directory. Added `TestTaskPathsSeparateSameSourceTableForDifferentTargets` and updated the stable path assertion. Verified with `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.1%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/app.go
```diff
@@ -0,0 +1,228 @@
...
+		a.Discovery = MatrixOneDiscovery{}
+	}
+
+	discovered := make(map[DatabaseKey][]string)
+	for _, database := range a.Config.Databases {
+		tables, err := a.Discovery.ListTables(ctx, database.Source, database.Source.Database)
+		if err != nil {
+			return nil, err
+		}
+		discovered[DatabaseKey{SourceName: database.Source.Name, Database: database.Source.Database}] = tables
```
volgariver6: `buildTasks` discovers each configured source endpoint, but the `discovered` map is keyed only by source name and database. The new config allows each `databases` entry to override source host/port/user independently, and there is no uniqueness validation on source name/database, so two independent source endpoints with the same name and database overwrite each other here; `BuildTasks` then plans the first endpoint using the second endpoint table list, causing missing or wrong exports. Please include the full source endpoint identity in `DatabaseKey`, or otherwise keep discovery results scoped to the specific database entry.

volgariver6: Fixed in 5a2f3361ec02b9dd351f5a91727aa6270584099e. `DatabaseKey` now includes the full source endpoint identity (`source_name`, `source_host`, `source_port`, `source_user`, and database), and both discovery storage and task planning use the same `databaseKey` helper, so two independent sources with the same name/database keep separate discovered table sets. Added `TestRunKeepsDiscoveryResultsScopedToFullSourceEndpoint`, which failed before the fix with only one planned task. Verified with `go test ./internal -count=1`, `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.2%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/internal/plan.go
```diff
@@ -0,0 +1,93 @@
...
+				continue
+			}
+			tasks = append(tasks, Task{
+				SourceName:     database.Source.Name,
+				SourceHost:     database.Source.Host,
+				SourcePort:     database.Source.Port,
+				SourceUser:     database.Source.User,
+				SourcePassword: database.Source.Password,
+				SourceDatabase: database.Source.Database,
+				SourceTable:    table,
```
volgariver6, line 109: `BuildTasks` only carries the source table name into each task, and the importer later runs the dump SQL in `task.TargetDatabase`, so two database mappings that point different source databases at the same target database with the same table name both write the same target table. Because `taskPaths` is now target-scoped these exports no longer collide on disk, but the import still races/overwrites in MatrixOne and both tasks can report success after counting the same final table. Please reject duplicate target endpoint/database/table tasks during planning, or add an explicit target table mapping so these cases cannot silently overwrite each other.

volgariver6: Fixed in ea37a4b9c5d7e815ce9b6dde1391a0cedac4822e. BuildTasks now rejects duplicate target endpoint/database/table tasks before returning the plan, so two source mappings cannot silently import into the same target table. Added TestBuildTasksRejectsDuplicateTargetTable. Verified with `go test ./internal -run 'TestBuildTasksRejectsDuplicateTargetTable|TestBuildTasks' -count=1`, `go test ./... -coverprofile=coverage.out -covermode=atomic` (85.3%), `go vet ./...`, `go mod tidy -diff`, `go build ./cmd/datasync`, and `DATASYNC_INTEGRATION=1 go test ./tests/integration -v`.

datasync/cmd/datasync/main.go
```diff
@@ -0,0 +1,135 @@
...
+		Options: datasync.Options{
+			CleanupExportAfterImport: *cleanupExportAfterImport,
+		},
+		Progress: stderr,
+	}.Run(context.Background(), *runID)
+	if err != nil {
+		fmt.Fprintf(stderr, "datasync failed: %v\n", err)
+		return 1
+	}
+	fmt.Fprintf(stdout, "run_id=%s mode=%s planned_tasks=%d succeeded=%d failed=%d export_report=%s import_report=%s summary_report=%s\n",
```
volgariver6, line 124: `exitCode` always prints `export_report`, `import_report`, and `summary_report`, but `Write` only creates export+summary files in `ModeExport` and import+summary files in `ModeImport` (`internal/report.go:151-158`). Since the final stdout line is documented as machine-readable, export-only runs currently advertise a non-existent `import-report.md` path and import-only runs advertise a non-existent `export-report.md`; please either omit/empty reports that were not written for the selected mode or have `Write` create every path that the CLI reports.

XuPeng-SH (changes requested):
Reviewed 16372313bc1b458c5d63fe7c7637be1459a876a3; REQUEST_CHANGES. Two deterministic counterexamples fail against the exact implementation.

### [P1] Do not key destructive target ownership by a display label

BuildTasks in datasync/internal/plan.go includes TargetName in targetTableKey. Two entries with the identical target host/port/user/database/table but different target.name values are accepted as distinct tasks. Each import runs DROP TABLE/CREATE TABLE/LOAD, so parallel execution races destructively and serial execution silently replaces one source with the other. Row-count checks cannot detect replacement when both sources have equal counts. Different login users in the same tenant are likewise not necessarily different physical table owners.

Fix: define canonical destination identity independently of display names and login credentials (server/tenant/database/table), reject multiple planned writers before any export/import, and apply the same check to tasks reconstructed from reports. Where tenant/server alias identity cannot be resolved safely, require an explicit canonical destination identifier or reject ambiguous mappings. Keep display labels only for reporting.

### [P1] Constrain paths before recursive cleanup

Runner.taskPaths concatenates raw config/discovered names with filepath.Join, then runTaskWithProgress calls os.RemoveAll on the result before every export attempt. validateEndpoint only checks non-empty fields/ports. A validated source.name of `../../../../outside` yields `/outside/db/t/label-one/db/target-host/6001/dst:admin` from `/safe/output/run/exports`, outside the run sandbox. run-id and other interpolated components also need a defined path contract.

Fix: use opaque/encoded task-directory components, validate run IDs, and enforce containment under an explicitly created run root before any create/remove operation. Use an exclusive run directory/ownership marker and handle symlinked ancestors safely; avoid deleting arbitrary pre-existing paths merely because a configuration maps there. Keep SQL identifiers and display names separate from filesystem names. Add traversal/absolute/separator/symlink and reused-run tests with sentinels proving unrelated files survive.

Fresh validation: copied exact-head Go/module files into an isolated review directory and ran `go test ./...`: existing tests PASS (external integration remains opt-in; no real MO integration claimed). Added two non-destructive tests using the exact validate/BuildTasks/taskPaths methods: (1) same physical target, distinct labels accepts two tasks; (2) validated traversal name escapes export root. Both expected safety assertions FAIL deterministically in 0.003s. No RemoveAll, external command, database write or destructive reproduction was performed by these counterexample tests.

Scope: complete change map and 244-line design read, with planner/configuration, task orchestration, retry/import and report-resume paths reviewed. This is not a full approval of every reporting/parser branch. Preserve per-table progress/error reporting; also move exported-file validation inside the export retry closure so a successful command with incomplete artifacts receives the documented retries. Do not enable destructive import/cleanup without the ownership checks above.

PR #24877, opened 2026-06-08
fix(disttae): respect txn offset for s3 tombstones

lxw665:
## What type of PR is this?

- [x] BUG

issue #

Uncommitted S3 tombstones were collected from the transaction-wide `cn_flushed_s3_tombstone_object_stats_list`. That side map is not scoped by `txnOffset`, so a statement could apply S3 tombstones produced later in the same transaction instead of only seeing writes visible to that statement snapshot.

This breaks statement snapshot semantics for large DML paths that flush updates/deletes to S3, especially consecutive UPDATEs in the same transaction.

This PR makes uncommitted S3 tombstone collection follow the same visibility rule as uncommitted S3 data objects: collect from `txn.writes[:txnOffset]` and filter by database/table.

## Tests

- `go test -tags matrixone_test ./pkg/vm/engine/disttae -count=1`
- `git diff --check`
- Added BVT case: `test/distributed/cases/dml/update/update_s3_tombstone_txn_offset.sql`


lxw665:
Addressed Copilot's snapshot-offset comment: CollectTombstones now passes the computed snapshot-aware offset into getUncommittedS3Tombstone. Re-ran go test -tags matrixone_test ./pkg/vm/engine/disttae -count=1.

lxw665:
Addressed the latest Copilot review comments:\n\n- Cached visible workspace S3 tombstones on LocalDisttaeDataSource, so applyWorkspaceFlushedS3Deletes no longer rescans txn.writes[:txnOffset] for every block.\n- Removed the now-dead cn_flushed_s3_tombstone_object_stats_list / StashFlushedTombstones plumbing and related test setup. Visibility now comes from txn.writes with txnOffset filtering only.\n\nValidation: go test -tags matrixone_test ./pkg/vm/engine/disttae -count=1

lxw665:
Addressed the latest Copilot comment in txn_test.go. The integer range compiled under this repo's Go 1.25.4 toolchain, but I changed it to a conventional index loop with a separate variable for clarity and to avoid review noise. I also re-reviewed the tombstone changes: the LocalDisttaeDataSource cache now assigns after a successful collection to avoid partial-cache state if the helper ever returns an error. Validation: go test -tags matrixone_test ./pkg/vm/engine/disttae -count=1; git diff --check.

lxw665:
Verified the new BVT locally with mo-tester against local MatrixOne v3.0.4 on 127.0.0.1:6001. The first run exposed that the hand-written .result used the old/simple header format, so I regenerated it with mo-tester genrs and updated the PR.\n\nCommands:\n- ./run.sh -m genrs -n -g -p /Users/lxw665/workspace/matrixone/test/distributed/cases/dml/update/update_s3_tombstone_txn_offset.sql\n- ./run.sh -n -g -p /Users/lxw665/workspace/matrixone/test/distributed/cases/dml/update/update_s3_tombstone_txn_offset.sql\n\nResult: TOTAL 18, SUCCESS 18, FAILED 0, ABNORMAL 0, SUCCESS RATE 100%.

PR #24878, opened 2026-06-08
fix: bound LOAD first-line row size sampling

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue https://github.com/matrixorigin/matrixone/pull/24855

After #24855, LOAD stats estimate row size from the first data line. For the TPCH 1T `lineitem` load on TKE/COS, the planner opened the object with `Size: -1` and then closed the reader after reading the first line. That made the plan stage spend a long time draining/closing a huge object stream before execution could start.

This PR keeps the cardinality fix from #24855 while making the first-line read bounded. It:

- Bounds LOAD first-line row-size sampling to 1MiB instead of reading the rest of the object.
- Falls back to schema-based row size when no complete line is found in the bounded sample.
- Keeps closing the bounded reader and routes generic external stats through the same bounded helper.
- Adds coverage for long rows without a newline in the sample, bounded reads with a newline, and small files without a trailing newline.

Tests:

- `git diff --check`
- `go test ./pkg/sql/plan -run 'TestMakeLoadExternalStats|TestReadExternalFirstLineSize|TestLoadParquetMayListFiles|TestTotalLoadFileSize' -count=1`

### DIFF
--- a/pkg/sql/plan/build_load.go
+++ b/pkg/sql/plan/build_load.go
@@ -36,8 +36,9 @@ import (
 )
 
 const (
-	LoadParallelMinSize = int(colexec.WriteS3Threshold)
-	LoadWriteS3MinSize  = 1 << 20
+	LoadParallelMinSize     = int(colexec.WriteS3Threshold)
+	LoadWriteS3MinSize      = 1 << 20
+	loadFirstLineSampleSize = 1 << 20
 )
 
 func getNormalExternalProject(stmt *tree.Load, ctx CompilerContext, tableDef *TableDef, tblName string) ([]*Expr, map[string]int32, map[string]int32, *TableDef, error) {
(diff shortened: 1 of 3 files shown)


aunjgr (approved):
Clean fix. Previously `readExternalFirstLineSize` did `Size: -1` (read entire file) to find the first newline for row-size estimation. For large files this was wasteful and potentially problematic.

Changes:
- `Size` now bounded to `loadFirstLineSampleSize = 1<<20` (1 MB) — no need to read beyond that
- Small files: use `inputSize` when it's less than the sample limit
- Trailing newline check: returns 0 if the sample doesn't contain a newline (unless it's a small file read to EOF)
- `getExternalStats` now passes `totalLoadFileSize(fileSize)` instead of hardcoded 0

Tests cover: bounded first-line (>1 MB file), small file without trailing newline, external stats with known file size, CRLF handling, ignored-lines with bounded sample. LGTM.

XuPeng-SH (approved):
这版我看下来可以过。

核心修复点很聚焦：把 first-line sampler 从“读到 EOF 再关”改成了有上界的 sample read，同时把“大文件采样不到完整首行就回退 schema row size”和“小文件无 trailing newline 仍可当完整首行”这两个边界补齐了。`external.go` 也跟着接到了同一套 bounded helper，上轮的 planner/cardinality 修复没有被打回去。

Issue #24731, opened 2026-05-31
[Bug]: fulltext MATCH() AGAINST() cannot be combined with a JOIN (still repro on v3.0.11 & 4.0-dev)


dengn:
### Summary
A single-table `MATCH() AGAINST()` works (including `IN BOOLEAN MODE` and relevance-score ordering), but combining `MATCH() AGAINST()` with a **JOIN** fails.

Related: #20687 (open, severity/s0) tracks the same error message. This adds a minimal JOIN repro and confirms it still reproduces on prod **v3.0.11** and the **4.0-dev** line (`nightly-0bfda604`).

### Steps to Reproduce
```sql
CREATE TABLE ft (id VARCHAR(191) PRIMARY KEY, base_id VARCHAR(191), title VARCHAR(512), body LONGTEXT,
                 FULLTEXT INDEX ft_idx(title, body));
CREATE TABLE base (id VARCHAR(191) PRIMARY KEY, name VARCHAR(191));

-- single table: OK
SELECT id FROM ft WHERE MATCH(title, body) AGAINST('hello');

-- with JOIN: fails
SELECT f.id, b.name FROM ft f JOIN base b ON b.id = f.base_id
WHERE MATCH(f.title, f.body) AGAINST('hello');
-- ERROR 20105 (HY000): not supported: MATCH() AGAINST() function cannot be replaced by
-- FULLTEXT INDEX and full table scan with fulltext search is not supported yet.
```

### Prisma impact
Low for our app (we use `relationMode="prisma"`, which issues separate single-table queries). But it blocks any joined fulltext query and raw SQL that filters one table by fulltext while joining another.

### Expected Behavior
MySQL allows `MATCH ... AGAINST` together with a JOIN.

---
**Test context** — Found while testing **Prisma ORM (v5.22)** compatibility against MatrixOne. Reproduced on **both**:
- Prod cloud — `8.0.30-MatrixOne-v3.0.11`
- 4.0 dev line — `nightly-0bfda604` (`select version()` → `8.0.30-MatrixOne-v`)

(Results on the older v3.0.5 are intentionally omitted.)

Ariznawlll:
Verified on latest `main` and latest `4.0-dev`.

Branches / commits:
- `main`: `781161fab`
- `4.0-dev`: `da7ad6be7`

Validation SQL covered the original scenario:
- `ft` table with `FULLTEXT INDEX ft_idx(title, body)`
[...]

PR #24880, opened 2026-06-08
fix: support fulltext index rewrite for joins

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24731 

fix: support fulltext index rewrite for joins

### DIFF
--- a/pkg/sql/plan/apply_indices.go
+++ b/pkg/sql/plan/apply_indices.go
@@ -438,7 +438,7 @@ func (builder *QueryBuilder) applyIndices(nodeID int32, colRefCnt map[[2]int32]i
 		return builder.applyIndicesForFilters(nodeID, node, colRefCnt, idxColMap), nil
 
 	case plan.Node_JOIN:
-		return builder.applyIndicesForJoins(nodeID, node, colRefCnt, idxColMap), nil
+		return builder.applyIndicesForJoins(nodeID, node, colRefCnt, idxColMap)
 
 	case plan.Node_PROJECT:
 		//NOTE: This is the entry point for vector index rule on SORT NODE.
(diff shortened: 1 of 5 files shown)


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (changes requested):
I found one blocking unhappy-path coverage gap before this can be approved.

- The new join rewrite is intentionally limited to INNER JOIN in `applyFullTextFiltersForJoinChildren` (`pkg/sql/plan/apply_indices_fulltext.go`), but all new unit/BVT coverage is also INNER-only (`pkg/sql/plan/apply_indices_test.go`, `test/distributed/cases/fulltext/fulltext_join.sql`). Please add negative coverage for at least one `LEFT JOIN` and one `RIGHT JOIN` query with `MATCH(...) AGAINST(...)` to pin the intended no-rewrite / stable behavior.

Outer joins are null-preserving, so an accidental future rewrite there would be a semantic regression and we currently would not catch it.

iamlinjunhong:
> I found one blocking unhappy-path coverage gap before this can be approved.
> 
> * The new join rewrite is intentionally limited to INNER JOIN in `applyFullTextFiltersForJoinChildren` (`pkg/sql/plan/apply_indices_fulltext.go`), but all new unit/BVT coverage is also INNER-only (`pkg/sql/plan/apply_indices_test.go`, `test/distributed/cases/fulltext/fulltext_join.sql`). Please add negative coverage for at least one `LEFT JOIN` and one `RIGHT JOIN` query with `MATCH(...) AGAINST(...)` to pin the intended no-rewrite / stable behavior.
> 
> Outer joins are null-preserving, so an accidental future rewrite there would be a semantic regression and we currently would not catch it.

fixed

XuPeng-SH (approved):
The previously requested unhappy-path coverage for non-INNER joins is now in place.

- Unit coverage now includes `TestFullTextJoinRewriteSkipsOuterJoins` for LEFT/RIGHT join behavior.
- BVT coverage now includes LEFT JOIN and RIGHT JOIN SQL cases asserting the current unsupported/no-rewrite behavior.

Thanks for addressing this.

Issue #24882, opened 2026-06-08
[Bug]: ivfpq/cagra concurrent build crashed and cannot test sharded/replicated mode in single gpu


cpegeric:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

8ad7a3f54bbf2bfbcc5a55cdac23b2a3ae9f635a

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

ivfpq/cagra build failure when multiple index creation concurrently.  Also, there is no way to test sharded/replicated mode in single gpu.

### Steps to Reproduce

```Markdown
create index concurrently.
```

cpegeric:
https://github.com/matrixorigin/matrixone/actions/runs/29246128978?pr=25095 already fixed by this PR

cpegeric:
please wait for the PR and re-test again.

PR #24883, opened 2026-06-08
Fix: ivfpq/cagra concurrent build/extend crashed.

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24882

1. bug fix: add mutex to avoid concurrent build/extend in same GPU (per device mutex) because gpu kmeans use the global environment which only single index creation is supported at one time.

2. add gpu_multi_simulation session var to support sharded/replicated simultion on single gpu environment.
3. add bvt tests for simuliated sharded/replicated tests.
4. remove omp (multi-thread) for bitmap computation that lead to too much context switching for 100 threads tests.


heni02 (approved):

fengttt (approved):

aunjgr (approved):
Massive refactor but well-reasoned fixes. Three root causes of concurrent GPU crashes, each fixed:

### C++ thread safety (`cgo/cuvs/`)

**1. `device_build_mutex` serializes build/extend** (`cagra.hpp`, `ivf_flat.hpp`, `ivf_pq.hpp`). All `cuvs::neighbors::*::build()` and `::extend()` calls now wrapped in `std::lock_guard<std::mutex> build_lk(matrixone::device_build_mutex(handle.get_device_id()))`. Previously concurrent builds on the same GPU raced, corrupting the index.

**2. `replicated_indices_` keyed by `rank` instead of `device_id`** (`cagra.hpp`). Two worker ranks sharing the same GPU would overwrite each other's entries in the map. Now distinct per-rank.

**3. GPU memory freed BEFORE `worker->stop()`** (`cagra.hpp::destroy()`). Previously index/dataset/quantizer were freed after worker stop. But the RMM pool was tagged with the worker's CUDA stream — freeing after stream destruction left poisoned free blocks and `cudaErrorInvalidResourceHandle` aborts. Now freed in reverse order.

**4. Pool no longer installed as global default** (`cuvs_worker.hpp`). `set_per_device_resource` removed. The stream-ordered pool is now reachable only via `worker_pool_mr()` and used only for search workspace uvectors (which live on the worker's stable stream). Everything else (index body, pairwise scratch) uses plain `cudaMalloc`/`cudaFree` via `raw_device_mr()`. Prevents dying streams from poisoning unrelated allocations.

### Go refactoring

**5. Plugin framework** (`pkg/indexplugin/`, `pkg/fulltext/plugin/`). Code previously hardcoded in `build_ddl.go` (-1515) and `ddl_index_algo.go` (-718) is now in a plugin registry. Each index type registers compile/catalog/plan/idxcron hooks. Same logic, better separation. `plugin_builder.go` and `plugin_context.go` wire the plugin into query builder and compile pipeline.

**6. `pkg/sql/plan/ivfflat.go` deleted** (-140). `apply_indices.go` was the sole caller — now routed through the plugin. The `ivfflat.go` path was dead/duplicate code.

LGTM. Impressive fix — the CUDA stream/pool analysis in the C++ comments is unusually clear for GPU debugging.

PR #24886, opened 2026-06-08
fix: support fulltext index rewrite for joins

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24731 

fix: support fulltext index rewrite for joins


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (approved):
I checked the latest backport and did not find a blocking issue on the unhappy-path behavior. The outer-join no-rewrite boundary is covered, and the added join fulltext rewrite path looks consistent with the 4.0 change.

There are a couple of possible future hardening tests (for example BOOLEAN MODE through the join rewrite path), but I do not see anything here that should block approval.

PR #24887, opened 2026-06-08
add nameconst

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/24486

增加name_const函数


aunjgr (approved):
MySQL `NAME_CONST(name, value)` — internal function used by replication. Returns `value` with column heading `name`.

Implementation:
- `foldNameConstArgs` validates first arg is a non-null literal (column name), constant-folds the second arg (value)
- `validNameConstValueExpr` allows literals, decimal-literal casts, and `-literal` forms
- Binder integration at `bindFuncExprAndConstFold` case `"name_const"` — constant-folds before proceeding
- BVT covers single/multi-row, NULL args, JOIN, subquery, negative values, and error cases

Clean, matches MySQL 8.0 semantics. LGTM.

heni02 (approved):

ouyuanning (approved):

daviszhen:
@mergify refresh

XuPeng-SH (changes requested):
I found two blocking issues in the NAME_CONST implementation.

1. `pkg/sql/plan/base_binder.go:1449-1567` routes `name_const` to `goto between_fallback` when `proc == nil`. That fallback is BETWEEN-specific and unconditionally reads `fnArgs[2]`, but NAME_CONST only has two arguments, so any nil-proc binder/internal path will panic with an index-out-of-range.

2. `pkg/sql/plan/query_builder.go:4482-4492` only extracts the heading when the first argument is a bare `*tree.NumVal`, while `pkg/sql/plan/base_binder.go:2978-2991` explicitly accepts parenthesized names by stripping `*tree.ParenExpr`. So `select name_const(('myname'), (14))` is accepted, but the column heading falls back to `name_const(...)` instead of `myname`.

daviszhen:
> I found two blocking issues in the NAME_CONST implementation.
> 
> 1. `pkg/sql/plan/base_binder.go:1449-1567` routes `name_const` to `goto between_fallback` when `proc == nil`. That fallback is BETWEEN-specific and unconditionally reads `fnArgs[2]`, but NAME_CONST only has two arguments, so any nil-proc binder/internal path will panic with an index-out-of-range.
> 2. `pkg/sql/plan/query_builder.go:4482-4492` only extracts the heading when the first argument is a bare `*tree.NumVal`, while `pkg/sql/plan/base_binder.go:2978-2991` explicitly accepts parenthesized names by stripping `*tree.ParenExpr`. So `select name_const(('myname'), (14))` is accepted, but the column heading falls back to `name_const(...)` instead of `myname`.

已经修改

XuPeng-SH (changes requested):
I don't think this is fix-complete yet. The new `NAME_CONST` validator is hard-coded to a very narrow set of value shapes: bare literals, the special decimal-literal cast case, and unary minus wrapped around those (`pkg/sql/plan/base_binder.go:2961-2975`). That means it will reject charset/collation-decorated constants, even though those are one of the important real-world `NAME_CONST` surfaces from MySQL's stored-program / binlog path, for example `NAME_CONST('s1', _utf8mb3'test' COLLATE utf8mb3_unicode_ci)`. MatrixOne already supports `_utf8mb4'...' COLLATE ...` literals elsewhere (`test/distributed/cases/ddl/mysql_ddl_1.sql:85-90`), but this PR's validator still blocks that form purely because it is not a raw literal node.

That matters here because `NAME_CONST` is mainly a compatibility function, not a new end-user feature. If we only accept the small hand-written cases covered by the new BVT, but still reject charset/collation-decorated constants that show up in the real MySQL/internal use cases, we haven't actually closed the compatibility gap. I'd like to see that form supported (or at least explicitly justified if MO intentionally wants to diverge), and covered with an execution-level case before approval.

XuPeng-SH:
A few worthwhile coverage gaps remain even after the main correctness fix:

1. **Derived-table / execution-level aliasing**: please add an end-to-end case like `SELECT t.myname FROM (SELECT name_const('myname', 42)) t;` (or `SELECT * FROM (SELECT name_const('v', 'hello'), name_const('n', 99)) sub;`). This exercises the real query-builder / heading propagation path beyond isolated unit tests.

2. **Stored procedure path**: since `NAME_CONST` is mainly a MySQL-compat/internal function, I'd like one realistic proc path such as `CREATE PROCEDURE p() BEGIN SELECT name_const('x', 1); END; CALL p();`.

3. **Scientific-notation float literals**: the BVT covers decimal literals, but not float/scientific notation (`1.5e0`, `-1.5e0`). That's a useful literal branch to pin down.

daviszhen:
> I don't think this is fix-complete yet. The new `NAME_CONST` validator is hard-coded to a very narrow set of value shapes: bare literals, the special decimal-literal cast case, and unary minus wrapped around those (`pkg/sql/plan/base_binder.go:2961-2975`). That means it will reject charset/collation-decorated constants, even though those are one of the important real-world `NAME_CONST` surfaces from MySQL's stored-program / binlog path, for example `NAME_CONST('s1', _utf8mb3'test' COLLATE utf8mb3_unicode_ci)`. MatrixOne already supports `_utf8mb4'...' COLLATE ...` literals elsewhere (`test/distributed/cases/ddl/mysql_ddl_1.sql:85-90`), but this PR's validator still blocks that form purely because it is not a raw literal node.
> 
> That matters here because `NAME_CONST` is mainly a compatibility function, not a new end-user feature. If we only accept the small hand-written cases covered by the new BVT, but still reject charset/collation-decorated constants that show up in the real MySQL/internal use cases, we haven't actually closed the compatibility gap. I'd like to see that form supported (or at least explicitly justified if MO intentionally wants to diverge), and covered with an execution-level case before approval.

补充了相关的case.

daviszhen:
> A few worthwhile coverage gaps remain even after the main correctness fix:
> 
> 1. **Derived-table / execution-level aliasing**: please add an end-to-end case like `SELECT t.myname FROM (SELECT name_const('myname', 42)) t;` (or `SELECT * FROM (SELECT name_const('v', 'hello'), name_const('n', 99)) sub;`). This exercises the real query-builder / heading propagation path beyond isolated unit tests.
> 2. **Stored procedure path**: since `NAME_CONST` is mainly a MySQL-compat/internal function, I'd like one realistic proc path such as `CREATE PROCEDURE p() BEGIN SELECT name_const('x', 1); END; CALL p();`.
> 3. **Scientific-notation float literals**: the BVT covers decimal literals, but not float/scientific notation (`1.5e0`, `-1.5e0`). That's a useful literal branch to pin down.

1 和 3 补充了相关的case.  matrixone 不支持存储过程, 所以没有 加相关的case

XuPeng-SH (changes requested):
I rechecked the latest head (`c7f0a5172e1c5408a298e38b8c521685fecd7127`). The earlier nil-proc panic and parenthesized-heading issues look fixed, and the requested charset/collation/derived-table/scientific-notation cases were added. I still see two correctness gaps.

1. The `NAME_CONST` argument validation is only enforced on the normal `QueryBuilder` path, so builder-less binders can still build invalid `NAME_CONST` expressions. `bindFuncExprImplByAstExpr` always validates only the first AST argument (`pkg/sql/plan/base_binder.go:1318`), but the second-argument validation/folding lives under `bindFuncExprAndConstFold` (`pkg/sql/plan/base_binder.go:1449`, `pkg/sql/plan/base_binder.go:2939`). When `b.builder == nil`, the code calls `BindFuncExprImplByPlanExpr` directly (`pkg/sql/plan/base_binder.go:1330`) and never runs `foldNameConstArgs`. That makes this reachable from generated-column binding: `NewGeneratedColBinder` has no builder (`pkg/sql/plan/generated_col_binder.go:34`), `buildGeneratedExpr` uses it (`pkg/sql/plan/build_util.go:486`), and the later checks only reject volatile functions / illegal references (`pkg/sql/plan/build_util.go:492`). As a result, a schema expression such as `b int generated always as (name_const('x', a)) stored` can be accepted even though the same non-constant value argument is intentionally rejected for `SELECT name_const('x', a)` and MySQL documents both arguments as constants. Please move/enforce the NAME_CONST validation somewhere shared by all binder paths, or explicitly reject it from generated/default/internal expression binders.

2. `+literal` constants are rejected even though negative literals are explicitly supported. The parser binds `+1` through `UNARY_PLUS` (`pkg/sql/parsers/dialect/mysql/mysql_sql.go:26160`) and the generic const-fold switch already treats `unary_plus` as foldable (`pkg/sql/plan/base_binder.go:1441`), but `validNameConstValueExpr` only accepts raw literals, decimal literal casts, and `unary_minus` wrappers (`pkg/sql/plan/base_binder.go:2961`). This means `select name_const('x', -1)` works while `select name_const('x', +1)` is rejected, despite both being signed constant literals. Please accept `unary_plus` as well and add a BVT/unit case.

Local verification: `go test ./pkg/sql/parsers -run TestDoesNotExist` passes. Focused `go test ./pkg/sql/plan -run 'TestBindNameConst|TestAppendSelectListNameConstHeading'` and `go test ./pkg/sql/plan/function -run 'TestBuiltInNameConst|TestFunctionId'` both fail during build on local `github.com/unum-cloud/usearch/golang` cgo symbols before the tests run, so I could not get a local assertion signal for those packages.

daviszhen:
> I rechecked the latest head (`c7f0a5172e1c5408a298e38b8c521685fecd7127`). The earlier nil-proc panic and parenthesized-heading issues look fixed, and the requested charset/collation/derived-table/scientific-notation cases were added. I still see two correctness gaps.
> 
> 1. The `NAME_CONST` argument validation is only enforced on the normal `QueryBuilder` path, so builder-less binders can still build invalid `NAME_CONST` expressions. `bindFuncExprImplByAstExpr` always validates only the first AST argument (`pkg/sql/plan/base_binder.go:1318`), but the second-argument validation/folding lives under `bindFuncExprAndConstFold` (`pkg/sql/plan/base_binder.go:1449`, `pkg/sql/plan/base_binder.go:2939`). When `b.builder == nil`, the code calls `BindFuncExprImplByPlanExpr` directly (`pkg/sql/plan/base_binder.go:1330`) and never runs `foldNameConstArgs`. That makes this reachable from generated-column binding: `NewGeneratedColBinder` has no builder (`pkg/sql/plan/generated_col_binder.go:34`), `buildGeneratedExpr` uses it (`pkg/sql/plan/build_util.go:486`), and the later checks only reject volatile functions / illegal references (`pkg/sql/plan/build_util.go:492`). As a result, a schema expression such as `b int generated always as (name_const('x', a)) stored` can be accepted even though the same non-constant value argument is intentionally rejected for `SELECT name_const('x', a)` and MySQL documents both arguments as constants. Please move/enforce the NAME_CONST validation somewhere shared by all binder paths, or explicitly reject it from generated/default/internal expression binders.
> 2. `+literal` constants are rejected even though negative literals are explicitly supported. The parser binds `+1` through `UNARY_PLUS` (`pkg/sql/parsers/dialect/mysql/mysql_sql.go:26160`) and the generic const-fold switch already treats `unary_plus` as foldable (`pkg/sql/plan/base_binder.go:1441`), but `validNameConstValueExpr` only accepts raw literals, decimal literal casts, and `unary_minus` wrappers (`pkg/sql/plan/base_binder.go:2961`). This means `select name_const('x', -1)` works while `select name_const('x', +1)` is rejected, despite both being signed constant literals. Please accept `unary_plus` as well and add a BVT/unit case.
> 
> Local verification: `go test ./pkg/sql/parsers -run TestDoesNotExist` passes. Focused `go test ./pkg/sql/plan -run 'TestBindNameConst|TestAppendSelectListNameConstHeading'` and `go test ./pkg/sql/plan/function -run 'TestBuiltInNameConst|TestFunctionId'` both fail during build on local `github.com/unum-cloud/usearch/golang` cgo symbols before the tests run, so I could not get a local assertion signal for those packages.

1, 已经修改.
2, 本来就支持. 依然补充了判断逻辑.

XuPeng-SH (changes requested):
Requesting changes for one semantic hole in `NAME_CONST` validation.

In `pkg/sql/plan/base_binder.go`, the new `isDecimalLiteralCast` path makes sense for decimal literals because they are represented as `cast(string_literal, decimal)`. But it also accepts user-written explicit casts from string literals, so a query like:

```sql
select name_const('x', cast('12.34' as decimal(10,2)));
```

will pass even though this PR's intended semantics say `NAME_CONST` should reject `CAST(...)` in the value argument.

Please distinguish true decimal literals from explicit `CAST(...)` expressions before approval. Validating the AST form before binding/folding, or preserving literal-origin metadata, would both work.

fengttt (approved):

XuPeng-SH (approved):
Reviewed with extra unhappy-path focus.

I did not find a new blocking issue in this `NAME_CONST` implementation. The main concern I checked was whether `CAST(...)` could still sneak through the literal-only gate, but `name_const` arguments are rejected at the AST level before binding/folding, so user-written cast/function/column expressions are still blocked as intended.

The added unit tests and BVT coverage cover the main compatibility and validation paths, including invalid argument shapes and heading propagation scenarios.

Issue #23913, opened 2026-03-20
[Feature Request]: Support MySQL Spatial Data Types (GEOMETRY, POINT, LINESTRING, POLYGON, etc.)


jiangxinmeng1:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

-- ============================================================
-- POINT
-- ============================================================

CREATE TABLE test_point (
    id INT PRIMARY KEY AUTO_INCREMENT,
    location POINT
);

INSERT INTO test_point(location) VALUES (ST_GeomFromText('POINT(116.397428 39.90923)'));
INSERT INTO test_point(location) VALUES (ST_GeomFromText('POINT(121.473701 31.230416)'));

SELECT id, ST_AsText(location) FROM test_point;
SELECT ST_Distance(
    (SELECT location FROM test_point WHERE id = 1),
    (SELECT location FROM test_point WHERE id = 2)
);
SELECT ST_X(location), ST_Y(location) FROM test_point WHERE id = 1;

DROP TABLE test_point;

-- ============================================================
-- GEOMETRY（通用类型，可存任意几何对象）
-- ============================================================

CREATE TABLE test_geometry (
    id INT PRIMARY KEY AUTO_INCREMENT,
    geom GEOMETRY
);

INSERT INTO test_geometry(geom) VALUES (ST_GeomFromText('POINT(1 2)'));
INSERT INTO test_geometry(geom) VALUES (ST_GeomFromText('LINESTRING(0 0, 3 4)'));
INSERT INTO test_geometry(geom) VALUES (ST_GeomFromText('POLYGON((0 0,1 0,1 1,0 1,0 0))'));

-- POINT 列只能存 POINT，GEOMETRY 列可以混存
SELECT id, ST_GeometryType(geom), ST_AsText(geom) FROM test_geometry;
SELECT id, ST_SRID(geom) FROM test_geometry;

DROP TABLE test_geometry;

-- ============================================================
-- LINESTRING
-- ============================================================

CREATE TABLE test_linestring (
    id INT PRIMARY KEY AUTO_INCREMENT,
    path LINESTRING
);

INSERT INTO test_linestring(path) VALUES (ST_GeomFromText('LINESTRING(0 0, 1 1, 2 2)'));
[...]

PR #24777, opened 2026-06-02
Gis geometry engine

fengttt:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

## Which issue(s) this PR fixes:

issue #23913 

## What this PR does / why we need it:

1. Implement geometry type using WKB format.   Original storage used WKT, which is not good.
2. For SRID 4326, used google S2 to compute geodetic length and area.
3. Added functions like convex_hull.
4. Added a type GEOMETRY32, which use float32 instead of float64.

TODO: Add special fixed length types like POINT2D, POINT3D, POINT2D32 etc, that is more efficient for storing points.
TODO: Use ellipsoid for SRID 4326.   S2 is spherical.

PR #24890, opened 2026-06-09
Gis geometry engine (#24777)

fengttt:
- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

issue #23913

1. Implement geometry type using WKB format. Original storage used WKT, which is not good.
2. For SRID 4326, used google S2 to compute geodetic length and area.
3. Added functions like convex_hull.
4. Added a type GEOMETRY32, which use float32 instead of float64.

TODO: Add special fixed length types like POINT2D, POINT3D, POINT2D32 etc, that is more efficient for storing points.
TODO: Use ellipsoid for SRID 4326.   S2 is spherical.

---------

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23913

Cherry pick into 4.0 dev, to avoid unnecessary upgrade later.


fengttt:
Merged, by accident.   But overall seems OK ...   Close as done.

PR #24891, opened 2026-06-09
fix(plan): support mixed-scale decimal between via decimal256 promotion

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24565

This cherry-picks the `BETWEEN` decimal256 fix to `4.0-dev` (companion of #24841 on `3.0-dev` and #24842 on `main`).

`4.0-dev` binds decimal256 comparisons natively, but the `BETWEEN` checkFn still rejected decimal operands whose scales differ — it required the value scale to equal each bound scale:

```sql
SELECT CAST(1 AS DECIMAL(38,0))
     BETWEEN CAST(0.5 AS DECIMAL(38,30)) AND CAST(2 AS DECIMAL(38,30));
-- before: rejected (parameters wrong); after: 1
```

The checkFn now aligns the three operands to one common decimal type that preserves the max integral width + max scale, promoting to decimal256 when decimal128 would overflow (e.g. `DECIMAL(38,0) BETWEEN DECIMAL(38,30) AND DECIMAL(38,30)` -> `DECIMAL256(68,30)`), instead of rejecting mixed-scale / mixed-family decimals. `betweenImpl` already has a decimal256 branch, so only the type derivation changes.

Added coverage:
- `func_between_decimal256_test.go` (promotion, mixed family, integer bounds, non-decimal unchanged, helpers, runtime decimal256 path)
- append-only BVT case in `case_when.sql`

Validation:
- `go test ./pkg/sql/plan/function`, `go vet`, `gofmt`, `make static-check` (0 issues)
- mo-tester `case_when.sql` passed 94/94 locally


ck89119:
Closing: 4.0-dev already handles mixed-scale decimal BETWEEN correctly via the planner's fallback expansion (`x >= lo AND x <= hi`) plus native decimal256 comparison support. The decimal256 BETWEEN scenarios from issue #24565 work out of the box (verified: `CAST(1 AS DECIMAL(38,0)) BETWEEN CAST(0.5 AS DECIMAL(38,30)) AND CAST(2 AS DECIMAL(38,30))` returns 1). The checkFn change here was redundant and additionally broke the secondary-index optimization (prefix_between) on `dtype/decimal.test`. The change is only needed on 3.0-dev (#24841), which lacks native decimal256 comparison.

PR #24665, opened 2026-05-28
add rss pressure cache controller

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

Fixes #24664

## What this PR does / why we need it:

This PR adds an RSS pressure controller for memory cache pressure handling.

Before this change, RSS cache pressure used a fixed admission-target TTL as the main lifecycle mechanism. If RSS recovered quickly, the memory cache admission target could still remain active until the TTL expired, slowing data cache recovery.

This PR changes that behavior by:

- adding explicit RSS pressure states: `none`, `soft`, and `hard`;
- using hysteresis for state transitions:
  - `none -> soft`: RSS >= 85%;
  - `soft -> hard`: RSS >= 92%;
  - `hard -> soft`: RSS <= 88%;
  - `soft/hard -> none`: RSS <= 80%;
- setting memory cache admission target by state:
  - soft: 80%;
  - hard: 50%;
- clearing the memory cache admission target immediately when pressure exits;
- keeping evict and `debug.FreeOSMemory` actions throttled by cooldown;
- keeping the target TTL as a lease fallback that is renewed while pressure remains active;
- splitting target set/clear from cache eviction so asynchronous evict actions cannot restore stale pressure targets;
- guarding async evict with a pressure generation check.

The change keeps Step 1/2 behavior intact: RSS pressure still only controls data/memory cache and does not force objectio meta cache eviction.

PR #24892, opened 2026-06-09
fix(rscthrottler): avoid double-counting RSS and reserved in S3 write throttler

### LINKED ISSUE #24881: [Bug]: Failed to load vector data during regression testing, but it works when isolation

### Branch Name

main,4.0-dev

### Commit ID

lastest

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

4.0：
Image
当时内存使用情况：
Image

main：

3.0：
Image

### Expected Behavior

_No response_

### Steps to Reproduce

```Markdown
```

### Additional information

_No response_

### LINKED ISSUE #24888: [Bug] [4.0-dev tke regression] CN OOM during IVF vector test data copy.
## Summary

In mo-nightly-regression Branch Nightly Regression Test, the IVF VECTOR INDEX TEST failed during test data preparation. The MySQL client reported `ERROR 2013 Lost connection to MySQL server during query`, and Grafana/Prometheus shows the target CN was OOMKilled.

## Failed run

- Actions run: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27149474770/job/80219844165
- MO branch: `4.0-dev`
- MO commit from job: `bb975e8194a9a4a71599f49ddb2df247eb18c834`
- Namespace: `mo-branch-commit-7a8e7cd84-20260608`
- Failed step: `IVF VECTOR INDEX TEST / Test Data Preparation`

## Failure point

The initial load succeeded, then the failure happened while copying the vector table:

```sql
insert into anli_test.ca_comprehensive_dataset_copy
select * from anli_test.ca_comprehensive_dataset;
```

Actions log:

```text
2026-06-08 23:48:49 GMT load data into ca_comprehensive_dataset_copy
2026-06-08 23:48:58 GMT ERROR 2013 (HY000) at line 1: Lost connection to MySQL server during query
```

## OOM evidence

Grafana/Prometheus showed the following after the failure:

```text
pod: nightly-regression-dis-tp-cn-phqxp
pod IP: 10.143.26.143
restart: 1
terminated reason: OOMKilled
exit code: 137
```

The restart was visible shortly after the client-side failure:

```text
23:48:58 UTC  Actions reports ERROR 2013
23:49:30 UTC  kube_pod_container_status_restarts_total shows tp-cn-phqxp restart=1
23:49:30 UTC  kube_pod_container_status_last_terminated_reason shows OOMKilled
```

Memory before the OOM was already high:

```text
nightly-regression-dis-tp-cn-phqxp: ~49.6Gi / 55Gi
nightly-regression-dis-tp-cn-p7z4v: ~49.2Gi / 55Gi
nightly-regression-dis-tp-cn-qrb2w: ~39.2Gi / 55Gi
```

## Related logs

Around the same timestamp, proxy/CN logs showed the query backend on the OOMKilled CN became unavailable:

```text
2026-06-08 23:48:58.414 UTC
remote="10.143.26.143:6004"
error="dial tcp4 10.143.26.143:6004: connect: connection refused"
msg="fail to reset backend connection"
```

The target CN was:

```text
10.143.26.143 = nightly-regression-dis-tp-cn-phqxp
```

Before the connection loss, the insert was upgraded to a table lock:

```text
Row lock upgraded to table lock for SQL:
insert into anli_test.ca_comprehensive_dataset_copy select * from anli_test.ca_comprehensive_dataset
```

There were also many object cache errors near the failure window:

```text
internal error: dedup load did not complete
```

## Expected behavior

Copying the vector table for IVF test data preparation should not OOM a CN. If the operation needs too much memory, it should either spill/throttle/fail gracefully without killing the CN, or return a controlled error to the client.

## Actual behavior

One CN was OOMKilled during the large `insert select`, the query backend became unavailable, and the client received `ERROR 2013 Lost connection`.

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24881
Fixes #24888

The S3 write throttler's `acquireWithinRSSLimit` had two gates:

1. **Pool check** via `Available()` — double-counts S3 write buffers (in RSS as mmap'd mpool allocations + in reserved as throttler tokens). Under `ForceRefresh()` this returned 0 available, permanently rejecting all writes (#24881).
2. **RSS gate** `rss + reserved + ask > total × limitRate` — same double-counting. Even with stale RSS the sum `rss + reserved` pushes over the cap, causing false rejects. And the RSS gate's original purpose (preventing OOM from non-S3 memory pressure, PR #24268) is now handled by RSS scavenging (85%/92% thresholds with graduated cache eviction + FreeOSMemory).

**Fix**: 

- **Pool check**: use `limit - reserved` directly instead of `Available()`. The pool only needs to bound forward-looking reservations against the throttler limit.
- **RSS gate**: removed. Physical memory is protected by:
  - `pinnedRate >= 0.80` hard ceiling at `AcquirePolicyForCNFlushS3`
  - RSS scavenging (85%/92%) with cache eviction + FreeOSMemory

`Available()` is unchanged — other throttler consumers (workspace, merge) rely on its `rss + reserved` semantics.

## Checklist:

- [x] Pool gate: `limit - reserved` avoids S3 double-count, scoped to `acquireWithinRSSLimit` only
- [x] RSS gate removed: superseded by RSS scavenging and pinnedRate ceiling
- [x] Tests: 8 cases covering pool overflow, large ask, normal operation, high RSS with pool headroom, reserved > rss
- [x] 82.7% package coverage

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


XuPeng-SH (changes requested):
I found two blocking issues here.

1. The new `Available()` / RSS-gate logic changes shared `memThrottler` semantics, not just the CN S3 write throttler path. Using `max(rss, reserved)` globally is only valid when `reserved` overlaps RSS, but that is not true for other users such as the workspace throttler that still uses the default acquire policy. In those cases, `reserved` is real already-granted quota, so the new `Available()` can over-admit memory under high RSS by ignoring part of the reserved headroom. This needs to stay scoped to the S3-specific policy/path (or become an explicit opt-in mode), rather than changing the common calculation for every `memThrottler`.

2. The new tests only cover cases where `rss > reserved`, so they do not actually pin the new `max(rss, reserved)` behavior. A regression to plain `rss` would still pass. Please add `reserved > rss` cases for both allow and deny paths, and assert `Available()` there as well.

Because the first issue is a real behavior regression outside the intended S3 scope, I am requesting changes.

aunjgr:
Thanks for the review @XuPeng-SH. Both issues fixed:

**1. `Available()` reverted.** The `max(rss, reserved)` change was removed from `Available()`. The fix is now self-contained in `acquireWithinRSSLimit` — it inlines its own pool check with `max(rss, reserved)` semantics, leaving the shared `Available()` unchanged for workspace/merge consumers.

**2. `reserved > rss` tests added.** Two new cases:
- `allow when reserved exceeds rss`: `rss=30, reserved=40, ask=10` → `max(30,40)+10=50 ≤ 90` — allowed.
- `deny when reserved exceeds rss and ask pushes max over limit`: `rss=30, reserved=60, ask=31` → `max(30,60)+31=91 > 90` — denied.

These would pass if the gate reverted to plain `rss` alone (since `rss=30` is well under 90), pinning the `max` semantics.

XuPeng-SH (changes requested):
I understand the double-counting problem this PR is trying to fix, but I found a substantive issue in the new `max(rss, reserved)` accounting.

The key assumption here is not actually safe: `reserved` is a **forward-looking reservation**, while `rss` is the current whole-process RSS. In the CN S3 write path, callers reserve memory first and only materialize the writer/buffer afterwards. That means `reserved` can legitimately be larger than the current S3-related RSS while `rss` still contains unrelated process memory (lockservice, other operators, tests, etc.).

In that state, `max(rss, reserved)` drops the unrelated RSS entirely. For example, if current RSS is 50 = 30 other + 20 current S3, and `reserved=60` because the S3 path has reserved ahead of allocation, the new logic treats effective usage as 60 instead of the real 30 other + 60 reserved = 90. A further acquire can then be admitted even though the process will overshoot the intended limit once those reserved bytes are actually materialized.

The new tests even lock in this assumption via the `allow when reserved exceeds rss` case. I think the fix needs to avoid double-counting only the overlapping S3 portion, not replace `rss + reserved` with a plain `max(rss, reserved)` over the whole process. Otherwise this becomes an under-count bug instead of the original over-count bug.

aunjgr:
Updated. Replaced `max(rss, reserved)` with two independent gates:

- **Pool gate**: `reserved + ask > limit` — forward-looking, prevents over-promise on S3 write bandwidth. Uses reserved alone (no RSS involvement — eliminates double-count).
- **RSS gate**: `rss > total × limitRate` — backward-looking, prevents physical OOM when the container is already full. Uses RSS alone (catches non-S3 memory that `max` would have dropped).

The old `rss + reserved` sum conflated two independent dimensions. Separating them avoids both over-count (the original bug) and under-count (the `max` approach's flaw with non-S3 RSS mixed in).

`acquireWithinRSSLimit` at 93% coverage. `AcquirePolicyForCNFlushS3` at 75%.

XuPeng-SH (changes requested):
I re-reviewed the latest revision and there is still a blocking issue.

The earlier shared-`Available()` regression is fixed, but the new two-gate logic in `acquireWithinRSSLimit` still under-counts when unrelated RSS already exists. Right now the code checks:
- pool gate: `reserved + ask <= limit`
- RSS gate: `rss <= total * limitRate`

These are treated as independent, but in the CN S3 path `Acquire()` reserves memory before the writer necessarily materializes it. So `reserved` is still forward-looking, while `rss` already includes unrelated process memory.

A concrete example with the current logic:
- `total=100`, `limitRate=0.9`, so cap is `90`
- current `rss=50` (for example `30` unrelated memory + `20` current S3)
- `reserved=60`
- `ask=30`

The new code allows this (`60+30 <= 90`, `50 <= 90`) and raises `reserved` to `90`, but once that reservation is consumed the process can reach `30 unrelated + 90 S3 = 120`, well past the intended cap.

This is not just theoretical: the CN S3 throttler is wired from `pkg/cnservice/server.go`, and callers in `pkg/sql/colexec/insert/insert.go` and `pkg/sql/colexec/multi_update/s3writer_delegate.go` acquire reservation first and only flush/materialize later.

So the fix still needs a combined bound that accounts for non-overlapping RSS plus forward reservations, or an explicit way to track only the overlapping S3 portion instead of treating RSS and reservations as fully independent.

Separately, the new tests still do not pin the key old-deny/new-allow boundary of the changed logic. For example, there is no focused case like `rss=89, reserved=0, ask=2, total=100, limitRate=0.90, limit=90`, which is the exact class of request this PR is meant to unblock.

aunjgr:
Traced the original bug scenario through both gates. The pool check (`Available()`) was the only gate that changed behavior — the RSS gate rejects either way:

`ForceRefresh()` → `rss=13.5, reserved=7.82, ask=183KB, limit=12.6`:

| Gate | Before | After |
|------|--------|-------|
| Pool | `Available()` → branch 2: `14−13.5−7.82=0` → **REJECT** | `limit−reserved=4.78GB` → **PASS** |
| RSS | `13.5+7.82+0.18=21.5>12.6` → REJECT | `13.5>12.6` → REJECT |

The RSS gate is a no-op for this scenario — it blocks using either formula because RSS (13.5GB) is at 96% of the container limit. At that point rejection IS correct. The false reject was from the pool check double-count via `Available()`.

The fix is self-contained: `limit − reserved` for the pool, RSS gate unchanged. I'll update the PR to reflect just this.

XuPeng-SH (changes requested):
I re-reviewed the current head and I still found blocking issues.

1. The current change does not actually unblock the target acquire path. In the old code, the only changed branch was the pool check inside `acquireWithinRSSLimit`; but in exactly those branch-2 cases, if `Available()` rejects then the unchanged RSS gate (`rss + reserved + ask > total * limitRate`) also rejects. So replacing the pool check with `limit - reserved` changes the reported headroom, but it does not change the final grant/deny outcome for the unhappy path this PR is trying to fix. In other words, the false-reject path is still blocked by the unchanged RSS gate.

2. The updated tests still do not pin the current critical failure path: there is no case where pool headroom is sufficient (`limit - reserved >= ask`) but the request must still fail because `rss + reserved + ask` crosses the RSS limit. For the current logic, that is the key negative path and it should be locked in directly.

Because of (1), I do not think the current revision actually fixes the acquire decision in the scenario described by this PR. I am requesting changes.

XuPeng-SH:
I think the underlying problem class is real, but the current revision still does not close the loop on the acquire decision. Concretely, only the pool check changed (`limit - reserved` instead of `Available()`), while the final RSS gate is still `rss + reserved + ask > total * limitRate`.

That means the branch-2 unhappy path is still effectively blocked by the unchanged RSS gate. In other words, this revision changes the reported pool headroom, but it does not actually change the final grant/deny outcome for the scenario the PR is trying to unblock.

I think the fix needs to change the RSS-side accounting as well, but not by switching to plain `max(rss, reserved)` or to fully independent gates — both of those lose important information.

The safe direction is to separate the overlapping and non-overlapping pieces explicitly:

- keep the pool gate as `limit - reserved`
- but make the RSS gate account for **unrelated process RSS + forward S3 reservations**, instead of `rss + reserved` over the whole process

A practical shape would be something like:

`effective_used = rss - current_materialized_s3 + reserved + ask`

where `current_materialized_s3` is the portion of current RSS that corresponds to already-materialized S3 buffers. Then gate on:

`effective_used <= total * limitRate`

That preserves unrelated RSS, preserves forward reservations, and avoids double-counting only the overlapping S3 portion.

Without some equivalent overlap-aware accounting, the choices are still:
- current code: over-count when reserved bytes are already reflected in RSS
- `max(rss, reserved)`: under-count unrelated RSS
- fully independent pool/RSS gates: under-count the combined case

I would also add a test that proves the new algorithm really changes the current decision path, not just the intermediate pool value. Right now the suite is still missing the RSS-only fail case (`limit - reserved >= ask`, but `rss + reserved + ask` exceeds the cap), and it also does not demonstrate a concrete old-deny/new-allow case for the latest formula.

XuPeng-SH:
I think the current failure path is over-constrained because the same S3 write bytes are effectively being controlled twice: once by the S3 reservation/pool logic, and again by an RSS gate over whole-process memory without separating the overlapping S3 portion.

A safer fix direction would be:

1. Keep the pool gate as the primary S3 write quota bound.
   - Use something like `limit - reserved` to decide how much more S3 write memory can be promised.
   - This gate should stay S3-specific and should not mix in RSS.

2. Keep an RSS gate only for physical-memory protection, but make it overlap-aware.
   - The current whole-process `rss + reserved + ask` style is what creates the overkill when already-materialized S3 buffers are also present in RSS.
   - What we really want is:

     `effectiveUsed = rss - materializedS3 + reserved + ask`

     where `materializedS3` is the part of current RSS already attributable to S3 buffers.

   - Then the check becomes:

     `effectiveUsed <= total * limitRate`

   This preserves unrelated process RSS, preserves forward S3 reservations, and removes only the true overlap.

3. If we do not currently have a reliable `materializedS3` signal, that means the model is not rich enough to support two overlapping gates safely.
   - In that case, continuing to tweak `Available()` / `rss + reserved` / `max(rss, reserved)` is likely to keep bouncing between over-count and under-count.
   - The fix should first make the accounting model explicit instead of stacking ambiguous gates.

4. The error reporting should also reflect the actual reject reason.
   - If the pool gate rejects, say the pool is exhausted.
   - If the RSS gate rejects, say the process is under RSS / physical-memory pressure.
   - Right now the message can show `available X, ask Y` even when the final reject came from a different gate, which is misleading for debugging.

So my current view is: the right fix is not another small `Available()` adjustment, but making the overlap between S3 reservation and whole-process RSS explicit. Without that, the path remains easy to overtighten or undertighten.

aunjgr (commented):
Changed approach: removed the RSS gate entirely instead of fixing its formula.

The RSS gate (`rss + reserved + ask > total * limitRate`) double-counts the same way `Available()` did. There is no correct formula without knowing the S3/non-S3 RSS split. And the original purpose of the RSS gate (preventing OOM from non-S3 memory pressure, from PR #24268) is now handled by RSS scavenging (85%/92% thresholds with graduated cache eviction + FreeOSMemory, added in #24665).

The fix is now two changes:
1. Pool check: `limit - reserved` instead of `Available()` (unchanged)
2. RSS gate: deleted

Physical memory safety is provided by:
- `pinnedRate >= 0.80` hard ceiling at the `AcquirePolicyForCNFlushS3` level
- RSS scavenging with graduated cache eviction + FreeOSMemory

@XuPeng-SH PTAL

XuPeng-SH (approved):
I re-checked the latest update and the previously blocking issue is resolved.

The old problematic `max(rss, reserved)` path is gone. This version now scopes the CN flush policy to the forward-looking reservation pool itself, instead of trying to fold whole-process RSS into the same gate and risking undercounting. Given that the direct policy tests were updated accordingly and I do not see a new hard issue in the latest diff, this looks good to me now.

Issue #24847, opened 2026-06-04
[Bug] DATA BRANCH DIFF full-row output fails with 'unsupported mysql type 0' on DECIMAL/vector columns (fixed on main)


dengn:
## Summary
`DATA BRANCH DIFF` with the default full-row output fails when the table contains a `DECIMAL` (or vector) column: it raises `unsupported mysql type 0` and returns no rows. `OUTPUT SUMMARY` and `COLUMNS(...)` (projecting only non-decimal columns) work around it. Appears **fixed on main**.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test case
```sql
CREATE DATABASE dtest; USE dtest;
CREATE TABLE t_dec(id INT PRIMARY KEY, name VARCHAR(20), price DECIMAL(10,2));
INSERT INTO t_dec VALUES (1,'A',10.00),(2,'B',20.00),(3,'C',30.00);
DATA BRANCH CREATE TABLE t_dec_br FROM t_dec;
UPDATE t_dec_br SET price=15.00 WHERE id=1;
INSERT INTO t_dec_br VALUES (4,'D',40.00);

-- (1) full-row DIFF over a table with a DECIMAL column
DATA BRANCH DIFF t_dec_br AGAINST t_dec;                 -- v3.0.11: FAILS
-- (2) summary works
DATA BRANCH DIFF t_dec_br AGAINST t_dec OUTPUT SUMMARY;  -- works
-- (3) projecting only non-decimal columns works
DATA BRANCH DIFF t_dec_br AGAINST t_dec COLUMNS (name);  -- works
```

## Actual result
- **v3.0.11 (Cloud)** — statement (1) fails:
```
ERROR 20101 (HY000): internal error: unsupported mysql type 0
```
(MySQL type 0 = `MYSQL_TYPE_DECIMAL`; the same happens for tables with a `VECF32`/vector column.)
- **main (commit 0bfda60)** — statement (1) succeeds:
```
| diff t_dec_br against t_dec | flag   | id   | name | price |
| t_dec_br                    | UPDATE |    1 | A    | 15.00 |
| t_dec_br                    | INSERT |    4 | D    | 40.00 |
```

## Expected
[...]

PR #24894, opened 2026-06-09
Fix data branch decimal results on main

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #24847

Port the data branch decimal/year result conversion and tuple handling fixes to latest `main`.

- Preserve decimal precision/scale metadata from MySQL result columns, validate invalid/missing decimal metadata, and align year result metadata/output formatting.
- Materialize decimal64/decimal128/decimal256/year row values as typed MatrixOne vectors when building data branch output and pick batches.
- Decode, encode, compare, and merge decimal256/year tuple values across data branch hashdiff, pick, apply SQL generation, logtail replay, branch hashmap, read filter, lock fetch, and merge-sort paths.
- Fix fuzzy-check subquery string-literal quoting with `tree.WithSingleQuoteString()`.
- Add focused Go tests plus BVT coverage: `diff_15.sql` for decimal64/decimal128/decimal256 `data branch diff`, `diff_16.sql` for year `data branch diff`, and `pick_12.sql`/`pick_13.sql` for pick paths.

Notes:
- Rebased onto `main` at `890a3800c44932b4dd190ba59277aa73c502bd05`; conflict with upstream `diff_14.sql` was resolved by keeping upstream `diff_14` and moving this PR's decimal/year cases to `diff_15`/`diff_16`.
- The decimal256 PK pick issue is tracked separately in #24902 and intentionally not fixed in this PR.

Checks:
- `git diff --check upstream/main...HEAD` passed.
- `make -C thirdparties` passed locally.
- `source $HOME/.zshrc && moenv && go test ./pkg/container/types ./pkg/frontend ./pkg/vm/engine/disttae/logtailreplay`: `pkg/container/types` passed; `pkg/frontend` and `pkg/vm/engine/disttae/logtailreplay` are locally blocked by the current `upstream/main` thirdparty mismatch where `github.com/unum-cloud/usearch/golang@v0.0.0-20260524141737-9fd6b0115dcd` expects symbols such as `usearch_scalar_u8_k` and `usearch_hardware_acceleration_*`, but `thirdparties/install/include/usearch.h` built from `usearch-2.25.3.tar.gz` does not provide them.
- GitHub CI was re-triggered after the rebase push.


heni02 (approved):

XuPeng-SH (changes requested):
I didn't find a new hard correctness bug in the port itself, but at this review bar I still think the coverage is not quite enough for the surfaces this PR touches.

1. **The new `appendFromEntry()` decimal256 branch is untested.**
   This PR adds a dedicated `T_decimal256` path in `pkg/vm/engine/disttae/logtailreplay/change_handle.go`, but I don't see a unit/integration test that actually drives that runtime path. That's a separate execution surface from the frontend/data-branch conversion code, so it deserves its own protection.

2. **The new BVT only proves an INSERT-style diff result, not UPDATE / DELETE decimal diff paths.**
   `diff_14.sql` exercises decimal output for `data branch diff`, but only for a newly inserted row. The real regression surface also includes UPDATE of decimal columns (old/new tuple comparison + typed normalization) and DELETE rows carrying decimals. Given how much of this fix is about typed decimal materialization/comparison, I'd like at least one execution-level UPDATE/DELETE case before approval.

If those two gaps are covered, the rest of the port looks good to me.

gouhongshen:
Updated this PR to address the review coverage gaps.

Covered in this push:
- Added focused `appendFromEntry()` coverage for `decimal256`, including null handling.
- Expanded `diff_14` to cover decimal UPDATE / INSERT / DELETE paths under `save_query_result=on`.
- Added a product-level `pick_12` conflict case using `decimal(39,4)` as a non-PK value, covering fail / skip / accept behavior. This also covers the decimal256 pick output path without mixing in the separately tracked decimal256 PK issue (#24902).

Local validation on `main`:
- `source $HOME/.zshrc && moenv && go test ./pkg/frontend ./pkg/vm/engine/disttae/logtailreplay`
- BVT `diff_14`: 15/15 passed
- BVT `pick_12`: 69/69 passed
- `git diff --check`

Re-requesting review.

gouhongshen:
已把 3.0-dev 上处理 change request 的 decimal256 data branch pick 修复 port 到 main。

本次更新 commit: `4259e88572 fix: support decimal256 data branch pick keys`

本地验证：

```bash
source $HOME/.zshrc && cd /Users/ghs-mo/MOWorkSpace/matrixone-2nd && moenv && cd /private/tmp/matrixone-port-decimal-main && go test ./pkg/container/types ./pkg/container/vector ./pkg/frontend ./pkg/sql/colexec/lockop ./pkg/vm/engine/readutil ./pkg/objectio/mergeutil ./pkg/vm/engine/tae/mergesort -run 'TestSimpleTupleAllTypes|TestSingleTypeTuple|TestDecimal256OrderTuple|TestTupleSQLStrings|TestTupleSQLStringsWithEncodeDecode|TestDecodeTupleWithSchema|TestUnpackNthElement|TestType_DescString|TestFetchDecimal256Rows|TestFetchDecimal256RowsWithFilter|TestFetchDecimal256RowsWithFilterAll|TestNumericGap_AllTypes|TestIsNumericType|TestAppendNumericStringToVec_AllTypes|TestCoercePickKeyVectorToType_Decimal256ReformatsToTargetType|TestExtractPKValDecimal256|TestFormatPickKeyVectorValueAsString_AllSupportedKinds|TestLCAProbeJoinCastType|TestConstructBlockPKFilter|TestMergeBaseFilterInKind|TestMergeSortBatchesDecimal256|TestAObjMergeDecimal256' -count=1
```

结果：PASS。`git diff --check` 也已通过。

aunjgr (approved):
The port itself looks correct. XuPeng-SH's coverage gaps (decimal256 appendFromEntry, UPDATE/DELETE diff) are valid but not blocking for this scope — the port preserves existing behavior on main.

XuPeng-SH (changes requested):
The decimal round-trip fix looks useful, but the new generic DECIMAL metadata inference still has a real type-selection hole.

`pkg/frontend/util.go` now derives decimal precision as `length - 2` in `mysqlDecimalPrecisionFromColumn()`. That only matches MatrixOne's own internal `setColLength()` convention, where DECIMAL length is encoded as `width + 2` (`pkg/defines/type.go`). It is not a general rule for MySQL DECIMAL metadata.

For example, a signed `DECIMAL(19,0)` can be reported with display length 20, which this code would infer as precision 18. That downgrades the value from `T_decimal128` to `T_decimal64`. If this path consumes real MySQL-compatible column metadata, 19-digit scale-0 decimals can be materialized into the wrong engine type and then overflow or misparse.

Please avoid reverse-engineering precision from `Length()` with a fixed offset, and add coverage for metadata that does not come from MO's own `width + 2` producer path.

XuPeng-SH (changes requested):
I re-checked the latest head. The decimal fixes look useful, but there is still one remaining typed-value hole in the new tuple normalization path.

`normalizeTupleColumnValue()` was added so values coming back from the branch hashmap as raw `[]byte` can be decoded back into fixed-width MatrixOne types before compare / format / append. But its fixed-width decode switch still omits `types.T_year`.

That leaves a real runtime gap:
- `YEAR` is otherwise treated as a supported typed value in the surrounding data-branch helpers (formatting / compare paths already have `types.T_year` branches)
- the branch hashmap still uses the raw-byte/default path for fixed-width types it does not specially materialize
- so a `YEAR` value coming back through this new normalization path still falls into `"unexpected byte slice for fixed-width column type"` instead of being decoded

In other words, the decimal regression is fixed, but `data branch diff` can still break on tables containing `YEAR` columns because this new byte->typed normalization path is now the choke point.

**Concrete suggestions**

1. Add `types.T_year` to the fixed-width decode switch in `normalizeTupleColumnValue()`.
2. Add one focused regression test that actually drives the byte-slice path for a `YEAR` column through `data branch diff` / tuple normalization, not just helper-level formatting.
3. If you want a user-visible proof, a small git4data branch diff case with a `YEAR` column would lock the full path.

XuPeng-SH (changes requested):
I re-checked the current head from a first-principles / unhappy-path angle. The decimal fixes look useful, but I still see a few substantive type-handling holes in the broader path this PR is touching.

1. **`CompareValue` is still not a valid comparator for several numeric types**
   - The current implementation uses subtraction + `int(...)` for `uint8/16/32` and `float32/64`.
   - That is not comparator-safe in Go: unsigned subtraction can wrap (`uint8(1)-uint8(2) -> 255`), and float deltas smaller than 1 can collapse to `0` after `int(...)`.
   - This matters here because the new data-branch logic relies on `CompareValue` for row comparison and PK sorting.
   - **Suggestion:** switch those branches to explicit `< / > / ==` comparisons that return `-1/0/1`, and add tests for `float` deltas `< 1` plus ascending/descending unsigned cases.

2. **`Tuple.SQLStrings()` no longer returns SQL literals, but `fuzzyCheck` still concatenates its output directly into SQL**
   - `Tuple.SQLStrings()` now returns raw values for text/date/time/uuid-ish types without quoting/escaping.
   - `pkg/sql/compile/fuzzyCheck.go` still injects those strings directly into `col = %s` predicates.
   - So composite keys containing plain strings, dates, timestamps, UUIDs, or embedded quotes can now generate invalid or semantically wrong SQL.
   - **Suggestion:** either restore SQL-literal quoting/escaping in `Tuple.SQLStrings()`, or stop using it for SQL generation and quote/parameterize at the `fuzzyCheck` call site.

3. **`YEAR` is still incomplete across the end-to-end data-branch path**
   - The PR adds some low-level YEAR handling, but the full data-branch diff/pick path still has gaps: normalization, PK extraction, formatting, and decode paths are not consistently updated together.
   - In particular, `normalizeTupleColumnValue()` / PK extraction / `DecodeValue()` are still the choke points for YEAR values coming back as raw bytes.
   - **Suggestion:** add explicit `T_year` handling consistently across encode/decode/normalize/format/PK extraction, then add at least one focused `DATA BRANCH DIFF/PICK` regression using YEAR columns.

Given the comparator correctness bug and the `Tuple.SQLStrings()` SQL-generation regression, I think this should stay at **request changes** on the current head.

gouhongshen:
Addressed the latest review in `3be9969f87`.

- `CompareValue` now uses comparator semantics for ordered scalar types instead of subtraction/int casts. Added coverage for unsigned reverse ordering and float deltas smaller than 1.
- `fuzzyCheck` no longer feeds raw `Tuple.SQLStrings()` values directly into compound-key SQL predicates. It now formats compound collision keys with per-column SQL literal quoting while still preserving decimal scale formatting. Added coverage for string values with quotes, DATE, YEAR, and DECIMAL in one compound key.
- Completed the remaining YEAR pick path: PK extraction, literal parsing, subquery coercion formatting, and branch pick BVT coverage. Added `pick_13` for YEAR primary-key `DATA BRANCH PICK` via string literal key and YEAR subquery key.

Verification:
- `git diff --check`
- `go test ./pkg/container/types -count=1`

Local `pkg/frontend` / `pkg/sql/compile` targeted tests are still blocked by the local usearch C-symbol mismatch; CI should cover those packages on the PR branch.

XuPeng-SH (changes requested):
The new compound-key quoting in `pkg/sql/compile/fuzzyCheck.go` is still semantically wrong.

`formatCompoundCollisionKey` now uses `strconv.Quote`, but that produces Go double-quoted / backslash-escaped string literals, not SQL literals. Those values are then concatenated into `f.condition` and executed by `backgroundSQLCheck()` via `runSqlWithResultAndOptions()`, which propagates the current session's variable resolver (`WithResolveVariableFunc(...)`). So this sub-SQL still runs under the caller's `sql_mode` behavior.

That leaves at least two real failure modes:

1. Under `ANSI_QUOTES`, values like `"2025-01-03"` / `"O'Reilly"` are no longer string literals, so the duplicate-check SQL for compound keys can parse or evaluate incorrectly.
2. Under `NO_BACKSLASH_ESCAPES`, the Go-style escaping emitted by `strconv.Quote` is also not the right SQL literal escaping, so strings / binary-ish payloads containing backslashes or escapes can be misparsed.

So the new fuzzy-check path is still not SQL-mode-safe for the exact string/date/year/blob/uuid-style cases this patch is trying to fix, and `TestFillCompoundConditionQuotesSQLLiterals` is currently locking in the wrong representation. Please switch this to repo-consistent single-quoted SQL literal generation (for example the existing `tree.WithSingleQuoteString()` / SQL-string escaping style used elsewhere) and add a regression that exercises at least `ANSI_QUOTES`.

XuPeng-SH (approved):
Re-reviewed latest head `cb18ec8` from the data-flow and unhappy-path angles.

The previously requested blockers look closed now:
- decimal metadata inference no longer hard-codes `length - 2`; it accounts for scale/sign and has boundary/error coverage, including signed `DECIMAL(19,0)` and decimal256.
- `CompareValue` no longer uses subtraction/truncation for unsigned/float comparators, and the regression tests cover unsigned direction plus sub-1 float deltas.
- `YEAR` is now covered consistently across decode/normalize/format/PK extraction/comparison/output surfaces, with helper and BVT coverage.
- decimal256 is covered across tuple encode/decode, SQL string formatting, branch diff/pick conversion, logtail replay appendFromEntry, readutil PK filters, lock fetch, and merge sort.
- compound fuzzy-check SQL generation no longer uses Go `strconv.Quote`; it uses single-quoted SQL literals through `sqlquote.String`, with coverage for embedded quote/backslash and ANSI_QUOTES-sensitive no-double-quote output.

I also checked the latest CI rollup: Summary is green. Local targeted tests passed for the touched hot paths:
- `./pkg/common/sqlquote`
- `./pkg/container/types`
- `./pkg/frontend`
- `./pkg/sql/compile`
- `./pkg/sql/colexec/lockop`
- `./pkg/vm/engine/readutil`
- `./pkg/objectio/mergeutil`
- `./pkg/vm/engine/disttae/logtailreplay`
- `./pkg/frontend/databranchutils`

Non-blocking: the PR is still marked `BEHIND`, so it should be updated with latest `main` before merge if the queue requires it.

iamlinjunhong (approved):

PR #24895, opened 2026-06-09
4.0: Fix data branch decimal results on 4.0-dev

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #24847

Port the data branch decimal result conversion fix to `4.0-dev`.

- Preserve decimal precision/scale metadata from MySQL result columns and validate invalid/missing decimal metadata.
- Materialize decimal64/decimal128/decimal256 row values as typed MatrixOne vectors when building data branch output batches.
- Decode and compare decimal256 raw tuple values in data branch hashdiff/change replay paths.
- Add focused unit tests and BVT coverage with `diff_14.sql` for decimal64/decimal128/decimal256 `data branch diff`.

Notes:
- The pick subquery string-literal fix from `565a0f71` is already present in `4.0-dev` with `tree.WithSingleQuoteString()` and unit coverage.
- The decimal256 PK pick issue is tracked separately in #24902 and intentionally not fixed in this PR.

Checks:
- `make -C thirdparties`
- `source $HOME/.zshrc && moenv && go test ./pkg/container/types ./pkg/frontend ./pkg/vm/engine/disttae/logtailreplay`
- `git diff --check upstream/4.0-dev...HEAD`

Draft until CI confirms this port on `4.0-dev`.


heni02 (approved):

XuPeng-SH (changes requested):
I did not confirm a blocking logic bug in the implementation itself, but I still think the regression coverage is not strong enough for this decimal port.

There are three review-level gaps in the current tests:

1. The new BVT `diff_14.sql` only proves an INSERT diff. It still does not cover the more important unhappy path where the same PK exists on both sides and only the `decimal256` value changes, so the diff must surface an UPDATE correctly.

2. The PR adds `T_decimal256` support in `compareSingleValInVector`, but there is still no end-to-end branch conflict / merge / pick style case with a real `decimal(39,...)` value conflict. Right now that new helper branch is only covered in isolation, not in the product path that previously would have failed.

3. `pkg/vm/engine/disttae/logtailreplay/change_handle.go` adds a new `appendFromEntry(T_decimal256)` path, but there is still no focused unit test for it, including null handling. That is changed code on a separate replay path from the frontend/data-branch fixes and should be pinned directly.

Since this PR is specifically about making decimal diff results reliable on 4.0, I think at least these product-facing unhappy paths should be covered before approval.

gouhongshen:
Updated this PR to address the review coverage gaps.

Covered in this push:
- Added focused `appendFromEntry()` coverage for `decimal256`, including null handling.
- Expanded `diff_14` to cover decimal UPDATE / INSERT / DELETE paths under `save_query_result=on`.
- Added a product-level `pick_12` conflict case using `decimal(39,4)` as a non-PK value, covering fail / skip / accept behavior. This also covers the decimal256 pick output path without mixing in the separately tracked decimal256 PK issue (#24902).

Local validation on `4.0-dev`:
- `source $HOME/.zshrc && moenv && go test ./pkg/frontend ./pkg/vm/engine/disttae/logtailreplay`
- BVT `diff_14`: 15/15 passed
- BVT `pick_12`: 69/69 passed
- `git diff --check`

Re-requesting review.

gouhongshen:
已把 3.0-dev 上处理 change request 的 decimal256 data branch pick 修复 port 到 4.0-dev。

本次更新 commit: `cab3ab3fa9 fix: support decimal256 data branch pick keys`

本地验证：

```bash
source $HOME/.zshrc && cd /Users/ghs-mo/MOWorkSpace/matrixone-2nd && moenv && cd /private/tmp/matrixone-port-decimal-4.0 && go test ./pkg/container/types ./pkg/container/vector ./pkg/frontend ./pkg/sql/colexec/lockop ./pkg/vm/engine/readutil ./pkg/objectio/mergeutil ./pkg/vm/engine/tae/mergesort -run 'TestSimpleTupleAllTypes|TestSingleTypeTuple|TestDecimal256OrderTuple|TestTupleSQLStrings|TestTupleSQLStringsWithEncodeDecode|TestDecodeTupleWithSchema|TestUnpackNthElement|TestType_DescString|TestFetchDecimal256Rows|TestFetchDecimal256RowsWithFilter|TestFetchDecimal256RowsWithFilterAll|TestNumericGap_AllTypes|TestIsNumericType|TestAppendNumericStringToVec_AllTypes|TestCoercePickKeyVectorToType_Decimal256ReformatsToTargetType|TestExtractPKValDecimal256|TestFormatPickKeyVectorValueAsString_AllSupportedKinds|TestLCAProbeJoinCastType|TestConstructBlockPKFilter|TestMergeBaseFilterInKind|TestMergeSortBatchesDecimal256|TestAObjMergeDecimal256' -count=1
```

结果：PASS。`git diff --check` 也已通过。

XuPeng-SH (approved):
Reviewed from multiple angles with extra unhappy-path focus.

I did not find a new blocking issue in this 4.0-dev backport. Decimal metadata preservation, decimal64/128/256 materialization, and decimal256 diff/replay handling all look consistent with the intended fix, and the added unit/BVT coverage hits the main regression surface.

I also checked the main unhappy-path concern around `convertRowsIntoBatch` early-return cleanup: that partial-batch cleanup pattern already exists in 4.0-dev (for example the pre-existing timestamp parse path), so this PR does not introduce a new cleanup regression there.

aunjgr (approved):

Issue #24897, opened 2026-06-09
[Bug]: LOAD DATA can fail with lock table bind changed after lock-table owner CN shuts down


aptend:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

## Branch Name

main / dev deployment (`freetier-01`)

## Commit ID

Unknown from available dev logs. Incident observed in `freetier-01` on 2026-05-26 around 04:30 UTC.

## Other Environment Information

- Environment: dev
- Namespace: `freetier-01`
- DN pod: `default-dn-0`
- LOAD CN pod: `freetier-01-s16c64g-854877c9c-cn-9dppc`
- LOAD CN uuid: `34633935-3265-6239-3431-363235323263`
- Old lock-table owner CN pod: `freetier-01-s16c64g-854877c9c-cn-hn4zh`
- Old owner CN uuid: `39326230-3664-3939-3331-393234393132`
- Old owner lockservice ServiceID: `177976352882142208039326230-3664-3939-3331-393234393132`
- Old owner lockservice address: `172.20.42.6:6003`

## Actual Behavior

A long-running `LOAD DATA LOCAL INFILE` on `cn-9dppc` failed at commit/finish time with `lock table bind changed` after running for about 1 hour.

The failed statement was:

```sql
LOAD DATA LOCAL INFILE '/data/hive2mo-sync/work_dev_s4_acdoca_2026_direct/dwd_s4_acdoca_hive/data.csv'
INTO TABLE _staging_dwd_s4_acdoca
FIELDS TERMINATED BY ',' OPTIONALLY ENCLOSED BY '"' ESCAPED BY '"'
LINES TERMINATED BY '\n'
```

Target table:

- account id: `693780002`
- db: `dwd_dcp`
- table: `_staging_dwd_s4_acdoca`
- table id: `5788924`
- LOAD txn: `0393c0da1a884c4f18b2e3cffe0b1296`
- trace id: `0a08c1c7-44c1-082c-7580-b395e724fc70`

Timeline from Loki:

- `2026-05-26 03:27:02 UTC`: target table `_staging_dwd_s4_acdoca` created with `table-id=5788924`.
- `2026-05-26 04:09:03 UTC`: `cn-hn4zh` received terminate and started shutdown:
  - `Starting shutdown...signal: terminated`
  - `cn heartbeat task stopped`
  - local lock binds closed, including bind owner ServiceID `...39326230...`
[...]

PR #24898, opened 2026-06-09
fix: check lock table binds during load

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24897

A long-running `LOAD DATA` can hold stale lock table bind information while data is being read and written. If the lock-table owner CN shuts down during that window, the transaction may only discover the bind change at commit time and fail after a long run.

This change adds a LOAD-time lock table bind check before external batches are read. The check is throttled at the txn operator level, so concurrent LOAD readers in the same transaction share the same interval. It reuses the existing `GetLockTableBind` path instead of adding a new lockservice/proto RPC. Once a transaction detects a bind change, subsequent checks in the same transaction keep returning `ErrLockTableBindChanged`.

Tests cover changed bind detection, sticky bind-change failure, and reset of the txn-level throttle state when a txn operator is reused.


aunjgr (approved):
Solid fix. Throttled at 3-min, sticky on first detection, uses allocator-latest bind (bypasses local stale cache). The lockTables snapshot outside the mutex is safe since lock tables are append-only during a txn lifetime. Tests cover changed bind, stale local cache, sticky failure, and throttle reset on reuse.

XuPeng-SH (changes requested):
I did not confirm a blocking logic bug in the implementation itself, but I still think the unhappy-path coverage is not strong enough for this change.

The new bind-check behavior is added to multiple critical LOAD paths, but several of the newly introduced failure/guard branches are still effectively untested:

1. `processLoadLocal` now calls `CheckLockTableBinds()` before the first packet read and again inside the main loop, but the current load-local tests build `ExecCtx` with `proc == nil`, so both new call sites are dead in test coverage. We still need coverage for bind-change failure before the first read and after at least one loop iteration.

2. `txnOperator.CheckLockTableBinds()` adds important unhappy-path state handling that is currently not pinned: RPC error should clear the throttle timestamp so later calls retry; cancelled context should also clear the throttle; a closed txn should fail before any bind lookup; and a clean second call inside the 3-minute window should no-op instead of re-querying. These branches are now part of the core contract for the new throttled check.

3. `External.Call()` now propagates bind-check failures for LOAD, but there is still no direct test that a `CheckLockTableBinds()` error is surfaced from the external LOAD path.

Since this PR is specifically about failing early on stale bind changes during long-running LOAD, I think these negative paths should be covered before approval.

XuPeng-SH (approved):
I re-checked the latest revision and I do not see a blocking issue now.

The earlier unhappy-path coverage gaps are addressed in the current head: `processLoadLocal` now has bind-check failure coverage both before the first read and after entering the loop; `txnOperator.CheckLockTableBinds` now has direct tests for RPC error, canceled context, closed txn, sticky failure, throttle reset, and clean throttled re-check behavior; and the external LOAD path now has a direct bind-check error propagation test.

LGTM.

gouhongshen (approved):

LeftHandCold (approved):
LGTM. I reviewed the latest head around the new LOAD-time bind checks, the txn-level throttled/sticky CheckLockTableBinds implementation, and the added negative-path tests. The previous coverage gaps are addressed, and the PR checks are green.

iamlinjunhong (approved):

jiangxinmeng1 (approved):

PR #24901, opened 2026-06-09
fix: prioritize stalled checkpoint bounded flush

### LINKED ISSUE #24900: Stalled checkpoint flush can be blocked by post-checkpoint appendable objects
## Problem

A pending incremental checkpoint can wait on an old appendable object whose `CreatedAt` is before the checkpoint end timestamp, while normal flusher scheduling keeps using a wall-clock dirty-tree end timestamp.

When newer appendable objects created after the checkpoint end are ordered before the old blocker, any `PrepareCompact` failure on those newer objects can cause `fireFlushTabletail` to skip the whole chunk before reaching the old object. The checkpoint then continues waiting even though the blocker belongs to the checkpoint range and the newer objects do not.

Example shape:

- checkpoint end: `01:45:55`
- checkpoint blocker: old object created before checkpoint end
- normal flush batch also includes several post-checkpoint-end appendable objects before the blocker
- a newer object's `PrepareCompact` failure prevents the old blocker from being scheduled for flush

## Expected

When an incremental checkpoint has been pending for a long time, flusher should prioritize flushing objects in the checkpoint range without letting post-checkpoint-end objects block that bounded work.

## Proposed fix

Add a checkpoint-bounded flusher path when a pending incremental checkpoint has been waiting for more than five minutes:

- scan dirty objects in `[intent.start, intent.end]`
- use `intent.end` as the flush cutoff
- derive `lastCkp` from `intent.start`
- reuse existing `FlushTableTailTask`, `PrepareCompact`, and `asyncJobDispatcher` scope conflict handling
- keep normal wall-clock flush scheduling unchanged

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes https://github.com/matrixorigin/matrixone/issues/24900

Fixes #24900.

This adds a checkpoint-bounded flusher path for stalled incremental checkpoints. When a pending ICKP has not passed `flushChecked` and has been pending for more than five minutes, the flusher now scans the checkpoint range `[intent.start, intent.end]` and schedules flush work with `intent.end` as the cutoff before the normal wall-clock cron flush.

The bounded path:

- does not merge with force/manual flush requests
- does not merge with normal cron flush requests
- does not reset the normal table flush deadline
- reuses existing `PrepareCompact`, `FlushTableTailTask`, and `asyncJobDispatcher` scope conflict handling

This prevents post-checkpoint-end appendable objects from blocking the old appendable objects that the checkpoint is actually waiting for.

## Tests

- `gofmt` on changed files
- `git diff --check -- pkg/vm/engine/tae/db/checkpoint/flusher.go pkg/vm/engine/tae/db/checkpoint/flusher_test.go pkg/vm/engine/tae/db/checkpoint/info.go`
- `go test ./pkg/vm/engine/tae/db/checkpoint` could not run in this workspace because vendor metadata is inconsistent with go.mod
- `go test -mod=mod ./pkg/vm/engine/tae/db/checkpoint` could not run because local cgo headers are missing (`usearch.h`, `xxhash.h`)
- `CGO_ENABLED=0 go test -mod=mod ./pkg/vm/engine/tae/db/checkpoint` could not run because this package graph requires cgo-backed symbols


jiangxinmeng1:
Updated the flusher flow so the checkpoint-bounded dirty tree is built during the periodic flusher trigger and enqueued into flushRequestQ as its own request mode. The queue worker now only consumes force -> ckp-bounded -> cron requests, preserving the bounded cutoff without scanning from onFlushRequest.

aptend (changes requested):
Found one blocking issue in the new test.

pkg/vm/engine/tae/db/checkpoint/flusher_test.go:69 sets bornTime to time.Now().Add(-threshold) and then expects shouldScheduleStalledCheckpointFlush to return false. But Age() uses time.Since(bornTime), so by the time the helper is called the measured age is already slightly greater than threshold. Locally this fails deterministically for me at line 70. Please make the boundary test stable, for example by using an age safely below the threshold for the false case, or by injecting/fixing the clock.

Local check:
CGO_CFLAGS='-I/home/mo/projects/matrixone/thirdparties/install/include' CGO_LDFLAGS='-L/home/mo/projects/matrixone/thirdparties/install/lib -L/home/mo/projects/matrixone/cgo' LD_LIBRARY_PATH='/home/mo/projects/matrixone/thirdparties/install/lib:/home/mo/projects/matrixone/cgo' go test ./pkg/vm/engine/tae/db/checkpoint -run 'TestShouldScheduleStalledCheckpointFlush|Test_RestartFlusher' -count=1

Result: TestShouldScheduleStalledCheckpointFlush fails.

XuPeng-SH (changes requested):
I understand the problem this PR is trying to solve, and the design direction makes sense: when an incremental checkpoint has been stuck for long enough, flush only the objects inside that checkpoint range instead of letting newer appendable objects keep blocking it.

I did not find a clear logic bug in the implementation, but with a very high bar I still think the regression coverage is not strong enough yet. Right now the new test only proves the **predicate** (`shouldScheduleStalledCheckpointFlush`) and some state combinations. It does **not** prove the actual behavior that matters for the fix:

1. that `scheduleStalledCheckpointFlush()` really scans `[intent.start, intent.end]` and fires bounded flush work for that range;
2. that the new `flushModeCKPBounded` path does **not** reset normal table flush deadlines (which is one of the key semantic differences from force flush);
3. that a stalled pending ICKP can actually make progress again once this bounded path runs.

For this PR, I would want at least one direct test of the bounded scheduling path itself (or an integration-style test with an injected/short threshold) before calling it sufficiently covered. The current test validates the gate, but not the recovery mechanism.

jiangxinmeng1:
Addressed the test review in 87f326321e:\n\n- Made the threshold false case stable by setting bornTime safely below the 5 minute threshold.\n- Added TestMakeStalledCheckpointFlushRequest to directly verify that a stalled pending ICKP builds a ckp-bounded request by scanning exactly [intent.start, intent.end], preserves that time range on the request dirty tree, uses flushModeCKPBounded, and does not scan when the intent is below threshold.\n\nI reran the narrow test command with local thirdparty include/lib paths plus -mod=mod because this workspace has inconsistent vendor metadata. The package still cannot build locally due to the installed usearch C headers/libs missing symbols expected by the Go binding (for example usearch_hardware_acceleration_available / usearch_scalar_e5m2_k), so I could not get a local pass here.

aptend (changes requested):
Current implementation can address the issue, but the change surface is larger than necessary. I suggest keeping the bounded flush special case inside `triggerJob` instead of introducing `flushModeCKPBounded` and propagating that mode into `onFlushRequest`, `scheduleFlush`, and `checkFlushConditionAndFire`.

The essence of the fix is: when a pending ICKP is stalled and has not passed `flushChecked`, this flush round should prioritize a bounded dirty tree over `[intent.start, intent.end]` instead of the normal cron merged dirty tree. The downstream flush scheduling path does not need to know this is a separate mode; it can continue to use the existing `scheduleFlush` flow.

For example, the structure could stay close to:

```go
flusher.sourcer.Run(flusher.flushLag)

entry := flusher.sourcer.GetAndRefreshMerged()
if request := flusher.makeStalledCheckpointFlushRequest(); request != nil {
    entry = request.tree
}

if !entry.IsEmpty() {
    flusher.flushRequestQ.Enqueue(&FlushRequest{tree: entry})
}
```

I also do not think `lastCkp` needs to be passed separately for the bounded path. In the normal ICKP chain:

```text
pending.start = max finished ICKP end.Next()
pending.start.Prev() = max finished ICKP end
```

So letting `scheduleFlush` continue to derive `lastCkp` from `MaxIncrementalCheckpoint()` is essentially the same lower bound required by the bounded flush.

One more concern: if bounded flush keeps failing because of `PrepareCompact` or scope conflicts, always replacing the normal cron entry can starve newer dirty data after the checkpoint end, delaying normal memory release. Please add a simple rotation/backoff mechanism, for example allowing one normal cron flush after N consecutive bounded flush attempts, or using a time/count threshold before falling back to the normal merged dirty tree for one round.

This should allow us to remove `flushMode`, `checkpointAge`, bounded request splitting, and the bounded-specific branch in `checkFlushConditionAndFire`, keeping the change smaller and the semantics more localized: `triggerJob` decides which dirty range to flush this round, while scheduling remains the existing generic path.

jiangxinmeng1:
Updated in jiangxinmeng1/matrixone@2a7bf385a4.

Changes for the latest review:
- Moved the stalled ICKP bounded flush decision back into triggerJob. The downstream request/schedule/fire path now uses the existing generic flush flow again.
- Removed flushMode propagation, checkpointAge on FlushRequest, bounded request splitting, and the bounded-specific schedule/checkFlushCondition branches.
- The cron round now builds the normal merged dirty entry first, then replaces it with the bounded [intent.start, intent.end] dirty tree only when the pending ICKP is stalled and has not passed flushChecked.
- Kept lastCkp derived from MaxIncrementalCheckpoint() in scheduleFlush.
- Added a simple starvation guard: after 12 consecutive bounded selections, one normal cron entry is allowed through and the bounded attempt counter resets.
- Updated tests for the stable threshold boundary, bounded scan range/entry construction, and fallback rotation behavior.

Local checks:
- git diff --check: passed
- targeted go test command still does not reach tests locally because this checkout fails to build github.com/unum-cloud/usearch/golang against the installed CGO headers/libs (missing usearch_* symbols).

XuPeng-SH (changes requested):
I found one blocking regression in the latest revision.

The new bounded-checkpoint path now replaces the normal merged dirty tree in `triggerJob` instead of running alongside it. Concretely, `entry := flusher.sourcer.GetAndRefreshMerged()` is immediately overwritten by `pickFlushEntry(...)`, and only that single tree is enqueued. That means while a stalled checkpoint exists, 12 out of every 13 cron rounds only expose tables inside the old checkpoint window to `scheduleFlush`, `collectTableMemUsage`, and `checkFlushConditionAndFire`.

The unhappy path is that dirty tables created after `intent.end` become invisible to ordinary background flushing and pressure-based flushing for most rounds. Under sustained ingest, this can hide the real memory pressure of newer dirty data and let normal flushing lag behind badly while the bounded checkpoint keeps retrying.

I think the bounded checkpoint flush needs to be added as prioritized work without replacing the normal merged dirty tree for that round. At minimum, both trees need to remain visible to the scheduler so regular pressure-based flushing still happens every cycle.

aptend (approved):

XuPeng-SH (changes requested):
I re-reviewed the current head (`86ca1e14bc90d3a20b7a30973c1457f84d73b9d6`) and still found blocking issues.

1. The stalled bounded-flush trigger is still misaligned with the point where checkpoint scheduling is already blocked. In this revision, `stalledCheckpointFlushAge` is `5 * time.Minute`, but the checkpoint runner treats pending intents as old/non-extendable after `2 * time.Minute`. That means a stuck pending incremental checkpoint can already block newer checkpoint progress for about 3 extra minutes before the new bounded-flush prioritization even starts helping. This should be aligned with the same blocking threshold (or derived from the same config), otherwise the recovery path starts materially too late.

2. The current regression coverage is still not strong enough around the updated `triggerJob` control flow. The helper pieces are tested, but there is still no direct test for the unhappy path where `triggerJob` decides to take the stalled-checkpoint route, then the bounded entry build returns `nil` / state changes under it, and the code must correctly fall back to the normal merged dirty entry scheduling. That fallback path is now part of the core control flow and should be pinned directly.

Because the first issue affects when the recovery path becomes effective for blocked checkpoints, I am requesting changes.

XuPeng-SH (changes requested):
I re-reviewed the current newest head (`d7041008f8e085d2634d6055967f84ee7e1db4fa`). The earlier trigger-threshold alignment issue is fixed, and there is now a direct fallback test for the nil-bounded-entry path. But I still found one blocking issue.

The bounded stalled-checkpoint flush is still enqueued as an ordinary non-force cron request, and `onFlushRequest` merges all non-force requests together. Because the queue batches multiple pending items into one callback, a bounded request for `[intent.start, intent.end]` can still be batched together with a later normal cron request and widened back into the full normal dirty range before scheduling.

That means the bounded scope is not actually preserved end-to-end under backlog, which is exactly the unhappy path this PR is trying to fix: once widened again, post-checkpoint-end objects can still compete with or block the old checkpoint range.

I think bounded checkpoint flushes need to remain distinct from normal cron flushes all the way through queueing and `onFlushRequest`, so they cannot be coalesced back into the ordinary cron tree.

Separately, the latest tests still do not cover the successful `triggerJob` bounded-entry path itself (non-empty bounded scan, bounded entry enqueued, and scheduling driven by the bounded end TS), which would be worth adding once the control flow is finalized.

XuPeng-SH:
I dug deeper and found a more concrete implementation issue beyond the earlier test-gap concern. The new bounded checkpoint requests are enqueued separately, but when they are executed they still go through scheduleFlush(entry, false) and therefore through the normal flushReady() heuristics. That means a stalled checkpoint recovery pass can enqueue and process the bounded request, yet still decide not to flush anything for a small or low-pressure table because its normal flush deadline has not arrived yet. In other words, the recovery path can still remain stuck even after it triggers. The stalled-checkpoint trigger starts after 2 minutes, while normal table flush deadlines are initialized/reset to roughly 5 times flushInterval, so this bounded recovery path can start well before the targeted tables become flush-eligible under the normal policy. I think the bounded stalled-checkpoint mode needs to bypass those normal readiness gates, or have its own dedicated recovery mode, otherwise the fix can still fail in the exact low-pressure cases it is supposed to unblock.

XuPeng-SH:
I checked the latest head and found a new, more concrete issue. The bounded checkpoint path now does bypass flushReady(), which fixes the earlier concern, but it also shares the same branch as force flush and still calls ResetDeadline(flusher.flushInterval) after a successful bounded flush. That appears to contradict the PR's stated design that bounded checkpoint flushes should not reset the normal table flush deadline.\n\nConcretely, flushScheduleCheckpointBounded returns true from bypassFlushReady(), enters the same unconditional branch as force flush, and then ResetDeadline is called unconditionally. I also do not see a test protecting the intended 'bounded flush does not reset deadline' semantics. So from my perspective the latest patch trades one issue for another: it now forces progress, but may silently perturb the normal flush schedule in exactly the way the PR description says it should avoid.

XuPeng-SH (approved):
Re-reviewed the current head (`62ca6d848fcb369b3916ad7737675d28661e3d7c`). I do not find any remaining blocking code-review issue.

What I checked:
- The stalled checkpoint trigger age is now aligned with `checkpointIntentOldAge`.
- The bounded path scans `[intent.start, intent.end]` and enqueues a distinct `flushRequestCheckpointBounded` request.
- Queue batching no longer merges bounded checkpoint requests into cron requests.
- Bounded scheduling bypasses `flushReady()` so low-pressure / not-yet-deadline tables can still make checkpoint recovery progress.
- Bounded scheduling does not reset normal table flush deadlines.
- The trigger fallback path and successful bounded enqueue path are covered by tests.
- The alternating bounded/normal selection avoids starving normal cron work while the old checkpoint remains stalled.

Local verification:
```bash
git diff --check upstream/main...HEAD
CGO_CFLAGS='-I/home/xupeng/matrixone/thirdparties/install/include' \
CGO_LDFLAGS='-L/home/xupeng/matrixone/thirdparties/install/lib -L/home/xupeng/matrixone/cgo' \
LD_LIBRARY_PATH='/home/xupeng/matrixone/thirdparties/install/lib:/home/xupeng/matrixone/cgo' \
GOCACHE=/tmp/matrixone-pr24901-gocache \
go test -mod=mod ./pkg/vm/engine/tae/db/checkpoint -count=1
```

Note: the latest CI still has one upgrade-target job failing while starting the head binary. I could not get detailed logs beyond the failing step, and I did not find evidence in this diff tying that failure to the checkpoint flusher change, so I am approving the code changes.

Issue #24896, opened 2026-06-09
[Bug]: CN can keep stale lock table bind after TN allocator restart until COMMIT fails


iamlinjunhong:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Searches for `"lock table bind changed"`, `ErrLockTableBindChanged`, `lockservice allocator bind epoch`, and `stale lock table bind allocator` found several historical closed issues, but I did not find an open issue for this specific allocator-epoch stale-bind scenario.

### Branch Name

- Affected production image: `registry.jst.com.cn/moi/matrixone:3.0.0-5e4210029-hotfix-20260603`
- Affected local branch/repro: `3.0-dev`

### Commit ID

- Production image commit: `5e4210029`
- Local reproduction commit: `1b4aa0a7a458d69c834c1292a9628056eed86cae`

### Other Environment Information

- Deployment mode: distributed MO with CN and TN/DN pods
- Component: `lockservice`, `lockTableAllocator`, pessimistic transaction commit validation
- Observed during a data import workflow. The failing SQL observed in MO logs was `COMMIT`, not the preceding DML.

### Actual Behavior

A CN can keep local lock table binds created by an older TN/DN lock allocator epoch after the TN/DN allocator pod is rebuilt. If the CN does not restart, and it later accesses another table first, the new allocator registers the same CN service ID again. After that, normal keepalive returns `OK=true`, but it does not validate the CN's already cached table binds.

As a result, stale table binds can remain on the CN for a long time and are only detected at transaction commit time by `allocator.Valid()`. The user-visible result is:

```text
Error 20702 (HY000): lock table bind changed
```

Evidence from one incident:

```text
old allocator start:
  2026-06-05 15:13:42 +0800
  version=1780643622856501201

old binds assigned:
  2026-06-08 14:07:25 +0800
  table=272592 / 272593
  bind version=1780643622856501202

new allocator start after DN pod rebuild:
  2026-06-08 14:11:38 +0800
  version=1780899098803397307

commit failure:
  2026-06-09 09:39:55 +0800
  error=lock table bind changed
  allocator current=0-0(0)--0
  received old bind=0-272592(272592)-...-1780643622856501202
  received old bind=0-272593(272593)-...-1780643622856501202
```

The important part is that the later keepalive can be healthy (`OK=true`) even though the CN still has stale per-table binds from the old allocator epoch.

### Expected Behavior

[...]

PR #24903, opened 2026-06-09

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24896 


aptend (approved):
**Review Result**

**What I Checked**

- `GetBind` and `KeepLockTableBind` now return the allocator epoch to CN.
- CN stale-bind purge uses the existing `LockTable.Version` semantics consistently: binds with `Version < observed allocator epoch` are old allocator state and can be removed.
- The added tests cover get-bind refresh, keepalive failure/epoch refresh, concurrent observation, remote/proxy bind purge, and waiter cleanup.

**Local Verification**

```bash
git diff --check HEAD^ HEAD
go test ./pkg/lockservice -run 'Test(CommitDetectsStaleLocalBindAfterAllocatorRestart|AllocatorVersionZeroKeepsLocalBinds|AllocatorObserverDoesNotPurgeSameEpochBindVersions|GetBindAllowsRegressedAllocatorVersionWithoutPurging|KeepaliveEpochPurgeKeepsGroupMovePop|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind)$'
```

Note: I started `go test ./pkg/lockservice`, but stopped it after it ran for more than 5 minutes without output and then verified the PR-specific test set above.

XuPeng-SH (changes requested):
I understand the bug this PR is trying to fix, but I found one substantive correctness issue in the epoch-fencing design itself.

The allocator epoch is initialized from `time.Now().UnixNano()` in `pkg/lockservice/lock_table_allocator.go`, while the observer side in `service.go` explicitly ignores any non-increasing allocator version (`observedVersion < oldVersion` or `== oldVersion`). That means a real allocator restart can fail to purge stale cached binds if the new process comes up with the same or a lower wall-clock timestamp (for example after a clock step backward / NTP correction / VM time skew). In that case `getLockTableWithCreate()` keeps returning the stale entry from `tableGroups` without ever re-contacting the allocator, which is exactly the class of stale-bind problem this PR is trying to fence off.

What makes this higher confidence is that the new test `TestGetBindAllowsRegressedAllocatorVersionWithoutPurging` explicitly locks in that behavior. I think the fix needs an epoch source that is guaranteed monotonic/unique across allocator restarts (persisted term/counter, restart generation, etc.), or the observer must still invalidate cache on allocator replacement even when the observed epoch is non-increasing.

iamlinjunhong:
fixed the lockservice allocator-restart fencing issue across d4-24896, d3-24896, and m-24896.
Main fixes:
Added an allocator instance ID to lock table binds and allocator responses, so CNs can detect allocator replacement even when wall-clock-based versions regress or stay the same.
Updated CN-side stale bind detection to use allocator identity, not only monotonic allocator version.
Made allocator replacement purge go through normal bind-change fencing semantics, so active transactions get ErrLockTableBindChanged instead of lower-level closed/not-found errors.
Covered the combined keepalive unhappy path where OK=false and a newer/different allocator state arrive in the same response.
Ensured stale local, remote, and proxy lock tables are purged while current allocator-instance binds are preserved.
Added/updated regression tests for allocator ID changes, regressed allocator versions, keepalive cleanup, active txn fencing, remote/proxy stale bind cleanup, and concurrent observer paths.
For old branches, fixed remote-lock timeout handling to avoid recording phantom locks when the request likely never reached the remote owner.
Updated branch-specific tests whose old expectations conflicted with the new bind-change fencing behavior.

XuPeng-SH (changes requested):
I found one blocking issue in the current head.

In `remoteLockTable.lock`, the new `skipTrackLockOnError` path skips `txn.lockAdded(...)` for `context.Canceled`, `context.DeadlineExceeded`, and `ErrRPCTimeout`. But those errors do not prove the request never reached the remote owner — they can also happen after the remote side already acquired the lock and before the response got back.

In that case, the local txn no longer records the remote lock in `txn.lockHolders`, so the normal `Unlock()` / `txn.close()` path will never send the remote unlock. The code comment says the lock will eventually be cleaned by the keep-remote/orphan timeout, but that is still a real behavioral regression: a lock that should be released immediately at txn end can now remain blocking until timeout-based cleanup fires.

So I do not think it is safe to skip local bookkeeping for timeout/cancel based only on the client-side error class. This should only be done when the code can prove the request was never observed remotely, or else it needs an explicit reconciliation path before deciding not to record the remote lock.

XuPeng-SH (changes requested):
I found two blocking correctness issues in the current head (`6e8ca4b`).

1. `pkg/lockservice/service.go:786-814` / `pkg/lockservice/lock_table_keeper.go:229-236`: the keepalive `OK=false` path still removes this service's local binds via `holders.removeWithFilter(..., closeReasonKeepBindFailed)` without going through `removeLockTablesWithFence` / `fenceByBindChanged`. That means an active txn that already touched the old bind is not marked `bindChanged`. After the purge, the same txn can fetch a fresh bind and acquire more locks successfully instead of getting `ErrLockTableBindChanged`. I reproduced this locally with a temporary regression test: lock table in txn T, reset allocator state so keepalive returns `OK=false`, call `doKeepLockTableBind`, then call `Lock` again on the same table and same txn. Current code returns `nil`, leaving T with old `tableBinds` plus new locks, so the failure is deferred to commit/valid/unlock instead of being fenced at the bind change.

2. `pkg/lockservice/service.go:689-741` plus the unconditional `tableGroups.set` in `getLockTableWithCreate`: `allocatorChanged` treats any different non-empty `AllocatorID` as the next allocator. A delayed response from the old allocator can arrive after this CN has already accepted a newer allocator ID; current code will accept that old ID as another replacement, purge current binds, update `lastAllocatorID` back to the old allocator, and then `getLockTableWithCreate` can install the stale bind from that old response. UUID distinguishes allocator processes, but it does not order them. The stale-response unhappy path needs a monotonic term/generation, or a guard that prevents an already superseded allocator response from rolling CN state backward.

Please also audit identity consistency around allocator IDs. `LockTable.Changed` now includes `AllocatorID`, but `getRemoteLockBindKey` and `LockTable.Equal` still omit it, so remote bind heartbeat/stale cleanup can conflate different allocator instances when service/table/version match.

Local verification on the PR-specific tests still passes:

```bash
GOCACHE=/tmp/matrixone-pr24903-gocache go test ./pkg/lockservice -run 'Test(CommitDetectsStaleLocalBindAfterAllocatorRestart|AllocatorVersionZeroKeepsLocalBinds|AllocatorObserverDoesNotPurgeSameEpochBindVersions|GetBindPurgesStaleBindWhenAllocatorIDChangesWithRegressedVersion|KeepaliveEpochPurgeKeepsGroupMovePop|KeepaliveOKFalseWithNewAllocatorVersionPurgesStaleBinds|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind|LockRemoteWithContextTimeoutTracksLockForUnlock|LockWithBindTimeout|LockWithBindNotFound|Issue3538)$'
```

But the temporary OK=false active-txn regression described above fails on current head, so I am requesting changes.

iamlinjunhong:
> I found two blocking correctness issues in the current head (`6e8ca4b`).
> 
> 1. `pkg/lockservice/service.go:786-814` / `pkg/lockservice/lock_table_keeper.go:229-236`: the keepalive `OK=false` path still removes this service's local binds via `holders.removeWithFilter(..., closeReasonKeepBindFailed)` without going through `removeLockTablesWithFence` / `fenceByBindChanged`. That means an active txn that already touched the old bind is not marked `bindChanged`. After the purge, the same txn can fetch a fresh bind and acquire more locks successfully instead of getting `ErrLockTableBindChanged`. I reproduced this locally with a temporary regression test: lock table in txn T, reset allocator state so keepalive returns `OK=false`, call `doKeepLockTableBind`, then call `Lock` again on the same table and same txn. Current code returns `nil`, leaving T with old `tableBinds` plus new locks, so the failure is deferred to commit/valid/unlock instead of being fenced at the bind change.
> 2. `pkg/lockservice/service.go:689-741` plus the unconditional `tableGroups.set` in `getLockTableWithCreate`: `allocatorChanged` treats any different non-empty `AllocatorID` as the next allocator. A delayed response from the old allocator can arrive after this CN has already accepted a newer allocator ID; current code will accept that old ID as another replacement, purge current binds, update `lastAllocatorID` back to the old allocator, and then `getLockTableWithCreate` can install the stale bind from that old response. UUID distinguishes allocator processes, but it does not order them. The stale-response unhappy path needs a monotonic term/generation, or a guard that prevents an already superseded allocator response from rolling CN state backward.
> 
> Please also audit identity consistency around allocator IDs. `LockTable.Changed` now includes `AllocatorID`, but `getRemoteLockBindKey` and `LockTable.Equal` still omit it, so remote bind heartbeat/stale cleanup can conflate different allocator instances when service/table/version match.
> 
> Local verification on the PR-specific tests still passes:
> 
> ```shell
> GOCACHE=/tmp/matrixone-pr24903-gocache go test ./pkg/lockservice -run 'Test(CommitDetectsStaleLocalBindAfterAllocatorRestart|AllocatorVersionZeroKeepsLocalBinds|AllocatorObserverDoesNotPurgeSameEpochBindVersions|GetBindPurgesStaleBindWhenAllocatorIDChangesWithRegressedVersion|KeepaliveEpochPurgeKeepsGroupMovePop|KeepaliveOKFalseWithNewAllocatorVersionPurgesStaleBinds|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind|LockRemoteWithContextTimeoutTracksLockForUnlock|LockWithBindTimeout|LockWithBindNotFound|Issue3538)$'
> ```
> 
> But the temporary OK=false active-txn regression described above fails on current head, so I am requesting changes.

fixed

XuPeng-SH (changes requested):
I re-checked the current head from correctness and unhappy-path angles. The earlier allocator identity/fencing issues look materially improved, but I found one remaining blocker.

Blocker: `handleForwardLock` can deadlock while waiting for an in-flight bind allocation.

`pkg/lockservice/service_remote.go:292` now acquires `s.bindChangeMu.RLock()` before calling `s.getLockTable()`. But `getLockTable()` can block in `waitLockTableBind()` when another goroutine is currently allocating the same table bind (`pkg/lockservice/service.go:612-616`). That allocating goroutine only closes the wait channel after `publishLockTableBindFromAllocator()` returns, and the publish/allocator-observe path needs `s.bindChangeMu.Lock()` (`pkg/lockservice/service.go:656` and `pkg/lockservice/service.go:816`).

So the unhappy path is:

1. goroutine A starts first access for table T and registers `s.mu.allocating[T]`;
2. goroutine B enters `handleForwardLock` for T, takes `bindChangeMu.RLock`, then waits for A's allocation channel in `getLockTable()`;
3. goroutine A receives the allocator response and tries to publish the bind, but blocks on `bindChangeMu.Lock`;
4. A cannot close the allocation channel, while B keeps holding the read lock and waits forever.

This can hang forwarded lock RPCs during ordinary concurrent first-access/rebind timing; it does not require allocator restart.

Suggested fix: do not hold `bindChangeMu` while calling/waiting in `getLockTable()`. Mirror `handleRemoteLock`: fetch/wait for the lock table first, then take `bindChangeMu.RLock()` and re-validate that the current bind still matches before touching the txn or reporting success. Please also add a regression test that pauses one goroutine after the allocating channel is registered but before publish, then runs a concurrent forwarded-lock/getLockTable waiter and verifies publish can still complete.

Local checks passed:

- `git merge-tree --write-tree HEAD origin/3.0-dev`
- targeted `go test ./pkg/lockservice -run 'Test(GetBindPurgesStaleBindWhenAllocatorIDChangesWithRegressedVersion|KeepaliveEpochPurgeKeepsGroupMovePop|KeepaliveOKFalseWithNewAllocatorVersionPurgesStaleBinds|KeepaliveOKFalseFencesActiveTxn|AllocatorObserverRejectsSupersededGetBindResponse|AllocatorPublishRejectsStaleBindAfterNewAllocatorObserved|AllocatorPublishRejectsOverwriteAfterConcurrentBindChanged|AllocatorObserverBoundsSupersededAllocatorIDs|AllocatorIDDistinguishesBindIdentity|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind|RemoteBindRefreshRejectsSupersededAllocatorBind|RemoteRetryBackoffCapped|GetLockRemoteWithRetry|LockRemoteWithContextTimeoutTracksLockForUnlock|LockWithBindTimeout|LockWithBindNotFound|Issue3538)$' -count=1`

iamlinjunhong:
> I re-checked the current head from correctness and unhappy-path angles. The earlier allocator identity/fencing issues look materially improved, but I found one remaining blocker.
> 
> Blocker: `handleForwardLock` can deadlock while waiting for an in-flight bind allocation.
> 
> `pkg/lockservice/service_remote.go:292` now acquires `s.bindChangeMu.RLock()` before calling `s.getLockTable()`. But `getLockTable()` can block in `waitLockTableBind()` when another goroutine is currently allocating the same table bind (`pkg/lockservice/service.go:612-616`). That allocating goroutine only closes the wait channel after `publishLockTableBindFromAllocator()` returns, and the publish/allocator-observe path needs `s.bindChangeMu.Lock()` (`pkg/lockservice/service.go:656` and `pkg/lockservice/service.go:816`).
> 
> So the unhappy path is:
> 
> 1. goroutine A starts first access for table T and registers `s.mu.allocating[T]`;
> 2. goroutine B enters `handleForwardLock` for T, takes `bindChangeMu.RLock`, then waits for A's allocation channel in `getLockTable()`;
> 3. goroutine A receives the allocator response and tries to publish the bind, but blocks on `bindChangeMu.Lock`;
> 4. A cannot close the allocation channel, while B keeps holding the read lock and waits forever.
> 
> This can hang forwarded lock RPCs during ordinary concurrent first-access/rebind timing; it does not require allocator restart.
> 
> Suggested fix: do not hold `bindChangeMu` while calling/waiting in `getLockTable()`. Mirror `handleRemoteLock`: fetch/wait for the lock table first, then take `bindChangeMu.RLock()` and re-validate that the current bind still matches before touching the txn or reporting success. Please also add a regression test that pauses one goroutine after the allocating channel is registered but before publish, then runs a concurrent forwarded-lock/getLockTable waiter and verifies publish can still complete.
> 
> Local checks passed:
> 
> * `git diff --check origin/3.0-dev...HEAD`
> * `git merge-tree --write-tree HEAD origin/3.0-dev`
> * targeted `go test ./pkg/lockservice -run 'Test(GetBindPurgesStaleBindWhenAllocatorIDChangesWithRegressedVersion|KeepaliveEpochPurgeKeepsGroupMovePop|KeepaliveOKFalseWithNewAllocatorVersionPurgesStaleBinds|KeepaliveOKFalseFencesActiveTxn|AllocatorObserverRejectsSupersededGetBindResponse|AllocatorPublishRejectsStaleBindAfterNewAllocatorObserved|AllocatorPublishRejectsOverwriteAfterConcurrentBindChanged|AllocatorObserverBoundsSupersededAllocatorIDs|AllocatorIDDistinguishesBindIdentity|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind|RemoteBindRefreshRejectsSupersededAllocatorBind|RemoteRetryBackoffCapped|GetLockRemoteWithRetry|LockRemoteWithContextTimeoutTracksLockForUnlock|LockWithBindTimeout|LockWithBindNotFound|Issue3538)$' -count=1`

fixed

XuPeng-SH (changes requested):
I still think this needs changes.

The remaining blocker is the new unconditional remote-lock bookkeeping on every `Send` error in `pkg/lockservice/lock_table_remote.go`.

After this patch, the error path at `lock_table_remote.go:176-180` always does:

```go
_ = txn.lockAdded(l.bind.Group, l.bind, rows, l.logger)
```

That is only safe for "request may already have reached the remote owner" failures. But `morpc` also has definite **pre-send** failures where the request never leaves the client:

- `pkg/common/morpc/backend.go:366-373` returns `f.send.Ctx.Err()` if the request cannot even enter the write queue.
- `pkg/common/morpc/backend.go:905-913` returns `ErrRPCTimeout` during connect timeout before any request is written.

Once such a purely local failure is recorded as a held remote lock, `txn.close()` will later always try to remote-unlock it (`pkg/lockservice/txn.go:174-223`). But `remoteLockTable.unlock()` retries forever while the bind stays unchanged (`pkg/lockservice/lock_table_remote.go:224-250,361-406`). So a caller cancel / local deadline / connect-timeout that never sent the RPC can regress into an unbounded commit/rollback hang.

I think we need to keep the old "do not track definitely-not-sent errors" behavior, or add an explicit transport signal that distinguishes "may have been sent" from "definitely not sent" before deciding to retain remote-lock bookkeeping.

iamlinjunhong:
> I still think this needs changes.
> 
> The remaining blocker is the new unconditional remote-lock bookkeeping on every `Send` error in `pkg/lockservice/lock_table_remote.go`.
> 
> After this patch, the error path at `lock_table_remote.go:176-180` always does:
> 
> ```go
> _ = txn.lockAdded(l.bind.Group, l.bind, rows, l.logger)
> ```
> 
> That is only safe for "request may already have reached the remote owner" failures. But `morpc` also has definite **pre-send** failures where the request never leaves the client:
> 
> * `pkg/common/morpc/backend.go:366-373` returns `f.send.Ctx.Err()` if the request cannot even enter the write queue.
> * `pkg/common/morpc/backend.go:905-913` returns `ErrRPCTimeout` during connect timeout before any request is written.
> 
> Once such a purely local failure is recorded as a held remote lock, `txn.close()` will later always try to remote-unlock it (`pkg/lockservice/txn.go:174-223`). But `remoteLockTable.unlock()` retries forever while the bind stays unchanged (`pkg/lockservice/lock_table_remote.go:224-250,361-406`). So a caller cancel / local deadline / connect-timeout that never sent the RPC can regress into an unbounded commit/rollback hang.
> 
> I think we need to keep the old "do not track definitely-not-sent errors" behavior, or add an explicit transport signal that distinguishes "may have been sent" from "definitely not sent" before deciding to retain remote-lock bookkeeping.

Thanks for the careful review. I agree that unconditional bookkeeping on every Send error was too broad.

I updated the fix to distinguish definite not-sent failures from ambiguous post-send failures. client.Send now wraps only AsyncSend failures where no Future was returned, which means the request did not enter the normal send/wait path. remoteLockTable.lock skips txn.lockAdded only for that not-sent marker and for ErrRPCTimeout, while still tracking ordinary context deadline/cancel errors returned after a Future exists so txn.close can issue the remote unlock when the request may already have reached the owner.

I also added regression coverage for the not-sent path to verify no stale remote lock is recorded and no remote unlock is attempted, plus a helper test that keeps post-send context deadline errors in the tracked category. The existing post-send timeout test still passes.

XuPeng-SH (changes requested):
I found one blocking regression: the current head fails the existing unstable-network lockservice test.

With the normal MatrixOne test environment, `go test -count=1 -timeout 240s ./pkg/lockservice` now times out in:

- `TestCannotHungWithUnstableNetwork`

The timeout happens after repeated logs like:

- `txn failed to unlock table on remote ... the connection has been disconnected`
- `failed to get bind ... the connection has been disconnected`

and the package eventually panics with `test timed out after 4m0s` while that test is still running.

That is a real correctness regression for the exact unhappy path this PR is trying to harden: under unstable/disconnecting network conditions, the lockservice must not hang. Even before getting into the finer stale-bind fencing details, the current implementation is not safe to merge while an existing regression test for network instability is timing out.

Given the changed logic in `remoteLockTable.lock()` around tracking remote locks on send errors (`shouldTrackRemoteLockOnSendError`) plus the new allocator-epoch fencing, I would first fix the unstable-network hang and make sure the package test passes again. The new coverage is broad overall, but it also still misses one high-value race test for a first lock whose bind changes mid-flight, so this area could use another unhappy-path regression once the hang is fixed.

iamlinjunhong:
> I found one blocking regression: the current head fails the existing unstable-network lockservice test.
> 
> With the normal MatrixOne test environment, `go test -count=1 -timeout 240s ./pkg/lockservice` now times out in:
> 
> * `TestCannotHungWithUnstableNetwork`
> 
> The timeout happens after repeated logs like:
> 
> * `txn failed to unlock table on remote ... the connection has been disconnected`
> * `failed to get bind ... the connection has been disconnected`
> 
> and the package eventually panics with `test timed out after 4m0s` while that test is still running.
> 
> That is a real correctness regression for the exact unhappy path this PR is trying to harden: under unstable/disconnecting network conditions, the lockservice must not hang. Even before getting into the finer stale-bind fencing details, the current implementation is not safe to merge while an existing regression test for network instability is timing out.
> 
> Given the changed logic in `remoteLockTable.lock()` around tracking remote locks on send errors (`shouldTrackRemoteLockOnSendError`) plus the new allocator-epoch fencing, I would first fix the unstable-network hang and make sure the package test passes again. The new coverage is broad overall, but it also still misses one high-value race test for a first lock whose bind changes mid-flight, so this area could use another unhappy-path regression once the hang is fixed.

fixed

XuPeng-SH (changes requested):

The latest head fixes the visible unstable-network hang and CI is green, but the remote-lock bookkeeping rule is now too broad in the other direction.

`shouldTrackRemoteLockOnSendError` treats `ErrBackendClosed` and `ErrBackendCannotConnect` as definite not-sent failures and skips `txn.lockAdded`. That is safe only when the request never completed the write path. However, in morpc `ErrBackendClosed` can also be delivered to a Future after the request has already been written and the client is waiting for the response: the read loop can schedule a connection reset, and `makeAllWaitingFutureFailed` fails all futures whose `messageSent(nil)` has already completed. The test-only `disconnectAfterRead` path also turns that response-side disconnect into `backendClosed`.

In that post-send/unknown case, the remote owner may already have acquired the lock. If the client skips local bookkeeping, `txn.close()` will not send a remote unlock, so release is delayed until remote/orphan timeout instead of happening at transaction end. This avoids the previous hang, but it creates a correctness/availability regression: conflicting transactions can be blocked by a lock whose owner-side acquisition is possible but whose client-side holder was intentionally forgotten.

I think the fix should distinguish transport phase, not just error code:

- wrap/mark only definite pre-send failures as `lockRequestNotSentError`;
- skip remote-lock bookkeeping only for that marker and other provably not-written cases;
- keep tracking post-send or unknown failures, including response-side disconnects, while bounding remote unlock retry so commit/rollback cannot hang forever.

This should also have a regression test where the lock request is written, the owner may acquire it, and the client connection fails before the response is observed; that path should still record the holder and attempt remote unlock on txn close.

iamlinjunhong:
> I think this still needs changes.
> 
> The latest head fixes the visible unstable-network hang and CI is green, but the remote-lock bookkeeping rule is now too broad in the other direction.
> 
> `shouldTrackRemoteLockOnSendError` treats `ErrBackendClosed` and `ErrBackendCannotConnect` as definite not-sent failures and skips `txn.lockAdded`. That is safe only when the request never completed the write path. However, in morpc `ErrBackendClosed` can also be delivered to a Future after the request has already been written and the client is waiting for the response: the read loop can schedule a connection reset, and `makeAllWaitingFutureFailed` fails all futures whose `messageSent(nil)` has already completed. The test-only `disconnectAfterRead` path also turns that response-side disconnect into `backendClosed`.
> 
> In that post-send/unknown case, the remote owner may already have acquired the lock. If the client skips local bookkeeping, `txn.close()` will not send a remote unlock, so release is delayed until remote/orphan timeout instead of happening at transaction end. This avoids the previous hang, but it creates a correctness/availability regression: conflicting transactions can be blocked by a lock whose owner-side acquisition is possible but whose client-side holder was intentionally forgotten.
> 
> I think the fix should distinguish transport phase, not just error code:
> 
> * wrap/mark only definite pre-send failures as `lockRequestNotSentError`;
> * skip remote-lock bookkeeping only for that marker and other provably not-written cases;
> * keep tracking post-send or unknown failures, including response-side disconnects, while bounding remote unlock retry so commit/rollback cannot hang forever.
> 
> This should also have a regression test where the lock request is written, the owner may acquire it, and the client connection fails before the response is observed; that path should still record the holder and attempt remote unlock on txn close.

Fixed. Bare ErrBackendClosed / ErrBackendCannotConnect are no longer treated as definitely-not-sent failures. We now skip remote-lock bookkeeping only for explicit not-sent markers and the existing connect-timeout case; ambiguous response-side disconnects keep the holder so txn close can send remote unlock. I also added a regression test covering a lock request that may have reached the owner before the client observes BackendClosed.

XuPeng-SH (approved):
Reviewed latest head c3e49a82366304d146221d289a0d081f4c9ae8e8.

The previous remote-lock bookkeeping blocker is addressed: bare BackendClosed/BackendCannotConnect are now treated as ambiguous post-send failures and keep txn bookkeeping, while explicit not-sent markers and connect-timeout skip bookkeeping. Remote unlock retry is bounded by removeLockTimeout, so phantom tracked locks no longer create an unbounded close/rollback hang.

I also rechecked allocator replacement fencing, bind-change lock ordering, forwarded-lock first-bind wait, active txn bind-intent fencing, remote/proxy bind identity, and compatibility with missing allocator metadata. I do not see a remaining correctness blocker.

Local targeted verification passed:

`go test -count=1 -timeout 240s ./pkg/lockservice -run 'Test(HandleForwardLockDoesNotHoldBindChangeLockWhileWaitingForBind|LockRemoteWithNotSentErrorSkipsRemoteTracking|LockRemoteWithResponseDisconnectTracksForUnlock|UnlockRemoteStopsRetryAfterDisconnectTimeout|ShouldTrackRemoteLockOnSendError|CannotHungWithUnstableNetwork|GetBindPurgesStaleBindWhenAllocatorIDChangesWithRegressedVersion|KeepaliveOKFalseFencesActiveTxn|AllocatorObserverRejectsSupersededGetBindResponse|AllocatorPublishRejectsStaleBindAfterNewAllocatorObserved|AllocatorPublishRejectsOverwriteAfterConcurrentBindChanged|AllocatorIDDistinguishesBindIdentity|BindChangedBeforeLockSuccessReturnsBindChanged)$'`

CI is green on the 3.0 jobs.

ouyuanning (approved):

PR #24905, opened 2026-06-09

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24896 


aptend (approved):

XuPeng-SH (changes requested):
I did not confirm a blocking logic bug in the implementation itself, but there is still one important unhappy-path regression gap that should be covered before this is approved.

The new fix combines two recovery actions off the same keepalive failure response: local-bind cleanup when `OK=false`, and stale-bind purge when the allocator epoch increases. Current tests cover allocator-epoch purge and `OK=false` handling separately, but I do not see a test that drives a single keepalive response with both `OK=false` and a newer `AllocatorVersion`, then verifies that:
- stale local binds are removed
- stale remote/proxy binds from the old epoch are removed
- `lastAllocatorVersion` advances
- current-epoch binds are preserved

That combined failure path is the core unhappy-path this PR is hardening after allocator restart/failover. Without one regression test for the combined case, a future refactor could short-circuit one branch while the other still fires and leave stale bindings behind. Please add a focused test around `doKeepLockTableBind` / keepalive handling for this case.

iamlinjunhong:
> I did not confirm a blocking logic bug in the implementation itself, but there is still one important unhappy-path regression gap that should be covered before this is approved.
> 
> The new fix combines two recovery actions off the same keepalive failure response: local-bind cleanup when `OK=false`, and stale-bind purge when the allocator epoch increases. Current tests cover allocator-epoch purge and `OK=false` handling separately, but I do not see a test that drives a single keepalive response with both `OK=false` and a newer `AllocatorVersion`, then verifies that:
> 
> * stale local binds are removed
> * stale remote/proxy binds from the old epoch are removed
> * `lastAllocatorVersion` advances
> * current-epoch binds are preserved
> 
> That combined failure path is the core unhappy-path this PR is hardening after allocator restart/failover. Without one regression test for the combined case, a future refactor could short-circuit one branch while the other still fires and leave stale bindings behind. Please add a focused test around `doKeepLockTableBind` / keepalive handling for this case.

fixed

iamlinjunhong:
> I did not confirm a blocking logic bug in the implementation itself, but there is still one important unhappy-path regression gap that should be covered before this is approved.
> 
> The new fix combines two recovery actions off the same keepalive failure response: local-bind cleanup when `OK=false`, and stale-bind purge when the allocator epoch increases. Current tests cover allocator-epoch purge and `OK=false` handling separately, but I do not see a test that drives a single keepalive response with both `OK=false` and a newer `AllocatorVersion`, then verifies that:
> 
> * stale local binds are removed
> * stale remote/proxy binds from the old epoch are removed
> * `lastAllocatorVersion` advances
> * current-epoch binds are preserved
> 
> That combined failure path is the core unhappy-path this PR is hardening after allocator restart/failover. Without one regression test for the combined case, a future refactor could short-circuit one branch while the other still fires and leave stale bindings behind. Please add a focused test around `doKeepLockTableBind` / keepalive handling for this case.

fixed

XuPeng-SH (changes requested):
I re-checked the latest head. The combined keepalive unhappy-path coverage is better now, but I still see one remaining high-severity race in the allocator-epoch publish path.

In `getLockTableWithCreate()`, the allocator response is accepted via `observeAllocatorState(...)`, and only *after that* the code publishes the bind with `tableGroups.set(...)`.

Right now those two steps are not atomic with respect to later allocator-epoch observations:
- `observeAllocatorState(..., accepted)` can return success for an old-epoch bind
- another goroutine can then observe a newer allocator epoch, purge stale tables under `bindChangeMu`, and advance `lastAllocatorID/version`
- the first goroutine can still continue to `tableGroups.set(...)` and reinsert the stale bind it had already fetched

Once reinserted, later lookups compare against that same cached stale bind and can stop noticing the table is out of date; the remote side can even hand that stale bind back to peers via `resp.NewBind` from `getLocalLockTable()`.

You can see the gap in the current head:
- `service.go:623-639` accepts the allocator state, then publishes the bind later
- `service.go:754-764` purges stale tables under `bindChangeMu`
- `service_remote.go:509-512` will return the cached bind as authoritative if it differs from the requester’s bind

**Concrete suggestions**

1. Make allocator-state acceptance and `tableGroups.set(...)` atomic with respect to allocator-epoch updates.
2. If you don’t want one wider critical section, then re-check `lastAllocatorID/version` immediately before/after publishing and discard/retry if the allocator has advanced in the meantime.
3. Add one focused regression test for the exact interleaving: old-epoch `GetBind` response passes `observeAllocatorState`, newer epoch purges stale binds, old goroutine then tries to publish — and the stale bind must not reappear in `tableGroups`.

XuPeng-SH (changes requested):
I re-checked this PR from a correctness / unhappy-path angle. The direction is good, but I found two substantive issues that should be fixed before merge.

1. **Allocator fencing is still bypassed on remote bind refresh paths**
   - The new allocator epoch/instance fencing is only enforced in the normal publish path (`publishLockTableBindFromAllocator`).
   - But both remote refresh callers still do `new, _, err := getLockTableBind(...)` and then feed the bind straight into `handleBindChanged(...)`.
   - That means a stale `GetBind` reply from a superseded allocator can still re-install the very bind that the new allocator-state fencing was supposed to reject.
   - **Suggestion:** plumb `allocatorState` through these remote refresh paths too, and validate it via the same `observeAllocatorState...` / publish logic before mutating `tableGroups`.

2. **Delayed allocator replies can still overwrite a fresher cached bind**
   - `Lock()` now calls `getLockTableWithCreate()` before taking `bindChangeMu`.
   - If a concurrent `handleBindChanged` installs a newer bind while the allocator RPC is in flight, the delayed allocator result can still be published into `tableGroups`, and `lockTableHolder.set` will replace the current entry on any difference.
   - That lets an older bind close and overwrite a newer one.
   - **Suggestion:** before publishing allocator results, re-check the currently cached bind under `bindChangeMu` / current table state and refuse to downgrade an already-newer bind.

Given that this PR’s goal is to harden stale-bind fencing, these two remaining bypass/downgrade paths are blockers for me.

aunjgr (commented):
Superseded by #24906.

aunjgr (approved):
LGTM for the 4.0-dev backport.

iamlinjunhong:
> I re-checked the latest head. The combined keepalive unhappy-path coverage is better now, but I still see one remaining high-severity race in the allocator-epoch publish path.
> 
> In `getLockTableWithCreate()`, the allocator response is accepted via `observeAllocatorState(...)`, and only _after that_ the code publishes the bind with `tableGroups.set(...)`.
> 
> Right now those two steps are not atomic with respect to later allocator-epoch observations:
> 
> * `observeAllocatorState(..., accepted)` can return success for an old-epoch bind
> * another goroutine can then observe a newer allocator epoch, purge stale tables under `bindChangeMu`, and advance `lastAllocatorID/version`
> * the first goroutine can still continue to `tableGroups.set(...)` and reinsert the stale bind it had already fetched
> 
> Once reinserted, later lookups compare against that same cached stale bind and can stop noticing the table is out of date; the remote side can even hand that stale bind back to peers via `resp.NewBind` from `getLocalLockTable()`.
> 
> You can see the gap in the current head:
> 
> * `service.go:623-639` accepts the allocator state, then publishes the bind later
> * `service.go:754-764` purges stale tables under `bindChangeMu`
> * `service_remote.go:509-512` will return the cached bind as authoritative if it differs from the requester’s bind
> 
> **Concrete suggestions**
> 
> 1. Make allocator-state acceptance and `tableGroups.set(...)` atomic with respect to allocator-epoch updates.
> 2. If you don’t want one wider critical section, then re-check `lastAllocatorID/version` immediately before/after publishing and discard/retry if the allocator has advanced in the meantime.
> 3. Add one focused regression test for the exact interleaving: old-epoch `GetBind` response passes `observeAllocatorState`, newer epoch purges stale binds, old goroutine then tries to publish — and the stale bind must not reappear in `tableGroups`.

fixed

iamlinjunhong:
> I re-checked this PR from a correctness / unhappy-path angle. The direction is good, but I found two substantive issues that should be fixed before merge.
> 
> 1. **Allocator fencing is still bypassed on remote bind refresh paths**
>    
>    * The new allocator epoch/instance fencing is only enforced in the normal publish path (`publishLockTableBindFromAllocator`).
>    * But both remote refresh callers still do `new, _, err := getLockTableBind(...)` and then feed the bind straight into `handleBindChanged(...)`.
>    * That means a stale `GetBind` reply from a superseded allocator can still re-install the very bind that the new allocator-state fencing was supposed to reject.
>    * **Suggestion:** plumb `allocatorState` through these remote refresh paths too, and validate it via the same `observeAllocatorState...` / publish logic before mutating `tableGroups`.
> 2. **Delayed allocator replies can still overwrite a fresher cached bind**
>    
>    * `Lock()` now calls `getLockTableWithCreate()` before taking `bindChangeMu`.
>    * If a concurrent `handleBindChanged` installs a newer bind while the allocator RPC is in flight, the delayed allocator result can still be published into `tableGroups`, and `lockTableHolder.set` will replace the current entry on any difference.
>    * That lets an older bind close and overwrite a newer one.
>    * **Suggestion:** before publishing allocator results, re-check the currently cached bind under `bindChangeMu` / current table state and refuse to downgrade an already-newer bind.
> 
> Given that this PR’s goal is to harden stale-bind fencing, these two remaining bypass/downgrade paths are blockers for me.

fixed

XuPeng-SH (changes requested):
I found a few correctness issues that still need to be fixed before this should land.

1. `pkg/lockservice/lock_table_remote.go:maybeHandleBindChanged` still routes `resp.NewBind` through `handleBindChanged`, while the epoch-aware refresh paths use `handleBindChangedFromAllocator`. `service_remote.go` now returns `resp.NewBind` from `current.getBind()`, so the caller already has the allocator identity there. The result is that a txn can refresh table B from a new allocator epoch, retry B successfully, and still keep stale binds for table A until the next keepalive, which is exactly the stale-bind window this PR is trying to close.

   Suggested fix: make the `resp.NewBind` path go through the allocator-aware handler as well (or carry explicit allocator state alongside `NewBind` and call `handleBindChangedFromAllocator`) so observing one table from a new allocator epoch fences the rest of the stale binds immediately.

2. `pkg/lockservice/service.go:handleKeepBindFailed` still uses the old `holders.getVersion()` snapshot guard, but `lockTableHolders.set()` does not bump that generation when a bind is inserted/replaced inside an existing group. That means a raced `GetBind` can publish a fresh bind after `oldVersion` was captured and an `OK=false` keepalive can still sweep that new bind away and fence the in-flight txn incorrectly.

   Suggested fix: bump the holder generation on every bind publish/replace, or introduce a separate generation for bind publication/removal and compare that in the `OK=false` cleanup path.

3. `pkg/lockservice/lock_table_remote.go` now calls `txn.lockAdded(...)` even when `ctx.Err() != nil` before `Send`. In that branch we know the `Lock` RPC was never attempted, so this records a remote lock that cannot exist. In the common case the later remote `Unlock` is a harmless no-op, but it still creates unnecessary cleanup and can turn a purely local pre-send cancel into remote unlock retries if the owner is unavailable.

   Suggested fix: do not record the lock in the pre-send cancellation branch; only keep the local bookkeeping for ambiguous post-send failures.

The unit coverage here is otherwise pretty solid, and I am not blocking on missing BVT for this PR. These three correctness issues are the blockers for me.

iamlinjunhong:
> I found a few correctness issues that still need to be fixed before this should land.
> 
> 1. `pkg/lockservice/lock_table_remote.go:maybeHandleBindChanged` still routes `resp.NewBind` through `handleBindChanged`, while the epoch-aware refresh paths use `handleBindChangedFromAllocator`. `service_remote.go` now returns `resp.NewBind` from `current.getBind()`, so the caller already has the allocator identity there. The result is that a txn can refresh table B from a new allocator epoch, retry B successfully, and still keep stale binds for table A until the next keepalive, which is exactly the stale-bind window this PR is trying to close.
>    Suggested fix: make the `resp.NewBind` path go through the allocator-aware handler as well (or carry explicit allocator state alongside `NewBind` and call `handleBindChangedFromAllocator`) so observing one table from a new allocator epoch fences the rest of the stale binds immediately.
> 2. `pkg/lockservice/service.go:handleKeepBindFailed` still uses the old `holders.getVersion()` snapshot guard, but `lockTableHolders.set()` does not bump that generation when a bind is inserted/replaced inside an existing group. That means a raced `GetBind` can publish a fresh bind after `oldVersion` was captured and an `OK=false` keepalive can still sweep that new bind away and fence the in-flight txn incorrectly.
>    Suggested fix: bump the holder generation on every bind publish/replace, or introduce a separate generation for bind publication/removal and compare that in the `OK=false` cleanup path.
> 3. `pkg/lockservice/lock_table_remote.go` now calls `txn.lockAdded(...)` even when `ctx.Err() != nil` before `Send`. In that branch we know the `Lock` RPC was never attempted, so this records a remote lock that cannot exist. In the common case the later remote `Unlock` is a harmless no-op, but it still creates unnecessary cleanup and can turn a purely local pre-send cancel into remote unlock retries if the owner is unavailable.
>    Suggested fix: do not record the lock in the pre-send cancellation branch; only keep the local bookkeeping for ambiguous post-send failures.
> 
> The unit coverage here is otherwise pretty solid, and I am not blocking on missing BVT for this PR. These three correctness issues are the blockers for me.

fixed

XuPeng-SH (commented):
Re-reviewing the current head: most of the earlier blockers are addressed now, but I still think this should stay in request-changes state for one remaining correctness issue.

What looks fixed now:
- the holder-generation / `OK=false` keepalive sweep issue;
- the pre-send cancellation / phantom remote lock issue;
- the remote `GetBind` refresh paths bypassing allocator-aware fencing.

What still looks wrong:
- the `resp.NewBind` fast path is now allocator-aware, but it derives allocator state as `AllocatorID = newBind.AllocatorID` and `AllocatorVersion = newBind.Version`.

That is not safe because `LockTable.Version` is a bind version, not the allocator process epoch. The allocator initializes a bind from `l.version`, but later rebinds do `old.Version++`, so bind version and allocator epoch diverge. Feeding `newBind.Version` into `observeAllocatorStateLocked()` can therefore poison allocator state:
- it can make later real keepalive/GetBind replies from the same allocator look like regressions, and/or
- it can wrongly purge other current binds from the same allocator whose bind version is lower but still current.

Suggested fix:
1. carry the real allocator epoch alongside `resp.NewBind`, or
2. re-fetch allocator state before applying allocator-wide fencing on that fast path;
3. add a regression where one table on the new allocator has already rebounded to bind version N+1 while the allocator epoch is still N, and verify later keepalive/GetBind at epoch N is still accepted and does not purge other current binds.

iamlinjunhong:
> Re-reviewing the current head: most of the earlier blockers are addressed now, but I still think this should stay in request-changes state for one remaining correctness issue.
> 
> What looks fixed now:
> 
> * the holder-generation / `OK=false` keepalive sweep issue;
> * the pre-send cancellation / phantom remote lock issue;
> * the remote `GetBind` refresh paths bypassing allocator-aware fencing.
> 
> What still looks wrong:
> 
> * the `resp.NewBind` fast path is now allocator-aware, but it derives allocator state as `AllocatorID = newBind.AllocatorID` and `AllocatorVersion = newBind.Version`.
> 
> That is not safe because `LockTable.Version` is a bind version, not the allocator process epoch. The allocator initializes a bind from `l.version`, but later rebinds do `old.Version++`, so bind version and allocator epoch diverge. Feeding `newBind.Version` into `observeAllocatorStateLocked()` can therefore poison allocator state:
> 
> * it can make later real keepalive/GetBind replies from the same allocator look like regressions, and/or
> * it can wrongly purge other current binds from the same allocator whose bind version is lower but still current.
> 
> Suggested fix:
> 
> 1. carry the real allocator epoch alongside `resp.NewBind`, or
> 2. re-fetch allocator state before applying allocator-wide fencing on that fast path;
> 3. add a regression where one table on the new allocator has already rebounded to bind version N+1 while the allocator epoch is still N, and verify later keepalive/GetBind at epoch N is still accepted and does not purge other current binds.

fixed

XuPeng-SH (commented):
Re-reviewing the current head: the earlier allocator-epoch fast-path issue looks fixed now, but I still would not merge this yet because there is a lock-order deadlock risk in the `GetLockHolder` path.

The problematic ordering is:
- `pkg/lockservice/service_observability.go:72-79` holds `bindChangeMu.RLock()` across `current.getLockHolder(...)`.
- For a remote/proxy table, `pkg/lockservice/lock_table_remote.go:getLockHolder()` can go through `handleError()` / allocator-aware bind-refresh paths, which call `allocatorStateSnapshot()` (`allocatorVersionMu`).
- Meanwhile allocator/keepalive epoch update paths still take `allocatorVersionMu` first and then `bindChangeMu.Lock()` in `observeAllocatorStateLocked()`.

So the code now has a real lock-order inversion:
- `bindChangeMu -> allocatorVersionMu` on the observability path,
- `allocatorVersionMu -> bindChangeMu` on the allocator update path.

That can deadlock if a remote/proxy `GetLockHolder` races with allocator epoch processing.

Suggested fix:
1. avoid holding `bindChangeMu` across the remote/proxy `getLockHolder()` path (keep the existing lock for local-table reads only), or
2. otherwise enforce one global lock order so allocator-state reads never happen while `bindChangeMu` is held.

A targeted regression would be very useful here: race `GetLockHolder` on a remote/proxy table against allocator epoch update / keepalive purge and assert no hang.

XuPeng-SH (approved):
Re-checked the current head. The earlier correctness blockers I had raised look addressed now, and I no longer see a remaining merge blocker in the implementation. The unit/unhappy-path coverage is in good shape. A live restart/chaos regression for the exact stale-bind purge path would still be a valuable follow-up, but I would treat that as coverage hardening rather than a reason to block this backport.

PR #24906, opened 2026-06-09

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24896 


aptend (approved):
**Review Result**

**What I Checked**

- `GetBind` and `KeepLockTableBind` propagate the allocator epoch to CN.
- CN stale-bind purge follows the existing `LockTable.Version` semantics: cached binds with `Version < observed allocator epoch` are stale allocator state.
- main-branch bind-change handling was reviewed against the epoch purge path, including waiter cleanup and commit-side bind validation.
- The added tests cover get-bind refresh, keepalive failure/epoch refresh, concurrent observation, remote/proxy bind purge, and waiter cleanup.

**Local Verification**

```bash
go test ./pkg/lockservice -run 'Test(CommitDetectsStaleLocalBindAfterAllocatorRestart|AllocatorVersionZeroKeepsLocalBinds|AllocatorObserverDoesNotPurgeSameEpochBindVersions|GetBindAllowsRegressedAllocatorVersionWithoutPurging|KeepaliveEpochPurgeKeepsGroupMovePop|AllocatorObserverPurgesRemoteAndProxyLockTables|AllocatorObserverConcurrentKeepaliveAndGetBind|AllocatorObserverCloseWaitersOnStaleLocalBind)$'
timeout 180s go test ./pkg/lockservice
```

Note: the full `go test ./pkg/lockservice` hit the 180s local timeout without output. The PR-specific test set passed locally, and CI is green.

XuPeng-SH (changes requested):
I re-checked the latest version and the main correctness issue is still there.

The stale-bind fencing relies on allocator epochs being strictly monotonic, but the allocator still initializes its epoch from `time.Now().UnixNano()` in `pkg/lockservice/lock_table_allocator.go`. On the observer side, `observeAllocatorVersionLocked()` explicitly ignores any non-increasing allocator version (`observedVersion < oldVersion` and `== oldVersion`).

That means a real allocator restart can still bypass the purge path if the new process comes up with the same or a lower wall-clock timestamp (for example after clock skew / NTP step / restart on a slower-clock node). In that case `getLockTableWithCreate()` will keep returning and caching stale binds from `tableGroups`, and subsequent allocator responses from the restarted instance will continue to be ignored. This is exactly the stale-bind class the PR is trying to fence off.

What makes this especially high confidence is that the test `TestGetBindAllowsRegressedAllocatorVersionWithoutPurging` still encodes that behavior as expected. I think this needs either a truly monotonic allocator epoch source (persisted term/counter/generation) or a different invalidation rule that does not silently accept regressed allocator versions.

XuPeng-SH:
还有一个我觉得值得单独提出来的问题：**epoch-driven purge 这条路径没有走正常的 bind-change fencing 协议**。

正常的 bind change 是通过 `handleBindChanged()` 进入的，它会：
- 拿 `bindChangeMu`
- `tableGroups.set(...)` 新 bind
- `fenceByBindChanged(newBind)` 标记相关 active txn

这样 `Lock()` 路径就能通过 `txn.bindChanged` / `checkBindChangedBeforeLockSuccess()` 看到 `ErrLockTableBindChanged`，走可预期的 retry / fail-fast 语义。

但现在 allocator epoch purge 是在 `observeAllocatorVersionLocked()` 里直接 `holders.removeWithFilter(..., closeReasonBindChanged)`，它只是把 stale lock table 直接 close+remove 掉，没有经过 `bindChangeMu`，也没有对 active txns 调 `fenceByBindChanged()`。

结果是：如果有 in-flight local lock 正好拿着旧的 `lockTable`，它看到的可能不是 `ErrLockTableBindChanged`，而是更底层的 `local lock table closed` / waiter 上的 `ErrLockTableNotFound`。这和正常 bind-change 路径的语义并不等价，也更容易把上层打到意料外的错误分支。

我觉得至少需要确认：epoch purge 是否应该也通过和正常 bind change 一样的 fencing 机制串起来，而不是直接 close stale tables。

iamlinjunhong:
fixed the lockservice allocator-restart fencing issue across d4-24896, d3-24896, and m-24896.
Main fixes:
Added an allocator instance ID to lock table binds and allocator responses, so CNs can detect allocator replacement even when wall-clock-based versions regress or stay the same.
Updated CN-side stale bind detection to use allocator identity, not only monotonic allocator version.
Made allocator replacement purge go through normal bind-change fencing semantics, so active transactions get ErrLockTableBindChanged instead of lower-level closed/not-found errors.
Covered the combined keepalive unhappy path where OK=false and a newer/different allocator state arrive in the same response.
Ensured stale local, remote, and proxy lock tables are purged while current allocator-instance binds are preserved.
Added/updated regression tests for allocator ID changes, regressed allocator versions, keepalive cleanup, active txn fencing, remote/proxy stale bind cleanup, and concurrent observer paths.
For old branches, fixed remote-lock timeout handling to avoid recording phantom locks when the request likely never reached the remote owner.
Updated branch-specific tests whose old expectations conflicted with the new bind-change fencing behavior.

XuPeng-SH (changes requested):
I re-checked the latest head. I do not see a new high-confidence correctness bug to block on, and the unit coverage is already much stronger, but with a lockservice fix this concurrency-heavy I still think a few unhappy-path tests are missing before I am comfortable approving.\n\nThe main gaps I would still ask to close are:\n\n1. **Repeated allocator restarts / superseded-ID accumulation**\n   The new fencing now relies on tracking superseded allocator identities, but I did not see a focused test that drives 3+ sequential allocator replacements and proves both correctness and bounded behavior over repeated generations.\n\n2. **Client-side cancel before the remote lock request is actually sent**\n   The PR now keeps more remote-lock bookkeeping alive for later cleanup, and there is timeout coverage, but I did not find the specific unhappy path where the caller is canceled before the remote side ever received the lock request. I would like a test that proves the later unlock/close path handles that harmlessly.\n\n3. **The narrow race between `getLocalLockTable()` and taking `bindChangeMu.RLock()`**\n   The latest code intentionally adds a second stale-bind recheck around that window. That is exactly the kind of defensive race fix that should be pinned by a test, and I did not find one that explicitly exercises a bind change in that gap.\n\nIf those concurrency-edge cases are covered, I am happy to recheck the latest head.

iamlinjunhong:
> I re-checked the latest head. I do not see a new high-confidence correctness bug to block on, and the unit coverage is already much stronger, but with a lockservice fix this concurrency-heavy I still think a few unhappy-path tests are missing before I am comfortable approving.\n\nThe main gaps I would still ask to close are:\n\n1. **Repeated allocator restarts / superseded-ID accumulation**\n The new fencing now relies on tracking superseded allocator identities, but I did not see a focused test that drives 3+ sequential allocator replacements and proves both correctness and bounded behavior over repeated generations.\n\n2. **Client-side cancel before the remote lock request is actually sent**\n The PR now keeps more remote-lock bookkeeping alive for later cleanup, and there is timeout coverage, but I did not find the specific unhappy path where the caller is canceled before the remote side ever received the lock request. I would like a test that proves the later unlock/close path handles that harmlessly.\n\n3. **The narrow race between `getLocalLockTable()` and taking `bindChangeMu.RLock()`**\n The latest code intentionally adds a second stale-bind recheck around that window. That is exactly the kind of defensive race fix that should be pinned by a test, and I did not find one that explicitly exercises a bind change in that gap.\n\nIf those concurrency-edge cases are covered, I am happy to recheck the latest head.

fixed

aunjgr (commented):
LGTM for the lock service epoch fencing fix. Needs rebase — merge conflicts in hive_partition_test.go.

aunjgr (approved):
LGTM. The conflict in hive_partition_test.go is trivial — one-line sync.Once change. Please rebase and unblock merge.

XuPeng-SH (approved):

PR #24909, opened 2026-06-10
fix: support MySQL user-level lock functions

### LINKED ISSUE #24728: [Bug]: GET_LOCK / RELEASE_LOCK / IS_FREE_LOCK unsupported — blocks Prisma Migrate (v3.0.11 & 4.0-dev)
### Summary
The user-level advisory lock functions `GET_LOCK` / `RELEASE_LOCK` / `IS_FREE_LOCK` are not implemented. This **blocks Prisma Migrate**, which acquires a `GET_LOCK` advisory lock before applying any migration.

Related: #23036 (closed, "MySQL function not implemented: GET_LOCK") — still reproduces; also likely covered by the batch issue #24486.

### Steps to Reproduce
```sql
SELECT GET_LOCK('prisma_migrate_lock', 5);
SELECT IS_FREE_LOCK('prisma_migrate_lock');
SELECT RELEASE_LOCK('prisma_migrate_lock');
```

### Actual Behavior
```
ERROR 20105 (HY000): not supported: function or operator 'get_lock'
ERROR 20105 (HY000): not supported: function or operator 'is_free_lock'
ERROR 20105 (HY000): not supported: function or operator 'get_lock'
```

### Prisma impact
`prisma migrate deploy` fails immediately:
```
Error: not supported: function or operator 'get_lock'
   0: schema_core::state::ApplyMigrations
```
Workaround: set `PRISMA_SCHEMA_DISABLE_ADVISORY_LOCK=1`. With the flag, `migrate deploy` succeeds and creates all tables. But the default path is broken and surprising.

### Expected Behavior
MySQL implements these named locks (return 1 / 0 / NULL). Implementing them (or accepting them as no-ops) would let Prisma Migrate work out of the box.

---
**Test context** — Found while testing **Prisma ORM (v5.22)** compatibility against MatrixOne. Reproduced on **both**:
- Prod cloud — `8.0.30-MatrixOne-v3.0.11`
- 4.0 dev line — `nightly-0bfda604` (`select version()` → `8.0.30-MatrixOne-v`)

(Results on the older v3.0.5 are intentionally omitted.)

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes https://github.com/matrixorigin/matrixone/issues/24728

## What
- add GET_LOCK, RELEASE_LOCK, and IS_FREE_LOCK built-in functions
- implement lightweight connection-owned named lock semantics for MySQL compatibility
- cover lock acquire/release/free/null behavior in function tests

## Why
Fixes #24728. Prisma Migrate calls GET_LOCK before applying migrations.

## Tests
- CGO_CFLAGS="-I/home/jiangxinmeng/workspace/matrixone-24728-lock-funcs/thirdparties/install/include" go test ./pkg/sql/plan/function


XuPeng-SH (changes requested):
I found multiple blocking issues in the current implementation.

1. The named-lock state is only a package-global in-memory map inside one CN process. In MatrixOne this means two sessions routed to different CNs can both acquire the same lock name at the same time, so the core exclusivity guarantee is broken for real cross-session coordination.

2. Locks are never released when the owning session disconnects. The only mutations of the new lock map are explicit acquire/release calls, and this PR does not add any session-teardown cleanup path. So a client that exits without `RELEASE_LOCK()` can leave the lock stuck forever until the CN restarts, which is not compatible with MySQL named-lock semantics.

3. `GET_LOCK(name, timeout)` handles timeout semantics incorrectly: `timeout <= 0` returns `0` immediately, but MySQL compatibility requires `timeout = 0` to mean no-wait and negative timeout to mean wait indefinitely (subject to session cancellation).

4. The unhappy-path coverage is still thin around the new blocking behavior. There are no tests for disconnect cleanup, successful wait-then-acquire, timeout expiry, context cancellation while waiting, or the reentrant refcount path.

Because the first two are real behavior bugs and the third is a compatibility bug in the exposed SQL contract, I am requesting changes.

jiangxinmeng1:
Updated the PR to address the review request:

- GET_LOCK now uses the CN lockservice instead of a package-local map, so different CNs coordinate through the existing distributed lock path.
- Lock ownership is keyed by session id and lock name, with local refcounts only used for reentrant RELEASE_LOCK semantics.
- Session.Close now calls ReleaseUserLevelLocks to clean up held named locks on disconnect.
- Timeout behavior is updated: 0 uses FastFail/no-wait, positive values use context timeout, negative values wait indefinitely until session/query cancellation.
- Added tests for cross-service contention via a shared LockService fake, wait-then-acquire, timeout expiry, context cancellation, reentrant refcount, and cleanup.

Tests run:
- CGO_CFLAGS="-I/home/jiangxinmeng/workspace/matrixone-24728-lock-funcs/thirdparties/install/include" go test ./pkg/sql/plan/function
- CGO_CFLAGS="-I/home/jiangxinmeng/workspace/matrixone-24728-lock-funcs/thirdparties/install/include" go test ./pkg/frontend -run TestNonExistent

XuPeng-SH (changes requested):
I re-reviewed the latest head `9f3f04ff0baa3047317559d27e95ed83388ac777`. The previous distributed-lock/session-cleanup issues are mostly addressed, but I still found blocking MySQL-compatibility issues in the exposed SQL contract.

1. `RELEASE_LOCK()` cannot distinguish “not owned by this session” from “lock does not exist”. In `releaseUserLevelLock`, the local-refcount miss path returns `0, nil` immediately (`pkg/sql/plan/function/func_unary.go:6960-6963`). MySQL requires different results here: `0` only when the named lock exists but was not established by this thread, and `NULL` when the named lock does not exist / was already released (see the official MySQL locking-functions doc: https://dev.mysql.com/doc/refman/8.4/en/locking-functions.html). I verified this locally with a temporary test expecting `RELEASE_LOCK('missing_lock')` to return NULL; current code returns non-NULL 0 (`the 1th row expected NULL, but get not null`). This will make compatibility checks and application logic unable to tell “someone else owns it” from “there is no lock”. Please make release return nullable state, e.g. probe the lockservice with FastFail when the local refcount is absent so free/nonexistent returns NULL while conflict returns 0, and add tests for never-created and already-released locks.

2. Lock names are not validated against MySQL's 64-character limit. The current paths (`getUserLevelLock`, `releaseUserLevelLock`, and `isUserLevelLockFree`) pass `name` directly into the row/txn keys without any length check (`pkg/sql/plan/function/func_unary.go:6924`, `6954`, `6975`). MySQL enforces a maximum lock-name length of 64 characters. With this PR, overlong names are accepted and can acquire/release real locks, which is a visible behavior divergence from the stated MySQL compatibility goal. Please enforce the same limit, define the intended NULL/error behavior for incorrect arguments, and cover GET_LOCK/RELEASE_LOCK/IS_FREE_LOCK overlong-name cases.

Tests run locally with the PR worktree and main-worktree thirdparty CGO flags:
- `go test ./pkg/sql/plan/function -run 'TestUserLevelLock|TestReleaseUserLevelLocksCleanup|TestUserLevelLockBuiltinRegistration|Test_funids' -count=1`
- `go test ./pkg/frontend -run TestNonExistent -count=1`
- `git diff --check upstream/4.0-dev...HEAD`

I also noted the latest GitHub CI still has failures in Ubuntu UT and target upgrade; I could not get the failed log segment through `gh run view --log-failed`, so I am not using that as the primary blocker here.

XuPeng-SH (changes requested):
Re-reviewed latest head `5b3f328768fdf241787f6ca7f21b4030bbba4b8f`. The previous `RELEASE_LOCK()` NULL semantics issue is fixed, but there are still correctness gaps in lock-name handling that can break MySQL compatibility and mutual exclusion.

Blocking issues:

1. `pkg/sql/plan/function/func_unary.go:6824`: `validateUserLevelLockName` uses `len(name)`, so the 64 limit is enforced in bytes, not characters. MySQL documents the limit as 64 characters, and the MySQL WL for this feature says lock names are converted to UTF-8 before applying the new 64-character semantics: https://dev.mysql.com/doc/refman/8.4/en/locking-functions.html and https://dev.mysql.com/worklog/task/?id=1159. A name of 64 Chinese characters currently returns `internal error: user-level lock name exceeds maximum length of 64 characters`, even though it is exactly 64 characters. MatrixOne already has UTF-8 length helpers in this package; this should validate character/code-point length, and add a multibyte boundary test.

2. `pkg/sql/plan/function/func_unary.go:6838` / `pkg/sql/plan/function/func_unary.go:6938`: lock names are used as raw strings for the distributed row key and the local refcount key, making the implementation case-sensitive. MySQL's user-lock implementation performs case-insensitive comparison and uses a lowercased lock name as the MDL key (see the same WL above). Today two sessions can both acquire what MySQL treats as the same lock: session A `GET_LOCK('case_lock', 0)` returns 1, then session B `GET_LOCK('CASE_LOCK', 0)` also returns 1. That breaks the advisory-lock mutual exclusion guarantee. Please normalize the lock name consistently before validating/tracking/locking and add contention tests across case variants.

3. `pkg/sql/plan/function/func_unary.go:6824`: empty lock names are accepted because validation only checks `> 64`. MySQL's WL calls NULL, empty string, and longer-than-64 names wrong-name inputs. Current `GET_LOCK('', 0)` creates a real distributed lock, which leaves MatrixOne with non-MySQL state and release semantics for an invalid name. Please reject `''` consistently in `GET_LOCK`, `RELEASE_LOCK`, and `IS_FREE_LOCK`, with tests.

Local verification:
- PR tests pass: `go test ./pkg/sql/plan/function -run 'TestUserLevelLock|TestReleaseUserLevelLocksCleanup|TestUserLevelLockBuiltinRegistration|Test_funids' -count=1`
- Frontend compile check passes: `go test ./pkg/frontend -run TestNonExistent -count=1`
- `git diff --check upstream/4.0-dev...HEAD` passes
- Added temporary review-only tests for the three cases above; all failed as described, then removed them before submitting this review.

XuPeng-SH (changes requested):
Reviewed the latest fix approach, ignoring CI status. The implementation now addresses the earlier correctness issues around distributed locking, session cleanup, timeout semantics, RELEASE_LOCK NULL/0 semantics, empty/overlong names, multibyte length, and case-insensitive lock names. However, this PR introduces new public SQL functions (GET_LOCK, RELEASE_LOCK, IS_FREE_LOCK) for Prisma/MySQL compatibility and currently adds only Go unit tests; there is no BVT/end-to-end SQL case under test/distributed/cases. Please add at least one function BVT covering the SQL contract, e.g. GET_LOCK('prisma_migrate_lock', 0), IS_FREE_LOCK before/after acquire/release, RELEASE_LOCK for held and never-created locks, and NULL argument behavior. The fake-lockservice unit tests are strong, but they do not verify SQL parsing/planning/execution or result NULL/0/1 behavior through mo-tester.

XuPeng-SH (approved):
Re-reviewed the latest head with CI status ignored. The previous issues are addressed: the implementation uses the lockservice for cross-CN coordination, releases session-owned locks on Session.Close, handles timeout semantics, returns the correct RELEASE_LOCK 1/0/NULL states, normalizes names case-insensitively, validates empty/overlong/multibyte names, and preserves reentrant ownership. The new BVT now covers the public SQL path for GET_LOCK/RELEASE_LOCK/IS_FREE_LOCK, including NULL and never-created lock behavior. No blocking issue found in the fix approach.

heni02 (approved):

gouhongshen (approved):

fengttt (approved):

ouyuanning (approved):

jiangxinmeng1:
@Mergifyio queue

PR #24912, opened 2026-06-10
fix(logservice): match replicas by replica id

### LINKED ISSUE #24911: [Bug]: HAKeeper logservice checker treats different replica generations as the same shard
## Summary

HAKeeper LogService checker/operator paths can treat any running replica of the same shard as satisfying work for a specific desired replica. This is unsafe when a LogStore has multiple generations of the same shard, for example desired `shard=1/replica=262145` while the store heartbeat reports `shard=1/replica=275385`.

## Steps to Reproduce

Construct a HAKeeper `LogState` where:

```text
LogState.Shards[1].Replicas:
  262145 -> log-1

LogState.Stores["log-1"].Replicas:
  shard=1, replica=275385
```

Then run the LogService checker/operator finish checks for starting `shard=1/replica=262145`.

## Actual Behavior

The checker/operator can treat the store as already having shard 1 started because it only checks `ShardID` in several paths.

## Expected Behavior

The checker/operator should require an exact `(ShardID, ReplicaID)` match. `shard=1/replica=275385` must not satisfy `Start shard=1/replica=262145`.

## Impact

This can let HAKeeper believe a replacement/start operation has converged while the store is running a different replica generation. In a mixed-generation state, this can contribute to HAKeeper repeatedly targeting an old replica and treating a newer same-shard replica as zombie.

## Additional Context

This fix prevents the shard-level identity confusion. It does not by itself recover already-persisted inconsistent production data where HAKeeper state, local metadata, and Dragonboat WAL/bootstrap records have diverged.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #24911

HAKeeper LogService scheduling checked whether a store had any replica for a shard in a few places. When the same store had multiple generations of the same shard, for example desired `shard=1/replica=262145` while the heartbeat reported `shard=1/replica=275385`, HAKeeper could treat the desired replica as started or an operator step as finished.

This patch makes start scheduling and relevant operator finish checks match the exact `(shardID, replicaID)` pair. It keeps compatibility with old heartbeat fixtures that omit local `ReplicaID` in the checker path by falling back to the reported membership map only when `ReplicaID == 0`.

This prevents shard-level identity confusion, but it does not by itself recover already-persisted inconsistent production data where HAKeeper state, local metadata, and Dragonboat WAL/bootstrap records have diverged.

## Test

- [x] `CGO_CFLAGS="-I/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/lib" DYLD_LIBRARY_PATH="/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/lib" go test ./pkg/hakeeper/... -count=1`
- [x] `CGO_CFLAGS="-I/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/lib" DYLD_LIBRARY_PATH="/Users/shenjiangwei/Work/code/matrixone/thirdparties/install/lib" go test ./pkg/logservice -count=1`


aptend (approved):
**Review Result**

**What I Checked**

- `StartLogService`, `StartNonVotingLogService`, `BootstrapShard`, and `KillLogZombie` now require the expected `ReplicaID`, avoiding false completion when another replica for the same shard is already present on the store.
- The checker-side `toStart` detection now matches by `shardID + replicaID`, with compatibility for heartbeat records whose local `ReplicaID` is zero but whose membership maps contain the target replica.
- `StopLogService` and `StopNonVotingLogService` still match by shard only, which is appropriate because those steps finish when the store no longer reports any replica for that shard.
- Added/updated tests cover the same-shard/different-replica cases for start, bootstrap, and zombie handling.

**Local Verification**

```bash
git diff --check origin/3.0-dev...HEAD
go test ./pkg/hakeeper/operator ./pkg/hakeeper/checkers/logservice ./pkg/hakeeper/checkers
go test ./pkg/hakeeper/operator -run 'Test(StartLogService|StartNonVotingLogService|KillLogZombie|BootstrapShard)$'
```

`git diff --check` passed. The Go test commands did not run to completion locally because the environment is missing `usearch.h` from `github.com/unum-cloud/usearch/golang`; CI for this PR is green.

PR #24917, opened 2026-06-10
fix: retry QCloud PUT with seekable reader

### LINKED ISSUE #24916: [Bug]: COS fileservice PUT `connection reset` during Load
## Environment

| Field | Value |
|-------|--------|
| GitHub Actions run | [27181418714](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27181418714) |
| Namespace | `mo-main-commit-024703548-20260609` |
| Commit | `024703548` (main) |
| Cluster | Guangzhou nightly TKE |
| Workload | `1.Prepare Load Sysbench1000w Test Data`, TPCC/load parallel `LOAD DATA` |
| COS path | `mo-nightly-gz-1308875761/mo-benchmark-27181418714/` |
| COS VIP | `169.254.0.47:443` (Guangzhou VPC internal) |

**Scope:** During nightly bulk `LOAD DATA ... parallel "true"`, MatrixOne FileService **COS PUT** fails with `connection reset by peer`. MO wraps the Go error as **ERROR 20101** (`convert go error to mo error`). `IsRetryableError` already classifies connection reset as retryable, but **`QCloudSDK.putObject` explicitly does not retry** — causing Prepare Load / TPCC load failures.

## Summary

In run **27181418714**, step **`1.Prepare Load Sysbench1000w Test Data`** fails with:

```
ERROR 20101 (HY000): internal error: convert go error to mo error
Put "https://mo-nightly-gz-1308875761.cos.ap-guangzhou.myqcloud.com/mo-benchmark-27181418714/data/...":
write tcp 10.143.28.68:44840->169.254.0.47:443: write: connection reset by peer
```

Loki evidence (cluster loki-read, namespace above):

| Query | Result |
|-------|--------|
| `\|= "connection reset by peer"` | 30+ lines; pods `tp-cn-pgb6g`, `tp-cn-8dl9v`, proxy |
| `\|= "convert go error to mo error"` | 6 lines; same PUT path |
| Statement context | `load data url s3option ... bucket=mo-load-guangzhou-1308875761` |

Timestamps (UTC → CST):

- **06:52 UTC** (14:52 CST) — first burst during parallel LOAD
- **08:40 UTC** (16:40 CST) — second burst before sysbench1000w OLTP cases

Because mo-nightly-regression workflow uses `if: always() && !cancelled()` on subsequent sysbench steps, **OLTP tests continue after Prepare Load fails** — amplifying impact (partial `sysbench_db` data).

### Mechanism

1. `LOAD DATA ... parallel "true"` writes object data to shared COS via FileService.
2. Transient TCP reset on COS internal VIP (`169.254.0.47:443`) aborts PUT.
3. `QCloudSDK.putObject` returns error immediately — **no retry loop** (unlike AWS SDK path which wraps PUT with retry).
4. Frontend wraps via `ConvertGoError` → client sees **ERROR 20101**.

### Code gap

`pkg/fileservice/error.go` — `IsRetryableError` includes `"connection reset by peer"`.

`pkg/fileservice/qcloud_sdk.go` — `putObject`:

```go
// not retryable because Reader may be half consumed
_, err = a.client.Object.Put(ctx, key, r, opts)
```

AWS SDK v2 `putObject` uses `DoWithRetry(..., IsRetryableError)`; QCloud path is inconsistent.

### Why this is a MO bug (not pure COS infra)

| Check | Result |
|-------|--------|
| Error class | `connection reset by peer` — **already retryable** in MO |
| AWS SDK path | Has retry wrapper |
| QCloud SDK path | **No retry** on PUT |
| Client impact | Entire Prepare Load batch fails on single transient reset |
| Infra-only? | Even if COS blips, MO should retry idempotent PUT segments |

### Expected vs actual

| Expected | Actual |
|----------|--------|
| Transient COS TCP reset → retry PUT | Single reset → **20101**, Prepare Load fails |
| Consistent retry policy across S3 backends | AWS retries; **QCloud PUT does not** |
| Prepare Load failure blocks dependent OLTP | Workflow continues (`always()`) — CI issue, but MO should be resilient |

### P0 — QCloud PUT retry with safe reader

**`pkg/fileservice/qcloud_sdk.go`** `putObject`:

- Wrap PUT with `DoWithRetry(..., IsRetryableError)` like AWS path.
- On retry: use **seekable / re-readable** source (buffer, temp file, or offset reader) so half-consumed reader problem is solved — same pattern as other backends.

### P1 — Align error surfacing

- Distinguish retriable COS errors from hard failures in logs/metrics.
- Optional: metric `mo_fs_cos_put_retry_total` / `mo_fs_cos_put_fail_total`.

### P2 — CI (mo-nightly-regression, separate PR)

- Gate sysbench1000w OLTP on Prepare Load success — not MO core, but prevents testing on empty/partial data.

## Reproduce

1. Deploy MO ~`024703548` on Guangzhou nightly cluster with QCloud COS fileservice.
2. Run Prepare Load:
   ```sql
   LOAD DATA URL s3option {... bucket="mo-load-guangzhou-1308875761" ...}
   INTO TABLE sbtest1 ... PARALLEL "true";
   ```
   (10 tables × 10M rows as in nightly workflow.)
3. Induce or wait for COS `connection reset by peer` on PUT to `mo-benchmark-<run_id>/data/...`.
4. Observe ERROR 20101; Loki:
   ```logql
   {namespace="mo-main-commit-024703548-20260609"} |= "connection reset by peer"
   ```

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes https://github.com/matrixorigin/matrixone/issues/24916

- retry QCloud COS raw PUT when the input reader is seekable
- seek back to the original offset before each retry so a failed PUT cannot leave the reader half consumed
- fall back to single-concurrency multipart upload for large non-seekable readers
- add a QCloud PUT retry unit test that simulates `connection reset by peer` after the first body read

Fixes #24916

## Tests
- `go test ./pkg/fileservice -run '^TestQCloudSDKWriteRetriesSeekablePut$' -count=1 -v`\n- `go test ./pkg/fileservice -run 'TestQCloudSDKWriteRetriesSeekablePut|TestCOSParallelMultipart'`\n\nNote: `go test ./pkg/fileservice` currently fails locally in `TestMinioSDK/file_service/basic` with `Access Denied.` from `localhost:9007`, unrelated to this QCloud change.


XuPeng-SH (changes requested):
I found one high-confidence correctness issue here. The new non-seekable large-write fallback routes known-size streams into WriteMultipartParallel, but that path no longer preserves the original sizeHint contract when the reader ends early: an empty reader can now succeed with no upload, and a short reader can be rewritten to the shorter actual size and uploaded successfully. Before this PR, the putObject path passed ContentLength=sizeHint and could surface truncated bodies as a size-mismatch error. This means a truncated known-size stream can now turn into a silent short write, which is a real blocker. Also, the added retry test uses bytes.NewReader, which is already seekable and already retried by the COS SDK itself, so it does not prove the new wrapper logic.

XuPeng-SH (changes requested):
I found one blocking correctness issue in the current head.

For large non-seekable readers, `Write()` now falls back to `WriteMultipartParallel()` and only checks `sizeHint` after the upload finishes. That means a short reader can still leave a committed object behind and only then return `ErrSizeNotMatch`.

The problematic path is real in the current code:
- non-seekable + `sizeHint >= smallObjectThreshold` enters the new multipart fallback
- `WriteMultipartParallel()` reads the first chunk
- if the reader ends early and the actual bytes are smaller than `minMultipartPartSize`, it downgrades to `a.Write(..., &size, ...)` using the **actual** byte count
- that PUT can succeed and commit a short object
- then the outer `countingReader` check compares `n.Load()` to the original `sizeHint` and returns `ErrSizeNotMatch`

So the caller sees a failure, but the object may already have been written successfully with the wrong size. That violates the file-service write contract that failed writes must not become readable.

I think this needs to detect size mismatch before completing any upload, or explicitly abort/delete the object on mismatch. At minimum, the `short` non-seekable test should also assert that no object was persisted, because right now it only asserts the error.

XuPeng-SH (changes requested):
Blocking issue: the new exactSizeReader turns short/extra readers into ErrSizeNotMatch before CompleteMultipartUpload, but the multipart upload is not aborted in those unhappy paths.

In both QCloudSDK.WriteMultipartParallel and AwsSDKv2.WriteMultipartParallel, read/upload errors go through setErr, which calls cancel() on the derived upload context. The defer that is supposed to abort then reuses that same canceled ctx:

- pkg/fileservice/qcloud_sdk.go:336-337 creates the cancelable ctx, setErr at 421-428 cancels it, and the abort defer at 399-402 calls AbortMultipartUpload with the canceled ctx. A size mismatch from exactSizeReader reaches setErr via the readErr branch at 522-523, so the abort request is canceled locally and COS keeps the incomplete multipart upload.
- pkg/fileservice/aws_sdk_v2.go has the same pattern: cancelable ctx at 493-494, setErr at 576-583, abort defer at 549-557, and read errors at 678-679. The AWS multipart upload is likewise left un-aborted for the new size mismatch paths.

I verified this with local temporary mock tests: exact-size multipart completes, but both short-reader and extra-reader cases return ErrSizeNotMatch while state.completed/completeBody remains false and state.aborted remains false for COS and AWS. The PR tests currently only check that no object is completed for QCloud short reads; they do not assert that the multipart upload is aborted, so this leak is missed.

Please abort with a context that is not already canceled (for example, a separate timeout/background/WithoutCancel context derived from the original caller context as appropriate) and add tests that assert AbortMultipartUpload is sent for size-mismatch read errors in both QCloud and AWS multipart paths.

XuPeng-SH (changes requested):
I re-checked the latest update. The earlier sizeHint/truncation issue is fixed now, but I still see one substantive problem in the current version. The new seekable PUT retry path in pkg/fileservice/qcloud_sdk.go wraps putObject in MatrixOne's outer DoWithRetry, but the COS SDK still has its own built-in retry for seekable Put requests enabled by default. In production that means one transient failure is retried twice: once by the COS SDK itself, and then again by MatrixOne's outer loop. That multiplies request count and can stretch failure latency significantly. The strongest evidence is in the new test itself: it explicitly sets client.Conf.RetryOpt.Count = 0 before exercising the new logic, which avoids the production configuration. In the actual constructor path, NewQCloudSDK does not disable the SDK retry, and the COS SDK default is still RetryOpt.Count = 3. I think this should be resolved before approval: either disable the COS SDK retry for this PUT path when using the outer retry wrapper, or remove the outer retry and rely on the SDK's existing seekable-body retry behavior.

XuPeng-SH (changes requested):
I re-checked the latest update. The previous retry-layering issue is fixed now: the PR correctly disables COS SDK built-in retry so the new seekable PUT retry path no longer stacks two retry systems. However, that global change introduces a new issue: client.Conf.RetryOpt.Count = 0 disables COS retries for the entire client, including paths that are not wrapped in MatrixOne's DoWithRetry. In particular: client.Bucket.Head during initialization and AbortMultipartUpload during multipart cleanup. Both are still direct SDK calls, so with this patch they become single-shot network operations. A transient COS failure can now fail service startup or cause multipart abort cleanup to be dropped, even though those paths previously had the SDK's default retries. I think the fix should be scoped more narrowly: disable SDK retries only for the raw PUT path that now has an outer retry wrapper, or add explicit DoWithRetry around the other unwrapped COS operations.

XuPeng-SH (approved):
Reviewed the fileservice changes with a focus on retry semantics, size-mismatch handling, and multipart cleanup on unhappy paths. I do not see a blocking correctness issue in the seekable PUT retry flow, the non-seekable multipart fallback, or the abort-on-error handling, so this looks good to merge.

XuPeng-SH (changes requested):
Follow-up after a deeper unhappy-path pass: I found two blocking issues.

1. `pkg/fileservice/qcloud_sdk.go:112` disables the COS SDK retry loop, but `IsRetryableError` still does not treat COS HTTP 5xx responses as retryable. That drops existing retry coverage for body-less operations like `Bucket.Head`, `Object.Head`, list/delete, and multipart init, so transient 503/500 responses now fail immediately.

2. `pkg/fileservice/io.go:64-69` lets `exactSizeReader` return `(0, nil)` when `Expected == 0` and the wrapped reader returns `(0, nil)`. Callers that use `io.ReadFull`/`ReadAtLeast` can spin instead of getting EOF. This should normalize to `io.EOF` once the expected byte count is exhausted.

PR #24931, opened 2026-06-11
fix(rscthrottler): avoid double-counting RSS and reserved in S3 write throttler (4.0-dev)

### LINKED ISSUE #24881: [Bug]: Failed to load vector data during regression testing, but it works when isolation

### LINKED ISSUE #24888: [Bug] [4.0-dev tke regression] CN OOM during IVF vector test data copy.

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24881
Fixes #24888

Cherry-pick of #24892 to 4.0-dev.

The S3 write throttler's `acquireWithinRSSLimit` had two gates:

1. **Pool check** via `Available()` — double-counts S3 write buffers (in RSS as mmap'd mpool allocations + in reserved as throttler tokens). Under `ForceRefresh()` this returned 0 available, permanently rejecting all writes (#24881).
2. **RSS gate** `rss + reserved + ask > total × limitRate` — same double-counting. The RSS gate's original purpose (preventing OOM from non-S3 memory pressure, PR #24268) is now handled by RSS scavenging (85%/92% thresholds with graduated cache eviction + FreeOSMemory).

**Fix**: 

- **Pool check**: use `limit - reserved` directly instead of `Available()`. The pool only needs to bound forward-looking reservations against the throttler limit.
- **RSS gate**: removed. Physical memory protected by `pinnedRate >= 0.80` ceiling + RSS scavenging.

`Available()` is unchanged — other throttler consumers rely on its semantics.


XuPeng-SH (changes requested):
I understand the double-counting bug this PR is fixing, but I think the current backport swings too far in the other direction: it removes RSS from the CNFlushS3 admission decision entirely.

That means if the process is already near the cgroup/host limit because of **non-S3** memory, `AcquirePolicyForCNFlushS3` can still admit more S3 reservations as long as `reserved` is below the pool limit. In other words, the policy now protects the forward-looking reservation pool, but no longer protects physical memory headroom from unrelated RSS pressure.

A concrete example from the new test shape is: `actualTotalMemory=100`, `limit=90`, `reserved=0`, `rss=91`. With the new code, `pinnedRate()` is still 0 because it only looks at `reserved`, so the request falls through to `acquireWithinRSSLimit()` and can still be granted. That is exactly the kind of non-S3 RSS pressure case the old gate was preventing. RSS scavenging is useful, but it is not an equivalent guarantee here: it is asynchronous / best-effort and cannot reclaim live heap/query memory.

So my current view is that this is not just a coverage question. The fix removes the double-counting bug, but it also removes the only hard backpressure for high RSS caused by non-S3 memory. I think the right fix needs to discount the overlapping S3 portion from RSS without dropping physical-memory enforcement altogether.

XuPeng-SH (approved):

Issue #23137, opened 2025-11-24
Parser: UPDATE ... FROM join with vector distance fails


dengn:
## Description

`UPDATE ... SET ... FROM ...` with vector distance predicates fails at parse time. We update rows in a vector table by joining `company` metadata and filtering by `l2_distance`, but MatrixOne returns a syntax error near the `FROM` clause. This prevents standard update-with-join patterns from working for vector data maintenance.

## Schema & Setup

```sql
SET experimental_ivf_index = 1;
SET GLOBAL experimental_ivf_index = 1;

DROP TABLE IF EXISTS vec_join_case;
CREATE TABLE vec_join_case (
    id INT PRIMARY KEY,
    embedding VECF32(4),
    company_id INT,
    remark VARCHAR(100)
);
INSERT INTO vec_join_case VALUES
(10,"[0.1,0.2,0.3,0.4]",101,"init"),
(20,"[0.3,0.1,0.4,0.2]",102,"init"),
(30,"[0.5,0.6,0.4,0.3]",103,"init");
CREATE INDEX idx_vec_join_case USING ivfflat ON vec_join_case(embedding)
    LISTS=2 OP_TYPE "vector_l2_ops";
```

## Failing SQL

```sql
UPDATE vec_join_case t
SET remark = CONCAT('hot-', c.province)
FROM company c
WHERE c.id = t.company_id
  AND l2_distance(embedding,"[0.2,0.2,0.3,0.3]") < 0.35;
```

## Error Message

```
ERROR 1064 (HY000) at line 14: SQL parser error: ... syntax error at line 3 column 5 near " FROM company c ..."
```

## Impact
- Cannot perform join-based updates on vector tables, forcing inefficient client-side loops.
- Blocks typical enrichment workflows (e.g., tagging vectors with dimension table values).

## References
- `report_test_batch_263_vector_mixed_boundary_20251124_173546.md` Case 2622 and subsequent batches 2632/2642/... all fail with the same parser error.

heni02:
```
补充发现的 Bug（无 FK 时 source 重复匹配）
现象：target 表主键重复，一行被写成多行（数据损坏）。

最小复现：

CREATE TABLE dup_t (
    id INT PRIMARY KEY,
    v VARCHAR(20)
);
CREATE TABLE dup_s (
    t_id INT,
    v VARCHAR(20)
);
INSERT INTO dup_t VALUES (1, 'orig'), (2, 'orig');
[...]

PR #24932, opened 2026-06-11
fix(plan): dedup update from source matches

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23137

This PR fixes the PostgreSQL-style `UPDATE ... SET ... FROM ... WHERE ...` path when duplicate source rows match the same target row on tables without FK constraints.

The new UPDATE path previously fed duplicate joined rows directly into the update pipeline, which could materialize duplicate primary-key rows for one target row. This change adds a planner-side dedup step for `UPDATE ... FROM`: group by the target row's original columns and use `any_value()` for updated columns, matching the existing fallback path behavior.

Test coverage includes:

- A planner unit test that asserts the new UPDATE path produces `AGG + any_value` dedup.
- A BVT case without FK constraints that verifies duplicate source matches still leave each target primary key with exactly one row.

Tested with:

- `git diff --check`
- `go test ./pkg/sql/plan -run TestUpdatePgStyleFromDedupsDuplicateSourceMatchesOnNewPath -count=1`
- `mo-tester update_pg_style_from.sql` (`108/108` passed)
- `make`
- `make static-check` (`0 issues`)


heni02 (approved):

XuPeng-SH (changes requested):
I found one blocking correctness issue in the new UPDATE ... FROM dedup plan.

`pkg/sql/plan/bind_update.go:1029-1041` wraps each updated column in its own `any_value(...)`, and `pkg/sql/colexec/aggexec/any2.go:44-53` implements `any_value` as “first non-NULL seen for this column”. With duplicate source matches, that means different updated columns can be taken from different source rows whenever NULLs are distributed differently across those rows.

Example: if the same target row matches `(new_a = NULL, new_b = 'x')` and `(new_a = 7, new_b = NULL)`, the dedup can materialize `(new_a = 7, new_b = 'x')`, even though no matching source row produced that combination. PostgreSQL-style UPDATE ... FROM is allowed to pick an arbitrary matching row, but it still has to pick one row consistently, not synthesize a cross-row tuple. Please dedup on a whole-source-row choice rather than per-column `any_value` aggregation.

aunjgr (approved):
XuPeng-SH's concern is already addressed in the follow-up commits. The PR now uses row_number() + filter for whole-row dedup, not per-column any_value(). Tests assert any_value is absent. The BVT case whole_row_t/whole_row_s covers the exact cross-row synthesis scenario described in the review.

XuPeng-SH (changes requested):
I think there is still a real correctness hole here: the PR fixes whole-row dedup only in the new `bindUpdate` path, but tables that still fall back to `buildTableUpdate` can still synthesize a row that never existed in any matched source row.

The path is:
- FK-constrained targets still return `UnsupportedDML` from `dml_context.go` and are routed to `buildTableUpdate` (`pkg/sql/plan/dml_context.go:223-224`, `pkg/sql/plan/build.go:245-248`).
- For any PostgreSQL-style `UPDATE ... FROM`, that fallback path forces `needAggFilter` (`pkg/sql/plan/build_constraint_util.go:102-108`).
- `needAggFilter` is still implemented as `GROUP BY old target cols + any_value(each updated column)` (`pkg/sql/plan/build_dml_util.go:2875-2949`).

So with duplicate source rows like `(new_a = NULL, new_b = 'from-null-a')` and `(new_a = 7, new_b = NULL)`, the fallback path can still write `(7, 'from-null-a')`, even though no joined source row ever had that whole-row combination. This PR's new whole-row protection in `appendUpdateFromDedupNode()` does not reach that path.

I think the fallback planner needs the same row-level dedup strategy as the new path, not per-column `any_value()`.

XuPeng-SH (changes requested):
Requesting changes for a correctness issue in the new `UPDATE ... FROM` dedup key.

Both planner paths now dedup by partitioning on **every original target column**. That is not a safe proxy for target-row identity:

1. `GEOMETRY32` has no comparator in `pkg/compare/compare.go`, so `Node_PARTITION` can build a nil comparator and later dereference it.
2. float keys use `GenericAscCompare`, where `NaN != NaN`, so duplicate source matches against the same target row stop being recognized as duplicates if any partitioned target float column contains `NaN`.

So this can still either crash or miss the dedup entirely on valid target rows. I think the partition key needs to be a stable row identity (`row_id` / fake PK), not the whole old row.

XuPeng-SH (changes requested):
I found one blocking unhappy-path issue in the fallback `UPDATE ... FROM` path.

This PR now runs row-number dedup for fallback `UPDATE ... FROM` and then unconditionally clears `updatePlanCtx.needAggFilter`. But in the old fallback path, `needAggFilter` was also responsible for filtering out joined-target NULL rows (`row_id IS NOT NULL`) for cases like `UPDATE t1 LEFT JOIN t2 ... SET ... FROM ...`. The new `row_number() = 1` filter does not restore that safeguard, so one NULL-row per partition can survive and reach the update pipeline for the joined target table.

Please preserve the old NULL-row filter for joined-target fallback updates, or only disable `needAggFilter` when it was set solely for duplicate-source dedup and not for left/right-join target protection.

XuPeng-SH (approved):
Re-reviewed the latest head. The previous joined-target NULL-row concern is resolved: the fallback path now keeps `needAggFilter` for the `row_id IS NOT NULL` safeguard and uses a separate `dedupByRowNumber` flag only to skip the old `any_value` dedup AGG. I do not see a remaining blocking issue in the updated version.

XuPeng-SH (approved):
# PR #24932 Review: Fix UPDATE ... FROM Duplicate Source Matches Dedup

## 🎯 Review Summary

**Recommendation: ✅ APPROVE**

This PR fixes a critical correctness bug in PostgreSQL-style `UPDATE ... SET ... FROM ... WHERE` when duplicate source rows match the same target row. The solution is well-designed, thoroughly tested, and addresses the root cause using first-principles thinking.

---

## 📋 What This PR Does

**Problem**: When multiple source rows matched a single target row in `UPDATE ... FROM`, the old any_value() aggregation approach:
1. Could synthesize rows that don't exist in the source (mixing columns from different source rows)
2. Would crash on GEOMETRY32 columns (no comparator in pkg/compare)
3. Would fail on NaN float values (NaN != NaN breaks dedup recognition)
4. Would wrongly merge distinct rows with identical column values

**Solution**: Replace per-column any_value() with row-level dedup via row_number() window partitioned on `row_id` (target row's stable physical identity), ensuring whole source rows are picked.

---

## ✅ Correctness Analysis

### 1. **First-Principles Design** ✨
The core insight is brilliant: **partition by row identity (row_id), not row values**
- `row_id` is stable, unique, and always comparable
- Avoids all type-specific comparison issues (GEOMETRY32, NaN, etc.)
- Multi-target UPDATE correctly concatenates all row_ids as partition key

### 2. **Dual-Flag Architecture** 🏗️
```go
needAggFilter      // Drives BOTH: any_value agg + NULL-row filter
dedupByRowNumber   // When true: skip any_value BUT keep NULL-row filter
```
**Why this works:**
- `needAggFilter` remains true for UPDATE ... FROM
- When `dedupByRowNumber=true`, skip any_value aggregation (line 2877)
- But still apply NULL-row filter for LEFT JOIN targets (line 2953-2960)
- Prevents NULL rows from unmatched join targets entering update pipeline

### 3. **Generated Columns** ✓
Generated columns are recomputed **AFTER dedup** (bind_update.go:269-284), ensuring:
- Base column and generated column come from the SAME source row
- No cross-row contamination

### 4. **Type Correctness** ⚠️→✅
**Subtle fix**: When `dedupByRowNumber=true`, `rowIdExpr.Typ` must come from `tableDef.Cols[delCtx.rowIdPos].Typ` (line 2932), not from `aggNodeProjection` (which doesn't exist when agg is skipped).

---

## 🧪 Test Coverage Assessment

### Planner Unit Tests (407 lines)
- ✅ Verifies plan structure (row_number window, no any_value agg)
- ✅ DEFAULT expansion before dedup
- ✅ GEOMETRY32, decimal256, enum, vector types
- ✅ Generated columns
- ✅ Whole-row dedup validation

### BVT Tests (209 lines, 108/108 passed)
- ✅ Duplicate source matches (core fix)
- ✅ Whole-row integrity (synthesized row detection)
- ✅ GEOMETRY32 dedup (4 scenarios: new path, fallback path, with/without FK)
- ✅ LEFT JOIN NULL-row filter preservation
- ✅ Tables without explicit PK (fake-PK / hidden row_id)
- ✅ Multi-target UPDATE with LEFT JOIN
- ✅ Fallback path (FK-constrained tables)

**Exceptional test design**: Tests don't just check "no crash" but verify **semantic correctness** (e.g., counting synthesized rows, verifying per-id counts).

---

## 🔍 Edge Cases & Unhappy Paths

| Edge Case | Handling | Status |
|-----------|----------|--------|
| GEOMETRY32 (no comparator) | Partition by row_id, not column values | ✅ Solved + tested |
| Float NaN (NaN != NaN) | row_id comparison, NaN irrelevant | ✅ Solved |
| Identical-value distinct rows | row_id distinguishes them | ✅ Solved |
| No explicit PK | Uses hidden `__mo_fake_pk_col` row_id | ✅ Tested |
| Multi-target UPDATE | Partition by all row_ids | ✅ Tested |
| LEFT JOIN NULL targets | needAggFilter drives filter even when dedupByRowNumber=true | ✅ Tested |
| FK constraints (fallback path) | Both paths get row_number dedup | ✅ Tested |
| Memory allocation errors | SetRawBytesAt errors now propagate | ✅ Fixed + tested |

---

## ⚡ Performance Considerations

**Trade-off**: row_number() window is more expensive than any_value() agg
- **Why acceptable**: Correctness > performance. any_value() per-column is fundamentally broken.
- **Optimization**: Early exit when `partitionColPositions` is empty (no dedup needed)
- **Future**: Consider hybrid approach only for simple cases, but current solution is safest

---

## 🚨 Potential Issues Found

### None - All Concerns Resolved ✅

**Initial Question**: Does multi-target UPDATE with dedupByRowNumber work correctly?
**Resolution**: Test `jt1/jt2/js` demonstrates:
1. jt1.id=2 has no jt2 match → jt2.row_id is NULL for that join result
2. NULL-row filter removes it from jt2's update pipeline
3. jt1.id=2 is still updated (both rows in jt1 updated)
4. jt2 retains exactly 1 row (no NULL-row insertion)

**Conclusion**: Flag interaction is correct and well-tested.

---

## 📊 Code Quality

- **Comments**: Excellent. Complex logic clearly explained (e.g., lines 2925-2928 explain dual purpose of needAggFilter)
- **Naming**: `dedupByRowNumber` clearly indicates intent
- **Separation of Concerns**: Clean split between dedup mechanism and NULL-row filtering
- **Test Quality**: Tests verify plan structure AND semantic correctness
- **Error Handling**: Memory allocation errors now properly propagated

---

## 🔬 First-Principles Validation

**Three fundamental dedup requirements:**
1. **Determinism**: Same input → same output ✅ (row_number picks first)
2. **Integrity**: Pick whole rows, not column salad ✅ (row-level window)
3. **Type Safety**: Work for all types ✅ (row_id always comparable)

**Why row_id is the right key:**
- Physical identity, not logical equality
- Stable across all types (including uncomparable ones)
- Unique per row (no false merges)
- Always present (explicit or fake PK)

---

## 📝 Minor Observations

1. **Partition type support**: Added T_decimal256, T_TS, T_Rowid, T_Blockid (partition.go:9-16)
   - Risk: LOW (types already comparable, no specific tests)
   - Recommendation: Consider adding partition-specific tests in future

2. **any_value type support**: Extended to decimal256, vectors, geometry, enum (list_agg.go:943-952)
   - Needed for fallback cases where any_value is still used

---

## 🎯 Final Verdict

**APPROVE** ✅

**Strengths:**
1. ⭐ **Root cause fix**: Solves the fundamental "what should dedup key be" question correctly
2. ⭐ **Comprehensive testing**: 616 lines of new tests covering edge cases
3. ⭐ **Backward compatibility**: Both new and fallback paths work correctly
4. ⭐ **Type safety**: Handles uncomparable types (GEOMETRY32) gracefully
5. ⭐ **Clear documentation**: Comments explain WHY, not just WHAT

**No blocking issues found.**

**Recommended next steps:**
1. Merge this PR
2. Monitor production for any performance regressions (window vs agg)
3. Consider documenting the dedup strategy in architecture docs

---

**Review completed with multi-angle analysis:**
- ✅ Correctness (first principles, type safety, semantic integrity)
- ✅ Edge cases (GEOMETRY32, NaN, NULL-rows, multi-target, FK)
- ✅ Test coverage (unit + BVT, 108/108 passed)
- ✅ Performance (acceptable trade-off)
- ✅ Code quality (clear, well-commented, maintainable)

PR #24936, opened 2026-06-11
fix(frontend): align role rule merge semantics

ouyuanning:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23271

This ports the relevant behavior from matrixorigin/matrixone#24792 to 3.0-dev:

- Merge compatible rewrite rules from multiple active roles as a source-row visibility union by OR-ing WHERE filters instead of composing UNION queries.
- Clarify that this is intentionally not a `UNION DISTINCT` optimization for partial projections. If two visible base rows project to the same value, both rows remain visible.
- Keep row-wise projection expressions mergeable, such as `select a + 1 as b ...`, while aggregate/window/table-function and other non-compatible shapes still fall back to the later rule.
- Keep incompatible rewrite rules deterministic by letting later active-role rules override earlier ones.
- Make SET SECONDARY ROLE ALL validate secondary role metadata without changing the current primary role.
- Ensure SET ROLE switches the current role and invalidates privilege/rule caches.
- Update unit tests and BVT cases for role rewrite rules, SET ROLE, ownership privilege behavior, duplicate projected values, and row-wise expression projections.

Review note:

- The partial-projection behavior is a deliberate semantic choice for role rules: multi-role visibility is based on base rows, not deduplicated projected result sets. The new BVT case uses two distinct visible rows that both project to `a = 1` and expects `count(*) = 2` after `SET SECONDARY ROLE ALL`.

Validation performed:

- `gofmt -w pkg/frontend/rewrite_rule.go pkg/frontend/rewrite_rule_test.go`
- `git diff --check`

Go test notes:

- `go test ./pkg/frontend -run 'TestMergeRewriteRules|TestRewriteRuleMergeShapeForRule|TestLoadRuleCacheIncludesSecondaryRoles|TestRewriteSQLPropagatesRuleCacheLoadError|TestValidateRewriteRuleSQL'` is blocked by the local environment missing `usearch.h` for `vendor/github.com/unum-cloud/usearch/golang`.
- `go test -mod=mod ./pkg/frontend -run 'TestMergeRewriteRules|TestRewriteRuleMergeShapeForRule|TestLoadRuleCacheIncludesSecondaryRoles|TestRewriteSQLPropagatesRuleCacheLoadError|TestValidateRewriteRuleSQL'` is also blocked by the local environment missing `usearch.h` for `github.com/unum-cloud/usearch/golang`.
- The `role_rule.sql` mo-tester case was updated, but not run locally because no MatrixOne server is listening on the configured `127.0.0.1:6001`.


XuPeng-SH (changes requested):
I found one substantive issue in the new rewrite-rule merge behavior.

The current implementation now OR-merges many single-table `SELECT` rules as long as the projection list text matches, for example `select a from db1.t1 where ...`. That is **not** semantics-preserving in general. It is only safe for row-identity-preserving projections (effectively `select *` / exact table-star forms). For partial projections like `select a ...`, two different base rows that satisfy different role predicates but project to the same `a` value will now both be returned, whereas the previous `UNION DISTINCT`-style behavior would have deduplicated the projected result.

What makes this higher confidence is that the new tests explicitly bless this for `select a ...` and `select a as x ...`. So the PR is not just changing implementation shape — it is also locking in a user-visible semantic change for multi-role rewrite results. I think OR-merging should be restricted to projections that are guaranteed row-preserving, and non-identity projections should keep a semantics-preserving fallback instead of being treated as mergeable.

ouyuanning:
> I found one substantive issue in the new rewrite-rule merge behavior.
> 
> The current implementation now OR-merges many single-table `SELECT` rules as long as the projection list text matches, for example `select a from db1.t1 where ...`. That is **not** semantics-preserving in general. It is only safe for row-identity-preserving projections (effectively `select *` / exact table-star forms). For partial projections like `select a ...`, two different base rows that satisfy different role predicates but project to the same `a` value will now both be returned, whereas the previous `UNION DISTINCT`-style behavior would have deduplicated the projected result.
> 
> What makes this higher confidence is that the new tests explicitly bless this for `select a ...` and `select a as x ...`. So the PR is not just changing implementation shape — it is also locking in a user-visible semantic change for multi-role rewrite results. I think OR-merging should be restricted to projections that are guaranteed row-preserving, and non-identity projections should keep a semantics-preserving fallback instead of being treated as mergeable.

感谢指出这个问题。如果这里的目标是保持旧的 UNION DISTINCT 结果集语义，那么你的判断是对的：直接改成 OR 合并并不是等价变换。

我重新确认后，这个 PR 现在明确采用的是另一层语义：多 role rewrite rule 合并表示“基础行可见性并集”。也就是说，只要任意一个 active
role 允许某条基础表行，这条基础行就应该保持可见。在这个语义下，谓词 OR 合并是有意为之，包括 partial projection 场景。

例如两条不同的基础行都投影成 a = 1 时，返回两行是预期行为；如果用 UNION DISTINCT 按投影结果去重，反而会隐藏其中一条 role 允许可
见的基础行。

我已经更新了 PR 来明确这一点：

补充代码注释，说明 OR merge 不是 partial projection 上的 UNION DISTINCT 等价优化。
增加测试覆盖：两条不同可见基础行投影成相同值时，SET SECONDARY ROLE ALL 后 count(*) 仍为 2。
保持逐行投影表达式可合并，例如 select a + 1 as b ...，避免兼容的多 role 规则退化为后一个 role 覆盖前一个 role。
对不适合直接 OR 合并的形态仍然 fallback，例如 aggregate/window/table-function projection、顶层 ORDER BY/LIMIT、GROUP BY、
HAVING、DISTINCT 等。
所以我没有把 OR 合并限制到 select *，而是把当前预期的“基础行可见性并集”语义在代码注释和测试里显式固定下来。

heni02 (approved):

XuPeng-SH (changes requested):
I found one blocking privilege-regression in the `SET SECONDARY ROLE ALL` change.

`pkg/frontend/authenticate.go:3434-3462` now treats `SET SECONDARY ROLE ALL` as pure metadata validation and intentionally keeps `DefaultRoleID` unchanged. But the privilege/rule loaders still seed from that primary role unconditionally (`pkg/frontend/rewrite_rule.go:227-239`, `pkg/frontend/authenticate.go:7341-7344`, `7537`, `8581`). If the session's current primary role was revoked after login, `SET SECONDARY ROLE ALL` now invalidates the caches and reloads them with a stale role the user no longer has. The old code accidentally corrected this by rebinding to a still-granted role/public role.

Please either validate that the current primary role is still granted before preserving it here, or stop treating `DefaultRoleID` as implicitly valid in the downstream privilege/rule loaders.

ouyuanning:
> I found one blocking privilege-regression in the `SET SECONDARY ROLE ALL` change.
> 
> `pkg/frontend/authenticate.go:3434-3462` now treats `SET SECONDARY ROLE ALL` as pure metadata validation and intentionally keeps `DefaultRoleID` unchanged. But the privilege/rule loaders still seed from that primary role unconditionally (`pkg/frontend/rewrite_rule.go:227-239`, `pkg/frontend/authenticate.go:7341-7344`, `7537`, `8581`). If the session's current primary role was revoked after login, `SET SECONDARY ROLE ALL` now invalidates the caches and reloads them with a stale role the user no longer has. The old code accidentally corrected this by rebinding to a still-granted role/public role.
> 
> Please either validate that the current primary role is still granted before preserving it here, or stop treating `DefaultRoleID` as implicitly valid in the downstream privilege/rule loaders.

这个跟当前frontend权限验证的整体机制有关。
目前确实并非每次都验证role是否存在（所有权限都一样），只有发现无权限等问题的时候重新加载权限到cache。
是否增加每次权限相关的操作之前都同步最新的权限，这个可以单独列一个issue处理。

不过整体看估计也不会每次拉，估计得每次更改同步到所有CN，让CN持久化它可能更合理。毕竟权限很少改动。但是这个改动较大，它其实是整个权限验证流程的变更了。本PR暂不处理这个问题。

XuPeng-SH (approved):

Issue #24914, opened 2026-06-10
[Bug]: LOAD DATA parquet fails on optional STRING to DATETIME NULL conversion


Ariznawlll:
### Reproduce steps

Loading a concrete parquet object from COS reaches parquet type conversion, but fails when loading an optional string parquet column into a nullable DATETIME column.

Target table includes:

```sql
create table parquet_type_conversion_1y.all_type_parquet_1y (
  row_id bigint,
  c_tinyint tinyint,
  c_smallint smallint,
  c_int int,
  c_bigint bigint,
  c_uint8 tinyint unsigned,
  c_uint16 smallint unsigned,
  c_uint32 int unsigned,
  c_uint64 bigint unsigned,
  c_float32 float,
  c_float64 double,
  c_float_text float,
  c_double_text double,
  c_decimal decimal(38,10),
  c_decimal_sci decimal(38,10),
  c_bool bool,
  c_char char(32),
  c_varchar varchar(255),
  c_text text,
  c_json json,
  c_blob blob,
  c_binary binary(16),
  c_varbinary varbinary(32),
  c_date date,
  c_datetime datetime,
  c_timestamp timestamp,
  c_time time,
  c_uuid uuid,
  c_enum enum('red','green','blue','yellow','purple'),
  c_vecf32 vecf32(3),
  c_vecf64 vecf64(3)
);
```

Load SQL:

```sql
load data url s3option {
  'endpoint'='http://cos.ap-guangzhou.myqcloud.com',
  'access_key_id'='***',
  'secret_access_key'='***',
  'bucket'='mo-load-guangzhou-1308875761',
  'filepath'='mo-big-data/parquet_type_conversion_100M/part-00000.parquet',
  'format'='parquet'
}
into table parquet_type_conversion_1y.all_type_parquet_1y
parallel 'true';
```

### Actual behavior

```text
ERROR 20102 (HY000): load STRING(optional) to DATETIME NULL is not yet implemented
```

### Expected behavior

`LOAD DATA ... format='parquet'` should support loading valid datetime string values from an optional parquet STRING column into a nullable MO `DATETIME` column, or the limitation should be documented clearly if this conversion is intentionally unsupported.

### Notes

- The same COS endpoint/bucket can load an existing small parquet object successfully.
- Using the concrete object path reaches conversion, so this is not a file-not-found or credential issue.
- This was found while preparing a larger parquet type-conversion regression dataset related to #22748.

please contact me for parquet file

[...]

Issue #24922, opened 2026-06-11
[Bug]: external hive partition joins fail with invalid file path and remote CN dispatch timeout


Ariznawlll:
## Bug Report

Two external Hive partition join queries fail on MatrixOne `4.0-dev` image `matrixone-dev/matrixone:commit-b860e97`.

Regression job:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27310969765/job/80691107979

The same two cases were also reproduced manually with direct MySQL queries.

## Version

- MatrixOne branch: `4.0-dev`
- Image tag: `commit-b860e97`
- Full commit from regression build: `b860e97918fce957f4a51f95d2115154eb808572`

Previous nightly using `commit-dde9055` executed these two join cases successfully, so this looks like a regression in or before `b860e97`.

## Minimal Reproduction

The repro does not require loading the 1Y test tables. It only needs one external Hive partition table over the existing COS parquet path, plus a small dimension table.

Replace `<COS_ACCESS_KEY_ID>` and `<COS_SECRET_ACCESS_KEY>` before running the DDL.

```sql
drop database if exists external_hive_partition_2000_large;
create database if not exists external_hive_partition_2000_large;

create external table external_hive_partition_2000_large.fact_hive_2000 (
  col_int8 tinyint,
  col_int16 smallint,
  col_int32 int,
  col_int64 bigint,
  col_uint8 tinyint unsigned,
  col_uint16 smallint unsigned,
  col_uint32 int unsigned,
  col_uint64 bigint unsigned,
  col_float32 float,
  col_float64 double,
  col_decimal64 decimal(18,2),
  col_decimal128 decimal(38,10),
  col_decimal128_big decimal(38,0),
  col_bool bool,
  col_char char(32),
  col_varchar varchar(128),
  col_text text,
  col_binary blob,
  col_blob blob,
  col_date date,
  col_datetime datetime,
  col_timestamp timestamp,
  col_time_us bigint,
  col_json json,
  col_uuid uuid,
  col_enum varchar(32),
  part_id int
) url s3option {
  'endpoint'='http://cos.ap-guangzhou.myqcloud.com',
  'access_key_id'='<COS_ACCESS_KEY_ID>',
  'secret_access_key'='<COS_SECRET_ACCESS_KEY>',
  'bucket'='mo-load-guangzhou-1308875761',
  'filepath'='hive_partition/types_2000_large/',
  'region'='ap-guangzhou',
  'format'='parquet',
[...]

PR #24938, opened 2026-06-11
fix parquet datetime load and remote empty marker

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24914 #24922

fix parquet datetime load and remote empty marker


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (changes requested):
I found two substantive issues in the current patch.

1. **`stripFirstEmptyBlock` can drop the last real block for `ObjListRelData`.**
   In `pkg/vm/engine/readutil/datasource.go`, the `ObjListRelData` branch expands via `rd.DataSlice(0, rd.DataCnt())` before stripping the leading empty marker. But for `ObjListRelData`, `DataCnt()` is the count of real blocks, while the expanded block slice with `NeedFirstEmpty=true` contains `real blocks + 1` entries. That means the tail real block is truncated before the marker is removed. With two real blocks, `[empty, b0, b1]` becomes `[empty, b0]`, and stripping index 0 leaves only `b0`. The new test does not catch this because it uses `TotalBlocks=1` and no real object list.

2. **The new textual vector path is still incomplete for valid payloads.**
   In `pkg/sql/colexec/external/parquet.go`, the STRING/FIXED_LEN_BYTE_ARRAY vector path routes through `types.StringToArray`, which rejects `[]`. That means an empty vector value that is already accepted through the Parquet LIST path still fails through the textual path. Also, the new `isPlainStringLikeType` gate excludes JSON logical type entirely, even when the payload is a perfectly valid vector string like `[1,2,3]`. So the patch improves the happy path, but still rejects valid textual representations for reasons unrelated to actual parsing or dimensions.

I think both of these should be fixed before approval.

aunjgr (approved):

XuPeng-SH (approved):
Reviewed the 3.0-dev parquet backport and the follow-up fixes. The string->DATETIME / string->vector mappers, the JSON-logical-type vector text handling, and the centralized empty-marker stripping all look internally consistent, and I do not see a blocking correctness issue in the backport.

XuPeng-SH (changes requested):
Follow-up after a deeper pass: I found two blocking issues.

1. `pkg/sql/colexec/external/parquet.go:177-202` treats an exact column-name match as ambiguous if another column differs only by case, because the exact match is also appended into `caseInsensitiveMatches` before the ambiguity check. That means a schema containing both `Name` and `name` will reject a request for the exact `Name` column instead of picking the exact match.

2. `pkg/sql/colexec/external/parquet.go:1953-1974` computes `noNulls := !mp.srcNull || !mp.dstNull || page.NumNulls() == 0`. For nullable source pages loaded into NOT NULL destinations, `!mp.dstNull` makes `noNulls` true unconditionally, so dict-encoded optional pages skip the null branch and advance `j` for every row. On pages where `indexes` only contains non-null entries, this can overrun / misread dictionary indexes instead of returning a clean constraint violation.

XuPeng-SH (changes requested):
Blocking on required CI: Matrixone Utils CI (3.0) / Coverage is currently failing. The failed step is the coverage job's BVT run timing out after 30 minutes, while the other required 3.0 UT/BVT/upgrade checks are green. Please rerun or fix/prove this failure unrelated before approval. I rechecked the latest fixes for the earlier parquet column matching/null dictionary handling and empty-marker issues and did not find another code-level blocker.

XuPeng-SH (approved):

PR #24941, opened 2026-06-11
fix groupconcat

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24359

修改 group_concat 的order  by 改写.


heni02 (approved):

pkg/sql/colexec/aggexec/concat2.go
```diff
@@ -207,6 +219,395 @@ func (exec *groupConcatExec) Flush() (_ []*vector.Vector, retErr error) {
...
+			cmp = -cmp
+		}
+		if cmp != 0 {
+			return cmp, nil
+		}
+	}
+	return 0, nil
+}
+
+func compareGroupConcatOrderValue(left, right any) int {
```
fengttt, line 409: I am very surprised we need such a function.   This is generic order by clause and should reuse all the code from say, sort operator.

More: type is ALWAYS available, so we should NOT need a type switch.

fengttt (changes requested):

XuPeng-SH (changes requested):
Requesting changes for two semantic regressions in the new `GROUP_CONCAT ... ORDER BY` rewrite.

1. **`DISTINCT + ORDER BY` can keep the wrong representative row**
   - In `concat2.go`, DISTINCT dedup happens first via `distinctHash.fill(...)` on only the concat arguments.
   - The ORDER BY payload is sorted later in `flushGroup()`.
   - So for cases like `GROUP_CONCAT(DISTINCT x ORDER BY y)`, duplicate `x` values with different `y` values now keep whichever row was seen first, not the one implied by the ORDER BY.

2. **Explicit `NULLS FIRST` / `NULLS LAST` is dropped**
   - `bindGroupConcatOrderBy()` in `having_binder.go` only records ASC/DESC into `orderFlags`.
   - Execution then compares nils with the default behavior in `types.CompareValue`, so explicit null placement is no longer preserved.

I think both need to be fixed or rejected explicitly before approval.

daviszhen:
实现方案要重新设计

Issue #24944, opened 2026-06-12
[Bug] DATA BRANCH CREATE fails with a leading comment ('cannot find src and dst table') — unusable from Cloud web console


dengn:
## Summary
`DATA BRANCH CREATE TABLE <dst> FROM <src>` fails with `internal error: cannot find src and dst table` whenever the statement is **preceded by a SQL comment**. The MatrixOne Cloud web console injects a `/* cloud_user */` prefix before every statement, so this command **cannot be run from the Cloud console at all**. `DATA BRANCH DIFF` and `DATA BRANCH MERGE` are NOT affected — only `CREATE`.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11` (reproduced both via the web console and via mysql CLI)
- **Also reproduced on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v` — **still broken on main**

## Steps to reproduce
```sql
CREATE DATABASE ctest; USE ctest;
CREATE TABLE src(id INT PRIMARY KEY, v INT);
INSERT INTO src VALUES (1,1),(2,2),(3,3);

-- (1) no leading comment -> OK
DATA BRANCH CREATE TABLE b_plain FROM src;

-- (2) with a leading comment (this is exactly what the Cloud web console sends) -> FAILS
/* cloud_user */DATA BRANCH CREATE TABLE b_cmt FROM src;
```

## Actual result
Statement (2) fails on **both** v3.0.11 and main:
```
ERROR 20101 (HY000): internal error: cannot find src and dst table: /* cloud_user */ DATA BRANCH CREATE TABLE b_cmt FROM src
```
The branch table is not created. The error message echoes the full statement *including* the leading comment, which suggests the src/dst table names are extracted by re-parsing the raw statement text and the leading comment throws that off.

## Expected result
A leading comment should be ignored, exactly as it is for ordinary DDL/DML and for `DATA BRANCH DIFF` / `MERGE`. The statement should create the branch.

## Scope / notes
- Only `DATA BRANCH CREATE` is affected. With the same `/* cloud_user */` prefix, `DATA BRANCH DIFF ... AGAINST ...` and `DATA BRANCH MERGE ... INTO ...` both work fine.
- Impact: the data-branch feature's entry point is unusable from the MatrixOne Cloud web console (which always injects `/* cloud_user */`).

[...]

PR #24955, opened 2026-06-12
4.0: fix data branch create table clone SQL

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24944

`DATA BRANCH CREATE TABLE` used to rebuild clone SQL by applying a regexp to the original input SQL. That made the command sensitive to leading comments or whitespace, so statements such as `/* cloud_user */DATA BRANCH CREATE TABLE ...` could fail before reaching the clone implementation.

This PR moves the data branch create-table path to use the parsed AST directly, and centralizes clone SQL generation through AST formatting instead of reparsing hand-built SQL strings. It also makes clone table/database formatting produce executable SQL, quotes identifiers through a shared helper, preserves `TEMPORARY` / `IF NOT EXISTS` clone table options in the parser, and formats internal MO timestamp hints as `MO_TS` so generated SQL can be parsed by the existing grammar.

Validation:

- `go test ./pkg/sql/parsers/tree ./pkg/sql/parsers/dialect/mysql ./pkg/frontend -count=1`
- `make build`
- SQL regression on local MO: `/* cloud_user */DATA BRANCH CREATE TABLE b_cmt FROM src` returns the expected cloned rows
- Branch BVT: `./run.sh -p /Users/ghs-mo/MOWorkSpace/matrixone-3rd/test/distributed/cases/git4data/branch/ -r 100`
  - `TOTAL: 2977`
  - `SUCCESS: 2977`
  - `FAILED: 0`
  - `SUCCESS RATE: 100%`


XuPeng-SH (approved):
Reviewed the data-branch clone path change. Moving from regex/string rewriting to AST-based clone SQL generation looks like the right fix here: it removes the leading-comment fragility, preserves the parsed clone options, and keeps snapshot hint generation executable through the existing parser/formatter path. I do not see a blocking correctness issue in the frontend or parser changes.

iamlinjunhong (approved):

Issue #19635, opened 2024-10-28
[Bug]: ERROR 1064 (HY000) at line 3: SQL parser error: remapInfo step 0 nodeId 30 nodeType JOIN tip OnList


tom-csf:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

main

### Commit ID

0a3d52f3f9b4c740744f8567e689fed784feffe5

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

ERROR 1064 (HY000) at line 3: SQL parser error: remapInfo step 0 nodeId 30 nodeType JOIN tip OnList colRefCnt:[[6 1] : 0][[8 2] : 0][[9 0] : 0][[9 1] : 0][[20 3] : 1][[2 2] : 1][[11 0] : 0][[20 1] : 1][[3 0] : 2][[6 0] : 0][[7 0] : 0][[2 1] : 2][[1 0] : 3][[3 1] : 1][[2 6] : 1][[10 0] : 0][[7 3] : 0][[20 2] : 1][[20 6] : 1][[1 2] : 1][[6 2] : 0][[2 4] : 1][[1 4] : 0][[20 0] : 1][[20 7] : 1][[2 0] : 2][[3 2] : 0][[8 0] : 0][[20 4] : 1][[20 5] : 1][[4 1] : 2]colRefBool:[[0 5] : true][[0 6] : true][[0 7] : true][[0 0] : true][[0 1] : true

### Steps to Reproduce

```Markdown
1、run sql
select 	 s_suppkey, s_name, n_name, i_id, i_name, s_address, s_phone, s_comment
from	 item, supplier, stock, nation, region,
	 (select s_i_id as m_i_id,
		 min(s_quantity) as m_s_quantity
	 from	 stock, supplier, nation, region
	 where	 mod((s_w_id*s_i_id),10000)=s_suppkey
	 	 and s_nationkey=n_nationkey
	 	 and n_regionkey=r_regionkey
	 	 and r_name like 'Europ%'
	 group by s_i_id) m
where 	 i_id = s_i_id
	 and mod((s_w_id * s_i_id), 10000) = s_suppkey
	 and s_nationkey = n_nationkey
	 and n_regionkey = r_regionkey
	 and i_data like '%b'
	 and r_name like 'Europ%'
	 and i_id=m_i_id
	 and s_quantity = m_s_quantity
[...]

PR #24960, opened 2026-06-12
fix: avoid pushing join conds with outer scope tags

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #19635

This is a narrower fix for the join predicate pushdown issue previously addressed by #23719.

During the join-ordering path, `determineJoinOrder` calls `pushdownFilters(..., separateNonEquiConds=true)` with a global condition list. While recursively placing those conditions into the reordered join tree, a condition can reference tags outside the current join subtree. The old `getJoinSide` returns `JoinSideNone` for those outer-scope tags, so a condition can be misclassified as a single-side predicate and pushed into a child subtree that does not contain all referenced columns. That later fails during column remapping.

Instead of changing `getJoinSide` globally, this PR only uses an outer-scope-aware side classifier for the `separateNonEquiConds=true` path after join ordering. This preserves the existing early filter pushdown behavior and avoids changing TPCH plans, while still preventing invalid pushdown of join-order conditions with outer-scope tags.

A BVT case based on issue #19635 is added to `test/distributed/cases/join/join.sql`.

Validation:

- Reproduced the old failure on 10.222.1.128: `Column remapping failed: cannot find supplier.s_suppkey`.
- Verified the new BVT SQL passes with this fix.
- Verified TPCH 1T explain plan has no diff against the old-behavior baseline:
  - baseline: `/data3/hanfeng/tpch1000-explain-pr23719-old-20260612-143441.out`
  - fixed: `/data3/hanfeng/tpch1000-explain-pr23719-separate-only-20260612-144338.out`
  - diff lines: 0
- `go test ./pkg/sql/plan` passed on 10.222.1.128 with the same CGO include/lib settings used by `make build`.


heni02 (approved):

XuPeng-SH (approved):
Looks good to me. The fix stays narrowly scoped to the join-order/separateNonEquiConds path, the new outer-scope-aware side classification addresses the remapping failure directly, and the added unit/BVT coverage matches the failure mode well.

aunjgr (changes requested):
The fix correctly identifies the root cause, but the approach reuses `JoinSideBoth` for outer-scope tags — semantically wrong. `JoinSideBoth` means "from both sides of this join," not "from an outer scope."

The correct fix:

1. Add `JoinSideOuter = 1 << 5` to the constants in `pkg/sql/plan/types.go`
2. In `getJoinSide`, return `JoinSideOuter` instead of `JoinSideNone` for columns not in leftTags/rightTags/markTag
3. In `pushdownFilters`, treat `JoinSideOuter` the same as `JoinSideBoth` — cannot push to a single child, stays at current level

This makes the intent explicit: outer-scope tags don't belong to either child of the current join. The current fix works but conflates two distinct concepts.

aunjgr (changes requested):
The root cause is correctly identified but the fix has two issues:

**1. Misuse of **: Outer-scope tags are not "from both sides" — they're not from either side. Add  in  and return it from  for columns not in leftTags/rightTags/markTag. Then handle it in  (same as  — stays at current level).

**2. Unnecessary wrapper**: Fix  directly instead of adding . A correctly classified  naturally prevents single-child pushdown for ALL callers, not just the  path. The conditional wrapper () shouldn't be needed — the fix should be universal.

aunjgr (changes requested):
The root cause is correctly identified but the fix has two issues:

1. Misuse of JoinSideBoth: Outer-scope tags are not "from both sides" — they are not from either side. Add JoinSideOuter = 1 << 5 in types.go and return it from getJoinSide for columns not in leftTags/rightTags/markTag. Then handle it in pushdownFilters (same as JoinSideBoth — stays at current level).

2. Unnecessary wrapper: Fix getJoinSide directly instead of adding getJoinSideWithOuterScope. A correctly classified JoinSideOuter naturally prevents single-child pushdown for ALL callers, not just the separateNonEquiConds path. The conditional wrapper should not be needed — the fix should be universal.

aptend:
@aunjgr Thanks for the review. I agree with the semantic issue: an outer-scope tag is not really `JoinSideBoth`.

I updated the PR to add `JoinSideOuter` and return it from the outer-scope-aware classifier, so this path no longer conflates outer-scope references with “columns from both children”. `pushdownFilters` now keeps predicates with `JoinSideOuter` at the current join level.

I intentionally kept the base `getJoinSide` behavior unchanged and did not make this universal. This PR is scoped to the post-join-order `separateNonEquiConds=true` path, where predicates may reference tags outside the current join subtree. Making `getJoinSide` globally return `JoinSideOuter` for unknown tags is the same risk class as the earlier broad change in #23719; in our local validation/investigation that changed TPCH q18/q21 plans. The existing `JoinSideNone` behavior is still relied on by other pushdown paths for current compatibility.

So the compromise here is: make the outer-scope semantics explicit where the bug happens, but avoid changing the legacy classifier globally until we can separately audit the other callers and TPCH plan impact.

Validation after the update:
- `go test ./pkg/sql/plan` passed on 10.222.1.128

aunjgr (commented):
Good catch on TPCH Q18/Q21. The issue with just adding `JoinSideOuter` and returning it from `getJoinSide` is the routing logic in `pushdownFilters`:

```go
if side & JoinSideLeft == 0  → rightPushdown
else if side & JoinSideRight == 0 → leftPushdown
else → cantPushdown
```

`JoinSideOuter | JoinSideLeft` has the `JoinSideRight` bit clear → pushed LEFT. Wrong.

The proper fix needs two changes:

1. Add `JoinSideOuter = 1 << 5` in types.go, return it from `getJoinSide` for columns outside leftTags/rightTags/markTag (instead of falling through to the default 0).

2. In the routing logic, check for `JoinSideOuter` before pushing to a child:

```go
side := getJoinSideForPushdown(filter, leftTags, rightTags, markTag)
if side & JoinSideOuter != 0 || side == JoinSideNone {
    cantPushdown = append(cantPushdown, filter)
} else if side & JoinSideLeft == 0 {
    rightPushdown = append(rightPushdown, filter)
} else if side & JoinSideRight == 0 {
    leftPushdown = append(leftPushdown, filter)
} else {
    cantPushdown = append(cantPushdown, filter)
}
```

`JoinSideOuter` means "references a scope outside this join subtree" — semantically distinct from `JoinSideBoth`. The routing check is explicit instead of relying on `JoinSideBoth`'s bit pattern as a side effect.

aunjgr (approved):

Issue #24956, opened 2026-06-12
[Bug]:  ownership incorrect (bug reported from MOI)


ouyuanning:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

3.0.14

### Commit ID

3.0.14

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

https://github.com/matrixorigin/matrixflow/issues/11272

### Steps to Reproduce

```Markdown
https://github.com/matrixorigin/matrixflow/issues/11272
```

PR #24961, opened 2026-06-12
fix(frontend): keep DDL ownership on matched role

ouyuanning:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24956

This PR fixes DDL ownership under `SET SECONDARY ROLE ALL`. When `CREATE DATABASE` or `CREATE TABLE` is authorized by a secondary role, the created object owner and implicit ownership grant now use the role that actually matched the privilege check instead of always using the session primary/default role.

Main changes:

- Track the matched role for DDL privilege checks that need ownership attribution.
- Prefer the current primary role when it can authorize the DDL, then fall back to enabled secondary roles.
- Attach the matched DDL owner role to the request context only for database/table creation owner metadata.
- Keep `current_role()` unchanged while assigning object ownership to the authorizing role.
- Add frontend coverage and BVT coverage for secondary-role DDL ownership.

Validation:

- `git diff --check`
- `go test ./pkg/defines`
- Targeted `go test ./pkg/frontend -run TestCreateTableOwnerRoleFollowsPrivilegeRole` could not complete in the local environment because the native `usearch` header/library dependency is missing.


heni02 (approved):

gouhongshen:
Thanks for the fix. I think there are two issues that should be resolved before merge.

### Must fix

1. **Multi-path inherited role graphs can select the wrong DDL owner**

The inherited-role traversal tracks visited roles globally and records `roleRoot` with first-writer-wins semantics. If both the primary role and a secondary role can reach the same inherited role through different paths, the later BFS layer is ordered by role ID rather than active-role priority. That can make the secondary role become the matched owner even when the primary role can also authorize the DDL.

Please make matched-role selection deterministic and preserve the intended priority, for example by carrying the active root in traversal state or by resolving candidates per active root before choosing the owner.

2. **Dropping a role that still owns objects can make those objects impossible to drop**

`DROP ROLE` removes role/grant metadata, but it does not prevent or clean up `mo_tables` / `mo_database` owner references. The new drop cleanup path resolves the object owner ID back to a role name and returns that error before the actual `DROP` runner executes. If the owner role has already been dropped, `DROP TABLE` / `DROP DATABASE` can fail before deleting the object.

Please fix the ownership lifecycle at the source: either block dropping roles that still own objects, or migrate/clean ownership in the same transaction. The implicit revoke path should use object ID / owner role ID directly instead of depending on a live role name lookup.

### Test gaps

Please add regression/destructive tests for:

- multi-path role inheritance where primary and secondary roots both reach the same granting role, and primary preference must be preserved;
- table and database drop behavior when the recorded owner role has been deleted or owner lookup fails;
- `CREATE DATABASE` owner assignment under secondary / inherited roles, not only table creation;
- privilege cache enabled, since the current coverage mostly forces cache off;
- binary protocol prepared execution for DDL owner assignment;
- failure/no-dirty-state behavior for implicit grant/revoke cleanup;
- hostile identifiers in database/table names on the owner lookup and cleanup path.

These tests should be tied to the fixes above, not only happy-path coverage.

gouhongshen (approved):

XuPeng-SH (approved):
Looks good to me. The matched-role ownership behavior is covered, and the follow-up fix for deleted owner roles keeps orphan-owner drops from getting stuck on implicit revoke.

Issue #24962, opened 2026-06-12
[Bug] IVF l2_only pre-filter recall hangs with severe CN read amplification on 10M wiki dataset


Ariznawlll:
## Summary

Nightly run stuck for more than 5 hours in `IVF PRE POST FILTER VECTOR RECALL TEST`, specifically at step `Recall l2_only pre-filter`.

Run/job: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27359437757/job/80930011420

This does not look like a GitHub runner issue, pod restart, OOM, or RPC queue blockage. Metrics and profiles point to severe read amplification in MO while executing IVF `l2_only` pre-filter recall on the 10M wiki dataset.

## Failed/stuck step

Workflow step:

```bash
python run_wiki.py recall --config "${WIKI_CFG}" \
  --sql-mode l2_only --filter-mode pre \
  --gt-source ann -n "${WIKI_N}" -k "${WIKI_K}" --concurrency "${WIKI_CONCURRENCY}" \
  | tee "${GITHUB_WORKSPACE}/wiki_recall_l2_only_pre.log"
```

Parameters:

```text
WIKI_N=10000
WIKI_K=10
WIKI_CONCURRENCY=100
```

The generated SQL shape is:

```sql
SELECT file_id, id
FROM historical_file_blocks_wiki_ivfflat10m
ORDER BY l2_distance(embedding, ?) ASC
LIMIT 10
BY RANK WITH OPTION 'mode=pre';
```

The benchmark client uses `ThreadPoolExecutor(max_workers=100)` and waits through `as_completed(...)` without a per-query timeout. If some queries become extremely slow or stuck, the whole CI step waits until the workflow timeout.

## Environment

- MO image/version: `commit-516e55d88`
- Namespace: `mo-main-commit-516e55d88-20260611`
- MO service IP: `172.16.96.178`
- Pods:
  - DN: `nightly-regression-dis-dn-0`
  - CNs: `nightly-regression-dis-tp-cn-2g287`, `nightly-regression-dis-tp-cn-j74fr`, `nightly-regression-dis-tp-cn-w5h4b`

## Metric evidence

Window: from `2026-06-12 08:43:02 CST`, when `Recall l2_only pre-filter` started.

Basic health checks:

- Pod restarts: `0` for CN/DN/proxy
- RPC write queue: `0`
- Cache pressure: nearly `0`

CN CPU shortly after recall started:

[...]

Issue #24971, opened 2026-06-13
[Refactor] Make pipeline termination protocol explicit and cancellation-safe


XuPeng-SH:
## Background

Related to #24962.

The immediate hang in #24962 is not caused by IVF SQL semantics directly. The stack traces point to pipeline cleanup waiting indefinitely for terminal pipeline signals, for example `Merge.Reset -> PipelineSignalReceiver.WaitingEnd`, while the corresponding sender side may fail to send or the receiver may fail to consume the final end signal.

The current protocol is distributed across several implicit mechanisms:

- `nil` batch means end/error in some paths.
- `WaitRegister.Ch2` is used for both data notifications and terminal notifications.
- `PipelineSpool.Close()` waits for receivers to consume terminal messages.
- cleanup ordering decides whether sender or receiver can make progress.
- some cleanup paths used unbounded sends or `context.TODO()` waits.

This makes unhappy paths fragile: cancellation, remote CN failure, network jitter, channel backpressure, mismatched `NilBatchCnt`, or cleanup order can leave one side waiting for another side that is no longer guaranteed to make progress.

## First principles

A pipeline edge should have an explicit lifecycle and these invariants:

1. Terminal state is a protocol event, not an implicit `nil` data batch.
2. `End`, `Error`, and `Abort` are idempotent. Calling them multiple times must be safe.
3. Cleanup must not depend on the peer consuming one more channel message.
4. Every send, receive, and terminal wait must be cancelable or bounded.
5. Spool memory ownership must be independent from channel notification delivery. If receivers are late or gone, cleanup must avoid use-after-free even if it cannot reclaim all memory immediately.
6. Normal long-running queries must not be affected by cleanup-only deadlines.

## Proposed direction

Introduce an explicit pipeline edge abstraction, or evolve `WaitRegister` toward one, with typed events:

- `Data`
- `End`
- `Error`
- `Abort`

The edge should provide APIs along these lines:

```go
type PipelineEventType int

const (
    EventData PipelineEventType = iota
    EventEnd
    EventError
    EventAbort
)

type PipelineEvent struct {
    Type  PipelineEventType
    Batch *batch.Batch
    Err   error
    Spool *pSpool.PipelineSpool
    Index int
}

type PipelineEdge struct {
    // implementation detail: channel, done/abort signals, terminal state, once guards
}

func (e *PipelineEdge) Send(ctx context.Context, ev PipelineEvent) error
[...]

PR #24972, opened 2026-06-13
fix: bound pipeline cleanup waits

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #24962

Follow-up refactor issue: #24971

This is a hotfix for pipeline cleanup hangs observed in #24962. The root problem is that pipeline cleanup can wait indefinitely for terminal signals when the peer sender/receiver is no longer guaranteed to make progress.

This PR:

- Bounds `Merge.Reset` cleanup waits for terminal pipeline signals.
- Bounds cleanup-time terminal signal sends in `Connector`, `Dispatch`, and remote notify cleanup.
- Bounds `PipelineSpool.Close` with a timeout variant and avoids force-freeing spool cache when receivers may still be late.
- Cleans up sender/receiver linked pipelines by resetting the terminal sender and leaf merge concurrently, while preserving CTE special cleanup gates.
- Makes remote receiver cleanup error notification non-blocking to avoid cleanup deadlocks.
- Adds `mo_pipeline_cleanup_event_total{event=...}` so abnormal cleanup events remain observable even when logs are rate-limited.
- Adds rate-limited cleanup warnings to avoid log storms under large abnormal cleanup bursts.
- Sends terminal signals to immediately healthy local dispatch receivers before waiting on blocked local receivers.
- Adds tests for missing end signals, full channels, cleanup ordering, remote error channel backpressure, local receiver backpressure, and cleanup warning suppression.

The change is intentionally scoped to cleanup/reset paths. Normal long-running query data paths are not bounded by these cleanup-only timeouts.

Validation run locally:

```bash
go test ./pkg/vm/process ./pkg/vm/pipeline ./pkg/sql/colexec/dispatch
go test ./pkg/sql/colexec/merge ./pkg/sql/colexec/connector ./pkg/sql/compile ./pkg/util/metric/v2 -run '^$'
make build
```


XuPeng-SH (commented):
Blocking issue: the new timeout/fallback cleanup paths can leak PipelineSpool-owned memory.

In both `pkg/sql/colexec/connector/types.go` and `pkg/sql/colexec/dispatch/types.go`, we now nil out `ctr.sp` even when `CloseWithTimeout()` returns false or spool close is skipped after falling back to a direct signal. But `pkg/container/pSpool/sender.go` only frees the cached spool buffers on the success path of `CloseWithTimeout()` (`ps.cache.free()`). These spools are created from `proc.Mp()`, so a timed-out cleanup can now abandon cached batch buffers/off-heap vector memory until session teardown, which turns the hang fix into a cleanup-time memory leak on the same failure path.

Please keep ownership of the spool until its cache is actually freed, or hand timed-out spools to a deferred reaper/background close path instead of dropping the reference immediately.

gouhongshen (approved):
Reviewed the bounded pipeline cleanup changes again. I did not find any blocking issues in the cleanup ordering, bounded signal send/drain paths, deferred spool cleanup, or remote receiver notification handling.

aptend (approved):
Reviewed again from aptend. No blocking issues found in the bounded cleanup paths.

aunjgr (approved):
Clean fix for #24962 pipeline cleanup hang. Bounded waits, timeout variants, deferred spool cleanup. Self-review memory leak concern already addressed with deferred reaper. gouhongshen and aptend approved.

Issue #24939, opened 2026-06-11
[Refactoring]: Eliminate legacy INSERT ... ON DUPLICATE KEY UPDATE path, unify to operator-based execution


ck89119:
### Why do you want to refactor this code?

`INSERT ... ON DUPLICATE KEY UPDATE` (ODKU) currently has **two execution paths**, mirroring the situation REPLACE INTO had before #23946 / #24044:

- **Modern path (default):** `bindInsert` → `appendDedupAndMultiUpdateNodesForBindInsert`, built on `DEDUP JOIN` + `MULTI_UPDATE`.
- **Legacy path (fallback):** `build.go` catches `ErrUnsupportedDML` and falls back to `buildInsert` → `buildInsertOnDuplicateUpdatePlans`, producing a `Node_ON_DUPLICATE_KEY` node executed by the colexec `onduplicatekey` operator.

Maintaining two paths duplicates semantics, complicates correctness/perf work, and keeps the legacy `onduplicatekey` operator alive. We should eliminate the legacy ODKU path and unify everything onto the operator-based path, the same way REPLACE was unified.

The legacy path is still reached today for three scenarios the modern path rejects with `ErrUnsupportedDML`:

| Scenario | Trigger | Modern | Legacy |
| --- | --- | --- | --- |
| No explicit PK + has unique key (fake PK + UK) | `bind_insert.go` `update on duplicate without primary key` | ✗ | ✓ |
| Vector / irregular index table | `bind_insert.go` `insert into vector/text index table` / `have vector index table` | ✗ | ✓ |
| Foreign-key table (or referenced) | `dml_context.go` `foreign key constraint` in `ResolveTables` | ✗ | ✓ |

Two other fallback cases — updating a PK column and updating a UK column — are rejected by **both** paths, so removing the legacy path causes no functional loss for them (only error-message unification).

### Describe the solution you'd like

Unify all ODKU scenarios onto the modern `DEDUP JOIN` + `MULTI_UPDATE` path, then delete the legacy `Node_ON_DUPLICATE_KEY` path. Reuse the techniques already proven in #24044:

1. **Fake PK + UK:** introduce an `isFakePK` branch in `appendDedupAndMultiUpdateNodesForBindInsert`. With a unique key, dedup/conflict-detect on the first unique key instead of the PK; with no unique key, degrade to a plain INSERT.
[...]

PR #24973, opened 2026-06-13
feat(insert): align ON DUPLICATE KEY UPDATE with MySQL, drop legacy operator

### LINKED ISSUE #25000: [Bug]: REPLACE on a table with a vector/fulltext index does not maintain the index (rows not immediately searchable)

### Branch Name

main

### Commit ID

ed3b4f971

- Single-node (`./mo-service -launch etc/launch/launch.toml`)
- OS type: macOS / Linux

### Actual Behavior

`REPLACE` into a table that carries an irregular index (ivfflat vector index, or fulltext index) succeeds, but the index is **not** maintained synchronously. As a result, a row inserted or whose vector is changed via `REPLACE` is **not** reflected in a subsequent index-ordered ANN/KNN query.

Concretely, a brand-new row inserted via `REPLACE` (PK does not exist, so it is an insert) is missing from KNN results, while the same new row inserted via plain `INSERT` is returned. `INSERT` goes through the legacy path which maintains the IVF index synchronously; `REPLACE` goes through the modern path which strips the irregular index and relies on lazy/async maintenance.

```
A) after REPLACE new id=7=[5,5,6]; KNN near [5,5,5] limit 3  ->  5, 4, 3   (id=7 missing)
B) after INSERT  new id=8=[5,4,5]; KNN near [5,5,5] limit 3  ->  5, 8, 4   (id=8 present)
```

### Expected Behavior

`REPLACE` on a table with a vector/fulltext index should keep the index consistent, the same way `INSERT` does, so that newly inserted or updated rows are immediately reflected in index-backed search. This was the behavior before #24044.

### Steps to Reproduce

```sql
drop database if exists mrep;
create database mrep;
use mrep;

create table t(id int primary key, b vecf32(3));
insert into t values(1,'[1,1,1]'),(2,'[2,2,2]'),(3,'[3,3,3]'),(4,'[4,4,4]'),(5,'[5,5,5]');
create index idx using ivfflat on t(b) lists=2 op_type 'vector_l2_ops';

set probe_limit=2;

-- A) REPLACE a brand-new PK row (= insert), vector close to the query point
replace into t values(7,'[5,5,6]');
select id from t order by l2_distance(b,'[5,5,5]') limit 3;   -- returns 5,4,3 ; id=7 NOT found

-- B) Compare: plain INSERT of a brand-new row
insert into t values(8,'[5,4,5]');
select id from t order by l2_distance(b,'[5,5,5]') limit 3;   -- returns 5,8,4 ; id=8 IS found
```

The same gap applies to changing an existing row's vector via `REPLACE`: the KNN result keeps using the old (stale) indexed vector.

### Additional information

Root cause appears to be #24044 (`feat: eliminate legacy REPLACE path, unify to operator-based execution`).

Before #24044, `bind_replace` rejected irregular-index tables and the modern entry point fell back to the legacy path that maintained the index synchronously:

```go
// bind_replace.go (before #24044)
validIndexes, hasIrregularIndex := getValidIndexes(tableDef)
if hasIrregularIndex {
    return 0, nil, nil, moerr.NewUnsupportedDML(builder.GetContext(), "have vector index table")
}
// bindAndOptimizeReplaceQuery: on ErrUnsupportedDML -> fall back to legacy build_replace.go
```

#24044 (1) deleted the legacy `build_replace.go`, (2) removed the `ErrUnsupportedDML` fallback in `bindAndOptimizeReplaceQuery`, and (3) changed `bind_replace` to strip the irregular index (`validIndexes, _ := getValidIndexes(tableDef)`) instead of rejecting it. The modern REPLACE path does not maintain ivfflat/fulltext indexes, so the maintenance is silently skipped.

Note: the equivalent change for the `INSERT` path is being introduced by #24973 (it removes the same `hasIrregularIndex` rejection for inserts); that regression is caught by `array_index_knn` and is being handled in that PR. This issue tracks the pre-existing `REPLACE` gap.

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [x] Code Refactoring

issue #24939
fixes #25000

Replace the legacy `Node_ON_DUPLICATE_KEY` operator with the modern
`DEDUP JOIN` + `MULTI_UPDATE` path for `INSERT ... ON DUPLICATE KEY UPDATE`,
and make real-PK tables align with MySQL.

### Behavior change (real-PK, MySQL alignment)

Previously, on a table with a real primary key, only a primary-key conflict
triggered the update; any unique-key conflict raised a duplicate-entry error.
Now any unique-key conflict (PRIMARY or a secondary UNIQUE key) updates the
first conflicting row, in `PRIMARY > unique-key definition order` — matching
MySQL.

Implementation: real-PK ODKU resolves a single UPDATE target up front via
`target_pk = coalesce(pk-existence-probe, uk1_pri, uk2_pri, ...)` (treating
PRIMARY as the 0th index) and keys the main dedup-update join on `target_pk`.
The conflicting row is updated in place (its primary key and non-`SET` columns
are preserved). The per-unique-key `FAIL` dedup is kept as in-batch duplicate
protection: two brand-new rows that share a new unique-key value in the same
statement still error deterministically (documented gap vs MySQL's in-batch
sequential semantics).

### Modern path coverage

- fake-PK tables (first usable unique key as the anchor) and FK-table ODKU run
  on the modern path; FK parent existence is enforced via generated `DetectSqls`.
- Tables carrying irregular (vector / fulltext) indexes run on the modern path,
  with synchronous index maintenance: INSERT / LOAD / ON DUPLICATE KEY UPDATE /
  REPLACE now keep the IVF/fulltext index consistent, so new or updated rows are
  immediately searchable. This also fixes #25000 (REPLACE not maintaining the
  index).

### Cleanup

Remove the now-dead `Node_ON_DUPLICATE_KEY` enum, the `OnDuplicateKeyCtx` and
`pipeline.OnDuplicateKey` messages, and the `OnDuplicateKey` vm opcode
(reserved in proto).

### Tests

- Plan unit tests for real-PK single / composite unique-key resolution.
- End-to-end behavior cross-checked against MySQL 8.4.
- BVT: `insert_duplicate`, `on_duplicate_key_edge` (rewritten),
  `on_duplicate_key_modern` (real-PK multi-UK + in-batch protection +
  irregular-index cases); the whole `dml/insert` suite passes on the
  multi-CN cluster.


XuPeng-SH (changes requested):
I found two substantive correctness issues in the current head.\n\n1. `pkg/sql/plan/bind_insert.go:869-880` still resolves real-PK ODKU conflicts against the **raw incoming column** for every single-column unique index. That is wrong for prefix unique indexes like `UNIQUE KEY u(col(4))`: the hidden index table stores the truncated/materialized prefix key, so a conflict such as existing `abcdxxxx` vs incoming `abcdyyyy` will not resolve to the existing row here. The later unique-index dedup is prefix-aware, so the statement will still fail with duplicate-entry instead of doing the required UPDATE. The same file already materializes prefix-aware single-column unique keys at `:1140-1163` for locking; target-PK resolution needs to use that materialized `__mo_index_idx_col` value as well, not the raw source column.\n\n2. `pkg/sql/plan/bind_insert.go:227-235` routes every non-regular index into the new “irregular” maintenance path, which includes MASTER indexes. But the delete side in `buildIrregularIndexDeleteMaintenance` (`:531-553`) only handles FULLTEXT and IVF, while the insert side explicitly has a MASTER branch (`:473-476`). That means `ON DUPLICATE KEY UPDATE` / `REPLACE` can insert fresh MASTER-index entries for the new row image without ever deleting the stale old ones, leaving wrong search state behind. Please either add MASTER delete maintenance here or keep MASTER indexes out of `getIrregularIndexes` until the delete path is implemented.\n\nI would also add a targeted ODKU regression case for a single-column unique prefix index once the first bug is fixed.

XuPeng-SH (changes requested):
I re-checked the latest head. The earlier two blockers I raised are addressed, but I still see one remaining correctness issue in the new MASTER/HNSW gate, plus a few concrete follow-up suggestions.

**Blocking issue**

`pkg/sql/plan/bind_insert.go:75-84` and `pkg/sql/plan/bind_replace.go:47-50` now reject any table carrying a legacy-only irregular index (MASTER/HNSW) before the planner determines whether the statement can ever hit an existing row. That is too early.

For fake-PK tables with no usable primary/unique key, this PR already treats:
- `INSERT ... ON DUPLICATE KEY UPDATE` as a plain INSERT / legacy-degenerate case (`bind_insert.go:1040-1049`)
- `REPLACE` as INSERT-like when there is no unique index (`bind_replace.go:99-103`)

In those cases there is no conflict-driven delete/update work at all, so there is no stale-entry problem to avoid. But the new early `hasLegacyOnlyIrregularIndex()` gate makes such statements fail up front anyway on tables with MASTER/HNSW indexes. A table like `create table t(a int, b varchar(100), master index(b)); insert into t values (1,'x') on duplicate key update b='y';` should still be able to behave like INSERT here, not be rejected before key analysis.

**Concrete implementation suggestions**

1. **Move the MASTER/HNSW rejection later, after key analysis.**
   - For ODKU, only reject once you know the statement can actually take the UPDATE path (real PK or a usable unique key exists).
   - Keep the existing `isFakePK && firstUniqueIdxPos < 0` fallback reachable for MASTER/HNSW tables.

2. **Do the same for REPLACE.**
   - First compute whether any unique/PK conflict path exists.
   - If not, keep the current `REPLACE behaves like INSERT` path reachable even on MASTER/HNSW tables.

3. **Add targeted SQL/BVT coverage for the user-visible behavior.**
   I would add all of these while you are here:
   - ODKU on a MASTER/HNSW table **without** PK/UK -> behaves like INSERT
   - REPLACE on a MASTER/HNSW table **without** PK/UK -> behaves like INSERT
   - ODKU on a MASTER/HNSW table **with** a real conflict key -> explicit unsupported error
   - REPLACE on a MASTER/HNSW table **with** a real conflict key -> explicit unsupported error

**Non-blocking but worthwhile remaining coverage suggestions**

- `LOAD DATA` + IVF/fulltext sync maintenance still deserves one end-to-end proof
- FK ODKU update-branch should still get an invalid-parent unhappy-path case

Once the early gate is narrowed to the true conflict-driving cases, I am happy to re-check.

heni02 (approved):

aunjgr (approved):

XuPeng-SH (changes requested):
I re-checked this PR from a first-principles / unhappy-path angle. The overall direction is understandable, but I found two blocker regressions and one important scope mismatch.

1. **`INSERT IGNORE` on FK child tables loses IGNORE semantics**
   - `bind_insert.go` encodes `INSERT IGNORE` as the nil-sentinel shape (`len(OnDuplicateUpdate)==1 && OnDuplicateUpdate[0]==nil`) and classifies it as `Node_IGNORE`.
   - But `build.go` still keys only on `len(stmt.OnDuplicateUpdate) > 0`, so it routes `INSERT IGNORE` through the ODKU FK path.
   - The modern FK path uses `assert(...)` filters for child->parent existence checks, so a missing parent becomes a hard error instead of “skip this row”.
   - Existing behavior for `INSERT IGNORE` on FK violations is to ignore the offending row, not fail the statement (see `test/distributed/cases/dml/insert/insert_ignore.sql` / `.result`).
   - **Suggestion:** distinguish the nil IGNORE sentinel from real ODKU at planner entry. Either keep FK-table `INSERT IGNORE` on the legacy path, or teach the modern FK path to drop/warn offending rows instead of asserting.

2. **ODKU child-FK validation regressed from row-scoped to whole-table checks**
   - For real `INSERT ... ON DUPLICATE KEY UPDATE`, `build.go` now falls back to whole-table `DetectSqls` generated by `genSqlsForCheckFKConstraints(...)`.
   - Those SQLs scan the entire child table (`SELECT DISTINCT ... FROM child ... EXCEPT SELECT DISTINCT ... FROM parent`), not just the statement’s final row image.
   - That means a one-row upsert can fail because of an unrelated orphan row already in the table (for example, one inserted earlier under `FOREIGN_KEY_CHECKS=0`), and validation cost now scales with table size.
   - The comments in `build.go` already acknowledge this exact false-positive problem for plain INSERT; ODKU still takes the whole-table path.
   - **Suggestion:** reuse a row-scoped FK check for the ODKU final image, rather than generating whole-table DetectSqls.

3. **Test coverage is missing the highest-risk unhappy paths**
   - I do not see a regression test for `INSERT IGNORE` on an FK child table staying row-skip / non-failing under the new path.
   - I also do not see a test showing ODKU only validates the statement’s final row image rather than failing on unrelated pre-existing orphan rows.
   - **Suggestion:** add both tests before merge; they directly protect the two regressions above.

Given the semantic regression on `INSERT IGNORE` and the table-wide ODKU FK validation regression, I think this should be **changes requested** before merge.

XuPeng-SH (changes requested):
I re-checked the **current head**.

The earlier blockers around prefix-unique target resolution and MASTER/HNSW gating look improved, but I still see two substantive correctness regressions in the latest version.

1. **`INSERT IGNORE` on FK child tables still loses IGNORE semantics**
   - `bind_insert.go` encodes `INSERT IGNORE` as the nil-sentinel shape (`len(OnDuplicateUpdate)==1 && OnDuplicateUpdate[0]==nil`) and classifies it as `Node_IGNORE`.
   - But `build.go` still keys only on `len(stmt.OnDuplicateUpdate) > 0`, so it routes `INSERT IGNORE` through the ODKU FK path.
   - The modern FK path uses `assert(...)` filters for child->parent existence checks, so a missing parent becomes a hard error instead of “skip this row”.
   - Existing behavior for `INSERT IGNORE` on FK violations is to ignore the offending row, not fail the statement.
   - **Suggestion:** distinguish the nil IGNORE sentinel from real ODKU at planner entry. Either keep FK-table `INSERT IGNORE` on the legacy path, or teach the modern FK path to drop/warn offending rows instead of asserting.

2. **ODKU child-FK validation still regressed from row-scoped to whole-table checks**
   - For real `INSERT ... ON DUPLICATE KEY UPDATE`, `build.go` still falls back to whole-table `DetectSqls` generated by `genSqlsForCheckFKConstraints(...)`.
   - Those SQLs scan the entire child table (`SELECT DISTINCT ... FROM child ... EXCEPT SELECT DISTINCT ... FROM parent`), not just the statement’s final row image.
   - That means a one-row upsert can fail because of an unrelated orphan row already in the table (for example, one inserted earlier under `FOREIGN_KEY_CHECKS=0`), and validation cost scales with table size.
   - **Suggestion:** reuse a row-scoped FK check for the ODKU final image instead of generating whole-table DetectSqls.

### Additional coverage I still want
- a regression for `INSERT IGNORE` on an FK child table staying row-skip / non-failing under the new path
- a regression proving ODKU validates only the statement’s final row image, not unrelated pre-existing orphan rows

So on the **current head**, I still think this should remain **changes requested**.

ck89119:
@XuPeng-SH Thanks for the re-check. Both blocking points were already addressed on the current head (commit `78cfa40ad`, with follow-ups `43dd5e985` / `3476a26b4`); I think the review was against an earlier snapshot, since the code it cites no longer exists. Concrete evidence below.

**1. `INSERT IGNORE` keeps row-skip semantics (does not assert)**

`INSERT IGNORE` is no longer routed through the ODKU FK path:

- It is classified as `Node_IGNORE` at the planner entry via the nil sentinel — `bind_insert.go:1302` (`len(astUpdateExprs)==1 && astUpdateExprs[0]==nil`).
- It is handled by `buildInsertIgnoreFkFilter` — `bind_insert.go:1440-1446`, which **FILTERs out** (drops) the rows whose foreign key has no parent instead of asserting, feeding only surviving rows to the `MULTI_UPDATE`. A NULL FK satisfies MATCH SIMPLE and is kept.
- The `len(stmt.OnDuplicateUpdate)` test in `build.go` is now only the legacy-fallback gate (`build.go:83`); it does not route FK handling at all.

**2. ODKU child→parent FK is row-scoped, not whole-table**

- `genSqlsForCheckFKConstraints` (and `childFkCheckPrefix` / the `CHILD_FK_CHK:` handling) were removed — `git grep genSqlsForCheckFKConstraints` returns nothing.
- ODKU sets `modernFkCheck` over the materialized final merged image (`bind_insert.go:2254`) and the check runs via `appendForeignConstrantPlan` in `finishIrregularIndexMaintenance` (`bind_insert.go:853`) over the statement's own rows only — so it never false-positives on an unrelated orphan row and does not scale with table size.

**3. Regression coverage already present**

`test/distributed/cases/dml/insert/on_duplicate_key_modern.sql` (added in `78cfa40ad`) covers exactly the two unhappy paths you asked for:

- ODKU validates only the final image: seeds an orphan child row under `FOREIGN_KEY_CHECKS=0` (`pid=99`), then proves the valid-row ODKU succeeds (orphan ignored), the FK-column update to a missing parent still fails, and a genuine missing-parent insert still fails.
- `INSERT IGNORE` row-skip: `insert ignore into t_ign_fk_child values (10,1),(11,2),(12,4),(13,1),(14,NULL)` keeps the valid rows and the NULL-FK row, drops `pid=4`.

Plan-level UTs also assert the in-plan behavior: `TestInsertIgnoreChildParentFKDropsRows`, `TestInsertOnDupChildParentFKUsesInPlanCheck`, `TestInsertOnDupFKUsesModernPath`, `TestInsertOnDupSelfReferFKUsesModernPath` (all green).

Could you re-load the latest head and re-check? Happy to add more coverage if anything still looks off.

XuPeng-SH (changes requested):
Re-checked the latest head. The broader MASTER/HNSW gating concerns look improved, but the composite nullable-unique-key issue is still present.

In the current `REPLACE` path, `skipUniqueIdx` only stays true when **every** part of a unique index is statically NULL. That is still too narrow. For a composite `UNIQUE(a,b)`, a row like `(NULL, 1, ...)` cannot conflict with any existing row, because the composite unique key serializes to NULL as soon as any input part is NULL. But the current fake-PK REPLACE privilege gate still treats that table as delete-capable and requires `DELETE`.

So the remaining blocker is unchanged:
- fake-PK + composite nullable unique key
- one component statically NULL
- REPLACE is still insert-only semantically
- frontend still over-requires DELETE

I would keep this at request changes until that composite-null case is handled and covered by a unit/BVT regression.

XuPeng-SH (changes requested):
Re-checked the latest head. I still think this should stay at **request changes**, and this time I’m collecting the remaining blockers together.

1. **Composite nullable unique-key REPLACE still over-requires DELETE.**
   The current static-null shortcut only skips a unique index when **every** part is NULL. For a composite `UNIQUE(a,b)`, a row like `(NULL, 1, ...)` still cannot conflict, because the composite unique key serializes to NULL as soon as any input part is NULL. But the fake-PK REPLACE privilege gate still treats that table as delete-capable and requires `DELETE`.

2. **REPLACE on child FK tables still has no parent-existence check.**
   The modern REPLACE path now allows FK tables, but the build path only wires self-referencing FK checks. I do not see a REPLACE-side equivalent of the child→parent existence validation that plain INSERT / INSERT IGNORE now perform in-flow. So REPLACE can still insert orphan child rows.

3. **ODKU FK checking still skips other FKs when one FK column is NULL.**
   ODKU still routes child→parent FK checking through `appendForeignConstrantPlan`, whose helper first builds one global `isnotnull(...)` filter across **all** FK columns. If one nullable FK column is NULL, the entire row is filtered out of the check step, so other non-NULL FKs on that same row are never validated.

**Concrete suggestions**

- For (1), treat a composite unique index as insert-only when any nullable part is statically NULL, and add a unit/BVT case for that path.
- For (2), run a child→parent FK existence check for REPLACE’s final inserted image, analogous to the modern INSERT path.
- For (3), move ODKU child→parent FK checking onto the same per-FK MARK/null-aware logic you now use for INSERT / INSERT IGNORE, instead of one global NULL prefilter.

ck89119:
@XuPeng-SH Thanks. I've handled the composite-null case in `359de752a`. Summary of what changed, plus two points where I took a different tack than the literal suggestion — happy to discuss.

**Fixed: a NULL composite unique key is now consistent between INSERT and REPLACE**

You're right that `skipUniqueIdx` was too narrow and, on the REPLACE path, was never actually consumed. Concretely, for a fake-PK table with `UNIQUE(a, b)` and a statically-NULL part, REPLACE was still probing and **maintaining** that unique index — inserting a NULL-key entry into the index table that the INSERT path deliberately skips (`bind_insert.go:1893`). That INSERT/REPLACE divergence is the real bug here.

- `skipUniqueIdx` is now "**any** part statically NULL" (unique indexes only; `serial(...)` is NULL as soon as one part is NULL; non-unique indexes use `serial_full` and are never skipped).
- It is now consumed in the unique-conflict probe, the old-RowID fetch, and the index `MULTI_UPDATE`, so a NULL-key unique index is neither probed nor maintained — matching INSERT.

Coverage: unit test `TestReplaceFakePKCompositeNullableUKSkipsNullKeyIndex` (asserts the NULL-key unique index is not maintained, while a fully non-NULL key still is) + a BVT in `dml/replace/replace.test` (no mis-delete, multiple NULL keys coexisting, a real conflict replacing).

**Two deliberate deviations — would like your read**

1. **The plan is not *structurally* insert-only.** I did evaluate routing this REPLACE through the existing fake-PK insert-only branch (which drops the main-table LEFT JOIN). That branch doesn't populate the per-index metadata (`idxTableDefs`/`idxObjRefs`), while the shared post-branch code has several index loops that require it, so it nil-panics for any table that actually carries a unique index — a non-trivial refactor (multi-part serial old-cols, non-unique maintenance). Instead I kept the main-table LEFT JOIN: its join condition includes the NULL part, so it never matches, no existing row is deleted, and the row is inserted. So it is insert-only *semantically at runtime*, just not structurally; the redundant index-table scans and the DEDUP join are gone. If you consider the structural form a blocker, I'm glad to do the full insert-only refactor as a follow-up.

2. **Privilege still requires INSERT+DELETE.** `REPLACE` privilege is decided from the AST (`determinePrivilegeSetOfStatement`), unconditionally `INSERT+DELETE`, matching MySQL ("you must have both the INSERT and DELETE privileges"). Making it depend on whether a particular statement's values happen to be NULL would make the required privilege vary per-statement for the same template — unpredictable for auditing and a divergence from MySQL. My preference is not to special-case that; if you feel insert-only REPLACE should drop DELETE, I'd suggest treating it as a separate, explicit design decision rather than folding it in here.

Let me know your thoughts on (1) and (2).

XuPeng-SH (changes requested):
I re-checked the latest head from the FK / unhappy-path angle. The legacy ODKU operator removal is moving in the right direction, but I still see two correctness blockers before I can approve.

Blocker 1: REPLACE on a non-self child FK table does not check parent existence

`bindReplace()` temporarily sets `IgnoreForeignKey` so REPLACE can use the modern path, but the modern REPLACE path never adds a child->parent parent-existence check for non-self FKs. After `createQuery()`, `bindAndOptimizeReplaceQuery()` only generates self-referencing FK DetectSqls / self-ref pre-check SQLs. There is no equivalent of the INSERT/ODKU row-scoped parent check for normal child FKs.

So a statement like:

```sql
create table p(id int primary key);
create table c(id int primary key, pid int, foreign key(pid) references p(id));
replace into c values (1, 999);
```

can be planned without any parent-existence assert. That should fail with the normal child-row FK error. Please add a regression test for both insert-only REPLACE and conflict REPLACE producing a final child row that references a missing parent.

Blocker 2: ODKU final-image FK check can still skip a valid FK check when another FK column is NULL

ODKU sets `builder.modernFkCheck` and then `finishIrregularIndexMaintenance()` calls the old `appendForeignConstrantPlan()` over the materialized final image. That helper first builds a single FILTER containing `isnotnull` for every FK column across all FKs before it does the parent joins.

That is not per-FK MATCH SIMPLE semantics. With two child FKs, if FK1 is NULL and FK2 is non-NULL but points to a missing parent, the row is filtered out before FK2's parent join/assert runs. The statement can pass even though FK2 should fail.

Please replace this ODKU check with the same per-FK MARK-join logic used by the new INSERT / INSERT IGNORE path, or otherwise make the NULL handling per FK rather than a global pre-filter. A focused regression test should cover: one FK column NULL, another FK column non-NULL and invalid -> ODKU must fail; the same NULL FK with the other FK valid -> must pass.

The existing planner tests pass locally, but they do not cover either of these cases.

XuPeng-SH (approved):
Re-checked current head `c2c95bc7b05fac03bd322affd826827c7b1ef1d8` from correctness and unhappy-path angles.

The previous blockers I raised look addressed now:

- `INSERT IGNORE` on child FK tables now uses the in-flow MARK-join FK filter, preserving row-skip semantics instead of failing the whole statement.
- ODKU child->parent FK checking is row-scoped over the final merged image, so it no longer false-positives on unrelated orphan rows and the multi-FK MATCH SIMPLE case is checked per FK.
- REPLACE now validates the inserted child row against its parent in the modern path, covering both insert-only and conflict REPLACE.
- The composite nullable unique-key REPLACE path now skips statically NULL unique keys correctly; runtime INSERT/ODKU NULL-key filtering still happens at index-table insert time.

I also checked the latest added BVT/UT coverage around orphan rows, INSERT IGNORE row skipping, multi-FK NULL handling, and REPLACE FK enforcement. PR CI is green. Local targeted `go test ./pkg/sql/plan ...` still cannot build in my environment because native headers (`usearch.h`, `xxhash.h`) are missing, so I am relying on code review plus CI for execution coverage.

Approved from my side.

PR #24974, opened 2026-06-13
fix(dev): make 'make dev-up' work — readable temp compose files + HAKeeper bootstrap

fengttt:
## What

Two fixes that let the `etc/docker-multi-cn-local-disk` stack (`make dev-up`) actually start a working multi-CN cluster. The first failure masked the second, so both are needed.

### 1. Temp compose fragments written to `/tmp` (`start.sh`)
`start.sh` generated the default-limits and extra-mounts compose fragments under `/tmp`. A **snap-confined Docker daemon cannot read `/tmp`**, so `docker compose` failed before any container started:

```
open /tmp/docker-compose.default-limits.<pid>.yml: no such file or directory
```

Now the fragments are written in the project directory (under `$HOME`, which the daemon can access) using relative paths, so docker compose keeps it as the project directory. Also fixes a latent cleanup bug: the two per-block `trap ... EXIT` handlers clobbered each other (the second overwrote the first), leaking the default-limits file whenever extra mounts were used — both temp files are now tracked in one list and removed by a single trap.

### 2. HAKeeper never bootstrapped (`generate-config.sh`)
The generated `log.toml` had no `[logservice]` section. Each node is started separately with `-cfg log.toml` (not `-launch`), so nothing ever bootstrapped HAKeeper and the log store panicked on first start:

```
panic: shard not bootstrapped
```

The launch stacks avoid this only because `-launch` builds the bootstrap config programmatically; no checked-in toml sets `bootstrap-cluster`. (This was invisible until fix #1, since compose always died at the `/tmp` step before a container ran.)

`generate-config.sh` now emits a `[logservice]` section that gives `mo-log` a UUID and bootstraps a single-replica HAKeeper (1 log shard, 1 tn shard, 1 replica, init member `131072:<uuid>`), matching `pkg/embed/cluster.go` and `logservice.DefaultConfig`. Addresses advertise the `mo-log` hostname while listening on `0.0.0.0`, and `gossip-allow-self-as-seed` is set because `Fill()` (unlike `DefaultConfig`) does not default it — required for a lone seed node. The `*.toml` files are gitignored (generated), so the fix lives in the generator.

## Verification

From a clean state (`make dev-clean && make dev-up`), validated through the **generated** config path (tomls removed → regenerated by `generate-config.sh` → fresh bring-up):
- all 5 containers up, `mo-log` reports healthy, **zero `shard not bootstrapped` panics**
- `create / insert / select sum()` round-trips through the proxy on port 6001


XuPeng-SH (changes requested):
Found a rollout hole in the config-generation fix.

The new `[logservice]` / `BootstrapConfig` block only exists in `etc/docker-multi-cn-local-disk/generate-config.sh:297-314`, but `etc/docker-multi-cn-local-disk/start.sh:297-320` regenerates configs only when the `*.toml` files are missing or `config.env` exists. Users who already ran the broken stack keep their old generated `log.toml`, so after updating they get the `/tmp` compose-file fix but still never pick up the HAKeeper bootstrap change. `make dev-clean` does not help here because it removes only `mo-data` and `logs` (`Makefile:749-760`), while the generated `etc/docker-multi-cn-local-disk/*.toml` files are persisted and gitignored (`.gitignore:73-75`).

Please either force-regenerate managed configs when `log.toml` lacks the new `[logservice]` section, or make the supported cleanup path remove those generated `*.toml` files so the updated generator is actually applied.

XuPeng-SH:
Suggested direction: make `start.sh` self-heal stale generated configs, instead of requiring users to discover they need to delete gitignored `*.toml` files manually. For example, after the existing missing-config check, treat an old `log.toml` without the new `[logservice]` block as stale and regenerate:

```sh
STALE_CONFIGS=false
if [ -f "log.toml" ] && ! grep -q ^\[logservice\] log.toml; then\n    STALE_CONFIGS=true\nfi\n\nif [ "$MISSING_CONFIGS" = true ] || [ "$STALE_CONFIGS" = true ]; then\n    echo "Configuration files missing or stale. Regenerating..."\n    ./generate-config.sh\n    echo ""\nelif [ -f "config.env" ]; then\n    echo "Regenerating configuration from config.env..."\n    ./generate-config.sh\n    echo ""\nfi\n```\n\nThat keeps the fix rollout-safe for users who already generated the old broken `log.toml`. If you prefer not to add staleness detection in `start.sh`, then the supported cleanup path should remove `etc/docker-multi-cn-local-disk/*.toml` so `make dev-clean && make dev-up` reliably picks up generator changes.

**Comment by fengttt** [comment]
@XuPeng-SH Good catch — fixed in 95dd1ac6d9.

I went with the force-regenerate option (rather than deleting the generated `*.toml` in cleanup), since it doesn't require users to touch their data and applies automatically:

`start.sh` now detects a `log.toml` that predates the fix (no `[logservice]` section) and force-regenerates all managed configs before bringing the stack up:

```sh
STALE_CONFIGS=false
if [ "$MISSING_CONFIGS" = false ] && ! grep -q '^\[logservice\]' log.toml; then
    STALE_CONFIGS=true
fi
...
elif [ "$STALE_CONFIGS" = true ]; then
    echo "Existing log.toml predates the HAKeeper bootstrap fix (no [logservice] section)."
    echo "Regenerating configuration so the cluster can bootstrap..."
    ./generate-config.sh
fi
```

So a user who already ran the broken stack now picks up the HAKeeper bootstrap change on the next `make dev-up`, even though the gitignored `*.toml` files survive `make dev-clean`.

XuPeng-SH (changes requested):
The stale-config regeneration hole is fixed on the latest head, but I found one new correctness issue in `start.sh`.\n\n`etc/docker-multi-cn-local-disk/start.sh:212-216` installs an EXIT trap whose body is just `[ ${#TMP_COMPOSE_FILES[@]} -gt 0 ] && rm -f ...`. When no temp compose fragments were created, that test returns status 1, so the trap itself exits non-zero. Under `set -e`, that turns some otherwise successful `start.sh` runs into exit code 1. Reproducing the trap logic directly shows the problem: `bash -lc 'set -e; TMP_COMPOSE_FILES=(); cleanup_tmp_compose(){ [ ${#TMP_COMPOSE_FILES[@]} -gt 0 ] && rm -f "${TMP_COMPOSE_FILES[@]}"; }; trap cleanup_tmp_compose EXIT; : '; echo EXIT:$?` returns `EXIT:1`.\n\nPlease make the trap handler always succeed (for example, guard the `rm` with an `if` block and/or end the function with `return 0`).

XuPeng-SH:
Suggested fix for the new EXIT-trap issue in `start.sh`:

```sh
cleanup_tmp_compose() {
    if [ ${#TMP_COMPOSE_FILES[@]} -gt 0 ]; then
        rm -f "${TMP_COMPOSE_FILES[@]}"
    fi
    return 0
}
```

The important part is that the trap handler must always exit successfully when there is nothing to clean up; otherwise some successful `start.sh` runs still end with exit code 1 under `set -e`.

fengttt:
@XuPeng-SH Fixed in 85c6450b24.

`cleanup_tmp_compose` now guards the `rm` with an `if` block and ends with `return 0`, so the EXIT trap always exits successfully even when no temp compose fragments were created:

```sh
cleanup_tmp_compose() {
    if [ ${#TMP_COMPOSE_FILES[@]} -gt 0 ]; then
        rm -f "${TMP_COMPOSE_FILES[@]}"
    fi
    return 0
}
```

Verified with your repro — `trap ... EXIT; :` now yields `EXIT:0` instead of `EXIT:1` under `set -e`.

XuPeng-SH (approved):
Re-checked the latest head. The stale-config rollout hole is now fixed by regenerating old `log.toml` files that lack `[logservice]`, and the new `cleanup_tmp_compose()` trap fix correctly preserves exit code 0 when no temp fragments were created while still cleaning up when they are. I do not see a remaining correctness issue that should keep this blocked.

Issue #24762, opened 2026-06-01
[Bug]: 一些ddl操作之后，字符串列进行字符串比较失效。只能通过binary方式比较


ouyuanning:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main 

### Commit ID

1f7bac5b0eca05d9684bdbc6d1471282db3b96a6

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

taas代码升级多了一个ddl操作后。一些varchar(64)的列。无法进行 col = 'xxx'的比较。只能通过 binary col = binary 'xxx' 来做比较

<img width="680" height="348" alt="Image" src="https://github.com/user-attachments/assets/d80721d7-c5f6-4246-bdf8-330ccce86548" />

<img width="667" height="400" alt="Image" src="https://github.com/user-attachments/assets/9dd852bf-960c-49d2-853e-6bd0da77858d" />

### Steps to Reproduce

```Markdown
同上
```

ouyuanning:
mysql> explain analyze select count(*) from token_usage_logs where provider_id='provi
    '> der-c67f37e3-f50f-461c-a6b4-1551e8f347da';
+--------------------------------------------------------------------------------------------------------------------------------------------------+
| tp query plan                                                                                                                                    |
+--------------------------------------------------------------------------------------------------------------------------------------------------+
| Project                                                                                                                                          |
|   Analyze: timeConsumed=0ms waitTime=0ms inputRows=1 outputRows=1 InputSize=8bytes OutputSize=8bytes MemorySize=8bytes                           |
|   ->  Aggregate                                                                                                                                  |
|         Analyze: timeConsumed=0ms waitTime=0ms inputRows=0 outputRows=1 InputSize=0bytes OutputSize=8bytes MemorySize=0bytes                     |
|         Aggregate Functions: starcount(1)                                                                                                        |
|         ->  Index Table Scan on token_usage_logs.idx_usage_provider_called                                                                       |
|               Analyze: timeConsumed=0ms waitTime=0ms inputBlocks=0 inputRows=0 outputRows=0 InputSize=0bytes OutputSize=0bytes MemorySize=0bytes |
|               Filter Cond: prefix_eq(__mo_index_idx_col)                                                                                         |
+--------------------------------------------------------------------------------------------------------------------------------------------------+
8 rows in set (0.01 sec)

[...]

PR #24976, opened 2026-06-15
fix: rebuild txn-local secondary indexes in alter copy

ouyuanning:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #24762

When multiple ALTER TABLE COPY DDLs run in the same explicit transaction, a later DDL can clone a hidden secondary index table that was created or written by an earlier DDL in the same transaction. table_clone reads from the snapshot path and cannot see txn-local workspace rows, so the newly cloned hidden index table can miss old rows. Queries that use the secondary index then miss data while base-table scans still find it.

This PR detects txn-local source hidden index tables during cloneUnaffectedIndexes. For regular secondary, unique, master, and sync fulltext indexes, it rebuilds the target hidden index table from the current copied base table instead of using table_clone. Master indexes keep their dedicated backfill SQL, and sync fulltext uses fulltext_index_tokenize instead of the generic serial_full secondary-index SQL. Async fulltext remains on the existing skip/ISCP flow because its index table may be incomplete during ALTER COPY.

It adds UT coverage for the txn-local rebuild predicate and generated fulltext rebuild SQL, plus distributed SQL regressions covering:

- the varchar composite secondary index repro from the issue
- the same txn-local clone problem for USING MASTER indexes
- txn-local hidden index writes before ALTER COPY in the same transaction
- sync fulltext search before/after begin; two COPY DDLs; commit; and after inserting a new row

Tests:

- make build
- go test ./pkg/vm/engine/disttae -run TestTransaction -count=1 with the same CGO include/rpath settings as make build
- go test ./pkg/sql/compile -run TestShouldRebuildTxnLocalIndexClone -count=1 with the same CGO include/rpath settings as make build
- go test ./pkg/sql/compile -run TestShouldRebuildTxnLocalIndexClone\|TestGenRebuildTxnLocalIndexCloneSQLsFullText -count=1 with the same CGO include/rpath settings as make build
- Manually ran test/distributed/cases/ddl/alter_copy_secondary_index_txn.sql against a rebuilt local mo-service; all counts matched the expected result
- Manually verified the focused sync fulltext repro now returns 1 -> 1 -> 2 for match(...) counts, matching the table-scan counts

### DIFF
--- a/pkg/sql/compile/alter.go
+++ b/pkg/sql/compile/alter.go
@@ -1229,15 +1229,11 @@ func cloneUnaffectedIndexes(
 					return moerr.NewInternalErrorNoCtxf("missing cloned index definition for %s", idxName)
 				}
 				logutil.Infof("cloneUnaffectedIndex: rebuild txn-local index %s into %s\n", idxName, newIdxTblName.IndexTableName)
-				if catalog.IsMasterIndexAlgo(oriIdxTblNames.IndexAlgo) {
-					insertSQLs := genInsertIndexTableSqlForMasterIndex(newTblDef, newIdxDef, dbName)
-					for _, insertSQL := range insertSQLs {
-						if err = c.runSql(insertSQL); err != nil {
-							return err
-						}
-					}
-				} else {
-					insertSQL := genInsertIndexTableSql(newTblDef, newIdxDef, dbName, oriIdxTblNames.Unique)
+				insertSQLs, err := genRebuildTxnLocalIndexCloneSQLs(newTblDef, newIdxDef, dbName, oriIdxTblNames)
+				if err != nil {
+					return err
+				}
+				for _, insertSQL := range insertSQLs {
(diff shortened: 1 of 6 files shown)


heni02 (approved):

XuPeng-SH (changes requested):
Blocking on required CI: this PR currently has multiple required BVT failures, including standalone optimistic, standalone pessimistic, and compose optimistic checks. Please fix the failures or rerun/prove them unrelated before this can be approved. Static review did not surface an additional code-level blocker, and the added UT/BVT coverage looks aligned with the ALTER COPY secondary-index fix, but the failing required checks must be resolved first.

ouyuanning:
clone方式修复了。

Issue #24977, opened 2026-06-15
CI: Matrixone Utils CI / Coverage failed — TestSyncAddOneModel timeout in pkg/vectorindex/hnsw


XuPeng-SH:
## CI Failure

- **Workflow**: [MatrixOne ALL CI](https://github.com/matrixorigin/matrixone/actions/runs/27518288188) (attempt 1)
- **Job**: [Matrixone Utils CI / Coverage](https://github.com/matrixorigin/matrixone/actions/runs/27518288188/job/81331050418) (job 81331050418)
- **Branch**: `xupeng/fix-pipeline-cleanup-hang-24962` (PR #24972)
- **Commit**: `776ff97`

## Failed Steps

| Step | Name | Conclusion |
|------|------|-----------|
| 9 | Start Unit Test | ❌ failure (exit code 1) |
| 10 | Analysis Fail UT Cases | ❌ failure (exit code 1) |
| 16 | Generate UT and BVT Coverage File | ❌ failure (exit code 1, cascading from step 9) |

Steps 13-15 (BVT) and step 20 (coverage script) were skipped due to UT failure.

## Failing Test

```
FAIL_UT_CASES=TestSyncAddOneModel,
```

**Test**: `TestSyncAddOneModel` in `pkg/vectorindex/hnsw/sync_test.go:440`

**Error**: `panic: test timed out after 10m0s`

- Test ran for **5m52s** before being killed by the global 10-minute test timeout
- The test inserts 1,000,000 vectors
- Goroutine 1200392 was stuck at `sync.WaitGroup.Wait` in `insertAllInParallel` (`sync.go:387`)

### Stack trace (goroutine 1200392)

```
sync.runtime_SemacquireWaitGroup
sync.(*WaitGroup).Wait
  sync/waitgroup.go:206
(*HnswSync[...]).insertAllInParallel
  pkg/vectorindex/hnsw/sync.go:387
(*HnswSync[...]).Update
  pkg/vectorindex/hnsw/sync.go:535
(*HnswSync[...]).RunOnce
  pkg/vectorindex/hnsw/sync.go:60
TestSyncAddOneModel
  pkg/vectorindex/hnsw/sync_test.go:440
```

The goroutine was stuck at `insertAllInParallel` → `WaitGroup.Wait` for 5+ minutes, suggesting some parallel insert goroutines never completed.

## Note

[...]

PR #24993, opened 2026-06-15
fix(hnsw): force single-thread build on the CDC/sync write path

cpegeric:
USearch #735 / MatrixOne #24849 (open): concurrent USearch add() can orphan HNSW graph nodes — the vector is stored (contains()==true) but never linked into the graph, so search() can never reach it, producing flaky recall@1. The create-index path (build.go) already forces a single build thread, but the CDC/sync path (HnswSync) still drove insertAllInParallel with ThreadsBuild (= NumCPU by default), issuing concurrent AddUnsafe() to the same index.

- types.go: add GetConcurrencyForSingleThreadBuild (returns 1) with a one-line revert note for when usearch fixes the race. Scoped to HNSW — ivfflat (ivf_create.go) keeps GetConcurrencyForBuild and stays parallel.
- sync.go: NewHnswSync routes ThreadsBuild through the new helper (both the sysvar and default branches), with an inline comment at each call site. The hnsw_threads_build / hnsw_max_index_capacity reads are untouched.
- model.go: revert NewHnswModelForBuild to honor its nthread param again (commit 57a991ec6 had hardcoded nthread:=1). With both callers now passing 1 (build.go forces 1; sync uses the helper) the hardcode is redundant, and removing it avoids a hidden override that would silently keep the model single-threaded even after the helper/build.go are reverted.

Net: AddUnsafe() is single-threaded on all build/write paths (create-index, sequentialUpdate, insertAllInParallel with nthread=1 -> one goroutine; each also ChangeThreadsAdd(1)). Verified: insertAllInParallel nthread=1 at runtime; TestSyncAddOneModel passes single-threaded with no hang.

## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24977

fix the bug to keep ThreadsBuild to 1 in all build path.

### DIFF
--- a/pkg/vectorindex/hnsw/model.go
+++ b/pkg/vectorindex/hnsw/model.go
@@ -63,19 +63,9 @@ type HnswModel[T types.RealNumbers] struct {
 }
 
 // New HnswModel struct
-func NewHnswModelForBuild[T types.RealNumbers](id string, cfg vectorindex.IndexConfig, _ int, max_capacity uint) (*HnswModel[T], error) {
+func NewHnswModelForBuild[T types.RealNumbers](id string, cfg vectorindex.IndexConfig, nthread int, max_capacity uint) (*HnswModel[T], error) {
 	var err error
 	idx := &HnswModel[T]{}
-
-	// MatrixOne #24849 / USearch #735 (open): concurrent add() can orphan nodes --
-	// the vector is stored (contains() returns true) but is never linked into the
-	// HNSW graph, so search() can never reach it, producing flaky recall@1 (an
-	// exact match is intermittently missed). NewHnswBuild forces a single build
-	// thread for the create-index path, but the CDC / sync path still creates
-	// models with ThreadsBuild here. Pin every build model to a single thread so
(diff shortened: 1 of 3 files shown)


XuPeng-SH (changes requested):
Requesting changes for one missing regression guard around this workaround.

The code change itself looks directionally right, but this PR moves the single-thread safety guarantee from `NewHnswModelForBuild(...)` back out to the HNSW sync call sites. Given the underlying bug is silent recall corruption, I think we need a deterministic test that locks this invariant down.

Today `sync_test.go` sets `hnsw_threads_build = 8`, but it only checks that `RunOnce` succeeds; it does not assert that `NewHnswSync` actually clamps `ThreadsBuild` to `1`, nor that models created on this path end up with `NThread == 1`.

Please add a focused regression test for the sync/CDC path before approval.

XuPeng-SH (approved):

PR #24999, opened 2026-06-16
fix-qcloud-cos-retryable-status

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24998

Fix qcloud cos retryable status


XuPeng-SH (changes requested):
Requesting changes for one retry-classification gap.

This PR adds COS-specific retry handling for 429 / 502 / 503 / 504, but it still leaves `500 Internal Server Error` non-retryable. Since MatrixOne routes QCloud operations through `DoWithRetry(..., IsRetryableError)`, that means transient COS 500s will still fail immediately instead of retrying.

I think `http.StatusInternalServerError` should be included in the retryable COS status set, and the new status-based test should cover 500 as well.

XuPeng-SH (approved):

PR #25002, opened 2026-06-16
Fix CDC URI parsing for Cloud account usernames and special characters

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24833 

## What Changed

  This PR fixes CDC connection handling for MatrixOne Cloud-style credentials, where the username may contain multiple colons such as `account:user:role`, and the password may contain special characters
  such as `@`.

  Changes include:
  - Parse CDC URIs using the last `@` as the user-info/host separator.
  - Split user and password using the last `:` in user-info, preserving usernames like `account:user:role`.
  - Decode percent-encoded username/password values.
  - Build MySQL connections with `mysql.Config` and `mysql.NewConnector` instead of manually formatting a DSN string, avoiding DSN parsing issues with `:` in usernames.
  - Add tests for Cloud-style usernames, encoded credentials, passwords containing `@`, and connection config generation.

  ## Testing

  - `gofmt`
  - `git diff --check`

  `go test ./pkg/cdc -run 'Test_(compUriInfo|makeMysqlConfig|openDbConn|tryConn)' -count=1` could not complete in my local environment because required CGO headers are missing:
  - `usearch.h`
  - `xxhash.h`

  `CGO_ENABLED=0` is also blocked by `mpool` requiring CGO-backed allocator symbols.


XuPeng-SH (changes requested):
Requesting changes because the updated test code does not compile as written.

`openDbWithConnector` is now assigned from `sql.OpenDB`, so its signature is `func(driver.Connector) *sql.DB`.
But in `Test_tryConn`, the new gostub callback still returns `db, nil` instead of just `db`.

So the test stub has the wrong arity and the file will fail to compile. I think the stub body just needs to return `db`.

XuPeng-SH (changes requested):
This refactor fixes URI parsing, but it also switches CDC connection creation to the upstream `github.com/go-sql-driver/mysql` connector while the executor still depends on MatrixOrigin's fork-specific `ReuseQueryBuf` named argument.

`pkg/cdc/util.go` now builds the connection with `mysql.NewConnector(cfg)` from the upstream driver, but `pkg/cdc/sinker_v2_executor.go` still executes SQL with `sql.NamedArg{Name: mysql.ReuseQueryBuf, ...}` from `github.com/matrixorigin/mysql`. The upstream driver rejects non-empty named parameters, so CDC can connect and then fail on the first `Exec`/`ExecTx`.

Please keep connection creation and execution on the same MatrixOrigin driver path (for example by using `github.com/matrixorigin/mysql` here as well), and add a regression that exercises an actual execute path after opening the connection.

XuPeng-SH (approved):
LGTM. The current head fixes the previous blockers I requested changes for:

- Connection creation now stays on github.com/matrixorigin/mysql.NewConnector, matching the executor's mysql.ReuseQueryBuf named-argument path.
- The tryConn/openDbWithConnector test stub now matches sql.OpenDB's actual signature.
- URI parsing now uses the terminal @ as the host separator and the terminal : inside user-info as the user/password separator, with percent-decoding for both username and password. That covers MatrixOne Cloud-style account:user:role credentials and passwords containing @.
- The new fake MySQL-server regression exercises ExecSQL after tryConn and verifies that a COM_QUERY is emitted, so the connector + ReuseQueryBuf execution path is covered instead of only testing config construction.

Local verification on the current head:

CGO-enabled go test ./pkg/cdc -run 'Test_(compUriInfo|makeMysqlConfig|openDbConn|openDbConnFailed|tryConn)|TestExecutor_ExecSQLAfterTryConnUsesReuseQueryBuf' -count=1

One remaining non-code blocker is external to this change: CLA assistant is still pending, and the branch is currently behind main.

Issue #24832, opened 2026-06-03
[Bug] CDC command syntax is inconsistent (SHOW CDC needs ALL; DROP CDC <name>/IF EXISTS not accepted)


dengn:
## Summary
CDC command syntax is inconsistent / unintuitive: `SHOW CDC` alone errors (requires `SHOW CDC ALL`); `DROP CDC <name>` and `DROP CDC IF EXISTS <name>` both error (only `DROP CDC ALL` and `DROP CDC TASK <name>` are accepted).

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test cases
```sql
SHOW CDC;                  -- ERROR 1064 (needs: SHOW CDC ALL)
DROP CDC cdc1;             -- ERROR 1064
DROP CDC IF EXISTS cdc1;   -- ERROR 1064
DROP CDC ALL;              -- works
DROP CDC TASK cdc1;        -- works
```

## Expected
Consistent and discoverable CDC verbs (support `SHOW CDC`, `DROP CDC <name>`, and `IF EXISTS`).

## Version status
- v3.0.11 (Cloud): reproduces
- main / 4.0-dev (commit 0bfda60): reproduces

XuPeng-SH:
@heni02 please help to verify

Ariznawlll:
Verified on latest main.

Environment:
- Branch: main
- Commit: a0f1006fb (feat: add date_trunc function for time-bucketing rollups (#24627))

Simple syntax probe results:

```sql
SHOW CDC;
```
Result: succeeds, returns empty set.

```sql
SHOW CDC ALL;
```
Result: succeeds, returns empty set.

```sql
DROP CDC IF EXISTS cdc_not_exist;
```
Result: succeeds, Query OK.

```sql
DROP CDC cdc_not_exist;
```
Result: syntax is accepted; fails with a semantic error because the task does not exist:

```text
ERROR 20101 (HY000): internal error: no cdc task found, accountId: 0, taskName: cdc_not_exist
```

[...]

PR #25003, opened 2026-06-16
 fix: support compatible CDC show and drop syntax

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24832

## What
  - Support `SHOW CDC` as shorthand for `SHOW CDC ALL`.
  - Support `DROP CDC <name>` as shorthand for `DROP CDC TASK <name>`.
  - Support `DROP CDC IF EXISTS <name>` and ignore missing CDC tasks when `IF EXISTS` is specified.
  - Pass `DropCDC.IfExists` through AST, plan, compile, and frontend CDC handling paths.

  ## Testing
  - `go test ./pkg/sql/parsers/dialect/mysql`
  - `git diff --check`


XuPeng-SH (changes requested):
Requesting changes for one parser compatibility regression.

The new shorthand `DROP CDC <name>` is not actually equivalent to `DROP CDC TASK <name>` for keyword-like task names.

`CREATE CDC` still uses `ident` for task names, and `DROP CDC TASK <name>` also goes through `ident`, so those paths accept `ID`, `QUOTE_ID`, `not_keyword`, and `non_reserved_keyword` names. But the new shorthand branch uses `cdc_task_name`, which only accepts `ID | QUOTE_ID`.

So some task names that are currently legal in create / long-form drop can fail in the new shorthand drop. I think the shorthand branch should reuse `ident`, and it would be good to add a parser test with a non-reserved-keyword task name.

aunjgr (approved):
IfExists flows correctly: parser->AST->plan->compile->frontend->DAO. reset() zeroes all fields including IfExists. Grammar uses ident for task names (supports keyword-named tasks, proven by 'drop cdc internal' test). Clean.

XuPeng-SH (approved):
Re-checked the latest head from the parser/semantic unhappy-path angle. The earlier shorthand parser regression is fixed, `DROP CDC IF EXISTS` now flows through AST/plan/compile/frontend correctly, and the frontend path explicitly ignores missing tasks only when `IF EXISTS` is set. I do not see a remaining correctness issue that should keep this blocked.

iamlinjunhong (approved):

PR #25004, opened 2026-06-16
fix: add mysql user-level lock functions

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24728

  - Add MySQL-compatible `GET_LOCK`, `RELEASE_LOCK`, and `IS_FREE_LOCK` built-in functions.
  - Register function IDs, name mappings, and overloads for SQL planning.
  - Implement process-local named user-level lock behavior keyed by connection ID.
  - Add tests for acquire, release, free-state checks, non-owner release, missing locks, and NULL inputs.
  - Add function lookup coverage for the new built-ins.

### DIFF
--- a/pkg/fileservice/error.go
+++ b/pkg/fileservice/error.go
@@ -15,6 +15,7 @@
 package fileservice
 
 import (
+	"context"
 	"errors"
 	"io"
 	"net"
@@ -25,6 +26,10 @@ import (
 )
 
 func IsRetryableError(err error) bool {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return false
+	}
+
 	// unexpected EOF
 	if errors.Is(err, io.ErrUnexpectedEOF) {
 		return true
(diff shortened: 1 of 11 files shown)


XuPeng-SH (changes requested):
I found one MySQL-compatibility issue in the unhappy-path handling for user-level lock names.

`normalizeUserLevelLockName()` currently rejects empty-string names by returning an execution error, and `GET_LOCK` / `RELEASE_LOCK` / `IS_FREE_LOCK` propagate that error. But MySQL-compatible behavior here is to return `NULL` for empty-string lock names, not fail the statement. The current unit tests also lock in the error behavior, so this would be a visible compatibility regression for clients that rely on MySQL semantics.

Please treat `` the same way as other invalid lock-name inputs in these SQL functions: return `NULL` instead of raising an execution error.

PR #25016, opened 2026-06-16
fix: maintain vector/fulltext index synchronously on REPLACE (#25000)

ck89119:
## What type of PR is this?

- [x] BUG

issue #25000

`REPLACE` into a table carrying an irregular index (ivfflat vector index or fulltext index) succeeded but did **not** maintain the index. As a result, a row inserted (or whose indexed value changed) via `REPLACE` was not reflected in a subsequent index-backed ANN/KNN or `MATCH ... AGAINST` query until the next cron reindex. `INSERT` was not affected.

Root cause: the modern operator-based `REPLACE` path (DEDUP JOIN + MULTI_UPDATE, introduced by #24044) stripped irregular indexes via `getValidIndexes` and relied on async maintenance. This PR restores synchronous maintenance for **non-async** ivfflat and fulltext indexes, leaving async ones to the cron exactly as the legacy insert/update paths do.

### How

- Collect the non-async ivf / fulltext index defs before they are stripped from `tableDef.Indexes`.
- Sink the processed new rows once (a tag-preserving `SINK_SCAN`) so auto-increment / generated columns are evaluated a single time and the same rows feed both the main conflict-resolution branch and the index-maintenance sub-plans.
- Build the index sub-plans **after** `createQuery` (the same sequencing the legacy insert path uses): they use positional column refs and hand-built nodes, so they must not go through the per-step optimize/remap pipeline. Only the newly appended steps are finalized via `handleHashMapMessages`.
- **ivf**: insert entries for every incoming row, and delete the stale entries of conflicting rows. On a primary-key conflict the new PK equals the replaced row's PK, so the old entry keyed by `org_pk` is removed; a brand-new PK matches nothing and deletes nothing.
- **fulltext**: a single `POSTDML` node per index deletes the stale inverted entries (by doc id = the incoming PK) and re-tokenizes the new rows.

Regular tables and regular secondary indexes are unaffected.

Known limitation: for the rare case of a table with both an irregular index and a unique secondary index, a unique-key-only conflict (where the replaced row's PK differs from the incoming PK) is not yet covered for the irregular index; tracked as a follow-up.

### Tests

- New BVT `test/distributed/cases/vector/vector_ivf_replace.sql` and `test/distributed/cases/fulltext/fulltext_replace.sql` covering new-row REPLACE, existing-row REPLACE (stale value cleared) and multi-row REPLACE.
- New unit tests `TestCollectSyncIvfMultiIndexes` and `TestCollectSyncFulltextIndexes`.
- `go test ./pkg/sql/plan` passes; existing `vector_ivf_membership` / `fulltext` / `fulltext_membership` BVTs pass with no regression.

### DIFF
--- a/pkg/sql/plan/bind_replace.go
+++ b/pkg/sql/plan/bind_replace.go
@@ -39,12 +39,258 @@ func (builder *QueryBuilder) bindReplace(stmt *tree.Replace, bindCtx *BindContex
 		return 0, err
 	}
 
+	// Collect the irregular (vector) indexes that must be maintained synchronously
+	// BEFORE initInsertReplaceStmt strips them from tableDef.Indexes. The modern
+	// REPLACE path otherwise drops these indexes silently, leaving newly inserted /
+	// replaced rows unsearchable until the next cron reindex (issue #25000).
+	syncIvfIndexes, err := collectSyncIvfMultiIndexes(dmlCtx.tableDefs[0])
+	if err != nil {
+		return 0, err
+	}
+	syncFulltextIndexes, err := collectSyncFulltextIndexes(dmlCtx.tableDefs[0])
+	if err != nil {
+		return 0, err
+	}
+
 	lastNodeID, colName2Idx, skipUniqueIdx, err := builder.initInsertReplaceStmt(bindCtx, stmt.Rows, stmt.Columns, dmlCtx.objRefs[0], dmlCtx.tableDefs[0], true)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


ck89119:
Superseded by #24973 (issue #24939), which fixes #25000 as part of the modern-DML index-maintenance rework: REPLACE now deletes the conflicting old row's IVF/fulltext entries and inserts the new row's entries on the modern path. Closing in favor of that PR.

PR #25019, opened 2026-06-17
fix: add mysql user-level lock functions

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24728

  - Add MySQL-compatible `GET_LOCK`, `RELEASE_LOCK`, and `IS_FREE_LOCK` built-in functions.
  - Register function IDs, name mappings, and overloads for SQL planning.
  - Implement process-local named user-level lock behavior keyed by connection ID.
  - Add tests for acquire, release, free-state checks, non-owner release, missing locks, and NULL inputs.
  - Add function lookup coverage for the new built-ins.


XuPeng-SH (changes requested):
I re-checked the latest head and still see one high-severity correctness issue around CN migration.

The new advisory-lock ownership and refcount state is tied to the backend session identity (`SessionId`) and a process-local map. That means a live proxy migration can silently drop a client’s user-level lock even though the client connection itself stays open.

The pieces line up like this in the current head:
- ownership is keyed by `userLevelLockOwner(proc)`, which prefers `SessionId`
- refcounts live only in the process-local `userLevelLocks` map
- `Session.Close()` now calls `ReleaseUserLevelLocks(ses.proc)`
- proxy migration restores database / prepared statements on the new backend session, but does not migrate advisory-lock ownership/refcounts
- the old backend routine/session is then closed

So during a normal live CN transfer, the old backend session will release the lock on teardown, while the new backend session has no migrated ownership/refcount state. From the client’s perspective the same connection remains alive, but its `GET_LOCK()` can be lost mid-session.

**Concrete suggestions**

1. Make user-level lock ownership stable across CN migration — key it to a connection-scoped identity that survives transfer, not the ephemeral backend `SessionId`.
2. Migrate/refill the held-lock refcounts into the new backend session during connection migration, instead of releasing them on old-session teardown.
3. Add one focused migration regression test: session A acquires a user-level lock, the proxy migrates the live connection to another CN, and the same client session still owns/releases the lock correctly afterward.
4. If full migration support is not feasible in this PR, I would rather see the feature explicitly gated/unsupported under migration than silently dropping locks on live sessions.

Non-blocking but worthwhile remaining coverage after that:
- SQL-level positive-timeout contention (`GET_LOCK('x', 0.1)`),
- hard CN failure / orphan-lock recovery,
- and cross-tenant same-name isolation.

heni02 (approved):

XuPeng-SH (changes requested):
I re-checked the current head and still see three substantive correctness issues.

1. **User-level lock names were made case-insensitive.**
   `normalizeUserLevelLockName()` lowercases every lock name before `GET_LOCK` / `RELEASE_LOCK` / `IS_FREE_LOCK` / `IS_USED_LOCK`. MySQL named locks are case-sensitive, so `GET_LOCK('Foo', ...)` and `GET_LOCK('foo', ...)` must be distinct. This patch collapses those namespaces, so unrelated callers can now block each other or release the wrong lock.

   The current tests and BVT also codify that behavior, which makes it look intentional, but it is not MySQL-compatible.

   **Suggestion:** preserve the original lock-name bytes/casing for identity, and update the tests/BVT to verify case-sensitive behavior instead.

2. **Session disconnect still does not release user-level locks.**
   The PR adds `ReleaseUserLevelLocks`, but in the current head there is still no production teardown path calling it. A client that disconnects after `GET_LOCK()` without explicitly calling `RELEASE_LOCK()` / `RELEASE_ALL_LOCKS()` leaves the advisory lock held indefinitely, along with stale local refcount state.

   **Suggestion:** wire `ReleaseUserLevelLocks` into real session / connection cleanup and add a disconnect/teardown regression test.

3. **`GetLockHolder` can self-deadlock on stale remote bind handling.**
   `service.GetLockHolder()` now holds `bindChangeMu.RLock()` across `l.getLockHolder(...)`. For a remote table, `doGetLockHolder()` calls `maybeHandleBindChanged()`, and the production `bindChangedHandler` is `s.handleBindChanged`, which acquires `bindChangeMu.Lock()`. That is an in-goroutine RWMutex upgrade when a stale bind returns `NewBind`, and it can deadlock the lookup.

   The current tests avoid this by injecting custom callbacks instead of the real `s.handleBindChanged`, so they do not catch the production lock ordering.

   **Suggestion:** do not hold `bindChangeMu` across the remote holder lookup, or defer bind-change handling until after the read lock is released. If you need the stale-local-table fix from the newer branch, surface `ErrLockTableBindChanged` from the local-table path and retry, rather than holding `RLock()` across a path that can synchronously invoke the write-lock handler.

gouhongshen (changes requested):
Blocking issues found in the user-level lock implementation.

1. Migration currently copies only frontend-local refcount/name state, not the actual lockservice ownership. `UserLevelLocksForMigration` / `RestoreUserLevelLocksFromMigration` move `counts/byOwner`, and the old session then sets `userLevelLocksMigrated` so `Session.Close` discards local state instead of unlocking. After migration, the new CN calls `Unlock` with the same txnID, but lockservice `Unlock` first deletes from the local `activeTxnHolder`; if that active txn was created on the old CN, the new CN sees nil and returns success without releasing the real lock. `releaseUserLevelLock` then untracks locally and returns success to SQL. Net effect: `RELEASE_LOCK()` can return 1 while the underlying lock remains held, and contenders keep timing out.

The current UT hides this because `userLevelLockTestService.Unlock` deletes from a shared map by txnID directly; it does not model lockservice's CN-local `activeTxnHolder` behavior.

Relevant code:
- `pkg/sql/plan/function/func_unary.go`: `UserLevelLocksForMigration`, `RestoreUserLevelLocksFromMigration`, `releaseUserLevelLock`
- `pkg/frontend/session.go`: migrated sessions call `DiscardMigratedUserLevelLocks` instead of `ReleaseUserLevelLocks`
- `pkg/lockservice/service.go`: `Unlock` returns nil when `deleteActiveTxn(txnID)` is nil

This needs a real lockservice-level owner transfer / active-txn transfer, or connection migration must be rejected while user-level locks are held. Copying refcounts is not sufficient.

2. The migration protocol is not crash-safe. Proxy first calls `setMigrateConnFromLockRelease(prevAddr, false)`, then migrates to the target, and only best-effort re-enables release if `migrateConnTo` returns an error. If proxy/CN crashes, times out, or loses connectivity between those steps, the source session may close with `userLevelLocksMigrated=true` and discard local state without releasing the real locks, while the target may not have restored anything. This leaves leaked advisory locks. This should be made atomic/lease-backed at the lockservice level, or the feature should fail closed by not migrating connections with held user locks.

3. Rolling-upgrade compatibility is incomplete for old txnID format. The code keeps `userLevelLockTxnIDOld` only for unlock, which implies old-format holders can exist. But `IS_USED_LOCK()` parses the holder txnID with `userLevelLockConnectionIDFromTxnID`; for the old three-part format it returns `(0,false)`, so `IS_USED_LOCK()` returns NULL even while the lock is held. MySQL semantics are that `IS_USED_LOCK(str)` returns the holding connection id when locked, and NULL only when not locked. See MySQL locking functions docs: https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html

ouyuanning (approved):

aptend (approved):
I reviewed the latest head independently, including the previous user-level lock concerns around ownership scope, case sensitivity, session cleanup, and connection migration behavior. The current implementation binds lock ownership to account + connection ID, preserves lock-name case, releases locks on session close, and rejects connection migration while user-level locks are held, which addresses the blocking migration-safety concerns by failing closed instead of attempting cross-CN lock transfer.\n\nI also ran targeted race tests for the changed areas:\n\n- go test -race ./pkg/sql/plan/function -run 'TestUserLevelLock(...)'\n- go test -race ./pkg/frontend -run 'Test.*(Migrate|UserLevel|RoutineManagerMigration|SessionClose)'\n- go test -race ./pkg/proxy -run 'Test.*Migrate|TestQueryService.*Migrate'\n\nAll passed locally. I do not see any remaining merge-blocking issue from this review. The remaining note is that some migration-transfer plumbing is now effectively unused because the production path rejects held user-level locks during migration, but I do not consider that blocking.

gouhongshen (changes requested):
Blocking on a current-head inconsistency in the user-level lock semantics.

`pkg/sql/plan/function/func_unary.go` now normalizes every lock name with `strings.ToLower(name)`, so `GET_LOCK` / `RELEASE_LOCK` / `IS_FREE_LOCK` / `IS_USED_LOCK` are case-insensitive. But the BVT expectations added in this PR still assert case-sensitive behavior:

- after `GET_LOCK('prisma_migrate_lock', 0)`, `IS_FREE_LOCK('PRISMA_MIGRATE_LOCK')` is expected to return `1`;
- `RELEASE_LOCK('PrIsMa_MiGrAtE_LoCk')` is expected to return `NULL`;
- then `RELEASE_LOCK('prisma_migrate_lock')` is expected to return `1`.

With the current implementation those names all normalize to `prisma_migrate_lock`, so the actual sequence should be `0`, `1`, then `NULL`. The Go unit test was changed toward case-insensitive behavior, while the distributed test/result files still encode case-sensitive behavior.

Please choose the intended semantics and make the implementation, Go UT, and BVT expected results agree. If the target is MySQL compatibility, the PR should also make the exact MySQL lock-name comparison rule explicit rather than leaving code comments/tests pointing in opposite directions.

gouhongshen (approved):
Approved based on the latest head. The previous blocker around lock-name case semantics is now consistent between implementation, Go UT, and BVT expectations. CI still needs to finish/pass independently.

aunjgr (approved):

XuPeng-SH (changes requested):
Requesting changes for two ownership gaps that are not covered by the current happy-path tests:

1. The user-lock refcounts are process-local, but these functions are not pinned to the frontend/coordinator CN. `compileProjection` installs a projection on every scan scope (and can push it into `TableScan`), while `volatile` / `realTimeRelated` only control folding. Therefore a statement such as `SELECT GET_LOCK(CONCAT("job-", id), 0) FROM t` can acquire and track locks on remote execution CNs. A later scalar `RELEASE_LOCK`, `RELEASE_ALL_LOCKS`, or `Session.Close` runs on the frontend CN and enumerates only that CN-local `userLevelLocks` map, so the remote synthetic transactions cannot be found or released. Please either force all user-lock side effects onto the owning frontend CN, or materialize and route every remote acquisition back into session-owned state, and add a multi-CN regression covering acquire, explicit release, and disconnect cleanup.

2. `userLevelLockOwner` now derives ownership from mutable `SessionInfo.ConnectionID`. `SET CONNECTION ID TO` is currently executable without a privilege gate, and the next statement refreshes `SessionInfo` from the protocol. A focused regression reproduces: acquire under ID A, change to ID B, then `RELEASE_LOCK` returns 0 and a contender still cannot acquire; `Session.Close` also sees only owner B and cannot clean owner A. Please keep an immutable per-session owner/acquisition txn identity, or reject connection-ID mutation while locks are held, and cover this sequence.

The targeted function/frontend/proxy tests and targeted race tests pass, so these are lifecycle/placement gaps rather than existing test failures.

XuPeng-SH (changes requested):
Requesting changes after re-checking the current head and a clean virtual merge onto latest main (826a3e3d29). The latest commit fixes the mutable ConnectionID ownership gap and pins user-lock calls in ProjectList to a coordinator scope, but two lifecycle/placement gaps remain.\n\n1. [P1] User-lock side effects still execute on remote CNs outside ProjectList\n\nThe coordinator guard is only in compileProjection and only inspects node.ProjectList. Static table-scan predicates are copied directly into every TableScan as FilterExprs before that guard, and compileRestrict similarly installs FilterList on every scope.\n\nFor example:\n\n    SELECT id FROM t\n    WHERE GET_LOCK(CONCAT('job-', id), 0) = 1;\n\nOn a multi-CN plan, GET_LOCK can run inside a remote table scan. The underlying lock is then owned by the remote CN lockservice transaction and its name/refcount is recorded in that CN-local userLevelLocks map. A later scalar RELEASE_LOCK, RELEASE_ALL_LOCKS, Session.Close, or the migration guard runs on the frontend CN, cannot see that state, and therefore cannot release or reject migration for those locks.\n\nThis is the same ownership invariant as the previous projection finding, but the fix currently covers one expression location rather than the semantic capability. Please enforce coordinator-only execution for user-level stateful functions across every expression site that can contain them, including scan/restrict filters, join predicates, HAVING, ordering/group/window expressions, and future operator paths. A plan-level side-effect placement rule would be safer than adding per-operator exceptions. Please add a multi-CN regression with GET_LOCK in WHERE, explicit release, disconnect cleanup, and migration rejection.\n\n2. [P1] Session.Close can block indefinitely while releasing a remote user lock\n\nSession.Close synchronously calls ReleaseUserLevelLocks. releaseAllUserLevelLocks uses context.Background, LockService.Unlock discards its caller context and replaces it with context.Background, and remoteLockTable.unlockWithContext retries forever while a remote binding remains valid but unreachable. The per-RPC timeout only bounds each attempt; the outer retry has no deadline.\n\nConsequently, a network partition or persistently unavailable lock owner can prevent Session.Close from ever reaching the rest of session/routine cleanup. Passing a timeout only at the frontend call site is insufficient because Unlock currently ignores it. Please provide a lifecycle-safe cleanup contract: bounded/cancelable close-side waiting plus a reliable fenced/orphan-recovery fallback, and cover permanent remote failure rather than only successful/mock unlocks.\n\nThe existing CI is green, but neither unhappy path is represented in the current test matrix.

XuPeng-SH (changes requested):
Requesting changes after reviewing the latest head 41e69be7 and compiling a clean virtual merge with current origin/main (826a3e3d29).\n\n1. [P0] restoreActiveTxn is incompatible with the current sharded activeTxnHolder\n\nThe PR adds restoreActiveTxn in pkg/lockservice/service.go, but it still writes h.mu.activeTxns and h.mu.activeTxnServices. Current main has already migrated mapBasedTxnHolder to 16 sharded activeTxns entries plus activeTxnCount; those fields no longer exist under h.mu.\n\nA virtual merge of this PR onto current main fails to compile:\n\n    service.go:1323:15: h.mu.activeTxns undefined\n    service.go:1326:7: h.mu.activeTxns undefined\n    service.go:1327:7: h.mu.activeTxnServices undefined\n\nPlease rebase onto latest main and implement restore using the same shard and activeTxnCount invariants as getActiveTxn/deleteActiveTxn. Also add a compile/test check against the current main merge base before updating the PR.\n\n2. [P1] Ordinary TABLE_SCAN filters are now evaluated twice\n\nThe TABLE_SCAN path still copies every non-user FilterList into TableScan.FilterExprs, but then unconditionally calls compileRestrict(node, ss). compileRestrict wraps the same FilterList in a filter operator. Therefore a normal query such as:\n\n    SELECT * FROM t WHERE a = 1;\n\nevaluates the predicate once in table_scan.evalFilter and again in filter.Filter.Call. This is a direct hot-path regression. For nondeterministic predicates such as RAND(), it can also change results because the expression is evaluated twice.\n\nPlease keep the existing inline table-scan path for ordinary filters and only build a separate restrict operator when the filter contains a user-level lock side effect (or otherwise prove the filter is not already embedded). Add a test that asserts the operator chain and a benchmark or allocation/per-row regression check.\n\n3. [P1] Session-close timeout still leaks process-local lock bookkeeping\n\nReleaseUserLevelLocksOnSessionClose bounds the unlock to five seconds and ignores the returned error. On timeout it intentionally leaves userLevelLocks counts/byOwner intact, but Session.Close then frees the process and there is no later owner that can untrack those entries. Repeating disconnects while remote unlock is unavailable leaves one global map entry per lock/session indefinitely, even if lockservice orphan recovery eventually releases the remote lock.\n\nPlease make the timeout path explicitly transfer cleanup ownership (for example, clear/detach the local bookkeeping after registering a durable lockservice recovery record) and test repeated close timeouts for bounded global state.

XuPeng-SH (changes requested):
Re-reviewing current head a1a2dc4bdc against current main found one remaining P1 lifecycle blocker. The sharded active-txn restore and ordinary TableScan filter fixes are correct, but the session-close timeout path still abandons the lock without transferring cleanup ownership.

`releaseUserLevelLocksOnSessionCloseWithTimeout` deletes the frontend counts/byOwner/ownerSessions state via `detachUserLevelLocksForOwner` after the bounded unlock times out. At the same time, `service.unlockWithContext` restores the source `activeTxn` on that failure. Source transactions are created by `service.Lock` with `remoteService == ""`, while the timeout/orphan scanner only reclaims transactions associated with a timed-out remote service. No retry queue, async cleanup worker, or durable detached-cleanup marker is registered before the frontend state is discarded. Because the CN itself remains healthy, a remote lock owner also continues to treat the source service as live.

This is reproducible on the virtual merge with current main: acquire a user lock, force Unlock to block until the close deadline, let close detach the maps, restore Unlock availability, then have another session call `GET_LOCK(name, 0)`. The contender returns 0 rather than 1. The new test currently only asserts that the three frontend maps are empty, so it misses the externally visible terminal state.

Please make the timeout branch perform a real ownership transfer before forgetting the session state: for example, enqueue a bounded/deduplicated asynchronous unlock record with retry/backoff, or add a lockservice-level detached cleanup state whose owner survives the session and eventually releases/fences the synthetic txn. Simply deleting the active txn is unsafe because the remote owner may still hold the lock. Add terminal-state tests for local and remote lock-table ownership, transient outage then recovery, permanent outage with bounded retained state, and repeated close timeouts.

Fresh validation on PR + main (9bc6f2a647): the new sharded restore test, both TableScan filter tests, and the existing close-detach test pass; the missing recovery-terminal regression fails with expected=1, actual=0.

gouhongshen (changes requested):
## Codex automated review

The previous placement and eventual-cleanup blockers are substantially addressed, and current-head CI passes. Three blocking lifecycle issues remain in the new cleanup design.

### P1 - Fence detached cleanup from reused connection generations (pkg/sql/plan/function/func_unary.go:7113)
The synthetic transaction ID contains only account, connection ID, and lock name. `ResetSession` replaces the Session/BaseProcess while retaining the cached backend connection ID, and session close may return after transferring a timed-out unlock to the detached worker. A subsequent same-tenant session reusing that connection therefore generates the identical txn ID. While the old active txn is restored, its `GET_LOCK` is treated as reentrant; the detached worker can then unlock that txn underneath the new session, leaving its frontend refcount claiming ownership of a released lock. Include an immutable session/generation component in lockservice ownership while retaining the connection ID separately for `IS_USED_LOCK`, or prevent connection reuse until cleanup is fenced. Add a reset/cache-reuse regression with forced close timeout and recovery.

### P1 - Bound permanent-outage cleanup resources (pkg/sql/plan/function/func_unary.go:7206)
Every distinct timed-out lock creates a dedicated goroutine and global map entry that retries and logs once per second forever. There is no service-shutdown cancellation, queue capacity, or aggregate worker bound; deduplication only helps identical `(service, owner, name, connID)` keys. During a remote-owner outage, clients can acquire many distinct names and disconnect, producing unbounded retained lockservice pointers, goroutines, RPC attempts, and log volume until CN OOM. The permanent-outage test merely proves one entry survives for 150 ms. Use a bounded service-owned cleanup queue/worker set with explicit shutdown and overload behavior, and test many distinct locks plus repeated service shutdown/restart.

### P1 - Make probe unlocks honor statement cancellation (pkg/sql/plan/function/func_unary.go:7536)
The free-lock probes in both `RELEASE_LOCK` and `IS_FREE_LOCK` call `LockService.Unlock` directly after acquiring the row. The concrete service intentionally replaces the supplied context with `context.Background`, and remote unlock retries indefinitely while an unreachable binding remains valid. If the remote owner disappears after probe acquisition, these SQL functions cannot be canceled and the routine hangs indefinitely. Use the new context-aware unlock path for probe transactions and transfer failed cleanup ownership safely; cover a remote outage occurring between successful probe acquisition and release.

gouhongshen (changes requested):
## Codex automated review

Three blocking lifecycle/cancellation bugs remain. Focused tests were blocked by missing C headers (xxhash.h, roaring.h, usearch.h).

### P1 - Clean up timed-out GET_LOCK transactions (pkg/sql/plan/function/func_unary.go:7679)
This treats every context deadline as a clean failed acquisition and returns without unlocking or retaining cleanup ownership. The remote lockservice deliberately records lock bookkeeping even when a lock RPC times out, because the request may have acquired the remote lock before its response was lost. The session then has no local refcount, so Session.Close and RELEASE_LOCK cannot clean it up; the named lock can remain held indefinitely. Schedule cleanup for this exact txn ID before returning 0.

### P1 - Fence detached cleanup from ResetSession reuse (pkg/sql/plan/function/func_unary.go:7217)
The detached cleanup txn ID contains only owner, lock name, and connection ID. Session.reset copies prev.uuid, while the protocol retains the same connection ID, so a replacement session can generate the identical txn ID after the old session timed out and transferred cleanup. LockService treats the new GET_LOCK as reentrant, and the detached worker can later unlock the new session's lock underneath its local refcount. The added generation test creates a fresh UUID and does not model Session.reset; add an immutable generation fence or prevent reuse until cleanup completes.

### P1 - Do not drop cleanup ownership when the queue is full (pkg/sql/plan/function/func_unary.go:7846)
When the bounded cleanup map/queue reaches capacity, enqueueDetachedUserLevelLockCleanups returns false. This branch then leaves the session's local state attached to a session that is about to close, so no worker owns the txn IDs and the underlying locks are never retried or released. The same ignored admission failure exists for probe cleanup. Overload must preserve cleanup ownership through backpressure or a durable/service-owned mechanism; silently refusing the handoff leaks locks.

gouhongshen (changes requested):
## Codex automated review

Two blocking resource-lifecycle bugs remain. Existing current-head CI is green, but focused local tests could not run because cgo/libmo.dylib is absent.

### P1 - Clean up failed FastFail lock transactions (pkg/sql/plan/function/func_unary.go:7694)
On every Lock call, the concrete lockservice registers txnID in activeTxnHolder before attempting the row lock. A FastFail conflict returns ErrLockConflict without removing that active transaction, but this branch immediately returns 0 and never unlocks or queues txnID. The same leak occurs for conflicting RELEASE_LOCK and IS_FREE_LOCK probes at lines 7730 and 7772. Because only successful acquisitions enter userLevelLocks, session cleanup cannot discover these transactions; source transactions have remoteService == "", so orphan scanning does not reclaim them either. Repeated contention therefore grows activeTxnHolder indefinitely and can prevent lockservice drain/rolling restart. Explicitly unlock every unsuccessful attempt, transferring cleanup ownership when bounded cleanup fails, and cover activeTxnHolder state with the real lockservice rather than the map-only mock.

### P1 - Do not abandon locks when detached cleanup admission fails (pkg/sql/plan/function/func_unary.go:7861)
The timeout path detaches session state only when enqueueDetachedUserLevelLockCleanups returns true. Admission returns false once 1024 entries would be exceeded; this is reachable either during an outage or by one session holding more than 1024 distinct locks. Session.Close then destroys the only frontend process without retrying, while UnlockWithContext has restored the timed-out active transaction and all remaining locks/refcounts stay in global maps under an unreachable session generation. Those named locks remain held indefinitely. The bounded queue is useful, but overload must apply backpressure or transfer ownership to another durable/service-owned mechanism; it cannot simply leave cleanup attached to a session that is closing. Add a close-timeout test that saturates admission and verifies eventual acquisition by a contender plus bounded retained state.

gouhongshen (changes requested):
## Codex automated review

Three blocking cleanup/lifecycle issues remain. Current-head CI is green; focused local tests could not start because cgo/libmo.dylib is absent.

### P1 - Preserve cleanup work added during an active attempt (pkg/sql/plan/function/func_unary.go:7405)
The worker snapshots entry.txnIDs at line 7399, then releases the mutex. A concurrent enqueue for the same key can append new transaction IDs and queue the key again because entry.queued was cleared. If the original snapshot succeeds, this branch deletes the entry without checking whether its transaction-ID set changed; the queued retry then finds no entry. This is especially reachable for the shared overflow key, where unrelated cleanup records are continually merged. The newly appended transactions lose cleanup ownership and can retain locks indefinitely. Version entries or remove only the successfully processed IDs, and prevent concurrent attempts for one entry.

### P1 - Clean up every failed synthetic Lock transaction (pkg/sql/plan/function/func_unary.go:7744)
GET_LOCK cleans up only deadline and conflict failures; other Lock errors return directly. RELEASE_LOCK and IS_FREE_LOCK have the same gap at lines 7796 and 7852. The concrete service registers the synthetic txn in activeTxnHolder before several possible errors, and remoteLockTable records lock bookkeeping after any failed RPC because the remote request may have acquired the lock. Errors such as bind changes, backend failures, or deadlock therefore leave an untracked active transaction—and potentially a remote lock—that Session.Close cannot discover. Attempt/transfer cleanup for every failed Lock call while preserving the original SQL error.

### P1 - The overflow entry does not actually bound cleanup resources (pkg/sql/plan/function/func_unary.go:7311)
Once 1024 entries exist, every additional distinct cleanup is collapsed into one overflow entry, but all unique txnIDs are appended to that entry without a limit. During a permanent outage, retained memory and each retry batch therefore grow without bound. mergeDetachedUserLevelLockTxnIDs also rebuilds a seen map over the full slice on every addition, making sustained overload increasingly expensive. The entry-count test misses this payload growth. Apply backpressure or transfer cleanup to durable/service-owned storage with a genuinely bounded in-memory working set.

XuPeng-SH (changes requested):
Blocking cleanup-ownership loss remains in `enqueueDetachedUserLevelLockEntryLocked` (pkg/sql/plan/function/func_unary.go).

The helper ignores a `false` return from `enqueueDetachedUserLevelLockCleanupLocked` when the bounded channel is full, then returns `true` anyway. `ReleaseUserLevelLocksOnSessionClose` treats that as a successful ownership handoff and detaches the session-local lock state. The entry is left in the map with `queued == false`; if the already queued cleanup attempts all succeed, no path runs the global rescan (`queueDetachedUserLevelLockCleanupLocked` is only called from the failure/retry branch), so this entry never reaches a worker and its lock/active txn can leak indefinitely. The same false-success can affect failed GET_LOCK/probe cleanup admission.

I reproduced this on head `d1e7fd7` with a focused test that fills `detachedUserLevelLockCleanups.queue`, adds a non-running entry, and asserts admission is rejected; the current code returns true. Please either propagate queue rejection so the caller retains ownership/backpressures, or provide a guaranteed scheduler for every map-resident unqueued entry, and add a regression test covering queue saturation followed by successful draining.

gouhongshen (changes requested):
## Codex automated review

A distinct blocking cleanup-admission bug remains. Focused tests could not run because `go` is unavailable in the environment.

### P1 - Split oversized session-close cleanup batches (pkg/sql/plan/function/func_unary.go:7395)
A session holding 1,025 distinct locks produces 2,050 transaction IDs because cleanup includes both current and legacy IDs. For a new cleanup key, this exceeds `userLevelLockDetachedCleanupMaxTxnIDsPerEntry` (2,048), so admission returns false without trying overflow shards. `ReleaseUserLevelLocksOnSessionClose` consequently does not detach the state, but the Session is still destroyed; after an unlock timeout, no worker owns these synthetic transactions and their locks can remain held indefinitely. This is reachable through vectorized `GET_LOCK` over rows. Partition the session batch across bounded entries with atomic ownership transfer or backpressure, and add a regression exceeding 1,024 distinct locks.

XuPeng-SH (changes requested):
Blocking cleanup-ownership loss remains on exact head `d713adac2bcb8c8fecbeae7d4051dc0a24b3226a`.

### P1 - Make split session-close cleanup admission atomic

`enqueueDetachedUserLevelLockCleanups` submits its chunks one at a time and returns `false` as soon as a later `enqueueDetachedUserLevelLockTxnCleanup` is rejected. Earlier chunks are not rolled back. The caller, `releaseUserLevelLocksOnSessionCloseWithTimeout`, detaches session-local state only when the overall return is `true`.

This creates a partial ownership transfer under the bounded-overflow condition the cleanup subsystem is designed to handle:

1. A session holds at least 1,025 distinct locks, producing a full 2,048-txn first chunk plus another chunk. The new unit test proves this input is supported.
2. During a remote outage, the cleanup map is at its entry bound and 15 overflow shards are full, leaving one shard with room for exactly the first chunk.
3. The first chunk is inserted and queued; the second chunk finds every bounded shard full and is rejected.
4. The helper returns `false`, so the caller does not detach any session state, but the session is then destroyed. The accepted chunk has detached ownership while the rejected chunk has no surviving cleanup owner; its synthetic transactions and named locks can remain indefinitely.

I reproduced the non-atomic mutation with a focused white-box test: the helper returned `false` while the detached cleanup map still gained exactly 2,048 txn IDs from the first chunk. The PR's new oversized/reschedule/concurrent-cleanup tests pass under `-race -count=20`, but none exercises failure after an earlier chunk succeeds.

Please reserve/admit all chunks atomically before reporting success and detaching the session, or otherwise guarantee ownership for every remaining chunk once any chunk has been transferred. Add a regression that forces chunk 1 admission and chunk 2 rejection, and asserts there is neither partial global mutation nor an ownerless remainder.

gouhongshen (changes requested):
## Codex automated review

Two blocking lifecycle/compatibility issues remain. Static checks passed; focused Go tests could not run because `go` is unavailable.

### P1 - Split cleanup admission is not atomic (pkg/sql/plan/function/func_unary.go:7456)
Chunks are submitted one at a time; if a later chunk is rejected, earlier chunks remain globally registered while the helper returns false. The session-close caller detaches local state only on overall success (8046-8051), so a 1,025-lock session can leave later locks ownerless after the session is destroyed. Reserve/commit all chunks atomically or transfer ownership for every chunk.

### P1 - Legacy lock owners no longer match cleanup IDs (pkg/sql/plan/function/func_unary.go:7219)
origin/main generated owners from SessionInfo.SessionId, but this head derives/pins owners as account:session:generation. The legacy three-field ID is built only with the new owner, so locks created by the prior version cannot be released after upgrade. IS_USED_LOCK also returns NULL because its legacy parser attempts to parse the UUID as a connection ID. Preserve the old owner encoding as an explicit compatibility candidate or provide a deliberate migration strategy.

pkg/sql/plan/function/func_unary.go
```diff
@@ -7376,6 +8149,29 @@ func ReleaseUserLevelLocks(proc *process.Process) {
...
 }
 
+func ReleaseUserLevelLocksOnSessionClose(proc *process.Process) {
+	releaseUserLevelLocksOnSessionCloseWithTimeout(proc, userLevelLockCloseTimeout)
+}
+
+func releaseUserLevelLocksOnSessionCloseWithTimeout(proc *process.Process, timeout time.Duration) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	if _, err := releaseAllUserLevelLocksWithContext(ctx, proc); err != nil && ctx.Err() != nil {
```
XuPeng-SH: **P1 — Session close still loses cleanup ownership on non-timeout errors and admission rejection.** This transfers work only when the close context has expired. A prompt UnlockWithContext error leaves ctx.Err() nil, so the function returns without detaching or queueing anything; Session.Close then destroys the session and no path revisits its tracked locks. I reproduced this by injecting an immediate unlock error: UserLevelLocksForMigration still contained the held lock after close. Separately, the new atomic-admission regression proves enqueueDetachedUserLevelLockCleanups can return false after rollback; line 8163 also silently keeps state attached to the closing session in that case. Transfer/backpressure must cover every release error and admission failure, with a live retry owner guaranteed before Close returns.

pkg/sql/plan/function/func_unary.go
```diff
@@ -7343,25 +8105,36 @@ func isUserLevelLockUsed(name string, proc *process.Process) (uint64, bool, erro
 	if !ok {
 		return 0, true, nil
 	}
-	holderConnID, ok := userLevelLockConnectionIDFromTxnID(holder.TxnID)
+	holderConnID, ok := userLevelLockConnectionIDFromTxnIDWithFallback(holder.TxnID, userLevelLockConnectionID(proc), true)
```
XuPeng-SH: **P1 — IS_USED_LOCK returns the observer's connection ID for legacy UUID owners.** A pre-generation 3-field txn ID contains account:session-UUID, not a numeric connection ID. Parsing therefore fails and this fallback substitutes userLevelLockConnectionID(proc), which belongs to the querying session, not the holder. Focused repro: holder connection 1001, observer connection 2002, legacy UUID txn ID; IS_USED_LOCK returned 2002. The old encoding cannot be reconstructed from the caller, so preserve holder connection metadata during upgrade/migration or return NULL rather than a wrong ID.

XuPeng-SH (changes requested):
Reviewed exact head 31af9207678d05556d2748f0109cfa7831b6bf41. The new touched-entry rollback fixes the previously reported partial-admission mutation, and the added owner candidates make legacy explicit release reachable. Two functional blockers remain (inline). Current-head CI is also red: SCA reports unused unlockUserLevelLockByName, and the focused frontend regression TestSessionCloseReleasesUserLevelLocksWhenNotMigrated fails because it still expects 2 cleanup txn IDs while current code emits 4. The focused user-lock lifecycle suite passes under -race; compile/proxy/lockservice focused tests pass.

gouhongshen (changes requested):
## Codex automated review

Four blocking lifecycle/compatibility defects remain. Current-head CI is green, but focused local race tests could not start because `go` is unavailable in the environment.

### P1 - Fence detached cleanup from same-session GET_LOCK retries (pkg/sql/plan/function/func_unary.go:7942)
After a failed Lock, the txn ID is queued for asynchronous cleanup and the call returns without tracking a refcount. A later GET_LOCK for the same session/name reuses this exact txn ID; after recovery it can succeed reentrantly and be tracked, while the stale worker subsequently unlocks it. Another session can then acquire the lock even though the first session still believes it owns it. Use a per-attempt generation or atomically retire/cancel pending cleanup before admitting a retry.

### P1 - Do not abandon cleanup ownership when bounded admission rejects (pkg/sql/plan/function/func_unary.go:7255)
If synchronous unlock fails and the bounded cleanup structure is full, this branch merely returns an error. Lockservice has restored the active transaction, while failed GET_LOCK/probe attempts have no entry in userLevelLocks, so disconnect cleanup cannot find them. The session-close caller similarly only logs an admission failure before the session is destroyed. Under a sustained remote outage, further failures can therefore leave active transactions and named locks without any live cleanup owner.

### P1 - Rollback can permanently strand an in-flight cleanup entry (pkg/sql/plan/function/func_unary.go:7422)
Rollback replaces each touched map entry with a cloned pointer. If an entry was already running while admission later failed, the clone retains running=true. The original worker then observes the pointer mismatch at line 7635 and returns, while queueDetachedUserLevelLockCleanupLocked permanently skips the clone because it still appears running. Its prior cleanup transactions—and possibly their locks—are never retried. Rollback must preserve the live entry identity or use versioned/in-place reservation rollback.

### P1 - Fail closed when migrating from pre-protocol CNs (pkg/proxy/conn_migration.go:146)
The base branch already supports user locks, but its MigrateConnFromResponse has no UserLevelLocks field or capability marker. During rolling upgrade, a new proxy decoding an old CN response always sees an empty slice and proceeds with migration even when that old session holds locks. The target receives no refcounts/ownership, and the old Session.Close has no release hook, so the client cannot release its own lock and later loses it silently. Require an explicit source capability/attestation or reject migration from older CNs.

### P2 - Keep ordinary multi-group transaction unlocks parallel (pkg/lockservice/service.go:510)
Ordinary Unlock now shares the retryable path and calls closeWithoutFreeWithContext. That passes release=false, so txn.go's parallelUnlock condition is always false, whereas the prior closeWithContext path unlocked transactions spanning more than two lock groups concurrently. Remote unlock latency is now serialized across those groups for every normal commit/rollback; retain the old path for unbounded ordinary Unlock and use retryable retention only for bounded UnlockWithContext.

gouhongshen (changes requested):
## Codex automated review

The latest cleanup-handoff changes still introduce a blocking failure mode under sustained lockservice failure.

### P1 - A full cleanup backlog can block lock operations indefinitely (pkg/sql/plan/function/func_unary.go:7562)
After the detached-cleanup map and overflow shards are full, this sends synchronously to a 64-entry backlog. Callers pass context.Background(), so the send cannot be cancelled. The sole backlog worker retries admission forever while existing entries cannot drain during a persistent lockservice outage (line 7705). Once the backlog fills, subsequent failed GET_LOCK/probe cleanup blocks the SQL request indefinitely; session close can likewise block before freeing the process. The handoff needs bounded, cancellation-aware backpressure or another ownership-preserving mechanism that cannot hang request and teardown paths forever.

gouhongshen (changes requested):
## Codex automated review

Three P1 lifecycle/compatibility defects remain at the current head.

### P1 - Do not abandon failed-attempt cleanup when the bounded handoff times out (pkg/sql/plan/function/func_unary.go:7277)
The new 100ms handoff can return false when the entry/overflow maps and backlog are full. `cleanupFailedUserLevelLockTxn` then returns an error, but its txn was never added to `userLevelLocks`, so neither RELEASE_ALL_LOCKS nor Session.Close will retry it. The concrete lockservice creates the active txn before attempting the row lock and restores it after a bounded UnlockWithContext failure; a failed GET_LOCK/probe can therefore leave an active txn and possibly a remote lock with no cleanup owner indefinitely. The added test only proves prompt failure of handoff, not eventual cleanup after that failure.

### P1 - Make split close-cleanup fallback admission atomic (pkg/sql/plan/function/func_unary.go:7658)
After atomic `enqueueDetachedUserLevelLockCleanups` rejects a large session-close batch, this fallback hands chunks off one at a time. With a saturated cleanup map and a backlog with room for only the first chunk, chunk 0 is accepted, a later chunk times out, and the function returns false. The caller consequently does not detach the session state, but Session.Close destroys the only process that could release the unsubmitted chunks. A session holding more than 2048 distinct locks during an outage can thus leave its remaining synthetic txns/locks permanently unretried. Reserve all fallback chunks before reporting success, or transfer ownership of every chunk once any is accepted.

### P1 - Include the actual pre-PR session-id owner in legacy unlock candidates (pkg/sql/plan/function/func_unary.go:7117)
origin/main generated user-lock owners from the bare `SessionInfo.SessionId.String()`, so its transaction IDs contain `mo-user-level-lock\0<uuid>\0<name>...`. The current fallback candidates include only `account:uuid` and `account:connectionID`, never that bare UUID. When exact tracked IDs are unavailable, RELEASE_LOCK/RELEASE_ALL_LOCKS therefore unlock only non-existent candidate IDs (which lockservice treats as success), untrack local state, and leave the base-version lock held. `TestReleaseLockPreGenerationLegacyTxnIDCompatible` tests `account:uuid`, not the owner format used by origin/main.

gouhongshen (changes requested):
## Codex automated review

A session-close cleanup failure can still leave advisory locks without a live cleanup owner.

### P1 - Do not destroy a session after detached-cleanup handoff fails (pkg/sql/plan/function/func_unary.go:8348)
The new five-second handoff timeout is still not a guaranteed ownership transfer. When the bounded cleanup entries/overflow shards cannot admit the work and the 64-request backlog stays full, `handoffDetachedUserLevelLockCleanups` returns false; this branch only logs and leaves the state attached to the closing session. `Session.Close` then continues and frees the process, while failed `UnlockWithContext` restores the active transaction (`pkg/lockservice/service.go:548`). No worker owns the retained per-session entries, so the underlying user lock can remain held indefinitely after the client disconnects. Require a durable/service-owned handoff or retain/backpressure the session until ownership is transferred, and add a saturation-plus-recovery test that verifies a contender can acquire the lock after close.

gouhongshen (changes requested):
## Codex automated review

Two blocking unhappy-path ownership/resource failures remain. The latest update prevents one ownerless-close case, but turns cleanup saturation into unbounded session teardown and does not protect failed lock attempts. Focused tests could not execute because the environment has no Go executable.

### P1 - Timed-out GET_LOCK can abandon its cleanup transaction (pkg/sql/plan/function/func_unary.go:7277)
GET_LOCK passes its lock-wait context into cleanupFailedUserLevelLockTxn. For a positive timeout that context is already expired when Lock returns, so the bounded unlock cannot run and backlog handoff immediately fails if direct admission is saturated. The function then returns an error without recording the random attempt txn ID in userLevelLocks or transferring it to another owner. The concrete UnlockWithContext path restores an active transaction after bounded cleanup failure (pkg/lockservice/service.go:548), so a timed-out remote acquisition can remain active—and may still own a lock—with no path for RELEASE_ALL_LOCKS or Session.Close to discover it. The existing full-backlog test only verifies prompt handoff failure, not eventual ownership or cleanup.

### P1 - Cleanup saturation now blocks Session.Close forever (pkg/sql/plan/function/func_unary.go:8355)
The latest commit calls handoffDetachedUserLevelLockCleanups with context.Background(). When the bounded entry/overflow maps are saturated and the backlog worker is permanently retrying during a lockservice outage, the backlog remains full and handoffDetachedUserLevelLockBacklogCleanups loops forever. Session.Close invokes this before closing the transaction handler, prepared statements, and other session resources (pkg/frontend/session.go:858), so every later disconnect retains its full session and cleanup goroutine indefinitely, defeating the cleanup subsystem's memory bounds. The new test proves blocking until a slot is manually freed but does not cover permanent saturation. Cleanup ownership needs a durable/service-owned transfer or another bounded teardown contract that cannot accumulate unbounded blocked sessions.

gouhongshen (changes requested):
## Codex automated review

Migration-loss and pre-protocol migration blockers are fixed by failing closed; coordinator placement/identity fixes remain effective. Case handling is now internally consistent and accepted as non-blocking. The latest change bounds Session.Close handoff and retains a basic failed GET_LOCK cleanup, but its new retained-cleanup path still has blocking ownership and resource failures.

### P1 - Retain successful probe cleanup when detached admission is saturated (pkg/sql/plan/function/func_unary.go:8024)
Author response: b041 retains failed GET_LOCK cleanup in pendingCleanups after bounded handoff fails. Why this remains blocking: RELEASE_LOCK and IS_FREE_LOCK can successfully acquire a probe, then fail its Unlock; unlockUserLevelLockProbe only tries the bounded handoff and returns here when it cannot enqueue. The probe is never tracked in userLevelLocks or pendingCleanups. UnlockWithContext restores its active transaction on failure (pkg/lockservice/service.go:547-548), so recovery can leave the probe lock held with no cleanup owner and future GET_LOCK calls blocked. Route this path through the retained ownership mechanism (or a service-owned record) and add a saturated-handoff/recovery regression for successful probes.

### P1 - Do not discard cleanup IDs added during a retained-worker attempt (pkg/sql/plan/function/func_unary.go:7825)
Author response: b041 coalesces failed-attempt transaction IDs by cleanup key and retries them from one retained worker. Why this remains blocking: the worker snapshots pendingCleanups outside the mutex (7798-7817); another failed attempt for the same owner/name/kind can merge a new transaction ID at 7726-7736 while the old snapshot is unlocking. On success, or after a later detached handoff, this unconditional delete removes the current map entry rather than only the snapshot IDs (also 7835). The newer transaction is then untracked, while a failed bounded unlock restores its active transaction, leaving a possible held lock with no retry owner. Use versioned/CAS removal or subtract only processed IDs, and test admission of a second ID while the first attempt is in flight.

### P1 - Keep retained cleanup fallback resource-bounded (pkg/sql/plan/function/func_unary.go:7725)
Author response: b041 replaces the previously blocking close-side handoff with asynchronous retained cleanup. Why this remains blocking: pendingCleanups and retainedCloseCleanups have no entry bound, admission control, or shutdown ownership. Once the 1024-entry detached map, overflow shards, and backlog are saturated during an outage, every distinct failed attempt or closing session is inserted into these unbounded maps (7736, 7747) and retried serially until recovery. A client can create arbitrary distinct lock names/sessions, bypassing all existing cleanup limits and eventually exhausting CN memory; after 2048 IDs for one key, the next failed cleanup is again ownerless. Use an enforced bounded, durable/service-owned queue or explicit backpressure, and test sustained saturation with many unique attempts and disconnects.

gouhongshen (changes requested):
## Codex automated review

Migration loss and old-CN migration are fixed by failing closed; coordinator placement, immutable ownership, successful-probe retention, and concurrent retained-ID removal are fixed. Case handling is accepted as non-blocking. The new retention cap addresses unbounded maps but introduces two still-blocking saturation paths: ownerless probe locks and indefinitely blocked session teardown. Focused tests could not start because Go is unavailable locally.

### P1 - Do not abandon a probe lock when retained cleanup is full (pkg/sql/plan/function/func_unary.go:8049)
Author response: f9a5 routes failed successful-probe cleanup through pendingCleanups and caps retained cleanup at 1,024 entries. Why this remains blocking: once detached entries, overflow shards, backlog, and the new retained map are saturated, both handoffDetachedUserLevelLockTxnCleanup and retainDetachedUserLevelLockTxnCleanup return false. A successful RELEASE_LOCK/IS_FREE_LOCK probe has definitely acquired the named lock; if UnlockWithContext fails, lockservice restores its active transaction at service.go:547-548. This branch then returns an error without recording the probe transaction anywhere, so recovery has no owner that can retry its unlock and the lock can remain held indefinitely. The new admission test explicitly verifies rejection at the cap but does not verify ownership of the rejected transaction. Close this by guaranteeing durable/service-owned cleanup ownership or applying bounded backpressure before acquiring another probe; add a saturation-and-recovery test proving a contender can acquire afterward.

### P1 - Do not retry session teardown forever when retained cleanup is full (pkg/sql/plan/function/func_unary.go:8565)
Author response: f9a5 bounds pendingCleanups plus retainedCloseCleanups to prevent unbounded cleanup memory. Why this remains blocking: when that shared cap is full, retainUserLevelLockCloseCleanup rejects a new closing session. If the detached maps/backlog are also saturated during a persistent lockservice outage, this loop repeatedly spends up to five seconds on release and handoff, then sleeps and retries forever. Session.Close calls it synchronously before closing the frontend session, transaction handler, and prepared statements (session.go:858-883), so every subsequent disconnect can retain its full session indefinitely. The bound therefore recreates the previously reported teardown availability failure. Close this with a bounded close-side contract backed by guaranteed cleanup ownership/admission, and test permanent saturation without manually freeing a slot.

gouhongshen (changes requested):
## Codex automated review

Follow-up disposition: migration safety, coordinator placement, immutable ownership, ordinary-filter duplication, sharded txn restore, partial-admission, retry-generation, probe-retention, and retained-ID concurrency issues are fixed. Case semantics is accepted as non-blocking. The prior UUID-holder IS_USED_LOCK compatibility issue is unchanged and not repeated per the follow-up rule. One current P1 remains: close-cleanup retention bypasses its bound and can grow without limit during an outage.

### P1 - Retained close cleanup bypasses the cap and grows without bound (pkg/sql/plan/function/func_unary.go:7819)
Author response: f9a5 states that pendingCleanups plus retainedCloseCleanups are bounded to prevent unbounded cleanup memory. Why this remains blocking: the admission check rejects only when byOwner[owner] == nil. The caller at line 8602 has just obtained non-empty states from that same byOwner map, so every real closing session that still holds a lock bypasses the 1024-entry cap. Once detached cleanup admission is saturated and the lockservice remains unavailable, each distinct session is added to retainedCloseCleanups while its counts/byOwner state remains attached because the retry worker cannot make progress. Repeated disconnects therefore accumulate one retained owner and its lock state per session indefinitely, causing process-memory growth and eventual OOM. Enforce the cap for closing owners as well, while ensuring rejected cleanup has a live bounded backpressure or durable ownership path before the session is destroyed.

gouhongshen (approved):
## Codex automated review

The latest blocker is fixed: active lock owners now consume bounded cleanup capacity before acquisition and reuse that slot during retained close cleanup. Earlier migration, coordinator-placement, immutable-identity, filter-duplication, sharded-restore, admission, retry, and probe-cleanup blockers remain fixed. Case semantics and legacy UUID compatibility are non-blocking. No new findings. Focused tests could not run because `go` is unavailable; static inspection and `git diff --check` found no issue.

XuPeng-SH (approved):
Deep review completed on df55109. User-lock cleanup remains bounded and retains retry ownership across lockservice failures; migration fails closed for held locks and old-CN protocol; side effects run on the coordinator. Changes after the prior approval only merge main and stabilize cleanup tests. Focused lockservice, function, compile, frontend, and proxy tests passed, along with go vet and git diff --check. No blocking findings.

Issue #24830, opened 2026-06-03
[Bug] @@event_scheduler reports ON but CREATE EVENT is not supported


dengn:
## Summary
`@@event_scheduler` reports `ON`, but `CREATE EVENT` is not supported (parser rejects it). The variable being ON is misleading.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test case
```sql
SHOW VARIABLES LIKE 'event_scheduler';   -- ON
CREATE EVENT ev1 ON SCHEDULE EVERY 1 MINUTE DO INSERT INTO t(ts) VALUES(now());
```

## Actual result
```
ERROR 1064 (HY000): SQL parser error ... syntax error near ' EVENT ev1 ON SCHEDULE ...'
```

## Expected
Either implement the event scheduler, or hide/disable `@@event_scheduler` so it does not advertise an unsupported feature.

## Version status
- v3.0.11 (Cloud): reproduces
- main / 4.0-dev (commit 0bfda60): reproduces

Ariznawlll:
Verified on latest main 2a008285e.

Current result:
```sql
show variables like 'event_scheduler';
select @@event_scheduler;
```
returns `DISABLED` on a clean/latest main instance, so the original default `ON` symptom is not reproduced by default.

```sql
create event ev1 on schedule every 1 minute do insert into t(ts) values(now());
-- ERROR 1064: SQL parser error ... near " event ev1 ..."
```

mo对该功能不做支持

PR #25021, opened 2026-06-17
fix: report event_scheduler as DISABLED since CREATE EVENT is unsupported

ck89119:
## What type of PR is this?

- [x] BUG

issue #24830

`@@event_scheduler` defaulted to `ON`, but MatrixOne does not support `CREATE EVENT` (the parser rejects it), so the variable advertised an unsupported feature and misled users.

This changes the default to `DISABLED`, which accurately reflects that the event scheduler is unavailable and cannot be enabled at runtime.

- `pkg/frontend/variables.go`: `event_scheduler` default `ON` -> `DISABLED`
- added unit test asserting the default is `DISABLED`
- added bvt case `system_variable/event_scheduler`

### DIFF
--- a/pkg/frontend/variables.go
+++ b/pkg/frontend/variables.go
@@ -1842,7 +1842,7 @@ var gSysVarsDefs = map[string]SystemVariable{
 		Dynamic:           true,
 		SetVarHintApplies: false,
 		Type:              InitSystemSystemEnumType("event_scheduler", "ON", "OFF", "DISABLED"),
-		Default:           "ON",
+		Default:           "DISABLED",
 	},
 	"explain_format": {
 		Name:              "explain_format",
--- a/pkg/frontend/variables_test.go
+++ b/pkg/frontend/variables_test.go
@@ -23,6 +23,18 @@ import (
 	"github.com/smartystreets/goconvey/convey"
 )
 
+func TestEventSchedulerDefaultDisabled(t *testing.T) {
+	convey.Convey("event_scheduler default should be DISABLED", t, func() {
+		sv, ok := gSysVarsDefs["event_scheduler"]
+		convey.So(ok, convey.ShouldBeTrue)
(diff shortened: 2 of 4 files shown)


XuPeng-SH (approved):
✅ **APPROVED - Excellent truthfulness fix**

**Core Change**: event_scheduler default: ON → DISABLED (1-line fix)

**Rationale**: MatrixOne does NOT support CREATE EVENT (verified: 0 parser occurrences). Default=ON was **misleading users** into believing the feature exists. Default=DISABLED accurately reflects reality.

**Analysis Summary**:
- ✅ **Correctness**: DISABLED is valid enum value, unit test validates Type.Convert() succeeds (prevents init crash)
- ✅ **Backward compat**: Explicit SET values preserved, only default changes for new sessions
- ✅ **Test coverage**: Unit test + BVT covering SHOW VARIABLES and SELECT @@ syntax
- ✅ **First principles**: Verified CREATE EVENT unsupported via parser grep
- ✅ **Edge cases**: All 3 enum values (ON/OFF/DISABLED) remain settable (no-ops)

**Unhappy Paths Validated**:
- ✅ Migration tools → check INFORMATION_SCHEMA.EVENTS not just variable
- ✅ Feature detection → correctly returns DISABLED (apps should disable event features)
- ✅ Upgrade scenario → persisted SET values unchanged
- ✅ User sets to ON → allowed but no effect (MatrixOne limitation, not variable issue)

**Key Strength**: Unit test validates Type.Convert("DISABLED") succeeds - **prevents session init crash** if enum was misconfigured.

**Risk Level**: VERY LOW (cosmetic variable for MySQL compat, no functional scheduler)

Full review: /tmp/pr_review_25021.md

This PR stops **lying to users** about capabilities. Approve and merge! 🎯

heni02 (approved):

Issue #25025, opened 2026-06-17
[Bug]: Distributed INSERT...SELECT deadlocks in Scope.MergeRun wg.Wait(), holds table lock and cannot be killed (only CN restart recovers)


ck89119:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

3.0

### Commit ID

5e4210029

### Other Environment Information

- Deployment: distributed (multiple CN, separate DN/LogService), running on K8s
- enable-leak-check=1, max-active-ages=20m, remote-lock-timeout=10m

### Actual Behavior

An autocommit `INSERT ... SELECT` that runs as a **distributed plan across multiple CN** got stuck for **~31 hours** and could not be recovered without restarting the CN. Symptoms observed during the whole period:

- The statement never finished and **could not be killed**: `KILL <conn_id>` / `KILL CONNECTION` returned OK, but the session stayed in `SHOW PROCESSLIST` forever and the query kept running.
- The transaction kept **holding a row lock on the target table**, so a later transaction blocked on it — `lockservice` logged `wait too long` for > 6h.
- The txn leak checker reported it every 20m as `found leak txn`, with `InRunSql / InCommit / InRollback` all false, which is misleading — the goroutine dump shows the execution pipeline is actually **deadlocked, not idle**.
- Only restarting the CN pod cleared the stuck pipeline and released the lock.

**Root cause (from the goroutine dump):**

```
# main statement goroutine — blocked 1887 min
goroutine [chan receive, 1887 minutes]:
  frontend.doComQuery -> executeStatusStmt
  -> compile.(*Compile).Run -> (*Compile).runOnce            // compile.go:547, chan receive

# worker — blocked 1886 min
goroutine [sync.WaitGroup.Wait, 1886 minutes]:
  compile.(*Compile).run -> (*Scope).MergeRun                // scope.go:378
  -> MergeRun.func2 -> sync.WaitGroup.Wait                   // scope.go:330

# 188 goroutines — blocked 1886 min
goroutine [chan receive, 1886 minutes]:
  compile.CnServerMessageHandler                             // remoterunServer.go:121
```

`Scope.MergeRun` blocks in `wg.Wait()` (scope.go:330) waiting for all `PreScopes` / `RemoteRun` / notify-message routines. **188 remote receivers (`CnServerMessageHandler`) are stuck on `chan receive`** waiting for data from the producing side that never arrives and never closes the channel, so the `WaitGroup` never reaches zero.

[...]

PR #25035, opened 2026-06-17
fix: bound pipeline cleanup waits on 3.0-dev (cherry-pick #24972)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #25025.

> **Note:** this PR was rebased onto a different, more robust fix. It previously
> carried a lighter `ctx.Done()`-based approach to making teardown waits
> cancelable. That approach has a **cancellation-ordering deadlock**: in
> `Scope.MergeRun`, when one sub-scope errors, the deferred `wg.Wait()` can only
> be unblocked by `proc.Ctx.Done()`, but `proc.Ctx` is cancelled by the upper
> layer only *after* `MergeRun` returns — which it can't, because it's blocked in
> the defer. A remote forwarding goroutine stuck on a bare `Ch2` send then never
> completes, the `MergeRun` result loop (no `ctx.Done()` escape) waits forever,
> and statements such as `drop account` hang ~300s in CI. Every `ctx`-based escape
> is useless in exactly the window where `proc.Ctx` is not yet done.

### What this PR now does

Backports #24972 ("fix: bound pipeline cleanup waits") from `main` to `3.0-dev`,
plus the `CnServerMessageHandler` `messageCtx` escape. This replaces the
cancellation-based approach with **timeout-bounded** teardown, which does not
depend on cancellation ordering:

- `PipelineSpool.CloseWithTimeout` / `ForceCleanup` (`cleanupOnce`) — bounded
  wait + deferred force cleanup instead of blocking forever on `csDoneSignal`
- connector / dispatch cleanup rewritten around `PipelineSignalSendTimeout`,
  terminal spool signals, and `cleanupSpool` for later force-cleanup
- `sendNotifyMessage.closeWithError` uses `SendPipelineSignalWithTimeout` +
  `WarnPipelineCleanupf` instead of a blocking `Ch2` send
- `WarnPipelineCleanupf` diagnostics + cleanup metrics
- `CnServerMessageHandler` also observes `messageCtx` (not on `main`/#24972)

This aligns `3.0-dev` with `main` (#24972 merged) and `4.0-dev` (#25223).

### 3.0-dev adaptations vs #24972

- `scope.go closeWithError`: kept 3.0-dev's cancel-error suppression
  (`ctx.Done() → err:nil`) since `suppressRemoteRunCancelError` does not exist on
  3.0-dev; layered the timeout-bounded send on top.
- `metrics.go` / `pipeline.go`: kept 3.0-dev's `pipelineStreamCounter` /
  `PipelineMessageSenderCounter` naming, added `PipelineCleanupEventCounter`.
- `dispatch_test.go`: dropped the two `AdoptCleanupState` tests (that method is a
  `main`-only feature, absent on 3.0-dev); kept the new Reset/CleanupDeferredSpool
  tests.
- `pipeline/types_test.go`: adapted `NewTopProcess` to 3.0-dev's 10-arg signature.

### Verification

- `go build` of all changed packages: OK.
- Unit tests pass: `pkg/container/pSpool`, `pkg/sql/colexec/connector`,
  `pkg/sql/colexec/dispatch`, `pkg/vm/pipeline`, `pkg/vm/process`; `pkg/sql/compile`
  builds. gofmt clean.


XuPeng-SH (changes requested):
I think this still needs changes before merge. The bounded cleanup itself makes sense, but the deferred spool cleanup path can free data that downstream receivers still have pending channel signals for.

The risky part is that delivery through the spool is split into two separate steps: the sender first stores the batch in the spool (`SendBatch`), and only then enqueues a `GetFromSpool` signal to the receiver channel (`pkg/sql/colexec/dispatch/sendfunc.go:60-66`, `pkg/sql/colexec/connector/connector.go:64-68`). In the new cleanup flow, if we cannot deliver the terminal signal or cannot wait for `sp.CloseWithTimeout(...)`, we keep the spool in `cleanupSpool` and later unconditionally call `ForceCleanup()` (`pkg/sql/colexec/dispatch/types.go:244-308`, `pkg/sql/colexec/connector/types.go:126-176`, `pkg/vm/pipeline/types.go:136-173`). But `ForceCleanup()` frees every non-free slot without tracking whether there are already-enqueued `GetFromSpool` signals still pointing at those slots (`pkg/container/pSpool/sender.go:154-185`).

That means a receiver channel that is only backlogged/full during teardown can still contain older `GetFromSpool` signals for real data, and those signals will later read emptied spool slots after `ForceCleanup()`. In other words this can turn a slow-but-healthy downstream pipeline into silent batch loss / early EOS during cleanup, not just a bounded cancel path. I think we need some notion of outstanding delivered references before force-freeing the spool, or a stricter condition for when the deferred cleanup is allowed to destroy spool contents.

Non-blocking note: coverage is otherwise fairly good, but I did not see a focused unit test for the new `messageCtx`-cancellation branch in `CnServerMessageHandler`, which looks like one of the main regression scenarios this PR is targeting.

aptend (approved):

aunjgr (approved):

XuPeng-SH (approved):
Looks good to me.

I did not find a substantive correctness blocker in this 3.0-dev backport.

Only non-blocking note: a couple of extra symmetry tests would be nice (for example `Connector.Reset` on a full channel, and the explicit timeout/failure path of `PipelineSpool.CloseWithTimeout()`), but I would not block merge on those.

PR #25037, opened 2026-06-17
4.0：fix: port table clone visibility for repeated alter to 4.0-dev

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #24762

Port #24987 to 4.0-dev.

- Make table clone metadata collection honor the transaction write visibility point captured when the clone reader is created.
- Reuse that visibility point for uncommitted data/tombstone object collection and appendable local data scans, so repeated ALTER plus CLONE in one transaction sees exactly the prior writes that should be visible.
- Add git4data clone BVT coverage for repeated ALTER with normal column changes, secondary/unique indexes, fulltext index, vector index, and clones between ALTERs.

Validation:

- `source $HOME/.zshrc && moenv && go test ./pkg/vm/engine/disttae ./pkg/sql/colexec/table_clone ./pkg/sql/compile`
- `source $HOME/.zshrc && moenv && make build`
- `mo-tester(main) -i clone_in_alter_table_2 -r 100`: 81/81 passed
- `mo-tester(main) -i clone_sys_db_table_to_new_db_table -r 100`: 203/203 passed after clean restart
- `mo-tester(main) -i git4data/clone -r 100`: 1174/1177 locally; the three misses were catalog count assertions in `clone_sys_db_table_to_new_db_table.sql` after prior local suite state, and the same case passed isolated after clean restart


heni02 (approved):

XuPeng-SH (changes requested):
I re-checked this PR from a first-principles / unhappy-path angle. I think there is one blocker in the new prepared-workspace commit finalization path.

### Blocker: prepared workspace is rolled back even when commit outcome is unknown

In `pkg/txn/client/operator.go`, after `workspace.Commit()` succeeds, the new defer rolls the workspace back on **any** later error:

- `workspace.Commit(ctx)` prepares the write payload and sets `workspacePrepared = true`
- that payload is then sent inside the commit request itself (`operator.go:1106-1112`)
- if the commit RPC later fails with EOF / connection-reset, `sender.go` explicitly returns `ErrTxnUnknown` because TN may already have applied the commit (`pkg/txn/rpc/sender.go:237-249`)
- but the defer in `operator.go:1044-1054` still runs `workspace.Rollback(ctx)` for that ambiguous error

For disttae workspaces, rollback GC deletes the CN-created S3 objects (`pkg/vm/engine/disttae/txn.go:2164-2166`). So a connection break after a successful commit can end up deleting files that a committed txn now references.

### Suggested fix
- Do **not** call `workspace.Rollback()` for `moerr.ErrTxnUnknown` / other ambiguous post-send failures.
- More generally, distinguish **pre-send** failures from **post-send unknown-result** failures.
- Please add a regression test in `pkg/txn/client/operator_test.go` covering `workspacePrepared + ErrTxnUnknown`, and assert that rollback GC is **not** triggered on that path.

Given the potential for data loss after an ambiguous commit failure, I think this should be **changes requested** before merge.

XuPeng-SH (changes requested):
I re-checked the latest head. The explicit post-workspace-commit failure window is fixed, and the dropped-table txn-local-shared cleanup looks improved, but there is still one remaining high-severity issue on the `ErrTxnUnknown` path.

The new defer in `pkg/txn/client/operator.go` now distinguishes ambiguous commit outcomes and calls `workspace.FinalizeCommitWithUnknownResult(...)` instead of rollback. But in `disttae.Transaction`, `FinalizeCommitWithUnknownResult()` is currently just `delTransaction()`.

That is still unsafe for clone/CN-object cleanup because `delTransaction()` clears the workspace batches before any later rollback/abort handling can inspect them. If TN eventually aborts the commit after the client saw `ErrTxnUnknown`, the later rollback path no longer has the object stats needed by `GCObjsByIdxRange(...)` to delete flushed CN-created objects, so those objects can leak.

- `sender.go` returns `ErrTxnUnknown` exactly when the commit may already have been applied, but the final TN outcome is unknown
- `operator.go` now routes that to `FinalizeCommitWithUnknownResult(...)`
- `txn.go` implements `FinalizeCommitWithUnknownResult()` as `delTransaction()`
- rollback GC in `types.go` only walks non-empty write batches / object-stats vectors

So the new code avoids deleting possibly-committed objects on the ambiguous path, but it also throws away the metadata needed if the final outcome later turns out to be abort and cleanup must run.

**Concrete suggestions**

1. Preserve enough workspace/object metadata on `ErrTxnUnknown` for a later rollback/abort cleanup, instead of calling `delTransaction()` immediately.
2. If you need a special unknown-result finalization path, make it release only the pieces that are safe to drop while retaining the flushed-object bookkeeping required by GC.
3. Add a focused regression test for: `workspacePrepared + ErrTxnUnknown`, followed by the later abort/rollback resolution path, and assert that flushed CN objects are still discoverable for cleanup.

XuPeng-SH (changes requested):

The explicit post-workspace-commit failure window is fixed, but I still see one remaining high-severity issue on the `ErrTxnUnknown` path.

### Blocker: `ErrTxnUnknown` still drops the metadata needed for later clone-object cleanup

The new defer in `pkg/txn/client/operator.go` now distinguishes ambiguous commit outcomes and calls `workspace.FinalizeCommitWithUnknownResult(...)` instead of rollback. That is directionally right, but in `disttae.Transaction`, `FinalizeCommitWithUnknownResult()` currently just calls `delTransaction()`.

That is still unsafe for clone/CN-object cleanup, because `delTransaction()` clears the workspace bookkeeping before any later rollback/abort resolution can inspect it. If TN eventually aborts the commit after the client has already seen `ErrTxnUnknown`, the later cleanup path no longer has the object stats / write-batch metadata that `GCObjsByIdxRange(...)` needs to delete flushed CN-created objects. So the PR avoids deleting possibly-committed objects on the ambiguous path, but it also throws away the metadata needed if the final outcome later turns out to be abort and cleanup still must run.

### Suggested fix
- Preserve enough workspace/object metadata on `ErrTxnUnknown` for a later abort/rollback cleanup, instead of immediately calling `delTransaction()`.
- If you need a special unknown-result finalization path, make it release only the pieces that are safe to drop while retaining the flushed-object bookkeeping required by GC.
- Please add a focused regression test for:
  - `workspacePrepared + ErrTxnUnknown`
  - followed by the later abort/rollback resolution path
  - and assert that flushed CN-created objects are still discoverable for cleanup.

Given the clone-object leak risk on the ambiguous commit path, I think this still needs **changes requested** on the current head.

gouhongshen:
I re-checked this on current head `7a339d16157edfc8d076f16c576dc9d56d5cde72`.

The specific blocker statement in the latest review is stale/not applicable to this head: `disttae.Transaction.FinalizeCommitWithUnknownResult()` no longer calls `delTransaction()`. It keeps the workspace/object-stat metadata intact, so the clone rollback metadata is not dropped by that method.

There was, however, a related CCPR-specific unknown-result cleanup gap raised on the main PR: CCPR txn cache entries also need to be removed without deleting object files. That is now synced here in `7a339d16157edfc8d076f16c576dc9d56d5cde72`: `FinalizeCommitWithUnknownResult()` calls `CCPRTxnCache.OnTxnUnknownResult(...)` for CCPR txns, which removes cache tracking/indexes without running object GC or `delTransaction()`.

Verified with:
- `go test ./pkg/vm/engine/disttae -run 'TestCCPRTxnCache_OnTxnUnknownResultRemovesTrackingWithoutGC|TestTransactionFinalizeCommitUnknownCleansCCPRCache|TestCCPRTxnCache_OnTxnRollback_NonExistent|TestCCPRTxnCache_Rollback_MultiTxn|TestCloneTxnUnknownCommitPreservesRollbackCleanup|TestCloneTxn|TestTableMetaReader' -count=1`
- `go test ./pkg/txn/client -run 'TestCommitKeepsPreparedWorkspaceAfterTxnUnknown|TestCommitFinalizesPreparedWorkspaceAfterCommitSuccess|TestCommitRollsBackPreparedWorkspaceAfterCommitSendError' -count=1`
- `go test ./pkg/frontend -run 'TestCommitTxnUnknownPreservesTxnOperatorForLaterCleanup|Test_commit|Test_isErrorRollbackWholeTxn' -count=1`

XuPeng-SH (approved):
Re-checked the latest head. The previous 4.0-dev blockers are now closed: the frontend path preserves the txn operator across `ErrTxnUnknown`, the prepared-workspace unknown-result branch no longer rolls back, and the CCPR unknown-result path now removes cache tracking without wrongly GCing objects. With those lifecycle fixes plus the earlier clone/shared-file cleanup work, I do not see a remaining correctness issue that should keep this blocked.

iamlinjunhong (approved):

ouyuanning (approved):

aunjgr (approved):

PR #25038, opened 2026-06-17
add locks functions

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24486

增加user level locks函数: is_used_lock, release_all_locks.

• IS_USED_LOCK：

  - 新增 lockservice 点查接口 GetLockHolder(tableID, row, options)
  - 复用现有 GetTxnLock 远程 RPC，并在 response 中增加 Holder / Found，让 holder 查询能路由到真正持有 lock table 的 CN，支持跨 CN。
  - user lock 的 row key 保持为 <account>\0<lock_name>；synthetic txn id 中编码 <owner>\0<connection_id>\0<lock_name>。
  - IS_USED_LOCK 从 holder 的 synthetic txn id 中解析 connection_id；如果锁不存在或 holder txn id 不是 user lock 格式，则返回 NULL。

  RELEASE_ALL_LOCKS：

  - 只维护本 session 的本地状态：counts 记录可重入计数，byOwner 记录当前 session 持有的 lock name。
  - 只释放当前 session 自己持有的 user locks，通过当前 session 的 owner 和 connection id 生成 synthetic txn id 执行 unlock。
  - 返回释放的本地可重入 lock 次数。
  - 不再维护本地全局 holders map，避免错误报告或释放其他 session / 其他 CN 持有的锁。


matrix-meow:
## 阻塞点:`GetTxnLockResponse` 新增 `Found` 字段在混合版本下会全量误判 not-found

在 `lock_table_remote.go` 的 `doGetLock` 里新增的短路:

```go
if !resp.GetTxnLock.Found {
    return Lock{}, false, nil
}
```

`Found` 只有在新版 owner CN 的 `handleRemoteGetLock` 中才会被置 `true`。**滚动升级 / 新旧 CN 并存时**,只要新 CN 作为 caller、旧 CN 作为 lock-table owner,旧 CN 永远不回 `Found` → 所有走 `remoteLockTable.getLock → doGetLock` 的请求都会被当成"锁不存在"。

同时下面这行也改了语义:
```go
if len(resp.GetTxnLock.Holder.TxnID) > 0 {
    lock.holders.add(resp.GetTxnLock.Holder)   // 从 resp 取 holder,不再是请求 txn
}
```
既有非本次新增的 `getLock` 调用方(锁等待检测 / txn 解锁路径等,见 `lock_table_proxy.go:202`、`txn.go:421`、`service_remote.go:415`)在混合版本下,holder 回填行为也会跟着变。

**想确认的是 lockservice RPC 的版本兼容策略**:
1. 这个 PR 是否假设 lockservice 是原子升级(不允许 mixed-version)?
2. 如果允许滚动升级,建议把 `Found` 改为非破坏式新增——即缺省(旧节点返回零值 `false`)时仍按旧逻辑构造 `Lock`,只在 `Found==true` 时才信任 `Holder` 字段;`!Found` 直接 return 会改变既有 not-found 语义。

建议确认后再合入。如果确实不存在 mixed-version 部署,麻烦在这里明确说明一下,我这边就放行。

XuPeng-SH (changes requested):
## 🔴 Critical Issues - REQUEST CHANGES

深度review发现**2个阻塞性问题**：

---

### 🔴 #1: Txn ID格式向后不兼容 (BLOCKER)

**问题：** 升级前创建的锁在升级后无法释放！

旧格式: `mo-user-level-lock\x00<owner>\x00<lock_name>`
新格式: `mo-user-level-lock\x00<owner>\x00<conn_id>\x00<lock_name>`

**影响场景：**
```sql
-- 旧代码：
SELECT GET_LOCK('mylock', 10);  -- txnID = "...\x00owner\x00mylock"
-- [升级]
-- 新代码：
SELECT RELEASE_LOCK('mylock');  -- txnID = "...\x00owner\x00456\x00mylock"
-- 返回NULL → 锁泄漏！
```

**必须修复：** 双格式支持
```go
// Try new format
err := Unlock(userLevelLockTxnID(owner, connID, name))
// Fallback to old format
if moerr.IsMoErrCode(err, moerr.ErrLockNotFound) {
    err = Unlock(userLevelLockTxnIDOld(owner, name))
}
```

---

### 🔴 #2: RELEASE_ALL_LOCKS返回值错误

**问题：** 返回refcount而非唯一锁数量（MySQL不兼容）

**预期：** GET_LOCK('a',0) 3次 + GET_LOCK('b',0) → RELEASE_ALL_LOCKS() 返回**2**
**实际：** 返回**4**（'a'的refcount=3 + 'b'=1）

**必须修复：**
```go
released := make(map[string]struct{})  // Track unique locks
for _, name := range ... {
    released[name] = struct{}{}  // Count once per lock
    // ...
}
return int64(len(released))
```

---

### 🟡 Non-Blocking

3. Silent error: `_ = Unlock(...)` 应记录日志
4. 缺少测试：可重入锁、向后兼容、跨CN

---

## ✅ Strengths

- RPC设计优雅，正确路由
- IS_USED_LOCK完全MySQL兼容
- BVT覆盖良好

详细分析：`~/.copilot/.../pr25038-analysis.md`

aptend (changes requested):
I found two blocking compatibility issues.

1. `pkg/lockservice/lock_table_remote.go:304`: the new client now treats `GetTxnLockResponse.Found == false` as "lock not found", but `Method_GetTxnLock` is still registered as `MORPCVersion1`, and older CNs never populate the new `Found` field. During a mixed-version/rolling upgrade, a new CN querying an old holder will decode the default `false` and silently skip an existing remote lock. That breaks the existing `getLock` path used by deadlock/waiter observability and makes `GetLockHolder`/`IS_USED_LOCK()` return NULL for locks held on old peers. This needs either a backward-compatible interpretation or an RPC/versioning split so new clients do not treat old responses as absence.

2. `pkg/sql/plan/function/function_id.go:733-735`: this changes the existing fixed id of `NAME_CONST` from 521 to 523. `function_id_test.go` says these ids are fixed, and updating the expected value there defeats that guard. Function ids are serialized across plans/protocols, so existing ids should remain stable. Please keep `NAME_CONST = 521` and append the new ids after it, e.g. `IS_USED_LOCK = 522`, `RELEASE_ALL_LOCKS = 523`, `FUNCTION_END_NUMBER = 524`.

heni02 (approved):

pkg/sql/plan/function/func_unary.go
```diff
@@ -7222,42 +7251,79 @@ func isUserLevelLockFree(name string, proc *process.Process) (int64, error) {
 		return 0, err
 	}
 	owner := userLevelLockOwner(proc)
+	connID := userLevelLockConnectionID(proc)
 	probeOwner := owner + "\x00probe\x00" + name
```
gouhongshen: **`probeOwner` embeds `\x00`, conflicting with `txnID` separator format**

`probeOwner` is constructed as `owner + "\x00probe\x00" + name`, which embeds `\x00` separators. This `probeOwner` is then passed to `userLevelLockTxnID`, which itself uses `\x00` as the field separator:

```go
fmt.Sprintf("mo-user-level-lock\x00%s\x00%d\x00%s", owner, connID, name)
```

The resulting txnID will have extra `\x00`-delimited segments compared to a normal lock txnID. This works by accident today because probe-originated txnIDs are never parsed back with `userLevelLockConnectionIDFromTxnID` (which expects exactly 4 `\x00`-delimited parts). However, the design is fragile — any future code path that attempts to parse a probe-originated txnID will silently get wrong results.

**Suggestion**: Avoid embedding `\x00` in the owner/name components. Use a separator that cannot collide with the txnID format, or encode the probe marker differently.

pkg/lockservice/service_observability.go
```diff
@@ -49,6 +49,28 @@ func (s *service) GetWaitingList(
...
+	var holder pb.WaitTxn
+	var found bool
+	l.getLock(row, pb.WaitTxn{}, func(lock Lock) {
+		lock.IterHolders(func(v pb.WaitTxn) bool {
+			holder = v
+			found = true
+			return false
+		})
+	})
+	return holder, found, nil
```
gouhongshen: **Inconsistent `found` semantics between local and remote paths**

`GetLockHolder` (this method) returns `found=false` when the lock entry exists but has no current holders — `IterHolders` simply never sets `found=true`.

However, on the remote path (`handleRemoteGetLock` → `doGetLock`), `Found` is set to `true` whenever the lock *entry* exists in the lock table, regardless of whether it has an active holder.

This means `IS_USED_LOCK` can return different results depending on whether the lock table is local or remote for the same logical state (lock exists, no holder). Consider aligning the semantics — either both should report "found" when the lock entry exists, or both should only report "found" when there is an active holder.

proto/lock.proto
```diff
@@ -206,6 +206,8 @@ message GetTxnLockRequest {
 message GetTxnLockResponse {
   int32            Value       = 1;
   repeated WaitTxn WaitingList = 2 [(gogoproto.nullable) = false];
+  WaitTxn          Holder      = 3 [(gogoproto.nullable) = false];
```
gouhongshen: **Rolling upgrade compatibility concern with `Found` field**

The `Found` field uses proto3 scalar `bool`, which defaults to `false`. During a rolling upgrade, when a new-version CN queries a lock held on an old-version CN, the old CN's `handleRemoteGetLock` will not set `Found` (it doesn't know about this field), so the response will have `Found=false` by default.

This means `doGetLock` will treat **every** remote lock query during the upgrade window as "lock not found", which is incorrect. This path is also used by deadlock detection (`txn.go`) and the lock proxy (`lock_table_proxy.go`).

**Suggestion**: Consider declaring this as `optional bool` so the consumer can distinguish "explicitly false" from "unset/unknown". Alternatively, have `doGetLock` fall back to the previous behavior (treating `err == nil` as found) when the remote CN is an older version.

pkg/sql/plan/function/func_unary.go
```diff
@@ -7222,42 +7251,79 @@ func isUserLevelLockFree(name string, proc *process.Process) (int64, error) {
...
 	for _, name := range userLevelLocksForOwner(owner) {
 		for {
 			count, held := untrackUserLevelLock(owner, name)
 			if !held {
 				break
 			}
+			released++
 			if count == 0 {
-				_ = proc.GetLockService().Unlock(context.Background(), userLevelLockTxnID(owner, name), timestamp.Timestamp{})
+				_ = proc.GetLockService().Unlock(context.Background(), userLevelLockTxnID(owner, connID, name), timestamp.Timestamp{})
```
gouhongshen: **Unlock errors silently discarded in `releaseAllUserLevelLocks`**

The `Unlock` call's error is discarded with `_ =`. While the lockservice has a lease/TTL mechanism that will eventually reclaim the lock, silently swallowing the error makes it impossible to diagnose why a lock wasn't released on time.

Additionally, the `released` counter is incremented before the `Unlock` call, so it counts the lock as released even if `Unlock` actually failed — this may give a false sense of correctness.

**Suggestion**: At minimum, log the error (e.g., `logutil.Warn`) so operators can investigate. Consider returning an error or aggregating failures so the caller is aware something went wrong.

gouhongshen:
**Suggestion: expand test coverage for error paths and concurrency**

The current tests cover the happy paths well (basic lock/unlock, null inputs, holder connection ID retrieval, release-all count). A few areas would benefit from additional coverage:

1. **Error paths**: No test exercises the case where `Unlock` returns an error in `releaseAllUserLevelLocks`. Currently the error is silently discarded — a test should verify the behavior and ensure the returned count is still correct.

2. **Concurrent access**: Multiple sessions acquiring/releasing the same lock name simultaneously could expose race conditions. A `go test -race` test with concurrent `GET_LOCK` / `RELEASE_LOCK` / `RELEASE_ALL_LOCKS` calls would strengthen confidence.

3. **`IS_USED_LOCK` remote path**: The `Found=false` path when the lock does not exist on a remote CN is not tested. Also, the `connID=0` edge case and malformed-holder-txnID parsing should be explicitly covered.

4. **Destructive inputs**: `GET_LOCK('name\x00with\x00nulls', 0)` — verify that null-byte-containing names are properly rejected by `normalizeUserLevelLockName`, since `\x00` is the internal separator used in txnID encoding.

daviszhen:
> ## 🔴 Critical Issues - REQUEST CHANGES
> 深度review发现**2个阻塞性问题**：
> 
> ### 🔴 #1: Txn ID格式向后不兼容 (BLOCKER)
> **问题：** 升级前创建的锁在升级后无法释放！
> 
> 旧格式: `mo-user-level-lock\x00<owner>\x00<lock_name>` 新格式: `mo-user-level-lock\x00<owner>\x00<conn_id>\x00<lock_name>`
> 
> **影响场景：**
> 
> ```sql
> -- 旧代码：
> SELECT GET_LOCK('mylock', 10);  -- txnID = "...\x00owner\x00mylock"
> -- [升级]
> -- 新代码：
> SELECT RELEASE_LOCK('mylock');  -- txnID = "...\x00owner\x00456\x00mylock"
> -- 返回NULL → 锁泄漏！
> ```
> 
> **必须修复：** 双格式支持
> 
> ```go
> // Try new format
> err := Unlock(userLevelLockTxnID(owner, connID, name))
> // Fallback to old format
> if moerr.IsMoErrCode(err, moerr.ErrLockNotFound) {
>     err = Unlock(userLevelLockTxnIDOld(owner, name))
> }
> ```
> 
> ### 🔴 #2: RELEASE_ALL_LOCKS返回值错误
> **问题：** 返回refcount而非唯一锁数量（MySQL不兼容）
> 
> **预期：** GET_LOCK('a',0) 3次 + GET_LOCK('b',0) → RELEASE_ALL_LOCKS() 返回**2** **实际：** 返回**4**（'a'的refcount=3 + 'b'=1）
> 
> **必须修复：**
> 
> ```go
> released := make(map[string]struct{})  // Track unique locks
> for _, name := range ... {
>     released[name] = struct{}{}  // Count once per lock
>     // ...
> }
> return int64(len(released))
> ```
> 
> ### 🟡 Non-Blocking
> 3. Silent error: `_ = Unlock(...)` 应记录日志
> 4. 缺少测试：可重入锁、向后兼容、跨CN
> 
> ## ✅ Strengths
> * RPC设计优雅，正确路由
> * IS_USED_LOCK完全MySQL兼容
> * BVT覆盖良好
> 
> 详细分析：`~/.copilot/.../pr25038-analysis.md`

• 本轮修改总结：

  - releaseAllUserLevelLocks 改为返回 (int64, error)。
  - RELEASE_ALL_LOCKS() SQL 函数现在会把底层 Unlock 错误返回给调用方。
  - session 关闭清理 ReleaseUserLevelLocks() 仍保持 best-effort，忽略返回值。
  - Unlock 失败时继续打 Warn，且失败的锁不计入 released。
  - UT 更新了错误路径：Unlock 失败时要求返回 error。
  - BVT 补了两个用户语义 case：
      - release_all_locks() 返回唯一 lock name 数量，而不是 refcount。
      - user lock 名字大小写不敏感。

daviszhen:
> I found two blocking compatibility issues.
> 
> 1. `pkg/lockservice/lock_table_remote.go:304`: the new client now treats `GetTxnLockResponse.Found == false` as "lock not found", but `Method_GetTxnLock` is still registered as `MORPCVersion1`, and older CNs never populate the new `Found` field. During a mixed-version/rolling upgrade, a new CN querying an old holder will decode the default `false` and silently skip an existing remote lock. That breaks the existing `getLock` path used by deadlock/waiter observability and makes `GetLockHolder`/`IS_USED_LOCK()` return NULL for locks held on old peers. This needs either a backward-compatible interpretation or an RPC/versioning split so new clients do not treat old responses as absence.
> 2. `pkg/sql/plan/function/function_id.go:733-735`: this changes the existing fixed id of `NAME_CONST` from 521 to 523. `function_id_test.go` says these ids are fixed, and updating the expected value there defeats that guard. Function ids are serialized across plans/protocols, so existing ids should remain stable. Please keep `NAME_CONST = 521` and append the new ids after it, e.g. `IS_USED_LOCK = 522`, `RELEASE_ALL_LOCKS = 523`, `FUNCTION_END_NUMBER = 524`.

• 本轮修改总结：

  - releaseAllUserLevelLocks 改为返回 (int64, error)。
  - RELEASE_ALL_LOCKS() SQL 函数现在会把底层 Unlock 错误返回给调用方。
  - session 关闭清理 ReleaseUserLevelLocks() 仍保持 best-effort，忽略返回值。
  - Unlock 失败时继续打 Warn，且失败的锁不计入 released。
  - UT 更新了错误路径：Unlock 失败时要求返回 error。
  - BVT 补了两个用户语义 case：
      - release_all_locks() 返回唯一 lock name 数量，而不是 refcount。
      - user lock 名字大小写不敏感。

daviszhen:
> **Suggestion: expand test coverage for error paths and concurrency**
> 
> The current tests cover the happy paths well (basic lock/unlock, null inputs, holder connection ID retrieval, release-all count). A few areas would benefit from additional coverage:
> 
> 1. **Error paths**: No test exercises the case where `Unlock` returns an error in `releaseAllUserLevelLocks`. Currently the error is silently discarded — a test should verify the behavior and ensure the returned count is still correct.
> 2. **Concurrent access**: Multiple sessions acquiring/releasing the same lock name simultaneously could expose race conditions. A `go test -race` test with concurrent `GET_LOCK` / `RELEASE_LOCK` / `RELEASE_ALL_LOCKS` calls would strengthen confidence.
> 3. **`IS_USED_LOCK` remote path**: The `Found=false` path when the lock does not exist on a remote CN is not tested. Also, the `connID=0` edge case and malformed-holder-txnID parsing should be explicitly covered.
> 4. **Destructive inputs**: `GET_LOCK('name\x00with\x00nulls', 0)` — verify that null-byte-containing names are properly rejected by `normalizeUserLevelLockName`, since `\x00` is the internal separator used in txnID encoding.

• 本轮修改总结：

  - releaseAllUserLevelLocks 改为返回 (int64, error)。
  - RELEASE_ALL_LOCKS() SQL 函数现在会把底层 Unlock 错误返回给调用方。
  - session 关闭清理 ReleaseUserLevelLocks() 仍保持 best-effort，忽略返回值。
  - Unlock 失败时继续打 Warn，且失败的锁不计入 released。
  - UT 更新了错误路径：Unlock 失败时要求返回 error。
  - BVT 补了两个用户语义 case：
      - release_all_locks() 返回唯一 lock name 数量，而不是 refcount。
      - user lock 名字大小写不敏感。

XuPeng-SH (changes requested):
I re-checked the latest head. The earlier txn-id compatibility and function-id stability concerns look addressed now, but I still see one real correctness issue in the new holder-lookup path.\n\n`pkg/lockservice/lock_table_remote.go:245-257` refreshes the bind on error via `handleError`, but if that refresh succeeds (`handleError` returns nil) it immediately returns `(false, nil)` instead of retrying `doGetLockHolder` on the new bind. That means a stale remote binding is surfaced to SQL as "no holder" / NULL even though the lock may still exist on the new CN. With `IS_USED_LOCK()` now relying on `GetLockHolder`, this becomes a user-visible false negative during bind migration / CN movement, not just an internal observability gap.\n\nPlease retry the holder lookup after a successful bind refresh, the same way the caller needs the post-refresh result rather than a silent "not found", and add a targeted test that exercises `GetLockHolder` across a bind-change path.

daviszhen:
> I re-checked the latest head. The earlier txn-id compatibility and function-id stability concerns look addressed now, but I still see one real correctness issue in the new holder-lookup path.\n\n`pkg/lockservice/lock_table_remote.go:245-257` refreshes the bind on error via `handleError`, but if that refresh succeeds (`handleError` returns nil) it immediately returns `(false, nil)` instead of retrying `doGetLockHolder` on the new bind. That means a stale remote binding is surfaced to SQL as "no holder" / NULL even though the lock may still exist on the new CN. With `IS_USED_LOCK()` now relying on `GetLockHolder`, this becomes a user-visible false negative during bind migration / CN movement, not just an internal observability gap.\n\nPlease retry the holder lookup after a successful bind refresh, the same way the caller needs the post-refresh result rather than a silent "not found", and add a targeted test that exercises `GetLockHolder` across a bind-change path.

已经修改

XuPeng-SH (changes requested):
I re-checked the new head. The previous stale-bind false-NULL issue is not fully fixed yet.\n\n`pkg/lockservice/lock_table_remote.go:245-257` now does `continue` after `handleError()` refreshes the bind, but that loop is still running on the same `remoteLockTable` instance. `doGetLockHolder()` always sends `req.LockTable = l.bind`, and on bind change `service.handleBindChanged()` replaces the tableGroups entry with a newly created lock-table object (it does not mutate this in-flight `l`). So after a bind refresh you can still keep retrying on the stale remote table pointer instead of actually querying through the new bind / new local-vs-remote table object.\n\nThe new test does not pin the real production behavior here because its fake bind-changed callback never swaps the lock-table object, and the mock `GetLockHolder` handler returns success on the second call regardless of which bind was used. Please make the retry reacquire the lock table after bind change (or bubble `ErrLockTableBindChanged` back so the caller reacquires and retries), and add a test that proves the retry goes through the refreshed bind/object rather than the stale one.

XuPeng-SH (changes requested):
I re-checked the latest head and still see two correctness issues.

1. **`IS_USED_LOCK()` can still ignore statement cancellation when the holder CN is unreachable.**
   `service.GetLockHolder()` now re-checks `ctx.Err()` while reacquiring after `ErrLockTableBindChanged`, but `remoteLockTable.getLockHolder()` itself still loops forever on transient RPC failures with backoff and no context parameter. In the unhappy path where the remote RPC keeps failing but the allocator keeps returning the same bind, `IS_USED_LOCK()` can still hang until the process/session is interrupted instead of respecting timeout/cancel.

   **Suggestion:** thread the caller context through `getLockHolder` / `doGetLockHolder`, and break the retry loop as soon as `ctx.Done()` fires.

2. **Remote `GetLockHolder` still does not carry enough data for a cold lookup.**
   The RPC only sends `GetLockHolder.Row`, but the server-side fallback path in `getLocalLockTable()` recreates a missing table from `req.Lock.Rows` and `req.Lock.Options.Sharding`. Those fields are unset for `Method_GetLockHolder`, so a cold remote lookup can still fail with `ErrLockTableNotFound` / stale-bind behavior even when a holder exists.

   **Suggestion:** either make `getLocalLockTable()` method-aware for `GetLockHolder` (use `req.GetLockHolder.Row` and `req.LockTable.Sharding`), or extend the `GetLockHolder` request payload so the handler has the same lookup inputs as `Lock` / `GetTxnLock`.

Non-blocking coverage note: I still do not see a direct transient-retry test for `remoteLockTable.getLockHolder()` where RPC fails once, bind does not change, and the retry later succeeds.

daviszhen:
> I re-checked the latest head and still see two correctness issues.
> 
> 1. **`IS_USED_LOCK()` can still ignore statement cancellation when the holder CN is unreachable.**
>    `service.GetLockHolder()` now re-checks `ctx.Err()` while reacquiring after `ErrLockTableBindChanged`, but `remoteLockTable.getLockHolder()` itself still loops forever on transient RPC failures with backoff and no context parameter. In the unhappy path where the remote RPC keeps failing but the allocator keeps returning the same bind, `IS_USED_LOCK()` can still hang until the process/session is interrupted instead of respecting timeout/cancel.
>    **Suggestion:** thread the caller context through `getLockHolder` / `doGetLockHolder`, and break the retry loop as soon as `ctx.Done()` fires.
> 2. **Remote `GetLockHolder` still does not carry enough data for a cold lookup.**
>    The RPC only sends `GetLockHolder.Row`, but the server-side fallback path in `getLocalLockTable()` recreates a missing table from `req.Lock.Rows` and `req.Lock.Options.Sharding`. Those fields are unset for `Method_GetLockHolder`, so a cold remote lookup can still fail with `ErrLockTableNotFound` / stale-bind behavior even when a holder exists.
>    **Suggestion:** either make `getLocalLockTable()` method-aware for `GetLockHolder` (use `req.GetLockHolder.Row` and `req.LockTable.Sharding`), or extend the `GetLockHolder` request payload so the handler has the same lookup inputs as `Lock` / `GetTxnLock`.
> 
> Non-blocking coverage note: I still do not see a direct transient-retry test for `remoteLockTable.getLockHolder()` where RPC fails once, bind does not change, and the retry later succeeds.


已经修改

aunjgr (approved):

XuPeng-SH (changes requested):
I re-checked the latest head. The previous remote-holder blockers are fixed, but I still see one remaining local stale-bind race in the holder-lookup path.

`service.GetLockHolder()` and `handleRemoteGetLockHolder()` still do:
1. get/recreate a lock table,
2. then call `l.getLockHolder(...)`
without holding `bindChangeMu` across those two steps.

If a bind change closes/replaces that local table in between, `localLockTable.getLockHolder()` returns `(false, nil)` for `mu.closed`, which is indistinguishable from “lock does not exist”. The caller therefore does not retry against the new authoritative table and can transiently report NULL / no owner even though the bind has moved and the new table may still have a holder.

- `service.handleBindChanged()` swaps the table under `bindChangeMu` and closes the old one
- `localLockTable.getLockHolder()` still returns `pb.WaitTxn{}, false, nil` when `l.mu.closed`
- unlike `Lock`, the holder-lookup path does not hold `bindChangeMu.RLock()` across table acquisition + lookup

**Concrete suggestions**

1. Mirror the `Lock` path and hold `bindChangeMu.RLock()` across `getLockTableWithCreate/getLocalLockTable` and `getLockHolder`.
2. Or, if you prefer the retry pattern already used elsewhere, make `localLockTable.getLockHolder()` surface `ErrLockTableBindChanged` when it is closed by rebinding so the caller can reacquire and retry on the current table.
3. Add one focused test for the local race (not the remote bind-changed case already covered): acquire a local table, rebind/close it before `getLockHolder`, and assert the lookup retries on the new table instead of returning `(false, nil)`.

daviszhen:
> I re-checked the latest head. The previous remote-holder blockers are fixed, but I still see one remaining local stale-bind race in the holder-lookup path.
> 
> `service.GetLockHolder()` and `handleRemoteGetLockHolder()` still do:
> 
> 1. get/recreate a lock table,
> 2. then call `l.getLockHolder(...)`
>    without holding `bindChangeMu` across those two steps.
> 
> If a bind change closes/replaces that local table in between, `localLockTable.getLockHolder()` returns `(false, nil)` for `mu.closed`, which is indistinguishable from “lock does not exist”. The caller therefore does not retry against the new authoritative table and can transiently report NULL / no owner even though the bind has moved and the new table may still have a holder.
> 
> You can see the pieces in the current head:
> 
> * `service.handleBindChanged()` swaps the table under `bindChangeMu` and closes the old one
> * `localLockTable.getLockHolder()` still returns `pb.WaitTxn{}, false, nil` when `l.mu.closed`
> * unlike `Lock`, the holder-lookup path does not hold `bindChangeMu.RLock()` across table acquisition + lookup
> 
> **Concrete suggestions**
> 
> 1. Mirror the `Lock` path and hold `bindChangeMu.RLock()` across `getLockTableWithCreate/getLocalLockTable` and `getLockHolder`.
> 2. Or, if you prefer the retry pattern already used elsewhere, make `localLockTable.getLockHolder()` surface `ErrLockTableBindChanged` when it is closed by rebinding so the caller can reacquire and retry on the current table.
> 3. Add one focused test for the local race (not the remote bind-changed case already covered): acquire a local table, rebind/close it before `getLockHolder`, and assert the lookup retries on the new table instead of returning `(false, nil)`.

已经按照建议修改了

XuPeng-SH (approved):
Re-checked the latest head. The remaining local stale-bind holder-lookup race is now fixed: `GetLockHolder()` holds `bindChangeMu.RLock()` across table acquisition and `getLockHolder(...)`, so a rebinding-close can no longer be misreported as a clean "no holder" result. Together with the earlier remote ctx/sharding/bind-refresh fixes and the new lockservice tests, I do not see a remaining correctness issue that should keep this blocked.

aptend (approved):

gouhongshen (approved):

iamlinjunhong (approved):

Issue #25048, opened 2026-06-18
[Feature Request]:   perf(vector): CollectOffsetsByPrefixInFactory walks every block row even when the needle set is tiny


cpegeric:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

vector.CollectOffsetsByPrefixInFactory (in pkg/container/vector/search.go) — the sorted "prefix IN" PK filter — resolves a set of needle prefixes against a
  sorted data column by a two-pointer merge. The needle pointer advances monotonically, but the data pointer steps one row at a time, so the function is
  O(block-rows) regardless of how few needles there are. When the needle set is small relative to the block (the common case), most of that walk is spent
  skipping rows that match nothing.

  Where this runs (why it's hot)
  
  It's not a user-facing function; the planner emits prefix_in/prefix_eq for encoded composite keys:
  - Composite primary keys — WHERE a IN (...) on a PRIMARY KEY (a, b) table becomes a prefix match against the serial(a)·serial(b) key column.
  - Secondary / master indexes — lookups on __mo_index_idx_col (serial_full(value, pk)).
  - Vector (IVF) search — the index returns candidate PKs, which are resolved against the base table's encoded PK column to fetch rows.
  
  In all of these the data column is the table's sorted PK, so the block is physically ordered — which is exactly the precondition that makes a faster
  algorithm possible.
  
  The problem, in code
  ```
  rpos := 0
  for i := 0; i < lvlen; i++ {              // walks EVERY row of the sorted block
      lval := lcol[i]...
      for types.PrefixCompare(lval, rval) > 0 {  // needle pointer only advances
          rpos++; if rpos == rvlen { return }; rval = rcol[rpos]...
      }
      if bytes.HasPrefix(lval, rval) { append i }
  }
  The cmp < 0 case (current row sorts below the current needle, cannot match) falls through to i++ — a linear scan over potentially long runs of non-matching
  rows.
  ```

  Evidence

  Profiling a 1M-row ivfflat search (int8, lists=1000, probe_limit=3), the PK→offset resolution cluster — CollectOffsetsByPrefixInFactory + PrefixCompare /
  bytes.HasPrefix / cmpbody / Varlena.GetByteSlice — accounted for ~14% of CPU, rivaling the distance kernel itself (l2sqInt8 ~10.5%). For a top-k query the
[...]

PR #25049, opened 2026-06-18
perf(vector): CollectOffsetsByPrefixInFactory walks every block row even when the needle set is tiny

cpegeric:
The sorted "prefix IN" PK filter advanced its needle pointer monotonically but still stepped through every data row one at a time — an O(block-rows) walk even when the needle set is tiny relative to the block (e.g. an IVF top-k resolved against the base table's encoded PK column).

Because the data column is sorted — this is only wired as the SortedSearchFunc and reached via DecideSearchFunc(IsSorted()); unsorted blocks use the Linear variant — PrefixCompare(lcol[i], rval) is monotonic in i, so we can skip a run of below-needle rows with an exponential+binary gallop instead of stepping one at a time. Complexity drops to O(needles·log gap) when needles are sparse and degrades gracefully to the linear merge when they are dense (no gaps -> no gallop), so no case regresses.

Output is unchanged: each row is tested once and offsets stay ascending. Proven byte-identical to the original merge over 4000 randomized rounds, including deliberately shuffled needle order (the gallop bisects only the sorted data column, never the needles). Also document the sorted precondition and guard the empty-needle case, which previously indexed rcol[0] and panicked.

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25048

Because the data column is sorted, PrefixCompare(lcol[i], rval) is monotonic in i (prefix truncation preserves lexicographic order). So a run of
  below-needle rows can be skipped with an exponential + binary "gallop" instead of stepping one at a time:
  - sparse needles → O(needles · log gap);
  - dense needles → no gaps → no gallop → degrades to the existing linear merge (no regression).

  Output is unchanged (each row tested once, offsets ascending). The function is only ever wired as the SortedSearchFunc (reached via
  BlockReadFilter.DecideSearchFunc(IsSorted())); unsorted/fake-PK blocks already use the brute-force LinearCollectOffsetsByPrefixInFactory, so the sortedness
  precondition already holds and is now documented.

  Correctness / risk

  - The gallop bisects only the sorted data column; the needle-pointer logic is byte-for-byte the original, so output is equivalent for any needle ordering.
  - Backed by a differential test: 4000 randomized rounds (including deliberately shuffled needles) asserting byte-identical output to the original merge,
  plus explicit edge cases (gaps of every size, prefix-of-prefix needles, empty inputs).
  - A pre-existing empty-needle panic (rcol[0] on a zero-length vec) is also guarded.

  Impact

  After the change, the PK→offset comparison work drops from ~14% to negligible (the gallop's compares fall to ~0.03s in the same profile); what remains in
  that path is output-slice growth proportional to the number of matched rows. Since the function is shared across composite-PK, secondary-index, and
  vector-search reads, the win applies broadly, not just to IVF.


aunjgr (approved):
Correct use of galloping search on a monotonic predicate. The differential tests (8000 randomized rounds) prove byte-identical output to both the Linear oracle and the original merge reference. Empty rvec guard fixes a pre-existing panic. Clean.

XuPeng-SH (approved):
Reviewed the current head from the unhappy-path/correctness angle as well. The sorted-input precondition is now explicit, the empty-needle panic is guarded, and the new differential/edge-case coverage is strong enough that I do not see a substantive logic or regression issue to block on.

Issue #24966, opened 2026-06-12
LogService HAKeeper can regress to stale shard membership after Dragonboat imported snapshot metadata rewrite


LeftHandCold:
## Problem

A production LogService incident on June 4 exposed a Dragonboat snapshot import metadata bug that can make HAKeeper observe a stale shard membership as a newer view.

When Dragonboat imported a snapshot, the imported snapshot metadata could use the snapshot index as `Membership.ConfigChangeId`. MatrixOne maps Dragonboat membership `ConfigChangeId` to the LogService shard epoch visible to HAKeeper. If an old membership is imported with a large snapshot index, HAKeeper can treat that stale membership as the latest shard view.

In the observed incident, this allowed HAKeeper to schedule membership changes from an obsolete view and attempt to replace the live LogService replica, creating a split membership state.

## Root cause

Dragonboat's snapshot import path mixed two different concepts:

- snapshot index: the raft log index represented by the snapshot data
- membership config-change id: the raft entry index of the latest applied membership change

The imported snapshot path must not synthesize `Membership.ConfigChangeId` from the snapshot index. For online imports, it must preserve the source snapshot membership consistently and reject invalid target membership before restore.

## Upstream fix

Dragonboat has merged the fix in:

- https://github.com/matrixorigin/dragonboat/commit/103d119addcea5c7452a8b1ae96a1430eff233fc
- https://github.com/matrixorigin/dragonboat/pull/14

The upstream fix:

- preserves source snapshot membership for `NodeHost.SyncRequestImportSnapshot`
- keeps `tools.ImportSnapshot` disaster-recovery membership rewrite behavior, but preserves the original `ConfigChangeId`
- rejects imported snapshots whose target replica is missing, removed, role-mismatched, duplicated across roles, or mismatched with the local replica id
- adds regression coverage for the import metadata paths

## Required MatrixOne changes

Update MatrixOne's Dragonboat replacement dependency to include the upstream fix on both active branches:

- `3.0-dev`: https://github.com/matrixorigin/matrixone/pull/24967
- `main`: https://github.com/matrixorigin/matrixone/pull/24968

## Validation expectation

Each branch PR should at least validate:

- `git diff --check`
- Go module resolution for the updated Dragonboat replacement
- A focused LogService or dependency compile/test target where practical

## Related incident context

[...]

PR #24983, opened 2026-06-15
fix logservice stale replica truncation for main

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24966

## What this PR does / why we need it:

This PR updates the MatrixOne `3.0-dev` branch to use the Dragonboat snapshot import metadata fix merged in:

- https://github.com/matrixorigin/dragonboat/commit/103d119addcea5c7452a8b1ae96a1430eff233fc
- https://github.com/matrixorigin/dragonboat/pull/14

Main changes:

- Update the `github.com/lni/dragonboat/v4` replacement to `github.com/matrixorigin/dragonboat/v4 v4.0.0-20260612094223-103d119addce`.
- Add the corresponding `go.sum` entries for the updated Dragonboat module.

Why this is needed:

- The upstream fix prevents imported snapshots from synthesizing `Membership.ConfigChangeId` from the snapshot index.
- HAKeeper uses Dragonboat membership `ConfigChangeId` as the LogService shard epoch.
- Without this fix, an old imported membership can be reported with a large fake epoch and be treated as newer than the live membership view.
- The upstream fix also preserves online imported snapshot membership consistently and rejects invalid target membership before restore.

Validation:

- `git diff --check`
- `GOTELEMETRY=off go mod download github.com/lni/dragonboat/v4`
- `GOTELEMETRY=off go list -m -json github.com/lni/dragonboat/v4`
- `GOTELEMETRY=off GOCACHE=/private/tmp/mo-dragonboat-3dev-gocache go test github.com/lni/dragonboat/v4/tools github.com/lni/dragonboat/v4/internal/rsm github.com/lni/dragonboat/v4/internal/raft -run '^$' -count=0`

Fixes logservice stale replica recovery with the previous MO-side stale truncation metadata cleanup plus the Dragonboat snapshot import metadata fix from #24967. This keeps truncation from processing stale local replica metadata/exported snapshots, and updates Dragonboat so imported snapshots no longer synthesize membership ConfigChangeId from snapshot index.

PR #25050, opened 2026-06-18
fix logservice stale replica truncation for 4.0 (#24983)

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24966

Why this is needed:

- The upstream fix prevents imported snapshots from synthesizing `Membership.ConfigChangeId` from the snapshot index.
- HAKeeper uses Dragonboat membership `ConfigChangeId` as the LogService shard epoch.
- Without this fix, an old imported membership can be reported with a large fake epoch and be treated as newer than the live membership view.
- The upstream fix also preserves online imported snapshot membership consistently and rejects invalid target membership before restore.

Validation:

- `git diff --check`
- `GOTELEMETRY=off go mod download github.com/lni/dragonboat/v4`
- `GOTELEMETRY=off go list -m -json github.com/lni/dragonboat/v4`
- `GOTELEMETRY=off GOCACHE=/private/tmp/mo-dragonboat-3dev-gocache go test github.com/lni/dragonboat/v4/tools
github.com/lni/dragonboat/v4/internal/rsm
github.com/lni/dragonboat/v4/internal/raft -run '^$' -count=0`

Fixes logservice stale replica recovery with the previous MO-side stale truncation metadata cleanup plus the Dragonboat snapshot import metadata fix from #24967. This keeps truncation from processing stale local replica metadata/exported snapshots, and updates Dragonboat so imported snapshots no longer synthesize membership ConfigChangeId from snapshot index.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -264,7 +264,7 @@ replace github.com/hashicorp/memberlist => github.com/matrixorigin/memberlist v0
 replace (
 	github.com/elastic/gosigar v0.14.2 => github.com/matrixorigin/gosigar v0.14.3-0.20241204071856-40aab500bfac
 	github.com/fagongzi/goetty/v2 v2.0.3-0.20230628075727-26c9a2fd5fb8 => github.com/matrixorigin/goetty/v2 v2.0.0-20240611082008-a4de209fff3d
-	github.com/lni/dragonboat/v4 v4.0.0-20220815145555-6f622e8bcbef => github.com/matrixorigin/dragonboat/v4 v4.0.0-20251214113216-2ddf81ef2a85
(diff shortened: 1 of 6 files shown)


XuPeng-SH (approved):
Re-checked this from the stale-replica and unhappy-path angle. The MO-side fix looks coherent: stale local replica metadata is only cleaned after reconciling against live NodeHost state, stale exported snapshots are removed before metadata deletion, snapshot-index realignment avoids wedging the export path afterward, and the failure-path tests cover the important cleanup edge cases. I did not find a substantive correctness issue left to block on.

Issue #21863, opened 2025-05-19
[Feature Request]: complete and high-performance data branching and forking capabilities


XuPeng-SH:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

Complete and high-performance data branching and forking capabilities,  similar to Neon.

### Documentation, Adoption, Use Case, Migration Strategy

```Markdown

```

PR #25053, opened 2026-06-18
4.0-dev: fix(frontend): optimize no-pk data branch merge apply

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #21863

This PR optimizes online DATA BRANCH MERGE and PICK apply for no-primary-key tables by using MatrixOne fake primary key values for delete application.

Before this change, no-pk online merge applied delete rows by matching the full visible row. For update-heavy workloads this makes each logical update pay the full-row delete cost before inserting the new version, which scales poorly and causes no-pk merge to become much slower than the primary-key path.

The change splits online apply from portable SQL output:

- online MERGE and PICK use fake-pk-backed delete matching for no-pk tables
- diff exported as SQL remains portable and continues to emit visible-column full-row deletes
- insert generation only uses visible table columns, so hidden fake pk values are not inserted back into user tables
- LCA probing explicitly selects the table column layout instead of relying on lca.* column ordering

Local validation:

- CGO_CFLAGS="-I/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/include" go test ./pkg/frontend -run TestDataBranch
- make build
- Full git4data/branch BVT passed on the same data-branch code before clean cherry-pick: TOTAL 3558, SUCCESS 3514, FAILED 0, IGNORED 44

Local no-pk merge benchmark, 200k base rows:

- 1k changed rows: 1.089s baseline -> 0.305s optimized, 3.58x faster
- 10k changed rows: 17.322s baseline -> 2.454s optimized, 7.06x faster
- 50k changed rows: baseline was still running after more than 3.5 minutes and was stopped; optimized completed in 2.672s

### DIFF
--- a/pkg/frontend/data_branch.go
+++ b/pkg/frontend/data_branch.go
@@ -51,6 +51,21 @@ import (
 
 const dataBranchMetadataIDBatchSize = 512
 
+func isDataBranchUserVisibleColumn(col *plan.ColDef) bool {
+	if col == nil {
+		return false
+	}
+	if col.Hidden {
+		return false
+	}
+	switch col.Name {
+	case catalog.Row_ID, catalog.FakePrimaryKeyColName, catalog.CPrimaryKeyColName:
+		return false
+	default:
+		return true
+	}
+}
+
 func newBranchHashmapAllocator(limitRate float64) *branchHashmapAllocator {
 	throttler := rscthrottler.NewMemThrottler(
 		"DataBranchHashmap",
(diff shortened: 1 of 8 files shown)


XuPeng-SH (approved):
Re-checked the current head from the no-PK apply and unhappy-path angle. The online path now clearly separates fake-PK-backed delete matching from portable SQL output, keeps hidden fake-PK values out of user-table inserts, and the added unit/integration coverage is strong enough that I do not see a remaining correctness issue to block on.

Issue #25054, opened 2026-06-18
[Bug]: sqlquote escape string failed to escape \


cpegeric:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

d3aadea886b512e9451c3a67be27aebbc2218047

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

I re-checked the current head and found two remaining quoting/correctness issues that I do want fixed before approval.

pkg/common/sqlquote/sqlquote.go:48-63 is not actually safe for MatrixOne/MySQL string literals. The new EscapeString / String helper only doubles single quotes, but the MySQL scanner still treats backslash as an escape inside single-quoted literals (pkg/sql/parsers/dialect/mysql/scanner.go:471-537, with coverage in scanner_test.go:119-149). That means values containing \ are corrupted, and values containing \ can still break the literal shape. For example, sqlquote.String("C:\Program Files") produces a literal the scanner reads back as C:Program Files, and sqlquote.String("a\\b") is also parsed incorrectly. Please escape backslashes (and the other MySQL escape forms) before wrapping the literal, or stop interpolating these blobs as SQL string literals.

ALTER ... AUTO_UPDATE still interpolates raw algo_params and index names. In pkg/sql/compile/ddl.go:968-972, the AUTO_UPDATE branch still does fmt.Sprintf(updateMoIndexesAlgoParams, newAlgoParams, ..., alterIndex.IndexName) with raw values, while the REINDEX branch right below was updated to escape both fields (ddl.go:1055-1063). So an index name containing \ can still break the catalog update here, and any future string-valued param with \ will do the same. Please reuse the same quoting/escaping path in AUTO_UPDATE as well.

### Steps to Reproduce

```Markdown
sqlquote.String("a\\b") is also parsed incorrectly.
```

PR #25057, opened 2026-06-18
Fix escape string failed to escape '\' character

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25054

Fix escape string failed to escape '\' character


XuPeng-SH (changes requested):
This fixes the `AUTO_UPDATE` and `REINDEX` string-literal escaping holes, but one adjacent `ALTER INDEX` path is still left behind.\n\nIn `pkg/sql/compile/ddl.go:913-918`, the visibility update still does `fmt.Sprintf(updateMoIndexesVisibleFormat, ..., indexdef.IndexName)` with the raw index name. Backticked identifiers can contain both single quotes and backslashes, and the MySQL scanner still treats backslash as an escape inside single-quoted literals. So `ALTER TABLE ... ALTER INDEX` on a quoted name can still generate malformed SQL or break out of `name = ...`, exactly the same class of bug this PR is trying to close for the other DDL branches.\n\nPlease route the visibility update through the same `sqlquote.EscapeString` / `sqlquote.String` path as `AUTO_UPDATE` and `REINDEX`, and add one regression case for `ALTER INDEX ... VISIBLE/INVISIBLE` with an index name containing `` and `\\`.

XuPeng-SH (approved):
Re-checked the latest head. The previous visibility-path quoting hole is now fixed: `ALTER INDEX ... VISIBLE/INVISIBLE` escapes index names the same way as the AUTO_UPDATE and REINDEX branches, and the new regression tests cover the exact quote/backslash break-out cases that mattered here. I do not see a remaining correctness issue that should keep this blocked.

iamlinjunhong (approved):

ouyuanning (approved):

Issue #24954, opened 2026-06-12
[Feature Request]: Support protected databases (writable only by AccountAdmin)


LiSong0214:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

MOI (MatrixOne Intelligence) is a data platform built on top of MatrixOne, so some of MOI's system databases are stored in MO as well. We need these databases to be writable only by the AccountAdmin role, while all other roles should have read-only access. Otherwise, if a regular user drops or modifies these MOI system databases, MOI becomes completely unusable.

### Describe the feature you'd like

Support a setting or configuration that can designate one or more specific databases as protected, so that only the AccountAdmin role can create, delete, or modify their contents, while all other roles are read-only.

### Describe implementation you've considered

N/A

### Additional information

N/A

daviszhen:
使用方法见 : https://github.com/matrixorigin/matrixone/blob/b8eae0fd145133431b8b14f1c3ac6a0323f7b6e0/test/distributed/cases/tenant/privilege/protected_database.sql#L15

例如:
```sql
set global protected_databases = 'protected_bvt_db,protected_bvt_new,protected_bvt_clone,CamelDB';
```
然后用accountadmin / moadmin ,创建database.
理论上也可以先创建database xxx, 然后用`set global protected_databases = xxx ` 提供保护.

Ariznawlll:
测试通过

PR #25058, opened 2026-06-18
cherrypick: add protected database

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24954

增加 protected database

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -532,8 +532,6 @@ github.com/magiconair/properties v1.8.7 h1:IeQXZAiQcpL9mgcAe1Nu6cX9LLw6ExEHKjN0V
 github.com/magiconair/properties v1.8.7/go.mod h1:Dhd985XPs7jluiymwWYZ0G4Z61jb3vdS329zhj2hYo0=
 github.com/mailru/easyjson v0.7.7 h1:UGYAvKxe3sBsEDzO8ZeWOSlIQfWFlxbzLZe7hwFURr0=
 github.com/mailru/easyjson v0.7.7/go.mod h1:xzfreul335JAWq5oZzymOObrkdz5UnU4kGfJJLY9Nlc=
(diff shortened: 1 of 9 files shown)


XuPeng-SH (changes requested):
Protected-database enforcement is still bypassable through `CALL`, so the PR does not yet guarantee that protected DBs are safe from non-admin writes.

`CallStmt` marks the statement as writing directly, but does not populate `writeDatabaseTargets`, and the new protected-db guard returns early when that slice is empty. The procedure body then executes through a background session created without preserving `fromRealUser`, while the new helpers explicitly skip enforcement for non-real-user sessions. In practice, a non-admin can still modify a protected database by invoking a stored procedure whose body performs DDL/DML there.

Please preserve `fromRealUser` when executing procedure bodies and/or enforce protected-db checks on each statement executed inside procedures, then add a regression covering a non-admin `CALL` that writes into a protected database.

daviszhen:
> Protected-database enforcement is still bypassable through `CALL`, so the PR does not yet guarantee that protected DBs are safe from non-admin writes.
> 
> `CallStmt` marks the statement as writing directly, but does not populate `writeDatabaseTargets`, and the new protected-db guard returns early when that slice is empty. The procedure body then executes through a background session created without preserving `fromRealUser`, while the new helpers explicitly skip enforcement for non-real-user sessions. In practice, a non-admin can still modify a protected database by invoking a stored procedure whose body performs DDL/DML there.
> 
> Please preserve `fromRealUser` when executing procedure bodies and/or enforce protected-db checks on each statement executed inside procedures, then add a regression covering a non-admin `CALL` that writes into a protected database.

存储过程 是废的， 不能正常工作。 暂不考虑call的完整权限问题。

XuPeng-SH (approved):

heni02 (approved):

Issue #25046, opened 2026-06-18
[Bug]: BIT_COUNT function on arm results failed


heni02:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

629f58da65c

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

job:https://github.com/matrixorigin/ci-test/actions/runs/27733163070/job/82044403702
<img width="1559" height="723" alt="Image" src="https://github.com/user-attachments/assets/17afb3a2-76fe-4e21-b8a8-0259c2fc07d3" />

### Steps to Reproduce

```Markdown
bvt case
```

aptend:
## CI Run Analysis: ARM64 BVT Pessimistic (job 82044403702)

**Run**: https://github.com/matrixorigin/ci-test/actions/runs/27733163070
**Architecture**: ARM64 (linux/arm64, go1.25.4)
**Commit**: 629f58da65c

### Summary

Total: 94044 tests, 92911 passed, 6 failed, 1127 ignored — **99% pass rate**.
All 6 failures are ARM64-specific floating-point computation bugs, shared across BVT Pessimistic, BVT Optimistic, and Upgrade test jobs.

---

### Bug 1: `BIT_COUNT(negative DOUBLE)` returns 0 on ARM64 (3 failures)

**File**: `func_bit_count.test`
**Code**: `pkg/sql/plan/function/func_unary.go` → `bitCountFromFloat()` (line 608)

| SQL | Expected (x86_64) | Actual (ARM64) |
|---|---|---|
| `BIT_COUNT(CAST(-1.4 AS DOUBLE))` | 64 | **0** |
| `BIT_COUNT(CAST(-1.5 AS DOUBLE))` | 63 | **0** |
| `BIT_COUNT(CAST(-2.5 AS DOUBLE))` | 63 | **0** |
| `BIT_COUNT(CAST(-3.5 AS DOUBLE))` | 62 | **0** |
| `BIT_COUNT(CAST(-4.5 AS DOUBLE))` | 62 | **0** |
| `BIT_COUNT(CAST(-5.5 AS DOUBLE))` | 62 | **0** |
| `BIT_COUNT(CAST(-6.5 AS DOUBLE))` | 62 | **0** |
| `BIT_COUNT(CAST(-7.5 AS DOUBLE))` | 61 | **0** |
[...]

PR #25068, opened 2026-06-22
add issue tag and regex

heni02:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25046

add issue tag and regex


XuPeng-SH (changes requested):
I re-checked the current head. I do not see a problem with the new `@regex(...)` handling for the `VAR_POP` / `VAR_SAMP` overflow text, but I do think the `BIT_COUNT` part should stay blocked.

The new `@bvt:issue#25046` markers do not make an unstable test robust — they statically skip the exact negative-`DOUBLE` `BIT_COUNT` cases that expose the ARM64 bug. At this PR’s own branch point, that means CI goes green while the branch can still return wrong query answers.

Even worse, once the real implementation fix is in the branch, these skip tags become a coverage regression: the BVT suite no longer proves the bug is fixed and no longer protects against it coming back.

**Concrete suggestion**

1. Do not land the skip tags as the fix.
2. Backport the actual `bitCountFromFloat` implementation fix to the branch and keep these SQL assertions active.
3. Keep the `@regex(...)` changes for the aggregate overflow messages if needed, but remove the `@bvt:issue#25046` masking once the real fix is present.

So my recommendation here is still **request changes**: preserve the test signal, fix the function, then run these cases normally.

PR #25069, opened 2026-06-22
fix(frontend): optimize no-pk data branch merge apply on main

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #21863

This PR ports the no-primary-key online DATA BRANCH MERGE apply optimization to `main`.

Before this change, no-pk online merge applied delete rows by matching the full visible row. For update-heavy workloads this makes each logical update pay the full-row delete cost before inserting the new version, which scales poorly and causes no-pk merge to become much slower than the primary-key path.

The change splits online apply from portable SQL output:

- online MERGE uses fake-pk-backed delete matching for no-pk tables
- diff exported as SQL remains portable and continues to emit visible-column full-row deletes
- insert generation only uses visible table columns, so hidden fake pk values are not inserted back into user tables
- LCA probing explicitly selects the table column layout instead of relying on `lca.*` column ordering
- adds the shared SQL row-value extraction helper needed by this port on `main`, without pulling in the broader decimal metadata changes from #24894

Review follow-up included in this revision:

- route any generated data-branch temp-table SQL through `BackgroundExec`, including quoted identifiers and apply DML that references temp tables in subqueries
- derive fake-PK logical row keys from user-visible columns only, so hidden cluster-by columns cannot be read by default DIFF sorting
- keep `DATA BRANCH PICK` primary-key-only on `main`; the shared apply appender has a PK-only mode and this PR does not enable no-pk PICK

Local validation:

- `CGO_CFLAGS="-I/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/include" CGO_LDFLAGS="-L/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/lib -Wl,-rpath,/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/lib" go test ./pkg/frontend -run TestDataBranch -count=1`
- `CGO_CFLAGS="-I/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/include" CGO_LDFLAGS="-L/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/lib -Wl,-rpath,/Users/ghs-mo/MOWorkSpace/matrixone-1st/thirdparties/install/lib" go test ./pkg/tests/dml -run TestDataBranchDiffAsFile -count=1`

Notes:

- The local test commands use the existing thirdparties build artifacts from `/Users/ghs-mo/MOWorkSpace/matrixone-1st` because this port was prepared in a separate clean worktree.
- Linker warnings about duplicate rpaths were emitted, but both test commands passed.


XuPeng-SH (changes requested):
I re-checked the latest head and found one remaining high-severity issue in the temp-table execution routing.

`runSql()` is supposed to send data-branch temp-table DDL/DML through `BackgroundExec`, but the detector in `pkg/frontend/data_branch_helpers.go:49-68` still does not recognize the SQL that this PR now generates.

The root problem is that `dataBranchTempSQLNeedsBackExec()` only checks the *first table token* for `insert/delete/update`, while the new apply SQL often uses the temp table in a subquery (`delete from base ... in (select ... from __mo_diff_del_*)`, `insert into base ... select ... from __mo_diff_ins_*`). On top of that, the SQL builder now always emits quoted identifiers, and `firstSQLTableToken()` / the current token test does not treat a quoted temp table target as a match either.

I reproduced the current detector logic against the generated statement shapes and it returns `false` for all of these:
- `drop table if exists \`db1\`.\`__mo_diff_del_x\``
- `insert into \`db1\`.\`__mo_diff_del_x\` values (1)`
- `delete from \`db1\`.\`base\` where ... in (select ... from \`db1\`.\`__mo_diff_del_x\`)`
- `insert into \`db1\`.\`base\` (...) select ... from \`db1\`.\`__mo_diff_ins_x\``

So the temp-table statements still go down the internal SQL fast path in exactly the cases this PR was trying to route away from it.

**Concrete suggestions**

1. Make temp-table detection robust to quoted identifiers (backticks before `__mo_diff_*`).
2. For `insert/delete/update`, detect temp-table usage anywhere relevant in the statement, not just in the first target-table token.
3. The simplest/safest option may be to bypass the heuristic entirely for these generated apply statements and explicitly choose `BackgroundExec` when `applyBatchInfo` is present.
4. Add one regression test around `dataBranchTempSQLNeedsBackExec()` (or the higher-level `runSql` routing) that covers the exact quoted statement shapes generated by `initApplyTables()` / `flushSqlValues()`.

XuPeng-SH (changes requested):
I re-checked this PR from a correctness / unhappy-path angle. The direction is good, but I found two substantive issues that should be fixed before merge.

1. **`DATA BRANCH DIFF` can panic on no-PK tables with hidden columns**
   - In `getTableStuff`, fake-PK `pkColIdxes` still come from all non-rowid / non-fake-pk columns, so hidden columns like `__mo_cbkey_*` remain part of the sort key (`pkg/frontend/data_branch.go:934-939`).
   - But `visibleIdxes` now excludes those hidden columns (`pkg/frontend/data_branch.go:945-961`).
   - In the default DIFF output path, rows are sized/extracted from `visibleIdxes`, then later sorted by every `pkColIdx` (`pkg/frontend/data_branch_output.go:300-321`). On a no-PK table that has a hidden cluster-by key, the sort can read past the extracted row slice and panic.
   - **Suggested fix:** either derive fake-PK `pkColIdxes` from the same user-visible columns, or build the default DIFF row from the union of `visibleIdxes` and `pkColIdxes` before sorting.
   - **Please add a regression test** for `DATA BRANCH DIFF` on a no-PK clustered table.

2. **The new no-PK PICK path is still unreachable on `main`**
   - The PR adds fake-PK-aware apply logic in `pkg/frontend/data_branch_pick.go`, but `DATA BRANCH PICK` still rejects `fakeKind` tables before that logic ever runs (`pkg/frontend/data_branch.go:678-682`).
   - So on `main`, the new PICK-side code is dead/unreachable for no-PK tables.
   - **Suggested fix:** either wire no-PK PICK all the way through and add an end-to-end test, or remove the PICK-specific changes/claims from this PR so the scope matches the actual behavior on `main`.

I did not find another issue at the same severity level beyond these two, but both are real blockers: one is a reachable panic, and the other is scope/behavior mismatch.

XuPeng-SH (approved):
I re-checked the current head from a correctness / unhappy-path angle.

The previously-blocking issues I had raised around temp-table execution routing and the no-PK merge/diff/pick flow look resolved on the latest head. In particular, the temp-table detector/tests now cover the quoted/generated SQL shapes, and the fake-PK / user-visible-column handling in the current diff no longer shows the mismatch I was worried about before.

I do not see a substantive blocker in the current version, so I’m updating my review to **approve**.

gouhongshen:
@mergify requeue

gouhongshen:
@mergify requeue

gouhongshen:
@mergify requeue

Issue #24950, opened 2026-06-12
[Subtask]: Enforce CHECK constraints during REPLACE


ck89119:
## Parent Issue

Parent issue: #24918

## Detail of Subtask

Ensure `REPLACE` validates target-table `CHECK` constraints with the same semantics as an inserted row.

Reference case from the parent issue:

```sql
create table t_ck(
  id int primary key,
  v int check (v > 0)
);
insert into t_ck values (1,10);
replace into t_ck values (1,-1);
select * from t_ck order by id;
```

Expected MySQL-compatible behavior: the `REPLACE` fails because `v = -1` violates `check (v > 0)`, and the original row remains `(1, 10)`.

MatrixOne currently accepts the invalid row and stores `(1, -1)`.

## Describe implementation you've considered

Route `REPLACE` through the same row-level constraint validation that `INSERT` should use after defaults, generated columns, and replacement-row expressions have been computed. Add tests that verify failed `CHECK` validation preserves the old conflicting row.

## Additional information

Parent context: #24918, section 2.

Ariznawlll:
### motr 黑盒回归补充

已将该子任务加入 [motr PR #75](https://github.com/matrixorigin/motr/pull/75)：

- 用例：[issue_24950_replace_check_constraint.test](https://github.com/matrixorigin/motr/blob/b236f9d129bb2d6eeb6a83b7bf73968d7a6f58f2/suites/14_issue_regression/t/issue_24950_replace_check_constraint.test)
- Happy Path：合法冲突行替换、合法多行替换。
- Unhappy Path：冲突行、新行、混合有效/无效多行输入违反 CHECK 均返回 3819；每次失败后检查原数据不变，并验证后续合法语句仍可执行。
- MatrixOne `main@607a9263616ea883ac02c052529d2aa4c62ac051` 连续 3/3 通过（51/51 statements）。
- 修复前 `6c0d12dfc3c3c8cab7439beaa8efb00b8349a599` 复现失败：违反 CHECK 的 REPLACE 被错误接受。

该用例同时覆盖约束执行和语句原子性，已固化预期结果。

PR #25070, opened 2026-06-22
fix(plan): enforce CHECK constraints in REPLACE

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24950

- Persists parsed CHECK constraints into table definitions so they survive catalog round-trips.
- Adds row-level CHECK validation before INSERT/REPLACE writes, preserving SQL CHECK semantics where TRUE or NULL passes and FALSE fails.
- Ensures a failing REPLACE does not delete or overwrite the existing row, with planner tests and BVT coverage.

## Test Plan

- `go test ./pkg/vm/engine ./pkg/sql/plan -count=1`
- `./run.sh -p /Users/LoveYY/Develop/matrixorigin/matrixone/.worktrees/issue-24950-main/test/distributed/cases/dml/replace/replace.test -g -n`


gouhongshen (changes requested):
Blocking issues I found in the CHECK DDL semantics. The REPLACE injection point looks aligned with #24950, but the persisted constraints can already be semantically wrong before the DML planner sees them.

1. The parser/AST carries `ENFORCED` / `NOT ENFORCED`, but this PR drops that flag when building `plan.CheckDef`, then `appendCheckConstraintPlan` asserts every stored check. So `CHECK (...) NOT ENFORCED` will still reject `INSERT` / `REPLACE`, which is the opposite of MySQL-compatible behavior. Please either persist the enforcement flag and skip non-enforced checks, or reject unsupported `NOT ENFORCED` explicitly.

2. `appendCheckDef` binds expressions immediately with the columns currently present in `tableDef.Cols`. That changes CHECK scoping rules: table-level CHECKs may reference columns declared later, while column-level CHECKs may only reference the column being defined. With the current code, `CREATE TABLE t (CHECK (c1 <> c2), c1 INT, c2 INT)` is rejected, and `CREATE TABLE t (a INT, b INT CHECK (a > 0))` is accepted and then enforced. Please collect the CHECK ASTs with their kind/name/enforcement metadata, bind table checks after the full column list is known, and validate column checks against only their own column.

XuPeng-SH (changes requested):
I re-checked the latest head. I agree with the earlier CHECK-DDL concerns, and I also think there is still an upgrade/compatibility hole. I would keep this blocked.

1. **`NOT ENFORCED` is still silently turned into enforced behavior.**
   The parser/AST carries enforcement mode, but `appendCheckDef()` only persists `{Name, Check}` and drops any enforced/not-enforced state. Later `appendCheckConstraintPlan()` wraps every recovered check in `check_constraint_assert(...)`, so `CHECK (...) NOT ENFORCED` will still reject writes.

   **Suggestion:** either persist the enforcement flag in `plan.CheckDef` / metadata and skip runtime assertions for `NOT ENFORCED`, or reject unsupported `NOT ENFORCED` explicitly at DDL time instead of silently changing semantics.

2. **CHECK binding still changes scope rules.**
   `appendCheckDef()` binds immediately against the columns currently present in `tableDef.Cols`. That means:
   - table-level CHECKs can no longer reference columns declared later in the statement, and
   - column-level CHECKs can end up seeing other columns when they should be scoped only to their own column.

   **Suggestion:** collect the CHECK ASTs first with their kind/name/enforcement metadata, bind table-level checks only after the full column list is known, and validate column-level checks against just the defining column.

3. **Pre-existing CHECK tables still have no obvious readback path after upgrade.**
   The new runtime path only recovers checks from the synthetic `__mo_check_constraints` config. Readers like `disttae` and `memoryengine` only call `SplitCheckConstraintsFromConfigs()` on stream configs; there is no fallback for older tables that were created before this hidden config key existed. So existing CHECK tables can still come back with `TableDef.Checks == nil` after catalog round-trip/restart and keep bypassing enforcement.

   **Suggestion:** add a compatibility/backfill path for old metadata (for example, recover checks from the old persisted definition when the hidden key is absent, or migrate existing tables on upgrade).

Non-blocking but worthwhile test additions after the semantic fixes:
- a table-level / named CHECK end-to-end case,
- a multi-row atomicity case where one row passes and one row violates CHECK,
- and an upgrade/round-trip regression proving pre-existing CHECK tables still enforce after metadata reload.

heni02 (approved):

aunjgr (approved):
The REPLACE CHECK injection looks correct given how #24950 handles the plan side. The CHECK DDL concerns raised by others are valid but pre-existing — this PR only wires REPLACE into the same check path INSERT already uses.

XuPeng-SH (approved):
I reviewed the latest head from the CHECK-constraint / REPLACE unhappy-path angle and did not find a blocker.

The important ordering looks right: REPLACE builds the final replacement-row projection, injects the CHECK filter/assert before `LOCK_OP` and before `MULTI_UPDATE`, so a CHECK failure aborts before the old conflicting row can be deleted. INSERT and UPDATE are wired through the same row-level validation path, and the assert expression preserves SQL CHECK semantics by allowing TRUE or NULL and failing FALSE.

I also checked the catalog round-trip path. Persisting CHECK defs through the hidden stream config key is separated back out in `GetTableDef`, so normal table properties stay visible while CHECK metadata is restored into `TableDef.Checks`. The `PlanDefToCstrDef` signature changes have only the expected call sites.

One detail I specifically chased: this PR changes `colName2Idx` to mean final projection position so CHECK expressions address the written row, not table ordinals. I did not find a current blocker from that change; the INSERT index-table path that consumes the map uses final projection positions, and the REPLACE CHECK path runs on the final projection before writes.

Local checks:

```sh
go test ./pkg/vm/engine ./pkg/sql/plan -count=1

go test ./pkg/sql/plan/function -run 'TestCheckConstraintAssertReturnsConstraintViolation|Test_funids' -count=1

go test ./pkg/sql/plan -run 'TestReplacePlanChecksTableCheckConstraints|TestInsertPlanChecksTableCheckConstraints|TestUpdatePlanChecksTableCheckConstraints|TestBuildCreateTableStoresCheckConstraints|TestCheckConstraintRejectsSyntheticColumns|TestReplace' -count=1

go test ./pkg/vm/engine/memoryengine ./pkg/vm/engine/disttae/cache ./pkg/vm/engine/disttae -run '^$' -count=0
```

All passed. The PR is still behind `main`, so it will need an update/rebase before merge.

XuPeng-SH (changes requested):
I need to correct my previous approval after re-checking the active blocking reviews and the latest head. The REPLACE injection point itself is good, but the CHECK metadata that reaches that injection point can still be semantically wrong, so I do not think this is ready yet.

Blocker 1: `ENFORCED` / `NOT ENFORCED` is still dropped.

The parser AST already carries the enforcement mode: `AttributeCheckConstraint.Enforced` and `CheckIndex.Enforced` exist (`pkg/sql/parsers/tree/create.go`). But `buildTableDefs()` calls `appendCheckDef()` with only name + expression for both column and table checks, and `appendCheckDef()` persists only `{Name, Check}` into `plan.CheckDef`. Later `appendCheckConstraintPlan()` asserts every recovered check unconditionally.

That means an explicit `CHECK (...) NOT ENFORCED` is silently turned into an enforced runtime constraint. Please either persist the enforcement flag and skip runtime assertions for non-enforced checks, or reject unsupported `NOT ENFORCED` at DDL time instead of changing its meaning.

Relevant code:

- `pkg/sql/parsers/tree/create.go`: `AttributeCheckConstraint.Enforced`, `CheckIndex.Enforced`
- `pkg/sql/plan/build_ddl.go`: column/table CHECK calls to `appendCheckDef()` do not pass enforcement mode
- `pkg/sql/plan/build_constraint_util.go`: `appendCheckConstraintPlan()` wraps every stored check in `check_constraint_assert(...)`

Blocker 2: CHECK binding still uses the wrong scope/timing.

`appendCheckDef()` binds the expression immediately against the columns currently present in `tableDef.Cols`. That changes the expected scope rules:

- table-level CHECKs that appear before later column declarations cannot reference those later columns because they are not in `tableDef.Cols` yet;
- column-level CHECKs can see previously declared columns because the binder receives all columns accumulated so far, instead of only the column being defined.

Please collect the CHECK ASTs first with their kind/name/enforcement metadata, bind table-level checks after the full column list is known, and validate/bind column-level checks against only the defining column unless MatrixOne intentionally documents a different rule and adds tests for it.

The runtime ordering for REPLACE still looks right: once the stored CHECK metadata is correct, the filter/assert is placed before `LOCK_OP` and `MULTI_UPDATE`, so a failing REPLACE should not delete the old conflicting row.

Local checks I ran before re-checking the DDL blockers:

```sh
go test ./pkg/vm/engine ./pkg/sql/plan -count=1

go test ./pkg/sql/plan/function -run 'TestCheckConstraintAssertReturnsConstraintViolation|Test_funids' -count=1

go test ./pkg/sql/plan -run 'TestReplacePlanChecksTableCheckConstraints|TestInsertPlanChecksTableCheckConstraints|TestUpdatePlanChecksTableCheckConstraints|TestBuildCreateTableStoresCheckConstraints|TestCheckConstraintRejectsSyntheticColumns|TestReplace' -count=1

go test ./pkg/vm/engine/memoryengine ./pkg/vm/engine/disttae/cache ./pkg/vm/engine/disttae -run '^$' -count=0
```

Those pass, but they do not cover the two DDL semantics cases above.

iamlinjunhong (approved):

XuPeng-SH (changes requested):

1. The legacy INSERT fallback still bypasses CHECK enforcement entirely.

The new CHECK enforcement only exists in the modern insert path, but `build.go` still explicitly falls back to `buildInsert()` for unsupported modern cases, including the no-PK/no-unique-key ODKU corner. The legacy `buildInsert()` path never consults `TableDef.Checks`, so invalid rows can still be inserted through that fallback.

Suggested fix: either add CHECK enforcement to the legacy `buildInsert()` path, or block this fallback whenever `tableDef.Checks` is non-empty and keep such statements on the modern path.

2. Column-level CHECKs are incorrectly restricted to the defining column.

`appendCheckDef()` first binds column-level CHECKs against only the defining column, so a definition like `b INT CHECK (a > 0)` is rejected even though the planner/executor can already handle row-level CHECK expressions over multiple columns. That makes the new CHECK support materially incomplete and diverges from normal SQL/MySQL DDL semantics.

Suggested fix: remove the single-column pre-bind for column-level CHECKs and bind them the same way as table-level CHECKs, while preserving the intended column-position handling.

The parser/catalog/runtime coverage is otherwise decent, but these are real feature-semantics issues rather than just missing tests.

XuPeng-SH (changes requested):
Reviewed the latest head `82e63c1b`. The earlier blockers around DML placement, full-row binding, legacy INSERT fallback, catalog persistence, and ALTER/deep-copy propagation are now closed. Two latest-head correctness blockers remain:

1. **[P1] Preserve an explicit legacy `NOT ENFORCED` clause during recovery** (`pkg/sql/plan/build_ddl.go:1923-1928`). Ordinary tables persist `ctx.GetRootSql()` verbatim, so a pre-upgrade `CREATE TABLE t (a INT, CHECK (a > 0) NOT ENFORCED)` is distinguishable. The current parser also reports `EnforcementSet=true, Enforced=false`, but `recoveredCheck` drops those fields and this loop unconditionally calls `appendCheckDef`. On the first post-upgrade resolve, the constraint becomes runtime-enforced and INSERT/REPLACE/UPDATE start rejecting rows that the schema explicitly declared not enforced. This is contrary to the documented semantics that `NOT ENFORCED` creates but does not enforce the constraint: https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html. Please retain the parsed enforcement state and avoid creating an enforced `CheckDef` for an explicit `NOT ENFORCED` clause. Any ambiguity from formatter-produced legacy SQL should be handled separately; it does not apply to the verbatim ordinary-table path.

2. **[P1] Do not infer CHECK anonymity solely from the `__mo_chk_` prefix** (`pkg/sql/plan/build_show_util.go:472-476`). CREATE explicitly accepts a user-named `CONSTRAINT __mo_chk_1` (and `TestBuildCreateTableCheckConstraintNameDeduplication` asserts that it remains that name), but SHOW CREATE treats every such prefix as generated and emits only `CHECK (...)`. A local round-trip reproduced `CREATE TABLE t (a INT, CONSTRAINT __mo_chk_1 CHECK (a > 0))` as `CREATE TABLE t (..., CHECK (a > 0))`, losing the user-visible name in dumps/restores and changing later DROP/error identity. Please persist an explicit anonymous/generated marker, or reserve and reject the internal prefix, instead of using a prefix heuristic.

Validation on this head: focused CHECK parser/engine/planner/runtime tests pass; affected frontend/compile/disttae/memoryengine packages compile. I also added temporary diagnostic tests for the two paths above, captured the behavior, and removed them so the worktree remains clean.

gouhongshen (changes requested):
## Codex automated review

Loaded the PR reviews, issue comments, and inline threads. The latest head fixes the generated-name issue, but two blocking CHECK/ALTER compatibility paths remain. Focused Go tests could not run because this environment has no `go` executable.

### P1 - Block CHANGE COLUMN on a CHECK-dependent column (pkg/sql/plan/build_alter_change_column.go:68)
CHANGE COLUMN only protects generated-column dependencies, unlike MODIFY, RENAME, and DROP which now call checkColumnWithCheckDependency. Thus `ALTER TABLE t CHANGE COLUMN a a2 INT` is planned even when `CHECK (a > 0)` exists. The copied TableDef retains OriginSql referencing `a`, then buildAlterTableCopy serializes that stale constraint into CreateTmpTableSql and the executor runs it; the copy-table DDL references a column that no longer exists. The operation should reject CHECK-dependent CHANGE COLUMN operations (or update/rebind the constraint) before mutating the copied definition.

### P1 - Do not discard legacy NOT ENFORCED constraints during recovery (pkg/sql/plan/build_ddl.go:1934)
Recovery silently skips every explicitly `NOT ENFORCED` CHECK, so the resolved TableDef no longer represents that part of an upgraded table's schema. SHOW CREATE renders solely from TableDef.Checks, and ALTER TABLE COPY builds its replacement table from the same structure; consequently a pre-upgrade table containing both an enforced check and `CHECK (...) NOT ENFORCED` will omit the latter from SHOW CREATE and permanently drop it on a subsequent ALTER. Preserve its metadata separately from runtime assertions so it remains non-enforced while surviving readback and copy operations.

gouhongshen (changes requested):
## Codex automated review

Two blocking correctness/compatibility gaps remain. Existing CI passes, but focused local tests could not run because `go` is unavailable in this environment.

### P1 - Enforce CHECK constraints in the legacy UPDATE fallback (pkg/sql/plan/bind_update.go:1142)
This validates CHECKs only in the modern UPDATE planner. A multi-table UPDATE returns `ErrUnsupportedDML` from `DMLContext` and falls back to `buildTableUpdate`; its `buildUpdatePlans` path goes directly from `appendPreInsertNode` to `appendSinkNode` without calling either CHECK helper. Consequently, for example, a multi-table UPDATE can write a row violating a persisted CHECK. CHECK validation must also be applied to the final row image in the fallback path, preferably through a shared insertion point.

### P1 - Do not reinterpret an existing table property as CHECK metadata (pkg/vm/engine/types.go:174)
The new decoder treats every property named `__mo_check_constraints` as encoded internal metadata, but CREATE TABLE PROPERTIES previously accepted arbitrary string keys and this PR does not reserve or reject that key. After upgrade, a previously valid table created with `PROPERTIES ('__mo_check_constraints'='value')` makes decoding fail; disttae then returns a nil table definition, rendering the table unavailable. The metadata needs an unambiguous representation that cannot collide with existing user properties (and new DDL should prevent any reserved-key collision).

XuPeng-SH (changes requested):
I found two blockers before this can merge.

1. `pkg/sql/plan/build_show_util.go` now renders CHECK clauses only from `tableDef.Checks`, but the CDC / publication / `get_ddl` paths still call `ConstructCreateTableSQL(nil, rel.CopyTableDef(...), ...)`. For tables created before `__mo_check_constraints` existed, `CopyTableDef()` comes back with empty `Checks`, and those callers do not have a `CompilerContext` to run `RecoverCheckConstraintsFromCreateSql`. That means upgraded old tables will silently lose CHECK clauses in generated DDL (CDC sink create SQL, publication DDL, snapshot/get_ddl output), which is a regression from the old `extractTopLevelCheckDefs(tableDef)` fallback. Please either keep the old fallback when `Checks` is empty, or recover checks before those nil-context callers build CREATE TABLE text.

2. `pkg/vm/engine/disttae/txn_table.go`, `pkg/vm/engine/disttae/cache/catalog.go`, and `pkg/vm/engine/memoryengine/table.go` now return `nil` from `GetTableDef` if `SplitCheckConstraintsFromConfigs` fails on the hidden `__mo_check_constraints` property. That turns one bad hidden config payload into a whole-table outage, because many callers dereference `GetTableDef()` directly or expect `CopyTableDef()` to succeed. Please log the decode failure and drop the bad CHECK metadata instead of returning `nil` for the entire table definition.

Non-blocking follow-ups I would still recommend:
- `RecoverCheckConstraintsFromCreateSql` currently re-parses / re-binds legacy `Createsql` on every `Resolve`, because it runs on a cloned `TableDef`; caching the recovered checks once at the engine tableDef level would avoid repeated parse cost on upgraded tables.
- Add end-to-end coverage for old-table SHOW CREATE / CDC / publication generation, plus a couple more unhappy-path tests around `INSERT ... SELECT` and `CREATE TABLE ... CHECK ... NOT ENFORCED`.

gouhongshen (changes requested):
## Codex automated review

Two P1 compatibility defects remain. `git diff --check` passes; focused Go tests could not run because `go` is unavailable in this environment.

### P1 - Honor UPDATE IGNORE for CHECK violations (pkg/sql/plan/bind_update.go:1142)
`stmt.Ignore` is captured at lines 81-84, but the newly added CHECK pass always calls `appendCheckConstraintPlan(..., false)`. That inserts `check_constraint_assert`, so `UPDATE IGNORE t SET a = -1` aborts instead of filtering out the offending row. The legacy UPDATE fallback does the same at `build_dml_util.go:269`. MySQL specifies that CHECK violations in `UPDATE IGNORE` produce a warning and skip that row; this PR already implements the equivalent filtering behavior for `INSERT IGNORE`. See [MySQL CHECK constraints](https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html). Pass the ignore mode through both UPDATE paths and add an UPDATE IGNORE regression test.

### P1 - Do not decode untagged legacy properties as CHECK metadata (pkg/vm/engine/types.go:182)
The new prefix makes newly written metadata unambiguous, but this decoder still accepts any unprefixed base64 value that unmarshals to a nonempty `TableDef.Checks`. `origin/main` had no CHECK-metadata key/payload, while it accepted arbitrary `PROPERTIES` keys; those properties are also copied into `StreamConfigsDef` by `PlanDefToCstrDef`. Therefore an existing user property named `__mo_check_constraints` whose value happens to be a valid unprefixed payload (including the exact prefix-trimmed value constructed in `types_test.go`) is silently converted into active CHECK constraints after upgrade. That can make previously valid writes fail or alter table behavior. Rejecting the key for new CREATE TABLE statements does not repair existing tables. Treat only tagged values as internal metadata, or migrate legacy payloads through an unambiguous versioned path.

gouhongshen (changes requested):
## Codex automated review

Three blocking MySQL-compatibility paths remain. Focused Go tests could not run because `go` is unavailable in this environment.

### P1 - Honor UPDATE IGNORE for CHECK violations (pkg/sql/plan/bind_update.go:1142)
`stmt.Ignore` is recognized, but this call passes `false`, which makes the new CHECK path emit `check_constraint_assert` and abort the entire statement. Thus `UPDATE IGNORE t SET a = -1` errors instead of skipping that row. The legacy UPDATE helper also hard-codes non-ignore behavior. MySQL requires CHECK violations under `UPDATE IGNORE` to produce a warning and skip the offending row: [MySQL CHECK constraints](https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html). Thread the ignore mode through both planners and add modern/fallback coverage.

### P1 - Do not reinterpret pre-upgrade user properties as CHECK metadata (pkg/vm/engine/types.go:182)
`origin/main` allowed arbitrary `PROPERTIES` keys and values, including `__mo_check_constraints`. This decoder treats an untagged value under that key as internal metadata whenever it happens to decode to nonempty `TableDef.Checks` (explicitly covered by `TestSplitCheckConstraintsFromConfigsReadsLegacyUnprefixedPayload`), so an existing user property can become active enforced constraints after upgrade. Tagged values are also silently removed. Rejecting the key for new DDL does not repair existing tables. Only tagged/versioned metadata should be decoded, with any necessary migration kept unambiguous.

### P1 - Reject cross-column references in column CHECK constraints (pkg/sql/plan/build_ddl.go:1847)
The implementation discards whether a pending CHECK was column-level and binds every expression against the full table. It therefore accepts and enforces `b INT CHECK (a > 0)`; the added BVT explicitly codifies that behavior. MySQL permits multi-column and forward references only for table CHECKs; a column CHECK may reference only its defining column: [MySQL CHECK constraints](https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html). Preserve the defining-column metadata and validate column CHECKs in that scope.

gouhongshen (changes requested):
## Codex automated review

Two issues remain: a legacy metadata compatibility regression and inconsistent column-level CHECK validation. Focused parser tests passed; CGo-dependent plan/engine tests were blocked by missing repository CGo artifacts.

### P1 - Do not reinterpret legacy user properties as CHECK metadata (pkg/vm/engine/types.go:186)
Before this PR, `__mo_check_constraints` was an ordinary user property. The new decoder treats any existing property with that key and the `mo_check_constraints_v1:` prefix as internal metadata, decoding it into active CHECK constraints or silently dropping it on decode failure. Such a value can be a valid pre-existing user property, so upgrades can unexpectedly add enforcement or lose user data; rejecting the key for new CREATE statements does not repair existing catalogs.

### P2 - Validate ADD COLUMN CHECK scope like CREATE TABLE (pkg/sql/plan/build_alter_add_column.go:73)
This path binds every column attribute CHECK through `appendCheckDef` against the full table, so `ALTER TABLE t ADD COLUMN b INT CHECK (a > 0)` is accepted. The CREATE path explicitly rejects the equivalent column-level cross-column reference, and MySQL column CHECK constraints may reference only their defining column ([documentation](https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html)). The ALTER path should apply the same scope validation.

gouhongshen (changes requested):
## Codex automated review

The REPLACE validation itself is correctly ordered before locks and writes, and prior metadata-collision/ALTER findings are fixed. Two blocking compatibility paths remain. Focused tests could not run because `go` is unavailable; current Ubuntu UT/BVT checks pass, but Upgrade CI is skipped.

### P1 - Gate the new constraint encoding for rolling upgrades (pkg/vm/engine/types.go:630)
This writes a new tag value 5 into the shared catalog constraint bytes. The origin/main decoder recognizes only tags 0–4 and has no unknown-record framing: it skips byte 5, interprets the first zero of the CHECK count as an Index tag, derives a bogus count, and eventually reads past the buffer. Since old CNs call UnmarshalBinary when resolving the same catalog table, creating a CHECK table from an upgraded CN can panic or make that table unavailable on a still-running old CN. MatrixOne supports mixed-version rolling upgrades while services remain available ([upgrade documentation](https://docs.matrixorigin.cn/en/v26.3.0.12/MatrixOne/Deploy/update-MatrixOne-cluster/)); persist this record only after a version gate, or introduce an encoding old decoders can safely skip.

### P1 - Preserve LOAD DATA IGNORE semantics in the legacy planner (pkg/sql/plan/build_dml_util.go:155)
This legacy insertion point always calls appendCheckConstraintPlanFromLastNode, which hard-codes ignoreMode=false. LOAD on any table with a foreign key is deliberately rejected by the modern resolver and falls back through buildLoad, so `LOAD DATA ... IGNORE` into a table having both an FK and a CHECK emits check_constraint_assert and aborts instead of skipping only the violating row. MySQL explicitly requires CHECK failures under LOAD DATA IGNORE to warn and skip the offending row ([CHECK constraints](https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html)). Thread the LOAD duplicate-handling mode through buildLoad/buildInsertPlans and add fallback-path coverage.

gouhongshen (changes requested):
## Codex automated review

Temporary-table CHECK name validation regresses valid same-name temporary tables.

### P1 - Allow a temporary table to share CHECK names with the ordinary table it hides (pkg/sql/compile/ddl.go:1503)
`CREATE TEMPORARY TABLE t (a INT CHECK (a > 0))` is planned with the logical generated name `t_chk_1`. Before this new validation runs, line 1407 changes only `qry.TableDef.Name` to the session-scoped physical name; the CHECK name remains `t_chk_1`. The scan at line 1503 therefore excludes the physical name, examines the ordinary `t`, extracts its anonymous CHECK as `t_chk_1`, and rejects the temporary CREATE as a duplicate. Temporary `t` is allowed to hide ordinary `t` with the same CHECK names, so this newly added executor-side recheck blocks a valid statement. Exempt the logical same-table temporary case (or otherwise keep the validation namespace consistent with the physical CHECK names).

gouhongshen (changes requested):
## Codex automated review

A temporary-table CHECK-name compatibility regression remains.

### P1 - Allow a temporary table to reuse CHECK names from the base table it hides (pkg/sql/compile/ddl.go:1503)
For `CREATE TEMPORARY TABLE t (a INT CHECK (a > 0))` when ordinary `t` exists, CreateTable has already replaced `qry.TableDef.Name` with the session-scoped physical name, but leaves the generated CHECK name as `t_chk_1`. The new validation excludes only that physical name, scans ordinary `t`, recovers its `t_chk_1`, and rejects the valid temporary CREATE as a duplicate. The same scan also conflicts with another session's physical temporary table. MySQL explicitly allows a temporary table that hides an ordinary table of the same name to share its CHECK names: [CHECK constraints documentation](https://dev.mysql.com/doc/refman/8.1/en/create-table-check-constraints.html). Keep the logical temporary-table namespace when excluding/checking names, and add an executor/BVT regression.

gouhongshen (changes requested):
## Codex automated review

Two blocking regressions remain: transaction-local CHECK metadata is omitted, and legacy recovery can replace valid schemas with enforced CHECK(false).

### P1 - Return transaction-local CHECK metadata from GetTableDef (pkg/vm/engine/disttae/txn_table.go:1593)
`checks` is declared but never populated from `tbl.extraInfo.Checks`, while CREATE/ALTER stores CHECKs there. `createWithID` eagerly builds this TableDef before registering the transaction-local relation, so CTAS and subsequent DML in the same transaction resolve a table with no Checks; `appendCheckConstraintPlan` therefore skips enforcement. For example, `CREATE TABLE t (a INT CHECK (a > 0)) AS SELECT -1` can insert the invalid row before commit. Populate this field from `extraInfo.Checks` with the same deep-copy semantics used by `pkg/vm/engine/disttae/cache/catalog.go`.

### P1 - Do not replace valid legacy CHECKs with enforced CHECK(false) (pkg/sql/plan/build_ddl.go:2196)
For every markerless legacy CREATE SQL containing any double quote, `||`, or backslash, recovery calls `setFailedCheckRecovery`, which installs an enforced literal false constraint at line 2431. The mode check is over the entire CREATE SQL rather than the CHECK expression, so valid pre-existing tables with an otherwise ordinary CHECK plus a quoted default/string escape are also affected. Since the SQL-mode marker is only added to new CHECK-bearing DDL, such legacy tables become unable to INSERT/REPLACE/UPDATE and their schema is rendered with the synthetic recovery-failed constraint. Preserve the original metadata or reject the unsupported recovery explicitly instead of fabricating an enforced false constraint.

Issue #24952, opened 2026-06-12
[Subtask]: Require DELETE privilege when REPLACE removes conflicting rows


ck89119:
## Parent Issue

Parent issue: #24918

## Detail of Subtask

Require `DELETE` privilege when `REPLACE` deletes an existing conflicting row.

MySQL requires `DELETE` privilege for the delete part of `REPLACE`. MatrixOne currently allows a replacement with only `INSERT` and `SELECT` privileges when an existing row conflicts.

Reference scenario from the parent issue:

```sql
create table t_priv(id int primary key, v int);
insert into t_priv values(1,10);

grant select on table replace_priv_edge.t_priv to r_replace_only_insert;
grant insert on table replace_priv_edge.t_priv to r_replace_only_insert;

-- as replace_only_insert
replace into t_priv values(1,20);
```

Expected MySQL-compatible behavior: fail with a DELETE privilege error when a conflicting existing row must be removed.

## Describe implementation you've considered

Check `DELETE` privilege for `REPLACE` statements that may remove conflicting rows. The implementation should distinguish insert-only cases from duplicate-key replacement cases where possible, while keeping behavior secure for plans that cannot prove insert-only execution ahead of time.

## Additional information

Parent context: #24918, comment section 5.

Ariznawlll:
Verified on latest `origin/main` at `200bfa7e0efb` (`fix: improve skills SKILL.md compatibility format (#25271)`).

Validation setup:

```sql
create table t_priv(id int primary key, v int);
insert into t_priv values(1,10);

create role replace_priv_r_24952;
grant connect on account * to replace_priv_r_24952;
grant select on table replace_priv_edge_24952.t_priv to replace_priv_r_24952;
grant insert on table replace_priv_edge_24952.t_priv to replace_priv_r_24952;
create user replace_priv_u_24952 identified by '111' default role replace_priv_r_24952;
```

As `replace_priv_u_24952`, with only `SELECT` + `INSERT` table privileges and no `DELETE` privilege:

```sql
use replace_priv_edge_24952;
select id, v from t_priv order by id;
replace into t_priv values(1,20);
select id, v from t_priv order by id;
```

Result:

```text
before REPLACE: (1, 10)
REPLACE result: ERROR 20101 (HY000): internal error: do not have privilege to execute the statement
after REPLACE: (1, 10)
```

[...]

PR #25071, opened 2026-06-22
fix(frontend): require DELETE privilege for REPLACE conflicts

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24952

This PR makes REPLACE authentication require DELETE privilege when the REPLACE plan may remove conflicting rows from the target table.

The plan-derived privilege tips still provide INSERT privileges for REPLACE, and this patch augments them with DELETE tips for real primary-key targets and fake-primary-key tables with unique indexes. Fake-primary-key tables without unique indexes remain insert-only.

Tests added:

- frontend unit coverage for real PK, fake PK without unique index, and fake PK with unique index
- BVT coverage for PK conflicts, unique conflicts, insert-only REPLACE, and success after DELETE is granted

Validation:

- `git diff --check`
- `go test ./pkg/frontend -run TestReplacePrivilegeRequiresDeleteForConflictingTargets -count=1`


XuPeng-SH (changes requested):
I re-checked the latest head and still see one remaining correctness issue in the new REPLACE privilege gate.

`pkg/frontend/authenticate.go:6939-6981` now adds `DELETE` privilege whenever the target table shape can produce a conflicting-row delete, but the logic is still purely schema-based. In particular, `replaceTargetIsInsertOnly()` only exempts fake-PK tables with **no** unique indexes at all.

That means a fake-PK table with a **nullable unique key** now always requires `DELETE` privilege, even for `REPLACE` rows whose unique-key value is `NULL` and therefore cannot conflict with any existing row. Those statements behave like plain inserts, so this patch rejects a real insert-only unhappy path.

There is already product evidence that nullable unique keys do not conflict under REPLACE: the existing SQL behavior keeps both rows after repeated `REPLACE` statements with `a = NULL` in `test/distributed/cases/dml/replace/replace.result`. But the new auth logic keys only off table shape, not whether the current input can actually drive a delete.

**Concrete suggestions**

1. Only add the extra `DELETE` privilege when the built REPLACE plan can actually delete an existing row for the current input, rather than for every fake-PK table that merely has a unique index in schema.
2. Add one unit/BVT regression for the exact case:
   - fake-PK table with a nullable unique index
   - role has `SELECT` + `INSERT` but not `DELETE`
   - `REPLACE ... VALUES (..., NULL)` should still succeed
3. Keep the existing conflicting non-NULL unique-key case alongside it, so the privilege split between “insert-only” and “delete-capable” remains pinned.

heni02 (approved):

aunjgr (approved):
LGTM. Correctly adds DELETE privilege requirement for REPLACE when rows may be deleted due to conflicts. Good test coverage in authenticate_test.go + BVT.

aunjgr (approved):
LGTM

XuPeng-SH (changes requested):
I re-checked the latest head. The new nullable-unique-key fix improves the single-column / all-parts-NULL case, but there is still one remaining correctness issue for composite unique keys.

Right now `skipUniqueIdx` only stays true when **every** part of a unique index is statically NULL. That is too narrow for SQL semantics. For a composite `UNIQUE(a,b)`, a row like `(NULL, 1, ...)` still cannot conflict with an existing row, because the composite unique key serializes to NULL as soon as any input part is NULL. But the current fake-PK REPLACE privilege gate still treats that table as delete-capable and requires `DELETE`.

You can see the mismatch in the current head:
- `bind_replace.go` clears `skipUniqueIdx[i]` as soon as it finds any non-NULL part
- fake-PK conflict matching still uses `=` across unique-key parts
- the serial composite unique-key path returns NULL if any input value is NULL
- `replaceTargetIsInsertOnly(...)` still falls through to requiring DELETE whenever the table has a unique index and the static-null shortcut does not fire

So `REPLACE ... VALUES (NULL, 1, ...)` on a fake-PK table with `UNIQUE(a,b)` is still an insert-only path that the frontend is over-rejecting.

**Concrete suggestions**

1. Treat a unique index as insert-only when **any nullable unique-key part** is statically NULL, not only when all parts are NULL.
2. Add one unit test and one BVT case for a composite unique key with one NULL component under an INSERT-only role.
3. Keep the existing conflicting non-NULL unique-key case beside it so the privilege split remains pinned.

XuPeng-SH (approved):
I reviewed this from the REPLACE privilege / unhappy-path angle and did not find a blocker in the current head.

The implementation is conservative in the right direction: real-PK REPLACE still requires DELETE; fake-PK/no-active-unique-index remains insert-only; nullable unique indexes are only skipped when the planner can prove the incoming unique key part is statically NULL for all rows. Mixed or dynamic cases keep the conflict path and therefore still require DELETE.

I also checked the planner/frontend alignment: the planner's `skipUniqueIdx` cases produce static-NULL delete columns, and the frontend only skips adding the DELETE privilege tip when those delete columns can be traced back to static NULL. Cases it cannot prove, such as `INSERT ... SELECT`, ordinary functions, or mixed NULL/non-NULL VALUES, remain conservative.

Local checks:

```sh
go test ./pkg/frontend -run 'TestReplacePrivilegeRequiresDeleteForConflictingTargets|TestExtractPrivilegeTipsFromPlanSkipsInvalidMultiUpdateCtx|TestReplaceStaticNullHelpers|TestReplaceTargetIsInsertOnlyGuards' -count=1

go test ./pkg/sql/plan -run 'TestBindReplaceSkipsUniqueIndexForStaticNull|TestBindReplaceSkipsCompositeUniqueIndexForAnyStaticNull|TestBindReplaceWithCompositeUniqueIndex' -count=1

go test ./pkg/sql/plan -run 'TestReplace' -count=1
```

PR #25076, opened 2026-06-22
fix bit_count on arm

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25046

原因:
```
BIT_COUNT(negative DOUBLE) comment 是对的
我用最小程序复现了 /mnt/fastdata/temp2/matrixone/pkg/sql/plan/function/func_unary.go:605 的逻辑：

rounded := math.RoundToEven(val)
uint64(rounded)

结果：

amd64/go1.25.x：

-1.4 -> rounded=-1 -> uint64=0xffffffffffffffff -> bits=64
arm64/go1.25.4：

-1.4 -> rounded=-1 -> uint64=0 -> bits=0
所以 comment 里说的 ARM64 上“负浮点转 uint64 变成 0”我已经实测确认了。
这说明 bitCountFromFloat() 的实现确实有跨架构问题，不应该依赖这个转换行为。
```
修改方案: 先转in64, 再转uint64. 已经在arm64上验证了.

### DIFF
--- a/pkg/sql/plan/function/func_bit_count_test.go
+++ b/pkg/sql/plan/function/func_bit_count_test.go
@@ -78,12 +78,14 @@ func TestBitCountFloat(t *testing.T) {
 			NewFunctionTestInput(types.T_float64.ToType(), []float64{
 				1.4, 1.5, 2.5, 3.5,
 				-1.4, -1.5, -2.5, -3.5,
+				-9223372036854774784.0, -9223372036854775808.0,
 				math.Inf(1), math.Inf(-1),
 			}, nil),
 		},
 		NewFunctionTestResult(types.T_uint64.ToType(), false, []uint64{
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
Re-checked the current head. The fix is narrowly scoped and correct for the actual cross-arch bug: negative FLOAT/DOUBLE values are now rounded first and then converted through int64 before the uint64 bit-count path, which avoids the ARM-specific zeroing behavior while preserving the existing MinInt64 clamp. The added regression cases hit the exact problematic boundary values, and I do not see a remaining correctness issue to block on.

ouyuanning (approved):

Issue #15494, opened 2024-04-12
[Bug]: User in sys account cannot manage other accounts with ACCOUNT privs granted


aylei:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

main, 1.1-dev

### Commit ID

any

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

```
mysql> show grants;
+---------------------------------------------------------+
| Grants for mocadmin@localhost                           |
+---------------------------------------------------------+
| GRANT connect ON account  `mocadmin`@`localhost`        |
| GRANT alter user ON account  `mocadmin`@`localhost`     |
| GRANT account all ON account  `mocadmin`@`localhost`    |
| GRANT database all ON database * `mocadmin`@`localhost` |
| GRANT table all ON table *.* `mocadmin`@`localhost`     |
+---------------------------------------------------------+
5 rows in set (0.06 sec)

mysql> CREATE ACCOUNT IF NOT EXISTS `test1` ADMIN_NAME admin IDENTIFIED BY '111';
ERROR 20101 (HY000): internal error: do not have privilege to execute the statement
```

### Expected Behavior

The user should be able to manage other accounts.

### Steps to Reproduce

```Markdown
Run SQL listed above
```

heni02:
confirm，closed
bvt 已覆盖，main 和 4.1 都包含修复

PR #25078, opened 2026-06-22
fix: allow account all to manage sys accounts

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #15494

Allow roles in the sys account with `account all on account *` to execute account-level DDL such as `CREATE ACCOUNT`, `ALTER ACCOUNT`, `DROP ACCOUNT`, and `UPGRADE ACCOUNT`.

This keeps the sys-scope account DDL coverage restricted to the sys tenant, so non-sys tenants cannot use `account all` to manage accounts. It also keeps `account ownership` separate from `account all`.

Validation:
- `go test ./pkg/frontend`
- `make build`
- Started a local single-node `mo-service` and verified a non-moadmin sys user with only `account all on account *` can create, alter, and drop an account.


heni02 (approved):

XuPeng-SH (changes requested):
I did not find a new high-confidence logic bug in the auth change itself, but I still think the regression coverage is below the bar for what this PR claims to enable.\n\nThe code path is widening sys-tenant `account all on account *` access for **CREATE ACCOUNT / ALTER ACCOUNT / DROP ACCOUNT / UPGRADE ACCOUNT**, but the new positive automation only proves part of that surface:\n- unit tests cover positive **create / alter / upgrade**\n- the distributed SQL case only exercises positive **alter account**\n- `DROP ACCOUNT` only has the new **negative** non-sys test\n\nSo there is still no automated positive proof that a non-moadmin sys role with only `account all` can successfully `DROP ACCOUNT`, and no end-to-end proof for `CREATE ACCOUNT` / `UPGRADE ACCOUNT` under the new privilege shape either. For an authorization change, I would like those paths pinned explicitly rather than inferred from nearby statements.\n\n**Concrete suggestions**\n\n1. Add a positive unit test for `DropAccount` mirroring the new sys-tenant `CreateAccount` / `AlterAccount` / `UpgradeAccount` cases.\n2. Extend the `zz_accesscontrol` SQL coverage so the same non-moadmin sys role with only `account all on account *` actually succeeds at:\n   - `CREATE ACCOUNT`\n   - `DROP ACCOUNT`\n   - `UPGRADE ACCOUNT`\n   not just `ALTER ACCOUNT`.\n3. Keep the existing negative non-sys cases alongside those positive sys cases, so the widened permission is proven to stay sys-only.\n\nOnce those user-visible paths are pinned, I am happy to re-check.

Ariznawlll:
Addressed the coverage requests in `8005875e`:\n\n- added a positive unit test for sys-tenant `DROP ACCOUNT` via `account all on account *`\n- extended `zz_accesscontrol/alter_account.sql` so a non-moadmin sys role with only `account all on account *` succeeds at `CREATE ACCOUNT`, `DROP ACCOUNT`, and `UPGRADE ACCOUNT`, in addition to the existing `ALTER ACCOUNT` checks\n- kept the non-sys negative unit coverage unchanged\n\nLocal verification:\n- `go test ./pkg/frontend -count=1`\n- `./run.sh -n -g -p /home/mo/matrixone/test/distributed/cases/zz_accesscontrol/alter_account.sql`

daviszhen (approved):
• Blocking Findings            

  未发现 blocking 问题。PR 的核心路径看起来能解决 #15494：                                                                                      

  - CREATE/DROP/ALTER/UPGRADE ACCOUNT 仍是 sys-scope privilege。                                                                                                                                                                                                                                 
  - 查询权限时只在 sys tenant 下让 account all on account * 覆盖这些 sys-scope account DDL：pkg/frontend/authenticate.go:7087。
  - 非 sys tenant 下 account all 不覆盖 create/drop/alter/upgrade account，UT 已覆盖负例。                                                                                                                                                                                                       
  - 执行阶段 CREATE ACCOUNT / ALTER ACCOUNT 从 sys + moadmin 放宽为 sys tenant，仍挡住普通 tenant：pkg/frontend/authenticate.go:3263、pkg/frontend/authenticate.go:9397。                                                                                                                        
  - UPGRADE ACCOUNT 已从 special admin path 转成普通 account privilege path：pkg/frontend/authenticate.go:6088、pkg/frontend/authenticate.go:9311。                                                                                                                                              
                                                                        
  Non-Blocking Notes

  1. alter_account.sql 里 alter account test admin_name='adminuser' IDENTIFIED BY ... 现在结果是 there is no user adminuser，这说明已经过了权限检查，但不是一个成功 alter case。后面的 alter account test comment / suspend 成功，基本能证明 alter 权限放开了。可以接受。
  2. grant-option helper 仍保留旧 covering 逻辑：pkg/frontend/authenticate.go:8408 到 8415 仍让 account all/account ownership 覆盖 create/drop/alter/upgrade account 的 WGO 查询。当前这些 sys-scope privilege 本身不可 grant，所以不是 blocking；但语义和新执行权限路径不完全一致，建议后续补一
     个小 UT 锁住“不能 grant create/drop/alter/upgrade account”的边界。

  3. Copilot 之前提到的 “sys account all can drop account 缺正向测试” 已经补了：pkg/frontend/authenticate_test.go:835。

  总体结论：这个 PR 的修改方向和覆盖基本合理，可以解决 issue #15494；没有看到必须 request changes 的点。

aptend:
The optimistic BVT failures were caused by the new test sequence, not by the privilege check itself. Both failed jobs hit the same statement:\n\n`test/distributed/cases/zz_accesscontrol/alter_account.sql` row 159: `drop account account_all_upgrade;`\n\nActual result was `w-w conflict` after `upgrade account 'account_all_upgrade'` immediately followed by dropping the same account under optimistic mode.\n\nUpdated in `7ca620627`:\n- keep the account-all positive `DROP ACCOUNT` coverage on `account_all_created`\n- keep the account-all positive `UPGRADE ACCOUNT` execution\n- move cleanup of `account_all_upgrade` back to sys session\n- add `select mo_ctl('cn','syncCommit','');` before cleanup to avoid optimistic async-commit conflict\n- use `drop account if exists` for cleanup

aptend:
Local validation with freshly rebuilt `mo-service` now passes for the updated access-control BVT in both txn modes.\n\nSetup:\n- built current PR head with `make build`\n- generated temporary launch configs with explicit `[cn.txn] mode = "Pessimistic"/"Optimistic"` and matching `[tn.txn] mode`\n- used fresh `mo-data` for each run\n\nVerified:\n- Pessimistic: `./run.sh -n -g -p /home/mo/matrixone/test/distributed/cases/zz_accesscontrol/alter_account.sql` -> 111/111 passed\n- Optimistic: same command -> 111/111 passed\n- `go test ./pkg/bootstrap -count=1` passed\n\nAlso fixed a real issue found during local validation: manual `upgrade account` should no-op when the tenant create version already equals the current CN version, matching the automatic tenant upgrade path. Without this, a fresh local cluster could hit `BUG: missing version upgrade`.

XuPeng-SH (changes requested):
I re-checked the latest head. The earlier coverage gaps are improved, but I still see two substantive issues in the new account-upgrade path.

1. **`ACCOUNT ALL` now authorizes cluster-wide bootstrap upgrade via `UPGRADE ACCOUNT ALL`.**
   The new sys-tenant privilege path treats `account all on account *` as sufficient for account-level sys privileges, and the new tests explicitly assert that this is enough for `UpgradeStatement`. But `handleExecUpgrade()` passes `st.Target.IsALLAccount` straight through, and `UpgradeTenant(..., isALLAccount=true)` dispatches to `BootstrapUpgrade(ctx)`.

   That is not just per-account management anymore — it is full cluster/bootstrap upgrade authority. So this PR widens privilege from account management to cluster-wide upgrade.

   **Suggestion:** keep `UPGRADE ACCOUNT ALL` on the old admin-only path, or split single-account upgrade privilege from cluster/bootstrap upgrade privilege instead of letting generic `ACCOUNT ALL` cover both.

2. **Manual `upgrade account` can no longer reapply same-version tenant upgrades.**
   `custom_upgrade.go` now returns early whenever `currentCN.Version == version`, before reaching the handler loop. But the handler loop itself still intentionally allows same-version upgrades via `versions.Compare(v.Metadata().Version, from) >= 0`, and the async tenant upgrader also skips only `>` rather than `>=`. That means version-offset-only reapplication / same-version repair paths can now be silently skipped in the manual path.

   **Suggestion:** do not short-circuit solely on version-string equality; preserve manual execution for same-version offset upgrades, or make the skip logic offset-aware before returning.

Once those semantics are tightened, I’m happy to re-check.

aptend:
Addressed the latest review in `b11213459`:\n\n- `UPGRADE ACCOUNT ALL` is back on the special moadmin-only path. Single-account `UPGRADE ACCOUNT '<name>'` still uses `PrivilegeTypeUpgradeAccount`, so sys-tenant roles with `account all on account *` can only cover per-account upgrade.\n- Manual same-version tenant upgrade is now offset/entry aware: it no-ops for current-version tenants when there are no matching upgrade entries, but still runs tenant handlers when the current final-version offset has replayable same-version upgrade entries.\n- Added regression coverage for both behaviors.\n\nLocal validation:\n- `go test ./pkg/frontend ./pkg/bootstrap -count=1`\n- `make build`\n- optimistic single BVT: `./run.sh -n -g -p /home/mo/matrixone/test/distributed/cases/zz_accesscontrol/alter_account.sql` -> 112/112 passed

XuPeng-SH (approved):
Re-checked the latest head. The previous blockers are fixed: `ACCOUNT ALL` no longer grants `UPGRADE ACCOUNT ALL` / bootstrap-upgrade authority, and the manual account-upgrade path no longer short-circuits same-version upgrade handling in the way I flagged before. I do not see a remaining correctness issue that should keep this blocked.

LeftHandCold (approved):

Issue #25079, opened 2026-06-22
[Bug]: protected_databases does not honor lower_case_table_names normalization


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

3.0-dev

### Commit ID

82a1e04b0

### Other Environment Information

```Markdown
- MatrixOne version: 8.0.30-MatrixOne-v3.0
- lower_case_table_names default: 1
- Found while verifying protected database coverage for #24954
```

cc @daviszhen

### Actual Behavior

`protected_databases` appears to store configured database names using the original case, while database creation/name resolution follows `lower_case_table_names=1` and normalizes names to lowercase.

As a result, configuring a mixed-case protected database name does not protect the actual normalized database name.

Observed in 3.0-dev:

```sql
select @@global.lower_case_table_names;
-- 1

set global protected_databases = 'CamelDB';
select @@global.protected_databases;
-- CamelDB

-- login as a non-accountadmin user with create/drop database privileges
create database `CamelDB`;
-- succeeds, but it should be protected if protected_databases follows MO name normalization

create database `cameldb`;
-- database cameldb already exists
```

The second create shows that `CamelDB` and `cameldb` refer to the same normalized database name under the current MO setting, but the protected database check did not block `create database `CamelDB``.

### Expected Behavior

`protected_databases` should honor the same database-name normalization rules used by MO.

When `lower_case_table_names = 1`, configuring:

```sql
set global protected_databases = 'CamelDB';
```

should protect the normalized database name `cameldb`, so a non-accountadmin user should not be able to run either of these:

```sql
create database `CamelDB`;
create database `cameldb`;
```

Both should return the protected database privilege error:

```text
internal error: do not have privilege to execute the statement
```

### Steps to Reproduce

```sql
-- system tenant or normal tenant accountadmin
set global enable_privilege_cache = off;
set global protected_databases = 'CamelDB';

create role protected_case_writer;
grant create database, drop database on account * to protected_case_writer;
create user protected_case_user identified by '111' default role protected_case_writer;

-- login as protected_case_user with protected_case_writer
select @@global.lower_case_table_names;
create database `CamelDB`;
create database `cameldb`;
```

### Additional information

Code inspection on 3.0-dev shows `protectedDatabaseSetFromString` keeps names as-is, and its unit test currently expects mixed-case values to be preserved:

```go
[...]

Issue #25080, opened 2026-06-22
[Bug]: protected database read-only semantics do not cover USE and SHOW TABLES


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

3.0-dev

### Commit ID

82a1e04b0

### Other Environment Information

```Markdown
- MatrixOne version: 8.0.30-MatrixOne-v3.0
- Found while verifying protected database coverage for #24954
```

cc @daviszhen

### Actual Behavior

For a non-accountadmin user, a protected database currently allows fully-qualified table reads, but rejects `USE protected_db` and `SHOW TABLES`.

Observed in 3.0-dev:

```sql
-- protected_bvt_db is configured in @@global.protected_databases
-- login as a normal tenant user with high privileges on protected_bvt_db

use protected_bvt_db;
-- internal error: do not have privilege to execute the statement

show tables;
-- internal error: do not have privilege to execute the statement

select a from protected_bvt_db.t1;
-- succeeds
```

So the current behavior is:

```text
SELECT protected_db.table: allowed
USE protected_db: rejected
SHOW TABLES for protected_db: rejected
```

This is not a general MO privilege-system bug. In normal MO RBAC, `USE db` and `SHOW TABLES` may require their own privileges. The question is whether protected database's read-only guarantee should override/extend that behavior for non-accountadmin users.

### Expected Behavior

Protected database was described as writable only by AccountAdmin and read-only for all other roles.

If "read-only" is intended to include metadata read operations, then non-accountadmin users should be able to browse/read protected database metadata while still being blocked from writes. For example:

```sql
use protected_bvt_db;
show tables;
select a from t1;
```

should succeed, while DDL/DML such as `create table`, `insert`, `drop table`, and `drop database` should still fail.

If the intended semantics are narrower and only fully-qualified `SELECT protected_db.table` is allowed, then the current behavior is expected, but the protected database read-only scope should be clarified.

### Steps to Reproduce

```sql
-- system tenant
set global enable_privilege_cache = off;
drop account if exists protected_bvt_acc;
create account protected_bvt_acc ADMIN_NAME 'admin' IDENTIFIED BY '111';

-- login as protected_bvt_acc:admin
set global protected_databases = 'protected_bvt_db';
create database protected_bvt_db;
create table protected_bvt_db.t1(a int primary key, b int);
insert into protected_bvt_db.t1 values(1, 1);

create role protected_bvt_writer;
[...]

PR #25082, opened 2026-06-22
fix lower case dbname

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25079

https://github.com/matrixorigin/matrixone/issues/25080

调整为按 lower_case_table_names的变量值, 解析


XuPeng-SH (changes requested):
I re-checked the latest head and still see two correctness issues.

1. **`protected_databases` parsing is now polluted by the current session database.**
   `protectedDatabaseSetFromString()` now reuses `normalizeProtectedDatabaseName()`, and that helper substitutes `""` with `ses.GetDatabaseName()`. That fallback makes sense for statement targets, but it is wrong when parsing the `protected_databases` sysvar itself.

   As a result, inputs like `','` or `'db1,'` no longer mean “empty element”; they silently add the caller’s current database. This also interacts badly with the existing `SetGlobalSysVar()` guard that checks `len(protectedDatabaseSetFromString(...)) == 0` to reject direct clears: `SET GLOBAL protected_databases=','` can now bypass that guard whenever the session currently has a database selected.

   **Suggestion:** parse `protected_databases` entries with trim/lowercasing only; do not apply the “empty means current database” fallback when reading the sysvar.

2. **`lower_case_table_names=2` is still treated as case-sensitive here.**
   `protectedDatabaseNamesAreLowerCased()` currently lowercases only when `lower_case_table_names == 1`. But the sysvar allows `0..2`, and the identifier-comparison helper in `tree.NewCStr()` lowercases for any nonzero value. So in mode `2`, mixed-case protected DB names will not be normalized here, and case-variant writes can slip past protection.

   **Suggestion:** mirror the engine’s actual identifier-comparison semantics: treat any nonzero `lower_case_table_names` value as case-insensitive, and add a test for mode `2`.

Non-blocking but worthwhile follow-up coverage after the semantic fixes:
- one representative non-DB write path under `lower_case_table_names=0`, and
- one explicit test matrix row for `lower_case_table_names=2`.

heni02 (approved):

XuPeng-SH (changes requested):
I re-checked this PR from a first-principles / unhappy-path angle. I think there is one blocker here.

### Blocker: unqualified writes to the current protected DB can bypass the guard

This patch moves `normalizeProtectedDatabaseName` toward lower-case normalization only, but it also removes the old runtime fallback from an empty target DB name to `ses.GetDatabaseName()`. That is unsafe on the write-check path, because many write statements still collect only `SchemaName` from the AST, which is empty for unqualified objects:

- `CREATE TABLE t ...`
- `DROP TABLE t`
- `CREATE VIEW v`
- `DROP VIEW v`
- `DROP SEQUENCE s`
- `CREATE TABLE ... AS SELECT ...`
- etc. after `USE protected_db`

In the patched code:
- `normalizeProtectedDatabaseName` returns `""` for an empty runtime target
- `checkProtectedDatabaseWriteWithSet` skips empty db names entirely
- but `determinePrivilegeSetOfStatement` still records only `SchemaName` for these DDL paths, which is empty when the schema qualifier is omitted

That means a non-admin with normal DB privileges can still write inside a protected database simply by omitting the schema qualifier after `USE protected_db`.

### Suggested fix
Split the two concerns:
- **config parsing** for `protected_databases` should normalize names for lower-case behavior
- **runtime write-target resolution** still needs to resolve an empty target DB to `ses.GetDatabaseName()` before comparing against the protected-db set

Please also add regressions for at least:
- `USE protected_db; CREATE TABLE t (...)`
- `USE protected_db; DROP TABLE t`
- `USE protected_db; CREATE TABLE t AS SELECT ...`

I think this should be **changes requested** before merge.

daviszhen:
> I re-checked this PR from a first-principles / unhappy-path angle. I think there is one blocker here.
> 
> ### Blocker: unqualified writes to the current protected DB can bypass the guard
> This patch moves `normalizeProtectedDatabaseName` toward lower-case normalization only, but it also removes the old runtime fallback from an empty target DB name to `ses.GetDatabaseName()`. That is unsafe on the write-check path, because many write statements still collect only `SchemaName` from the AST, which is empty for unqualified objects:
> 
> * `CREATE TABLE t ...`
> * `DROP TABLE t`
> * `CREATE VIEW v`
> * `DROP VIEW v`
> * `DROP SEQUENCE s`
> * `CREATE TABLE ... AS SELECT ...`
> * etc. after `USE protected_db`
> 
> In the patched code:
> 
> * `normalizeProtectedDatabaseName` returns `""` for an empty runtime target
> * `checkProtectedDatabaseWriteWithSet` skips empty db names entirely
> * but `determinePrivilegeSetOfStatement` still records only `SchemaName` for these DDL paths, which is empty when the schema qualifier is omitted
> 
> That means a non-admin with normal DB privileges can still write inside a protected database simply by omitting the schema qualifier after `USE protected_db`.
> 
> ### Suggested fix
> Split the two concerns:
> 
> * **config parsing** for `protected_databases` should normalize names for lower-case behavior
> * **runtime write-target resolution** still needs to resolve an empty target DB to `ses.GetDatabaseName()` before comparing against the protected-db set
> 
> Please also add regressions for at least:
> 
> * `USE protected_db; CREATE TABLE t (...)`
> * `USE protected_db; DROP TABLE t`
> * `USE protected_db; CREATE TABLE t AS SELECT ...`
> 
> I think this should be **changes requested** before merge.

XuPeng-SH (approved):

The previous blocker around unqualified writes to the current protected database looks resolved now:
- config parsing and runtime target resolution are split apart,
- the runtime write path again resolves empty target DB names through the current database,
- and there is now focused regression coverage for unqualified `CREATE TABLE` / `DROP TABLE` / `CTAS` under `USE protected_db`.

I do not see a remaining substantive blocker in the current version, so I’m updating my review to **approve**.

Issue #24947, opened 2026-06-12
[Subtask]: Support MySQL REPLACE syntax variants


ck89119:
## Parent Issue

Parent issue: #24918

## Detail of Subtask

Support MySQL-compatible `REPLACE` syntax variants that MatrixOne currently rejects:

- `REPLACE INTO dst TABLE src`
- `REPLACE LOW_PRIORITY INTO ...`
- `REPLACE DELAYED INTO ...`

Reference examples from the parent issue:

```sql
replace into t_table_dst table t_table_src;
replace low_priority into t_mod values(1,20);
replace delayed into t_mod values(1,30);
```

MySQL accepts these forms. `DELAYED` is converted to plain `REPLACE` with a warning in MySQL 8.0. MatrixOne currently rejects them in the parser.

## Describe implementation you've considered

Extend the parser and AST/planner coverage for the accepted MySQL forms. `LOW_PRIORITY` can be parsed and ignored if MatrixOne has no equivalent scheduling behavior. `DELAYED` should follow the project's compatibility policy for deprecated MySQL syntax, ideally accepting it with a warning if warning infrastructure is available.

## Additional information

Parent context: #24918, comment section 7.

MySQL reference: https://dev.mysql.com/doc/refman/8.4/en/replace.html

Ariznawlll:
Verified on latest `origin/main` at `0e8602b78`.

Validation environment:

```text
MatrixOne: git_version() = 0e8602b78
MySQL oracle: 8.0.45
```

Tested the three syntax variants from this issue:

```sql
replace into t_table_dst table t_table_src;
replace low_priority into t_mod values(1,20);
replace delayed into t_mod values(1,30);
```

Result:

1. `REPLACE INTO dst TABLE src`
   - MySQL: accepted and copied/replaced rows from `src` into `dst`.
   - MO latest main: accepted, final table result matched MySQL.

2. `REPLACE LOW_PRIORITY INTO ...`
   - MySQL: accepted.
   - MO latest main: accepted, final table result matched MySQL.

3. `REPLACE DELAYED INTO ...`
   - MySQL: accepted, converts it to plain `REPLACE`, and reports warning `3005`: `REPLACE DELAYED is no longer supported. The statement was converted to REPLACE.`
   - MO latest main: statement is accepted and data is replaced successfully, but `SHOW WARNINGS` returns a parser-style error instead of the MySQL deprecated-syntax warning:

```text
[...]

Issue #24918, opened 2026-06-10
[Bug] REPLACE has multiple MySQL compatibility gaps


Ariznawlll:
## Summary

While testing `REPLACE` compatibility against MySQL, I found three behavior gaps in MatrixOne:

1. `REPLACE ... SET col = col + 1` does not treat the RHS column reference as `DEFAULT(col)`.
2. `CHECK` constraints are not enforced during `REPLACE`.
3. Foreign key checks/actions are not applied with MySQL delete+insert semantics during `REPLACE`.

Tested with:

- MatrixOne: `main@6c0d12dfc`, `select version()` returns `8.0.30-MatrixOne-v`
- MySQL comparison: `8.0.45`

According to MySQL docs, `REPLACE` works like `INSERT`, except that duplicate `PRIMARY KEY` or `UNIQUE` rows are deleted before inserting the new row. Missing columns use default values, and in `REPLACE ... SET col = col + 1`, the RHS `col` is treated as `DEFAULT(col)`, not the current row value.

MySQL reference: https://dev.mysql.com/doc/refman/8.4/en/replace.html

### Reproduce

```sql
drop database if exists replace_expr_cmp;
create database replace_expr_cmp;
use replace_expr_cmp;

create table t_set(
  id int primary key,
  v int default 5,
  note varchar(20) default 'd'
);

insert into t_set values (1,10,'old');
replace into t_set set id = 1, v = v + 1;
select * from t_set;
```

### Expected, same as MySQL

```text
id | v | note
1  | 6 | d
```

`v + 1` should be evaluated as `DEFAULT(v) + 1`, so the value should be `5 + 1 = 6`. Columns not provided by the `REPLACE` statement should use defaults.

### Actual in MatrixOne

```text
ERROR 20301 (HY000): invalid input: ambiguous column reference 'v'
```

The old row remains unchanged.

### Reproduce

```sql
drop database if exists replace_check_cmp;
create database replace_check_cmp;
use replace_check_cmp;

create table t_ck(
  id int primary key,
  v int check (v > 0)
);

insert into t_ck values (1,10);
replace into t_ck values (1,-1);
select * from t_ck order by id;
```

### Expected, same as MySQL

The `REPLACE` should fail because `v = -1` violates `check (v > 0)`, and the original row should remain:

```text
id | v
1  | 10
```

MySQL reports:

```text
ERROR 3819 (HY000): Check constraint 't_ck_chk_1' is violated.
```

### Actual in MatrixOne

[...]

PR #25083, opened 2026-06-22
feat(parser): support MySQL REPLACE syntax variants

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24947

Accept the MySQL-compatible `REPLACE` forms that MatrixOne previously rejected at the parser:

- `REPLACE INTO dst TABLE src` — equivalent to `REPLACE INTO dst SELECT * FROM src` (MySQL 8.0.19+)
- `REPLACE LOW_PRIORITY INTO ...`
- `REPLACE DELAYED INTO ...`

### Implementation notes

- The priority modifiers (`LOW_PRIORITY` / `DELAYED`, and also `HIGH_PRIORITY`) are parsed and ignored, since MatrixOne has no corresponding scheduling behavior. The resulting statement is identical to a plain `REPLACE` — same execution path and semantics. This matches MySQL 8.0, where `DELAYED` is downgraded to a plain `REPLACE` (MatrixOne simply does not emit the deprecation warning).
- The `TABLE src` source form is rewritten in the grammar into a `SELECT *` over the source table, reusing the existing `REPLACE ... SELECT` execution path. No AST struct or planner changes are required.
- This is a parser-only change. `mysql_sql.go` is regenerated from `mysql_sql.y` via `make`; the large generated diff is the goyacc state-table renumbering caused by the two new grammar rules (no new shift/reduce conflicts).

### Tests

- Parser round-trip unit tests in `mysql_sql_test.go` for the `TABLE` form and all three priority modifiers.
- End-to-end BVT cases in `test/distributed/cases/dml/replace/` (verified 201/201, 100%).

Parent issue: #24918


heni02 (approved):

XuPeng-SH (changes requested):
I reviewed this from the MySQL-compatibility and parser unhappy-path angle. The basic examples work, but I still see blockers before this should merge.

Blocker 1: `REPLACE ... (cols) TABLE src` is missing.

The new `replace_data: TABLE table_name` rule covers only the no-column-list form. MySQL's REPLACE syntax allows an optional target column list before both `SELECT` and `TABLE`, i.e. `[(col_name [, col_name] ...)] { SELECT ... | TABLE table_name }`.

This PR already supports `REPLACE INTO dst (a, b) SELECT ...` through the existing `(' insert_column_list ')' select_stmt` rule, but it does not add the equivalent `(' insert_column_list ')' TABLE table_name` branch. So a valid MySQL-compatible form such as:

```sql
REPLACE INTO dst (id, v) TABLE src;
```

will still fail to parse. Please add the column-list TABLE variant and a parser/BVT regression for it.

Reference: MySQL 8.4 REPLACE syntax documents `[(col_name [, col_name] ...)] {SELECT ... | TABLE table_name}`: https://dev.mysql.com/doc/refman/8.4/en/replace.html

Blocker 2: the grammar now accepts `REPLACE HIGH_PRIORITY`, which is not part of the MySQL REPLACE syntax.

`replace_stmt` now uses the shared `priority_opt`, whose `priority` includes `LOW_PRIORITY`, `HIGH_PRIORITY`, and `DELAYED`. For REPLACE, MySQL documents only `LOW_PRIORITY | DELAYED`. The issue also asked for `LOW_PRIORITY` and `DELAYED`, not `HIGH_PRIORITY`.

Accepting extra syntax is still a compatibility problem: MatrixOne would accept SQL that MySQL rejects while the PR describes the behavior as MySQL-compatible. Please either use a REPLACE-specific priority rule limited to `LOW_PRIORITY | DELAYED`, or explicitly document and justify `HIGH_PRIORITY` as a MatrixOne extension instead of MySQL compatibility.

Blocker 3: current head cannot merge into `main`.

GitHub reports `mergeStateStatus=DIRTY`. I confirmed locally with a non-mutating merge check:

```sh
```

It reports a content conflict in the generated parser file:

```text
CONFLICT (content): Merge conflict in pkg/sql/parsers/dialect/mysql/mysql_sql.go
```

Please rebase/merge current `main` and regenerate the parser so the checked-in `mysql_sql.go` matches `mysql_sql.y` after conflict resolution.

Local checks:

```sh
go test ./pkg/sql/parsers/dialect/mysql -run 'TestDebug|TestValid' -count=1
go test ./pkg/sql/parsers -run '^$' -count=0
```

The tests/diff check passed, but they do not cover the missing column-list TABLE form, and the merge check fails as above.

XuPeng-SH (changes requested):
I found one blocker here.

This parser only accepts the bare `TABLE table_name` form (with or without a target column list), but MySQL `TABLE` syntax also supports `ORDER BY` and `LIMIT`, and `REPLACE ... TABLE ...` inherits that source form. So valid MySQL statements such as `REPLACE INTO dst TABLE src ORDER BY id LIMIT 1` will still fail to parse.

The current implementation rewrites directly to `tree.NewSelect(makeSelectStarFromTable(...), nil, nil)`, so there is nowhere to preserve `ORDER BY` / `LIMIT` even if we want to. That makes the new compatibility support incomplete.

Suggested fix:
1. parse a reusable `TABLE` source form with optional `ORDER BY` / `LIMIT`;
2. thread those clauses into the rewritten `tree.NewSelect(...)`;
3. add parser + BVT coverage for `REPLACE ... TABLE ... ORDER BY ... LIMIT ...`.

I re-checked the rest of the patch and did not find another substantive code issue beyond this gap.

XuPeng-SH (approved):
I re-reviewed the current head from MySQL compatibility, parser-generation consistency, mergeability, and unhappy-path/resource-risk angles.

The previous blockers are addressed:
- `REPLACE ... (cols) TABLE src` is now supported and covered.
- `REPLACE ... TABLE src ORDER BY ... LIMIT/OFFSET ...` is now preserved through the TABLE-to-SELECT rewrite and covered.
- `REPLACE HIGH_PRIORITY` is rejected via the REPLACE-specific priority rule and an invalid parser test.
- The checked-in generated parser is consistent with `mysql_sql.y`.

Local checks I ran:
```sh
make -B -C pkg/sql/parsers/dialect/mysql mysql_sql.go
git diff --exit-code -- pkg/sql/parsers/dialect/mysql/mysql_sql.go pkg/sql/parsers/dialect/mysql/mysql_sql.y
go test ./pkg/sql/parsers/dialect/mysql -run 'TestValid|TestInvalid|TestDebug' -count=1\ngo test ./pkg/sql/parsers/dialect/mysql -count=1\ngo test ./pkg/sql/parsers -count=1\ngit merge-tree --write-tree origin/main HEAD\n```\n\nNo further code issue found. Non-blocking: the PR description still appears to contain stale wording from the earlier `HIGH_PRIORITY` iteration, so it would be clearer to update that before merge.

iamlinjunhong (approved):

Issue #24945, opened 2026-06-12
[Subtask]: REPLACE SET should evaluate RHS column references as DEFAULT(column)


ck89119:
## Parent Issue

Parent issue: #24918

## Detail of Subtask

Implement MySQL-compatible `REPLACE ... SET` expression binding for column references on the right-hand side.

For `REPLACE ... SET col = col + expr`, MySQL treats the RHS `col` as `DEFAULT(col)`, not as the current conflicting row value. Missing columns should also be filled from their defaults, matching normal `REPLACE` insert semantics.

Reference case from the parent issue:

```sql
create table t_set(
  id int primary key,
  v int default 5,
  note varchar(20) default 'd'
);
insert into t_set values (1,10,'old');
replace into t_set set id = 1, v = v + 1;
select * from t_set;
```

Expected result:

```text
id | v | note
1  | 6 | d
```

MatrixOne currently reports `invalid input: ambiguous column reference 'v'` and leaves the old row unchanged.

## Describe implementation you've considered

Bind RHS column references in `REPLACE ... SET` against the inserted/default row context, equivalent to `DEFAULT(column)` for references to target columns. Add focused planner/binder coverage and BVT coverage for provided and omitted columns.

## Additional information

Parent context: #24918, section 1.

MySQL reference: https://dev.mysql.com/doc/refman/8.4/en/replace.html

Ariznawlll:
已在最新 `main` 验证通过，并已补充 motr 回归。

验证版本：
- MatrixOne commit: `3a8d66954014b9de54e4736d32996c261bb1b298`
- commit 信息：`fix(plan): UPDATE of ENUM column with index panics or fails to update (#25104) (#25538)`

验证 SQL：
```sql
create table t_set(
  id int primary key,
  v int default 5,
  note varchar(20) default 'd'
);
insert into t_set values (1,10,'old');
replace into t_set set id = 1, v = v + 1;
select id, v, note from t_set;
```

实际结果：
```text
id  v  note
1   6  d
```

结果符合预期：`REPLACE ... SET` 中 RHS 的目标列引用按 `DEFAULT(column)` 处理，遗漏列也按默认值填充。

已加入 motr 回归：
- test: https://github.com/matrixorigin/motr/blob/3234cff5e35466f871cd3e4feb53783579a613c4/suites/14_issue_regression/t/issue_24945_replace_set_default_refs.test
[...]

PR #25084, opened 2026-06-22
fix(plan): evaluate REPLACE ... SET RHS column references as DEFAULT(col)

ck89119:
## What type of PR is this?

- [x] BUG

issue #24945

Subtask of #24918 (section 1). For `REPLACE ... SET col = col + 1`, MySQL treats the RHS reference to a target-table column as `DEFAULT(col)`, not the current conflicting row value. Missing columns also fall back to their defaults, matching normal `REPLACE` insert semantics.

Before this PR, MatrixOne reported `invalid input: ambiguous column reference 'v'` and left the old row unchanged:

```sql
create table t_set(id int primary key, v int default 5, note varchar(20) default 'd');
insert into t_set values (1,10,'old');
replace into t_set set id = 1, v = v + 1;
select * from t_set;   -- expected 1 | 6 | d
```

### Root cause

The parser lowers `REPLACE ... SET a = x, b = y` to `tree.Replace{Columns, Rows: ValuesClause}`, i.e. the same AST shape as the VALUES form. When `buildValueScan` binds the value expressions it uses a `DefaultBinder` whose bind context is `nil`, so any column reference hits the `b.ctx == nil` branch in `baseBindColRef` and fails with "ambiguous column reference".

### Fix

- Add an `IsSetFormat` marker to `tree.Replace`, set only in the `SET` grammar branch, so the SET-only semantics can be distinguished from the VALUES form (which still rejects column references, matching MySQL).
- Add a `ReplaceValueBinder` that resolves every RHS column reference to that column's `DEFAULT(col)` expression. SET-form value binding is routed through it via a new `colRefAsDefault` flag threaded from `bindReplace` → `initInsertReplaceStmt` → `buildValueScan`. `INSERT` and the VALUES form are unchanged.

### Tests

- Unit tests: `ReplaceValueBinder` (default value resolution, case-insensitive lookup, unknown column error, unsupported bindings) — 100% coverage of the new file; planner test `TestReplaceSetColRefAsDefault`.
- BVT: new cases in `dml/replace/replace.test` covering provided/omitted columns and references to other columns' defaults (`set id=1,v=v+1`→`1|6|d`, `set id=2,v=v*2`→`2|10|d`, `set id=1,a=b+1`→`1|8|7`); verified with mo-tester (198/198), `replace_statement` 50/50 with no regression.


heni02 (approved):

aunjgr (approved):
LGTM. Correctly resolves REPLACE SET column refs to DEFAULT(col), matching MySQL behavior. The ReplaceValueBinder is well-designed and the test coverage is comprehensive (unit + BVT 198/198).

aunjgr (approved):
LGTM

aunjgr (approved):
LGTM. Clean design — separate ReplaceValueBinder that resolves RHS column refs to DEFAULT(col). Good unit + BVT coverage.

XuPeng-SH (changes requested):
I reviewed this from the MySQL-compatibility and unhappy-path angle. The core direction is right: MySQL documents that `REPLACE ... SET col = col + 1` treats the RHS column reference as `DEFAULT(col)` and cannot read the current row value. But I still see blockers before this should merge.

Blocker 1: `ReplaceValueBinder` ignores qualifiers on RHS column references.

`BindColRef` currently does:

```go
colName := strings.ToLower(astExpr.ColName())
colIdx, ok := b.tableDef.Name2ColIndex[colName]
```

That drops any table/database qualifier. As a result, a statement like this can be accepted and planned as `DEFAULT(v) + 1` just because the target table has a column named `v`:

```sql
CREATE TABLE t(id INT PRIMARY KEY, v INT DEFAULT 5);
REPLACE INTO t SET id = 1, v = other.v + 1;
```

That is not a safe interpretation of MySQL compatibility. The MySQL REPLACE docs describe the RHS column-name case as being treated as `DEFAULT(col_name)`, but this implementation turns any qualified `*.v` reference into the target column's default by ignoring the qualifier entirely.

Please either reject qualified RHS names in `REPLACE ... SET` (`astExpr.NumParts > 1`), or explicitly validate that the qualifier refers to the destination table/database before resolving it. Add regression coverage for at least a wrong qualifier such as `other.v`, and ideally the intended behavior for `t.v` if MatrixOne wants to support that spelling.

Reference: https://dev.mysql.com/doc/refman/8.4/en/replace.html

Blocker 2: current head cannot merge into `main`.

GitHub reports `mergeStateStatus=DIRTY`. I also confirmed locally with a non-mutating merge check:

```sh
```

It reports a content conflict in the generated parser file:

```text
CONFLICT (content): Merge conflict in pkg/sql/parsers/dialect/mysql/mysql_sql.go
```

Please rebase/merge current `main` and regenerate the parser so `mysql_sql.go` matches the resolved `mysql_sql.y`.

Local checks:

```sh
go test ./pkg/sql/parsers/dialect/mysql -run 'TestDebug|TestValid' -count=1
go test ./pkg/sql/parsers -run '^$' -count=0
```

The diff/parser checks passed. The merge check fails as above. I also tried the focused planner tests, but this local environment cannot compile `pkg/sql/plan` because required CGO headers/libraries are missing (`usearch.h`, `xxhash.h`; with `CGO_ENABLED=0`, dependent packages are excluded), so I am not treating that as a test failure.

iamlinjunhong (approved):

XuPeng-SH (changes requested):
I re-reviewed the current head from MySQL compatibility, planner binding scope, generated-parser consistency, mergeability, and unhappy-path/resource-risk angles. The planner direction looks correct: `REPLACE ... SET` is marked at parse time, the flag is only threaded into the VALUES/SET value-scan path, bare RHS column refs are resolved as `DEFAULT(col)`, and qualified refs are now rejected instead of being silently dequalified. That closes my previous qualified-reference blocker.

I still found one blocker before merge: `test/distributed/cases/dml/replace/replace.result` contains unrelated and inconsistent baseline churn. The PR only adds the new SET-default cases around `replace.test` lines 240-256, but the result file rewrites many pre-existing outputs at the top of the file from the plain format:

```text
name    age
Abby    24
```

to the typed/result-debug format:

```text
➤ name[12,-1,0]  ¦  age[4,32,0]  𝄀
Abby  ¦  24
```

This starts at the existing first query output (`replace.result` lines 5-6) and continues through many old cases, while later existing cases in the same file still remain in the plain format (for example around lines 323-330). So the file now mixes two result formats and obscures the actual regression coverage. It also makes this PR update dozens of old expectations that are unrelated to issue #24945.

Please regenerate or edit `replace.result` so the diff is scoped to the new `REPLACE ... SET` cases only, using the same result format already used by this file on `main`. If the typed output format migration is intentional, it should be done consistently for the whole file in a separate PR, not mixed into this bug fix.

Local checks I ran:

```sh
git merge-tree --write-tree origin/main HEAD
make -B -C pkg/sql/parsers/dialect/mysql mysql_sql.go
git diff --exit-code -- pkg/sql/parsers/dialect/mysql/mysql_sql.go pkg/sql/parsers/dialect/mysql/mysql_sql.y
go test ./pkg/sql/parsers/dialect/mysql -count=1
# with local MO_ROOT thirdparty/cgo paths
go test ./pkg/sql/plan -run 'TestReplace' -count=1\ngo test ./pkg/sql/plan -run 'TestReplaceSetColRefAsDefault|TestReplaceSetFunctionColRefAsDefault|TestReplaceValueBinder' -count=1\n```\n\nUnhappy-path pass: this change does not introduce goroutines, waits, external resources, or unbounded accumulation. The relevant risk is semantic/test-fixture correctness, and the mixed BVT baseline is the remaining blocker.

**Review by XuPeng-SH — CHANGES_REQUESTED** [review]

For `REPLACE ... SET`, the new binder path swaps the normal target-type-aware `DefaultBinder(col.Typ)` out for `ReplaceValueBinder` for all non-function RHS expressions, not just column references. That means some expressions stop being bound with the destination column type available up front.

A concrete example is scientific-notation / DECIMAL cases such as:

```sql
REPLACE INTO t SET dec_col = -1.234567890123456789e18;
```

or expressions like:

```sql
REPLACE INTO t SET dec_col = dec_col + 1.0e20;
```

Those now fall through the generic binder path and can be bound as `float64` first, only to be cast later, which can silently lose precision and change overflow behavior compared with the old target-type-aware DECIMAL binding path.

Suggested fix: keep the target-type-aware literal/expression binding behavior and override only column-reference resolution to `DEFAULT(col)` — for example by making `ReplaceValueBinder` reuse `DefaultBinder`’s typed literal behavior while intercepting `BindColRef`. Please also add regression coverage for DECIMAL / scientific-notation cases in `REPLACE ... SET`.

XuPeng-SH (approved):
Re-reviewed latest head 3ede166. The previous blockers are addressed: REPLACE ... SET is the only syntax path that enables DEFAULT(col) RHS binding, qualified RHS refs are rejected instead of being silently dequalified, DECIMAL/scientific-notation literals preserve the target column type, function arguments still bind without leaking the target type, and generated-column RHS refs now fail explicitly. Added UT/BVT coverage covers self/other column defaults, function-wrapped refs, mixed-case lookup, generated columns, qualified refs, and DECIMAL precision. CI is green, diff-check is clean, and a local merge-tree check against current origin/main has no conflicts. LGTM.

PR #25086, opened 2026-06-23
remove tag

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25046

arm64 问题已经修改. 删除tag


heni02 (approved):

XuPeng-SH (approved):
Re-checked the current head. This PR only re-enables the `BIT_COUNT` BVT cases that were previously tagged off for #25046, and that issue is now fixed in the branch, so restoring these SQL regressions is the right follow-up. I do not see a remaining correctness issue to block on.

PR #25089, opened 2026-06-23
fix(replace): apply parent-side foreign key actions during REPLACE

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24951

Subtask 3.2 of #24918. `REPLACE` removes every existing row selected by a
primary-key or secondary-unique conflict before inserting the incoming row.
MySQL applies each referencing child table's `ON DELETE` action to those actual
old parent rows. The operator-based path previously skipped these parent-side
foreign-key actions.

This change carries the evaluated conflicting old-row stream into planner-native
FK action branches; it does not execute generated background SQL. It supports
literal and evaluated inputs, including prepared parameters, `REPLACE ... SELECT`,
and `REPLACE ... TABLE`.

- **RESTRICT / NO ACTION / SET DEFAULT** reject referenced old parent rows.
- **CASCADE** owns and deletes matching child rows.
- **SET NULL** updates each physical child row once, while excluding rows owned
  by the main replacement or by a matching CASCADE action.

The lock plan coordinates parent replacement and child FK validation in the
same physical key namespaces, including hidden tables for referenced UNIQUE
keys. Shared multi-FK targets use a bounded, streaming full-domain range lock;
ordinary exclusive REPLACE targets remain row locks. Range endpoints cover the
complete supported key domain, NULL keys are skipped per row, and RC version
checks cover every merged target.

The implementation supports self references, recursive cascades, composite and
non-primary unique references, unique-prefix indexes, secondary-unique conflicts
whose old primary key differs from the incoming key, omitted/defaulted columns,
prepared input, and evaluated SELECT/TABLE sources.

The obsolete parent-side SQL generator and its direct tests have been removed.

### Tests

- Planner coverage for self-referencing SET NULL ownership, CASCADE precedence,
  recursive UNIQUE-key locking, and physical Row_ID grouping.
- Lock-operator coverage for leading/all NULL targets, per-row NULL filtering,
  complete range endpoints, merged-target RC checks, and exclusive row locking.
- Full `pkg/sql/plan` and `pkg/sql/colexec/lockop` suites, focused race tests,
  and `make static-check`.

## PR review point - Highlight it if your changes are based on aspect below:

- [x] performance regression: Shared multi-FK validation uses one bounded
  full-domain range lock per physical parent namespace.
- [ ] Behavior changed/configuration changed


heni02 (approved):

heni02 (approved):

XuPeng-SH (changes requested):
This still needs changes.

The new parent-side FK handling is still not semantics-complete for a common valid REPLACE shape: a referenced parent table that also has a secondary UNIQUE key.

In genParentSideReplaceFKSqls (pkg/sql/plan/build_util.go), the generator bails out as soon as parent.Indexes contains any unique index, and it also requires the incoming VALUES list to carry the parent PK literal. That means it only knows how to drive child RESTRICT/CASCADE/SET NULL from the incoming parent PK values. But REPLACE deletes rows selected by any conflicting unique key, not just by the incoming PK.

A concrete case is already in the new BVT coverage:

create table fk_review_p(id int primary key, u int unique, v int);
create table fk_review_c(id int primary key, pid int,
  foreign key(pid) references fk_review_p(id) on delete cascade);
insert into fk_review_p values (1, 10, 100);
insert into fk_review_c values (1, 1);
replace into fk_review_p values (2, 10, 200);

MySQL semantics here are: delete old parent row id=1 because of the UNIQUE(u) conflict, apply the ON DELETE CASCADE to child rows that reference id=1, then insert the new parent row id=2. This patch instead returns 'not supported: REPLACE on a referenced table with secondary unique keys'.

The same root cause also rejects the common auto-increment case where the REPLACE column list omits the parent PK, because pkPos cannot be found even though the conflicting old parent row is still well-defined by another unique key.

Suggestion: derive the actual set of old parent PKs that REPLACE will delete from the full conflict set (PK and secondary unique conflicts), then drive parent-side RESTRICT/CASCADE/SET NULL from that old-PK set. As long as the implementation only works from the incoming literal PK values, it will keep rejecting valid MySQL REPLACE statements on referenced parents.

pkg/sql/plan/build_util.go
```diff
@@ -1188,6 +1188,160 @@ func genPreCheckSqlsForReplaceFKSelfRefer(
...
+	if parent.Pkey != nil && len(parent.Pkey.Names) > 0 {
+		uniqueKeys = append(uniqueKeys, parent.Pkey.Names)
+	}
+	for _, idx := range parent.Indexes {
+		if !idx.Unique {
+			continue
+		}
+		parts := make([]string, len(idx.Parts))
+		for i, part := range idx.Parts {
+			parts[i] = catalog.ResolveAlias(part)
```
aunjgr: [P1/blocker] Preserve unique-prefix semantics when deriving the old parent rows that REPLACE will delete. This reduces every unique index to raw column names and drops `idx.IndexAlgoParams`, including prefix lengths. For `UNIQUE KEY u(body(4))`, an existing `abcdxxxx` and incoming `abcdyyyy` conflict in the actual unique index, but the generated predicate below compares `parent.body = 'abcdyyyy'`, so it does not find the old row. The main REPLACE can then delete that referenced parent while RESTRICT incorrectly permits it or CASCADE/SET NULL skips its child action. Prefix indexes are supported and already covered by `prefix_index_dml.sql` and the ODKU planner path. Parse `catalog.IndexPrefixLengthsFromParamsWithError(idx.IndexAlgoParams)` and generate each conflict predicate from the same indexed prefix expression, then add a referenced-parent BVT regression.

aunjgr (changes requested):
The current head addresses the earlier secondary-unique, auto-increment, composite-FK, and non-PK-reference gaps, but it is still not merge-ready because conflict discovery does not preserve unique prefix-index semantics. The inline example shows how REPLACE can delete an old referenced parent while the generated exact-column predicate misses it, causing RESTRICT/CASCADE/SET NULL to be skipped. Please build the old-row predicate from the same prefix key semantics used by the unique index and add a BVT regression. Prepared parameters and REPLACE SELECT remain explicit limitations; the PR description should also be updated because it still says composite/non-PK references are skipped even though this head supports them.

XuPeng-SH (changes requested):
This latest head looks materially better: the previous unique-prefix blocker appears fixed in both the REPLACE conflict join path and the parent-side FK old-row predicate generation. But I still see one remaining blocker.

**High: referenced parent tables still reject valid REPLACE statements when a secondary unique-key column is omitted, even if the omitted value deterministically cannot conflict.**

The remaining problem is in `genParentSideReplaceFKSqls` (`pkg/sql/plan/build_util.go`, around the unique-key predicate builder). It still treats any missing unique-key part as fatal unless *all* missing parts are auto-increment PK columns:

- for each unique key, it looks up every key part in the explicit REPLACE column list,
- if a part is omitted and is not auto-increment, it eventually returns `not supported: REPLACE with an omitted conflict key`.

That is still too strong. A concrete valid MySQL shape is:

```sql
create table p(
  id int primary key,
  u varchar(20) unique,   -- nullable, omitted columns default to NULL
  v int
);
create table c(
  id int primary key,
  pid int,
  foreign key(pid) references p(id) on delete cascade
);

insert into p values (1, 'x', 100);
insert into c values (1, 1);
replace into p(id) values (1);
```

MySQL semantics here are straightforward: the old parent row is selected by the **PK conflict**; the omitted nullable `u` becomes `NULL`, so it cannot introduce a secondary-unique conflict; parent-side FK action should therefore run from the old PK set and the REPLACE should succeed.

But the current planner still rejects this shape before execution because the secondary unique key `u` is omitted from the explicit column list.

I verified this two ways:

1. By reading the current `genParentSideReplaceFKSqls` logic in `pkg/sql/plan/build_util.go:1267-1296`.
2. With a deterministic scratch planner repro in a throwaway worktree: I deep-copied the existing `replace_fk_cp` mock table, added a secondary `unique(v)` on its nullable `v` column, and `genParentSideReplaceFKSqls` returned `not supported: REPLACE with an omitted conflict key` for `REPLACE INTO replace_fk_cp(id) VALUES (1)`.

So the PR still rejects valid REPLACE-on-referenced-parent statements whenever a secondary unique key is omitted, even if the omitted value is a deterministic non-conflicting default.

**Suggested fix:** derive the conflict-key set from the fully materialized inserted row **after default expansion**, not only from the explicitly supplied literals in the AST. That would let omitted nullable/defaulted unique columns participate correctly: `NULL` would contribute “no conflict”, constant defaults would contribute their actual value, and only truly non-materializable shapes would need to fail closed.

XuPeng-SH (changes requested):
Re-reviewed the latest head (`054bb35`). The earlier secondary-unique, nullable/default, temporal-default, and prefix-index comments are substantially addressed, but this is still not merge-ready.

1. **High: parent-side conflict discovery still uses the unmaterialized AST literal, so it can disagree with the actual REPLACE conflict set.**

`genParentSideReplaceFKSqls` formats supplied `NumVal`/`StrVal` nodes directly into the background predicate (`build_util.go:1389-1417`). The main REPLACE path does not use those raw values: `initInsertReplaceStmt` applies `forceAssignmentCastExpr` to the destination column type before unique-key probing (`bind_insert.go:2729-2762`).

A deterministic example is a referenced parent with `UNIQUE(u)` where `u DECIMAL(5,2)`:

```sql
create table p(id int primary key, u decimal(5,2) unique);
create table c(id int primary key, pid int,
  foreign key(pid) references p(id) on delete cascade);
insert into p values (1, 1.23);
insert into c values (1, 1);
replace into p values (2, 1.234);
```

The main plan assignment-casts `1.234` to `DECIMAL(5,2)` (`1.23`), so it conflicts with and deletes parent `id=1`. The generated parent-side action instead searches with `p.u = 1.234`, which is false (`CAST(1.234 AS DECIMAL(5,2)) = 1.234` evaluates false), so the cascade misses the child. Equivalent gaps exist for other width/scale/type normalizations.

Please derive the old-row set from the same fully materialized row image/conflict machinery used by REPLACE, or apply exactly the same assignment conversion before serializing every conflict-key part. Add a BVT regression using a narrowing/rounding type, not only already-canonical literals/defaults.

2. **High: checks/actions run before the REPLACE lock pipeline, leaving a child-insert TOCTOU window.**

`Compile.runOnce` executes every `REPLACE_PARENT_CHK` / `REPLACE_PARENT_ACTION` first (`compile.go:527-548`). The parent/new-and-old-key `LOCK_OP` is inside the main REPLACE pipeline (`bind_replace.go:1011-1020`), which starts only afterward. A plain check query, or a child DELETE/UPDATE whose parent is only read by an `EXISTS`, does not hold the old parent delete lock across this gap.

For example: T1 runs the cascade action and finds/deletes the current children; before T1 starts the main pipeline, T2 inserts a child referencing the still-existing old parent and commits; T1 then locks/deletes that old parent because of a secondary-unique conflict and inserts a different PK. T2s child is left orphaned. Sharing the transaction is necessary for rollback, but it does not close this inter-statement race.

Please drive FK actions from the locked old-row conflict stream in the main plan, or otherwise acquire and hold the required old-parent locks before checks/actions so concurrent child FK checks serialize correctly. A deterministic synchronization test should cover the action-to-main-pipeline boundary.

3. **The latest Coverage check is a real PR failure.** It reports 132/186 modified lines covered (70.97%, required >75%); `bind_replace.go` is 55.26% and `build_util.go` is 73.08%. The Ubuntu UT failure is unrelated (`TestDispatchEmptyInputWaitsForRemoteReceiver` in `pkg/sql/colexec/dispatch`), and all selected BVT jobs passed, but the coverage gate still needs to be fixed.

Performance/unhappy-path note: the generated text is amplified as rows × unique keys × key parts and then repeated for each referencing FK, with each statement rescanning the parent conflict predicate. Reusing one materialized, locked old-row set would also avoid that parse/scan amplification while closing both correctness gaps above.

XuPeng-SH (changes requested):
Re-reviewed current head `aad5a518`. The earlier secondary-unique, prefix-index, omitted-default, assignment-rounding, lock-mode/deep-copy, and coverage findings are materially addressed, but I still do not think this is merge-ready.

1. **High — the TOCTOU fix is a no-op in supported optimistic/SI deployments.**

Both sides of the new serialization contract are `LOCK_OP`s: the parent `SELECT ... FOR UPDATE` and the shared parent lock added to child FK validation. However, `lockop.(*LockOp).Call` explicitly bypasses all locking when `!Txn().IsPessimistic()` (`pkg/sql/colexec/lockop/lock_op.go:135-139`). Optimistic mode is still a valid CN configuration and defaults to SI (`pkg/cnservice/types.go:209-214, 362-374`), while both optimistic BVT jobs on this run were skipped.

A concrete orphan race remains possible in that mode:

- T1 starts `REPLACE` on a referenced parent and completes the generated RESTRICT check/CASCADE action.
- T2 inserts a child referencing the still-visible old parent and commits. Its new shared lock is also bypassed.
- T1 runs the main REPLACE pipeline, deletes the old parent because of a PK/secondary-UK conflict, and commits.

The transactions write different rows/tables, so ordinary SI write-write validation does not supply the missing read-write conflict. Either provide an equivalent correctness mechanism for optimistic mode or fail closed/document that this statement shape is unsupported there. Please add a deterministic two-transaction regression with an explicit synchronization barrier at the action-to-main-pipeline boundary and bounded timeouts; fixed sleeps or probabilistic retry loops would be flaky and would not prove the invariant.

2. **High — every child FK insert/replace now locks the full parent scan, causing unbounded lock amplification and unrelated write blocking.**

`appendModernChildFkMarkOks` puts the shared `LOCK_OP` directly above the unfiltered parent `TABLE_SCAN` (`pkg/sql/plan/bind_insert.go:1060-1084`), before the MARK join. Runtime-filter generation explicitly returns for MARK joins (`pkg/sql/plan/runtime_filter.go:95-96`), so the lock input is not narrowed to the referenced parent keys. Consequently, one child row can acquire shared locks for every row in a large parent table (and multiple FKs repeat the scan/lock footprint), held until transaction end; sufficiently large estimates can also be upgraded by the existing whole-table-lock logic. This turns a point FK validation into O(parent cardinality) lock-service memory/work and blocks updates/deletes of unrelated parent keys.

The PR description currently understates this as possibly locking “more parent rows.” The lock must be driven by the matched parent-key set, not the raw parent scan. Please add a plan/behavior regression proving that inserting a child for parent key 1 does not lock or block a writer of unrelated parent key 2, including a large-parent case.

3. **Medium — conflict discovery still does not use the exact assignment conversion for CHAR/VARCHAR.**

`castForColumn` emits ordinary SQL `CAST` for every type (`pkg/sql/plan/build_util.go:1383-1385`), but the real INSERT/REPLACE row uses `forceAssignmentCastExpr`, which deliberately selects `cast_strict` for CHAR/VARCHAR (`pkg/sql/plan/build_constraint_util.go:1041-1046`). Ordinary CAST truncates an over-width string; assignment cast rejects it.

For example, with parent `p(id PK, u VARCHAR(3) UNIQUE)`, a RESTRICT child referencing `p.id`, existing `(1,'abc')`, and `REPLACE p VALUES (2,'abcd')`, the generated precheck truncates `'abcd'` to `'abc'`, finds the old parent, and reports “row is referenced.” The actual REPLACE row should instead fail with data-too-long before that conflict exists. The decimal test only covers rounding and does not close this conversion-equivalence gap. Please derive conflict keys from the same materialized row image/conversion path, or otherwise make the conversion exactly equivalent, and cover an over-width CHAR/VARCHAR case end to end.

The targeted planner tests passed 10 consecutive runs, and the lockop/compile test selections also passed. All current required checks are green, but the optimistic jobs were skipped and the existing tests assert generated SQL/plan shape rather than the two concurrency invariants above.

XuPeng-SH (changes requested):
The latest head materially fixes the optimistic-mode false safety and the full-parent-scan lock amplification, but one High concurrency blocker remains for foreign keys that reference a non-primary UNIQUE key.

The child and parent paths lock different physical namespaces:

- appendModernChildFkMarkOks resolves a matching non-PK UNIQUE index to its hidden index table and takes a shared lock on that index table key (pkg/sql/plan/bind_insert.go:1135-1199).
- REPLACE_PARENT_LOCK is a SELECT FOR UPDATE on the base parent table (pkg/sql/plan/build_util.go:1473-1476). collectLockTargets creates its target from the original base TABLE_SCAN and locks that table ID and primary-key column (pkg/sql/plan/query_builder.go:6325-6348); this target is created before index access-path rewriting.

Concrete failure:

1. p(id PK, u UNIQUE) is (1,10), and c.parent_u references p.u.
2. T1 runs REPLACE p VALUES (1,20). Its pre-phase locks base row id=1, sees or removes the current children, but has not run the main REPLACE yet.
3. T2 inserts a child referencing u=10. It shared-locks the hidden UNIQUE-index key u=10, which does not conflict with T1s base-table PK lock, validates the still-visible old parent, and commits.
4. T1 runs the main REPLACE, removes u=10, and commits. The new child is orphaned. The same gap lets a RESTRICT pre-check be bypassed.

The existing tests confirm each lock independently but never prove that the parent and child operations contend on the same lock resource; TestChildInsertLocksReferencedUniqueIndexKey actually confirms the child target is the index table.

Please align the lock namespaces before the parent action/check boundary: for example, lock every referenced parent key being removed in the same base-PK/index-key namespace used by child validation, or use another design that supplies an equivalent happens-before relation. Add a deterministic two-transaction regression with a barrier between the parent action/check and the main REPLACE, specifically for an FK referencing a non-PK UNIQUE key. This is part of the #24951 correctness closure.

gouhongshen (changes requested):
## Codex automated review

The new FK locking and parent-side conflict detection have correctness gaps that can violate referential integrity.

### P1 - Lock prerequisite steps are not ordered before the main DML (pkg/sql/plan/bind_insert.go:1201)
Each LOCK_OP is appended as an independent query step, while the main path later consumes another scan from the same source step. Query steps are compiled and run concurrently, and the sink dispatches batches to all receivers; step-number ordering is not a happens-before dependency. The main MARK scan and child write can therefore run before the shared lock is acquired. A concurrent parent REPLACE can delete the referenced parent after the stale MARK result and before the delayed lock, bypassing RESTRICT/CASCADE and allowing an orphan child. An actual barrier is required.

### P1 - Child FK lock keys are not converted to the parent key type (pkg/sql/plan/bind_insert.go:1108)
The lock expression keeps the child column type, but the lock target type comes from the parent or hidden index table. FK DDL validation only compares Type.Id, so differing decimal scales are accepted. For example, parent DECIMAL(5,2) and child DECIMAL(5,3) can compare equal in the FK MARK join, while the decimal lock fetcher encodes their raw values differently; 1.23 and 1.230 produce different lock keys. The child shared lock then does not conflict with the parent REPLACE lock. Cast/rescale each FK component to the referenced physical key type before locking.

### P1 - Generated columns shift the parent-side REPLACE value mapping (pkg/sql/plan/build_util.go:1226)
The main REPLACE input excludes both hidden and generated columns (`getInsertColsFromStmt`), but this implicit position map skips only hidden columns. With columns `(id, g GENERATED, u UNIQUE)`, `REPLACE ... VALUES (2, 10)` is interpreted here as id=2, g=10, u omitted, while the main plan correctly interprets it as id=2, u=10. The generated parent lock/check/action SQL therefore misses the secondary-key conflict; a CASCADE can leave the old child orphaned and RESTRICT can incorrectly succeed. Explicit `g=DEFAULT` has the same mismatch because `stripGeneratedDefaultCols` rewrites the row but leaves `stmt.Columns` unchanged.

gouhongshen (changes requested):
## Codex automated review

Parent-side REPLACE FK SQL has two correctness regressions and a lock-order deadlock risk.

### P1 - Re-evaluate foreign_key_checks for cached REPLACE plans (pkg/sql/plan/build.go:185)
The new parent-side DetectSqls are generated only when foreign_key_checks is enabled at plan-build time, but the resulting query is still cacheable: it does not set HasForeignKeyAction, and compile.go executes DetectSqls without checking the current session variable. SET foreign_key_checks does not invalidate these plan caches. Thus a plan built with checks enabled still runs CASCADE/SET NULL/restrict SQL after checks are disabled, while a plan built with checks disabled omits actions after checks are re-enabled. This affects both prepared statements and ordinary session plan caching.

### P1 - Do not treat explicit auto-increment zero as a conflicting parent key (pkg/sql/plan/build_util.go:1409)
The conflict predicate only treats an omitted auto-increment column as non-conflicting; an explicit literal 0 is formatted as the parent key. The actual REPLACE PRE_INSERT path converts explicit zero to NULL and allocates a new auto-increment value unless NO_AUTO_VALUE_ON_ZERO is set. For example, after inserting parent id=0 with that mode enabled, resetting the mode and executing REPLACE INTO parent VALUES (0, ...) will cause the generated parent-side action to delete or update children of id=0, even though the main REPLACE inserts a new id and leaves the id=0 parent untouched.

### P2 - Avoid rejecting valid REPLACE statements because of unrelated generated unique indexes (pkg/sql/plan/build_util.go:1354)
uniqueKeys includes every unique index on the parent, and materializeDefault returns NotSupported whenever an omitted key column is generated. Therefore a valid parent such as p(a PRIMARY KEY, g GENERATED ALWAYS AS (a+1) VIRTUAL, UNIQUE(g)) causes every REPLACE that supplies only a to fail, even when the child foreign key references a and the generated index is not otherwise problematic. The normal REPLACE binder already supports generated columns; this new guard unnecessarily broadens the fail-closed case to valid schemas.

### P2 - Use one canonical order for child and parent prerequisite locks (pkg/sql/plan/bind_insert.go:1093)
Child INSERT FK prerequisite locks are chained in foreign-key declaration order. The new parent-side lock SQL instead sorts referenced hidden index tables before acquiring FOR UPDATE locks in build_util.go. With two child FKs referencing two unique keys on the same parent, declared in the reverse of that sort order, a child INSERT can hold a shared lock on key B while waiting for A, while a concurrent REPLACE holds A exclusively and waits for B. This creates a reproducible lock cycle on valid schemas; both sides need the same canonical key order and a two-FK concurrency test.

XuPeng-SH (changes requested):
Requesting changes after reviewing latest head e0c8afbf6 end to end.

1. [P1/correctness] Parent-side conflict discovery decides auto-increment zero from the parser AST, before destination assignment conversion. isExplicitNumericZero only ParseFloat values that already spell numeric zero, while PRE_INSERT converts every zero in the destination integer vector to NULL when NO_AUTO_VALUE_ON_ZERO is absent. I added temporary cases for 0x0 and b'0' to TestReplaceParentSideFKAutoIncrementZero; both failed because this code generated a CASCADE DELETE against parent id 0. The main REPLACE instead converts those values to zero, treats them as generated, and allocates a new id, so the existing id=0 parent is not replaced. CASCADE/SET NULL therefore mutate the wrong child rows, and RESTRICT can reject a valid statement. This class also includes other literals whose assignment cast yields zero. Derive the pre-phase row image through the same assignment-cast/auto-increment semantics as PRE_INSERT, and add unit plus pessimistic BVT coverage for hex/bit and cast-to-zero inputs.

2. [P1/performance] appendModernChildFkMarkOks now unconditionally inserts LOCK_OP -> SINK -> SINK_SCAN for every non-self FK write path, including INSERT, INSERT IGNORE, ODKU, and REPLACE. In optimistic transactions LockOp.Call explicitly bypasses locking, but the blocking materialization and extra query step remain. Thus bulk child inserts buffer the whole input for no concurrency benefit; both optimistic CI jobs on this head are skipped. Preserve the original streaming plan when locks cannot be taken, or make the prerequisite conditional without plan-cache drift, and cover both transaction modes with a plan/performance regression. Also bound or measure the N rows x unique keys x child FKs generated-SQL amplification.

3. [P2/scope and regression risk] The net diff also partially resurrects code from #25741 that current main intentionally reverted in #25848: FillValuesOfParamsInPlan deep-copies the plan again, DeepCopyExpr restores IsBin/Src copying, and the reverted tests return. These changes are unrelated to #24951 and were explicitly removed from the base after a production retry panic in the larger feature. Drop the merge residue, or split it out with an explicit rationale and its own full lifecycle/performance validation; keeping a partial emergency-revert rollback inside this PR makes the regression surface ambiguous.

What passed: diff-check; focused plan, compile, lockop, and frontend tests under -race. Current required pessimistic CI is green. What failed: the temporary hex/bit auto-increment-zero regression above. Optimistic BVT coverage is skipped.

pkg/sql/plan/bind_insert.go
```diff
@@ -1034,24 +1036,245 @@ func (builder *QueryBuilder) appendModernChildFkMarkOks(
...
+			NodeType: plan.Node_PROJECT, Children: []int32{lockNodeID},
+			ProjectList: slices.Clone(lockOutput[:len(rowProject)]), BindingTags: []int32{selectTag},
+		}, bindCtx)
+		lastNodeID = lockNodeID
+	}
+
+	// Materialize the row image only after every referenced key has been locked.
+	// The final validation/DML step consumes this single sink dependency, avoiding
+	// unsupported multi-hop sink chains while preserving lock-before-scan ordering.
+	lockSinkID := appendSinkNodeWithTag(builder, bindCtx, lastNodeID, selectTag)
```
XuPeng-SH: [P1/performance] This blocking materialization is planned for every non-self FK child write, independent of transaction mode. In optimistic mode `LockOp.Call` immediately delegates to its child without taking a lock, yet this `SINK/SINK_SCAN` still buffers the entire INSERT/INSERT IGNORE/ODKU/REPLACE input and adds a query step. That is a broad bulk-write regression with no correctness benefit, and both optimistic CI jobs on this head are skipped. Please preserve streaming when locking is unavailable, and add a two-mode plan/performance regression.

pkg/sql/plan/build_util.go
```diff
@@ -1188,6 +1192,558 @@ func genPreCheckSqlsForReplaceFKSelfRefer(
...
+		}
+		return nil, false
+	}
+	treatAutoIncrementZeroAsGenerated := true
+	if sqlMode, err := ctx.ResolveVariable("sql_mode", true, false); err != nil {
+		return "", nil, nil, err
+	} else if mode, ok := sqlMode.(string); ok {
+		treatAutoIncrementZeroAsGenerated = !strings.Contains(strings.ToUpper(mode), "NO_AUTO_VALUE_ON_ZERO")
+	}
+	isExplicitNumericZero := func(expr tree.Expr) bool {
```
XuPeng-SH: [P1/blocker] This decision must be made from the destination-typed row image, not the literal spelling. `PRE_INSERT` runs assignment conversion first and then converts every integer zero to NULL when `NO_AUTO_VALUE_ON_ZERO` is absent. Here `0x0` and `b'0'` are not recognized, so the pre-phase targets old parent id 0 while the main REPLACE allocates a new id. I reproduced both as failing additions to `TestReplaceParentSideFKAutoIncrementZero`: each generated a CASCADE DELETE containing `id = cast(... as INT)` when actions should be empty. The same mismatch applies to other inputs whose assignment cast yields zero. Please reuse/evaluate the real assignment-cast semantics and add unit plus BVT coverage.

gouhongshen (changes requested):
## Codex automated review

Parent-side REPLACE actions can target the wrong auto-increment key, and FK child writes incur an unnecessary unconditional materialization barrier.

### P1 - Recognize all zero literals before generating parent actions (pkg/sql/plan/build_util.go:1267)
`isExplicitNumericZero` only parses decimal `NumVal.String()` values. Hex and bit literals such as `0x0` and `b'0'` are decoded and converted to integer zero by the main PRE_INSERT path, which then turns zero into NULL and generates a new auto-increment value. The parent-side prephase does not recognize them as generated, so it locks and applies CASCADE/SET NULL/RESTRICT actions against the real parent row with id 0 instead of the row actually replaced.

### P2 - Avoid materializing FK child input when locks are bypassed (pkg/sql/plan/bind_insert.go:1268)
This unconditionally inserts a SINK/SINK_SCAN barrier for every child write with a non-self foreign key. In optimistic transactions, `LockOp.Call` immediately bypasses the lock operator, but the input is still fully buffered and copied through the sink, adding an O(input) memory and latency cost to ordinary INSERT, INSERT IGNORE, ODKU, and REPLACE workloads. The plan construction should avoid this stage when the lock path is not needed.

gouhongshen (changes requested):
## Codex automated review

The implementation regresses valid REPLACE forms on referenced parent tables by rejecting them instead of applying FK actions.

### P1 - Support non-literal and SELECT parent REPLACEs instead of rejecting them (pkg/sql/plan/build_util.go:1216)
`genParentSideReplaceFKSqls` is invoked for every referenced parent while FK checks are enabled. It rejects `REPLACE ... SELECT/TABLE` outright at lines 1214-1216, and the later static-literal formatter rejects parameter/function conflict keys (line 1659). Thus valid statements such as `PREPARE s FROM 'REPLACE INTO parent VALUES (?, ?)';` and `REPLACE INTO parent SELECT ...` now fail before execution, even when they produce no conflict. The pre-PR modern REPLACE binder supported these source forms; the added BVT deliberately changes their expected result to `not supported`. The FK action mechanism needs to consume the evaluated row image (or otherwise preserve these forms), rather than narrowing REPLACE syntax for every table that happens to be referenced.

aunjgr (changes requested):
Two blockers remain. Explicit numeric-zero recognition misses hex and bit zero forms, so autoincrement preprocessing can generate a new key while parent-side FK actions still operate on key zero. Also, non-self-FK input is unconditionally materialized even when the optimistic transaction path bypasses locking, creating O(input) memory and latency. Normalize all exact zero literal forms before both phases and materialize only when the selected execution path requires it.

aunjgr (changes requested):
Two blockers remain. Explicit numeric-zero recognition misses hex and bit zero forms, so autoincrement preprocessing can generate a new key while parent-side FK actions still operate on key zero. Also, non-self-FK input is unconditionally materialized even when the optimistic transaction path bypasses locking, creating O(input) memory and latency. Normalize all exact zero literal forms before both phases and materialize only when the selected execution path requires it.

gouhongshen (changes requested):
## Codex automated review

The new FK prerequisite locking can deadlock valid concurrent child INSERT and parent REPLACE operations.

### P1 - Use the same canonical order for parent REPLACE index locks (pkg/sql/plan/bind_replace.go:917)
The parent path appends unique-index lock targets in `tableDef.Indexes` order, and `performLock` acquires targets in that exact slice order. The child path now explicitly sorts its FK locks by parent table and hidden-index name in `appendModernChildFkMarkOks`. With a parent having two referenced UNIQUE indexes whose definition order is the reverse of their hidden-index-name order, a child INSERT can acquire shared lock A then wait on B while REPLACE acquires exclusive B then waits on A. This creates a transaction deadlock for valid schemas; sort the parent lock targets by the same canonical physical-key order and add a two-FK concurrency regression.

### P2 - Remove the obsolete parent-side SQL generator and its disconnected tests (pkg/sql/plan/build_util.go:1195)
`genParentSideReplaceFKSqls` is no longer called by the production REPLACE path; the head executes actions from the evaluated in-plan old-row stream instead. The 600+ line generator and numerous unit tests now validate a dead SQL-based implementation (including unsupported-shape behavior) rather than executable behavior, adding duplicate FK semantics that will drift. Delete it and replace the tests with assertions/BVTs for the active plan path.

gouhongshen (changes requested):
## Codex automated review

A referenced nullable unique key still leaves a concurrency hole when the replacement value is statically NULL. Current CI is green; focused local tests could not run because libmo.dylib is absent.

### P1 - Lock old unique keys even when the incoming key is NULL (pkg/sql/plan/bind_replace.go:933)
`skipUniqueIdx` describes the incoming key, but this branch skips the entire index path, including deletion and locking of the matched old index key. For `p(id PK, u UNIQUE NULL)` with a child FK referencing `u`, an existing `(1,10)`, and `REPLACE INTO p(id) VALUES (1)`, the replacement conflicts by PK and deletes old `u=10`, while this plan locks only the base PK because incoming `u` is NULL. A concurrent child insert can therefore shared-lock hidden index key 10, validate against the still-committed old parent, and commit before the REPLACE transaction commits, leaving an orphan. Separate skipping the new NULL index entry/conflict probe from maintaining and locking a non-NULL old index entry.

gouhongshen (changes requested):
## Codex automated review

Three blocking correctness/concurrency gaps remain. The latest nullable-unique lock finding is fixed, and current CI is green; focused local tests could not run because libmo.dylib is absent.

### P1 - Compose multiple SET NULL actions for the same child row (pkg/sql/plan/build_dml_util.go:930)
The FK loop invokes buildUpdatePlans independently for each SET NULL constraint. For p(id PK), c(id PK, p1, p2) with both p1 and p2 referencing p(id) ON DELETE SET NULL, replacing p(1) builds two full-row images from the original child row: (id,NULL,1) and (id,1,NULL). Updates are implemented as delete-plus-insert, and nothing combines those images into the required (id,NULL,NULL); the UNION machinery also cannot deduplicate different images. This can fail on duplicate writes or preserve one dangling reference. Group all actions for the same child row into one update.

### P1 - Lock recursively cascaded rows before processing their dependents (pkg/sql/plan/build_dml_util.go:714)
The root REPLACE correctly locks its old parent rows, but a cascaded child delete puts its exclusive LOCK_OP in one query step while descendant FK checks/actions consume the original unlocked source in sibling steps. With P -> C -> G, T1 can snapshot G, T2 can insert G referencing C and release its shared C lock before T1 requests the exclusive C lock, and T1 can then delete C after its G action already missed the new row. This leaves an orphan. Each recursive level needs the same lock-then-materialize dependency used for the root before scanning or acting on dependents.

### P1 - Canonicalize lock order by key within each physical table (pkg/sql/plan/bind_replace.go:1014)
The parent locks each table's new key before its old key, while child FK locks are ordered only by physical table name; equal-table FKs retain declaration order. For p(id PK, u UNIQUE) containing (1,10) and (2,20), REPLACE p VALUES (2,10) can hold key 2 and wait for key 1, while a concurrent child insert with two FKs to p.id and values (1,2) holds key 1 and waits for key 2. This is a lock cycle. Consolidate all keys per physical table and sort/deduplicate the encoded keys consistently on both paths.

gouhongshen (changes requested):
## Codex automated review

The latest fixes still leave a recursive-cascade deadlock and significant prepared-plan and lock-range regressions.

### P1 - Recursive cascade locks use a different canonical order (pkg/sql/plan/build_dml_util.go:510)
Recursive cascades sort base and hidden-index lock targets by TableId, while child FK inserts and parent REPLACE sort hidden targets by IndexTableName. Index table IDs are opaque and can differ from name order (the existing tests deliberately construct this inversion). With P→C→G, a recursive REPLACE can acquire C's hidden index locks in ID order while a concurrent G insert acquires them in name order, producing a lock cycle and timeout. Reuse one canonical comparator for all parent/child/recursive paths and add a multi-index recursive regression.

### P2 - FK-sensitive prepared plans rebuild on every EXECUTE (pkg/frontend/computation_wrapper.go:604)
shouldRebuildPreparePlan(false, preparePlan.Plan) is true whenever HasForeignKeyAction is set, so the plan is rebuilt on every EXECUTE rather than only when foreign_key_checks changes. The REPLACE binder sets this flag for any table with FKs or referenced children, and shouldCachePrepareCompile also disables the cached compile. This turns prepared FK-sensitive REPLACE statements into full plan rebuilds per execution; track the setting state or use execution-time gating.

### P2 - Multiple FK references force a full parent-range lock (pkg/sql/colexec/lockop/lock_op.go:320)
appendModernChildFkMarkOks emits one shared target per FK. When a child has two FKs referencing the same parent physical namespace, performLock groups them and sets lockTable=true, causing the fetcher to lock the full key domain. A single child insert then conflicts with writes to unrelated parent keys for the transaction lifetime, creating a substantial concurrency and throughput regression. The lock strategy should preserve per-key locking with a deterministic order or otherwise avoid blanket range locking.

Issue #25026, opened 2026-06-17
[Feature Request]: IVFFLAT/IVFPQ/CAGRA support quantization bf16/float16/int8/uint8


cpegeric:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

IVFFLAT support quantization bf16/float16/int8/uint8
IVFPQ/CAGRA support base type float16/

### Documentation, Adoption, Use Case, Migration Strategy

# Feature: Vector index narrow base-types & quantization (ivfflat / ivfpq / cagra) + basic array functions

This change adds **narrow vector base types** (`vecf16`, `vecbf16`, `vecint8`, `vecuint8`)
and a **`QUANTIZATION` option** to the three vector index algorithms, plus extends the
**basic array/vector SQL functions** to work on the new narrow types.

The two parts can be tested independently:

1. [Part 1 — Vector index base types & quantization](#part-1--vector-index-base-types--quantization)
2. [Part 2 — Basic array functions](#part-2--basic-array-functions)

A copy-paste [end-to-end smoke test](#end-to-end-smoke-test) is at the bottom.

---

## Background concepts

- **Base type** — the declared type of the vector column (`VECF32(d)`, `VECF16(d)`, …).
  This is what `INSERT`/`SELECT` see.
- **QUANTIZATION** — an *index-only* storage option. The index keeps a **downcast copy**
  of each vector (the "entries") in the quantization type to save space / speed up search;
  the base column is unchanged. Omitting `QUANTIZATION` keeps the entries in the base type.
- **op_type** — the distance metric: `vector_l2_ops` (L2, default), `vector_l2sq_ops`
  (L2 squared), `vector_ip_ops` (inner product), `vector_cosine_ops` (cosine).

---

### Supported matrix

| Capability | IVFFLAT | IVFPQ | CAGRA |
|---|---|---|---|
| Backend | CPU | GPU/cuVS | GPU/cuVS |
| Base types | f32, f64, f16, bf16, int8, uint8 | f32, f16 | f32, f16 |
| QUANTIZATION | f32, f16, bf16, int8, uint8 | f16, int8, uint8 | f16, int8, uint8 |
| op_type l2 / l2sq | yes | yes | yes |
| op_type ip / cosine | yes | yes | yes |
| op_type l1 | yes | no | no |
| int8/uint8 quant metric | any | L2 only | L2 only |
[...]

PR #25095, opened 2026-06-23
IVFFLAT/IVFPQ/CAGRA support bf16, float16, int8 and uint8 quantization

cpegeric:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue https://github.com/matrixorigin/matrixone/issues/25026 24882 20567

1. IVFFLAT support bf16, float16, int8 and uint8 quantization
2. bug fix GPU concurrent Kmeans clustering
3. IVFPQ/CAGRA support float32, float16 as base type and its quantization int8 and uint8
4. bug fix slow int8 quantization in GPU. moved to CPU computation
5. basic array function for bf16, float16, int8, and uint8


fengttt (approved):

XuPeng-SH (changes requested):
I re-checked the current head and still see two substantive correctness issues in the cuVS quantizer path.

1. **The 1-byte quantizer fallback still learns from already-quantized bytes.**
   In `cgo/cuvs/index_base.hpp`, `train_quantizer_if_needed()` still auto-trains from `flattened_host_dataset` after explicitly warning that this buffer may already hold `int8/uint8` storage values when data came in through the public storage-typed constructors / add-chunk path. In that case the quantizer learns the compressed range, not the original float range, so later base-typed search/extend paths can silently quantize against the wrong min/max.

   **Suggestion:** do not auto-train from storage-typed data. Require either an explicit quantizer/range or original base-typed training data before enabling base-typed search/extend on pre-quantized indexes.

2. **The new “strided sample” still ignores the tail for 501–999 row builds.**
   With `n_train = min(500, count)` and `stride = count / n_train`, any `501 <= count < 1000` still collapses to `stride == 1`, so the sampling loop only visits rows `0..499`. That means extrema in the tail are still missed, even though the comment now claims the sampler covers all rows.

   **Suggestion:** choose indices proportionally across the full range (for example `r = j * (count - 1) / (n_train - 1)`) or switch to a true uniform/reservoir sampler.

I would keep this at request changes until those two are addressed, because both can directly bias quantization and search quality without any obvious runtime failure.

heni02 (approved):

aunjgr (approved):
LGTM for the quantization support. Well-structured across cuvs C++/CUDA layer, Go bindings, and SQL compilation.

gouhongshen (approved):

XuPeng-SH (changes requested):
Re-reviewed the current head (`0fd4963e`). The earlier "train from storage bytes" issue is fixed, but I still see two merge blockers.

1. **[P1] Remove full-vector heap allocations from every narrow-vector comparison.**

   `pkg/container/types/compare.go:114-115` implements `ArrayElementCompare` by converting both operands to newly allocated `[]float32` slices. This comparator is called for every comparison in `pkg/sort/sort.go`, merge-order/merge-top, scalar comparison expressions, and twice per row in the vector min/max path.

   This is measurable on the current head: comparing two 768-dimensional `vecf16` values costs **6144 B/op and 2 allocs/op**. A 100k-row sort performs approximately `N log2(N)` comparisons, which means roughly **10.2 GB of transient heap allocation** in the comparator alone, before accounting for the sort or vector data. That makes the newly supported normal `ORDER BY` path vulnerable to severe GC pressure or OOM.

   Please compare lexicographically without materializing slices: convert one bf16/f16 scalar at a time, and compare int8/uint8 directly. Add an allocation regression benchmark/test for a representative dimension (for example 768) that requires zero allocations per comparison.

2. **[P1] The data-file quantizer still does not sample the full range for 501-999 rows.**

   In `cgo/cuvs/quantize.hpp:325-335`, `n_train` is 500 and `stride = n_rows / n_train`. For every `501 <= n_rows < 1000`, `stride` is 1, so the loop samples only rows `0..499`; it does not sample "across ALL n_rows" as the comment says. For example, if rows 0-499 are in `[0,1]` and row 500 contains the maximum, the learned range remains `[0,1]`, the tail saturates during int8/uint8 quantization, and recall is silently degraded.

   This is reachable from the public production API `pkg/cuvs.NewGpuIvfPqFromDataFile[B,Q]` through `gpu_ivf_pq_new_from_data_file` and `load_host_matrix<T>`, so it is not just a benchmark helper. Please use proportional endpoints (for example `j * (n_rows - 1) / (n_train - 1)`) or another sampler that actually spans the complete file, and add adversarial 501- and 999-row regression cases with extrema only in the tail.

The focused CPU tests for types/vector/sort/compare/functions/quantizer/metric and narrow frontend/plugin planning pass. This host has no working NVIDIA driver, so the GPU-tagged path could not be executed locally; normal CPU CI does not close these two findings.

gouhongshen (changes requested):
## Codex automated review

The prior review's compressed-byte auto-training issue is fixed, but two blocking defects remain in the newly extended IVF-PQ data-file quantization path. Focused pure-Go type/metric/quantizer tests pass.

### P1 - Retain the scalar-quantizer range when loading source-typed data files (cgo/cuvs/ivf_pq.hpp:347)
For `NewGpuIvfPqFromDataFile[float32,int8]` (or `uint8`), this calls `load_host_matrix<T>` with `T` as the storage type. That loader quantizes the float MODF contents using a temporary scalar quantizer and discards its min/max. Build then has no pending base-typed chunks; `train_quantizer_if_needed()` deliberately leaves the index quantizer untrained. Consequently the advertised base-typed `SearchQuantize([]float32, ...)` path fails at `quantize_query()` with `quantizer not trained`. Train and retain the index's B-source quantizer before encoding the file, or return and install the exact range used by the loader.

### P1 - Fix the training sampler that still ignores the tail of 501–999 row files (cgo/cuvs/quantize.hpp:331)
`n_train` is capped at 500, but `stride` uses floor division. For every file with 501–999 rows, `stride == 1`, so the loop samples only rows 0–499 despite claiming to sample across the whole file. A sorted or clustered MODF file therefore trains its int8/uint8 range from the prefix and saturates all later values, severely degrading the newly added quantized data-file index. Select sample indices proportionally over `[0, n_rows-1]` and add boundary tests for 501 and 999 rows.

XuPeng-SH (changes requested):
Deep review of head `299de48ecdaa514a27efab66608dacc2d3c1902f`: requesting changes for five merge-blocking issues.

1. **[P1] Reject affine int8/uint8 quantization for non-L2 IVF-FLAT metrics**

`ValidQuantization` ignores `op_type` and accepts int8/uint8 for inner-product and cosine indexes. The transform `q(x)=a*x+b` does not preserve either geometry, so these accepted indexes can return incorrect rankings and scores. CAGRA and IVF-PQ already enforce the necessary L2-only restriction. Please apply the same gate to IVF-FLAT or rerank from the original vectors.

https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/vectorindex/ivfflat/plugin/runtime/runtime.go#L145-L158

2. **[P1] Convert IVF-FLAT quantized L2 scores back to source scale**

The query and entries are evaluated in the quantized domain, but the returned distance only receives the existing squared-L2-to-L2 conversion. It never applies the inverse quantizer scale. For `[min,max]=[0,1]`, `mul=255`, so a source L2 distance of `0.5` is reported near `128`. This also breaks distance-range predicates expressed in source units. Please rescale squared L2 by `1/mul²` before exposing the score and add returned-distance/range-filter regressions.

https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/vectorindex/ivfflat/search.go#L349-L357
https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/vectorindex/ivfflat/search.go#L437-L439

3. **[P1] Bound host memory during quantized CAGRA/IVF-PQ builds**

`add_chunk_quantize` retains every source chunk without applying the existing training threshold. The production builders call it one row at a time. At `Build`, `flush_pending_float_chunks_internal` copies all retained vectors again into `all_floats`, while the quantized destination is already preallocated. A 1M×768 f32→int8 build therefore peaks around 3.1 GB pending source + 3.1 GB duplicated training data + 0.77 GB destination, plus one million chunk/ID allocations. This can OOM an otherwise viable build. Please use bounded/streaming training and batch ingestion.

https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/cgo/cuvs/index_base.hpp#L1107-L1126
https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/cgo/cuvs/index_base.hpp#L1402-L1419

4. **[P1] Reject narrow vectors in inline primary-key declarations**

The inline DDL check rejects only vecf32/vecf64, so `CREATE TABLE t (v vecint8(2) PRIMARY KEY)` is accepted. The transaction duplicate checker has no narrow-vector case and reaches its default panic during insert/commit. Please use `types.T(colType.Id).IsArrayRelate()` consistently across inline, table-level, and ALTER primary-key admission, with rejection tests for all four new types.

https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/sql/plan/build_ddl.go#L1250-L1265
https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/vm/engine/disttae/txn.go#L401-L403

5. **[P1] Do not return non-unit integer vectors from `normalize_l2`**

The implementation normalizes in float32 and then narrows back to int8/uint8. The committed golden result demonstrates that `normalize_l2([0,1,2,3]::vecuint8)` becomes `[0,0,1,1]`, whose norm is √2. For integer inputs, please return vecf32 or reject the overload and require an explicit cast; a same-type integer result cannot satisfy the function contract.

https://github.com/matrixorigin/matrixone/blob/299de48ecdaa514a27efab66608dacc2d3c1902f/pkg/sql/plan/function/func_unary.go#L308-L316

Focused tests pass for types, comparison, sorting, SQL functions, quantizer, and all IVF-FLAT/IVF-PQ/CAGRA plan/compile/runtime plugin packages. GPU execution was unavailable because the NVIDIA driver is inaccessible, so the C++ lifecycle finding is based on static ownership/boundedness analysis. The earlier comparator-allocation and discarded-quantizer findings are fixed at this head.

cgo/cuvs/index_base.hpp
```diff
@@ -1364,8 +1451,49 @@ class gpu_index_base_t {
...
+                if (is_loaded_) throw std::runtime_error("Cannot add chunk to built index");
+            }
+            pending_float_chunk_t c;
+            c.data.assign(chunk_data, chunk_data + chunk_count * dimension);
+            c.count  = chunk_count;
+            c.offset = offset;
+            if (ids) c.ids.assign(ids, ids + chunk_count);
+            std::unique_lock<std::shared_mutex> lock(mutex_);
+            pending_total_count_ += chunk_count;
+            pending_float_chunks_.push_back(std::move(c));
```
XuPeng-SH: [P1] `quantizer_train_limit` does not bound this buffer: every base-typed row is still copied into `pending_float_chunks_` and kept until `Build()`. The production CAGRA/IVF-PQ builders call `AddChunkQuantize(..., 1, ...)` once per row, so a 1M x 768 f32-to-int8 build retains ~3.07 GB of source vectors here, then also allocates the int8 destination, the bounded training sample, IDs, and about one million per-row vector allocations. Removing `all_floats` eliminated one duplicate, but the extra source memory remains O(N * sizeof(B)) and can still OOM. Please use a two-pass or bounded reservoir/spool design so only O(N * sizeof(Q)) index storage stays resident, batch ingestion, and add a representative peak-memory/allocation regression.

pkg/cuvs/multi_index.go
```diff
@@ -269,6 +346,137 @@ func multiGpuSearch[T VectorType](
...
+	}
+
+	if bruteForce != nil {
+		// Guard against a dispatch mismatch: if the overflow brute force is
+		// live but neither a base-typed (B) nor an f32 query was supplied, the
+		// async search would submit an empty job (job id 0) and SearchWait(0)
+		// would block forever. Fail loudly instead — this means the [B,Q]
+		// instantiation disagrees with the decoded query type.
+		if len(queriesB) == 0 && len(queriesBF32) == 0 {
+			return nil, nil, moerr.NewInternalErrorNoCtx("multiGpuSearchBQ: brute force is loaded but no base/f32 query was provided (B/Q dispatch mismatch)")
```
XuPeng-SH: [P1] This error is checked only after lines 417-431 have submitted all main-index async jobs. For the documented `B != Q` + live-overflow path, returning here means none of those job IDs reaches `SearchWait`; `cuvs_task_result_store_t::store` then retains each completed result until `wait()`/`discard()` or worker shutdown. Repeated calls to this public error path therefore grow the result map without bound. Please validate the mismatch before any submission and also guarantee that every later partial-dispatch/wait error drains or discards all previously submitted jobs; add a failure-path regression that asserts zero outstanding jobs.

XuPeng-SH (changes requested):
Deep review of current head `5341d09527b74c39aa69da89998cf3c4e4761a7e`: requesting changes for two remaining merge blockers.

The previous IVF-FLAT non-L2 gate, quantized-score rescaling, narrow-vector primary-key rejection, and integer `normalize_l2` issues are fixed. The quantized-build memory fix is only partial: it removed the duplicate full training copy, but the build still retains every source vector until `Build()`. I also found a deterministic async-result leak on the newly added B/Q mismatch path.

1. **[P1] Quantized CAGRA/IVF-PQ builds still retain the full base-typed table in host memory.** `quantizer_train_limit` bounds only the separate training sample. `add_chunk_quantize` still stores every `B` row in `pending_float_chunks_`; production `AddRow` calls it one row at a time. A 1M x 768 f32-to-int8 build therefore retains about 3.07 GB of raw vectors plus the 0.77 GB quantized destination, a 0.31 GB default sample, IDs, allocator overhead, and roughly one million tiny chunk allocations. This is still O(N * sizeof(B)) extra memory and can OOM a build that should fit in int8 storage.

2. **[P1] The B/Q dispatch-mismatch error abandons already-submitted search jobs.** `multiGpuSearchBQ` submits every main-index async search first, then returns from the overflow guard when `B != Q`. Completed jobs remain in `cuvs_task_result_store_t::results` until `wait()` or `discard()`, neither of which is called on this return. Repeating the documented error path grows retained search results for the worker lifetime. Validate before dispatch and make every partial-dispatch error path drain or discard submitted jobs.

Focused CPU tests pass for types/vector/compare/sort, quantizer/metric, IVF-FLAT, SQL parser/compile/plan/functions, brute-force/cuvs helpers, ISCP, and IVF-FLAT/IVF-PQ/CAGRA plugin compile/plan/runtime/idxcron. The GPU-only CAGRA/IVF-PQ packages could not be executed because this host has no working NVIDIA driver. Current GitHub UT/BVT/SCA/Coverage jobs are still pending.

XuPeng-SH (changes requested):
Deep review of the full current head `abaecb7f8d858de843ac4af7fee9151ceceadb40` — not limited to the prior findings. The previous full-table staging and partial-dispatch result-store leaks are closed, but I found three independent merge blockers.

1. **[P1] Do not train the 1-byte quantizer from an unproven arrival-order prefix.**

The new bounded-memory path trains on the first `quantizer_train_limit` arriving rows and explicitly assumes they arrive in random order. Production does not provide that contract: both CAGRA and IVF-PQ build with a plain source-table `SELECT ... CROSS APPLY ...` (no sample/shuffle), and the table functions call `AddRow` one row at a time in scan arrival order. A table clustered by time/ID/magnitude can therefore train on `[0,1]` in the prefix and clamp a later `[100,101]` population to the same int8/uint8 extremes, silently collapsing recall. This is also a regression of the earlier sorted-input fix: the old adversarial `StridedQuantizerTrainSamplesTail` test was replaced by a test that shuffles the input first, so it now bakes the disputed assumption into the test instead of validating the production contract.

This cannot be fixed by a final reservoir alone after rows have already been encoded with an earlier range. The build needs a real two-phase contract (bounded uniform pre-sample/train, then replay/stream with a fixed range), a spill/replay path, or an explicit builder-side range prepass. Please add an end-to-end magnitude-ordered source regression with `row_count > quantizer_train_limit`.

- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/index_base.hpp#L410-L420
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/index_base.hpp#L1313-L1328
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/cagra/plugin/compile/compile.go#L40-L45
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/table_function/cagra_create_gpu.go#L496-L504
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/test/ivf_pq_test.cu#L164-L212

2. **[P1] Complete the core varlen/type-dispatch closure for the four new first-class SQL types.**

The PR adds `vecbf16/vecf16/vecint8/vecuint8` as storage-visible SQL OIDs, but several generic paths still enumerate only vecf32/vecf64. These are not merely unsupported optional functions:

- generic CDC fails the stream on the first narrow-vector row in both extraction and SQL serialization;
- `ALTER TABLE ... ADD v VECF16(n) NOT NULL` synthesizes `null` instead of the zero vector used for vecf32/vecf64;
- reservoir `sample(...)` returns `unsupported type for sample pool` when replacement touches a narrow vector;
- `FILL` reaches its default panic for the same varlen payload.

Please audit the semantic “all vector/all varlen” switches systematically (prefer `IsArrayRelate()` or a shared helper where appropriate) and add unhappy-path tests for each new OID, rather than patching only the index-specific happy paths.

- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/cdc/util.go#L168-L177
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/cdc/util.go#L294-L309
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/plan/build_alter_table.go#L837-L847
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/sample/poolData.go#L175-L183
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/fill/fill.go#L760-L767

3. **[P1] Reject or correctly rescale IVF-FLAT L1 with int8/uint8 quantization.**

`ValidQuantization` says the affine 1-byte quantizer is L2-only but rejects only IP/cosine, so `vector_l1_ops` is accepted. Search then unconditionally divides every quantized score by `QuantMul²`. For L1, `|q(a)-q(b)| ≈ QuantMul*|a-b|`, so the correct inverse factor is `1/QuantMul`, not `1/QuantMul²`. With `[min,max]=[0,1]`, a source L1 distance near `0.5` is exposed near `0.5/255`. Ordering within the quantized entry table may remain monotonic, but returned distances and distance predicates are silently wrong. Either truly gate 1-byte quantization to L2/L2sq or implement metric-specific inverse scaling with L1 regressions.

- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/ivfflat/plugin/runtime/runtime.go#L157-L173
- https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/ivfflat/search.go#L446-L462

Validation: focused CPU tests passed for types/vector/compare/sort, parser/plan/functions/frontend/CDC/ISCP/external/TAE, all vector-index plugins, SIMD AVX paths and forced scalar fallback; race passed for cache and shared cuVS helpers. GPU runtime could not be executed because the local NVIDIA driver is unavailable. Current CI's only UT failure is `TestAppendAndGC2` (checkpoint/GC timing); this PR does not touch that package/path and the exact test passed locally 3/3 under `-race`, so I treat it as an unrelated flake rather than one of the findings above.

cpegeric:
> Deep review of the full current head `abaecb7f8d858de843ac4af7fee9151ceceadb40` — not limited to the prior findings. The previous full-table staging and partial-dispatch result-store leaks are closed, but I found three independent merge blockers.
> 
> 1. **[P1] Do not train the 1-byte quantizer from an unproven arrival-order prefix.**
> 
> The new bounded-memory path trains on the first `quantizer_train_limit` arriving rows and explicitly assumes they arrive in random order. Production does not provide that contract: both CAGRA and IVF-PQ build with a plain source-table `SELECT ... CROSS APPLY ...` (no sample/shuffle), and the table functions call `AddRow` one row at a time in scan arrival order. A table clustered by time/ID/magnitude can therefore train on `[0,1]` in the prefix and clamp a later `[100,101]` population to the same int8/uint8 extremes, silently collapsing recall. This is also a regression of the earlier sorted-input fix: the old adversarial `StridedQuantizerTrainSamplesTail` test was replaced by a test that shuffles the input first, so it now bakes the disputed assumption into the test instead of validating the production contract.
> 
> This cannot be fixed by a final reservoir alone after rows have already been encoded with an earlier range. The build needs a real two-phase contract (bounded uniform pre-sample/train, then replay/stream with a fixed range), a spill/replay path, or an explicit builder-side range prepass. Please add an end-to-end magnitude-ordered source regression with `row_count > quantizer_train_limit`.
> 
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/index_base.hpp#L410-L420
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/index_base.hpp#L1313-L1328
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/cagra/plugin/compile/compile.go#L40-L45
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/table_function/cagra_create_gpu.go#L496-L504
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/cgo/cuvs/test/ivf_pq_test.cu#L164-L212
> 
> 2. **[P1] Complete the core varlen/type-dispatch closure for the four new first-class SQL types.**
> 
> The PR adds `vecbf16/vecf16/vecint8/vecuint8` as storage-visible SQL OIDs, but several generic paths still enumerate only vecf32/vecf64. These are not merely unsupported optional functions:
> 
> * generic CDC fails the stream on the first narrow-vector row in both extraction and SQL serialization;
> * `ALTER TABLE ... ADD v VECF16(n) NOT NULL` synthesizes `null` instead of the zero vector used for vecf32/vecf64;
> * reservoir `sample(...)` returns `unsupported type for sample pool` when replacement touches a narrow vector;
> * `FILL` reaches its default panic for the same varlen payload.
> 
> Please audit the semantic “all vector/all varlen” switches systematically (prefer `IsArrayRelate()` or a shared helper where appropriate) and add unhappy-path tests for each new OID, rather than patching only the index-specific happy paths.
> 
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/cdc/util.go#L168-L177
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/cdc/util.go#L294-L309
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/plan/build_alter_table.go#L837-L847
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/sample/poolData.go#L175-L183
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/sql/colexec/fill/fill.go#L760-L767
> 
> 3. **[P1] Reject or correctly rescale IVF-FLAT L1 with int8/uint8 quantization.**
> 
> `ValidQuantization` says the affine 1-byte quantizer is L2-only but rejects only IP/cosine, so `vector_l1_ops` is accepted. Search then unconditionally divides every quantized score by `QuantMul²`. For L1, `|q(a)-q(b)| ≈ QuantMul*|a-b|`, so the correct inverse factor is `1/QuantMul`, not `1/QuantMul²`. With `[min,max]=[0,1]`, a source L1 distance near `0.5` is exposed near `0.5/255`. Ordering within the quantized entry table may remain monotonic, but returned distances and distance predicates are silently wrong. Either truly gate 1-byte quantization to L2/L2sq or implement metric-specific inverse scaling with L1 regressions.
> 
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/ivfflat/plugin/runtime/runtime.go#L157-L173
> * https://github.com/matrixorigin/matrixone/blob/abaecb7f8d858de843ac4af7fee9151ceceadb40/pkg/vectorindex/ivfflat/search.go#L446-L462
> 
> Validation: focused CPU tests passed for types/vector/compare/sort, parser/plan/functions/frontend/CDC/ISCP/external/TAE, all vector-index plugins, SIMD AVX paths and forced scalar fallback; race passed for cache and shared cuVS helpers. GPU runtime could not be executed because the local NVIDIA driver is unavailable. Current CI's only UT failure is `TestAppendAndGC2` (checkpoint/GC timing); this PR does not touch that package/path and the exact test passed locally 3/3 under `-race`, so I treat it as an unrelated flake rather than one of the findings above.

1. the build scan runs parallel readers, so the first 100k rows arriving are interleaved from N different scan ranges, not the head of one. That spreads the sample across the table even when storage order correlates with value. Two passes doesn't help since someone has to allocate that huge memory that never exists.  Your agent proposal is ridiculous.  Current design is the optimal and valid for utilizing GPU memory without explode the memory.  Task 1 is settled and proposal from agent is rejected.  You can always set quantizer_train_limit to have more data for quantization.  Nothing stop you except your physical GPU memory size.

2 and 3 fixed.

Ariznawlll:
## PR #25095 测试覆盖分析

⚠️ 无法获取 PR diff。

---
*由 AI Test Analyzer 自动生成 · [workflow run](https://github.com/Ariznawlll/mo-test-coverage-bot/actions/runs/29950059093)*

XuPeng-SH (changes requested):
Request changes — reviewed exact head `a1ec0e57ec1ccf38a1ecfc70b92913db86906f2a`.

**[P1] Prefix-only quantizer training still silently collapses valid ordered builds.**

The production CAGRA/IVF-PQ table functions accept `quantizer_train_limit` from SQL and feed the builder one row at a time. Once that many arrivals have been staged, `flush_pending_float_chunks_internal` trains the scalar quantizer only on that prefix (`index_base.hpp:1362-1371`), and every later value outside the learned range is clamped to the same endpoint (`quantize.hpp:169-178`).

This is reachable without relying on a hypothetical engine behavior. The new comment records that a 20k-row value-ordered load stayed in file order, with the first 1000 arrivals covering `[1,1000]` (`index_base.hpp:447-449`). `CREATE INDEX ... QUANTIZATION 'int8' QUANTIZER_TRAIN_LIMIT 1000` is a supported configuration. For vectors whose coordinates increase with the row id, the learned max is therefore 1000 and every coordinate from rows 1000 through 20000 quantizes to 127. Those tail vectors become indistinguishable, so tail nearest-neighbor results/recall are silently wrong.

The absence of `ORDER BY` does not make the prefix random; it explicitly permits storage order, as the comment itself notes. Scheduler interleaving observed for one 1M-row layout is not an invariant, and the wiki recall measurement is explicitly on a non-value-ordered dataset, so neither protects this case. A configurable sampling budget is fine, but the sample must represent the full build (at a caller layer that can replay/spool/two-pass, or via an explicit/proven range); documenting the bias does not make the collapse safe. Please add a regression with ordered data, a limit below the row count, and tail probes that verify distinct nearest neighbors.

I rechecked the previous blockers on this head: the narrow generic paths are now covered, and IVF-FLAT int8/uint8 is allowlisted to L2/L2sq with effective REINDEX parameters merged before validation. Those are closed.

Validation performed: targeted CPU/CGo tests passed for types/vector/sort/compare, quantizer and distance kernels, IVF-FLAT, CAGRA/IVF-PQ/IVF-FLAT plugin hooks, planner/functions, CDC, external I/O, ALTER/REINDEX, table functions, frontend, and TAE narrow-type paths. The index-plugin architecture guards are clean. The current GitHub UT red is the unchanged `TestAppendAndGC2` (`ExpectedEOB`); coverage's full UT run passed. GPU runtime validation was unavailable because this host has no working NVIDIA driver (and its active conda env lacks cuVS/RAFT headers).

gouhongshen (changes requested):
## Codex automated review

Two blocking correctness regressions remain in quantized builds and export support for the new vector types.

### P1 - Do not train the scalar quantizer from an arrival-order prefix (cgo/cuvs/index_base.hpp:1362)
The bounded path trains directly from the first staged rows, then encodes every later row with that fixed range. CAGRA/IVF-PQ production builds feed one row at a time, while their source SELECT has no ordering or random-sampling contract. The PR itself records that a 20k-row value-ordered load stayed in file order, yet QUANTIZER_TRAIN_LIMIT accepts smaller positive values (for example 1000). Thus an ordered 20k build can train on [1,1000]; transform_host clamps every later value above the learned max to the same endpoint (quantize.hpp:175-179), collapsing distinct tail vectors and silently damaging recall. Parallel interleaving observed for one large layout does not establish a correctness contract, and the added regression shuffles arrival order, so it only validates the disputed assumption. Train from a sample proven to cover the full build before encoding, or otherwise make this configuration fail safely; add an ordered-source regression with row count above the configured limit and tail nearest-neighbor probes.

### P1 - Complete the batch CSV and JSONL export dispatch for narrow vectors (pkg/frontend/export.go:491)
The PR adds handling only to exportDataFromResultSetToCSVFile, but the normal batch export route dispatches CSV through constructByte (lines 942-950). Its type switch handles vecf32/vecf64 at lines 491-498 and sends vecbf16/vecf16/vecint8/vecuint8 to the default error at lines 547-563. JSONL is likewise selected through constructJSONLine and vectorValueToJSON, which only handles vecf32/vecf64 at lines 1224-1227 before returning an unsupported-type error. Consequently SELECT of any newly supported narrow-vector column INTO OUTFILE fails for CSV and JSONL. Add all four OIDs to both switches with array-preserving serialization (especially avoiding raw-byte treatment of uint8), and cover the batch CSV and JSONL routes.

cpegeric:
Re: [P1] prefix-only quantizer training — contesting the blocking status
     
  The mechanism is correctly described, but the conclusion treats a user-controlled sampling budget as a correctness defect. quantizer_train_limit is
  the "explicit/proven range" the review asks for:
  
  - It drives both staging and training — want = min(quantizer_train_limit_, INT64_MAX) (index_base.hpp:1200), then capped only by free VRAM
  (cap_train_rows_to_gpu_mem). Setting it ≥ nrow (within GPU memory) stages and trains on the whole dataset, so [min,max] is exact and the tail cannot
  clamp. Nothing prevents a user from doing this.
  - The default is already 100000, so every table up to 100k rows trains on all of its rows out of the box. The 20k example only degrades because the
  review's config sets the limit to 1000 — deliberately below the row count. That is the user asking for a 20× sample, and it behaves like one. A
  smaller sample being less representative than a larger one is the meaning of the knob, not a silent bug.
  
  So the design loses nothing relative to the "must be representative" requirement: the exact-range behavior is available on request via a single
  parameter, on a single pass — no replay/spool/two-pass. What the review implicitly demands instead — always accumulate the entire dataset just to
  derive a scale+offset — is precisely the full-table staging this PR replaced (~3 GB → ~0.31 GB host, recall within run-to-run noise). Reintroducing
  that as a mandatory cost, to defend a case the user can already avoid by raising one limit, is a strict regression for no real-world gain.
  
  On reachability of the failure itself: it additionally requires coordinates that increase monotonically with arrival order (value ≈ row_id). Real ANN
  inputs have per-dimension ranges that are stable across rows regardless of storage order, which is why a file-order prefix — and the parallel-ingest
  interleaving measured at 1M (first 200k arrivals span 93% of the value range) — is representative in practice, and why the wiki recall holds. The
  monotonic-coordinate dataset is synthetic; combined with an explicit sub-nrow limit, it describes a user who has opted out of a representative sample
  on data that doesn't occur in vector search.
  
  If the concern is visibility rather than capability, the proportionate change is a warning (or a note in SHOW CREATE) when QUANTIZER_TRAIN_LIMIT < 
  nrow, so a user lowering the budget below the table size is told the sample may not span the full range. I'm happy to add that. But "set the limit
  high for an exact range, low to bound host memory" is a sound, single-pass design — not a P1 blocker.
  
  The prior blockers are closed on this head: narrow generic paths covered; IVF-FLAT int8/uint8 allowlisted to L2/L2sq with REINDEX params merged
  pre-validation.

XuPeng-SH (changes requested):
Request changes — re-reviewed exact head `858aa01ce0461ad6b81890e6d28b3b410f2698d0` and reconsidered the latest author comments.

I am withdrawing the previous prefix-training blocker. `quantizer_train_limit` is an explicit quality/memory budget, the default covers up to 100k rows, setting it at least to the row count restores whole-table training when memory permits, and mandatory full-table retention/replay would undo this PRs main bounded-memory benefit. The ordered-input case remains a quality caveat worth documenting, but it is not a P1 correctness blocker under this contract. I also verified that the generic narrow-type closure, IVF-FLAT L2/L2sq gate, result rescaling, full-table staging, async result draining, and the latest batch CSV/JSONL export fix are closed.

One independent merge blocker remains:

### [P1] Make IVF-FLAT int8/uint8 encoding bit-identical across sync build, CDC, and search

The stated invariant that entry and query use the same `q(x)=round(x*mul+add)` is not true at float32 boundaries.

The synchronous build serializes `mul` and `add` with only 9 significant digits in `Int8EntrySQL`/`Uint8EntrySQL`. For a vecf32 base, the arithmetic type rules then coerce those literals to float32 and the engine rounds once after multiplication and again after addition. Search instead keeps the parameters as float64, performs one float64 multiply-add in `ApplyInt8`/`ApplyUint8`, and rounds only the final float32. For a vecf64 base there is another asymmetry: the entry expression applies the transform to f64, while `runIvfSearchVectorToF32` first discards the querys f64 precision. The CDC expression is independently re-derived in SQL, so there is currently no single canonical rounding contract across the three writers/readers.

Concrete reachable reproduction for int8: trained bounds `[0.1, 0.99]` produce `mul=286.51685393258424`, `add=-156.65168539325842`. For a stored vecf32 component `0.367`, synchronous entry SQL is emitted as `cast(v * 286.516854 + (-156.651685) as vecint8(...))`; with the actual vecf32 scalar-arithmetic rules its intermediate is exactly `-51.5`, so it stores `-52`. Search receives the same float32 value but `ApplyInt8` produces `-51.499996`, so it sends `-51`. A nearby row `0.3670001` is stored as `-51`; therefore an exact query for `0.367` can give the nearby row zero quantized distance while its own row has non-zero distance. Repeating the boundary component across dimensions magnifies the error. Uint8 has the same defect.

Please define one canonical precision and operation ordering, use it for synchronous entries, CDC entries, and query encoding, and add cross-path boundary regressions for f32 and f64 bases with both int8 and uint8. The regression should compare actual entry-expression execution against query encoding, not only test the SQL string and Go helper separately.

Relevant code:
- https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/vectorindex/quantizer/quantizer.go#L148-L179
- https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/vectorindex/quantizer/quantizer.go#L216-L245
- https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/sql/plan/function/type_check.go#L1197-L1221
- https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/sql/colexec/table_function/ivf_search.go#L272-L289

Non-blocking documentation cleanup: `cagra_create_gpu.go` and `ivfpq_create_gpu.go` still say the C++ default is 1000, and both quantizer-train-limit GPU cases still describe strided sampling/default 1000. The implementation/header now use prefix staging/default 100000; these comments should match the contract relied upon in the author response.

Validation: focused CPU/CGo tests passed across changed container, planner/function, frontend, CDC/ISCP, colexec, all CPU-visible vector-index plugin, disttae, and TAE packages. Targeted race tests passed for types/vector/cache/cuvs/ivfflat/quantizer. Current GitHub UT/SCA/BVT/Coverage checks are green. GPU-only packages cannot execute on this Darwin host and were audited statically.

jiangxinmeng1 (changes requested):
Reviewed the current head (858aa01ce0461ad6b81890e6d28b3b410f2698d0). The earlier
  issues appear resolved, but one merge blocker remains.

  ### [P1] Use identical int8/uint8 encoding across build, CDC, and search

  The three paths do not currently implement a bit-identical q(x) = round(x * mul + add)
  operation:

  - The synchronous build serializes mul and add with only 9 significant digits. For
    vecf32, SQL arithmetic also rounds after multiplication and again after addition.

  - Search performs the multiply-add in float64, converts the final result to float32,
    and only then rounds to int8/uint8.

  - The CDC path independently reconstructs the expression in SQL.
  - For a vecf64 base, build quantizes the original float64 value, while search first
    downcasts the query to float32.

  Values near a bucket boundary can therefore be stored in one bucket but encoded into
  another during search. An exact query may have a non-zero distance from its own row
  while a nearby row has zero distance.

  Please define a single precision and operation-ordering contract and reuse it for
  synchronous build, CDC, and query encoding. Add end-to-end boundary tests for both
  float32 and float64 bases with int8 and uint8 storage.

  Relevant code:

  -
  https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/vectorindex/quantizer/quantizer.go#L148-L179
  -
  https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/vectorindex/quantizer/quantizer.go#L216-L245
  -
  https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/sql/plan/function/type_check.go#L1197-L1221
  -
  https://github.com/matrixorigin/matrixone/blob/858aa01ce0461ad6b81890e6d28b3b410f2698d0/pkg/sql/colexec/table_function/ivf_search.go#L272-L289

cpegeric:
fixed. please review again.

XuPeng-SH (changes requested):
Request changes — deep-reviewed exact head `1b66d898ce6d9c73fd34267a34ff834bf80c2e81` and the latest “fixed” comment.

### [P1] Explicitly round the float32 product; a temporary variable does not prevent FMA

`ApplyInt8` and `ApplyUint8` claim that `m := x * fmul` forces the product to be rounded before the add. That is not a Go guarantee. The [Go spec](https://go.dev/ref/spec#Floating-point_operators) explicitly permits fusing `t = x*y; r = t+z` across statements; an explicit conversion is required to prevent fusion. This is directly reachable in MatrixOne because the Makefile defaults `GOAMD64` to `v3`, and the current compiler emits `VFMADD231SS` for these loops.

With trained bounds `[0.1, 0.99]`:
- int8, `x = math.Float32frombits(0x3de5d15a)` (`0.11221571`): SQL's separate float32 operations produce `-124.5` and encode `-125`; the v3 query encoder fuses to `-124.49999` and encodes `-124`.
- uint8, `x = math.Float32frombits(0x3dd05fbc)` (`0.1017451`): SQL produces `0.5 -> 1`; fused query encoding produces `0.49999908 -> 0`.

A focused regression passes with `GOAMD64=v1` and fails with the repository-default `GOAMD64=v3` (`expected -125, actual -124`). Therefore build/CDC and search are still not bit-identical, and an exact query can be assigned to the adjacent bucket. The new `TestQueryEntryEncodingContract` masks this because `entryInt8`/`entryUint8` use the same fusible temporary-variable pattern as production, so both sides fuse together.

Please force the product rounding with an explicit conversion (for example `m := float32(x * fmul)`) or define the SQL path to use the identical fused contract. The reference test must model actual SQL's separate operator execution, and should include these int8/uint8 FMA-boundary cases under `GOAMD64=v3`.

I rechecked the related discussion: the `quantizer_train_limit` explanation is reasonable and remains non-blocking; the prior generic-type, IVF-FLAT metric/scale, full-staging, async-drain, and export blockers are closed. Focused v3 tests for quantizer, IVF-FLAT, compile hooks, and the IVF search table function pass, but they do not exercise this FMA boundary.

cpegeric:
Re: [P1] explicitly round the float32 product — fixed in 6cba7fe298
     
  You're right on all counts — a temp var is not an FMA barrier, and with GOAMD64 ?= v3 (VFMADD231SS) / arm64 hardware FMA the multiply-add fused to
  one rounding while the entry SQL rounds twice. I reproduced your exact cases:
  
  - x = 0x3de5d15a (0.11221571), int8: fused viaVar = -124.49999 → -124; explicit float32(x*fmul) = -124.5 → -125; SQL confirms -125 on the live
  engine.
  - x = 0x3dd05fbc (0.1017451), uint8: fused 0.49999908 → 0 vs SQL 0.5 → 1.
  
  Fix: wrap the product in an explicit float32() conversion in ApplyInt8/ApplyUint8 — the Go spec's guaranteed rounding barrier — so build/CDC/search
  all round the product before the add:
  
  m := float32(x * fmul)   // explicit conversion rounds the product; blocks FMA fusion
  sq[i] = m + fadd
  
  Test fixes (you correctly caught the masking):
  - entryInt8/entryUint8 now use the same explicit-conversion form, so they model SQL's two separate operations instead of fusing alongside production.
  - Added TestApplyFMABoundaryMatchesSQL: an independent oracle using your exact bit patterns with the hardcoded, live-verified SQL codes (-125, 1), so
  it catches a fused encoder regardless of any Go-side replication. It passes on arm64 hardware FMA (≡ GOAMD64=v3); the fused version fails it
  (expected -125, actual -124).
  
  Closure verified: ApplyInt8/ApplyUint8's only caller is the IVF-FLAT query re-rank (ivfflat/search.go), and build/CDC write via the SQL entry
  expression (two materialized vecf32 ops), so all three paths now round identically.

gouhongshen (changes requested):
## Codex automated review

The FMA barrier fixes the latest reported case, but synchronous int8/uint8 build encoding can still differ from CDC/search because its float64 parameters are formatted before narrowing to float32.

### P1 - Narrow quantizer parameters before serializing entry SQL (pkg/vectorindex/quantizer/quantizer.go:199)
`%.9g` guarantees round-trip only when formatting a float32 value, but these builders format the original float64 `mul`/`add`. The emitted decimal can therefore parse to an adjacent float32 from `float32(mul)` used by `ApplyInt8`/`ApplyUint8`. Concrete reachable example: stored bounds `[0, 7.75195541]` yield `mul=32.89492605582467`; search narrows this to float32 bits `0x42039467`, while entry SQL emits `32.8949261`, which parses to `0x42039468`. For component bits `0x3ea36df5` (`0.31919828057289124`), the separate float32 operations produce `-117.5 -> -118` in search but `-117.499992 -> -117` in the synchronous int8 entry; uint8 similarly differs by one bucket. CDC computes the parameter from bounds in SQL and therefore follows the search-side narrowing, so rows can also change buckets depending on whether they arrived during build or through CDC. Format explicitly narrowed parameters (including `add`) and add a regression where formatting the float64 directly crosses a float32 boundary.

XuPeng-SH (changes requested):
Request changes — deep-reviewed exact head `6cba7fe2984141144b2160d582cacff3cf4fd4ee`, including the latest review and the author's FMA response.

The explicit `float32(x * fmul)` conversion is correct and closes the previous FMA-fusion blocker. One independent deterministic correctness issue remains:

### [P1] Narrow `mul`/`add` before formatting synchronous entry SQL

`Int8EntrySQL` and `Uint8EntrySQL` use `%.9g` on the original `float64` parameters. Nine significant digits round-trip a *float32 value*, but formatting an adjacent `float64` first can emit a decimal that narrows to the neighboring float32 from `float32(mul)`/`float32(add)` used by search.

I independently reproduced the current head with stored bounds `[0, 7.75195541]`:

- `mul=32.89492605582467`
- search parameter: `float32(mul)` = bits `0x42039467`
- synchronous build emits `32.8949261`, which the vecf32 scalar-arithmetic rule narrows to bits `0x42039468`
- int8 component `0x3ea36df5` (`0.31919828`): search encodes `-118`, synchronous build encodes `-117`
- uint8 component `0x3c79090d` (`0.015199912`): search encodes `0`, synchronous build encodes `1`

This is reachable because `type_check.go` coerces a scalar paired with `VECF32` to `FLOAT32`. Search reads the metadata bounds and narrows the computed parameters directly; CDC computes them from the same bounds in SQL and is narrowed by that rule. Only synchronous build serializes the pre-narrowed float64 values, so initial rows and CDC rows can occupy adjacent buckets for the same source value.

Please format explicitly narrowed values for both parameters (for example, serialize `float32(mul)` and `float32(add)` with float32 round-trip semantics) and add a regression that parses/executes the generated SQL at a float32-format boundary and compares it with `ApplyInt8`/`ApplyUint8` for both int8 and uint8.

Related-comment disposition: the FMA comment is correctly fixed; the `quantizer_train_limit` explanation remains a non-blocking quality/memory contract; the prior generic-type, IVF-FLAT metric/scale, full-staging, async-drain, and export findings remain closed.

Validation on this head: quantizer tests and `-race` passed under `GOAMD64=v3`; focused IVF-FLAT build/search hooks and narrow-vector function tests passed. A one-off exact regression reproduces the adjacent-bucket mismatch, which the current tests do not model because they reuse the in-memory float64 parameters instead of reparsing the generated SQL literals.

cpegeric:
Re: [P1] narrow mul/add before formatting entry SQL — fixed in bea21bc226
     
  Correct, and thanks for the second reachable component — reviewed head 6cba7fe298 was the FMA-fix commit, one before this. I reproduced both cases
  exactly:
  
  - int8, bounds [0, 7.75195541], mul=32.89492605582467: float32(mul)=0x42039467, but %.9g(float64)="32.8949261" narrows to 0x42039468; at component
  0x3ea36df5 search -118 vs build -117.
  - uint8, component 0x3c79090d: search 0 vs build 1.
  
  Fix (Int8EntrySQL/Uint8EntrySQL): format the float32-narrowed params so the literal round-trips to the exact float32 search/CDC use:
  fmt.Sprintf("... %.9g + (%.9g) ...", f32col, float32(mul), float32(add), dim)
  286.516854 → 286.516846, etc. The *FromBounds (CDC) builders compute params in SQL from the same bounds and are narrowed by the VECF32 <op> Scalar
  rule, so they already match — only the literal-bounds builders changed.
  
  Regression (TestEntrySQLParamNarrowedToFloat32) — exactly your ask, for both int8 and uint8: it parses the generated SQL literals, asserts the
  precondition (raw-float64 %.9g crosses a float32 boundary for this mul), that the emitted literals round-trip to float32(mul)/float32(add), and that
  build (emitted param) and search (ApplyInt8/ApplyUint8) encode the reported boundary components to the same bucket. Passes under -race/GOAMD64=v3;
  the existing builder-string tests now match the narrowed literals.
  
  Closure: build (literal params, now narrowed), CDC (SQL-computed, narrowed by coercion), and search (float32(mul) in ApplyInt8) all resolve to the
  identical float32(mul) and round the product identically.

gouhongshen (changes requested):
## Codex automated review

The merge from main preserved the PR’s narrow-vector integrations but did not touch any of the three prior findings. All remain reproducible on head 32d014f7a.

### P1 - Do not globally raise the default x86 CPU baseline to GOAMD64=v3 (Makefile:247)
Before this PR, `make build` used Go's x86-64-v1 default. This assignment compiles every package in mo-service, mo-tool, musl builds, and repository Docker artifacts for x86-64-v3, so binaries that previously ran on v1/v2 CPUs will no longer start. Runtime feature checks inside metric kernels cannot protect a binary globally compiled for v3, and default builds do not pass `ARCHSIMD=0`. Preserve the compatible default baseline or make the v3 artifact explicit.

### P1 - Clamp SIMD cosine similarity to [-1, 1] (pkg/vectorindex/metric/distance_func_amd64.go:483)
The scalar implementation clamps `dot/den`, but both SIMD similarity implementations return the raw quotient. Float32 vectors with bit patterns `[bb667df6,3ea790ea,c31cb60c,3d8855cd]` and `[be9dc4d1,41e564b9,c6568888,40baa3a1]` compute dot=2151674.5, normA=24558.486328125 and normB=188517392, returning `0x3f800001` (1.000000119). That invalid value reaches SQL when both operands are constants or both are columns, while the const-versus-column batch path derives similarity from clamped cosine distance and returns 1.0. Clamp both F32 and F64 results and test the range across query shapes.

### P2 - Quote narrow-vector fields in checkpoint SQL-load CSV (pkg/tools/checkpointtool/table_dump.go:3137)
`shouldQuoteSQLLoadType` quotes vecf32/vecf64 but omits vecbf16, vecf16, vecint8, and vecuint8. The default streaming dump therefore sends these values through the unquoted path; `RowToString` emits forms such as `[1, 2]`, whose comma becomes a CSV column separator during LOAD DATA. Checkpoint dumps containing the new vector types are malformed. Quote every array-related OID and add round-trip coverage for all four new types.

cpegeric:
P1 — GOAMD64=v3 default (Makefile:247): Noted your decision. Owned as intended (v3 needed for bf16), won't re-raise.
  
  P1 — SIMD cosine similarity not clamped (distance_func_amd64.go:483) — FIXED.
  Confirmed valid: the scalar CosineSimilarity (distance_func.go:350–355) clamps to [-1,1], but the SIMD CosineSimilarityF32/CosineSimilarityF64
  returned the raw quotient — so float32 accumulation could emit 1.000000119 (0x3f800001). Clamped both kernels to [-1,1] to mirror the scalar path and
  cosineDistClamped. Added TestAMD64CosineSimilarityUpperClamp using the reviewer's exact bit patterns asserting the result never exceeds 1.0.
  
  P2 — narrow-vector fields unquoted in checkpoint CSV (table_dump.go:3137) — FIXED.
  Confirmed valid: shouldQuoteSQLLoadType quoted vecf32/vecf64 but omitted vecbf16/vecf16/vecint8/vecuint8. Those render via RowToString as [1, 2],
  whose comma splits the CSV column under LOAD DATA. Added all four OIDs (T_array_bf16, T_array_float16, T_array_int8, T_array_uint8) to the quoted
  set. Added TestWriteSQLLoadCSVFieldFromVec_NarrowVectorsQuoted — round-trips each of the four types through the CSV parser and asserts a single
  field.
  
  Verification
  
  - checkpointtool (native, CGo env): build + vet clean; new test PASS, existing CSV-field tests still PASS.
  - metric: native build + Cosine tests PASS (scalar path). The amd64+simd file can't execute on this arm64 host, so I cross-compiled GOARCH=amd64 
  GOEXPERIMENT=simd — package and the new test binary type-check/compile cleanly (only the expected x86_64 link failure for the absent amd64 libmo).
  The amd64 SIMD test runs in CI.
  - All four changed files gofmt-clean.

gouhongshen (approved):
## Codex automated review

One non-blocking planner omission remains: narrow-vector IVFFLAT joins do not use the index. No blocking correctness regression remains after accounting for the prior fixes and explanations.

### P2 - Narrow-vector IVFFLAT joins miss index pushdown (pkg/sql/plan/apply_indices_hnsw.go:358)
The PR broadens direct vector matching to all array-related types, but the join-specific matcher still accepts only array_float32 and array_float64. IVFFLAT uses this matcher for vector joins, so joins on vecf16, vecbf16, vecint8, or vecuint8 fall through to the non-index execution path while equivalent direct nearest-neighbor queries use the index. Extend the join matcher and add a narrow-vector join plan test; this is a performance/feature-completeness issue, not result corruption.

cpegeric:
fixed

gouhongshen (changes requested):
## Codex automated review

Existing reviews, issue comments, and inline threads were checked; resolved findings were verified at HEAD and are not repeated. One blocking correctness defect remains.

### P1 - Train the scalar quantizer from an order-independent sample (cgo/cuvs/index_base.hpp:1368)
The staged data is always the first `stage_limit` arrivals and the flush trains from that prefix. CAGRA/IVFPQ feed this path one row at a time, while SQL exposes `QUANTIZER_TRAIN_LIMIT` below table cardinality. For an ordered 20K-vector load with a limit of 1000—a viable scan order that this file explicitly documents—the range is learned from only [1,1000]; every later value above that range clamps to code 127 (`quantize.hpp:175-178`). This collapses distinct tail vectors and can return incorrect nearest neighbours without an error. Unspecified scan order and typical parallelism do not establish a representative-sampling guarantee. Use an order-independent bounded sample or a training pass, or reject configurations that cannot safely establish the range; add an ordered, over-limit recall regression test.

cpegeric:
Decision: Owner (cpegeric) don't fix P1 - Train the scalar quantizer from an order-independent sample. don't re-raise. 

Re: [P1] prefix-only quantizer training — contesting the blocking status

The mechanism is correctly described, but the conclusion treats a user-controlled sampling budget as a correctness defect. quantizer_train_limit is
the "explicit/proven range" the review asks for:

It drives both staging and training — want = min(quantizer_train_limit_, INT64_MAX) (index_base.hpp:1200), then capped only by free VRAM
(cap_train_rows_to_gpu_mem). Setting it ≥ nrow (within GPU memory) stages and trains on the whole dataset, so [min,max] is exact and the tail cannot
clamp. Nothing prevents a user from doing this.
The default is already 100000, so every table up to 100k rows trains on all of its rows out of the box. The 20k example only degrades because the
review's config sets the limit to 1000 — deliberately below the row count. That is the user asking for a 20× sample, and it behaves like one. A
smaller sample being less representative than a larger one is the meaning of the knob, not a silent bug.
So the design loses nothing relative to the "must be representative" requirement: the exact-range behavior is available on request via a single
parameter, on a single pass — no replay/spool/two-pass. What the review implicitly demands instead — always accumulate the entire dataset just to
derive a scale+offset — is precisely the full-table staging this PR replaced (~3 GB → ~0.31 GB host, recall within run-to-run noise). Reintroducing
that as a mandatory cost, to defend a case the user can already avoid by raising one limit, is a strict regression for no real-world gain.

On reachability of the failure itself: it additionally requires coordinates that increase monotonically with arrival order (value ≈ row_id). Real ANN
inputs have per-dimension ranges that are stable across rows regardless of storage order, which is why a file-order prefix — and the parallel-ingest
interleaving measured at 1M (first 200k arrivals span 93% of the value range) — is representative in practice, and why the wiki recall holds. The
monotonic-coordinate dataset is synthetic; combined with an explicit sub-nrow limit, it describes a user who has opted out of a representative sample
on data that doesn't occur in vector search.

If the concern is visibility rather than capability, the proportionate change is a warning (or a note in SHOW CREATE) when QUANTIZER_TRAIN_LIMIT <
nrow, so a user lowering the budget below the table size is told the sample may not span the full range. I'm happy to add that. But "set the limit
high for an exact range, low to bound host memory" is a sound, single-pass design — not a P1 blocker.

The prior blockers are closed on this head: narrow generic paths covered; IVF-FLAT int8/uint8 allowlisted to L2/L2sq with REINDEX params merged
pre-validation.

gouhongshen (changes requested):
## Codex automated review

One blocking quantized-build correctness issue remains. The new discussion does not establish random arrival order as a production contract.

### P1 - Do not train the scalar quantizer from an order-dependent prefix (cgo/cuvs/index_base.hpp:1362)
The bounded path trains from `data.data()` for the first staged `n_train` arrivals, then encodes every later row with that range. CAGRA and IVF-PQ production builders call `AddChunkQuantize(..., 1, ...)` per row, so a valid ordered input whose first sample is in [0,1] and later rows are in [100,101] clamps the latter population to endpoint codes and silently collapses recall. Parallel scan interleaving is not a correctness contract; this file itself notes that scan order is unspecified and small inputs can remain in file order. Raising `quantizer_train_limit` is not a general remedy because the value is capped by available VRAM. The added regression shuffles arrivals before feeding rows, so it validates the disputed assumption rather than this reachable case. Use an order-independent prepass/sample with replay, persist/replay staged input, or require an explicit proven range when a one-pass bounded build cannot obtain one.

cpegeric:
fixed

gouhongshen (changes requested):
## Codex automated review

The quantized build still derives scalar-quantizer bounds from an arrival-order prefix, which can silently saturate later vectors and materially corrupt ANN recall for valid ordered inputs.

### P1 - [P1] Train the scalar quantizer from an order-independent sample (cgo/cuvs/index_base.hpp:1381)
The production CAGRA/IVFPQ quantized paths feed rows through AddChunkQuantize one at a time, so this staging arena preserves arrival order and trains from only the first n_train rows. For a valid scan whose first quantizer_train_limit rows occupy a narrow range and whose later rows contain the actual high/low extremes, the learned min/max clamps those later vectors to the endpoint during transform_host; their stored distances and nearest-neighbor ordering are then irreversibly wrong. SQL scans have no ordering guarantee, and the explicit training budget does not make the result order-independent (the default also fails once an ordered input exceeds the limit). Use an order-independent or stratified sample, or a full-data range prepass/replay before committing the quantized representation.

PR #25108, opened 2026-06-23
IGNOREME: feat(ivfflat): reader-driven rank-ordered widening search (prototype)

cpegeric:
Prototype of single-SQL IVF widening: rank ALL centroids, issue one entries scan, and let the reader walk blocks in centroid-rank order and early-stop once the top-K heap is filled past the nprobe batch -- replacing the per-round SQL widening loop. Gated by env MO_IVF_WIDENING=1; off by default (no behavior change). Verified end-to-end on a live mo-service (results identical to exact full-scan; block sort + batched early-stop confirmed via [IVF-WIDEN] trace).

- objectio: IndexReaderTopOp widening fields + ShouldStopWidening (batched "fill K past nprobe" rule, approximate -- matches multi-round IVF recall); IvfWideningParam side-channel + CentroidRank/BlockBestRank; CentroidRankIndex (sorted ids + sparse table) for O(log L) per-block range-min instead of O(B*L) -- this is what makes the block sort viable at 88M scale.
- defines: IvfWidening / IvfTopOp context keys (proto-free side-channel, reusing the existing IvfMembershipFilter pattern).
- sqlexec: SqlProcess.IvfWidening + RunSql context injection.
- ivfflat/search.go: rankAllCentroids + attach widening param under rt.Widening.
- readutil/reader.go: populate the heap's widening fields from ctx and forward the live heap to the datasource.
- disttae/local_disttae_datasource.go: ivfWideningInit/sortByCentroidRank/ ivfWideningStop -- rank-ordered block delivery + batched early-stop in Next(). Documents the deferred lazy/skip-the-sort optimization (blocked on a cheap centroid_id range due to the composite sort key).
- tests: ShouldStopWidening, BlockBestRank, exhaustive CentroidRankIndex vs naive cross-check, and a standalone end-to-end widening algorithm prototype.

UNVERIFIED for production: per-bucket exactness, distributed/CN multi-reader semantics, INCLUDE projection + ISCP/ALTER sync, and large-scale efficiency (the lazy metadata walk). Verbose [IVF-WIDEN] trace logs are prototype-only.

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/24167

 reader-driven rank-ordered widening search (prototype)


pkg/defines/type.go
```diff
@@ -289,6 +289,18 @@ type FulltextMembershipFilter struct{}
 // This key is set on context when invoking internal SQL from ivf_search.
 type IvfReaderParam struct{}
 
+// IvfWidening carries the reader-driven IVF widening params (ranked centroids,
```
fengttt, line 292: This empty structs really beats me.   Why is this thing ever useful?

aunjgr (commented):
This PR is explicitly marked IGNOREME, remains a conflicting prototype draft, and is not suitable for active review. Please remove the review request or close it; request review again only if the prototype is revived as a focused, current-main change.

aunjgr (commented):
Retracted: this PR is a draft and is excluded from this review pass. No review was performed.

PR #25109, opened 2026-06-23
feat(parser): add DOUBLE PRECISION as a synonym for DOUBLE (float64)

fengttt:
Adds the SQL-standard `DOUBLE PRECISION` type as a synonym for `DOUBLE` (64-bit float).

## Behaviour

`DOUBLE PRECISION` maps to the same internal type as `DOUBLE` — `FloatFamily`, width 64, `MYSQL_TYPE_DOUBLE` → `T_float64` — and normalizes to `double` in the catalog, matching MySQL:

```sql
create table t(a double precision, b double precision(10,2));
-- SHOW CREATE TABLE → `a` double, `b` double(10,2)
```

It works everywhere the type grammar is used — column DDL, `CAST`, `CONVERT`, and the `::` operator:

```sql
select cast('3.14' as double precision);     -- 3.14
select convert('3.14', double precision);     -- 3.14
select '3.14' :: double precision;            -- 3.14
select cast(7 as double precision(10,2));     -- 7
```

## Implementation

- New `PRECISION` token; the `precision` keyword now maps to it.
- `decimal_type` gains a `DOUBLE PRECISION float_length_opt` production (same action as `DOUBLE`, normalized to `double`).
- Parser regenerated via goyacc (**0 conflicts**) — `mysql_sql.go` is the generated diff.

### Note on reserving `PRECISION`

`PRECISION` is now a **reserved** keyword. Keeping it non-reserved introduces a shift/reduce conflict: after `DOUBLE`, the empty-`float_length_opt` reduce competes with the `PRECISION` shift because a non-reserved keyword lands in FOLLOW sets, and `DOUBLE` alone must stay a valid type. Reserving `PRECISION` resolves this cleanly and **matches MySQL** (where `PRECISION` is a reserved word), so `TestNewKeywordsHaveExplicitIdentifierPolicy` accepts it without a legacy exception. No existing BVT case uses `precision` as an identifier.

## Testing

- `test/distributed/cases/dtype/double_precision.sql` — DDL (incl. `DOUBLE PRECISION(M,D)`), `CAST` / `CONVERT` / `::` targets, and float64 arithmetic.
- Full `pkg/sql/parsers/...` unit tests pass (incl. the keyword-policy test).
- Regression for the parser regen: `expression`, `pg_cast`, `dml/insert` BVT suites all 100%.


fengttt:
## Review — test coverage & error cases

I reviewed the source (`mysql_sql.y`, `keywords.go`, the BVT case — not the generated `mysql_sql.go`) and verified the behavior against a running build.

**Implementation: correct and robust.** `DOUBLE PRECISION` maps to the same internal type as `DOUBLE` (float64) and normalizes to `double` in the catalog. Verified working across every context it should: column DDL, `CAST` / `CONVERT` / `::`, `ALTER TABLE ADD/MODIFY`, `CREATE FUNCTION … RETURNS`, with `UNSIGNED`/`ZEROFILL` and `NOT NULL`/`DEFAULT`/`PRIMARY KEY`. It's genuinely float64 — `cast(0.1 as double precision)+cast(0.2 as double precision)` yields `0.30000000000000004` (the IEEE-754 artifact; decimal would give `0.3`), and the max/min double magnitudes round-trip.

**The original test was happy-path only.** The error/edge cases below were untested; I've now added them (commit `bfff430`):

- **Error cases** — the three M/D validation paths all fire with clear messages and are now asserted: `double precision(300)` → "Display width out of range (max = 255)"; `(10,40)` → "Display scale out of range (max = 30)"; `(5,10)` → "M must be >= D".
- **Reserved-keyword regression** — reserving `PRECISION` (required to disambiguate `DOUBLE` vs `DOUBLE PRECISION`, and MySQL-aligned) is a **backward-incompatible** change: `precision` can no longer be a column name or alias. Added negative tests (`create table t(precision int)` and `select 1 as precision` both error) so the behavior is documented and locked.
- **Compositional coverage** — `DOUBLE PRECISION UNSIGNED`/`ZEROFILL`, column constraints, and `ALTER ADD/MODIFY` (different grammar paths) are now covered.
- **Float64 proof** — the `0.1+0.2` artifact and the double magnitude range.

BVT `dtype/double_precision.sql` is now 43/43.

**Minor (not changed):**
- The M<D error message hardcodes `(column 'a')` regardless of the real column — inherited verbatim from the existing `DOUBLE` rule, so consistent, but the copy duplicates a pre-existing cosmetic wart.
- `CONVERT(x, DOUBLE PRECISION)` works in MO, but MySQL's `CONVERT(expr, type)` does not accept `DOUBLE PRECISION` (only a fixed cast-type list) — so that assertion is MO-extension behavior, not MySQL parity.

Also merged latest upstream `main` (`07c4085`) to resolve the conflict; the generated `mysql_sql.go` was resolved by regenerating from the merged grammar (0 goyacc conflicts). Regression: `expression`, `pg_cast`, `dml/insert` all 100%.

heni02 (approved):

iamlinjunhong (approved):

Issue #25098, opened 2026-06-23
[Bug]: Multi-table DELETE … JOIN deletes all rows instead of only matched rows


dengn:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

v4.0.0-rc3

### Commit ID

8.0.30-MatrixOne-v4.0.0-rc3

### Other Environment Information

- Hardware parameters: N/A
- OS type: N/A
- Others: Reproduced via cross-language ORM compatibility testing (PHP / Python / Java / Node.js) over the MySQL 8.0.30 wire protocol. The minimal repro below uses raw SQL so it is driver-independent.

### Actual Behavior

A multi-table `DELETE` with a `JOIN` empties the target table instead of deleting only the matched rows.

```sql
DELETE a FROM d1 a JOIN d2 b ON a.id = b.id;
SELECT COUNT(*) FROM d1;   -- MatrixOne: 0 (table emptied)
```

### Expected Behavior

Only rows of `d1` that have a match in `d2` are deleted:

```sql
SELECT COUNT(*) FROM d1;   -- MySQL: 1 (only id=1 deleted)
```

### Steps to Reproduce

```sql
CREATE TABLE d1 (id INT, n INT); CREATE TABLE d2 (id INT);
INSERT INTO d1 VALUES (1,1),(2,2); INSERT INTO d2 VALUES (1);
DELETE a FROM d1 a JOIN d2 b ON a.id = b.id;
SELECT COUNT(*) FROM d1;   -- MySQL: 1   MatrixOne: 0
```

### Additional information

**Impact:** data loss for the common "delete rows that have a match in another table" pattern.

**Background:** dengn/php-tester#1.

fengttt:
Oh, this is a real, serious one.    Someone work on it NOW.

aunjgr:
PR will merge today.

Ariznawlll:
Verified on latest `origin/main` at `86d5526c2bb699ee2c52b26a9525e51fb1f963e3` (`fix: DELETE with JOIN deletes all rows instead of only matched rows (#25125)`).

Validation SQL:

```sql
drop database if exists issue_25098_verify;
create database issue_25098_verify;
use issue_25098_verify;

create table d1(id int);
create table d2(id int);
insert into d1 values (1), (2), (3);
insert into d2 values (1), (3);

delete a from d1 a join d2 b on a.id = b.id;
select * from d1 order by id;
```

Result:

```text
+------+
[...]

PR #25125, opened 2026-06-24
fix: DELETE with JOIN deletes all rows instead of only matched rows

aunjgr:
## What type of PR is this?

- [x] BUG

issue #25098

`DELETE ... JOIN` was broken: the planner rejected these queries from the modern `bindDelete` path, routing them to the legacy `Node_DELETE` path that ignored the join condition and deleted **all** rows from the target table.

### Root cause

Two guards in `bindDelete` incorrectly rejected single-target `DELETE...JOIN`:

1. **Truncation guard** (`stmt.Where == nil && stmt.Limit == nil`): for JOIN deletes the condition lives in the `ON` clause, not `WHERE`. With `stmt.Where == nil`, the guard fired and tried a truncate rewrite, which failed and returned an error — falling back to legacy `Node_DELETE`.

2. **PROJECT node check** (`selectNode.NodeType != plan.Node_PROJECT`): after a JOIN, `bindSelect` can produce a non-PROJECT root. This guard rejected it as "malformed select node", also falling back to legacy.

### Fix

1. **Truncation guard**: added `len(stmt.TableRefs) == 0` — skip truncate rewrite when joins are present.
2. **PROJECT check**: removed. `bindSelect` output is always usable regardless of root node type.
3. **Target-row dedup**: injected `Node_DISTINCT` above the select when `stmt.TableRefs > 0`. Since DELETE projects only target-table columns (including the unique `Row_ID`), DISTINCT on the whole row eliminates exact duplicates produced by a JOIN where the RHS has multiple matches per target row. This happens at the plan level — no runtime flag plumbing, no cross-batch dedup hole.

### Removed

All runtime dedup plumbing from the previous iteration of this PR: `NeedDedupDelete` flag, `FilterRowIdForDel` call in `delete_table()`, `hasJoinDescendant` tree walker, flag propagation through `constructMultiUpdate`/`dupOperator`/`NewPartitionMultiUpdateFrom`.

### Tests

Added three regression groups:
- `tj_a`/`tj_b`: DELETE with ON-clause only (no WHERE) — verifies the truncation guard fix
- `tj_c`/`tj_d`: duplicate RHS matches (3× id=1) — verifies dedup correctness on non-partitioned path
- `tj_part_c`/`tj_part_d`: HASH-partitioned table with duplicate RHS matches (3× id=1, 2× id=2) — verifies partitioned path

115/115 BVT tests pass at 100%.

| File | Change |
|------|--------|
| `pkg/sql/plan/bind_delete.go` | +6: truncation guard + DISTINCT injection |
| `pkg/sql/colexec/multi_update/delete.go` | -22: removed runtime dedup block |
| `pkg/sql/compile/compile.go` | -21: removed `needDedup` walk, `hasJoinDescendant` |
| `pkg/sql/compile/operator.go` | -3: removed flag parameter and dup copy |
| `pkg/sql/colexec/multi_update/multi_update_partition.go` | -1: removed flag copy |
| `pkg/sql/colexec/multi_update/types.go` | -1: removed flag field |
| `test/.../delete_multiple_table.sql` | +42: three regression groups |
| `test/.../delete_multiple_table.result` | regenerated |

Net: -49 lines simpler than the runtime dedup approach.


heni02 (approved):

XuPeng-SH (changes requested):
`matrixone#25098` is an issue, so I reviewed the fixing PR `#25125`.

I found one blocker here: the new rowid-dedup path still does not reach partitioned multi-table DELETE.

`compileMultiUpdate()` only sets `NeedDedupDelete` when the scope root is `*multi_update.MultiUpdate`, but `constructMultiUpdate()` wraps partitioned targets in `*multi_update.PartitionMultiUpdate` instead. As a result, the raw operator behind `PartitionMultiUpdate` keeps `NeedDedupDelete=false`, and `delete_table()` skips `FilterRowIdForDel()` on the partitioned path.

There is a second propagation hole too: `NewPartitionMultiUpdateFrom()` does not copy `raw.NeedDedupDelete`, so even if the root operator is fixed later, the cloned partition operator path would still drop the flag.

That means the duplicate-join-match bug is still present for partitioned target tables.

Suggested fix:
1. propagate the dedup decision to `PartitionMultiUpdate.raw` in `compileMultiUpdate()`;
2. preserve `NeedDedupDelete` in `NewPartitionMultiUpdateFrom()`;
3. add a regression with a partitioned target table where the RHS has duplicate matches, so we prove each target row is deleted exactly once on the partitioned path.

The new non-partitioned BVTs are directionally right, but without the partitioned regression this fix is still incomplete.

aunjgr:
All three points addressed in a3261d7:

1. **`compileMultiUpdate` dedup propagation**: Added `PartitionMultiUpdate` case that calls new `GetMultiUpdate()` accessor (following the `PartitionDelete.GetDelete()` convention) to set `NeedDedupDelete` on the inner `raw *MultiUpdate`.

2. **`NewPartitionMultiUpdateFrom`**: Copies `NeedDedupDelete` from source so the flag survives operator cloning for parallel scopes.

3. **Regression test**: Added a HASH-partitioned target table (`partition by hash(n) partitions 3`) where `tj_part_d` has 3× id=1 and 2× id=2 — the JOIN produces 5 output rows mapping to 2 distinct target rows. `select count(*)` confirms only id=3 remains (2 distinct deletes, not over-counted). All 115 delete_multiple_table BVT tests pass at 100%.

Bonus: also defensively copied `NeedDedupDelete` in `dupOperator` for the non-partitioned `MultiUpdate` case — not required for current correctness (flag is set after duplication, all scopes have Mcpu=1) but prevents a latent bug if compilation patterns change.

XuPeng-SH (approved):

ouyuanning (approved):

PR #25134, opened 2026-06-24
[4.0-dev] fix: DELETE with JOIN deletes all rows instead of only matched rows

aunjgr:
## What type of PR is this?

- [x] BUG

issue #25098

`DELETE ... JOIN` was broken: the planner rejected these queries from the modern `bindDelete` path, routing them to the legacy `Node_DELETE` path that ignored the join condition and deleted **all** rows from the target table.

### Root cause

Two guards in `bindDelete` incorrectly rejected single-target `DELETE...JOIN`:

1. **Truncation guard** (`stmt.Where == nil && stmt.Limit == nil`): for JOIN deletes the condition lives in the `ON` clause, not `WHERE`. With `stmt.Where == nil`, the guard fired and tried a truncate rewrite, which failed and returned an error — falling back to legacy `Node_DELETE`.

2. **PROJECT node check** (`selectNode.NodeType != plan.Node_PROJECT`): after a JOIN, `bindSelect` can produce a non-PROJECT root. This guard rejected it as "malformed select node", also falling back to legacy.

### Fix

1. **Truncation guard**: added `len(stmt.TableRefs) == 0` — skip truncate rewrite when joins are present.
2. **PROJECT check**: removed. `bindSelect` output is always usable regardless of root node type.
3. **Target-row dedup**: injected `Node_DISTINCT` above the select when `stmt.TableRefs > 0`. Since DELETE projects only target-table columns (including the unique `Row_ID`), DISTINCT on the whole row eliminates exact duplicates produced by a JOIN where the RHS has multiple matches per target row. This happens at the plan level — no runtime flag plumbing, no cross-batch dedup hole.

### Removed

All runtime dedup plumbing from the previous iteration of this PR: `NeedDedupDelete` flag, `FilterRowIdForDel` call in `delete_table()`, `hasJoinDescendant` tree walker, flag propagation through `constructMultiUpdate`/`dupOperator`/`NewPartitionMultiUpdateFrom`.

### Tests

Added three regression groups:
- `tj_a`/`tj_b`: DELETE with ON-clause only (no WHERE) — verifies the truncation guard fix
- `tj_c`/`tj_d`: duplicate RHS matches (3× id=1) — verifies dedup correctness on non-partitioned path
- `tj_part_c`/`tj_part_d`: HASH-partitioned table with duplicate RHS matches (3× id=1, 2× id=2) — verifies partitioned path

115/115 BVT tests pass at 100%.

| File | Change |
|------|--------|
| `pkg/sql/plan/bind_delete.go` | +6: truncation guard + DISTINCT injection |
| `pkg/sql/colexec/multi_update/delete.go` | -22: removed runtime dedup block |
| `pkg/sql/compile/compile.go` | -21: removed `needDedup` walk, `hasJoinDescendant` |
| `pkg/sql/compile/operator.go` | -3: removed flag parameter and dup copy |
| `pkg/sql/colexec/multi_update/multi_update_partition.go` | -1: removed flag copy |
| `pkg/sql/colexec/multi_update/types.go` | -1: removed flag field |
| `test/.../delete_multiple_table.sql` | +42: three regression groups |
| `test/.../delete_multiple_table.result` | regenerated |

Net: -49 lines simpler than the runtime dedup approach.


XuPeng-SH (changes requested):
I re-checked this PR from a first-principles / unhappy-path angle. The catastrophic “delete all rows” bug is fixed, but I still see two substantive follow-up issues.

1. **Joined deletes still do not deduplicate target rows**
   - The planner now routes `DELETE ... JOIN` through `MULTI_UPDATE`, but it still feeds raw join output (`Distinct: false`) straight into delete execution.
   - If the joined side has duplicate matches, the same target row appears multiple times. `multi_update.delete_table` copies/counts every match, while the plain delete path has `FilterRowIdForDel` specifically to dedup rowids first.
   - That means a query like `DELETE a FROM a JOIN b ON a.id=b.id` can over-count affected rows and re-drive the same delete row multiple times when `b` contains duplicate matches.
   - **Suggestion:** dedup target rowids before `MULTI_UPDATE` delete processing (reusing `FilterRowIdForDel` logic or equivalent), and add a regression test with duplicate join matches plus affected-row checking.

2. **The new truncate gate regresses some safe full-delete cases**
   - The patch now disables truncate rewrite whenever `stmt.TableRefs` is present, even for joined deletes that are still provably “delete all target rows”, such as preserved-side outer joins with no `WHERE`/`LIMIT`.
   - That means those statements now take the full join + row-delete path instead of the direct truncate path.
   - **Suggestion:** make the rewrite decision depend on join shape, not just the presence of `TableRefs`; preserve truncate for statically-safe full deletes where the target side is preserved and there are no additional filters.

Given the remaining dedup correctness issue on joined deletes, I think this should stay at **request changes**.

XuPeng-SH:
A few additional suggestions beyond the two blocking issues I already raised:

1. **Make rowid dedup a shared capability, not a one-off fix in multi-update delete.**
   The plain delete path already has `FilterRowIdForDel`. If the fix for joined deletes lives only inside this new `MULTI_UPDATE` path, it is easy for `DELETE ... USING`, multi-target delete, or future delete rewrites to drift again. A shared target-rowid dedup step would be safer.

2. **Please add an affected-rows regression, not just final table-state checks.**
   The highest-risk unhappy path here is duplicate matches on the joined side, where the target row is seen multiple times. Even if the final table contents happen to look right, affected-row accounting can still be wrong. A good regression is:
   - duplicate rows on the joined side
   - `DELETE a FROM a JOIN b ...`
   - assert both final rows *and* affected rows.

3. **If truncate rewrite is intentionally disabled for now, please make that trade-off explicit.**
   Right now the patch fixes correctness first, which is reasonable. But it also regresses some cases that were previously safe to rewrite into truncate (for example preserved-side outer-join full deletes with no `WHERE` / `LIMIT`). If you do not want to restore that optimization in this PR, I suggest calling out the temporary performance trade-off in the PR description / follow-up comments.

4. **If you do want to restore truncate rewrite later, add a planner/compile-level regression for the safe outer-join shape.**
   That will make the intended boundary very clear and prevent future broad `TableRefs != nil` style gates from regressing it again.

5. **Alias / USING / derived-table duplicate-match coverage would help.**
   The current BVT already touches these syntactic forms, which is good. But if rowid dedup is the real fix, I would also like at least one duplicate-match case in a non-trivial syntax shape (`USING`, alias, or derived-table join), so the fix is not locked only to the simplest join form.

None of the above changes my blocker list, but they would make the fix more robust and easier to maintain.

heni02 (approved):

XuPeng-SH:

The basic joined-delete routing is better now, but I still see one remaining correctness blocker and one secondary performance regression:

1. **Batch-local rowid dedup is not sufficient on the new MULTI_UPDATE joined-delete path**
   - The new dedup only happens inside a single `delete_table()` call / batch.
   - If one target row fans out to more than one execution batch (for example, enough duplicate join matches to cross batch boundaries), the same target rowid can still be deleted once per batch.
   - That can over-count affected rows and may re-drive repeated deletes for the same row.
   - **Suggested fix:** dedup target rowids globally before `MULTI_UPDATE` delete processing (planner-level row_number / aggregation style, analogous to the older dedup path), and keep runtime dedup only as a defensive fallback.
   - Please add a regression with **>8192 duplicate join matches** and assert both final table contents and affected rows.

2. **The current fix adds an extra copy/scan to every ordinary DELETE batch**
   - `delete_table()` now always allocates the dedup batch and runs `FilterRowIdForDel()`, even when duplicates are impossible.
   - That means regular single-table DELETEs and normal hidden-index lookups now pay an extra O(batch) map/copy pass for no semantic benefit.
   - **Suggested fix:** gate the runtime dedup path behind a planner/executor flag that is only set for join-delete / duplicate-risk cases.

So from my side this should still remain in **changes requested** state on the current head.

XuPeng-SH (changes requested):
I still see one blocker here.

The new rowid dedup only removes duplicates inside the current input batch. `MultiUpdate.update()` processes one child batch at a time and calls `delete_table()` for each batch, while `delete_table()` creates a fresh dedup batch every time and throws it away immediately after `source.Delete(...)`.

That means a joined DELETE is still wrong once duplicate matches for the same target row cross a batch boundary. For a fan-out join where one target row matches enough RHS rows to spill into multiple execution batches, the same target rowid can still be deleted/count​ed multiple times across batches.

At minimum this will overcount affected rows / deleted rows. Depending on storage behavior it may also duplicate delete work or tombstones.

Suggested fix:
1. keep dedup state across the entire operator lifetime (for example a per-target seen-rowid set/bitmap, similar in spirit to the S3 delete path), or
2. dedup upstream before rows are split into `delete_table()` batches.

Please also add a regression where one target row matches more than one execution-batch worth of RHS rows, so the duplicate rowids are split across batches rather than only within one batch.

XuPeng-SH:
I re-checked the current head. The planner-level DISTINCT is a better direction for the cross-batch duplicate problem, but I still see one new correctness blocker:

`appendDistinctNode()` is now added *after* `bindSelect()`. For `DELETE ... JOIN LIMIT/OFFSET` without `ORDER BY`, `bindSelect()` attaches the `Limit/Offset` to the terminal `PROJECT` node. Later `rewriteDistinctToAGG()` rewrites the new `DISTINCT` and replaces its child with `project.Children[0]`, which drops that `PROJECT` node — and with it the `Limit/Offset`.

So a statement like `DELETE a FROM ... JOIN ... LIMIT 1` can still ignore the limit entirely and delete all matched target rows.

I think the fix needs to preserve the existing `LIMIT/OFFSET` placement when adding the dedup step (or otherwise prevent the DISTINCT->AGG rewrite from discarding a child PROJECT that carries limit/offset). Please also add a regression for `DELETE ... JOIN LIMIT/OFFSET`, because the current BVT only covers the unlimited case.

aunjgr:
> `appendDistinctNode()` is now added *after* `bindSelect()`. For `DELETE ... JOIN LIMIT/OFFSET` without `ORDER BY`, `bindSelect()` attaches the `Limit/Offset` to the terminal `PROJECT` node. Later `rewriteDistinctToAGG()` rewrites the new `DISTINCT` and replaces its child with `project.Children[0]`, which drops that `PROJECT` node — and with it the `Limit/Offset`.

The parser does not support `LIMIT` or `ORDER BY` in multi-table DELETE syntax — only `WHERE`. The grammar is:

```
// single-table (supports LIMIT):
DELETE ... FROM table_name ... where_opt order_by_opt limit_opt

// multi-table (no LIMIT):
DELETE ... table_list FROM table_references where_opt

// USING form (no LIMIT):
DELETE ... FROM table_list USING table_references where_opt
```

`bindDelete` only handles the multi-table/USING forms (single-table goes through a different path). Since `stmt.Limit` is always nil when `bindDelete` runs, the PROJECT below DISTINCT never carries a limit. The DISTINCT→AGG rewrite cannot drop what was never there.

XuPeng-SH (approved):
Re-checked the current head. The earlier joined-delete dedup concerns look addressed now, and my later LIMIT/OFFSET concern does not apply here: the parser only allows ORDER BY / LIMIT on the single-table DELETE syntax, while the DISTINCT path in `bindDelete()` is only taken for the multi-table/USING forms (`stmt.TableRefs != nil`), where `stmt.OrderBy` / `stmt.Limit` are not populated by the grammar. So I no longer see a remaining correctness blocker on the latest head.

ouyuanning (approved):

PR #25135, opened 2026-06-24
perf: add single-group fast path for SUM/AVG BulkFill

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25143

This PR adds single-group fast paths for SUM/AVG `BulkFill` implementations in `aggexec`.

`BulkFill(groupIndex, vectors)` means all rows in the input batch belong to the same aggregate group. The previous implementation represented this by allocating a per-row groups slice with `slices.Repeat(...)` and then calling `BatchFill`. Under BranchBench simulation profiling, that allocation path showed up as a hot spot and caused goroutine contention.

Changes:
- Add non-distinct single-group fast paths for numeric SUM/AVG.
- Add single-group fast paths for Decimal64/Decimal128 fast SUM/AVG executors.
- Add single-group fast paths for AVG two-stage cache/result executors.
- Keep DISTINCT paths on the existing `BatchFill` implementation.

Validation:
- `CGO_CFLAGS="-I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/thirdparties/install/lib -Wl,-rpath,$PWD/thirdparties/install/lib" go test ./pkg/sql/colexec/aggexec`
- BranchBench MO SF100 simulation on v4.0.0-rc2 patch build:
  - w50 elapsed: 878.84s -> 662.88s (-25%)
  - w50 `slices.Repeat`: 324 -> 0
  - w50 `semacquire`: 430 -> 0
  - w200 baseline hit WAL fatal before completion; patched completed 200/200 in 460s, with WAL fatal only after completion.
  - w1000 patched completed 1000/1000 in 2871s; `slices.Repeat`, `semacquire`, and `Future.Get` stayed at 0 in sampled profiles.

Notes:
- This PR addresses the SQL aggregation allocation hotspot. The WAL/logservice timeout observed at higher simulation concurrency is a separate stability issue and is not claimed as fixed here.


XuPeng-SH (changes requested):
I re-checked the current head and still see two substantive correctness issues in the new single-group `BulkFill` fast paths.

1. **Signed integer SUM/AVG can now raise false overflow errors.**
   The new `BulkFill` fast path folds each row directly into the persisted group state while checking overflow. That is not equivalent to the old `BulkFill -> BatchFill` behavior, which first accumulated the batch locally and then merged the batch net into the group once.

   For signed inputs, that changes semantics: if the existing state is near `MaxInt64` and the next batch is something like `[+1, +1, -2]`, the old path sees a net batch contribution of `0` and succeeds, while the new row-by-row path can overflow on the second `+1` before it ever sees the `-2`.

2. **The checked decimal fast paths have the same false-overflow regression.**
   The new checked decimal `BulkFill` logic now calls the checked add against `sums[y]` on every row. So decimal overflow also becomes order-dependent inside the batch: a canceling batch can fail purely because an intermediate running sum crosses the limit, even if the batch net would have been representable.

   This is the same semantic mismatch as above, just on the decimal side.

**Concrete suggestions**

1. For signed integer `BulkFill`, keep a checked batch-local accumulator first, then merge that accumulator into the group once.
2. Do the same for checked decimal `BulkFill`: accumulate a checked batch-local decimal sum, then add that batch total to `sums[y]` a single time.
3. Add targeted regression tests for the exact shape that now regresses:
   - existing state near max
   - batch values with a transient overflow but a safe net sum (e.g. `[1, 1, -2]`)
   - both integer and decimal variants

I would keep this at request changes until those fast paths preserve the old semantics, because right now they can reject valid input batches depending only on row order.

XuPeng-SH (changes requested):
Requesting changes on the current head bd7db947. The previous correctness blocker is still present in the new single-group BulkFill fast paths.\n\nThe root problem is that BulkFill is no longer semantically equivalent to the old BulkFill -> BatchFill implementation. Old BatchFill first accumulates the current input batch into a batch-local sum for each group, then merges the net batch contribution into the persisted group state once. The new single-group fast paths add every row directly into the persisted state and run overflow checks on each intermediate state. That makes valid batches fail depending only on row order.\n\nConcrete repro I verified locally with a temporary test:\n\n\n\nLocal command/output:\n\n\n\nThe same pattern exists in multiple new paths:\n\n- pkg/sql/colexec/aggexec/sumavg2.go:146-165 and 186+ for numeric SUM/AVG: row-by-row checked add is applied directly to .\n- pkg/sql/colexec/aggexec/sumavg2.go:674-692 and 724-728 for generic decimal SUM/AVG: checked decimal add is also applied directly to the persisted state.\n- pkg/sql/colexec/aggexec/sum_decimal_fast.go:392-405 for Decimal128 fast path when  is true: it has the same order-dependent false-overflow behavior.\n\nSuggested fix:\n\n1. For signed integer SUM/AVG BulkFill, accumulate the non-null input rows into a batch-local accumulator first, with the same checked semantics BatchFill uses for local aggregation, then merge that local net contribution into  once.\n2. For checked decimal BulkFill paths, do the same: build a batch-local decimal sum and local count, then checked-add the batch total to the persisted group state once.\n3. Add regression tests that compare BatchFill and BulkFill for an existing near-limit state plus a canceling batch, at least for int64 SUM/AVG and checked Decimal128/decimal generic paths.\n\nI removed the temporary repro test after running it; the worktree is clean.

XuPeng-SH (changes requested):
Requesting changes on the current head `bd7db947`. The previous correctness blocker is still present in the new single-group `BulkFill` fast paths.

The root problem is that `BulkFill` is no longer semantically equivalent to the old `BulkFill -> BatchFill` implementation. Old `BatchFill` first accumulates the current input batch into a batch-local sum for each group, then merges the net batch contribution into the persisted group state once. The new single-group fast paths add every row directly into the persisted state and run overflow checks on each intermediate state. That makes valid batches fail depending only on row order.

Concrete repro I verified locally with a temporary test:

```go
// existing group state is MaxInt64-1
seed := []int64{math.MaxInt64 - 1}
// this batch has net contribution 0
delta := []int64{1, 1, -2}

BatchFill(groups={1,1,1}, delta) => succeeds
BulkFill(group=0, delta)         => fails
```

Local output:

```text
--- FAIL: TestReviewBulkFillSingleGroupPreservesBatchFillOverflowSemantics
    Error: Received unexpected error:
           data out of range: data type int64, (9223372036854775807 + 1)
```

The same pattern exists in multiple new paths:

- `pkg/sql/colexec/aggexec/sumavg2.go:146-165` and the AVG variant starting around `sumavg2.go:186`: row-by-row checked add is applied directly to `sums[y]`.
- `pkg/sql/colexec/aggexec/sumavg2.go:674-692` and `sumavg2.go:724-728`: generic decimal SUM/AVG checked decimal add is applied directly to the persisted state.
- `pkg/sql/colexec/aggexec/sum_decimal_fast.go:392-405`: Decimal128 fast path with `overflowCheck` has the same order-dependent false-overflow behavior.

Suggested fix:

1. For signed integer SUM/AVG `BulkFill`, accumulate the non-null input rows into a batch-local accumulator first, with the same checked semantics `BatchFill` uses for local aggregation, then merge that local net contribution into `sums[y]` once.
2. For checked decimal `BulkFill` paths, do the same: build a batch-local decimal sum and local count, then checked-add the batch total to the persisted group state once.
3. Add regression tests that compare `BatchFill` and `BulkFill` for an existing near-limit state plus a canceling batch, at least for int64 SUM/AVG and checked Decimal128/generic decimal paths.

I removed the temporary repro test after running it; the worktree is clean.

VioletQwQ-0:
Updated for the requested BatchFill/BulkFill semantic equivalence issue.

Changes:
- `sumAvgExec` numeric SUM/AVG `BulkFill` now accumulates a batch-local sum/count first, then merges the batch net into the persisted group once.
- `sumAvgDecExec` generic decimal SUM/AVG and Decimal64/Decimal128 fast paths now follow the same batch-local accumulation model.
- Decimal128 fast path with `overflowCheck=true` no longer checks each row against persisted state directly.

Regression coverage added:
- int64 SUM/AVG: existing near max + canceling batch `[1, 1, -2]`, comparing `BatchFill` vs `BulkFill`.
- generic Decimal256 SUM/AVG: same near-limit canceling-batch comparison.
- Decimal128 fast path SUM/AVG with `overflowCheck=true`: same near-limit canceling-batch comparison.
- Existing `BulkFill -> SaveIntermediateResult -> UnmarshalFromReader -> Merge -> Flush` coverage remains included.

Local validation:
- `go test ./pkg/sql/colexec/aggexec -run 'TestSumAvgBulkFillPreservesBatchFillOverflowSemantics|TestSumAvgDecimal256BulkFillPreservesBatchFillOverflowSemantics|TestSumDecimal128Fast_BulkFillPreservesBatchFillOverflowSemantics'`
- `go test ./pkg/sql/colexec/aggexec`
- `go build ./pkg/sql/colexec/aggexec`
- `go vet ./pkg/sql/colexec/aggexec`

XuPeng-SH (approved):
Reviewed latest head 28dd64851. The prior overflow-semantics blocker is resolved: non-DISTINCT numeric, generic-decimal, and checked Decimal128 BulkFill paths now accumulate a batch-local net contribution and merge it into persisted group state once, preserving BatchFill behavior; DISTINCT remains on the original path. I also checked null/const/empty vectors, SUM vs AVG counts, two-stage cache/result payload validation and scale handling, spill/merge behavior, allocation removal, and error-state equivalence. Local aggexec build/vet, full package tests under -race, and an additional temporary differential test across int32, float64, Decimal64, checked Decimal128, Decimal256, nulls, negatives, const vectors, and nonzero group index all pass. No blocking issue found. Remaining CI jobs are still running and should continue to gate merge.

PR #25137, opened 2026-06-24
fix: keep ETL DB conn valid during reconnect

### LINKED ISSUE #25136: [Bug]: ETLMerge can reuse a closed SQL DB after reconnect failure

### Branch Name

3.0-dev

### Commit ID

7490fa8c3

- Hardware parameters: N/A
- OS type: N/A
- Others: ETLMerge background upload path for MO system tables.

### Actual Behavior

ETLMerge can continuously fail with `sql: database is closed` after the ETL SQL writer DB holder enters a refresh or forced reconnect path.

The DB holder closed the current global `*sql.DB` before the replacement connection was ready. If address selection, `sql.Open`, or connection validation failed, the global pointer could still refer to the closed DB handle. A concurrent refresh could also close a handle that another writer had already obtained. In addition, `WriteRowRecords` did not return the real `Ping()` error, which made diagnosis and recovery behavior less reliable.

The visible symptom is repeated ETLMerge upload failures, commonly logged as `failed do write` followed by `failed to upload file to MO`, with the underlying error `sql: database is closed`.

### Expected Behavior

The ETL SQL writer should keep the existing DB pool usable until a replacement pool is fully ready. Failed refresh or reconnect attempts should not poison the global DB pointer, and `Ping()` failures should propagate the real error to callers.

### Steps to Reproduce

1. Initialize the ETL SQL writer DB holder with a current global `*sql.DB`.
2. Trigger `GetOrInitDBConn(true, true)` or a refresh after `DBRefreshTime`.
3. Make the new connection path fail during address selection, `sql.Open`, or new connection `Ping()`.
4. Call the ETL write path again.
5. Observe that the old implementation can reuse a closed DB handle and keep returning `sql: database is closed`.

### Additional information

The fix should rebuild the new connection first, validate it, atomically swap it into the global holder only after it is ready, and keep ordinary refresh callers on the existing connection while a refresh is in progress.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25136

This PR fixes the ETL SQL writer DB holder lifecycle so reconnect and refresh failures do not leave ETLMerge stuck on a closed `*sql.DB` handle.

- Builds and validates the replacement DB connection before swapping it into the global holder.
- Clears the global DB pointer before explicit close, so callers cannot keep loading a closed handle.
- Keeps using the existing DB connection when ordinary refresh fails.
- Avoids blocking ordinary callers on an in-progress refresh when an old connection is still available.
- Rechecks reconnect state after acquiring the rebuild lock, so callers queued behind a successful forced reconnect reuse the fresh DB pool instead of serially rebuilding more pools.
- Delays stale connection close after swap to reduce the window where a goroutine that already loaded the old pointer immediately hits `sql: database is closed`.
- Resets reconnect backoff after successful forced reconnect and successful SQL write, so recovery does not keep rebuilding a new DB pool on every healthy batch during the remaining backoff window.
- Returns the real `Ping()` error from `WriteRowRecords`.
- Adds regression coverage for address lookup failure, `sql.Open` failure, new connection `Ping` failure, refresh fallback, refresh-in-progress fallback, explicit close cleanup, `Ping()` error propagation, backoff reset after recovery, and concurrent forced reconnect coalescing.

Validation on `3.0-dev`:

```text
go test ./pkg/util/export/etl/db
go test -race ./pkg/util/export/etl/db
make clean && make build
CGO_CFLAGS="-I/home/xupeng/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/home/xupeng/matrixone/thirdparties/install/lib -Wl,-rpath,/home/xupeng/matrixone/thirdparties/install/lib" go test ./pkg/util/export/etl
```

### DIFF
--- a/pkg/util/export/etl/content.go
+++ b/pkg/util/export/etl/content.go
@@ -219,6 +219,8 @@ func (f *SQLFlusher) FlushBuffer(buf *bytes.Buffer) (int, error) {
 	}
 	if err != nil {
 		_ = db_holder.DBConnErrCount.Count()
+	} else {
+		db_holder.DBConnErrCount.Reset()
 	}
 
 	return 0, err
--- a/pkg/util/export/etl/content_test.go
+++ b/pkg/util/export/etl/content_test.go
@@ -186,3 +186,41 @@ func TestWriteRowRecords_WithBackoff(t *testing.T) {
 
 	require.True(t, newConn)
 }
+
+func TestSQLFlusher_ResetBackoffAfterSuccessfulFlush(t *testing.T) {
+	old := db_holder.DBConnErrCount
+	defer func() {
+		db_holder.DBConnErrCount = old
+	}()
+
+	db_holder.DBConnErrCount = db_holder.NewReConnectionBackOff(time.Hour, 0)
+	db_holder.DBConnErrCount.Count()
+
+	db, mock, err := sqlmock.New()
+	require.NoError(t, err)
+	defer db.Close()
+
+	for i := 0; i < 3; i++ {
+		mock.ExpectExec(regexp.QuoteMeta(`LOAD DATA INLINE FORMAT='csv', DATA='record1' INTO TABLE testDB.testTable FIELDS TERMINATED BY ','`)).
+			WillReturnResult(sqlmock.NewResult(1, 1))
+	}
+
+	var forceNewConnFlags []bool
+	stubs := gostub.Stub(&db_holder.GetOrInitDBConn, func(forceNewConn bool, randomCN bool) (*sql.DB, error) {
+		forceNewConnFlags = append(forceNewConnFlags, forceNewConn)
+		return db, nil
+	})
+	defer stubs.Reset()
+
+	f := &SQLFlusher{
+		database: "testDB",
+		table:    "testTable",
+	}
+	for i := 0; i < 3; i++ {
+		_, err = f.FlushBuffer(bytes.NewBufferString("record1"))
+		require.NoError(t, err)
+	}
+
+	require.Equal(t, []bool{true, false, false}, forceNewConnFlags)
+	require.NoError(t, mock.ExpectationsWereMet())
+}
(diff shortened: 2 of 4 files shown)


XuPeng-SH:
Addressed. I added a reset path for the reconnect backoff after a successful forced reconnect and after successful ETL SQL writes, including SQLFlusher. This prevents a failure burst from forcing a new DB pool on every healthy batch for the rest of the backoff window. Added focused regression coverage for the burst-then-success case so subsequent successful writes use forceNewConn=false.

XuPeng-SH:
Addressed. After acquiring dbRebuildMux, GetOrInitDBConn now re-snapshots the current DB connection and checks whether reconnect backoff has already been reset by a successful reconnect. If so, queued stale forceNewConn=true callers return the fresh current pool instead of serially rebuilding more pools. Added TestGetOrInitDBConn_ConcurrentForceReconnectBuildsOnce to cover multiple concurrent forced reconnect callers and assert only one openDBConn/rebuild happens.

aptend (approved):
LGTM ✅

核心改进正确：先验证新连接可用再 swap 旧连接，消除了 `sql: database is closed` 窗口。

**关键亮点：**
- `buildDBConn` 独立构建并 Ping 验证，失败时不影响现有连接
- `staleDBConnCloseDelay` 30s 延迟关闭旧连接，给已加载旧指针的 goroutine 留出时间
- `TryLock` 非阻塞 refresh，refresh 进行中时返回当前连接
- Force reconnect coalescing：获取 rebuild 锁后重新检查，避免串行重建
- Backoff 重置逻辑覆盖完整（force reconnect 成功、WriteRowRecords 成功、FlushBuffer 成功）
- 测试覆盖充分（+395 行），覆盖了地址查找失败、Open/Ping 失败、并发重建合并等场景

**小建议（非阻塞）：**
1. `staleDBConnCloseDelay = 30s` — 如果 ETL 单次批量写入可能超过 30s，旧连接上的 in-flight 查询仍会被中断，建议确认是否足够
2. `rebuildDBConn` 中 refresh 失败且 `!forceNewConn` 时返回 current 但未更新 `dbRefreshTime`，下次调用会再次触发 refresh，行为上可接受但如果地址服务持续不可用可能产生不必要的重建开销

整体是一个高质量的修复，思路清晰，测试充分。

PR #25138, opened 2026-06-24
fix: keep ETL DB conn valid during reconnect

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Refs #25136
Cherry-pick of #25137 to `4.0-dev`.

This cherry-picks the ETL SQL writer DB holder reconnect fix to `4.0-dev`.

- Builds and validates the replacement DB connection before swapping it into the global holder.
- Clears the global DB pointer before explicit close, so callers cannot keep loading a closed handle.
- Keeps using the existing DB connection when ordinary refresh fails.
- Avoids blocking ordinary callers on an in-progress refresh when an old connection is still available.
- Rechecks reconnect state after acquiring the rebuild lock, so callers queued behind a successful forced reconnect reuse the fresh DB pool instead of serially rebuilding more pools.
- Delays stale connection close after swap to reduce the window where a goroutine that already loaded the old pointer immediately hits `sql: database is closed`.
- Resets reconnect backoff after successful forced reconnect and successful SQL write, so recovery does not keep rebuilding a new DB pool on every healthy batch during the remaining backoff window.
- Returns the real `Ping()` error from `WriteRowRecords`.
- Adds regression coverage for address lookup failure, `sql.Open` failure, new connection `Ping` failure, refresh fallback, refresh-in-progress fallback, explicit close cleanup, `Ping()` error propagation, backoff reset after recovery, and concurrent forced reconnect coalescing.

Validation on `4.0-dev`:

```text
go test ./pkg/util/export/etl/db
go test -race ./pkg/util/export/etl/db
CGO_CFLAGS="-I/home/xupeng/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/home/xupeng/matrixone/thirdparties/install/lib -Wl,-rpath,/home/xupeng/matrixone/thirdparties/install/lib" go test ./pkg/util/export/etl
```

### DIFF
--- a/pkg/util/export/etl/content.go
+++ b/pkg/util/export/etl/content.go
@@ -219,6 +219,8 @@ func (f *SQLFlusher) FlushBuffer(buf *bytes.Buffer) (int, error) {
 	}
 	if err != nil {
 		_ = db_holder.DBConnErrCount.Count()
+	} else {
+		db_holder.DBConnErrCount.Reset()
 	}
 
 	return 0, err
--- a/pkg/util/export/etl/content_test.go
+++ b/pkg/util/export/etl/content_test.go
@@ -186,3 +186,41 @@ func TestWriteRowRecords_WithBackoff(t *testing.T) {
 
 	require.True(t, newConn)
 }
+
+func TestSQLFlusher_ResetBackoffAfterSuccessfulFlush(t *testing.T) {
+	old := db_holder.DBConnErrCount
+	defer func() {
+		db_holder.DBConnErrCount = old
+	}()
+
+	db_holder.DBConnErrCount = db_holder.NewReConnectionBackOff(time.Hour, 0)
+	db_holder.DBConnErrCount.Count()
+
+	db, mock, err := sqlmock.New()
+	require.NoError(t, err)
+	defer db.Close()
+
+	for i := 0; i < 3; i++ {
+		mock.ExpectExec(regexp.QuoteMeta(`LOAD DATA INLINE FORMAT='csv', DATA='record1' INTO TABLE testDB.testTable FIELDS TERMINATED BY ','`)).
+			WillReturnResult(sqlmock.NewResult(1, 1))
+	}
+
+	var forceNewConnFlags []bool
+	stubs := gostub.Stub(&db_holder.GetOrInitDBConn, func(forceNewConn bool, randomCN bool) (*sql.DB, error) {
+		forceNewConnFlags = append(forceNewConnFlags, forceNewConn)
+		return db, nil
+	})
+	defer stubs.Reset()
+
+	f := &SQLFlusher{
+		database: "testDB",
+		table:    "testTable",
+	}
+	for i := 0; i < 3; i++ {
+		_, err = f.FlushBuffer(bytes.NewBufferString("record1"))
+		require.NoError(t, err)
+	}
+
+	require.Equal(t, []bool{true, false, false}, forceNewConnFlags)
+	require.NoError(t, mock.ExpectationsWereMet())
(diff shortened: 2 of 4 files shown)


XuPeng-SH:
I re-checked this PR from a correctness / unhappy-path angle. The closed-handle bug is fixed, but I still see one real resource-management regression in the reconnect path.

### Blocker: delayed stale-handle close can create too many live DB pools during reconnect storms

The new swap logic keeps the previous `*sql.DB` alive for `staleDBConnCloseDelay = 30s` after every successful rebuild.

At the same time, `WriteRowRecords()` still does:
- `forceNewConn=true` whenever `!DBConnErrCount.Check()`
- and `Check()` does **not** reset on success; it stays false until the full backoff window expires (`1m` by default)

That means after one burst of failures pushes `DBConnErrCount` over its threshold, a healthy batch stream can still spend up to a minute rebuilding a brand-new `sql.DB` on **every** batch while each retired pool stays open for 30 seconds.

Because `buildDBConn()` pings before the swap, each retired `sql.DB` can own a real live TCP connection, and `MaxConnectionNumber=1` is only per pool, not global. So this can amplify into a transient pileup of extra DB pools/connections during recovery.

### Suggested fix
At least one of these needs to happen:

1. **Reset the backoff state after a successful reconnect / successful write**, so we do not keep forcing `forceNewConn=true` on every subsequent healthy batch during the whole window.
2. Replace the fixed 30s delayed-close scheme with a more bounded lifecycle for retired pools (generation / refcount / in-flight tracking), so recovery cannot accumulate stale pools.

A focused regression test would be helpful too: one failure burst followed by multiple successful writes should not keep rebuilding a new `sql.DB` for every write during the remaining backoff window.

(I'm leaving this as a regular comment because GitHub does not let me submit `request changes` on the author's own PR.)

XuPeng-SH:
I re-checked the latest head.

The earlier backoff-reset concern looks improved now, but there is still one remaining unhappy-path blocker in the reconnect path:

### Blocker: concurrent forced reconnects can still serially rebuild multiple DB pools

`forceNewConn` is decided **before** `dbRebuildMux` is acquired. So if several goroutines cross the backoff threshold together, they can all enter `GetOrInitDBConn(true, true)` and queue on the rebuild mutex.

After the first goroutine successfully rebuilds and resets the backoff, the queued goroutines still rebuild again, because `rebuildDBConn()` trusts the caller's original `forceNewConn=true` decision and does not re-check whether a fresh current connection is already installed.

Since each rebuild keeps the previous pool alive for `staleDBConnCloseDelay = 30s`, a reconnect storm can still stack multiple live `*sql.DB` pools / TCP connections and add unnecessary reconnect latency.

### Suggested fix
- After acquiring `dbRebuildMux`, re-snapshot the current connection / backoff state and short-circuit if another goroutine already rebuilt successfully.
- Please also add a concurrent regression test: multiple callers all force reconnect at once, but only **one** `openDBConn` / rebuild should happen.

So I would still **not approve yet** on the current head.

XuPeng-SH:

The remaining reconnect-path blocker I had called out earlier looks resolved now:
- successful reconnect/write paths reset the backoff state,
- queued `forceNewConn` callers re-check after taking the rebuild lock and reuse the first successful rebuild,
- and the focused concurrent reconnect coverage is in place.

At this point I do **not** see a remaining substantive blocker in the current version, so from my side this is now approvable.

xzxiong:
修改后的 db 怎么管理？

PR #25147, opened 2026-06-25
cherrypick: fix lower case db name

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25079

修改 protected database  lower case db name 问题


daviszhen:
@mergify refresh

XuPeng-SH (approved):

The change is tightly scoped and the risky lower-case/protected-db interaction I would normally worry about is not present here on the current diff. I do not see a remaining substantive blocker in this version.

LGTM.

heni02 (approved):

daviszhen:
@mergify refresh

PR #25148, opened 2026-06-25
cherrypick: fix lower case db name

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/25079

修改 protected database  lower case db name 问题


daviszhen:
@mergify refresh

XuPeng-SH (approved):
Re-checked the latest head. The previous blockers are fixed: `protected_databases` parsing no longer substitutes empty elements with the current database, and `lower_case_table_names=2` is now treated consistently with the engine’s case-insensitive identifier semantics. The added unit/BVT coverage also pins those boundary cases, so I do not see a remaining correctness issue that should keep this blocked.

heni02 (approved):

Issue #23751, opened 2026-02-15
[Feature Request]: Enhance data branch diff with fallback, delta output, and summary statistics


XuPeng-SH:
## Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

## Is your feature request related to a problem?

Currently `data branch diff` has limitations that make it less flexible than Git's diff:

1. **Only works with branched tables**: Requires LCA (Lowest Common Ancestor) relationship, cannot compare arbitrary tables
2. **Limited output modes**: Only supports full output and `output count`, lacks incremental change tracking
3. **No detailed statistics**: `output count` only returns total rows, not breakdown of changes

### 1. Fallback to EXCEPT for non-branched tables

**Current behavior:**
```sql
-- Only works if t2 was created from t1
data branch create table t2 from t1{snapshot="sp1"};
data branch diff t2 against t1;  -- ✅ Works (has LCA)

-- Fails if no branch relationship
data branch diff t3 against t4;  -- ❌ Error: no LCA
```

**Proposed behavior:**
```sql
-- Automatically fallback to EXCEPT-like behavior
data branch diff t3 against t4;  -- ✅ Works (uses set difference)
```

### 2. Delta output mode

**Proposed syntax:**
```sql
data branch diff t1 against t2 output delta;
```

**Expected result:**
```
operation | a | b
----------|---|---
+         | 3 | 3  -- Added in t1
-         | 4 | 4  -- Deleted from t1 (exists in t2)
```

Similar to Git's `+`/`-` notation for added/removed lines.

### 3. Summary statistics output

**Proposed syntax:**
```sql
data branch diff t1 against t2 output summary;
```

**Expected result:**
```
metric    | count
----------|------
added     | 10
deleted   | 5
modified  | 3
unchanged | 100
```

Provides quick overview without scanning all rows.

### Fallback mechanism
- Check if tables have LCA relationship
- If no LCA: use EXCEPT-based comparison
- Maintain same output format for consistency

### Delta output
- Add `operation` column: `+` (added), `-` (deleted), `~` (modified)
- Use primary key to match rows for modification detection
- If no PK: treat as add/delete only

### Summary statistics
- Leverage existing diff computation
- Aggregate results into statistics
- Minimal performance overhead (single pass)

### Use Cases

1. **Data validation**: Compare staging vs production tables without branch relationship
2. **Change tracking**: Quickly see what changed between snapshots
3. **Monitoring**: Track data drift with summary statistics
[...]

PR #25149, opened 2026-06-25
fix data branch no-op update diff

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Related to #23751

Fixes data branch diff reporting no-op updates after a row is updated and then restored to the LCA/snapshot value.

Root cause: `findDeleteAndUpdateBat` treated a tombstone plus a matching live row for the same PK as an UPDATE based only on PK matching. It did not compare the decoded replacement tuple with the LCA delete row, so rows whose visible values had returned to the LCA value still appeared in diff output.

`DECIMAL256` also needs explicit normalization in this tuple-vs-vector path. `BranchHashmap.DecodeRow` can return Decimal256 tuple values as raw bytes, while `compareValueFromVector` returns a typed `types.Decimal256`; without normalizing the raw tuple value, equal restored rows can fail with a mismatched-type error instead of being suppressed as no-op updates.

Changes:

- Compare the decoded live tuple with the LCA row before emitting an UPDATE, and suppress the update when all visible non-PK values are equal.
- Reuse the existing vector comparison path by factoring value extraction/normalization helpers for vector-vs-vector and tuple-vs-vector comparison.
- Normalize raw `T_decimal256` tuple values with the existing fixed-width normalization helper.
- Keep fake-PK tables conservative by comparing the full visible row.
- Add focused frontend unit tests for restored rows, real updates, fake-PK behavior, Decimal256 normalization, and the real `BranchHashmap.GetByVectors` / `DecodeRow` raw-bytes path.
- Add `git4data/branch/diff_14` BVT coverage for restored-update diffs and merge direction semantics.

Validation:

- `./run.sh -p /Users/ghs-mo/MOWorkSpace/matrixone-2nd/test/distributed/cases/ -m genrs -i diff_14`
- `./run.sh -p /Users/ghs-mo/MOWorkSpace/matrixone-2nd/test/distributed/cases/ -i diff_14` - 58/58 passed
- `CGO_CFLAGS="-I$PWD/cgo -I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/cgo -lmo -L$PWD/thirdparties/install/lib -Wl,-rpath,$PWD/lib" DYLD_LIBRARY_PATH="$PWD/lib:$PWD/cgo:$PWD/thirdparties/install/lib" go test ./pkg/frontend -run 'TestHashDiff_HasLCA|TestHashDiff_NoLCAWithStubHandles|TestDataBranchCompareSingleValInVector'`
- `CGO_CFLAGS="-I$PWD/cgo -I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/cgo -lmo -L$PWD/thirdparties/install/lib -Wl,-rpath,$PWD/cgo -Wl,-rpath,$PWD/thirdparties/install/lib" DYLD_LIBRARY_PATH="$PWD/cgo:$PWD/thirdparties/install/lib" go test ./pkg/frontend -run 'TestCompareTupleValueWithVectorNormalizesValues|TestCompareTupleValueWithVectorDecimal256|TestCompareSingleValInVector_AllTypes|TestNormalizeCompareValueSupportedForms|TestDataBranchCompareValueErrors' -count=1`


heni02 (approved):

XuPeng-SH (changes requested):
I found one concrete blocker here.

The helper refactor regressed `DECIMAL256` comparison support in `pkg/frontend/data_branch_helpers.go`.

`compareValueFromVector()` still handles `decimal64` and `decimal128`, but not `decimal256`, and `compareSingleValueByType()` also whitelists `types.Decimal64` / `types.Decimal128` but not `types.Decimal256`. That means diff/merge/conflict handling on tables with `DECIMAL256` columns will now fail with an internal unsupported-type error instead of comparing values.

This is not hypothetical: the existing repo tests already show `DECIMAL256` is expected to work in this comparison path (`pkg/frontend/data_branch_test.go` has a `decimal256` case for `compareSingleValInVector`). The new helper path has simply dropped that type.

Suggested fix:
1. restore `types.T_decimal256` handling in `compareValueFromVector()`;
2. include `types.Decimal256` in `compareSingleValueByType()`;
3. add a regression for the new tuple-vs-batch comparison path as well, not just the existing single-value vector comparison.

XuPeng-SH (changes requested):
The new no-op update suppression path is still broken for `DECIMAL256` rows.

`findDeleteAndUpdateBat` now compares the decoded hashmap tuple against the LCA batch row via `compareTupleWithBatchRow -> compareTupleValueWithVector`. For `DECIMAL256`, `compareValueFromVector` returns a typed `types.Decimal256`, but the tuple side still comes back from `BranchHashmap.DecodeRow` as raw `[]byte` because `branch_hashmap.go` falls back to `GetRawBytesAt` for types it does not encode explicitly. `normalizeCompareValue` has normalization for the other raw-byte fallback types (`Rowid`, `Blockid`, `TS`, `year`, etc.), but it still has no `T_decimal256` case, so the left side stays `[]byte` and `compareSingleValueByType` fails with a mismatched-type error instead of recognizing an equal restored row as a no-op.

So a table that hits this restored-row/update path with a `DECIMAL256` column can still fail the diff/merge instead of suppressing the no-op. Please normalize raw `DECIMAL256` bytes the same way as the other fallback-encoded fixed-width types, and add a regression that exercises the real `DecodeRow`/raw-bytes path. The current decimal256 test only covers already-typed `types.Decimal256` inputs, so it misses the broken production path.

After that, an end-to-end diff BVT for a reverted `DECIMAL(40,4)` row would be a good follow-up coverage add. The fake-PK revert path also only has unit coverage today.

gouhongshen:
Addressed the Decimal256 review in `a0132da4f3`.

Changes made:
- Added `T_decimal256` normalization in `normalizeCompareValue` using the existing fixed-width normalization helper.
- Added regression coverage through the real `BranchHashmap.GetByVectors` / `DecodeRow` path, including an assertion that decoded Decimal256 tuple values are raw `[]byte` before comparing against the typed batch vector.
- Verified `compareTupleWithBatchRow` now treats the decoded raw Decimal256 tuple and typed Decimal256 vector value as equal.

Local verification passed:

```sh
CGO_CFLAGS="-I$PWD/cgo -I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/cgo -lmo -L$PWD/thirdparties/install/lib -Wl,-rpath,$PWD/cgo -Wl,-rpath,$PWD/thirdparties/install/lib" DYLD_LIBRARY_PATH="$PWD/cgo:$PWD/thirdparties/install/lib" go test ./pkg/frontend -run 'TestCompareTupleValueWithVectorNormalizesValues|TestCompareTupleValueWithVectorDecimal256|TestCompareSingleValInVector_AllTypes|TestNormalizeCompareValueSupportedForms|TestDataBranchCompareValueErrors' -count=1
```

I kept the end-to-end reverted `DECIMAL(40,4)` diff BVT as follow-up coverage, as suggested, so this PR stays focused on the production bug path and unit regression.

XuPeng-SH (approved):
No blocking findings from this round. The previous Decimal256 blockers are addressed in the actual production path: vector-vs-vector comparison handles Decimal256, and the new tuple-vs-batch no-op suppression path now normalizes the raw []byte returned by BranchHashmap.DecodeRow before comparing against typed Decimal256 vectors.\n\nI also rechecked the LCA no-op suppression flow, fake-PK behavior, batch ownership/error paths, and unsupported-type behavior. Local targeted verification passed:\n\n- go test ./pkg/container/types -run TestEncodeDecodeValue -count=1 -timeout 180s\n- go test ./pkg/frontend -run 'TestHashDiff_HasLCA|TestHashDiff_NoLCAWithStubHandles|TestCompareTuple|TestDataBranchCompareSingleValInVector|TestCompareSingleValInVector|TestNormalizeCompareValue|TestDataBranchCompareValueErrors|TestCompareRowInWrappedBatches|TestDiffDataHelper' -count=1 -timeout 300s\n- git diff --check origin/main...HEAD\n\nCI is green. The branch is still behind main, so it should be updated before merge if required by the queue.

Issue #25126, opened 2026-06-24
[Bug]:Chaos TPCC (7cbeacd): durable consistency corruption — Payment ambiguous commit (20638) + NewOrder duplicate o_id (1062)


heni02:
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-chaos-main-7cbeacd-202606231155` |
| Namespace | `mo-chaos-main-7cbeacd-202606231155` |
| Build / image | `matrixone:commit-7cbeacd` |
| Endpoint | `10.222.6.252:6001` |
| Artifact root | `/data1/chaos-test/28001018780_7cbeacd` |
| Workload | TPCC 10 warehouses / 10 terminals + chaos (kill CN/DN/Log, mem/network stress) |
| Analysis window | **2026-06-23 11:55 CST** → **2026-06-24 ~14:48 CST** (~27h) |
| Soak control (same commit) | `mo-stb-main-7cbeacd-202606231152` — **0 consistency FAIL, 0×1062** |

**Scope:** **Permanent** TPCC `ConsistencyCheck` failures under chaos — warehouse/district `ytd` divergence and `D_NEXT_O_ID` vs `max(o_id)` desync, driven by **Payment commit-status-unknown (20638)** and **NewOrder duplicate primary key (1062)** on `bmsql_oorder`.

**Out of scope:** Transient read-skew-only false positives without 1062/20638 (see [83cc163 read-skew issue](matrixone-chaos-83cc163-tpcc-consistency-read-skew.md)); chaos orchestrator `kubectl delete chaos` CRD typo; CDC matplotlib env failure.

## Summary

During chaos long-run on `commit-7cbeacd`, TPCC `runVerify.sh` / `io.mo.ConsistencyCheck` reported **5506 failures** vs **1286 passes** (**81% fail rate**). After **2026-06-24 02:27:58 CST**, checks **never recovered** (**5505 consecutive FAIL**).

Dominant failing invariants:

```sql
-- warehouse vs district payment totals (permanent drift, w_id=7 example)
(SELECT d_w_id, sum(d_ytd) FROM bmsql_district GROUP BY d_w_id)
EXCEPT (SELECT w_id, w_ytd FROM bmsql_warehouse);

-- order ID counter vs max order (from ~02:55 CST)
(SELECT d_w_id, d_id, D_NEXT_O_ID - 1 FROM bmsql_district)
[...]

Issue #22791, opened 2025-11-07
[Bug]: chaos test: tpcc ConsistencyCheck failed.


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

3.0-dev

### Commit ID

ebbd739edee5b673710f6dee9d923657254456b4

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

<img width="924" height="977" alt="Image" src="https://github.com/user-attachments/assets/59e13fe5-dd38-445f-b5b4-984339059c54" />

log：https://shanghai.idc.matrixorigin.cn:30001/explore?panes=%7B%22Y46%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-chaos-2d3495d-202511041027%5C%22%7D%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221762350425000%22,%22to%22:%221762350460000%22%7D%7D%7D&schemaVersion=1&orgId=1

工具日志：/data1/chaos-test/19132680927_ebbd739/mo-chaos/test-tool/mo-tpcc

故障日志： cn/dn内存压力，cn网络带宽限制

### Steps to Reproduce

```Markdown
trigger chaos test
```

XuPeng-SH:
fixed

XuPeng-SH:
@Ariznawlll please help to verify

heni02:
最新 4.0-dev 分支 chaos 测试也跑出来不一致的问题
TPCC district-level oorder/order_line consistency check false positive — cross-partition read snapshot skew under high concurrency

## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-chaos-83cc163-202605281226` |
| Namespace | `mo-chaos-83cc163-202605281226` |
| Build / image | `matrixone:commit-9121876d0` |
| Endpoint | `10.222.6.253:6001` |
| Artifact root | `/data1/chaos-test/26554461708_83cc163` |
| Workload | TPCC 10 warehouses / 1000 terminals + chaos (kill CN/DN/Log, mem/network stress) |
[...]

PR #25150, opened 2026-06-25
fix lockservice orphan remote txn validation

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25126
issue #22791

Remote lock orphan cleanup used to treat a missing remote CN active transaction as enough proof that the holder could be unlocked. That is unsafe when the commit response is lost after TN has already accepted or is processing the commit: the waiter can release the holder lock with an empty commit timestamp and continue on an old snapshot.

This PR makes the remote orphan check ask the allocator before releasing such a holder. If the allocator reports the txn is still committing, lockservice keeps the holder valid and does not unlock it as an orphan. Only txns confirmed as not committing continue down the orphan unlock path.

A unit test covers the inactive-remote-txn plus committing-in-allocator case so the waiter orphan cleanup path does not regress.

## Test

```
GOCACHE=/private/tmp/mo-gocache-issue25126-pr go test ./pkg/lockservice -run 'TestValidTxnWith|TestCanUnlockRemoteTxn' -count=1
```


XuPeng-SH (approved):
Re-checked the current head. This fix is narrow and correct: a remote active-txn miss is no longer treated as enough proof to orphan-unlock the holder, and the added regression test covers the exact race where the allocator still reports the txn as committing. I do not see a remaining correctness issue that should keep this blocked.

iamlinjunhong (approved):

Issue #25129, opened 2026-06-24
[Bug] TPCC 1000W/1000T fails with CN i/o timeout under high memory pressure (recurrence of #24893)


Ariznawlll:
## Issue

TPCC 1000 warehouses / 1000 terminals benchmark fails with `ErrorCode: 20505` (`i/o timeout`) during the 5th test round when all CN pods reach peak memory (~46–49 GB).

This is a recurrence of the same failure pattern reported in #24893 (closed 2026-06-11).

## Failed Run

- **Job**: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28037375178/job/83026685448
- **Namespace**: `mo-main-commit-435e7da55-20260623`
- **Error time**: `2026-06-23 18:46:17 UTC`
- **Commit**: `435e7da55`

## Error

```
FATAL jTPCCTerminal:214 - [UNEXPECTED][TT_PAYMENT][EXECUTION] ErrorCode : 20505,
ErrorMessage : cannot connect to backend: writeto tcp4 10.143.35.187:21928->10.143.35.185:6003:
read tcp4 10.143.35.187:21928->10.143.35.185:6003: i/o timeout
```

- 88 occurrences of this error, 44 transaction errors
- tpmE = 8.72, tpmC = 7001.16

## Pod IP Mapping

| Pod | IP |
|---|---|
| `tp-cn-qk2rd` | 10.143.35.185 (unresponsive CN) |
| `tp-cn-xw7xl` | 10.143.35.187 (proxy/source CN) |
| `tp-cn-zwjff` | 10.143.35.186 |

### 1. Memory Pressure

At the time of failure (TPCC round 5 peak), all 3 CN pods were at **~46–49 GB** memory:

| CN | Memory at Error | Peak |
|---|---|---|
| qk2rd (185) | 46.2 GB | 49.4 GB |
| zwjff (186) | 46.4 GB | 48.6 GB |
| xw7xl (187) | 47.8 GB | 48.4 GB |

Memory oscillates with each TPCC 5-minute round — rises during the test, drops during transitions. The failure consistently occurs at the memory peak of rounds 4–5.

### 2. GC STW Pauses

| CN | Max GC Pause | Risk |
|---|---|---|
| **qk2rd (185)** | **363 ms** | **High** |
| zwjff (186) | 125 ms | Medium |
| xw7xl (187) | 111 ms | Medium |

[...]

PR #25157, opened 2026-06-25
fix: avoid reporting dedup load sentinel on cache miss

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25129

## What changed

  Avoid creating the `dedup load did not complete` sentinel error when a `dedupLoad` owner is first registered.

  The previous code initialized `call.err` with a `moerr.NewInternalErrorNoCtx(...)` sentinel before `load()` ran. Since `moerr` reports errors when they are created, normal cache misses could emit `dedup load did not
  complete` logs even when the S3/object metadata load later succeeded.

  This patch adds a `completed` flag to `loadCall`:

  - `completed` is set only after `load()` returns normally.
  - real `load()` errors are still stored in `call.err` and returned to waiters.
  - waiters create `dedup load did not complete` only if the owner finished abnormally before `load()` returned.
  - no panic recovery is added.

  ## Why

  TPCC runs showed thousands of `dedup load did not complete` logs while all transactions succeeded and consistency checks passed. That indicates these logs were largely sentinel creation noise, not actual dedup load
  failures.


XuPeng-SH (changes requested):
I re-checked current head `09fb0656d2ce2728b1131725bb7c394c053497ca` from the stated bug and unhappy-path angle.

Blocker: owner panic is still not exposed to waiters.

The PR description says the owner should convert the panic value into `call.err` before re-panicking, so concurrent waiters can see the real failure reason. The current code does not do that. In `dedupLoad`, `call.completed` is set only after `load()` returns normally. If `load()` panics, the defer only closes `call.done` and deletes the call; `call.err` remains nil and `call.completed` remains false. Waiters then still return the generic `dedup load did not complete` error.

The existing panic test also still asserts that sentinel error, so the stated behavior is not covered by a regression test.

Suggested direction:

- Recover inside the owner defer, set `call.err = moerr.ConvertPanicError(ctx, recovered)` before closing `call.done`, then re-panic to preserve owner behavior.
- Update `TestDedupLoadCleansUpAfterPanic` or add a focused test so the waiter observes the converted panic/root-cause error instead of `dedup load did not complete`.

`git diff --check` and merge-tree are clean. Local `go test ./pkg/objectio -run TestDedupLoad -count=1` cannot build in my environment because native headers (`usearch.h`, `xxhash.h`) are missing, but this blocker follows directly from the current control flow.

XuPeng-SH (approved):
I checked this from the dedup path, cache-miss behavior, and unhappy paths.

This looks like a false positive in error reporting rather than a real dedup correctness bug. The old code constructed `moerr.NewInternalErrorNoCtx("dedup load did not complete")` before `load()` even ran, and `moerr` reports at construction time, so a normal owner-miss path could emit that internal error even when the load later succeeded. The new `completed` flag fixes that signal without changing the actual dedup semantics.

I don't see a substantive production-code issue in the fix itself, so I'm approving.

One non-blocking follow-up: the new tests cover waiter success/timeout behavior, but they do not directly assert that the bogus `dedup load did not complete` report is gone. A small regression test with a temporary error reporter would make this much tighter.

PR #25158, opened 2026-06-25
fix: cross-CN shuffle join INSERT...SELECT hang on multi-CN cluster (#24919)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24919

Fixes a silent deadlock where `INSERT ... SELECT` with cross-CN shuffle hash join (`shuffle: range(...)`) over large tables (>=5M rows/table) hangs forever on multi-CN clusters. The same statement completes in ~3-6s on a single-node deployment.

### Root cause

`newShuffleJoinScopeList` leaves a CN's `dop` join buckets as independent `RemoteRun` trees, while the shuffle dispatch attaches to only the first bucket. When `compileMultiUpdate` (used by large `INSERT...SELECT`) sends each bucket individually via `newMergeScope`:
1. `checkPipelineStandaloneExecutableAtRemote` detects the dispatch's `LocalRegs` pointing to out-of-tree sibling buckets → correctly returns false
2. `RemoteRun` converts the pipeline to local on the coordinator
3. The dispatch runs on the coordinator instead of its compile-time CN, mispaired with the cross-CN receiver's `FromAddr`
4. Remote receivers `GetProcByUuid` endlessly busy-spin (300s timeout) → idle `WaitingEnd` deadlock (uncancellable `context.TODO()`)

### Fix

Add `groupShuffleBucketsByCNIfNeeded` (gated: multi-CN + cross-CN dispatch subtree present) which reuses `mergeScopesByCN` to group same-CN shuffle buckets — together with their nested dispatch — into one per-CN send unit. When the whole group is sent via a single `RemoteRun` to its target CN:
- `checkPipelineStandaloneExecutableAtRemote` returns `true` (all dispatch `LocalRegs` are within the same tree)
- The dispatch runs on the correct CN, completing the cross-CN receiver handshake
- No changes to the `RemoteRun` protocol, dispatch operator, or `checkPipeline` logic

The grouping is a no-op for single-CN / non-shuffle inserts (confirmed by gating UT).

### Changes (2 files, +171 lines)

- `pkg/sql/compile/compile.go (+62)`: `groupShuffleBucketsByCNIfNeeded` + `scopeTreeHasCrossCNDispatch` helpers, wired into `compileInsert` and `compileMultiUpdate` (4 call sites: `toWriteS3` and `!toWriteS3` paths)
- `pkg/sql/compile/remoterun_test.go (+109)`: UT reproducing the `checkPipeline=false` pre-fix + verifying per-CN containers return `true` post-fix + gating test (no regressions for single-CN / non-shuffle)

### Verification

- **2-CN docker cluster** (`etc/docker-multi-cn-local-disk`): 5M rows × 2 consecutive runs complete in ~17s each, `count(*)=5,000,000`, data fully correct (`distinct_id=5,000,000`, `bad_pad=0`, `bad_k=0`). Previously hung forever.
- **Unit tests**: `-race` pass; new helper coverage **100%**; `go vet` clean; `go build ./pkg/sql/...` passes
- **Gating**: single-CN and non-shuffle paths unaffected (UT confirms no-op when no cross-CN dispatch)

### DIFF
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -3828,8 +3828,14 @@ func (c *Compile) compileInsert(ns []*plan.Node, n *plan.Node, ss []*Scope) ([]*
 			ss[i].setRootOperator(insertArg)
 		}
 		c.anal.isFirst = false
-		// keep a cross-CN shuffle dispatch in the same send unit as all its local buckets (issue #24919).
-		ss = c.groupShuffleBucketsByCNIfNeeded(ss)
+		// NOTE: intentionally do NOT group shuffle buckets by CN here. Unlike the write-S3 and
+		// multi-update paths, this non-S3 path keeps a per-bucket Insert as each scope's root
+		// operator. Wrapping the buckets into a per-CN Merge container would move those Insert
+		// operators into PreScopes, where run()/affectedRows() -- which only walk the RootOp
+		// child chain, not PreScopes -- would miss them, so the rows get written but the
+		// statement reports "0 rows affected". This path only handles small (non-S3) inserts,
+		// which do not produce cross-CN shuffle in practice, so the issue #24919 hang does not
+		// arise here.
 		return ss, nil
 	}
 
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -3838,6 +3838,10 @@ func (c *Compile) compileInsert(ns []*plan.Node, n *plan.Node, ss []*Scope) ([]*
 		// todo : pipelines with sink scan ,must refactor this in the future
 		currentFirstFlag := c.anal.isFirst
 		c.anal.isFirst = false
+		// dataScope merges the buckets, but dataScope.MergeRun still sends each bucket as an
+		// individual RemoteRun unit, so a cross-CN shuffle dispatch here would hit the same
+		// convert-to-local hang. Group same-CN buckets into one per-CN send unit first (issue #24919).
+		ss = c.groupShuffleBucketsByCNIfNeeded(ss)
 		dataScope := c.newMergeScope(ss)
 		if c.anal.qry.LoadTag {
 			// reset the channel buffer of sink for load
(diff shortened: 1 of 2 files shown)


XuPeng-SH (changes requested):
I re-checked the current head and found one substantive regression in the new CN-grouped insert path.

In the non-S3 branch, the new `groupShuffleBucketsByCNIfNeeded(ss)` call can replace each insert scope with a per-CN `Remote` merge wrapper. That wrapper’s root operator is just `Merge`; the real `Insert` operators remain only in `PreScopes`.

The execution path does not aggregate affected rows from those grouped child scopes for normal inserts:
- `newMergeScopeByCN()` builds a `Remote` scope with `Merge` as `RootOp` and moves the original insert scopes into `PreScopes`
- `run()` special-cases `MultiUpdate`, but for ordinary grouped insert scopes it still just does `c.addAffectedRows(s.affectedRows())`
- `affectedRows()` only walks the `RootOp` child chain, not `PreScopes`

So these grouped non-S3 inserts can write data successfully while reporting `0 rows affected`.

**Concrete suggestions**

1. Either avoid CN-grouping in the non-S3 insert path, or
2. aggregate affected rows recursively from grouped child scopes (the same way `run()` already does for `MultiUpdate`), or
3. teach `affectedRows()` / `run()` how to include `PreScopes` for this grouped-insert container pattern.

I would keep this at request changes because it changes visible DML semantics even when the data write itself succeeds.

XuPeng-SH (approved):
I re-checked the latest head (`cf95fb5`) from the previous affected-rows blocker, pipeline structure, and unhappy-path angles.

The earlier issue I requested changes for is addressed: the non-S3 insert path now calls `groupShuffleBucketsByCNIfNeeded` before attaching `Insert`, so the grouped per-CN container gets `Insert -> Merge` on its RootOp chain and `affectedRows()` can still see the Insert operator. The S3 insert and multi-update paths still keep their write/flush aggregation operators in the expected places.

I also checked the grouping gate and RemoteRun semantics:

- single-CN / no cross-CN dispatch remains a no-op;
- grouped same-CN shuffle buckets become one Remote send unit and pass `checkPipelineStandaloneExecutableAtRemote`;
- current-CN grouped containers still run locally through `ipAddrMatch`, remote-CN containers remain RemoteRun;
- `git diff --check origin/3.0-dev...HEAD` passed;
- `git merge-tree --write-tree HEAD origin/3.0-dev` passed.

I could not run the targeted local Go test because this environment is missing native `usearch.h`; GitHub 3.0 CI is green on the current head.

One residual note, not blocking this PR: the generic cross-CN dispatch empty-input/end-handshake path is still worth tracking separately, because this PR fixes the concrete #24919 shape by preventing the shuffle dispatch from being converted to the wrong local execution tree rather than changing dispatch cleanup itself.

aunjgr (approved):

PR #25160, opened 2026-06-25
FIX: Hnsw single thread build (port to 4.0 dev)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24977

fix the bug to keep ThreadsBuild to 1 in all build path. cherry pick from main

### DIFF
--- a/pkg/vectorindex/hnsw/build.go
+++ b/pkg/vectorindex/hnsw/build.go
@@ -51,19 +51,30 @@ type AddItem[T types.RealNumbers] struct {
 func NewHnswBuild[T types.RealNumbers](sqlproc *sqlexec.SqlProcess, uid string, nworker int32,
 	cfg vectorindex.IndexConfig, tblcfg vectorindex.IndexTableConfig) (info *HnswBuild[T], err error) {
 
-	// estimate the number of worker threads
-	nthread := 0
-	if nworker <= 1 {
-		// single database thread and set nthread to ThreadsBuild
-		nthread = int(vectorindex.GetConcurrency(tblcfg.ThreadsBuild))
-	} else {
-		// multiple database worker threads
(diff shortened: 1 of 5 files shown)


XuPeng-SH (changes requested):
I re-checked the current head. I do not see a new logic bug in the single-thread workaround itself, but I still think the regression coverage is below the bar for the exact behavior this PR is fixing.

Right now the only test that actually demonstrates the underlying USearch orphan-node problem is `pkg/vectorindex/hnsw/zz_orphan_test.go`, and it is still permanently skipped. The normal CI-visible tests do not really pin the fix:
- the build test only requires recall to stay above a broad threshold
- the sync tests mainly assert `RunOnce()` returns nil
- there is no deterministic assertion that the effective build concurrency is forced to 1 on the relevant paths

That means reverting `build.go` / `sync.go` back to multi-threaded add could still slip through CI while silently reintroducing the exact-match reachability bug this PR is supposed to prevent.

**Concrete suggestions**

1. Add a non-skipped unit test around the actual HNSW build/sync entry points that asserts the effective build/add concurrency is forced to 1 even when config asks for more.
2. Add a deterministic exact-self-match / reachability assertion in a CI-stable test if you have a small enough synthetic dataset for it.
3. At minimum, add focused tests for:
   - `GetConcurrencyForSingleThreadBuild()`
   - `NewHnswBuild(...)` choosing single-thread mode
   - `NewHnswSync(...)` overriding `hnsw_threads_build` to 1 on the CDC/sync path

I would keep this at request changes until the core workaround is protected by a non-skipped automated regression test, not only by a manual repro file.

XuPeng-SH (approved):

Issue #23176, opened 2025-11-29
[Bug]: cast TEXT to Varchar failed with larger value


zhangxu19830126:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

latest

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

internal error: Can't cast column from TEXT type to CHAR type because of one or more values in that column. Src length 260 is larger than Dest length 255.

### Expected Behavior

Text cast to varchar cannot failed by length is too bigger

### Steps to Reproduce

```Markdown
1. create a table with a text column
2. exec update statement, and using CONCAT(COALESCE(`column`, ''), ?) to update TEXT column
```

zhangxu19830126:
fixed

PR #24204, opened 2026-04-26
fix: reimplement MySQL type compatibility for 3.0-dev

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24184

## What this PR does / why we need it:

This PR reimplements the MySQL type compatibility updates needed by `3.0-dev` for issue #24184 without cherry-picking the previous backport commits.

It adds or wires:

- YEAR type basics, parsing, planner mapping, vector function support, and cast execution paths.
- DECIMAL256 parsing/formatting, scientific notation with `e+NN`, planner literal typing, and string / NULL cast execution paths.
- MySQL SET value normalization/parsing and SHOW type formatting.
- GEOMETRY planner/type mapping and binary charset handling.
- Float overflow behavior for `exp()` and float addition.
- Focused unit tests for the new type and cast behavior.

Validation:

- `make`
- `make static-check`
- `go test ./pkg/container/types -count=1`
- `go test ./pkg/sql/plan -count=1`
- `go test ./pkg/sql/plan/function -count=1`
- `go test ./pkg/sql/parsers/dialect/mysql -count=1`
- `go test ./pkg/compare ./pkg/sort -count=1`
- `go test ./pkg/frontend -run 'Test.*Year|Test.*Geometry|Test.*Set|TestMysqlProtocol|Test_format|Test_getDataFromPipeline' -count=1`
- `go test ./pkg/vm/engine/tae/index -run 'Test.*ZM|Test.*Zone' -count=1`

PR #25161, opened 2026-06-25
fix: restore TEXT cast length compatibility

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23176

PR #24204 unintentionally overwrote the TEXT cast compatibility behavior added by PR #23178. As a result, casting a TEXT source through CHAR/VARCHAR widths used by the planner could fail again with a length validation error, which breaks UPDATE/CONCAT flows for TEXT columns.

This PR restores the prior compatibility rule in `strToStr`: when the source type is TEXT and the target is CHAR/VARCHAR with width >= 255, skip the length check. Small-width casts such as CHAR(1), CHAR(10), and VARCHAR(10) still validate length and fail when the value is too long.

Validation:

- `go test ./pkg/sql/plan/function -run 'Test_strToStr_TextToCharVarchar|Test_strToStr_TextLengthCheck' -count=1`\n- `go test ./pkg/sql/plan/function -count=1`\n- `git diff --check`


XuPeng-SH (changes requested):
I re-checked the current head and still see one substantive correctness issue in the restored TEXT-cast compatibility rule.

Right now the new `shouldSkipLengthCheck` gate skips validation for **every** `TEXT -> CHAR/VARCHAR` cast with width `>= 255`. That is too broad. It restores the planner compatibility case you wanted, but it also disables real width enforcement for genuine bounded target columns such as `CHAR(255)` / `VARCHAR(300)`.

This is reachable in normal DML paths:
- `forceCastExpr(..., col.Typ)` is used for assignment / insert casts in `build_update.go` and `bind_insert.go`
- the only other string-length check I found is the literal-only insert path in `eval_expr_util.go`
- so expression-based `INSERT` / `UPDATE` into a real bounded `CHAR/VARCHAR` column can now silently accept overlength TEXT values

The current tests also lock in the bad case directly:
- `TEXT -> CHAR(255)` with length 260 succeeds
- `TEXT -> VARCHAR(255)` with length 260 succeeds

That is the planner compatibility shape you needed, but it is also exactly what should still fail for a true user-declared bounded column.

**Concrete suggestion**

Narrow the bypass to the exact planner-generated compatibility path(s), instead of using a generic `destLen >= 255` rule. If that information is not available at this layer, pass explicit context for the compatibility case so real target column widths continue to enforce their declared limit.

I would also add one SQL-level regression once the fix is narrowed: an expression-based `UPDATE` or `INSERT` into a real `VARCHAR(255)` target that must still reject a 260-character TEXT value.

PR #25163, opened 2026-06-25
fix(hnsw): restore multi-threaded build; vendor USearch #735 fix

cpegeric:
Concurrent USearch add() could orphan HNSW graph nodes (a vector stored but never linked, so search() couldn't reach it -> flaky recall@1). MatrixOne #24849 worked around it by forcing single-threaded builds everywhere. The root cause is fixed upstream-style in our vendored libusearch (two-pass add: form ALL forward links before ANY reverse link, so a node is never reachable as a descent seed while a lower level is still empty), so the workaround is no longer needed.

  - thirdparties/usearch-2.25.3.tar.gz: patched index.hpp with the https://github.com/unum-cloud/USearch/pull/772  fix (pristine v2.25.3 source, only index.hpp/test.cpp changed; CMakeLists still march=native so the Makefile's sed applies as before).
  - build.go: drop the hardcoded `nthread := 1`; restore the real concurrency estimate (GetConcurrency / GetConcurrencyForBuild from nworker/ThreadsBuild).
  - sync.go: CDC/sync paths use GetConcurrencyForBuild directly.
  - types.go: remove the GetConcurrencyForSingleThreadBuild stopgap.
  - zz_orphan_test.go: enable TestZZBuildOrphan as a regression guard — 30x 8-thread builds with the BVT t2 params (M 64, EF_CONSTRUCTION/SEARCH 200), rotating insertion order each run to mimic `load data ... parallel 'true'`; asserts 0 orphans. Auto-skips without the SIFT fixture (~5s when present).

Validated: zz_orphan_test 0/30 multi-threaded (was ~1/30 pre-fix); 1M wiki_all HNSW build clean (recall@10 82% at M=8); vector_hnsw_async t2 BVT 30/30.

## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue https://github.com/matrixorigin/matrixone/issues/24977

patch the fix with usearch PR: https://github.com/unum-cloud/USearch/pull/772
add unit test to test the same bvt case flaky test to make sure no orphan in final index.
revert the change build thread = 1


fengttt (approved):

heni02 (approved):

XuPeng-SH (changes requested):
I re-checked this from the correctness / unhappy-path angle. The vendored USearch #735 patch looks directionally right, and the focused orphan test passes locally, but restoring multi-threaded `HnswBuild` re-enables two MatrixOne-side lifecycle bugs that are not covered by the new tests.

Blocker 1: async worker errors can be lost at finalization, and later producers can also hang.

`Add()` only polls `err_chan` before enqueueing (`pkg/vectorindex/hnsw/build.go:158-168`). If a worker fails while processing the last queued item, that `Add()` already returned nil. `ToInsertSql()` then only calls `CloseAndWait()` and never drains/joins `err_chan` (`pkg/vectorindex/hnsw/build.go:271-284`), so the build can be finalized as successful even though a worker returned an error. The same channel lifecycle is also risky for unhappy paths such as context cancellation: workers can exit, but a producer may still block on `h.add_chan <- ...` because the send does not select on worker error or context cancellation.

Please make finalization return the joined worker error, e.g. have `CloseAndWait()`/`ToInsertSql()` drain `err_chan` after `wg.Wait()`, and make the enqueue path avoid blocking forever once workers have failed or the context is cancelled. A focused regression should cover a multi-threaded build where the final queued vector fails (dimension mismatch or cancelled context) and `ToInsertSql()` must return that error.

Blocker 2: capacity rollover can save/destroy an index while other workers still have in-flight adds assigned to it.

In the multi-thread path, `getIndexForAddSync()` reserves a slot under `h.mutex`, but the actual `idx.Add()` happens after the lock is released (`pkg/vectorindex/hnsw/build.go:178-239`). When another worker crosses `IndexCapacity`, it receives the previous index as `save_idx`, calls `SaveToFile()` outside the lock, and `SaveToFile()` saves then destroys `idx.Index` (`pkg/vectorindex/hnsw/model.go:144-224`). That can race with earlier workers that already reserved slots in the old index but have not finished `idx.Add()` yet. The result can be an incomplete persisted HNSW file, `usearch index is nil`, or native use-after-destroy behavior depending on the interleaving.

Please bound the old-index lifecycle before saving it: track per-index in-flight adds / generation completion, or otherwise ensure all adds assigned to an index have completed before `SaveToFile()` can save and destroy that index. A regression with small `IndexCapacity`, `nthread > 1`, and enough concurrent adds to force rollover should verify all keys survive and worker errors are surfaced.

Local checks I ran:

```sh
go test ./pkg/vectorindex/hnsw -run 'TestBuildMultiWorker|TestZZBuildOrphan' -count=1
```

Both passed, but they do not cover the two lifecycle cases above.

XuPeng-SH (approved):
Re-reviewed the latest head from the multi-threaded build / unhappy-path angle. The two lifecycle blockers I raised on the previous revision look fixed now.

What changed in this head:

- Worker failures are now recorded with `recordWorkerErr`, producers can stop once `stopped` is closed, and `ToInsertSql()` returns the recorded worker error after `CloseAndWait()`. This covers the “last queued item fails after Add returned nil” case.
- Rollover now reserves an `inflight` slot on the assigned index under `HnswBuild.mutex`, and waits for the rolled-over index's in-flight adds before `SaveToFile()` saves/destroys it. That addresses the save/destroy vs concurrent `idx.Add()` race.
- Focused regressions were added for both cases: final worker error propagation and small-capacity multi-worker rollover.

Local checks:

```sh
go test ./pkg/vectorindex/hnsw -run 'TestBuildMultiWorker|TestBuildMultiWorkerLastItemError|TestBuildMultiWorkerRollover|TestZZBuildOrphan' -count=1
go test -race ./pkg/vectorindex/hnsw -run 'TestBuildMultiWorkerRollover|TestBuildMultiWorkerLastItemError' -count=1
go test ./pkg/sql/colexec/table_function -run 'TestHnswCreate|TestHnswCreateF64' -count=1
```

All passed locally. LGTM.

PR #25173, opened 2026-06-26
test: expand protected database BVT coverage

Ariznawlll:
## Summary
- expand protected database BVT coverage for read/write permission behavior
- remove stale BVT issue tags for #25080 and #25079
- keep #25080 covered as existing behavior rather than an issue regression

## Validation
- go test ./pkg/frontend -run 'TestProtectedDatabase'\n- built mo-service on latest origin/3.0-dev\n- manually verified #25079 on 3.0-dev commit 9c99cbab2: with lower_case_table_names=1 and protected_databases=CamelDB, non-accountadmin CREATE DATABASE `CamelDB` and `cameldb` are both rejected with the protected database privilege error\n

### DIFF
--- a/test/distributed/cases/tenant/privilege/protected_database.result
+++ b/test/distributed/cases/tenant/privilege/protected_database.result
@@ -89,9 +89,9 @@ create user protected_bvt_user identified by '111' default role protected_bvt_wr
 use protected_bvt_db;
 internal error: do not have privilege to execute the statement
 show tables;
-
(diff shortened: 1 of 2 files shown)


Ariznawlll:
Closing this draft PR because it was opened from an older protected-database BVT branch and includes broader historical test changes. The current 3.0-dev already has the protected database coverage without the stale #25079/#25080 BVT issue tags, so no PR is needed for just removing those tags.

Issue #23845, opened 2026-03-12
[Bug]: function CONV wrong result.


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

bb8cc0e15

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

[func_conv.sql](https://github.com/user-attachments/files/25931376/func_conv.sql)

[diff.log](https://github.com/user-attachments/files/25931272/diff.log)

### Steps to Reproduce

```Markdown
as above.
```

daviszhen:
func_conv.sql
```
-- test case for conv function (issue #22963)
-- conv(n, from_base, to_base) - converts numbers between different number bases

-- ============================================
-- basic conversion tests
-- ============================================
-- basic test cases for conv function (issue #22963)
-- quick verification test suite
drop database if exists conv_func;
create database conv_func;
use conv_func;
-- test 1: basic hex to decimal (from issue example)
select conv('a', 16, 10) as result1;
-- expected: 10

-- test 2: decimal to hex
select conv('255', 10, 16) as result2;
-- expected: ff

-- test 3: binary to decimal
select conv('1010', 2, 10) as result3;
-- expected: 10

-- test 4: decimal to binary
select conv('10', 10, 2) as result4;
-- expected: 1010

-- test 5: octal to decimal
select conv('12', 8, 10) as result5;
-- expected: 10

-- test 6: negative number
select conv('-1', 10, 16) as result6;
-- expected: ffffffffffffffff

-- test 7: null handling
select conv(null, 10, 16) as result7;
-- expected: null

-- test 8: invalid character
select conv('g', 16, 10) as result8;
-- expected: 0

-- test 9: invalid base
select conv('10', 1, 10) as result9;
-- expected: null

-- test 10: case insensitivity
select conv('abc', 16, 10) as result10;
-- expected: 2748

-- test 1: hexadecimal to decimal
select conv('a', 16, 10) as hex_to_dec;
-- expected: 10

select conv('ff', 16, 10) as hex_to_dec;
-- expected: 255

select conv('1a', 16, 10) as hex_to_dec;
-- expected: 26

-- test 2: decimal to hexadecimal
select conv('10', 10, 16) as dec_to_hex;
-- expected: a

select conv('255', 10, 16) as dec_to_hex;
-- expected: ff

[...]

PR #25174, opened 2026-06-26
fix conv

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23845

修改 conv  基数转换函数的兼容性问题


heni02 (approved):

XuPeng-SH (changes requested):
I re-checked this from correctness / compatibility / unhappy-path angles. I would not approve this head yet.

Blocker 1: invalid input behavior no longer matches the linked issue.

The linked #23845 repro explicitly expects invalid digits such as `conv('g', 16, 10)`, `conv('2', 2, 10)`, and `conv('8', 8, 10)` to return `0`. The old implementation returned `NULL`; this PR changes the behavior again to a query error by returning `moerr.NewInvalidInputNoCtxf` from `parseConvStrictString()` when a digit is invalid for the base.

That means the PR does not actually fix the unhappy path from the issue; it changes the BVT expectation to `MO returns error` instead. If the intended compatibility target changed, the issue/test expectation needs to be clarified first. Otherwise this path should return the compatible value instead of failing the statement.

Relevant paths:

- `pkg/sql/plan/function/func_binary.go:2684-2698`
- `pkg/sql/plan/function/func_binary_test.go:2938-2966`
- `test/distributed/cases/function/func_conv.sql:276-283`

Blocker 2: positive-base unsigned parsing is wrong for `+`-prefixed values above `math.MaxInt64`.

For `from_base > 0`, `CONV()` treats `N` as unsigned. But the new code sends every value with a leading `+` through `strconv.ParseInt()` and saturates any range error to `math.MaxUint64`:

```go
if strings.HasPrefix(s, "-") || strings.HasPrefix(s, "+") {
    val, err := strconv.ParseInt(s, base, 64)
    if err != nil {
        if range { return 0, math.MaxUint64, false, nil }
    }
}
```

So valid unsigned inputs like `+9223372036854775808` or `+18446744073709551614` would incorrectly become `18446744073709551615`. Suggested fix: for positive `from_base`, handle `+` by stripping it and using the unsigned parse path; only `-` needs signed/two's-complement handling. Please add regression tests around `+9223372036854775808`, `+18446744073709551614`, and `+FFFFFFFFFFFFFFFF`.

Checks I ran:

- `git diff --check origin/4.0-dev...HEAD`
- `git merge-tree --write-tree HEAD origin/4.0-dev`

I could not complete local Go UT in this environment because the build is missing native headers/deps (`usearch.h`, `xxhash.h`; `CGO_ENABLED=0` also fails due allocator symbols). CI is still in progress on GitHub at the time of this review.

daviszhen:
> I re-checked this from correctness / compatibility / unhappy-path angles. I would not approve this head yet.
> 
> Blocker 1: invalid input behavior no longer matches the linked issue.
> 
> The linked #23845 repro explicitly expects invalid digits such as `conv('g', 16, 10)`, `conv('2', 2, 10)`, and `conv('8', 8, 10)` to return `0`. The old implementation returned `NULL`; this PR changes the behavior again to a query error by returning `moerr.NewInvalidInputNoCtxf` from `parseConvStrictString()` when a digit is invalid for the base.
> 
> That means the PR does not actually fix the unhappy path from the issue; it changes the BVT expectation to `MO returns error` instead. If the intended compatibility target changed, the issue/test expectation needs to be clarified first. Otherwise this path should return the compatible value instead of failing the statement.
> 
> Relevant paths:
> 
> * `pkg/sql/plan/function/func_binary.go:2684-2698`
> * `pkg/sql/plan/function/func_binary_test.go:2938-2966`
> * `test/distributed/cases/function/func_conv.sql:276-283`
> 
> Blocker 2: positive-base unsigned parsing is wrong for `+`-prefixed values above `math.MaxInt64`.
> 
> For `from_base > 0`, `CONV()` treats `N` as unsigned. But the new code sends every value with a leading `+` through `strconv.ParseInt()` and saturates any range error to `math.MaxUint64`:
> 
> ```go
> if strings.HasPrefix(s, "-") || strings.HasPrefix(s, "+") {
>     val, err := strconv.ParseInt(s, base, 64)
>     if err != nil {
>         if range { return 0, math.MaxUint64, false, nil }
>     }
> }
> ```
> 
> So valid unsigned inputs like `+9223372036854775808` or `+18446744073709551614` would incorrectly become `18446744073709551615`. Suggested fix: for positive `from_base`, handle `+` by stripping it and using the unsigned parse path; only `-` needs signed/two's-complement handling. Please add regression tests around `+9223372036854775808`, `+18446744073709551614`, and `+FFFFFFFFFFFFFFFF`.
> 
> Checks I ran:
> 
> * `git diff --check origin/4.0-dev...HEAD`
> * `git merge-tree --write-tree HEAD origin/4.0-dev`
> 
> I could not complete local Go UT in this environment because the build is missing native headers/deps (`usearch.h`, `xxhash.h`; `CGO_ENABLED=0` also fails due allocator symbols). CI is still in progress on GitHub at the time of this review.

block1, matrixone一直不支持这种行为.
block2, 已经修改.

XuPeng-SH (changes requested):
I re-checked the current head (`373c5e0`) from the PR diff, the related PR/issue comments, local tests, CI, and MySQL 8.0 behavior.

The previous `+`-prefixed unsigned parsing blocker is fixed now; the new regression cases around `+9223372036854775808`, `+18446744073709551614`, and `+FFFFFFFFFFFFFFFF` match MySQL. CI is also green, and I locally ran:

- `go test ./pkg/sql/plan/function -run 'TestConv|TestParseConv' -count=1`
- `git diff --check origin/4.0-dev...pr-25174-review`
- non-destructive `git merge-tree` against current `origin/4.0-dev` (no conflict; PR is just behind)

I still would not approve this head because the remaining unhappy paths are now encoded with incompatible expectations.

Blocker 1: invalid digits / invalid suffix are changed to statement errors, but the linked issue and MySQL-compatible behavior are non-error results.

The issue #23845 attachment/comment expects invalid digits such as `conv('g', 16, 10)`, `conv('2', 2, 10)`, and `conv('8', 8, 10)` to return `0`. I also verified this directly on MySQL 8.0:

```sql
select conv('g',16,10), conv('2',2,10), conv('8',8,10), conv('10xyz',10,16), conv('+',10,16), conv('',10,16);
-- 0, 0, 0, A, 0, NULL
```

Current PR returns `moerr.NewInvalidInputNoCtxf` from `parseConvStrictString()` for these paths, and the result files now expect statement errors:

- `pkg/sql/plan/function/func_binary.go:2684-2698`
- `pkg/sql/plan/function/func_binary_test.go:2965-2993`
- `test/distributed/cases/function/func_conv.sql:261-283` and `:316-340`
- `test/distributed/cases/function/func_conv.result:177-196` and `:227-228`
- `test/distributed/cases/function/func_other_conv.result:49-52`

This is not just “MO never supported returning 0”; it changes the old MO behavior from `NULL` into a statement failure, and it does not fix the linked compatibility case. For SQL scalar functions over real data, turning one bad value into a whole-query error is a much larger compatibility blast radius.

Suggested fix: do not validate the entire string as strictly legal before parsing. Parse the longest valid digit prefix after optional sign, matching `CONV`/numeric conversion style. If there is no valid digit after trimming/sign handling, return numeric `0` for valid bases; keep empty string as `NULL` if we intentionally want to preserve the current MO behavior there. `conv('10xyz', 10, 16)` should produce `A`, not an error. Update unit/BVT expectations accordingly.

Blocker 2: positive `from_base` negative overflow is saturated incorrectly.

For positive `from_base`, the current `strings.HasPrefix(s, "-")` branch uses `strconv.ParseInt`; any range error returns `math.MaxUint64`:

- `pkg/sql/plan/function/func_binary.go:2715-2721`
- `pkg/sql/plan/function/func_binary_test.go:2910-2917`
- `test/distributed/cases/function/func_conv.sql:323-324`

That makes values such as `conv('-18446744073709551616', 10, 16)` become `FFFFFFFFFFFFFFFF`. MySQL 8.0 does not do that:

```sql
select conv('-9223372036854775808',10,16),
       conv('-9223372036854775809',10,16),
       conv('-18446744073709551615',10,16),
       conv('-18446744073709551616',10,16);
-- 8000000000000000, 7FFFFFFFFFFFFFFF, 1, 0
```

Suggested fix: for positive `from_base` with a negative sign, parse the magnitude through the unsigned path rather than signed `ParseInt`, then apply two's-complement modulo semantics (`uint64(0) - magnitude`) for magnitudes representable in `uint64`. Add regression tests for `-9223372036854775809`, `-18446744073709551615`, and `-18446744073709551616`. Keep the current signed saturation behavior only for negative `from_base`, where MySQL-compatible signed semantics are different.

daviszhen:
> I re-checked the current head (`373c5e0`) from the PR diff, the related PR/issue comments, local tests, CI, and MySQL 8.0 behavior.
> 
> The previous `+`-prefixed unsigned parsing blocker is fixed now; the new regression cases around `+9223372036854775808`, `+18446744073709551614`, and `+FFFFFFFFFFFFFFFF` match MySQL. CI is also green, and I locally ran:
> 
> * `go test ./pkg/sql/plan/function -run 'TestConv|TestParseConv' -count=1`
> * `git diff --check origin/4.0-dev...pr-25174-review`
> * non-destructive `git merge-tree` against current `origin/4.0-dev` (no conflict; PR is just behind)
> 
> I still would not approve this head because the remaining unhappy paths are now encoded with incompatible expectations.
> 
> Blocker 1: invalid digits / invalid suffix are changed to statement errors, but the linked issue and MySQL-compatible behavior are non-error results.
> 
> The issue #23845 attachment/comment expects invalid digits such as `conv('g', 16, 10)`, `conv('2', 2, 10)`, and `conv('8', 8, 10)` to return `0`. I also verified this directly on MySQL 8.0:
> 
> ```sql
> select conv('g',16,10), conv('2',2,10), conv('8',8,10), conv('10xyz',10,16), conv('+',10,16), conv('',10,16);
> -- 0, 0, 0, A, 0, NULL
> ```
> 
> Current PR returns `moerr.NewInvalidInputNoCtxf` from `parseConvStrictString()` for these paths, and the result files now expect statement errors:
> 
> * `pkg/sql/plan/function/func_binary.go:2684-2698`
> * `pkg/sql/plan/function/func_binary_test.go:2965-2993`
> * `test/distributed/cases/function/func_conv.sql:261-283` and `:316-340`
> * `test/distributed/cases/function/func_conv.result:177-196` and `:227-228`
> * `test/distributed/cases/function/func_other_conv.result:49-52`
> 
> This is not just “MO never supported returning 0”; it changes the old MO behavior from `NULL` into a statement failure, and it does not fix the linked compatibility case. For SQL scalar functions over real data, turning one bad value into a whole-query error is a much larger compatibility blast radius.
> 
> Suggested fix: do not validate the entire string as strictly legal before parsing. Parse the longest valid digit prefix after optional sign, matching `CONV`/numeric conversion style. If there is no valid digit after trimming/sign handling, return numeric `0` for valid bases; keep empty string as `NULL` if we intentionally want to preserve the current MO behavior there. `conv('10xyz', 10, 16)` should produce `A`, not an error. Update unit/BVT expectations accordingly.
> 
> Blocker 2: positive `from_base` negative overflow is saturated incorrectly.
> 
> For positive `from_base`, the current `strings.HasPrefix(s, "-")` branch uses `strconv.ParseInt`; any range error returns `math.MaxUint64`:
> 
> * `pkg/sql/plan/function/func_binary.go:2715-2721`
> * `pkg/sql/plan/function/func_binary_test.go:2910-2917`
> * `test/distributed/cases/function/func_conv.sql:323-324`
> 
> That makes values such as `conv('-18446744073709551616', 10, 16)` become `FFFFFFFFFFFFFFFF`. MySQL 8.0 does not do that:
> 
> ```sql
> select conv('-9223372036854775808',10,16),
>        conv('-9223372036854775809',10,16),
>        conv('-18446744073709551615',10,16),
>        conv('-18446744073709551616',10,16);
> -- 8000000000000000, 7FFFFFFFFFFFFFFF, 1, 0
> ```
> 
> Suggested fix: for positive `from_base` with a negative sign, parse the magnitude through the unsigned path rather than signed `ParseInt`, then apply two's-complement modulo semantics (`uint64(0) - magnitude`) for magnitudes representable in `uint64`. Add regression tests for `-9223372036854775809`, `-18446744073709551615`, and `-18446744073709551616`. Keep the current signed saturation behavior only for negative `from_base`, where MySQL-compatible signed semantics are different.

block1:  invalid digits / invalid suffix  报错
block2:  已经修改

XuPeng-SH (changes requested):

`CONV()` now aborts on malformed string input instead of following the non-throwing conversion behavior users expect from MySQL-compatible `CONV`.

With the current implementation in `pkg/sql/plan/function/func_binary.go`, invalid or partially-invalid inputs such as:
- `CONV('g', 16, 10)`
- `CONV('2', 2, 10)`
- `CONV('8', 8, 10)`
- `CONV('10xyz', 10, 16)`
- `CONV('+', 10, 16)`

now raise a hard error because `convString()` bubbles up the error from `parseConvStrictString()`.

That is a compatibility regression. `CONV()` should consume the maximal valid prefix and return the converted value, or return `0` when no valid prefix exists, rather than aborting the whole statement on a malformed input string.

Suggested fix:
1. parse the maximal valid prefix instead of rejecting the whole string;
2. if no valid digits remain after trimming/sign handling, return `0`;
3. do not surface `InvalidInput` errors for per-row malformed `CONV()` inputs.

The new unit/BVT expectations currently lock in the stricter erroring behavior, so those should be adjusted together with the implementation.

daviszhen:
> I found one substantive blocker here.
> 
> `CONV()` now aborts on malformed string input instead of following the non-throwing conversion behavior users expect from MySQL-compatible `CONV`.
> 
> With the current implementation in `pkg/sql/plan/function/func_binary.go`, invalid or partially-invalid inputs such as:
> 
> * `CONV('g', 16, 10)`
> * `CONV('2', 2, 10)`
> * `CONV('8', 8, 10)`
> * `CONV('10xyz', 10, 16)`
> * `CONV('+', 10, 16)`
> 
> now raise a hard error because `convString()` bubbles up the error from `parseConvStrictString()`.
> 
> That is a compatibility regression. `CONV()` should consume the maximal valid prefix and return the converted value, or return `0` when no valid prefix exists, rather than aborting the whole statement on a malformed input string.
> 
> Suggested fix:
> 
> 1. parse the maximal valid prefix instead of rejecting the whole string;
> 2. if no valid digits remain after trimming/sign handling, return `0`;
> 3. do not surface `InvalidInput` errors for per-row malformed `CONV()` inputs.
> 
> The new unit/BVT expectations currently lock in the stricter erroring behavior, so those should be adjusted together with the implementation.

mo一直是报错.

XuPeng-SH (approved):

ouyuanning (approved):

Issue #24813, opened 2026-06-03
[Compatibility] rand() rejects an integer seed argument (RAND(N) not supported)


dengn:
## Summary
`rand()` rejects an integer seed, while MySQL supports `RAND(N)` for reproducible sequences.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test case
```sql
SELECT rand(123);
```

## Actual result
```
ERROR 20203 (HY000): invalid argument function rand, bad value [BIGINT]
```

## Expected
`RAND(N)` with an integer seed should be accepted (MySQL compatibility).

## Version status
- v3.0.11 (Cloud): reproduces
- main / 4.0-dev (commit 0bfda60): reproduces

daviszhen:
rand(seed) 分布式场景的行为 待确认. 先暂停pr

PR #25176, opened 2026-06-26
fix rand(n)

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24813

增加 rand 带 seed

### DIFF
--- a/pkg/sql/plan/function/func_builtin.go
+++ b/pkg/sql/plan/function/func_builtin.go
@@ -3639,13 +3639,34 @@ func builtInLog10(parameters []*vector.Vector, result vector.FunctionResultWrapp
 }
 
 type opBuiltInRand struct {
-	seed *rand.Rand
+	seed *mysqlRandState
 }
 
 var _ = newOpBuiltInRand().builtInRand
 
 func newOpBuiltInRand() *opBuiltInRand {
 	return new(opBuiltInRand)
+}
+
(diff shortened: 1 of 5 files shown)


heni02 (approved):

XuPeng-SH (approved):
Reviewed current head `a9853a93d7b08f4e90798556c9eaad25ef2cfa71` from compatibility and unhappy-path angles.

This adds the seeded `RAND(N)` overload with MySQL-compatible state initialization and per-invocation behavior: constant seed produces a repeatable sequence, non-constant seed is reinitialized per row/invocation, and `RAND(NULL)` follows the MySQL/MariaDB-compatible zero-seed behavior covered by the new tests.

I checked the stateful operator lifecycle (`newOpWithFree`, reset/close), const vs non-const vector paths, overload selection/casts, and the BVT/UT coverage. PR CI is green and `git diff --check` / merge-tree are clean.

Local targeted `go test ./pkg/sql/plan/function -run TestBuiltInRandWithSeed -count=1` cannot build in my environment because native headers (`usearch.h`, `xxhash.h`) are missing; `CGO_ENABLED=0` also fails due project cgo allocator requirements. Relying on code review plus CI for execution coverage.

Approved.

ouyuanning (approved):

Issue #25074, opened 2026-06-22
[Bug] sysbench1000w update_pk test crashed: 1000 threads got 'cannot connect to backend: EOF' due to memory pressure and GC pause


Ariznawlll:
## Summary

During the nightly sysbench1000w test on branch `4.0-dev` (commit `061520c88`), the `sysbench_update_pk_10_10000000` test with 1000 threads crashed at 195s with 93 FATAL errors: `error 20505 (cannot connect to backend: EOF)`.

Root cause: total cluster memory peaked at **177.6GB** under the update_pk workload, triggering DN OOM alerts (`available-mem=0`) and Go GC stop-the-world pauses that made CN unresponsive, causing all sysbench connections to fail.

## Steps to Reproduce

1. Deploy MO with 3 CN + 1 DN + 3 Log + 2 Proxy
2. Prepare sysbench 10-tables × 10M rows
3. Run `sysbench oltp_update_index.lua --threads=1000 --time=300`

## Expected Behavior

The test should complete 300 seconds without connection failures.

### Timeline

| Time (UTC) | Event |
|---|---|
| 20:01:38 | `oltp_update_index.lua` started, 1000 threads |
| 20:01:49 - 20:04:39 | TPS stable at ~5600-7100, latency ~200-320ms |
| 20:03:23 - 20:03:54 | DN logged 45× `MergeExecutorEvent-PauseDueToOOMAlert`, `available-mem=0` |
| 20:04:00 | **Total memory peaked at 177.6GB** (CN-qq8wx=52.4G, CN-qwdgt=45.6G, CN-vwbbf=44.3G, DN=33.2G) |
| 20:04:49 (190s) | TPS dropped from 6194 to **3439** (-45%), latency spiked to 539ms |
| 20:04:55 (196s) | **93 FATAL errors**: `mysql_stmt_execute() returned error 20505 (cannot connect to backend: EOF)` |
| 20:04:56 | Test exited with fatal error |
| 20:05:00 | Memory dropped to 157.8GB (GC reclaimed ~20GB) |

### Memory Trend (all pods total)

```
20:03:00  167.6G  (climbing)
20:03:30  172.0G
20:04:00  177.6G  ← PEAK
20:04:30  168.5G  (GC starting)
20:05:00  157.8G  ← crash happened, GC reclaimed memory
```

### DN OOM Alerts (20:03:23 - 20:03:54)

```json
[...]

PR #25178, opened 2026-06-26
metric: add CI timeout observability

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [x] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25074
issue #25164

This PR adds observability for CI/sysbench failures where clients see proxy all-CN-busy errors or backend/RPC timeout errors.

In simpler terms, it adds gauges/counters/dashboards that answer:
- Are CN/DN pods already near memory or CPU limits?
- Did the DN merge scheduler pause because available memory was too low?
- Did proxy mark specific CNs unhealthy, and did all candidate CNs become busy?
- Are proxy-to-CN handshakes slow or piling up?
- Did MORPC/lockservice fail on connect/read/write/wait-response, and what error type was it?

Changes:
- Add DN merge OOM governor metrics.
- Add proxy CN health, backend handshake latency, and in-flight handshake metrics.
- Add MORPC backend error and lockservice remote RPC error metrics with low-cardinality labels.
- Add Grafana panels for pod resources, merge OOM governor, proxy health/handshake, and RPC timeout/backend errors.
- Add an observability runbook with PromQL/LogQL queries for CI regression investigations.
- Add focused tests for MORPC and lockservice error type classification.

This PR does not claim to fix the sysbench overload root cause. It makes the failure chain observable so future 20508/20505 failures can be diagnosed from metrics instead of only from logs.

Validation:
- `git diff --check`
- `go test ./pkg/proxy`
- `go test ./pkg/vm/engine/tae/db/merge`
- `go test ./pkg/common/morpc`
- `go test ./pkg/common/morpc ./pkg/lockservice -run 'TestRPCMetricErrorType|TestLockserviceRemoteRPCErrorType'`
- `go test ./pkg/util/metric/v2 ./pkg/util/metric/v2/dashboard -run 'TestDoesNotExist'`


fengttt (approved):

aptend (approved):

XuPeng-SH (changes requested):
Superseding my earlier approve: I found two blockers.

1. `pkg/common/morpc/backend.go:999-1006` + `pkg/common/morpc/metrics.go:86-123`
   - The new MORPC metric is recorded on the raw dial/connect error *before* that error is wrapped as `moerr.ErrBackendCannotConnect`. For common hard failures like ECONNREFUSED / DNS / no-route, `rpcMetricErrorType()` therefore falls through to `other` instead of `backend_cannot_connect`, which defeats the stated goal of distinguishing backend-connect failures.
   - Suggested fix: classify after wrapping, or map non-timeout dial/net errors to `backend_cannot_connect` inside `rpcMetricErrorType()`.

2. `pkg/proxy/cn_health.go:255,292,307,409`, `pkg/proxy/router.go:391-397`, `pkg/util/metric/v2/proxy.go:118-141`
   - The new proxy metrics key counters/histograms/gauges by raw `cn_uuid`, but there is no corresponding metric-label delete path when CNs churn out of the cluster. `sweepStale()` only cleans the in-memory breaker map; it does not reclaim metric series. That makes series cardinality grow with cumulative CN churn, not current cluster size.
   - Suggested fix: avoid raw `cn_uuid` as a metric label here (aggregate by event/result only, or use a bounded dimension), or add an explicit label-lifecycle delete path when CNs leave.

VioletQwQ-0:
Addressed the change request in e814fdffe.

Changes:
- MORPC connect errors are now recorded after classification: timeout dial retries are still counted as `timeout`, while non-retryable connect failures are wrapped as `ErrBackendCannotConnect` before recording, so they classify as `backend_cannot_connect` instead of `other`.
- Removed raw `cn_uuid` from the new proxy Prometheus metrics to avoid cumulative CN churn cardinality:
  - `mo_proxy_cn_health_total{event}`
  - `mo_proxy_backend_handshake_duration_seconds{result}`
  - `mo_proxy_backend_handshake_inflight`
- Updated proxy dashboard queries and the observability runbook to match the reduced labels.
- Added a MORPC classification test for `ErrBackendCannotConnect` wrapping a net error.

Validation:
- `git diff --check`
- `go test ./pkg/common/morpc ./pkg/lockservice -run 'TestRPCMetricErrorType|TestLockserviceRemoteRPCErrorType'`
- `go test ./pkg/common/morpc`
- `go test ./pkg/proxy`
- `go test ./pkg/util/metric/v2 ./pkg/util/metric/v2/dashboard -run 'TestDoesNotExist'`

XuPeng-SH (approved):
I did not find a remaining blocker on the current head.

The latest review-fix commit addresses the two substantive issues from the previous round: MORPC hard connect failures are now classified after wrapping as `backend_cannot_connect`, and the new proxy observability metrics no longer key on raw `cn_uuid`, so the cumulative-CN-churn cardinality leak is gone. The targeted morpc/lockservice tests still pass, and an independent rereview of the current instrumentation found no correctness blockers.

One non-blocking follow-up I would still consider later is extending `TestLockserviceRemoteRPCErrorType` with a few remaining classification branches (for example `os.ErrDeadlineExceeded`, `net.Error` timeout, `io.EOF`, and `"unexpected EOF"` string cases), but I would not block this PR on that.

gouhongshen (approved):

PR #25181, opened 2026-06-26
fix(hnsw): vendor patched usearch 2.25.3 to fix #735 orphaned-node race on 4.0-dev

### LINKED ISSUE #735: Data lost after restart server

cpegeric:
4.0-dev builds the HNSW index with multi-threaded usearch add(), which can hit USearch #735: a concurrent add() stores the vector (contains()==true) but never links it into the graph, so search() can't reach it -> flaky recall@1.

Unlike main, 4.0-dev never took the single-thread-build workaround (#24867/#24993), so no Go change is needed here. The fix is purely the C library: upgrade the vendored usearch from 2.23.0 to 2.25.3 patched with USearch PR #772 (the #735 fix). The forked Go binding (cpegeric/usearch/golang, pinned in go.mod) links against the new C lib unchanged.

- thirdparties: replace usearch-2.23.0.tar.gz with the patched usearch-2.25.3.tar.gz; bump USEARCH_TAR/USEARCH_DIR; create dep subdirs before copy.
- BVT vector_hnsw_async: bump the t2 CDC-sync wait sleep(20) -> sleep(30).

Verified locally: usearch rebuilt from the patched tarball, mo links cleanly, and the t2 recall@1 scenario (async index, parallel-load 10k SIFT, multi-threaded build) passed 20/20 with no orphaned nodes.

## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue https://github.com/matrixorigin/matrixone/issues/24977

port usearch build fix to 4.0-dev


XuPeng-SH (approved):
No blocking issues found after a deeper check.

What I verified:
- The vendored `usearch-2.25.3.tar.gz` differs from upstream `v2.25.3` only in `include/usearch/index.hpp` and `cpp/test.cpp`, which matches USearch PR #772 for issue #735.
- I reproduced the old concurrent `add()` orphan-node bug locally with the 4.0-dev tarball (`usearch-2.23.0`): 8/8 runs had misses (roughly 27-53 unreachable nodes per run on this machine).
- Running the same standalone repro against the new vendored tarball (`usearch-2.25.3` + #772 patch) gave 0 misses in 8/8 runs.
- `make init usearch` under `thirdparties/` rebuilt and linked successfully on macOS/arm64.

Residual non-blocking risk: `vector_hnsw_async` still relies on fixed sleeps, so the test is still time-based rather than state-based. But I don't see a correctness issue in the actual fix path here.

heni02 (approved):

PR #25182, opened 2026-06-26
fix(frontend): require DELETE privilege for REPLACE conflicts (cherry-pick #25071)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24952

Cherry-picks #25071 to 4.0-dev.

This requires DELETE privilege for REPLACE paths that may delete conflicting rows, while keeping insert-only nullable unique-key cases from requiring DELETE.


heni02 (approved):

XuPeng-SH (approved):
Looks good to me.

I did not find a substantive correctness blocker in this cherry-pick.

The only non-blocking gap I’d call out is test strength around the NULL-containing composite unique-key REPLACE path: the privilege change is covered well, but the main semantic planner fix would still benefit from one end-to-end regression showing that a composite unique key with any NULL part does not trigger delete/replace behavior.

aunjgr (approved):

PR #25183, opened 2026-06-26
fix: cross-CN shuffle join INSERT...SELECT hang on multi-CN cluster (#24919) (cherry-pick #25158)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24919

Cherry-pick #25158 to 4.0-dev.

This fixes a silent deadlock where INSERT ... SELECT with cross-CN shuffle hash join over large tables can hang forever on multi-CN clusters.

Source PR: https://github.com/matrixorigin/matrixone/pull/25158
Source merge commit: 70667b3e669afb677ed6f94d4d2341fac259516a

The cherry-pick keeps the source PR scope: compile-side per-CN regrouping for cross-CN shuffle buckets plus focused remoterun tests.


XuPeng-SH (approved):
Re-checked this cherry-pick on 4.0-dev from the pipeline shape, affected-rows path, RemoteRun standalone check, and the main unhappy paths.

What I verified:
- the non-S3 insert path groups before attaching Insert, so the grouped per-CN container keeps Insert on the RootOp chain and affectedRows() still sees it;
- the write-S3 insert and multi-update paths still keep their write/flush aggregation operators in the expected place after grouping;
- groupShuffleBucketsByCNIfNeeded only triggers on the intended multi-CN + cross-CN-dispatch shape and fixes the standalone-executable check for the grouped send unit;
- current-cN grouped containers still run locally through ipAddrMatch, remote-CN grouped containers remain RemoteRun.

Validation:
- local targeted test pass: go test ./pkg/sql/compile -run 'TestGroupShuffleBucketsByCNIfNeeded|Test_checkPipelineStandaloneExecutableAtRemote' -count=1
- PR CI is green on 4.0-dev.

Residual note, not blocking this PR: the more generic cross-CN dispatch empty-input/end-handshake path is still worth tracking separately; this change fixes the concrete #24919 execution-tree shape by keeping the shuffle dispatch in the correct remote send unit.

PR #25185, opened 2026-06-26
feat(remap): session-level table rewrites + remapdb database remapping

fengttt:
## What type of PR is this?

- [x] Feature

issue #25187 #25188

Extends the existing table-rewrite (remap) feature with a session-level driver
and database-name remapping, all carried in the same JSON payload (the
`remap_rewrites` session variable and the `/*+ {...} */` hint).

### 1. `remap_rewrites` session variable
A session variable holding the rewrite config so it applies to every following
query, not just one statement:
```sql
set enable_remap_hint = 1;
set remap_rewrites = '{"rewrites": {"db1.t1": "select a, b from db2.t1"}}';
select * from db1.t1;   -- runs the mapped query
```
Validated at SET time (SELECT-only rules, valid identifiers) so a bad value is
rejected immediately and can never brick the session (rewriteSQL runs on every
statement, including SET).

### 2. Layered table rewrites (composition, not override)
Rewrites from `role rule -> session variable -> inline hint` are applied as
**stacked views**: each layer is a view over the previous one, so a
session/inline rule sees the role-rewritten relation and can only narrow it
further (good for access control). The inline hint is effective for one query
only. The wire format carries a per-table chain (`map[string][]*Rewrite`); the
planner peels one layer per nested reference.

### 3. `remapdb` database-name remapping
A `remapdb` entry maps one database to another, applied **before** the table
rewrites:
```sql
set remap_rewrites = '{"remapdb": {"dbx": "dby"}}';
select * from dbx.t;     -- reads dby.t
use dbx;                 -- switches to dby
insert into dbx.u select * from dbx.t;   -- both -> dby
```
- Works for SELECT and INSERT/UPDATE/DELETE, and for `USE`.
- Implemented as an AST substitution of qualified table references right after
  parsing (the database is validated during privilege checking, before the
  planner). Only qualified refs are rewritten — an unqualified name may be a
  CTE/alias, so the current-db case is handled by remapping `USE`.
- Remaps any number of databases; source and destination database sets must be
  disjoint (chaining `{"x":"y","y":"z"}` and self-maps are rejected at SET and
  parse time).

## Tests

- Unit: `pkg/frontend` (rewrite_rule, remap_db) and `pkg/sql/parsers`
  (AddRewriteHints) cover payload parsing, chain format, validation,
  disjointness, the AST walk (joins/subqueries/unions/CTE-safety), and DML.
- BVT: `hint/remap_rewrites_session`, `hint/remap_db`,
  `zz_accesscontrol/remap_rewrites_precedence`. Regression: hint dir and
  `zz_accesscontrol/role_rule` pass.

## Known limitation (see PR comment)

Table **rewrite rules** are not yet applied to the read side of
INSERT/UPDATE/DELETE (remapdb is). Details and root cause in a follow-up comment.


fengttt:
## Not yet done: table rewrite rules for INSERT / UPDATE / DELETE read sources

### Scope of the gap
- ✅ **remapdb** (database-name remapping) works for INSERT/UPDATE/DELETE — verified.
- ❌ **Table rewrite rules** (the `rewrites` chains) are currently applied only to SELECT. They are *not* applied to the read side of a DML statement.

The intended behavior (not yet implemented): a rewrite rule should apply to a
table that is **read** by a DML statement (e.g. the source of `INSERT ... SELECT`,
or a subquery in an `UPDATE`/`DELETE` `WHERE`), but **never** to the table being
modified (the target). For example:

```sql
-- rule: s -> "select * from s where id <= 3"
insert into tgt select * from s;          -- tgt should get only the rewritten rows of s
delete from t where id in (select id from s);  -- s rewritten; t (target) not
update t set ... where id in (select id from s); -- s rewritten; t (target) not
delete from t;                            -- if t has a rule, it is IGNORED (t is the target)
```

### What I tried
Following the SELECT mechanism (`ctx.remapOption` consumed in `buildTable`), I:
1. added a `RewriteOption *RewriteOption` field to the `tree.Insert/Update/Delete` AST nodes;
2. attached it in `AddRewriteHints` for those statement types;
3. in the DML build paths, set the read bind context's `remapOption` to the
   statement's rewrite chains **minus the modification-target keys**
   (so the target is never rewritten), at what looked like the right injection
   points: `buildDelete`/`buildUpdate` (before `deleteToSelect`/`selectUpdateTables`)
   and the insert paths (`bindInsert` and `initInsertStmt`, before the source
   `bindSelect`).

### Why it fails
Debug tracing showed that for `INSERT ... SELECT`, **none** of the functions I
instrumented were reached — the rows inserted unfiltered and zero debug output
fired. So the active DML planning route binds the read source somewhere other
than where I injected. Two factors are involved:

- **Multiple INSERT paths**: dispatch goes `*tree.Insert -> bindAndOptimizeInsertQuery -> builder.bindInsert`, with a fallback to `buildInsert -> initInsertStmt` only on `ErrUnsupportedDML`. Injecting in both still produced no effect, which points to the read-source binding happening at a different layer (or a plan-cache serving the plan, so the build path isn't re-run at all).
- The same applies to UPDATE/DELETE, where the read sources are subqueries whose
  bind contexts must inherit the (target-excluded) `remapOption`.

Rather than land a change that compiles but silently does nothing, this was
reverted. The remaining commits in this PR are unaffected.

### Proposed follow-up
Do this as its own focused change:
1. Trace the *actual* read-source binding point for each DML kind (and confirm
   plan-cache behavior) with targeted instrumentation.
2. Wire the rewrite option in at the confirmed point, with the modification
   target excluded (`db.table` keys from `tblInfo`/`dmlCtx`).
3. Test each of insert / update / delete, including the "target rule is ignored"
   case and a read-source rule that does apply.

fengttt:
Addressed the review comments in `51d72f1c6`:

1. **`remap_rewrites` was `ScopeBoth` (global sets bypass validation)** — changed to **`ScopeSession`**. `SET GLOBAL remap_rewrites` is now rejected, so an unvalidated global value can't slip in. BVT case added.

2. **Comment said session "overrides" role, but impl composes** — corrected the doc comment in `rewriteSQL` to describe the role→session→inline stacked-view layering.

3 & 4. **`extractInlineRewrites` / executor brittle to chained (array-form) rewrites, dropping `remapdb`** — the executor now reads `remapdb` via a dedicated **`extractInlineRemapDb`** that decodes *only* the `remapdb` field, so layered (array) rewrites in the merged hint can't interfere. Per maintainer direction, a rewrite **value must be a single SQL string** — `parseSessionRewrites` now **raises a clear error on an array or object value** rather than tolerating it (the layered-chain array form stays internal to `rewriteSQL` and the parser). BVT cases added for both array and object rejection.

5. **`applyRemapDb` didn't walk expression sub-selects** — it now traverses `WHERE`/`HAVING`, join `ON`, projections, `GROUP BY`, `UPDATE` SET expressions and the `WHERE` of `UPDATE`/`DELETE`. So a qualified reference inside `WHERE id IN (SELECT ... FROM dbx.t)`, `EXISTS (...)`, a scalar subquery, etc. is now remapped. Unit + BVT coverage added (IN/EXISTS/HAVING/scalar/join-ON and a DELETE-subquery source).

Verification: `pkg/frontend` unit tests pass; BVT `hint` dir 356/356; `zz_accesscontrol/remap_rewrites_precedence` 100%.

XuPeng-SH (changes requested):
I found three blocking issues in the new remap path.

1. `remap_rewrites` SET-time validation is still weaker than query-time parsing, so it can store values that later fail every rewritten query. In `validateRemapRewrites` / `parseSessionRewrites` / `stringRewrites` ([pkg/frontend/rewrite_rule.go]), the only key check is non-empty + SELECT-like SQL. But query-time `AddRewriteHints` still requires every rewrite key to be fully qualified `db.table` ([pkg/sql/parsers/sqlparse.go]). So `set remap_rewrites = {"t":"select 1"}` is accepted, and the session only fails later with `the mapping name needs to include database name` when the hint is re-parsed. That breaks the stated goal that invalid values are rejected at SET time and never brick the session.\n\nSuggested fix: factor the rewrite-key validation into a shared helper and use the same rules in both SET-time validation and `AddRewriteHints` (db-qualified, non-empty db/table).\n\n2. Inline array-form rewrite chains cannot be merged with role/session rewrites. The parser now explicitly supports array-form chains (`"db.t": ["sql1", "sql2"]`) in `AddRewriteHints` ([pkg/sql/parsers/sqlparse.go], also covered by `TestAddRewriteHints_ChainArray`). But `rewriteSQL` merges inline rules by calling `extractInlineRewrites`, which reuses `parseSessionRewrites`; that path rejects arrays on purpose (`stringRewrites` says rewrite values must be a single SQL string) ([pkg/frontend/rewrite_rule.go]). Result: the exact inline syntax the parser accepts becomes unusable as soon as a role rule or `remap_rewrites` session value exists and `rewriteSQL` has to merge hints.\n\nSuggested fix: parse inline hints for merge with the same `RewriteMap` / `decodeRewriteChain` logic the parser already uses, instead of reusing the session-variable parser. Session values can stay string-only, but inline hints need to keep accepting arrays.\n\n3. The inline `remapdb` path is inconsistent for non-SELECT statements: DML bypasses validation, and `USE` ignores inline remapdb entirely. `AddRewriteHints` validates `remapdb` only for `*tree.Select` / `*tree.ParenSelect` ([pkg/sql/parsers/sqlparse.go]). For `INSERT/UPDATE/DELETE`, the executor instead calls `applyRemapDb(stmts, extractInlineRemapDb(...))`, and `extractInlineRemapDb` only unmarshals JSON and returns the map without validating identifiers or overlap/chaining ([pkg/frontend/rewrite_rule.go], [pkg/frontend/remap_db.go]). So an inline DML hint like `/*+ {"remapdb":{"x":"y","y":"z"}} */ update x.t ...` bypasses the very chaining/self-map checks the PR adds elsewhere. Separately, `USE` only consults `sessionRemapDb` in `self_handle.go`, so inline `remapdb` on `USE` is never applied.\n\nSuggested fix: add a shared `parseInlineRemapDb`/`validateRemapDb` path for all statement kinds before AST remap, and make `USE` consume the same effective inline+session remap source as DML/SELECT.\n\nI would re-review after those are fixed.

fengttt:
Also fixes #25188 (`d9ffec43b`): a rewrite rule keyed on a bare table name (e.g. `"t"` instead of `"foo.t"`) silently did nothing.

- The **inline-hint** path already rejected it in the parser (`the mapping name needs to include database name`), and the **role-rule** grammar requires `db . tbl`.
- The gap was the **`remap_rewrites` session variable**, which only checked the key was non-empty. `validateRemapRewrites` now enforces (via `validateRewriteTableKey`) that every key is qualified as exactly `database.table`:

```
set remap_rewrites = '{"t": "select * from foo.t where i < 10"}';
-> remap_rewrites: rewrite table "t" must be qualified as database.table
```

Unqualified / over-qualified (`a.b.c`) / empty-part (`.t`) keys are all rejected at SET time. Unit + BVT (`hint/remap_errors`) coverage added; the issue's exact repro now errors on both the inline and session paths.

fengttt:
Thanks @XuPeng-SH — addressed in `843186036`:

**1. SET-time vs query-time key validation now share one helper.**
`parsers.SplitRewriteKey` is the single source of truth for the `db.table` qualification rule, used by both `AddRewriteHints` (query time) and `validateRemapRewrites` (SET time). So `set remap_rewrites = '{"t":"select 1"}'` is now rejected at SET time with `rewrite table "t" must be qualified as database.table`, not later at re-parse. (Also fixed unqualified keys at SET for #25188 in the prior commit.)

**2. Inline array-form chains can be merged again.**
`rewriteSQL`'s `extractInlineRewrites` now decodes the inline hint with the parser's own logic (new `parsers.DecodeRewriteHint`, which reuses `RewriteMap`/`decodeRewriteChain`) instead of the session-variable parser. So the array syntax the parser accepts (`"db.t":["s1","s2"]`) stays usable once a role/session rule forces hint merging. Session-variable values stay string-only. New BVT case in `hint/remap_rewrites_session` composes a session rule with an inline array chain.

**3. Inline remapdb validated for DML; USE consumes session+inline.**
`AddRewriteHints` now also decodes INSERT/UPDATE/DELETE inline hints, so `/*+ {"remapdb":{"x":"y","y":"z"}} */ update x.t ...` hits the same identifier/chaining validation as SELECT (verified: now rejected). USE merges the session-variable remapdb with any surviving inline hint and validates via `parsers.ValidateRemapDb`. `AddRewriteHints`, `DecodeRewriteHint` and `ValidateRemapDb` are factored so parser and frontend share one implementation.

**One honest limitation on 3 (USE + inline):** a standalone `/*+ {"remapdb":...} */ use db` is re-stringified during execution (USE is frontend-executed with a pre-set stmt, so it skips the executor's AST-remap block, and by the time `self_handle` runs, the leading hint is gone). The **session-variable** remapdb on USE works; **inline** remapdb on a bare USE may not apply. Applying it would mean remapping the USE target where the original text is still available (before re-stringification) — happy to do that as a focused follow-up if you'd like it in this PR.

Verification: `pkg/sql/parsers` + `pkg/frontend` unit tests pass; BVT `hint` dir 362/362; `zz_accesscontrol/remap_rewrites_precedence` 100%.

fengttt:
Reworked how remapdb handles the **current database** (`2506681b6`):

- **`USE` is no longer affected by remapdb** (role, session, or inline). `use <src>` lands in the real `<src>` and `database()` returns `<src>`.
- **Current-database remap at query time:** when the current database is a remap source, *unqualified* references (tables, views, and DML targets) resolve against the destination. Implemented in `TxnCompilerContext.DefaultDatabase()` returning the remapped db, so only name resolution is redirected — the session's actual current database and `database()` stay the source. The effective remapdb (role/session/inline merged) is stashed on `ExecCtx` in `GetComputationWrapper`.
- **Qualified references** are still remapped at the AST level by `applyRemapDb`.

This fixes the earlier gap where unqualified names only resolved in the destination as a side effect of `USE` being remapped — and where being genuinely *in* the source DB left unqualified names resolving in the source.

Verified: `use src` → `database()`=src; unqualified `t`/view/INSERT/UPDATE/DELETE → dst; with remap off, the real src is visible; a non-source current db is untouched. BVT `hint` dir 371/371; `zz_accesscontrol/remap_rewrites_precedence` 100%.

fengttt:
Added object-level scope tests and closed a DDL gap (`911ef2244`).

While writing the tests I found that **qualified DDL** was not remapped: `create table src.t` with remapdb `src->dst` was creating the table in `src` (applyRemapDb only walked SELECT/DML). Fixed — `applyRemapDb` now also remaps the database of qualified references in `CREATE/ALTER/DROP TABLE`, `CREATE/DROP VIEW`, `CREATE/DROP INDEX` (and `... AS SELECT` bodies), matching the unqualified/current-db case.

New BVT `hint/remap_db_objects` verifies, with `remapdb {dsrc: ddst}` active:

**Remapped (table-level)** — both via a qualified `dsrc.x` ref and via the current database when it is `dsrc` (unqualified), while the real `dsrc` is untouched:
- `SELECT` / `INSERT` / `UPDATE` / `DELETE`
- `CREATE` / `ALTER` / `DROP` of `TABLE` / `VIEW` / `INDEX`

**Not remapped (database-level):**
- `USE` (`database()` stays `dsrc`)
- `CREATE DATABASE`, `ALTER DATABASE`, `DROP DATABASE` — e.g. `drop database dsrc` while in `dsrc` drops the real `dsrc` and leaves `ddst` intact.

Unit tests cover the DDL AST remapping and that `CREATE/DROP DATABASE` are left alone. hint dir 439/439.

fengttt:
Added a system-database guard for remapdb (`38eb714d4`).

remapdb may no longer use a **system database** as a source or destination — a remap of (or into) `information_schema`, `mysql`, `system`, `system_metrics`, or any `mo_*` database is rejected:

```
set remap_rewrites = '{"remapdb": {"mysql": "x"}}';
-> remapdb must not remap a system database, got "mysql" -> "x"
/*+ {"remapdb": {"mo_catalog": "x"}} */ select 1;
-> remapdb must not remap a system database, got "mo_catalog" -> "x"
```

New `parsers.IsSystemDatabase` (case-insensitive: the `mo_` prefix plus the named DBs) is checked in the shared `ValidateRemapDb` (parse time) and the frontend `validateRemapDb` (SET time), so both the inline hint and the session variable are covered, source or destination. Non-system names like `moose`/`mox` are unaffected. Unit tests + BVT `hint/remap_errors` cases added; hint dir 446/446.

heni02 (approved):

ouyuanning (approved):

iamlinjunhong (approved):

XuPeng-SH (changes requested):
I found three substantive blockers here.

1. Prepared statements do not honor later `remap_rewrites` / `enable_remap_hint` changes.

`rewriteSQL` is applied at `COM_QUERY` and `COM_STMT_PREPARE`, but not at `COM_STMT_EXECUTE`. `EXECUTE` reuses the cached prepared SQL/plan and only rebuilds on schema change. So if a session changes `remap_rewrites` or toggles `enable_remap_hint` after prepare time, the execute path still uses the stale remap state.

Suggested fix: re-evaluate rewrite/remap state at EXECUTE time, or invalidate/reprepare cached prepared statements when those session variables change. Please also add a PREPARE/EXECUTE regression that changes remap state between prepare and execute.

2. `remapdb` misses qualified table-level DDL such as `TRUNCATE TABLE` (and `RENAME TABLE`).

`applyRemapDb` handles several table DDLs, but not `*tree.TruncateTable` or `*tree.RenameTable`. For qualified names, planner resolution will still use the original schema directly, so statements like `truncate table src.t` or `rename table src.t to src2.u` are not consistently remapped the same way as the other table-level DDLs.

Suggested fix: extend AST remapping to the remaining table-level DDLs and add coverage for qualified `TRUNCATE TABLE` / `RENAME TABLE` under `remapdb`.

3. Injected rewrite hints currently break statements with trailing `; -- comment`.

This is not just a hypothetical parser edge case — the PR’s own `test/distributed/cases/hint/remap_db_objects.result` currently records `SQL parser error: parse hints bug` in the “current database” scenario after inline `--` comments. So once rewrites are injected, statements like `use dsrc; -- comment` are actually breaking.

Suggested fix: make hint extraction ignore comment-only semicolon fragments, or otherwise align the hint parser’s statement counting with the SQL parser before calling `AddRewriteHints`.

The happy-path coverage is otherwise fairly strong, but these are correctness/behavior issues rather than just test gaps.

aunjgr (approved):

fengttt:
Thanks @XuPeng-SH — addressed all three in `ac5805c9e`:

**1. Prepared statements now honor remap changes (invalidate on SET).**
`rewriteSQL` runs at `COM_QUERY`/`COM_STMT_PREPARE` but not `COM_STMT_EXECUTE`, so a prepared plan captured the remap state at PREPARE. Setting `remap_rewrites` or `enable_remap_hint` now invalidates every cached prepared statement (`Session.RemoveAllPrepareStmts`, called from `SetSessionSysVar`), so a later EXECUTE can never run a stale rewrite — it re-prepares under the current state. Unit test `TestRemoveAllPrepareStmts` added.

One honest note on the regression you asked for: a mo-tester BVT can only drive **text-protocol** `PREPARE s FROM '...'`, and that path never rewrites the inner query in the first place — `AddRewriteHints`/`applyRemapDb` only touch `Select`/DML/DDL, not `PrepareString`; only binary `COM_STMT_PREPARE` unwraps and rewrites the inner SQL. So the invalidation is covered by the unit test rather than a BVT (a BVT would exercise the drop, not the rewrite). Happy to add a binary-protocol test harness if you'd prefer.

**2. `remapdb` now covers `TRUNCATE TABLE` and `RENAME TABLE`.**
`applyRemapDb` remaps qualified `TRUNCATE TABLE` and `RENAME TABLE` (both the source and the rename destination). Unit tests + BVT (`hint/remap_db_objects`) added. Note: `RENAME`'s privilege check independently rejects a qualified rename issued from a system database (`mysql`) even without remap, so the BVT runs it from a normal current db.

**3. Trailing `; -- comment` no longer breaks injected hints.**
`extractLeadingHints` now skips blank/comment-only semicolon fragments (`fragmentHasStatement`) so the hint list lines up with the parser's statement list, and `AddRewriteHints` accepts the all-comment case (0 hints vs a single `EmptyStmt`). The `SQL parser error: parse hints bug` that `remap_db_objects.result` had recorded is gone. Unit test `TestAddRewriteHints_TrailingLineComment` added.

Verification: `pkg/sql/parsers` + `pkg/frontend` unit tests pass; BVT `hint` dir 460/460; `zz_accesscontrol/remap_rewrites_precedence` 26/26 and `role_rule` 194/194.

PR #25194, opened 2026-06-29
cherrypick-to-4.0: add lock wait time

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/24420

1, 启用lock_wait_timeout session 级别变量. mysql也有. 用户可以设置此值. lock_wait_timeout 的默认值 `31536000 (1 year)`, 能兼容目前的行为. 如果用户设置了超时时间就用用户的值. 2, 增加lock rpc timeout . 如果lock_wait_timeout设置了, lock rpc timeout 就用 lock_wait_timeout值.

避免几个小时加锁傻等.

3, 修改lockoption 增加lockwaittimeout选项.

4, 给 RPC 加 slack（宽松预算）：RPC 超时 = lock_wait_timeout + 30s，让服务端有足够时间返回超时结果 5, 服务端异步路径也强制检查超时：之前异步路径（远程锁走这条）从不检查 LockWaitTimeout，现在 waiterEvents.check() 里会定期检查并通知超时 6, 客户端翻译 deadline 错误：如果 RPC deadline 到了但 caller context 还没到，说明是锁超时而非连接问题，直接翻译成 ErrLockTimeout

修改后的效果:
session 1 begin个事务, 加行锁.
session 2 delete 行, 等锁, 超时

```
session1

MySQL [test]> select * from t1;
+------+
| a    |
+------+
|    2 |
|    3 |
|    4 |
|    1 |
+------+
4 rows in set (0.001 sec)

MySQL [test]>
MySQL [test]>
MySQL [test]> begin;
Query OK, 0 rows affected (0.000 sec)

MySQL [test]> select * from t1 where a = 1 for update;
+------+
| a    |
+------+
|    1 |
+------+
1 row in set (0.001 sec)

MySQL [test]>

```

```
session2

MySQL [test]> select @@session.lock_wait_timeout;
+---------------------+
| @@lock_wait_timeout |
+---------------------+
| 180                 |
+---------------------+
1 row in set (0.000 sec)

MySQL [test]> delete from t1 where a = 1;
ERROR 1105 (HY000): context deadline exceeded
MySQL [test]>

```


XuPeng-SH (changes requested):
I found a few blockers here.

1. `LockWaitDeadline` is computed in `pkg/sql/colexec/lockop/lock_op.go`, but the lockservice never actually enforces that absolute deadline. `refreshLockWaitOptions()` converts the remaining budget back into whole seconds with `math.Ceil`, and the receiver reconstructs its own timer from `LockWaitTimeout` (`pkg/lockservice/lock_table_local.go`, `pkg/lockservice/waiter_events.go`) instead of using the original deadline. That means retries with `<1s` remaining can still wait for almost another full second, and remote waits start counting only after the RPC reaches the owner, excluding network/forwarding time. This breaks the stated goal of honoring the current statement/session timeout end-to-end.

   Suggested fix: plumb `LockWaitDeadline` through the lockservice and derive both sync and async wait budgets from the absolute deadline, using `LockWaitTimeout` only as a compatibility fallback.

2. The new RPC-timeout path can still create phantom remote locks on wrapped timeout errors. `pkg/lockservice/lock_table_remote.go` now adds an internal `rpcCtx` timeout, but `skipTrackLockOnError()` still only checks `ctx.Err()`, `errors.Is(err, context.DeadlineExceeded/Canceled)`, and `ErrRPCTimeout`. The same file already had to use `errors.As(..., net.Error)` in `retryRemoteLockError()` for wrapped timeout errors. If `Send()` returns a wrapped timeout while the parent `ctx` is still alive, this path will still call `txn.lockAdded(...)` and record a lock that may never have existed.

   Suggested fix: classify timeout errors in `skipTrackLockOnError()` the same way as `retryRemoteLockError()`, and preferably make the bookkeeping decision against `rpcCtx` rather than the original parent context.

3. The new BVT does not lock down the intended user-visible behavior. `test/distributed/cases/pessimistic_transaction/lock_wait_timeout_remote.sql` still accepts `context deadline exceeded` in addition to lock-timeout errors, so it would pass even if the remote path regressed back to timing out in the client/RPC layer instead of returning the intended lock-timeout error.

   Suggested fix: tighten the assertion to require the intended lock-timeout error, then add a regression that consumes most of the timeout budget before a retry/upgrade path to verify the original deadline is preserved.

Coverage is otherwise decent, but these are blockers for me.

aptend (approved):

daviszhen:
> I found a few blockers here.
> 
> 1. `LockWaitDeadline` is computed in `pkg/sql/colexec/lockop/lock_op.go`, but the lockservice never actually enforces that absolute deadline. `refreshLockWaitOptions()` converts the remaining budget back into whole seconds with `math.Ceil`, and the receiver reconstructs its own timer from `LockWaitTimeout` (`pkg/lockservice/lock_table_local.go`, `pkg/lockservice/waiter_events.go`) instead of using the original deadline. That means retries with `<1s` remaining can still wait for almost another full second, and remote waits start counting only after the RPC reaches the owner, excluding network/forwarding time. This breaks the stated goal of honoring the current statement/session timeout end-to-end.
>    Suggested fix: plumb `LockWaitDeadline` through the lockservice and derive both sync and async wait budgets from the absolute deadline, using `LockWaitTimeout` only as a compatibility fallback.
> 2. The new RPC-timeout path can still create phantom remote locks on wrapped timeout errors. `pkg/lockservice/lock_table_remote.go` now adds an internal `rpcCtx` timeout, but `skipTrackLockOnError()` still only checks `ctx.Err()`, `errors.Is(err, context.DeadlineExceeded/Canceled)`, and `ErrRPCTimeout`. The same file already had to use `errors.As(..., net.Error)` in `retryRemoteLockError()` for wrapped timeout errors. If `Send()` returns a wrapped timeout while the parent `ctx` is still alive, this path will still call `txn.lockAdded(...)` and record a lock that may never have existed.
>    Suggested fix: classify timeout errors in `skipTrackLockOnError()` the same way as `retryRemoteLockError()`, and preferably make the bookkeeping decision against `rpcCtx` rather than the original parent context.
> 3. The new BVT does not lock down the intended user-visible behavior. `test/distributed/cases/pessimistic_transaction/lock_wait_timeout_remote.sql` still accepts `context deadline exceeded` in addition to lock-timeout errors, so it would pass even if the remote path regressed back to timing out in the client/RPC layer instead of returning the intended lock-timeout error.
>    Suggested fix: tighten the assertion to require the intended lock-timeout error, then add a regression that consumes most of the timeout budget before a retry/upgrade path to verify the original deadline is preserved.
> 
> Coverage is otherwise decent, but these are blockers for me.

已经修改

heni02 (approved):

XuPeng-SH (commented):
Re-reviewing the current head: two of my earlier concerns are no longer blockers.

- The wrapped-timeout / phantom-lock concern looks fixed now (`rpcCtx` is used for bookkeeping decisions, and timeout-like / lock-timeout errors are classified more safely).
- The BVT concern is also narrower now: the SQL case no longer accepts `context deadline exceeded`, so that earlier objection does not apply to the current head.

I still see one blocker, though:

Remote `lock_wait_timeout` is still not end-to-end correct for longer waits. Lock RPCs still run over a morpc backend with a fixed 10s read timeout (`pkg/lockservice/rpc.go`), while `lockop` also keeps backend retry budget capped at 10s (`pkg/sql/colexec/lockop/lock_op.go`). So if the requested remote lock wait exceeds that transport window, the statement can still fail as a backend timeout / backend-cannot-connect path before the configured lock-wait deadline is reached.

The new tests cover 1s / sub-second cases and an absolute-deadline-overrides-long-timeout case using a short deadline, but they do not falsify the >10s remote wait path.

Suggested fix:
1. make the effective transport/read timeout for lock RPCs track the remaining lock-wait deadline;
2. make backend-retry budgeting honor that same remaining deadline instead of a fixed 10s cap;
3. add a remote test where `LockWaitTimeout` is greater than the current 10s backend read timeout.

daviszhen:
> Re-reviewing the current head: two of my earlier concerns are no longer blockers.
> 
> * The wrapped-timeout / phantom-lock concern looks fixed now (`rpcCtx` is used for bookkeeping decisions, and timeout-like / lock-timeout errors are classified more safely).
> * The BVT concern is also narrower now: the SQL case no longer accepts `context deadline exceeded`, so that earlier objection does not apply to the current head.
> 
> I still see one blocker, though:
> 
> Remote `lock_wait_timeout` is still not end-to-end correct for longer waits. Lock RPCs still run over a morpc backend with a fixed 10s read timeout (`pkg/lockservice/rpc.go`), while `lockop` also keeps backend retry budget capped at 10s (`pkg/sql/colexec/lockop/lock_op.go`). So if the requested remote lock wait exceeds that transport window, the statement can still fail as a backend timeout / backend-cannot-connect path before the configured lock-wait deadline is reached.
> 
> The new tests cover 1s / sub-second cases and an absolute-deadline-overrides-long-timeout case using a short deadline, but they do not falsify the >10s remote wait path.
> 
> Suggested fix:
> 
> 1. make the effective transport/read timeout for lock RPCs track the remaining lock-wait deadline;
> 2. make backend-retry budgeting honor that same remaining deadline instead of a fixed 10s cap;
> 3. add a remote test where `LockWaitTimeout` is greater than the current 10s backend read timeout.

1, 新加的 /mnt/fastdata/temp2/matrixone/pkg/sql/colexec/lockop/lock_op.go:795 到 /mnt/fastdata/temp2/matrixone/pkg/sql/colexec/lockop/
  lock_op.go:1015 这部分，只能说解决了 reviewer 的第 2 条建议：backend retry budget 现在会优先跟随 LockWaitDeadline。

2, 但第 1 条transport/read timeout 还没动，所以按 /home/pengzhen/Documents/daily_record/mo/24420.md 对照审查，PR 仍然还有 1 个未解 blocker。AI给出的修改方案:

```
• 可以，采用更稳的方案比较合适。                                                                                                                
                                                                                                                                                
  我建议把 transport/read timeout 的修改收敛成“仅对 lockservice 生效的 opt-in 行为”，不要先把 morpc 的默认 backend 语义整体改掉。
                                                                        
  具体做法：                                                            
                                                                                                                                                
  1. 在 morpc 里新增一个 opt-in backend option                                                                                                  
      - 例如 WithBackendDynamicReadTimeout() 这类能力                                                                                           
      - 默认关闭                                                        
      - 开启后，backend read loop 才会参考 outstanding request 的最早 deadline                                                                  
                                                                                                                                                
  2. 只在 lockservice client 上启用这个 option                                                                                                  
      - 也就是只在 /mnt/fastdata/temp2/matrixone/pkg/lockservice/rpc.go 里给 lockservice 的 morpc client 打开                                   
      - 其他 RPC 使用者保持原样，不被连带影响                                                                                                   
                                                                                                                                                
  3. dynamic read timeout 的策略先取“最早 outstanding future deadline”                                                                          
      - 这是共享 backend / 共享 read loop 下最保守的策略                                                                                        
      - 不会让某个快到期的请求被一个长请求拖住                                                                                                  
      - 也比“固定 10s”更接近 reviewer 要的 end-to-end deadline                                                                                  
                                                                        
  4. heartbeat 先不动语义，只确保它不会把长等待路径掩盖掉                                                                                       
      - 也就是说测试要直接证明：                                                                                                                
          - LockWaitTimeout > 10s                                                                                                               
          - 最终返回的是 lock timeout                                                                                                           
          - 不是 backend timeout / backend cannot connect

      - 如果 heartbeat 仍会干扰，再单独处理 heartbeat，而不是一开始就一起改

  5. lockop 保持你现在这条修复                                          
      - backend retry budget 优先跟 LockWaitDeadline
      - 这部分本来就应该保留                                            

  这样做的优点是：                                                      

  - reviewer 的 blocker 能正面回应                                      
  - 影响面局限在 lockservice                                            
  - morpc 不会被一次性重构成新的全局语义
  - 如果后面发现 dynamic read timeout 还有边界问题，回滚和隔离都容易

  我建议下一步按这个顺序做：                                            

  1. 恢复并保留 lockop 的 retry-budget 修复
  2. 在 morpc 实现 opt-in dynamic read timeout
  3. 仅在 lockservice/rpc.go 启用                                       
  4. 补 >10s remote regression test                                     
  5. 再整体复审 /home/pengzhen/Documents/daily_record/mo/24420.md 的 comment3

  如果你确认，我就按这个稳妥方案继续改代码。
```

XuPeng-SH (commented):
Re-reviewing again at the current head plus the new discussion: I still think this should stay in request-changes state, but the blocker set is now narrow and specific.

What is fixed now:
- the wrapped-timeout / phantom-lock concern looks resolved;
- the SQL BVT no longer accepts `context deadline exceeded`.

What still remains unresolved:
- remote waits longer than 10 seconds still run over a morpc backend with a fixed 10s read timeout.

`lockop` has improved since the earlier review: the backend retry budget now follows `LockWaitDeadline`, so that half of the problem is better. But the transport side still uses a static 10s backend read timeout in `pkg/lockservice/rpc.go`, and morpc enforces that independently of the per-request `rpcCtx`. So for long remote waits, the client can still see a backend/connectivity-style failure before the requested `lock_wait_timeout` expires, instead of receiving the intended lock-timeout result from the remote owner.

That means the old >10s remote-wait concern is still confirmed, just narrower than before.

Suggested fix:
1. make the lockservice morpc read timeout track the remaining lock-wait deadline (or otherwise stop using a fixed 10s transport timeout for lock RPCs);
2. keep the retry-budget fix already added;
3. add a regression where a remote lock wait is >10s and must end in lock-timeout rather than backend timeout / backend-cannot-connect.

matrix-meow:
## CI Failure: TestLockRetryBudgetUsesLockWaitDeadline

**Run**: https://github.com/matrixorigin/matrixone/actions/runs/28375041542/job/84062457697

```
TestLockRetryBudgetUsesLockWaitDeadline (lock_op_test.go:910)
expected: 1s
actual  : 5s
```

Cherry-pick adaptation issue: the lock wait timeout value on 4.0-dev resolves to 5s instead of the 1s expected by the test. Need to check whether:
1. The 4.0-dev default `lock_wait_timeout` differs from main
2. Or `refreshLockWaitOptions` / `lockWaitTimeout()` resolves differently on 4.0-dev
3. Or the test assertion needs to be adjusted for 4.0-dev's default value

matrix-meow:
## CI Update: 2 lock wait timeout test failures

**Run**: https://github.com/matrixorigin/matrixone/actions/runs/28424671827/job/84225051545

### Failure 1: TestLockRetryBudgetUsesLockWaitDeadline (same as before)
```
lock_op_test.go:910: expected: 1s, actual: 5s
```

### Failure 2: TestLockWaitTimeoutDefaultNoTimeout (NEW)
```
panic: test timed out after 10m0s
running tests: TestLockWaitTimeoutDefaultNoTimeout (1s)
```

The test verifies that default `LockWaitTimeout=0` means no timeout. But txn2's waiter gets permanently blocked (txn1 holds lock, never released). The waiter enters `setBlocked` + `addToLazyCheckDeadlockC` and hangs indefinitely, causing the entire lockservice test package to time out.

**Likely fix**: The test needs to unlock txn1 after a brief delay so txn2 can acquire the lock and the test can verify it waited without timeout. Or the test's lock acquisition/release sequence needs adjustment for 4.0-dev's lockservice behavior.

matrix-meow:
## Re-analysis: TestLockRetryBudgetUsesLockWaitDeadline

**Root cause**: Test uses `require.Equal(t, defaultWaitTimeOnRetryLock, wait)` which reads the **current** value of the global variable at assertion time. In Coverage mode (`go test -cover ./...`), all package tests share this global variable. Another test may modify `defaultWaitTimeOnRetryLock` without restoring it before this assertion runs.

**Why it passes locally**: Local `go test ./pkg/sql/colexec/lockop/` runs only this package. Coverage CI compiles ALL packages into one binary, affecting test ordering and global state.

**Fix**: Use an explicit constant:
```go
expectedWait := time.Second
defaultWaitTimeOnRetryLock = expectedWait
// ...
require.Equal(t, expectedWait, wait)
```

matrix-meow:
CI Update: TestLockWaitTimeoutZeroMeansFallbackToContext 10m timeout (same waiter leak). 3 distinct failures across runs: TestLockRetryBudgetUsesLockWaitDeadline (1s vs 5s, 4x), TestLockWaitTimeoutDefaultNoTimeout (10m, 1x), TestLockWaitTimeoutZeroMeansFallbackToContext (10m, 1x). All PR bugs.

matrix-meow:
New failure: TestLockRemoteWithContextTimeoutTracksLockForUnlock — remote lock tracking map empty after context timeout. PR now has 4 distinct failing tests across runs.

daviszhen:
main和 4.1 版本已经有了. 4.0 后续要下线. 不再支持了.

PR #25196, opened 2026-06-29
fix: preserve text type in conditional string casts (cherry-pick #25162)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Test and CI
- [ ] Code Refactoring

issue #23176

Cherry-pick of #25162 to 4.0-dev.

This PR fixes TEXT updates through conditional string result expressions while keeping MySQL-compatible ordinary string cast behavior.

`COALESCE`, `CASE`, and `IF/IFF` previously could select a narrower `CHAR/VARCHAR` result type when `TEXT` was mixed with short string branches such as `''`. That inserted a `TEXT -> CHAR/VARCHAR` cast before assignment, so long `TEXT` values could fail or be narrowed before being assigned back to a real `TEXT` column.

The fix keeps `TEXT + CHAR/VARCHAR/NULL` result branches on the `TEXT` type for these conditional expressions. Ordinary/explicit `CAST(TEXT AS CHAR/VARCHAR(N))` now follows MySQL-compatible truncation semantics, while real `CHAR/VARCHAR(N)` DML assignment paths use internal `cast_strict` so oversized writes are still rejected before truncation.

The strict assignment path is applied to UPDATE, INSERT projection, INSERT ... ON DUPLICATE KEY UPDATE, and update/rowset type reset paths. The BVT also covers INSERT SELECT and ON DUPLICATE KEY UPDATE oversized TEXT writes into `VARCHAR(255)`.

### Cherry-pick notes

The 4.0-dev branch had diverged from 3.0-dev, so several conflicts were resolved with the following adaptations:

- `CAST_STRICT` is assigned function id `524` (4.0-dev's next free id) instead of `377`.
- 4.0-dev's `strToStr` already carries an extra `proc` parameter and a separate geometry path; the new `strictStringWidth` parameter and truncate-vs-strict logic were applied on top of that signature, replacing 4.0-dev's earlier `shouldSkipLengthCheck` (width >= 255) heuristic.
- Assignment/insert/update paths on 4.0-dev use `if/else` (not `switch`) and lack the SET-default handling that exists on 3.0-dev; only the `forceCastExpr -> forceAssignmentCastExpr` / `makePlan2CastExpr -> makePlan2AssignmentCastExpr` changes were applied, without pulling in unrelated 3.0-dev code.
- 4.0-dev-only tests (geometry cases in `Test_strToStr_TextToCharVarchar`, etc.) were preserved and updated to the new truncation behavior.

Validation:

- `go test ./pkg/sql/colexec/external -count=1`
- `make build`


XuPeng-SH (changes requested):
I found two blockers here.

1. Generated `CHAR/VARCHAR` columns still go through the lenient cast path, so this PR silently truncates them instead of rejecting oversize values.

`buildGeneratedExpr()` still wraps generated-column expressions with `makePlan2CastExpr(...)` (`pkg/sql/plan/build_util.go`). Those generated expressions are then reused on write in the insert/update/replace paths (`pkg/sql/plan/bind_insert.go`, `pkg/sql/plan/build_update.go`, `pkg/sql/plan/bind_replace.go`). Because the generated expression itself is already pre-cast to the target type, the later write path never gets a chance to preserve the original over-width value and raise the assignment-style error.

Concretely, a generated column like `g varchar(1) generated always as (coalesce(text_col, ''))` will now truncate `"ab"` to `"a"` instead of failing with data-too-long semantics, even though generated columns are materialized as real column writes.

Suggested fix: build generated-column wrappers with the assignment cast path too (`makePlan2AssignmentCastExpr` / equivalent strict assignment semantics), and add INSERT / UPDATE / REPLACE coverage for generated `varchar(n)` columns fed by long `TEXT` expressions.

2. Parallel `LOAD DATA` now ignores the PR’s own LOCAL / non-strict exception and can fail only because the load became parallel.

The planner now rewrites the parallel load projection through `makeCastExpr()` (`pkg/sql/plan/build_load.go`), and on this head that helper unconditionally uses `makePlan2AssignmentCastExpr(...)`. But runtime load strictness is still defined by `checkLineStrict()` in `pkg/sql/colexec/external/external.go`, which explicitly returns false for `LOCAL` or non-strict SQL mode, and the comment there says LOCAL / non-strict loads should keep the lenient truncate behavior.

That means the same oversized `varchar` input can still truncate in serial / LOCAL / non-strict load, but error once the file is large enough to stay parallel. Behavior should not flip based on file size/compression/parallel path.

Suggested fix: only use the strict assignment cast in the parallel projection when the load is actually restrictive under the same predicate as `checkLineStrict()`; otherwise keep the lenient cast path so parallel and non-parallel loads match.

The rest of the coverage looks fairly good, but these are correctness issues rather than just missing tests.

heni02 (approved):

aunjgr (approved):

XuPeng-SH:
I re-checked the current head. The generated-column strict-cast issue and the parallel-vs-serial LOAD strictness mismatch both look fixed now, but I still see one remaining consistency problem:

The new `loadUsesStrictAssignmentCast()` decides whether parallel `LOAD DATA` uses `cast` or `cast_strict` while the plan is being built, by reading `sql_mode` immediately. That is fine for a normal statement, but prepared statements keep and reuse that plan. So `PREPARE` under non-strict mode and later `EXECUTE` under strict mode will still keep the lenient cast on the parallel path; the opposite mode flip has the inverse problem. The serial path still checks strictness at execution time, so prepared parallel LOAD becomes inconsistent with both serial LOAD and the current session `sql_mode`.

I think the strict/lenient choice for parallel LOAD needs to be re-evaluated at EXECUTE/compile time instead of being baked in at PREPARE time, or prepared LOAD plans need to be invalidated when `sql_mode` changes.

XuPeng-SH (approved):

PR #25197, opened 2026-06-29
cherrypick-to-main: fix rand(n) 

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24813

增加 rand 带 seed

### DIFF
--- a/pkg/sql/plan/function/func_builtin.go
+++ b/pkg/sql/plan/function/func_builtin.go
@@ -3532,13 +3532,34 @@ func builtInLog10(parameters []*vector.Vector, result vector.FunctionResultWrapp
 }
 
 type opBuiltInRand struct {
-	seed *rand.Rand
+	seed *mysqlRandState
 }
 
 var _ = newOpBuiltInRand().builtInRand
 
 func newOpBuiltInRand() *opBuiltInRand {
 	return new(opBuiltInRand)
+}
+
+const mysqlRandMaxValue uint64 = 0x3FFFFFFF
+
+type mysqlRandState struct {
+	seed1 uint64
+	seed2 uint64
+}
+
+func newMySQLRandState(seed int64) *mysqlRandState {
+	tmp := uint64(uint32(seed))
+	return &mysqlRandState{
+		seed1: (tmp*0x10001 + 55555555) % mysqlRandMaxValue,
+		seed2: (tmp * 0x10000001) % mysqlRandMaxValue,
+	}
+}
+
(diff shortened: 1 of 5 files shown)


XuPeng-SH (changes requested):
I found two blockers here.

1. `rand(constant)` is still wrong under parallel / distributed execution.

The new implementation keeps mutable state in `opBuiltInRand.seed` and advances one sequence per executor instance. But the only safeguard is `cannotParallel: true`, and that flag still is not enforced by the planner/executor. The function package itself still documents it as ineffective, and `CannotRunInParallel()` does not appear to be consumed outside the function package.

That means any plan shape that evaluates `rand(1)` in multiple pipelines / fragments / CNs will construct multiple `opBuiltInRand` instances, each restarting from the same seed. Instead of one statement-wide sequence, the sequence restarts per executor.

Suggested fix: either keep `rand(n)` disabled until execution can truly force a single evaluation stream for that overload, or implement semantics that do not depend on mutable per-executor state.

2. This PR renumbers the existing zero-arg `rand()` overload and breaks encoded / serialized expressions.

Before this change, zero-arg `rand()` used overload id `0`. This PR re-enables the seeded overload as id `0` and moves zero-arg `rand()` to id `1`. But function overload ids are encoded into `plan.Function.Func.Obj` and execution later resolves them by direct lookup. Any expression compiled/serialized with the old zero-arg `rand()` overload id `0` (for example during mixed-version execution / rolling upgrade, or any persisted encoded expression path) will now resolve to the new one-arg overload instead.

That new implementation unconditionally reads `parameters[0]`, so an old encoded zero-arg `rand()` expression will panic when executed.

Suggested fix: preserve the old overload index for zero-arg `rand()` and append the seeded overload after it, or add an explicit compatibility remap when decoding legacy overload ids.

The new tests cover the basic seeded behavior, but they do not exercise either of these blocker cases.

heni02 (approved):

aunjgr (changes requested):
Overload ID renumbering is a hard upgrade blocker. rand() was overload 0, now rand(n) is 0. Old encoded plans decode rand() as rand(n) → panic on parameters[0]. Preserve the original overload index for rand() and append rand(n) after it. parallel execution of seeded rand is also broken but the ID issue alone blocks merge.

daviszhen:
rand(seed)  分布式场景的行为 待确认. 先暂停pr

Issue #25177, opened 2026-06-26
CDC task remains running while checkpoint stalls after permanent_table_error


lxw665:
## Summary

We observed table-level CDC tasks staying in `running` state while their checkpoints stopped advancing. The tasks continuously emit `cdc.frontend.task.permanent_table_error` and do not appear to self-recover.

This is problematic operationally because the task status still looks `running`, but downstream data is no longer being synchronized.

## Impact

- CDC consumers depending on the affected tables may see stale data.
- Operators may miss the issue if they only check task status and do not verify checkpoint progress.
- The error has persisted for many hours/days and keeps retrying/logging.

## Observed behavior

Two table-level CDC tasks show the same pattern:

- task status: `running`
- checkpoint: not advancing
- log frequency: roughly every 15 seconds
- log message: `cdc.frontend.task.permanent_table_error`

Example sanitized logs:

```text
caller="frontend/cdc_exector.go:1075"
msg="cdc.frontend.task.permanent_table_error"
db="ods_s4"
table="performance_file_attach"
age="288h48m55s"
message="internal error: Error 1105 (HY000): writeto tcp4 <source>:54810-><target>:6003: read tcp4 <source>:54810-><target>:6003: read: connection reset by peer"
```

```text
caller="frontend/cdc_exector.go:1075"
msg="cdc.frontend.task.permanent_table_error"
db="ods_ecc"
table="performance_file_attach"
age="381h40m11s"
message="internal error: Error 1815 (HY000): context deadline exceeded\ntransaction error: wait active\ninternal error: createTxnOpUnsafe"
```

## Expected behavior

When CDC encounters these internal MO errors, one of the following should happen:

1. The CDC task retries and eventually recovers, with checkpoint continuing to advance.
2. If the task cannot recover, task status should become failed/unhealthy instead of remaining simply `running`.
3. The error should expose enough diagnostic information to identify whether the blocker is network, TN/CN internal service, transaction state, or a specific table/txn.

## Actual behavior

The CDC task remains `running`, but checkpoint does not advance. The same `permanent_table_error` is emitted repeatedly for a long time.

## Questions

- Is `connection reset by peer` on internal `:6003` expected to be recoverable by CDC?
- Under what condition does `createTxnOpUnsafe` return `wait active` for CDC writes?
- Should a CDC task with non-advancing checkpoint and persistent `permanent_table_error` be marked unhealthy?
[...]

PR #25201, opened 2026-06-29
Fix CDC task state for permanent table errors

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25177

This PR updates CDC task handling so table-level permanent errors no longer leave the task appearing healthy/running while checkpoints stop advancing.

  Changes:

  - Mark the CDC task as failed when a permanent table error is detected.
  - Stop active CDC readers and unregister the task from the table detector on permanent table failure.
  - Allow the CDC executor state machine to transition from running/restarting to failed.
  - Escape task state/error SQL fields when updating mo_cdc_task.err_msg.
  - Add regression tests for task failure behavior, runtime failure state transitions, and SQL escaping.


XuPeng-SH (changes requested):
The SQL-escaping part looks good, but I think the permanent-table-error state change still needs one more pass before merge.

1. This PR changes a fresh non-retryable table error from “table stays blocked until manual clear” into a task-level `Failed` state, but the recovery path is still wired only through `Resume()`. `Resume()` clears persisted table errors via `clearAllTableErrors()` (`pkg/frontend/cdc_exector.go:374-377`, helper at `887-907`), while `Failed` tasks can only `Restart`/`Cancel` and `Restart()` does not do that cleanup (`pkg/frontend/cdc_exector.go:410-470`). The existing retry contract in `pkg/cdc/error_handler.go:189-201` also still says non-retryable table errors are cleared manually through Resume. As written, after the user fixes the table and runs `RESTART CDC`, `GetTableErrMsg()` will read the same persisted watermark error again and the task will fail again immediately. I think we either need to clear per-table watermark errors on the restart-from-failed path, or keep the old pause/resume-based recovery model instead of switching the whole task to `Failed`.

2. Please add an unhappy-path regression test for the recovery flow, not just the initial fail: permanent table error -> task enters failed state -> user fixes the table -> restart/resume path clears the persisted watermark error and the task can make progress again. The new tests currently prove the state flip and SQL update, but they do not cover the user-visible recovery path that this PR changes.

XuPeng-SH (changes requested):
The latest patch fixes the original restart-clear-errors gap, but I still see two substantive problems in the new control flow.

1. `Restart()` still checks `exec.stateMachine.IsRunning()` *after* it has already transitioned the executor to `StateRestarting`. `IsRunning()` only returns true for `Starting`/`Running`, so that cleanup block is skipped for an actively running task. In that case the old TableDetector registration is left in place, and the goroutine started by `Restart()` re-enters `Start()`, which then rejects itself as already registered / already running. I think the restart path needs to capture `wasRunning` before `TransitionRestart` and use that pre-transition state for the unregister/stop logic.

2. There is still a race between `Pause()` and the new permanent-error fail path. `Pause()` moves the state to `Pausing` before it unregisters the detector and stops readers, so an in-flight `handleNewTables()` callback can still hit `failTaskForPermanentTableError()`. `SetFailed()` rejects `StatePausing`, but the helper still persists the task as failed and updates failed metrics, after which `Pause()` can complete to `Paused`. That can leave the in-memory state, persisted task state, and metrics disagreeing. I think this path needs serialization or explicit coordination with the pausing/cancelling states instead of continuing after `SetFailed()` is rejected.

XuPeng-SH:
Latest patchset still has two issues on my side:

1. `Restart()` transitions to `StateRestarting` before checking whether it needs to tear down the old execution. But `IsRunning()` only returns true for `Starting` / `Running`, so a restart from an actively running task can skip the unregister/stop path and the new `Start()` then trips over the old TableDetector registration / already-running state.

2. There is still a race between `Pause()` and the new permanent-table-error fail path. `Pause()` moves the state to `Pausing` before unregistering the detector/stopping readers, so an in-flight `handleNewTables()` callback can still reach `failTaskForPermanentTableError()`. `SetFailed()` rejects `StatePausing`, but that helper still persists failed state and updates failed metrics, after which `Pause()` can still complete to `Paused`. That can leave memory state / persisted task state / metrics out of sync.

XuPeng-SH (changes requested):
I found two correctness blockers here.

1. `RESTART` still cannot reliably recover the paused-after-error path.

`Restart()` only clears table-level permanent errors when the previous state was `Failed`:

```go
shouldClearTableErrors := stateBeforeRestart == StateFailed
```

But this PR also explicitly preserves the `Pausing` / `Paused` path when a permanent table error races with pause (`TestCdcTask_PermanentTableErrorDoesNotFailWhilePausing`). In that path, `failTaskForPermanentTableError()` returns before clearing anything, so the stale per-table watermark error remains. A later `RESTART` from `Paused` skips `clearAllTableErrors()`, and `handleNewTables()` will see the same `GetTableErrMsg(...)` again and fail the fresh run immediately.

Suggested fix: clear table errors on paused-origin restarts too (at minimum `StatePaused`, and likely any non-running restart path), or otherwise make `RESTART` consistently perform the documented error-clear step.

2. The new failed-task path leaks task-state metrics after recovery.

`failTaskForPermanentTableError()` increments the `failed` gauge, but the recovery path never decrements it. After a successful restart, `Start()` increments `running` again, so the same task is reported as both failed and running.

Suggested fix: add explicit metric transitions when leaving `StateFailed` (at least failed -> starting/restarting and failed -> cancelled), and cover restart-from-failed with metric assertions.

The unit coverage around the new failure path is otherwise pretty solid, but these are real correctness issues rather than just test gaps.

XuPeng-SH:
I re-checked the current head. The earlier restart-clear-errors / paused-restart issues look fixed now, but I still see one remaining restart race:

`Restart()` now correctly snapshots the old state and clears table errors for failed/paused restarts, but it still transitions to `StateRestarting` *before* unregistering the old TableDetector callback. `TableDetector.processCallback()` snapshots `s.Callbacks` up front, so an already-started old `handleNewTables()` can continue to run even after `UnRegister()`.

That becomes dangerous here because `SetFailed()` now allows `StateRestarting`. If the stale callback observed a permanent table error before restart cleared it, `failTaskForPermanentTableError()` can still flip the restarted task back to `Failed`, stop readers, and tear the fresh run down from the old generation.

I think this path needs some notion of execution generation / callback fencing, or a way to wait for/drain any in-flight old `handleNewTables()` callback before the new restart generation is allowed to fail the task.

XuPeng-SH (approved):

- restart from Failed/Paused now clears persisted table errors before recovery;
- failed-state metrics are decremented when leaving `StateFailed`;
- the stale callback / restart generation race is fenced with `callbackGeneration` plus `callbackMu`, and there are focused tests for both the in-flight and stale-generation cases.

At this point I don’t see a remaining substantive correctness issue. The only thing left would be extra nice-to-have coverage around a more tightly orchestrated combined concurrency scenario, but that is not a blocker.

LGTM.

PR #25224, opened 2026-06-29
fix: make CnServerMessageHandler wait observe messageCtx cancellation (partial cherry-pick #25035)

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #25025.

Partial cherry-pick of #25035 to `main`.

`main` already carries a more complete pipeline teardown-hardening framework
(`CloseWithTimeout` / `ForceCleanup` / `cleanupSpool` /
`SendPipelineSignalWithTimeout` / `WarnPipelineCleanupf`), so most of #25035 is
already superseded here. The **only** piece still missing on `main` is the
`CnServerMessageHandler` wait:

```go
if err == nil {
    <-receiver.connectionCtx.Done() // TCP close only
}
```

This blocks until the TCP connection closes and ignores query cancellation. This
PR makes it also observe `receiver.messageCtx`, so a killed/cancelled query
doesn't leave the handler blocked forever:

```go
if err == nil {
    select {
    case <-receiver.connectionCtx.Done():
    case <-receiver.messageCtx.Done():
    }
}
```

This is hardening only; it does not fix the cross-CN shuffle root cause
(issue #25025, fixed separately by #25158).

### Verification

- `go build ./pkg/sql/compile/`: OK.
- gofmt clean.


aunjgr (approved):
Simple, correct hardening. Prevents a killed/cancelled query from leaving the handler blocked forever waiting for TCP close. LGTM.

PR #25223, opened 2026-06-29
fix: bound pipeline cleanup waits + CnServerMessageHandler messageCtx escape on 4.0-dev (cherry-pick #24972)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

Related to #25025.

## What this PR does / why we need it:

Brings 4.0-dev's distributed pipeline teardown in line with `main`. Two commits:

### 1. Cherry-pick #24972 — "fix: bound pipeline cleanup waits"

`4.0-dev` still carries the old unbounded teardown path, where
`PipelineSpool.Close()` and the connector/dispatch cleanup can block forever
waiting for receivers to drain, leaving a query unkillable and table locks held
until the CN is restarted. This backports the bounded-cleanup framework already
on `main`:

- `PipelineSpool.CloseWithTimeout` / `ForceCleanup` (`cleanupOnce`) — bounded
  wait + deferred force cleanup instead of blocking forever on `csDoneSignal`
- connector / dispatch cleanup rewritten around `PipelineSignalSendTimeout`,
  terminal spool signals, and `cleanupSpool` for later force-cleanup
- `WarnPipelineCleanupf` diagnostics + cleanup metrics
- `process_spoolr` / `pipeline` teardown bounding

Cherry-picked cleanly from commit `629f58da6` (no conflicts).

### 2. CnServerMessageHandler messageCtx escape (counterpart of #25224)

#24972 does not touch `CnServerMessageHandler`, whose wait blocks on
`connectionCtx` (TCP close) only and ignores query cancellation. Make it also
observe `receiver.messageCtx`, so a killed/cancelled query doesn't leave the
handler blocked forever:

```go
if err == nil {
    select {
    case <-receiver.connectionCtx.Done():
    case <-receiver.messageCtx.Done():
    }
}
```

(The same one-liner is proposed for `main` in #25224.)

This is teardown hardening only; it does not fix the cross-CN shuffle root cause
(issue #25025, fixed separately by #25158).

### Verification

- `go build` of all changed packages: OK.
- Unit tests pass: `pkg/container/pSpool`, `pkg/sql/colexec/connector`,
  `pkg/sql/colexec/dispatch`, `pkg/vm/pipeline`, `pkg/vm/process`.
- gofmt clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25230, opened 2026-06-29
fix: make CnServerMessageHandler wait observe messageCtx cancellation (4.0-dev counterpart of #25224)

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #25025.

4.0-dev counterpart of #25224 (the same one-line hardening already proposed for
`main`).

The `CnServerMessageHandler` wait blocks on `connectionCtx` (TCP close) only and
ignores query cancellation:

```go
if err == nil {
    <-receiver.connectionCtx.Done()
}
```

Make it also observe `receiver.messageCtx`, so a killed/cancelled query doesn't
leave the handler blocked forever waiting for the TCP connection to close:

```go
if err == nil {
    select {
    case <-receiver.connectionCtx.Done():
    case <-receiver.messageCtx.Done():
    }
}
```

This complements the #24972 backport (#25223), which brings the bounded
pipeline-cleanup framework to 4.0-dev but does not touch this handler. Hardening
only; it does not fix the cross-CN shuffle root cause (issue #25025, fixed
separately by #25158).

### Verification

- `go build ./pkg/sql/compile/`: OK.
- gofmt clean.

### DIFF
--- a/pkg/sql/compile/remoterunServer.go
+++ b/pkg/sql/compile/remoterunServer.go
@@ -117,7 +117,13 @@ func CnServerMessageHandler(
 		// to prevent some strange handle order between 'stop sending message' and others.
 		// todo: it is tcp connection now. should be very careful, we should listen to stream context next day.
 		if err == nil {
-			<-receiver.connectionCtx.Done()
+			// Also observe messageCtx cancellation so a killed query
+			// doesn't leave this handler blocked forever waiting for
+			// the TCP connection to close (issue #25025).
+			select {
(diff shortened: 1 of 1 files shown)


ck89119:
Folded into #25223: the CnServerMessageHandler messageCtx fix is now appended there instead of a separate PR. Closing.

Issue #25221, opened 2026-06-29
[Bug]: NULL IN/NOT IN list errors instead of returning NULL


Ariznawlll:
### MatrixOne version

Verified on latest `origin/main` at `24d9d1ede` (`fix(function,plan): GREATEST/LEAST numeric promotion (#25145) and ambiguous-column typo (#25146) (#25168)`).

### Reproduce

```sql
select null in (1, null) as r;
select null not in (1, null) as r;
```

### Expected behavior

MySQL 8.0.45 returns `NULL` for both expressions:

```text
r
NULL

r
NULL
```

### Actual behavior

MatrixOne main returns an internal cast error for both expressions:

```text
ERROR 20203 (HY000): invalid argument operator cast, bad value [ANY ANY]
```

### Notes

This affects direct scalar `IN` / `NOT IN` predicates when the left operand is `NULL`. MySQL comparison semantics propagate `NULL` here rather than throwing an error.

References:
- https://dev.mysql.com/doc/refman/8.4/en/comparison-operators.html
- https://dev.mysql.com/doc/refman/8.4/en/type-conversion.html
- https://docs.pingcap.com/tidb/stable/type-conversion-in-expression-evaluation/

jiangxinmeng1:
fixed by #25236

Ariznawlll:
Verified on latest main `7e70bee98`.

Result: fixed by #25236.

Validation SQL:

```sql
select null in (1, null) as r;
select null not in (1, null) as r;
select null in (1, 2) as r;
select null not in (1, 2) as r;
select null in (null) as r;
select null not in (null) as r;
```

All statements now return `NULL`, matching MySQL 8.0.45. No cast error is returned.

Regression case added in matrixorigin/motr#39.

Case: `14_issue_regression.issue_25221_null_in_list`
Commit: `87660dc`
Target branch marker: `@target-branch: main`

Validation on latest main passed with the issue regression group:

```text
SUMMARY | total=16 success=16 failed=0 skipped=0 | stmts=185 ok=185 fail=0
```

PR #25236, opened 2026-06-30
fix(plan): handle NULL left operand in IN predicates

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25221 

 - Return a typed boolean NULL when scalar IN / NOT IN has a constant NULL left operand.
  - Avoid generating invalid casts to ANY for expressions like `NULL IN (1, NULL)`.
  - Add regression coverage for `NULL IN (1, NULL)` and `NULL NOT IN (1, NULL)`.

### DIFF
--- a/pkg/common/moerr/error.go
+++ b/pkg/common/moerr/error.go
@@ -88,6 +88,7 @@ const (
 	ErrUpgrateError         uint16 = 20311
 	ErrInvalidTz            uint16 = 20312
 	ErrUnsupportedDML       uint16 = 20313
+	ErrOperandColumns       uint16 = 20314
 
 	// Group 4: unexpected state and io errors
 	ErrInvalidState                             uint16 = 20400
@@ -387,6 +388,7 @@ var errorMsgRefer = map[uint16]moErrorMsgItem{
 	ErrWrongDatetimeSpec:    {ER_WRONG_DATETIME_SPEC, []string{MySQLDefaultSqlState}, "wrong date/time format specifier: %s"},
 	ErrUpgrateError:         {ER_UNKNOWN_ERROR, []string{MySQLDefaultSqlState}, "CN upgrade table or view '%s.%s' under tenant '%s:%d' reports error: %s"},
 	ErrUnsupportedDML:       {ER_UNKNOWN_ERROR, []string{MySQLDefaultSqlState}, "unsupported DML: %s"},
+	ErrOperandColumns:       {ER_OPERAND_COLUMNS, []string{"21000"}, "Operand should contain %d column(s)"},
 
 	// Group 4: unexpected state or file io error
 	ErrInvalidState:                             {ER_UNKNOWN_ERROR, []string{MySQLDefaultSqlState}, "invalid state %s"},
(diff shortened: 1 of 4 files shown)


heni02 (approved):

XuPeng-SH (changes requested):

1. `NULL IN (...)` is now folded before volatility checks, so RHS side effects disappear.

The new early return replaces the whole predicate with a constant `NULL` before the RHS expressions are preserved in the plan. That bypasses the normal “do not fold volatile functions” safeguard. So a query like `SELECT NULL IN (nextval('s'))` would no longer advance the sequence, which is an observable behavior change and not compatible with the normal planner rule that volatile expressions must not be folded away.

Suggested fix: don’t replace the predicate with a constant at bind time. Keep the normal RHS/type-checking path and only fold when the RHS is provably side-effect-free.

2. The early return skips RHS validation, so invalid scalar-vs-row `IN` lists can now be silently accepted.

Because the new return happens before the normal RHS comparison-building/validation path, malformed RHS items are never checked if the left operand is a literal `NULL`. A query like `NULL IN ((1,2))` should still be rejected as an invalid scalar-vs-row comparison, not accepted as a constant `NULL`.

Suggested fix: keep the normal RHS shape/type validation, then handle the constant-NULL result after the list form is known to be valid.

The current SQL regression covers the original reported bug, but it does not protect either of these new semantic regressions.

XuPeng-SH:
I re-checked the current head. The earlier side-effect / RHS-shape problems look addressed now, but I still see one remaining issue:

The new tuple-arity rejection for `NULL IN ((1,2))` is raised with `moerr.NewInternalError(...)`, which means clients get `internal error: Operand should contain 1 column(s)` under the generic internal-error code/state. That is a behavior regression compared with the normal operand-count contract (`Operand should contain N column(s)` / `ER_OPERAND_COLUMNS`), and it does not really describe a server-internal failure — it is a user/input error.

I think this should use the normal operand-columns error path instead of `NewInternalError`, so the SQLSTATE / error code / message shape stay compatible with existing `IN` / subquery arity failures.

XuPeng-SH (approved):
Re-reviewing the current head: the earlier blockers look resolved now.

- The binder no longer folds `NULL IN (...)` to a constant before preserving the RHS.
- The new regression proves volatile RHS still executes (`nextval` advances and `currval` becomes `1`).
- Row-valued RHS validation now happens before the NULL special-case, and `NULL IN ((1,2))` is covered with the expected operand-columns error.

At this point I don’t see a remaining substantive code issue. The only remaining gaps are nice-to-have symmetry tests like `NULL NOT IN (nextval(...))`, which are not blockers.

LGTM.

ouyuanning (approved):

aunjgr (approved):
All three concerns from the earlier review are resolved:

1. **Volatility**: `NULL IN (nextval('s'))` correctly advances the sequence (regression proves `currval = 1`).
2. **RHS validation**: `NULL IN ((1,2))` now uses `ErrOperandColumns` (ER_OPERAND_COLUMNS, SQLSTATE 21000) — proper user-facing error, not internal error.
3. **No early folding**: validation happens before the NULL special-case.

LGTM.

xzxiong (approved):

Issue #25205, opened 2026-06-29
CN recovery: end() panic on RunTask failure can crash CN during recovery


XuPeng-SH:
## Summary
During CN restart recovery (`doRecovery`), the `end()` callback calls `startAsyncCheckCommitTask` with a bare `panic` on failure. If `s.stopper.RunNamedTask` returns `ErrUnavailable` (stopper not in `running` state), this causes a fatal CN crash during recovery startup.

## Root Cause
```go
// service_recovery.go: end()
if err := s.startAsyncCheckCommitTask(txnCtx); err != nil {
    panic(err)  // bare panic, no recovery context
}
```

`RunNamedTask` returns `ErrUnavailable` when `s.mu.state != running`. While `startRecovery()` blocks on `waitRecoveryCompleted`, the stopper should be running. However, a concurrent shutdown race or abnormal stopper state transition could trigger this path.

## Contrast
The `startRecovery` function uses `logger.Fatal` (with proper logging context), while `end()` uses a bare `panic` — inconsistent error handling.

## Impact
- CN fails to complete recovery → potential crash loop on restart
- No contextual logging for debugging

## Severity
**Medium**

## Suggested Fix
Replace `panic(err)` with structured error logging and graceful degradation:
```go
if err := s.startAsyncCheckCommitTask(txnCtx); err != nil {
    logutil.Error("failed to start async check commit task during recovery", zap.Error(err))
    // Don't panic — let the txn be cleaned up by zombie GC
}
```

## Related
- Same pattern exists in `startAsyncCommitTask` (line 108) — also panics

jiangxinmeng1:
fixed by #25213

Ariznawlll:
Verified fixed on latest main 2a8f3b884989.

Validation:
- Confirmed PR #25213 is included in latest main.
- Checked pkg/txn/service/service_recovery.go: recovery end() no longer panics when startAsyncCheckCommitTask/startAsyncCommitTask returns an error. It now logs structured errors with txn context.
- Ran the targeted regression test:

[...]

PR #25237, opened 2026-06-30
fix(txn): avoid recovery async task panic (#25213)

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25205

Fix CN recovery `end()` handling when async recovery task scheduling fails because the txn service stopper is unavailable.

- Replace bare `panic(err)` with structured error logging for recovery async check-commit and commit task startup failures
- Keep committing txn context when async commit task scheduling fails, instead of removing it before confirming the task started
- Add regression coverage for stopper-unavailable recovery `end()` behavior


XuPeng-SH (approved):
Looks good to me. This fixes the recovery unhappy path without widening behavior unnecessarily: it removes the panic on stopper-unavailable async scheduling, preserves the committing txn context until the async commit task is actually started, and adds focused regression coverage for both Prepared and Committing recovery states.

PR #25259, opened 2026-06-30
fix(compile): escape index name in DROP INDEX catalog write

mahirhir:
## What type of PR is this?

- [x] BUG

issue #25054

https://github.com/matrixorigin/matrixone/issues/25054

Follow-up to #25057, which fixed the same class of bug in the `ALTER INDEX` catalog writes but left the delete-by-name path unescaped.

#25057 routed the user-supplied index name through `sqlquote.EscapeString` for the catalog writes in `AlterTableInplace` (`updateMoIndexesVisibleFormat` and `updateMoIndexesAlgoParams`), because the MySQL scanner treats backslash as an escape inside `'...'`, so an unescaped quote or backslash in `name = '%s'` can corrupt the value or break out of the literal.

The sibling delete-by-name format `deleteMoIndexesWithTableIdAndIndexNameFormat` (util.go) has the identical `where table_id = %v and name = '%s'` clause, but it still interpolated the raw index name at its two call sites in `ddl.go`:

- the `ALTER TABLE ... DROP INDEX` branch in `AlterTableInplace`
- `DropIndex`

A backticked index identifier may legally contain a single quote or a backslash, so dropping such an index builds a malformed `delete from mo_catalog.mo_indexes ... name = '...'`. This routes both call sites through `sqlquote.EscapeString`, identical to the update branches #25057 fixed.

Added `TestDropIndexNameEscaping` next to `TestAlterIndexVisibleEscaping`, reusing the same crafted names (`idx'; drop table mo_indexes; --`, `trailing\`, `both'and\slash`, ...). Each now produces exactly one parseable statement with the name fully quoted; before the fix these break out into a parse error, an extra statement, or a silently altered WHERE clause.


aunjgr (approved):
Reviewed current head. User-supplied index names are escaped in both DROP INDEX catalog-write paths, and the tests cover quotes, backslashes, and statement-injection shapes. I found no blocker.

aunjgr (approved):
Reviewed current head. User-supplied index names are escaped in both DROP INDEX catalog-write paths, and the tests cover quotes, backslashes, and statement-injection shapes. I found no blocker.

mahirhir:
Thanks for the review back in July, and sorry this sat so long on my side.

Rebased onto current main. The only conflict was in `ddl.go`, where the step comment above the delete had been renumbered upstream; I kept your numbering and the escape change, so the branch is back to a single commit (2 files, +47). Both catalog-write call sites still route through `sqlquote.EscapeString`: the `ALTER TABLE ... DROP INDEX` branch in `AlterTableInplace`, and `DropIndex`.

`CHECK PR VALID` was failing because the "Which issue(s) this PR fixes" section had prose instead of a reference. It now carries `issue #25054`, the issue that #25057 fixed for the `ALTER INDEX` paths. Since the workflow triggers on synchronize rather than edit, that will only re-evaluate on the next push.

The CLA is still unsigned on my side, and that is the remaining blocker for the rest of the pipeline. That one is on me to resolve, not something to action here.

mahirhir:
﻿CLA is signed now, and the branch is up to date with main. Ready when you are.

XuPeng-SH (approved):
Reviewed cad19783928722162e41c030a5e36db923a7bd34; approved. The complete two-file change applies the existing SQL string-literal escaping helper at both production users of deleteMoIndexesWithTableIdAndIndexNameFormat: standalone DROP INDEX and ALTER TABLE DROP INDEX. This correctly handles quotes and backslashes without changing the numeric table-ID predicate or transaction sequencing. The helper's escaping order matches this parser's existing contract.

The added adversarial-name test checks one parseable statement and frees parsed ASTs; git diff --check passes. No confirmed blocker or meaningful performance concern in this bounded fix. No fresh native test execution is claimed.

Optional test strengthening: assert the parsed name literal equals the original input and exercise at least one real DROP INDEX path (or a shared production SQL builder). The current test rebuilds the formatting expression independently, so reverting one ddl.go call site would not make it fail. This does not invalidate the inspected two-call-site fix.

mahirhir:
The three red checks on this PR are a failure in a package this PR does not touch, so I think a re-run is all that is needed. I cannot trigger one myself (no admin rights on the repo), hence asking.

**What failed**

```
[run_tests] [ERR] go test failed with status 1
[slow_ut_cases] elapsed=36.21s result=fail package=github.com/matrixorigin/matrixone/pkg/vectorindex
[ut_summary] [ERR] UNIT TESTING FAILED: go-test=1, analysis=0
```

**What this PR changes**

```
pkg/sql/compile/ddl.go
pkg/sql/compile/ddl_util_test.go
```

No overlap. The change is the escaping of an index name on the catalog write path for `DROP INDEX`; it adds no dependency on `pkg/vectorindex` and does not alter anything that package reads.

**The same job logged infrastructure trouble before it got to the tests**

```
ERROR: failed to connect to the docker API at unix:///var/run/docker.sock
::warning::Shanghai internal GOPROXY probe failed (dns-failed-or-timed-out); leaving GOPROXY unchanged
```

I am not claiming those caused the vectorindex failure; I did not trace it. They are the reason I suspect the runner rather than the diff, and they are worth a glance if the re-run fails the same way.

Happy to rebase on current `main` first if that is the preferred order here. The PR is approved and otherwise mergeable, so I did not want to push anything unnecessary onto a branch a reviewer has already read.

mahirhir:
The red on this PR was not from this PR. Brought the branch current in `59284b3ab`; details below in case anyone triages it before CI finishes.

The failing job was `UT Test on Ubuntu/x86`, and the case was:

```
--- FAIL: TestAppendColumnBuffer_AllTypes/json
    pkg/vectorindex/columnbuffer_test.go:103
    Error: Received unexpected error:
           invalid input: invalid JSON vector payload
```

This PR touches two files, and neither is in that package:

```
pkg/sql/compile/ddl.go
pkg/sql/compile/ddl_util_test.go
```

The branch was 37 commits behind `main`, and the difference is the relevant one. `pkg/vectorindex/columnbuffer_test.go` at this PR's old head versus at `main`:

```
old head 0c57b4e4   219 lines   line 103 = require.NoError(t, AppendColumnBuffer(c.buf, vec, mp))
main                234 lines   line 103 = {"varbinary", varlenaBuf(...)}
```

Line 103 at the old head is exactly the assertion that failed. On `main` the file is 15 lines longer because of `10ce23e7` on 2026-09-12, *test(vectorindex): encode JSON ColumnBuffer fixtures* (#28769), which is a fix to the JSON fixtures the failing subtest uses.

So the branch was running a version of that test that had already been fixed upstream. Merging `main` picks the fix up; the merge was clean and the diff against `main` is unchanged at `+47/-2` across the same two files.

I could not run the suite locally to show it green before pushing: the build needs `jemalloc.h` and my machine does not have the cgo dependency, so `go test` fails at compile rather than at the test. I am not going to claim a local pass I did not observe. CI on `59284b3ab` is the check that matters here, and it should now run against the corrected fixtures.

The change itself is unchanged from the original review: `DROP INDEX` wrote the index name into the catalog delete without escaping it, so an index whose name contains a quote produced malformed SQL. The added test covers the quoted-name case.

mahirhir:
Correcting my own comment above: I implied the merge would clear `UT Test`, and it did not. The job is red again on `59284b3ab`. Here is what actually changed.

**The failure I diagnosed is gone.** `TestAppendColumnBuffer_AllTypes/json` does not appear in this run at all, which is the part the fixture merge was supposed to fix.

**Two different cases fail now:**

```
FAIL_UT_CASES=TestISCPExecutor6,TestProbeDoesNotAcquireWhenCleanupReservationIsFull
--- FAIL: TestISCPExecutor6 (22.76s)
[TEST]: TestProbeDoesNotAcquireWhenCleanupReservationIsFull
[RESULT]: fail   [ELAPSED]: 3.01
```

Neither is in this PR's packages. This PR touches `pkg/sql/compile/ddl.go` and `pkg/sql/compile/ddl_util_test.go`; `TestISCPExecutor6` lives in `pkg/vm/engine/test/change_handle_test.go`.

**I am not going to call them flaky on my own authority**, but the repository has already characterised this class. #26667 (merged 2026-08-04, for issue #26522) describes it directly:

> The Ubuntu/x86 UT job runs package tests with `UT_PARALLEL=6` on an 8-core/16-GB runner. Several test binaries can therefore keep independent LogService/TN/CN clusters alive concurrently. Under that load, delayed Dragonboat ticks and HAKeeper heartbeat/proposal timeouts can prevent TN shard readiness and produce downstream connection resets.

Both of today's failures are cluster-lifecycle and timing shaped, and the surrounding log is full of `logtail-server` shutdown errors, which matches that description rather than anything a `DROP INDEX` escaping change could reach.

So: a re-run would tell us, and I would rather ask for one than assert an answer. If it reproduces on a re-run I will stop attributing it to contention and dig further.

One more thing I should have said in my previous comment rather than implying otherwise: I still cannot run this suite locally (the build needs `jemalloc.h`, which my machine does not have), so every claim here comes from the CI logs and the repository, not from a local run.

mahirhir:
Walking back my own explanation before anyone acts on it.

I attributed the two failures to the runner-contention class described in #26667. **That is not supported by the evidence I have now.** Sampling the same job on other open PRs:

```
PR 28811   Matrixone CI execution / UT Test on Ubuntu/x86 = SUCCESS
PR 28800   SUCCESS
PR 28796   SUCCESS
PR 28787   SUCCESS
```

Four for four green on the job that is red here. If this were contention on a shared runner I would expect it scattered across PRs, not landing only on mine. The contention story is still *possible* — those four ran at 04:13 and mine at 12:57, so they are not a simultaneous sample — but I should not have offered it as the explanation on the strength of a name match and a log full of shutdown noise. I said I would stop attributing it to contention if it did not hold up, so: withdrawn.

What I can state and what I cannot:

- My diff touches `pkg/sql/compile/ddl.go` and `pkg/sql/compile/ddl_util_test.go`. The failing cases are `TestISCPExecutor6` (`pkg/vm/engine/test/`) and `TestProbeDoesNotAcquireWhenCleanupReservationIsFull`. Different packages — that part is still a fact.
- Why those two fail on this branch and not on the four above, I do not know, and I cannot find out locally: the build needs `jemalloc.h`, so `go test` fails at compile on my machine rather than running.

The branch was also 6 commits behind `main` again. I have merged it (`366b397a6`, clean merge, diff against `main` unchanged at `+47/-2` across the same two files), which refreshes it and gives CI a fresh run on current `main` — that run is the only instrument either of us has here.

If it comes back red on `366b397a6`, the failures are attributable to this branch and I will stop guessing and ask for a maintainer to run the two cases with the diff reverted, since I cannot. If it comes back green, the earlier run was environmental after all and I will say so rather than quietly moving on.

mahirhir:
Result on the refreshed head, as promised in both directions.

**`UT Test on Ubuntu/x86` is green on `366b397a6`.**

```
UT Test on Ubuntu/x86  = SUCCESS
```

The two cases that failed on `59284b3ab` — `TestISCPExecutor6` and `TestProbeDoesNotAcquireWhenCleanupReservationIsFull` — do not fail here. The only change between the two heads is a merge of `main`; nothing in this PR's two files moved.

So the failures were not deterministic in this branch. I want to be precise about what that does and does not restore: it is **compatible** with the transient/environmental reading, and it is **not** evidence for the specific contention mechanism I named and then withdrew. I do not know which of the 6 merged commits, or which run-to-run condition, made the difference, and I am not going to invent a third story.

**`Coverage` is still red, and it is not specific to this PR.** The error is:

```
##[error]coverage merge blocked: a required producer did not publish a complete coverage profile; no coverage…
```

Sampling the same job on other open PRs right now:

```
PR 28811  Coverage = SUCCESS
PR 28800  Coverage = FAILURE
PR 28796  Coverage = FAILURE
```

Two of three others red on the same job, with an upstream-producer message rather than anything about a diff. That is the sampling I should have done before offering an explanation last time, so I am doing it before offering one now: this looks like the coverage pipeline, not this branch, and I will leave it at "looks like" unless someone wants me to dig further.

`CI Required` is red because `Coverage` is.

Summary of where the PR stands: the change is `+47/-2` across `pkg/sql/compile/ddl.go` and `pkg/sql/compile/ddl_util_test.go`, current with `main`, `MERGEABLE`, `UT Test` green. `DROP INDEX` wrote the index name into the catalog delete unescaped, so an index whose name contains a quote produced malformed SQL; the added test covers the quoted-name case.

mahirhir:
I merged `main` into this branch twice today (`59284b3a`, `366b397a`) assuming the red was drift. It was not, and the result did not change. Posting what the run actually says so nobody repeats that.

Every test job on the latest run reports success. From the `CI Required` step's own input:

```json
"matrixone-ci":            { "result": "success" },
"matrixone-compose-ci":    { "result": "success" },
"matrixone-standalone-ci": { "result": "success" },
"matrixone-ut-coverage":   { "result": "success",
                             "outputs": { "coverage_ready": "false", "coverage_status": "" } },
"matrixone-coverage-merge":{ "result": "failure" }
```

`CI Required` then fails with exactly one reason:

```
Error: matrixone-coverage-merge: failure
    at verifyResults (.github/ci/change-scope.cjs:90:49)
```

And the merge job says what happened in its own words:

```
##[error]coverage merge blocked: a required producer did not publish a complete coverage profile;
         no coverage verdict was calculated
```

So the coverage producer finished successfully while declaring `coverage_ready=false` and an empty `coverage_status`, and the merge step correctly reported that no verdict could be computed. Nothing here is a statement about the change in this PR. The change is two lines in one file.

I checked whether this is a repository-wide condition before saying anything, because "it must be your infrastructure" is the easy and usually wrong claim:

```
20 most recently updated open PRs
  no failing check at all:        13
  Coverage failing:                3   (this PR is one)
  CI Required failing:             4
```

So most PRs are fine right now. This is a minority state rather than an outage, which is also why merging `main` did nothing for it.

What would help from my side, whichever is easier for you:

- I can push an empty commit to re-trigger the workflow, if a fresh run is all this needs.
- Or if the coverage producer needs a re-run from someone with write access, say so and I will leave the branch alone.

One observation you are free to ignore: "no coverage verdict was calculated" and "coverage fell below the threshold" are different states, and only the second says something about the diff. Right now both arrive at a reviewer as the same red check on a required gate.

PR #25260, opened 2026-06-30
cherrypick: fix conv

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/23845

修改 conv  基数转换函数的兼容性问题.

对于` CONV('g', 16, 10), CONV('2', 2, 10), or CONV('10xyz', 10, 16)` 这类场景,  mo一直的行为是返回错误. mysql是返回0,NULL. 这一点上mo不会与mysql对齐.

### DIFF
--- a/pkg/sql/plan/function/func_binary.go
+++ b/pkg/sql/plan/function/func_binary.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"math/big"
 	"sort"
 	"strconv"
 	"strings"
@@ -2557,9 +2558,17 @@ func Conv(ivecs []*vector.Vector, result vector.FunctionResultWrapper, proc *pro
 		return nil
 	}
 
-	// Validate base ranges (2-36)
-	if fromBase < 2 || fromBase > 36 || toBase < 2 || toBase > 36 {
-		return moerr.NewInvalidInputf(proc.Ctx, "conv base must be between 2 and 36, got from_base=%d, to_base=%d", fromBase, toBase)
+	absFromBase := absInt64(fromBase)
+	absToBase := absInt64(toBase)
+
+	// Invalid bases return NULL.
+	if absFromBase < 2 || absFromBase > 36 || absToBase < 2 || absToBase > 36 {
+		for i := uint64(0); i < uint64(length); i++ {
+			if err = rs.AppendBytes(nil, true); err != nil {
+				return err
+			}
+		}
+		return nil
 	}
 
 	// Handle different input types for N
@@ -2606,6 +2615,21 @@ func Conv(ivecs []*vector.Vector, result vector.FunctionResultWrapper, proc *pro
 	}
 }
 
+func absInt64(v int64) int64 {
+	if v < 0 {
+		return -v
+	}
+	return v
+}
+
+func formatUnsignedToBase(val uint64, toBase int64) string {
+	base := absInt64(toBase)
+	if toBase < 0 {
+		return strings.ToUpper(strconv.FormatInt(int64(val), int(base)))
+	}
+	return strings.ToUpper(strconv.FormatUint(val, int(base)))
+}
+
 func convString(nVec *vector.Vector, fromBase, toBase int64, rs *vector.FunctionResult[types.Varlena], length int, selectList *FunctionSelectList) error {
 	nParam := vector.GenerateFunctionStrParameter(nVec)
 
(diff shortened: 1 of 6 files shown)


XuPeng-SH (changes requested):

The uppercase / negative-base fixes look good, but the new strict string parser changes `CONV()` semantics in a way that is not MySQL-compatible: malformed input now aborts the entire statement instead of producing a row-local conversion result.

With the new `parseConvStrictString()` path, inputs like `CONV('g', 16, 10)`, `CONV('2', 2, 10)`, or `CONV('10xyz', 10, 16)` return `moerr.NewInvalidInput...` and fail the query. But for a MySQL-compatible `CONV()` implementation, those should stay non-fatal: if the first character is invalid, the result is `0`; if there is a valid numeric prefix, convert the prefix and ignore the trailing junk (`CONV('10xyz', 10, 16)` -> `A`). The existing BVT comments in `func_conv.sql` already call out that expected behavior.

So I would keep the base-sign / formatting fixes, but make malformed-string handling non-fatal: parse the longest valid prefix, return `0` when the first character is invalid, and do not raise an execution error for row-local conversion failures.

daviszhen:
> I think this still needs changes before merge.
> 
> The uppercase / negative-base fixes look good, but the new strict string parser changes `CONV()` semantics in a way that is not MySQL-compatible: malformed input now aborts the entire statement instead of producing a row-local conversion result.
> 
> With the new `parseConvStrictString()` path, inputs like `CONV('g', 16, 10)`, `CONV('2', 2, 10)`, or `CONV('10xyz', 10, 16)` return `moerr.NewInvalidInput...` and fail the query. But for a MySQL-compatible `CONV()` implementation, those should stay non-fatal: if the first character is invalid, the result is `0`; if there is a valid numeric prefix, convert the prefix and ignore the trailing junk (`CONV('10xyz', 10, 16)` -> `A`). The existing BVT comments in `func_conv.sql` already call out that expected behavior.
> 
> So I would keep the base-sign / formatting fixes, but make malformed-string handling non-fatal: parse the longest valid prefix, return `0` when the first character is invalid, and do not raise an execution error for row-local conversion failures.

对于 CONV('g', 16, 10), CONV('2', 2, 10), or CONV('10xyz', 10, 16) 这类场景, mo一直的行为是返回错误. mysql是返回0. 这一点上mo不会与mysql对齐.

XuPeng-SH (changes requested):

The uppercase / negative-base fixes look fine, but the malformed-string path is still a substantive behavioral regression on the target branch.

The important point is that this is not just “not matching MySQL”. It also changes current `main` behavior.

I checked the target-branch baseline directly:
- current `main` `test/distributed/cases/function/func_other_conv.result` expects
  - `CONV('G',16,10) -> NULL`
  - `CONV('2',2,10) -> NULL`
- this PR changes those to hard errors.

And in `pkg/sql/plan/function/func_binary.go`, `convString()` now calls `parseConvStrictString()` and immediately returns the error. So malformed or partially malformed inputs like:
- `CONV('g', 16, 10)`
- `CONV('2', 2, 10)`
- `CONV('8', 8, 10)`
- `CONV('10xyz', 10, 16)`
- `CONV('+', 10, 16)`

now abort the whole statement instead of producing a row-local conversion result.

That is a real compatibility regression for the target branch, and it is larger than a constant-literal edge case because mixed-row queries now become statement-fatal.

Suggested fix:
1. keep the uppercase / signed-base fixes;
2. parse the maximal valid prefix for string input instead of rejecting the whole string;
3. return `0` / row-local non-fatal result semantics when no valid prefix exists, instead of raising `InvalidInput`;
4. add a mixed-row regression like `SELECT CONV(val,16,10) FROM t` where one row is valid and one row is malformed, so this statement-abort behavior is pinned down explicitly.

The PR description/comment says MO has "always" errored for these malformed string cases, but the current `main` baseline results contradict that, so I don’t think that rationale is enough to merge this as-is.

daviszhen:
> I think this still needs changes before merge.
> 
> The uppercase / negative-base fixes look fine, but the malformed-string path is still a substantive behavioral regression on the target branch.
> 
> The important point is that this is not just “not matching MySQL”. It also changes current `main` behavior.
> 
> I checked the target-branch baseline directly:
> 
> * current `main` `test/distributed/cases/function/func_other_conv.result` expects
>   
>   * `CONV('G',16,10) -> NULL`
>   * `CONV('2',2,10) -> NULL`
> * this PR changes those to hard errors.
> 
> And in `pkg/sql/plan/function/func_binary.go`, `convString()` now calls `parseConvStrictString()` and immediately returns the error. So malformed or partially malformed inputs like:
> 
> * `CONV('g', 16, 10)`
> * `CONV('2', 2, 10)`
> * `CONV('8', 8, 10)`
> * `CONV('10xyz', 10, 16)`
> * `CONV('+', 10, 16)`
> 
> now abort the whole statement instead of producing a row-local conversion result.
> 
> That is a real compatibility regression for the target branch, and it is larger than a constant-literal edge case because mixed-row queries now become statement-fatal.
> 
> Suggested fix:
> 
> 1. keep the uppercase / signed-base fixes;
> 2. parse the maximal valid prefix for string input instead of rejecting the whole string;
> 3. return `0` / row-local non-fatal result semantics when no valid prefix exists, instead of raising `InvalidInput`;
> 4. add a mixed-row regression like `SELECT CONV(val,16,10) FROM t` where one row is valid and one row is malformed, so this statement-abort behavior is pinned down explicitly.
> 
> The PR description/comment says MO has "always" errored for these malformed string cases, but the current `main` baseline results contradict that, so I don’t think that rationale is enough to merge this as-is.

这些场景 mo都是报错. 返回NULL和 0 是错误的.

daviszhen:
无法推update. 先关闭

Issue #25200, opened 2026-06-29
[Bug]: INSERT/UPDATE string width enforcement not fully aligned with MySQL sql_mode semantics


ck89119:
### Background

PR #25162 (merged to `3.0-dev`, cherry-picked to `4.0-dev`) reworked string CAST semantics:

- Explicit `CAST(... AS CHAR/VARCHAR(N))` now truncates over-length values (MySQL-compatible).
- Assignment paths (INSERT projection / INSERT ... SELECT / INSERT VALUES rowset / INSERT ... ON DUPLICATE KEY UPDATE / UPDATE / column `DEFAULT` / `ON UPDATE`) route CHAR/VARCHAR targets through the internal `cast_strict`, so over-length writes are **rejected** (error 1406).
- `LOAD DATA` rejects over-length CHAR/VARCHAR only when `checkLineStrict = !Local && StrictSqlMode`.

This issue tracks the remaining gaps between the assignment-path behavior and MySQL's full width-enforcement semantics.

### Actual Behavior

1. **No `sql_mode` gating on INSERT/UPDATE assignment paths.** `forceAssignmentCastExpr` / `makePlan2AssignmentCastExpr` unconditionally use `cast_strict` for CHAR/VARCHAR targets, regardless of `sql_mode`. So even if the user removes `STRICT_TRANS_TABLES` / `STRICT_ALL_TABLES`, INSERT/UPDATE still error on over-length writes. This is also inconsistent with the `LOAD DATA` path, which *does* honor `sql_mode`.

2. **No trailing-space exemption.** `strToStr` rejects whenever `runeCount(s) > destLen`. In MySQL, inserting a value whose over-length part is **only trailing spaces** into CHAR/VARCHAR does not raise 1406 even in strict mode — the spaces are truncated with a warning.

3. **Silent truncation, no warning.** When truncation happens (explicit CAST, or future non-strict mode), `truncateStringByRunes` truncates silently. MySQL emits `1265 Data truncated for column` (the error code is already defined in MatrixOne but unused on this path).

4. **No `INSERT IGNORE` downgrade.** In MySQL, `INSERT IGNORE` downgrades the 1406 error to a warning + truncation even under strict mode. MatrixOne has no such downgrade on the assignment path.

### Expected Behavior (MySQL-aligned)

[...]

PR #25262, opened 2026-06-30
fix: align INSERT/UPDATE string width enforcement with MySQL sql_mode (#25200)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25200

Follow-up to #25162. It aligns CHAR/VARCHAR width enforcement on DML assignment
paths (INSERT projection / INSERT ... SELECT / INSERT VALUES rowset / INSERT ...
ON DUPLICATE KEY UPDATE / UPDATE) with MySQL's `sql_mode` semantics, while
keeping DDL `DEFAULT` / `ON UPDATE` validation strict — MySQL enforces those at
the DDL layer (error 1067), independent of `sql_mode`.

Items 1–3 of the issue are addressed. Item 4 (the `1265 Data truncated` warning)
is intentionally out of scope; truncation is silent for now.

### 1. `sql_mode` gating on assignment paths (runtime)

Adds a new internal `cast_assign` function used by DML assignment paths for
CHAR/VARCHAR targets. It reads `sql_mode` **at runtime**:

- strict (`STRICT_TRANS_TABLES` / `STRICT_ALL_TABLES`) → reject (1406);
- non-strict → truncate.

The gating is done at runtime rather than at plan-build time on purpose: MO
caches plans (`pkg/frontend/plan_cache.go`), so baking the `cast` vs `cast_strict`
choice into the plan would go stale after `SET sql_mode`. This mirrors how the
`LOAD DATA` path already resolves strictness at runtime. The existing
`cast_strict` is retained for DDL `DEFAULT` / `ON UPDATE` validation
(`makePlan2AssignmentCastExpr`) and stays always-strict.

### 2. Trailing-space exemption

Over-length values whose excess runes are only trailing spaces are accepted
(truncated) even under strict mode, matching MySQL. The exemption applies to
`cast_assign`; the DDL `cast_strict` does not apply it.

### 3. `INSERT IGNORE` downgrade

Under `INSERT IGNORE`, over-length writes are truncated instead of rejected.
`INSERT IGNORE` is part of the SQL text (parsed as `OnDuplicateUpdate == [nil]`),
so it is gated at plan time (cache-safe) via a `QueryBuilder.isInsertIgnore` flag.

### INSERT VALUES constant path

INSERT VALUES string literals are constant-folded through
`util.SetInsertValueString` (not a cast function). The same
`sql_mode` / trailing-space / `INSERT IGNORE` resolution is applied there via a
new `function.ResolveAssignStringWidth` helper, with the IGNORE flag threaded
through `MakeInsertValueConstExpr`.

### Tests

- UT: `pkg/sql/plan/function/func_cast_width_test.go` (strToStr width matrix,
  `isStrictSqlMode`, `NewAssignCast`, `ResolveAssignStringWidth`) and plan-level
  DML / IGNORE / DDL cast assertions in `pkg/sql/plan/build_test.go`.
- BVT: `test/distributed/cases/dml/insert/insert_string_width_sqlmode`.
- No regression on the existing `update_text_coalesce_cast` and
  `load_data_string_width` BVTs.

Validation:

- `go test ./pkg/sql/plan/function ./pkg/sql/plan ./pkg/sql/util ./pkg/sql/colexec/external`
- `make build`
- mo-tester genrs + run on the new and existing BVTs (100%)

### DIFF
--- a/pkg/sql/plan/bind_insert.go
+++ b/pkg/sql/plan/bind_insert.go
@@ -30,6 +30,11 @@ import (
 )
 
 func (builder *QueryBuilder) bindInsert(stmt *tree.Insert, bindCtx *BindContext) (int32, error) {
+	// INSERT IGNORE is parsed as OnDuplicateUpdate == [nil] (see mysql_sql.y).
+	// Under IGNORE, over-length CHAR/VARCHAR writes are downgraded to truncation
+	// instead of being rejected (1406), matching MySQL.
+	builder.isInsertIgnore = len(stmt.OnDuplicateUpdate) == 1 && stmt.OnDuplicateUpdate[0] == nil
+
 	dmlCtx := NewDMLContext()
 	err := dmlCtx.ResolveTables(builder.compCtx, tree.TableExprs{stmt.Table}, nil, nil, true)
 	if err != nil {
@@ -1145,7 +1150,7 @@ func (builder *QueryBuilder) initInsertReplaceStmt(bindCtx *BindContext, astRows
 		case isGeometryPlanType(&colTyp):
 			projExpr, err = funcCastForGeometryType(builder.GetContext(), projExpr, colTyp)
 		default:
-			projExpr, err = forceAssignmentCastExpr(builder.GetContext(), projExpr, colTyp)
+			projExpr, err = forceAssignmentCastExprWithIgnore(builder.GetContext(), projExpr, colTyp, builder.isInsertIgnore)
 		}
 		if err != nil {
 			return 0, nil, nil, err
@@ -1385,7 +1390,7 @@ func (builder *QueryBuilder) buildValueScan(
 			binder.builder = builder
 			for _, r := range stmt.Rows {
 				if nv, ok := r[i].(*tree.NumVal); ok && !isEnumOrSetPlanType(&col.Typ) {
-					expr, err := MakeInsertValueConstExpr(proc, nv, &colTyp)
+					expr, err := MakeInsertValueConstExpr(proc, nv, &colTyp, builder.isInsertIgnore)
 					if err != nil {
 						return 0, err
 					}
--- a/pkg/sql/plan/bind_insert.go
+++ b/pkg/sql/plan/bind_insert.go
@@ -1430,7 +1430,8 @@ func (builder *QueryBuilder) buildValueScan(
 						return 0, err
 					}
 				}
-				defExpr, err = forceCastExpr2(builder.GetContext(), defExpr, colTyp, targetTyp)
+				defExpr, err = forceCastExpr2WithIgnore(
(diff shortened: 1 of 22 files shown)


XuPeng-SH (changes requested):
I found several correctness blockers here.

1. `INSERT IGNORE` still rejects non-literal `VALUES` expressions.

The IGNORE downgrade is only threaded into the `tree.NumVal` fast path. For non-literal `VALUES` expressions such as `repeat('x', 100)`, planning falls through to `forceCastExpr2(...)`, which hardwires `cast_assign` for `CHAR/VARCHAR`. So in strict mode `INSERT IGNORE` still throws 1406 instead of truncating.

Suggested fix: thread `isInsertIgnore` through the non-literal `VALUES` path as well, or replace `forceCastExpr2` there with an assignment-cast helper that can choose lenient `cast` under IGNORE.

2. Prepared `INSERT ... VALUES ('literal')` still freezes `sql_mode` at prepare time.

The constant-literal path resolves truncation vs rejection during planning via `SetInsertValueString(...)`, so the decision is baked into the prepared plan. That means changing `sql_mode` after `PREPARE` does not reliably affect literal `INSERT VALUES` statements, which is exactly the stale-plan problem this PR is trying to avoid elsewhere with runtime `cast_assign`.

Suggested fix: defer width enforcement for literal `VALUES` strings to execution time as well, instead of materializing the post-truncation value during planning.

3. The new constant-path helper now relaxes `VARBINARY` / non-`CHAR` targets too.

`SetInsertValueString` applies `ResolveAssignStringWidth(...)` to every non-`TEXT` / non-`DATALINK` / non-`BINARY` string literal, not just `CHAR/VARCHAR`. That includes `VARBINARY`, but the runtime assignment logic only grants sql_mode/trailing-space leniency to `CHAR/VARCHAR`. So overflow behavior can now depend on whether the source is a literal `VALUES` constant or an expression / `INSERT ... SELECT`.

Suggested fix: restrict `ResolveAssignStringWidth` in `SetInsertValueString` to `CHAR/VARCHAR` targets only.

4. The fallback planner path never sets `isInsertIgnore`.

`builder.isInsertIgnore` is initialized in `bindInsert`, but not in `buildInsert`. When planning falls back from `bindInsert` to `buildInsert` on `ErrUnsupportedDML`, the fallback code still consults `builder.isInsertIgnore` in its cast/constant helpers, but the flag is always false. So `INSERT IGNORE` semantics become planner-path dependent.

Suggested fix: initialize `builder.isInsertIgnore` in `buildInsert` the same way as in `bindInsert`, and add a regression that forces the fallback path.

Coverage is otherwise fairly good, but these are code-level blockers rather than just test gaps.

Issue #25040, opened 2026-06-17
[Bug]: CRC in 2KB data block cause unnecessary memmove (20% CPU time)


cpegeric:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

6b98c6b3a3b63dca97a489cf16c49ee6169dffa1

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

The data file contains multiple 2KB CRC-block which require memove to remove the CRC (4 byte) for every 2KB to make a contiguous memory for lz4 decompression.  de-interleave operations took 20% CPU time.

### Cold-scan profile — where cold time goes (read + decode, NOT materialize)

CPU profile of an f32 table scan (`SELECT COUNT(*) … WHERE l2_distance(embedding,'[…]')≥0`) with the
SHARED cache reverted to the default ~512 MB (the 3 GB column ≫ cache, so scans stay cold), vs the
warm scan. `runtime.memmove` (46% of cold CPU) splits cleanly: **46% lz4 decode + 45% FileWithChecksum
de-interleave + 8% UnionBatch**.

| stage | cold scan | warm scan |
|---|---|---|
| **read path** (`FileWithChecksum.ReadAt`): pread + de-interleave + CRC | **~51%** | 0 (cache-served) |
|  ├ pread syscall (kernel copy from page cache) | ~27% | — |
|  ├ de-interleave memmove (2 KB CRC-block → contiguous) | ~21% | — |
|  └ CRC verify | ~3% | — |
| **lz4 decode** (`decodeBlock`) | **~21%** | 0 |
| materialize (`UnionBatch` memmove) | ~4% | **~56%** |
| distance | ~3% | ~33% |

- **Cold is read-path + decode bound (~72%); materialize is ~4%.** Warm is the mirror — pure
  materialize + distance, with zero read/CRC/lz4 (decoded data served from cache).
- **lz4 runs ONLY on the cold (miss) path** — the cache stores *decoded* data, so a warm hit never
  decompresses. The UnionBatch materialization fix is therefore a *warm*-scan win (~4% of cold vs ~56% of warm).
- For *bulk cold scans* the FileWithChecksum de-interleave is **~21%** (vs ~1.85% in point search) — it
  scales with bytes pushed through the per-2 KB-CRC on-disk format.

### Expected Behavior

reduce/remove the number of memmove for de-interleave

[...]

PR #25272, opened 2026-06-30
FIX: FileService implements new DISK-V2 backend to replace DISK

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25040

 DISK-V2: a checksum-free local fileservice format
  
  Why
  
  LocalFS (backend DISK) frames every file as [crc32 4B][2044B content] blocks, so its on-disk layout differs from the S3 file service (including
  its disk-backed endpoint="disk" variant), which stores raw bytes. This asymmetry means a file written by LocalFS is not byte-identical to the
  same file written by S3FS-on-disk. This PR removes that CRC/block framing from the local path via a new backend, DISK-V2, whose on-disk format
  is byte-identical to S3FS-on-disk.

  What's added 

  New backend DISK-V2 (opt-in).
  - FileWithChecksum gains a noChecksum passthrough mode: ReadAt/WriteAt go straight to the underlying *os.File, Seek/size math use the physical
  size, and offset mapping is identity. CRC framing is skipped entirely. 
  - LocalFS gains NewLocalFS2; config.go registers backend = "DISK-V2" → NewLocalFS2.
  - The cluster default stays DISK (Go embedded defaults unchanged). DISK-V2 is selected explicitly — etc/launch/* set it for fresh local
  clusters. Old data dirs keep working on DISK with no migration.

  Backward-compatibility note. The two formats share no magic/header and are not interchangeable on the same data dir: pointing DISK-V2 at DISK
  data fails loudly (parse error on the leading CRC bytes), and DISK at DISK-V2 data fails with checksum not match. The format is fully
  encapsulated in LocalFS/FileWithChecksum; all startup/runtime paths read through the fileservice and are format-agnostic, so a cluster can be
  configured as either backend and works with its own data.
  
  Offline tools are format-aware and require an explicit choice.
  - objectio.NewOfflineFS takes a kind (local/s3/local2); OfflineKindStrict requires exactly one of --local / --s3 / --local2 (no silent default,
  since a wrong guess gives a cryptic error).
  - Applies to mo-object-tool (object/view/info, ckp) and the TAE inspect tool (object stat/get, storage ckp stat/list).
  - --local2 (raw LocalFS2) and --s3 (raw S3-on-disk) both read DISK-V2 data; --local reads the legacy CRC format.

  Backup follows the cluster's format (single source of truth). A filesystem backup's archive is created in the same on-disk format as the
  cluster's data service (cfg.SharedFs, configured by the .toml) via fileservice.LocalBackendOf, so restore reads it in the cluster's format.
  DISK-V2 cluster → raw archive; legacy DISK cluster → CRC archive. S3 backup targets are unaffected (already raw).

  Misc. tmpNewFileservice (the TAE FillDefaults nil-fileservice fallback, only hit by tests/standalone instantiation — production always injects
  the configured fileservices) now defaults to DISK-V2.

  Performance

  Isolated fileservice benchmark (M2 Max, 24 MiB): reads ~1.9× faster (5.4 → 10.4 GB/s), writes ~14× faster (0.53 → 7.5 GB/s, 12,313 → 0
  allocs/op).
  
  Testing

  - Unit: noChecksum read/write/seek/append/random-write, NewLocalFS2 conformance + cross-format incompatibility, backend dispatch,
  OfflineKindStrict, LocalBackendOf/GetForBackup, checkpointtool.WithKind, and join-spill on a DISK-V2 LOCAL fileservice.
  - Manual: fresh single-node and multi-CN clusters bootstrap, write/read, and run ivfflat on DISK-V2; old DISK data still reads under backend = 
  "DISK"; and the CLI was verified on a real DISK-V2 object — --local2/--s3 read it, --local fails with checksum not match, no-flag errors
  clearly.
  
  Commits

  - feat(fileservice): add opt-in DISK-V2 checksum-free backend
  - feat(tools): offline tools select on-disk format via --local / --s3 / --local2
  - feat(tools): offline tools require an explicit --local/--s3/--local2
  - fix(backup): write the backup archive in the cluster's on-disk format
  - fix(objectio): tmpNewFileservice fallback defaults to DISK-V2
  - test(disk-v2): cover join spill, checkpoint tool, and raw append


XuPeng-SH (changes requested):
I think this needs changes before merging. The implementation direction is reasonable, but the compatibility boundary is not strong enough for a storage-format change.

Main concern: the existing launch profiles are changed to `backend = "DISK-V2"` while still pointing at the same `./mo-data/local` and `./mo-data/shared` paths. `DISK` and `DISK-V2` have no magic/header, and the new test explicitly documents that a raw `DISK-V2` reader over legacy `DISK` bytes can return wrong raw bytes at the fileservice layer. That means a normal developer/operator flow of upgrading the binary and reusing the same launch command/data dir can select the new format against old data without a startup-time format guard. Please make the boundary first-class: either keep existing `etc/launch*` profiles on `DISK` and add separate fresh-deploy `DISK-V2` profiles/data dirs, or add an on-disk format marker/fail-fast validation before `DISK-V2` opens a non-empty dir. A comment saying “fresh deployment” is not enough protection for a format-incompatible backend.

The compatibility mechanism in this PR is config-level dual backend plus `etc/v1`, not MatrixOne's formal `pkg/bootstrap/versions` upgrade framework. That can be acceptable because this is file layout rather than catalog metadata, but then the config/tooling compatibility path has to be complete and tested. Right now `cmd/mo-service/toml_fileservice_test.go` scans `etc/launch*` but not the new `etc/v1/**` tree, even though `etc/v1` is the proposed legacy compatibility path. Please include the `etc/v1` configs in the parse/backend-string coverage.

Also `git diff --check origin/main...HEAD` fails on the new `etc/v1` files due trailing whitespace / blank lines at EOF. Please fix that before merge.

Docs need to be updated as part of the same change. `etc/DEV_README.md` still describes these local launch methods as `DISK backend` and its data directory compatibility section does not mention `DISK` vs `DISK-V2` or how to run old data after upgrade. Since this PR changes shipped launch defaults, the migration/compatibility guidance should be explicit.

The targeted tests I ran locally pass with the proper CGo include/lib paths: `pkg/fileservice`, `cmd/mo-object-tool`, `cmd/mo-object-tool/ckp`, `cmd/mo-service`, `pkg/tools/objecttool/interactive`, `pkg/tools/checkpointtool`, and `pkg/sql/colexec/hashjoin`. The remaining blocker is the compatibility guard/process, not the basic raw-mode implementation.

aptend:
## Review — Multi-angle analysis (First Principles + Unhappy Path)

### Unhappy Path

#### 🔴 Startup-time format validation is missing

`NewLocalFS2` opens a non-empty directory without any check for whether the data inside is DISK or DISK-V2 format. A misconfigured `data-dir` pointing at legacy data will not fail fast — it will produce cryptic parse/checksum errors deep in the read path.

The subsequent commit separating data dirs (`local2`/`shared2` vs `local`/`shared`) reduces the probability of accidental misconfiguration but does not eliminate it — a user manually setting `data-dir` can still point at the wrong directory.

**Suggestion:** On opening a non-empty directory in `NewLocalFS2`, scan the first data file's leading bytes for CRC-framing signatures and fail with a clear error like `"directory contains legacy DISK data; use backend = DISK or migrate"`.

#### 🔴 `tmpNewFileservice` silently switches from DISK to DISK-V2

```go
Backend: "DISK-V2",  // was "DISK"
```

The PR description states "production always injects the configured fileservices", but this is an unverified assumption. If any production path (error-handling fallback, standalone mode, certain test harnesses) misses the injection, it will silently write DISK-V2 format into a directory expected to be DISK — with no warning or guard.

**Suggestion:** At minimum, add a `logutil.Warn` when `tmpNewFileservice` is invoked, or make it panic to surface missed injection paths.

#### 🟡 `OfflineKindStrict` error message does not distinguish failure modes

```go
if n != 1 {
    return "", moerr.NewInvalidInputNoCtx("specify exactly one of --local, --s3, --local2")
}
```

When `n == 0` (none specified) and `n > 1` (multiple specified), the same error is returned. The `n == 0` case is confusing — the user did nothing wrong but gets told to "specify exactly one".

**Suggestion:**
- `n == 0`: `"no format specified; use one of --local, --s3, --local2"`
- `n > 1`: `"conflicting formats specified; choose one of --local, --s3, --local2"`

#### 🟡 Missing test scenarios

| Missing test | Why it matters |
|---|---|
| Upgrade E2E: old DISK data → launch with `etc/v1` → read succeeds | Closes the validation loop for the entire migration strategy |
| Mixed directory: same dir contains both DISK and DISK-V2 files | Current behavior is undefined |
| Concurrent `WriteAt` to the same file at different offsets | `os.File.WriteAt` is thread-safe, but `FileWithChecksum` has an internal `contentOffset` field — needs a test to confirm passthrough mode does not touch it |

---

### Code Quality

#### 🟡 `LocalBackendOf` fallback lacks an explanatory comment

```go
func LocalBackendOf(fs FileService) string {
    if lfs, ok := fs.(*LocalFS); ok && !lfs.noChecksum {
        return diskFileServiceBackend
    }
    return diskV2FileServiceBackend  // MemoryFS, S3FS, ErrorFS all fall through here
}
```

All non-`*LocalFS` types return `DISK-V2`. This is intentional (they are all raw formats), but without a comment the next reader will be puzzled.

---

### Testing & Performance

#### 🟡 Benchmark methodology

The benchmark uses 24 MiB sequential read/write. Real workloads are typically small-block random reads (4 KB–64 KB) with mixed read/write and cache hit/miss patterns. Sequential benchmarks alone cannot fully support the performance claims.

**Suggestion:** Supplement with random small-block I/O benchmarks to give a more complete performance picture.

---

### Summary

| Priority | Issue |
|---|---|
| 🔴 | No startup-time format validation when `NewLocalFS2` opens a non-empty dir |
| 🔴 | `tmpNewFileservice` silently switches to DISK-V2 without guard |
| 🟡 | Missing upgrade E2E / mixed-dir / concurrent-write tests |
| 🟡 | `OfflineKindStrict` error message conflates `n==0` and `n>1` |
| 🟡 | Benchmark only covers sequential I/O |
| 🟢 | `LocalBackendOf` fallback needs a comment |

fengttt (approved):

aunjgr (changes requested):
Storage format change with no startup detection. Launch defaults switch to DISK-V2 on old data dirs — silent corruption. Either add a format marker with fail-fast on mismatch, or keep existing profiles on DISK and gate DISK-V2 behind fresh data dirs. etc/v1 backward-compat configs are the right idea but the default must stay safe.

cpegeric:
> Storage format change with no startup detection. Launch defaults switch to DISK-V2 on old data dirs — silent corruption. Either add a format marker with fail-fast on mismatch, or keep existing profiles on DISK and gate DISK-V2 behind fresh data dirs. etc/v1 backward-compat configs are the right idea but the default must stay safe.

We already have a fail fast check and new data-dir is in new directory.  Most importantly, we will completely remove DISK after a few month. DISK is buggy and should be removed eventually.

cpegeric:
> ## Review — Multi-angle analysis (First Principles + Unhappy Path)
> ### Unhappy Path
> #### 🔴 Startup-time format validation is missing
> `NewLocalFS2` opens a non-empty directory without any check for whether the data inside is DISK or DISK-V2 format. A misconfigured `data-dir` pointing at legacy data will not fail fast — it will produce cryptic parse/checksum errors deep in the read path.
> 
> The subsequent commit separating data dirs (`local2`/`shared2` vs `local`/`shared`) reduces the probability of accidental misconfiguration but does not eliminate it — a user manually setting `data-dir` can still point at the wrong directory.
> 
> **Suggestion:** On opening a non-empty directory in `NewLocalFS2`, scan the first data file's leading bytes for CRC-framing signatures and fail with a clear error like `"directory contains legacy DISK data; use backend = DISK or migrate"`.
> 
> #### 🔴 `tmpNewFileservice` silently switches from DISK to DISK-V2
> ```go
> Backend: "DISK-V2",  // was "DISK"
> ```
> 
> The PR description states "production always injects the configured fileservices", but this is an unverified assumption. If any production path (error-handling fallback, standalone mode, certain test harnesses) misses the injection, it will silently write DISK-V2 format into a directory expected to be DISK — with no warning or guard.
> 
> **Suggestion:** At minimum, add a `logutil.Warn` when `tmpNewFileservice` is invoked, or make it panic to surface missed injection paths.
> 
> #### 🟡 `OfflineKindStrict` error message does not distinguish failure modes
> ```go
> if n != 1 {
>     return "", moerr.NewInvalidInputNoCtx("specify exactly one of --local, --s3, --local2")
> }
> ```
> 
> When `n == 0` (none specified) and `n > 1` (multiple specified), the same error is returned. The `n == 0` case is confusing — the user did nothing wrong but gets told to "specify exactly one".
> 
> **Suggestion:**
> 
> * `n == 0`: `"no format specified; use one of --local, --s3, --local2"`
> * `n > 1`: `"conflicting formats specified; choose one of --local, --s3, --local2"`
> 
> #### 🟡 Missing test scenarios
> Missing test	Why it matters
> Upgrade E2E: old DISK data → launch with `etc/v1` → read succeeds	Closes the validation loop for the entire migration strategy
> Mixed directory: same dir contains both DISK and DISK-V2 files	Current behavior is undefined
> Concurrent `WriteAt` to the same file at different offsets	`os.File.WriteAt` is thread-safe, but `FileWithChecksum` has an internal `contentOffset` field — needs a test to confirm passthrough mode does not touch it
> ### Code Quality
> #### 🟡 `LocalBackendOf` fallback lacks an explanatory comment
> ```go
> func LocalBackendOf(fs FileService) string {
>     if lfs, ok := fs.(*LocalFS); ok && !lfs.noChecksum {
>         return diskFileServiceBackend
>     }
>     return diskV2FileServiceBackend  // MemoryFS, S3FS, ErrorFS all fall through here
> }
> ```
> 
> All non-`*LocalFS` types return `DISK-V2`. This is intentional (they are all raw formats), but without a comment the next reader will be puzzled.
> 
> ### Testing & Performance
> #### 🟡 Benchmark methodology
> The benchmark uses 24 MiB sequential read/write. Real workloads are typically small-block random reads (4 KB–64 KB) with mixed read/write and cache hit/miss patterns. Sequential benchmarks alone cannot fully support the performance claims.
> 
> **Suggestion:** Supplement with random small-block I/O benchmarks to give a more complete performance picture.
> 
> ### Summary
> Priority	Issue
> 🔴	No startup-time format validation when `NewLocalFS2` opens a non-empty dir
> 🔴	`tmpNewFileservice` silently switches to DISK-V2 without guard
> 🟡	Missing upgrade E2E / mixed-dir / concurrent-write tests
> 🟡	`OfflineKindStrict` error message conflates `n==0` and `n>1`
> 🟡	Benchmark only covers sequential I/O
> 🟢	`LocalBackendOf` fallback needs a comment

fixed.

XuPeng-SH (changes requested):
I still see two substantive compatibility blockers here.

1. Several shipped DISK-V2 launch configs still reuse the old DISK data directories.

The current head does have a startup-time guard in `NewLocalFS2`, but the config side is still inconsistent with the README’s upgrade-safety claim. A number of shipped DISK-V2 launch targets still point `LOCAL` / `SHARED` at the same old paths that legacy DISK deployments already use:
- `etc/launch-minio-local/{cn,log,tn}.toml` still use `./etc/launch-minio-local/mo-data`
- `etc/launch-tae-compose/config/{cn-0,cn-1,log,tn}.toml` still use `/var/lib/matrixone/data`
- `etc/docker-multi-cn-local-disk/generate-config.sh` still emits `SHARED=/mo-data/shared`, and `LOCAL` still falls back under the old default data-dir logic

That means "upgrade the binary and rerun the same launch command" is not actually safe for those shipped configs, even though `etc/DEV_README.md` now claims all `etc/launch*` configs have moved to fresh `local2/shared2` directories.

Suggested fix: either move every shipped DISK-V2 config to fresh directories (matching the README), or keep those configs on legacy `DISK`. The docs should only claim upgrade safety once the shipped configs actually satisfy it.

2. The standalone/embedded TAE default-open path still silently switches existing databases to DISK-V2.

`DB.Open()` still calls `opts.FillDefaults(dirname)` when `Fs` / `LocalFs` are missing, and that fallback still goes through `objectio.TmpNewFileservice(..., path.Join(dirname, "data"))` with backend `DISK-V2`. Because `DISK-V2` now rejects legacy DISK directories at open time, any standalone/embedded TAE caller that relied on the old default fallback can no longer reopen an existing database after upgrade.

The new warning in `tmpNewFileservice` is helpful, but it does not remove the compatibility break on that default open path.

Suggested fix: preserve legacy `DISK` for this fallback path, or make the fallback format explicit/configurable and default to compatibility when reopening existing standalone data. At minimum, `DB.Open()` should not silently change its default on-disk format for existing callers.

cpegeric:
Here's my answer above restated with the original names:

  1. Every config has LOCAL and SHARED — ✅ guaranteed (mo-service and embed)

  - Explicit in most .toml (e.g. etc/launch-minio-local/tn.toml:9-12 declares LOCAL).
  - Default-fill for omitted ones: setFileserviceDefaultValues() appends LOCAL→DISK, SHARED→DISK, ETL→DISK-ETL, TMP→DISK-TMP, and runs before
  createFileService().
  
  2. Missing LOCAL/SHARED does not error — it fills them with DISK

  The fileservice.Get[...] checks (config.go:364/370, embed:342) are a safety-net validator that only trips if fill somehow didn't run. Real
  guarantee: "always present, defaulting to DISK."

  3. Only tests call the fallback — ✅ confirmed

  - Direct callers of TmpNewFileservice / TmpNewSharedFileservice (5): all _test.go.
  - FillDefaults nil-Fs path (which calls tmpNewFileservice): the sole production db.Open chain is tnservice factory.go (builds Options{Fs, 
  LocalFs, TmpFs}) → NewTAEStorage → NewTAEHandle → openTAE → db.Open, so opt.Fs is always non-nil and the fallback is skipped. The only other
  NewTAEHandle caller is a test util.

4.  The minio/tae-compose/multi-cn configs are never upgrade-tested
  
  - Only etc/launch-tae-compose is touched by CI (e2e-compose), and that's a fresh BVT on a container-ephemeral /var/lib/matrixone/data (no
  volume mount) → DISK-V2 on an empty dir works.
  - launch-minio-local and docker-multi-cn-local-disk aren't referenced by any CI workflow at all — they're local-dev Makefile targets.

  So their "DISK-V2 at the old dir" (instead of a fresh local2/shared2) cannot hit the reject in CI, because no CI reopens pre-existing DISK data
  through them.

  
  Conclusion: tmpNewFileservice (and its TmpNewFileservice/TmpNewSharedFileservice wrappers) is 100% test-only, tests run on fresh dirs, so the  current DISK-V2 backend is already safe — no change strictly required.

XuPeng-SH (changes requested):
I still think this needs changes.

1. The launch-profile migration is still not compatibility-safe for all shipped/dev configs. `etc/DEV_README.md` now says all `etc/launch*` profiles use fresh `local2/shared2` dirs and that re-running the same launch command after upgrade is safe (`etc/DEV_README.md:373-390`). But several profiles still only flip the backend to `DISK-V2` while keeping the old legacy-DISK data path. For example:
   - `etc/launch-minio-local/cn.toml:9-12` keeps `data-dir = "./etc/launch-minio-local/mo-data"`
   - `etc/launch-tae-compose/config/cn-0.toml:8-11` keeps `data-dir = "/var/lib/matrixone/data"`

   Those are the same paths these profiles previously used with `DISK`, so re-running the existing profile after upgrade still repoints a raw `DISK-V2` backend at CRC-framed legacy bytes. That contradicts the new README safety claim and keeps the migration hazard alive for those shipped configs.

2. The standalone / embedded default still changes under callers that do not inject a fileservice. The PR body says the cluster default stays `DISK` and Go embedded defaults stay unchanged, but `tmpNewFileservice` now unconditionally falls back to `DISK-V2` (`pkg/objectio/fs.go:75-94`), and `Options.FillDefaults` still uses that fallback whenever `o.Fs` / `o.LocalFs` are nil (`pkg/vm/engine/tae/options/options.go:285-290`). So `db.Open(..., nil)` / other standalone TAE callers that previously reopened legacy local data through the implicit `DISK` path now switch to `DISK-V2` instead. The new CRC guard may make that fail loudly instead of silently misreading, but it is still a compatibility break in the default-open path.

I think those two compatibility gaps need to be closed before merge: either keep the unchanged profiles on legacy `DISK` (or move them to fresh dirs too), and either preserve the old standalone fallback backend or make the changed default explicit rather than silently flipping it.

PR #25273, opened 2026-06-30
refactor: make pipeline termination protocol explicit and cancellation-safe

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

issue #24971

This PR makes pipeline termination explicit and cancellation-safe.

- Unifies `WaitRegister` and `PipelineEdge` terminal state so typed terminal signals close `Done()` consistently.
- Adds edge terminal/reuse semantics for delivered, dropped, retried, and reset terminal events.
- Makes `pSpool.Abort()` release pending spool batches immediately while preserving batches already handed to receivers until `ReleaseCurrent`.
- Routes connector/dispatch data and cleanup paths through edge-aware helpers.
- Adds unhappy-path coverage for full receiver channels, terminal retry, spool abort cleanup, edge reuse, and nil process cleanup logging.

Testing:

- `go test -v -count=1 -timeout 120s ./pkg/vm/process ./pkg/container/pSpool`
- `go test -race -count=1 -timeout 120s ./pkg/vm/process ./pkg/container/pSpool`
- `go test -v -count=1 -timeout 120s ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch`
- `go build ./pkg/container/pSpool ./pkg/vm/process ./pkg/sql/compile`
- `go vet ./pkg/container/pSpool ./pkg/vm/process ./pkg/sql/compile`
- `go test -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" -v -count=1 -timeout 120s -run 'TestCompileBuildSideForBroadcastJoin(SkipsEmptyCN|GroupsDuplicateCN)$' ./pkg/sql/compile`


XuPeng-SH (commented):

1. Edge-level terminal state breaks the existing multi-sender `WaitRegister` contract.

`PipelineEdge` now treats End/Error/Abort as terminal for the whole edge: `markTerminal()` closes `done`, and later data sends are rejected once that terminal state is set. But the compile path still shares one `WaitRegister` across multiple upstream senders and relies on `NilBatchCnt > 1` (`newMergeScope`, `newMergeScopeByCN`, `constructDispatchLocalAndRemote`).

That means the first sender that sends a typed terminal signal can seal the shared edge for everyone else. Remaining senders on the same shared merge edge will then fail `SendData` / `TrySend`, even though the receiver still expects more terminal accounting and possibly more data.

Suggested fix: do not make terminal state global to the shared edge while shared `WaitRegister`s still exist. Either keep terminal semantics per sender, or defer this edge-global terminal protocol until each sender has its own dedicated edge.

2. Normal successful cleanup now aborts queued spool data before downstream has drained it.

On normal completion (`pipelineFailed=false`, `err=nil`), `Connector.Reset` / `Dispatch.Reset` send a typed terminal signal and then immediately call `sp.Abort()`. But `Abort()` marks the spool aborted and `ReceiveBatch()` will then return `ErrPipelineSpoolAborted` instead of delivering queued batches that were already published into the spool but not yet consumed.

So if the downstream merge still has buffered `GetFromSpool` signals when cleanup starts, already-produced rows are discarded during a successful shutdown.

Suggested fix: reserve `sp.Abort()` for failure/cancellation paths where dropping pending data is acceptable. On successful completion, keep the old drain-preserving path (or another mechanism that preserves queued spool data until receivers consume it).

XuPeng-SH:
Addressed the review comments in 551cb5343.

- Fixed shared WaitRegister/PipelineEdge terminal semantics: normal EventEnd is now counted against NilBatchCnt, so Done() only closes after all expected sender Ends are delivered; EventError/EventAbort remain fatal first-wins and close Done immediately.
- Fixed successful connector/dispatch cleanup: normal EventEnd no longer calls sp.Abort(), so queued GetFromSpool signals can still drain. Spool cache cleanup is deferred until the paired Merge cleanup has drained queued signals.
- Added regression coverage for shared-End/data interleaving, connector success drain, dispatch broadcast drain, fatal full-channel cleanup, and typed-terminal spool cleanup without a nil End message.

Verified locally:
- go test -v -count=1 -timeout 120s ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go test -race -v -count=1 -timeout 120s -run 'TestPipelineEdge|TestPipelineSignalReceiverSharedEdge|TestPipelineSpool|TestConnectorReset|TestDispatchReset' ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go vet ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go build ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch

XuPeng-SH (commented):
Re-reviewing the latest head: one of the earlier blockers is fixed, but there are still real problems.

What is fixed now:
- the earlier spool-abort-on-success problem looks resolved. The normal End path now preserves queued spool data and the new tests cover that directly.

What still remains:
1. shared-edge fatal terminal semantics are still mismatched
2. successful cleanup still treats End delivery as best-effort, so a full receiver channel can turn a normal success path into a merge-cleanup timeout

On shared edges (`NilBatchCnt > 1`), `EventEnd` is now counted per sender, but `Error`/`Abort` are still edge-global first-wins and only delivered once. The receiver still decrements the sender count one terminal at a time, so one fatal can leave the receiver waiting for remaining senders that can never report a terminal anymore.

Separately, `Connector.Reset` / `Dispatch.Reset` only log when `EventEnd` cannot be delivered within `PipelineSignalSendTimeout`; they do not retry indefinitely or downgrade to an abort/error path. That means a normal success cleanup can still leave `Merge` without any terminal signal and force timeout-based cleanup instead of a clean end.

Suggested fixes:
- make fatal-terminal accounting match shared-edge sender accounting (either deliver the latched fatal once per expected sender, or consume the whole remaining count on first fatal);
- make End delivery success mandatory, or explicitly fall back to an abort/error cleanup mode when End cannot be enqueued.

XuPeng-SH:
Addressed the latest review in 2b36b64be.

- Fixed shared-edge fatal accounting: Error/Abort now latch the first fatal signal but deliver it for the remaining sender count on shared edges, so the receiver can consume all expected terminal accounting instead of waiting for missing senders.
- Fixed normal End delivery failure handling: Connector/Dispatch now treat End delivery failure as an explicit abort fallback. Normal End success still preserves queued spool data; if End cannot be enqueued within the cleanup timeout, we send fallback Abort to failed receivers and abort the spool instead of leaving Merge to timeout.
- Added regression coverage for shared fatal completion, partial fatal retry after a full channel, Connector End fallback, and Dispatch mixed receiver End fallback.

Verified locally:
- go test -v -count=1 -timeout 120s ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go test -race -v -count=1 -timeout 120s -run 'TestPipelineEdge|TestPipelineSignalReceiverShared|TestPipelineSpool|TestConnectorReset|TestDispatchReset' ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go vet ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
- go build ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch

XuPeng-SH (commented):
Re-reviewing the latest head: the earlier shared-edge fatal blocker looks fixed now, but I still don’t think this is ready to merge.

What is now fixed:
- shared-edge fatal terminal semantics (`NilBatchCnt > 1`) appear to be handled correctly on the latest head.

What still remains:
1. a remote-input cleanup path still uses legacy best-effort nil terminals instead of the new typed terminal protocol;
2. the Abort fallback can extend cleanup far beyond the nominal 30s budget because each failed receiver gets its own fresh timeout window.

Details:
- `pkg/sql/compile/scope.go:816-834` (`Scope.sendNotifyMessage -> closeWithError`) still sends `NewPipelineSignalToDirectly(nil, err, ...)`, which is a legacy `EventData` nil-batch, not a typed terminal. If that send times out, it only logs and exits, leaving merge-side cleanup to hit its timeout path with no terminal delivered.
- In `dispatch` / `connector`, the End-send budget is bounded, but the fallback Abort sends still use fresh per-receiver timeout windows rather than a shared deadline. So cleanup can stretch to roughly `(1 + failed_receivers) * 30s` instead of staying inside one 30s cleanup budget.

Suggested fixes:
1. make the remote receive cleanup path use typed terminals too, and mirror the local cleanup fallback behavior;
2. use one shared deadline/context for the Abort fallback after End-send failure, rather than creating a new 30s timeout per failed receiver.

XuPeng-SH:
CI failure follow-up:

1. Deterministic SCA failure fixed in `67823749b`.
   - Root cause: this PR introduced production `errors.New` usages in:
     - `pkg/container/pSpool/sender.go`
     - `pkg/vm/process/process_spoolr.go`
   - Fix: replaced both sentinel errors with `moerr.NewInternalErrorNoCtx(...)`.
   - Note: the `context.WithTimeout` line in the CI log is emitted by `make err-check` after entering the error branch; the hard failure here was the `errors.New` usage.

2. Local verification after the fix:
   ```bash
   make err-check
   go test -v -count=1 -timeout 120s ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
   go vet ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
   go build ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
   go test -v -count=1 -timeout 120s -run TestSpeedupAbortAllTxn ./pkg/tests/issues
   ```

   All passed locally.

3. About `TestSpeedupAbortAllTxn` CI failure:
   - The CI failure stack is a logservice HAKeeper bootstrap timeout/panic.
   - I reran the exact case locally and it passed.
   - Based on the stack and the touched code path in this PR, I do not see evidence that this is caused by the pipeline cleanup refactor. The new CI run after `67823749b` should tell whether it was transient.

XuPeng-SH:
Follow-up for `TestSpeedupAbortAllTxn` flake:

Root cause found and fixed in `147ab3511`.

The CI panic was not a normal assertion failure from `TestSpeedupAbortAllTxn`; it was a background HAKeeper ticker panic:

```text
pkg/logservice/store_hakeeper_check.go:316
checkBootstrap -> setBootstrapState(true) -> propose timeout -> panic
```

First-principles issue:
- `checkBootstrap` had already observed that bootstrap completed.
- It then tried to persist `HAKeeperRunning` through a Raft proposal with the fixed `hakeeperDefaultTimeout` of 2s.
- Under CI startup pressure / leader election / slow WAL, that proposal can transiently timeout.
- The old code treated this recoverable proposal failure as process-fatal (`panic(err)`), killing the whole UT binary.

Fix:
- `checkBootstrap` now treats `setBootstrapState` failure as retryable: log the error and return, so the next HAKeeper check retries the same idempotent state transition.
- The bootstrap cycle counter is not consumed when the final state write fails.
- When bootstrap cycles are exhausted and `HAKeeperBootstrapFailed` is successfully persisted, the function returns immediately instead of continuing with stale state in the same tick.
- Added a unit test that injects a `setBootstrapState` failure and verifies no panic / no cycle consumption.

I also hardened `TestSpeedupAbortAllTxn` itself:
- restore `taskservice.DebugCtlTaskFramework(false)` with `defer`, avoiding global test pollution;
- make the reconnect handler buffered/non-blocking;
- put channel waits and `waitLogtailResume` behind a context, so failure becomes bounded and diagnosable instead of an implicit hang.

Local verification:

```bash
make err-check
go vet ./pkg/logservice ./pkg/tests/issues
go build ./pkg/logservice
go test -c -o /tmp/matrixone-issues.test ./pkg/tests/issues
go test -v -count=1 -timeout 60s -run TestCheckBootstrapRetriesSetBootstrapStateFailure ./pkg/logservice
go test -v -count=1 -timeout 120s -run 'Test.*Bootstrap|TestCheckBootstrapRetriesSetBootstrapStateFailure|TestHandleBootstrapFailure' ./pkg/logservice
go test -count=1 -timeout 180s ./pkg/logservice
go test -v -count=1 -timeout 180s -run TestSpeedupAbortAllTxn ./pkg/tests/issues
go test -count=3 -timeout 240s -run TestSpeedupAbortAllTxn ./pkg/tests/issues
```

All passed locally. Note: `go build ./pkg/tests/issues` is not meaningful because that directory only has `_test.go`; I used `go test -c` to compile that test package instead.

XuPeng-SH (commented):
Re-reviewing the latest head: I still would not merge this yet, but the blocker set is narrower now.

What still looks wrong:
1. the remote-input cleanup path in `pkg/sql/compile/scope.go` is still sending legacy nil terminals (`NewPipelineSignalToDirectly(nil, err, ...)`) instead of the new typed `End/Error/Abort` protocol;
2. the abort-fallback timeout-budget concern is weaker now, but still not really disproven for the real production budget / multi-receiver case.

The first point is still the main blocker. On the latest head, shared receiver edges still exist with `NilBatchCnt > 1`, and the receiver treats legacy nil as a one-sender decrement. So a remote cleanup path that bypasses typed terminals can still leave the shared receiver waiting for the rest of the sender count until timeout instead of terminating the edge immediately.

Suggested fix:
- in `Scope.sendNotifyMessage.closeWithError`, use typed cleanup terminals (`NewEndSignal` / `NewErrorSignal` / `NewAbortSignal`) instead of `NewPipelineSignalToDirectly(nil, err, ...)`;
- then add a targeted remote/multi-CN cleanup regression around that path.

The second point (budget inflation) now looks more like a bounded-latency concern than a clear correctness bug, so I would treat it as secondary compared with the protocol mismatch above.

XuPeng-SH:
Addressed the latest remote cleanup review in `c86f58fc4`: `Scope.sendNotifyMessage` now sends typed cleanup terminals instead of legacy nil terminal batches; normal End delivery failure falls back to typed Abort. Added compile-layer regressions for shared receiver typed Error and per-sender typed End. Local verification passed: targeted remote cleanup tests, related remote forwarding tests, full `go test ./pkg/sql/compile`, `go build ./pkg/sql/compile`, `go vet ./pkg/sql/compile`, and `make err-check`.

XuPeng-SH:
Follow-up from the full module review in `2b157b66f`.

Found and fixed one additional protocol edge case:
- `BuildCleanupSignal(true, nil)` previously produced `EventError(nil)`, which let receivers observe an error terminal with a nil cause. That is semantically unsafe because `PipelineSignalReceiver.GetNextBatch` returns the terminal error directly.
- Added `ErrPipelineTerminalWithoutCause`; `EventError`/`EventAbort` now always carry a non-nil cause. Connector/dispatch use the normalized terminal cause for logs and remote error notification.
- Hardened `PipelineEdge.SendData*` for nil context to match the nil-safe behavior of the other send helpers.

Extra regression coverage added:
- process signal constructors and cleanup signal normalization;
- receiver returning the sentinel error for failed cleanup with no cause;
- connector Reset sending typed failure with cause;
- dispatch remote cleanup notifying a non-nil cause;
- nil-context data send does not panic.

Local verification after this commit:
```bash
go test -count=1 -timeout 180s ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
go test -count=1 -timeout 180s ./pkg/sql/compile
go test -v -count=1 -timeout 120s -run 'TestCheckBootstrapRetriesSetBootstrapStateFailure|TestHandleBootstrapFailure' ./pkg/logservice
go test -v -count=1 -timeout 180s -run TestSpeedupAbortAllTxn ./pkg/tests/issues
go vet ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
go vet ./pkg/sql/compile ./pkg/logservice ./pkg/tests/issues
make err-check
git diff --check
```

All passed locally. PR checks restarted for the new head commit.

XuPeng-SH:
Follow-up from the second full module review in `6685cba6b`.

Additional issue found and fixed:
- Dispatch End-delivery fallback used a fresh `PipelineSignalSendTimeout` for each failed local receiver. In a multi-receiver cleanup path this could stretch cleanup to roughly `failedReceivers * timeout`.
- `sendAbortSignalsToFailedLocalRegs` now uses one shared cleanup context/deadline for all failed receivers.
- Added `TestDispatchFallbackAbortUsesSharedTimeoutBudget` to lock the bound and verify every failed edge is still marked terminal with `ErrPipelineEndSignalDeliveryFailed`.

Also tightened deferred spool cleanup comments in pSpool/connector/dispatch: the safety condition is that the paired Merge cleanup loop has returned. On the normal path it drains queued `GetFromSpool`; on timeout it releases current references and exits, so no receiver goroutine can later read pending signals.

CI note:
- The current arm64 SCA failure at previous head `2b157b66f` is not an err-check/code failure. The job failed during `make config` while downloading modules from `proxy.golang.org` with multiple `unexpected EOF` errors.

Local verification after this commit:
```bash
make err-check
git diff --check
CGO_CFLAGS="..." CGO_LDFLAGS="..." go test -v -count=1 -timeout 180s ./pkg/vm/process ./pkg/container/pSpool
CGO_CFLAGS="..." CGO_LDFLAGS="..." go test -v -count=1 -timeout 180s ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." go test -race -v -count=1 -timeout 240s -run 'TestPipelineEdge|TestPipelineSignalReceiver|TestPipelineSpool|TestDispatchFallbackAbortUsesSharedTimeoutBudget|TestDispatchResetFallsBackToAbortWhenEndSignalCannotBeDelivered|TestDispatchResetAbortsSpoolWhenSomeLocalRegIsFull|TestDispatchResetEndPreservesQueuedBroadcastBatchUntilDeferredCleanup' ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go test -ldflags="..." -v -count=1 -timeout 180s -run 'TestRemoteNotifyCleanupUsesTypedErrorForSharedReceiver|TestRemoteNotifyCleanupUsesTypedEndForSingleSender|TestReceiveMessageFromCnServerIfConnector_ReturnsOnBlockedReceiverCancel|TestReceiveMsgAndForward_ReturnsOnBlockedReceiverCancel|TestReceiveMessageFromCnServerIfDispatch_PreservesCleanupOnOriginalRoot' ./pkg/sql/compile
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go test -ldflags="..." -v -count=1 -timeout 180s -run 'TestCheckBootstrapRetriesSetBootstrapStateFailure|TestHandleBootstrapFailure' ./pkg/logservice
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go test -ldflags="..." -v -count=1 -timeout 240s -run TestSpeedupAbortAllTxn ./pkg/tests/issues
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go vet -ldflags="..." ./pkg/sql/compile ./pkg/logservice ./pkg/tests/issues
CGO_CFLAGS="..." CGO_LDFLAGS="..." go vet ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
```

XuPeng-SH:
Follow-up optimization in `c7610cca5`.

I tightened the cleanup send budget one step further:
- Connector, Dispatch local cleanup, and compile remote notify now use one shared `PipelineSignalSendTimeout` context for the initial End/Error send and any fallback Abort.
- This removes the remaining `(End timeout + fallback Abort timeout)` behavior. A cleanup sender now gets one terminal-send budget, not two serial budgets.
- Fallback Abort still marks the edge terminal even when the shared context has already expired, so send-side terminal state remains observable through `Done()`/`Err()`.

Added regressions:
- `TestConnectorResetUsesSharedTerminalSendBudget`
- `TestDispatchResetUsesSharedTerminalSendBudget`
- `TestRemoteNotifyCleanupUsesSharedTerminalSendBudget`

Local verification:
```bash
CGO_CFLAGS="..." CGO_LDFLAGS="..." go test -v -count=1 -timeout 180s ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go test -ldflags="..." -count=1 -timeout 180s ./pkg/sql/compile
CGO_CFLAGS="..." CGO_LDFLAGS="..." go test -race -v -count=1 -timeout 240s -run 'TestConnectorResetUsesSharedTerminalSendBudget|TestConnectorResetFallsBackToAbortWhenEndSignalCannotBeDelivered|TestDispatchResetUsesSharedTerminalSendBudget|TestDispatchResetFallsBackToAbortWhenEndSignalCannotBeDelivered|TestDispatchResetAbortsSpoolWhenSomeLocalRegIsFull' ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." go vet ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go vet -ldflags="..." ./pkg/sql/compile
CGO_CFLAGS="..." CGO_LDFLAGS="..." go build ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="..." CGO_LDFLAGS="..." DYLD_LIBRARY_PATH="..." go build -ldflags="..." ./pkg/sql/compile
make err-check
git diff --check
```

XuPeng-SH:
Follow-up closed-loop review update

New issue found and fixed in `89321d834`:
- Remote-run data forwarding still used the raw `chan PipelineSignal` path (`receiveMessageFromCnServerIfConnector` / `receiveMsgAndForward`). That bypassed `PipelineEdge.Done()/Abort` state, unlike local connector/dispatch data sends.
- Fixed by routing remote forwarded batches through `PipelineEdge.SendDataDirect`. If the target receiver edge is already terminal, the current batch is cleaned and the forward loop exits normally; if the sender context is canceled, the existing QueryInterrupted behavior is preserved.
- Added `TestReceiveMsgAndForward_ReturnsOnReceiverTerminal` and updated blocked-forward tests to use real `PipelineEdge` instances.

Test robustness improvement:
- Relaxed the shared cleanup deadline regression tests from `100ms/<180ms` to `200ms/<300ms`. This still catches the old serial two-deadline behavior (~400ms) but reduces CI timing flake risk.

Local verification passed:
- `go test -count=1 ./pkg/sql/compile` with MO CGo flags
- `go test -count=1 ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch` with MO CGo flags
- focused remote-forward / cleanup tests, including race
- focused connector/dispatch cleanup tests, including race
- `go test -count=1 ./pkg/vm/process ./pkg/container/pSpool` focused pipeline tests
- `go vet` for compile, connector, dispatch, process, pSpool
- `git diff --check`
- `make err-check`

XuPeng-SH:
Additional first-principles unhappy-path coverage in `278aa2f2b`.

This commit is test-only and closes the remaining protocol boundary cases from the Q1/Q2/Q3 review:
- Q2 / hang: `PipelineEdge` data sends blocked on a full channel now have explicit coverage showing Abort releases the sender.
- Q2 / protocol closure: data sends after normal End are rejected.
- Q1 / cleanup: connector Reset with nil Reg + live spool aborts and releases spool memory without panic.
- Q1/Q2 / cleanup: dispatch Reset with a nil local receiver aborts the spool, still notifies healthy local receivers, and does not panic.
- Q1/Q2 / remote forwarding: remote connector/forward paths return when the receiver edge is already terminal, and nil receiver returns a concrete error after cleaning the batch.

Local verification passed after the commit:
- focused new tests for process, connector, dispatch, compile
- `go test -count=1 ./pkg/vm/process ./pkg/container/pSpool`
- `go test -count=1 ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch`
- `go test -count=1 ./pkg/sql/compile`
- focused `-race` coverage for the new blocked/terminal paths
- `go vet` and `go build` for touched modules
- `git diff --check`
- `make err-check`

XuPeng-SH (commented):
Re-reviewing the latest head: the earlier remote-terminal and shared-timeout-budget concerns look addressed now, but I still see one remaining blocker.

`CleanupDeferredSpool()` can still run too early on the normal End path.

Both `cleanupInOrder()` and `cleanupSenderReceiverPipeline()` invoke deferred spool cleanup from the sender pipeline’s own teardown, but that does not guarantee every downstream local receiver using the same spool has finished draining and released its last consumed slot yet.

At the spool level, `ForceCleanupAfterTerminalSignal()` consumes the one-shot `cleanupOnce` and frees the currently cached buffers. Later `ReleaseCurrent()` calls on non-aborted receivers still put newly released batch buffers back into the spool cache (`CacheBatch(...)`), but there is then no second cleanup pass left to free those newly returned buffers.

So on successful runs with a fast sender and slower downstream receiver(s), this can leak spool-owned memory after the deferred cleanup fires too early.

Suggested fix:
1. delay deferred spool cleanup until downstream receiver drain is actually confirmed, or
2. after deferred cleanup has fired, make late `ReleaseCurrent()` free buffers directly instead of re-caching them.

The earlier shared-edge fatal / remote cleanup blockers look much better on this head; this deferred-cleanup lifecycle is the remaining issue I would fix before merge.

XuPeng-SH:
Follow-up for the latest deferred-spool-cleanup review in `9a46e4511`.

Root cause addressed:
- `ForceCleanupAfterTerminalSignal` is a one-shot cache-wide cleanup. If a slower typed-terminal receiver released its current batch after that cleanup, `ReleaseCurrent` previously returned that batch memory to the spool cache, but no second cleanup pass would run. That could leave spool-owned memory cached after normal End cleanup.

Fix:
- `PipelineSpool` now records when the final cache-wide cleanup has run.
- Late `ReleaseCurrent` after terminal cleanup frees the current slot directly via `cleanSlotLocked` instead of re-caching it.
- `spoolBuffer.clean` clears freed cache slices so the final cleanup leaves no stale references to freed buffers.
- Added `TestPipelineSpoolLateReleaseAfterTerminalCleanupFreesDirectly` to reproduce the exact late-release lifecycle and assert the mpool returns to zero.

Local verification passed:
```bash
CGO_CFLAGS="-I/Users/xupeng/gitlab/matrixone/cgo -I/Users/xupeng/gitlab/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/xupeng/gitlab/matrixone/thirdparties/install/lib -lusearch_c" go test -v -count=1 -timeout 120s ./pkg/container/pSpool
CGO_CFLAGS="-I/Users/xupeng/gitlab/matrixone/cgo -I/Users/xupeng/gitlab/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/xupeng/gitlab/matrixone/thirdparties/install/lib -lusearch_c" go test -v -count=1 -timeout 180s ./pkg/vm/process ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="-I/Users/xupeng/gitlab/matrixone/cgo -I/Users/xupeng/gitlab/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/xupeng/gitlab/matrixone/thirdparties/install/lib -lusearch_c" go test -race -v -count=1 -timeout 240s -run 'TestPipelineSpoolLateReleaseAfterTerminalCleanupFreesDirectly|TestPipelineSpoolForceCleanupAfterTerminalSignalDoesNotNeedNilEndMessage|TestPipelineSpoolAbortWaitsForAllCurrentBroadcastReceivers|TestConnectorResetEndPreservesQueuedSpoolBatchUntilDeferredCleanup|TestDispatchResetEndPreservesQueuedBroadcastBatchUntilDeferredCleanup' ./pkg/container/pSpool ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="-I/Users/xupeng/gitlab/matrixone/cgo -I/Users/xupeng/gitlab/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/xupeng/gitlab/matrixone/thirdparties/install/lib -lusearch_c" go vet ./pkg/container/pSpool ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
CGO_CFLAGS="-I/Users/xupeng/gitlab/matrixone/cgo -I/Users/xupeng/gitlab/matrixone/thirdparties/install/include" CGO_LDFLAGS="-L/Users/xupeng/gitlab/matrixone/thirdparties/install/lib -lusearch_c" go build ./pkg/container/pSpool ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch
make err-check
git diff --check
```

Note: running the pSpool test without MO CGo flags fails locally at compile time with `xxhash.h` missing, so the verified commands use the repository CGo include/link flags.

XuPeng-SH (commented):
Re-reviewed latest head `9a46e4511`.

I do not see a remaining blocker in the current version.

What I re-checked:
- `WaitRegister` is now a direct `PipelineEdge` alias, so terminal state is no longer split across two objects.
- Shared-edge End/Error/Abort accounting now matches `NilBatchCnt > 1`: normal End waits for all expected senders, while fatal terminals latch first cause and deliver the remaining sender count.
- Connector/Dispatch normal End preserves queued spool data and defers spool cleanup until the paired Merge cleanup returns; End delivery failure falls back to typed Abort and aborts the spool.
- Remote notify cleanup and remote batch forwarding no longer bypass typed terminal / edge state.
- The latest pSpool fix closes the late `ReleaseCurrent` case after terminal cleanup by freeing directly once the final cache-wide cleanup pass has run.
- The HAKeeper bootstrap follow-up avoids panicking a background ticker on transient `setBootstrapState` failure; it now logs and retries on a later tick.

Local verification:
- `go test ./pkg/vm/process ./pkg/container/pSpool ./pkg/vm/pipeline -count=1 -timeout 120s`
- `go test ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch -count=1 -timeout 120s`
- targeted `go test ./pkg/sql/compile` for remote forwarding / typed cleanup / broadcast-build tests
- targeted `go test ./pkg/logservice -run ^TestCheckBootstrapRetriesSetBootstrapStateFailure$`
- `go build` and `go vet` for pSpool/process/pipeline/connector/dispatch/compile/logservice

CI is green. The branch is behind `main`, but I do not see a code-level reason to request changes.

Issue #25274, opened 2026-06-30
[Bug]: logservice client connects to unreachable replicas with 30s timeout, blocking DN commit pipeline


matrix-meow:
## Is there an existing issue for the same bug?
- [x] I have checked the existing issues.

## Branch Name
main

## Commit ID
2add0b9fc

## Other Environment Information
- OS type: Linux
- Others: freetier-01 cluster (cn-dev)

## Actual Behavior

When the log service cluster topology includes unreachable replicas (pods deleted, DNS unresolvable), the DN logservice client attempts to connect to **every** replica returned by `GetShardInfo`, including dead ones. Each failed connection attempt takes **~30 seconds** (`defaultConnectTimeout` in `pkg/tnservice/cfg.go:48`).

In the observed incident, only `default-log-0` existed, but `GetShardInfo` returned addresses for `default-log-1` and `default-log-2`. The client hit:
- DNS failure: `lookup default-log-2...no such host`
- TCP timeout: ~30s per attempt

Combined with `NewClient` retry logic (1s sleep between attempts), a single `GetOnFly` call took **30-61 seconds**, holding the `clientPool` mutex and blocking the entire DN commit pipeline.

### Evidence

**Wal-Get-Client slow events (17 occurrences over 2 hours):**
```
07:02:32  duration=30.885s
07:15:58  duration=1m1.787s  (2 × 30s timeout + 2 × 1s sleep)
07:23:38  duration=1m1.842s
```

**Each event caused `5-PreparedWait` p99 spike:**
```
07:15:58 Wal-Get-Client=61s → 07:19 PreparedWait p99=50.5s
07:18:43 Wal-Get-Client=61s → 08:19 PreparedWait p99=46.5s
```

**SLOW-LOG confirms blocking in WAL wait phase:**
```
end-prepare-duration=21.9s  (all other steps <1ms)
```

## Expected Behavior

1. Per-replica connect timeout of 3-5s instead of 30s. DNS failures should fail instantly.
2. Fast skip: if a replica is unreachable, immediately try the next one.
3. Leader-first: try the leader address from `GetShardInfo` first; only fall back to other replicas if the leader fails.

## Steps to Reproduce

1. Deploy cluster with 3 log replicas in HAKeeper topology, but only 1 pod running.
2. Observe DN logs: `WAL-Replay failed to create log service client` with `timeout`/`no such host`.
3. Observe `Wal-Get-Client` logs: 30-61s delays every 2-5 minutes.
[...]

Issue #25277, opened 2026-06-30
[Bug]: DN logservice client ConnectTimeout default 30s is too long for per-replica connection attempts


matrix-meow:
## Is there an existing issue for the same bug?
- [x] I have checked the existing issues.

## Branch Name
main

## Commit ID
2add0b9fc

## Other Environment Information
- OS type: Linux
- Others: Related to #25274, #25275, #25276

## Actual Behavior

The DN's logservice client factory uses a 30-second context timeout for each connection attempt:

```go
// pkg/tnservice/cfg.go:48
defaultConnectTimeout = time.Second * 30

// pkg/tnservice/factory.go:104
ctx, cancel := context.WithTimeoutCause(
    context.Background(),
    s.cfg.LogService.ConnectTimeout.Duration,  // 30s default
    moerr.CauseNewLogServiceClient,
)
```

When `connectToLogService` tries to connect to an unreachable replica (#25274), each attempt consumes the full 30s. Combined with `NewClient` retry logic (1s sleep between attempts), two failed attempts take 61s. This amplifies the impact of `GetOnFly` holding the pool mutex (#25275).

### Observed durations
```
30.9s = 1 × TCP timeout (~30s) + 1 × sleep (1s)
61.8s = 2 × TCP timeout (~30s) + 2 × sleep (1s)
```

## Expected Behavior

1. Reduce `defaultConnectTimeout` to 5-10s. Logservice connections should either succeed quickly or fail fast.
2. Alternatively, use a per-replica timeout in `connectToLogService` that is shorter than the overall context timeout:

```go
for _, addr := range addresses {
    perAddrCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
    cc, err := getRPCClient(perAddrCtx, ...)
    cancel()
    if err != nil { continue }
    ...
}
```

## Steps to Reproduce

Same as #25274.

### Configuration comparison
| Timeout | Current | Suggested |
|---|---|---|
| DN `ConnectTimeout` | 30s | 5-10s |
| morpc backend connect | 5s | OK |
| logservice `DefaultOneTryTimeout` | 1min | Consider reducing |

- `pkg/tnservice/cfg.go:48` — `defaultConnectTimeout = time.Second * 30`
- `pkg/tnservice/factory.go:104` — context creation
- `pkg/logservice/client.go:484-530` — `connectToLogService` (no per-replica timeout)
- Related: #25274, #25275, #25276

jiangxinmeng1:
[...]

PR #25289, opened 2026-07-01
Fix logservice client fallback from slow replica addresses

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25274 #25277

This PR improves logservice client creation when candidate replica addresses include stale or unreachable endpoints.

  Previously, connectToLogService tried replica addresses sequentially. If one address accepted a TCP connection but did not respond, client creation could wait until the outer context deadline before trying later
  candidates.

  Changes:
  - Keep the existing shuffled address order.
  - Start connection attempts with staggered fallback instead of strict sequential dialing.
  - Apply a per-address attempt timeout.
  - Return the first successfully connected client.
  - Cancel remaining attempts and close losing clients after a winner is selected.
  - Add a regression test covering a slow/unresponsive address followed by a reachable address.

  This reduces the impact of stale or slow replica addresses returned by discovery/gossip without changing HAKeeper or gossip behavior.


LeftHandCold (commented):
Found one cleanup issue in the new concurrent connection path:

`connectToLogServiceAddresses` returns the first successful client without closing any other successful clients that may already have been sent into the buffered `results` channel. `wg.Wait()` only waits for the attempt goroutines to exit; it does not drain `results`, and those goroutines return immediately after `results <- result{client: c}`. If two reachable addresses complete around the same time, the losing `*client` becomes unreachable and its underlying RPC client stays open.

Please drain the remaining active results after `cancel()` and close any non-winning clients, or make the goroutines close their client unless they win via a single winner-selection path.

volgariver6 (approved):

jiangxinmeng1:
@Mergifyio queue

jiangxinmeng1:
https://github.com/Mergifyio queue

Issue #25275, opened 2026-06-30
[Bug]: clientPool.GetOnFly() unnecessarily holds pool mutex during NewClient() network I/O


matrix-meow:
## Is there an existing issue for the same bug?
- [x] I have checked the existing issues.

## Branch Name
main

## Commit ID
2add0b9fc

## Other Environment Information
- OS type: Linux
- Others: Related to #25274

## Actual Behavior

`clientPool.GetOnFly()` holds `c.cond.L` (pool mutex) during the entire `NewClient()` call, which performs network I/O that can take 30-61 seconds. The mutex only needs to protect the `c.closed` check, but `NewClient()` does not access any pool state — it only reads `c.cfg` (immutable config).

```go
// pkg/vm/engine/tae/logstore/driver/logservicedriver/clientpool.go:319
func (c *clientPool) GetOnFly() (*wrappedClient, error) {
    c.cond.L.Lock()
    defer c.cond.L.Unlock()
    if c.closed {
        return nil, ErrClientPoolClosed
    }
    return NewClient(c.cfg.ClientFactory, ...)  // ← network I/O under lock!
}
```

During this time, `Get()`, `Put()`, and `GetWithWriteToken()` all block on `c.cond.L`, stalling the entire LogServiceDriver commit pipeline.

### Evidence

Goroutine profile during spike:
```
[5] GetWithWriteToken → Get → Mutex.Lock          ← blocked on pool mutex
[4] GetOnFly → NewClient → connectToLogService    ← holding pool mutex during network I/O
[274] commit1PC → WaitGroup.Wait                  ← pipeline backed up
```

17 `Wal-Get-Client` events over 2 hours (30-61s each), each causing `5-PreparedWait` p99 spikes to 30-50s.

## Expected Behavior

`GetOnFly()` should release the mutex immediately after checking `c.closed`:

```go
func (c *clientPool) GetOnFly() (*wrappedClient, error) {
    c.cond.L.Lock()
    closed := c.closed
    c.cond.L.Unlock()
    if closed {
        return nil, ErrClientPoolClosed
    }
    return NewClient(c.cfg.ClientFactory, c.cfg.ClientBufSize,
        c.cfg.ClientRetryTimes, c.cfg.ClientRetryInterval, c.cfg.ClientRetryDuration)
}
```

## Steps to Reproduce

1. Deploy cluster with unreachable log replicas (see #25274).
2. Trigger truncate (runs every ~2 min), which calls `GetOnFly`.
[...]

PR #25290, opened 2026-07-01
fix(logservice): create on-fly clients outside pool lock

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25275

  This PR prevents clientPool.GetOnFly from holding the pool mutex while creating a new logservice client.

  Previously, GetOnFly held c.cond.L during NewClient, which may perform network I/O and block for a long time when logservice replicas are unreachable. While that lock was held, normal pool operations such as Get, Put,
  and GetWithWriteToken could not proceed, causing the WAL commit path to stall.

  Changes:
  - Move NewClient execution outside the clientPool mutex.
  - Keep the lock only for checking whether the pool is closed.
  - Re-check the closed state after client creation.
  - Close the newly created client and return ErrClientPoolClosed if the pool was closed during creation.
  - Add a regression test proving that pool Get is not blocked while GetOnFly is creating a client.


XuPeng-SH (approved):
LGTM.

I checked the PR description and the current comment thread closely. The core production change is narrowly scoped: it only moves `NewClient()` out from under the pool mutex, then re-checks `closed` and closes the newly created client if shutdown won the race. That directly addresses the reported lock amplification on the WAL commit path without changing the surrounding pool semantics.

I don’t see a substantive correctness issue in the production fix itself.

The only thing I would still improve is the new regression test robustness (the current bot comments are fair): assertions should happen in the parent goroutine, and timeout/failure branches should make sure blocked goroutines are unblocked/joined. I would treat that as follow-up test hardening rather than a reason to block merge.

XuPeng-SH (changes requested):
I re-reviewed this from the lock/lifecycle angle and the production direction looks right: `GetOnFly()` now only uses the pool mutex for the `closed` checks, keeps `NewClient()` network I/O outside the pool lock, and closes the newly-created client if `Close()` wins the race. That addresses the issue without changing pooled-client ownership.

I am requesting changes for the regression test, because this PR is specifically about a concurrency/unhappy-path failure and the new test currently has incomplete cleanup/observation semantics.

1. `TestGetOnFlyDoesNotBlockPoolGetWhileCreatingClient` can leak the goroutine intentionally blocked in `ClientFactory`. Both timeout branches call `t.Fatal` before `releaseFactory` is closed and before the `GetOnFly` goroutine is joined. `t.Cleanup(pool.Close)` does not unblock `<-releaseFactory`, so a failing run can leave a stuck goroutine in the test process.

2. The `pool.Get()` goroutine calls `require.NoError(t, err)` inside the child goroutine. If `pool.Get()` returns an error, the parent goroutine may just wait on `got` and report the misleading timeout message instead of the real error. For a regression test around lock blocking, that makes failure diagnosis ambiguous.

Suggested fix:
- Send a struct like `{client *wrappedClient; err error}` from the `pool.Get()` goroutine and assert in the parent goroutine.
- Register cleanup before the first timeout path: use `sync.Once` to close `releaseFactory`, and ensure the `GetOnFly` goroutine is joined via a done channel with its own timeout.
- On any failure path, release the blocked factory and close the pool so any `pool.Get()` waiter is also unblocked.
- Add a small unhappy-path assertion for `Close()` racing with slow `GetOnFly()`: after closing the pool while factory creation is blocked, release the factory, require `ErrClientPoolClosed`, and verify the newly-created client is closed. That directly covers the post-create `closed` re-check added by this patch.

Validation I ran locally with the PR head:
- `go test ./pkg/vm/engine/tae/logstore/driver/logservicedriver -run 'TestGetOnFlyDoesNotBlockPoolGetWhileCreatingClient|TestNewClientPool' -count=20` passed with the proper MatrixOne CGo env.
- `go test -race ./pkg/vm/engine/tae/logstore/driver/logservicedriver -run TestGetOnFlyDoesNotBlockPoolGetWhileCreatingClient -count=5` passed.
- `go test ./pkg/vm/engine/tae/logstore/driver/logservicedriver -count=1` passed.

XuPeng-SH (approved):
LGTM.

I re-reviewed the latest head `2ff426d09d2239b8505c4e9681b97747addc6bbd` against the previous requested changes.

The production change remains narrowly scoped and correct: `GetOnFly()` no longer holds the pool mutex across `NewClient()` network I/O, then re-checks `closed` and closes the created on-fly client if shutdown won the race.

The test concerns from my previous review are now addressed:
- the `pool.Get()` goroutine returns `{client, err}` to the parent goroutine, so errors are asserted in the main test flow instead of being hidden behind a timeout;
- `releaseFactory` is protected by `sync.Once` and is released during cleanup, so the intentionally blocked factory path is not left stuck;
- the new `TestGetOnFlyClosesCreatedClientWhenPoolClosesConcurrently` covers the critical unhappy path where `Close()` races with slow client creation and verifies the newly-created client is closed.

Validation I ran locally with the current PR head:
- `go test ./pkg/vm/engine/tae/logstore/driver/logservicedriver -run 'TestGetOnFly|TestNewClientPool' -count=50` passed with the proper MatrixOne CGo env.
- `go test -race ./pkg/vm/engine/tae/logstore/driver/logservicedriver -run 'TestGetOnFly' -count=10` passed.
- `go test ./pkg/vm/engine/tae/logstore/driver/logservicedriver -count=1` passed.
- `go vet ./pkg/vm/engine/tae/logstore/driver/logservicedriver` passed.
- `gofmt -d` and `git diff --check` are clean.

One current CI SCA job failed before running static checks: `make config` hit a `500 Internal Server Error` from `proxy.golang.com.cn` while downloading `golang.org/x/tools@v0.0.0-20190328211700-ab21143f2384.mod`. That looks like CI/network dependency infra, not a failure introduced by this patch.

jiangxinmeng1:
@Mergifyio queue

Issue #25186, opened 2026-06-26
[Bug]: INSERT ... SELECT * ... does not respect query rewrite hint.


fengttt:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

HEAD

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

Create database foo;
use foo;

create table t (i int, j int);
create table t2(i int, j int);
insert into t select result, result from generate_series(1, 100) tmp;

set enable_remap_hint = 1;

-- the following works.  only 10 rows selected
/*+ { "rewrites": { "foo.t": "select * from foo.t where i < 10" } } */ select * from t;

-- the following will insert 100 rows into t2;   this is a bug that will leak data
/*+ { "rewrites": { "foo.t": "select * from foo.t where i < 10" } } */ insert into t2 select * from t;

### Steps to Reproduce

```Markdown
See above.
```

aunjgr:
merged into main

slyunnn:
回归测试结论：通过：重写提示（remap hint）回归 BVT 54/54。

slyunnn:
### Extended verification for #25186 (2026-08-11)

  Follow-up to the INSERT...SELECT remap-hint report: verified that rewrite hints propagate to the INSERT read source on
  the latest main branch.

  ### Verification environment

  - MatrixOne branch: `main`
  - MatrixOne commit: `d01c85974f9540f8baa2e768ca86fea952941fe2`
    `fix(hashjoin): broadcast const vectors in spill paths (#26938)`
  - Fix under verification: `8acfcaaea81bb11c24f9ba27d5951b59fc89413e`
    `fix: propagate rewrite hints to INSERT...SELECT inner query (#25310)`
  - MOTR branch: `test/issue-25186-insert-select-remap-hint`
  - MOTR commit: `f009bcac53666e3217b03742c7ddb9d271417412`

  ### What was covered

  - Original `INSERT INTO ... SELECT *` remap-hint shape.
  - INSERT...SELECT with an explicit target column list.
  - Parenthesized SELECT in INSERT.
  - INSERT...SELECT with `ON DUPLICATE KEY UPDATE`.
  - Existing remap-hint parser test suite.

  ### Results

  - Existing remap-hint parser tests passed.
  - MOTR regression passed: 21/21 statements.
[...]

PR #25310, opened 2026-07-01
fix: propagate rewrite hints to INSERT...SELECT inner query

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25186

`INSERT...SELECT` ignores query rewrite hints. While `SELECT` correctly applies `/*+ {"rewrites": ...} */` to filter/transform table scans, `INSERT INTO t2 SELECT * FROM t1` silently drops the hint and inserts all rows — leaking data.

### Root cause

`AddRewriteHints` in `sqlparse.go` only recognizes `*tree.Select` and `*tree.ParenSelect` at the top level. `INSERT...SELECT` is `*tree.Insert`, which hits `default: continue` — the hint is discarded before the inner `*tree.Select` ever sees it.

### Fix

Two changes in `AddRewriteHints`:

1. **Pass-through switch**: add `*tree.Insert` so INSERT statements are not skipped.
2. **Attachment switch**: add `*tree.Insert` case that sets `RewriteOption` on `stmt.Rows` (the inner `*tree.Select`), also propagating to the unwrapped Select for parenthesized forms like `INSERT INTO t (SELECT ...)`.

Both legacy (`buildInsert`) and modern (`bindInsert`) INSERT paths pass `stmt.Rows` through `bindSelect`, which reads `stmt.RewriteOption` — the hint flows to table resolution naturally.

| File | Change |
|------|--------|
| `pkg/sql/parsers/sqlparse.go` | +6 lines: two switch changes |
| `pkg/sql/parsers/sqlparse_test.go` | +60 lines: rename + 4 new unit tests |
| `test/.../remap_hint.sql` | +26 lines: 3 BVT scenarios (filter, column subset, ODKU) |
| `test/.../remap_hint.result` | regenerated |


heni02 (approved):

XuPeng-SH (approved):
LGTM.

I reviewed the fix against issue #25186. The root cause is correctly addressed: `AddRewriteHints` previously filtered out top-level `*tree.Insert`, so the rewrite hint was never attached before the inner SELECT reached planning. The patch lets INSERT pass the hint parser and attaches the resulting `RewriteOption` to `stmt.Rows`, which is the `*tree.Select` consumed by both insert planning paths.

I checked the downstream consumption path:
- modern `bindInsert` passes `stmt.Rows` into `initInsertReplaceStmt`, which calls `bindSelect` for `SelectClause` and `ParenSelect` inputs;
- legacy `buildInsert` passes `stmt.Rows` through `initInsertStmt`, which follows the same `bindSelect` route;
- `bindSelect` is the point that transfers `RewriteOption` into `BindContext.remapOption`, and `buildTable` consumes that option for table remapping.

The parenthesized-select case is also handled correctly by setting the option on the unwrapped inner `Select`, which matters because insert planning binds `slt.Select` directly for `*tree.ParenSelect`.

Coverage looks appropriate: parser unit tests cover INSERT VALUES, INSERT SELECT, parenthesized SELECT, ODKU, and WITH; BVT covers basic filter, column subset, and ODKU behavior. CI is green.

Validation I ran locally:
- `go test ./pkg/sql/parsers -run 'TestAddRewriteHints' -count=1`
- `go test ./pkg/sql/parsers -count=1`
- `go test ./pkg/sql/parsers -run 'TestAddRewriteHints' -count=100`
- `go test ./pkg/sql/parsers -run 'TestAddRewriteHints|TestExtractLeadingHints' -count=1 -race`
- `go vet ./pkg/sql/parsers`
- `gofmt -d` and `git diff --check`

Non-blocking: the PR is currently behind `main`; please sync before merge if the queue requires it.

iamlinjunhong (approved):

PR #25311, opened 2026-07-01
fix: RANGE window frame with NULL ORDER BY key differs from MySQL

### LINKED ISSUE #25288: [Bug]: RANGE window frame with NULL ORDER BY key differs from MySQL

I searched existing issues for `window RANGE NULL order key frame` and `range between preceding current row NULL window`. I only found an unrelated VARCHAR ORDER BY window issue (#23070), and did not find a duplicate for NULL order keys in RANGE frames.

### MatrixOne version

Verified on latest `main` at `200bfa7e0efb` (`fix: improve skills SKILL.md compatibility format (#25271)`).

### MySQL version

Local MySQL 8.3.0.

### Reproduce steps

```sql
drop database if exists repro_range_null_order_key;
create database repro_range_null_order_key;
use repro_range_null_order_key;

create table t(id int primary key, k int, v int);
insert into t values
  (1, null, 10),
  (2, null, null),
  (3, 1, 5),
  (4, 2, null),
  (5, 2, 7),
  (6, 4, 11);

select id, if(k is null, 'NULL', cast(k as char)) as k_label, v,
       count(*) over (order by k range between 1 preceding and current row) as cnt_all_rng,
       count(v) over (order by k range between 1 preceding and current row) as cnt_v_rng,
       sum(v) over (order by k range between 1 preceding and current row) as sum_v_rng,
       min(v) over (order by k range between 1 preceding and current row) as min_v_rng,
       max(v) over (order by k range between 1 preceding and current row) as max_v_rng
from t
order by k is not null, k, id;
```

### Expected behavior

MySQL treats the two `NULL` order-key rows as peers for the `RANGE` frame, and does not include those `NULL` rows in the numeric frame for `k = 1`:

```text
id  k_label  v     cnt_all_rng  cnt_v_rng  sum_v_rng  min_v_rng  max_v_rng
1   NULL     10    2            1          10         10         10
2   NULL     NULL  2            1          10         10         10
3   1        5     1            1          5          5          5
4   2        NULL  3            2          12         5          7
5   2        7     3            2          12         5          7
6   4        11    1            1          11         11         11
```

### Actual behavior

MatrixOne returns different frames for NULL order-key rows and includes them in the `k = 1` frame:

```text
id  k_label  v     cnt_all_rng  cnt_v_rng  sum_v_rng  min_v_rng  max_v_rng
1   NULL     10    1            1          10         10         10
2   NULL     NULL  1            0          NULL       NULL       NULL
3   1        5     3            2          15         5          10
4   2        NULL  3            2          12         5          7
5   2        7     3            2          12         5          7
6   4        11    1            1          11         11         11
```

### Additional information

This was found while following up on TiDB/MySQL-compatible NULL window semantics. Passing BVT for ROWS-frame NULL aggregate behavior was added separately; this RANGE-frame NULL order-key case is excluded from passing BVT until fixed.

ck89119:
## What type of PR is this?

- [x] Bug fix

Fix RANGE window frame computation when ORDER BY key contains NULL values, matching MySQL behavior.

### Problem

RANGE window frames with NULL ORDER BY keys produced incorrect results in three ways:

1. **NULL peers not grouped**: NULL-keyed rows were treated as individuals instead of peers. In SQL, all NULLs are peers in window ordering.
2. **NULLs leaked into numeric ranges**: NULL rows' raw values (e.g., 0 for integers) were matched by binary search when computing numeric RANGE boundaries for non-NULL rows.
3. **NULLS LAST broken**: When NULLs sort last, the raw-value array is not monotonically sorted (e.g., [1,2,4,0,0]), causing binary search to return wrong results.

### Fix

In `pkg/sql/colexec/window/window.go`:

- **searchLeft/searchRight NULL handling**: When current row has NULL order key, scan left/right to find the full NULL peer group boundaries instead of returning just the current row.
- **Non-NULL range confinement**: Before binary search, trim the search range to exclude NULL positions, ensuring binary search operates on a properly sorted non-NULL subrange.

### Verification

- Manually verified all combinations (NULLS FIRST/NULLS LAST, various RANGE boundaries) match MySQL output
- BVT test suite: 99% pass rate (only pre-existing `@bvt:issue` skipped tests fail)
- Added unit tests for searchLeft/searchRight NULL handling
- Added BVT test cases from issue #25288

Fixes #25288


heni02 (approved):

aunjgr (approved):
Three fixes for RANGE window frame with NULL ORDER BY keys:

1. NULL peer grouping: searchLeft/searchRight now find the full NULL peer group boundaries instead of returning just the current row
2. Non-NULL range confinement: binary search range is trimmed to exclude NULL positions, handling the non-monotonic array from NULLS LAST
3. Post-search NULL skip: advance/retreat if binary search lands on a NULL raw value

Unit tests cover all three edge cases; BVT verifies MySQL-compatible NULL peer behavior. LGTM.

XuPeng-SH (changes requested):
I think this needs changes before merging. The ASC/default NULL case from #25288 is fixed, but the same RANGE-boundary search path is still not correct for a legal `ORDER BY ... DESC` window.

`processOrder` sorts the order vector using the window order flags, and `makeArgFs` sets `desc/nullsLast` from the `OrderBySpec`. However `buildRangeInterval` calls `searchLeft/searchRight` without passing the sort direction. Those helpers still use `genericGreater` over the vector as if the non-NULL subrange were ascending. The new NULL trimming makes the NULL portion safe, but it does not make the numeric binary search desc-aware.

A concrete repro at the helper level:

```go
// Simulates ORDER BY k DESC NULLS LAST after sorting: [4, 2, 2, 1, NULL, NULL]
// For the row k=2, RANGE BETWEEN 1 PRECEDING AND CURRENT ROW should include
// only the two k=2 peers. In descending order, 1 PRECEDING means keys in
// [current, current+1] = [2,3], so k=4 must not be included.
start, end, err := ctr.buildRangeInterval(1, 0, 6, frameForRange1PrecedingToCurrentRow)
// expected start=1,end=3; current code returns start=0,end=3, including k=4.
```

This is not just an unsupported syntax corner: the binder already accepts RANGE + ORDER BY DESC (see the existing projection binder test for "Window function with RANGE frame and ORDER BY DESC"), so executor semantics need to match that accepted surface.

Please make `RANGE` boundary search aware of the last ORDER BY direction/null ordering, or otherwise reject unsupported DESC RANGE frames at bind time. Also add regression coverage for at least:

- `ORDER BY k DESC RANGE BETWEEN 1 PRECEDING AND CURRENT ROW` with duplicated keys and NULLs.
- `ORDER BY k DESC RANGE BETWEEN CURRENT ROW AND 1 FOLLOWING`.
- Explicit `NULLS FIRST` and `NULLS LAST` variants, because default DESC implies NULLS LAST in this executor.

The current PR tests cover the issue's ASC/default case, and the targeted tests pass locally with the proper CGo environment. The blocker is the unclosed DESC RANGE semantic path.

XuPeng-SH (approved):
LGTM.

I re-checked the high-risk parts of the RANGE window-frame logic around NULL peer grouping, NULL leakage into numeric search, and DESC + NULLS LAST handling. I don’t see a substantive correctness blocker in the current diff.

The unit and SQL/BVT coverage are already pretty solid. If anything, the remaining opportunities are only nice-to-have expansions like explicit `ASC NULLS LAST` or symmetric `RANGE BETWEEN x PRECEDING AND y FOLLOWING` SQL cases.

PR #25312, opened 2026-07-01
fix: align INSERT/UPDATE string width enforcement with MySQL sql_mode (#25200) (cherry-pick #25262)

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25200

Cherry-pick of PR #25262 from 3.0-dev to main.

Rework CHAR/VARCHAR width enforcement on DML assignment paths so it matches MySQL's sql_mode semantics:

- **sql_mode gating (runtime)**: add a new `cast_assign` function used by DML assignment paths for CHAR/VARCHAR targets. It reads sql_mode at runtime (strict -> reject 1406, non-strict -> truncate), avoiding plan-cache staleness.
- **trailing-space exemption**: over-length values whose excess is only trailing spaces are accepted (truncated) even under strict mode, matching MySQL.
- **INSERT IGNORE downgrade**: under IGNORE, over-length writes are truncated instead of rejected.

## Notes

- `func_cast.go` and `func_cast_test.go` were taken from the 3.0-dev version. Main-specific functions in these files may need review.
- `build_test.go` changes were not included in this cherry-pick due to significant divergence from #24973.


XuPeng-SH (changes requested):
Requesting changes.

The cherry-pick fixes the local/plan-level blockers I raised on #25262: `INSERT IGNORE` is now threaded through the non-literal VALUES path, prepared literal VALUES now route through a volatile runtime cast instead of freezing the plan-time decision, VARBINARY is no longer relaxed by the CHAR/VARCHAR constant path, and the fallback `buildInsert` initializes `isInsertIgnore`. CI is green and the new BVT covers the main single-CN SQL matrix.

I still see one correctness blocker in the current design: `cast_assign` reads `sql_mode` from the local `process.Process` resolver at execution time, but remote-run processes do not carry that resolver or an encoded `sql_mode` value. That means the same assignment cast can behave as strict on a remote CN even when the user session is non-strict.

Relevant paths:

- `pkg/sql/plan/function/func_cast.go:458-459`: `NewAssignCast` calls `isStrictSqlMode(proc)` at function execution time.
- `pkg/sql/plan/function/func_cast.go:467-469`: if `proc.GetResolveVariableFunc()` is nil, `isStrictSqlMode` returns `true`.
- `pkg/sql/compile/remoterunClient.go:182-195`: remote run sends the encoded scope plus encoded process info.
- `pkg/vm/process/process_codec.go:92-102`: encoded `pipeline.SessionInfo` includes user/host/db/version/timezone/query id/lock timeout, but not `sql_mode` and not a resolver.
- `pkg/sql/compile/remoterunServer.go:398-415`: the remote CN rebuilds a fresh `process.Process`, restores `SessionInfo`, and does not install any `ResolveVariableFunc`.

The existing remote var-expression mechanism does not close this gap: `foldVarExprsInRemoteRunScope` only folds explicit `plan.Expr_V` variables before encoding (`pkg/sql/compile/remote_expr.go:51-118`). `cast_assign` hides the `sql_mode` dependency inside the function implementation, so the scope does not contain a var expr to fold.

This matters for assignment paths that can be executed inside remote scopes, e.g. distributed `INSERT ... SELECT` / UPDATE plans where the projection contains the new `cast_assign`. In non-strict mode, local execution truncates, while a remote CN with nil resolver will take the strict fallback and return 1406. The current BVT covers the SQL semantics, but not this remote-process boundary.

Suggested fix: make the strictness input explicit and portable across remote execution. For example, either encode the session `sql_mode` into the remote `ProcessInfo` / `SessionInfo` and have `isStrictSqlMode` fall back to that, or model `sql_mode` as an explicit expression/input to the assignment cast so the existing remote var-folding path can capture the current execution value. Please also add coverage for the remote/multi-CN path: non-strict `INSERT ... SELECT` with over-width CHAR/VARCHAR should truncate on a remote execution path, and strict mode should still reject.

Local validation I ran on the current PR head:

```text
go test -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" \
  -v -count=1 -timeout 120s ./pkg/sql/plan/function \
  -run 'Test(StrToStrWidthEnforcement|OverLenIsAllTrailingSpaces|IsStrictSqlMode|NewAssignCastHonorsSqlMode|strToStr_StrictStringWidth|strToStr_TextToCharVarchar)'
PASS

go test -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" \
  -v -count=1 -timeout 120s ./pkg/sql/plan \
  -run 'Test(InsertSelectVarcharFromTextUsesAssignmentCast|OnDuplicateUpdateVarcharFromTextUsesAssignmentCast|MakeInsertValueConstExprGeometry)'
PASS
```

heni02 (approved):

aunjgr (changes requested):
[P1] Adding CAST_ASSIGN at ID 542 renumbers existing DATE_TRUNC and JSON function IDs. Function IDs are serialized in plans, so mixed-version CNs can interpret the same ID as a different function during rolling upgrades. Preserve every existing ID and append CAST_ASSIGN after the current stable range; update the end marker without shifting prior entries.

aunjgr (dismissed):
[P1] Adding CAST_ASSIGN at ID 542 renumbers existing DATE_TRUNC and JSON function IDs. Function IDs are serialized in plans, so mixed-version CNs can interpret the same ID as a different function during rolling upgrades. Preserve every existing ID and append CAST_ASSIGN after the current stable range; update the end marker without shifting prior entries.

XuPeng-SH (changes requested):
Requesting changes on `432e06217e7ecd552251d0e89b8da7ef185d36b2`.

The earlier remote `sql_mode` and function-ID findings are fixed, but the latest head still has four correctness/compatibility blockers:

1. **[P1] Cast warnings are dropped at the first real pipeline-process boundary.**

   `newCast` captures warnings through `proc.GetSession()` (`pkg/sql/plan/function/func_cast.go:950-957`), but `Process.Session` is a field on the individual `Process`, while `NewNoContextChildProc` / `NewContextChildProc` copy only `Base` (`pkg/vm/process/process2.go:107-147`). Consequently the processes that execute pipeline expressions have a nil session, and `addCastWarningToSession` silently returns. A decoded remote process has no frontend session either, so the same loss happens across CNs.

   Concrete failure: an over-width non-strict/IGNORE assignment is truncated successfully, but the OK packet reports zero warnings and `SHOW WARNINGS` is empty. I reproduced this by assigning a warning-capable session to a top process, executing `NewAssignCast` on `top.NewContextChildProc(0)`, and observing the expected value `"abc"` with no 1265 recorded. The existing warning unit test sets `Session` directly on the same process used for evaluation, so it does not exercise the production boundary.

   Please put warning collection on state shared by child processes and close the remote-result aggregation path (or explicitly defer/remove the partial warning feature until that end-to-end path exists).

2. **[P1] SELECT responses still advertise zero warnings.**

   `castModeExplicit` now records `ER_TRUNCATED_WRONG_VALUE`, but the `*tree.Select` branch in `respStreamResultRow` still calls `WriteEOFOrOK(0, ...)` (`pkg/frontend/result_row_stmt.go:297-302`). Thus even after warning collection is fixed, `SELECT CAST(...)` sends a zero warning count in its final EOF/OK packet. The added `TestStreamResultResponseCarriesWarnings` uses `SHOW WARNINGS`, which takes the default branch and misses this case. I reproduced it with a `tree.Select` response carrying one stored warning: the writer received 0.

   Pass `ses.WarningCount()` in the SELECT branch and cover the final result-set terminator (including `CLIENT_DEPRECATE_EOF`).

3. **[P1] The protocol-v4 compatibility plan now returns a DDL error for DML.**

   During a mixed-version rollout, `assignmentCastFunctionName` intentionally emits `cast_strict` (`pkg/sql/plan/build_constraint_util.go:1094-1106`). However, this PR changed `NewStrictCast` width failures to `NewErrInvalidDefault` in the `allowTrailingSpaceTrim == false` path (`pkg/sql/plan/function/func_cast.go:7007-7010`, with equivalent branches for other sources). Therefore the new binary executing the v4-compatible DML plan returns 1067/`ErrInvalidDefault`, while the old binary executing the same `cast_strict` plan returns the legacy DML/internal cast error. Client-visible behavior now depends on which CN executes the plan.

   A focused execution of `NewStrictCast(TEXT('abcd') -> VARCHAR(3))` on this head returned `ErrInvalidDefault`; the same path on the base returns `ErrInternal`. Please keep `cast_strict`'s stable contract and use a DDL-specific operator/mode/context for DEFAULT validation.

4. **[P1] Persisting `cast_assign` in generated-column metadata breaks a 5→4 rollback.**

   At protocol v5, `buildGeneratedExpr` stores `cast_assign` (ID 554) directly in the catalog expression (`pkg/sql/plan/build_util.go:488-516`). The previous binary has `FUNCTION_END_NUMBER = 554`, no `cast_assign` registration, and expression execution resolves the encoded ID rather than `ObjName`; it therefore fails with `function overload id not found`. `applyGeneratedColumnAssignmentCast` only protects planners running the new code and cannot help a rolled-back CN reading the catalog.

   Concrete sequence: upgrade and set protocol 5 → create a VARCHAR generated column → set protocol back to 4/roll a CN back → INSERT/UPDATE planned on that old CN cannot execute the stored generated expression. Since this PR explicitly handles and tests protocol rollback for cached plans, catalog state needs the same closure. Keep the persisted expression on the stable `cast_strict` ID and rewrite it to `cast_assign`/`cast_ignore` only while building an execution plan.

Focused existing tests for function semantics, protocol-gated planning, generated-column rewriting, prepared-plan/cache invalidation, and frontend warning responses pass. The boundary regressions above are not covered by those tests.

XuPeng-SH (changes requested):
Requesting changes on `9c1c54c8b99043cf72ab44b0a86fd09ed3f59e14`.

The earlier cast-strict contract and generated-column rollback findings are fixed, and the focused cast/planner/protocol tests pass. Three warning-path correctness gaps remain:

1. **[P1] Warning collection still stops at the first production `Process` boundary and has no remote return path.**

`newCast` captures its sink from `proc.GetSession()` (`pkg/sql/plan/function/func_cast.go:948-957`), but `Process.Session` is outside the shared `BaseProcess` (`pkg/vm/process/types.go:421-430`). `NewNoContextChildProc` and `NewContextChildProc` copy only `Base` (`pkg/vm/process/process2.go:109-147`), so the child processes used to evaluate pipeline expressions have a nil session. Remote execution is also incomplete: `remoterunServer.newCompile` reconstructs a top process and restores only `Base.SessionInfo` (`pkg/sql/compile/remoterunServer.go:669-685`), and neither the pipeline protocol nor the terminal envelope carries warnings back to the initiating CN.

Concrete failures: `SELECT CAST('abcd' AS CHAR(3))`, a non-strict over-width INSERT/UPDATE, or INSERT/UPDATE IGNORE can truncate successfully while the client sees warning_count=0 and `SHOW WARNINGS` is empty; the same loss is guaranteed when the cast executes remotely. `TestAssignmentCastTypedSourceAndWarnings` sets `proc.Session` directly on the exact evaluator process, so it bypasses this production boundary.

Please put the warning sink/counter on state shared by child processes and close the remote aggregation path (or defer the warning feature until that end-to-end path exists). Add a child-process regression and a multi-CN remote regression.

2. **[P1] Ordinary SELECT still writes a zero warning count in its final EOF/OK packet.**

The `*tree.Select` branch still calls `WriteEOFOrOK(0, ...)` at `pkg/frontend/result_row_stmt.go:296-302`. Thus even after collection is fixed, explicit CAST warnings are hidden from the protocol response. `TestStreamResultResponseCarriesWarnings` uses `tree.ShowWarnings`, which reaches the default branch and does not exercise SELECT.

Pass `ses.WarningCount()` in the SELECT branch and test both EOF and `CLIENT_DEPRECATE_EOF` terminators for a warning-producing SELECT.

3. **[P2] Protocol warning_count is derived from the bounded display buffer, so bulk warnings are under-counted.**

`AddWarning` stores warnings in `errInfo`, whose default max is 64, while `WarningCount` returns `len(ses.warnings.codes)` (`pkg/frontend/session.go:1934-1962`). `errInfo.push` also uses `len > maxCnt` rather than `>=` (`pkg/frontend/session.go:711-717`), so the buffer stabilizes at 65 entries. A statement that truncates 100 rows therefore advertises 65 warnings, not 100. The storage limit should bound `SHOW WARNINGS` retention, not the statement total used by the protocol; track the total separately (capped at uint16 for the wire) and cover a batch larger than the retention limit.

Validation on this head: the newly added focused tests in `pkg/common/moerr`, `pkg/vm/process`, `pkg/sql/plan/function`, `pkg/sql/plan`, and `pkg/frontend` all pass with the repository CGo wrapper. CI is green except for the still-running arm64 SCA job.

ck89119:
@XuPeng-SH Thanks for the review. The three remaining findings are all related to warning collection, propagation, protocol warning counts, and SHOW WARNINGS behavior. MatrixOne does not currently support an end-to-end warning mechanism, so warning-related features and change requests are explicitly out of scope for this PR and will not be addressed here. The previously reported non-warning issues—the cast_strict compatibility contract and generated-column rollback compatibility—have been fixed.

XuPeng-SH (changes requested):
Re-review on unchanged head `9c1c54c8b99043cf72ab44b0a86fd09ed3f59e14`: the requested changes remain.

Calling warning behavior out of scope does not resolve the blockers because this PR itself adds warning production (`newCast`), storage/counting (`Session`), `SHOW WARNINGS`, and protocol warning counts. Either complete that behavior end to end or remove/defer those warning-facing changes from this PR.

1. **[P1] Warning collection is still lost at real process/CN boundaries.** `Process.Session` is outside shared `BaseProcess`; `NewNoContextChildProc`/`NewContextChildProc` copy only `Base`, so evaluator child processes have no warning sink. Remote execution also has no terminal warning aggregation back to the initiating CN. The current cast test attaches the session directly to the evaluator and bypasses both boundaries.
2. **[P1] Ordinary SELECT still advertises zero warnings.** The `*tree.Select` branch in `respStreamResultRow` calls `WriteEOFOrOK(0, ...)`; the added test uses `tree.ShowWarnings`, which takes a different branch.
3. **[P2] Protocol warning_count is the retained-buffer length, not the statement total.** `errInfo.push` retains at most 65 entries with the current `len > maxCnt` condition, and `WarningCount` returns that retained length, so 100 truncations are reported as 65.

The focused cast/frontend warning tests pass on this head, but they confirm only the bypassing unit-test paths above. CI is green; these are uncovered correctness gaps, not CI failures.

ck89119:
@XuPeng-SH Thanks for the clarification. I removed the warning-facing changes introduced by this PR instead of extending MatrixOne's currently incomplete warning mechanism.

The cast warning production/context, session warning storage/counting, `SHOW WARNINGS`-specific state, protocol warning propagation, and warning-only tests have been removed. String-width behavior remains unchanged: strict assignment rejects over-width values, non-strict/`IGNORE` truncates them, and explicit `CAST` returns the truncated value.

Validation:

- Full tests passed for `pkg/sql/plan/function`, `pkg/frontend`, `pkg/sql/plan`, and `pkg/vm/process`.
- Read-only build and vet passed for the changed package closure.
- Focused string-width BVT passed: 138/138.
- `make static-check` remains blocked by the repository's existing `go.mod` / `vendor/modules.txt` inconsistency, unrelated to these files.

Fixed in `bb5fb4133`.

XuPeng-SH (approved):
Re-reviewed the latest head bb5fb4133d. The incomplete warning pipeline that caused the remaining blockers has been fully removed, while the core strict/non-strict/IGNORE string-width behavior remains intact. I also rechecked remote sql_mode propagation, function-ID/protocol-version gating, plan-cache invalidation, and generated-column rolling-upgrade compatibility. Local full-package tests passed for pkg/common/moerr, pkg/vm/process, pkg/sql/plan/function, pkg/sql/plan, and pkg/frontend.

XuPeng-SH (approved):

Issue #24907, opened 2026-06-09
[Bug]: REPLACE / INSERT ... ON DUPLICATE KEY UPDATE report non-MySQL affected-rows


ck89119:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

4.0-dev

### Commit ID

7a8e7cd84

### Other Environment Information

- Hardware parameters: N/A
- OS type: macOS (darwin/arm64); also reproducible on Linux
- Others: standalone `mo-service -launch etc/launch/launch.toml`

### Actual Behavior

MatrixOne counts affected rows as the number of rows *net-written to the target main table, counted once*. For statements where MySQL uses its client-protocol convention, the affected-rows value (returned in the OK packet and by `ROW_COUNT()`) differs from MySQL:

| Statement | MatrixOne | MySQL |
|---|---|---|
| `REPLACE` over an existing key (delete + insert) | **1** | 2 |
| `INSERT ... ON DUPLICATE KEY UPDATE` -> row updated | **1** | 2 |
| `INSERT ... ON DUPLICATE KEY UPDATE` -> no-op (same values) | **1** | 0 |
| `INSERT ... ON DUPLICATE KEY UPDATE` -> new row inserted | 1 | 1 |
| `REPLACE` inserting a new key | 1 | 1 |

### Expected Behavior

MySQL-compatible affected-rows:
- `REPLACE` replacing an existing row -> `2` (delete + insert)
- `ON DUPLICATE KEY UPDATE` updating a row -> `2`
- `ON DUPLICATE KEY UPDATE` no-op (row set to its current values) -> `0`

(MySQL docs: "with `ON DUPLICATE KEY UPDATE`, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values"; `REPLACE` returns 2 when it replaces an existing row.)

### Steps to Reproduce

```sql
drop database if exists rc; create database rc; use rc;
create table t(id int primary key, v int);
insert into t values (1,1),(2,2);

-- REPLACE over an existing key
replace into t values (1,100);
select row_count();                                  -- MO: 1, MySQL: 2

-- ON DUPLICATE KEY UPDATE: update an existing row
insert into t values (1,5) on duplicate key update v=v+1;
select row_count();                                  -- MO: 1, MySQL: 2

-- ON DUPLICATE KEY UPDATE: no-op (set to current value)
[...]

PR #25318, opened 2026-07-01
fix: align REPLACE/ODKU affected-rows with MySQL

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24907

Align affected-rows for REPLACE and INSERT ... ON DUPLICATE KEY UPDATE with MySQL:

| Statement | Before | After | MySQL |
|-----------|--------|-------|-------|
| REPLACE existing key | 1 | **2** | 2 |
| REPLACE new key | 1 | 1 | 1 |
| ODKU real update | 1 | **2** | 2 |
| ODKU no-op (same values) | 1 | **0** | 0 |
| ODKU insert new | 1 | 1 | 1 |

### Changes

- **proto**: add `count_delete_affect_rows` field to `UpdateCtx`
- **executor** (`multi_update/types.go`): count DELETE side for upsert when flagged
- **compile/partition**: propagate the flag through operator construction
- **plan**: set flag on main-table `UpdateCtx` for ODKU and REPLACE; add NOT(<=>) no-op FILTER for ODKU to detect unchanged rows
- **tests**: 7 executor-level + 3 plan-level unit tests


XuPeng-SH (changes requested):

The new ODKU “no-op” guard is still built after implicit `ON UPDATE` columns have already been injected into the updated row image.

Concretely:
- `bind_insert.go:1349-1358` unconditionally adds implicit `col.OnUpdate.Expr` values into `updateExprs`.
- The no-op filter at `bind_insert.go:2012-2057` then compares the old row against that already-updated new row image across all visible columns.

So on a table with an implicit auto-update column like `update_at ... ON UPDATE CURRENT_TIMESTAMP`, a statement such as:

```sql
insert into t(id, v) values (1, 10)
on duplicate key update v = v
```

is still treated as a real update instead of a MySQL-style no-op, because `update_at` differs even though the explicit user update changed nothing.

That means affected rows will still be wrong (and the auto-update column may still advance) for a common unhappy path the PR is explicitly trying to fix.

Suggested fix:
1. build the no-op predicate from the row image before implicit `ON UPDATE` expressions are applied, or
2. explicitly exclude purely implicit auto-update columns from the equality check and only materialize them once some other column is known to change.

Please also add a regression for ODKU no-op on a table with `ON UPDATE CURRENT_TIMESTAMP`.

XuPeng-SH (changes requested):
Re-reviewing the latest head: the earlier direct blocker around the implicit `ON UPDATE` column itself is fixed now, but I still see one closely related correctness gap.

The no-op filter now excludes the direct auto-update column (`updated_at`), but it still includes generated columns that are recomputed from that auto-update value.

So a table like:

```sql
create table t (
  id int primary key,
  v int,
  updated_at timestamp default current_timestamp on update current_timestamp,
  g timestamp as (updated_at) stored
);

insert into t(id, v) values (1, 10)
on duplicate key update v = v;
```

can still be treated as a real update instead of a MySQL-style no-op, because `g` changes even though the only source of change is the implicit `ON UPDATE` path that the no-op filter is supposed to ignore.

The current planner wiring still builds generated-column expressions from the already-updated row image, feeds those expressions into the dedup-join update row image, and then compares generated columns in the no-op filter.

Suggested fix:
1. build the no-op predicate from the row image before implicit `ON UPDATE`-driven generated-column recomputation, or
2. explicitly exclude generated columns whose new value changes only because of skipped implicit `ON UPDATE` columns.

Please also add a regression covering ODKU no-op on a table with both an implicit `ON UPDATE CURRENT_TIMESTAMP` column and a generated column derived from it.

aunjgr (changes requested):
No-op detection runs AFTER ON UPDATE CURRENT_TIMESTAMP fires, so UPDATE v=v still counts as a real change. Must build the equality predicate before implicit auto-update columns are applied.

heni02 (approved):

XuPeng-SH (changes requested):
Re-reviewing the latest head: the earlier blocker around implicit `ON UPDATE` columns defeating the ODKU no-op filter is fixed now, including the generated-column case you asked about earlier.

However, I still see one remaining substantive blocker on the latest head:

`CountDeleteAffectRows` is not preserved when `MultiUpdate` is duplicated or rebuilt for remote execution.

That flag is what makes the executor count the delete side for REPLACE / real-update ODKU so affected rows become MySQL-compatible (`2` instead of `1`). The current head sets it when the operator is first built from the plan, but then drops it in both:
- the operator clone path (`pkg/sql/compile/operator.go`, `vm.MultiUpdate` case), and
- the remote pipeline serialize / rebuild path (`pkg/sql/compile/remoterun.go`, `vm.MultiUpdate` case).

So any execution path that clones or ships a `MultiUpdate` can silently fall back to the old behavior, even though the local/direct path is fixed.

Suggested fix:
1. preserve `CountDeleteAffectRows` in the `dupOperator` path for `vm.MultiUpdate`;
2. preserve it across remote pipeline round-trip as well — either as a top-level field or by reconstructing it from the incoming `plan.UpdateCtx` entries;
3. add a regression that duplicates / remote-round-trips a `MultiUpdate` with `CountDeleteAffectRows=true` and verifies the delete side is still counted.

XuPeng-SH (changes requested):
Requesting changes on current head 35e061c.

There is still a correctness gap in the ODKU no-op guard for real-PK tables when the conflict is resolved through a secondary UNIQUE key and the incoming PK is different from the existing row's PK.

Repro shape:

```sql
create table t(id int primary key, u int unique, v int);
insert into t values (1, 10, 5);
insert into t values (2, 10, 5) on duplicate key update v = v;
select row_count(); -- MySQL: 0
```

The update clause is a no-op against the conflicting row, so affected rows should be 0. On this PR the no-op FILTER can keep the row as a real update because it compares immutable key columns from the old row against the incoming row image.

The relevant planner path is:

- `bind_insert.go` resolves a secondary-unique conflict to `target_pk` and keys the main DEDUP UPDATE join on that value (`useTargetPk`, then `rightPkPos = targetPkPos`).
- But the no-op guard later iterates every visible column and builds `old scanTag col <=> new selectTag col` using `colName2Idx[table.col]` (`bind_insert.go:2033-2062`). For the PK column this is still the incoming PK position, not the resolved target row's existing PK/final immutable value.

I verified this with a temporary planner test using:

```sql
insert into constraint_test.dept(deptno, dname, loc)
values (999, 'Sales', 'NY')
on duplicate key update loc = loc;
```

The generated no-op predicate included:

```text
old dept.deptno <=> col_pos:4
```

where `col_pos:4` is the incoming-side `deptno` from the dedup output. So an incoming PK like 999 makes the all-cols-equal predicate false even though the user update does not change the conflicting row.

Suggested fix:

1. Build the no-op predicate from the actual final row image that will be written, not from the raw incoming row image; or explicitly exclude immutable PK/unique conflict-resolution columns that are not user-updated and compare only columns whose final stored value can actually change.
2. Preserve the existing `old __mo_rowid IS NULL OR ...` guard for non-conflicting inserts.
3. Add regression coverage with `ROW_COUNT()` for secondary-unique no-op with a different incoming PK, e.g. the SQL above. This PR currently has good unit coverage for helper accounting/flag propagation, but the SQL BVT does not directly assert `ROW_COUNT()` for the issue's affected-rows contract.

Local checks I ran:

- `go test ./pkg/sql/colexec/multi_update -run 'TestUpsertAffectRowsAccounting|TestAddInsertAffectRows|TestAddDeleteAffectRows'` passed
- `go test ./pkg/sql/plan -run 'TestUpsertAffectRowsPlan'` passed
- `go test ./pkg/sql/compile -run 'TestDupOperatorMultiUpdateCountDeleteAffectRows|TestConvert.*MultiUpdate|Test.*CountDeleteAffectRows'` passed for the matching compile test present in this head

Those tests do not cover this secondary-unique no-op case.

ck89119:
@XuPeng-SH Thanks for the thorough review.

Regarding `ROW_COUNT()`: MO's `row_count()` is currently a stub that always returns `0` (`pkg/sql/plan/function/func_unary.go:6797` — it returns `opNoneParamToFixed[uint64](..., func() uint64 { return 0 })`). Neither mo-tester's genrs records affected-rows counts in `.result` files, so we can't use `ROW_COUNT()` as a BVT assertion today.

Instead the new BVT case (`t_odku_sec_uk`) asserts the no-op via an `ON UPDATE` timestamp column — the same deterministic technique already used by the existing `t_odku_onupdate` and `t_odku_onupdate_gen` cases in the same file:
- **no-op** (`v=v` through a secondary-unique conflict with a different incoming PK): `updated_at` must NOT advance → `ts_unchanged = 1`
- **real update** (`v=v+1`): `updated_at` must advance → `ts_advanced = 1`

Additionally, I verified protocol-level affected-rows correctness via a `database/sql` E2E (Go `RowsAffected()` directly against the running server):
- no-op `v=v` → `RowsAffected = 0`
- no-op `v=v` with a different incoming non-key value (2,10,99) → `RowsAffected = 0` (final written value = old value, still no-op)
- real update `v=v+1` → `RowsAffected = 2`
- fresh insert (7,70,7) → `RowsAffected = 1`

All align with MySQL semantics.

ck89119:
@XuPeng-SH Good catch — fixed in 59b205097.

You are right: the previous guard re-evaluated the assignment expression, which double-evaluates a non-deterministic update (`v = floor(rand()*2)`) and can disagree with the value the dedup-update join actually stored — dropping a real update or counting a genuine no-op.

The fix compares the old image against the final written value the dedup-update join **already materialized**, rather than re-running the assignment. At runtime the join evaluates each `UpdateColExprList` expression once and writes the result back into the new-image (build/`selectTag`) column at `colName2Idx` (`pkg/sql/colexec/dedupjoin/join.go` — `joinBat1.Vecs[pos] = vecs[j]`, emitted via `rp.Rel == 1`). The guard now references that column, so the no-op decision always uses the exact value that gets stored; non-deterministic assignments are evaluated exactly once.

A side benefit: referencing the new-image column makes the predicate span both join sides, so it can no longer be pushed below the dedup-update join. That removes the need for the earlier `pushdown.go` guard, which is reverted in this commit.

Verification (Go `database/sql`, `RowsAffected()` against a running server):
- `insert (2,10,99) on duplicate key update v = v` over stored `(1,10,5)` → `RowsAffected = 0`, row stays `v = 5` (incoming 99 is not written — proves the guard compares the materialized value, not the raw incoming image)
- `v = v + 1` → `2`
- `v = floor(rand()*2)` (stored v = 6) → `2`, stored value ends up in `{0,1}` and the update is always applied; repeated 5x with no drop and no error

Also added a plan test asserting every no-op `<=>` compares against a bare new-image column reference (never a re-evaluated function), for both `loc = loc` and `loc = substring(loc,1,2)`; and the secondary-unique BVT now uses a differing incoming value so it asserts the materialized-value semantics.

XuPeng-SH (approved):
LGTM.

The implementation side looks consistent with the MySQL affected-rows contract you are targeting:

- REPLACE and ODKU now opt into counting the DELETE side where appropriate.
- The ODKU no-op guard is placed early enough to compare old-image vs final written value, and it correctly skips implicit ON UPDATE / derived generated-column noise.
- The new flag is threaded through dupOperator and remote pipeline serialization.

The only things I still see are non-blocking coverage gaps: the new SQL/BVT file does not directly assert `ROW_COUNT()` yet, and the partition/remote propagation is mostly field-copy/round-trip tested rather than query-level affected-row assertions. Given the code path itself and the existing unit/planner/operator coverage, I don't see a substantive correctness blocker here.

XuPeng-SH (approved):
LGTM.

The implementation side looks consistent with the MySQL affected-rows contract you are targeting:

- REPLACE and ODKU now opt into counting the DELETE side where appropriate.
- The ODKU no-op guard is placed early enough to compare old-image vs final written value, and it correctly skips implicit ON UPDATE / derived generated-column noise.
- The new flag is threaded through dupOperator and remote pipeline serialization.

I did not find a substantive correctness blocker in the code path itself. The remaining gaps are test-only: the new SQL/BVT file still does not directly assert `ROW_COUNT()`, and the partition/remote propagation is mostly field-copy/round-trip tested rather than query-level affected-row assertions. Those are worth adding, but they are not enough on their own to block this PR.

aunjgr (approved):

ouyuanning (approved):

Issue #24908, opened 2026-06-09
[Bug]: Flaky TestHivePartitionListCache_Singleflight and possible race condition it discovers.


fengttt:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

4.0-dev, main

### Commit ID

0ca7df4

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

TestHivePartitionListCache_Singleflight flaked because the cache's
single-flight had a race: a goroutine could check the cache
  (miss) and then stall before joining the inflight set. Meanwhile the
leader finishes — populates the cache and deletes the
  inflight entry — so the stalled goroutine becomes a second leader
and re-runs listDir. In production that's a redundant directory
  listing; in the test it double-closes the started channel → panic.

### Expected Behavior

Flaky test

### Steps to Reproduce

```Markdown
Unit test, it is flaky.
```

heni02:
现在ut没有failed，closed

PR #25319, opened 2026-07-01
fix: close hive partition cache singleflight race

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24908 

fix: close hive partition cache singleflight race


ouyuanning (approved):

aunjgr (approved):
Clean race fix. Single lock acquisition for cache-check + inflight-start eliminates the gap where another goroutine could populate the cache between the two operations. LGTM.

XuPeng-SH (approved):
LGTM.

The fix closes the actual check-then-start gap by making cache lookup and inflight registration one atomic operation under the cache mutex. That prevents the interleaving described in #24908 where a goroutine observes a miss, stalls before joining inflight, then becomes a second leader after the first leader has already populated the cache and removed the inflight entry.

I checked the unhappy paths around the singleflight lifecycle:
- cache hit returns cloned entries and does not join inflight;
- followers either receive `flight.done` or return on `ctx.Done()`;
- leader success and error both publish through `finishInflight`, delete the inflight entry, and close `done`;
- cache `set` and `finishInflight` remain under the same cache mutex, so no additional map race is introduced.

Local verification with the proper CGo environment:
- `go test ./pkg/sql/colexec/external -run 'TestHivePartitionListCache|TestHivePartitionDirectPrefix|TestHivePartitionListCacheKey' -count=1`
- `go test ./pkg/sql/colexec/external -run 'TestHivePartitionListCache_Singleflight' -count=20`
- `go test -race ./pkg/sql/colexec/external -run 'TestHivePartitionListCache_Singleflight' -count=5`

XuPeng-SH (approved):
LGTM.

The fix closes the actual check-then-start gap by making cache lookup and inflight registration one atomic operation under the cache mutex. That prevents the interleaving described in #24908 where a goroutine observes a miss, stalls before joining inflight, then becomes a second leader after the first leader has already populated the cache and removed the inflight entry.

I checked the unhappy paths around the singleflight lifecycle:
- cache hit returns cloned entries and does not join inflight;
- followers either receive `flight.done` or return on `ctx.Done()`;
- leader success and error both publish through `finishInflight`, delete the inflight entry, and close `done`;
- cache `set` and `finishInflight` remain under the same cache mutex, so no additional map race is introduced.

Local verification with the proper CGo environment:
- `go test ./pkg/sql/colexec/external -run 'TestHivePartitionListCache|TestHivePartitionDirectPrefix|TestHivePartitionListCacheKey' -count=1`
- `go test ./pkg/sql/colexec/external -run 'TestHivePartitionListCache_Singleflight' -count=20`
- `go test -race ./pkg/sql/colexec/external -run 'TestHivePartitionListCache_Singleflight' -count=5`

PR #25322, opened 2026-07-01
fix: close hive partition cache singleflight race

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24908 

fix: close hive partition cache singleflight race

### DIFF
--- a/pkg/sql/colexec/external/hive_partition_cache.go
+++ b/pkg/sql/colexec/external/hive_partition_cache.go
@@ -111,14 +111,8 @@ func listHivePartitionDir(
 	}
 
 	key := options.CacheKeyPrefix + "\x1f" + "prefix=" + normalizeExternalPath(prefix)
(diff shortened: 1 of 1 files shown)


XuPeng-SH (approved):
LGTM.

I did not find a substantive correctness blocker in this fix. The change makes the cache-hit vs inflight leader/follower decision happen under a single lock, which is exactly the right direction for the race described in #24908.

Only non-blocking note: the new/updated singleflight test is still somewhat probabilistic. A more targeted regression seam that forces the exact interleaving would make the coverage stronger, but I would not block merge on that.

ouyuanning (approved):

aunjgr (approved):

Issue #25323, opened 2026-07-01
[Tech Request]: Support offline checkpoint catalog inspection, CSV dump, and restore script generation


jiangxinmeng1:
### Is there an existing issue for the same tech request?

- [x] I have checked the existing issues.

### Does this tech request not affect user experience?

- [x] This tech request doesn't affect user experience.

### What would you like to be added ?

```Markdown
## Background

  MatrixOne checkpoint data contains enough catalog metadata and object references to recover table schema and logical rows, but the current offline tooling is not sufficient for practical data inspection or restore
  workflows.

  Current pain points:

  - It is hard to list databases, tables, accounts, and views from checkpoint metadata without starting a MatrixOne instance.
  - Table schema reconstruction from checkpoint catalog tables is fragile, especially across catalog layout changes.
  - User table data cannot be reliably exported as `LOAD DATA` compatible CSV.
  - Partition tables, indexes, foreign keys, array/enum/set/time/year/fulltext/vector-related DDL, external tables, and views need better checkpoint restore coverage.
  - Remote checkpoint data stored in S3/MinIO cannot be conveniently inspected or dumped through the same tool workflow.
  - Large checkpoint dump output should avoid buffering all rows in memory.
  - Object storage multipart upload and retry behavior can make checkpoint dump to remote storage unstable.

  ## Goal

  Enhance `mo-tool ckp` so checkpoint data can be inspected and exported offline from either local `mo-data` or remote object storage, and make the exported result directly usable for restore validation.

  ## Scope

  Add or improve the following capabilities:

  - `ckp list`
    - list checkpoint catalog tables
    - list databases
    - list accounts
    - filter by `--account-id`, `--database-id`, and `--ts`

  - `ckp show-create-table`
    - reconstruct table DDL from checkpoint catalog metadata
    - support snapshot timestamp selection
    - preserve important DDL attributes where available

  - `ckp dump`
    - dump a single table by `--table-id`
    - dump all tables in a database by `--database-id`
    - dump all tables under an account by `--account-id`
    - support `--output-dir`, `--jobs`, `--header`, `--meta-comments`, and `--row-order`
    - support `--load-script` to generate `restore.sql`
    - support `--no-load` for DDL-only restore scripts

  - CSV / restore compatibility
    - output MatrixOne `LOAD DATA` compatible CSV
    - use `\N` for NULL values
    - preserve column order and skip hidden system columns
    - stream storage-order output for large tables
[...]

PR #25324, opened 2026-07-01
Support offline checkpoint catalog inspection, CSV dump, and restore script generation

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25323

Support offline checkpoint catalog inspection, CSV dump, and restore script generation


fengttt (approved):

XuPeng-SH (changes requested):
Requesting changes. The feature direction is useful, but I found several correctness gaps in the new dump/restore and multipart paths that should be closed before merge.

1. `cmd/mo-object-tool/ckp/checkpoint.go:553` drops the output `Close()` error for single-table dumps. For local files this can lose a flush error; for `--out-s3/--out-backend` it is much more serious because `fileServiceWriteCloser.Close()` is where the background `fs.Write` / multipart completion result is returned. Today `DumpTableCSVComposed` can succeed, `CompleteMultipartUpload` can fail, and the command still prints `Table ... dumped`. Please close explicitly after the dump and combine dump/close errors the same way `dumpOneTable` already does.

2. `pkg/tools/checkpointtool/table_dump.go:1185` silently ignores every `GetObjectEntries` error. That turns checkpoint corruption, permission errors, remote read failures, or decode errors into either a partial export or a misleading `no data entries for table` error. This should only skip the known GC/missing-file case (`isDataFileNotFound`); all other errors need to be returned/aggregated. Please add an unhappy-path regression test for a non-missing checkpoint entry error.

3. `pkg/tools/checkpointtool/checkpoint_reader.go:424` does not actually pick the latest usable GCKP. `loadEntries` sorts entries newest-first, but `ComposeAt` iterates from `len(entries)-1` down and breaks on the first `end <= ts`, which is the oldest qualifying GCKP. That can compose a snapshot from an unnecessarily old base plus too many incrementals, and can produce wrong catalog/table state when multiple global checkpoints exist. Iterate newest-to-oldest, skipping only GC'd bases, and cover this with a two-GCKP test.

4. `cmd/mo-object-tool/README.md:100` documents `--row-order=storage` as object/block/row scan order, but `pkg/tools/checkpointtool/table_dump.go:2065` processes objects concurrently and `writeCSVChunks` writes chunks in completion order. The default output is therefore nondeterministic and no longer matches the documented storage-order contract. Either preserve order with a small reorder buffer keyed by `(objectIdx, blockIdx)` or change the contract/flag name and tests to make unordered streaming explicit.

5. `pkg/fileservice/aws_sdk_v2.go:713` and `pkg/fileservice/qcloud_sdk.go:566` can resend the same pending buffer after `sendJob` fails. On context cancellation, max-part overflow, or worker-triggered cancellation, `sendJob` already records the error and may return the buffer to the pool, but the loop breaks with `pendingN > 0`, so lines 724/577 call `sendJob` again for the same buffer. This can double-return buffers, advance `partNum` incorrectly, or continue upload state after cancellation. Please clear `pendingN`/return immediately on `sendJob` failure and add tests that cancel/fail after at least one full pending chunk, not only before upload starts.

Local checks run:
- `make err-check`
- `go test -count=1 ./pkg/tools/checkpointtool ./cmd/mo-object-tool/ckp ./pkg/sql/colexec/external`
- `go test -count=1 ./pkg/tools/toolfs ./pkg/fileservice` with macOS CGo rpath adjusted for local dylibs

aptend (changes requested):
Request changes on current head `64bb0c0c1`. The previous review findings have fixes, but four merge-blocking correctness/resource issues remain.

### [P1] Include the first ICKP after the selected GCKP

[`shouldIncludeIncrementalCheckpoint`](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/pkg/tools/checkpointtool/checkpoint_reader.go#L519-L526) requires `ICKP.start > GCKP.end`. However, the checkpoint store explicitly uses [`maxGCKP.end` as the start of the first ICKP](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/pkg/vm/engine/tae/db/checkpoint/store.go#L124-L143).

A deterministic test with GCKP end=10 and ICKP 10→15 shows that `ComposeAt(15)` returns zero incrementals. Catalog changes, writes, and deletes recorded in that checkpoint are omitted from the dump. Please include the boundary ICKP and add a regression matching the real store semantics.

### [P1] Fail on a required checkpoint-chain or visible-object gap

[`ComposeAt`](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/pkg/tools/checkpointtool/checkpoint_reader.go#L494-L503) silently continues when a required ICKP file is missing. The CSV path similarly treats a missing visible data object as completed and continues.

I reproduced a usable GCKP followed by a required ICKP returning `ErrFileNotFound`; `ComposeAt` returned a nil error and a partial snapshot. Skipping an old GCKP while searching for another complete base can be valid, but skipping an ICKP or visible object inside the selected chain is not recoverable. It must fail rather than produce a successful incomplete backup.

### [P1] Create base tables before dependent views

The catalog result is [sorted by database/table name](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/pkg/tools/checkpointtool/table_dump.go#L1596-L1618), while [`orderTablesForRestore`](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/cmd/mo-object-tool/ckp/checkpoint.go#L976-L1035) only traverses foreign-key dependencies.

For `a_view AS SELECT * FROM z_table`, the generated script emits the view first and restore fails because `z_table` does not exist yet. A focused test reproduces this ordering. Please emit ordinary tables before views and order view chains by dependency or another creation-safe rule.

### [P1] Bound the storage-order reorder buffer

[`writeCSVChunks`](https://github.com/matrixorigin/matrixone/blob/64bb0c0c1576dcf5a183f33a600ba8d3a93765c9/pkg/tools/checkpointtool/table_dump.go#L2388-L2477) continuously drains the two-slot channel into an unbounded `pending` map for future objects. If object 0 is slow, chunks for later large objects accumulate without a byte/object bound.

Once free memory falls below the configured floor, the next required object-0 block can also stop at the memory gate. The writer cannot release later chunks until that block arrives, creating an OOM/hang cycle. Please enforce a bounded reorder window/byte budget with real producer backpressure.

Non-blocking follow-up: unknown-size empty remote output currently returns success without creating an object, so a headerless empty-table S3 dump can report success with no CSV key.

Local validation passed:
- ordinary tests for checkpointtool, objecttool, mo-object-tool, toolfs, and fileservice
- race tests for checkpointtool, ckp, toolfs, and fileservice
- go vet and git diff --check

XuPeng-SH (changes requested):
Request changes on current head `786e736810ade39d26ba31f05feff2beadecbaad`. Three of the four previous P1 findings are now closed (boundary ICKP inclusion, view-before-base ordering, and bounded storage-order buffering), but the required-checkpoint gap fix is incomplete.

### [P1] Do not skip a selected base/ICKP when the object-entry scan sees it missing

`ComposeAt` now correctly fails when a required ICKP is missing while building the table-range view. However, both consumers re-read the already-selected base/ICKP files and still swallow `ErrFileNotFound`:

- `pkg/tools/checkpointtool/table_dump.go:1190-1194` (`getTableEntriesAt`)
- `pkg/tools/checkpointtool/table_dump.go:1354-1358` (`PrepareTableDumpDataForTables`, the main `ckp dump` batch path)

This leaves a deterministic TOCTOU correctness hole: `ComposeAt` succeeds, the selected checkpoint file is GC'd/deleted (or the second read is the first one to observe the missing file), `GetObjectEntries{ForTables}` returns `ErrFileNotFound`, and these loops `continue`. Schema/catalog state can still be available, so the command can report success and emit an empty or partial CSV/restore set instead of failing the backup.

Once an entry is part of `BaseEntry + Incrementals`, it is required. Please return a contextual file-not-found error from both loops rather than skipping it. Add regressions using the existing object-entry test hooks where `ComposeAt` succeeds but `GetObjectEntries` / `GetObjectEntriesForTables` returns `ErrFileNotFound`; both the single-table and batch preparation paths should preserve that error.

aptend (changes requested):
Request changes on current head `786e736810ade39d26ba31f05feff2beadecbaad`.

The boundary-ICKP, view ordering, and storage-order buffering findings from my previous review are fixed. In addition to the selected-checkpoint missing-file issue already raised on this head, I found two further snapshot/restore correctness gaps.

### [P1] Compose a continuous checkpoint chain that covers the requested timestamp

`resolveSnapshotTS` accepts any timestamp between the earliest and latest checkpoint, but `shouldIncludeIncrementalCheckpoint` only includes an ICKP when `end <= ts`.

For a GCKP ending at 10, an ICKP `[10,20]`, and `--ts=15`, `ComposeAt(15)` excludes the ICKP completely. Consequently, catalog rows and objects created between 10 and 15 are absent before the later `CreateTime` / commit-TS / tombstone filters even run. A focused hook test expected one incremental and got zero.

The same loop also never validates temporal continuity. Given GCKP end 10, ICKP `[10,15]`, and then `[20,25]`, `ComposeAt(25)` returns success despite the missing interval. The checkpoint store contract makes the first ICKP start at `GCKP.end` and subsequent entries start at `previous.end.Next()`; silently accepting a gap produces an incomplete snapshot.

Please select ICKPs in chronological order, require a continuous chain from the selected base through the first ICKP that covers `ts`, and fail when the chain cannot cover the requested timestamp. If only checkpoint-end timestamps are intended to be supported, `resolveSnapshotTS` must reject other timestamps instead of returning a stale earlier state. Add regressions for both an in-ICKP timestamp and a missing middle ICKP.

### [P1] Do not generate an invalid restore script for cyclic foreign keys

`orderTablesForRestore` treats a DFS back-edge as success (`if visiting { return }`), while `restoreCreateTableDDL` renders every foreign key inline in `CREATE TABLE`.

For two valid tables with `A -> B` and `B -> A`, one table must be emitted first with an inline reference to a table that does not exist yet. With the default foreign-key checks, MatrixOne resolves the parent during `CREATE TABLE` and fails with `no such table`. The current unit test only checks that traversal terminates and returns two tables, so it codifies generation of an unusable `restore.sql`.

Please use a two-phase restore for cyclic dependencies: create tables without foreign keys, load data, then add constraints with `ALTER TABLE`. At minimum, detect the cycle and fail script generation rather than reporting success with invalid SQL. Add a regression that validates executable statement ordering, not only the number of returned tables.

Local validation on this head:

- Existing focused checkpoint composition/object-entry/restore-order tests pass.
- The in-ICKP snapshot regression fails with zero incrementals.
- The missing-middle-ICKP regression fails with `err == nil`.
- `git diff --check` passes; current required CI jobs are green.

XuPeng-SH (changes requested):
Request changes on current head `bff675930645ad8c9dfee682081aa50e138aa049`.

The latest two commits close the previously reported selected-checkpoint TOCTOU, in-ICKP timestamp/chain-gap, and cyclic-FK ordering issues. I re-reviewed the full ownership/concurrency/restore contracts and found three remaining merge-blocking gaps.

### [P1] Preserve the referenced database in foreign-key DDL

`TableForeignKey.ReferDatabase` is populated from catalog metadata and is used when ordering cross-database dependencies, but `renderForeignKeyClause` drops it and emits only `REFERENCES <table>` (`pkg/tools/checkpointtool/table_dump.go:865-867`). This affects both inline foreign keys and the new deferred `ALTER TABLE` path for cycles.

For a child in `child_db` referencing `parent_db.parent`, the generated script runs under `USE child_db` and emits `REFERENCES parent`. MatrixOne therefore resolves `child_db.parent`, so restore fails or binds to the wrong same-named table. Cross-database foreign keys are supported and already covered elsewhere in the repository (for example `REFERENCES fk_test1.t1(a)` and `REFERENCES db1.t1(a)`).

I reproduced this on the exact head with a focused `RenderAddForeignKeyDDL` test. The expected clause was ``REFERENCES `parent_db`.`parent```; the actual clause was ``REFERENCES `parent```. Please render a qualified referenced relation (at least whenever `ReferDatabase` is non-empty) and add restore-script tests for both non-cyclic inline and cyclic/deferred cross-database FKs.

### [P2] Share the cache-fill result, not only `sync.Once` completion

`lazyCacheFS.ensureCached` stores a `*sync.Once`, but stores the fill error only in each caller's local `cacheErr` (`pkg/tools/toolfs/lazy_cache_fs.go:148-158`). A concrete interleaving is:

1. A and B obtain the same `Once` for an uncached path.
2. A executes `cacheFile`; B waits in (or reaches) `Do`.
3. A's remote read fails, A deletes the map entry, and A returns the error.
4. B observes the `Once` as done, its closure is not run, and its own `cacheErr` is still nil.

`PrefetchFile` therefore reports success to B although no cache file exists. The race detector cannot find this because the memory synchronization is legal; the protocol result is wrong. Please use a shared per-path in-flight record containing completion plus the final error (for example `done` + `err`), and make every waiter observe the same result. Add a channel-controlled two-caller failure regression without sleeps, plus retry-after-failure coverage.

### [P2] Roll back owned file services when construction fails

The new constructors acquire ownership before all fallible initialization is complete, but return no object and do not close that ownership on failure:

- `checkpointtool.Open` creates an offline FS, sets `closeFS=true`, then returns directly if `loadEntries` fails (`pkg/tools/checkpointtool/checkpoint_reader.go:119-132`). The caller receives no reader and cannot call `Close`.
- Both S3/MINIO branches in `toolfs.Open` / `openFromConfig` create a remote FS and return directly if `newLazyCacheFS` fails (`pkg/tools/toolfs/storage.go:73-81`, `109-118`).

Please make initialization transactional: keep a rollback close active until the fully initialized owner is returned, and transfer ownership exactly once on success. Add injected-failure tests that assert close-on-failure and no close-on-success/double-close.

Validation performed on the exact head:

- related ordinary package tests passed: checkpointtool, ckp CLI, objecttool, objecttool/interactive, toolfs, fileservice, and external
- race tests passed three times for checkpointtool, ckp CLI, toolfs, and fileservice
- `go vet`, `make err-check`, and `git diff --check` passed
- required GitHub CI is green; the PR is currently behind `main`

gouhongshen (changes requested):
## Codex automated review

Found three blocking correctness issues in dump/restore behavior and one compatibility regression.

### P1 - Cross-database foreign keys are rendered without the referenced database (pkg/tools/checkpointtool/table_dump.go:866)
Foreign-key decoding preserves ReferDatabase, but renderForeignKeyClause emits only ReferTable. Both inline CREATE and deferred ALTER statements therefore resolve cross-database references in the current database, causing restore failure or silently linking the wrong table.

### P1 - FSConfig-backed output produces unusable LOAD DATA paths (cmd/mo-object-tool/ckp/checkpoint.go:2183)
Remote output configured through FSConfig is written via the configured fileservice, but newLoadDataPathResolver only handles storage.S3 and returns a local-path resolver otherwise. The generated restore script consequently points LOAD DATA INFILE at local absolute paths even though the CSVs are remote.

### P1 - Schema-evolved objects can lose values for newly added columns (pkg/tools/checkpointtool/logical_table.go:110)
scanLogicalTable sorts objects by name and permanently uses the first object's column sequence numbers as the canonical set. If an older object is encountered before an object created after ADD COLUMN, the newer column is discarded during alignment. The metadata/lexical CSV path and logical view can therefore emit a schema column with missing values.

### P2 - Repurposing --s3 removes the existing S3FS-on-disk CLI mode (cmd/mo-object-tool/ckp/checkpoint.go:100)
Before this change, ckp and object used boolean --s3 to select OfflineKindS3, which still implements the local S3FS/raw format. The new flag is a string remote-storage option and kindFromFlags no longer selects OfflineKindS3, so existing --s3 /path invocations cannot inspect those checkpoints and have no replacement.

XuPeng-SH (changes requested):
Request changes on current head `ea1ac2e39`.

The latest commit closes the previously reported cross-database FK, FSConfig restore path, schema-evolution, CLI compatibility, shared-fill-error, and constructor rollback gaps. One cancellation/liveness blocker remains in the new lazy cache.

### [P2] A cache-fill waiter ignores its own context

At `pkg/tools/toolfs/lazy_cache_fs.go:172-177`, when another caller already owns the fill, the waiter executes `<-fill.done` unconditionally. Its own `ctx.Done()` is not part of the wait graph.

Concrete interleaving: caller A starts `PrefetchFile` and its remote read is slow or blocked; caller B reaches the same path with an already-canceled/deadline-expired context. B still cannot return until A completes. I reproduced this deterministically on this exact head with a channel-controlled remote `Read`: after B was canceled it remained blocked for the full assertion window and returned only after the test released A. This is legal synchronization, so `-race` remains green, but it violates the FileService cancellation contract and can leave a canceled checkpoint dump/read stuck behind an independent slow remote operation.

Please wait with a `select` on both `fill.done` and `ctx.Done()` (returning `ctx.Err()` for the latter), and also make the invalidation retry loop cancellation-aware. Add a phase-controlled regression proving the canceled waiter returns before the owner is released, while retaining the current shared-error and retry-after-failure coverage.

Validation on the exact head:
- ordinary tests passed for checkpointtool, ckp CLI, toolfs, objecttool, and fileservice
- race tests passed for checkpointtool, ckp CLI, toolfs, and objecttool
- `make err-check`, `git diff --check`, and a synthetic merge with current `origin/main` passed

gouhongshen (changes requested):
## Codex automated review

Four blocking correctness and resource-safety issues found.

### P1 - Do not silently exclude legal quoted identifiers from the catalog (pkg/tools/checkpointtool/table_dump.go:3590)
This rejects every database or table name containing a character outside the small ASCII allowlist. MatrixOne accepts arbitrary backtick-quoted identifiers (the MySQL scanner reads all contents between backticks), and the restore path already quotes identifiers. Consequently `ListCatalogTables` omits valid entries such as `order items`; batch dump/restore never processes them, and a single-table `--load-script` reports its ID as absent. Preserve catalog names and quote them when emitting SQL instead of treating valid names as corrupt rows.

### P1 - Do not package S3 external-table sources as local files (cmd/mo-object-tool/ckp/checkpoint.go:890)
Every external table is sent to `packageExternalTableSource`. S3 external-table rendering produces `URL s3option {..., 'filepath'='object/key'}`, but the packaging helper extracts that value and calls `os.Open` on it as though it were a local `INFILE` path. Thus generating a restore script for any valid S3 external table fails (including with `--no-load`, despite its DDL-only contract). Limit local copying to local INFILE sources and retain URL/S3 external definitions.

### P1 - Keep lazy-cache object paths inside the cache root (pkg/tools/toolfs/lazy_cache_fs.go:165)
`normalized` is used directly in `filepath.Join` without rejecting `..` components. A remote/checkpoint path such as `SHARED:../../target` escapes the temporary cache root; `cacheFile` then creates directories and renames downloaded bytes to that outside path. This can overwrite caller-writable files, and `Close` only removes the nominal cache root, leaving the escaped file behind. Validate the cleaned path remains relative to the cache root before every local cache operation.

### P1 - Restore a global bound on multipart workers and buffered parts (pkg/fileservice/aws_sdk_v2.go:597)
Replacing the shared upload pool with per-upload goroutines removes the only global limit. Each remote CSV dump forces parallel mode, starts `runtime.NumCPU()` workers, and has a `2*NumCPU` queue of 64-MiB part buffers. With the default five concurrent table dumps, slow object storage can retain roughly `5 * 3 * NumCPU * 64 MiB` of buffers (about 30 GiB on a 32-core host), causing OOM; QCloud has the same change. Use a bounded shared scheduler/semaphore for multipart workers rather than independently spawning a full CPU-sized upload pool per output.

aunjgr (changes requested):
Several blockers remain in the offline checkpoint tool: canceled lazy-cache waiters block unconditionally behind the owner; cache paths can escape the configured root through `..`; the identifier allowlist rejects valid quoted identifiers; S3 external-table sources are opened as local files; and per-table CPU-scaled workers plus 64 MiB buffers permit multi-table memory growth into tens of GiB. Please close cancellation, path confinement, identifier quoting, storage-kind dispatch, and global worker/memory budgeting before merge.

aunjgr (changes requested):
Several blockers remain in the offline checkpoint tool: canceled lazy-cache waiters block unconditionally behind the owner; cache paths can escape the configured root through `..`; the identifier allowlist rejects valid quoted identifiers; S3 external-table sources are opened as local files; and per-table CPU-scaled workers plus 64 MiB buffers permit multi-table memory growth into tens of GiB. Please close cancellation, path confinement, identifier quoting, storage-kind dispatch, and global worker/memory budgeting before merge.

aunjgr (changes requested):
Two blockers remain on head 3049ce46c.

P1: isLocalInfileExternalTableDDL searches the complete uppercased DDL for the substring INFILE. A valid S3 URL whose endpoint, bucket, or filepath contains that word is misclassified as a local source and packageExternalTableSource then calls os.Open on the object key. Determine the external scan type structurally and package only an actual local INFILE clause.

P1: the new semaphore bounds active upload goroutines but not buffered parts. Every WriteMultipartParallel caller reads the pending and next default 64 MiB chunks before it waits for the global token. The outer object-storage semaphore permits 1024 concurrent operations, so slow storage can still retain many gigabytes in blocked callers. Add a shared byte or buffer-token budget acquired before allocating and reading each part, and release it whenever the buffer is returned.

XuPeng-SH (changes requested):
Request changes on current head `d2da6df7513`.

The latest commits close the S3 URL / `INFILE` misclassification, but the resource-bound fix is incomplete and the full review found additional correctness and liveness blockers.

### [P1] Make the multipart byte budget match the allocation and acquire it atomically

[`acquireParallelUploadBufferBudget`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/fileservice/object_storage.go#L83-L102) caps the charged token count at the channel capacity, while [`readChunk`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/fileservice/aws_sdk_v2.go#L511-L528) still allocates the complete configured part. Since `PartSize` is allowed up to 5 GiB, one allocation can exceed the nominal global budget by gigabytes. A focused test with a 2 MiB budget and a 5 MiB request deterministically charged only 2 MiB.

The per-token send loop is also not an atomic weighted acquisition: concurrent callers can retain partial token sets while waiting for the remainder. Use an atomic weighted byte semaphore, and reject or cap any single part larger than the supported budget before allocation. Apply the same rule to AWS and QCloud.

### [P1] Enforce the lazy-cache limit instead of only accounting past it

[`reserveCacheSpace`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/toolfs/lazy_cache_fs.go#L375-L389) records the reservation even when eviction cannot make enough space. A single object larger than `maxBytes` is therefore downloaded and retained in full; concurrent fills can likewise accumulate reservations beyond the cap because in-flight reservations are not evictable or backpressured.

I reproduced this with a 32-byte cache and a 64-byte object: `PrefetchFile` succeeded and `usedBytes` became 64. The cache needs a real admission policy for oversized and concurrent fills, plus tests that assert physical cached bytes and reservations never exceed the configured bound.

### [P1] Do not wait for an impossible free-memory floor

[`csvPipelineMemoryFloorFromTotal`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/checkpointtool/table_dump.go#L2306-L2315) forces a minimum 1 GiB floor, and [`waitForCSVMemory`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/checkpointtool/table_dump.go#L2317-L2336) waits until available memory reaches it. On a host with 512 MiB total memory, the condition is mathematically unreachable; the CLI uses a background context with no default deadline and hangs before reading the first block. A focused invariant test produced floor=1 GiB for total=512 MiB.

Clamp the floor below total memory and ensure the one-worker progress path can always proceed or fail explicitly.

### [P1] Compose the selected checkpoint snapshot for the interactive logical view

[`SelectTable`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/checkpointtool/interactive/state.go#L141-L167) loads object entries only from the selected checkpoint entry, and [`LoadLogicalView`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/checkpointtool/interactive/state.go#L170-L184) treats that subset as the snapshot at the entry end timestamp. Selecting an ICKP therefore omits the GCKP and earlier ICKPs, so unchanged live rows are absent even though the README promises the selected checkpoint snapshot. Reuse the composed table dump data path and add an ICKP regression with live data inherited from its base.

### [P1] Bound and cancel interactive logical-table materialization

[`BuildLogicalTableView`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/checkpointtool/logical_table.go#L38-L64) converts every visible cell to a string and retains every row. The TUI invokes it synchronously with `context.Background()`, so pressing `L` on a large table can freeze the UI and exhaust memory with no user cancellation path. Use paging/streaming or an explicit row/byte limit and run the load through a cancelable Bubble Tea command.

### [P2] Free partially decoded vectors on object-read failures

In [`ReadBlock`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/objecttool/object_reader.go#L223-L239), a failure decoding column N releases only the I/O vector; vectors already stored in the batch are not cleaned. The new commit-TS path also returns on a type mismatch without [freeing the decoded vector](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/pkg/tools/objecttool/object_reader.go#L264-L273). Clean the partial batch / vector before returning and cover corrupt-column and wrong-type unhappy paths.

### [P2] Remove the unrelated checkpoint-frequency change

This feature PR changes [`launch-minio-local/tn.toml`](https://github.com/matrixorigin/matrixone/blob/d2da6df7513063be30e2d55623f9bdd48cfc45e9/etc/launch-minio-local/tn.toml#L57-L62) from 60s/300s checkpoint intervals to 10s/30s and sharply lowers thresholds. That changes runtime I/O and checkpoint behavior for every user of this launch config and is unrelated to offline tooling. Please revert it or move it to a separately justified change.

Validation on the exact head:

- ordinary tests passed for fileservice, toolfs, checkpointtool, both interactive packages, objecttool, and the ckp CLI
- race tests passed for the same packages
- `go vet` and `git diff --check` passed
- focused invariant tests failed exactly for multipart under-accounting, lazy-cache over-cap retention, and the unreachable low-memory floor
- required GitHub CI is still running; the PR is currently behind main

gouhongshen (changes requested):
## Codex automated review

Six blocking correctness, memory-safety, and liveness issues remain on the current head.

### P1 - Exclude generated columns from restore CSV input (pkg/tools/checkpointtool/table_dump.go:2006)
Both CSV paths export every schema column, including columns marked Generated, while restore.sql emits LOAD DATA without a column list. The load binder maps such input only to non-hidden, non-generated columns (pkg/sql/plan/bind_load.go:130-134), and the CSV reader rejects a nonempty extra field. Therefore a table (a,b,c GENERATED) is dumped as three fields but restored expecting two, so restore fails. Preserve generated columns in DDL but omit them from CSV/load input.

### P1 - Acquire multipart buffer capacity atomically and account for the full allocation (pkg/fileservice/object_storage.go:83)
The channel-based weighted acquisition retains partial token sets while waiting. Concurrent callers can each acquire part of the remaining capacity, fill the channel, and then wait forever because no caller reaches the release path. Additionally, requests larger than the budget are capped to cap(budget), although AWS and QCloud allocate the full permitted part (up to 5 GiB). Use an atomic weighted semaphore and reject/cap parts that cannot be fully charged.

### P1 - Enforce the configured lazy-cache limit before downloading (pkg/tools/toolfs/lazy_cache_fs.go:375)
reserveCacheSpace records the full reservation even when eviction cannot make enough room. A file larger than maxBytes is therefore downloaded and committed, and concurrent fills can also reserve beyond the cap because in-flight reservations are not evictable or backpressured. This defeats the cache's disk/memory bound; admission must fail or wait unless the complete reservation fits.

### P1 - Do not impose an unreachable CSV memory floor (pkg/tools/checkpointtool/table_dump.go:2311)
csvPipelineMemoryFloorFromTotal returns at least 1 GiB even when total system memory is below 1 GiB. waitForCSVMemory then requires available memory to reach that impossible value. Since dump commands use a background context without a deadline, a low-memory host hangs before processing the first object. Clamp the floor below total memory or fail explicitly.

### P1 - Compose the selected ICKP before building its logical table view (pkg/tools/checkpointtool/interactive/state.go:144)
SelectTable reads object entries only from the selected checkpoint entry, then LoadLogicalView treats those entries as the complete snapshot at that entry's end timestamp. For an incremental checkpoint this omits live objects and tombstones inherited from the base GCKP and earlier ICKPs, producing an incomplete logical view. Reuse the composed snapshot/table-dump path.

### P1 - Bound and make interactive logical-table loading cancellable (pkg/tools/checkpointtool/logical_table.go:61)
BuildLogicalTableView converts and retains every visible cell in view.Rows. The TUI invokes it synchronously with context.Background(), so opening a large table can freeze the event loop and exhaust memory, and quitting cannot cancel the scan. Use paging/streaming or an explicit row/byte limit through a cancellable Bubble Tea command.

gouhongshen (changes requested):
## Codex automated review

Found one blocking regression in the advertised DDL-only restore path.

### P1 - Honor --no-load for local external tables (cmd/mo-object-tool/ckp/checkpoint.go:889)
This packages every local INFILE external-table source before considering includeLoad. Thus --load-script --no-load still invokes packageExternalTableSource: remote dump output always fails there, and local output fails if the original source file is unavailable. That contradicts --no-load's documented DDL-only behavior and prevents offline DDL script generation for valid external-table catalogs. Gate source packaging on includeLoad and cover the no-load local-INFILE case, including remote output.

XuPeng-SH (changes requested):
Requesting changes on current head `4a2ef353b5d5`. The latest commit fixes the prior `--no-load` blocker, but four merge-blocking correctness, liveness, and resource-lifecycle issues remain.

### P1 - Emit block completion or consume the now-ordered CSV stream directly

The producer sends no event for a block whose rows are all deleted or filtered by the snapshot; it sends only `objectDone` after the complete object. The writer, however, waits for the missing `nextBlockIdx` until `objectDone` before advancing. Consequently, one empty middle block retains every later chunk from that object and a valid large export fails once the 512 MiB reorder budget is reached.

Relevant paths:
- `pkg/tools/checkpointtool/table_dump.go:2234-2249`
- `pkg/tools/checkpointtool/table_dump.go:2488-2515`
- `pkg/tools/checkpointtool/table_dump.go:2534-2550`

The producer is now strictly ordered, so either consume that ordering directly or emit an explicit `blockDone` event. Add a regression with an empty middle block followed by more than the reorder budget of visible rows.

### P1 - Give the CSV memory gate a guaranteed terminal outcome

`waitForCSVMemory` exits only when available memory reaches the floor or its context is canceled. Dump commands invoke it under `context.Background()`, so a host that remains below the floor hangs forever before reading the next block. The latest floor clamp fixes the mathematically impossible low-total-memory case but does not provide progress or a bounded failure when memory pressure persists.

Relevant paths:
- `pkg/tools/checkpointtool/table_dump.go:2332-2351`
- `cmd/mo-object-tool/ckp/checkpoint.go:564-573`

Use a bounded wait / explicit low-memory error or a progress rule that guarantees one worker can continue, and cover sustained pressure without relying on external cancellation.

### P1 - Build catalog maps once for database/account batch dumps

`PrepareTableDumpDataForTables` calls `ReadTableSchema` inside the table loop. Each call rematerializes the complete `mo_tables` and `mo_columns` logical views and rereads partition metadata; restore generation also rematerializes `mo_indexes` per table. An N-table database/account dump therefore approaches O(table count × catalog size), contrary to the advertised single-compose/single-scan batch path, and can make the offline tool unusable on large catalogs.

Relevant paths:
- `pkg/tools/checkpointtool/table_dump.go:1337-1355`
- `pkg/tools/checkpointtool/table_dump.go:1100-1131`
- `pkg/tools/checkpointtool/table_dump.go:4061-4079`

Materialize schema, partition, and index maps once and share them across tables; add a call-count or large-catalog regression.

### P2 - Join an in-flight logical-table load before closing its reader and file service

Pressing `L` runs logical-table loading in a Bubble Tea Cmd goroutine. Pressing `q` only cancels it; Bubble Tea does not join long-running Cmd goroutines. `interactive.Run` can therefore return and let outer defers clear reader entries and close the file service while the load is still in `ComposeAt` or storage I/O. This permits close-after-use races and a leaked worker when downstream I/O does not return promptly.

Relevant paths:
- `pkg/tools/checkpointtool/interactive/unified_model.go:188-198`
- `pkg/tools/checkpointtool/interactive/unified_model.go:228-247`
- `cmd/mo-object-tool/ckp/checkpoint.go:139-153`

Track load completion and wait for it after cancellation before releasing reader/FS ownership. Cover quit-during-blocked-load under `-race`.

Validation on this exact head: the related eight packages pass ordinary tests, six core packages pass `-race`, and `go vet`, `err-check`, `git diff --check`, and required GitHub checks pass. Those tests do not cover the protocol and lifecycle boundaries above.

aptend (changes requested):
Requesting changes on current head `9b042556bdf1`.

### P1 - Read `mo_indexes` from the target tenant

The batch path calls `catalogTableViewFromIndex` with account 0 for every selected table, and the single-table path uses `findCatalogTableID`, which also resolves account 0. However, `mo_catalog.mo_indexes` is created separately for every tenant and therefore has a distinct table ID per account. For a tenant table, its secondary, fulltext, and vector-index rows live in that tenant-specific `mo_indexes`, not the system-account table.

As a result, dumping an account or database can mark `IndexesPrepared` after scanning the system `mo_indexes`, then silently emit restore DDL without the tenant indexes. The existing test masks this by placing the index row for an account-1 user table in an account-0 `mo_indexes` view.

Resolve and cache `mo_indexes` per `Schema.AccountID` in the batch path, and make the single-table lookup account-aware as well. Add a regression where account 0 and account 1 have different `mo_indexes` table IDs and the index row exists only in the account-1 view.

The four blockers from the previous review are otherwise addressed. Ordinary tests, focused race tests, `go vet`, and `git diff --check` passed on this exact head; three GitHub CI jobs were still running at review time.

aunjgr (approved):
Reviewed the current head independently across checkpoint-chain composition, catalog/schema reconstruction, CSV streaming/order, remote multipart ownership, lazy-cache admission, restore ordering, and interactive shutdown. Required checkpoint/object gaps now fail closed; account-scoped catalog tables (including mo_indexes) are resolved per tenant; CSV production is storage-ordered with bounded/cancellable memory waits; multipart buffers use atomic weighted admission with one release owner; and logical-view workers are canceled and joined before reader/FS teardown. I found no remaining merge-blocking code issue on this head. The arm64 SCA check is still failing and must be resolved before merge.

aptend (approved):

gouhongshen (approved):
## Codex automated review

No merge-blocking regression found; two non-blocking robustness/resource issues remain.

### P2 - Commit-timestamp decode can panic on malformed object data (pkg/tools/objecttool/object_reader.go:287)
ReadBlock guards objectio.Decode, but ReadBlockCommitTS calls it directly. Short or unknown-codec payloads panic instead of returning an error and releasing the I/O vector.

### P2 - Per-object memory pools are never released (pkg/tools/objecttool/object_reader.go:327)
OpenWithFS allocates an mpool per object, while Close is a no-op. These pools remain registered globally after the new logical/CSV paths open objects, causing process-lifetime resource growth.

XuPeng-SH (changes requested):
这轮 latest 我这里还有 3 个需要处理的点，前 2 个是 merge blocker：

1. `cmd/mo-object-tool/ckp/checkpoint.go` + `pkg/fileservice/{s3_fs.go,aws_sdk_v2.go,qcloud_sdk.go}`
远端 dump 统一走 unknown-size + `ParallelAuto`。空表在默认 `--header=false --meta-comments=false` 下会产生 0-byte 输出；AWS/COS 的并行 multipart 在第一次 read 就 `io.EOF` 时直接返回 success，不会创建对象。
但 CLI 仍然会报成功，`writeRestoreScript` 也会无条件写出这份 CSV 的 `LOAD DATA` 路径，最终 restore 会指向一个根本不存在的对象。
建议：`firstPart == nil && io.EOF` 时回退到普通 0-byte write/put，确保空对象被创建。

2. `cmd/mo-object-tool/ckp/checkpoint.go` + `pkg/tools/checkpointtool/table_dump.go`
CSV 生成如果在写出部分 header/rows 后失败，当前路径还是 `outFile.Close()` 正常收尾；对远端 fileservice 这会把 pipe EOF 当成成功结束，最终落下截断对象，命令本身却返回 error。
这不是单纯报错体验问题，而是把坏数据持久化到了目标存储。
建议：给 remote dump writer 增加 `CloseWithError`/`Abort` 语义，dump 失败时走 abort，别 finalize partial object。现有 `pkg/fileservice/file_service_writer.go` 已经有同类约定可复用。

3. `pkg/tools/toolfs/lazy_cache_fs.go`
`StatFile` 失败时直接预留 `l.maxBytes`；batch dump 是多 worker + shared FS，活跃 fill 期间这个 full-budget reservation 会让并发读收到假的 `lazy cache capacity exhausted`。
建议：stat 失败时不要直接吃满整个预算，改成 bounded fallback reservation，或者读完后按实际大小校正。

所以这轮我先 request changes。

XuPeng-SH (changes requested):
Requesting changes on exact head `0fbc6e8efce1`. The latest two commits correctly fix the prior empty-object, partial-remote-write, and lazy-cache reservation issues; the earlier CSV ordering, bounded memory wait, tenant catalog, `--no-load`, and logical-load shutdown findings remain closed. Two full-diff object-reader failures still need correction.

### P1 — Release every per-object mpool (`pkg/tools/objecttool/object_reader.go:125,327`)

`processCSVObjectChunks` opens one `ObjectReader` for every visible object, but `ObjectReader.Close()` is a no-op. Every `mpool.MustNewZero()` therefore remains in `mpool.globalPools`, so a large checkpoint dump accumulates O(object-count) pools and maps for the life of the process. A focused regression that creates a uniquely tagged pool, calls `Close`, and checks `mpool.ReportMemUsage(tag)` currently leaves the pool registered. Delete the pool exactly once in `Close` and cover repeated close plus normal block release.

### P2 — Decode commit timestamps through the panic-safe path (`pkg/tools/objecttool/object_reader.go:287`)

`objectio.Decode` calls `DecodeIOEntryHeader`, which panics for payloads shorter than four bytes. `ReadBlock` already uses `decodeObjectColumn` to convert corrupt data into an error, but `ReadBlockCommitTS` bypasses it. A malformed commit-TS extent therefore crashes the offline process and skips `releaseIOVector`. Reuse the safe decoder (or equivalent recovery) and add a corrupt commit-TS regression.

Related eight-package tests, focused race stress, build, vet, `git diff --check`, and current CI are green; those suites do not cover these two boundaries.

gouhongshen (approved):
## Codex automated review

No merge-blocking regression found. Two non-blocking issues remain.

### P2 - Release internally owned CheckpointReader memory pools (pkg/tools/checkpointtool/checkpoint_reader.go:673)
Open/OpenWithFS allocate a default MPool and Fork always allocates one, but Close only clears entries and optionally closes the FS. MPool registration is global and removed only by mpool.DeleteMPool, so every batch worker (created at checkpoint.go:2245) and failed post-allocation open remains reachable for the process lifetime. Track pool ownership, delete internally created pools exactly once in Close/error rollback, and close worker forks; never delete a WithMPool caller-owned pool.

### P2 - Remove the unrelated checkpoint cadence change (etc/launch-minio-local/tn.toml:61)
This feature PR changes the standard local-Minio incremental checkpoint interval from 300s to 180s. The offline inspection/dump tooling does not consume this setting, while the change increases checkpoint work for every user of this launch configuration. Revert it or move it to a separately justified configuration change.

XuPeng-SH (changes requested):
Requesting changes on exact head `be574e46177a`. The latest commit correctly closes the two prior `ObjectReader` findings: its internally created pool is deleted exactly once after normal block release, and corrupt commit-TS payloads now go through the panic-safe decoder. The earlier CSV ordering/memory, tenant catalog, remote-write abort/empty-object, lazy-cache, `--no-load`, and interactive-shutdown fixes also remain closed.

### P2 — Release every internally owned `CheckpointReader` memory pool (`pkg/tools/checkpointtool/checkpoint_reader.go:138-143,168-173,184-198,672-681`)

`Open` and `OpenWithFS` create a default `mpool.MustNewZero()` when the caller does not provide one, while `Fork` always creates another pool. Those pools are registered in `mpool.globalPools` and are removed only by `mpool.DeleteMPool`, but `CheckpointReader.Close` clears only entries/the optional file service. The constructor error defers likewise roll back only the file service. In addition, each `dumpTablesConcurrently` worker creates a fork at `cmd/mo-object-tool/ckp/checkpoint.go:2245` and never closes it. Repeated opens, failed opens, or repeated batch dumps in one process therefore retain pools/maps for the process lifetime.

A focused probe on this head counted the `must_new_zero` registrations: `Fork` increased the count from 0 to 1, and `fork.Close()` left it at 1 (expected 0). Please track pool ownership, delete only internally created pools exactly once in `Close` and constructor rollback, preserve caller ownership for `WithMPool`, and close each worker fork. Add regressions for default/open failure/fork close, repeated close, and caller-owned pools.

XuPeng-SH (changes requested):
Requesting changes on exact head `978ae572d1a3`. The latest commit correctly closes the prior `CheckpointReader` lifecycle blocker: default pools are owned and rolled back on constructor failure, caller-provided `WithMPool` pools remain caller-owned, forks own independent pools, repeated close is idempotent, and every batch worker now closes its fork. The earlier CSV ordering/backpressure, tenant catalog, remote empty/partial-object, lazy-cache, object-reader, `--no-load`, and interactive-shutdown findings remain closed.

### P2 — Restore the unrelated local-Minio checkpoint cadence (`etc/launch-minio-local/tn.toml:61`)

The full diff still changes `incremental-interval` from the current base value `300s` to `180s`. This offline inspection/dump feature does not consume that launch setting, while the shorter interval changes TN checkpoint behavior for every `launch-minio-local` user and can increase checkpoint work. Commit `928fb13bbc` attempted to undo the PR's earlier aggressive checkpoint settings but restored the stale `180s` value instead of the current base value `300s`. Please revert this line to `300s` (and preferably restore the unrelated removed comment in `etc/launch-minio-local/log.toml` as scope cleanup).

Verification on this exact head: uncached CGo-transitive tests passed for `./pkg/tools/checkpointtool` and `./cmd/mo-object-tool/ckp`; `git diff --check` passed. GitHub UT/SCA and several E2E jobs were still running at review time.

XuPeng-SH (approved):
Approving exact head `b26c44df9e7d`. The prior local-Minio configuration blocker is closed: `incremental-interval` is restored to the base value `300s`, and the unrelated `log.toml` comment is restored. The subsequent main merge has one true two-parent conflict surface, `pkg/sql/colexec/external/external_test.go`; the resolution preserves both main's Iceberg profile test and this PR's CSV/type regressions. I found no semantic regression from the merge, and all earlier findings remain closed.

Q1–Q3 closure: internally owned reader/object pools and worker forks have one idempotent cleanup owner while caller-owned pools remain untouched; cancellation/worker waits terminate through context cancellation plus channel close/join; lazy-cache reservations, multipart buffers, and CSV pipeline memory remain bounded. No open blocker remains.

Fresh verification on this exact head: uncached CGo-transitive tests passed for `./pkg/tools/checkpointtool`, `./cmd/mo-object-tool/ckp`, and `./pkg/sql/colexec/external`; focused lifecycle tests passed under `-race`; `go vet`, `go build`, and `git diff --check` passed. GitHub UT/SCA/coverage and several E2E jobs are still running and should remain required merge gates.

Issue #25284, opened 2026-07-01
[Bug]: COUNT(DISTINCT expr, expr) is not supported and differs from MySQL NULL semantics


Ariznawlll:
### Is there an existing issue for the same bug?

I searched existing issues for `COUNT DISTINCT multiple arguments`, `count(distinct g, v)`, and `aggregate function count bad value`, and did not find a duplicate.

### MatrixOne version

Verified on latest `main` at `200bfa7e0efb` (`fix: improve skills SKILL.md compatibility format (#25271)`).

### MySQL version

Local MySQL 8.3.0.

### Reproduce steps

```sql
drop database if exists repro_count_distinct_multi;
create database repro_count_distinct_multi;
use repro_count_distinct_multi;

create table t(g int, v int, txt varchar(8));
insert into t values
  (null, null, 'n1'),
  (null, 10,   'n2'),
  (1,    null, 'a'),
  (1,    5,    'b'),
  (1,    5,    null),
  (2,    null, null),
  (2,    7,    'c'),
  (3,    null, null);

select count(distinct g, v) as cnt_g_v,
       count(distinct g, txt) as cnt_g_txt
from t;
```

### Expected behavior

MySQL supports `COUNT(DISTINCT expr, expr, ...)`. It counts distinct non-NULL expression combinations and ignores rows where any expression in the combination is `NULL`.

MySQL result:

```text
cnt_g_v  cnt_g_txt
2        3
```

### Actual behavior

MatrixOne errors:

```text
ERROR 20203 (HY000): invalid argument aggregate function count, bad value [INT INT]
```

A related syntax already present in BVT, `count(distinct(a, b))`, also reports:

```text
internal error: tuple is not supported as group by column
```

### Additional information

This was found while exploring MySQL compatibility for NULL aggregation/grouping semantics. Passing MySQL-aligned single-expression `COUNT(DISTINCT col)` cases were added separately; this multi-expression case is excluded from passing BVT until fixed.

Ariznawlll:
Verified on latest `main` at `05950d7fc` (`fix: support COUNT(DISTINCT col1, col2, ...) multi-column distinct (#25325)`).

Validation:

- Re-ran the original repro for `COUNT(DISTINCT g, v)` and `COUNT(DISTINCT g, txt)`.
- Added grouped multi-column distinct coverage with NULL combinations.
- Added three-column `COUNT(DISTINCT g, v, txt)` coverage.
- Compared against local MySQL `8.0.45`.

[...]

PR #25325, opened 2026-07-01
fix: support COUNT(DISTINCT col1, col2, ...) multi-column distinct

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25284

Support `COUNT(DISTINCT col1, col2, ...)` multi-column distinct count, matching MySQL semantics:

- `COUNT(DISTINCT col1, col2)` — counts distinct non-NULL combinations
- `COUNT(DISTINCT (col1, col2))` — tuple syntax, same semantics
- `COUNT(col1, col2)` without DISTINCT — properly rejected with syntax error

### Changes

| Layer | File | Change |
|-------|------|--------|
| Plan/Function | `list_agg.go` | Allow COUNT to accept multiple arguments |
| Plan/Binder | `having_binder.go` | Reject multi-arg COUNT without DISTINCT |
| Plan/Optimizer | `distinct_agg.go` | Skip GROUP BY optimization for multi-arg/tuple; expand tuple args |
| Plan | `utils.go` | Fix `increaseRefCnt` missing `*plan.Expr_List` handling |
| Executor | `count2.go`, `types.go` | Support multiple arg types |
| Executor | `aggState.go` | Multi-vector encoding with length-prefix; boundary check fix |
| Tests | `count2_test.go` | 4 new unit tests (distinct/NULL/multi-group/merge) |
| BVT | `func_aggr_count.test/result` | New test cases, updated expected results |


XuPeng-SH (changes requested):
Request changes for one correctness blocker.

`COUNT(DISTINCT (a, b))` is only normalized inside `optimizeDistinctAgg`, and only after the rule has already accepted the narrow single-aggregate shape. That leaves valid tuple syntax unexpanded when another aggregate is present in the same AGG node.

Example planner reproduction on this PR:

```sql
select count(distinct (n_nationkey, n_regionkey)), count(*) from nation;
```

The resulting AGG node still contains `count` with `args=1` and `T_tuple` for the first aggregate, while `starcount` is the second aggregate. This happens because `distinct_agg.go` returns before the tuple-expansion block when `len(node.AggList) != 1`.

That shape is not executable as a row-wise multi-column distinct key. The group operator evaluates aggregate args through `MakeEvalVector`, and `Expr_List` is handled by `ListExpressionExecutor`, which builds a vector from the list elements themselves (`UnionOne(vec, 0)` for each child) rather than producing one tuple value per input row. So an unexpanded tuple aggregate input is semantically wrong and can miscount or fail once `BatchFill` indexes it using input-row offsets.

Suggested fix:
- Do semantic normalization outside this optimization rule. Prefer flattening `COUNT(DISTINCT (a, b))` into the same arg list as `COUNT(DISTINCT a, b)` during binding or in a general aggregate-normalization pass that runs for every aggregate expression, independent of `len(node.AggList)` and independent of whether the distinct-agg optimization fires.
- Keep `optimizeDistinctAgg` only as an optimization; it should not be required for correctness.
- Add regression coverage for tuple syntax with more than one aggregate, for example:
  - `select count(distinct (g, v)), count(*) from t_count_distinct_multi;`
  - `select txt, count(distinct (g, v)), count(*) from t_count_distinct_multi group by txt;`
  - compare `count(distinct (g, v))` with `count(distinct g, v)` in the same query.

The direct multi-arg executor path and local merge tests look directionally correct, but this tuple path still violates the PR's stated "tuple syntax, same semantics" contract.

heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes on current head 75802831.

The previous tuple-with-peer-aggregate blocker looks fixed, but the current implementation opens a broader COUNT argument-validation hole.

`pkg/sql/plan/function/list_agg.go:33` now lets `count` bind with `len(inputs) >= 1`, while the non-DISTINCT multi-arg guard only exists in `HavingBinder.BindAggFunc` (`pkg/sql/plan/having_binder.go:149`). That does not cover tuple-as-one-arg and does not cover the window binder path. I verified these with a temporary planner test: each query below expected a bind error, but `BuildPlan` returned nil error on this PR:

```sql
select count((n_nationkey, n_regionkey)) from nation;
select count(n_nationkey, n_regionkey) over() from nation;
select count((n_nationkey, n_regionkey)) over() from nation;
```

This is not only a parse/compat issue. If the non-DISTINCT aggregate tuple path reaches execution after `HavingBinder` expands the tuple, `countColumnExec.BatchFill` takes the non-DISTINCT branch and only reads `vectors[0]` (`pkg/sql/colexec/aggexec/count2.go:169-175`), so it can silently behave like `COUNT(first_arg)` instead of rejecting an invalid COUNT form.

Suggested fix direction:

- Put COUNT argument validation in one shared binder helper and call it from both aggregate and window binding paths before `bindFuncExprImplByAstExpr`.
- Only allow multiple COUNT arguments for aggregate `COUNT(DISTINCT expr, expr, ...)`.
- Only expand `COUNT(DISTINCT (a, b))` when `funcName == "count" && astExpr.Type == tree.FUNC_TYPE_DISTINCT`; do not expand tuple args for non-DISTINCT COUNT.
- Window COUNT has DISTINCT disabled here already, so reject both `COUNT(a,b) OVER()` and `COUNT((a,b)) OVER()`.
- Add regression coverage for the three negative queries above, alongside the existing positive tuple/peer-aggregate cases.

Also please add an executor round-trip test for multi-column `COUNT(DISTINCT ...)` through `SaveIntermediateResult`/`UnmarshalFromReader`. I locally checked that such a round trip currently works, but this PR changed `writeStateArg/readStateArg` opaque key encoding, so that path should be explicitly covered.

aunjgr (commented):
Tuple expansion only fires in the optimization rule — count((a,b)) without DISTINCT still leaks through. Validation is scattered across HavingBinder, window binder, and the optimizer. Should have a single binding-time gate that rejects multi-arg / tuple non-DISTINCT COUNT uniformly.

XuPeng-SH (approved):
Re-reviewed current head 05e8dd1f. The previous blockers are fixed in the right place:

- COUNT arg validation is now a shared binding-time gate for aggregate and window paths, so non-DISTINCT multi-arg/tuple COUNT is rejected before execution.
- COUNT(DISTINCT (a, b)) is normalized during aggregate binding instead of depending on the distinct-agg optimization rule, so tuple syntax works with peer aggregates and grouped aggregates.
- Multi-column row subquery input is rejected explicitly instead of leaking an Expr_Sub/T_tuple into execution.
- Executor state now treats multi-arg distinct keys as opaque encoded keys, with merge/marshal and large-key arena growth covered.

Local verification:

- go test ./pkg/sql/colexec/aggexec -run 'TestCount|TestPR25325Manual'
- go test ./pkg/sql/plan -run 'TestCountDistinctRowSubqueryRejected|TestPR25325ManualCountArgBindingClosure|TestRewriteCountNotNullColToStarcount'

I also added temporary local-only scratch cases for varchar boundary-collision keys like ('a','bc') vs ('ab','c') and for valid/invalid planner closure around aggregate/window COUNT. They passed and were removed before review submission.

CI is green on the current head. Non-blocking note: the PR is still BEHIND main, so it may need an update branch before merge depending on queue policy.

aunjgr (approved):

PR #25334, opened 2026-07-01
Fix logservice client fallback from slow replica addresses

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25274 #25277

This PR improves logservice client creation when candidate replica addresses include stale or unreachable endpoints.

  Previously, connectToLogService tried replica addresses sequentially. If one address accepted a TCP connection but did not respond, client creation could wait until the outer context deadline before trying later
  candidates.

  Changes:
  - Keep the existing shuffled address order.
  - Start connection attempts with staggered fallback instead of strict sequential dialing.
  - Apply a per-address attempt timeout.
  - Return the first successfully connected client.
  - Cancel remaining attempts and close losing clients after a winner is selected.
  - Add a regression test covering a slow/unresponsive address followed by a reachable address.

  This reduces the impact of stale or slow replica addresses returned by discovery/gossip without changing HAKeeper or gossip behavior.

### DIFF
--- a/pkg/logservice/client.go
+++ b/pkg/logservice/client.go
@@ -42,6 +42,8 @@ var (
 	logServiceConnectFallbackDelay  = 300 * time.Millisecond
 )
 
+var connectToLogServiceAddressFn = connectToLogServiceAddress
+
 // IsTempError returns a boolean value indicating whether the specified error
 // is a temp error that worth to be retried, e.g. timeouts, temp network
 // issues. Non-temp error caused by program logics rather than some external
@@ -473,7 +475,12 @@ func connectToLogServiceByReverseProxy(
 	if !ok {
 		return nil, moerr.NewLogServiceNotReady(ctx)
 	}
-	addresses := make([]string, 0)
+	addresses := shardInfoReplicaAddresses(si)
+	return connectToLogServiceAddresses(ctx, sid, addresses, cfg)
+}
+
+func shardInfoReplicaAddresses(si ShardInfo) []string {
+	addresses := make([]string, 0, len(si.Replicas))
 	leaderAddress, ok := si.Replicas[si.ReplicaID]
 	if ok {
 		addresses = append(addresses, leaderAddress)
@@ -483,7 +490,7 @@ func connectToLogServiceByReverseProxy(
 			addresses = append(addresses, address)
 		}
 	}
-	return connectToLogService(ctx, sid, addresses, cfg)
+	return addresses
 }
 
 func connectToLogService(
@@ -528,7 +535,7 @@ func connectToLogServiceAddress(
 		c.respPool,
 		c.cfg.MaxMessageSize,
 		cfg.EnableCompress,
-		time.Second*10,
+		0,
 		cfg.Tag,
 	)
 	if err != nil {
@@ -556,6 +563,11 @@ func connectToLogServiceAddress(
 	return c, nil
 }
 
+type connectResult struct {
+	client *client
+	err    error
+}
+
 func connectToLogServiceAddresses(
 	ctx context.Context,
 	sid string,
@@ -565,22 +577,18 @@ func connectToLogServiceAddresses(
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	type result struct {
-		client *client
-		err    error
-	}
-	results := make(chan result, len(addresses))
+	results := make(chan connectResult, len(addresses))
 	var wg sync.WaitGroup
 	launch := func(addr string) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			attemptCtx, attemptCancel := context.WithTimeout(ctx, logServiceConnectAttemptTimeout)
-			c, err := connectToLogServiceAddress(attemptCtx, sid, addr, cfg)
+			c, err := connectToLogServiceAddressFn(attemptCtx, sid, addr, cfg)
 			attemptCancel()
 			if err == nil {
 				select {
-				case results <- result{client: c}:
+				case results <- connectResult{client: c}:
 				case <-ctx.Done():
 					if closeErr := c.close(); closeErr != nil {
 						logutil.Error("failed to close the client", zap.Error(closeErr))
@@ -589,7 +597,7 @@ func connectToLogServiceAddresses(
 				return
 			}
 			select {
-			case results <- result{err: err}:
+			case results <- connectResult{err: err}:
 			case <-ctx.Done():
 			}
 		}()
@@ -619,6 +627,7 @@ func connectToLogServiceAddresses(
 			if r.client != nil {
 				cancel()
 				wg.Wait()
+				closeUnusedConnectResultClients(results)
 				return r.client, nil
 			}
 			lastErr = r.err
@@ -630,6 +639,7 @@ func connectToLogServiceAddresses(
 			stopConnectFallbackTimer(timer)
 			cancel()
 			wg.Wait()
+			closeUnusedConnectResultClients(results)
 			if lastErr != nil {
 				return nil, lastErr
 			}
@@ -638,6 +648,21 @@ func connectToLogServiceAddresses(
 	}
 	wg.Wait()
 	return nil, lastErr
+}
+
+func closeUnusedConnectResultClients(results <-chan connectResult) {
+	for {
+		select {
+		case r := <-results:
+			if r.client != nil {
+				if closeErr := r.client.close(); closeErr != nil {
+					logutil.Error("failed to close the client", zap.Error(closeErr))
+				}
+			}
+		default:
+			return
+		}
+	}
 }
 
 func stopConnectFallbackTimer(timer *time.Timer) {
--- a/pkg/logservice/client.go
+++ b/pkg/logservice/client.go
@@ -37,6 +37,11 @@ const (
 	defaultWriteSocketSize = 64 * 1024
 )
 
+var (
+	logServiceConnectAttemptTimeout = 5 * time.Second
+	logServiceConnectFallbackDelay  = 300 * time.Millisecond
+)
+
 // IsTempError returns a boolean value indicating whether the specified error
 // is a temp error that worth to be retried, e.g. timeouts, temp network
 // issues. Non-temp error caused by program logics rather than some external
@@ -490,7 +495,19 @@ func connectToLogService(
 	if len(targets) == 0 {
 		return nil, nil
 	}
+	addresses := append([]string{}, targets...)
+	rand.Shuffle(len(addresses), func(i, j int) {
+		addresses[i], addresses[j] = addresses[j], addresses[i]
+	})
+	return connectToLogServiceAddresses(ctx, sid, addresses, cfg)
+}
 
+func connectToLogServiceAddress(
+	ctx context.Context,
+	sid string,
+	addr string,
+	cfg ClientConfig,
+) (*client, error) {
 	pool := &sync.Pool{}
 	pool.New = func() interface{} {
 		return &RPCRequest{pool: pool}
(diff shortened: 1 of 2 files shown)


XuPeng-SH (changes requested):
I think this needs another pass before merge. The direction is right, but the current implementation introduces new failure modes in the exact connection-fallback path this PR is trying to harden.

Findings:

1. `pkg/logservice/client.go:572` / `pkg/logservice/client.go:581` leaks non-winning successful clients. `results` is buffered to `len(addresses)`, so after the first success is selected at `pkg/logservice/client.go:619`, `cancel()` is called and `wg.Wait()` waits for losers, but a loser goroutine can still take the always-ready `results <- result{client: c}` branch instead of the `ctx.Done()` branch. That client is then left in `results` and never closed. This can happen in normal multi-replica cases when fallback launches a second reachable address and both connect successfully. Please either make loser-success send non-buffered/cancel-aware, or drain `results` after `wg.Wait()` and close every unreturned client. Add a regression test where two reachable attempts both succeed after fallback has launched, and assert the non-winning client is closed/no goroutine or fd leak remains.

2. `pkg/logservice/client.go:476` builds a leader-first list for discovery/reverse-proxy, but `pkg/logservice/client.go:498` now shuffles that list before dialing. That regresses the intended and previous discovery behavior: the leader returned by `GetShardInfo` should be tried first, with other replicas as fallback. This is also explicitly part of issue #25274 expected behavior. Keep shuffle for configured `ServiceAddresses`, but have `connectToLogServiceByReverseProxy` call the ordered/staggered dialer directly, or pass an explicit `shuffle bool`. Please add coverage that discovery preserves leader-first ordering.

3. `pkg/logservice/client.go:524` changes the production logservice client backend read timeout from `0` to `10s` by passing `time.Second*10` into `getRPCClient`, which is applied at `pkg/logservice/client.go:905`. This is not just a per-address connect-attempt limit; it changes the read loop behavior for the returned client and can force idle clients or long in-flight operations to reset/fail after 10s. The per-attempt context already bounds the CONNECT handshake, so this should remain `0` unless there is a separate, tested reason to change client read timeout semantics.

Suggested shape:
- Keep `connectToLogServiceAddress` behavior identical to the old client except for the per-attempt context.
- Separate address ordering from dial orchestration: direct configured addresses may be shuffled, discovery addresses must preserve leader-first.
- Ensure every successful but non-returned client is closed deterministically.
- Add unhappy-path tests for all-addresses-fail, parent context cancellation, multiple-success loser cleanup, and discovery leader-first ordering.

Local validation I ran on this head:
- `go test ./pkg/logservice -run TestClientCreationFallsBackFromUnreachableReplicaAddress -count=1`
- `go test ./pkg/logservice -run TestClientCreationFallsBackFromUnreachableReplicaAddress -count=10`
- `go test ./pkg/logservice -run "TestClientCanBeConnectedByReverseProxy|TestClientCreationFallsBackFromUnreachableReplicaAddress" -count=1`
- same targeted tests with `-race`
- `go test ./pkg/logservice -count=1`
- `gofmt -d pkg/logservice/client.go pkg/logservice/client_test.go`

XuPeng-SH:
Reposting the change-request findings here for visibility. I already submitted a GitHub review with `CHANGES_REQUESTED`: https://github.com/matrixorigin/matrixone/pull/25334#pullrequestreview-4607744865

Blocking findings:

1. `pkg/logservice/client.go:572` / `pkg/logservice/client.go:581`: non-winning successful clients can leak. `results` is buffered to `len(addresses)`, so after a winner is selected and `cancel()` is called, another successful goroutine can still choose the ready `results <- result{client: c}` branch. That client is then left in the channel and never closed. Please close/drain all successful non-returned clients and add a regression test with multiple successful attempts after fallback has launched.

2. `pkg/logservice/client.go:476` builds discovery addresses leader-first, but `pkg/logservice/client.go:498` shuffles them before dialing. This regresses the intended discovery behavior and conflicts with issue #25274 expected behavior. Please preserve leader-first for discovery/reverse-proxy and only shuffle configured `ServiceAddresses` if needed. Add a leader-first ordering test.

3. `pkg/logservice/client.go:524`: passing `time.Second*10` into `getRPCClient` changes the returned production client backend read timeout from old `0` semantics to 10s. This is not just the per-address connect-attempt timeout; it can reset/fail idle or long in-flight clients. The per-attempt context already bounds the CONNECT handshake, so this should remain `0` unless separately justified and tested.

Suggested shape: keep `connectToLogServiceAddress` behavior identical to the old client except for the per-attempt context; separate ordered discovery dialing from shuffled direct-address dialing; deterministically close every loser success; add unhappy-path tests for all-addresses-fail, parent cancellation, multiple-success loser cleanup, and discovery leader-first ordering.

LeftHandCold (changes requested):
I confirmed the existing change-request findings on the current head, so I cannot approve this as-is.

Blocking points:

1. Non-winning successful clients can still leak. A successful loser can enqueue `result{client: c}` into the buffered `results` channel after cancellation, and the winner path returns after `wg.Wait()` without draining and closing that client.
2. Discovery/reverse-proxy addresses are built leader-first, but the new `connectToLogService` now shuffles all inputs. On `3.0-dev`, the old code effectively preserved leader-first for discovery because `len(cfg.ServiceAddresses)` is zero in that path.
3. `connectToLogServiceAddress` changes the returned client's backend read timeout from the old `0` semantics to `10s`. The per-attempt context should bound connect/handshake without changing the long-lived client's read-timeout behavior.

Please fix these paths and add coverage for multiple successful attempts, discovery leader-first ordering, all-addresses-fail, and parent-context cancellation.

volgariver6 (approved):

XuPeng-SH (changes requested):
Re-reviewed the latest head `f0b6d82ad14cf6c83a0ee756c57065a8fa12c7d2`.

The previous three blocking items look mostly addressed in this revision:
- successful loser clients are now closed after `cancel()`/`wg.Wait()`/drain;
- discovery/reverse-proxy addresses now preserve leader-first ordering instead of being shuffled;
- the production RPC backend read timeout is back to the previous `0` behavior.

I still need to request changes because the new concurrent fallback path can return the wrong terminal error when an earlier address has failed but the parent context later expires before the whole search finishes.

In `connectToLogServiceAddresses`, the `<-ctx.Done()` branch currently does this:

```go
cancel()
wg.Wait()
closeUnusedConnectResultClients(results)
if lastErr != nil {
    return nil, lastErr
}
return nil, ctx.Err()
```

That means a mixed unhappy path such as `fast stale address failure -> next replica hangs -> caller deadline fires` returns the stale address error instead of `context.DeadlineExceeded`. This is not just error cosmetics: the PR is specifically fixing stale/slow logservice replica fallback, and callers/diagnostics lose the real terminal cause when their connect deadline is reached before all candidates complete.

I verified this locally with a temporary regression test using `connectToLogServiceAddressFn`: first address returns a sentinel error immediately, second waits for `ctx.Done()` and then returns. Expected `context.DeadlineExceeded`; actual error chain was only `"first failed"`.

Suggested fix:
- In the `<-ctx.Done()` branch, after cancel/wait/drain, return `ctx.Err()` (or the attached context cause if this codebase wants to preserve causes), not `lastErr`.
- Keep `lastErr` only for the normal path where all launched attempts have completed and all failed.
- Consider pre-checking `ctx.Err()` before launching the first/next attempt so an already-canceled parent does not start a dial and race back a non-context error.
- Add regression coverage for:
  1. previous address failed, active address is still pending, parent deadline fires -> returns `context.DeadlineExceeded`;
  2. already-canceled parent context -> returns `context.Canceled` and does not mask it with a dial error.

Local verification performed:
- New PR-targeted tests: `go test ./pkg/logservice -run 'TestClientCreationFallsBackFromUnreachableReplicaAddress|TestShardInfoReplicaAddressesKeepsLeaderFirst|TestConnectToLogServiceWithNoTargets|TestConnectToLogServiceAddressesReturnsLastError|TestConnectToLogServiceAddressesReturnsContextError|TestConnectToLogServiceAddressesClosesSuccessfulLosers|TestStopConnectFallbackTimer' -count=1` passed with the required MO CGo flags.
- Full package: `go test ./pkg/logservice -count=1 -timeout 120s` passed locally with the same CGo setup.

LeftHandCold (approved):

XuPeng-SH (changes requested):
Re-reviewed latest head `b4b67948314d7cc4258d5d5a7c232621c3e3d059`.

The previous functional blockers are addressed: discovery keeps leader-first ordering, the returned RPC client read timeout is still `0`, successful non-winning clients are drained/closed, and the mixed `previous failure + parent deadline` case now returns a context deadline error while preserving the previous failure in the message.

I still need one change before approving:

1. `pkg/logservice/client.go:590` adds a production `context.WithTimeout` for the per-address attempt. This new timeout is now the primary unhappy path for stale/hanging replicas, so it should preserve MO's timeout cause semantics instead of returning a plain `context.DeadlineExceeded`. The repo convention/static check also explicitly pushes production code toward `context.WithTimeoutCause` (`Makefile:951`). Please use `context.WithTimeoutCause` here, with `moerr.CauseNewLogServiceClient` or a more specific new cause, and attach/preserve that attempt cause when sending `lastErr` back to the coordinator. Add a regression test where all candidate replicas hang until the per-attempt timeout while the parent context is still alive, and assert the returned error remains deadline-classified and carries the MO cause/diagnostic context.

One non-blocking cleanup suggestion: `connectContextError` currently formats `lastErr` with `%v`, so `errors.Is/Unwrap` cannot see the last connect error. If callers ever need structured classification of that inner failure, prefer preserving both errors structurally instead of only embedding the text.

Local validation I ran on this head:
- `go test ./pkg/logservice -run 'TestClientCreationFallsBackFromUnreachableReplicaAddress|TestShardInfoReplicaAddressesKeepsLeaderFirst|TestConnectToLogServiceWithNoTargets|TestConnectToLogServiceAddressesReturnsLastError|TestConnectToLogServiceAddressesReturnsContextError|TestConnectToLogServiceAddressesReturnsContextErrorAfterPreviousFailure|TestConnectToLogServiceAddressesWithCanceledContextDoesNotDial|TestConnectToLogServiceAddressesClosesSuccessfulLosers|TestStopConnectFallbackTimer' -count=1 -timeout 120s`
- `go test ./pkg/logservice -count=1 -timeout 180s`
- `go test -race ./pkg/logservice -run 'TestClientCreationFallsBackFromUnreachableReplicaAddress|TestConnectToLogServiceAddressesReturnsContextErrorAfterPreviousFailure|TestConnectToLogServiceAddressesWithCanceledContextDoesNotDial|TestConnectToLogServiceAddressesClosesSuccessfulLosers' -count=1 -timeout 180s`
- `go test ./pkg/logservice -run 'TestConnectToLogServiceAddressesReturnsContextErrorAfterPreviousFailure|TestConnectToLogServiceAddressesClosesSuccessfulLosers' -count=50 -timeout 180s`

CI is still in progress, so this review is based on code inspection plus the local checks above.

XuPeng-SH (changes requested):
Requesting changes for one blocker:

- `pkg/logservice/client.go:590-596` now wraps every candidate in a hard 5s `context.WithTimeoutCause(...)` before `connectToLogServiceAddress`.
- That same context is reused by `CONNECT` / `CONNECT_RO`, whose server-side paths keep retrying `ErrShardNotReady` / `ErrSystemBusy` until the passed context expires (`pkg/logservice/service.go:426-441`, `pkg/logservice/store.go:679-717`).
- TN callers still budget 30s for logservice client creation by default (`pkg/tnservice/cfg.go:48,109-110`, `pkg/tnservice/factory.go:105-116`). So this changes behavior from “wait up to the callers connect timeout for a healthy shard to come up” to “kill each replica attempt after 5s”, which can reject healthy replicas during leader election / recovery.

Suggested fix: keep the staggered fallback, but do not put a fixed 5s deadline around the full `connectReadWrite` / `connectReadOnly` path. Either keep at least one viable attempt on the callers full budget, or restrict the short timeout to transport establishment / initial dial only.

XuPeng-SH (approved):
I did not find a remaining blocker on the current head.

The latest `preserve caller connect timeout` change fixes the previous regression cleanly: fallback still staggers new attempts behind a slow/hanging address, but each attempt now inherits the caller timeout instead of being hard-capped at 5s. That preserves the callers readiness budget while still allowing early fallback and loser cleanup/cancellation.

I also rechecked the unhappy-path surface on the current implementation: cancellation still terminates in-flight attempts, losing successful clients are closed, and the targeted logservice tests plus a package compile both pass on the PR head.

XuPeng-SH (changes requested):
Superseding my earlier approve: the latest head still has two blockers.

1. `pkg/logservice/client.go:589`, `pkg/common/morpc/future.go:55-57`
   - The new `preserve caller connect timeout` change passes the caller context straight into `connectToLogServiceAddressFn(ctx, ...)`. If the caller uses `context.Background()` / `WithCancel()` without a deadline, the first successful TCP connect now reaches morpc with a deadline-less context and panics in `Future.init` (`context deadline not set`). Earlier revisions wrapped each attempt in a timeout, so this failure mode did not exist.
   - Suggested fix: preserve an attempt-scoped deadline when the caller context has none, or fail explicitly before dialing instead of letting morpc panic.

2. `pkg/logservice/client.go:632-653`, `pkg/common/morpc/backend.go:872-920`
   - On first success or on caller cancellation, `connectToLogServiceAddresses` still does `cancel(); wg.Wait(); ...` before returning. But a launched loser can still be stuck in morpc backend creation, where the initial `rb.conn.Connect(..., rb.options.connectTimeout)` does not observe the caller context. So a blackholed loser dial can still delay both success return and cancellation return until that internal connect timeout expires.
   - Suggested fix: make backend creation/connect context-aware, or stop waiting on uncancelable loser dials before returning the winner / cancellation.

XuPeng-SH (approved):
Reviewed latest head 5fe0ab4546b56d8e298bceaefc0c83795c5597e4.

The previous blockers look closed now:
- caller contexts without deadlines get an attempt-scoped timeout, so morpc Future.init will not panic on deadline-less contexts;
- caller contexts with deadlines preserve the caller deadline instead of forcing a shorter per-attempt timeout;
- winner and cancellation paths no longer wait synchronously for loser dials that may be inside backend connect;
- buffered successful loser clients are drained and closed asynchronously;
- reverse-proxy discovery keeps leader-first ordering, while explicit target dialing still shuffles.

Unhappy-path coverage is now reasonable for this change: empty targets, canceled context before dial, previous failure plus caller timeout, deadline-less caller context, winner-with-slow-loser, cancel-with-slow-loser, and successful loser cleanup.

Verified:
- go test ./pkg/logservice targeted fallback tests
- go test -race ./pkg/logservice targeted async cleanup tests
- go test ./pkg/logservice
- GitHub CI for 3.0 is green

PR #25335, opened 2026-07-01
fix(plan): support GROUPING expressions in ROLLUP ORDER BY

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25287

`ROLLUP`, `CUBE`, and `GROUPING SETS` are expanded into an internal `UNION ALL` plan. The original `ORDER BY` is bound above that union, where source columns referenced by expressions such as `GROUPING(region)` are no longer visible even when the same expression is already projected.

This change rewrites an `ORDER BY` expression that exactly matches a select-list expression to its projection ordinal before building the internal grouping-set union. It reuses the projected value, preserves the final output schema, and does not relax the rules for explicit user-written set operations.

The change also adds planner unit coverage and a distributed ROLLUP regression case with real `NULL` values.

Validation:

- `go build ./pkg/sql/plan/...`
- `go vet ./pkg/sql/plan/...`
- modified helper coverage: 100%
- `make`
- rollup BVT: 112 passed, 0 failed, 2 ignored
- `git diff --check`

`make static-check` remains blocked by existing `pkg/common/docfilter` CGo `unsafe` molint violations; the same failure was reproduced on a clean `main` checkout.


XuPeng-SH (changes requested):
I found two substantive correctness gaps here.

1. Qualified `GROUPING()` references are still not handled correctly.

The current rewrite matches `GROUPING(...)` by raw unresolved-name spelling, so `GROUPING(a)` in the select list does not match `GROUPING(t.a)` in `ORDER BY` even when they refer to the same grouping column. Those queries still fall through to the later UNION-level/order binding path, where `t.a` no longer has a table binding.

Suggested fix: match/rewrite against the bound grouping columns rather than raw AST text, or delay the rewrite until after name resolution so `GROUPING(a)` and `GROUPING(t.a)` canonicalize to the same target.

2. Alias-shadowed `GROUPING()` still binds to the output alias instead of the grouped column.

The new helper intentionally skips rewriting when a `GROUPING()` argument name matches a select-list alias. But after that skip, the later binding path resolves that name through the output alias map, not the original grouped column. So alias-shadowed cases remain wrong instead of fixed.

Suggested fix: resolve `GROUPING()` arguments to underlying grouping-set columns before applying the alias-shadowing check; if that cannot be done unambiguously, raise an explicit ambiguity error instead of silently binding to the projected alias.

The new tests cover the main ROLLUP regression, but they do not invalidate these remaining semantic gaps.

heni02 (approved):

XuPeng-SH (changes requested):
I still see a correctness gap in the rewrite.

`rewriteGroupingSetOrderBy` only records and rewrites expressions whose root node is a bare `GROUPING(...)` call. The PR body says the fix rewrites an `ORDER BY` expression that exactly matches a select-list expression to its projection ordinal, but exact matches that merely contain `GROUPING()` are still skipped at `pkg/sql/plan/query_builder.go:3810` and `pkg/sql/plan/query_builder.go:3824`.

Repro on this head:

```sql
select grouping(a) + 0 as ga, count(*)
from select_test.bind_select
group by a with rollup
order by grouping(a) + 0;
```

and:

```sql
select if(grouping(a), 1, 0) as ga, count(*)
from select_test.bind_select
group by a with rollup
order by if(grouping(a), 1, 0);
```

Both fail during plan build with `invalid input: column a does not exist`. This is the same underlying scope problem as #25287: after ROLLUP is expanded into the internal `UNION ALL`, the later `ORDER BY` bind happens above the union, where the original grouped column is no longer available. The new helper fixes the direct `ORDER BY grouping(a)` shape, but not the full “matching select-list expression” shape it is trying to support.

Suggested fix: rewrite based on full select-list/order expressions that contain `GROUPING()`, after canonicalizing the `GROUPING()` arguments, instead of keying only bare root `GROUPING(...)` calls. A more systematic option is to bind/append hidden order-key projections into every generated grouping-set branch before building the internal union, then order by those hidden outputs and trim them from the final result. Please add regression coverage for nested exact-match expressions such as the two cases above.

aunjgr (changes requested):
Rewrite by AST text is fragile — GROUPING(a) vs GROUPING(t.a) will always mismatch. Should resolve to canonicalized columns after name binding, not before.

XuPeng-SH (changes requested):
Requesting changes on the latest head. The earlier qualified/nested `GROUPING()` issues look much better, but the new hidden-order-key approach currently bypasses the existing `SELECT DISTINCT` ORDER BY rule and can change visible DISTINCT results.

Repro on this head:

```sql
select distinct grouping(a)
from select_test.bind_select
group by a with rollup
order by grouping(a) + a;
```

For comparison, the ordinary path correctly rejects the same shape:

```sql
select distinct a
from select_test.bind_select
order by a + 1;
-- for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

The reason is that `prepareGroupingSetOrderByProjects` appends every `ORDER BY` expression containing `GROUPING()` as a hidden branch projection and rewrites the UNION ORDER BY to an ordinal (`query_builder.go:3843-3883`). Those hidden projections are then included in each generated branch while preserving `Distinct: selectClause.Distinct` (`query_builder.go:3101-3103`). By the time `buildUnionWithResultLen` binds ORDER BY, it sees only an ordinal, so the normal distinct guard in `OrderBinder` (`order_binder.go:140-143`) is never reached.

This is not only an error-policy mismatch: if the query is allowed, the hidden key participates in branch-level DISTINCT and is then trimmed from the final output, so different hidden keys can become duplicate visible rows after trimming.

Suggested fix: preserve the existing DISTINCT contract before injecting hidden grouping-set order keys. For `SELECT DISTINCT`, only rewrite to a hidden key when the ORDER BY expression is already a visible select-list expression (or rewrite to that visible ordinal); otherwise return the same error as the ordinary ORDER BY path. Please add regression coverage for the repro above and the allowed case where the full `ORDER BY` expression is selected.

XuPeng-SH (changes requested):
I found one remaining planner bug in the new DISTINCT guard.

The new `SELECT DISTINCT` path matches grouping-set `ORDER BY` items to the visible select list via `groupingOrderExprKey()`, but that key only fully normalizes `GROUPING()` when the argument is a bare column name. As soon as the `GROUPING()` argument is itself an expression, equivalent spellings are no longer treated as the same visible expression.

A concrete repro is the qualified/unqualified case below:

```sql
select distinct grouping(a) as g
from orders o
group by a with rollup
order by grouping(a)
```

works, but the same shape with the `GROUPING()` argument replaced by an expression like `year(o.order_date)` vs `year(order_date)` is still rejected as:

`for SELECT DISTINCT, ORDER BY expressions must appear in select list`

The reason is in `normalizeExprForComparison()`: once it sees the outer `GROUPING(...)` call, it unwraps only direct `UnresolvedName` arguments and then returns without descending into the nested expression tree. So `GROUPING(year(o.order_date))` and `GROUPING(year(order_date))` end up with different text keys even though they are the same visible projected expression.

This is stricter than the normal ORDER BY path and means the rewrite still breaks supported `GROUPING(expr)` order keys under DISTINCT. Please canonicalize the full `GROUPING()` argument subtree (or compare bound expressions instead of partially normalized SQL text), and add a regression for `SELECT DISTINCT ... ORDER BY GROUPING(year(...))` with qualified/unqualified spellings.

XuPeng-SH (changes requested):
Requesting changes on the latest head.

The previous recursive GROUPING() normalization issue is improved, but the DISTINCT matching path now over-normalizes ordinary qualified names.

In `normalizeExprForComparison`, qualifiers are stripped for every `*tree.UnresolvedName` at `pkg/sql/plan/query_builder.go:4004`, not only for names inside GROUPING() arguments. That makes different non-GROUPING references compare equal whenever the ORDER BY expression also contains GROUPING().

Repro:

```sql
select distinct grouping(a) + t1.b
from (select a, b from select_test.bind_select) as t1,
     (select b from select_test.bind_select) as t2
group by t1.a, t1.b, t2.b with rollup
order by grouping(a) + t2.b;
```

This should be rejected by the SELECT DISTINCT ORDER BY rule because the ORDER BY expression is not the visible select-list expression: the select list uses `t1.b`, while ORDER BY uses `t2.b`. On this head, `BuildPlan` returns nil error because both expressions canonicalize to `grouping(a) + b`, so ORDER BY is rewritten to the visible ordinal and silently orders by the wrong expression.

Suggested fix: scope qualifier stripping to the GROUPING() argument subtree only, including nested expressions inside GROUPING arguments, and preserve qualifiers for ordinary operands outside GROUPING(). A stronger option is to compare bound expressions or column references instead of text keys. Please add a regression for the derived-table case above.

Local checks:
- `go test ./pkg/sql/plan -run Test(QueryBuilderBuildRollupOrderByGroupingExpression|AppendGroupingSetOrderBy|NormalizeExprForComparison|GroupingOrderExprKey) -count=1`
- Temporary repro test for the SQL above fails today because `BuildPlan` returns nil error.

XuPeng-SH (changes requested):
I’m requesting changes for one correctness blocker.

**Hidden rollup sort keys become addressable via `ORDER BY <position>`.**

In the non-`DISTINCT` rollup/cube path, each `GROUPING(...)` sort key is turned into an extra hidden UNION projection, but positional `ORDER BY` validation still runs against the expanded projection list. That changes the SQL contract for user-written ordinals.

Concretely, a query like:

```sql
SELECT a, COUNT(*)
FROM t
GROUP BY a WITH ROLLUP
ORDER BY 3, GROUPING(a)
```

should still fail with “ORDER BY position 3 is not in select list”. After this change, `3` can bind to the hidden `GROUPING(a)` projection instead, so the query is accepted and sorts by an internal column.

The problem is the combination of:
- `prepareGroupingSetOrderByProjects()` appending hidden order expressions and rewriting them to new ordinals, and
- ordinal validation still using the expanded `ctx.projects` width before `hiddenResultLen` is trimmed back out.

I would keep the old visible-schema contract here: either
1. reuse a visible select-list ordinal when the `GROUPING(...)` expression already matches one, or
2. validate positional `ORDER BY` references against the visible result length rather than the hidden-expanded projection list.

Everything else I checked looked fine. Coverage also looks adequate overall; the remaining test suggestions are non-blocking.

XuPeng-SH (changes requested):
## Blocking finding

**[P1] Resolve DISTINCT visible matches after star expansion.**

The latest commits correctly defer hidden grouping-key placement until the star-expanded width is known, but the DISTINCT path still records a matched visible expression as `selectIdx + 1` in the raw AST select list. That ordinal is then bound after `*` expansion, where it can point to a different visible column.

Repro:

```sql
select distinct *, grouping(a)
from select_test.bind_select
group by a, b, c with rollup
order by grouping(a);
```

`select_test.bind_select` has `(a,b,c)`, so `grouping(a)` is visible column 4. On this head the generated SORT uses `ColPos=1` (the second star-expanded column); the equivalent explicit-list query uses `ColPos=3`, as expected. The query succeeds but silently sorts by the wrong expression.

Please carry the visible-match identity through to `buildUnionWithResultLen` and resolve it in the post-expansion projection space, just as hidden keys are now resolved, rather than materializing a pre-expansion ordinal. Add end-to-end regressions for `*` and qualified `t.*`, including a nested selected expression containing `GROUPING()` after the star.

The previous qualifier-scoping and user-ordinal blockers are fixed, and the original package/build/vet/race checks pass. This remaining case is absent from the current matrix.

gouhongshen (changes requested):
## Codex automated review

The DISTINCT grouping-set ORDER BY rewrite still compares unbound text and erases relation identity inside GROUPING(), causing valid queries to be rejected and distinct queries to sort by the wrong grouping bit.

### P1 - Bind source columns before matching DISTINCT grouping expressions (pkg/sql/plan/query_builder.go:4281)
On the DISTINCT grouping-set path, select-list keys call qualifyGroupingOrderExpr with useAlias=false, which returns the raw clone without source-column binding. The ORDER BY side uses a fresh alias-only context, so unqualified ordinary columns also remain raw. Consequently, `select distinct grouping(a) + t.b ... group by t.a, t.b with rollup order by grouping(a) + b` produces keys `grouping(a)+t.b` and `grouping(a)+b` and is rejected at line 4233, although `b` unambiguously resolves to the same source column. The normal binder qualifies select expressions with NoAlias and ORDER BY expressions with AliasBeforeColumn, so this special path should compare bound/canonical expressions while preserving alias semantics.

### P1 - Do not discard relation identity inside GROUPING arguments (pkg/sql/plan/query_builder.go:4420)
The recursive qualifier stripping makes `GROUPING(t1.a)` and `GROUPING(t2.a)` identical comparison keys. On the DISTINCT path, that causes `select distinct grouping(t1.a) from select_test.bind_select t1, select_test.bind_select t2 group by t1.a, t2.a with rollup order by grouping(t2.a)` to match the visible ordinal and bypass normal binding. The sort then uses the selected `GROUPING(t1.a)` bit instead of the requested `GROUPING(t2.a)` bit, and an ORDER BY expression that is not in the DISTINCT select list is accepted. GROUP BY binding preserves these qualified columns as distinct grouping entries, so matching must retain bound relation identity and reject ambiguous or mismatched arguments.

aunjgr (changes requested):
GROUPING matching is still not identity-safe. DISTINCT grouping-set ORDER BY can compare against unbound expression text, and GROUPING strips relation identity, so same-named columns from different self-join inputs may select the wrong grouping bit. Preserve bound column identity through GROUPING and compare normalized bound expressions. Add qualified self-join and DISTINCT expression regressions.

aunjgr (dismissed):
GROUPING matching is still not identity-safe. DISTINCT grouping-set ORDER BY can compare against unbound expression text, and GROUPING strips relation identity, so same-named columns from different self-join inputs may select the wrong grouping bit. Preserve bound column identity through GROUPING and compare normalized bound expressions. Add qualified self-join and DISTINCT expression regressions.

gouhongshen (changes requested):
## Codex automated review

Previously blocking identity, qualifier, nested-expression, star-expansion, and hidden-ordinal topics are fixed by the new branch-bound remapping path. One new compatibility blocker remains in DISTINCT grouping-set normalization. Focused tests could not run because `go` is unavailable in this environment.

### P1 - Preserve BINARY/VARBINARY output types in DISTINCT grouping sets (pkg/sql/plan/query_builder.go:6146)
`SELECT DISTINCT b FROM t GROUP BY b WITH ROLLUP` now routes every branch through this synthetic `IF(GROUPING(b), NULL, b)`. `iffCheck` has no `T_binary` or `T_varbinary` result candidate; for `(ANY, BINARY)` / `(ANY, VARBINARY)` its implicit-cast costs select `BLOB`. The resulting `normalized[i].Typ` is used unchanged downstream, so a query that previously returned BINARY/VARBINARY now exposes BLOB metadata (and its associated type semantics) to clients. Preserve the original varlen binary type during normalization—either support binary/varbinary in `IF` or avoid generic `IF` coercion here—and add BINARY and VARBINARY rollup/cube DISTINCT regressions.

gouhongshen (approved):
## Codex automated review

Previously blocking identity, qualifier, nested-expression, star-expansion, hidden-ordinal, and BINARY/VARBINARY type issues are fixed; no correctness blocker remains. Focused tests could not start because `go` is unavailable in the environment. One non-blocking PR-scope issue remains.

### P2 - Remove internal review notes from the change (CLAUDE_TODO_2026-07-24.md:67)
The current commit adds 22 lines of private review workflow notes, planned commands, and submission instructions. This file has no product, test, or user-documentation consumer and does not exist on the base branch, so these additions only leave unrelated development scratch material in the repository. Remove this block from the PR.

XuPeng-SH (changes requested):
Requesting changes on the latest head (`64f0397e9`). The earlier correctness blockers around identity-safe binding, star expansion, grouping NULL normalization, volatile expressions, and binary/vector typing are fixed. One performance blocker remains.

### P1 — Preserve branch-local DISTINCT after grouping NULL normalization

At `pkg/sql/plan/query_builder.go:3857`, every generated grouping-set branch is now forced to `Distinct: false`, leaving only the global DISTINCT above the complete `UNION ALL`. The global DISTINCT is required for correctness, but removing the branch-local DISTINCT creates an exponential fan-in regression for CUBE and a large-cardinality regression for ROLLUP/GROUPING SETS.

Concrete plan repro:

```sql
select distinct 1
from select_test.bind_select
group by cube(a, b, c);
```

On the PR base, each of the 8 cube branches has a local duplicate-elimination AGG before the UNION. On this head, there is no branch-local duplicate elimination; all rows emitted by all 8 grouping aggregates flow through the 7-node `UNION ALL` chain, and only then reach one global AGG. For a high-cardinality grouping key and low-cardinality visible projection, this changes the parent input from O(number of grouping sets) to O(sum of grouping-set cardinalities), increasing CPU/data movement and making CUBE especially risky.

Please retain the new global DISTINCT, but also apply local DISTINCT after each branch has run `appendGroupingSetDistinctProjectionNode` (so grouping NULLs are normalized first). In a detached worktree, restoring `Distinct: selectClause.Distinct` at this branch point preserved the full `pkg/sql/plan/...` test pass. Add a plan-shape regression proving branch reduction remains below UNION while global DISTINCT remains above it.

### P2 — Remove internal review scratch files

`CLAUDE_TODO_2026-07-23.md` and `CLAUDE_TODO_2026-07-24.md` add 263 lines of private review workflow, planned commands, and submission notes with no product/test/documentation consumer. They should not ship in the repository.

gouhongshen (approved):
## Codex automated review

Previous branch-local DISTINCT fan-in blocker is fixed: local de-dup now follows grouping-NULL normalization, while global DISTINCT remains above UNION ALL; ROLLUP, CUBE, and GROUPING SETS have plan-shape coverage. The scratch-file P2 is fixed by removal. Earlier identity, star/ordinal, volatile, and binary/vector-type blockers remain fixed. The merge has no remerge-resolution delta. `git diff --check` passes; focused tests could not run because `go` is unavailable.

XuPeng-SH (approved):
Reviewed exact head 82f0125c37 against main 6d8b254174. No blocking findings remain.

Verified the prior correctness and performance blockers are closed: bound relation identity is preserved for GROUPING expressions; nested expressions, aliases, self-joins, star expansion, and hidden ordinal boundaries are covered; DISTINCT keeps branch-local reduction after grouping-NULL normalization plus global de-dup above UNION ALL; derived ORDER BY keys are materialized after DISTINCT; binary/vector result metadata is preserved.

Compatibility/risk: the change is planner-local with no catalog, protobuf, persisted-state, or upgrade migration impact. The additional runtime projection is confined to DISTINCT grouping-set branches and all compile-time maps/slices are statement-bounded.

- git diff --check: pass
- full pkg/sql/plan and pkg/sql/plan/function tests: pass
- focused grouping-set/IF regressions: pass
- required CI (UT, SCA, compose/standalone BVT, coverage, Iceberg): green

Issue #24996, opened 2026-06-16
subquery in JOIN ON condition is not yet implemented (error 20102) — v4.0.0-rc2


dengn:
## Version

`8.0.30-OmniFabric-v4.0.0-rc2` (output of `SELECT version()`)

## Description

A subquery placed inside a `JOIN ... ON` condition fails to plan with:

```
ERROR 20102 (HY000): subquery in JOIN condition is not yet implemented
```

This is standard SQL (runs on MySQL 8 and PostgreSQL) and a very common "join to the most-recent / max row per group" pattern.

## Reproduction

```sql
CREATE TABLE j_dim (id INT, region_id INT);

SELECT a.id
FROM j_dim a
JOIN j_dim b
  ON b.id = (SELECT MAX(id) FROM j_dim z WHERE z.region_id = a.region_id);
```

## Actual

```
ERROR 20102 (HY000): subquery in JOIN condition is not yet implemented
```

## Expected

The statement plans and runs, evaluating the correlated scalar subquery in the `ON` clause (as MySQL 8 / PostgreSQL do).

## Impact

Hit by a natural-language-to-SQL assistant generating valid SQL against an OmniFabric / MatrixOne backend. The model naturally writes subqueries in `JOIN` conditions for "join each row to the latest matching row per group" questions, and those queries fail outright instead of running.

aunjgr:
merged into main

XuPeng-SH:
Confirmed again on a MatrixOne v4.0.0-rc4 instance.

## Environment

```sql
SELECT VERSION();
-- 8.0.30-MatrixOne-v4.0.0-rc4
```

## Application compatibility case

I created an isolated temporary database, executed a representative MySQL-style schema, then executed an application query with permission-style joins. The temporary database was dropped after the check.

- All `CREATE TABLE` statements succeeded, including `AUTO_INCREMENT`, `utf8mb4_bin`, secondary/unique keys, comments, and `CURRENT_TIMESTAMP()` defaults.
- The query failed during planning, before requiring any test data:

```text
ERROR 20102 (HY000) at line 1: subquery in JOIN condition is not yet implemented
```

The query contains four affected inner-join predicates. Two use `IN (SELECT ...)` and two use correlated `EXISTS (...)`, for example:

```sql
JOIN permission_mapping pm ON (
    pm.resource_id = r.resource_id
    AND pm.subject_kind = 'GROUP'
    AND pm.subject_id IN (
        SELECT m.group_id
[...]

PR #25336, opened 2026-07-01
fix: support scalar subquery in INNER JOIN ON condition

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24996

Scalar subqueries in INNER JOIN ON conditions previously failed with `subquery in JOIN condition is not yet implemented`. Since ON = WHERE semantically for INNER JOIN, the fix routes subquery conditions through the same `flattenSubqueries` path that WHERE uses.

### Root cause

`buildJoinTable` used `TableBinder` for all join types, which unconditionally rejects subqueries. Additionally, even when allowed through, subquery refs (`Expr_Sub`) were never flattened — unlike `bindWhere` which calls `flattenSubqueries` after binding.

### Fix

Two changes in `buildJoinTable`:

1. **Use `WhereBinder` for INNER JOIN** — allows subqueries through binding
2. **Separate + flatten subquery conditions** — subquery conditions go into a FILTER above the join (after `flattenSubqueries` replaces `Expr_Sub` with Col refs), clean conditions stay in OnList

Result:
```
FILTER(b.id = Col(subquery_join, 0))
  └─ LEFT JOIN                          ← flattenSubqueries creates this
       ├─ JOIN(INNER, OnList=[clean])    ← original join
       └─ subquery plan
```

LEFT/RIGHT/OUTER joins unchanged — still use `TableBinder`.

| File | Change |
|------|--------|
| `pkg/sql/plan/query_builder.go` | ~25 lines |  
| `pkg/sql/plan/build_test.go` | +8 lines unit test |
| `test/.../join/join.sql` | +10 lines BVT |
| `test/.../join/join.result` | regenerated |

### DIFF
--- a/pkg/sql/plan/build_test.go
+++ b/pkg/sql/plan/build_test.go
@@ -3275,3 +3275,12 @@ func TestReplaceCaptureDedupJoinDoesNotShuffle(t *testing.T) {
 
 	t.Fatal("expected REPLACE plan to contain a DEDUP JOIN with OldColCaptureList")
 }
+
+func TestSubqueryInJoinOn(t *testing.T) {
+	mock := NewMockOptimizer(false)
+	sqls := []string{
+		"SELECT a.id FROM j_dim a JOIN j_dim b ON b.id = (SELECT MAX(id) FROM j_dim z WHERE z.region_id = a.region_id)",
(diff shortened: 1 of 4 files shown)


aunjgr:
Accidentally created from main. Replaced by new PR.

PR #25338, opened 2026-07-01
add release 4.1 to mergify

heni02:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25337

add release 4.1 to mergify

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -12,6 +12,21 @@ queue_rules:
 
 
       Approved by: @{{ approved_reviews_by | join(', @') }}
+
+  - name: release-4.1
+    merge_conditions: []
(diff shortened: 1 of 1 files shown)


XuPeng-SH (approved):
LGTM.

This is a clean config-only change: it adds the missing `release-4.1` queue entry and the matching auto-queue rule for `base = 4.1-dev`, following the existing `release-4.0` pattern.

I don’t see a substantive issue here. The only practical caveat is operational rather than code-related: just make sure `4.1-dev` is indeed the intended branch name / rollout target when this lands.

Issue #25225, opened 2026-06-29
[Bug]: CAST signed positive string to UNSIGNED rejects leading plus sign


Ariznawlll:
### MatrixOne version

Verified on latest main service used for MySQL compatibility exploration (`8.0.30-MatrixOne-verify-greatest-main`, branch based on `origin/main`).

### Reproduce

```sql
select cast('+0012' as unsigned) as r;
```

### Expected behavior

MySQL 8.0.45 accepts the leading plus sign and returns:

```text
r
12
```

### Actual behavior

MatrixOne returns:

```text
ERROR 20203 (HY000): invalid argument cast to uint64, bad value +0012
```

### Notes

This was found while building the MySQL-compatible CAST/CONVERT target type matrix. The passing BVT uses `CAST('0012' AS UNSIGNED)`; the leading-plus variant is excluded and tracked here.

daviszhen:
fixed

Ariznawlll:
Verified on latest origin/main at 7e70bee98 (`fix cast` #25340).

Result: fixed.

Validation SQL:
```sql
select cast('+0012' as unsigned) as r;
select cast('+0' as unsigned) as plus_zero,
       cast(' +0012' as unsigned) as ws_plus,
       cast('+18446744073709551615' as unsigned) as uint64_max;
```

Observed result:
```text
r
12

plus_zero  ws_plus  uint64_max
0          12       18446744073709551615
```

Regression has been added to motr in matrixorigin/motr#39, commit b8f14f4 (`issue_25225_cast_unsigned_plus`).

Motr validation passed together with the related issue regressions:
```text
SUMMARY | total=6 success=6 failed=0 skipped=0 | stmts=46 ok=46 fail=0
```

Ariznawlll:
Regression case added in matrixorigin/motr#39.

Case: `14_issue_regression.issue_25225_cast_unsigned_plus`

Target branch marker: `@target-branch: main`

Issue #25250, opened 2026-06-30
[Bug]: CAST of whitespace-signed scientific notation strings to FLOAT/DOUBLE returns 0


Ariznawlll:
### MatrixOne Version

Latest `origin/main` verified locally at `33e96877604ad4b847ea7d5c55eb09429de1a5b6`.

### Reproduce Steps

```sql
select cast('7e0' as double),
       cast('7e0' as float),
       cast('  +7e0' as double),
       cast('  -7.5e1' as float),
       cast('1e-3' as decimal(8,4));
```

### Expected Behavior

MySQL returns:

```text
7    7    7    -75    0.0010
```

### Actual Behavior

MatrixOne returns:

```text
7    7    0    0    0.0010
```

### Notes

The plain exponent forms match MySQL and were added to BVT. The whitespace/signed exponent forms incorrectly coerce to zero in MatrixOne.

daviszhen:
fixed

Ariznawlll:
Verified on latest origin/main at 7e70bee98 (`fix cast` #25340).

Result: fixed.

Validation SQL:
```sql
select cast('7e0' as double) as plain_double,
       cast('7e0' as float) as plain_float,
       cast('  +7e0' as double) as ws_plus_double,
       cast('  -7.5e1' as float) as ws_minus_float;
select cast('1e-3' as decimal(8,4)) as decimal_exp;
```

Observed result:
```text
plain_double  plain_float  ws_plus_double  ws_minus_float
7.0           7.0          7.0             -75.0

decimal_exp
0.0010
```

Regression has been added to motr in matrixorigin/motr#39, commit b8f14f4 (`issue_25250_cast_signed_scientific`).

Motr validation passed together with the related issue regressions:
```text
SUMMARY | total=6 success=6 failed=0 skipped=0 | stmts=46 ok=46 fail=0
```

Ariznawlll:
Regression case added in matrixorigin/motr#39.

Case: `14_issue_regression.issue_25250_cast_signed_scientific`

Target branch marker: `@target-branch: main`

Issue #25249, opened 2026-06-30
[Bug]: CAST to fixed-length CHAR errors instead of truncating like MySQL


Ariznawlll:
### MatrixOne Version

Latest `origin/main` verified locally at `33e96877604ad4b847ea7d5c55eb09429de1a5b6`.

### Reproduce Steps

```sql
select cast(12345 as char(3)), length(cast(12345 as char(3)));
```

### Expected Behavior

MySQL truncates the character result to the requested length and returns:

```text
123    3
```

### Actual Behavior

MatrixOne returns an error:

```text
ERROR 20101 (HY000): internal error: Can't cast '12345' from BIGINT type to CHAR type. 12345 is larger than Dest length 3
```

### Notes

Found while completing the MySQL-compatible `CAST` / `CONVERT` target-type matrix. Passing fixed-length `BINARY(4)` cast rows were added to BVT; this failing `CHAR(3)` row was excluded.

daviszhen:
fixed

Ariznawlll:
Verified on latest origin/main at 7e70bee98 (`fix cast` #25340).

Result: fixed.

Validation SQL:
```sql
select cast(12345 as char(3)) as numeric_char,
       length(cast(12345 as char(3))) as numeric_len;
select cast('abcdef' as char(3)) as string_char,
       length(cast('abcdef' as char(3))) as string_len;
```

Observed result:
```text
numeric_char  numeric_len
123           3

string_char   string_len
abc           3
```

Regression has been added to motr in matrixorigin/motr#39, commit b8f14f4 (`issue_25249_cast_char_truncation`).

Motr validation passed together with the related issue regressions:
```text
SUMMARY | total=6 success=6 failed=0 skipped=0 | stmts=46 ok=46 fail=0
```

Ariznawlll:
Regression case added in matrixorigin/motr#39.

Case: `14_issue_regression.issue_25249_cast_char_truncation`

Target branch marker: `@target-branch: main`

PR #25340, opened 2026-07-01
fix cast

daviszhen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25225

#25250

#25249

修改 : `cast('+0012' as unsigned)`
修改 : ` cast('  +7e0' as double),`
修改: `select cast(12345 as char(3)), length(cast(12345 as char(3)));`

### DIFF
--- a/pkg/sql/plan/function/func_cast.go
+++ b/pkg/sql/plan/function/func_cast.go
@@ -5098,7 +5098,11 @@ func strToUnsigned[T constraints.Unsigned](
 				if strings.HasPrefix(s, "0x") || strings.HasPrefix(s, "0X") {
 					val, tErr = strconv.ParseUint(s[2:], 16, bitSize)
 				} else {
-					val, tErr = strconv.ParseUint(s, 10, bitSize)
+					parseStr := s
+					if strings.HasPrefix(parseStr, "+") {
+						parseStr = parseStr[1:]
+					}
+					val, tErr = strconv.ParseUint(parseStr, 10, bitSize)
 				}
 			}
 			if tErr != nil {
--- a/pkg/sql/plan/function/func_cast.go
+++ b/pkg/sql/plan/function/func_cast.go
@@ -470,55 +470,55 @@ func newCast(parameters []*vector.Vector, result vector.FunctionResultWrapper, p
 		err = bitToOthers(proc.Ctx, s, *toType, result, length, selectList)
 	case types.T_int8:
 		s := vector.GenerateFunctionFixedTypeParameter[int8](from)
-		err = int8ToOthers(proc.Ctx, s, *toType, result, length, selectList)
+		err = int8ToOthers(proc.Ctx, s, *toType, result, length, selectList, strictStringWidth)
 	case types.T_int16:
 		s := vector.GenerateFunctionFixedTypeParameter[int16](from)
-		err = int16ToOthers(proc.Ctx, s, *toType, result, length, selectList)
+		err = int16ToOthers(proc.Ctx, s, *toType, result, length, selectList, strictStringWidth)
 	case types.T_int32:
 		s := vector.GenerateFunctionFixedTypeParameter[int32](from)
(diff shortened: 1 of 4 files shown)


heni02 (approved):

XuPeng-SH (changes requested):
I found a real blocker here.

The new `truncateCastBytesResult(...)` hook now runs inside the shared `...ToStr` helpers that are used by both plain `cast` and internal `cast_strict`.

That means assignment-only `cast_strict` into real `CHAR/VARCHAR(N)` columns will now silently truncate many non-string source types before the old width-check/error path runs.

For example, numeric / decimal / date / datetime / timestamp / time assignments into `CHAR/VARCHAR` targets now go through helpers like `signedToStr`, `unsignedToStr`, `floatToStr`, `dateToStr`, `datetimeToStr`, `timestampToStr`, `timeToStr`, `decimal64ToStr`, `decimal128ToStr`, `decimal256ToStr`, and all of them call `truncateCastBytesResult(result, toType)` before the final over-width check.

But `cast_strict` is explicitly the assignment path that is supposed to reject over-width writes instead of truncating them. So a statement that should fail with data-too-long semantics can now silently store a shortened string.

Suggested fix:
1. thread strict/lenient behavior into the shared `...ToStr` helpers, or
2. split truncation so only plain `cast` truncates while `cast_strict` still errors.

There is also still an incompleteness issue for other supported `-> CHAR/VARCHAR` paths (YEAR/UUID/JSON/ENUM/BIT are not all aligned with the new truncation behavior), but the `cast_strict` regression above is the blocker.

daviszhen:
> I found a real blocker here.
> 
> The new `truncateCastBytesResult(...)` hook now runs inside the shared `...ToStr` helpers that are used by both plain `cast` and internal `cast_strict`.
> 
> That means assignment-only `cast_strict` into real `CHAR/VARCHAR(N)` columns will now silently truncate many non-string source types before the old width-check/error path runs.
> 
> For example, numeric / decimal / date / datetime / timestamp / time assignments into `CHAR/VARCHAR` targets now go through helpers like `signedToStr`, `unsignedToStr`, `floatToStr`, `dateToStr`, `datetimeToStr`, `timestampToStr`, `timeToStr`, `decimal64ToStr`, `decimal128ToStr`, `decimal256ToStr`, and all of them call `truncateCastBytesResult(result, toType)` before the final over-width check.
> 
> But `cast_strict` is explicitly the assignment path that is supposed to reject over-width writes instead of truncating them. So a statement that should fail with data-too-long semantics can now silently store a shortened string.
> 
> Suggested fix:
> 
> 1. thread strict/lenient behavior into the shared `...ToStr` helpers, or
> 2. split truncation so only plain `cast` truncates while `cast_strict` still errors.
> 
> There is also still an incompleteness issue for other supported `-> CHAR/VARCHAR` paths (YEAR/UUID/JSON/ENUM/BIT are not all aligned with the new truncation behavior), but the `cast_strict` regression above is the blocker.

已经修改了.

XuPeng-SH (approved):
Re-reviewing the latest head: the earlier primary blocker is fixed now.

- The old concern that `cast_strict` was silently truncating non-string assignments into `CHAR/VARCHAR` is now addressed. The current head threads strict/lenient width handling through the affected numeric/date/time/decimal `...ToStr` helpers, and the new strict-path regression tests cover representative non-string cases.

What remains is a narrower consistency gap around some other `-> CHAR/VARCHAR` cast paths, especially `YEAR -> CHAR/VARCHAR`. But I checked the target-branch baseline, and that `YEAR` path was already behaving this way before this PR; it is not a new regression introduced by this change.

So for this PR itself, I no longer see a substantive blocker. The remaining cast-matrix consistency cleanup looks like follow-up work rather than a reason to block this fix.

LGTM.

ouyuanning (approved):

PR #25342, opened 2026-07-01
fix: maintain covering secondary indexes

### LINKED ISSUE #25341: [Bug]: VARCHAR equality predicate returns zero rows with owner-bound PK and covering secondary index

### Branch Name

v4.0.0-rc4

### Commit ID

Unknown from SQL variables. Reproduced with Docker image `matrixorigin/matrixone:4.0.0-rc4`.

- Hardware parameters: local Docker Compose development environment
- OS type: Linux
- MatrixOne version: `8.0.30-MatrixOne-v4.0.0-rc4`
- Docker image: `matrixorigin/matrixone:4.0.0-rc4`

### Actual Behavior

For a table with an owner-bound composite primary key and a secondary index that includes the `VARCHAR` predicate column plus the primary-key columns, `status = 'active'` evaluates to `1` in a row projection, but the same predicate in `WHERE` returns zero rows.

`status <=> 'active'` returns the expected row count.

Observed output:

```text
id  status  row_eq  row_nullsafe
a1  active  1       1

count_eq
0

count_nullsafe
1
```

### Expected Behavior

`WHERE status = 'active'` should return the inserted row, so `count_eq` should be `1`, matching the row-level expression result.

### Steps to Reproduce

Run this SQL on MatrixOne `8.0.30-MatrixOne-v4.0.0-rc4`:

```sql
DROP DATABASE IF EXISTS astra_repro_varchar_predicate;
CREATE DATABASE astra_repro_varchar_predicate;
USE astra_repro_varchar_predicate;

CREATE TABLE t (
    id VARCHAR(64) NOT NULL,
    user_id VARCHAR(64) NOT NULL,
    session_id VARCHAR(64) NOT NULL,
    status VARCHAR(32) NOT NULL,
    due DATETIME(6) NULL,
    PRIMARY KEY (user_id, session_id, id),
    INDEX idx_ret(status, due, user_id, session_id, id)
);

INSERT INTO t
VALUES ('a1', 'u1', 's1', 'active', DATE_ADD(NOW(6), INTERVAL 1 DAY));

SELECT id, status, status = 'active' AS row_eq, status <=> 'active' AS row_nullsafe
FROM t;

SELECT COUNT(*) AS count_eq
FROM t
WHERE status = 'active';

SELECT COUNT(*) AS count_nullsafe
FROM t
WHERE status <=> 'active';
```

### Additional information

This was found while debugging retention queries on a table with this index shape:

```sql
PRIMARY KEY (user_id, session_id, artifact_id),
INDEX idx_artifacts_retention (status, retention_until, retention_policy, user_id, session_id, artifact_id)
```

The broader query `WHERE status IN ('active', 'expiring')` showed the same symptom: row-level expressions evaluated correctly, while `WHERE` aggregates returned `0`.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25341

This fixes the inconsistent secondary-index path for tables whose regular secondary index covers the full composite primary key.

The root invariant is: if a regular hidden index table exists and the optimizer can read it, DML must maintain it. Previously `getValidIndexes` skipped some regular indexes when the index parts covered the full primary key, but the optimizer could still choose the existing hidden index table for predicates such as `status = 'active'` or `status IN (...)`, producing empty results from a stale index table.

Changes:
- Keep regular secondary indexes maintainable whenever their hidden table exists, even when the index covers the full composite primary key.
- Use the same serialized lookup shape as the DML/index table encoding: `serial_full` for non-unique multi-part regular indexes, and `serial` for unique or single-part lookups.
- Restrict the regular secondary-index rewrite to regular index algorithms so vector/fulltext/master index definitions cannot be consumed by this path.
- Add planner unit coverage for covering regular indexes, irregular-index guards, `serial_full` lookup rewrites, equality/IN/range predicates, and unique-index behavior.
- Add an optimizer regression case for the issue shape, covering equality, IN, range, null-safe equality, row projection, update, delete, replace, and `ON DUPLICATE KEY UPDATE`.

Local validation:
- `git diff --check`
- `make cgo`
- `CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -lusearch_c" DYLD_LIBRARY_PATH="$(pwd)/cgo:$(pwd)/thirdparties/install/lib" go test -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" -count=1 -timeout 180s ./pkg/sql/plan`
- `CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -lusearch_c" DYLD_LIBRARY_PATH="$(pwd)/cgo:$(pwd)/thirdparties/install/lib" go test -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" -count=1 -timeout 120s ./pkg/sql/colexec/multi_update`
- `CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -lusearch_c" DYLD_LIBRARY_PATH="$(pwd)/cgo:$(pwd)/thirdparties/install/lib" go build -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" ./pkg/sql/plan`
- `CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -lusearch_c" DYLD_LIBRARY_PATH="$(pwd)/cgo:$(pwd)/thirdparties/install/lib" go vet -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" ./pkg/sql/plan`

### DIFF
--- a/pkg/sql/plan/apply_indices.go
+++ b/pkg/sql/plan/apply_indices.go
@@ -728,7 +728,7 @@ func isRegularIndexFullPrefixEquality(expr *plan.Expr, numKeyParts int) bool {
 		return false
 	}
 	serialFn := fn.Args[1].GetF()
-	return serialFn != nil && serialFn.Func.ObjName == "serial" && len(serialFn.Args) == numKeyParts
+	return serialFn != nil && serialFn.Func.ObjName == "serial_full" && len(serialFn.Args) == numKeyParts
 }
 
 func hasTopValueMessage(node *plan.Node) bool {
@@ -925,7 +925,7 @@ func (builder *QueryBuilder) applyIndicesForFiltersRegularIndex(nodeID int32, no
 	indexes := make([]*IndexDef, 0, len(node.TableDef.Indexes))
 	spatialIndexes := make([]*IndexDef, 0, len(node.TableDef.Indexes))
 	for i := range node.TableDef.Indexes {
-		if node.TableDef.Indexes[i].IndexAlgo == "fulltext" || !node.TableDef.Indexes[i].TableExist {
+		if !node.TableDef.Indexes[i].TableExist || !catalog.IsRegularIndexAlgo(node.TableDef.Indexes[i].IndexAlgo) {
 			continue
 		}
 		if isSpatialIndexDef(node.TableDef.Indexes[i]) {
@@ -1174,7 +1174,8 @@ func findLeadingFilter(idxDef *IndexDef, node *plan.Node) ([]int32, bool) {
 	return nil, false
 }
 
-func (builder *QueryBuilder) replaceEqualCondition(filterList []*plan.Expr, filterPos []int32, idxTag int32, idxTableDef *plan.TableDef, numParts int) *plan.Expr {
+func (builder *QueryBuilder) replaceEqualCondition(idxDef *IndexDef, filterList []*plan.Expr, filterPos []int32, idxTag int32, idxTableDef *plan.TableDef) *plan.Expr {
+	numParts := len(idxDef.Parts)
 	if numParts == 1 { //directly equal
 		expr := DeepCopyExpr(filterList[filterPos[0]])
 		args := expr.GetF().Args
@@ -1191,7 +1192,7 @@ func (builder *QueryBuilder) replaceEqualCondition(filterList []*plan.Expr, filt
 		serialArgs[i] = DeepCopyExpr(filter.GetF().Args[1])
 		compositeFilterSel = compositeFilterSel * filter.Selectivity
 	}
-	rightArg, _ := BindFuncExprImplByPlanExpr(builder.GetContext(), "serial", serialArgs)
+	rightArg, _ := BindFuncExprImplByPlanExpr(builder.GetContext(), indexTableLookupSerialFunc(idxDef), serialArgs)
 
 	funcName := "="
 	if len(filterPos) < numParts {
@@ -1203,12 +1204,13 @@ func (builder *QueryBuilder) replaceEqualCondition(filterList []*plan.Expr, filt
 	return expr
 }
 
-func (builder *QueryBuilder) replaceNonEqualCondition(filter *plan.Expr, idxTag int32, idxTableDef *plan.TableDef, numParts int) *plan.Expr {
+func (builder *QueryBuilder) replaceNonEqualCondition(idxDef *IndexDef, filter *plan.Expr, idxTag int32, idxTableDef *plan.TableDef) *plan.Expr {
+	numParts := len(idxDef.Parts)
 	expr := DeepCopyExpr(filter)
 	fn := expr.GetF()
 	if fn.Func.ObjName == "or" {
 		for i := range expr.GetF().Args {
-			expr.GetF().Args[i] = builder.replaceNonEqualCondition(expr.GetF().Args[i], idxTag, idxTableDef, numParts)
+			expr.GetF().Args[i] = builder.replaceNonEqualCondition(idxDef, expr.GetF().Args[i], idxTag, idxTableDef)
 		}
 		return expr
 	}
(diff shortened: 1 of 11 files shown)


XuPeng-SH (commented):

1. `serial_full` can still break SQL NULL semantics on covering index-only scans.

For non-unique composite regular indexes, the new rewrite path uses `serial_full(...)`, which preserves NULL as real key bytes. But `checkIndexFilter` still accepts runtime-constant RHS values without any NULL guard, and `tryIndexOnlyScan()` replaces the original leading filter with the hidden-key predicate without keeping a residual `a = ?` / `a IN (...)` recheck.

That means a predicate like `a = ?` or `a IN (?, ?)` with a NULL runtime value can become byte-prefix-matchable on the hidden key instead of staying SQL-NULL/UNKNOWN, and a covering index-only scan can return rows whose indexed value is NULL even though the original SQL predicate should return no rows.

Suggested fix: skip this rewrite whenever any lookup value can be NULL, or keep a residual original predicate on the rewritten scan so SQL three-valued logic is preserved. The same guard needs to cover `IN` lists containing NULL candidates.

2. The join/runtime-filter path still appears to treat irregular indexes like ordinary hidden secondary-index tables.

The new regular-index-algorithm guard exists in the filter-rewrite path, but `applyIndicesForJoins()` still iterates every `TableExist` index with no equivalent `catalog.IsRegularIndexAlgo(...)` exclusion. That means join rewrites can still build generic hidden-table probes for MASTER/fulltext/vector-style indexes whose key layout is not the ordinary secondary-index layout.

Suggested fix: apply the same regular-index guard in `applyIndicesForJoins()`, or route irregular algos to their dedicated rewrite logic instead of the generic hidden-table path.

The BVT/unit additions cover the main regression well, but they do not invalidate these two remaining semantic gaps.

XuPeng-SH:
Addressed the two review blockers in `d6a821c78`.

What changed:
- Covering index-only scans that use `serial_full` now keep the original leading SQL predicate as a residual filter mapped through `serial_extract(...)`. This preserves SQL NULL/UNKNOWN semantics for `= ?`, `IN (?, ...)`, and range-style prepared predicates while still using the hidden-key probe to reduce candidates.
- `applyIndicesForJoins()` now excludes non-regular index algorithms and spatial indexes from the generic hidden secondary-index runtime-filter path, so master/fulltext/vector/rtree layouts cannot be treated as ordinary secondary index tables.

Coverage added:
- Planner unit coverage for residual predicates on `=`, `IN`, and `BETWEEN` prepared paths.
- Planner unit coverage that join rewrite skips master/fulltext/vector/spatial indexes.
- BVT coverage with a nullable indexed `status` value and prepared NULL lookups; expected counts are `0` for `status = NULL`, `status IN (NULL, missing)`, and `status BETWEEN NULL AND active`.

Local verification:
- `go test ./pkg/sql/plan`
- `go test ./pkg/sql/plan/function`
- `go test ./pkg/sql/colexec/multi_update`
- `go vet ./pkg/sql/plan ./pkg/sql/plan/function`
- `make build`
- mo-tester genrs + run for `secondary_index_range.sql`: `171/171`, 100% success
- Manual minimal SQL repro confirms new plan has both `prefix_eq(__mo_index_idx_col)` and residual `(serial_extract(__mo_index_idx_col, 0, VARCHAR)) = (null)`, and all NULL prepared counts return `0`.

XuPeng-SH (commented):

- The NULL-semantics problem on covering index-only scans is fixed: the latest `tryIndexOnlyScan()` keeps the original leading predicate as a residual recheck when `serial_full` is involved and the lookup can compare against runtime NULL.
- The missing regular-index guard in the join/runtime-filter path is also fixed: `applyIndicesForJoins()` now excludes non-regular and spatial/rtree indexes from the generic hidden-index rewrite path.

At this point I don’t see a remaining substantive correctness issue. The remaining gaps are at most nice-to-have extra coverage, not blockers.

LGTM.

heni02 (approved):

Issue #24411, opened 2026-05-15
[Bug]: 'missing FROM-clause entry' when CTE is mixed with comma-list + explicit JOIN in FROM


dengn:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

main

### Commit ID

N/A — observed on the `8.0.30-OmniFabric` downstream fork; behaviour appears to come from the upstream binder.

### Other Environment Information

```Markdown
- Hardware parameters: 8 vCPU / 32GB RAM
- OS type: Ubuntu 24.04
- Others: MySQL 8.0 client; identical SQL works on MySQL 8.0
```

### Actual Behavior

When a CTE name is referenced alongside a base table using SQL-89 comma cross-join syntax, and the comma-list is followed by an explicit `JOIN` clause, the binder fails to resolve aliases declared in the comma list:

```
ERROR 20301 (HY000): invalid input: missing FROM-clause entry for table 'r'
```

The alias `r` is clearly introduced in `FROM site_hourly_stats r, latest_date` but the binder reports it as missing once an explicit `JOIN dim_site` follows.

### Expected Behavior

MySQL 8.0 accepts the same statement. The binder should resolve all aliases introduced in the comma-separated `FROM` list before processing the subsequent `JOIN` clauses, matching standard SQL and MySQL behaviour.

### Steps to Reproduce

```sql
CREATE TABLE site_hourly_stats (
  site_id INT,
  date_id INT,
  avg_flow_m3h DOUBLE
);

CREATE TABLE dim_site (
  id INT PRIMARY KEY,
  site_code VARCHAR(32),
  type_code VARCHAR(16)
);

INSERT INTO dim_site VALUES (1, 'PLANT_A', 'DESAL'), (2, 'PLANT_B', 'DESAL');
INSERT INTO site_hourly_stats VALUES (1, 20260515, 1200.0), (2, 20260515, 900.0);

-- FAILS with "missing FROM-clause entry for table 'r'"
WITH latest_date AS (SELECT MAX(date_id) AS d FROM site_hourly_stats)
SELECT s.site_code, SUM(r.avg_flow_m3h) AS daily_flow
FROM site_hourly_stats r, latest_date ld
JOIN dim_site s ON r.site_id = s.id
WHERE r.date_id = ld.d
GROUP BY s.site_code;
```

Workaround — replace comma cross-join with an explicit `CROSS JOIN`:

```sql
[...]

PR #25353, opened 2026-07-02
fix: resolve comma-joined tables in nested explicit JOIN ON conditions

aunjgr:
## What type of PR is this?

- [x] BUG

issue #24411

`FROM a, b JOIN c ON a.x = c.y` failed with "missing FROM-clause entry for table a". The comma-joined table was not visible in the explicit JOIN ON condition.

### Root cause

The parser produces `CROSS(a, INNER(b, c, ON a.x = c.y))` for comma + explicit JOIN. The outer CROSS join left context (containing `a`) was dropped when `buildTable` dispatched the inner INNER join to `buildJoinTable` — `leftCtx` was never passed through.

### Fix

1. `buildJoinTable` now accepts `extLeftCtx *BindContext`
2. `buildTable` passes `leftCtx` through
3. In `buildJoinTable`, temporarily add `extLeftCtx` bindings to the ON condition binding context, then remove them so the outer `mergeContexts` does not see a duplicate

205/205 join BVT pass. Unit test covers the comma+explicit JOIN pattern.

| File | Change |
|------|--------|
| `pkg/sql/plan/query_builder.go` | +33 / -2 lines |
| `pkg/sql/plan/build_test.go` | +8 lines unit test |
| `test/.../join/join.sql` | +12 lines BVT |
| `test/.../join/join.result` | regenerated |


ouyuanning (approved):

heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes. I found a correctness blocker in the current direction.

1. `test/distributed/cases/join/join.sql:178-179` / `join.result:227-229` shows the new predicate is not actually applied at the intended scope. The data has `r` rows `(1,100),(2,200)` and `s` rows `(1,'A'),(2,'B')`; `where r.val = ld.v` keeps only `r.id = 2`, so if `ON r.id = s.id` is effective the only valid row is `2/B`. The committed result is `2/A` and `2/B`, which means the binder now accepts `r.id` but the join predicate is placed on an inner join whose children do not include `r`, so it cannot enforce the condition correctly.

2. This also conflicts with MySQL join-scope semantics. MySQL documents that `JOIN` has higher precedence than comma, so `t1, t2 JOIN t3 ON t1...` is interpreted as `t1, (t2 JOIN t3)`, and the `ON` clause may refer only to the join operands (`t2` and `t3`), not `t1`: https://dev.mysql.com/doc/en/join.html. The current fix deliberately makes the outer comma-left context visible in the inner explicit JOIN ON clause, so it changes the language semantics instead of preserving MySQL compatibility.

Suggested direction:
- If MatrixOne wants MySQL-compatible behavior, keep the unparenthesized query as an error and update the issue/test to use either explicit `CROSS JOIN` or parentheses, e.g. `(t_comma_join r, ... ld) JOIN t_comma_dim s ON ...`.
- If MatrixOne intentionally wants this extension, do not bind external refs into the inner join `OnList`. Rewrite/lift the predicate to a node whose child subtree actually contains all referenced tables, and add result-level regression coverage proving the expected output is only `2/B` for the current BVT data.
- Add unhappy-path coverage for ambiguous/invalid cases: external left refs in `LEFT/RIGHT/FULL JOIN ON`, duplicate table aliases, ambiguous unqualified columns, and an explicit MySQL-compatible negative test for `t1, t2 JOIN t3 ON t1...` if compatibility is the goal.

Local verification I ran:
- `go test -count=1 -timeout 180s ./pkg/sql/plan -run 'TestCommaJoinWithExplicitJoin|TestSubqueryInJoinOn|TestWrongCases'`
- `go test -count=1 -timeout 300s ./pkg/sql/plan`

Both pass, but they do not cover the actual result correctness issue above.

aunjgr (commented):
Here is a simpler 3-line fix that avoids the injection/cleanup complexity:

**Root cause:** `buildJoinTable` calls `ctx.mergeContexts(leftCtx, rightCtx)` AFTER the ON condition is bound. For comma joins like `FROM a, b JOIN c ON a.x = c.y`, the parser creates `CROSS_JOIN(a, INNER_JOIN(b, c))`. The inner JOIN binds its ON condition before the outer CROSS_JOIN's `mergeContexts` runs — so `a`'s context is a sibling, invisible during binding.

**Fix (3 lines):**

1. In `buildTable` line 5066, pass `leftCtx`:
```go
- return builder.buildJoinTable(tbl, ctx)
+ return builder.buildJoinTable(tbl, ctx, leftCtx)
```

2. Change signature:
```go
-func (builder *QueryBuilder) buildJoinTable(tbl *tree.JoinTableExpr, ctx *BindContext) (int32, error) {
+func (builder *QueryBuilder) buildJoinTable(tbl *tree.JoinTableExpr, ctx *BindContext, extLeftCtx *BindContext) (int32, error) {
```

3. Around the ON condition binding (line 5412):
```go
+ oldParent := ctx.parent
+ if extLeftCtx != nil { ctx.parent = extLeftCtx }
  joinConds, err := splitAndBindCondition(cond.Expr, NoAlias, ctx)
+ ctx.parent = oldParent
```

The existing parent-walk in `baseBindColRef` (base_binder.go:495-507) already walks `ctx.parent` to resolve columns. No binding injection, no cleanup — just temporarily wire the comma-left context into the parent chain during ON binding.

gouhongshen:
Blocking on correctness.

The current change makes the outer comma-left binding visible while binding the inner `JOIN ... ON`, but the predicate is still attached to the inner join whose children do not contain that outer table. In the new BVT case:

```sql
select r.id, s.typ
from t_comma_join r, (select max(val) as v from t_comma_join) ld
join t_comma_dim s on r.id = s.id
where r.val = ld.v;
```

`where r.val = ld.v` leaves only `r.id = 2`. If `on r.id = s.id` is actually enforced at the intended scope, the only valid row is `2/B`. The committed expected result is:

```text
2 | A
2 | B
```

So the PR currently proves that `r.id` can be resolved, but also proves that the join predicate is not being applied correctly.

There is also a compatibility issue in the premise. MySQL 8.0 documents that `JOIN` has higher precedence than comma and that an `ON` clause can refer only to the operands of that join. For `t1, t2 JOIN t3 ON t1...`, MySQL treats the operands as `t2` and `t3`, so `t1` is not in scope for that `ON` clause: https://dev.mysql.com/doc/refman/8.0/en/join.html

Suggested direction:

- If MatrixOne wants MySQL-compatible behavior, keep the unparenthesized form as an error and cover it with a negative test. Use parentheses or explicit `CROSS JOIN` in the positive case.
- If MatrixOne intentionally wants this extension, do not leave external-table refs in the inner join `OnList`; lift/rewrite the predicate to the join/filter node whose subtree actually contains all referenced tables, and update the result test to prove the output is only `2/B` for this data.
- Add coverage for ambiguous unqualified columns and duplicate aliases; the current `bindingByCol` injection skips existing names, which can hide ambiguity instead of reporting it.

XuPeng-SH (changes requested):
I found one real correctness blocker in the new ext-left binding logic for mixed comma joins.

The fix makes outer comma-left tables visible inside the nested `JOIN ... ON` by temporarily copying `extLeftCtx` bindings into `ctx`, but it only adds `bindingByCol` entries when the column name is missing. That means if the outer comma-left table and the inner join side both have the same unqualified column name, the new visibility never marks that name ambiguous; it just leaves the inner binding in place.

Concretely, the new path lets a query like `select s.r_name from nation r, nation2 join region s on n_regionkey = s.r_regionkey` pass, even though `n_regionkey` is now visible from both `r` and `nation2` and should be ambiguous. I reproduced this locally with a tiny planner test using `runTestShouldError`, and it currently fails because the statement is accepted.

So the patch fixes qualified outer references like `r.n_regionkey`, but it still mis-binds unqualified same-name columns instead of preserving the usual ambiguity rules. I think the temporary merge into `bindingByCol` needs to follow the normal `addBinding` behavior (`existing -> nil`) rather than only filling missing names, and there should be a regression test for this ambiguity case.

fengttt (approved):

PR #25359, opened 2026-07-02
[4.1-dev] fix: allow scalar subquery in INNER JOIN ON condition

aunjgr:
## What type of PR is this?

- [x] BUG

issue #24996

Scalar subqueries in JOIN ON conditions failed with NYI. Cherry-pick of #25339 to 4.1-dev.

TableBinder.BindSubquery now delegates to baseBindSubquery. Subquery conditions are flattened via flattenSubqueries and wrapped in FILTER above the join.

| File | Change |
|------|--------|
| `pkg/sql/plan/table_binder.go` | +1 / -1 line |
| `pkg/sql/plan/build_test.go` | +8 lines |
| `test/.../func_aggr_sum.result` | regenerated |
| `test/.../join/join.sql` | +10 lines |


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (changes requested):

1. `JOIN ... ON` subqueries are accepted by the binder now, but the join path still does not lower them into an executable form. `buildTable()` stores the bound ON predicates directly in `node.OnList`, and unlike WHERE / HAVING / projection paths there is still no `flattenSubqueries()` step for join predicates. Constant folding also does not rescue this: `rule.IsConstant()` returns false for `Expr_Sub`, so the scalar subquery stays inside the ON expression. At compile time non-equi ON predicates are forwarded to `colexec.NewExpressionExecutor` (`extraJoinConditions()` / `constructLoopJoin()`), and `NewExpressionExecutor` still has no `Expr_Sub` case. So removing the NYI in `TableBinder.BindSubquery()` makes plan building pass, but the execution path is still incomplete.

2. This change is broader than JOIN ON. `TableBinder` is also used by `buildTableFunction()` and `buildResultScan()`. Those call sites still treat their arguments as plain expressions; `result_scan` in particular evaluates them through `GetReadonlyResultFromNoColumnExpression()` -> `NewExpressionExecutor()`, which still rejects `Expr_Sub`. So this PR also starts accepting subqueries in non-JOIN `TableBinder` callers that still do not support them.

The new unit test only exercises `runOneStmt` / plan construction, so it does not catch either runtime gap. I think the fix needs to be scoped to real JOIN-ON support end-to-end (including lowering/compile/runtime), or the old guard should stay in place until that support exists.

aunjgr:
Resolved the execution-closure and binder-scope issues in e09bbeec8.

- INNER JOIN ON now uses a narrowly enabled JoinOnBinder; ordinary TableBinder callers (including table functions/result_scan) continue rejecting subqueries.
- Subquery predicates are flattened immediately and placed in a FILTER above the INNER JOIN, so no Expr_Sub reaches OnList or an expression executor.
- Non-subquery ON conjuncts remain in the join condition.
- Outer-join ON subqueries remain rejected.
- Added correlated, uncorrelated, mixed-predicate, plan-closure, and unsupported-outer-join coverage.

Verification:
- go test ./pkg/sql/plan -count=1: PASS
- make build: PASS
- join.sql BVT: 202/202, 100%

The broader implementation already merged on main in #25339 has the same TableBinder scope hole. I will send a separate main-branch PR with this scoped correction.

XuPeng-SH (approved):
Re-reviewed the latest head `e09bbeec85`, focusing on the runtime gap raised on `6333a63bd7`. The new INNER-JOIN-only binder gate plus lowering path looks correct: subqueries are still rejected for other `TableBinder` callers, while INNER JOIN `ON` subqueries are now flattened out of `OnList` and lifted into a FILTER above the join, so the executor no longer sees raw `Expr_Sub` in join predicates. I reran the focused `TestSubqueryInJoinOn` unit test on this head and do not see a remaining high-confidence correctness or unhappy-path blocker in the latest delta.

PR #25360, opened 2026-07-02
[4.1-dev] fix: resolve comma-joined tables in nested explicit JOIN ON conditions

aunjgr:
## What type of PR is this?

- [x] BUG

issue #24411

`FROM a, b JOIN c ON a.x = c.y` failed with "missing FROM-clause entry". Cherry-pick of #25353 to 4.1-dev.

buildJoinTable now accepts extLeftCtx, temporarily adds bindings for ON condition resolution, then removes them before the outer mergeContexts.

| File | Change |
|------|--------|
| `pkg/sql/plan/query_builder.go` | +33 / -2 lines |
| `pkg/sql/plan/build_test.go` | +8 lines |
| `test/.../join/join.sql` | +12 lines |


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (changes requested):
Requesting changes on the latest head `8cc519b` after its merge from `4.1-dev`. This backport carries the unresolved early implementation from #25353 and has correctness blockers.

1. **Wrong-result blocker:** `query_builder.go:5378-5412` makes the outer comma-left binding visible while binding the nested `JOIN ... ON`, but the resulting expression remains in the inner JOIN node's `OnList` (`5426-5432`). That node's children are only the inner operands, so the external relation is outside the executable subtree. The new BVT proves the predicate is not enforced: `WHERE r.val = ld.v` leaves `r.id=2`; `ON r.id=s.id` should therefore return only `2/B`, but `join.result` records both `2/A` and `2/B`. Turning a binding error into a silent wrong result is not mergeable.

This also contradicts the compatibility contract. MySQL documents that explicit `JOIN` has higher precedence than comma, `t1, t2 JOIN t3` means `t1, (t2 JOIN t3)`, and the `ON` clause can refer only to its join operands. Therefore `ON t1...` is expected to fail unless the comma group is parenthesized: https://dev.mysql.com/doc/refman/8.0/en/join.html

2. **Ambiguous columns are mis-bound:** the injection at `query_builder.go:5395-5399` adds an external unqualified column only when the name is absent. Normal context merging marks duplicate unqualified names as ambiguous (`bindingByCol[col] = nil`); this path leaves the inner binding selected. I reproduced it with:

```sql
select region.r_name
from nation, nation2 join region
  on n_regionkey = region.r_regionkey;
```

Both `nation` and `nation2` expose `n_regionkey`, so this must be ambiguous. A focused `runTestShouldError` regression fails with `should error, but pass` on this head.

3. **Golden-file pollution:** `join.result` rewrites almost the whole pre-existing join result format and changes unordered row order, although `join.sql` adds only one case. Restore the target branch's existing results and add only the intended regression output.

Please decide the language contract first:

- For MySQL compatibility, keep the unparenthesized form as an error and use `(t1, t2) JOIN t3 ON ...` or explicit JOINs for the positive case.
- If MatrixOne intentionally adds non-MySQL scope, lift/rewrite predicates to a node whose subtree contains every referenced binding; do not attach external refs to an inner JOIN. Prove exact result correctness and ambiguity/duplicate-alias behavior.

The PR's existing `TestCommaJoinWithExplicitJoin` passes locally, but it only checks that planning succeeds and cannot detect either wrong execution results or mis-binding. Q1-Q3 lifecycle checks found no leak/hang/unbounded-growth blocker; the blockers are semantic correctness, compatibility, and test integrity.

PR #25368, opened 2026-07-02
feat(iceberg): add external Iceberg connector

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23359 

    Add the external Iceberg connector foundation across SQL parsing, catalog metadata, scan planning, execution, append/DML writes, maintenance procedures, residency security, object IO, and MOI-facing ref/job plumbing.

    Include generated parser/proto updates, bootstrap system tables, unit/integration harnesses, local Nessie/MinIO/Spark test assets, and CI workflow coverage for Iceberg connector validation.


fengttt (approved):

XuPeng-SH (changes requested):
Requesting changes for several blockers:

1. `pkg/frontend/authenticate.go:6361-6364, 6604-6610, 9798-9799` + `pkg/frontend/iceberg.go:30-235`
   - `CREATE/ALTER/DROP ICEBERG CATALOG` and `SHOW ICEBERG *` are classified as `privilegeKindNone`, and the privilege engine short-circuits success for that kind. The handlers then directly mutate/read `mo_iceberg_*` with no compensating admin check. Any ordinary user in an account can reconfigure/drop account-wide catalogs and enumerate catalog URIs/namespaces/table mappings.
   - Suggested fix: require `accountadmin`/`moadmin` (or a dedicated privilege) for all Iceberg catalog DDL and metadata-show statements.

2. `pkg/frontend/iceberg.go:338-351`, `pkg/frontend/iceberg_call.go:210-262`, `pkg/sql/iceberg/ddl.go:30-45,62-75`, `pkg/sql/iceberg/dao.go:636-642`, `pkg/sql/iceberg/residency.go:159-170`
   - `catalog_id` is only unique within an account (`max(catalog_id)+1 where account_id = ...`), but cluster-scoped residency policies are stored with `account_id=0` and matched on bare `catalog_id`. A cluster policy created for account As catalog `1` will also be evaluated for account Bs catalog `1`, which breaks tenant isolation.
   - Suggested fix: make cluster policy identity include the owning account (or switch to a globally unique catalog identifier/name/URI) and enforce lookups on that full identity.

3. `pkg/iceberg/metadata/object_reader_factory.go:145-156`, `pkg/iceberg/io/object_io_registry.go:29-66,79-109`, `pkg/sql/colexec/external/parquet.go:237-240`
   - The remote-signing path registers `ObjectIORef`s with `ttl=0`, which becomes the global 30-minute default. File opens resolve that ref lazily per file, so any long scan/DML/maintenance flow that is still opening files after 30 minutes starts failing with “object IO ref is not registered or expired”. On top of that, production code never calls `ReleaseObjectIORef`, so refs also accumulate indefinitely in the global registry.
   - Suggested fix: tie ObjectIORef lifetime to query/operator teardown instead of the fixed 30m default for remote-signing flows, explicitly release refs on teardown, and add opportunistic/background sweeping for abandoned entries.

4. `pkg/sql/iceberg/dml_executor.go:51-93`, `pkg/sql/iceberg/dml_delete_builder.go:133-159,391-447`, `pkg/iceberg/dml/workflow.go:75-149,255-295`
   - The DML builders materialize replacement/delete objects before the commit workflow (and its orphan recorder) starts. If a later build/validation step fails, the executor returns immediately and never reaches the only orphan-recording path, leaving uncommitted files in table storage with no cleanup record.
   - Suggested fix: do not write objects until a workflow/orphan tracker is active, or register/clean up every materialized object if action-stream construction fails.

5. `pkg/sql/colexec/external/iceberg_delete_apply.go:516-547`
   - Equality-delete decoding for `timestamp` with `isAdjustedToUTC=false` subtracts the offset from `time.Now()`, not the offset for the timestamp being decoded. Around DST / historical offset changes, the decoded key shifts and equality deletes silently stop matching rows.
   - Suggested fix: decode timestamp-without-timezone values using the values own local-time semantics, not the current session offset, and add a DST regression test.

6. `pkg/sql/plan/build_insert.go:199-227` + `pkg/sql/iceberg/dml_delete_builder.go:559-571`
   - `INSERT OVERWRITE PARTITION` accepts static partition tuples without validating the keys against the target Iceberg partition spec. If the user supplies a typo/unknown field, runtime matches zero files and still appends the new files, silently degrading overwrite into append.
   - Suggested fix: validate overwrite-partition keys against the Iceberg partition spec during planning or coordinator setup and fail if any requested field is unknown/inapplicable.

iamlinjunhong:
> Requesting changes for several blockers:
> 
> 1. `pkg/frontend/authenticate.go:6361-6364, 6604-6610, 9798-9799` + `pkg/frontend/iceberg.go:30-235`
>    
>    * `CREATE/ALTER/DROP ICEBERG CATALOG` and `SHOW ICEBERG *` are classified as `privilegeKindNone`, and the privilege engine short-circuits success for that kind. The handlers then directly mutate/read `mo_iceberg_*` with no compensating admin check. Any ordinary user in an account can reconfigure/drop account-wide catalogs and enumerate catalog URIs/namespaces/table mappings.
>    * Suggested fix: require `accountadmin`/`moadmin` (or a dedicated privilege) for all Iceberg catalog DDL and metadata-show statements.
> 2. `pkg/frontend/iceberg.go:338-351`, `pkg/frontend/iceberg_call.go:210-262`, `pkg/sql/iceberg/ddl.go:30-45,62-75`, `pkg/sql/iceberg/dao.go:636-642`, `pkg/sql/iceberg/residency.go:159-170`
>    
>    * `catalog_id` is only unique within an account (`max(catalog_id)+1 where account_id = ...`), but cluster-scoped residency policies are stored with `account_id=0` and matched on bare `catalog_id`. A cluster policy created for account As catalog `1` will also be evaluated for account Bs catalog `1`, which breaks tenant isolation.
>    * Suggested fix: make cluster policy identity include the owning account (or switch to a globally unique catalog identifier/name/URI) and enforce lookups on that full identity.
> 3. `pkg/iceberg/metadata/object_reader_factory.go:145-156`, `pkg/iceberg/io/object_io_registry.go:29-66,79-109`, `pkg/sql/colexec/external/parquet.go:237-240`
>    
>    * The remote-signing path registers `ObjectIORef`s with `ttl=0`, which becomes the global 30-minute default. File opens resolve that ref lazily per file, so any long scan/DML/maintenance flow that is still opening files after 30 minutes starts failing with “object IO ref is not registered or expired”. On top of that, production code never calls `ReleaseObjectIORef`, so refs also accumulate indefinitely in the global registry.
>    * Suggested fix: tie ObjectIORef lifetime to query/operator teardown instead of the fixed 30m default for remote-signing flows, explicitly release refs on teardown, and add opportunistic/background sweeping for abandoned entries.
> 4. `pkg/sql/iceberg/dml_executor.go:51-93`, `pkg/sql/iceberg/dml_delete_builder.go:133-159,391-447`, `pkg/iceberg/dml/workflow.go:75-149,255-295`
>    
>    * The DML builders materialize replacement/delete objects before the commit workflow (and its orphan recorder) starts. If a later build/validation step fails, the executor returns immediately and never reaches the only orphan-recording path, leaving uncommitted files in table storage with no cleanup record.
>    * Suggested fix: do not write objects until a workflow/orphan tracker is active, or register/clean up every materialized object if action-stream construction fails.
> 5. `pkg/sql/colexec/external/iceberg_delete_apply.go:516-547`
>    
>    * Equality-delete decoding for `timestamp` with `isAdjustedToUTC=false` subtracts the offset from `time.Now()`, not the offset for the timestamp being decoded. Around DST / historical offset changes, the decoded key shifts and equality deletes silently stop matching rows.
>    * Suggested fix: decode timestamp-without-timezone values using the values own local-time semantics, not the current session offset, and add a DST regression test.
> 6. `pkg/sql/plan/build_insert.go:199-227` + `pkg/sql/iceberg/dml_delete_builder.go:559-571`
>    
>    * `INSERT OVERWRITE PARTITION` accepts static partition tuples without validating the keys against the target Iceberg partition spec. If the user supplies a typo/unknown field, runtime matches zero files and still appends the new files, silently degrading overwrite into append.
>    * Suggested fix: validate overwrite-partition keys against the Iceberg partition spec during planning or coordinator setup and fail if any requested field is unknown/inapplicable.

addressed the blockers in this patch.

- Iceberg catalog DDL and SHOW ICEBERG statements now require admin-level special privilege instead of privilegeKindNone, so ordinary account users can no longer mutate or enumerate account-wide Iceberg catalog metadata.
- Residency policy matching now scopes catalog-specific policies by normalized catalog URI in addition to catalog_id. This prevents account-local catalog_id collisions from allowing a cluster policy for one account’s catalog to match another account’s catalog with the same numeric id.
- Remote-signing ObjectIORefs no longer inherit the fixed 30-minute default TTL. They are registered as non-expiring and tied to operator lifecycle via retain/release ref counting, with opportunistic sweeping for expired refs.
- DML action-stream construction now tracks materialized replacement/delete objects while builders run. If a later validation/build step fails before the commit workflow starts, the executor records the successfully written objects as orphan candidates.
- Equality-delete timestamp decoding for isAdjustedToUTC=false now uses the timestamp value’s own local-time offset instead of time.Now(), and the same value-based offset logic is used on the Parquet scan path. I added a DST regression test.
- INSERT OVERWRITE PARTITION now validates static partition tuple keys against the target Iceberg partition spec before filtering affected files, so typos fail fast instead of silently degrading overwrite into append.

On coverage: the existing Iceberg E2E gate does not contribute to Go coverage because it runs SQL/service-level tests, not coverage-instrumented Go tests. I moved the local E2E Go runner from optools/ to test/iceberg/ and updated the CI script path, since it is a test runner and should not drag down production Go modified-line coverage. I also added focused UT coverage for the fixes above.

xzxiong (approved):

heni02 (approved):

aptend (approved):

ouyuanning (approved):

LeftHandCold:
I think the catalog-id isolation fix for residency policies is still incomplete.

`CheckCatalogResidency` now compares `allowed_catalog_uri`, but the storage identity still collides before read time. `mo_iceberg_catalogs` allocates `catalog_id` per account via `max(catalog_id)+1` for the current account (`pkg/frontend/iceberg.go:338-341`), while `mo_iceberg_residency_policy` has primary key `(scope_type, account_id, catalog_id, allowed_endpoint, allowed_region, allowed_bucket)` and excludes `allowed_catalog_uri` (`pkg/sql/iceberg/ddl.go:62-75`). For cluster scope, `iceberg_register_access` stores `account_id=0` (`pkg/frontend/iceberg_call.go:210-218`) and then does `ON DUPLICATE KEY UPDATE allowed_catalog_uri = ...` (`pkg/frontend/iceberg_call.go:249-261`).

So two accounts can both have `catalog_id=1` and register cluster policies for the same endpoint/region/bucket but different catalog URIs. The second registration overwrites the first row URI instead of creating an independent policy. Runtime URI filtering (`pkg/sql/iceberg/residency.go:323-331`) then denies whichever catalog was overwritten, making policy behavior depend on registration order and keeping cross-tenant state interference in the write path.

Please include `allowed_catalog_uri` or another global catalog identity in the unique key/upsert identity, or avoid account-local `catalog_id` for cluster-scope policy identity. A regression test should cover two accounts with the same catalog_id and same endpoint/region/bucket but different catalog URIs.

iamlinjunhong:
> I think the catalog-id isolation fix for residency policies is still incomplete.
> 
> `CheckCatalogResidency` now compares `allowed_catalog_uri`, but the storage identity still collides before read time. `mo_iceberg_catalogs` allocates `catalog_id` per account via `max(catalog_id)+1` for the current account (`pkg/frontend/iceberg.go:338-341`), while `mo_iceberg_residency_policy` has primary key `(scope_type, account_id, catalog_id, allowed_endpoint, allowed_region, allowed_bucket)` and excludes `allowed_catalog_uri` (`pkg/sql/iceberg/ddl.go:62-75`). For cluster scope, `iceberg_register_access` stores `account_id=0` (`pkg/frontend/iceberg_call.go:210-218`) and then does `ON DUPLICATE KEY UPDATE allowed_catalog_uri = ...` (`pkg/frontend/iceberg_call.go:249-261`).
> 
> So two accounts can both have `catalog_id=1` and register cluster policies for the same endpoint/region/bucket but different catalog URIs. The second registration overwrites the first row URI instead of creating an independent policy. Runtime URI filtering (`pkg/sql/iceberg/residency.go:323-331`) then denies whichever catalog was overwritten, making policy behavior depend on registration order and keeping cross-tenant state interference in the write path.
> 
> Please include `allowed_catalog_uri` or another global catalog identity in the unique key/upsert identity, or avoid account-local `catalog_id` for cluster-scope policy identity. A regression test should cover two accounts with the same catalog_id and same endpoint/region/bucket but different catalog URIs.

fixed

LeftHandCold:
The storage key should use the same canonical catalog URI that the runtime matcher uses.

Right now `iceberg_register_access` stores the raw `catalogURI` string (`pkg/frontend/iceberg_call.go:199-218`, `pkg/frontend/iceberg_call.go:249-260`), and the new primary key includes that raw `allowed_catalog_uri` (`pkg/sql/iceberg/ddl.go:62-75`). But residency checks normalize both the request URI and each policy URI before comparing (`pkg/sql/iceberg/residency.go:303-331`), and disabled rows are skipped (`pkg/sql/iceberg/residency.go:104-107`, `pkg/sql/iceberg/residency.go:155-157`).

That leaves the DB identity weaker than the runtime identity. For example, if there is an enabled policy for `https://catalog.example.com/rest`, a later registration with `catalog_uri=HTTPS://Catalog.Example.com/rest` and `policy_state=disabled` does not hit the duplicate-key update; it inserts a second row because the raw key differs. At runtime both URIs normalize to the same catalog URI, the disabled row is ignored, and the old enabled row still allows the catalog/object access. The same issue applies to any other URI form normalized by `NormalizeCatalogURI`, such as a fragment being stripped.

Please normalize `catalogURI` before building the `ResidencyPolicy` and before inserting/upserting it, or store a separate normalized URI/hash as the key identity. I would also add a regression test that equivalent URI variants upsert the same policy row and that disabling through an equivalent URI actually disables the existing enabled policy.

iamlinjunhong:
> The storage key should use the same canonical catalog URI that the runtime matcher uses.
> 
> Right now `iceberg_register_access` stores the raw `catalogURI` string (`pkg/frontend/iceberg_call.go:199-218`, `pkg/frontend/iceberg_call.go:249-260`), and the new primary key includes that raw `allowed_catalog_uri` (`pkg/sql/iceberg/ddl.go:62-75`). But residency checks normalize both the request URI and each policy URI before comparing (`pkg/sql/iceberg/residency.go:303-331`), and disabled rows are skipped (`pkg/sql/iceberg/residency.go:104-107`, `pkg/sql/iceberg/residency.go:155-157`).
> 
> That leaves the DB identity weaker than the runtime identity. For example, if there is an enabled policy for `https://catalog.example.com/rest`, a later registration with `catalog_uri=HTTPS://Catalog.Example.com/rest` and `policy_state=disabled` does not hit the duplicate-key update; it inserts a second row because the raw key differs. At runtime both URIs normalize to the same catalog URI, the disabled row is ignored, and the old enabled row still allows the catalog/object access. The same issue applies to any other URI form normalized by `NormalizeCatalogURI`, such as a fragment being stripped.
> 
> Please normalize `catalogURI` before building the `ResidencyPolicy` and before inserting/upserting it, or store a separate normalized URI/hash as the key identity. I would also add a regression test that equivalent URI variants upsert the same policy row and that disabling through an equivalent URI actually disables the existing enabled policy.

fixed

LeftHandCold:
The latest patch fixes canonical `allowed_catalog_uri`, but the same storage-vs-runtime identity mismatch still exists for other key fields.

`allowed_endpoint` and `allowed_region` are part of the residency policy primary key (`pkg/sql/iceberg/ddl.go:62-75`), and `iceberg_register_access` still stores the raw trimmed input for both (`pkg/frontend/iceberg_call.go:179-181`, `pkg/frontend/iceberg_call.go:253-262`). At runtime, however, endpoint is canonicalized by `NormalizeEndpoint` and region is lower-cased before matching (`pkg/sql/iceberg/residency.go:287-309`, `pkg/sql/iceberg/residency.go:351-360`).

So an existing enabled row for endpoint `s3.me-central-1.amazonaws.com`, region `me-central-1` will not be hit by a later `policy_state=disabled` registration using an equivalent endpoint such as `https://S3.ME-CENTRAL-1.AMAZONAWS.COM/` or region `ME-CENTRAL-1`; the duplicate-key update misses, a disabled variant is inserted, disabled rows are skipped, and the old enabled row still allows access.

Please normalize the full storage identity before insert/upsert, not just `AllowedCatalogURI`: at least `AllowedEndpoint = NormalizeEndpoint(...)` and `AllowedRegion = strings.ToLower(strings.TrimSpace(...))` (bucket is already only trim-sensitive in runtime matching). The regression should cover disabling through equivalent endpoint/region variants.

XuPeng-SH (changes requested):
I’m requesting changes. I found several substantive blockers across security, runtime correctness, and SQL surface:

1. **Write / maintenance paths bypass the new principal-mapping model.**
   The scan path runs `CheckCatalogAccess` and propagates `ExternalPrincipal`, but append / DML / maintenance catalog requests do not. That means writes can hit the external catalog under the shared catalog token even when the MatrixOne principal has no `mo_iceberg_principal_map` entry or should map to a less-privileged external identity. Please run the same principal-map check on write / maintenance paths and require `CatalogRequest.ExternalPrincipal` for all external catalog operations.

2. **Write / maintenance paths also skip catalog residency enforcement.**
   Scan planning installs residency validation, but append / DML / maintenance can still call arbitrary registered catalog URIs before any residency check. That defeats the new residency-policy model and creates an outbound catalog-access / SSRF-style hole. Please enforce catalog residency before the first outbound catalog request on every write / maintenance path, not just scans.

3. **`iceberg_register_access` has a privilege bypass for cluster-scope policy creation.**
   In `pkg/frontend/iceberg_call.go`, a non-system tenant can create `scope_type=cluster` rows whenever `MO_ICEBERG_ALLOW_PLAIN_HTTP=1` is set. That env var is a transport/local-dev flag (see `pkg/sql/compile/iceberg_default_planner.go`), not an authorization control, but it currently weakens a global policy-management check. Please keep cluster-scope registration strictly moadmin-only and do not key that privilege decision off the plain-HTTP dev flag.

4. **MERGE `DEFAULT` semantics are broken.**
   `pkg/sql/plan/iceberg_dml_merge.go` keeps raw `DEFAULT` AST nodes in MERGE update/insert expressions instead of lowering them to the target column default expression. The UPDATE path already expands `DEFAULT`; MERGE should go through the same `getDefaultExpr` + assignment-cast path. As written, `SET col = DEFAULT` / `VALUES (DEFAULT)` will not produce the column’s declared default.

5. **Equality deletes break when one scan sees delete files with different equality-field layouts.**
   `pkg/sql/colexec/external/iceberg_delete_apply.go` unions all equality field IDs into one probe tuple shape, but the delete index stores only raw equality keys, not the originating field layout. Legal combinations like `(id)` and `(id, region)` in the same scan can no longer both match correctly, so deleted rows may leak through. Please keep separate equality indexes per equality-field layout (or per delete task/schema) instead of collapsing them into one global tuple shape.

6. **MERGE does not reject multiple source rows matching the same target row.**
   The planner/runtime path treats MERGE as a plain join and never enforces one source row per target row. That allows duplicate UPDATE/DELETE actions against the same `(data_file_path, row_ordinal)`, which is not safe. Please detect duplicate target matches before commit and raise a MERGE cardinality error.

7. **`SHOW ICEBERG ... LIKE/WHERE` is parsed but ignored.**
   The AST carries `Like` / `Where` on `ShowIcebergCatalogs`, `ShowIcebergNamespaces`, and `ShowIcebergTables`, but the frontend handlers in `pkg/frontend/iceberg.go` never apply those filters. So the syntax is accepted while returning unfiltered results. Please either implement LIKE/WHERE filtering or reject the unsupported forms.

8. **Coverage is still weak at the distributed SQL layer.**
   Unit / embedded coverage is broad, but I did not find distributed BVT coverage for the core Iceberg SQL surface (`CREATE/ALTER/DROP ICEBERG CATALOG`, `SHOW ICEBERG ...`, duplicate/missing-option errors, and SQL-level non-admin denial). Given how much user-facing SQL and metadata this PR touches, please add at least one distributed happy-path + unhappy-path catalog test and one SQL-level privilege-denial test.

XuPeng-SH:
One more metadata/compatibility pass turned up a few additional issues that were not in my earlier review:

1. **The Iceberg feature flags do not actually gate the read/catalog surface.**
   `frontend.iceberg.enable` / `enable-per-account` are exposed as rollout controls, but catalog DDL/SHOW, Iceberg external-table DDL, and scan compilation do not enforce them. On top of that, write/maintenance paths hardcode `AccountConfig{Enable: true}` (`pkg/sql/iceberg/append_runtime_factory.go`, `pkg/sql/iceberg/dml_delete_runtime_factory.go`, `pkg/cnservice/server.go`). Please enforce the real `cfg.Enable` / `EffectiveForAccount(...)` checks consistently across catalog DDL, table DDL, scan compile, and runtime entry points.

2. **`DROP ICEBERG CATALOG` can strand orphan-cleanup metadata.**
   The drop path only checks `mo_iceberg_tables`, then deletes the catalog row. It does not guard against dependent rows like `mo_iceberg_orphan_files`, so later orphan cleanup can no longer resolve the catalog and the underlying objects are left behind. Please either block DROP while catalog-keyed metadata exists, or cascade/retire those rows transactionally.

3. **The global `FOR ICEBERG` lexer rewrite regresses existing `FOR iceberg` syntax.**
   `mysql_lexer.go` rewrites any `FOR` followed by `ICEBERG` into `FOR_ICEBERG` before parser context is known. That breaks existing forms that legitimately use `iceberg` as an identifier immediately after `FOR` (for example `SET PASSWORD FOR iceberg` / `SHOW GRANTS FOR iceberg` when the user or role name is `iceberg`). This needs to be context-sensitive in the grammar instead of a global lexer rewrite.

4. **`INSERT OVERWRITE PARTITION` can silently miss mixed-case partition fields.**
   The overwrite path lowercases partition field names in `build_insert.go`, but matching later does exact key lookups in `dml_delete_builder.go`. If the partition spec preserves original case, overwrite can fail to match any target files and behave like append. Please canonicalize overwrite keys against the partition spec case-insensitively or reuse the same case-insensitive lookup helper used by pruning.

iamlinjunhong:
> One more metadata/compatibility pass turned up a few additional issues that were not in my earlier review:
> 
> 1. **The Iceberg feature flags do not actually gate the read/catalog surface.**
>    `frontend.iceberg.enable` / `enable-per-account` are exposed as rollout controls, but catalog DDL/SHOW, Iceberg external-table DDL, and scan compilation do not enforce them. On top of that, write/maintenance paths hardcode `AccountConfig{Enable: true}` (`pkg/sql/iceberg/append_runtime_factory.go`, `pkg/sql/iceberg/dml_delete_runtime_factory.go`, `pkg/cnservice/server.go`). Please enforce the real `cfg.Enable` / `EffectiveForAccount(...)` checks consistently across catalog DDL, table DDL, scan compile, and runtime entry points.
> 2. **`DROP ICEBERG CATALOG` can strand orphan-cleanup metadata.**
>    The drop path only checks `mo_iceberg_tables`, then deletes the catalog row. It does not guard against dependent rows like `mo_iceberg_orphan_files`, so later orphan cleanup can no longer resolve the catalog and the underlying objects are left behind. Please either block DROP while catalog-keyed metadata exists, or cascade/retire those rows transactionally.
> 3. **The global `FOR ICEBERG` lexer rewrite regresses existing `FOR iceberg` syntax.**
>    `mysql_lexer.go` rewrites any `FOR` followed by `ICEBERG` into `FOR_ICEBERG` before parser context is known. That breaks existing forms that legitimately use `iceberg` as an identifier immediately after `FOR` (for example `SET PASSWORD FOR iceberg` / `SHOW GRANTS FOR iceberg` when the user or role name is `iceberg`). This needs to be context-sensitive in the grammar instead of a global lexer rewrite.
> 4. **`INSERT OVERWRITE PARTITION` can silently miss mixed-case partition fields.**
>    The overwrite path lowercases partition field names in `build_insert.go`, but matching later does exact key lookups in `dml_delete_builder.go`. If the partition spec preserves original case, overwrite can fail to match any target files and behave like append. Please canonicalize overwrite keys against the partition spec case-insensitively or reuse the same case-insensitive lookup helper used by pruning.

fixed

iamlinjunhong:
> I’m requesting changes. I found several substantive blockers across security, runtime correctness, and SQL surface:
> 
> 1. **Write / maintenance paths bypass the new principal-mapping model.**
>    The scan path runs `CheckCatalogAccess` and propagates `ExternalPrincipal`, but append / DML / maintenance catalog requests do not. That means writes can hit the external catalog under the shared catalog token even when the MatrixOne principal has no `mo_iceberg_principal_map` entry or should map to a less-privileged external identity. Please run the same principal-map check on write / maintenance paths and require `CatalogRequest.ExternalPrincipal` for all external catalog operations.
> 2. **Write / maintenance paths also skip catalog residency enforcement.**
>    Scan planning installs residency validation, but append / DML / maintenance can still call arbitrary registered catalog URIs before any residency check. That defeats the new residency-policy model and creates an outbound catalog-access / SSRF-style hole. Please enforce catalog residency before the first outbound catalog request on every write / maintenance path, not just scans.
> 3. **`iceberg_register_access` has a privilege bypass for cluster-scope policy creation.**
>    In `pkg/frontend/iceberg_call.go`, a non-system tenant can create `scope_type=cluster` rows whenever `MO_ICEBERG_ALLOW_PLAIN_HTTP=1` is set. That env var is a transport/local-dev flag (see `pkg/sql/compile/iceberg_default_planner.go`), not an authorization control, but it currently weakens a global policy-management check. Please keep cluster-scope registration strictly moadmin-only and do not key that privilege decision off the plain-HTTP dev flag.
> 4. **MERGE `DEFAULT` semantics are broken.**
>    `pkg/sql/plan/iceberg_dml_merge.go` keeps raw `DEFAULT` AST nodes in MERGE update/insert expressions instead of lowering them to the target column default expression. The UPDATE path already expands `DEFAULT`; MERGE should go through the same `getDefaultExpr` + assignment-cast path. As written, `SET col = DEFAULT` / `VALUES (DEFAULT)` will not produce the column’s declared default.
> 5. **Equality deletes break when one scan sees delete files with different equality-field layouts.**
>    `pkg/sql/colexec/external/iceberg_delete_apply.go` unions all equality field IDs into one probe tuple shape, but the delete index stores only raw equality keys, not the originating field layout. Legal combinations like `(id)` and `(id, region)` in the same scan can no longer both match correctly, so deleted rows may leak through. Please keep separate equality indexes per equality-field layout (or per delete task/schema) instead of collapsing them into one global tuple shape.
> 6. **MERGE does not reject multiple source rows matching the same target row.**
>    The planner/runtime path treats MERGE as a plain join and never enforces one source row per target row. That allows duplicate UPDATE/DELETE actions against the same `(data_file_path, row_ordinal)`, which is not safe. Please detect duplicate target matches before commit and raise a MERGE cardinality error.
> 7. **`SHOW ICEBERG ... LIKE/WHERE` is parsed but ignored.**
>    The AST carries `Like` / `Where` on `ShowIcebergCatalogs`, `ShowIcebergNamespaces`, and `ShowIcebergTables`, but the frontend handlers in `pkg/frontend/iceberg.go` never apply those filters. So the syntax is accepted while returning unfiltered results. Please either implement LIKE/WHERE filtering or reject the unsupported forms.
> 8. **Coverage is still weak at the distributed SQL layer.**
>    Unit / embedded coverage is broad, but I did not find distributed BVT coverage for the core Iceberg SQL surface (`CREATE/ALTER/DROP ICEBERG CATALOG`, `SHOW ICEBERG ...`, duplicate/missing-option errors, and SQL-level non-admin denial). Given how much user-facing SQL and metadata this PR touches, please add at least one distributed happy-path + unhappy-path catalog test and one SQL-level privilege-denial test.

fixed

iamlinjunhong:
> One more metadata/compatibility pass turned up a few additional issues that were not in my earlier review:
> 
> 1. **The Iceberg feature flags do not actually gate the read/catalog surface.**
>    `frontend.iceberg.enable` / `enable-per-account` are exposed as rollout controls, but catalog DDL/SHOW, Iceberg external-table DDL, and scan compilation do not enforce them. On top of that, write/maintenance paths hardcode `AccountConfig{Enable: true}` (`pkg/sql/iceberg/append_runtime_factory.go`, `pkg/sql/iceberg/dml_delete_runtime_factory.go`, `pkg/cnservice/server.go`). Please enforce the real `cfg.Enable` / `EffectiveForAccount(...)` checks consistently across catalog DDL, table DDL, scan compile, and runtime entry points.
> 2. **`DROP ICEBERG CATALOG` can strand orphan-cleanup metadata.**
>    The drop path only checks `mo_iceberg_tables`, then deletes the catalog row. It does not guard against dependent rows like `mo_iceberg_orphan_files`, so later orphan cleanup can no longer resolve the catalog and the underlying objects are left behind. Please either block DROP while catalog-keyed metadata exists, or cascade/retire those rows transactionally.
> 3. **The global `FOR ICEBERG` lexer rewrite regresses existing `FOR iceberg` syntax.**
>    `mysql_lexer.go` rewrites any `FOR` followed by `ICEBERG` into `FOR_ICEBERG` before parser context is known. That breaks existing forms that legitimately use `iceberg` as an identifier immediately after `FOR` (for example `SET PASSWORD FOR iceberg` / `SHOW GRANTS FOR iceberg` when the user or role name is `iceberg`). This needs to be context-sensitive in the grammar instead of a global lexer rewrite.
> 4. **`INSERT OVERWRITE PARTITION` can silently miss mixed-case partition fields.**
>    The overwrite path lowercases partition field names in `build_insert.go`, but matching later does exact key lookups in `dml_delete_builder.go`. If the partition spec preserves original case, overwrite can fail to match any target files and behave like append. Please canonicalize overwrite keys against the partition spec case-insensitively or reuse the same case-insensitive lookup helper used by pruning.

fixed

XuPeng-SH (changes requested):
I am requesting changes after another first-principles pass. Several earlier blockers look substantially improved (principal/residency is now wired into write/maintenance paths, SHOW ICEBERG LIKE/WHERE is rejected instead of ignored, the FOR ICEBERG lexer rewrite is narrower, and MERGE duplicate matched-target detection was added), but I still see blockers:

1. **High: multi-CN Iceberg scan fanout can send an unusable process-local `ObjectIORef` to remote CNs.**

`compileExternScanIcebergFileFanout` still creates remote shard scopes when there is more than one shard (`pkg/sql/compile/compile.go`), then `attachIcebergRuntimeToExternal` stores the same `runtime.objectIORef` on each external operator and only retains it in the coordinator process. Remote-run serialization copies only the string (`pkg/sql/compile/remoterun.go`). The registry behind `ResolveObjectIORef` is just a process-local global map (`pkg/iceberg/io/object_io_registry.go`), so a remote CN process that receives the string has no registered provider to resolve.

The policy gate in `validateIcebergRemoteFanoutPolicy` allows remote fanout when `ProtectedCNToCN` is enabled, and also allows the no-credential-scope remote-signing case when `EnableRemoteSign` is enabled. Neither condition registers/rebuilds the provider on the remote CN. Remote signing helps the coordinator-side provider sign URLs, but the remote CN still only has an opaque ref string.

Please either force Iceberg scans to the current CN until the handoff is implemented, or make the remote operator rebuild/register an equivalent object IO provider on that CN under the protected transport model. The regression needs to exercise real remote-run semantics with an isolated remote registry; encode/decode roundtrip tests that only preserve the string are not enough.

2. **High: `DROP ICEBERG CATALOG` leaves catalog-scoped metadata behind, and `catalog_id` can be reused.**

The drop path now checks `mo_iceberg_tables` and `mo_iceberg_orphan_files`, then deletes only the catalog row (`pkg/frontend/iceberg.go`). It does not block or clean rows in `mo_iceberg_principal_map`, `mo_iceberg_residency_policy`, `mo_iceberg_refs`, `mo_iceberg_publish_jobs`, or `mo_iceberg_maintenance_jobs`. Since new catalog ids are allocated with `max(catalog_id)+1` within the account, dropping the last catalog allows the same numeric id to be reused.

That means stale principal mappings can silently authorize a newly-created, unrelated catalog with the reused id; stale residency/ref/job metadata can also interfere with the new catalog. This is especially risky because principal-map identity is only `(account_id, catalog_id, role/user)` and has no catalog URI/name discriminator.

Please make catalog drop transactional over the full catalog-scoped metadata set: either block while any dependent metadata exists, or cascade/retire every dependent row deliberately. Add a regression that creates a catalog, registers access, drops it, recreates a catalog that reuses the id, and proves the old principal/residency state cannot apply to the new catalog.

3. **Medium: MERGE INSERT column-list semantics use NULL for omitted columns instead of column defaults.**

`icebergMergeInsertExprMap` fills every omitted nullable target column with an explicit NULL and rejects omitted NOT NULL columns (`pkg/sql/plan/iceberg_dml_merge.go`). That is not the normal INSERT/MERGE column-list contract: omitted columns should use the target column default; only columns with no default and nullable type should become NULL. The PR even adds a test that codifies the NULL behavior, which will make this wrong semantics permanent.

Please use the same default expansion path already used by explicit MERGE `DEFAULT` (`getDefaultExpr` + assignment cast) for omitted columns. Add tests for a nullable column with a non-NULL default, a NOT NULL column with a default, and a nullable column with no default.

CI is also not finished on the current head, so this is not approvable yet even aside from the blockers above.

iamlinjunhong:
> I am requesting changes after another first-principles pass. Several earlier blockers look substantially improved (principal/residency is now wired into write/maintenance paths, SHOW ICEBERG LIKE/WHERE is rejected instead of ignored, the FOR ICEBERG lexer rewrite is narrower, and MERGE duplicate matched-target detection was added), but I still see blockers:
> 
> 1. **High: multi-CN Iceberg scan fanout can send an unusable process-local `ObjectIORef` to remote CNs.**
> 
> `compileExternScanIcebergFileFanout` still creates remote shard scopes when there is more than one shard (`pkg/sql/compile/compile.go`), then `attachIcebergRuntimeToExternal` stores the same `runtime.objectIORef` on each external operator and only retains it in the coordinator process. Remote-run serialization copies only the string (`pkg/sql/compile/remoterun.go`). The registry behind `ResolveObjectIORef` is just a process-local global map (`pkg/iceberg/io/object_io_registry.go`), so a remote CN process that receives the string has no registered provider to resolve.
> 
> The policy gate in `validateIcebergRemoteFanoutPolicy` allows remote fanout when `ProtectedCNToCN` is enabled, and also allows the no-credential-scope remote-signing case when `EnableRemoteSign` is enabled. Neither condition registers/rebuilds the provider on the remote CN. Remote signing helps the coordinator-side provider sign URLs, but the remote CN still only has an opaque ref string.
> 
> Please either force Iceberg scans to the current CN until the handoff is implemented, or make the remote operator rebuild/register an equivalent object IO provider on that CN under the protected transport model. The regression needs to exercise real remote-run semantics with an isolated remote registry; encode/decode roundtrip tests that only preserve the string are not enough.
> 
> 2. **High: `DROP ICEBERG CATALOG` leaves catalog-scoped metadata behind, and `catalog_id` can be reused.**
> 
> The drop path now checks `mo_iceberg_tables` and `mo_iceberg_orphan_files`, then deletes only the catalog row (`pkg/frontend/iceberg.go`). It does not block or clean rows in `mo_iceberg_principal_map`, `mo_iceberg_residency_policy`, `mo_iceberg_refs`, `mo_iceberg_publish_jobs`, or `mo_iceberg_maintenance_jobs`. Since new catalog ids are allocated with `max(catalog_id)+1` within the account, dropping the last catalog allows the same numeric id to be reused.
> 
> That means stale principal mappings can silently authorize a newly-created, unrelated catalog with the reused id; stale residency/ref/job metadata can also interfere with the new catalog. This is especially risky because principal-map identity is only `(account_id, catalog_id, role/user)` and has no catalog URI/name discriminator.
> 
> Please make catalog drop transactional over the full catalog-scoped metadata set: either block while any dependent metadata exists, or cascade/retire every dependent row deliberately. Add a regression that creates a catalog, registers access, drops it, recreates a catalog that reuses the id, and proves the old principal/residency state cannot apply to the new catalog.
> 
> 3. **Medium: MERGE INSERT column-list semantics use NULL for omitted columns instead of column defaults.**
> 
> `icebergMergeInsertExprMap` fills every omitted nullable target column with an explicit NULL and rejects omitted NOT NULL columns (`pkg/sql/plan/iceberg_dml_merge.go`). That is not the normal INSERT/MERGE column-list contract: omitted columns should use the target column default; only columns with no default and nullable type should become NULL. The PR even adds a test that codifies the NULL behavior, which will make this wrong semantics permanent.
> 
> Please use the same default expansion path already used by explicit MERGE `DEFAULT` (`getDefaultExpr` + assignment cast) for omitted columns. Add tests for a nullable column with a non-NULL default, a NOT NULL column with a default, and a nullable column with no default.
> 
> CI is also not finished on the current head, so this is not approvable yet even aside from the blockers above.

fixed

aunjgr (approved):

XuPeng-SH (changes requested):
Requesting changes on the latest head after verifying that the three blockers from my Jul 9 review are fixed.

**High — the configured manifest-cache memory limit is not enforced, so a long-lived CN can retain Iceberg metadata without bound.**

`pkg/config/configuration.go:251-253` defaults `iceberg.scan.manifest_cache_bytes` to 256 MiB, and the value is carried into `cfg.Scan.ManifestCacheBytes`, but `pkg/sql/compile/iceberg_default_planner.go:83` constructs the process-wide cache with only the TTL. `pkg/iceberg/metadata/cache.go:61-73,136-145` has no byte/capacity accounting or eviction at all; `SizeBytes` is stored but never used to bound the map. Moreover, `Get` deliberately keeps expired entries that have an ETag (`cache.go:94-98`) for revalidation, so a manifest key that is never requested again is never removed.

Every new snapshot normally introduces new manifest-list and manifest-path keys, whose values include cloned metadata JSON, manifest lists, and all manifest entries. On a long-lived CN that scans evolving tables, those keys accumulate across snapshots/catalogs/principals with no hard bound. The advertised 256 MiB setting therefore gives a false safety guarantee and this becomes an eventual OOM path.

Please wire `ManifestCacheBytes` into the cache and enforce a real weighted bound (including replacement accounting and deterministic/concurrency-safe eviction). Expired stale entries may be retained for ETag revalidation only while they fit inside that bound. Please add tests covering: weighted eviction below the configured limit, replacement accounting, oversized single entries, expired ETag entries under key churn, and concurrent Get/Put/Invalidate.

Audit notes: creation→destruction is closed for ObjectIO refs and failed write artifacts; waits terminate by cancellation/ticker stop; the remaining failed proposition is accumulation→bound for the manifest cache. Targeted core Iceberg tests, previous-blocker tests, and coordinator cache race tests pass, and the latest CI is green.

iamlinjunhong:
> Requesting changes on the latest head after verifying that the three blockers from my Jul 9 review are fixed.
> 
> **High — the configured manifest-cache memory limit is not enforced, so a long-lived CN can retain Iceberg metadata without bound.**
> 
> `pkg/config/configuration.go:251-253` defaults `iceberg.scan.manifest_cache_bytes` to 256 MiB, and the value is carried into `cfg.Scan.ManifestCacheBytes`, but `pkg/sql/compile/iceberg_default_planner.go:83` constructs the process-wide cache with only the TTL. `pkg/iceberg/metadata/cache.go:61-73,136-145` has no byte/capacity accounting or eviction at all; `SizeBytes` is stored but never used to bound the map. Moreover, `Get` deliberately keeps expired entries that have an ETag (`cache.go:94-98`) for revalidation, so a manifest key that is never requested again is never removed.
> 
> Every new snapshot normally introduces new manifest-list and manifest-path keys, whose values include cloned metadata JSON, manifest lists, and all manifest entries. On a long-lived CN that scans evolving tables, those keys accumulate across snapshots/catalogs/principals with no hard bound. The advertised 256 MiB setting therefore gives a false safety guarantee and this becomes an eventual OOM path.
> 
> Please wire `ManifestCacheBytes` into the cache and enforce a real weighted bound (including replacement accounting and deterministic/concurrency-safe eviction). Expired stale entries may be retained for ETag revalidation only while they fit inside that bound. Please add tests covering: weighted eviction below the configured limit, replacement accounting, oversized single entries, expired ETag entries under key churn, and concurrent Get/Put/Invalidate.
> 
> Audit notes: creation→destruction is closed for ObjectIO refs and failed write artifacts; waits terminate by cancellation/ticker stop; the remaining failed proposition is accumulation→bound for the manifest cache. Targeted core Iceberg tests, previous-blocker tests, and coordinator cache race tests pass, and the latest CI is green.

fixed

XuPeng-SH (changes requested):
I re-reviewed the latest head with focus on the post-Jul-13 delta. The earlier manifest-cache byte-bound blocker looks substantially fixed, but I still found a merge blocker.

**High: Nessie branch/tag scans can reuse cached metadata from a different ref.**

The latest head rewrites Nessie refs into the catalog prefix, but the metadata/manifests cache key still does not include that effective prefix.

Concrete failure path:

1. `RuntimeScanPlanner.resolveCatalogPrefix` rewrites a request like `tag:release` or `tag:hotfix` into different `req.Prefix` values (for example `release|s3://warehouse` vs `hotfix|s3://warehouse`) and then forces `req.Ref = ""` plus `req.Snapshot.RefName = "main"` (`pkg/iceberg/metadata/runtime_planner.go:143-149`).
2. `LocalScanPlanner.PlanScan` normalizes that rewritten request and passes `ExternalRef: selector.RefName` into `CachedTableMetadataLoader` (`pkg/iceberg/metadata/scan_planner.go:70-85`). After the rewrite, `selector.RefName` is just `main` for both requests.
3. `CachedTableMetadataLoader.baseKey` builds the metadata/location cache key from account/catalog/table/namespace/ref/snapshot/principal/credential hash, but it does **not** include `req.Prefix` (`pkg/iceberg/metadata/cache_loader.go:227-242`).
4. The cache hit path returns cached metadata **before any second `LoadTable` call** (`pkg/iceberg/metadata/cache_loader.go:72-80`). So a first scan of `tag:release` can populate the cache under `Ref=main`; a later scan of `tag:hotfix` with the same table/principal/credentials will hit that entry and reuse the `release` metadata until TTL/invalidation, producing the wrong snapshot/manifest set for `hotfix`.

I verified this both by reading the code path and with a deterministic scratch repro in a throwaway worktree: two post-rewrite requests that differed only by `Prefix` (`release|...` vs `hotfix|...`) produced the exact same `metadata_location` cache key and the second load hit cache with only one catalog call.

Please include the effective catalog prefix (or the original external ref before the Nessie rewrite) in the metadata/location and manifest cache keys. The regression should scan the same table across two Nessie refs with shared credentials and prove the second ref does not reuse the first ref's cached metadata/manifests.

iamlinjunhong:
> I re-reviewed the latest head with focus on the post-Jul-13 delta. The earlier manifest-cache byte-bound blocker looks substantially fixed, but I still found a merge blocker.
> 
> **High: Nessie branch/tag scans can reuse cached metadata from a different ref.**
> 
> The latest head rewrites Nessie refs into the catalog prefix, but the metadata/manifests cache key still does not include that effective prefix.
> 
> Concrete failure path:
> 
> 1. `RuntimeScanPlanner.resolveCatalogPrefix` rewrites a request like `tag:release` or `tag:hotfix` into different `req.Prefix` values (for example `release|s3://warehouse` vs `hotfix|s3://warehouse`) and then forces `req.Ref = ""` plus `req.Snapshot.RefName = "main"` (`pkg/iceberg/metadata/runtime_planner.go:143-149`).
> 2. `LocalScanPlanner.PlanScan` normalizes that rewritten request and passes `ExternalRef: selector.RefName` into `CachedTableMetadataLoader` (`pkg/iceberg/metadata/scan_planner.go:70-85`). After the rewrite, `selector.RefName` is just `main` for both requests.
> 3. `CachedTableMetadataLoader.baseKey` builds the metadata/location cache key from account/catalog/table/namespace/ref/snapshot/principal/credential hash, but it does **not** include `req.Prefix` (`pkg/iceberg/metadata/cache_loader.go:227-242`).
> 4. The cache hit path returns cached metadata **before any second `LoadTable` call** (`pkg/iceberg/metadata/cache_loader.go:72-80`). So a first scan of `tag:release` can populate the cache under `Ref=main`; a later scan of `tag:hotfix` with the same table/principal/credentials will hit that entry and reuse the `release` metadata until TTL/invalidation, producing the wrong snapshot/manifest set for `hotfix`.
> 
> I verified this both by reading the code path and with a deterministic scratch repro in a throwaway worktree: two post-rewrite requests that differed only by `Prefix` (`release|...` vs `hotfix|...`) produced the exact same `metadata_location` cache key and the second load hit cache with only one catalog call.
> 
> Please include the effective catalog prefix (or the original external ref before the Nessie rewrite) in the metadata/location and manifest cache keys. The regression should scan the same table across two Nessie refs with shared credentials and prove the second ref does not reuse the first ref's cached metadata/manifests.

fixed

pkg/sql/colexec/icebergwrite/icebergwrite.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (w *IcebergWrite) appendBatch(proc *process.Process, bat *batch.Batch) error {
+	if coordinator, ok := w.Coordinator.(ProcessAwareCoordinator); ok {
+		return coordinator.AppendWithProcess(proc, bat)
+	}
+	return w.Coordinator.Append(proc.Ctx, bat)
+}
+
+func (w *IcebergWrite) Reset(proc *process.Process, pipelineFailed bool, err error) {
+	w.abortOpen(proc, err)
```
LeftHandCold, line 122: [High] `Reset` makes the operator look unopened again but retains the same `Coordinator`. Cached prepared pipelines are explicitly reused via `Compile.Reset`; after the first execution, the runtime shared coordinator has `commitAttempted=true`, so the second execution reaches `Prepare` and fails in `Begin` as already committed. There is a second identity problem on the same path: `Request.StatementID` / `IdempotencyKey` were captured when the PREPARE pipeline was compiled, not from the current EXECUTE process, so merely rebuilding the coordinator would let later executions reuse the first execution’s commit key. Please discard/recreate factory-owned coordinator state on reset and refresh the execution-specific statement/idempotency identity before calling the factory. Add a regression that executes the same cached prepared Iceberg INSERT twice and verifies two independent commits.

pkg/sql/iceberg/dml_executor.go
```diff
@@ -0,0 +1,325 @@
...
+		req.SnapshotID = e.SnapshotID
+	}
+	if err := e.validateCommitPrerequisites(ctx, req.TableLocation, req.SnapshotID); err != nil {
+		return DMLCommitActionStreamResult{}, err
+	}
+	tracker := newDMLMaterializedObjectTracker()
+	wrapDMLDeleteRequestWriters(&req, tracker)
+	stream, err := BuildDMLDeleteActionStream(ctx, req)
+	if err != nil {
+		_ = e.recordMaterializedOrphans(ctx, req.TableLocation, req.Base, tracker.paths())
```
LeftHandCold: [High] This is the only recovery path for objects materialized before action-stream construction fails, but the orphan-recorder error is discarded. If the store is unavailable, the method returns only the build error and leaves the uploaded objects untracked, so the sweeper can never discover them. The same `_ = e.recordMaterializedOrphans(...)` pattern exists in all four Commit methods. Please preserve both failures (for example, `errors.Join(err, recordErr)`) and add a recorder-failure regression.

pkg/iceberg/io/signed_http_fs.go
```diff
@@ -0,0 +1,340 @@
...
+	if resp.StatusCode == http.StatusOK && entry.Offset > 0 {
+		if _, err := io.CopyN(io.Discard, resp.Body, entry.Offset); err != nil {
+			return nil, io.ErrUnexpectedEOF
+		}
+	}
+	reader := io.Reader(resp.Body)
+	if entry.Size >= 0 {
+		reader = io.LimitReader(reader, entry.Size)
+	}
+	data, err := io.ReadAll(reader)
```
LeftHandCold: [High] `entry.Size == -1` leaves this `ReadAll` completely unbounded. That is a production path: manifest/DML readers and native maintenance issue `Read(..., 0, -1)`, and remote-signing resolves them to this FileService. A catalog/object endpoint can therefore return an arbitrarily large body and OOM the CN; the REST client already has a response cap, but this path does not. Please stream when the caller requested `ReadCloserForRead`, and enforce a configured/hard maximum for materialized unknown-length reads (including a regression with an oversized response).

LeftHandCold (changes requested):
Re-reviewed the latest head after the recent write-planning memory changes. The earlier cache/ref blockers look addressed, and the targeted Iceberg suites pass, but three production blockers remain: cached prepared executions reuse terminal write state and a stale idempotency identity; DML build-failure cleanup discards orphan-recorder failures; and signed-HTTP reads still allow an unbounded response for Size=-1. Please address the inline comments and add the requested regressions.

iamlinjunhong:
> Re-reviewed the latest head after the recent write-planning memory changes. The earlier cache/ref blockers look addressed, and the targeted Iceberg suites pass, but three production blockers remain: cached prepared executions reuse terminal write state and a stale idempotency identity; DML build-failure cleanup discards orphan-recorder failures; and signed-HTTP reads still allow an unbounded response for Size=-1. Please address the inline comments and add the requested regressions.

fixed

XuPeng-SH (approved):
Approved on latest head `5b0921e29434b2c0e0f56c660e82081dca78eb34`.

I re-verified the previous blockers end to end:

- prepared/cached Iceberg writes discard terminal factory coordinators on reset and refresh statement/idempotency identity for each execution;
- all four DML build-failure paths track materialized objects and preserve orphan-recorder failures with the original build error;
- signed HTTP unknown-length reads stream when requested and otherwise enforce a hard aggregate materialization bound with response-body ownership closed;
- the metadata cache is weighted/LRU bounded and Nessie effective prefixes participate in metadata and manifest keys;
- active ObjectIO refs survive TTL sweeping, are released by operator teardown, and process-local refs are explicitly prevented from remote-CN fanout;
- earlier privilege, residency, catalog-drop dependency, MERGE default/cardinality, equality-delete layout, timestamp, and overwrite-partition issues remain fixed after the rebase.

Unhappy-path audit: creation-to-destruction, wait-to-termination, and accumulation-to-bound all close on the reviewed paths. Focused Iceberg/frontend/compile/operator tests pass, the cache/ObjectIO/coordinator race matrix passes, `go vet` is clean, and current CI including Iceberg E2E is green.

LeftHandCold (approved):
Re-reviewed latest head 5c76e1fb2bbeaddfd7d0f8962477d31398b0947b, focusing on the one new post-approval commit and its regression surface.

The DML change now keeps coordinator construction execution-scoped in IcebergWrite.Prepare while giving the sink its own retained ObjectIORef, so scan teardown cannot invalidate commit-time reads and cached executions do not capture compile-time coordinator state.

The maintenance change durably records post-commit cleanup intent before any object/catalog mutation. Failed or still-referenced candidates remain retryable and rotate through the queue; the cleaner continues to prove non-reference across retained snapshots before deleting. The idempotent orphan upsert and optimistic version transitions are consistent with partial recorder retries and concurrent stale candidates.

I also rechecked the single-writer boundary and normal append/DML/maintenance paths for regressions. No actionable blocker found. Targeted tests passed for pkg/iceberg/maintenance, pkg/sql/colexec/icebergwrite, pkg/sql/iceberg, and pkg/sql/compile.

Issue #25242, opened 2026-06-30
[Bug]: REGEXP and RLIKE accept empty pattern unlike MySQL


Ariznawlll:
### Environment

- Branch: latest `main`
- Commit: `33e96877604ad4b847ea7d5c55eb09429de1a5b6`
- MySQL oracle: 8.0.45
- MO: `8.0.30-MatrixOne-v1.3.0`

### Steps to Reproduce

```sql
select 'abc' regexp '' as regexp_empty_pattern;
select 'abc' rlike '' as rlike_empty_pattern;
```

### MySQL Result

```text
ERROR 3685 (HY000): Illegal argument to a regular expression.
```

### MatrixOne Result

```text
regexp_empty_pattern
1

rlike_empty_pattern
1
```

### Expected Behavior

MatrixOne should reject an empty regular expression pattern the same way MySQL does instead of returning a match result.

Ariznawlll:
Verified on latest main 2a008285e.

Result: fixed.

Validation:
```sql
select 'abc' regexp '';
select 'abc' rlike '';
```

MO now matches MySQL: both empty-pattern `REGEXP` and `RLIKE` raise error 3685.

Regression added in matrixorigin/motr#39, commit 75bf03f:
- `suites/14_issue_regression/t/issue_25242_regexp_empty_pattern.test`
- `suites/14_issue_regression/r/issue_25242_regexp_empty_pattern.result`

MOTR validation passed:
```
[ISSUE_25242_REGEXP_EMPTY_PATTERN] success
```

PR #25372, opened 2026-07-02
fix: reject empty regexp patterns

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Feature
- [ ] Improvement
- [ ] Documentation
- [ ] Enhancement
- [ ] Other

issue #25242

MatrixOne accepted empty patterns for `REGEXP`, `RLIKE`, and `REGEXP_LIKE` because Go's regular expression engine treats an empty pattern as a valid zero-width match.

This change adds predicate-specific empty-pattern checks that return MySQL-compatible error 3685 (`HY000`). It covers `REGEXP`, `RLIKE`, `NOT REGEXP`, and both the two-argument and three-argument forms of `REGEXP_LIKE`, while leaving the shared matcher used by other regular expression functions unchanged. NULL and valid-pattern behavior remain unchanged.

The change also adds focused execution tests and BVT coverage for all affected forms.

## Testing

- `make`
- `go build ./pkg/common/moerr ./pkg/sql/plan/function`
- `go vet ./pkg/common/moerr ./pkg/sql/plan/function`
- `go test -count=1 ./pkg/common/moerr`
- `go test -count=1 ./pkg/sql/plan/function`
- `mo-tester`: 67/67 cases passed


XuPeng-SH (approved):
Reviewed the latest head (`e836558d0`) with the issue scope in mind.

The fix is scoped correctly for #25242: `REGEXP` / `RLIKE` / `NOT REGEXP` now route through a match-specific compiler guard, `REGEXP_LIKE` rejects empty patterns explicitly, and NULL inputs still short-circuit before the empty-pattern check. The shared regexp compiler remains unchanged for the other regexp functions, so this does not accidentally change `REGEXP_SUBSTR` / `REGEXP_REPLACE` / `REGEXP_INSTR` behavior.

Unhappy-path pass:

- empty non-NULL pattern returns MySQL error 3685 / HY000.
- NULL pattern still returns NULL.
- valid positive/negative regexp paths still work.
- no new resource ownership, wait, goroutine, or unbounded accumulation path introduced.

Local verification:

- `go test -count=1 -timeout 120s ./pkg/common/moerr`
- `go test ... ./pkg/sql/plan/function -run 'Test_BuiltIn_.*Regexp|Test_BuiltIn_RegMatch|Test_BuiltIn_Regular'`
- `go test ... ./pkg/sql/plan/function`
- `go build ... ./pkg/common/moerr ./pkg/sql/plan/function`
- `go vet ... ./pkg/common/moerr ./pkg/sql/plan/function`

Approving. CI still had several jobs in progress when I checked, so merge should still wait for the normal required checks.

heni02 (approved):

aunjgr (approved):
Clean, well-scoped. Empty regexp → error 3685, NULL still NULL.

PR #25373, opened 2026-07-02
fix: use proper MySQL error for scalar subquery returning multiple rows

aunjgr:
## What type of PR is this?

- [x] BUG

issue #23062

Scalar subquery returning multiple rows used `internal error:` prefix (wrong error class). MySQL uses `ER_SUBQUERY_NO_1_ROW` (1242, SQLSTATE 21000).

### Fix

1. Registered `ErrSubqueryNo1Row` in `errorMsgRefer` map with proper error code (1242) and SQLSTATE (21000)
2. Added `NewErrSubqueryNo1Row` helper
3. Replaced 9 `NewInternalError` calls in hashjoin (7) and loopjoin (2)

Error message retained the "scalar" qualifier for clarity: `ERROR 1242 (21000): scalar subquery returns more than 1 row`

| File | Change |
|------|--------|
| `pkg/common/moerr/error.go` | +5 lines |
| `pkg/sql/colexec/hashjoin/join.go` | 7 locations |
| `pkg/sql/colexec/loopjoin/join.go` | 2 locations |
| `test/.../cte.result` | regenerated |
| `test/.../seq_func2.result` | regenerated |
| `test/.../subquery-with-in.result` | regenerated |


XuPeng-SH (changes requested):
The core direction is correct: all nine known execution sites now map the cardinality violation to MySQL error 1242 / SQLSTATE 21000, and I do not see a meaningful hot-path, allocation, or concurrency regression. However, this is not ready to merge for the following reasons.

1. **High: the regenerated result files bless unrelated behavior regressions.**

   `test/distributed/cases/sequence/seq_func2.result` is rewritten end-to-end (64 additions / 64 deletions) even though this PR does not change sequence behavior. More importantly, `subquery-with-in.result` changes 1,053 lines and replaces unrelated successful-query baselines with errors, for example:

   - `SELECT * from t2 where topic IN (SELECT SUM(topic) FROM t1)` now expects `correlated columns in aggregate function is not yet implemented`.
   - A pre-existing nested tuple-IN query now expects `unsupported expression executor ... is not yet implemented`.

   Those failures are unrelated to ER_SUBQUERY_NO_1_ROW and must not be silently accepted by this PR. Please restore the unrelated baselines and update only the error lines exercised by this fix plus the newly added cases. Also preserve the file's existing line endings; the new SQL block currently makes `git diff --check` report CRLF/trailing-whitespace noise.

2. **Medium: the user-visible message is not the MySQL-compatible message this issue asks for.**

   MySQL ER_SUBQUERY_NO_1_ROW and this repository's generated `mysql_error_define.go` both define the message as exactly `Subquery returns more than 1 row`. The new mapping instead emits `scalar subquery returns more than 1 row`. Since #23062 was found by the official MySQL compatibility suite, please use the canonical spelling/capitalization rather than retaining a MatrixOne-specific variant.

3. **Medium: the protocol contract is not tested.**

   The new unit test only checks the MatrixOne-internal code. The BVT result only checks display text, so neither test would catch a regression back to the wrong MySQL code or SQLSTATE. Please assert at least:

   - `MySQLCode() == ER_SUBQUERY_NO_1_ROW` (1242)
   - `SqlState() == "21000"`
   - the exact canonical message
   - preservation across `MarshalBinary` / `UnmarshalBinary`

   A client/protocol-level assertion would provide the strongest closure, because that is the externally reported bug.

4. **Medium: most modified execution paths remain untested, and Coverage is red.**

   The Coverage job reports 40% overall for the modified blocks and only 20% in `hashjoin/join.go`: six of the seven changed hashjoin branches were never executed. The added SQL cases all being green does not prove the left/right SINGLE variants are covered; the planner selected only one hashjoin path. Add deterministic operator tests for the distinct invariants: unique/PK right SINGLE duplicate detection, non-unique left SINGLE with multiple selections, non-equi left/right SINGLE, and multi-worker `syncBitmap` duplicate detection. Alternatively, centralize construction/checking enough that the cardinality invariant has one directly testable error path.

Please keep the final diff scoped to the error mapping, all actual throw sites, precise contract tests, and focused BVT cases.

aunjgr:
Thanks for the detailed review. The current head is `710c7b6dc` after merging the latest `upstream/main`.

Regarding the four items:

1. The result-file rewrite is intentional output regenerated by the newest mo-tester, rather than manually edited or selectively blessed output. The decorated result format is already used by current BVT results. Also, `subquery-with-in.sql` already uses CRLF throughout, so the `git diff --check` reports are caused by that existing file convention rather than newly mixed line endings. I therefore kept the generated baselines.

2. Fixed. `ErrSubqueryNo1Row` now uses the canonical MySQL text exactly: `Subquery returns more than 1 row`.

3. Fixed. `TestErrSubqueryNo1RowContract` now asserts:
   - MO error code
   - MySQL error 1242 / `ER_SUBQUERY_NO_1_ROW`
   - SQLSTATE `21000`
   - exact canonical message
   - `MarshalBinary` / `UnmarshalBinary` round-trip preservation

4. Fixed. Added deterministic operator tests covering all distinct modified paths:
   - PK/unique right SINGLE, with and without a non-equi condition
   - non-unique left SINGLE with multiple selections
   - non-unique right SINGLE duplicate detection
   - non-equi left/right SINGLE
   - multi-worker `syncBitmap` duplicate detection
   - both loopjoin SINGLE error paths

Local PR changed-line coverage is now 10/10 executable production blocks (100%): moerr 1/1, hashjoin 7/7, loopjoin 2/2. The new CI Coverage job is currently running.

Post-merge verification is green:
- unit tests, build, and vet for moerr/hashjoin/loopjoin
- `cte.sql`: 145 passed, 0 failed, 1 ignored
- `seq_func2.sql`: 75 passed, 0 failed
- `subquery-with-in.sql`: 744 passed, 0 failed, 21 ignored

Please take another look when convenient.

XuPeng-SH (approved):
Re-reviewed the latest head `710c7b6dce`. The prior blockers look addressed: the user-visible message is now the canonical MySQL text, `NewErrSubqueryNo1Row` is covered for code/SQLSTATE/message plus marshal round-trip, the BVT baselines are reduced to the intended scalar-subquery error lines, and the hashjoin/loopjoin unit tests now cover the left/right single duplicate paths plus the multi-worker bitmap merge path. I reran `go test ./pkg/common/moerr ./pkg/sql/colexec/hashjoin ./pkg/sql/colexec/loopjoin -count=1` and focused coverage for hashjoin/loopjoin, and I do not see a remaining high-confidence correctness, compatibility, performance, or unhappy-path blocker on this head.

heni02 (approved):

PR #25391, opened 2026-07-02
fix str_to_date fractional seconds precision

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25305

fix str_to_date fractional seconds precision


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (approved):
Reviewed the scale-propagation chain for `str_to_date` / `to_date` from binder through old-function type inference to function return type selection.

I don't see a blocking issue in the current fix.

What I checked:
- `%f` is now detected at plan time without being confused by escaped `%%f`
- the hidden third argument now carries both the inferred temporal kind and the fractional-second scale
- the datetime/time overloads return the normalized scale from that hidden argument instead of hard-coding scale 0
- the new planner/unit tests and BVT output line up with the intended precision-preservation behavior

Approving.

aunjgr (approved):

Issue #25294, opened 2026-07-01
[Bug]: CHAR() does not match MySQL argument coercion and NULL skipping


Ariznawlll:
### Environment

- Branch: latest `main`
- Commit tested locally: `200bfa7e0efb` (`fix: improve skills SKILL.md compatibility format (#25271)`)
- MySQL oracle: local MySQL 8.x via `mysql -u root -p...`
- MO: `8.0.30-MatrixOne-v1.3.0`

### Steps to Reproduce

```sql
select hex(char(77,77.3,'77.3')) as char_noninteger;
select hex(char(null,65,66)) as char_skip_null;
```

### MySQL Result

```text
char_noninteger
4D4D4D

char_skip_null
4142
```

MySQL `CHAR()` interprets arguments as integers and skips `NULL` arguments.

### MatrixOne Result

```text
select hex(char(77,77.3,'77.3'));
ERROR 20203 (HY000): invalid argument cast to int, bad value 77.3

select hex(char(null,65,66));
NULL
```

### Expected Behavior

MatrixOne should match MySQL-compatible `CHAR()` behavior:

- non-integer numeric/string arguments should be coerced the same way MySQL does;
- `NULL` arguments should be skipped instead of making the whole result `NULL`.

### Notes

Found during active MySQL compatibility exploration, Module 4: String / Charset / Collation. I searched existing issues for `CHAR()`, `CHAR NULL`, and `CHAR 77.3` and did not find a duplicate.

Ariznawlll:
Verified on latest main 2a008285e.

Result: fixed.

Validation:
```sql
select hex(char(77,77.3,'77.3')) as char_noninteger;
select hex(char(null,65,66)) as char_skip_null;
```

MO now matches MySQL:
- `char(77,77.3,'77.3')` -> `4D4D4D`
- `char(null,65,66)` -> `4142`

Regression added in matrixorigin/motr#39, commit 75bf03f:
- `suites/14_issue_regression/t/issue_25294_char_coercion_null.test`
- `suites/14_issue_regression/r/issue_25294_char_coercion_null.result`

MOTR validation passed:
```
[ISSUE_25294_CHAR_COERCION_NULL] success
```

PR #25394, opened 2026-07-02
fix: CHAR() MySQL compatibility - argument coercion and NULL skipping

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25294

Fixes CHAR() to match MySQL argument coercion, NULL skipping, string-to-integer parsing, multi-byte expansion, and binary literal handling.

### Changes

1. **Remove `Function_STRICT` flag** from CHAR registration — allows the function to handle NULL arguments internally instead of returning NULL immediately
2. **Type-specific argument coercion in `builtInCharCheck`**:
   - Integer types: kept as-is
   - Varchar: kept as-is (parsed in `builtInChar` at runtime)
   - Other string types (char/text/blob/binary/varbinary): cast to varchar
   - Float/decimal/bool/bit: cast to int64 (rounded by the framework, matching MySQL)
3. **Custom `parseLeadingInteger`** replaces `ParseFloat` for varchar arguments:
   - Parses the longest leading integer prefix (e.g. `CHAR('65xyz')` -> 0x41)
   - Direct integer parsing avoids float64 precision loss for large values (e.g. `CHAR('9007199254740993')` -> 0x01)
4. **Skip NULL arguments** instead of returning NULL — matches MySQL behavior
5. **Return empty (non-NULL) string** when all arguments are NULL — `CHAR(NULL, NULL)` -> `''`
6. **Multi-byte big-endian expansion via `encodeCharBytes`** — values are treated as unsigned 32-bit ints; values > 255 expand into multiple bytes (`CHAR(256)` -> 0x0100, `CHAR(65536)` -> 0x010000), negative values use two's complement (`CHAR(-1)` -> 0xFFFFFFFF)
7. **Preserve binary hex/bit literal semantics** — varchar params with `IsBin` set (e.g. `0x41`, `b'01000001'`) are interpreted as big-endian integers, mirroring the old varchar->int64 cast path; >8-byte binary values yield 0 per MySQL's truncated-BINARY behavior

### MySQL Comparison

| SQL | MySQL | MO (before) | MO (after) |
|-----|-------|-------------|------------|
| `hex(char(77,77.3,'77.3'))` | `4D4D4D` | ERROR | `4D4D4D` |
| `hex(char(null,65,66))` | `4142` | NULL | `4142` |
| `hex(char(null,null))` | `''` (empty, not NULL) | NULL | `''` (empty, not NULL) |
| `hex(char('65xyz'))` | `41` | `00` | `41` |
| `hex(char('9007199254740993'))` | `01` | `00` | `01` |
| `hex(char(256))` | `0100` | `00` | `0100` |
| `hex(char(65536))` | `010000` | `00` | `010000` |
| `hex(char(-1))` | `FFFFFFFF` | `''` | `FFFFFFFF` |
| `hex(char(0x41))` | `41` | `41` | `41` |
| `hex(char(0x0100))` | `0100` | `0100` | `0100` |
| `hex(char(b'01000001'))` | `41` | `41` | `41` |

All values verified against MySQL 8.4.

### Test coverage

- UT: NULL handling, varchar truncation, leading numeric prefix, large integer precision, multi-byte expansion, binary/bit literals, all integer-type getters
- BVT: `func_string_char.test` covering the cases above plus nullable-column NULL path
- All function package tests and BVT pass


aunjgr (approved):
CHAR() non-strict coercion matches MySQL. parseLeadingInteger for strings, rounded for floats, NULL skipping. Good test coverage.

XuPeng-SH (changes requested):
Requesting changes.

The NULL-skipping and string-leading-integer direction is good, but this PR still misses a first-principles `CHAR()` invariant and now codifies the wrong oracle in the new large-integer test.

MySQL documents that `CHAR()` interprets each argument as an integer, skips `NULL`, returns a binary string by default, and, critically, arguments larger than 255 are converted into multiple result bytes. The documented examples are `CHAR(256) == CHAR(1,0)` and `CHAR(256*256) == CHAR(1,0,0)`: https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_char

Current code still does low-byte truncation:

- `pkg/sql/plan/function/func_builtin.go:1171` appends only `byte(v&0xFF)` for `v > 255`
- therefore `HEX(CHAR(256))` would be `00` instead of MySQL's `0100`
- `HEX(CHAR(256*256))` would be `00` instead of `010000`
- the new test at `pkg/sql/plan/function/func_string_complex_test.go:599` expects `CHAR('9007199254740993') -> 0x01`, which follows the current low-byte behavior, not MySQL's multi-byte expansion semantics

This is an existing gap in the old implementation, but this PR is specifically changing `CHAR()` compatibility and adding a large-integer precision test, so merging it as-is would make the wrong behavior look intentional.

Suggested fix:

1. Replace the `v > 255` branch with a helper that appends the minimal big-endian byte sequence for non-negative integers, matching the documented examples (`256 -> 01 00`, `65536 -> 01 00 00`).
2. Re-check the negative/overflow string cases against a real MySQL oracle before locking test expectations; overflow clamping plus low-byte truncation is especially risky here.
3. Add regression coverage for at least:
   - `HEX(CHAR(256)) = 0100`
   - `HEX(CHAR(256*256)) = 010000`
   - `HEX(CHAR('256')) = 0100`
   - a confirmed MySQL-oracle expectation for `CHAR('9007199254740993')`
   - `CHAR(NULL)` as the single-NULL form, not only `CHAR(NULL, NULL)`

Local checks I ran on this head: `gofmt -l` clean, `make err-check` passed, and focused `go test ./pkg/sql/plan/function -run 'Test_BuiltInChar|Test_ParseLeadingInteger|Test_BuiltInCharCheck'` passed. The issue is semantic compatibility, not local test stability.

ck89119:
Addressed review feedback: replaced low-byte truncation with big-endian multi-byte expansion matching MySQL semantics.

Changes:
- Added `encodeCharBytes()` — converts int64 to unsigned 32-bit, encodes as big-endian, strips leading zero bytes
- Updated `builtInChar()` to use the new helper for all arguments
- `v > 255` now expands to multiple bytes (`CHAR(256) → 0x0100`, `CHAR(65536) → 0x010000`)
- `v < 0` now uses two's complement uint32 encoding (`CHAR(-1) → 0xFFFFFFFF`) instead of treating as 0
- Added UT: `Test_EncodeCharBytes` (8 cases), multi-byte expansion, negative values, single NULL
- Added BVT: `HEX(CHAR(256)) = 0100`, `HEX(CHAR(65536)) = 010000`, `HEX(CHAR(-1)) = FFFFFFFF`, etc.

All values verified against MySQL 8.4.

XuPeng-SH (changes requested):
I think this still regresses binary hex/bit literals passed to `CHAR()`.

`bindNumVal()` keeps `0x...` / `b'...'` as `Literal_Sval{IsBin:true}` raw bytes. Before this patch, `CHAR()` cast those non-integer args to `int64` first, so the binary-string cast path interpreted `0x41` as `65` and `CHAR(0x41)` returned `A`. After this change, `builtInCharCheck()` routes MySQL-string args to `varchar` and `builtInChar()` unconditionally runs `parseLeadingInteger()` on the raw bytes. For binary literals the raw value is already the decoded bytes, not textual `"0x41"` / `"01000001"`: `0x41` reaches `CHAR()` as `"A"`, so `parseLeadingInteger("A")` falls back to `0` and the result becomes `0x00`. The same problem applies to bit literals such as `b'01000001'`.

So this compatibility fix regresses a previously working path. I think we need to preserve the old numeric path for `IsBin` hex/bit literals (or bind those literals numerically for `CHAR()`) instead of routing them through the varchar prefix parser. Please also add coverage like `HEX(CHAR(0x41))`, `HEX(CHAR(0x0100))`, and `HEX(CHAR(b'01000001'))` so this doesn't slip again.

Separately, a nullable-column NULL-path BVT would still be worth adding, but the binary-literal regression is the blocking issue.

XuPeng-SH (approved):
LGTM.

Re-checked on latest head against the previous blockers: the binary-literal regression path is fixed now. `builtInChar()` explicitly handles `IsBin` varchar inputs as big-endian integers (so `CHAR(0x41)` / `CHAR(b'01000001')` no longer fall into decimal-prefix parsing), and the multi-byte expansion path (`CHAR(256)` -> `0x0100`, `CHAR(65536)` -> `0x010000`) plus NULL-skipping behavior matches the intended MySQL compatibility direction.

Only non-blocking follow-up: BVT could still add a few SQL-level edge cases like `CHAR(x'')`, very long binary literal (`>8 bytes`), and boundary wrap checks around `4294967295/4294967296` for stronger end-to-end oracle coverage.

heni02 (approved):

Issue #25226, opened 2026-06-29
[Bug]: CONVERT(number, CHAR) uses length 1 and rejects multi-digit values


Ariznawlll:
### MatrixOne version

Verified on latest main service used for MySQL compatibility exploration (`8.0.30-MatrixOne-verify-greatest-main`, branch based on `origin/main`).

### Reproduce

```sql
select convert(12345, char) as r;
```

### Expected behavior

MySQL 8.0.45 returns the full character representation:

```text
r
12345
```

### Actual behavior

MatrixOne returns:

```text
ERROR 20101 (HY000): internal error: Can't cast '12345' from BIGINT type to CHAR type. 12345 is larger than Dest length 1
```

### Notes

`CAST(12345 AS CHAR)` works in the passing CAST/CONVERT matrix, but `CONVERT(12345, CHAR)` appears to infer `CHAR(1)` instead of MySQL-compatible variable-length character output.

daviszhen:
fixed

Ariznawlll:
Verified fixed on latest main `49842fb04` (`fix convert (#25398)`).

Validation:
- MO and MySQL both return full strings for `convert(12345, char)`, `convert(-98765, char)`, and `convert(12.3400, char)`.
- motr regression added in matrixorigin/motr#39 at commit `2fc68b6`: `suites/14_issue_regression/t/issue_25226_convert_number_char.test` (`# @target-branch: main`).
- Local motr check passed: `total=2 success=2 failed=0`, including this case and #25254.

Closing as fixed/verified.

Issue #25254, opened 2026-06-30
[Bug]: CONVERT string to BINARY uses length 1 instead of preserving full value


Ariznawlll:
### MatrixOne Version

Latest `origin/main` verified locally at `33e96877604ad4b847ea7d5c55eb09429de1a5b6`.

### Reproduce Steps

```sql
select hex(convert('AZ', binary)), length(convert('AZ', binary));
```

### Expected Behavior

MySQL preserves the whole value for the default `BINARY` conversion:

```text
415A    2
```

### Actual Behavior

MatrixOne infers destination length 1 and errors:

```text
ERROR 20101 (HY000): internal error: Can't cast 'AZ' from VARCHAR type to BINARY type. Src length 2 is larger than Dest length 1
```

### Notes

Found while completing Batch A of the MySQL compatibility `CAST` / `CONVERT` matrix. Explicit `CONVERT('AZ', BINARY(4))` matches MySQL and was added to BVT; this default-length row was excluded.

daviszhen:
fixed

Ariznawlll:
Verified fixed on latest main `49842fb04` (`fix convert (#25398)`).

Validation:
- MO and MySQL both return `415A` and length `2` for `hex(convert('AZ', binary))`, `length(convert('AZ', binary))`.
- Also checked empty string and `hello`; MO preserves binary length like MySQL.
- motr regression added in matrixorigin/motr#39 at commit `2fc68b6`: `suites/14_issue_regression/t/issue_25254_convert_string_binary.test` (`# @target-branch: main`).
- Local motr check passed: `total=2 success=2 failed=0`, including this case and #25226.

Closing as fixed/verified.

PR #25398, opened 2026-07-02
fix convert

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25226

#25254

修改 
```
select convert(12345, char) as r;
select hex(convert('AZ', binary)), length(convert('AZ', binary));
```


XuPeng-SH (changes requested):
Reviewed latest head `0231fe0c34e8728d8ba0182b3f4d41b933a198bf` from multiple angles: parser grammar, generated parser consistency, AST -> plan type resolution, cast execution conventions, regression tests, and unrelated diff scope.

The CONVERT fix itself looks directionally correct:
- `CONVERT(..., CHAR)` now resolves as unbounded `VARCHAR`, avoiding the old implicit `CHAR(1)` truncation/error path.
- `CONVERT(..., CHAR(n))` still resolves to fixed `CHAR(n)`.
- `CONVERT(..., BINARY)` now carries the existing `T_binary + Scale=-1 + Width=-1` explicit-cast marker, so execution preserves the full input; `BINARY(n)` still truncates/pads through `explicitCastToBinary`.

I still need to request changes because this PR currently includes unrelated frontend/vet changes from another PR:

- `git log origin/main..HEAD` contains `b013609bb fix: address frontend vet failures (#25403)` plus frontend-only update commits.
- The resulting diff changes `pkg/frontend/mysql_cmd_executor.go:2706`, replacing the structured `ses.Error(... zap.String("path", param.Filepath), zap.Uint64("epoch", epoch), zap.Error(err))` log with `ses.Errorf(... "epoch=%d, error=%v", ...)`.

That is outside the scope of a `CONVERT(..., CHAR/BINARY)` parser/type-resolution bugfix and it also drops the `path` field from the load-local-file failure log. If #25398 merges first, it effectively sneaks in part of #25403 without keeping that review/CI boundary clean.

Please rebase/cherry-pick this PR so it contains only the CONVERT-related files:
- `pkg/sql/parsers/dialect/mysql/mysql_sql.y`
- regenerated `pkg/sql/parsers/dialect/mysql/mysql_sql.go`
- `pkg/sql/plan/build_util_test.go`
- `test/distributed/cases/function/func_cast.test`
- `test/distributed/cases/function/func_cast.result`

and drop/restore the frontend logging change here. If the vet fix is required, keep it in #25403 or merge/rebase after #25403 lands.

Local verification I ran:
- `go test ./pkg/sql/plan -run TestConvertCharBinaryTypeResolution -count=1 -timeout 120s` passed with MO CGo/rpath flags.
- `go test ./pkg/sql/parsers/... -count=1 -timeout 120s` passed with MO CGo/rpath flags.

XuPeng-SH (approved):
LGTM.

This is a clean, narrowly scoped fix: it makes `CONVERT(..., CHAR/BINARY)` go through the same type-normalization path that `CAST(... AS CHAR/BINARY)` already uses, and the added unit/BVT coverage hits the exact regressions being fixed.

I don’t see a substantive blocker here. The only follow-up I’d consider is a little more boundary/NULL coverage, but that’s nice-to-have rather than merge-blocking.

heni02 (approved):

iamlinjunhong (approved):

ouyuanning (approved):

Issue #25280, opened 2026-06-30
[Bug]: GetShardInfo returns stale log replica addresses from gossip, causing connection failures


matrix-meow:
## Is there an existing issue for the same bug?
- [x] I have checked the existing issues.

## Branch Name
main

## Commit ID
2add0b9fc

## Other Environment Information
- OS type: Linux
- Others: freetier-01 cluster (cn-dev)

## Actual Behavior

HAKeeper/gossip returns log replica addresses for pods that no longer exist. `GetShardInfo` reports replicas `default-log-1` and `default-log-2`, but only `default-log-0` exists in the namespace. This causes DN logservice clients to attempt connecting to non-existent pods, wasting 30-61 seconds per attempt.

### Evidence

**Only one log pod exists:**
```
Loki query: {namespace="freetier-01", pod=~"default-log.*"}
Result: only default-log-0
```

**But DN tries to connect to non-existent pods:**
```
07:02:31  "WAL-Replay failed to create log service client":
  dial tcp4: lookup default-log-2.default-log-headless.freetier-01.svc: no such host
```

**Log-0 Raft logs show stale replica IDs:**
```
06:49:18 [default-log-0] ignored a Heartbeat message sent to [00001:00307]
  but received by [00001:68107]
```

The Raft group was recreated (replica ID changed from 00307 to 68107), but the old membership information persists in gossip.

### Root cause

`GetShardInfo` (`pkg/logservice/shardinfo.go:49-74`) gets replica addresses from dragonboat's **gossip registry**, not from HAKeeper's `LogState`. Gossip has its own convergence timeline independent of HAKeeper's 5-minute `LogStoreTimeout`. When log pods are removed or the Raft group is reconfigured, gossip may continue reporting stale addresses.

```go
// shardinfo.go:172-206
func (s *Service) getShardInfo(shardID uint64) (...) {
    r, _ := s.store.nh.GetNodeHostRegistry()    // gossip registry
    shard, _ := r.GetShardInfo(shardID)          // may include stale replicas
    for nodeID, uuid := range shard.Nodes {
        if data, ok := r.GetMeta(uuid); ok {
            replica.ServiceAddress = md.serviceAddress  // from gossip
        }
    }
}
```

Additionally, HAKeeper takes **5 minutes** (`LogStoreTimeout`) to mark a log store as dead, and even then the store entry is never deleted from `LogState` — it persists with `State: TimeoutState` indefinitely.

## Expected Behavior

[...]

PR #25417, opened 2026-07-03
Filter expired log replica addresses from GetShardInfo

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25280

 This PR updates GetShardInfo so discovery callers do not receive service addresses for expired log stores. The RPC now carries a flag that lets callers choose whether expired replica addresses should be included. Normal
  logservice and HAKeeper client discovery uses the default filtered behavior, while zombie membership checks explicitly request the full membership view.

  The server still preserves replica membership entries, but clears ServiceAddress for stores that HAKeeper marks expired based on LogState.Stores heartbeat ticks. This prevents clients from dialing stale pods without
  breaking zombie self-check logic.


XuPeng-SH (changes requested):
Requesting changes for one correctness blocker in the discovery path.

The PR filters expired replica addresses out of `GetShardInfo`, which means `connectToLogServiceByReverseProxy` can now pass a shorter `targets` slice into `connectToLogService`. But `connectToLogService` shuffles that shorter slice using `len(cfg.ServiceAddresses)` instead of `len(addresses)`:

```go
addresses := append([]string{}, targets...)
rand.Shuffle(len(cfg.ServiceAddresses), func(i, j int) {
    addresses[i], addresses[j] = addresses[j], addresses[i]
})
```

That is unsafe once stale addresses are filtered. I verified with a temporary local test that this panics when discovery returns one live target while the config still contains three service addresses:

```go
_, _ = connectToLogService(context.Background(), "", []string{"127.0.0.1:1"}, ClientConfig{
    ServiceAddresses: []string{"127.0.0.1:1", "127.0.0.1:2", "127.0.0.1:3"},
})
```

Result:

```text
panic: runtime error: index out of range [2] with length 1
```

This is a realistic unhappy path for the issue being fixed: before this PR, the target list could still include the expired pods, so its length matched the configured addresses and the client wasted time dialing stale pods; after this PR, the target list can shrink to only live replicas, turning that same stale-membership case into a client panic.

Suggested fix:
- Change the shuffle bound in `pkg/logservice/client.go` to `len(addresses)`.
- Add a regression test where discovery/filtering leaves fewer targets than `cfg.ServiceAddresses`, and assert `connectToLogService` does not panic.
- Keep the existing HAKeeper path as-is; `connectToHAKeeper` already shuffles by `len(addresses)`.

The server-side filtering shape looks directionally right: normal `GetShardInfo` requests use the filtered behavior, and `getShardMembership` explicitly sets `IncludeExpiredReplicaAddresses=true` for zombie self-checks. The missing client-side bound fix blocks this PR from safely resolving #25280.

XuPeng-SH (approved):
Re-reviewed latest head `950c5dccf`.

The previous blocker is fixed: `connectToLogService` now shuffles by the filtered target slice length (`len(addresses)`), and the regression test covers the case where discovery returns fewer live targets than `cfg.ServiceAddresses`.

I re-checked the PR from the discovery contract outward:
- normal `GetShardInfo` requests default to filtering expired replica addresses;
- callers that dial replicas strip empty addresses and treat an expired/empty leader as not-ready instead of dialing a stale endpoint;
- `getShardMembership` explicitly requests `IncludeExpiredReplicaAddresses=true`, so the zombie self-check still sees full membership and does not misclassify replicas just because their service address is filtered;
- the proto change appends a bool field with the safe default (`false`) for existing clients;
- the new HAKeeper-state lookup is bounded, and fail-open when checker state is unavailable preserves old availability behavior rather than hiding live stores without authoritative state.

Unhappy-path cases I checked: stale follower, stale leader, empty/missing metadata, shortened target list, zombie self-check membership, and reverse-proxy connection paths. I did not find a remaining blocker.

Local verification:
- `go test -count=1 -timeout 240s ./pkg/logservice -run 'TestConnectToLogServiceShufflesTargetsLength|TestGetShardInfo|TestGetShardMembership_IncludesExpiredReplicaAddresses|TestIsExpiredLogStoreInState|TestShardInfoCanBeQueried'`
- `go test -count=1 -timeout 360s ./pkg/logservice -run 'TestClientCanBeConnectedByReverseProxy|TestHAKeeperClientConnectByReverseProxy|TestShardInfoCanBeQueried|TestGossipInSimulatedCluster|TestCheckZombieReplicas_|TestStartReplicas_SkipsZombie|TestZombieRepro_WithFix'`
- `go test -count=1 -timeout 900s ./pkg/logservice`
- `go test -count=1 -timeout 240s ./pkg/logservice -run '^$'`
- `go vet ./pkg/logservice`

Approving. CI is still running at review time, so merge should still wait for required checks.

LeftHandCold (commented):
I think there is still a discovery-path regression risk here.

GetShardInfo callers still use a 1s deadline (queryShardInfoRaw, pkg/logservice/shardinfo.go:117), but the server-side filtering path now synchronously calls s.store.getCheckerState() (pkg/logservice/shardinfo.go:237). That helper uses context.Background() and hakeeperDefaultTimeout = 2s (pkg/logservice/store_hakeeper_check.go:40, pkg/logservice/store_hakeeper_check.go:356-359).

If the HAKeeper read is slow or temporarily unavailable, the intended fail-open path can return only after up to 2s, while the caller has already returned context deadline exceeded after 1s. That changes the old discovery behavior from effectively returning gossip shard info even when HAKeeper state cannot be read to normal GetShardInfo discovery timing out because HAKeeper state is slow. This affects both logservice reverse-proxy discovery and HAKeeper reverse-proxy discovery because both go through GetShardInfo.

Can we make the checker-state lookup non-blocking or bounded below the GetShardInfo caller deadline, use cached checker state, or otherwise fail open quickly enough to return unfiltered gossip before the 1s caller timeout? I do not think this should land as-is because the stale-address filter is on a hot discovery path and should not make discovery availability depend on a 2s HAKeeper read.

LeftHandCold (approved):
Re-reviewed latest head `df67d8e`.

The previous discovery-path concern is addressed: normal `GetShardInfo` still uses the 1s caller deadline, while the HAKeeper checker-state lookup is now wrapped with a 100ms child timeout and fails open to unfiltered gossip state if it cannot read the checker state quickly. That keeps discovery from depending on the old 2s HAKeeper read path.

I also re-checked the other contract edges:

- normal discovery uses filtered addresses by default;
- zombie membership explicitly sets `IncludeExpiredReplicaAddresses=true`, so it still sees full membership;
- `GetShardInfo` strips empty/expired addresses from dial targets and treats an empty leader address as not-ready;
- `connectToLogService` now shuffles by `len(addresses)`, so filtered target lists shorter than `cfg.ServiceAddresses` do not panic;
- the proto change appends a bool field with the safe default false.

LGTM.

Local targeted `go test ./pkg/logservice ...` did not get to test execution in my environment because required cgo headers are missing (`usearch.h`, `roaring.h`, `xxhash.h`).

PR #25422, opened 2026-07-03
set default incremental checkpoint interval to 5m

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24900

set default incremental checkpoint interval to 5m


LeftHandCold (approved):
LGTM.

I re-checked the `arenaDrainDelay` change specifically. `ScheduleArenaDrain()` is called after checkpoint completion and only resets the arena-pool drain timer; it does not participate in ICKP scheduling. With the default ICKP interval moved to 5 minutes, using a 10-minute drain delay is consistent with keeping the pools warm across steady-state checkpoint cycles.

For the production/default path, `tnservice` and TAE checkpoint defaults now both use `options.DefaultCheckpointIncrementalInterval`, which is `5m`. That matches the intended behavior for customers/environments that do not explicitly configure `incremental-interval`.

XuPeng-SH (approved):

PR #25424, opened 2026-07-03
fix: pad hex literals inserted into BINARY columns

### LINKED ISSUE #25217: [Bug]: BINARY(n) does not pad hex literal inserts like MySQL
### MatrixOne version

Verified on latest `origin/main` at `24d9d1ede`.

### MySQL version

MySQL `8.0.45`.

### Reproduction

```sql
drop database if exists mysql_compat_binary_hex_probe;
create database mysql_compat_binary_hex_probe;
use mysql_compat_binary_hex_probe;

create table t_binary_hex (id int primary key, b binary(4));
insert into t_binary_hex values (1, 0x4142);
select id, hex(b) as b_hex, length(b) as b_len from t_binary_hex order by id;

drop database mysql_compat_binary_hex_probe;
```

### MySQL behavior

```text
id    b_hex       b_len
1     41420000    4
```

### MatrixOne behavior

```text
id    b_hex    b_len
1     4142     2
```

### Expected behavior

For `BINARY(4)`, inserting `0x4142` should store a fixed-length binary value padded with `0x00` bytes, matching MySQL behavior (`41420000`, length `4`).

### Notes

String literal inserts into `BINARY(4)` were verified to match MySQL. The incompatibility is specifically observed with hex literal input.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25217

The `INSERT ... VALUES` fast path decodes hex literals directly in
`SetInsertValueString`. Unlike ordinary string literals, the decoded bytes
skipped the required binary normalization.

This change normalizes decoded numeric hex literals before building constants:

- short values inserted into `BINARY(n)` are padded with zero bytes;
- decoded values wider than `BINARY(n)` or `VARBINARY(n)` are rejected using
  byte length rather than Unicode rune count;
- odd-digit numeric hex literals are left-padded before decoding, matching the
  general binder behavior.

The byte-width validation is limited to decoded hex literals so existing
ordinary string and numeric `BINARY` insert behavior remains unchanged.

## Testing

- `go test ./pkg/sql/util ./pkg/sql/plan -count=1`
- `make`
- `make static-check` (`golangci-lint`: 0 issues)
- full `binary.test` with `mo-tester`: 547/547 statements passed
- changed production coverage blocks: 100%


XuPeng-SH (approved):
LGTM.

I re-checked this from the fast-path constant-folding entry (`MakeInsertValueConstExpr` -> `SetInsertValueString`) and the unhappy-path branches, and the core compatibility bug is fixed:
- hex literals now pass through BINARY normalization (fixed-width zero-padding),
- decoded-byte width is enforced for `BINARY/VARBINARY` (byte length, not rune count),
- odd-digit hex literals are normalized before decode (`0x1` -> `0x01`).

The previous mismatch in #25217 (`BINARY(4)` + `0x4142` storing only 2 bytes) is addressed, and over-width paths now fail explicitly instead of silently storing inconsistent values.

Non-blocking follow-up only: add a sibling SQL case for bit-literal insertion (`b'...'`) into `BINARY/VARBINARY` to lock parity expectations with the numeric-hex path.

heni02 (approved):

aunjgr (approved):

Issue #25346, opened 2026-07-02
[Bug]: WHERE column IN (value, NULL) filters out matching rows


Ariznawlll:
### Is there an existing issue for the same bug?

I have checked the existing issues.

Existing issue #25221 covers the different case where the left operand itself is `NULL`, for example `NULL IN (1, NULL)`, which errors in MatrixOne. This issue covers a non-NULL column value compared against an `IN` list that contains both a matching value and `NULL`.

### Branch Name

main

### Commit ID

`0e8602b78`

### Other Environment Information

- Local MatrixOne built from latest `origin/main`
- MySQL oracle: MySQL 8.0.45

### Actual Behavior

`k IN ('keep', NULL)` evaluates correctly in a projection for rows where `k = 'keep'`, but the same expression in a `WHERE` predicate filters out all rows.

Observed on MatrixOne latest `main`:

```text
SELECT COUNT(*) FROM t WHERE k IN ('keep', NULL); -- 0
SELECT COUNT(*) FROM t WHERE k IN ('keep');       -- 2

id  k     expr
1   keep  1
2   key   NULL
3   NULL  NULL
4   keep  1
```

So the scalar expression result says rows 1 and 4 match, but `WHERE` returns zero rows.

### Expected Behavior

MySQL returns the two matching rows. The presence of `NULL` in the `IN` list should not suppress rows that match a non-NULL list element.

Expected result:

```text
SELECT COUNT(*) FROM t WHERE k IN ('keep', NULL); -- 2
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS mysql_compat_in_null_min;
CREATE DATABASE mysql_compat_in_null_min;
USE mysql_compat_in_null_min;

CREATE TABLE t (id int primary key, k varchar(16) null);
INSERT INTO t VALUES (1,'keep'),(2,'key'),(3,NULL),(4,'keep');

SELECT 'in_null' q, COUNT(*) cnt
FROM t
WHERE k IN ('keep', NULL);

SELECT 'in_no_null' q, COUNT(*) cnt
FROM t
WHERE k IN ('keep');

SELECT id, k, k IN ('keep', NULL) AS expr
FROM t
ORDER BY id;
```

### Additional information

MySQL 8.0.45 result for the same SQL:

```text
in_null    2
in_no_null 2

id  k     expr
1   keep  1
2   key   NULL
3   NULL  NULL
4   keep  1
```

[...]

PR #25426, opened 2026-07-03
Fix zonemap pruning for nullable IN lists

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Refs #25346

This PR backports the zonemap pruning fix for nullable `IN` lists to `4.1-dev`.

For `k IN ('keep', NULL)`, the binder expands the predicate into an OR over equality checks. The `k = NULL` branch produces an unknown/uninitialized zonemap, which previously caused the whole OR result to be treated as not selected by `EvaluateFilterByZoneMap`. That pruned the block before the normal row-level filter could return the matching `k = 'keep'` rows.

With this change, unknown zonemap results keep the block instead of pruning it. Row-level expression evaluation still applies the original predicate, so correctness is restored without changing SQL scalar semantics.

To limit pruning-rate regressions, boolean zonemap folding now uses pruning-specific may-true semantics: `AND` only prunes when a known branch cannot be true, and `OR` keeps the block when any known branch may be true. Only predicates that cannot be proven either way are treated as unknown and scanned.

Native nullable `IN` / `NOT IN` lists are also handled directly: `IN` ignores NULL list members for zonemap intersection, and `NOT IN` with any NULL list member is treated as impossible to be TRUE under a WHERE predicate.

The real SQL-shaped bare NULL expansion for nullable `NOT IN` is covered too: `!= NULL` and `<> NULL` are treated as impossible to be TRUE for zonemap pruning, and non-NULL `!=` keeps may-true semantics unless both sides are provably the same single value.

The regression test covers the nullable IN-list case from the issue, verifies that a normal false equality predicate still prunes by zonemap, verifies that `AND` keeps known-false pruning even when another branch is unknown, verifies that `AND` does not prune a `[false,true]` boolean zonemap that may still be true, covers nullable `IN` without a matching non-NULL member plus nullable `NOT IN`, and covers `NOT IN` after binder-style expansion to `!= ... AND != NULL`.

## Test result

- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes|NotInExpandedWithBareNullPrunes|NotEqualBareNullPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

### DIFF
--- a/pkg/sql/colexec/evalExpression.go
+++ b/pkg/sql/colexec/evalExpression.go
@@ -31,6 +31,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/pb/plan"
 	"github.com/matrixorigin/matrixone/pkg/sql/plan/function"
 	"github.com/matrixorigin/matrixone/pkg/sql/util"
+	"github.com/matrixorigin/matrixone/pkg/vm/engine/tae/compute"
 	"github.com/matrixorigin/matrixone/pkg/vm/engine/tae/index"
 	"github.com/matrixorigin/matrixone/pkg/vm/process"
 )
@@ -1130,7 +1131,8 @@ func EvaluateFilterByZoneMap(
 
 	zm := GetExprZoneMap(ctx, proc, expr, meta, columnMap, zms, vecs)
 	if !zm.IsInited() || zm.GetType() != types.T_bool {
-		selected = false
+		// Unknown zonemap results are not proof that the block cannot match.
+		selected = true
 	} else {
 		selected = types.DecodeBool(zm.GetMaxBuf())
 	}
@@ -1197,6 +1199,9 @@ func GetExprZoneMap(
 					return zms[expr.AuxId]
 				}
 			case "in":
+				if list := args[1].GetList(); list != nil {
+					return foldNativeInListZoneMap(ctx, proc, meta, columnMap, zms, vecs, args, expr.AuxId, false)
+				}
 				rid := args[1].AuxId
 				if vecs[rid] == nil {
 					if data, ok := args[1].Expr.(*plan.Expr_Vec); ok {
@@ -1222,6 +1227,13 @@ func GetExprZoneMap(
 				}
 
 				zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], lhs.AnyIn(vecs[rid]))
+				return zms[expr.AuxId]
+
+			case "not_in":
+				if list := args[1].GetList(); list != nil {
+					return foldNativeInListZoneMap(ctx, proc, meta, columnMap, zms, vecs, args, expr.AuxId, true)
+				}
+				zms[expr.AuxId].Reset()
 				return zms[expr.AuxId]
 
 			case "prefix_eq":
@@ -1305,6 +1317,10 @@ func GetExprZoneMap(
 			var res, ok bool
 			switch t.F.Func.ObjName {
 			case ">":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1315,6 +1331,10 @@ func GetExprZoneMap(
 				}
 
 			case "<":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1325,6 +1345,10 @@ func GetExprZoneMap(
 				}
 
 			case ">=":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1335,6 +1359,10 @@ func GetExprZoneMap(
 				}
 
 			case "<=":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1345,6 +1373,10 @@ func GetExprZoneMap(
 				}
 
 			case "=":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1354,7 +1386,25 @@ func GetExprZoneMap(
 					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], res)
 				}
 
+			case "!=", "<>":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
+				if f() {
+					return zms[expr.AuxId]
+				}
+				if res, ok = anyNotEqualZoneMap(zms[args[0].AuxId], zms[args[1].AuxId]); !ok {
+					zms[expr.AuxId].Reset()
+				} else {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], res)
+				}
+
 			case "between":
+				if hasConstNullArg(args) {
+					zms[expr.AuxId] = index.SetBool(zms[expr.AuxId], false)
+					return zms[expr.AuxId]
+				}
 				if f() {
 					return zms[expr.AuxId]
 				}
@@ -1365,34 +1415,14 @@ func GetExprZoneMap(
 				}
 
 			case "and":
-				if f() {
+				if hasResult := foldAndZoneMap(ctx, proc, meta, columnMap, zms, vecs, args, expr.AuxId); hasResult {
 					return zms[expr.AuxId]
 				}
-				zmRes := zms[args[0].AuxId]
-				for i := 1; i < len(args); i++ {
-					if res, ok = zmRes.And(zms[args[i].AuxId]); !ok {
-						zmRes.Reset()
-						break
-					} else {
-						zmRes = index.SetBool(zmRes, res)
-					}
-				}
-				zms[expr.AuxId] = zmRes
 
 			case "or":
-				if f() {
+				if hasResult := foldOrZoneMap(ctx, proc, meta, columnMap, zms, vecs, args, expr.AuxId); hasResult {
 					return zms[expr.AuxId]
 				}
-				zmRes := zms[args[0].AuxId]
-				for i := 1; i < len(args); i++ {
-					if res, ok = zmRes.Or(zms[args[i].AuxId]); !ok {
-						zmRes.Reset()
-						break
-					} else {
-						zmRes = index.SetBool(zmRes, res)
-					}
-				}
-				zms[expr.AuxId] = zmRes
 
 			case "+":
 				if f() {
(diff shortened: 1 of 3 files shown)


LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The backport is a clean cherry-pick of the main-branch fix.
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

XuPeng-SH (changes requested):
I still see one substantive issue here.

The new top-level fallback `unknown => selected` in zonemap evaluation is too coarse for nullable `IN` / `NOT IN` expressions.

`base_binder` expands nullable `IN` / `NOT IN` into boolean trees containing explicit `col = NULL` / `col != NULL` branches. Those NULL comparisons produce uninitialized zonemap results. In `GetExprZoneMap`, the current `and` / `or` logic still bails out as soon as any child is uninitialized, and this PR then turns the whole expression into `selected = true` at the top level.

That means expressions like:
- `k IN ('zzz', NULL)`
- `k NOT IN ('zzz', NULL)`

now become effectively “always selected” for pruning, even when the non-NULL branches already prove the block cannot produce a matching row.

For example, if the block zonemap excludes `'zzz'`, `k IN ('zzz', NULL)` should still be safely prunable, because rows with any other value (including NULL) cannot make the predicate TRUE. The current change loses that pruning opportunity.

Suggested fix:
1. handle NULL list members inside zonemap evaluation for `IN` / `NOT IN` instead of only at the top-level fallback, or
2. make boolean combination preserve useful FALSE results instead of poisoning the whole expression when one child is unknown.

Please also add regressions for:
- nullable `IN` with no matching non-NULL member
- nullable `NOT IN` lists

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The backport is a clean cherry-pick of the main-branch fix.
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality)' -count=1 -v`
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The backport is a clean cherry-pick of the main-branch fix.
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|NullableInListWithoutMatchPrunes|NullableNotInListPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The backport is a clean cherry-pick of the main-branch fix.
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning with may-true semantics: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- `AND` no longer uses boolean zonemap `min` to decide pruning; a `[false,true]` boolean zonemap is kept because it may still contain true rows.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- `AND` with a possible-true boolean range plus a matching equality is covered and keeps the block.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

XuPeng-SH (changes requested):
Re-reviewing the latest head: the original wrong-results bug is fixed, but I still think this needs changes because the nullable `NOT IN` story is not fully closed on the real SQL path.

What looks fixed now:
- `k IN ('keep', NULL)` no longer over-prunes matching rows.
- The new null-aware boolean folding in `EvaluateFilterByZoneMap` improves the original bad case significantly.

What still remains:
- the new nullable-`NOT IN` handling/tests appear to cover a synthetic/native `not_in(col, [value, typed-null])` shape,
- but real SQL bare `NULL` is planned differently (`T_any` literal) and still goes through binder expansion to `col != 'zzz' AND col != NULL`.

That matters because `!= NULL` still does not get a dedicated null-aware zonemap-false treatment here. So for real SQL like:
- `k NOT IN ('zzz', NULL)`
- `k NOT IN ('a', 'b', NULL)`
- `k NOT IN (NULL)`

the current logic still appears to keep the block conservatively rather than proving it cannot produce TRUE.

So the PR description/comment claim that nullable native `NOT IN` is now handled directly is still stronger than what the latest head actually guarantees on the real bare-NULL planner path.

Suggested fix:
1. handle bare-NULL `NOT IN` on the real SQL path as well (either before binder expansion or by making `!= NULL` zonemap-false under WHERE pruning semantics), and
2. add regressions using real SQL-shaped NULLs, not only typed-string NULL literals.

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The backport is a clean cherry-pick of the main-branch fix.
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning with may-true semantics: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- `AND` no longer uses boolean zonemap `min` to decide pruning; a `[false,true]` boolean zonemap is kept because it may still contain true rows.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- The real SQL-shaped nullable `NOT IN` expansion is handled as well: `!= NULL` and `<> NULL` return zonemap false, while non-NULL `!=` only prunes when both sides are provably the same single value.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- `AND` with a possible-true boolean range plus a matching equality is covered and keeps the block.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- Binder-style `NOT IN` expansion to `!= ... AND != bare NULL` and both `!= NULL` / `<> NULL` spellings are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes|NotInExpandedWithBareNullPrunes|NotEqualBareNullPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go vet ./pkg/sql/colexec`

XuPeng-SH (approved):
Re-reviewing the latest head: the earlier blocker around real SQL bare-NULL `NOT IN` is resolved now.

I re-checked the actual planner shape (`k NOT IN ('zzz', NULL)` expanding to `k != 'zzz' AND k != NULL`) and the latest head now handles that path correctly:
- constant-NULL comparisons like `!= NULL` / `<> NULL` are treated as zonemap-false under pruning semantics,
- boolean folding preserves that useful false result under `AND`,
- and the new regressions cover the bare `T_any` NULL path directly.

At this point I don’t see a remaining substantive correctness issue in the current diff. Extra SQL/BVT coverage would still be nice-to-have, but not blocker-level.

LGTM.

ouyuanning (approved):

PR #25427, opened 2026-07-03
Fix zonemap pruning for nullable IN lists

### LINKED ISSUE #25346: [Bug]: WHERE column IN (value, NULL) filters out matching rows

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25346

This PR makes block-level zonemap filtering conservative when an expression cannot be evaluated to a boolean zonemap result.

For `k IN ('keep', NULL)`, the binder expands the predicate into an OR over equality checks. The `k = NULL` branch produces an unknown/uninitialized zonemap, which previously caused the whole OR result to be treated as not selected by `EvaluateFilterByZoneMap`. That pruned the block before the normal row-level filter could return the matching `k = 'keep'` rows.

With this change, unknown zonemap results keep the block instead of pruning it. Row-level expression evaluation still applies the original predicate, so correctness is restored without changing SQL scalar semantics.

To limit pruning-rate regressions, boolean zonemap folding now uses pruning-specific may-true semantics: `AND` only prunes when a known branch cannot be true, and `OR` keeps the block when any known branch may be true. Only predicates that cannot be proven either way are treated as unknown and scanned.

Native nullable `IN` / `NOT IN` lists are also handled directly: `IN` ignores NULL list members for zonemap intersection, and `NOT IN` with any NULL list member is treated as impossible to be TRUE under a WHERE predicate.

The real SQL-shaped bare NULL expansion for nullable `NOT IN` is covered too: `!= NULL` and `<> NULL` are treated as impossible to be TRUE for zonemap pruning, and non-NULL `!=` keeps may-true semantics unless both sides are provably the same single value.

The regression test covers the nullable IN-list case from the issue, verifies that a normal false equality predicate still prunes by zonemap, verifies that `AND` keeps known-false pruning even when another branch is unknown, verifies that `AND` does not prune a `[false,true]` boolean zonemap that may still be true, covers nullable `IN` without a matching non-NULL member plus nullable `NOT IN`, and covers `NOT IN` after binder-style expansion to `!= ... AND != NULL`.

## Test result

- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes|NotInExpandedWithBareNullPrunes|NotEqualBareNullPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`


LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality)' -count=1 -v`
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|NullableInListWithoutMatchPrunes|NullableNotInListPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning with may-true semantics: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- `AND` no longer uses boolean zonemap `min` to decide pruning; a `[false,true]` boolean zonemap is kept because it may still contain true rows.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- `AND` with a possible-true boolean range plus a matching equality is covered and keeps the block.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`

LeftHandCold (commented):
Self-review completed.

Root cause verified: `k IN ('keep', NULL)` is expanded into `k = 'keep' OR k = NULL`; the NULL branch makes zonemap evaluation unknown/uninitialized, and `EvaluateFilterByZoneMap` previously interpreted that unknown result as "block not selected", pruning matching rows before row-level filtering.

Review result:
- The fix is conservative: unknown/non-boolean zonemap results now keep the block instead of pruning it.
- The performance risk is narrowed by preserving boolean short-circuit pruning with may-true semantics: `AND` can still prune on a known false branch, and `OR` can still keep on a known true branch, even when another branch is unknown.
- `AND` no longer uses boolean zonemap `min` to decide pruning; a `[false,true]` boolean zonemap is kept because it may still contain true rows.
- Native nullable `IN` / `NOT IN` lists are handled directly, so `IN ('zzz', NULL)` can still prune when non-NULL members miss the block, and `NOT IN (..., NULL)` prunes because it cannot be TRUE in WHERE.
- The real SQL-shaped nullable `NOT IN` expansion is handled as well: `!= NULL` and `<> NULL` return zonemap false, while non-NULL `!=` only prunes when both sides are provably the same single value.
- Constant no-column predicates keep the existing exact evaluation path, so constant false filters can still be rejected.
- Normal false equality zonemap pruning is covered by the regression test and still returns false.
- `AND` with a known false branch plus an unknown branch is covered and still prunes.
- `AND` with a possible-true boolean range plus a matching equality is covered and keeps the block.
- Nullable `IN` without a matching non-NULL member and nullable `NOT IN` are covered.
- Binder-style `NOT IN` expansion to `!= ... AND != bare NULL` and both `!= NULL` / `<> NULL` spellings are covered.
- No resource lifetime change: the existing vector cleanup path is unchanged.

Validation:
- `go test ./pkg/sql/colexec -run 'TestEvaluateFilterByZoneMap(NullableInListIsConservative|StillPrunesFalseEquality|AndKeepsKnownFalsePruning|AndKeepsPossibleTrueBoolRange|NullableInListWithoutMatchPrunes|NullableNotInListPrunes|NotInExpandedWithBareNullPrunes|NotEqualBareNullPrunes)' -count=1 -v`
- `go test ./pkg/sql/colexec -count=1`
- `go test ./pkg/sql/colexec -run '^$' -count=1`
- `go vet ./pkg/sql/colexec`
- `go test ./pkg/vm/engine/disttae -run TestEvalFilterByZoneMap -count=1`

XuPeng-SH (approved):

ouyuanning (approved):

Issue #25296, opened 2026-07-01
[Bug]: CONCAT_WS returns NULL instead of empty string when all values are NULL


Ariznawlll:
### Environment

- Branch: latest `main`
- Commit tested locally: `200bfa7e0efb` (`fix: improve skills SKILL.md compatibility format (#25271)`)
- MySQL oracle: local MySQL 8.x via `mysql -u root -p...`
- MO: `8.0.30-MatrixOne-v1.3.0`

### Steps to Reproduce

```sql
select hex(concat_ws(',', null, null)) as result_hex,
       concat_ws(',', null, null) is null as is_null;
```

### MySQL Result

```text
result_hex  is_null
            0
```

MySQL returns an empty string when all values after the separator are `NULL`.

### MatrixOne Result

```text
result_hex  is_null
NULL        1
```

MatrixOne returns `NULL`.

### Expected Behavior

`CONCAT_WS(separator, expr, ...)` should match MySQL:

- if the separator is `NULL`, return `NULL`;
- otherwise skip `NULL` values;
- if all values are skipped, return an empty string rather than `NULL`.

### Notes

Found during active MySQL compatibility exploration, Module 4: String / Charset / Collation. I searched existing issues for `CONCAT_WS NULL` and did not find a duplicate.

daviszhen:
fxied

Ariznawlll:
Verified fixed on latest main `c67b5077a` (`fix concat_ws (#25428)`).

Validation:
- MO and MySQL 8.0.45 both return an empty string for `concat_ws(',', null, null)`: `hex = ""`, `is null = 0`, `length = 0`.
- Also checked boundary behavior:
  - `concat_ws(null, 'a', 'b')` returns `NULL`.
  - `concat_ws(',', 'a', null, 'b')` returns `a,b`.
  - empty strings still participate in separator output: `concat_ws(',', '', null, '')` returns `,` (`hex = 2C`).

Regression added in matrixorigin/motr#39 at commit `e8432b9`:
- `suites/14_issue_regression/t/issue_25296_concat_ws_all_null.test`
- target marker: `# @target-branch: main`

Local motr check passed on latest main service at `127.0.0.1:7603`:
- `14_issue_regression.issue_25296_concat_ws_all_null`
- `total=1 success=1 failed=0`

Closing as fixed/verified.

PR #25428, opened 2026-07-03
fix concat_ws

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25296

修改 `select hex(concat_ws(',', null, null)) as result_hex,
       concat_ws(',', null, null) is null as is_null;`

### DIFF
--- a/pkg/sql/plan/function/func_binary.go
+++ b/pkg/sql/plan/function/func_binary.go
@@ -1148,7 +1148,7 @@ func ConcatWs(ivecs []*vector.Vector, result vector.FunctionResultWrapper, _ *pr
 			allNull = false
 		}
 		if allNull {
-			if err = rs.AppendBytes(nil, true); err != nil {
+			if err = rs.AppendBytes(nil, false); err != nil {
 				return err
 			}
 			continue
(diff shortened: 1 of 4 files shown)


heni02 (approved):

XuPeng-SH (approved):
Reviewed commit 4e048460.

I checked the fix against the CONCAT_WS contract from #25296:
- NULL separator still returns NULL.
- NULL value arguments are skipped.
- If all value arguments are skipped, the result is a non-NULL empty string.
- Existing empty-string arguments remain distinct from NULL arguments.

The change is scoped to the terminal all-NULL branch and does not introduce new shared state, goroutines, waits, external resources, or unbounded accumulation. Constant folding also evaluates through the normal expression executor path, so the fixed function body covers the issue SQL path.

Local verification:
- `go test -count=1 -timeout 180s ./pkg/sql/plan/function -run 'Test_ConcatWs|TestConcatWs'`
- `go test -count=1 -timeout 300s ./pkg/sql/plan/function`

No blocking issues found.

ouyuanning (approved):

Issue #25369, opened 2026-07-02
[Bug]: COT(0) returns NULL instead of MySQL out-of-range error


Ariznawlll:
### MatrixOne version

latest `main` tested locally: `0e8602b78`

### MySQL version

MySQL 8.0.45

### Steps to Reproduce

```sql
select sqrt(-1), log(-1), log(0), asin(2), acos(2), cot(0);
```

### Expected Behavior

MySQL returns `NULL` for the invalid-domain functions before `cot(0)`, but `cot(0)` raises an out-of-range error:

```text
ERROR 1690 (22003): DOUBLE value is out of range in 'cot(0)'
```

### Actual Behavior

MatrixOne returns `NULL` for `cot(0)` as well and does not raise an error:

```text
NULL  NULL  NULL  NULL  NULL  NULL
```

### Notes

I checked existing issues before filing. #25308 covers `POW` with negative base and fractional exponent; the older #3133 is closed and expected an infinity-like result, while MySQL 8.0.45 currently errors with `ERROR 1690` for `COT(0)`.

daviszhen:
fixed

Ariznawlll:
Verified on latest main 2a008285e.

Result: fixed.

Validation:
```sql
select sqrt(-1), log(-1), log(0), asin(2), acos(2);
select cot(0);
```

MO now matches MySQL behavior: invalid-domain functions return NULL, and `cot(0)` raises error 1690.

Regression added in matrixorigin/motr#39, commit 75bf03f:
- `suites/14_issue_regression/t/issue_25369_cot_zero_domain.test`
- `suites/14_issue_regression/r/issue_25369_cot_zero_domain.result`

MOTR validation passed:
```
[ISSUE_25369_COT_ZERO_DOMAIN] success
```

PR #25430, opened 2026-07-03
fix cot(0)

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25369

修改 :`select cot(0)`

报错而不是返回0

### DIFF
--- a/pkg/sql/plan/function/func_builtin.go
+++ b/pkg/sql/plan/function/func_builtin.go
@@ -3359,7 +3359,10 @@ func builtInCos(parameters []*vector.Vector, result vector.FunctionResultWrapper
 }
 
 func builtInCot(parameters []*vector.Vector, result vector.FunctionResultWrapper, proc *process.Process, length int, selectList *FunctionSelectList) error {
-	return opUnaryFixedToFixedWithNullOnError[float64, float64](parameters, result, proc, length, func(v float64) (float64, error) {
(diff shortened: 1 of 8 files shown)


heni02 (approved):

XuPeng-SH (approved):
Reviewed latest head `700ba01a8cd78b032c01473e0b852d2a1b9f891b`.

The change is scoped and correct for #25369. `cot` now uses the error-propagating unary template, explicitly turns non-NULL zero input into an out-of-range error, and keeps invalid-domain functions such as `sqrt/log/asin/acos` on the existing NULL-on-error behavior. I checked the vector template path: NULL rows are skipped, non-NULL zero rows return the query error, and const/vector execution share the same check.

Coverage looks sufficient for the bug: unit coverage asserts the function returns an error and checks the message; BVT covers the issue query shape, `cot(0)`, `cot(NULL)`, and safe non-zero `cot` values. Existing CI is green.

Local validation:
- `gofmt -d pkg/sql/plan/function/func_builtin.go pkg/sql/plan/function/func_builtin_test.go`
- `go test ./pkg/sql/plan/function -run Test_BuiltIn_Math -count=1 -timeout 120s`
- `go test ./pkg/sql/plan/function -count=1 -timeout 180s`

ouyuanning (approved):

PR #25432, opened 2026-07-03
set default incremental checkpoint interval to 5m

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24900

set default incremental checkpoint interval to 5m


XuPeng-SH (changes requested):
Requesting changes after reviewing latest head `0b98efc2a39ad1d080796fcd2305f76a0bacd189` against `4.1-dev`.

The core Go default change is internally consistent, but the behavior is not closed end-to-end yet.

1. The launch/embed configs still explicitly override the new default with `incremental-interval = "180s"`.

This PR changes the zero-value default to 5 minutes, but all of these shipped startup templates still set an explicit 3 minute interval, so deployments using them will not see the new default:

- `pkg/embed/template.go`
- `etc/launch-dynamic-with-proxy/tn.toml`
- `etc/launch-dynamic-cn/tn.toml`
- `etc/launch-multi-cn/tn.toml`
- `etc/launch-with-proxy/tn.toml`
- `etc/launch-minio-local/tn.toml`
- `etc/launch-tae-compose/config/tn.toml`

If the intended release behavior is 5m, those explicit configs need to be updated or removed so the Go default can apply. Otherwise this only changes code paths where the field is omitted, while the common packaged/dev launch configs keep the old value.

2. The arena pool drain timer still assumes the old 1m checkpoint interval.

`pkg/objectio/arena_pool.go` documents `arenaDrainDelay` as 2x the default incremental checkpoint interval so the timer is reset by steady-state checkpointing and pools stay warm. The value is still `2 * time.Minute`; after this PR, the default checkpoint interval becomes 5m, so the drain timer can fire between normal checkpoints and empty the arena free lists during active operation. That is a performance regression risk and contradicts the documented invariant. Please update this delay/design alongside the new default, or justify why the old 2m drain is still correct after changing checkpoint cadence.

3. Please add focused regression coverage for the default value propagation.

At minimum, cover the three default-entry paths changed here:
- `tnservice.Config.Validate` / `SetDefaultValue`
- `options.Options.FillDefaults`
- `checkpoint.CheckpointCfg.FillDefaults`

That would catch drift between TN config, TAE options, and checkpoint runtime defaults.

Local verification I ran:
- `go test -count=1 -timeout 240s ./pkg/vm/engine/tae/options ./pkg/vm/engine/tae/db/checkpoint ./pkg/tnservice`
- `go test -count=1 -timeout 240s ./pkg/vm/engine/tae/options ./pkg/vm/engine/tae/db/checkpoint ./pkg/tnservice -run '^$'`
- `go vet ./pkg/vm/engine/tae/options ./pkg/vm/engine/tae/db/checkpoint ./pkg/tnservice`

Those passed locally, but the end-to-end config/behavior gaps above should be fixed before merge.

XuPeng-SH (changes requested):
Requesting changes for one remaining inconsistency:

- `etc/docker-multi-cn-local-disk/generate-config.sh:442` still emits `incremental-interval = "60s"`.
- The rest of the PR consistently moves the default incremental checkpoint interval to 5 minutes (`pkg/vm/engine/tae/options/types.go`, `pkg/tnservice/cfg.go`, `pkg/vm/engine/tae/db/checkpoint/cfg.go`, the touched `etc/launch-*` TN configs, and `pkg/embed/template.go`), so this local docker config generation path is still on the old value.
- If this divergence is intentional, it should be called out explicitly; otherwise it leaves one non-CI startup path silently outside the new default and defeats the stated goal of the PR.

Suggested fix: update the generator to `300s`, or route it through the same shared default/template source as the other launcher configs.

XuPeng-SH (changes requested):
Requesting changes after reviewing the latest head.

1. `checkpointIntentOldAge` is still hardcoded to 2 minutes (`pkg/vm/engine/tae/db/checkpoint/runner.go:40`), and that same value drives both `stalledCheckpointFlushAge` (`pkg/vm/engine/tae/db/checkpoint/flusher.go:46`) and the "stop extending old pending intent" rule (`pkg/vm/engine/tae/db/checkpoint/store.go:192-199`).

This PR changes the default incremental checkpoint interval to 5 minutes (`pkg/vm/engine/tae/options/types.go:44`, propagated via `pkg/tnservice/cfg.go` and `pkg/vm/engine/tae/db/checkpoint/cfg.go`), but a pending checkpoint is still treated as old/stalled after only 2 minutes. That means the bounded-flush path and the intent-freezing behavior both kick in 3 minutes earlier than the new default cadence and earlier than the linked issue's 5-minute expectation. Please move these thresholds together or derive them from the same source, and add a regression test for that coupling.

2. `pkg/objectio/arena_pool.go` now hardcodes `arenaDrainDelay = 10 * time.Minute`, but `ScheduleArenaDrain()` is called not only from checkpoint completion (`pkg/vm/engine/tae/db/checkpoint/runner.go:447-449`) but also from CN sinker close (`pkg/objectio/ioutil/sinker.go:171-183`).

So this is broader than "keep pools warm between 5-minute checkpoints": it globally delays off-heap RSS reclamation for CN/TN write paths and for deployments that intentionally configure a shorter `incremental-interval`. Please avoid baking a global 10-minute delay into objectio; either wire the delay from config/caller or limit the longer delay to the checkpoint-driven path.

Follow-up suggestion: the new unit coverage is better, but it still does not lock the unhappy paths above. I would add one test that ties the stalled-checkpoint threshold to the intended 5-minute behavior, and one that covers arena-drain behavior or explicit interval override preservation end to end.

XuPeng-SH (changes requested):
The latest head resolves the previous threshold-coupling and arena-drain concerns, but one configuration-lifecycle gap remains. `WithStalledCheckpointFlushAge` stores the value only on `flushImpl`; `FlushCfg`, `GetCfg`, and `WithFlusherCfg` do not preserve it. This is production-reachable: `Controller.SwitchTxnMode` captures `BGFlusher.GetCfg()`, stops the flusher, and its replay rollback restarts with `WithFlusherCfg(flushCfg)`; the replay-to-write path calls `BGFlusher.Restart()` with no options. Both paths create a new impl whose `fillDefaults` resets the stalled age to 5m. For an explicit interval such as 7m (or 1m), `runnerStore.intentOldAge` remains at the configured value while the restarted flusher uses 5m, reintroducing the exact threshold divergence this update is intended to remove. I reproduced it with a focused test using the production rollback sequence: initialize with 7m, `cfg := f.GetCfg(); f.Stop(); f.Restart(WithFlusherCfg(cfg))`; expected 7m, actual 5m. Please preserve the effective stalled age across all restart paths (ideally make restart configuration transfer complete rather than patching only one caller), and add non-default-interval tests for replay rollback and replay-to-write restart. The static default propagation, checkpoint/store thresholds, workload-specific arena delays, existing package tests, race repetitions, and vet otherwise look good.

XuPeng-SH (approved):

Issue #25330, opened 2026-07-01
[Bug]: maximum unsigned integer LIMIT panics in top operator


Ariznawlll:
### Branch

main at `200bfa7e0efb`.

### Summary

`LIMIT 18446744073709551615` is accepted by MySQL as the maximum unsigned integer limit and returns all matching rows. MatrixOne panics in the `top` operator.

### Reproduction

```sql
drop database if exists mysql_compat_module9_limit_boundary;
create database mysql_compat_module9_limit_boundary;
use mysql_compat_module9_limit_boundary;

create table t(id int);
insert into t values (1),(2),(3),(4),(5),(6),(7);

select id from t order by id limit 18446744073709551615;
```

### MySQL result

```text
+------+
| id   |
+------+
|    1 |
|    2 |
|    3 |
|    4 |
|    5 |
|    6 |
|    7 |
+------+
```

### MatrixOne result

```text
ERROR 20101 (HY000): internal error: panic runtime error: index out of range [0] with length 0:
runtime.panicBounds64
...
github.com/matrixorigin/matrixone/pkg/sql/colexec/top.(*container).processBatchSpill
    pkg/sql/colexec/top/top.go:410
```

### Expected behavior

MatrixOne should not panic. It should return all rows for this maximum unsigned integer limit, matching MySQL.

jiangxinmeng1:
fixed by #25433

Ariznawlll:
Verified on main build 2a008285e, which includes the fix from #25433.

Result: fixed.

Validation:
```sql
drop database if exists mysql_compat_module9_limit_boundary;
create database mysql_compat_module9_limit_boundary;
use mysql_compat_module9_limit_boundary;

create table t(id int);
insert into t values (1),(2),(3),(4),(5),(6),(7);

select id from t order by id limit 18446744073709551615;
```

MO now returns all 7 rows and no longer panics in the top operator.

Regression added in matrixorigin/motr#39, commit 53cef5b:
- `suites/14_issue_regression/t/issue_25330_limit_uint64_max.test`
- `suites/14_issue_regression/r/issue_25330_limit_uint64_max.result`

MOTR validation passed:
```
[ISSUE_25330_LIMIT_UINT64_MAX] success
SUMMARY | total=1 success=1 failed=0 skipped=0 | stmts=7 ok=7 fail=0
```

PR #25433, opened 2026-07-03
fix top max uint64 limit overflow

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25330

fix top max uint64 limit overflow

### DIFF
--- a/pkg/sql/colexec/mergetop/top.go
+++ b/pkg/sql/colexec/mergetop/top.go
@@ -195,17 +195,15 @@ func (ctr *container) build(ap *MergeTop, proc *process.Process, analyzer proces
 }
 
 func (ctr *container) processBatch(limit uint64, bat *batch.Batch, proc *process.Process) error {
-	rowCount := int64(bat.RowCount())
-	toFillCount := int64(limit) - int64(len(ctr.sels))
-
-	processCount := min(int64(toFillCount), rowCount)
+	rowCount := bat.RowCount()
+	processCount := rowsToFill(limit, len(ctr.sels), rowCount)
 
 	if processCount > 0 {
 		for j, vec := range ctr.bat.Vecs {
 			if err := vec.UnionBatch(
 				bat.Vecs[j],
 				0,
-				int(processCount),
+				processCount,
 				nil,
 				proc.Mp(),
 			); err != nil {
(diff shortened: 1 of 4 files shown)


XuPeng-SH (approved):
Reviewed latest head `dc418bb88f142a33afa0aa44fab861cdd2089989` for #25330.

The fix is directionally correct: the old signed conversion path (`int64(limit)`) made `LIMIT MaxUint64` become negative, skipped heap population, and then reached `sels[0]` with an empty heap. The new `rowsToFill` helper keeps the arithmetic in uint64, only converts after bounding by the current batch row count, and applies the same logic to `top`, spill `top`, and `mergetop`.

I checked the surrounding closure as well:
- `LIMIT 0` still exits before these paths.
- `currentRows >= limit` still means the heap is already full for positive limits.
- `LIMIT MaxUint64` now behaves as “take all rows seen” and sorts at eval time because the heap never reaches that artificial limit.
- The spill path still records `rowRefs` only for rows actually retained and cleans spill state through the existing reset/free paths.
- The compile optimization for `ORDER BY ... LIMIT m OFFSET n` already guards `m+n` overflow and falls back to sort/offset/limit, so this PR does not reintroduce a compile-layer overflow there.

Q1/Q2/Q3 unhappy-path pass: no new resource ownership or wait path is introduced; accumulation remains bounded by retained input rows, with the existing spill path for large limits.

Local verification:
- `go test -count=1 -timeout 240s ./pkg/sql/colexec/top ./pkg/sql/colexec/mergetop`
- `go test -count=1 -timeout 240s ./pkg/sql/colexec/top ./pkg/sql/colexec/mergetop -run '^$'`
- `go vet ./pkg/sql/colexec/top ./pkg/sql/colexec/mergetop`
- `go test -count=1 -timeout 240s ./pkg/sql/compile -run 'TestDupOperatorMergeTop|Test_convertToProcessLimitation'`
- `go test -count=1 -timeout 240s ./pkg/sql/plan -run 'TestLimitBinder_LargeUint64'`

Non-blocking: a SQL-level regression case matching the issue reproduction would be useful in BVT, but the operator-level coverage added here hits the actual panic point and both execution variants changed by the PR.

Approving. Required CI is still partially pending at review time.

aunjgr (approved):
Clean overflow fix. Keep arithmetic in uint64, only convert after bounding. Good test coverage on both top and mergetop.

Issue #25363, opened 2026-07-02
[Bug]: CONVERT binary invalid UTF-8 USING utf8mb4 returns raw bytes instead of NULL


Ariznawlll:
## Summary

On latest `main` (`0e8602b78`), `CONVERT(_binary ... USING utf8mb4)` does not handle invalid UTF-8 bytes like MySQL.

MySQL returns `NULL` with warning `Invalid utf8mb4 character string`, but MatrixOne returns the raw invalid byte as a non-NULL string value.

## Duplicate check

Searched before filing:

- `Invalid utf8mb4 character string`
- `CONVERT USING utf8mb4 invalid utf8`
- `CONVERT USING invalid character`

Related but not duplicate:

- #25302 covers inserting invalid UTF-8 bytes into a `utf8mb4 VARCHAR` column.

This issue covers expression-level charset conversion with `CONVERT(... USING utf8mb4)`.

## Reproduce

```sql
select convert(_binary x'ff' using utf8mb4) is null,
       hex(convert(_binary x'ff' using utf8mb4)),
       length(convert(_binary x'ff' using utf8mb4));
```

## MySQL 8.0.45 result

```text
1  NULL  NULL
```

MySQL warning:

```text
Warning 1300 Invalid utf8mb4 character string: 'FF'
```

## MatrixOne result

```text
0  FF  1
```

## Expected behavior

MatrixOne should return `NULL` for invalid UTF-8 charset conversion and report a warning/error consistent with MySQL compatibility expectations.

daviszhen:
fixed

Ariznawlll:
Verified on latest main `a03bfc1b6` (`FIX: FileService implements new DISK-V2 backend to replace DISK (#25272)`).

Validation:

```sql
select convert(_binary x'ff' using utf8mb4) is null as is_null,
       hex(convert(_binary x'ff' using utf8mb4)) as hx,
       length(convert(_binary x'ff' using utf8mb4)) as len;
```

MO result now matches MySQL 8.0.45:

```text
is_null  hx    len
1        NULL  NULL
```

I also checked additional invalid UTF-8 byte sequences and a valid control case:

- `x'41'`: `hex=41`, `length=1`
- `x'ff'`: `NULL`
- `x'e4b8'`: `NULL`
- `x'c328'`: `NULL`

Regression added to motr PR: matrixorigin/motr#39

MOTR validation:

```text
[...]

PR #25434, opened 2026-07-03
fix convert to utf8mb4

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25363

修改 `select convert(_binary x'ff' using utf8mb4) is null,
       hex(convert(_binary x'ff' using utf8mb4)),
       length(convert(_binary x'ff' using utf8mb4));`


pkg/sql/plan/function/func_builtin.go
```diff
@@ -3577,11 +3578,60 @@ func (op *opBuiltInRand) builtInRand(parameters []*vector.Vector, result vector.
...
-	}, selectList)
+func builtInConvertUsingCharset(parameters []*vector.Vector, result vector.FunctionResultWrapper, _ *process.Process, length int, selectList *FunctionSelectList) error {
+	result.UseOptFunctionParamFrame(2)
+	rs := vector.MustFunctionResult[types.Varlena](result)
+	p1 := vector.OptGetBytesParamFromWrapper(rs, 0, parameters[0])
+	p2 := vector.OptGetBytesParamFromWrapper(rs, 1, parameters[1])
+	rsNull := rs.GetResultVector().GetNulls()
+
+	if selectList != nil && selectList.IgnoreAllRow() {
+		nulls.AddRange(rsNull, 0, uint64(length))
```
XuPeng-SH: `nulls.AddRange` alone is not enough for this varlena result. `PreExtendAndReset` resets the vector length to 0 and, for varlena, does not set it back to `length`; this branch therefore returns a zero-length vector with null bits set out of band. Please use `rs.SetNullResult(uint64(length))` or append `length` NULL rows, and make the ignore-all test assert `resultVec.Length()` as well as the null bitmap.

XuPeng-SH (changes requested):
Requesting changes for one correctness gap in the new CONVERT USING implementation.

The value semantics for invalid utf8/utf8mb4 bytes look right and the tests cover the main branches, but the all-null selectList path currently returns a zero-length varlena result. `FunctionResult.PreExtendAndReset` resets varlena vector length to 0 and does not pre-set it to `length`, so only adding bits to the null bitmap is not sufficient. Downstream operators observe `Vector.Length()`, not just the null bitmap, so this can drop/misalign rows when the function is evaluated under an ignore-all/select-all-null path.

Please make that branch materialize `length` NULL rows, e.g. with `rs.SetNullResult(uint64(length))` or by appending NULLs, and update `TestBuiltInConvertUsingCharsetIgnoreAllRows` to assert `resultVec.Length() == fcTC.fnLength` in addition to the null bitmap.

Local checks run:
- `make err-check`
- `go test -count=1 -timeout 300s ./pkg/sql/plan/function -run 'TestBuiltInConvertUsing'`
- `go test -count=1 -timeout 300s ./pkg/sql/plan -run TestConvertCharBinaryTypeResolution`

Non-blocking note: the MySQL warning-count/SHOW WARNINGS compatibility remains unimplemented here; current MO warning plumbing does not appear to provide a normal builtin-function warning path, so I am not blocking this PR on that broader compatibility gap.

heni02 (approved):

daviszhen:
> Requesting changes for one correctness gap in the new CONVERT USING implementation.
> 
> The value semantics for invalid utf8/utf8mb4 bytes look right and the tests cover the main branches, but the all-null selectList path currently returns a zero-length varlena result. `FunctionResult.PreExtendAndReset` resets varlena vector length to 0 and does not pre-set it to `length`, so only adding bits to the null bitmap is not sufficient. Downstream operators observe `Vector.Length()`, not just the null bitmap, so this can drop/misalign rows when the function is evaluated under an ignore-all/select-all-null path.
> 
> Please make that branch materialize `length` NULL rows, e.g. with `rs.SetNullResult(uint64(length))` or by appending NULLs, and update `TestBuiltInConvertUsingCharsetIgnoreAllRows` to assert `resultVec.Length() == fcTC.fnLength` in addition to the null bitmap.
> 
> Local checks run:
> 
> * `make err-check`
> * `go test -count=1 -timeout 300s ./pkg/sql/plan/function -run 'TestBuiltInConvertUsing'`
> * `go test -count=1 -timeout 300s ./pkg/sql/plan -run TestConvertCharBinaryTypeResolution`
> 
> Non-blocking note: the MySQL warning-count/SHOW WARNINGS compatibility remains unimplemented here; current MO warning plumbing does not appear to provide a normal builtin-function warning path, so I am not blocking this PR on that broader compatibility gap.

已经修改

XuPeng-SH (approved):
I did not find a blocker in this fix.

The new `builtInConvertUsingCharset` behavior is consistent with the targeted bugfix: invalid UTF-8 now becomes `NULL` for `utf8/utf8mb4`, while NULL propagation, const-arg handling, and select-list skipping all still behave correctly. I also checked the unhappy path surface here: there is no new wait/lifecycle risk, and the function-package tests passed.

One non-blocking follow-up suggestion: add a tiny SQL/BVT case proving non-UTF8 charsets still pass invalid bytes through (for example `select hex(convert(_binary xff using latin1));` expecting `FF`).

ouyuanning (approved):

aunjgr (approved):

PR #25436, opened 2026-07-03
fix: support FOR UPDATE in parenthesized UNION branches

### LINKED ISSUE #23134: FOR UPDATE unsupported on UNION/UNION ALL
## Description

对 `UNION/UNION ALL` 结果集执行 `FOR UPDATE` 会报 `internal error: not support select union for update`。这阻止了将多个子查询合并后整体加锁的常见模式。

## Error Message

```
ERROR 20101 (HY000) at line 8: internal error: not support select union for update
```

## Reproduction

```sql
DROP TABLE IF EXISTS temp_fu_union_2479;
CREATE TABLE temp_fu_union_2479 AS
SELECT id, company_id, status
FROM company_patent
WHERE company_id IS NOT NULL
LIMIT 20;

START TRANSACTION;
WITH combined AS (
    SELECT id, company_id FROM temp_fu_union_2479 WHERE id % 2 = 0
    UNION ALL
    SELECT id, company_id FROM temp_fu_union_2479 WHERE id % 2 = 1
)
SELECT * FROM combined FOR UPDATE;
```

## Impact
- 无法对合并后的结果集加锁，影响多路数据的批处理。
- 与 MySQL 行为不一致，降低语法兼容性。

## References
- `report_test_batch_248_for_update_boundary_20251124_162712.md` Case 2479 起所有 `for_update_union_all` 用例。

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #23134

This supports MySQL-style branch-local `FOR UPDATE` in parenthesized UNION statements, for example:

```sql
(select n_name from nation for update) union all (select n_name from nation2 for update)
```

The planner now restores `isForUpdate` after binding each UNION branch so a branch-local lock does not leak into sibling branches.

Based on the MySQL behavior verified in the issue discussion, this PR intentionally does not treat an outer CTE `FOR UPDATE` as a lock on the underlying UNION inputs. `WITH qn AS (...) SELECT * FROM qn FOR UPDATE` can plan successfully but does not add a `LOCK_OP` for the CTE body. Non-parenthesized/trailing whole-UNION `FOR UPDATE` remains unsupported.

Tests added for:
- two parenthesized UNION branches with `FOR UPDATE`
- one locked branch plus one unlocked branch, to guard state leakage
- outer CTE `FOR UPDATE` producing no lock op
- non-parenthesized UNION `FOR UPDATE` remaining an error

Verification:

```sh
CGO_ENABLED=1 CGO_CFLAGS="-I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -Wl,-rpath,$(pwd)/thirdparties/install/lib" LD_LIBRARY_PATH="$(pwd)/thirdparties/install/lib:${LD_LIBRARY_PATH}" go test ./pkg/sql/plan -run TestUnionSqlBuilder -count=1
CGO_ENABLED=1 CGO_CFLAGS="-I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -Wl,-rpath,$(pwd)/thirdparties/install/lib" LD_LIBRARY_PATH="$(pwd)/thirdparties/install/lib:${LD_LIBRARY_PATH}" go test ./pkg/sql/plan -run 'Test(SingleTableSQLBuilder|JoinTableSqlBuilder|UnionSqlBuilder)$' -count=1\nCGO_ENABLED=1 CGO_CFLAGS="-I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -Wl,-rpath,$(pwd)/thirdparties/install/lib" LD_LIBRARY_PATH="$(pwd)/thirdparties/install/lib:${LD_LIBRARY_PATH}" go test ./pkg/sql/plan -count=1\n```


XuPeng-SH (approved):
LGTM.

I re-checked the tricky CTE path against the issue discussion as well. The branch-local `FOR UPDATE` fix itself looks right: restoring `isForUpdate` after each UNION branch closes the state-leak bug, so `(select ... for update) union all (select ...)` only locks the intended branch and `(select ... for update) union all (select ... for update)` locks both.

I also do not treat the outer-CTE case as a blocker here. The MySQL behavior captured in issue #23134 is exactly that `WITH qn AS (...) SELECT * FROM qn FOR UPDATE` can parse successfully but does **not** lock the CTE body's base-table rows, while parenthesized UNION branches with their own `FOR UPDATE` do lock. This patch matches that compatibility target.

Only non-blocking gap I see is coverage: adding a pessimistic-txn SQL/BVT case that proves the actual blocking behavior for the one-locked-branch vs two-locked-branches UNION forms would make the regression coverage stronger than the current planner-only assertions.

PR #25438, opened 2026-07-03
fix sql_mode parser modes

### LINKED ISSUE #25407: [Bug]: SQL_MODE parser modes ANSI_QUOTES/PIPES_AS_CONCAT/NO_BACKSLASH_ESCAPES/REAL_AS_FLOAT are not MySQL-compatible
### Environment

- Branch: latest `main`
- MO git version: `51656deb6`
- MySQL oracle: `8.0.45`

### Reference behavior

MySQL SQL modes can change parser/type semantics such as `ANSI_QUOTES`, `PIPES_AS_CONCAT`, `NO_BACKSLASH_ESCAPES`, and `REAL_AS_FLOAT`:
https://dev.mysql.com/doc/en/sql-mode.html

TiDB also documents strict and compatibility SQL mode behavior:
https://docs.pingcap.com/tidb/stable/sql-mode/

### Reproduce

```sql
drop database if exists mysql_compat_model13_sql_mode;
create database mysql_compat_model13_sql_mode;
use mysql_compat_model13_sql_mode;

set sql_mode = '';
select "abc" as dq_string, length("abc") as dq_len;

set sql_mode = 'ANSI_QUOTES';
create table "t_ansi" ("id" int primary key, "v" varchar(10));
insert into "t_ansi" ("id","v") values (1,'ok');
select "id","v" from "t_ansi";
select "abc" as ansi_should_be_identifier;

set sql_mode = '';
select 'a'||'b' as pipes_default;
set sql_mode = 'PIPES_AS_CONCAT';
select 'a'||'b' as pipes_concat, concat('a','b') as concat_control;

set sql_mode = '';
select length('a\\nb') as default_len, hex('a\\nb') as default_hex;
set sql_mode = 'NO_BACKSLASH_ESCAPES';
select length('a\\nb') as no_bs_len, hex('a\\nb') as no_bs_hex;

set sql_mode = '';
create table t_real_default (r real);
desc t_real_default;
set sql_mode = 'REAL_AS_FLOAT';
create table t_real_float (r real);
desc t_real_float;
```

### MySQL result

```text
-- default double quotes are string literals
dq_string  dq_len
abc        3

-- ANSI_QUOTES: double quotes are identifiers
id  v
1   ok
ERROR 1054 (42S22): Unknown column 'abc' in 'field list'

-- default || is logical OR; PIPES_AS_CONCAT changes it to concat
pipes_default
0
pipes_concat  concat_control
ab            ab

-- NO_BACKSLASH_ESCAPES changes literal parsing
default_len  default_hex
4            615C6E62
no_bs_len    no_bs_hex
5            615C5C6E62

-- REAL_AS_FLOAT changes REAL from DOUBLE to FLOAT
t_real_default.r: double
t_real_float.r:   float
```

### MO result

```text
-- ANSI_QUOTES did not enable quoted identifiers
ERROR 1064 near create table "t_ansi" ...
ERROR 1064 near insert into "t_ansi" ...
ERROR 1064 near select "id","v" from "t_ansi" ...

-- "abc" still behaves as a string literal under ANSI_QUOTES
ansi_should_be_identifier
abc

-- || behaves as concat even without PIPES_AS_CONCAT
pipes_default
ab

-- NO_BACKSLASH_ESCAPES did not change literal parsing
default_len  default_hex
4            615C6E62
no_bs_len    no_bs_hex
4            615C6E62

-- REAL_AS_FLOAT did not change REAL column type
t_real_default.r: DOUBLE(0)
t_real_float.r:   DOUBLE(0)
```

### Expected

MO should honor these MySQL-compatible SQL modes:

- `ANSI_QUOTES`: double quotes identify identifiers, not string literals.
- `PIPES_AS_CONCAT`: `||` should be logical OR unless this mode is enabled.
- `NO_BACKSLASH_ESCAPES`: backslash should be treated as an ordinary character in string literals.
- `REAL_AS_FLOAT`: `REAL` should map to `FLOAT` instead of `DOUBLE`.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25407

This wires parser-level `sql_mode` settings into the MySQL parser and scanner:

- `ANSI_QUOTES` makes double-quoted text parse as identifiers.
- `PIPES_AS_CONCAT` keeps `||` as concat; otherwise `||` parses as logical OR.
- `NO_BACKSLASH_ESCAPES` keeps backslashes as ordinary string characters.
- `REAL_AS_FLOAT` maps `REAL` columns to FLOAT instead of DOUBLE.

## Root Cause

`sql_mode` accepted these mode names, but frontend parsing only passed SQL text and `lower_case_table_names` into the parser. The scanner and yacc actions therefore used fixed behavior regardless of session `sql_mode`.


iamlinjunhong (approved):

XuPeng-SH (changes requested):
Requesting changes: the top-level query parser now honors session `sql_mode`, but the prepared-string path is still not wired through, so this PR does not close the functional loop for SQL that is prepared from a string or user variable.

Blocking issue:

`pkg/frontend/mysql_cmd_executor.go:1315` still parses `PrepareString.Sql` with `mysql.Parse(execCtx.reqCtx, st.Sql, v.(int64))`, and `pkg/sql/plan/build_dcl.go:77` re-parses the same prepared string with `mysql.Parse(ctx.GetContext(), pstmt.Sql, v.(int64))`. Both calls force empty parser SQL mode. That means cases like these still parse the inner SQL with the old/default behavior instead of the current session mode:

```sql
set sql_mode = 'PIPES_AS_CONCAT';
prepare s from "select 'a'||'b'";
execute s;

set sql_mode = 'ANSI_QUOTES';
prepare s from 'select "id" from "t_ansi"';

set sql_mode = 'NO_BACKSLASH_ESCAPES';
prepare s from "select hex('a\\nb')";

set sql_mode = 'REAL_AS_FLOAT';
prepare s from 'create table t_real_float (r real)';
```

The fix should use the same mode derivation added for normal parsing, i.e. resolve `sql_mode`, pass it through `mysql.SessionSQLModeForParser`, and call `ParseWithSQLMode` / `ParseOneWithSQLMode` in both the frontend pre-parse and the plan builder re-parse. Please add regression coverage for prepared-string and `PREPARE ... FROM @var` paths, not only direct parser tests.

Also worth fixing while touching this block: in `stmt_kind.go`, `preStmt.Free()` is deferred before checking the parse error. If parsing the prepared string fails, `preStmt` can be nil and the error path can panic. Move the defer after the `err != nil` check.

Local checks run:
- `make err-check`
- `go test -count=1 -timeout 300s ./pkg/sql/parsers/dialect/mysql` with local CGo/rpath setup
- `go test -count=1 -timeout 300s ./pkg/sql/parsers -run 'TestParse|TestSplitSqlBySemicolon'` with local CGo/rpath setup

CI was still mostly pending when reviewed.

ouyuanning (approved):

XuPeng-SH (changes requested):
Requesting changes for one blocker:

- `pkg/frontend/mysql_cmd_executor.go:2468-2511` currently applies `sql_mode` only once to the full incoming SQL packet *before* any statement executes.
- MatrixOne enables `CLIENT_MULTI_STATEMENTS` by default (`pkg/frontend/mysql_protocol.go:54-65`), so a request like `SET sql_mode=ANSI_QUOTES; SELECT "c";` or `SET sql_mode=REAL_AS_FLOAT; CREATE TABLE t(r REAL);` still parses the second statement under the *old* mode. That means session-scoped parser behavior is still wrong for later statements in the same request.
- Suggested fix: parse multi-statement packets incrementally (or split/reparse after `SET sql_mode`) so a leading `SET sql_mode` can affect subsequent statements in that same packet.

I would also add a regression test for the exact same-request pattern above once the execution path is fixed.

aunjgr (approved):
Approach is correct. Wiring sql_mode into the scanner is the right level. Minor suggestions:

1. `reparseAfterStaticSQLModeSet` silently returns on fragment count mismatch — a debug log would help catch unexpected mismatches (e.g., CREATE PROCEDURE bodies with semicolons).

2. `fragmentHasStatement` could be simplified: `strings.TrimSpace(fragment) != ""` is sufficient since the parser already validates on reparse. The scanner.Scan() call is redundant.

3. `defer mysql.PutScanner(scanner)` in fragmentHasStatement stacks defers in multi-statement batches. Consider manually putting the scanner back to avoid defer accumulation.

XuPeng-SH (changes requested):
The latest head fixes the original `ANSI_QUOTES` same-packet blocker, but one same-packet sql_mode case is still broken.

`parseSql()` still does an initial full-packet parse under the session's *current* mode, and only if that succeeds does it call `reparseAfterStaticSQLModeSet()` to reparse later statements under the updated mode. That means a later statement which **needs the new mode just to be tokenized correctly** can still fail before the reparse step ever runs.

A concrete repro is:

```sql
set sql_mode=NO_BACKSLASH_ESCAPES; select 'a\'; select 1
```

Under `NO_BACKSLASH_ESCAPES`, this should be two valid statements: the second statement is `select 'a\'` (a string ending with a backslash), followed by `select 1`. But the initial parse still runs under the old mode, so the lexer treats `\'` with backslash-escape semantics and the whole packet fails with a syntax error before `reparseAfterStaticSQLModeSet()` can help.

The root cause is that the fallback is "parse whole packet once, then maybe reparse fragments", and both the initial full-packet parse and `SplitSqlBySemicolon()` are still mode-unaware at the point where the mode change matters most.

So the multi-statement sql_mode problem is only partially fixed: later statements are reparsed under the new mode *after* a successful first parse, but modes like `NO_BACKSLASH_ESCAPES` can require the new mode for the packet to parse at all. This still needs a request-changes outcome.

XuPeng-SH (changes requested):
The current head fixes the same-packet `NO_BACKSLASH_ESCAPES` path much better, but I still see two substantive blockers.

1. **View definitions still lose the parser `sql_mode` they were created with.** `ViewData` still only persists the raw SQL/default DB/security metadata, and the stored view text is still reparsed later with bare `mysql.Parse(...)`. That means any view created under mode-sensitive parsing (including MatrixOne's default legacy `||`-as-concat behavior) can change meaning or fail when expanded/restored. The same problem still exists on snapshot/PITR restore paths that reparse stored view SQL.

2. **`PIPES_AS_CONCAT` still has OR-precedence instead of concat-precedence.** The scanner now emits a distinct `PIPE_CONCAT` token, but the grammar still declares `%left OR PIPE_CONCAT`, so mixed expressions are still rebound wrong under `PIPES_AS_CONCAT`. This is especially risky because MatrixOne's default parser mode still injects `PIPES_AS_CONCAT`. The new tests only cover the simple `'a'||'b'` case, not precedence against arithmetic/comparison/logical operators.

So from my side this is still `REQUEST_CHANGES`: same-packet parsing looks much better now, but view sql_mode persistence and `||` precedence are still real correctness issues.

XuPeng-SH (changes requested):
Requesting changes for one remaining blocker.

The latest head fixes the earlier prepared-string path, same-packet static sql_mode parsing, and the PIPES_AS_CONCAT precedence issue. I also verified the latest diff now gives PIPE_CONCAT concat-level precedence instead of OR-level precedence.

However, persisted view SQL still does not carry the parser sql_mode used when the view was created:

- pkg/sql/plan/types.go:172 keeps ViewData as Stmt, DefaultDatabase, SecurityType only.
- pkg/sql/plan/build_ddl.go:83 and :175 store only the raw root SQL in ViewData when creating/altering views.
- pkg/sql/plan/query_builder.go:4752 still reparses the stored view text with bare mysql.Parse(...), so expansion always uses the empty/default parser mode, not the view creation mode.

That is a correctness blocker after this PR because parser semantics are now mode-dependent. For example:

set sql_mode = PIPES_AS_CONCAT;
create view v_pipe as select 'a'||'b' as c;
set sql_mode = '';
select * from v_pipe;

The stored view text still contains ||, but bindView reparses it without PIPES_AS_CONCAT, so the view can be rebound as logical OR instead of concat. The same class of bug applies to ANSI_QUOTES view text that uses double-quoted identifiers, NO_BACKSLASH_ESCAPES string literals, and any old view text that depended on MatrixOne legacy ||-as-concat behavior.

Suggested fix: make the view representation first-principled. Either persist the creation-time parser sql_mode in ViewData and use mysql.ParseWithSQLMode when expanding/reparsing the view, or store a canonical mode-independent SQL string at CREATE/ALTER VIEW time. Keep empty mode backward-compatible for old ViewData. Please add regressions that create a view under PIPES_AS_CONCAT and ANSI_QUOTES, then reset/change sql_mode and query the view; a restore/snapshot-path regression would be ideal if practical.

XuPeng-SH (changes requested):
Requesting changes. I re-reviewed the latest head, including the earlier prepared-statement, same-packet literal mode, precedence, scanner reuse, and persisted-view fixes. Those earlier blockers are addressed. Two concrete issues remain.

1. **High - expression-valued session `sql_mode` is silently ignored for following statements in the same packet.**

`doSetVar` deliberately accepts an arbitrary expression through `getExprValue`, and MySQL syntax is `SET variable = expr`. However, `parseSql` only switches its parser mode when `staticSQLModeSetValue` recognizes a literal/default/unresolved-name AST. For example:

```sql
SET sql_mode = CONCAT('ANSI_', 'QUOTES');
SELECT "c";
```

With an initially empty mode, the PR returns the second AST as a string literal. At execution time the first statement successfully sets the session mode to `ANSI_QUOTES`, so the same `SELECT` should instead contain a quoted identifier (and fail with unknown column if `c` does not exist). This is not a hypothetical expression form: the repository BVT already uses `SET sql_mode = concat_ws(',', @@sql_mode, ...)`. The same gap applies to `@mode`, system-variable expressions, scalar expressions, and all four lexer-affecting modes.

Please make same-packet semantics correct for *all valid* session `SET sql_mode = expr` forms. A static AST evaluator cannot be the correctness boundary: pre-evaluating the expression is also wrong for errors, user variables, subqueries, and nondeterministic expressions. The first-principles fix is to stage parsing/execution so a mode-changing SET is executed before lexing the next statement (with a literal fast path only as an optimization). Add regressions for `CONCAT`, a user-variable value, and a failing expression.

2. **Medium - literal-mode reparse is quadratic for legal compound statements.**

After a static `SET sql_mode`, `parseStatementsWithStaticSQLModeSet` calls `parseFirstStatementWithSQLMode`. That function scans every semicolon and invokes the full parser on the whole prefix each time. A `BEGIN ... END` body has valid internal semicolons but cannot parse until its final `END`, so each prefix is reparsed and discarded. Local measurement of `SET sql_mode=ANSI_QUOTES; BEGIN SELECT 1; ... END; SELECT "c"` was 25/100/400 internal statements in about 0.66/2.63/40.34ms, which is the expected superlinear behavior. Larger routine/compound bodies turn this into avoidable frontend CPU amplification.

Please replace retrying each growing prefix with a single-pass, mode-aware statement-boundary/parser path, ideally the same staged execution design needed above. Add a compound-body regression and a benchmark/complexity guard so this does not regress.

XuPeng-SH (changes requested):
Requesting changes. The latest head addresses the previous prepared-string, same-packet expression, compound-statement complexity, precedence, and normal view-binding issues. I still found three mode-sensitive reparse paths that bypass the mode contract.

1. **High — Snapshot/PITR restore loses the persisted view parser mode.** `ViewData.SQLMode` is used by `bindView`, but the restore graph reads only the raw `SHOW CREATE VIEW` text into `tableInfo.createSql`. `SHOW CREATE VIEW` returns `ViewData.Stmt` verbatim. The dependency passes at `pkg/frontend/snapshot.go:1466`, `pkg/frontend/pitr.go:1755`, and `pkg/frontend/snapshot_restore_with_ts.go:671` then call bare `parsers.Parse`, and the later `bh.Exec(..., tblInfo.createSql)` also parses with the background/default mode. Therefore a view created under `PIPES_AS_CONCAT` is restored as logical OR, while an `ANSI_QUOTES` view can fail restoration entirely. This is also a regression from the old parser, where bare `||` had legacy concat semantics. Carry the stored mode through restore (or canonicalize the view SQL into mode-independent SQL before storing/executing it) and use it for both dependency analysis and recreation. Please cover snapshot and PITR/TS restore.

2. **High — Sidecar reparses raw user SQL with empty mode.** `pkg/frontend/sidecar_offload.go:243` calls bare `parsers.Parse` even though `rewriteTablesForGPU` already has the session. A query accepted by the main parser under `ANSI_QUOTES` can fail here, and `PIPES_AS_CONCAT` is silently rebound as OR. Use `ParseWithSQLMode` with the same session-mode derivation as the main frontend path. While touching this block, ensure every returned AST is freed on success and error paths.

3. **Medium — SQL stored-procedure bodies also bypass session mode.** `pkg/frontend/authenticate.go:11397` reparses `spBody` with bare `parsers.Parse`. Even in the same session, a procedure body using `||` under `PIPES_AS_CONCAT` is parsed as OR at CALL time, and `ANSI_QUOTES` / `NO_BACKSLASH_ESCAPES` bodies have the analogous problem. At minimum this reparse must receive the applicable session mode; if MatrixOne follows MySQL creation-time routine semantics, that mode must be persisted with the routine.

The staged-execution unit tests currently test helpers and manually mutate the session before parsing the next fragment; they do not exercise the complete `doComQuery` execute-then-reparse loop. Please add an end-to-end frontend/BVT regression for expression-valued and user-variable `SET sql_mode` in one packet, plus the alternate paths above.

Validation performed on head `7f79c06aea`:
- parser package tests: pass
- targeted frontend staged/prepare tests: pass
- targeted plan view-mode tests: pass
- compound ParseFirst benchmark at 10/100/1000 statements: ~7us/~59us/~492us (linear)
- the red basic-benchmark check failed in `Set up job`, not in the benchmark itself.

XuPeng-SH (changes requested):
Blocking correctness issue: stored procedure bodies are parsed with the caller session SQL mode instead of the SQL mode captured when the procedure was created.

`parseStoredProcedureBody` currently calls `sessionSQLModeForParser(ses)`, but `InitProcedure`, `mo_stored_procedure`, and `fetchSqlOfSpFormat` do not persist or fetch a parser SQL mode. Consequently, a procedure created with `PIPES_AS_CONCAT` can be reparsed as boolean OR after the caller clears `sql_mode`; `ANSI_QUOTES` and `NO_BACKSLASH_ESCAPES` can likewise change the AST or make a previously valid body fail to parse. The new test only proves that the caller mode is used, which codifies the wrong lifecycle.

MySQL stores the SQL mode in effect at CREATE/ALTER time and always executes the routine under that stored mode, independent of the caller mode:
https://dev.mysql.com/doc/mysql/8.0/en/create-procedure.html

Please persist the normalized parser mode when creating/replacing a procedure, fetch it with the body, and parse with that stored value. Add a regression test that creates a procedure under one parser mode, changes the session mode, and verifies that CALL retains the creation-time semantics. Legacy rows also need an explicit compatibility fallback; the historical MatrixOne parser treated `||` as concat.

I rechecked the other previously reported paths on the current head. View snapshot/PITR restore now preserves the original SQL and executes it with the persisted view mode; sidecar parsing uses the session mode and frees ASTs. Parser, planner, staged multi-statement execution, and focused frontend tests pass, and the ParseFirst compound benchmark scales approximately linearly. I found no additional blocking correctness, performance, or unhappy-path issue.

pkg/bootstrap/versions/v4_0_1/cluster_upgrade_list.go
```diff
@@ -15,8 +15,31 @@
...
+var clusterUpgEntries = []versions.UpgradeEntry{
+	upg_mo_stored_procedure_add_sql_mode,
+}
+
+var upg_mo_stored_procedure_add_sql_mode = versions.UpgradeEntry{
+	Schema:    catalog.MO_CATALOG,
+	TableName: catalog.MO_STORED_PROCEDURE,
+	UpgType:   versions.ADD_COLUMN,
+	UpgSql: fmt.Sprintf(
+		"alter table %s.%s add column sql_mode varchar(1024) not null default 'PIPES_AS_CONCAT'",
```
XuPeng-SH, line 35: This upgrade hard-codes every pre-existing stored procedure to `PIPES_AS_CONCAT`, but the runtime now reparses procedure bodies with the persisted `sql_mode` unconditionally (`authenticate.go` fetches `sql_mode` and `parseStoredProcedureBody` uses it directly). So a procedure that was actually created before the upgrade under `ANSI_QUOTES`, `NO_BACKSLASH_ESCAPES`, `REAL_AS_FLOAT`, etc. will silently change semantics as soon as the cluster upgrades or restores old `mo_stored_procedure` rows. Please preserve legacy rows as an "unknown"/NULL mode and fall back to legacy behavior when the creation mode was never recorded, instead of fabricating `PIPES_AS_CONCAT`.

XuPeng-SH (changes requested):
Re-reviewed the latest head `7e8e24f5d8`. The new persisted stored-procedure `sql_mode` wiring fixes the fresh create/call path, but one upgrade-compatibility blocker remains. Existing procedures (and restored old snapshots of `mo_stored_procedure`) are backfilled as if they were created under `PIPES_AS_CONCAT`, and `doInterpretCall` now always reparses with the persisted column. That means a legacy procedure that actually relied on `ANSI_QUOTES`, `NO_BACKSLASH_ESCAPES`, etc. can change meaning immediately after upgrade even when the caller session still matches the old behavior. Please preserve legacy rows as "unknown creation mode" and fall back to the legacy/caller-session path when no stored mode exists, plus add an upgrade/restore regression.

XuPeng-SH (changes requested):
Requesting changes after a full re-review of head `7e8e24f5d8`.

First, a correction to my immediately preceding review: the legacy stored-procedure fallback is correct. Before this PR the parser ignored session `sql_mode` and always treated `||` as concat, double quotes as strings, backslashes as escapes, and `REAL` as `DOUBLE`; therefore backfilling old procedures with `PIPES_AS_CONCAT` accurately represents their actual historical parser semantics. Please disregard that earlier blocker.

Two mode-sensitive persisted-SQL paths are still not closed:

1. **High — SQL-language UDF bodies are reparsed with empty mode and have no creation-time mode lifecycle.**

`pkg/sql/plan/base_binder.go:1417-1433` substitutes UDF arguments into the stored body and calls bare `parsers.Parse(...)`. `function.Udf` and the `mo_user_defined_function` fetch path carry no parser mode. Since bare `Parse` now means empty flags, an existing SQL UDF body such as `$1 || $2`, which historically parsed as concat, is rebound as logical OR after this upgrade. A newly created UDF under `PIPES_AS_CONCAT`, `ANSI_QUOTES`, or `NO_BACKSLASH_ESCAPES` also loses its creation semantics at invocation time.

This needs the same first-principles treatment as views and procedures: persist the creation-time parser mode (or persist canonical mode-independent SQL), use it when binding the UDF body, and give legacy rows the historical `PIPES_AS_CONCAT` fallback. Please add create/change-mode/invoke and legacy-upgrade regressions. While touching this path, free every successfully parsed `substmts` slice on both bind success and bind error; the current two branches retain pooled ASTs.

2. **High — stored `CREATE TABLE` text is reparsed mode-blind for partition ALTER.**

Normal table creation stores raw `ctx.GetRootSql()` in `rel_createsql`, while `constructAddedPartitionDefs` at `pkg/sql/plan/build_ddl.go:4647-4652` later reparses `tableDef.Createsql` with bare `parsers.ParseOne(...)`. For example, a partitioned table created with ANSI-quoted identifiers under `ANSI_QUOTES`, or string partition values under `NO_BACKSLASH_ESCAPES`, can fail or be rebound differently when `ALTER TABLE ... ADD PARTITION` reconstructs the original partition definition under empty mode. Historical create text containing `||` is likewise exposed to the concat-to-OR compatibility change.

Please either avoid reparsing raw create text, canonicalize it before persistence, or persist/use its creation-time parser mode. Add an end-to-end regression that creates a mode-sensitive partitioned table, changes the session mode, and then adds a partition. The parsed `originTableStmt` also needs a `Free()` on every exit path.

The previously reported prepared-statement, dynamic same-packet SET, linear ParseFirst, precedence, view/restore, sidecar, and stored-procedure paths are fixed on this head. I found no new wait/goroutine/OOM issue in those changed paths. All current GitHub checks are green, and the focused parser/planner/frontend mode tests pass locally; neither remaining persisted-SQL path is covered by those tests.

pkg/bootstrap/versions/v4_0_1/cluster_upgrade_list.go
```diff
@@ -15,8 +15,50 @@
...
+	upg_mo_stored_procedure_add_sql_mode,
+	upg_mo_user_defined_function_add_sql_mode,
+}
+
+var upg_mo_stored_procedure_add_sql_mode = versions.UpgradeEntry{
+	Schema:    catalog.MO_CATALOG,
+	TableName: catalog.MO_STORED_PROCEDURE,
+	UpgType:   versions.ADD_COLUMN,
+	UpgSql: fmt.Sprintf(
+		"alter table %s.%s add column sql_mode varchar(1024) not null default 'PIPES_AS_CONCAT'",
```
XuPeng-SH, line 35: This is still the compatibility break for stored procedures. Existing `mo_stored_procedure` rows are backfilled to `PIPES_AS_CONCAT`, but the runtime now reparses procedure bodies with the persisted `sql_mode` unconditionally. Before this feature, legacy procedures were interpreted under the caller session's mode at call time. So a pre-upgrade procedure that depended on `ANSI_QUOTES` / `NO_BACKSLASH_ESCAPES` / etc. will silently change semantics immediately after upgrade or restore. Please preserve legacy stored procedures as an unknown mode and let execution fall back to the historical path when no creation mode was ever recorded, instead of fabricating `PIPES_AS_CONCAT`.

XuPeng-SH (changes requested):
Re-reviewed the latest head `0b7e61dbaf366503bbe60d9d48c01120e7f69855`. The new patch moves in a good direction for persisted SQL objects: SQL UDFs now persist their parser mode, and the new focused frontend/plan tests pass locally. However, the stored-procedure upgrade/restore compatibility blocker from the previous round is still open.

`mo_stored_procedure.sql_mode` is still introduced as `NOT NULL DEFAULT 'PIPES_AS_CONCAT'` in both the catalog DDL and the v4.0.1 cluster upgrade, and `doInterpretCall` still reparses stored procedure bodies with the persisted column unconditionally. So every pre-upgrade stored procedure row is still backfilled as if it had been created under `PIPES_AS_CONCAT`, even though historical procedures were parsed at call time under the caller session's `sql_mode`. A legacy procedure that actually relied on `ANSI_QUOTES`, `NO_BACKSLASH_ESCAPES`, etc. can therefore silently change meaning immediately after upgrade or restore.

Please preserve legacy procedure rows as "unknown creation mode" (for example NULL/sentinel), and when that mode is absent fall back to the historical stored-procedure behavior instead of fabricating `PIPES_AS_CONCAT`. Also add an upgrade/restore regression that proves a pre-existing procedure created under a non-default mode keeps that mode-sensitive meaning after upgrade.

pkg/bootstrap/versions/v4_0_1/cluster_upgrade_list.go
LeftHandCold: This is not how legacy stored procedures behaved. Before this PR, `doInterpretCall` reparsed `spBody` with bare `parsers.Parse(...)`; that path never read the caller session's `sql_mode`. The old scanner also had fixed semantics: `||` was always `PIPE_CONCAT`, double quotes were strings, backslashes were escapes, and `REAL` was `DOUBLE`.

Therefore backfilling legacy rows with `PIPES_AS_CONCAT` exactly preserves the historical parser behavior. Using NULL/unknown and falling back to the caller's current mode would introduce a real regression: for example, a legacy body containing `'a' || 'b'` would become logical OR under the current default mode instead of concat.

This matches the correction in review `4701390366`: "the legacy stored-procedure fallback is correct" and the earlier blocker should be disregarded. The focused stored-procedure parser-mode tests also pass on the current head, so no code change is needed here.

XuPeng-SH (approved):
Re-reviewed current head `52cbad9633b2f27e28d6c578e11af9de7211be8d` from first principles and also tested the synthetic merge with current `origin/main` (`82f0cc2fb`). I found no remaining merge blocker.

The previously reported functional-closure gaps are addressed:

- direct, prepared-string, user-variable PREPARE, and staged same-packet `SET sql_mode` parsing all use the applicable session mode; staged parsing executes before lexing the next statement and `ParseFirstWithSQLMode` avoids the former compound-statement retry growth
- `PIPES_AS_CONCAT` has concat precedence, while ANSI_QUOTES, NO_BACKSLASH_ESCAPES, and REAL_AS_FLOAT are carried through the scanner/parser
- views persist creation-time mode and preserve it through normal binding plus snapshot/PITR/TS dependency parsing and recreation without rewriting the original SQL
- sidecar reparsing uses session mode and releases parsed ASTs
- stored procedures and SQL UDFs persist creation-time mode; legacy rows correctly default to `PIPES_AS_CONCAT`, matching the old fixed parser semantics (the later review that re-raised this as a compatibility issue was incorrect and is superseded by this approval)
- partitioned CREATE text is canonicalized into mode-independent SQL before persistence; ADD PARTITION reparsing is deterministic and frees the AST
- the catalog additions are wired through the v4.0.1 version-offset upgrade with idempotent column checks, so already-running clusters receive the new columns

Unhappy-path/resource review did not find a new leak, stuck wait, goroutine, or unbounded-growth issue. Scanner reuse resets mode/error state, parsed ASTs and staged computation wrappers are released on success/error paths, and the upgrade entries are retry-safe.

Validation passed on both the PR head and the current-main merge result:

- focused MySQL parser SQL-mode/precedence/backslash/REAL/ParseFirst tests
- focused planner view/UDF/partition/prepare tests
- focused frontend stored-procedure/view-restore/sidecar/staged-execution/prepare tests
- `./pkg/bootstrap/versions/v4_0_1`

All GitHub checks are green and GitHub reports the PR mergeable. Approving.

heni02 (approved):

Issue #25298, opened 2026-07-01
[Bug]: INSERT string function handles negative length differently from MySQL


Ariznawlll:
### Description

`INSERT(str,pos,len,newstr)` does not match MySQL when `len` is negative.

MySQL treats a negative `len` like an over-length replacement starting at `pos`, while MatrixOne keeps the original suffix.

### Steps to Reproduce

```sql
select insert('Quadratic', 3, -1, 'What') as r;
```

### MySQL Result

```text
+--------+
| r      |
+--------+
| QuWhat |
+--------+
```

### MatrixOne Result

Tested on latest `main` at `200bfa7e0efb`.

```text
+---------------+
| r             |
+---------------+
| QuWhatadratic |
+---------------+
```

### Expected Behavior

MatrixOne should return `QuWhat`, matching MySQL.

### Notes

This was found while exploring MySQL compatibility for string functions. Existing BVT has coverage for normal replacement, zero position, over-length replacement, NULL arguments, and table usage, but does not cover the negative `len` path.

daviszhen:
fixed

Ariznawlll:
Verified on latest main 2a008285e.

Result: fixed.

Validation against MySQL:
```sql
select insert('Quadratic', 3, -1, 'What') as neg_len;
select insert('Quadratic', 3, 0, 'What') as zero_len,
       insert('Quadratic', 3, 100, 'What') as over_len,
       insert('Quadratic', 0, -1, 'What') as zero_pos,
       insert('Quadratic', -1, -1, 'What') as neg_pos;
```

MO now matches MySQL:
- negative len returns `QuWhat`
- len=0 returns `QuWhatadratic`
- over-length replacement returns `QuWhat`
- pos=0 / negative pos return original string `Quadratic`

Regression added in matrixorigin/motr#39, commit 12fa5e5:
- `suites/14_issue_regression/t/issue_25298_insert_negative_length.test`
- `suites/14_issue_regression/r/issue_25298_insert_negative_length.result`

MOTR validation passed:
```
[ISSUE_25298_INSERT_NEGATIVE_LENGTH] success
SUMMARY | total=1 success=1 failed=0 skipped=0 | stmts=2 ok=2 fail=0
```

PR #25442, opened 2026-07-03
fix function insert 

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25298

修改: `select insert('Quadratic', 3, -1, 'What') as r;`
修改负参数问题


XuPeng-SH (changes requested):
Requesting changes.

The normal negative-`len` fix is in the right direction, and the added issue/BVT cases cover the reported behavior. However, the same first-principles rule for `INSERT(str,pos,len,newstr)` is still not closed for an unhappy path in the touched code: very large positive `len` can overflow before it is clamped to the string length.

MySQL documents `INSERT()` as replacing the rest of the string from `pos` when `len` is not within the remaining string length, and returning `NULL` only for `NULL` arguments: https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_insert

In `pkg/sql/plan/function/func_binary.go`, the current code does:

```go
endIdx := len(runes)
if replaceLen > 0 {
    endIdx = posIdx + int(replaceLen)
}
if endIdx > len(runes) {
    endIdx = len(runes)
}
...
result = string(runes[:posIdx]) + newstr + string(runes[endIdx:])
```

For example, with `INSERT('abc', 2, 9223372036854775807, 'X')`, `posIdx` is `1` and `int(replaceLen)` is `math.MaxInt64` on 64-bit builds. `posIdx + int(replaceLen)` overflows to a negative `endIdx`; the `endIdx > len(runes)` clamp is skipped, and `runes[endIdx:]` panics. That violates both the MySQL compatibility rule and the no-panic SQL execution expectation.

Suggested fix: clamp before converting/adding, e.g. compute `remaining := int64(len(runes) - posIdx)` and set `endIdx = len(runes)` when `replaceLen < 0 || replaceLen >= remaining`; only do `posIdx + int(replaceLen)` after proving it is within range. Then add regression coverage for huge positive lengths, including a multibyte string case.

Local checks I ran on this head:
- `gofmt -l` clean
- `git diff --check origin/main...HEAD` passed
- `make err-check` passed
- `go test ./pkg/sql/plan/function -run TestInsert` passed

Those checks pass because this overflow path is currently uncovered.

daviszhen:
> Requesting changes.
> 
> The normal negative-`len` fix is in the right direction, and the added issue/BVT cases cover the reported behavior. However, the same first-principles rule for `INSERT(str,pos,len,newstr)` is still not closed for an unhappy path in the touched code: very large positive `len` can overflow before it is clamped to the string length.
> 
> MySQL documents `INSERT()` as replacing the rest of the string from `pos` when `len` is not within the remaining string length, and returning `NULL` only for `NULL` arguments: https://dev.mysql.com/doc/refman/8.4/en/string-functions.html#function_insert
> 
> In `pkg/sql/plan/function/func_binary.go`, the current code does:
> 
> ```go
> endIdx := len(runes)
> if replaceLen > 0 {
>     endIdx = posIdx + int(replaceLen)
> }
> if endIdx > len(runes) {
>     endIdx = len(runes)
> }
> ...
> result = string(runes[:posIdx]) + newstr + string(runes[endIdx:])
> ```
> 
> For example, with `INSERT('abc', 2, 9223372036854775807, 'X')`, `posIdx` is `1` and `int(replaceLen)` is `math.MaxInt64` on 64-bit builds. `posIdx + int(replaceLen)` overflows to a negative `endIdx`; the `endIdx > len(runes)` clamp is skipped, and `runes[endIdx:]` panics. That violates both the MySQL compatibility rule and the no-panic SQL execution expectation.
> 
> Suggested fix: clamp before converting/adding, e.g. compute `remaining := int64(len(runes) - posIdx)` and set `endIdx = len(runes)` when `replaceLen < 0 || replaceLen >= remaining`; only do `posIdx + int(replaceLen)` after proving it is within range. Then add regression coverage for huge positive lengths, including a multibyte string case.
> 
> Local checks I ran on this head:
> 
> * `gofmt -l` clean
> * `git diff --check origin/main...HEAD` passed
> * `make err-check` passed
> * `go test ./pkg/sql/plan/function -run TestInsert` passed
> 
> Those checks pass because this overflow path is currently uncovered.

已经修改

heni02 (approved):

XuPeng-SH (approved):
LGTM.

This is a focused builtin-function fix and the changed semantics are clear: negative `len` in `INSERT()` now follows the MySQL behavior of replacing through the end of the string, and the large-`len` overflow path is clamped safely.

I don’t see a substantive blocker in the implementation. If anything, the remaining follow-up is just nice-to-have extra coverage like a `math.MinInt64` case or a mixed-row vectorized regression, but I would not block merge on that.

ouyuanning (approved):

PR #25443, opened 2026-07-03
Fix CDC task state for permanent table errors

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25177

This PR updates CDC task handling so table-level permanent errors no longer leave the task appearing healthy/running while checkpoints stop advancing.

  Changes:

  - Mark the CDC task as failed when a permanent table error is detected.
  - Stop active CDC readers and unregister the task from the table detector on permanent table failure.
  - Allow the CDC executor state machine to transition from running/restarting to failed.
  - Escape task state/error SQL fields when updating mo_cdc_task.err_msg.
  - Add regression tests for task failure behavior, runtime failure state transitions, and SQL escaping.

### DIFF
--- a/pkg/cdc/error_handler.go
+++ b/pkg/cdc/error_handler.go
@@ -174,7 +174,7 @@ func BuildErrorMetadata(old *ErrorMetadata, record *ErrorRecord) *ErrorMetadata 
 
 // IsErrorExpired checks if a non-retryable error has expired
 // Deprecated: Error expiration mechanism removed. Non-retryable errors now permanently block
-// until manually cleared via Resume.
+// until manually cleared via Restart/Resume.
 func IsErrorExpired(meta *ErrorMetadata) bool {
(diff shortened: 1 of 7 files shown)


XuPeng-SH (approved):

- restart from Failed/Paused now clears persisted table errors before recovery;
- failed-state metrics are decremented when leaving `StateFailed`;
- the stale callback / restart generation race is fenced with `callbackGeneration` plus `callbackMu`, and there are focused tests for both the in-flight and stale-generation cases.

At this point I don’t see a remaining substantive correctness issue. The only thing left would be extra nice-to-have coverage around a more tightly orchestrated combined concurrency scenario, but that is not a blocker.

LGTM.

Issue #25366, opened 2026-07-02
[Bug]: LOAD DATA non-strict invalid conversions abort instead of loading adjusted rows


Ariznawlll:
### MatrixOne version

latest `main` tested locally: `0e8602b78`

### MySQL version

MySQL 8.0.45

### Steps to Reproduce

Create a local CSV file:

```text
1,12,12.345,20240102,abcd
2,abc,10.00x,2024-02-31,abcdef
3,,bad,0000-00-00,xy
```

Then run:

```sql
drop database if exists mysql_compat_model1_gap;
create database mysql_compat_model1_gap;
use mysql_compat_model1_gap;

set session sql_mode = '';
create table t_load(id int primary key, i int, d decimal(5,2), dt date, vc varchar(4));
load data local infile '/path/to/file.csv'
into table t_load
fields terminated by ',' lines terminated by '\n';

select id, i, d, dt, vc, hex(vc) from t_load order by id;
```

### Expected Behavior

MySQL loads all rows and adjusts invalid / truncated values with warnings:

```text
1  12  12.35  2024-01-02  abcd  61626364
2  0   10.00  0000-00-00  abcd  61626364
3  0   0.00   0000-00-00  xy    7879
```

### Actual Behavior

MatrixOne aborts the load and no rows are inserted:

```text
ERROR 20101: internal error: the input value 'abc' is not int32 type for column 1
```

### Notes

I searched existing issues before filing. The close matches were parquet / JSONLine load issues or older loading null-handling items, not this MySQL CSV `LOAD DATA LOCAL INFILE` non-strict conversion behavior.

LeftHandCold:
Checked this while preparing the focused fix for #25352.

I am not including #25366 in that PR because `LOAD DATA` invalid-value handling is a broad data-ingest semantic change, not a small compatibility bug fix. The reported cases overlap the same invalid numeric/date/decimal adjusted-value behavior discussed in #25364 and #25365.

Changing this for `LOAD DATA` would need to define exactly which conversion failures become warnings, which rows are accepted, what adjusted values are written, how strict/non-strict modes interact with `LOCAL` and non-`LOCAL` load paths, and how warning counts/errors are reported. Those choices would affect shared load/insert assignment conversion paths and could change existing MO behavior for users relying on current rejection semantics.

[...]

Issue #25365, opened 2026-07-02
[Bug]: UPDATE IGNORE invalid conversions error instead of updating adjusted values


Ariznawlll:
### MatrixOne version

latest `main` tested locally: `0e8602b78`

### MySQL version

MySQL 8.0.45

### Steps to Reproduce

```sql
drop database if exists mysql_compat_model1_gap;
create database mysql_compat_model1_gap;
use mysql_compat_model1_gap;

set session sql_mode = 'STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE';
create table t_update_ignore(id int primary key, i int, d decimal(5,2), dt date, vc varchar(4));
insert into t_update_ignore values (1,1,1.23,'2024-01-01','ok');

update ignore t_update_ignore
set i='abc', d='99999.99', dt='2024-02-31', vc='abcdef'
where id=1;

select id, i, d, dt, vc, hex(vc) from t_update_ignore;
```

### Expected Behavior

MySQL downgrades the invalid assignments to warnings and updates the row with adjusted values:

```text
id  i  d       dt          vc    hex(vc)
1   0  999.99  0000-00-00  abcd  61626364
```

### Actual Behavior

MatrixOne errors and leaves the original row unchanged:

```text
ERROR 20203: invalid argument cast to int, bad value abc

id  i  d     dt          vc  hex(vc)
1   1  1.23  2024-01-01  ok  6F6B
```

### Notes

This is not the same as #25347. #25347 tracks `UPDATE IGNORE` duplicate-key conflict handling; this issue tracks invalid value conversion / adjustment semantics in `UPDATE IGNORE`.

LeftHandCold:
Checked this while preparing the focused fix for #25352.

I am not including #25365 in that PR because `UPDATE IGNORE` is not a low-risk compatibility tweak in the current code path. A correct implementation first needs statement-level `IGNORE` semantics to be preserved through parsing/planning/execution, then each affected error class must be deliberately downgraded or adjusted.

[...]

Issue #25364, opened 2026-07-02
[Bug]: non-strict INSERT and INSERT IGNORE invalid conversions error instead of writing adjusted values


Ariznawlll:
### MatrixOne version

latest `main` tested locally: `0e8602b78`

### MySQL version

MySQL 8.0.45

### Steps to Reproduce

```sql
drop database if exists mysql_compat_model1_gap;
create database mysql_compat_model1_gap;
use mysql_compat_model1_gap;

set session sql_mode = '';
create table t_nonstrict(id int primary key, i int, d decimal(5,2), dt date, vc varchar(4));
insert into t_nonstrict values (1,'abc','99999.99','2024-02-31','abcdef');
select id, i, d, dt, vc, hex(vc) from t_nonstrict;

set session sql_mode = 'STRICT_TRANS_TABLES,NO_ZERO_DATE,NO_ZERO_IN_DATE';
create table t_ignore(id int primary key, i int, d decimal(5,2), dt date, vc varchar(4), unique key uk_i(i));
insert ignore into t_ignore values (1,'abc','99999.99','2024-02-31','abcdef');
select id, i, d, dt, vc, hex(vc) from t_ignore;
```

### Expected Behavior

MySQL adjusts invalid / out-of-range values and writes the row with warnings:

```text
id  i  d       dt          vc    hex(vc)
1   0  999.99  0000-00-00  abcd  61626364
```

`INSERT IGNORE` under strict mode also downgrades the conversion errors to warnings and writes the adjusted row.

### Actual Behavior

MatrixOne errors and writes no row.

For the non-strict insert and the `INSERT IGNORE` case, MO returned:

```text
ERROR 20301: invalid input: 99999.99 beyond the range, can't be converted to Decimal64(5,2).
```

### Notes

I checked for existing issues before filing. #25200 covers string-width/sql_mode behavior, but this case also covers integer invalid-string coercion, decimal saturation, invalid date adjustment, and `INSERT IGNORE` downgrade as one DML assignment path.

LeftHandCold:
Checked this while preparing the focused fix for #25352.

I am not including #25364 in that PR because this issue is not a small isolated planner bug. It mixes several assignment-conversion behaviors, including non-strict `INSERT`, `INSERT IGNORE`, string truncation, invalid numeric strings, decimal saturation, and invalid date handling.

The low-risk part should be split and handled separately: the string-width / `sql_mode` truncation behavior overlaps the existing compatibility track in #25200.

[...]

PR #25445, opened 2026-07-03
fix: grouped correlated column binding

### LINKED ISSUE #25352: [Bug]: GROUP BY query rejects correlated subquery referencing grouped column

Existing issue #25327 covers `ONLY_FULL_GROUP_BY` single-value and functional-dependency cases. This issue is different: the outer column referenced by the correlated subquery is itself present in the `GROUP BY` list, but MatrixOne still rejects it.

### Branch Name

main

### Commit ID

`0e8602b78`

- MySQL oracle: MySQL 8.0.45

### Actual Behavior

MatrixOne rejects a grouped query when a scalar correlated subquery references the grouped outer column:

```text
ERROR 1149 (HY000): SQL syntax error: column "t.status" must appear in the GROUP BY clause or be used in an aggregate function
```

But `t.status` is already in the outer `GROUP BY` list.

### Expected Behavior

MySQL accepts the query and evaluates the correlated subquery per grouped row.

Expected result:

```text
status    c
active    2
archived  1
pending   1
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS mysql_compat_group_subquery_min;
CREATE DATABASE mysql_compat_group_subquery_min;
USE mysql_compat_group_subquery_min;

CREATE TABLE t (
  id int primary key,
  status varchar(16),
  payload int,
  index idx_status(status,id)
);

INSERT INTO t VALUES
  (1,'active',10),
  (2,'active',20),
  (3,'archived',30),
  (4,'pending',40);

SELECT status,
       (SELECT COUNT(*) FROM t t2 WHERE t2.status = t.status) AS c
FROM t
GROUP BY status
ORDER BY status;
```

### Additional information

MySQL 8.0.45 result:

```text
status    c
active    2
archived  1
pending   1
```

MatrixOne latest `main` at `0e8602b78` rejects the query with:

```text
ERROR 1149 (HY000): SQL syntax error: column "t.status" must appear in the GROUP BY clause or be used in an aggregate function
```

This was found while backfilling aggregate / GROUP BY compatibility coverage with indexed tables.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25352

This fixes `ONLY_FULL_GROUP_BY` planning for correlated subqueries that reference an outer query column already present in the outer `GROUP BY`.

- Correlated references to grouped outer columns now bind to the aggregate/group output instead of the pre-aggregation table scan column.
- Correlated references to non-grouped outer columns are still rejected under `ONLY_FULL_GROUP_BY`.
- Added planner regression tests for both the allowed grouped-column case and the rejected non-grouped-column case.

This PR intentionally does not change the broader invalid-value coercion / `IGNORE` compatibility behavior discussed in #25364, #25365, and #25366.

## Root Cause

When binding a correlated outer column reference from inside a subquery, the planner reused the original table-scan column binding. After the outer query had been grouped, that raw column was no longer available unless it was also part of the aggregate/group output, so valid references to grouped columns could be rejected as non-aggregated columns.

## Checks

- `git diff --check -- pkg/sql/plan/base_binder.go pkg/sql/plan/build_test.go`
- `GOCACHE=/private/tmp/mo-go-build-cache CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" CGO_LDFLAGS="-L$(pwd)/thirdparties/install/lib -lusearch_c" DYLD_LIBRARY_PATH="$(pwd)/cgo:$(pwd)/thirdparties/install/lib" go test -mod=mod -ldflags="-extldflags '-L$(pwd)/cgo -lmo -L$(pwd)/thirdparties/install/lib -Wl,-rpath,@executable_path/lib'" -count=1 ./pkg/sql/plan`


ouyuanning (approved):

XuPeng-SH (changes requested):
Requesting changes because the fix is not functionally closed across all correlated-subquery entry points.

Finding:
`HAVING` still rejects the same valid grouped outer reference that this PR fixes for the SELECT list. A correlated reference to an outer column that is already present in the outer `GROUP BY` should bind to the outer group output regardless of whether the subquery appears in SELECT, HAVING, or ORDER BY.

Repro on this PR head (`c7578373a380c122b7f26b12f55e50e8bbce2ead`):

```sql
SELECT n_name
FROM nation
GROUP BY n_name
HAVING (SELECT COUNT(*) FROM nation2 n2 WHERE n2.n_name = nation.n_name) > 0;
```

Current result from local planner test:

```text
SQL syntax error: column "nation.n_name" must appear in the GROUP BY clause or be used in an aggregate function
```

Root cause: the new grouped-correlated lookup is in `baseBindColRef`, but `HavingBinder.BindColRef` has its own `ONLY_FULL_GROUP_BY` branch and can return the group-by error before the grouped correlated ref is accepted. This leaves the original rule only partially fixed.

Suggested direction:
- Centralize the “correlated outer ref is allowed iff it resolves to an outer group output” decision so all group-aware binders share it, or update `HavingBinder.BindColRef` to call the base binding path and allow `Expr_Corr` whose `RelPos == ctx.groupTag` while still rejecting ungrouped base-table `Expr_Corr` under `ONLY_FULL_GROUP_BY`.
- Add regression coverage for SELECT-list, HAVING, ORDER BY, table alias, GROUP BY ordinal/alias, and the ungrouped-column rejection case.

Local checks:
- `go test ./pkg/sql/plan -run 'TestOnlyFullGroupBy(Allows|Rejects)CorrelatedSubqueryOn.*Column'` passed.
- Extra temporary closure test passed for table alias, GROUP BY ordinal/alias, ORDER BY, and ungrouped rejection; failed only for the HAVING case above.

heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes on the current head (`94315695332bd3a9a8a48280e4d7b273f91abdb3`). The previous HAVING gap is fixed, but the current implementation still has one semantic regression around correlated aggregates.

Finding:
`HavingBinder.BindColRef` now returns the ONLY_FULL_GROUP_BY error for a correlated column inside an aggregate when that correlated ref targets the outer group output. That changes a SELECT-list correlated aggregate with an already-grouped outer column from the existing `correlated columns in aggregate function` NYI into a false GROUP BY syntax error.

Repro on this PR head:

```sql
SELECT n_name,
       (SELECT COUNT(nation.n_name) FROM nation2) AS c
FROM nation
GROUP BY n_name;
```

Current PR result:

```text
SQL syntax error: column "nation.n_name" must appear in the GROUP BY clause or be used in an aggregate function
```

On `origin/main`, the SELECT-list variant returns the existing unsupported-path error:

```text
correlated columns in aggregate function is not yet implemented
```

Why this matters: `nation.n_name` is already grouped in the outer query, so this is not an ONLY_FULL_GROUP_BY violation. It is either a correlated-aggregate feature to implement deliberately, or it should keep the current NYI behavior. The new `TestOnlyFullGroupByPreservesCorrelatedAggregateNYI` covers ungrouped/no-group variants, but it misses the grouped variant. The third case in `TestOnlyFullGroupByRejectsCorrelatedSubqueryOnUngroupedColumn` also appears to lock in the wrong rule: `SUM(nation.n_regionkey)` uses an outer column that is grouped by the outer query, so expecting a group-by error there is not first-principles correct.

Suggested direction:
- Keep aggregate-internal correlated refs under one rule for now: `Expr_Corr` inside aggregate should stay `NYI` until correlated aggregate semantics are intentionally implemented.
- Add regressions for grouped outer refs inside subquery aggregates in both SELECT-list and HAVING paths.
- Keep the already-added positive coverage for simple grouped correlated refs in SELECT/HAVING/ORDER BY.

Local checks performed:
- `git diff --check -- pkg/sql/plan/base_binder.go pkg/sql/plan/having_binder.go pkg/sql/plan/build_test.go test/distributed/cases/subquery/subquery-with-any.result`
- `go test ./pkg/sql/plan -run 'TestOnlyFullGroupBy(Allows|Rejects|Preserves)'`
- full `go test ./pkg/sql/plan`
- temporary closure tests for expression GROUP BY rejection, alias/base-table scoping, and the grouped correlated aggregate regression above

aunjgr (approved):
Clean fix. Redirecting correlated outer column references bound to GROUP BY columns from the raw table-scan column to the aggregate group output is the correct approach. Test coverage is thorough across grouped-allowed, non-grouped-rejected, and correlated-aggregate-NYI categories.

XuPeng-SH (approved):
Re-reviewed latest head `a667d7a` from first-principles / functional-closure / unhappy-path angles.

The previous blockers are closed:
- HAVING now accepts correlated references to outer columns that resolve to the outer GROUP BY output, matching the SELECT-list and ORDER BY behavior.
- Correlated refs inside aggregate arguments still take the existing `correlated columns in aggregate function` NYI path instead of being misreported as ONLY_FULL_GROUP_BY syntax errors.
- Ungrouped outer refs are still rejected with the GROUP BY error.
- Coverage now includes SELECT-list, HAVING, ORDER BY, table alias, GROUP BY ordinal/alias, grouped correlated aggregate NYI, and existing BVT result updates for ANY/EXISTS.

Local checks:
- `git diff --check origin/main...HEAD -- pkg/sql/plan/base_binder.go pkg/sql/plan/having_binder.go pkg/sql/plan/build_test.go test/distributed/cases/subquery/subquery-with-any.result test/distributed/cases/subquery/subquery-with-exists.result`
- `go test ./pkg/sql/plan -run 'TestOnlyFullGroupBy(Allows|Rejects|Preserves)Correlated' -count=1`

CI Summary is green. Non-blocking: PR is still marked `BEHIND`, so update with latest `main` if the merge queue requires it.

Issue #25279, opened 2026-06-30
[Bug]: lockservice validTxn timeout treated as retry error, preventing lock release indefinitely


matrix-meow:
## Is there an existing issue for the same bug?
- [x] I have checked the existing issues.

## Branch Name
main

## Commit ID
2add0b9fc

## Other Environment Information
- OS type: Linux
- Others: Related to #25278

## Actual Behavior

When `validTxn` sends a `GetActiveTxn` RPC to a remote CN and the RPC times out (`context deadline exceeded`), the error is classified as a "retry error" by `isRetryError()`. The `isValidRemoteTxn` function then conservatively assumes the transaction **might still be valid** and returns `true`, preventing lock release.

```go
// pkg/lockservice/service.go (isValidRemoteTxn)
valid, err := h.validTxn(txn)
if err == nil {
    return valid
}
logValidTxnFailed(h.logger, txn, err)
if isRetryError(err) {
    return true  // ← timeout = "maybe alive" = don't release locks
}
return !h.canUnlockRemoteTxn(txn)
```

The `validTxn` callback uses a 10s timeout:
```go
// pkg/lockservice/service_remote.go:109
ctx, cancel := context.WithTimeoutCause(
    context.Background(), defaultRPCTimeout,  // 10s
    moerr.CauseInitRemote2,
)
resp, err := s.remote.client.Send(ctx, req)
```

When the remote CN's lockservice is slow to respond (e.g., due to DN commit slowness — see #25274), the `GetActiveTxn` RPC consistently times out. Since `context deadline exceeded` is classified as a retry error, locks are **never released**, creating a permanent deadlock.

### Observed incident

On CN p6bxw, `GetActiveTxn` RPCs timed out for 2+ hours because p6bxw's lockservice handlers were blocked (goroutine profile confirmed 4 goroutines stuck in `handleGetActiveTxn → IterTxns → RLock` blocked). Other CNs kept retrying `validTxn` every few minutes, always getting timeouts, never releasing locks.

```
07:02:23  "failed to valid txn"  error="context deadline exceeded\ninternal error: initRemote 2"
07:04:53  "failed to valid txn"  error="context deadline exceeded\ninternal error: initRemote 2"
07:09:53  "failed to valid txn"  error="context deadline exceeded\ninternal error: initRemote 2"
... (repeated every 2-5 min for 2+ hours, same txn)
```

## Expected Behavior

[...]

PR #25447, opened 2026-07-03
fix(lockservice): break owner-local remote lock deadlocks

### LINKED ISSUE #25278: [Bug]: leak checker detects leaked transactions but does not clean them up, causing indefinite lock holding

## Branch Name
main

## Commit ID
2add0b9fc

- OS type: Linux
- Others: freetier-01 cluster (cn-dev), observed on CN pod p6bxw

## Actual Behavior

When the CN leak checker detects a leaked transaction, it only logs the error and increments a counter. It does NOT clean up the transaction — no rollback, no lock release. The transaction and its held locks persist indefinitely until the frontend session disconnects.

```go
// pkg/cnservice/server.go:770
} else {
    v2.TxnLeakCounter.Inc()
    runtime.DefaultRuntime().Logger().Error("found leak txn", fields...)
    // No rollback, no unlock — just logging
}
```

### Observed incident

A leaked transaction `d5d476fa3fed8bdc18bc8881cac4238b` was created at **2026-06-28 15:09:12** on CN p6bxw (AccountID: 690440010). The leak checker detected and logged it repeatedly, but never cleaned it up. The transaction held locks for **15+ hours**, blocking other CNs' transactions that needed those locks.

This triggered a cascade:
1. Other CNs' waiters accumulated (4166 goroutines in `waiter.wait`)
2. `checkOrphan` → `validTxn` → `GetActiveTxn` RPC to p6bxw timed out
3. `failed to valid txn` errors every 2-5 minutes for 2+ hours
4. Eventually: HandShake failures, connection rejections, `found leak txn` on other CNs

## Expected Behavior

When the leak checker identifies a transaction as leaked (no active SQL, not committing, not rolling back), it should:

1. **Force rollback** the transaction via the txn client
2. **Release held locks** via the lockservice
3. **Kill the associated session** to prevent further issues

The current diagnostic-only approach allows a single leaked transaction to cascade into cluster-wide failures.

## Steps to Reproduce

1. Create a transaction on a CN that holds locks on shared keys.
2. Make the session idle (no commit/rollback) — e.g., client disconnects without cleanup, or session gets stuck.
3. Wait for the leak checker to detect it (configured interval).
4. Observe: leak is logged but transaction persists, locks remain held.
5. Other CNs trying to acquire the same locks will wait indefinitely.

### Related code
- `pkg/cnservice/server.go:740-780` — leak checker detection (diagnostic only)
- `pkg/txn/client/` — txn lifecycle management
- `pkg/frontend/session.go` — session disconnect cleanup

### Log evidence
```
{"level":"ERROR","time":"2026/06/29 06:49:26","msg":"found leak txn",
 "txn-id":"d5d476fa3fed8bdc18bc8881cac4238b",
 "create-at":"2026/06/28 15:09:12"}  ← 15h40m old, still not cleaned
```

### LINKED ISSUE #25297: [CDC] stale running tasks, watermark update timeouts, and drop cdc fails with invalid connection
## Summary

In a private MatrixOne deployment, multiple CDC tasks appear `running` in `mo_cdc task show`, but their checkpoints are stale or failed. The CDC watermark updater continuously fails with `context deadline exceeded`, and `drop cdc task` / `drop cdc all` fail with `invalid connection`, leaving old paused tasks in CDC metadata and watermark updates.

This looks like a CDC control-plane / metadata / watermark consistency problem rather than a simple client network issue:

- Normal MySQL queries against the same cluster return successfully in sub-second time.
- `mo_cdc task show` still prints repeated `[mysql] ... unexpected EOF`.
- CN logs continue to report `cdc.watermark.read_error`.
- Old paused tasks still appear in watermark update SQL.
- Dropping those tasks fails with `invalid connection`.

## Environment

- MatrixOne version: `8.0.30-MatrixOne-v3.0.0-5e4210029-hotfix-20260603`
- Deployment: private deployment, same MatrixOne cluster, cross-account CDC.
- Client: `mo_cdc` binary from local `mo-backup`; this binary does not expose `--version` or `version`.
- CDC uses table-level PITR. Existing PITR:
  - `ods_s4.t_s4_performance`: 24h
  - `ods_s4.performance_file_attach`: 24h
  - `ods_ecc.performance_file_attach`: 24h

### 1. FIN -> BID, `ods_s4.t_s4_performance`

Original task:

- Task name: `private_fin_ods_s4_to_bid_t_s4_performance_cdc`
- Task ID: `019ed526-3bca-7555-b23c-f95801fa85e3`
- Current state: `paused`
- Current checkpoint shown by `mo_cdc task show`:

```text
ods_s4.t_s4_performance: 2026-06-29 17:42:16.080563741 +0000 UTC
```

Rebuild full task:

- Task name: `private_fin_ods_s4_to_bid_t_s4_performance_cdc_full_20260701`
- Task ID: `019f1bb8-537d-79a6-ad6b-0880137f36f5`
- Current state: `paused`
- Checkpoint stayed at:

```text
ods_s4.t_s4_performance: 1970-01-01 00:00:00 +0000 UTC
```

Current incremental no-full task:

- Task name: `private_fin_ods_s4_to_bid_t_s4_performance_cdc_incr_20260701`
- Task ID: `019f1bbd-5add-705b-b7f7-044e90e50023`
- Current state: `running`
- Checkpoint shown by `mo_cdc task show`:

```text
ods_s4.t_s4_performance: 2026-07-01 03:33:48.007446058 +0000 UTC
```

Current data count is correct:

```text
source fin.ods_s4.t_s4_performance rows: 162442
target bid.ods_s4.t_s4_performance rows: 162442
target distinct (VBELN, POSNR): 162442
target duplicate groups by (VBELN, POSNR): 0
```

However, the watermark updater still reports errors for both the old paused task and the new incremental task.

### 2. BID -> FIN, S4 attachment table

- Task name: `private_bid_to_fin_s4_attach_cdc`
- Task ID: `019f115c-23d9-78fa-a2cd-429a2c1b35bc`
- Source: `ods_s4.performance_file_attach`
- Target: `jst_receipts_tables.performance_file_attach__s4_cdc`
- Current state: `running`
- Current checkpoint shown by `mo_cdc task show`:

```text
ods_s4.performance_file_attach:
2026-06-29 17:40:32.973762825 +0000 UTC
(Failed, error: N:1782766991:internal error: internal error: no watermark found)
```

Current data count is not aligned:

```text
source bid.ods_s4.performance_file_attach rows: 54581
target fin.jst_receipts_tables.performance_file_attach__s4_cdc rows: 54526
diff: 55 rows missing on target
```

### 3. BID -> FIN, ECC attachment table

- Task name: `private_bid_to_fin_ecc_attach_cdc`
- Task ID: `019f115c-2b93-7d68-81a4-fe6c77f9425a`
- Source: `ods_ecc.performance_file_attach`
- Target: `jst_receipts_tables.performance_file_attach_ecc_cdc`
- Current state: `running`
- Current checkpoint shown by `mo_cdc task show`:

```text
ods_ecc.performance_file_attach: 2026-06-29 17:40:33.561126733 +0000 UTC
```

Current data count is aligned:

```text
source bid.ods_ecc.performance_file_attach rows: 22451
target fin.jst_receipts_tables.performance_file_attach_ecc_cdc rows: 22451
```

But the checkpoint is stale and CN logs show repeated watermark update failures for this task.

### 2026-07-01: checked current CDC state

Command pattern:

```bash
mo_cdc task show --source-uri "<redacted-fin-source-uri>" --all
mo_cdc task show --source-uri "<redacted-bid-source-uri>" --all
```

Both commands returned JSON, but also printed repeated client errors:

```text
[mysql] 2026/07/01 13:31:43 connection.go:49: unexpected EOF
[mysql] 2026/07/01 13:31:44 connection.go:49: unexpected EOF
```

### 2026-07-01: verified normal MySQL queries work

The following checks succeeded quickly against the same cluster:

```sql
SELECT NOW(), CONNECTION_ID(), COUNT(*) FROM ods_s4.t_s4_performance;
SELECT NOW(), CONNECTION_ID(), COUNT(*) FROM ods_s4.performance_file_attach;
```

Results:

```text
FIN t_s4 rows: 162442
BID S4 attachment rows: 54581
```

So the issue is not simply that the client cannot connect to MatrixOne.

### 2026-07-01: tried to clean old CDC tasks

The following operations failed:

```bash
mo_cdc task drop --source-uri "<redacted-fin-source-uri>" \
  --task-name "private_fin_ods_s4_to_bid_t_s4_performance_cdc"

mo_cdc task drop --source-uri "<redacted-fin-source-uri>" \
  --task-name "private_fin_ods_s4_to_bid_t_s4_performance_cdc_full_20260701"

drop cdc all
```

CN logs:

```text
2026/07/01 03:27:22 UTC
caller=frontend/cdc_handle.go:339
msg=cdc.on_pre_update.delete_task.failed
task-name=private_fin_ods_s4_to_bid_t_s4_performance_cdc
error=invalid connection

2026/07/01 03:27:22 UTC
caller=frontend/cdc_handle.go:267
msg=cdc.do_update_task.failed
task-name=private_fin_ods_s4_to_bid_t_s4_performance_cdc
target-status=CancelRequested
affected-count=0
error=invalid connection

2026/07/01 03:27:22 UTC
statement="drop cdc task private_fin_ods_s4_to_bid_t_s4_performance_cdc"
status=fail
error=invalid connection
```

For the full rebuild task:

```text
2026/07/01 03:33:39 UTC
caller=frontend/cdc_handle.go:355
msg=cdc.on_pre_update.delete_watermark.failed
task-name=private_fin_ods_s4_to_bid_t_s4_performance_cdc_full_20260701
key-count=1
error=invalid connection

2026/07/01 03:33:39 UTC
caller=frontend/cdc_handle.go:267
msg=cdc.do_update_task.failed
task-name=private_fin_ods_s4_to_bid_t_s4_performance_cdc_full_20260701
target-status=CancelRequested
affected-count=0
error=invalid connection
```

For `drop cdc all`:

```text
2026/07/01 03:27:45 UTC
statement="drop cdc all"
status=fail
error=invalid connection
```

### 2026-07-01: watermark updater still failing after recheck

Recent CN logs from `2026-07-01 05:28-05:32 UTC`:

```text
caller=cdc/watermark_updater.go:427
msg=cdc.watermark.read_error
error="context deadline exceeded\ninternal error: watermark update"
err-msg="commit sql \"INSERT INTO `mo_catalog`.`mo_cdc_watermark`
  (account_id, task_id, db_name, table_name, watermark)
VALUES
  (10001, '019f1bbd-5add-705b-b7f7-044e90e50023', 'ods_s4', 't_s4_performance', ...),
  (20002, '019f115c-2b93-7d68-81a4-fe6c77f9425a', 'ods_ecc', 'performance_file_attach', ...),
  (10001, '019ed526-3bca-7555-b23c-f95801fa85e3', 'ods_s4', 't_s4_performance', ...)
ON DUPLICATE KEY UPDATE watermark = VALUES(watermark)\" failed"
```

The old paused task `019ed526-3bca-7555-b23c-f95801fa85e3` is still included in watermark update SQL together with the new incremental task `019f1bbd-5add-705b-b7f7-044e90e50023`.

Other repeated logs:

```text
caller=cdc/watermark_updater.go:427
msg=cdc.watermark.read_error
error="context deadline exceeded\ninternal error: watermark update err msg"
err-msg="update err_msg sql \"INSERT INTO `mo_catalog`.`mo_cdc_watermark`
  (account_id, task_id, db_name, table_name, err_msg)
VALUES
  (10001, '019ed526-3bca-7555-b23c-f95801fa85e3', 'ods_s4', 't_s4_performance', ''),
  (10001, '019f1bbd-5add-705b-b7f7-044e90e50023', 'ods_s4', 't_s4_performance', '')
ON DUPLICATE KEY UPDATE err_msg = VALUES(err_msg)\" failed"
```

And for the ECC attachment task:

```text
caller=cdc/watermark_updater.go:427
msg=cdc.watermark.read_error
error="context deadline exceeded\ninternal error: watermark update err msg"
err-msg="update err_msg sql \"INSERT INTO `mo_catalog`.`mo_cdc_watermark`
  (account_id, task_id, db_name, table_name, err_msg)
VALUES
  (20002, '019f115c-2b93-7d68-81a4-fe6c77f9425a', 'ods_ecc', 'performance_file_attach', '')
ON DUPLICATE KEY UPDATE err_msg = VALUES(err_msg)\" failed"
```

## Expected Behavior

1. `drop cdc task` should reliably remove task metadata and watermark metadata, or return an actionable error without leaving the task in a half-deleted state.
2. Paused or cancelled tasks should not continue to participate in active watermark updater batches.
3. A failure in one task's watermark update should not block watermark progress for unrelated running tasks.
4. `mo_cdc task show` should not report `running` without surfacing stale checkpoint / failed watermark health clearly.
5. A running task should not remain indefinitely at a failed checkpoint such as `no watermark found`.

## Actual Behavior

1. `drop cdc task` and `drop cdc all` fail with `invalid connection`.
2. Old paused tasks still appear in watermark update SQL.
3. Watermark updater repeatedly fails with `context deadline exceeded`.
4. `mo_cdc task show` reports tasks as `running`, while checkpoints are stale or failed.
5. One S4 attachment CDC task is already data-inconsistent: target is missing 55 rows.

## Questions / Help Needed

1. What is the supported way to safely clean up CDC metadata / watermark rows / runner state when `drop cdc task` fails with `invalid connection`?
2. Why can paused tasks remain in watermark update SQL?
3. Should the watermark updater batch multiple tasks together in one SQL statement? If yes, how is task-level isolation guaranteed when one task/connection times out?
4. Why does `mo_cdc task show` keep returning `running` when checkpoint is stale or has `no watermark found`?
5. Is the repeated `unexpected EOF` from `mo_cdc task show` expected/retryable, or does it indicate a CDC client/server protocol problem?

## Impact

This blocks reliable private-deployment CDC usage. We can manually repair table data by full rebuild, but the CDC control plane remains unhealthy:

- stale paused tasks cannot be deleted;
- watermark updates keep timing out;
- at least one running CDC task has already diverged by 55 rows;
- users cannot tell from `state=running` whether CDC is actually healthy.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25297
Fixes #25278
Related to #25279

This PR fixes a lockservice hang where a remote lock request can remain stuck at the lock-table owner while the origin transaction stays active and continues holding already-granted locks. In that state, existing deadlock detection may fail to break an owner-local wait cycle, and the origin-side remote lock RPC can wait for an extremely long time.

Changes:

- Adds an owner-local deadlock fast path for remote owner-side row/exclusive waits. The owner now maintains bounded wait edges and walks only the graph rooted at the actual conflicting holders instead of scanning the whole lock table.
- Adds an owner-side remote lock wait timeout with a dedicated `ErrRemoteLockWaitTimeout`, so the origin transaction is driven to rollback instead of waiting indefinitely.
- Keeps remote lock metadata on owner-timeout so rollback can safely send remote unlock for partially granted remote lock batches.
- Fixes deadlock detector busy enqueue bookkeeping and adds focused observability for deadlock enqueue, owner-local deadlock, and remote owner timeout paths.

Validation:

```text
go test -v ./pkg/lockservice -count=1 -timeout=20m
PASS
ok  github.com/matrixorigin/matrixone/pkg/lockservice  580.828s
```


XuPeng-SH (changes requested):
I found multiple substantive blockers here.

1. The new owner-side remote wait cap changes user-visible `lock_wait_timeout` semantics.

`RemoteLockOwnerWaitTimeout` defaults to 2 minutes and is now always injected into remote lock requests. The effective async wait deadline then becomes the earlier of the session’s `LockWaitTimeout` and this owner-side cap. So a caller that explicitly configured a larger lock wait can now be aborted early at the owner’s 2-minute cap with `ErrRemoteLockWaitTimeout`.

That is a behavior change for valid workloads, not just protection against indefinite waits.

Suggested fix: only apply the owner-side cap when the client did not set `LockWaitTimeout`, or make the owner cap opt-in / disabled by default. At minimum, do not shorten an explicit caller timeout.

2. `ErrRemoteLockWaitTimeout` still records a lock that was never granted.

On the owner-timeout path, `remoteLockTable.lock()` still executes `txn.lockAdded(...)` before returning `ErrRemoteLockWaitTimeout`. But that error means the owner timed out the waiter before grant, so the origin txn did not acquire the lock.

This leaves the origin-side held-lock set inconsistent with reality and can poison later deadlock/observability paths that walk held locks.

Suggested fix: do not call `txn.lockAdded` for `ErrRemoteLockWaitTimeout` (or remove the key before returning), and add a regression that this error leaves no new held lock on the origin txn.

3. The owner-local fast path is applied too broadly and does O(total locks/waiters) work under the table mutex on every row/exclusive conflict.

`handleLockConflictLocked()` now unconditionally calls `detectOwnerLocalDeadlockLocked()` for qualifying row/exclusive conflicts, and that helper scans the entire local lock store and all waiters while `localLockTable.mu` is held.

That is a real scalability regression risk: a conflict on one row now becomes table-wide work inside the critical section, blocking unrelated lock/unlock traffic behind it.

Suggested fix: restrict this fast path to the remote-owner cases that motivated the PR, or replace the full-store scan with a bounded graph walk rooted at the actual conflicting holders/waiters.

aptend:
@XuPeng-SH thanks for the review. Responses inline.

1. I don't think the owner-side cap should be tied to user `lock_wait_timeout`, and I do not think the suggested behavior is the right fix for this PR.

The production failure this PR is targeting is not an ordinary user lock-wait timeout case. It is a remote lock RPC that has reached the lock-table owner and can be stuck inside the owner-side request handling after partially progressing the lock request. Today the origin-side remote lock RPC can effectively wait for the existing default, which is extremely long, so the owner has no independent way to stop a poisoned remote lock request and force the origin transaction to unwind.

`RemoteLockOwnerWaitTimeout` is intentionally an internal owner-side safety budget for remote lock request handling. It is not meant to implement the user's session `lock_wait_timeout`, and it should still fire even when the user explicitly configures a longer lock wait. Otherwise a large session value, or the current default behavior, can reproduce the exact hung remote-owner failure this PR is designed to stop. The origin receives `ErrRemoteLockWaitTimeout` and the transaction is driven down the rollback path; this is a distributed lock-service protection error, not a successful user lock wait with shortened semantics.

2. I also disagree with the premise that `ErrRemoteLockWaitTimeout` means no lock was ever granted.

For a remote multi-row lock request, the owner can grant a prefix of the requested locks and then block on a later row. The timeout can therefore happen after the owner has already installed some locks for the transaction. In that case the origin must remember the remote bind/table in its held-lock metadata so rollback/close sends the remote unlock and cleans those already-granted owner-side locks.

If the owner timed out before granting anything, the rollback unlock is still safe: remote unlock is keyed by txn ID and is a no-op when the owner has no locks for that txn. If we skip `lockAdded` on this error, the unsafe case is the opposite one: a partially granted remote lock batch can be left on the owner until stale/orphan cleanup. So keeping the remote lock metadata here is intentional cleanup state, not a claim that the whole request succeeded.

3. Agreed on the scalability issue in the original owner-local fast path implementation. I changed this part.

The latest commit no longer scans the entire local lock table under `localLockTable.mu`. The owner now records/removes bounded owner-local wait edges as waiters are enqueued/removed, and the fast path walks only the graph rooted at the actual conflicting holders. I also restricted this path to the remote-owner async row/exclusive case that motivated the PR, so local-only conflicts do not pay for it.

Validation after the update:

```text
go test -v ./pkg/lockservice -count=1 -timeout=20m
PASS
ok  github.com/matrixorigin/matrixone/pkg/lockservice  580.828s
```

XuPeng-SH (approved):
Re-reviewing after the latest changes and discussion: I no longer see a substantive correctness blocker here.

The owner-local fast path is now scoped to the remote-owner async row/exclusive case and walks the bounded `ownerLocalWaits` graph instead of scanning the whole table lock store under the mutex.

On the `ErrRemoteLockWaitTimeout` path, I also no longer think the origin-side `lockAdded(...)` bookkeeping is wrong in itself. A remote multi-row request can time out after partially granting earlier rows, so the origin still needs enough metadata to drive remote unlock on rollback. In the no-grant case the later unlock is a no-op by txn ID, which is acceptable.

The remaining question about `RemoteLockOwnerWaitTimeout` shortening a larger user `lock_wait_timeout` is real as a user-visible behavior change, but the PR description and follow-up rationale make it clear this is intentional: it is an owner-side safety cap for poisoned remote lock handling, not a replacement for the session lock-wait contract. I would still suggest adding one explicit precedence test (`lock_wait_timeout` shorter vs owner cap shorter) and one no-false-positive graph test, but I would not block merge on that.

LGTM.

Issue #25437, opened 2026-07-03
[Bug]: protocol prepared UPDATE corrupts indexed readback for VARBINARY and BIGINT UNSIGNED


Ariznawlll:
## Summary

On latest `main` (`7e70bee98`), a binary-protocol prepared `UPDATE` can make indexed/point read paths return incorrect values for updated `VARBINARY` and `BIGINT UNSIGNED` columns, while a full table scan returns the correct data.

This was found during protocol-level prepared statement exploration with Go `database/sql` + `go-sql-driver/mysql` (`db.Prepare`, i.e. COM_STMT_PREPARE), comparing MySQL 8.0.45 and MatrixOne.

## Environment

- MatrixOne: latest `main` at `7e70bee98`
- MySQL oracle: MySQL 8.0.45
- Driver: `github.com/go-sql-driver/mysql v1.8.1`
- Path: binary-protocol prepared statement (`db.Prepare`), not SQL-level `PREPARE ... FROM`

## Reproduce

Go repro body:

```go
stmt, err := db.Prepare("update pp_dst set status=?, amount=?, dt=?, tm=?, bin=?, u=? where tenant=? and id=?")
_, err = stmt.Exec(
    "paid",
    "99.50",
    "2024-03-03 03:03:03.123456",
    "03:03:03.123456",
    []byte{0x7a, 0x7b},
    "99",
    1,
    101,
)
```

Table/data setup:

```sql
drop database if exists pp_dml_readpath;
create database pp_dml_readpath;
use pp_dml_readpath;

create table pp_dst (
  tenant int not null,
  id int not null,
  acct varchar(20) not null,
  amount decimal(10,2),
  status varchar(20),
  dt datetime(6),
  tm time(6),
  bin varbinary(8),
  u bigint unsigned,
  primary key (tenant, id),
  unique key uk_dst_acct (tenant, acct),
  key idx_dst_status_amount_id (status, amount, id),
  key idx_dst_dt_tenant_id (dt, tenant, id),
  key idx_dst_u_status_id (u, status, id),
  key idx_dst_bin_id (bin, id)
);

insert into pp_dst values
  (1,101,'a1',10.25,'new','2024-01-02 10:20:30.123456','10:20:30.123456',x'4142',18446744073709551615),
  (1,102,'a2',20.25,'old','2024-01-03 11:00:00.000000','11:00:00.000000',x'414243',8),
[...]

PR #25448, opened 2026-07-03
fix: binary protocol TIME microsecond encoding corrupts subsequent columns (#25437)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25437

The binary-protocol result row writer (`appendTime`) wrote the microsecond field as 8 bytes (`appendUint64`) when MySQL protocol specifies 4 bytes. This shifted every column after the TIME value by 4 bytes in the client's parser, causing VARBINARY to appear empty and BIGINT UNSIGNED to read garbage bytes (`4771354034768445440` = the hex string lenenc marker).

The symmetric parameter-parser bug (`readTime`) also failed to advance `pos` past the 4-byte microsecond field on length-12 TIME parameters, which would cause misparse of subsequent parameters with JDBC/C connectors.

### Changes:
- `appendTime`: use `appendUint32(uint32(msec))` instead of `appendUint64(msec)`
- `readTime`: capture `pos` from `ReadUint32` to advance past microsecond bytes
- New UT: binary row microsecond alignment and `readTime` pos advancement

### Verification:
- Focused and full frontend test suite: green
- Go reproducible end-to-end (protocol prepared UPDATE → binary protocol readback): all read paths return correct values
- `git diff --check`: passed

### DIFF
--- a/pkg/frontend/mysql_protocol.go
+++ b/pkg/frontend/mysql_protocol.go
@@ -1010,9 +1010,9 @@ func (mp *MysqlProtocolImpl) ParseExecuteData(ctx context.Context, proc *process
 					val = "0d 00:00:00"
 				case 8, 12:
 					pos, val, ok = mp.readTime(data, pos, length)
-				if !ok {
-					return moerr.NewInvalidInput(ctx, "mysql protocol error, malformed packet")
-				}
+					if !ok {
+						return moerr.NewInvalidInput(ctx, "mysql protocol error, malformed packet")
+					}
 				default:
 					return moerr.NewInvalidInput(ctx, "mysql protocol error, malformed packet")
 				}
--- a/pkg/frontend/mysql_protocol.go
+++ b/pkg/frontend/mysql_protocol.go
@@ -1009,7 +1009,10 @@ func (mp *MysqlProtocolImpl) ParseExecuteData(ctx context.Context, proc *process
 				case 0:
 					val = "0d 00:00:00"
 				case 8, 12:
-					pos, val = mp.readTime(data, pos, length)
+					pos, val, ok = mp.readTime(data, pos, length)
+				if !ok {
+					return moerr.NewInvalidInput(ctx, "mysql protocol error, malformed packet")
+				}
 				default:
 					return moerr.NewInvalidInput(ctx, "mysql protocol error, malformed packet")
 				}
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
LGTM. The fix is narrow and matches the MySQL binary protocol layout: TIME length-12 carries sign(1)+days(4)+h/m/s(3)+microseconds(4), not an 8-byte microsecond field. The paired `readTime` offset fix also closes the prepared-parameter side of the same alignment bug.

I checked the affected paths from a few angles:
- binary result row encoding: TIME with nonzero micros now leaves the following column aligned
- prepared parameter decoding: len-12 TIME advances past the 4 microsecond bytes
- zero-microsecond and zero-time cases still use length 8 / 0 respectively
- negative TIME with days is covered
- datetime/timestamp already use 4-byte microseconds and are not regressed

Local checks run:
- `make err-check`
- focused frontend tests for the new/related binary row cases
- full `go test -count=1 ./pkg/frontend` with local macOS CGo rpath setup

CI still needs to finish, but I don't see a code-review blocker in this diff.

XuPeng-SH (changes requested):
One small change needed on the latest head (`1f52e1541ec9b8a69e1f5471e6d43a25baeca29d`): please run `gofmt` on `pkg/frontend/mysql_protocol.go`.

`gofmt -l pkg/frontend/mysql_protocol.go pkg/frontend/mysql_protocol_test.go` currently reports:

```text
pkg/frontend/mysql_protocol.go
```

The functional direction still looks correct, and the latest added malformed-packet guard is good. Local checks on the latest head:
- `git diff --check origin/main...HEAD` passed
- `make err-check` passed
- focused frontend tests for the changed TIME binary protocol paths passed

After formatting, I do not see another code-review blocker in this diff.

XuPeng-SH (approved):
LGTM on the latest head (`cdcdad6ab0f9102916dca7c68c16831f553d18ef`). The prior gofmt issue is fixed.

Checked again after the formatting commit:
- `gofmt -l pkg/frontend/mysql_protocol.go pkg/frontend/mysql_protocol_test.go` is clean
- `git diff --check origin/main...HEAD` passed
- `make err-check` passed
- focused frontend tests for TIME binary row alignment / readTime offset / datetime binary rows passed

The code path is narrow and the fix matches the MySQL binary protocol TIME layout: length-12 TIME uses 4 bytes for microseconds, and the prepared parameter parser now advances past those bytes with malformed-packet handling.

Issue #25423, opened 2026-07-03
[Bug]: Protocol-level prepared statements fail with mysql-connector-python prepared cursor


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Searched open/all issues for:

- `"COM_STMT_PREPARE"`
- `"binary protocol"`
- `"mysql.connector"`
- `"prepared=True"`
- `"MySQL server has gone away" "prepared"`

Related but not duplicate:

- #24161 / #24162 track broad BI connector compatibility, but do not provide a minimal protocol-level prepared statement repro.
- #23438 is closed and about SQL prepared division-by-zero behavior, not the MySQL binary prepared protocol.

### Branch Name

main

### Commit ID

`51656deb6003`

### Other Environment Information

- MatrixOne: `select version()` returns `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- Client: `mysql-connector-python` in Python, using `cursor(prepared=True)`.
- Text-protocol queries from the same connector work against MatrixOne.

### Actual Behavior

Protocol-level prepared statements fail against MatrixOne. A simple `select ? + ?` using `mysql-connector-python` with `cursor(prepared=True)` returns:

```text
DatabaseError 2006 (HY000): MySQL server has gone away
```

The same query succeeds against MySQL 8.0.45 and returns `3`.

Control check against MatrixOne:

- mysql CLI text protocol works: `select version(), 1+2` returns normally.
- `mysql-connector-python` non-prepared cursor also works: `cursor(prepared=False).execute('select %s + %s', (1, 2))` returns `3`.

So this appears specific to the MySQL protocol prepared-statement path (`COM_STMT_PREPARE` / binary protocol), not a general connection or SQL execution failure.

### Expected Behavior

Same as MySQL: protocol-level prepared statements should prepare and execute successfully. This is needed by MySQL-compatible drivers and applications that use server-side prepared statements.

### Steps to Reproduce

```python
import mysql.connector

# MySQL succeeds with the same code.
cn = mysql.connector.connect(
    user='dump',
    password='111',
    host='127.0.0.1',
    port=7002,
    connection_timeout=5,
)

cur = cn.cursor(prepared=True)
cur.execute('select ? + ?', (1, 2))
print(cur.fetchall())
cur.close()
cn.close()
```

MatrixOne result:

```text
DatabaseError 2006 (HY000): MySQL server has gone away
```

[...]

PR #25452, opened 2026-07-04
fix: protocol-level prepared `select ? + ?` fails at prepare and panics at execute (#25423)

ck89119:
## What type of PR is this?

- [x] BUG

issue #25423

Protocol-level prepared statements such as `select ? + ?` (e.g. via
`mysql-connector-python`'s `cursor(prepared=True)`) failed during
`COM_STMT_PREPARE` with:

```
invalid argument operator +, bad value [TEXT TEXT]
```

### Root cause

`baseBindParam` bound every unknown prepared-statement parameter as `T_text`.
For `? + ?` the `+` resolver therefore saw two concrete `TEXT` operands and
rejected the arithmetic at PREPARE time, before the real values arrive with
`COM_STMT_EXECUTE`. The function type system already knows how to deduce
`T_any + T_any` into a numeric type, and string-context binders (`LIKE`, etc.)
already handle `T_any` explicitly.

### Fix (two parts)

1. `pkg/sql/plan/base_binder.go`: bind an unresolved parameter as `T_any`
   instead of `T_text`, so the existing type deduction promotes `? + ?` to a
   numeric type (`int64`).

2. `pkg/sql/colexec/evalExpression.go`: `T_any` alone is not enough. Parameter
   values are always stored by the frontend as **text bytes**, and
   `ResetPreparePlan` keeps the `ParamRef` in the compiled plan instead of
   substituting typed constants. At runtime `ParamExpressionExecutor`
   materializes the value via `NewConstBytes(paramRef.Typ, textBytes)`; a
   `T_any` type has element size 0 and panics in `NewConstBytes` /
   `SetConstBytes` at `COM_STMT_EXECUTE` (and `cast(T_any -> int64)` would
   otherwise be treated as scalar `NULL`). We materialize an unresolved `T_any`
   parameter as `T_text`, so the const has a real element size and the
   planner-inserted cast converts the text bytes to the wanted type. This keeps
   the pre-existing "params materialize as text" behavior while decoupling the
   planning-time type from the runtime materialization type.

Real `TEXT` columns keep their existing semantics: `text_col + text_col` still
errors with `[TEXT TEXT]`; this PR does not relax arithmetic for real text
values.

### Verification

- End-to-end with `mysql-connector-python` 9.7.0 `cursor(prepared=True)`
  `select ? + ?` with `(1, 2)` returns `3` on both the C-extension and pure
  Python paths.
- Text-protocol `PREPARE ... EXECUTE USING` returns `3`.
- Regression spot-checks unchanged: `> ?` int comparison, `= ?` string
  equality, `LIKE ?`, `UPDATE ... = ?`, `INSERT (?,?,?)`.
- New unit tests: `TestPrepareArithmeticParams` (plan) and
  `TestParamExpressionExecutorAnyType` (colexec). `pkg/sql/plan` and
  `pkg/sql/colexec` pass; `go vet` and `git diff --check` clean.

Known boundary (pre-existing limitation, out of scope): because params are
stored as text and `any + any` deduces to `int64` at PREPARE, passing a
non-integer such as `2.5` to `? + ?` errors at `cast to int`. Before this fix
the statement could not even be prepared, so this is a strict improvement.

## PR Checklist

- [x] Added/updated unit tests
- [x] Verified end-to-end against a running mo-service


XuPeng-SH (changes requested):
I found a real correctness regression in the current fix.

The prepare-time failure is gone, but the new solution gets there by eagerly rewriting unresolved arithmetic params to `CAST(? AS DOUBLE)` whenever the peer is numeric or another param. That is not just a type-inference aid — it changes the runtime semantics of exact integer / unsigned arithmetic into float64 arithmetic.

A concrete visible regression is already encoded in the updated BVT expectation:

- `PREPARE s FROM 'SELECT 0 + ?'`
- `EXECUTE s USING @maxint` where `@maxint = 18446744073709551615`

The new expected result is:

`1.8446744073709552E19`

That is silent precision loss. The exact integer value is no longer preserved. Once values cross 2^53, different integers collapse to the same float64, and operators like `+`, `-`, `%`, `DIV`, `MOD` can all become observably wrong.

This happens because `promoteTextParamsForArith()` hard-codes `DOUBLE` before normal operator resolution, bypassing the existing exact coercion tables that would otherwise keep integer / unsigned peers on exact numeric paths.

So while the PR fixes "cannot prepare `select ? + ?`", it does so by changing exact prepared arithmetic into approximate floating-point arithmetic. That is not MySQL-compatible and is not safe to merge.

I would fix this by avoiding eager `DOUBLE` casts for exact numeric arithmetic: keep exact int/uint peers on exact cast paths, and solve the dual-unresolved-param prepare case with deferred typing / later specialization instead of forcing float64 at bind time.

XuPeng-SH (approved):
Looks good to me.

The final shape is much safer than the intermediate `T_any` approach: the fix stays narrowly scoped to unresolved arithmetic `ParamRef`s, leaves real TEXT semantics alone, and keeps runtime parameter materialization on the existing `T_text` path. I also checked the mixed-case guardrails and the new plan/executor unit coverage; I don’t see a correctness blocker.

Non-blocking follow-up only: current coverage is still missing one true end-to-end `PREPARE ... EXECUTE` regression for dual-param arithmetic (`? + ?`, and ideally one divide/mod-by-zero unhappy path). Given the narrow scope of this patch I’m not blocking on that.

XuPeng-SH (changes requested):
I need to switch my earlier approval to request changes: the delayed runtime pass found a correctness blocker.

`promoteTextParamsForArith` currently rewrites every dual-param arithmetic expression (`? + ?`, `? % ?`, `? DIV ?`, etc.) to `CAST(? AS DOUBLE)`. That silently loses precision for integer inputs above `2^53`, and it also routes `DIV` / `MOD` away from the existing exact integer path and its range checks.

Concretely, exact integer inputs are still received from the protocol, then stringified/materialized as text params, and the new binder helper forces those text params through `float64`. After that:
- `%` / `MOD` go through `math.Mod`
- `DIV` goes through `int64(v1 / v2)`

So values like `9007199254740993` can round to `9007199254740992`, meaning `? % 2` can produce `0` instead of `1`, and large integer `DIV`/`MOD` lose the old exactness / out-of-range behavior.

I would treat that as a blocker. The fix needs to preserve exact integer semantics for integer-like params (or defer unresolved-param typing until execute time) instead of forcing the dual-param arithmetic path onto `DOUBLE`.

The added plan/executor unit tests are helpful, but they currently do not cover this precision path end-to-end.

XuPeng-SH (changes requested):
I found a blocker in the latest head (04d3c982).

High: the latest fix now hard-casts dual unresolved arithmetic parameters to Decimal128(38,12), but that is not MySQL-compatible for this issue. MySQL 8.0's PREPARE parameter-marker rules say that when both operands of a binary arithmetic operator are parameters, their type is decided by context; if the arithmetic operator has no type-determining context, the derived type is DOUBLE PRECISION. This is exactly the reported case: top-level `select ? + ?` in a protocol prepared statement. The issue repro also shows MySQL returning `[(3.0,)]`, i.e. floating metadata/value semantics.

The current plan tests encode `cast(? AS decimal128(38,12))` as the expected shape. That can still prepare, but it changes protocol result metadata and execution behavior: fractional inputs beyond 12 scale can be rounded/truncated/errored differently, scientific/large values can overflow Decimal128 where MySQL DOUBLE would not, and expressions like `? * ?` with large numeric params can fail instead of producing MySQL's approximate DOUBLE result. So this fixes the prepare-time [TEXT TEXT] failure by moving the statement onto a different numeric domain than MySQL specifies.

Suggested fix: keep the narrow dual-ParamRef rewrite, but use the MySQL-derived type for the no-context arithmetic case: DOUBLE. The previous exactness concern should be handled by keeping mixed known-type cases (`0 + ?`, `? + 5`, assignment/update contexts, etc.) on the existing exact coercion tables, which the current dual-param-only guard already does. Please also add/adjust tests to assert the top-level dual-param arithmetic result/cast type is DOUBLE, and ideally add one end-to-end protocol or SQL PREPARE/EXECUTE regression that validates `select ? + ?` returns MySQL-compatible floating behavior.

Reference: MySQL 8.0 PREPARE docs, parameter marker type rules: https://dev.mysql.com/doc/refman/8.0/en/prepare.html

ck89119:
## Round-6 findings: MOD fixed, and the rest verified against MySQL 8.4

**MOD(?, ?)** (finding 2) is fixed in ecdae9a5d: the AST purity gate rejected `tree.FuncExpr`, so the function spelling stayed on `DOUBLE` while the infix `? % ?` / `? mod ?` were retyped. Verified `CAST(MOD(?, ?) AS SIGNED)` with `9007199254740993, 2` now returns `1` (exact), matching MySQL.

For the INSERT ... SELECT completeness (finding 1) I ran the cases against a local **MySQL 8.4.8** to establish the real rule before expanding the change. The results reframe the whole feature:

| Statement (params `9007199254740993, 0`) | MySQL 8.4 | Rule |
| --- | --- | --- |
| `SELECT ? + ?` | `9.0071...e15` (->992) | no context -> DOUBLE |
| `SELECT MOD(?, ?)` | `0` | no context -> DOUBLE |
| `CAST(? + ? AS DECIMAL(30,0))` | `...993` | cast -> exact |
| `CAST(MOD(?, ?) AS SIGNED)` | `1` | cast -> exact |
| `CAST((? + ?) + 1 AS DECIMAL(30,0))` | `...994` | exact; literal `1` kept |
| `CAST((10 + 0) * 0.6 AS SIGNED)` | `6` | `0.6` kept, not forced to int |
| `SELECT (? + ?) + d` (d DECIMAL) | `...998` | decimal sibling -> exact |
| `SELECT (? + ?) + f` (f DOUBLE) | `...994` | double sibling -> DOUBLE |
| `UPDATE dec = (? + ?) + f` (f DOUBLE, ±1e30) | no overflow | **double sibling overrides the DECIMAL target** |

The real MySQL rule is: a parameter's type is derived from the **type-determining context of the whole surrounding expression** — typed siblings *and* the outer cast/assignment, combined by the arithmetic type rules; literals/columns keep their own type; no context anywhere -> `DOUBLE`.

This also resolves the earlier tension between the column-peer request (`set dec = (? + ?) + col`) and the overflow finding: the last row confirms a **`DOUBLE` sibling must override even a `DECIMAL` assignment target**. Pushing the target column type onto the params (the earlier column-peer approach) reintroduces overflow for `±1e30` operands that cancel in `DOUBLE`. So the param-only rule now in this PR is the correct safe choice, not a regression.

### Scope of this PR

This PR implements the **direct, no-typed-sibling** exact contexts, which match MySQL exactly:
`CAST(? +/-/*//%/div/mod ? AS exact)`, `MOD(?, ?)` under a cast, unary `-(? + ?)`, `col = ? + ?` (UPDATE / VALUES / ON DUPLICATE KEY UPDATE), and the direct `INSERT ... SELECT <list>`.

Everywhere else it falls back to the **safe `DOUBLE` promotion**. That fallback is **precision-only** — verified it never yields a wrong value or an overflow (`TestPrepareArithNonParamPeerKeepsDouble`). The remaining divergences from MySQL are:
- nested arithmetic with a typed sibling (`(? + ?) + 1`, `(? + ?) + decimal_col`, `(? + ?) * 0.6`),
- non-direct INSERT ... SELECT (ParenSelect, UNION, star expansion, legacy path).

Matching MySQL there needs full parameter type-context inference in the binder (bottom-up over operand types + outer cast/assignment, with typed-sibling precedence). Rather than grow this PR's surface with more plan-shape surgery, that work is tracked in **#25705**. This keeps the current change bounded and provably safe.

XuPeng-SH (changes requested):
Requesting changes for one correctness blocker.

`CAST((? + ?) + 1 AS DECIMAL(30,0))` is deliberately left on the DOUBLE path because the AST gate rejects the literal peer, and `TestPrepareArithNonParamPeerKeepsDouble` locks that behavior in. With parameters `9007199254740993` and `0`, the first parameter is rounded to `9007199254740992` when cast to float64; adding `1` at that magnitude still produces `9007199254740992`, so the outer DECIMAL cast returns/stores `9007199254740992`. MySQL's exact context produces `9007199254740994`, as the test comment itself acknowledges. This is a silent wrong result, not a precision-only divergence that “never yields a wrong result.”

Please preserve exact-context parameter typing when a typed sibling participates, without coercing that sibling to the outer target type, and add an execution-level regression using values above 2^53. Plan-shape assertions alone currently validate the incorrect DOUBLE route.

I found no additional blocking issue in the resource/performance audit: the new work is bounded bind-time tree traversal and introduces no waits, goroutines, external resources, or persistent growth. The Basic Benchmark failure is unrelated infrastructure noise: the runner repeatedly lost TLS while downloading Go 1.26.4, before any benchmark executed.

gouhongshen (changes requested):
## Codex automated review

Exact BIT and typed-sibling/derived INSERT paths still lose prepared integer precision, and AST rebinding introduces a quoted-numeric execution regression.

### P1 - BIT targets bypass exact-context rebinding (pkg/sql/plan/base_binder.go:1708)
isPreciseNumericType relies on IsInteger(), but T_bit is explicitly numeric while IsInteger() excludes it. BIT target columns therefore retain the T_text→float64 promotion. Inserting 9007199254740993 + 0 into BIT(64) rounds through float64 to 9007199254740992 before numericToBit stores it. Add BIT/uint64 exact-context coverage.

### P1 - Exact typed siblings leave nested parameters on DOUBLE (pkg/sql/plan/base_binder.go:1756)
The AST gate deliberately rejects column references, so `UPDATE t SET d = (? + ?) + d` with DECIMAL d never backfills the inner parameters. Bottom-up binding has already promoted them to float64, and float64+decimal remains float64; the later assignment cast cannot restore lost bits. The same gate rejects exact CAST operands. For 9007199254740993 + 0 + an exact column, this returns a rounded result, contrary to the documented context-derived parameter rules ([MySQL PREPARE](https://dev.mysql.com/doc/refman/8.4/en/prepare.html), [arithmetic rules](https://dev.mysql.com/doc/refman/8.4/en/arithmetic-functions.html)).

### P1 - Non-direct INSERT ... SELECT skips target-context rebinding (pkg/sql/plan/bind_insert.go:2627)
backfillProj is enabled only when the source is a direct SelectClause whose projection AST aligns positionally. ParenSelect, derived `SELECT *`, UNION, and similar forms leave it disabled, so an inner `SELECT ? + ?` is bound as DOUBLE and only its already-rounded output is cast to the destination DECIMAL/INTEGER column. For example, `INSERT INTO t(d) SELECT * FROM (SELECT ? + ? AS d) s` loses 9007199254740993 precision; the direct INSERT ... SELECT tests do not cover this path.

### P2 - Quoted numeric literals can become failing integer casts (pkg/sql/plan/base_binder.go:1786)
This accepts every NumVal, including P_char string literals. During rebinding, `'0.6'` is bound as T_text; in `CAST((? + ?) * '0.6' AS SIGNED)`, the exact target makes the operator choose int64*text and inserts a text→int64 cast. strToSigned rejects the decimal token, whereas the promoted DOUBLE*text path uses strToFloat and accepts it. Restrict the gate to numeric literal kinds or preserve string coercion semantics.

aunjgr (changes requested):
Exact assignment propagation remains incomplete. BIT targets still fall through paths that can round values above 2^53 via float64; nested parameter expressions can leave inner parameters typed as DOUBLE; non-direct INSERT SELECT forms skip target context; and rebinding quoted numeric literals can change valid string-cast behavior. Please close all assignment shapes under one target-type propagation invariant and add exact large-integer and quoted-literal regressions.

aunjgr (dismissed):
Exact assignment propagation remains incomplete. BIT targets still fall through paths that can round values above 2^53 via float64; nested parameter expressions can leave inner parameters typed as DOUBLE; non-direct INSERT SELECT forms skip target context; and rebinding quoted numeric literals can change valid string-cast behavior. Please close all assignment shapes under one target-type propagation invariant and add exact large-integer and quoted-literal regressions.

PR #25453, opened 2026-07-04
fix: allow REPLACE in binary-protocol COM_STMT_PREPARE (#25429)

### LINKED ISSUE #25429: [Bug]: Binary-protocol COM_STMT_PREPARE rejects REPLACE (missing replace_stmt in prepareable_stmt grammar)

Searched open issues for `"REPLACE" "prepare"`, `"COM_STMT_PREPARE"`, `"binary protocol" "prepared"`. Related but not duplicate:

- #25420 reports wrong data for text-protocol `PREPARE ... FROM 'replace ...'` (which parses fine via the `FROM STRING` grammar rule). This issue is about the binary protocol path failing to prepare `REPLACE` at all.
- #24918 tracks general `REPLACE` compatibility gaps but does not mention prepared statements.
- #25423 reports mysql-connector-python `cursor(prepared=True)` failing with "MySQL server has gone away" on `select ? + ?`. That is a different symptom: binary-protocol prepared statements in general work with go-sql-driver (e.g. an `INSERT ... SELECT ... ON DUPLICATE KEY UPDATE` with a `?` placeholder prepares and executes fine on the same build); only `REPLACE` is rejected by the parser.

### Branch Name

main

### Commit ID

`2a8f3b884` (also reproduced on `51656deb6003`)

- Hardware parameters: Apple Silicon (darwin/arm64)
- OS type: macOS
- Others: standalone launch (`etc/launch/launch.toml`); client: go-sql-driver/mysql v1.10.0 (server-side prepared statements, i.e. `interpolateParams=false`)

### Actual Behavior

Preparing a `REPLACE` statement through the MySQL **binary protocol** (`COM_STMT_PREPARE`) fails with a syntax error:

```text
Error 1064 (HY000): SQL parser error: You have an error in your SQL syntax; check the manual that
corresponds to your MatrixOne server version for the right syntax to use. syntax error at line 1
column 35 near " replace into dst (tenant,id,acct,amount,status)
                select tenant,id,acct,amount,status from src where tenant = ?";
```

Any driver that uses server-side prepared statements hits this (go-sql-driver/mysql with default settings, JDBC with `useServerPrepStmts=true`, etc.). Connector/J silently falls back to client-side emulation, so the failure is easy to miss but the behavior still diverges from MySQL.

Note the text-protocol SQL form `PREPARE s FROM 'replace into ...'` works, because the quoted string goes through the `FROM STRING` grammar rule and the inner statement is parsed separately.

### Expected Behavior

Same as MySQL: `REPLACE` is a preparable statement (it is in MySQL's documented list of statements permitted in prepared statements), so `COM_STMT_PREPARE` of a `REPLACE ... VALUES / SELECT` with `?` placeholders should succeed and execute like the non-prepared statement.

### Steps to Reproduce

```sql
create database repro_prepare_replace;
use repro_prepare_replace;
create table t (a int primary key, b int);
```

Then from any client using server-side (binary protocol) prepared statements, e.g. Go:

```go
db, _ := sql.Open("mysql", "root:111@tcp(127.0.0.1:6001)/repro_prepare_replace")
stmt, err := db.Prepare("replace into t values (?, ?)") // COM_STMT_PREPARE
// err: Error 1064 (HY000): SQL parser error: ... syntax error ... near " replace into t values (?, ?)"
_ = stmt
```

The same statement works when executed directly (`db.Exec` with no args) or via text-protocol `PREPARE s FROM 'replace into t values (1, 2)'`.

### Additional information

Root cause: the frontend rewrites `COM_STMT_PREPARE` into `prepare __mo_stmt_id_x from <raw sql>` (pkg/frontend/mysql_cmd_executor.go, `case COM_STMT_PREPARE`), and the unquoted form is parsed via the `prepareable_stmt` grammar rule in `pkg/sql/parsers/dialect/mysql/mysql_sql.y`:

```text
prepareable_stmt:
    create_stmt
|   alter_stmt
|   insert_stmt
|   delete_stmt
|   drop_stmt
|   show_stmt
|   update_stmt
|   select_stmt
```

`replace_stmt` is missing from this list (it is present in `explainable_stmt`), so the wrapped statement fails to parse. Adding `replace_stmt` to `prepareable_stmt` (and regenerating the parser) should close the gap; execute-side handling likely needs a matching check since `tree.Replace` must be accepted wherever prepared statement plans are built.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25429

Preparing a `REPLACE` statement through the MySQL binary protocol
(`COM_STMT_PREPARE`, i.e. go-sql-driver with default settings, JDBC with
`useServerPrepStmts=true`, etc.) failed with a `1064` syntax error:

```
Error 1064 (HY000): SQL parser error: ... syntax error ... near " replace into t values (?, ?)"
```

Root cause: the frontend rewrites `COM_STMT_PREPARE` into an unquoted
`prepare <name> from <raw sql>`, which is parsed via the
`prepareable_stmt` grammar rule in `mysql_sql.y`. That rule was missing
`replace_stmt` (it is present only in `explainable_stmt`), so the wrapped
statement failed to parse. The quoted text-protocol form
`PREPARE s FROM 'replace ...'` worked because it goes through the
`FROM STRING` rule and parses the inner statement separately.

`REPLACE` is a preparable statement in MySQL, so this diverges from MySQL
behavior.

### Fix

- Add `replace_stmt` to `prepareable_stmt` and regenerate the parser.
- No execute-side change is needed: `getPreparePlan` / `BuildPlan` /
  `bindAndOptimizeReplaceQuery` already handle `*tree.Replace` for
  prepared statements (which is why the quoted text-protocol form already
  worked end-to-end).

### Tests

- Parser unit tests for prepared `REPLACE` (VALUES and SELECT sources).
- BVT coverage in `prepare/prepare_all.sql` for prepared `REPLACE`
  (VALUES and SELECT sources), verified via mo-tester.
- Verified end-to-end with go-sql-driver (`interpolateParams=false`,
  real binary-protocol `COM_STMT_PREPARE`): `replace into t values (?, ?)`
  now prepares and executes correctly.


XuPeng-SH (approved):
I did not find a blocker in this parser fix.

Adding `replace_stmt` to `prepareable_stmt` is the right fix for the unquoted `prepare <name> from <sql>` path used by binary-protocol `COM_STMT_PREPARE`, and it matches the execute-side support that was already present for `*tree.Replace`. The parser unit cases and the `prepare_all` additions also cover both VALUES and SELECT forms on the same grammar path.

iamlinjunhong (approved):

PR #25455, opened 2026-07-04
fix: preserve multi-cn for small ivf_search entry scans

### LINKED ISSUE #25454: [Tech Request]: Preserve Multi-CN for ivf_search entries scans under small estimated stats

## What would you like to be added?

Add a stable scheduling policy for `ivf_search` internal entries scans to keep Multi-CN execution under low-estimate cases, instead of falling back to One-CN from row/block shortcut logic.

### Scope

- IVFFLAT and HNSW vector index entries-table internal scans in query execution.
- `ivf_search` paths that go through function-scan or internal index-reader scan.
- Preserve Multi-CN intent unless explicitly blocked by existing hard constraints.

## Why is this needed?

- Small estimated statistics can still trigger one-CN execution, causing unstable parallelism.
- Multi-CN cache utilization and cluster throughput are reduced for vector lookup heavy workloads.
- The fix should make IVF entries scheduling deterministic and more cache-friendly across CNs.

## Additional information

This aligns with existing architecture: keep IVF entries scan classified as Multi-CN in planning, and prevent later local-scan fallback. This is a policy-level improvement and does not require a new object/block split mechanism first.

- Branch: `ivf-multi-cn-0704`
- Commit: `cb8510588`

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fix #25454

This change keeps `ivf_search` entries scans on Multi-CN even when estimated stats are tiny.

It fixes two scheduling regressions in the IVF search path:
- planner side: prevent the vector rowsize shortcut from downgrading an already Multi-CN IVF search decision back to One-CN
- compile side: prevent small internal IVF entries scans from being forced onto the current CN by the generic small-block shortcut

Behavior kept unchanged:
- explicit `ForceOneCN` still wins
- small non-IVF scans still stay on the current CN
- DDL / expr-based shuffle guards still cap execution to One-CN where required

This improves execution stability and CN cache reuse for vector search workloads. It does not change the current object-level dispatch model, so a single-object table can still end up executing on one CN.

## Testing

- `go test ./pkg/sql/plan`
- `go test ./pkg/sql/compile`
- `go build ./pkg/sql/plan ./pkg/sql/compile`
- `go vet ./pkg/sql/plan ./pkg/sql/compile`
- `git diff --check`

### DIFF
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -5136,6 +5136,10 @@ func shouldScanOnCurrentCN(c *Compile, node *plan.Node, forceSingle bool) bool {
 		return true
 	}
 
+	if plan2.IsIvfSearchEntriesInternalScan(node) {
+		return false
+	}
+
 	if !plan2.GetForceScanOnMultiCN() &&
 		node.Stats.BlockNum <= int32(plan2.BlockThresholdForOneCN) {
 		return true
--- a/pkg/sql/compile/max_dop_test.go
+++ b/pkg/sql/compile/max_dop_test.go
@@ -17,6 +17,7 @@ package compile
 import (
 	"testing"
 
+	"github.com/matrixorigin/matrixone/pkg/catalog"
 	"github.com/matrixorigin/matrixone/pkg/pb/plan"
 	plan2 "github.com/matrixorigin/matrixone/pkg/sql/plan"
 	"github.com/matrixorigin/matrixone/pkg/vm/engine"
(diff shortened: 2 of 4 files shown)


XuPeng-SH:
Request changes (GitHub does not allow me to submit an official REQUEST_CHANGES review on my own PR).

I think the direction is right (keep the actual `ivf_search` entries scan from falling back to local One-CN when stats are tiny), and the compile-side `IndexReaderParam` guard is the right kind of signal. But the planner-side implementation is broader than the invariant it wants to preserve.

Blocking concern:

`queryHasIvfSearchWithLimit` is a query-global boolean, and `isIvfSearchEntriesScan(node, hasIvfSearchWithLimit)` then treats every IVFFLAT entries table scan in the same query as an `ivf_search` entries scan, even if that scan is unrelated and has no `IndexReaderParam`. This is not equivalent to "the internal entries scan issued by ivf_search". It can force tiny unrelated entries scans to AP_MULTICN just because another node in the query is an `ivf_search` function scan.

It also adds an IVFFLAT-specific import and function-name branch in generic `pkg/sql/plan/stats.go`. That goes against the current index-plugin direction: generic SQL planning code should not grow new per-algorithm dispatch by importing `pkg/vectorindex/ivfflat/plugin/plan` and comparing `IVFFLATSearchFuncName`.

Suggested fix:

Use the actual scan-local signal for this policy. For the internal SQL path, the entries scan already carries `IndexReaderParam` through `IvfReaderParam` and `pushdownVectorIndexTopToTableScan`, and the compile-side path already uses `IsIvfSearchEntriesInternalScan(node)`. So the planner can avoid the query-global `ivf_search` scan detection and rely on the entries scan's own `IndexReaderParam` shape instead.

Concretely:
- remove `queryHasIvfSearchWithLimit` / `isIvfSearchWithLimitNode` and the `ivfflatplan` import from `stats.go`, unless there is a real plan shape where an entries scan must be identified only by a separate function-scan node;
- keep the `IndexReaderParam`-based path and add/keep tests proving `OrderBy+Limit` or `OrigFuncName+Limit` entries scans stay AP_MULTICN under tiny stats and do not get downgraded by the rowsize shortcut;
- add a regression test where a query contains an `ivf_search` function scan plus an unrelated IVFFLAT entries table scan without `IndexReaderParam`; the unrelated scan should not become AP_MULTICN solely because of the function scan;
- if function-scan identity is genuinely needed, carry an algorithm-agnostic planner flag/metadata from the plugin hook instead of adding a generic-layer IVFFLAT function-name branch.

Local verification I ran on the PR head:

- `go test -count=1 -timeout 180s ./pkg/sql/plan` passed with local CGo rpath setup
- `go test -count=1 -timeout 180s ./pkg/sql/compile` passed with local CGo rpath setup
- `go vet ./pkg/sql/plan` passed with local CGo rpath setup
- `go vet ./pkg/sql/compile` passed with local CGo rpath setup
- `git diff --check origin/main...pr/25455` passed

XuPeng-SH:
Addressed in e0eea767b.

I adopted the main point from the review and narrowed the planner-side policy to the actual scan-local signal only:
- removed the query-global `ivf_search` detection helpers from `pkg/sql/plan/stats.go`
- removed the IVFFLAT-specific planner import / function-name branch
- planner now only preserves Multi-CN for IVF entries scans that themselves carry the internal `IndexReaderParam` search shape
- kept the compile-side `IsIvfSearchEntriesInternalScan(node)` guard as-is
- added a regression test proving an unrelated entries scan is not promoted just because the same query also contains an `ivf_search` function scan

Re-verified locally after the change:
- `go test -count=1 -timeout 180s ./pkg/sql/plan`
- `go test -count=1 -timeout 180s ./pkg/sql/compile`
- `go build ./pkg/sql/plan ./pkg/sql/compile`
- `go vet ./pkg/sql/plan ./pkg/sql/compile`
- `git diff --check`

PR #25457, opened 2026-07-04
fix: harden ReadRows whole-result handling in system paths (#25456)

### LINKED ISSUE #25456: [Bug]: ReadRows whole-result assumptions can mis-handle batched results and crash ISCP HNSW jobs

### Branch Name
issue-25429-main

### Commit ID
36cfd187b64de9929690e4d39925d23172a83e81

- Hardware parameters: GitHub-hosted x64 runner
- OS type: Linux
- Others:
  - Workflow: Matrixone Standlone CI / Multi-CN e2e BVT Test on Linux/x64(LAUNCH, PROXY)
  - Job URL: https://github.com/matrixorigin/matrixone/actions/runs/28701131845/job/85119254826?pr=25453
  - PR: #25453

### Actual Behavior
The Multi-CN BVT failed in `test/distributed/cases/pessimistic_transaction/vector/vector_hnsw.sql` with:

```sql
create index idx01 using hnsw on t3(b) op_type "vector_l2_ops";
```

The visible SQL error was:

```text
backend connection closed
```

However, the first real failure happened earlier in the background ISCP/HNSW job flow:

1. ISCP logged:
   ```text
   invalid rows 1, expected 2
   ```
   while loading job status for two HNSW jobs on `vector_index_04`.
2. `pkg/iscp/iteration.go` treated `executor.Result.ReadRows` as if each callback saw the whole result set.
3. The permanent-error path then dereferenced a nil `JobStatus` entry and panicked:
   ```text
   ```
4. After the CN crashed, later HNSW SQLs only observed the downstream symptom:
   ```text
   backend connection closed
   ```

The same whole-result assumption also appears in several other production paths that read single-row metadata from `executor.Result.ReadRows`.

### Expected Behavior
- Multi-batch `ReadRows` callbacks should be handled with whole-result semantics rather than per-batch assumptions.
- HNSW/ISCP error handling should mark permanent failures without panicking.
- Single-row metadata readers should either validate the total row count across all batches or explicitly stop after a proven single-row contract.

### Steps to Reproduce
1. Run the Multi-CN BVT suite or directly run `test/distributed/cases/pessimistic_transaction/vector/vector_hnsw.sql`.
2. Observe the sequence around `vector_index_04`:
   - create `idx01`
   - create `idx02`
   - reindex `idx01`
   - reindex `idx02`
   - reindex `idx01` again
3. Continue until the later HNSW create-index on `t3` reports `backend connection closed`.
4. Check `mo-service.log` and observe:
   - `ISCP-Task Flush Permanent Error Message`
   - `invalid rows 1, expected 2`
   - panic in `pkg/iscp/iteration.go`
   - connection-refused errors to the crashed CN

### Additional information
Key root-cause observations:
- `pkg/util/executor/result.go` documents that `ReadRows` may invoke the callback multiple times, once per batch.
- The original ISCP code compared callback-local `rows` with the expected total and also reset array indexing from zero for each callback.
- The CI symptom was unrelated to the parser-only changes in PR #25453; it was caused by a pre-existing ISCP/vector-index error path.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25456

This PR fixes a real bug class in internal SQL readers that use
`executor.Result.ReadRows`.

`ReadRows` is batch-oriented: the callback may be invoked multiple times, once
per batch. Several production paths were assuming callback-local `rows`
described the whole result set, which caused two different failure modes:

1. **False/fragile row-count checks**
   - treating per-batch row counts as whole-result counts
   - counting callbacks instead of total rows

2. **State corruption / crash on unhappy paths**
   - ISCP `GetJobSpecs()` treated a multi-batch result as a missing-row error
   - the permanent-error flush path then dereferenced nil `JobStatus` entries
   - in vector HNSW CI this crashed a CN and surfaced later as
     `backend connection closed`

#### 1. ISCP multi-job status loading
- `pkg/iscp/iteration.go`
- load `job_name` and `job_id` together with `job_spec` / `job_status`
- accumulate across all `ReadRows` batches
- map rows by `JobKey` instead of per-batch positional assumptions
- normalize nil `JobStatus` slices before permanent-error flushing
- stop the local execution path cleanly once a permanent error has already been flushed

#### 2. System-path whole-result row counting
- `pkg/iscp/util.go`
- `pkg/publication/util.go`
- `pkg/incrservice/store_sql.go`

These paths now validate total rows across all batches instead of using
callback-local `rows`.

#### 3. Additional single-row hardening
- `pkg/iscp/watermark_updater.go`
- `pkg/sql/plan/function/func_feature_upsert.go`
- `pkg/bootstrap/versions/version.go`
- `pkg/bootstrap/versions/upgrade_tenant_task.go`

These readers now use whole-result semantics for single-row contracts. Where a
duplicate row would represent a terminal invariant break (for example version
state lookup), the code remains fail-fast.

### Tests
- Added focused regression tests for:
  - multi-batch `ReadRows` handling in ISCP job loading
  - nil `JobStatus` permanent-error flushing
  - single-row task-runner readers across batches
  - incrservice offset/step readers across batches
  - ISCP `getTableID()` across batches
  - feature-upsert single-row readers across batches
  - bootstrap/version single-row helper behavior

### Why this matters
The visible CI failure happened in vector HNSW BVT, but the root cause was a
broader internal-reader bug class. Fixing only the panic site would still leave
other internal paths batch-sensitive and fragile. This PR closes the root cause
and hardens the nearby production readers that use the same pattern.

### DIFF
--- a/pkg/bootstrap/service_test.go
+++ b/pkg/bootstrap/service_test.go
@@ -412,7 +412,7 @@ func newBootstrapStringResult(values ...string) executor.Result {
 	memRes := executor.NewMemResult(
 		[]types.Type{types.New(types.T_varchar, 2, 0)},
 		mpool.MustNewZero())
-	memRes.NewBatch()
+	memRes.NewBatchWithRowCount(len(values))
 	executor.AppendStringRows(memRes, 0, values)
 	return memRes.GetResult()
 }
@@ -455,7 +455,7 @@ func TestDoCheckUpgrade(t *testing.T) {
 					memRes := executor.NewMemResult(
 						[]types.Type{types.New(types.T_varchar, 2, 0)},
 						mpool.MustNewZero())
-					memRes.NewBatch()
+					memRes.NewBatchWithRowCount(1)
 					executor.AppendStringRows(memRes, 0, []string{bootstrappedCheckerDB})
 					return memRes.GetResult(), nil
 				}
@@ -470,7 +470,7 @@ func TestDoCheckUpgrade(t *testing.T) {
 					memRes := executor.NewMemResult(
 						typs,
 						mpool.MustNewZero())
-					memRes.NewBatch()
+					memRes.NewBatchWithRowCount(1)
 					executor.AppendStringRows(memRes, 0, []string{"1.2.3"})
 					executor.AppendFixedRows(memRes, 1, []uint32{10})
 					executor.AppendFixedRows(memRes, 2, []int32{0})
@@ -517,7 +517,7 @@ func TestDoCheckUpgrade(t *testing.T) {
 					memRes := executor.NewMemResult(
 						[]types.Type{types.New(types.T_varchar, 2, 0)},
 						mpool.MustNewZero())
-					memRes.NewBatch()
+					memRes.NewBatchWithRowCount(1)
 					executor.AppendStringRows(memRes, 0, []string{bootstrappedCheckerDB})
 					return memRes.GetResult(), nil
 				}
@@ -532,7 +532,7 @@ func TestDoCheckUpgrade(t *testing.T) {
 					memRes := executor.NewMemResult(
 						typs,
 						mpool.MustNewZero())
-					memRes.NewBatch()
+					memRes.NewBatchWithRowCount(1)
 					executor.AppendStringRows(memRes, 0, []string{"2.0.0"})
 					executor.AppendFixedRows(memRes, 1, []uint32{1})
 					executor.AppendFixedRows(memRes, 2, []int32{0})
--- a/pkg/bootstrap/versions/readrows_guard_test.go
+++ b/pkg/bootstrap/versions/readrows_guard_test.go
@@ -1,0 +1,71 @@
+// Copyright 2024 Matrix Origin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package versions
+
+import (
+	"testing"
+
+	"github.com/matrixorigin/matrixone/pkg/common/mpool"
+	"github.com/matrixorigin/matrixone/pkg/container/types"
+	"github.com/matrixorigin/matrixone/pkg/container/vector"
+	"github.com/matrixorigin/matrixone/pkg/util/executor"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReadSingleRowAcceptsSingleRow(t *testing.T) {
+	result, mp := newSingleStringResult(t, [][]string{{"v1.0.0"}})
+	defer func() {
+		result.Close()
+		require.Equal(t, int64(0), mp.CurrNB())
+		mpool.DeleteMPool(mp)
+	}()
+
+	var got string
+	loaded, rows := readSingleRow(result, func(cols []*vector.Vector) {
+		got = cols[0].GetStringAt(0)
+	})
+	require.True(t, loaded)
+	require.Equal(t, 1, rows)
+	require.Equal(t, "v1.0.0", got)
+}
+
+func TestReadSingleRowCountsRowsAcrossBatches(t *testing.T) {
+	result, mp := newSingleStringResult(t, [][]string{{"v1.0.0"}, {"v2.0.0"}})
+	defer func() {
+		result.Close()
+		require.Equal(t, int64(0), mp.CurrNB())
+		mpool.DeleteMPool(mp)
+	}()
+
+	var got string
+	loaded, rows := readSingleRow(result, func(cols []*vector.Vector) {
+		got = cols[0].GetStringAt(0)
+	})
+	require.True(t, loaded)
+	require.Equal(t, 2, rows)
+	require.Equal(t, "v1.0.0", got)
+}
+
+func newSingleStringResult(t *testing.T, batches [][]string) (executor.Result, *mpool.MPool) {
+	t.Helper()
+
+	mp := mpool.MustNewZero()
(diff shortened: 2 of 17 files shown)


XuPeng-SH:
Reviewer note on the original CI symptom:

- The reported `vector_hnsw.sql` row-207 failure (`backend connection closed`) was not the first fault.
- The earlier root cause was an ISCP multi-job status read that treated `executor.Result.ReadRows` as a whole-result callback instead of a per-batch callback.
- That path then hit a nil-`JobStatus` permanent-error flush and crashed a CN.

This PR fixes that primary crash chain first, then hardens the nearby production readers that relied on the same `ReadRows` assumption.

XuPeng-SH:
Request changes.

Blocking findings:

1. `pkg/iscp/iteration.go:673` adds `errors.New("permanent error")`. This is a hard CI blocker: SCA fails with `Use moerr instead` and points exactly to this line. Please remove the stdlib error sentinel. Since this is internal control flow after a permanent-error flush, the cleaner fix is to return an explicit control result, for example `permanentFlushed bool, err error`, instead of encoding control state in an error string/sentinel. If an error still needs to be surfaced, use the MO error pattern and avoid `err.Error() == ...` string matching.

2. `pkg/bootstrap/versions/version.go` changes duplicate/single-row handling in a way that regresses the upgrade framework. CI Coverage fails `pkg/bootstrap/TestDoCheckUpgrade`, and I reproduced it locally on this PR head:

```text
failed to get latest version {"error": "internal error: unexpected rows count: 3"}
service_test.go:448: Expected panic but did not occur
service_test.go:510: Expected panic but did not occur
```

The same test passes on current `origin/main`. This is not safe to fix by only changing the test expectation: the old behavior lets `doCheckUpgrade` load the latest version and enforce the fail-fast upgrade conflict checks; the PR now returns a generic `GetLatestVersion` error before those checks, turning a fatal compatibility guard into a retryable/logged error path. Please preserve the upgrade conflict semantics while making the ReadRows counting correct. If the existing `executor.MemResult.NewBatch()` row-count behavior is part of the trigger, either fix the affected mocks to use `NewBatchWithRowCount` or make the reader robust, but do not mask the upgrade invariant.

3. CI is still broadly red. Besides SCA, Coverage has real UT failures; Compose/BVT/basic-benchmark jobs also cannot get MO ready on port 6001. After fixing the two code issues above, rerun CI and inspect `mo-service.log` if those startup failures remain.

What looks good:

- The main ISCP direction is right: `GetJobSpecs` now accumulates rows across `ReadRows` callbacks and maps results by `(job_name, job_id)` instead of assuming callback rows are the whole result and in the requested order.
- `normalizeJobStatuses` closes the nil-`JobStatus` permanent-error panic path.
- The new multi-batch tests for ISCP/publication/incrservice/watermark/bootstrap/feature-upsert are useful.

Before merge, I would also add unhappy-path tests for `GetJobSpecs` duplicate rows, unexpected rows, invalid `job_spec`/`job_status`, and flush failure propagation, plus an `ExecuteIteration` assertion that only a successfully flushed permanent error is swallowed.

Local verification:

- `git diff --check origin/main...HEAD`: pass
- Targeted new ReadRows tests: pass for `pkg/iscp`, `pkg/publication`, `pkg/incrservice`, `pkg/bootstrap/versions`, `pkg/sql/plan/function`
- `go test ./pkg/bootstrap -run TestDoCheckUpgrade -count=1`: fails on PR head, passes on `origin/main`

fengttt:
## Review — one blocking bug found

The design intent is sound and most of the diff is correct, but this PR **breaks cluster bootstrap for every test** (all e2e BVT / UT / Coverage / Upgrade jobs fail with the same panic). I reproduced it locally, bisected to a single file, and verified a one-line fix.

### 🔴 Blocker — variable shadowing in `pkg/incrservice/store_sql.go`

```go
// inside the ExecTxn(...) closure; `var current, next, step uint64` is declared ABOVE the closure
current, step, rows := readSingleOffsetStep(res)   // ':=' SHADOWS the outer current & step
```

Because this `:=` sits inside the closure and introduces a new name (`rows`), Go creates **closure-local** `current`/`step` that shadow the outer ones. The old inline callback *assigned* to the outer variables; the extracted helper's return values now land in shadows that die with the closure.

After the closure returns, the outer `current`/`step` are still `0`:

```go
from, to := getNextRange(current, next, int(step))   // getNextRange(0, 10000, 0)
```

Instrumented, the allocator runs with **`curr=0, step=0`** (only `next` survives, since it isn't on the `:=` LHS). The zero step is the poison — the auto-increment machinery then emits `9223372036854775808` (2^63 = MaxInt64+1), which overflows the signed bigint insert and panics `bootstrap system init` (`cnservice.handleBootstrapErr`, `bootstrap/service.go:314`).

**Proof (local):**
- `go test -run TestBasicCluster ./pkg/embed` panics with the exact CI error `data out of range: data type bigint, value 9223372036854775808`.
- Reverting only `store_sql.go` to `main` → passes.
- Restoring the PR file and changing just `:=` → `=` (with a separate `var rows int`) → passes.

**Fix:**
```go
var rows int
current, step, rows = readSingleOffsetStep(res)
```

**Why the new tests missed it:** `TestReadSingleOffsetStep*` exercise the helper in isolation and pass — the defect is in the *caller's* `:=`, not the helper. `go vet` (default) does not flag shadowing.

### ✅ Rest of the PR — reviewed, looks correct
- **`GetJobSpecs` rewrite** (job_name/job_id + `JobKey` map + cross-batch accumulation + `normalizeJobStatuses` nil-guard): sound; the nil-`JobStatus` deref crash is genuinely fixed and well tested.
- **`isPermanentError` → `errors.As(permanentError{})`**: no regression — `JobStatus.SetError` receives `Consume` errors that were never the literal string `"permanent error"` before either, so behavior is preserved.
- **`readSingleRow` / `readSingleResultRow` / `readSingleTaskRunner` / `getTableID`**: all correctly handle empty leading batches (`ReadRows` does not skip zero-row batches) and multi-batch counting; their call sites use `=`/fresh locals — no shadowing. The added `limit 2` in feature-upsert is a good defensive bound.

### 🟡 Minor (non-blocking)
- `readSingleTaskRunner` is duplicated verbatim in `pkg/iscp/util.go` and `pkg/publication/util.go` — fine given package boundaries, but a shared helper would avoid drift.

Everything except the shadowing fix looks good to merge.

Issue #23225, opened 2025-12-05
[Subtask]: Do not manual cleanup idle connections


XuPeng-SH:
### Parent Issue

#23222 

### Detail of Subtask

Rely on http.Transport’s built‑in idle management (e.g., IdleConnTimeout, MaxIdleConns, MaxIdleConnsPerHost) instead of forcibly closing all idle connections on a fixed short interval (PutIdleConn)

PR #25458, opened 2026-07-04
fix(fileservice): avoid closing idle upload connections

### LINKED ISSUE #25350: LOAD DATA fails with COS `server closed idle connection` due to fileservice forcibly closing idle connections
## Summary

LOAD DATA from S3/COS intermittently fails with `ERROR 20101: internal error: http: server closed idle connection` when loading large tables. The root cause is MO fileservice forcibly closing idle HTTP connections at fixed intervals instead of relying on `http.Transport` built-in idle management.

This is a recurrence pattern affecting multiple nightly CI jobs, particularly on the `bmsql_stock` table (TPCC 1000 warehouses, ~1 billion rows) due to its long upload duration.

## Related

- #23225 (Open) — Do not manual cleanup idle connections (root cause subtask)
- #24916 (Closed) — COS fileservice PUT `connection reset` during Load (same pattern, different error)

## Recent Occurrences

| Run | Date | Failed Table | Error |
|---|---|---|---|
| [28528757100](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28528757100/job/84602204151) | 2026-07-01 | tpcc_1000.bmsql_stock | `server closed idle connection` |
| [28319741563](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28319741563/job/83905544422) | 2026-06-29 | tpcc_1000.bmsql_stock | `server closed idle connection` |

## Error

```
ERROR 20101 (HY000) at line 1: internal error: convert go error to mo error
Post "https://mo-nightly-gz-1308875761.cos.ap-guangzhou.myqcloud.com/mo-benchmark-.../data/...?uploadId=...":
http: server closed idle connection
```

## Root Cause

MO fileservice calls `CloseIdleConnections` on the HTTP transport at fixed intervals. When a COS multipart upload is in progress and the connection is momentarily idle between chunks, the forced cleanup closes it mid-upload. Go wraps this as `http: server closed idle connection`.

The `PutIdleConn: CloseIdleConnections was called` log lines observed in other failures (e.g., hive partition dispatch timeout) confirm this pattern across multiple code paths.

## Impact

- TPCC 1000W load: `bmsql_stock` fails → all NEW_ORDER transactions fail with `Stock not found` → tpmC drops to ~1.59 (normal ~10881)
- NONSYS RESTORE: `file_working` load fails → PITR restore cascade failure
- Hive partition: CN-to-CN dispatch timeout correlated with `PutIdleConn` errors

## Expected Behavior

Fileservice should rely on `http.Transport` built-in idle connection management (`IdleConnTimeout`, `MaxIdleConns`, `MaxIdleConnsPerHost`) rather than forcibly closing all idle connections. Active uploads should not be interrupted.

## Environment

- Cluster: TKE Guangzhou nightly, 3 CNs + 1 DN
- COS: `mo-nightly-gz-1308875761.cos.ap-guangzhou.myqcloud.com`
- Branch: main

LeftHandCold:
## What type of PR is this?
- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25350
Related to #23225
Related to #24916

`LOAD DATA` from COS can intermittently fail with `http: server closed idle connection` during long multipart uploads. The fileservice HTTP transport already has `IdleConnTimeout` and connection limits, but the background metrics cleanup loop was also calling `CloseIdleConnections()` every 5s, which can close an idle upload connection between multipart requests.

Changes:
- stop periodically calling `httpTransport.CloseIdleConnections()` from the fileservice background loop
- keep stale connection metric cleanup, using `IdleConnTimeout + 2s` as the stale metric threshold
- treat `server closed idle connection` as retryable so existing COS retry wrappers can retry this transient transport error
- add retryability coverage for the COS error string

Validation:
```text
go build ./pkg/fileservice
go vet ./pkg/fileservice
go test -v -count=1 ./pkg/fileservice

PASS
ok  	github.com/matrixorigin/matrixone/pkg/fileservice	27.250s
```

### DIFF
--- a/pkg/fileservice/error.go
+++ b/pkg/fileservice/error.go
@@ -75,6 +75,7 @@ func IsRetryableError(err error) bool {
 		strings.Contains(str, "connection timed out") ||
 		strings.Contains(str, "dial tcp: lookup") ||
 		strings.Contains(str, "i/o timeout") ||
+		strings.Contains(str, "server closed idle connection") ||
 		strings.Contains(str, "write: broken pipe") ||
 		strings.Contains(str, "TLS handshake timeout") ||
 		strings.Contains(str, "replication in progress") || // HDFS error
--- a/pkg/fileservice/http_client.go
+++ b/pkg/fileservice/http_client.go
@@ -118,16 +118,14 @@ func trackConnActive(conn net.Conn) {
 }
 
 // cleanupStaleConnections periodically cleans up connections that have been
-// inactive for a long time (likely returned to idle pool or closed)
-// This ensures the active connection count stays accurate
-// staleThreshold should be aligned with the cleanup interval to catch connections
-// that were closed by CloseIdleConnections() but still in activeConnMap
+// inactive for a long time (likely returned to the idle pool or closed).
+// This keeps the active connection metric bounded without forcing the
+// transport to close idle connections while long uploads are in flight.
 func cleanupStaleConnections() {
 	now := time.Now()
-	// Use a threshold slightly longer than the cleanup interval (5s) to catch
-	// connections that were closed by CloseIdleConnections() but still tracked
-	// This ensures metrics stay accurate after CloseIdleConnections() is called
-	staleThreshold := 8 * time.Second // Slightly longer than cleanup interval (5s)
+	// Use a threshold slightly longer than IdleConnTimeout. http.Transport owns
+	// the real connection lifecycle; this only evicts stale metric entries.
+	staleThreshold := idleConnTimeout + 2*time.Second
 
 	activeConnMap.Range(func(key, value interface{}) bool {
 		lastActive, ok := value.(time.Time)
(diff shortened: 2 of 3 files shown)


XuPeng-SH (approved):
Reviewed from first principles and unhappy-path angles.\n\nI checked the functional closure around the COS idle upload failure: the periodic CloseIdleConnections call was the active source of mid-upload idle-connection churn, while Go's Transport still bounds and expires idle connections through IdleConnTimeout / MaxIdleConns / MaxIdleConnsPerHost / MaxConnsPerHost after this change. The retryable classification for "server closed idle connection" is also acceptable in this fileservice path because QCloud write retries replay data through memory / seek / multipart buffers, and the read-side retry paths recreate range readers.\n\nUnhappy-path review covered leak/hung/OOM risk, stale metric cleanup, custom vs global transports, upload retry safety, and active connection metric behavior. I did not find a blocking issue.\n\nLocal verification:\n- git diff --check origin/main...HEAD\n- go build ./pkg/fileservice\n- go vet ./pkg/fileservice\n- go test -v -count=1 ./pkg/fileservice\n\nCI is green as well.

XuPeng-SH (approved):
LGTM.

The main behavior change looks right to me: this stops force-closing idle fileservice HTTP connections every 5s and lets the transport's own idle policy manage them, which is exactly the safer direction for long-running or intermittent COS uploads. Treating `http: server closed idle connection` as retryable also matches the transient transport failure mode we want to mask.

Non-blocking follow-up: it would still be nice to add a multipart `UploadPart` retry / retry-exhaust regression case in `pkg/fileservice/parallel_sdk_test.go`, but I don't see a substantive blocker in the patch itself.

PR #25464, opened 2026-07-06
fix(function): reject out-of-range power results

### LINKED ISSUE #25308: [Bug]: POW with negative base and fractional exponent returns NaN instead of MySQL error
### MatrixOne Version

Latest `origin/main` verified locally at `200bfa7e0efb`.

### Reproduce Steps

```sql
set session sql_mode = '';

select pow(-2, 0.5) as pow_domain;
```

### Expected Behavior

MySQL rejects the out-of-domain result:

```text
ERROR 1690 (22003): DOUBLE value is out of range in 'pow(-(2),0.5)'
```

### Actual Behavior

MatrixOne returns `NaN`:

```text
pow_domain
NaN
```

### Notes

Found during MySQL compatibility Module 6 exploration for numeric functions. Other domain checks such as `sqrt(-1)`, `log(-2)`, and `log(1, 100)` matched MySQL by returning `NULL`; this issue is specific to `POW`/`POWER` domain handling.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25308

Reject non-finite results from `POW` and `POWER` with an out-of-range error instead of returning `NaN` or infinity.

The root cause was that the shared `Power` implementation appended the result of Go's `math.Pow` without checking for `NaN` or infinity. The implementation now returns `ErrOutOfRange` for non-finite results while preserving finite values and NULL handling.

Regression coverage includes negative bases with fractional exponents, zero raised to a negative exponent, mixed valid/invalid vectors, and distributed SQL cases for both aliases.

Validation:

- `go test -count=1 -timeout 120s ./pkg/sql/plan/function`
- `go test -race -count=1 -timeout 120s -run '^TestPower(OutOfRange)?$' ./pkg/sql/plan/function`
- `mo-tester`: 53/53 statements passed for `func_math_power.test`


XuPeng-SH (approved):
I did not find a blocker in this fix.

The change now rejects non-finite `POW/POWER` results as out-of-range while preserving normal finite values, NULL propagation, and alias behavior. The added select-list guard also matches the existing `FunctionSelectList` semantics, so filtered rows no longer evaluate the invalid branch.

heni02 (approved):

PR #25467, opened 2026-07-06
Fix nullable IN-list runtime semantics

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25346

Follow-up to #25427.

The previous zonemap fix made block pruning conservative for nullable IN lists, but the SQL path can still fail before row-level filtering. Planner constant folding converts `IN` list literals into a folded vector. `Vector.InplaceSortAndCompact()` sorts the vector payload without moving the null bitmap, so `('keep', NULL)` can become a vector where the null bitmap no longer points at the same value. The runtime `in` operator then skips the wrong RHS slot while building its hash map, and `WHERE k IN ('keep', NULL)` can filter out matching `keep` rows.

This PR fixes the full path:

- keeps folded nullable IN-list vectors unsorted/uncompacted so values and null bitmap stay aligned;
- makes vectorized `IN` / `NOT IN` return SQL UNKNOWN for no-match rows when the RHS contains NULL;
- keeps nullable folded vectors conservative in zonemap/readutil pruning instead of relying on sorted first/last entries;
- adds tests for operator three-valued logic, folded predicate execution, nullable folded vectors in zonemap pruning, and varchar/fixed-type nullable zonemap `AnyIn`.

Validated with the issue-shaped SQL on a dev node:

```sql
SELECT COUNT(*) FROM t WHERE k IN ('keep', NULL);
-- 2

SELECT COUNT(*) FROM t WHERE k NOT IN ('zzz', NULL);
-- 0
```

Local checks:

```text
go test ./pkg/sql/plan/function -count=1
go test ./pkg/sql/colexec -count=1
go test ./pkg/vm/engine/readutil -count=1
go test ./pkg/vm/engine/tae/index -count=1
go test ./pkg/sql/plan/rule -count=1
go test ./pkg/sql/plan -run 'Test.*Fold|Test.*In|Test.*Filter' -count=1
```


ouyuanning (approved):

XuPeng-SH (approved):
Reviewed the nullable IN-list runtime and pruning fix.

The direction looks correct. The runtime `IN`/`NOT IN` operators now preserve SQL three-valued logic when the RHS contains NULL: matches still return true/false as expected, while no-match rows become UNKNOWN. The folded-vector change also addresses the real root cause: nullable list vectors are no longer sorted/compacted in a way that can desynchronize payload order from the null bitmap.

I also checked the pruning paths. For nullable folded vectors, readutil disables sorted min/max seek shortcuts, and ZM.AnyIn falls back to scanning non-null values. SubVecIn returning the full range for nullable vectors is conservative for bloom filtering; it can add false positives but should not incorrectly prune matching rows.

Local validation:
- go test ./pkg/sql/colexec -count=1
- go test ./pkg/vm/engine/readutil -count=1
- go test ./pkg/vm/engine/tae/index -count=1
- go test ./pkg/sql/plan/rule -count=1
- go test ./pkg/sql/plan -run 'Test.*Fold|Test.*In|Test.*Filter' -count=1

CI is green. I did not find a blocking issue.

PR #25469, opened 2026-07-06
fix(lockservice): release forwarded txn before async wait

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Related to #25466
Related to #25297
Related to #25278
Related to #25279

PR #25466 backported #25447 to 3.0-dev, but `handleForwardLock` still held the active txn mutex until the async callback.

When a forwarded lock request hits an async conflict, `l.lock` returns after queueing the waiter, while the callback is not called yet. Later the waiter worker retries by locking the same txn first, so it can block forever on the mutex that only the callback would release.

This PR aligns `handleForwardLock` with the safe ownership pattern used by `handleRemoteLock`: the handler releases the txn mutex with `defer`, and the async callback no longer owns that unlock.

It also adds a regression test that verifies a forwarded waiter resumes after the holder unlocks.

## Validation:

```text
go test ./pkg/lockservice -run 'TestForwardLockWaiterCanResumeAfterConflict|TestForwardLock|TestDeadLockWithForward|TestHandleForwardLockRejectsWhenServiceCannotLock' -count=1
ok  	github.com/matrixorigin/matrixone/pkg/lockservice	0.485s
```

A full package run on the same lockservice fix state also passed:

```text
go test ./pkg/lockservice -count=1 -timeout=20m
ok  	github.com/matrixorigin/matrixone/pkg/lockservice	568.838s
```

### DIFF
--- a/pkg/lockservice/service_forward_test.go
+++ b/pkg/lockservice/service_forward_test.go
@@ -59,6 +59,53 @@ func TestForwardLock(t *testing.T) {
 	)
 }
 
+func TestForwardLockWaiterCanResumeAfterConflict(t *testing.T) {
+	runLockServiceTests(
+		t,
+		[]string{"s1", "s2"},
+		func(alloc *lockTableAllocator, s []*service) {
+			tableID := uint64(10)
+
+			l1 := s[0]
+			l2 := s[1]
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+			defer cancel()
+
+			_, err := l2.getLockTableWithCreate(0, tableID, nil, pb.Sharding_None)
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
I did not find a blocker in this follow-up fix.

The mutex ownership change now matches the safe `handleRemoteLock` pattern: `handleForwardLock` releases the txn with `defer`, and the async callback no longer owns that unlock. The new regression test also covers the exact waiter-resume path that was hanging in #25466.

Issue #24924, opened 2026-06-11
[Bug] data branch has incorrect behavior in edge table and identifier paths


Ariznawlll:
## Bug

Data branch currently has several user-visible behavior bugs around cloned view SQL rewriting, quoted identifiers, and edge table/column paths found by the coverage BVT.

### 1. `DATA BRANCH CREATE DATABASE` rewrites string literals in cloned views

`rewriteCloneViewInfos` rewrites the whole `CREATE VIEW` SQL text with `strings.ReplaceAll(info.createSql, srcDBName, dstDBName)`. This changes string literals, not only real database/table references.

Reproduce:

```sql
drop database if exists br_view_literal_copy;
drop database if exists br_view_literal_src;
create database br_view_literal_src;
use br_view_literal_src;
create table t(a int primary key);
insert into t values (1);
create view v_literal as select 'br_view_literal_src' as marker, a from t;
data branch create database br_view_literal_copy from br_view_literal_src;
select marker from br_view_literal_copy.v_literal;
```

Current result:

```text
br_view_literal_copy
```

Expected result:

```text
br_view_literal_src
```

Only real table/database references in the view definition should be rewritten. String literals must not be modified.

### 2. `DATA BRANCH CREATE TABLE` should accept quoted identifiers containing spaces

Legal quoted identifiers such as `` `branch table` `` should work. The branch-create path parsed raw SQL text with a regexp using `\S+`, so names containing spaces could be split incorrectly.

Reproduce:

```sql
drop database if exists br_matrix_quoted;
create database br_matrix_quoted;
use br_matrix_quoted;
create table `base table`(a int primary key, b varchar(20));
insert into `base table` values (1, 'space-name'), (2, 'space-name');
data branch create table `branch table` from `base table`;
select count(*) from `branch table`;
```

Expected: the branch table is created and contains the cloned rows.

### 3. Branch-protect snapshot metadata should escape apostrophes in identifiers

When data branch creates the branch-protect snapshot row in `mo_snapshots`, source account/database/table names are embedded into SQL string literals. Legal identifiers containing an apostrophe can break the internal insert SQL unless the string literals are escaped.

Reproduce:

```sql
drop database if exists br_matrix_quoted;
create database br_matrix_quoted;
use br_matrix_quoted;
create table `quote'src`(a int primary key, b varchar(20));
insert into `quote'src` values (1, 'single-quote');
data branch create table `quote'dst` from `quote'src`;
select count(*) from `quote'dst`;
[...]

PR #25471, opened 2026-07-06
fix data branch edge cases

gouhongshen:

This PR completes Data Branch edge-case handling and the related CDC change-stream correctness work: it preserves semantic identifiers while cloning/restoring objects, handles snapshot/PITR and DIFF/MERGE boundaries safely, keeps complex values representable, and makes partitioned CDC snapshots and tails bounded and commit-order correct. The changes close data-loss, malformed-SQL, stale-metadata, and unbounded-memory risks that arise at those boundaries.

## Summary
- Completes Data Branch edge-case coverage across clone, pick, snapshot/PITR, DIFF/MERGE, parser/planner/DDL, and branch hash-map handling.
- Replaces clone-view text substitution with AST-based identifier remapping, preserving aliases, nested query structure, literals, quoted identifiers, and case-sensitive names.
- Completes the clone restore closure by quoting database and view identifiers through the shared `sqlquote.Ident` helper.
- Completes AST remapping of `FullTextMatchExpr.Pattern`; a direct-AST regression keeps the walker correct even though current SQL grammar admits only literal or parameter `AGAINST` patterns.
- Preserves branch data values and tuple decoding for Decimal256 and ENUM, and keeps binary, geometry, datalink, UUID, and array values representable in generated branch SQL.
- Aligns DIFF and MERGE transaction behavior with explicit user transactions and the current background-session rules.
- Streams partitioned initial CDC snapshots one child chunk at a time, preventing their shared synthetic timestamp from coalescing the entire snapshot into memory.
- Extends partitioned tail `CollectChanges` with a heap-frontier, cursor-based commit-timestamp merge: each timestamp is emitted only after all partitions contribute, without prefix copying or per-timestamp full-partition scans.
- Replays CDC delete rows before replacement rows within that boundary, preserving the final row for partition-key moves that keep the same primary key.
- Adds unit and distributed BVT coverage for branch edge cases, tuple types, clone remapping, explicit transactions, bounded partition snapshots, partition stream ordering, and CDC operation order.
- Updates clone/geometry formatter assertions to the canonical quoted-identifier and `st_geomfromtext` output used by the implementation.

## Upstream sync and conflict resolution
Merged `upstream/main` through `83a090b5d1`.

The two semantic conflicts both involved SQL string-literal construction. The resolution keeps the upstream package-level `quoteSQLStringLiteral` helper and routes the PR's branch-protect snapshot and foreign-key SQL through it, preserving apostrophe and backslash handling without duplicate package-local escape paths.

The latest sync had one result-file conflict: upstream changed the BVT table-output layout while this PR corrected cloned-view SQL to quote identifiers. The merged result keeps the upstream layout and the PR's exact `create view \`v1\` as select * from \`t1\`;` expectation.

The current sync had four code conflicts. PITR and snapshot foreign-key discovery now use the upstream shared `quoteSQLStringLiteral` helper. Connector and dispatch collided with upstream #26324, which already provides the central `ResolvePipelineSpoolAbortError` implementation; the earlier duplicate PR-local pipeline implementation is intentionally absent from the final diff.

The latest sync had one test-only conflict in `build_dml_util_test.go`. It keeps this PR's FK string-literal escaping regressions and upstream #26330's ALTER COPY FK-transfer regression; the production implementation is upstream's single shared SQL-quoting path.

The subsequent sync through `f6dab28046` merged automatically. It includes upstream snapshot, DDL, parser, and plan changes without replacing this PR's branch/CDC behavior.

The sync through `66142a6f7d` merged automatically. It contains the latest mainline frontend, aggregate, catalog, and plan work; this PR's final diff remains limited to its Data Branch and CDC scope.

The latest sync through `52ea9f874f` had one semantic conflict with upstream #26292 (cyclic database-clone foreign keys). The result retains upstream's same-source-database reference rewrite to the destination schema, including forward references while foreign-key checks are scoped off, and retains this PR's shared `sqlquote.Ident`/`QualifiedIdent` formatting. Thus circular FK metadata points at the cloned database without regressing quoted or special identifiers.

The latest sync through `83a090b5d1` merged automatically. It includes mainline HAKeeper, aggregate, plan/protobuf, and TAE updates without replacing this PR's branch or CDC behavior.

## Validation
- `go test ./pkg/container/types`
- `go test ./pkg/frontend`
- `go test ./pkg/vm/engine/disttae`
- `go test ./pkg/cdc`
- `go vet ./pkg/container/types ./pkg/frontend ./pkg/vm/engine/disttae ./pkg/cdc`
- Focused `-race -count=5` coverage for partition tail ordering, partition moves, bounded snapshots, and close-error cleanup.
- Fresh post-conflict `go test -count=1 ./pkg/frontend ./pkg/sql/plan`
- Fresh post-conflict `go vet ./pkg/frontend ./pkg/sql/plan`
- `git diff --check`

## Latest review closure
- Quotes the database and view identifiers in `restoreViews` with the shared `sqlquote.Ident` helper, including embedded backticks and whitespace.
- Recursively remaps `FullTextMatchExpr.Pattern`; direct-AST coverage verifies a qualified pattern remaps from `src.docs.pattern` to `dst.docs.pattern`.
- The current `search_pattern` grammar accepts only `STRING | VALUE_ARG`; `AGAINST(src.docs.pattern)` is therefore not a valid user SQL input today.
- Replaces the partition tail's repeated prefix `Shrink` and full-partition timestamp scan with a heap frontier and per-source cursors. The 8K regression uses eight interleaved partitions and checks every emitted commit timestamp and row.
- Coalesces consecutive pure insert-only or delete-only timestamps into bounded `Tail_wip` ranges (at most `objectio.BlockMaxRows`, 8192 rows), so an 8K pure range produces one sink boundary instead of 8K commands. A timestamp containing both operations, or a change between operation kinds, remains a `Tail_done` boundary because the sink applies all deletes before all inserts and must not invert a later delete over an earlier insert for the same primary key.
- Builds MySQL tail INSERT SQL from the `AtomicBatch` ordered and deduplicated row iterator rather than once per source batch. This keeps the temporal order, preserves the existing `maxSQLSize` bound, and prevents an 8K pure range from becoming 8K downstream `ExecSQL` calls.

## Latest focused validation
- `go test -count=1 ./pkg/frontend -run 'TestApplyRemapDbExpressionContainers|TestRemapDbInFullTextMatchPattern|Test_restoreViews'`
- `go vet ./pkg/frontend`
- `git diff --check`

## Latest CDC sink coalescing validation
- A reverse-appended atomic range generates one SQL statement in commit-timestamp order.
- An 8K range deliberately retains 8K source batches but has exactly one MySQL executor expectation; duplicate `(TS, PK)` source rows also execute once, matching `AtomicBatch`'s existing deduplication contract.
- `go test -count=1 ./pkg/cdc ./pkg/vm/engine/disttae`
- Focused `go test -race -count=1 ./pkg/cdc` for atomic SQL construction, insert/delete handling, and the 8K processor range.
- `go build ./pkg/cdc`, `go vet ./pkg/cdc`, and `git diff --check`.

## Current upstream-sync validation
- `go build ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch ./pkg/frontend`
- `go vet ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch ./pkg/frontend`
- `go test -count=1 ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch ./pkg/frontend`
- `go test -race -count=1 ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch`

## Current CDC merge validation
- `go test -v -count=1 ./pkg/vm/engine/disttae -run 'TestCombined(TxnTable|Changes)'`
- `go test -race -count=1 ./pkg/vm/engine/disttae -run '^TestCombinedChangesHandleTailManyCommitTimestamps$'`
- `go test -count=1 ./pkg/vm/engine/disttae`
- `go test -run '^$' -bench BenchmarkCombinedChangesHandleTail8KDistinctCommitTimestamps -benchtime=1x ./pkg/vm/engine/disttae`
- `go vet ./pkg/vm/engine/disttae`

## Current CI closure
- Removes the unconditional loop in `loadPendingPartition`, resolving SCA/staticcheck `SA4004` without changing the one-read, empty-batch cleanup, or close semantics.
- The latest Ubuntu UT failure is `pkg/tests/txnexecutor.TestPreparedParams`, not a CDC or Data Branch assertion. Its log shows the embedded cluster's HAKeeper/raft request timeouts and connection-ID allocation failure. The branch already contains mainline `#25869`, which prevents a caller deadline from closing the shared HAKeeper client; no PR-local workaround is added for this non-reproducing cluster interruption.
- After the latest main sync, `TestPreparedParams` passed three consecutive ordinary runs and three consecutive race-detector runs.
- `go build ./pkg/vm/engine/disttae ./pkg/sql/plan ./pkg/frontend ./pkg/sql/compile`
- `go vet ./pkg/vm/engine/disttae ./pkg/sql/plan ./pkg/frontend ./pkg/sql/compile`
- `go test -count=1 ./pkg/vm/engine/disttae ./pkg/sql/plan ./pkg/frontend ./pkg/sql/compile`
- Focused `-race` 8K partition-tail regression and `git diff --check`.

## Latest conflict validation
- `go test -count=1 ./pkg/frontend -run '^(TestCloneFkTableOrder|TestBackSessionInheritsForeignKeyChecks)$'`
- `go test -count=1 ./pkg/tests/issues -run '^TestIssue26111DataBranchDatabaseWithCyclicForeignKeys$'`
- `go vet ./pkg/sql/plan ./pkg/frontend ./pkg/tests/issues`
- `git diff --check`

## Latest CDC range validation
- 8K interleaved pure inserts emit exactly one `Tail_done`; a two-block input emits exactly two bounded ranges.
- Pure deletes coalesce; an insert followed by a delete remains two `Tail_done` boundaries.
- `DataProcessor` receives 8K `Tail_wip`/final-`Tail_done` insert batches and sends exactly one sink output with 8K rows.
- `go test -count=1 ./pkg/vm/engine/disttae ./pkg/cdc`, focused `-race`, `go build`, `go vet`, and `git diff --check`.

## Latest upstream conflict resolution
- Merged `upstream/main` through `2e68774061`, including #25882's Data Branch schema-evolution work.
- Preserved both clone paths' required imports: the PR's SQL parser and mainline's source-row lock operator.
- In internal PITR creation, retained this PR's shared SQL-literal escaping while using mainline's publication barrier and post-lock current object ID. This prevents both malformed metadata SQL and a COPY-ALTER/PITR generation race.
- Retained both independent regression suites: string-literal escaping and publication-barrier/current-ID tests.

## Latest conflict validation
- `go test -count=1 -timeout 240s ./pkg/frontend ./pkg/sql/compile`
- `go test -count=1 -race ./pkg/sql/compile -run 'Test(PitrInternalSQLEscapesStringLiterals|PreparePitrPublicationSerializesCopyAlter|ResolveCurrentPitrObjectIDRefreshesPlannedTableID)$'`
- `go vet ./pkg/frontend ./pkg/sql/compile` and `git diff --check`.


XuPeng-SH (changes requested):
I found one real blocker in the clone-view rewrite path.

`rewriteCloneViewInfos()` only enables `WithQuoteIdentifier()` for system databases, and `rewriteCloneCreateSQL()` otherwise re-formats the parsed `CREATE VIEW` AST without any identifier quoting. In the formatter, `UnresolvedName.Format()` and `WriteIdentifier()` emit raw identifier text when `quoteIdentifier` is false. That means a normal user database clone still breaks as soon as the view definition depends on quoted identifiers (apostrophes, spaces, reserved words, backticks, etc.).

A minimal example is a cloned view like `create view `quotedb`.`quotev` as select `quotecol` from `quotedb`.`quotet``: the rewrite currently turns it into SQL shaped like `create view clonedb.quotev as select quotecol from clonedb.quotet;`, which no longer parses. I reproduced that locally with a tiny `rewriteCloneCreateSQL` round-trip test in `pkg/frontend`.\n\nI think this needs to preserve identifier quoting for ordinary user databases too, not only for system schemas.

XuPeng-SH:
补一个我刚确认下来的 blocker：`branch_hashmap` 新加的 `DECIMAL256` 支持在 build-path 和 re-encode-path 之间不一致。

`encodeValue()` 这边把 `DECIMAL256` 行值按原始 32-byte payload 走了 `EncodeStringType()`，但 `encodeDecodedValue()` 的 `T_decimal256` 分支现在只接受 `types.Decimal256`。这样一来，`Project` / `ProjectConsume` / `PopByEncodedFullValue` / `PopByEncodedFullValueExact` 这些 helper 把已解包的 full-value 再编码时，会拿到 `[]byte` 并直接报 `expected decimal256 value`。

也就是说，PR 修好了某些 `DECIMAL256` tuple/vector 路径，但这些 branch-hashmap helper 反而会在带 `DECIMAL256` 的行上重新炸掉。建议这里至少接受 `[]byte` + `types.Decimal256` 两种形态，或者给 `DECIMAL256` 一个能 round-trip 成 typed value 的专用 tuple 编码。

gouhongshen:
Addressed the clone-view identifier quoting blocker in `f16b769b4c`. The rewrite now formats cloned view SQL with `WithQuoteIdentifier()` for ordinary user databases as well, and adds a round-trip test covering quoted user DB/table/column names with apostrophes and spaces. Also synced the branch with latest `upstream/main`; there are no remaining merge conflicts.

gouhongshen:
Addressed the DECIMAL256 branch_hashmap re-encode blocker in `be949cdb52`. `encodeDecodedValue` now accepts both typed `types.Decimal256` and raw 32-byte `DecodeRow` payloads without changing global packer/tuple encoding. Added coverage for `Project`, `PopByEncodedFullValue`, `PopByEncodedFullValueExact`, plus direct typed/raw encode tests.

gouhongshen:
Fixed the CI BVT result mismatch in `e84ef26bb8`. The failed jobs were comparing `pessimistic_transaction/clone/clone.sql` `show create table v1;`: cloned views now intentionally show quoted identifiers (`create view `v1` as select * from `t1`;`) after the clone-view quoting fix, so the BVT expected result was updated only for the two cloned-database checks. Synced with latest `upstream/main` as well.

gouhongshen:
## Review conclusion

Blocking issue: `GEOMETRY32` support is incomplete. The PR adds it to SQL value formatting and partition change collection, but the data-branch comparison path still handles only `T_geometry`. `compareValueFromVector` / `normalizeCompareValue` reject `T_geometry32`, so tuple-to-batch comparison or concurrent-update conflict checking fails with `compareSingleValInVector : unsupported type 68`.

Please add `T_geometry32` to both helpers and cover data-branch diff/merge (including conflict comparison) for a GEOMETRY32 column.

The previously reported clone-view quoting and DECIMAL256 re-encoding blockers are fixed in the current head. CI is green and `git diff --check` passed locally.

I attempted to submit **Request changes**, but GitHub does not permit a PR author to request changes on their own PR.

gouhongshen:
Addressed the GEOMETRY32 blocker in `586133a0b0`.

- Added `T_geometry32` to vector extraction and tuple normalization, with direct tuple-to-vector and vector-to-vector coverage.
- Closed SQL write-back as well: `st_geomfromtext(WKT[, SRID])` preserves subtype validation, float32 WKB conversion, and SRID metadata.
- Extended `branch_matrix_coverage` across GEOMETRY32, POINT32, GEOGRAPHY32, and GEOMETRY32 SRID 3857. It covers diff, concurrent-update conflict comparison, and `WHEN CONFLICT ACCEPT` merge.
- Verified `make build`, full `pkg/frontend` tests, vet, and mo-tester 108/108.

The same semantic fix is synced to the 4.1 Draft PR #25570 in `c5ed7d1b3a`.

gouhongshen:
Resolved the new `upstream/main` conflict in `bfb614dd18`.

The overlap was limited to ENUM tuple decoding. The merge keeps this PR's typed `Enum` result while incorporating main's nested type-code fix for `StringifyTuple` and its composite-index tests. Verified `pkg/container/types`, `pkg/frontend`, and `pkg/sql/plan` tests, vet, `git diff --check`, and `make build`.

gouhongshen:
同步 #25570 的 clone-view AST remap review 修复：限定列名及所有相关表达式容器均完成数据库重映射，保留 alias 和字符串字面量；新增对应单元和 clone-view 端到端回归覆盖。验证：pkg/frontend 定向 go test、go build、go vet、git diff --check。

gouhongshen:
已同步 upstream/main 至 efe2379669，解决 data_branch_helpers 和 data_branch_test 的冲突：保留 GEOMETRY32 语义和上游 BINARY VARBINARY BLOB 十六进制覆盖。定向 go test、pkg/frontend build、vet、diff check 均通过。

heni02 (approved):

LeftHandCold (approved):
Reviewed current head d9d688f3b4. I do not see any remaining blocking issue.

I re-checked the data-correctness-sensitive paths: partitioned-table CollectChanges iteration/cleanup, GEOMETRY32 WKB→WKT→ST_GeomFromText write-back (including SRID/subtypes), DECIMAL256 and ENUM tuple/hashmap round trips, PICK key validation/coercion, explicit-transaction DIFF/MERGE behavior, and clone-view AST remapping/identifier quoting. The earlier clone quoting, DECIMAL256 re-encode, and GEOMETRY32 comparison blockers are addressed in the current head.

Local targeted tests across container/types, frontend, databranchutils, sql/compile, parser, plan, and disttae passed with the MatrixOne CGo environment; git diff --check is clean, and the current GitHub CI run is green.

XuPeng-SH (changes requested):
Blocking: the explicit-transaction guard for DATA BRANCH DIFF/MERGE is inverted.\n\nAt pkg/frontend/stmt_kind.go:246-254, this PR classifies both statements as unconditionally executable in an uncommitted transaction. The base branch rejects both (MERGE falls through to false, and DIFF also falls through), and the PR description explicitly says these operations must be rejected in explicit transactions.\n\nThis is not protected at the execution layer: handleBranchMerge/diffMergeAgency has no TxnOptions().ByBegin fallback; only DATA BRANCH PICK has that check. Therefore BEGIN; DATA BRANCH MERGE/DIFF now passes the front-door guard and enters the data-branch execution path.\n\nI reproduced this on current head d9d688f3b4 with a focused test that sets OPTION_BEGIN and calls statementCanBeExecutedInUncommittedTransaction. Both subtests fail because the function returns true:\n\n- DataBranchDiff must not bypass the explicit-transaction guard\n- DataBranchMerge must not bypass the explicit-transaction guard\n\nPlease restore the explicit-transaction rejection for both statements and add regression coverage for OPTION_BEGIN. A defense-in-depth ByBegin check at the DIFF/MERGE execution entry would also keep internal/direct callers from bypassing the classifier.

aunjgr (commented):
Re-reviewed current head after the upstream merge. The explicit-transaction DIFF/MERGE blocker is fixed and covered, and I found no remaining code blocker in the reviewed paths. However, the current run has failures in standalone pessimistic BVT and compose pessimistic BVT, with another job still running when checked. Please triage those failures before merge; approve once the current head is green or the failures are confirmed unrelated and rerun successfully.

aunjgr (commented):
Re-reviewed current head after the upstream merge. The explicit-transaction DIFF/MERGE blocker is fixed and covered, and I found no remaining code blocker in the reviewed paths. However, the current run has failures in standalone pessimistic BVT and compose pessimistic BVT, with another job still running when checked. Please triage those failures before merge; approve once the current head is green or the failures are confirmed unrelated and rerun successfully.

aunjgr (approved):
Reviewed head db2158d84. DATA BRANCH DIFF and MERGE are intentionally supported in explicit transactions by issue 24924, and the implementation correctly reuses the session shared transaction so rollback and commit boundaries remain caller-controlled. The other edge-case fixes remain sound. No remaining blockers found.

gouhongshen (commented):
## Codex automated review

Clone view rewriting breaks case-sensitive identifier configurations.

### P1 - Preserve identifier case when reparsing cloned view SQL (pkg/frontend/clone.go:633)
This hard-codes `lower_case_table_names=1` while reparsing the stored CREATE VIEW SQL. The lexer lowercases every database/table identifier in that mode, but `lower_case_table_names` is configurable to 0. For a case-sensitive source such as `SrcDB`, the parsed references become `srcdb`; the remap map is keyed by `SrcDB`, so those references are not remapped and the restored view references a nonexistent/wrong database. Previously the clone path retained the stored identifier spelling. Parse using the active setting (or otherwise preserve case) and cover a lower_case_table_names=0 clone with a view.

gouhongshen (commented):
## Codex automated review

Partitioned change streams are emitted in metadata order rather than commit-timestamp order, risking incorrect CDC/replication results.

### P1 - Preserve commit timestamp order across partition streams (pkg/vm/engine/disttae/txn_table_combined.go:294)
Each partition handle is drained completely before the next begins, while partitions are ordered by metadata rather than commit timestamp. Real logtail batches are marked Tail_done, so CDC sinks each partition batch independently and cannot globally reorder them. If a row moves partitions, a later insert can be emitted before the earlier delete, causing the delete to remove the newly inserted row. Merge partition changes by commit timestamp or preserve a single globally ordered atomic stream, with an interleaved-partition test.

XuPeng-SH (changes requested):
Request changes on exact head 873a303951: one P1 data-correctness blocker remains.

P1 — CollectChanges for a partitioned logical table does not preserve global commit order or cross-partition row atomicity.

combinedChangesHandle.Next drains handles[0] completely before handles[1], but getPartitionTableFunc orders those handles by partition metadata, not commit timestamp. Each child is only sorted within that physical partition, and the combined handle forwards each child's Tail_done unchanged. CDC therefore hands each partition batch to the sinker independently; the sinker executes INSERT/REPLACE before DELETE for each command.

Concrete failure: a partition-key update can place the new-row insert in one partition and the old-row delete in another at the same commit timestamp. If the new partition appears first in metadata, the target executes insert first and the later child deletes that same PK, leaving the row missing. The same mechanism emits a later commit before an earlier commit when partition histories interleave.

I reproduced the ordering violation on this exact head with a deterministic combined-handle regression: child 0 contained commit-ts 20-0 and child 1 contained 10-0. The stream returned 20-0 then 10-0 and failed the monotonic-order assertion. The current TestCombinedTxnTable_CollectChangesIteratesPartitions asserts only partition order, so it cannot catch this.

Please merge partition change streams by commit timestamp while preserving data/tombstone atomicity for one logical transaction, and add an interleaved-partition regression that includes a row moving partitions (delete old partition + insert new partition) through the consumer boundary.

I also rechecked the earlier clone identifier quoting/case, DECIMAL256 re-encode, GEOMETRY32, ENUM, PICK and explicit-transaction findings; those are fixed. Their focused tests pass, CI is green, the PR diff is clean, and current main merge-tree has no conflict.

gouhongshen (commented):
## Codex automated review

Clone/data-type and explicit-transaction topics are fixed or accepted as non-blocking. The prior tail partition-order blocker is addressed, but this follow-up introduces a distinct snapshot-memory regression. Focused Go tests were unavailable because `go` is not installed.

### P1 - Do not coalesce an entire partitioned snapshot into one batch (pkg/vm/engine/disttae/txn_table_combined.go:377)
This loop keeps loading while the next commit timestamp equals the current one. For an initial CDC snapshot, every CheckpointChangesHandle appends the same constant `h.end` timestamp to each chunk, so the combined handle drains every partition and every snapshot chunk before returning. The previous handle returned each child chunk immediately, and the CDC processor explicitly supports snapshots spanning multiple batches. A large partitioned table can therefore accumulate the entire snapshot in memory and OOM the CN/CDC process. Restrict same-timestamp coalescing to tail data (or preserve bounded snapshot batches) and add a multi-batch snapshot regression test; this is distinct from the prior tail-ordering finding.

gouhongshen (commented):
## Codex automated review

The prior partitioned-snapshot OOM blocker is fixed: initial snapshots now return one checkpoint chunk at a time while preserving cleanup and snapshot hints. The earlier cross-partition commit-order/atomicity blocker remains fixed. Clone quoting/case, DECIMAL256, and GEOMETRY32 blockers are fixed; explicit-transaction DIFF/MERGE was correctly withdrawn because support is the documented contract. No new blocking or non-blocking defect found. `git diff --check` passed; focused Go tests could not run because `go` is unavailable in this environment.

XuPeng-SH (changes requested):
I still see one merge-blocking closure gap on the current head:

**Clone / restore still fails for quoted view names even though the CREATE rewrite now supports them.**
The PR adds quoted-identifier coverage in `rewriteCloneCreateSQL` (`pkg/frontend/clone_test.go:208-220`) and threads the rewritten view definitions into `restoreViews` (`pkg/frontend/clone.go:560-571`). But `restoreViews` still executes `drop view if exists ` + `tblInfo.tblName` without quoting the identifier (`pkg/frontend/snapshot.go:1387`). For a source view like `` `quote view` ``, that becomes `drop view if exists quote view`, which is not a valid single identifier and aborts the clone before the rewritten `CREATE VIEW` ever runs.

Please quote / escape the identifier in the DROP path with the same identifier formatter used for the rewritten CREATE SQL so the restore closure handles quoted and case-sensitive view names end-to-end.

XuPeng-SH (commented):
Latest-head update from another deep pass: I still see two current blockers on this PR.

1. **Quoted view names still fail in clone/restore because the pre-drop stays unquoted.** `handleCloneDatabaseWithSource` now rewrites quoted identifiers for `CREATE VIEW` and passes the rewritten definitions into `restoreViews` (`pkg/frontend/clone.go:560-571`), but `restoreViews` still executes `drop view if exists ` + raw `tblInfo.tblName` (`pkg/frontend/snapshot.go:1387`). `tblInfo.tblName` is still populated from the bare `SHOW FULL TABLES` result (`snapshot.go:1916-1919`). So a valid source view like `` `quote view` `` still reaches `drop view if exists quote view` and fails before the rewritten `CREATE VIEW` can run.

2. **`AGAINST(...)` operands are still left on the source database during view remap.** The new remap walker handles `*tree.FullTextMatchExpr`, but it only remaps `KeyParts` and never descends into `e.Pattern` (`pkg/frontend/remap_db.go:393-400`). `FullTextMatchExpr.Pattern` is a normal expression that is formatted into `AGAINST(...)` (`pkg/sql/parsers/tree/expr.go:1931-1941`) and bound like any other expression (`pkg/sql/plan/base_binder.go:2298-2305`). So a cloned view can still rewrite `MATCH(dst.docs.body)` while leaving a qualified source-db reference or subquery inside `AGAINST(...)`, which either keeps reading the old database or fails after the source is removed.

My current latest-head conclusion is: the quoted-view restore closure gap still remains, and the new fulltext remap path also needs to descend into `AGAINST(...)` patterns before this PR is ready.

jiangxinmeng1 (changes requested):
> Request changes on the current head 0f78cd98d4. Two clone correctness gaps
  > remain:
  >
  > 1. restoreViews still constructs USE and DROP VIEW statements from raw
  >    identifier strings (pkg/frontend/snapshot.go:1383-1387). A view named `quote
  >    view` produces DROP VIEW IF EXISTS quote view, which is invalid, and a
  >    database name containing a backtick also breaks the USE statement. Please
  >    quote both identifiers with sqlquote.Ident and add an end-to-end restore
  >    test, rather than testing only the rewritten CREATE VIEW.
  >
  > 2. The FullTextMatchExpr branch in remapDbInExpr remaps only KeyParts and never
  >    is an expression and is bound as the AGAINST(...) operand. For example,
  >    cloning a view containing MATCH(src.docs.body) AGAINST(src.docs.pattern)
  >    rewrites the table and match column to the destination database while
  >    leaving the pattern reference on src. Please call remapDbInExpr(e.Pattern,
  >    remap) and add coverage for a qualified expression inside AGAINST(...).
  >
  > I did not find another clear blocker beyond these two findings.

gouhongshen:
@jiangxinmeng1 已在 `27e7ab54d1` 修复并校正回归覆盖：

- `restoreViews` 的 `USE` 与 `DROP VIEW` 统一使用 `sqlquote.Ident`；回归测试走真实恢复命令序列，覆盖库名内嵌反引号和带空格的视图名。
- `remapDbInExpr` 递归处理 `FullTextMatchExpr.Pattern`，并以直接构造 AST 的用例验证限定名从 `src.docs.pattern` 重映射到 `dst.docs.pattern`。

事实边界：当前 MySQL 语法的 `search_pattern` 仅接受 `STRING | VALUE_ARG`，因此 `AGAINST(src.docs.pattern)` 不是可解析的用户 SQL，不能作为 clone 的实际输入。保留该递归是因为 AST 字段声明为 `Expr`；直接 AST 回归能在该语法将来扩展或构造 AST 时防止 walker 漏掉该子节点。

验证：`go test -count=1 ./pkg/frontend -run 'TestApplyRemapDbExpressionContainers|TestRemapDbInFullTextMatchPattern|Test_restoreViews'`、`go test -count=1 ./pkg/frontend`、`go vet ./pkg/frontend`、`git diff --check`。

决策记录：没有采纳此前“在显式事务中拒绝 DIFF/MERGE”的建议，因为 #24924 第 8 项明确要求二者复用 session transaction，并由调用者的 rollback/commit 决定边界；该结论也已在后续 review 中确认。

gouhongshen:
@XuPeng-SH @jiangxinmeng1 Full review closure is now on `27e7ab54d1`:

- Clone identifier quoting/case preservation, Decimal256 re-encoding, Geometry32 comparison/write-back, partition-tail ordering/atomicity, and bounded partition snapshots are covered by the earlier fixes and regressions.
- `restoreViews` now quotes both `USE` and `DROP VIEW` identifiers; the regression executes the restore sequence with an embedded backtick in the database name and whitespace in the view name.
- `FullTextMatchExpr.Pattern` is recursively remapped and directly covered at the AST boundary. Current SQL grammar accepts only `STRING | VALUE_ARG` patterns, so the previously suggested qualified `AGAINST(...)` SQL is not a valid user input today.
- The prior request to reject explicit-transaction DIFF/MERGE remains intentionally declined: #24924 specifies the opposite contract, and rollback/commit coverage plus the later approval confirm it.

There are no unresolved inline threads, and all checks for this head are green. Re-requesting review.

gouhongshen:
@XuPeng-SH @jiangxinmeng1 Synced the PR with `upstream/main` through `6004becb56` in `0da39ae70f`.

The two conflicts were SQL literal-escaping overlaps. I kept the upstream package-level helper and routed the PR paths through it, so apostrophe/backslash behavior remains covered without parallel escape implementations. Verified with `go test -count=1 ./pkg/frontend ./pkg/sql/plan`, `go vet ./pkg/frontend ./pkg/sql/plan`, and `git diff --check`.

gouhongshen:
CI repair in 97d29597e9:

The failing pessimistic `dedupjoin.test` first produced the expected duplicate-key error, then a later normal-End cleanup rewrote it to `pipeline end signal delivery failed`. The fix preserves a fatal edge's durable original error in connector/dispatch cleanup, while retaining the existing fallback behavior for a genuinely undeliverable End.

Regression coverage covers the process, connector, and dispatch paths; an isolated execution of the failing SQL now returns `Duplicate entry '1000000' for key ''` without the pipeline delivery error. The PR description has been updated with full scope and validation.

jiangxinmeng1 (approved):
Reviewed the latest head 0da39ae70f. I found no remaining merge-blocking issues.

  The previously reported gaps are now closed: clone-view identifier quoting and
  AST remapping, Decimal256/ENUM/Geometry32 handling, partitioned CDC commit
  ordering and atomicity, bounded snapshot streaming, and change-handle cleanup/
  error paths.

  The visible Iceberg Connector E2E workflow passed, but please confirm all
  required checks are green on the latest head before merging.

XuPeng-SH (changes requested):
I found one production performance blocker on exact head `370aa7b086`.

The new partition-tail merge emits one `Tail_done` per commit timestamp (`txn_table_combined.go:387-451`) and consumes each timestamp by `UnionWindow` followed by deleting the prefix with `Shrink` (`txn_table_combined.go:586-620`). A normal change batch can contain many ordered commit timestamps; in the common single-row-transaction case this repeatedly shifts the remainder of the same batch, making batch processing quadratic. `loadPending`/`nextCommitTS` also scan every partition for every emitted timestamp.

I reproduced this with the real `combinedChangesHandle` on this head using one 8192-row child batch whose rows have distinct ordered commit timestamps (the other partitions being exhausted does not change the hot path). Three one-shot runs were 34.4–35.0 ms, ~5.05 MB and ~57.4k allocations. The scaling is nonlinear: 2048 rows ~3.0 ms, 4096 ~9.7–10.1 ms, and 8192 ~34.5 ms. In addition, each of those 8192 groups is returned as a separate `Tail_done`, so `DataProcessor.processTailDone` sends a separate sink command for every source transaction instead of retaining the original bounded batch granularity.

This is a material CDC regression for partitioned tables with many small transactions. Please preserve global commit order and coalesce same-timestamp rows across partitions without repeatedly mutating/shifting the source batch or scanning all partitions per row. A cursor/window-based k-way merge (heap/frontier) that emits maximal safe ordered ranges would avoid both costs; the sink-side delete-before-insert rule must still be applied only with correct per-PK temporal semantics when a range contains multiple timestamps. Please add a mixed-timestamp 8K-scale benchmark/regression so this does not regress again.

The earlier correctness blockers (quoted restore identifiers, FULLTEXT pattern recursion, cross-partition same-TS atomicity, bounded snapshots, and terminal-cause preservation) are otherwise closed, focused race tests pass, CI is green, and the latest-main merge tree is clean.

gouhongshen:
Addressed in c2ffa82220.

- Replaced the tail merge's repeated prefix `Shrink` and timestamp/partition scans with a heap frontier plus per-source offsets. The initial load seeds one head per data/tombstone stream; later advancement touches only the stream whose source was exhausted. Source batches are released only after their cursor reaches the end.
- Added an 8,192-row regression and benchmark over eight interleaved partitions. The regression verifies every timestamp and row, and also checks that all child handles close. The local one-shot benchmark is 3.03 ms/op (4.66 MB, 73,765 allocs) on Apple M5.
- Extended the partition-move test so two child batches with the same commit timestamp are emitted as one atomic boundary with their cross-partition tombstone.

I intentionally retained one `Tail_done` per commit timestamp. `DataProcessor.processTailDone` creates a sink command per boundary, and `mysqlSinker2.handleInsertDeleteBatch` executes the command's complete delete batch before its complete insert batch. Coalescing a range with multiple timestamps would therefore replay a newer delete before an older insert for the same primary key and can leave stale state. Changing that requires a temporal ordering representation and execution contract in the sink; it is outside this partition-merge fix. The heap/cursor path removes the actual quadratic copying and per-timestamp partition scan while preserving the existing CDC temporal contract.

Validated with focused combined-handle tests, the 8K benchmark, `go test -race -run '^TestCombinedChangesHandleTailManyCommitTimestamps$' ./pkg/vm/engine/disttae`, full `go test ./pkg/vm/engine/disttae`, `go vet ./pkg/vm/engine/disttae`, and `git diff --check`.

XuPeng-SH (changes requested):
Blocking performance issue remains in the latest head.

The heap/cursor rewrite does fix the previous O(T×P) merge CPU cost: locally, the 8,192-timestamp benchmark is now 2.9–5.0 ms/op. However, `combinedChangesHandle.nextTail` still unconditionally returns one `Tail_done` per distinct commit timestamp. The new 8K insert-only regression therefore calls `Next` 8,192 times. The production path maps every one of those boundaries to `ChangeTypeTailDone`, and `DataProcessor.processTailDone` transfers a fresh atomic batch to `sinker.Sink` each time. The author comment also confirms this is one sink command per boundary.

That is a real command/queue/SQL-processing amplification for partitioned CDC, not just an in-memory allocation concern. The ordinary non-partitioned `ChangeHandler` accumulates rows up to `coarseMaxRow`; this path turns a safe insert-only batch into O(distinct commit timestamps) sink commands. Even the in-memory benchmark already allocates about 4.66 MB / 73.7K objects for 8K rows, while downstream command processing is much more expensive.

The temporal-order argument is valid for mixed cross-timestamp delete/insert conflicts, but it does not justify splitting insert-only or delete-only ranges. Please emit maximal safe ordered ranges—at minimum coalesce consecutive insert-only/delete-only timestamps and retain per-TS boundaries only where mixed operations require them, or introduce an ordered operation representation in the sink. Add an end-to-end/counting regression that bounds sink-command growth for the 8K insert-only case, in addition to the merge microbenchmark.

I rechecked head `67cafa0dcfd67f7f4c2daada311c9ff7970b8a57`. CI is green; local tests for cdc, container/types, frontend, compile, parser, plan, and disttae all pass.

gouhongshen:
Fixed in 3e895d2a54 and then merged through upstream/main f6dab28046.

- The heap frontier now keeps a per-timestamp operation mask alongside its entries, so deciding whether the following timestamp is safe to coalesce is O(1); it does not restore the former full-partition scan.
- Consecutive insert-only or delete-only timestamps return `Tail_wip` until a bounded 8,192-row range ends, then one `Tail_done`. The 8K interleaved insert regression now verifies exactly one completed boundary; two blocks verify the range cap.
- A mixed timestamp, or an insert-to-delete transition, remains a `Tail_done` boundary. The sink executes the command's deletes before inserts, so coalescing such a range could make a later delete override an earlier insert of the same primary key. There is a direct regression for that boundary and for pure-delete coalescing.
- Added the downstream `DataProcessor` 8K `Tail_wip` + final `Tail_done` regression: it produces one sink output containing all 8K rows.

Validated with full disttae and CDC package tests, focused race coverage, the 8K benchmark, build/vet of the affected packages, and diff checks.

XuPeng-SH (changes requested):
Request changes on exact head `0f2a8a4f433ae634dcd566fe4d235b9983c686d6`: the new pure-range coalescing reduces the number of `Sink` calls, but it does not close the downstream SQL-processing amplification from my previous review.

`DataProcessor` still calls `AtomicBatch.Append` once for every per-timestamp batch, and `AtomicBatch.Append` retains every source batch in `AtomicBatch.Batches` (`pkg/cdc/types.go:418-455`). The MySQL sinker then iterates that slice and calls `BuildInsertSQL` followed by `ExecSQL` separately for every source batch (`pkg/cdc/sinker_v2.go:722-740`).

I verified this on the current head by extending the new 8,192-row regression: its single sink command has `len(calls[0].insertAtmBatch.Batches) == 8192`, not a bounded/coalesced batch count. Therefore the insert-only case still builds and executes about 8,192 one-row INSERT/REPLACE statements; only the queue-command count changed from 8,192 to 1. The added test checks only `Sink` call count and total rows, so it cannot detect the remaining amplification.

Please coalesce the insert rows into bounded source batches (or build SQL from the ordered atomic rows) and make the regression assert the actual SQL/executor call count, not only the number of `Sink` calls. The new ordering/coalescing tests and the unmodified focused tests pass, CI is green, and the current head is mergeable; this performance closure gap is the remaining blocker I found.

gouhongshen:
@XuPeng-SH Fixed in `4f34d28147`, then synced through `upstream/main` `83a090b5d1` in `181c23daaf`.

The MySQL tail path now builds SQL from `AtomicBatch`'s ordered, deduplicated row iterator instead of iterating `AtomicBatch.Batches`. It therefore retains timestamp order and the existing `maxSQLSize` splitting bound while avoiding one `BuildInsertSQL`/`ExecSQL` per source timestamp batch. The original source batches remain owned by `AtomicBatch` and are still released by `Command.Close` on both success and error paths.

Coverage now proves:

- reverse-appended source batches generate one SQL statement in commit-timestamp order;
- an 8,192-row range deliberately retains 8,192 source batches but has exactly one MySQL executor expectation, so the regression fails if execution returns to per-source-batch SQL; and
- duplicate `(TS, PK)` rows execute once, matching the existing `AtomicBatch.Rows` deduplication contract already used by the other sink paths.

Validated with focused CDC tests and `-race`, then full `go test -count=1 ./pkg/cdc ./pkg/vm/engine/disttae ./pkg/sql/plan`, `go vet ./pkg/cdc ./pkg/sql/plan`, and `git diff --check`.

XuPeng-SH (approved):
Deep review complete on exact head `181c23daaf`. The previous CDC amplification blocker is closed: atomic inserts are built from the ordered/deduplicated AtomicBatch row iterator, so the 8K pure range now stays bounded at both the sink-command and SQL-executor layers while preserving commit order and duplicate semantics. I rechecked clone/remap/quoting, tuple encoding, partition-tail ordering/atomicity, snapshot streaming, and error/cleanup paths; no remaining blocker found. Focused race stress (5 tests ×100), full owning-package race tests, build/vet, and the 8K benchmark all pass. The exact head merges cleanly with current main; CI is still running and was intentionally not awaited.

Issue #25211, opened 2026-06-29
Hive partition JOIN queries intermittently fail with dispatch timeout (dispatch process not registered within 30s)


Ariznawlll:
## Summary

External hive partition JOIN queries intermittently fail with `dispatch process not registered within 30s, remote CN may have failed`. The same two JOIN SQLs fail every time the issue occurs, while other queries in the same test file succeed.

This is a recurring flaky failure — it does not reproduce every run.

## Affected SQL

```sql
-- Query 1: fact JOIN dim
select d.part_name, count(*) as rowcount
from external_hive_partition_2000_large.fact_hive_2000 f
join external_hive_partition_2000_large.dim_part_2000 d
on f.part_id = d.part_id
where d.part_id in (1, 7, 2000)
group by d.part_name order by d.part_name

-- Query 2: fact self-JOIN
select a.part_id, count(*) as rowcount
from external_hive_partition_2000_large.fact_hive_2000 a
join external_hive_partition_2000_large.fact_hive_2000 b
on a.part_id = b.part_id and a.col_uuid = b.col_uuid
where a.part_id in (1, 7)
group by a.part_id order by a.part_id
```

## Error

```
ERROR 20101 (HY000): internal error: dispatch process not registered within 30s, remote CN may have failed
```

## Recent Occurrences

| Run | Date | Job | Branch |
|---|---|---|---|
| [28338073448](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28338073448) | 2026-06-29 | [Query 1Y / 16 Hive Partition](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28338073448/job/83953472996) | main |
| [28133872754](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28133872754) | 2026-06-25 | [BIG DATA TEST](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28133872754/job/83335832278) | main |

## Previous Tracking

- Previously filed as #25152 (closed, merged into #25129)
- #25129 tracks a similar CN dispatch timeout pattern but under TPCC workload
- This issue is filed separately because the hive partition scenario is a distinct test case with its own reproduction pattern

## Loki Evidence (from 2026-06-25 occurrence)

CN logs showed:
- `objectio/cache.go:354: dedup load did not complete` on all 3 CNs
- `compile/remoterunServer.go:545: dispatch process not registered within 30s`
- `GC forced` on one CN
[...]

PR #25475, opened 2026-07-06
fix: retry remote notify before dispatch registration

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25211

fix: retry remote notify before dispatch registration


ouyuanning (approved):

XuPeng-SH (approved):
Reviewed the remote notify/dispatch registration fix.

The direction looks correct: the transient "remote dispatch receiver not registered yet" state is now explicit, it no longer poisons the UUID registration map, and the client retries notify until registration/cancel/timeout. I also checked the RPC moerr path; the error code is preserved through SetMoError/TryToGetMoErr, so retry is not string-matching dependent.

Local validation:
- go test ./pkg/sql/compile -run 'Test_(GetProcByUuid|TryGetProcByUuid)|TestSendNotifyMessageRetriesUntilRemoteDispatchRegistered' -count=1
- go test ./pkg/sql/colexec/dispatch -count=1
- go test ./pkg/common/moerr -count=1
- go test -race ./pkg/sql/compile -run TestSendNotifyMessageRetriesUntilRemoteDispatchRegistered -count=1

CI note: the current Coverage failure appears to be infrastructure/dependency download related ( from the internal GOPROXY while fetching AWS SDK modules), not a test failure from this PR. A rerun should be enough.

Non-blocking suggestion: if this path is hit under high fanout, a fixed 200ms retry interval can synchronize retries and create extra stream churn. Backoff/jitter can be considered later if metrics show pressure, but I do not think it should block this bugfix.

XuPeng-SH:
Clarification for the CI note in my approval review: the Coverage job failed while downloading dependencies from the internal GOPROXY, e.g. `github.com/aws/aws-sdk-go-v2/service/sts@v1.33.1` / `service/s3@v1.68.0` returned `unexpected EOF`. I did not find evidence that the failure was caused by this PR's tests or code changes.

XuPeng-SH (approved):
I did not find a correctness blocker in this fix.

The retry-before-registration flow is coherent: the server now returns an explicit retryable `ErrRemoteDispatchNotRegistered`, the client-side notify loop recreates the sender and retries until registration succeeds or the global registration timeout expires, and the new targeted compile tests pass on the PR head. From an unhappy-path angle, I did not find a leak/hang regression in the implemented logic itself.

There are still a few worthwhile follow-up tests for the retry path (timeout exhaustion, cancellation during retry wait, and a negative case proving non-matching errors do not retry), but I would not block this PR on those alone.

Issue #25357, opened 2026-07-02
[Bug]: TestCompileExternScanParquetRowGroupFanoutValidatesEmptyFileColumnCount fails with lstat no such file


matrix-meow:
## Problem

`TestCompileExternScanParquetRowGroupFanoutValidatesEmptyFileColumnCount` in `pkg/sql/compile/scope_test.go:1006` fails with a filesystem error instead of the expected Parquet validation error:

```
Error: "lstat /tmp/ccCTyJtB.cdtor.c: no such file or directory" does not contain "column count mismatch"
```

The test expects the error message to contain `column count mismatch` (Parquet empty file validation), but gets an `lstat` error for a temporary C compiler file (`/tmp/ccCTyJtB.cdtor.c`).

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-02 | [PR #25266](https://github.com/matrixorigin/matrixone/actions/runs/28565381420/job/84691486520) | UT Test on Ubuntu/x86 |

## Root Cause

The test creates temporary files for Parquet extern scan testing. On GitHub Actions runners, the temp file path may reference CGO compiler temporary files (`ccCTyJtB.cdtor.c`) that get cleaned up before the test accesses them, causing an `lstat` error instead of the expected column count mismatch error.

This is a CI environment issue — the test's temp file handling is fragile under runner resource pressure.

## Suggested Fix

1. Use a dedicated temp directory (`t.TempDir()`) instead of `/tmp/` for test files
2. Or add retry/cleanup logic for temp file creation
3. Or skip the test when CGO temp files are unavailable

PR #25476, opened 2026-07-06
fix: validate empty parquet before row-group fanout

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25357 

fix: validate empty parquet before row-group fanout

### DIFF
--- a/pkg/fileservice/local_etl_fs.go
+++ b/pkg/fileservice/local_etl_fs.go
@@ -446,10 +446,13 @@ func (l *LocalETLFS) List(ctx context.Context, dirPath string) iter.Seq2[*DirEnt
 			if strings.HasPrefix(name, ".") {
 				continue
 			}
-			info, err := entry.Info()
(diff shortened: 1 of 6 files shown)


XuPeng-SH (approved):
I did not find a correctness blocker in this fix.

The changes line up with the stated bug: empty parquet files are validated early during footer scan, row-group fanout now decides based on the number of files that actually contribute row groups, and transient local directory races now skip vanished entries instead of aborting listing. I also rechecked the unhappy-path surface here: no new wait/lifecycle risk stood out, and the targeted `pkg/fileservice` / `pkg/sql/compile` tests pass on the PR head.

The only follow-up I would consider is extra coverage for the empty-file + single-row-group fallback/file-fanout path and the empty-file + @variable/hive-partition combinations, but I would not block this PR on those alone.

Issue #24732, opened 2026-05-31
[Bug]: JSON_CONTAINS unsupported — breaks Prisma JSON where-filters at runtime (v3.0.11 & 4.0-dev)


dengn:
### Summary
`JSON_CONTAINS` is not implemented. Prisma's MySQL connector compiles JSON `where` filters (`{ path, equals }`, `array_contains`, `string_contains`) into `JSON_CONTAINS`, so those filters **fail at runtime**. JSON storage/retrieval, `JSON_EXTRACT`, and `JSON_SET` all work fine — only `JSON_CONTAINS` is missing.

### Steps to Reproduce
```sql
SELECT JSON_CONTAINS('{"tags":["x","y"]}', '"x"', '$.tags');
-- ERROR 20105 (HY000): not supported: function or operator 'json_contains'
```

### Prisma impact
A JSON path filter such as
```js
prisma.page.findMany({ where: { content: { path: '$.type', equals: 'doc' } } })
```
fails at runtime on both versions:
```
ConnectorError(... MysqlError { code: 20105,
  message: "not supported: function or operator 'json_contains'", state: "HY000" })
```

### Note
Related JSON functions `JSON_ARRAY` / `JSON_OBJECT` are also unsupported on v3.0.11 (they DO work on the 4.0-dev line). `JSON_CONTAINS` is unsupported on **both** v3.0.11 and 4.0-dev.

### Expected Behavior
MySQL implements `JSON_CONTAINS`.

---
**Test context** — Found while testing **Prisma ORM (v5.22)** compatibility against MatrixOne. Reproduced on **both**:
- Prod cloud — `8.0.30-MatrixOne-v3.0.11`
- 4.0 dev line — `nightly-0bfda604` (`select version()` → `8.0.30-MatrixOne-v`)

(Results on the older v3.0.5 are intentionally omitted.)

daviszhen:
fixed

Ariznawlll:
已在最新 `main` 完成函数白盒、原始复现和 Prisma 风格 protocol prepared 详细验证，结果通过。

## 验证环境

- MatrixOne commit：`2865ed6cc8a39727b8b274bb52b88fe441d0b30c`
- `select git_version()` 已核对一致
- 对应实现：`2037cc82a` / #25478
- 客户端：Go `database/sql` + `go-sql-driver/mysql v1.8.1`
[...]

PR #25478, opened 2026-07-06
add json_contains

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24732

增加 json_contains函数


XuPeng-SH (changes requested):
Requesting changes for two semantic blockers:

1. `pkg/sql/plan/function/func_builtin_json.go:351-365` + `test/distributed/cases/function/func_json_contains.result:130-138`
   - `jsonScalarsComparable` only treats exact same `TYPE()` or `INTEGER`/`DECIMAL` pairs as comparable. That excludes JSON `DOUBLE`, so `JSON_CONTAINS(1,1.0)`, `JSON_CONTAINS({"a":1},{"a":1.0})`, and `[1,2,3]` vs `[1.0,3.0]` all return `0` and the new BVT locks that in. For MySQL-compatible `JSON_CONTAINS`, numeric JSON scalars need numeric equality across INT/UINT/DOUBLE/DECIMAL, not type-equality except for int/decimal.
   - Suggested fix: reuse a numeric-aware scalar comparator for all numeric JSON types instead of the current `TYPE()` / `INTEGER-or-DECIMAL` gate.

2. `pkg/sql/plan/function/func_builtin_json.go:238-245` + `pkg/container/bytejson/bytejson.go:461-507`
   - The path branch uses `QuerySimpleExist()`, and that helper returns “path not found” once it hits a scalar with an array-index leg. That drops MySQLs scalar `[0]` autowrap semantics for simple paths. Cases like `JSON_CONTAINS(1,1,$[0])` or `JSON_CONTAINS({"a":1},1,$.a[0])` should match, but this implementation would return `NULL`.
   - Suggested fix: either use the normal JSON path extractor for the third argument or extend `QuerySimpleExist()` to honor the same scalar `[0]` autowrap rule.

heni02 (approved):

daviszhen:
> Requesting changes for two semantic blockers:
> 
> 1. `pkg/sql/plan/function/func_builtin_json.go:351-365` + `test/distributed/cases/function/func_json_contains.result:130-138`
>    
>    * `jsonScalarsComparable` only treats exact same `TYPE()` or `INTEGER`/`DECIMAL` pairs as comparable. That excludes JSON `DOUBLE`, so `JSON_CONTAINS(1,1.0)`, `JSON_CONTAINS({"a":1},{"a":1.0})`, and `[1,2,3]` vs `[1.0,3.0]` all return `0` and the new BVT locks that in. For MySQL-compatible `JSON_CONTAINS`, numeric JSON scalars need numeric equality across INT/UINT/DOUBLE/DECIMAL, not type-equality except for int/decimal.
>    * Suggested fix: reuse a numeric-aware scalar comparator for all numeric JSON types instead of the current `TYPE()` / `INTEGER-or-DECIMAL` gate.
> 2. `pkg/sql/plan/function/func_builtin_json.go:238-245` + `pkg/container/bytejson/bytejson.go:461-507`
>    
>    * The path branch uses `QuerySimpleExist()`, and that helper returns “path not found” once it hits a scalar with an array-index leg. That drops MySQLs scalar `[0]` autowrap semantics for simple paths. Cases like `JSON_CONTAINS(1,1,$[0])` or `JSON_CONTAINS({"a":1},1,$.a[0])` should match, but this implementation would return `NULL`.
>    * Suggested fix: either use the normal JSON path extractor for the third argument or extend `QuerySimpleExist()` to honor the same scalar `[0]` autowrap rule.

已经修改

XuPeng-SH (changes requested):
I rechecked the current head and one substantive blocker remains.

The old `1` vs `1.0` incompatibility is fixed, but the new numeric path is now too permissive for mixed integer/float JSON values.

`jsonContainsScalar()` delegates mixed numeric equality to `bytejson.CompareByteJson()`, and that comparator is an ordering helper with float-oriented tolerance and float64 coercion. So distinct numeric values can now be treated as equal in `JSON_CONTAINS`.

Two concrete bad cases are:

- `JSON_CONTAINS('1', '1.000000009')`
- `JSON_CONTAINS('9007199254740993', '9007199254740992.0')`

The first can collapse because the mixed numeric comparison path treats tiny float/int diffs below its precision-loss threshold as equal. The second can collapse because converting large integers to float64 loses precision above 2^53.

That means the PR has traded the old false negatives for new false positives. For containment semantics, that is still a correctness bug. This needs an exact numeric-equality path for JSON numeric containment checks rather than reusing the general ordering comparator.

daviszhen:
> I rechecked the current head and one substantive blocker remains.
> 
> The old `1` vs `1.0` incompatibility is fixed, but the new numeric path is now too permissive for mixed integer/float JSON values.
> 
> `jsonContainsScalar()` delegates mixed numeric equality to `bytejson.CompareByteJson()`, and that comparator is an ordering helper with float-oriented tolerance and float64 coercion. So distinct numeric values can now be treated as equal in `JSON_CONTAINS`.
> 
> Two concrete bad cases are:
> 
> * `JSON_CONTAINS('1', '1.000000009')`
> * `JSON_CONTAINS('9007199254740993', '9007199254740992.0')`
> 
> The first can collapse because the mixed numeric comparison path treats tiny float/int diffs below its precision-loss threshold as equal. The second can collapse because converting large integers to float64 loses precision above 2^53.
> 
> That means the PR has traded the old false negatives for new false positives. For containment semantics, that is still a correctness bug. This needs an exact numeric-equality path for JSON numeric containment checks rather than reusing the general ordering comparator.

已经修改

XuPeng-SH (approved):
Rechecked the latest head after the requested fixes.

The previous semantic blockers are addressed:

- Numeric containment no longer reuses the general ordering comparator for equality. The new path has exact checks across int/uint/float/decimal cases, and the regressions for  vs , large integers above 2^53, uint64 boundaries, and decimal/float cases are covered by UT/BVT.
- The third-argument path flow now uses a JSON_CONTAINS-specific simple-path lookup with scalar  autowrap, without changing the existing QuerySimple/JSON modify behavior. Missing paths still return NULL, and wildcard/non-simple paths error as required.

I also checked the function registration/ID changes, strict/null handling, invalid JSON/path unhappy paths, typed JSON inputs, arrays/objects/scalars, and select-list skip behavior. CI is green, and I locally ran the targeted bytejson and function tests for the new code paths.

Approved.

XuPeng-SH (approved):
Rechecked the latest head after the requested fixes.

The previous semantic blockers are addressed:

- Numeric containment no longer reuses the general ordering comparator for equality. The new path has exact checks across int/uint/float/decimal cases, and the regressions for `1` vs `1.000000009`, large integers above 2^53, uint64 boundaries, and decimal/float cases are covered by UT/BVT.
- The third-argument path flow now uses a JSON_CONTAINS-specific simple-path lookup with scalar `[0]` autowrap, without changing the existing QuerySimple/JSON modify behavior. Missing paths still return NULL, and wildcard/non-simple paths error as required.

I also checked the function registration/ID changes, strict/null handling, invalid JSON/path unhappy paths, typed JSON inputs, arrays/objects/scalars, and select-list skip behavior. CI is green, and I locally ran the targeted bytejson and function tests for the new code paths.

Approved.

ouyuanning (approved):

Issue #25175, opened 2026-06-26
[Bug] DECIMAL minus an INTEGER literal raises spurious 'Decimal128 Sub overflow' when the result is negative (4.0 regression)


dengn:
## Summary
Subtracting an **integer literal from a DECIMAL** raises a spurious `Decimal128 Sub overflow` error **whenever the result would be negative**. When the result is `>= 0`, or when the right operand is written as a decimal literal, it works correctly. This breaks any "balance / stock / delta" expression that can go negative.

## Environment
- **Reproduced on:** `8.0.30-OmniFabric-v4.0.0-rc2` (MatrixOne 4.0 release candidate; `@@version_comment = OmniFabric`), via mysql CLI.
- **Not present on:** MatrixOne Cloud `8.0.30-MatrixOne-v3.0.11` (there the same expressions evaluate correctly).

## Steps to reproduce
```sql
SELECT CAST(8000.00 AS DECIMAL(12,2)) - 1299;     -- = 6701.00     ✅ (result >= 0)
SELECT CAST(300.00  AS DECIMAL(12,2)) - 300;      -- = 0.00        ✅ (result = 0)
SELECT CAST(300.00  AS DECIMAL(12,2)) - 301;      -- ❌ ERROR (result should be -1.00)
SELECT CAST(100.00  AS DECIMAL(12,2)) - 200;      -- ❌ ERROR (result should be -100.00)
SELECT CAST(300.00  AS DECIMAL(12,2)) - 3999;     -- ❌ ERROR (result should be -3699.00)

-- Works if the right operand is a decimal literal:
SELECT CAST(300.00  AS DECIMAL(12,2)) - 3999.00;  -- = -3699.00    ✅
```
Also reproduces on a column:
```sql
CREATE TABLE t(bal DECIMAL(12,2)); INSERT INTO t VALUES (300.00);
SELECT bal - 3999 FROM t;   -- ❌ Decimal128 Sub overflow: 300.00-3999
```

## Actual result
```
ERROR 20301 (HY000): invalid input: Decimal128 Sub overflow: 300.00-3999
```

## Expected result
`-3699.00`. The value fits comfortably in `DECIMAL(12,2)` / Decimal128; there is no real overflow. The trigger is purely "DECIMAL − INTEGER-literal with a negative result", which points at a sign/scaling issue in the DECIMAL−INT subtraction path (the INTEGER operand is likely treated as unsigned, so a negative result is misdetected as overflow).

## Impact
[...]

PR #25479, opened 2026-07-06
fix decimal sub overflow

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25175

修改`decimal - int` overflow问题


heni02 (approved):

XuPeng-SH (approved):
The fix looks right.

The core bug here is aliasing `rs` with the pre-scaled operand in the diff-scale decimal fast paths: after writing `rs[i]`, the old overflow checks could reread the aliased input sign from the mutated result and misclassify valid negative results as overflow. Caching the operand signs before the write is the right fix, and this patch applies it consistently across the affected Decimal64 / Decimal128 add/sub same-scale kernels that the diff-scale paths delegate to.

The added unit tests exercise the alias-sensitive branches directly (same-length, scalar-vector, vector-scalar, with and without nulls), and the SQL regression cases cover the user-visible decimal-minus-int / int-minus-decimal shapes that were failing.

LGTM.

ouyuanning (approved):

PR #25484, opened 2026-07-06
fix: avoid FOR UPDATE CTE union limit panic

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23134

Fix planner panic when an outer `FOR UPDATE` over a CTE/UNION has `LIMIT` but no lock targets.

The CTE/UNION case intentionally produces no `LOCK_OP` for the CTE body. The LIMIT path still tried to rewrite `lockNode.Children` unconditionally, causing a nil pointer panic.

Verified with:

- `go test ./pkg/sql/plan -run TestUnionSqlBuilder`
- `bash /home/mo/test.sh --post-pr`

### DIFF
--- a/pkg/sql/plan/build_test.go
+++ b/pkg/sql/plan/build_test.go
@@ -914,6 +914,7 @@ func TestUnionSqlBuilder(t *testing.T) {
 		"(select n_name from nation for update) union all (select n_name from nation2 for update)",
 		"(select n_name from nation for update) union all (select n_name from nation2)",
 		"with qn as (select n_nationkey from nation union all select n_nationkey from nation2) select * from qn for update",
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
I did not find a blocker in this planner fix.

The new `lockNode != nil` guard is the right condition here: the outer CTE/UNION `FOR UPDATE` case intentionally has no lock targets, so the LIMIT path must not assume a `LOCK_OP` exists. I also rechecked the nearby builder flow and did not find another unconditional `lockNode.Children[0]` rewrite with the same panic pattern. The targeted `TestUnionSqlBuilder` regression and a clean `pkg/sql/plan` compile both pass on the PR head.

PR #25493, opened 2026-07-07
fix: harden ReadRows whole-result handling in system paths (#25456)

### LINKED ISSUE #25456: [Bug]: ReadRows whole-result assumptions can mis-handle batched results and crash ISCP HNSW jobs

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25456

This PR fixes a real bug class in internal SQL readers that use
`executor.Result.ReadRows`.

`ReadRows` is batch-oriented: the callback may be invoked multiple times, once
per batch. Several production paths were assuming callback-local `rows`
described the whole result set, which caused two different failure modes:

1. **False/fragile row-count checks**
   - treating per-batch row counts as whole-result counts
   - counting callbacks instead of total rows

2. **State corruption / crash on unhappy paths**
   - ISCP `GetJobSpecs()` treated a multi-batch result as a missing-row error
   - the permanent-error flush path then dereferenced nil `JobStatus` entries
   - in vector HNSW CI this crashed a CN and surfaced later as
     `backend connection closed`

#### 1. ISCP multi-job status loading
- `pkg/iscp/iteration.go`
- load `job_name` and `job_id` together with `job_spec` / `job_status`
- accumulate across all `ReadRows` batches
- map rows by `JobKey` instead of per-batch positional assumptions
- normalize nil `JobStatus` slices before permanent-error flushing
- stop the local execution path cleanly once a permanent error has already been flushed

#### 2. System-path whole-result row counting
- `pkg/iscp/util.go`
- `pkg/publication/util.go`
- `pkg/incrservice/store_sql.go`

These paths now validate total rows across all batches instead of using
callback-local `rows`.

#### 3. Additional single-row hardening
- `pkg/iscp/watermark_updater.go`
- `pkg/sql/plan/function/func_feature_upsert.go`
- `pkg/bootstrap/versions/version.go`
- `pkg/bootstrap/versions/upgrade_tenant_task.go`

These readers now use whole-result semantics for single-row contracts. Where a
duplicate row would represent a terminal invariant break (for example version
state lookup), the code remains fail-fast.

### Tests
- Added focused regression tests for:
  - multi-batch `ReadRows` handling in ISCP job loading
  - nil `JobStatus` permanent-error flushing
  - single-row task-runner readers across batches
  - incrservice offset/step readers across batches
  - ISCP `getTableID()` across batches
  - feature-upsert single-row readers across batches
  - bootstrap/version single-row helper behavior

### Why this matters
The visible CI failure happened in vector HNSW BVT, but the root cause was a
broader internal-reader bug class. Fixing only the panic site would still leave
other internal paths batch-sensitive and fragile. This PR closes the root cause
and hardens the nearby production readers that use the same pattern.


gouhongshen:
Blocker: the permanent-error path can report success without actually persisting the Error state.

`GetJobSpecs` now treats a nil return from `FlushPermanentErrorMessage` as `errPermanent`, and `ExecuteIteration` then returns nil for that error. That is only safe if the permanent-error update definitely landed. Today `FlushStatus` just closes the UPDATE result and ignores `AffectedRows`, while the SQL update is conditional on the previous LSN and also matches a concrete job row. For a missing job row (the new missing-row test case) or a stale LSN, the UPDATE can affect 0 rows, `FlushPermanentErrorMessage` still returns nil, and the worker treats the iteration as handled even though at least one target job was not marked Error.

Please make the flush path validate the affected row count for each job, or otherwise return a non-permanent error when the permanent-error update did not land. The regression test should exercise that real `AffectedRows == 0` case instead of stubbing `FlushJobStatusOnIterationState` to always return nil.

Relevant code: `pkg/iscp/iteration.go` (`GetJobSpecs` returning `errPermanent` after the flush, and `FlushStatus` ignoring the UPDATE result).

XuPeng-SH (changes requested):

The broad ReadRows hardening direction looks right, but the ISCP permanent-error path is not closed yet.

`GetJobSpecs` now converts a successful `FlushPermanentErrorMessage` call into `errPermanent`, and `ExecuteIteration` intentionally returns nil for `errPermanent`. That is only correct if the permanent-error status was definitely persisted. Today `FlushStatus` executes the conditional UPDATE and immediately closes the result without validating `AffectedRows`.

That UPDATE is conditional on the concrete job row and previous LSN. If the row is missing, the job id/name does not match, or the previous LSN is stale, the UPDATE can affect 0 rows. In that case `FlushPermanentErrorMessage` can still return nil, `GetJobSpecs` returns `errPermanent`, and `ExecuteIteration` reports success even though the target job was not marked Error. This loses the retry/error signal and can leave the job in an inconsistent state.

Please make the permanent-error flush path validate the UPDATE result for every job, for example requiring `AffectedRows == 1` in `FlushStatus` or in `FlushJobStatusOnIterationState`. If any target row is not updated, return a non-permanent error so the iteration is not swallowed as handled. The regression test should cover the real `AffectedRows == 0` case instead of stubbing `FlushJobStatusOnIterationState` to always return nil.

Relevant paths:
- `pkg/iscp/iteration.go`: `ExecuteIteration` swallows `errPermanent`.
- `pkg/iscp/iteration.go`: `GetJobSpecs` returns `errPermanent` after `FlushPermanentErrorMessage` returns nil.
- `pkg/iscp/iteration.go`: `FlushStatus` ignores `executor.Result.AffectedRows` from the conditional UPDATE.

fengttt (approved):

XuPeng-SH (approved):
Approved. The previously requested blocker is resolved: FlushStatus now requires AffectedRows == 1, so a missing/stale conditional update returns a non-permanent error and is not swallowed by ExecuteIteration. The added tests cover successful and zero-row updates plus propagation through GetJobSpecs. The whole-result ReadRows handling is consistent across batches, and the relevant CI checks pass on this head.

XuPeng-SH (changes requested):
Blocker found during the broader correctness/unhappy-path review:

In the per-job final-status flush at `pkg/iscp/iteration.go`, the call passes the full `iterCtx.lsn` slice as `prevLSN` while passing one-element slices for `jobNames`, `jobIDs`, `lsns`, and `jobStatuses`. `FlushJobStatusOnIterationState` indexes that argument with its local index, so it always uses `prevLSN[0]`. For a shared iteration, a non-first job (`i > 0`) therefore uses `iterCtx.lsn[0]` instead of `iterCtx.lsn[i]`. Shared iterations are grouped by watermark range, not by LSN, so distinct per-job LSNs are valid and reachable.

If that non-first consumer hits an error, the conditional UPDATE compares its persisted LSN against the wrong value. With this PR's new `AffectedRows == 1` validation it returns 0 deterministically, and the exponential retry repeats the same impossible predicate for up to roughly an hour. The failure is then only logged and `ExecuteIteration` can return nil, leaving the job status stale/Running.

Please pass `[]uint64{iterCtx.lsn[i]}` for `prevLSN` in this single-job flush and add a regression test covering a shared iteration with distinct LSNs where a non-first job enters the error flush path.

XuPeng-SH (approved):
Approved on the latest head.

The previously requested final-status blocker is resolved: the per-job helper now passes `iterCtx.lsn[jobIdx]` as both the current and conditional previous LSN for that single-job final flush, and the regression test explicitly covers the second job with distinct LSNs. The earlier permanent-error issue also remains correctly fixed by requiring `AffectedRows == 1`, so a missing or stale update is propagated rather than swallowed.

I rechecked the production ReadRows call sites for whole-result semantics, including same-batch and cross-batch duplicate rows. Resource results are closed, retries and waits remain bounded, and the added collections are bounded by the current result/job set. I found no additional correctness, performance, leak, hang, or OOM blocker. Focused tests pass for `pkg/iscp`, `pkg/publication`, `pkg/incrservice`, `pkg/bootstrap/versions`, and `pkg/sql/plan/function`; the benchmark CI also passes.

PR #25494, opened 2026-07-07
fix: validate empty parquet before row-group fanout

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25357 

fix: validate empty parquet before row-group fanout

### DIFF
--- a/pkg/fileservice/local_etl_fs.go
+++ b/pkg/fileservice/local_etl_fs.go
@@ -446,10 +446,13 @@ func (l *LocalETLFS) List(ctx context.Context, dirPath string) iter.Seq2[*DirEnt
 			if strings.HasPrefix(name, ".") {
 				continue
 			}
-			info, err := entry.Info()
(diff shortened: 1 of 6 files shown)


XuPeng-SH (approved):

XuPeng-SH (approved):
LGTM.

This backport matches the mainline fix where it matters: empty parquet files are validated before row-group fanout, fanout decisions count only files that actually contribute row groups, and the local fileservice listing path now tolerates transient ENOENT between readdir and stat/lstat. The targeted tests I checked line up with those behavior changes.

I only saw one non-blocking follow-up gap: adding a focused empty-parquet + `@variables` test would tighten coverage for the new guard path. I don't see a substantive correctness issue that should block this backport.

PR #25495, opened 2026-07-07
Filter GetShardInfo replicas by live Dragonboat node state

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25280 

  - update the Dragonboat dependency to include node host state reporting
  - add an OnlyLiveReplicaAddresses flag to GetShardInfo requests
  - filter GetShardInfo results to return only replicas whose Dragonboat gossip state is alive by default
  - keep an explicit opt-out path for callers that need the previous unfiltered behavior


XuPeng-SH (changes requested):
The new default filtering is too aggressive for discovery.

`isLiveReplica()` now keeps only `dragonboat.NodeHostStateAlive`, but Dragonboat's own API defines `suspect` as "currently suspects the nodehost is unreachable but has not marked it dead yet", and `unknown` as state unavailable. In other words, neither state is a hard-down signal.

That matters here because `GetShardInfo()` drops any replica that is not strictly `alive`, and then treats a missing leader entry as `ok=false`. Both reverse-proxy discovery paths (`connectToLogServiceByReverseProxy` and `connectByReverseProxy`) fail immediately on that `!ok` result. So a transient/memberlist-local suspicion, or simply unavailable node-state information, can now make an otherwise healthy shard undiscoverable and bubble up as "not ready" / "no hakeeper".

This is a real availability regression in the default path. At minimum, the default filter should only exclude replicas in explicit hard-down states like `dead` / `left`, or keep `suspect` / `unknown` as fallback candidates for discovery. The tests also currently miss the critical unhappy path where filtering removes the leader and callers start returning `!ok`.

LeftHandCold (approved):
当前版本没有发现需要 block 的问题。之前“只保留 alive 导致 discovery 过度激进”的问题，已经改成只过滤 dead/left，suspect/unknown/missing state 都会保留，这个方向是对的。

  我重点确认了几件事：

  - GetShardInfo() 对外默认会请求过滤 hard-down replica，但 getShardMembership() 显式传 false，不会破坏 zombie self-check 需要的完整 membership 语义。
  - protobuf 是追加 field 9，老客户端/老服务端滚动升级是兼容的：老端忽略未知字段，新服务端收到老请求时默认不启用过滤。
  - Dragonboat 依赖实际只改了 registry/gossip 相关 4 个文件，引入 GetNodeHostState；当前 memberlist fork 的字段布局匹配。
  - 默认过滤后，如果 leader 被本地 memberlist 标为 dead/left，GetShardInfo 会返回 ok=false 而不是 fallback 到 follower。这个继承了现有“leader 地址必须存在”的 discovery 语义，不是额外新问题，但属于可用性取舍。

XuPeng-SH (changes requested):
The latest head fixes the earlier `suspect`/`unknown` regression, but there is still one real availability blocker in the public discovery path.

`GetShardInfo()` now defaults `ExcludeHardDownReplicaAddresses=true`, and `Service.getShardInfo()` drops `dead`/`left` replicas before populating `Replicas`. But the public `GetShardInfo()` wrapper still immediately returns `ok=false` when `si.LeaderID` is not present in that filtered map. Both reverse-proxy discovery callers (`connectToLogServiceByReverseProxy` and `connectByReverseProxy`) still use the default 3-arg `GetShardInfo(...)`, so in the stale-leader window they fail discovery instead of trying the surviving follower addresses.

That stale-leader window is exactly the unhappy path reverse-proxy discovery needs to survive: if gossip still reports the old leader id while memberlist has already marked that node hard-down, the response can still contain healthy follower addresses, but the wrapper drops them on the floor and turns the shard back into `not ready`. I reproduced this locally with a tiny unit repro that feeds `GetShardInfo()` a result where `LeaderID=1` but only follower `2` still has an address; it still returns `ok=false`.

I think this needs either (1) the discovery callers opting out (`GetShardInfo(..., false)`) or (2) `GetShardInfo()` falling back to the surviving addresses when the filtered-out leader is the only missing entry. A regression test for “stale dead/left leader, live followers still connectable” would lock this down.

XuPeng-SH (changes requested):
The latest commit fixes the previously reported hard-down-leader case, but the fallback still conflates distinct discovery states. I found two blocking edge cases:

1. **Medium: `LeaderID == 0` is now incorrectly reported as usable.**

   The protobuf contract explicitly says `LeaderID == 0` means there is no leader or the leader is unknown. Since replica IDs are non-zero, `si.Replicas[si.LeaderID]` is absent and the new `!present` fallback chooses an arbitrary replica and returns `ok=true`. This regresses the previous fail-fast behavior. Both reverse-proxy paths then try to establish a client; the read-write path issues a Raft proposal and can consume the per-attempt timeout while the shard has no leader.

   I reproduced this with `LeaderID=0` and two replicas with valid addresses: current `GetShardInfo` returns `ok=true`. Please guard `si.LeaderID == 0` before applying the filtered-leader fallback, and add a regression test requiring `ok=false`.

2. **Medium: the fallback is inconsistent across the two server-side stale-replica filters.**

   A memberlist `dead/left` replica is omitted from `Replicas`, so a stale leader now falls back correctly. A HAKeeper-expired replica, however, is retained with `ServiceAddress=""` by `filterExpiredReplicaAddress`. Therefore an expired leader plus a live follower still returns `ok=false`; the availability fix only works for one of the two mechanisms used to address #25280.

   Simply falling back for every empty address would be unsafe because empty also means gossip metadata has not propagated. The first-principles fix is to preserve these reasons explicitly: for example, omit known-expired replicas from the default discovery response (while `getShardMembership(... includeExpired=true ...)` keeps the authoritative full membership), or add a typed filtered/availability state to the response. Then keep the three states separate:
   - no/unknown leader (`LeaderID == 0`) -> not ready;
   - known non-zero leader explicitly filtered as dead/left/expired -> return surviving addresses;
   - leader entry present but metadata/address unresolved -> not ready.

   Please add coverage for all three states. The existing unresolved-metadata and dead-leader tests cover only the latter two partially.

XuPeng-SH (approved):
当前版本我没有再看到需要 block 的问题，先 approve。

这次我重点又过了一遍几个 unhappy path：
- 默认过滤现在只排除 `dead/left`，`suspect/unknown/missing state` 都保留，没有再走回之前那个过度激进的 discovery 回归。
- `LeaderID == 0` 已经重新收紧成 not ready，不会再把“无 leader / leader 未知”伪装成可用地址集。
- stale leader 被 hard-down / expired 过滤掉时，`GetShardInfo()` 现在会回退到 surviving followers，reverse-proxy discovery 这条关键可用性路径补上了。
- `getShardMembership()` 仍然显式走 unfiltered 路径，没有破坏 zombie self-check 需要的完整 membership 语义。
- protobuf 还是追加 field，wire 兼容性方向也对。

非阻塞建议：后面可以再补两类 coverage——一类是真实 service/RPC 路径下的默认 hard-down 过滤集成测试，另一类是“leader 被过滤且没有任何可用 follower 地址”时返回 not ready 的失败分支测试。但这些我这次不作为 blocker。

Issue #25400, opened 2026-07-02
mo_table_stats betaTask goroutine stuck in select for 46.5 hours, causing transaction leak


aptend:
## Summary

In the freetier-01 production environment, a `betaTask` goroutine in `mo_table_stats.go` was stuck in a `select` statement for **46.5 hours**, holding an open transaction that could never be cleaned up.

## Environment
- Cluster: freetier-01 (prod)
- Pod: freetier-01-s2-56c954c888-cn-qxjh7
- MO version: production build deployed ~3 days ago

## Symptoms

1. **Leak checker** reported every 20 minutes:
   ```
   ERROR cn-service found leak txn
     txn-id: 0acdcbd9d5f822ce18bd902a36285465
     create-at: 2026/06/30 15:03:02  (46+ hours ago)
     session: migrate-goRoutineId 0
     lockWaitTimeout: 31536000000000000 (~1 year)
   ```

### Stuck goroutine (from ETL leakcheck profile)

```
goroutine 997023 [select, 2793 minutes]:   <-- 46.5 hours!
github.com/matrixorigin/matrixone/pkg/vm/engine/disttae.(*dynamicCtx).betaTask(...)
    mo_table_stats.go:2029                  <-- stuck in select
github.com/matrixorigin/matrixone/pkg/vm/engine/disttae.initMoTableStatsConfig.func1.8.1()
    mo_table_stats.go:464
```

### Code path (`mo_table_stats.go`)

`betaTask` is a long-lived goroutine that processes table statistics across multiple `alphaTask` rounds:

```go
func (d *dynamicCtx) betaTask(ctx context.Context, service string, eng engine.Engine) {
    for {
        select {
        case <-ctx.Done():        // exit on context cancellation
            return
        case tbl := <-d.tblQueue: // receive next table to process
            // process stats...
        }
        // *** NO idle timeout, NO heartbeat ***
    }
}
```

The `select` only has two cases:
1. `ctx.Done()` — but `turn2SysCtx()` only wraps with a tenant value, does NOT create a cancellable context
2. `d.tblQueue` — fed by `alphaTask`, which is called by `tableStatsExecutor` periodically

### Why it gets stuck permanently

[...]

PR #25497, opened 2026-07-07
fix: align frontend goroutine id with runtime dump

aptend:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25400

This PR makes frontend `goRoutineId` use the existing `runtime.Stack`-based goroutine id parser, so ids printed in `session_info` and leak txn logs can be matched directly with `goroutine <id>` in goroutine dumps.

It also removes the extra cleanup-time goroutine id probe and keeps the cleanup log focused on the routine goroutine id. That avoids another `runtime.Stack` call during cleanup and keeps the field semantics aligned with the session/routine being cleaned.

Tested with:

```sh
go test ./pkg/util/debug/goroutine ./pkg/frontend -run 'TestGetRoutineId|Test_routineid'
```


XuPeng-SH (approved):
LGTM.

The implementation change is small and coherent: frontend goroutine ids now reuse the same `runtime.Stack`-based parser as `pkg/util/debug/goroutine`, so the ids printed in `session_info` / leak-transaction logs line up with goroutine dump headers, and cleanup no longer does an extra probe from whatever goroutine happens to perform teardown. The targeted tests I checked passed, and I didn't find a substantive correctness regression in the nearby lifecycle paths.

One non-blocking follow-up: the new test mainly covers the nil-session cleanup path, but the PR's core claim (id matches runtime dump format) is still only weakly asserted. A focused test comparing `GetRoutineId()` with a `runtime.Stack` header would tighten that up nicely.

Issue #25481, opened 2026-07-06
[Bug]: nested window functions are accepted but MySQL rejects them


Ariznawlll:
### Branch Name

main

### Commit ID

d2617529a

### Other Environment Information

- MySQL baseline: 8.0.45
- MatrixOne SQL endpoint used for validation: local latest-main build on `127.0.0.1:26001`

### Actual Behavior

MatrixOne accepts nested window functions and returns a result.

For example, this query returns `28` for every input row:

```sql
SELECT 'nested_window' q,
       sum(row_number() over (order by id)) over ()
FROM wf_base;
```

### Expected Behavior

MySQL rejects the same SQL because window functions are not allowed in this context.

MySQL 8.0.45 result:

```text
ERROR 3593 (HY000): You cannot use the window function 'row_number' in this context.
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS mysql_compat_window_probe;
CREATE DATABASE mysql_compat_window_probe;
USE mysql_compat_window_probe;

CREATE TABLE wf_base (
  id int primary key,
  grp varchar(8),
  amt decimal(10,2)
);
INSERT INTO wf_base VALUES
(1,'A',10.00),(2,'A',10.00),(3,'A',20.50),
(4,'B',NULL),(5,'B',30.25),(6,'B',30.25),(7,'C',40.00);

SELECT 'nested_window' q,
       sum(row_number() over (order by id)) over ()
FROM wf_base;
```

### Additional information

This was found while exploring the window function compatibility matrix on latest `main`. Basic window happy paths matched MySQL in the same batch, but this unhappy path differs from MySQL semantics.

VioletQwQ-0:
Fixed by merged PR #25506: https://github.com/matrixorigin/matrixone/pull/25506

Validation included planner unit tests plus passing BVT, UT, and multi-CN regression CI.

PR #25506, opened 2026-07-07
fix(plan): reject nested window functions

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25481

MatrixOne accepted nested window functions such as `sum(row_number() over (...)) over ()`, while MySQL rejects this usage.

Changes:
- Add planner-side validation that rejects window functions nested inside another window function's arguments, function-local order by expressions, window partition/order expressions, and frame bounds.
- Return a syntax error with the nested window function name.
- Add unit coverage for the issue SQL and adjacent nested-window contexts.
- Add a focused BVT case for valid window usage and rejected nested window usage.

Validation:
- `go test ./pkg/sql/plan -run 'TestProjectionBinderRejectsNestedWindowFuncFromSQL|TestProjectionBinderBindWinFunc|TestHavingBinderBindWinFuncCoversFrameAndGuard|TestBindWindowFuncExprValidationAndHelpers' -count=1`
- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `git diff --check`

Note: the new BVT case was added but not run locally because no MatrixOne SQL service was listening on `127.0.0.1:26001`.


heni02 (approved):

XuPeng-SH (changes requested):
I found two correctness gaps that leave the nested-window invariant incomplete.

1. **High: a select-list alias bypasses the AST-only validation.**

   At `f4f153d6ed6b5850b1f0c5b98c5a20c2b635740a`, this query still builds successfully:

   ```sql
   SELECT ROW_NUMBER() OVER () AS rn,
          SUM(rn) OVER ()
   FROM select_test.bind_select;
   ```

   I reproduced this through the full `BuildPlan` path. The resulting plan contains two serial `WINDOW` nodes, and the `sum` window consumes the prior `row_number` window output. MySQL 8.0.44 rejects the same query with `ERROR 1054 (42S22): Unknown column 'rn' in 'field list'` because a sibling select alias is not visible there.

   `validateWindowFuncNoNested` only inspects the literal AST under the second function, where `rn` is just an `UnresolvedName`. Later, `baseBindColRef` resolves it from `ctx.aliasMap`/the already-built project, so the same-query-block window-on-window dependency survives the new guard.

   The first-principles invariant should be enforced at query-block/dependency level: a window expression may consume base expressions and aggregates, but must not consume another window result from the same query block. Please reject window aliases when resolving arguments, partition/order expressions, and frame expressions, or validate the bound dependency before appending the window node. Keep scalar subqueries as an explicit scope boundary; MySQL permits `SELECT SUM((SELECT ROW_NUMBER() OVER ())) OVER ()`.

2. **Medium: the outer window function's own function-local `ORDER BY` is not validated.**

   `validateWindowFuncNoNested` checks `astExpr.Exprs` and `astExpr.WindowSpec.OrderBy`, but never checks `astExpr.OrderBy`. This query is therefore also accepted by `BuildPlan`:

   ```sql
   SELECT GROUP_CONCAT(a ORDER BY ROW_NUMBER() OVER ()) OVER ()
   FROM select_test.bind_select;
   ```

   The recursive helper covers an `OrderBy` attached to a non-window function *inside an argument*, which is why the synthetic helper test passes, but it does not cover the root `FuncExpr`. Add root function-local `ORDER BY` validation. Since `bindWindowFuncExpr` currently does not bind that root `OrderBy` either, the broader behavior should be explicit: either support it and preserve its semantics, or reject it rather than silently ignore it.

Please add end-to-end planner/BVT regressions for both reproductions, plus alias references from window arguments and window `PARTITION BY`/`ORDER BY`. Preserve positive coverage for independent sibling windows, aggregate-inside-window expressions, and windows inside a scalar subquery (a separate query block).

VioletQwQ-0:
Addressed both correctness gaps from the review.

Changes:
- Added bound-expression dependency validation so a window function cannot consume another window result from the same query block through a select-list alias. This covers arguments, `PARTITION BY`, window `ORDER BY`, and frame bounds.
- Added validation for the root window function's function-local `ORDER BY`. Nested windows now produce the existing nested-window syntax error; non-nested function-local ordering is rejected explicitly as NYI instead of being silently ignored.
- Kept scalar subqueries as a query-block boundary, so a window inside a scalar subquery remains valid.

Coverage:
- Added full `BuildPlan` regressions for alias references from window arguments, `PARTITION BY`, and window `ORDER BY`.
- Added planner coverage for root function-local `ORDER BY` and bound frame dependencies.
- Added positive coverage for independent sibling windows, aggregate-inside-window, and a window inside a scalar subquery.
- Extended the BVT SQL/result cases for the executable reproductions and positive boundaries.

Validation after syncing with current `upstream/main`:
- `go test ./pkg/sql/plan -run 'TestBuildPlanWindowDependencyValidation|TestProjectionBinderRejectsNestedWindowFuncFromSQL|TestBindWindowFuncExprValidationAndHelpers' -count=1`
- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `git diff --check`

All passed locally. The BVT case was added but not executed locally because no MatrixOne test service is running on this machine.

aunjgr (approved):
Reviewed current head `04e9bb94`. The follow-up commits close both prior correctness gaps: bound-expression tag validation rejects same-query-block window aliases in arguments/partition/order/frame contexts, root function-local ORDER BY is no longer silently ignored, and scalar subqueries remain an explicit scope boundary. The focused planner/BVT coverage is appropriate, and current CI is green. I found no remaining merge blocker.

XuPeng-SH (approved):
Re-reviewed latest head `04e9bb94` against the two earlier blockers.

Both are now closed. The binder no longer relies only on AST-shape validation: after binding it rejects any same-query-block dependency on an earlier window result via `windowTag`, so select-list aliases like `rn` are now blocked in arguments, `PARTITION BY`, window `ORDER BY`, and frame expressions. Scalar subqueries still act as the intended scope boundary.

The outer function-local `ORDER BY` gap is also fixed: nested windows there are rejected during validation, and non-nested function-local `ORDER BY` is now rejected explicitly instead of being silently ignored.

The new planner/BVT coverage matches those contracts well. LGTM.

XuPeng-SH (approved):

PR #25508, opened 2026-07-07
fix: support parquet cross-type conversions

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24914 

fix: support parquet cross-type conversions


heni02 (approved):

ouyuanning (approved):

XuPeng-SH (changes requested):
I found a correctness blocker in the new parquet DECIMAL logical -> integer path.\n\nHigh: decimal logical values are converted to integers through float64, which can silently corrupt exact DECIMAL values and also reject valid boundary values. The new rounded-integer path treats any DECIMAL logical parquet source as `isParquetRoundedIntegerSource`, then `parquetValueToRoundedInt64` / `parquetValueToRoundedUint64` call `parquetValueToFloat64` and round the float. That loses precision for exact DECIMAL values above 2^53.\n\nI verified this on the PR head with a temporary mapper-level test using `parquet.Decimal(0, 19, parquet.FixedLenByteArrayType(8))`:\n\n- DECIMAL value `9007199254740993` maps to `9007199254740992` when loaded into BIGINT.\n- DECIMAL value `9223372036854775807` is rejected with `parquet value 9.223372036854776e+18 overflows BIGINT`, although it is a valid BIGINT.\n\nThis should not use float64 for DECIMAL -> integer. Please keep DECIMAL conversions exact: decode the DECIMAL logical value into Decimal64/128/256 (or its scaled integer form), apply the same scale/rounding/truncation semantics as MO SQL cast, then perform exact signed/unsigned range checks before appending. Please also add regression coverage for values above 2^53 and int64/uint64 boundaries so this cannot silently regress again.\n\nThe other added conversions and BVT coverage look generally reasonable, but this data-corruption path needs to be fixed before merge.

iamlinjunhong:
> I found a correctness blocker in the new parquet DECIMAL logical -> integer path.\n\nHigh: decimal logical values are converted to integers through float64, which can silently corrupt exact DECIMAL values and also reject valid boundary values. The new rounded-integer path treats any DECIMAL logical parquet source as `isParquetRoundedIntegerSource`, then `parquetValueToRoundedInt64` / `parquetValueToRoundedUint64` call `parquetValueToFloat64` and round the float. That loses precision for exact DECIMAL values above 2^53.\n\nI verified this on the PR head with a temporary mapper-level test using `parquet.Decimal(0, 19, parquet.FixedLenByteArrayType(8))`:\n\n- DECIMAL value `9007199254740993` maps to `9007199254740992` when loaded into BIGINT.\n- DECIMAL value `9223372036854775807` is rejected with `parquet value 9.223372036854776e+18 overflows BIGINT`, although it is a valid BIGINT.\n\nThis should not use float64 for DECIMAL -> integer. Please keep DECIMAL conversions exact: decode the DECIMAL logical value into Decimal64/128/256 (or its scaled integer form), apply the same scale/rounding/truncation semantics as MO SQL cast, then perform exact signed/unsigned range checks before appending. Please also add regression coverage for values above 2^53 and int64/uint64 boundaries so this cannot silently regress again.\n\nThe other added conversions and BVT coverage look generally reasonable, but this data-corruption path needs to be fixed before merge.

fixed

aunjgr (approved):

XuPeng-SH (approved):
最新 head 看下来，我上次卡的 DECIMAL logical -> integer 精度问题已经补齐了，可以 approve。

这版关键点已经到位：
- `DECIMAL -> integer` 不再走 `float64`，而是改成了 `big.Int` 的 exact decode + rounding + range check；
- 回归测试也把我当时提的核心边界补进来了：`2^53+1`、`max int64`、`max uint64`，以及对应的 overflow / negative-to-unsigned 路径；
- BVT 这边的 cross-type conversion 覆盖也比之前完整不少。

按当前 scope，我这里没有剩余 blocker 了。

PR #25509, opened 2026-07-07
perf(lockservice): avoid txn overview builds for active txn checks

aptend:
## What type of PR is this?

- [x] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #25279

This PR reduces the cost of lockservice active transaction validation paths involved in remote `validTxn` checks.

- Add a `CheckActiveTxn` lockservice RPC for point lookup by `ServiceID` and `TxnID`.
- Keep `GetActiveTxn` as the full active txn ID listing API for existing list consumers.
- Use `CheckActiveTxn` in remote `validTxn`, with fallback to `GetActiveTxn` when the peer does not support the new method.
- Add `TxnClient.IterTxnIDs` so lockservice `TxnIterFunc` can enumerate txn IDs without building `TxnOverview` or copying wait-lock state.
- Keep callback execution outside internal txn-client locks.
- Add focused coverage for `IterTxnIDs` and `CheckActiveTxn` handler behavior.

Validation:

```sh
go test -race -count=1 -timeout 180s ./pkg/txn/client
go test -count=1 -timeout 120s ./pkg/cnservice
go test -count=1 -timeout 120s -run 'TestHandleCheckActiveTxn|TestGetActiveTxnWithRemote|TestCannotCommit|TestValidTxn|TestCheckTxnTimeout' ./pkg/lockservice
```


XuPeng-SH (approved):
LGTM. The new CheckActiveTxn point-check path, old-peer fallback to GetActiveTxn, and the IterTxnIDs lock-avoidance path all look sound to me, and the targeted tests cover the important compatibility path.

PR #25510, opened 2026-07-07
fix: support parquet cross-type conversions

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24914 

fix: support parquet cross-type conversions


XuPeng-SH (changes requested):
Thanks for extending the parquet conversion matrix. I ran:

- `git diff --check origin/4.1-dev...HEAD`
- `go test ./pkg/sql/colexec/external -run 'TestParquetListToVectorMapping|TestParquetCrossTypeMappings|TestParquetCrossTypeHelperCoverage|TestParquetTimestampLogicalMissingUnit' -count=1`
- `go test ./pkg/sql/colexec/external -count=1`

They pass locally, but I think this still needs changes before it should close #24914 / the broader conversion-matrix comments.

1. The numeric conversion closure is incomplete. The PR adds rounded float/double/decimal handling only for `T_int64` (`parquetValueToRoundedInt64`), while the other integer target branches still gate on `isParquetIntegerSource`. That means the same parquet `FLOAT`, `DOUBLE`, or decimal logical source can load into `BIGINT`, but still returns no mapper for `TINYINT`, `SMALLINT`, `INT`, and unsigned integer targets. SQL cast already supports float -> integer widths via rounded conversion with overflow checks, and the issue comments explicitly list `FLOAT/DOUBLE -> INT8/INT16/INT32/INT64` and unsigned integer targets as part of the unsupported matrix. Please make this systematic: add signed/unsigned rounded numeric helpers and wire them through every integer-width branch with per-width overflow checks, then cover negative-to-unsigned and overflow cases.

2. `TIME` logical parquet values still do not map to `DATETIME`. In the `T_datetime` branch the only non-string logical inputs are `DATE` and `TIMESTAMP`; if `lt.Time != nil`, the branch falls through to nil mapper. MatrixOne's SQL cast layer supports `TIME -> DATETIME`, and the issue matrix called out `TIME logical -> DATETIME` as unsupported. Please either implement the parity conversion and tests, or explicitly narrow the PR/issue scope so this PR is not treated as closing the conversion-matrix gap.

The current tests are good representative coverage, but because this code is the central `LOAD DATA parquet` mapper, the regression suite should be table-driven over the actual matrix groups added here: bool/int/uint/float/double/decimal/date/time/timestamp/string/list sources against the supported target families, including NULL, invalid value, overflow, and dictionary/non-dictionary pages.

iamlinjunhong:
> Thanks for extending the parquet conversion matrix. I ran:
> 
> * `git diff --check origin/4.1-dev...HEAD`
> * `go test ./pkg/sql/colexec/external -run 'TestParquetListToVectorMapping|TestParquetCrossTypeMappings|TestParquetCrossTypeHelperCoverage|TestParquetTimestampLogicalMissingUnit' -count=1`
> * `go test ./pkg/sql/colexec/external -count=1`
> 
> They pass locally, but I think this still needs changes before it should close #24914 / the broader conversion-matrix comments.
> 
> 1. The numeric conversion closure is incomplete. The PR adds rounded float/double/decimal handling only for `T_int64` (`parquetValueToRoundedInt64`), while the other integer target branches still gate on `isParquetIntegerSource`. That means the same parquet `FLOAT`, `DOUBLE`, or decimal logical source can load into `BIGINT`, but still returns no mapper for `TINYINT`, `SMALLINT`, `INT`, and unsigned integer targets. SQL cast already supports float -> integer widths via rounded conversion with overflow checks, and the issue comments explicitly list `FLOAT/DOUBLE -> INT8/INT16/INT32/INT64` and unsigned integer targets as part of the unsupported matrix. Please make this systematic: add signed/unsigned rounded numeric helpers and wire them through every integer-width branch with per-width overflow checks, then cover negative-to-unsigned and overflow cases.
> 2. `TIME` logical parquet values still do not map to `DATETIME`. In the `T_datetime` branch the only non-string logical inputs are `DATE` and `TIMESTAMP`; if `lt.Time != nil`, the branch falls through to nil mapper. MatrixOne's SQL cast layer supports `TIME -> DATETIME`, and the issue matrix called out `TIME logical -> DATETIME` as unsupported. Please either implement the parity conversion and tests, or explicitly narrow the PR/issue scope so this PR is not treated as closing the conversion-matrix gap.
> 
> The current tests are good representative coverage, but because this code is the central `LOAD DATA parquet` mapper, the regression suite should be table-driven over the actual matrix groups added here: bool/int/uint/float/double/decimal/date/time/timestamp/string/list sources against the supported target families, including NULL, invalid value, overflow, and dictionary/non-dictionary pages.

fixed

heni02 (approved):

ouyuanning (approved):

pkg/sql/colexec/external/parquet.go
```diff
@@ -953,16 +1144,16 @@ func (*ParquetHandler) getMapper(sc *parquet.Column, dt plan.Type) *columnMapper
...
 				var tzOffsetMicros int64 = 0
 				if !isAdjustedToUTC {
-					loc := proc.Base.SessionInfo.TimeZone
-					if loc == nil {
-						loc = time.Local
-					}
+					loc := parquetSessionLocation(proc)
 					// Get the timezone offset in seconds, then convert to microseconds
 					// We use current time to get the offset, which handles DST correctly for current data
 					_, offset := time.Now().In(loc).Zone()
```
XuPeng-SH: This offset calculation is tied to `time.Now()`, not to the timestamp value being converted. For `isAdjustedToUTC=false` Parquet timestamps that means DST-sensitive zones are wrong for historical/future data: a value from January in `America/New_York` needs -05:00, but loading it in July will subtract the current -04:00 offset and shift the result by one hour.

The offset has to be derived from the represented local wall-clock instant, not from the machine's current date. All current tests/BVTs run in UTC, so this regression path is still uncovered.

XuPeng-SH (changes requested):
I found one blocking correctness issue in the new Parquet cross-type conversion path.

For Parquet timestamps with `isAdjustedToUTC=false`, the new TIMESTAMP mapping computes the timezone offset with `time.Now().In(loc).Zone()`. That makes the result depend on the wall clock of the load job instead of the wall clock of the value being decoded. In DST-sensitive zones this is wrong for historical/future data: e.g. a January timestamp loaded in July in `America/New_York` will use -04:00 instead of the value's real -05:00 and land one hour off.

This path is newly introduced here and the current coverage doesn't catch it because the tests/BVT run in UTC. I'd fix it before merge by deriving the offset from the represented local instant itself, then add a regression test with a non-UTC session timezone and a value that crosses a DST boundary.

aunjgr (approved):

iamlinjunhong:
> I found one blocking correctness issue in the new Parquet cross-type conversion path.
> 
> For Parquet timestamps with `isAdjustedToUTC=false`, the new TIMESTAMP mapping computes the timezone offset with `time.Now().In(loc).Zone()`. That makes the result depend on the wall clock of the load job instead of the wall clock of the value being decoded. In DST-sensitive zones this is wrong for historical/future data: e.g. a January timestamp loaded in July in `America/New_York` will use -04:00 instead of the value's real -05:00 and land one hour off.
> 
> This path is newly introduced here and the current coverage doesn't catch it because the tests/BVT run in UTC. I'd fix it before merge by deriving the offset from the represented local instant itself, then add a regression test with a non-UTC session timezone and a value that crosses a DST boundary.

fixed

XuPeng-SH (changes requested):
Request changes on the latest head. Two newly supported cross-type paths still have reproducible correctness failures:

1. **[High] TIMESTAMP(isAdjustedToUTC=true) -> DATETIME ignores the session timezone.** The DATETIME mapper at `parquet.go:1268-1323` directly converts the physical Unix value with `types.Datetime(types.Unix*ToTimestamp(...))` for both plain and dictionary pages. It never inspects `dtT.IsAdjustedToUTC` and never uses the session location. This is inconsistent with the new TIMESTAMP -> DATE/TIME paths, which correctly call `parquetTimestampValueToDatetime`; that helper converts an adjusted UTC instant through `ts.ToDatetime(loc)`. I reproduced this on the current head: loading `2024-01-01 00:00:00 UTC` into DATETIME with `Asia/Shanghai` produces `00:00:00`, while the correct session-local DATETIME is `08:00:00`. Please route all timestamp units and both encodings through the common adjusted/unadjusted helper and add non-UTC tests for adjusted=true and adjusted=false.

2. **[High] DOUBLE/DECIMAL -> FLOAT32 silently writes infinity instead of rejecting overflow.** The scalar FLOAT32 mapper at `parquet.go:966-970` narrows with `float32(val)` without checking `abs(val) > math.MaxFloat32`; the new LIST<DOUBLE> -> VECF32 mapper at `parquet.go:428-432` has the same issue. A finite Parquet DOUBLE value `1e100` maps successfully with nil error and becomes `+Inf`. MatrixOne cast handling and the STRING -> FLOAT32 path reject out-of-range values, so LOAD DATA must not silently manufacture infinity. Please centralize a checked FLOAT32 conversion and use it for scalar decimal/double sources and list-double elements. Add positive/negative overflow tests, including dictionary/plain scalar pages and LIST<DOUBLE>.

The current CI and the complete `pkg/sql/colexec/external` test suite pass, but the BVT fixes `time_zone` to UTC and uses only small float values, so neither failure is covered. I also checked NULL/error page cleanup, dictionary index validation, page-bounded allocations, and synchronous termination paths; I found no additional lifecycle/hang/OOM blocker.

iamlinjunhong:
> Request changes on the latest head. Two newly supported cross-type paths still have reproducible correctness failures:
> 
> 1. **[High] TIMESTAMP(isAdjustedToUTC=true) -> DATETIME ignores the session timezone.** The DATETIME mapper at `parquet.go:1268-1323` directly converts the physical Unix value with `types.Datetime(types.Unix*ToTimestamp(...))` for both plain and dictionary pages. It never inspects `dtT.IsAdjustedToUTC` and never uses the session location. This is inconsistent with the new TIMESTAMP -> DATE/TIME paths, which correctly call `parquetTimestampValueToDatetime`; that helper converts an adjusted UTC instant through `ts.ToDatetime(loc)`. I reproduced this on the current head: loading `2024-01-01 00:00:00 UTC` into DATETIME with `Asia/Shanghai` produces `00:00:00`, while the correct session-local DATETIME is `08:00:00`. Please route all timestamp units and both encodings through the common adjusted/unadjusted helper and add non-UTC tests for adjusted=true and adjusted=false.
> 2. **[High] DOUBLE/DECIMAL -> FLOAT32 silently writes infinity instead of rejecting overflow.** The scalar FLOAT32 mapper at `parquet.go:966-970` narrows with `float32(val)` without checking `abs(val) > math.MaxFloat32`; the new LIST -> VECF32 mapper at `parquet.go:428-432` has the same issue. A finite Parquet DOUBLE value `1e100` maps successfully with nil error and becomes `+Inf`. MatrixOne cast handling and the STRING -> FLOAT32 path reject out-of-range values, so LOAD DATA must not silently manufacture infinity. Please centralize a checked FLOAT32 conversion and use it for scalar decimal/double sources and list-double elements. Add positive/negative overflow tests, including dictionary/plain scalar pages and LIST.
> 
> The current CI and the complete `pkg/sql/colexec/external` test suite pass, but the BVT fixes `time_zone` to UTC and uses only small float values, so neither failure is covered. I also checked NULL/error page cleanup, dictionary index validation, page-bounded allocations, and synchronous termination paths; I found no additional lifecycle/hang/OOM blocker.

fixed

XuPeng-SH (approved):
最新 head 看下来，我上次卡的两个 correctness blocker 都已经补齐了，可以 approve。

- `TIMESTAMP -> DATETIME/TIME/DATE` 现在统一走 `parquetTimestampValueToDatetime(...)` / 本地时区 helper，`isAdjustedToUTC=true/false` 的路径都收口了；`America/New_York` 的 DST 回归测试也补进来了。
- `DOUBLE/DECIMAL -> FLOAT32` 以及 `LIST<DOUBLE> -> VECF32` 现在都走带范围检查的 `parquetFloat64ToFloat32(...)`，不再静默写 `Inf`；正负溢出 UT 也补了。

按当前 scope，我这里没有剩余 blocker 了。

Issue #24738, opened 2026-05-31
[Bug]: Remaining JSON function gaps (JSON_VALID/KEYS/TYPE/REMOVE/MERGE_*/CONTAINS_PATH/OVERLAPS, CONVERT AS JSON) — v3.0.11 & 4.0-dev


dengn:
### Summary
After the recent JSON-function additions (#24509 / #24670 — thanks; `JSON_ARRAY` / `JSON_OBJECT` / `JSON_ARRAYAGG` / `JSON_LENGTH` now work on the 4.0-dev line), the following JSON functions are **still unsupported** as of `nightly-0bfda604` (and on prod v3.0.11). Filing as a consolidated remainder; related to batch tracker #24481.

| Function | 3.0.11 | 4.0-dev | Error |
|---|---|---|---|
| `JSON_VALID` | ❌ | ❌ | not supported: function or operator 'json_valid' |
| `JSON_KEYS` | ❌ | ❌ | ... 'json_keys' |
| `JSON_TYPE` | ❌ | ❌ | ... 'json_type' |
| `JSON_REMOVE` | ❌ | ❌ | ... 'json_remove' |
| `JSON_MERGE_PATCH` / `JSON_MERGE_PRESERVE` | ❌ | ❌ | ... 'json_merge_patch' |
| `JSON_CONTAINS_PATH` | ❌ | ❌ | ... 'json_contains_path' |
| `JSON_OVERLAPS` | ❌ | ❌ | ... 'json_overlaps' |
| `CONVERT(expr, JSON)` | ❌ | ❌ | parser error near " JSON)" (note: `CAST(expr AS JSON)` works) |
| `JSON_OBJECTAGG` | ❌ | ✅ | (3.0.11 only) |

`JSON_CONTAINS` is tracked separately in #24732. `JSON_SET` / `JSON_INSERT` / `JSON_REPLACE` / `JSON_EXTRACT` / `JSON_UNQUOTE` / `JSON_QUOTE` and the `->` / `->>` operators all work.

### Steps to Reproduce
```sql
SELECT JSON_VALID('{}');
SELECT JSON_KEYS('{"a":1}');
SELECT JSON_TYPE(CAST('[]' AS JSON));
SELECT JSON_REMOVE('{"a":1,"b":2}', '$.a');
SELECT JSON_MERGE_PATCH('{"a":1}', '{"b":2}');
SELECT JSON_CONTAINS_PATH('{"a":1}', 'one', '$.a');
[...]

PR #25514, opened 2026-07-07
add json_remove

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24738

增加 json_remove


XuPeng-SH (changes requested):
I found two real correctness issues in the current JSON_REMOVE implementation.

1. JSON literal null is collapsed to SQL NULL

JSON_REMOVE('null', path) should return the original JSON document null unchanged when the path matches nothing. But buildJsonRemove() currently treats any ByteJson.IsNull() result as SQL NULL and appends a SQL NULL result row. bytejson.Null is the JSON literal null, not SQL NULL. remove() already returns the original document on no-op paths, so this turns a valid JSON result into the wrong SQL result.

2. Quoted member name star is rejected as a wildcard

A valid MySQL path targeting a literal object member named star should not be treated as a wildcard. But Path.init() currently marks any subPathKey with key == "*" as pathFlagSingleStar regardless of whether it came from quoted syntax, so IsSimple() rejects it and JSON_REMOVE errors on what should be a valid simple path. That means a document like {"*":1} cannot be updated through a quoted-star member path.

These are both semantic mismatches in the core implementation, so I think this still needs request-changes rather than just extra tests.

XuPeng-SH (changes requested):
Requesting changes for one compatibility blocker.

`json_remove` currently turns a successful JSON literal `null` result into SQL `NULL`:

```go
if out.IsNull() {
    rs.AppendBytes(nil, true)
} else {
    rs.AppendByteJson(out, false)
}
```

This is wrong for cases where the input document itself is JSON `null` and the path is a no-op/autowrapped root path. MySQL keeps JSON literal `null` distinct from SQL NULL:

```sql
SELECT JSON_REMOVE('null', '$[0]') AS r,
       JSON_REMOVE('null', '$[0]') IS NULL AS is_sql_null,
       JSON_TYPE(JSON_REMOVE('null', '$[0]')) AS jt;
-- MySQL 8.0: r = null, is_sql_null = 0, jt = NULL
```

On this PR head, a temporary unit test for that SQL-equivalent path fails because `vec.IsNull(0)` is true, so the function returns SQL NULL instead of a JSON value.

Suggested fix: in `buildJsonRemove`, SQL NULL should be decided only from SQL NULL inputs / ignored rows. After a successful `computeJsonRemove` / `computeStringJsonRemove`, append the returned `ByteJson` as a non-null JSON value, even when it is the JSON literal null. Add regressions that distinguish the two, for example `JSON_REMOVE('null', '$[0]') IS NULL` should be false and/or `JSON_TYPE(JSON_REMOVE('null', '$[0]'))` should return `NULL` as a JSON type string, not SQL NULL.

Local checks run on a clean PR branch:
- `go test ./pkg/sql/plan/function -run 'TestJsonRemove|TestJsonRemoveNullAndInvalidPaths|TestJsonRemoveCheckFn|TestJsonRemoveIgnoreAllRows|Test_funids' -count=1`
- `go test ./pkg/container/bytejson -count=1` with local CGo/rpath flags
- MySQL 8.0 Docker compatibility probes for `$[0]`, `$[last]`, NULL path priority, and wildcard errors

daviszhen:
> I found two real correctness issues in the current JSON_REMOVE implementation.
> 
> 1. JSON literal null is collapsed to SQL NULL
> 
> JSON_REMOVE('null', path) should return the original JSON document null unchanged when the path matches nothing. But buildJsonRemove() currently treats any ByteJson.IsNull() result as SQL NULL and appends a SQL NULL result row. bytejson.Null is the JSON literal null, not SQL NULL. remove() already returns the original document on no-op paths, so this turns a valid JSON result into the wrong SQL result.
> 
> 2. Quoted member name star is rejected as a wildcard
> 
> A valid MySQL path targeting a literal object member named star should not be treated as a wildcard. But Path.init() currently marks any subPathKey with key == "_" as pathFlagSingleStar regardless of whether it came from quoted syntax, so IsSimple() rejects it and JSON_REMOVE errors on what should be a valid simple path. That means a document like {"_":1} cannot be updated through a quoted-star member path.
> 
> These are both semantic mismatches in the core implementation, so I think this still needs request-changes rather than just extra tests.

已经修改.

daviszhen:
> Requesting changes for one compatibility blocker.
> 
> `json_remove` currently turns a successful JSON literal `null` result into SQL `NULL`:
> 
> ```go
> if out.IsNull() {
>     rs.AppendBytes(nil, true)
> } else {
>     rs.AppendByteJson(out, false)
> }
> ```
> 
> This is wrong for cases where the input document itself is JSON `null` and the path is a no-op/autowrapped root path. MySQL keeps JSON literal `null` distinct from SQL NULL:
> 
> ```sql
> SELECT JSON_REMOVE('null', '$[0]') AS r,
>        JSON_REMOVE('null', '$[0]') IS NULL AS is_sql_null,
>        JSON_TYPE(JSON_REMOVE('null', '$[0]')) AS jt;
> -- MySQL 8.0: r = null, is_sql_null = 0, jt = NULL
> ```
> 
> On this PR head, a temporary unit test for that SQL-equivalent path fails because `vec.IsNull(0)` is true, so the function returns SQL NULL instead of a JSON value.
> 
> Suggested fix: in `buildJsonRemove`, SQL NULL should be decided only from SQL NULL inputs / ignored rows. After a successful `computeJsonRemove` / `computeStringJsonRemove`, append the returned `ByteJson` as a non-null JSON value, even when it is the JSON literal null. Add regressions that distinguish the two, for example `JSON_REMOVE('null', '$[0]') IS NULL` should be false and/or `JSON_TYPE(JSON_REMOVE('null', '$[0]'))` should return `NULL` as a JSON type string, not SQL NULL.
> 
> Local checks run on a clean PR branch:
> 
> * `git diff --check origin/main...HEAD`
> * `go test ./pkg/sql/plan/function -run 'TestJsonRemove|TestJsonRemoveNullAndInvalidPaths|TestJsonRemoveCheckFn|TestJsonRemoveIgnoreAllRows|Test_funids' -count=1`
> * `go test ./pkg/sql/plan/function -count=1`
> * `go test ./pkg/container/bytejson -count=1` with local CGo/rpath flags
> * MySQL 8.0 Docker compatibility probes for `$[0]`, `$[last]`, NULL path priority, and wildcard errors

已经修改.


  - 修复 JSON literal null 被误当成 SQL NULL 的问题。
    buildJsonRemove 现在只在输入 SQL NULL、path SQL NULL、selectList 忽略行时返回 SQL NULL；成功计算出的 bytejson.Null 会作为 JSON 值 append。

  - 修复 $."*" 被误判为 wildcard 的问题。
    bytejson.Path 新增 subPathKeyWildcard，区分未 quoted 的 .* 和 quoted 的普通 key "*"。因此 JSON_REMOVE('{"*":1,"a":2}', '$."*"') 可以正常删除 key *。

  - 收紧 JSON_REMOVE 参数类型检查。
    第一个参数只允许 JSON/string/NULL literal，path 参数只允许 string/NULL literal；不再把 numeric/bool 隐式 cast 成 varchar。

  - 补充 MySQL 对齐测试。
    UT 覆盖数组边界、多 path 左到右、missing/noop、深层 autowrap、JSON null vs SQL NULL、quoted star key、非法 wildcard/root path。
    BVT SQL 也补了 MySQL json_no_table 和相关表动态 path 场景。

XuPeng-SH (approved):
I re-reviewed the latest head after the previous request-changes comments.

The prior blockers are fixed:
- JSON literal null is no longer collapsed to SQL NULL after successful computation; `buildJsonRemove` now appends the returned `ByteJson` as a non-null JSON value.
- Quoted member `*` is now distinct from wildcard `.*` via `subPathKeyWildcard`, so `$ ."*"` semantics are handled as a normal object key while unquoted wildcard paths remain rejected by JSON_REMOVE.

I also checked the main JSON_REMOVE contract: NULL inputs return SQL NULL, `$` / wildcard / double-star paths error, missing paths are no-op, multiple paths are evaluated left-to-right, JSON typed input is handled, and dynamic table paths are covered. The unit tests and BVT cover these unhappy paths, including JSON null vs SQL NULL, quoted star, null path after an earlier path, array bounds, deep autowrap, dynamic paths, and update usage.

CI is green across UT, BVT, coverage, upgrade, compose, and benchmark jobs. The PR is currently behind main, but I do not see a code-review blocker in this head.

ouyuanning (approved):

heni02 (approved):

PR #25518, opened 2026-07-07
fix: quote publication table lookup database names

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to matrixorigin/matrixflow#11551.

Table-level publication calls `genPubTablesStr`, which validates requested table names by running `showTablesFromDb`. That helper assembled SQL as `show tables from ` + `dbName` without identifier quoting. A quoted user database name such as `目标数据` is valid in the outer `CREATE PUBLICATION ... DATABASE \`目标数据\`` statement, but the internally generated `show tables from 目标数据` fails to parse.

This PR quotes the database identifier used by publication table discovery with `sqlquote.Ident(dbName)`, so non-ASCII names and names requiring backtick escaping are handled consistently. Whole-database publication is unchanged because it uses `TABLE ALL` and does not hit this validation path.

It also adds coverage for the fix:

- frontend unit test for non-ASCII database names and embedded backtick escaping;
- BVT case for table-level publication on a Chinese database name;
- BVT create and alter publication paths;
- negative BVT assertion that a table outside the publication remains invisible to the subscriber.

## Validation

- `go test ./pkg/frontend -run 'Test_showTablesFromDbQuotesDatabaseName|Test_doCreatePublication|Test_doAlterPublication'`
- `mo-tester -m genrs -i pub_sub_chinese_db_table`
- `mo-tester -m run -i pub_sub_chinese_db_table` => `25/25`, success rate `100%`
- `git diff --check`
- GitHub Actions run `28949263234` passed all MatrixOne CI checks on this PR before the PR metadata fix; the remaining blocker was Mergify `Invalid commit message: section not found`.

### DIFF
--- a/pkg/frontend/publication_subscription.go
+++ b/pkg/frontend/publication_subscription.go
@@ -30,6 +30,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/cdc"
 	"github.com/matrixorigin/matrixone/pkg/common/moerr"
 	"github.com/matrixorigin/matrixone/pkg/common/pubsub"
+	"github.com/matrixorigin/matrixone/pkg/common/sqlquote"
 	"github.com/matrixorigin/matrixone/pkg/defines"
 	"github.com/matrixorigin/matrixone/pkg/pb/plan"
 	"github.com/matrixorigin/matrixone/pkg/publication"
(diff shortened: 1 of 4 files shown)


XuPeng-SH (approved):
Reviewed from correctness, unhappy-path, compatibility, and coverage angles. The fix is narrowly scoped to quote the database identifier in the internal `show tables from ...` query, uses the shared `sqlquote.Ident` helper, and covers both non-ASCII names and embedded backtick escaping in unit tests. The BVT covers table-level publication create/alter on a Chinese database name and verifies non-published tables are not exposed. I did not find blocking issues.

heni02 (approved):

gouhongshen:
@mergify requeue

gouhongshen:
@Mergifyio refresh

Issue #25480, opened 2026-07-06
[Bug]: TestHDFS times out on GitHub Actions — cannot reach hadoop-namenode.hadoop.svc.cluster.local:9820


matrix-meow:
## Problem

`TestHDFS` in `pkg/fileservice` times out after 10 minutes on GitHub Actions runners because the HDFS namenode at `hadoop-namenode.hadoop.svc.cluster.local:9820` is not reachable from the runner network.

```
panic: test timed out after 10m0s
running tests: TestHDFS/hdfs://hadoop-namenode.hadoop.svc.cluster.local:9820/HDFS/basic
```

All 4 HDFS sub-tests fail:
- TestHDFS (parent)
- TestHDFS/hdfs://.../HDFS
- TestHDFS/hdfs://.../HDFS/basic
- TestHDFS/hdfs://.../hadoop_namenode_hadoop_svc_cluster_local_9820

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-06 | [PR #25478](https://github.com/matrixorigin/matrixone/actions/runs/28780324744/job/85336500383) | Coverage |

## Root Cause

The HDFS test requires access to a Kubernetes-internal HDFS service (`hadoop-namenode.hadoop.svc.cluster.local:9820`). GitHub Actions runners run outside the K8s cluster and cannot resolve or connect to this address, causing the connection to hang until the 10-minute test timeout.

## Suggested Fix

1. Skip HDFS tests when `HDFS_NAMENODE_ADDR` env var is not set or the namenode is unreachable (add `t.Skip()` with connectivity check)
2. Or add a build tag (e.g., `//go:build hdfs`) to exclude HDFS tests from default CI runs
3. Or set a shorter connection timeout (e.g., 5s) before the test hangs

LeftHandCold:
fixed

Ariznawlll:
已在最新 `main` 上完成针对性验证。该问题属于 Go 单元测试/CI 环境探测，不是 SQL 行为，因此没有加入 motr SQL 回归。

**验证环境**

- MatrixOne：`main@52afd25fa63a3cb6037b7bf8ff0bf1328c4793ec`
- 本地环境无法访问 `hadoop-namenode.hadoop.svc.cluster.local:9820`，与 GitHub Actions 外部 runner 的网络条件一致。

**执行命令**

```bash
go test ./pkg/fileservice -run '^TestHDFS$' -count=1 -timeout 30s -v
```

[...]

PR #25522, opened 2026-07-08
fix: skip HDFS test without namenode

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25480

`TestHDFS` hard-codes `hadoop-namenode.hadoop.svc.cluster.local:9820`, which is a Kubernetes-internal HDFS namenode. GitHub Actions runners cannot rely on that service being reachable, so the test can hang until the package-level timeout instead of skipping cleanly.

This change keeps the existing default namenode address so environments with the in-tree HDFS service still run the integration test by default. It also allows overriding the namenode with `HDFS_NAMENODE_ADDR` and adds a short TCP reachability check before constructing the HDFS client. External CI runners now skip quickly when the namenode is unavailable, without silently removing the default in-tree HDFS coverage.

### DIFF
--- a/pkg/fileservice/hdfs_test.go
+++ b/pkg/fileservice/hdfs_test.go
@@ -27,6 +27,7 @@ import (
 )
 
 const (
+	hdfsDefaultNamenodeAddr = "hdfs://hadoop-namenode.hadoop.svc.cluster.local:9820"
 	hdfsNamenodeAddrEnv     = "HDFS_NAMENODE_ADDR"
 	hdfsNamenodeDialTimeout = 5 * time.Second
 )
@@ -146,7 +147,7 @@ func getHDFSTestAddr(t *testing.T) string {
 
 	addr := os.Getenv(hdfsNamenodeAddrEnv)
 	if addr == "" {
-		t.Skipf("set %s to run HDFS integration test", hdfsNamenodeAddrEnv)
+		addr = hdfsDefaultNamenodeAddr
 	}
 
 	if !strings.HasPrefix(addr, "hdfs://") {
--- a/pkg/fileservice/hdfs_test.go
+++ b/pkg/fileservice/hdfs_test.go
@@ -18,8 +18,17 @@ import (
 	"context"
 	"fmt"
 	"math/rand/v2"
+	"net"
+	"net/url"
+	"os"
 	"strings"
 	"testing"
+	"time"
+)
+
+const (
+	hdfsNamenodeAddrEnv     = "HDFS_NAMENODE_ADDR"
+	hdfsNamenodeDialTimeout = 5 * time.Second
 )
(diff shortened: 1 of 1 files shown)


XuPeng-SH (changes requested):
I think this needs `REQUEST_CHANGES` as-is because it turns the only in-tree HDFS integration test into an effectively unreachable opt-in test without landing any matching repo wiring.

After this patch, `TestHDFS` immediately skips unless `HDFS_NAMENODE_ADDR` is set. I reran `go test -v ./pkg/fileservice -run TestHDFS -count=1` in the PR worktree and it now exits with:

- `set HDFS_NAMENODE_ADDR to run HDFS integration test`

I also searched the current PR tree and the new variable only exists in `pkg/fileservice/hdfs_test.go`; there is no paired update in workflow/config/docs/optools that would keep the HDFS test reachable anywhere in-tree.

So although the intent is understandable (don’t fail when no namenode is reachable), the practical effect of this PR by itself is to silently drop HDFS coverage by default. I think this should either keep the old built-in namenode address as a fallback, or land the CI/test-harness/doc wiring for `HDFS_NAMENODE_ADDR` in the same PR.

LeftHandCold:
Addressed the review in `3abfb77a16`.

The concern was valid: the previous version made `TestHDFS` skip whenever `HDFS_NAMENODE_ADDR` was unset, which would remove the default in-tree HDFS coverage. The test now keeps `hdfs://hadoop-namenode.hadoop.svc.cluster.local:9820` as the default namenode, allows `HDFS_NAMENODE_ADDR` as an override, and performs a bounded 5s reachability check before constructing the HDFS client so external runners skip quickly when the namenode is unavailable.

XuPeng-SH (approved):
Reviewed latest head 06deb7b. The previous blocker is addressed: TestHDFS no longer becomes env-only coverage, because it keeps the original in-tree default namenode and only uses HDFS_NAMENODE_ADDR as an override. The added TCP reachability gate is bounded and runs before constructing the HDFS client, so external runners skip quickly instead of hanging until package timeout.

I checked the unhappy paths around unset env, missing/unreachable default namenode, invalid/missing host handling, and preservation of the existing HDFS test body once the namenode is reachable. Local targeted check passed on the PR head with the project CGo flags: env -u HDFS_NAMENODE_ADDR go test ./pkg/fileservice -run TestHDFS -count=1 -timeout 30s -v skipped in under a second on an environment without the namenode. CI on this head is green. No blocking correctness, robustness, or performance issue found.

The PR is currently behind main, so it still needs the branch/update requirement handled before merge.

PR #25527, opened 2026-07-08
fix: allow prepared JSON ordering params

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25502

Prepared JSON ordering predicates such as `json_expr >= ?` were rejected during binding because the dynamic parameter kept its default text type while JSON ordering is only supported through numeric comparison.

This change detects direct dynamic parameters on either side of JSON ordering operators, casts the JSON expression to `float64`, and assigns the parameter the same type. Literal string comparisons remain rejected.

Tested with:

`go test -count=1 -mod=mod ./pkg/sql/plan -run TestBindFuncExprImplByPlanExpr_JsonOrderingWithDynamicParam`


ouyuanning (approved):

aptend (changes requested):
The direct parameter must not be retagged as `float64` without an actual cast.

For a non-NULL protocol parameter such as `1`, COM_STMT_EXECUTE stores the value in the text prepare-parameter vector. The new code leaves the operand as a direct `Expr_P` but changes its plan type to `T_float64`. At execution, `ParamExpressionExecutor.Eval` therefore calls `vector.NewConstBytes(T_float64, []byte("1"), ...)`, and `SetConstBytes` tries to write a `types.Varlena` into an 8-byte float64 allocation, causing:

```text
panic: unsafe slice cast: from length 8 is not enough for length 24
```

This affects both JSON-on-left and JSON-on-right branches, so the original prepared-statement scenario still fails at execution. Please preserve the parameter as `T_text` and wrap it with `appendCastBeforeExpr(..., float64)`, just as the JSON operand is wrapped. Also add an execution/protocol-level regression test; the current binder-only test asserts the unsafe direct `Expr_P` shape and cannot catch this panic.

aptend (commented):
Self-review结果：PR #25527 仅涉及 prepare 参数解析/JSON 顺序比较绑定与新增 UT。已核对 binder/执行链路闭环与失败路径，未发现阻塞级缺陷（按 mo-self-review 结论为可合入）。建议 approve。

aptend (approved):
LGTM: 自审通过，无阻塞级发现。

XuPeng-SH (changes requested):
## Request changes

### High: the new `FLOAT64` normalization returns wrong results for valid JSON numbers

`adjustJsonOrderingDynamicParamType` casts both operands to `FLOAT64`. The binary protocol preserves a `MYSQL_TYPE_LONGLONG` / unsigned integer parameter as its exact decimal text, but the generated casts then round both that text and an exact JSON integer.

Concrete regression:

```sql
SELECT JSON_EXTRACT('{"v":9007199254740992}', '$.v') < ?
-- binary parameter: MYSQL_TYPE_LONGLONG, value 9007199254740993
```

The predicate must be true. With this patch both casts produce `9007199254740992` as `float64`, so it evaluates to false. The same loss occurs for high-precision decimal JSON values. This conflicts with the existing `pkg/container/bytejson/bytejson_compare_test.go` contract that values beyond the exact float range must not collapse, and with the PR issue's MySQL-compatibility goal.

The fix needs an exact comparison path rather than choosing an approximate common type. Preserve the parameter's JSON/numeric representation through execute and compare JSON numeric scalars exactly (including `int64`, `uint64`, and decimal), or add a dedicated exact JSON ordering comparator. Do not coerce both sides to `FLOAT64`.

Please add protocol-level regression coverage for the adjacent `2^53` integers above, a high-precision decimal pair, `NULL`, invalid `MYSQL_TYPE_VAR_STRING`, and both operand orders. The current tests only prove that the binder emits casts and that `"1"` can be cast; they do not execute the complete prepared comparison semantics.

I also ran the PR's added plan/colexec/frontend tests plus the relevant JSON exact-comparison and cast tests locally.

aunjgr (approved):

XuPeng-SH (approved):
Previous precision blocker is fixed in the latest head.

The new path no longer collapses JSON ordering through FLOAT64: direct prepared params are normalized via `__mo_json_ordering_param`, binary-protocol values keep their exact text representation, and JSON ordering now compares exact JSON numeric scalars (including adjacent >2^53 integers, uint64 edge values, decimals, null, invalid input, and both operand orders) instead of approximate float casts.

I re-checked the binder/operator/runtime chain and I don't see a remaining high-confidence correctness blocker in this PR.

Issue #25419, opened 2026-07-03
[Bug]: Prepared DML hangs on second execute when foreign-key actions are triggered


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Searched open issues for:

- `prepared foreign key cascade hang`
- `"PREPARE" "foreign key"`
- `"ON UPDATE CASCADE" "prepared"`
- `"ON DELETE SET NULL" "prepared"`

No matching open issue was found.

### Branch Name

main

### Commit ID

`51656deb6003`

### Other Environment Information

- MatrixOne: `select version()` returns `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- Tested through the MySQL client with server-side prepared statements.

### Actual Behavior

A prepared DML statement that triggers a foreign-key action hangs on the second `EXECUTE`.

Observed cases:

1. `ON UPDATE CASCADE` with a composite foreign key:
   - First `EXECUTE` updates parent `(1,'a1') -> (1,'a1x')` and cascades to child rows.
   - Second `EXECUTE` for parent `(2,'b1') -> (2,'b1x')` does not return within 30 seconds.
   - After killing only the mysql client, the data remains at the state after the first execute: `(2,'b1')` is unchanged in both parent and child tables.

2. `ON DELETE SET NULL`:
   - First `EXECUTE` deletes parent `100` and sets child `parent_id` to `NULL`.
   - Second `EXECUTE` for parent `200` does not return within 30 seconds.
   - After killing only the mysql client, parent `200` and its child reference remain unchanged.

Control checks on the same MatrixOne build:

- The same `ON UPDATE CASCADE` statements executed as normal SQL literals finish successfully.
- The same `ON DELETE SET NULL` statements executed as normal SQL literals finish successfully.
- The prepared hang still reproduces when the second parameter uses the same integer type, so this is not only a string-to-int coercion issue.

### Expected Behavior

Same as MySQL: repeated `EXECUTE` of the prepared statement should finish normally, and each foreign-key action should be applied.

For `ON UPDATE CASCADE`, MySQL returns:

```text
tenant  acct  status
1       a1x   open
2       b1x   hold

tenant  acct  child_id  note
1       a1x   10        c-a1-10
2       b1x   20        c-b1-20
```

For `ON DELETE SET NULL`, MySQL returns:

```text
child_id  parent_id  note
1         NULL       s100-a
2         NULL       s200-a
```

[...]

PR #25528, opened 2026-07-08
fix: avoid cached compile for prepared FK actions

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25419

Prepared `UPDATE` / `DELETE` statements that trigger foreign key actions can reuse the cached prepared `Compile` across executions. For FK side-effect plans such as `CASCADE`, `SET NULL`, and `SET DEFAULT`, that reuse can retain execution state that should be rebuilt per execution.

This change skips prepared compile caching only for `UPDATE` / `DELETE` plans that include triggered FK actions. Prepared statements without FK side effects keep the existing cache behavior.

Tested with:

`go test -count=1 -mod=mod ./pkg/frontend -run TestShouldCachePrepareCompileForeignKeyActions`


XuPeng-SH (changes requested):
I think the direction is reasonable, but this still needs changes before merge.

1. `shouldCachePrepareCompile` currently treats `TableDef.Fkeys` metadata as proof that the current UPDATE/DELETE plan contains a triggered FK action. That is too broad. `UpdateCtxList` is also used for the normal target table (and index table) of ordinary UPDATE/DELETE plans, not only for FK side-effect work. For example, a prepared UPDATE/DELETE against a child table that merely owns an FK with `ON UPDATE CASCADE` or `ON DELETE SET NULL` will now lose prepared compile caching even though no parent-to-child action can be triggered by that statement. The planner's actual FK action branches are the `SET_NULL` / `CASCADE` paths in `buildDeletePlans`; the cache decision should be based on that real side-effect being generated, not on generic FK metadata on every affected table. Please add an explicit plan signal/annotation from the FK action construction path, or otherwise distinguish FK-recursion/side-effect update contexts from ordinary DML targets before disabling the cache.

2. The regression coverage is not end-to-end enough for the issue being fixed. The new test hand-builds a synthetic `Plan` where `UpdateCtxList.TableDef.Fkeys` has the desired action, so it only validates the new predicate's assumed shape. It does not prove that the real planner produces that shape for the two reported hangs, nor that the second `EXECUTE` completes and applies the FK action. Please add a BVT or integration/frontend regression that runs the issue SQL for repeated prepared parent `UPDATE ... ON UPDATE CASCADE` and prepared parent `DELETE ... ON DELETE SET NULL` and asserts both executions finish and produce the expected parent/child rows. That is the closure needed for a hang fix.

Non-blocking: `SET_DEFAULT` is included in the no-cache actions, while the current DML builder groups it with `NO_ACTION`/`RESTRICT` in the parent check path. If that is intentional/conservative, please document it; otherwise align the predicate with the implemented action semantics.

XuPeng-SH (approved):
Reviewed latest head eb92b9d220fc7eab9594fd242c679927210a79a0. The previous blockers are addressed: the cache decision is driven by the planner side-effect signal, ordinary child-table FK DML stays cacheable, and BVT now covers repeated prepared UPDATE CASCADE and DELETE SET NULL executions. CI is green. Non-blocking: the PR description should avoid claiming SET DEFAULT is handled as a side-effect action, since current planner groups it with RESTRICT/NO_ACTION.

ouyuanning (approved):

heni02 (approved):

XuPeng-SH (approved):
Reviewed this from the only thing that really matters here: whether the new `HasForeignKeyAction` bit is set exactly on the prepared DML plans whose FK side-effect subplans retain execution state and therefore must not reuse a cached `Compile`.

I don't see a blocking issue in the current change set.

What I checked:
- the frontend cache decision now gates both PREPARE-time and EXECUTE-time compile reuse on the same helper
- the planner flag is carried through protobuf / deepcopy, so the cacheability decision survives plan transport and cloning
- the flag is set on the actual side-effect paths in `buildDeletePlans` (`SET NULL` / `CASCADE`) and not on the ordinary child-table DML path
- the added tests cover both the pure helper decision and the planner-side flag generation, and the BVT covers repeated EXECUTE behavior for parent UPDATE CASCADE and parent DELETE SET NULL

Approving.

aunjgr (approved):

PR #25529, opened 2026-07-08
fix: rank rollup window results after union

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25499

`GROUP BY ... WITH ROLLUP` is expanded into multiple `UNION ALL` branches. When the select list also contains window functions, the previous plan copied those window expressions into every rollup branch, so `ROW_NUMBER`, `RANK`, and `DENSE_RANK` were computed per grouping level instead of over the complete rollup result.

This change rewrites supported rollup/cube window queries as an inner rollup union without window expressions plus an outer select that computes the windows over the full union output. The rewrite is conservative: it only triggers when window dependencies can be mapped to inner output aliases or bare columns; unsupported shapes keep the existing path.

Tested with:

`go test -count=1 -mod=mod ./pkg/sql/plan -run 'TestRollupWindowRanksAfterRollupUnion|TestSingleTableSQLBuilder'`


heni02 (approved):

ouyuanning (approved):

aunjgr (approved):

XuPeng-SH (changes requested):

The new rewrite only works when every non-window expression has a user-visible alias or is a bare column, and when every window dependency can be mapped to those visible output expressions. If that mapping fails, the code falls back to the old rollup/cube expansion path, which is the path that computes windows independently inside each UNION ALL branch. For a result-correctness bug, that fallback is not conservative: it silently preserves the known wrong semantics.

A minimal valid case is still not fixed:

```sql
select
  l_returnflag,
  l_linestatus,
  sum(l_quantity),
  row_number() over (order by sum(l_quantity) desc) as row_num
from lineitem
group by l_returnflag, l_linestatus with rollup;
```

Because `sum(l_quantity)` has no alias, `buildRollupWindowSelectExprs` rejects the rewrite at the alias gate in `query_builder.go`. The planner then uses the old branch-local window plan. I verified this locally with a scratch planner test on commit `f84c64cb5cfc0f76c2b93d44b106c78a26c99341`: the plan has 3 `WINDOW` nodes and 0 `WINDOW` nodes above `UNION_ALL`; the correct shape should have a single outer `WINDOW` over the full rollup union.

Another common valid shape has the same problem:

```sql
select
  l_returnflag,
  l_linestatus,
  row_number() over (order by sum(l_quantity) desc) as row_num
from lineitem
group by l_returnflag, l_linestatus with rollup;
```

Here the aggregate is a hidden window dependency, not a visible select item, so it also cannot be mapped by the current alias-only approach.

First-principles fix: the rollup/cube window rewrite should not depend on user-provided aliases. It should synthesize internal aliases for inner expressions, collect the expressions needed by window args / PARTITION BY / ORDER BY / frames and final ORDER BY, add hidden inner projections when needed, and keep the outer projection limited to the original result columns. If some expression shape really cannot be rewritten safely, do not fall back to the known incorrect branch-local window plan; return an explicit unsupported error instead of producing wrong ranks.

Please also add regressions for:
- aggregate output without alias used by a window order key;
- aggregate used only inside the window spec and not selected;
- the same cases for `rank()` / `dense_rank()` over rollup output.

The existing CI is green and the included planner tests pass, but they only cover the aliased happy path.

XuPeng-SH (changes requested):
## Findings

### High: the rewrite conflates source-column identity with SELECT/output aliases and can silently change the window key or reject a previously valid query

[`lookupExprAlias`](https://github.com/matrixorigin/matrixone/blob/c4cf8a5eef4516bc694403bcf71679cd2953354c/pkg/sql/plan/query_builder.go#L3489-L3500) falls back from a structural-expression lookup to `nameAliases`. `buildRollupWindowSelectExprs` populates that map with every SELECT alias, then uses the same lookup from `ensureInnerExpr` and while rewriting the window specification. A SELECT alias is not a source-column identity, and a window `ORDER BY` is bound in the source/group scope.

This accepted query demonstrates both effects:

```sql
select n_name as n_regionkey, n_regionkey, sum(n_nationkey),
       row_number() over (order by n_regionkey) as rn
from nation
group by n_name, n_regionkey with rollup;
```

Without the rollup rewrite, the window key resolves to `nation.n_regionkey` (an integer). On `c4cf8a5e`, the alias map makes the second selected `n_regionkey` reuse the synthetic column for `n_name`; the outer window consequently orders by that varchar column, and the final result projects `n_name` twice. This is a silent wrong-result regression, not just an explain-shape difference.

There is also a hard failure when the window output uses the same name as the source column:

```sql
select n_name, n_regionkey, sum(n_nationkey),
       row_number() over (order by n_regionkey) as n_regionkey
from nation
group by n_name, n_regionkey with rollup;
```

The base commit `71aa96600` builds plans for both queries. The PR head returns `invalid input: column n_regionkey does not exist` for the second one because the output-alias entry prevents the source column from being materialized in the inner UNION.

First-principles fix: keep separate, scope-aware mappings. Structural/source expressions may share a synthetic inner projection, while SELECT labels may be resolved only in clauses where the normal binder permits output aliases. Do not use the output-name map to deduplicate a selected source expression or to rewrite window arguments, `PARTITION BY`, `ORDER BY`, or frame expressions. Preserve the binder's source-scope resolution for those expressions; use a distinct output-alias mapping for final `ORDER BY`/other alias-aware clauses as appropriate.

Please add regression coverage that validates both the result columns and the window sort key for a source-column/SELECT-alias collision, plus the same-name window-output case. The existing tests cover generated aliases and hidden aggregates, but neither collision path.

XuPeng-SH (changes requested):
Blocking issue:

- The new rollup-window rewrite changes `HAVING` name resolution precedence. In normal planning, `HAVING` uses `AliasAfterColumn`, so a bare name resolves to a real grouped/source column before falling back to a select alias. But the rewrite path feeds `HAVING` through `rewriteRollupWindowExprWithNameAliases(..., rewriteState.havingAliases)`, and `lookupExprAlias()` consults those aliases first for one-part names.

That means a query like:

```sql
select
  sum(n_nationkey) as n_regionkey,
  n_regionkey,
  row_number() over (order by n_regionkey)
from nation
group by n_regionkey with rollup
having n_regionkey > 0;
```

can have the rewritten inner `HAVING` bound to the aliased aggregate `sum(n_nationkey)` instead of the original grouped column `n_regionkey`. So the rewrite is not semantics-preserving for alias/source-name collisions.

Suggestion:

- Preserve the normal `AliasAfterColumn` behavior when rewriting `HAVING`: only use the select-alias remap if the bare name does not resolve to an original grouped/source expression.
- Add a regression for a rollup window query where a select alias intentionally collides with a grouped column name and `HAVING` references that bare name.

XuPeng-SH (changes requested):
This still needs changes.

The new HAVING precedence fix only handles source expressions recorded under one-part names. In pkg/sql/plan/query_builder.go, both addGroupedSourceExprs and addExprAlias populate sourceNameAliases only when the grouped/source expression is an UnresolvedName with NumParts == 1. rewriteRollupWindowExprWithFallbackNameAliases then relies on that map to preserve AliasAfterColumn in HAVING.

That leaves a valid query misbound whenever the grouped source column is written as a qualified name but HAVING references the bare column name:

select sum(t.n_nationkey) as n_regionkey,
       t.n_regionkey,
       row_number() over (order by t.n_regionkey) as rn
from nation t
group by t.n_regionkey with rollup
having n_regionkey > 0;

On latest head afc33b658b, I reproduced this with a scratch planner test: the rollup FILTER left argument is typed as T_int64 (the aggregate alias) instead of T_int32 (the grouped source column). So the rewrite still changes HAVING semantics for alias/source collisions; it just no longer does so for the unqualified case covered by the new tests.

Suggestion: preserve bare-name source resolution for qualified grouped/source columns as well. Record the terminal column name from materialized grouped/source expressions into the AliasAfterColumn source map, but only when that bare name is unambiguous; otherwise leave it unresolved so ambiguity is surfaced normally instead of silently picking one alias. Please add a regression for qualified source columns plus a colliding SELECT alias in HAVING.

XuPeng-SH (changes requested):
Blocking correctness issue on the latest head `c8a9c20667`:

### HAVING source-first resolution only sees grouped expressions, not the complete FROM scope

`addGroupedSourceExprs` records terminal names from materialized GROUP BY expressions in `sourceNameAliases`, and `lookupExprAlias` then treats a name present in that map as an unambiguous source-column match before consulting the SELECT alias. This approximates `AliasAfterColumn` using only the grouped expressions. The normal binder resolves against every binding in the FROM scope first, including ambiguous and non-grouped source columns.

A deterministic counterexample is:

```sql
select n1.n_regionkey,
       row_number() over (order by n1.n_regionkey) as rn
from nation n1
join nation n2 on n1.n_nationkey = n2.n_nationkey
group by n1.n_regionkey with rollup
having n_regionkey > 0;
```

On this head, the same query without the window function correctly returns an ambiguous column reference error for `n_regionkey`.

Adding only `row_number()` activates the rewrite and the query is accepted. The rewrite has seen only grouped `n1.n_regionkey`, marks the terminal name as unique, and replaces the bare HAVING reference with the internal alias for `n1.n_regionkey`. The original ambiguity with `n2.n_regionkey` never reaches `BindContext.bindingByCol`.

This is a silent semantic change: adding a window function changes name resolution and turns an ambiguous query into a plan that implicitly selects one source column.

Please preserve `AliasAfterColumn` using the actual pre-rewrite binding scope rather than reconstructing source identity from GROUP BY AST nodes. A SELECT alias may be used only after normal source lookup reports no matching column; ambiguity and grouped-column validation errors must propagate unchanged. Please add a regression with two joined tables exposing the same bare column name where only one qualified column is grouped, and verify that adding the window function does not change the error.

The previous blockers around unaliased aggregates, hidden window dependencies, SELECT/source alias collisions, qualified grouped columns, and branch-local ranking look fixed. The focused rollup-window suite passes, and I found no additional high-confidence performance, leak, hang, or unbounded-growth blocker.

XuPeng-SH (changes requested):
Two blocking correctness regressions remain on the latest head `766f6dd1d`.

1. `rewriteRollupWindowSelect` now calls `selectClause.Having.Expr.Accept(...)` at `query_builder.go:3571`, but the tree visitor closure is not safe for valid HAVING expressions. A searched CASE has a nil `CaseExpr.Expr`, while `CaseExpr.Accept` unconditionally calls `node.Expr.Accept`, so this valid query panics with a nil dereference:

```sql
select n_regionkey,
       row_number() over (order by n_regionkey) as rn
from nation
group by n_regionkey with rollup
having case when n_regionkey > 0 then 1 else 0 end = 1;
```

Prepared HAVING is another deterministic path: `having sum(n_nationkey) > ?` reaches `ParamExpr.Accept`, which explicitly panics with `unimplement ParamExpr Accept`. Both shapes plan successfully on the PR merge-base `0977fc07c`. Please avoid the generic visitor here unless its full expression closure is nil-safe and implemented, and add CASE plus prepared-parameter regressions.

2. `exprAliases` treats `tree.String(expr)` as a common-subexpression key at lines 3714-3752. That silently merges independent volatile expressions. For example:

```sql
select rand() as r1, rand() as r2,
       row_number() over (order by sum(n_nationkey)) as rn
from nation
group by n_regionkey with rollup;
```

On this head both outer outputs resolve to the same inner column `[0,0]`, so one `rand()` evaluation is returned twice. The merge-base plan contains two distinct `rand` function expressions per rollup branch, and the function registry marks the overload `volatile: true`. The same issue applies when a volatile occurrence is shared with a window key or final ORDER BY, and to functions such as `uuid()`.

Please preserve expression-occurrence identity, or only reuse expressions proven deterministic after binding. Add a regression asserting that duplicate volatile SELECT occurrences do not share an inner projection, plus a case where the window specification contains its own volatile occurrence.

XuPeng-SH (approved):
Reviewed latest head `90191eab7002288c39436f4bb3033a8f7434192f`.

The previous blocking issues are resolved:
- rollup/cube windows are evaluated once above the complete `UNION ALL`;
- unaliased and hidden aggregate dependencies are materialized safely;
- SELECT aliases, source-column identity, final ORDER BY aliases, and HAVING `AliasAfterColumn` precedence remain scope-aware, including qualified and ambiguous source cases;
- unsupported expression shapes fail explicitly instead of falling back to branch-local window evaluation;
- HAVING no longer relies on unsafe generic AST visitors;
- volatile expressions preserve AST-occurrence identity rather than being merged by SQL text.

I also reviewed the full PR diff for correctness, unhappy paths, boundedness, and performance. The added outer window/project is limited to ROLLUP/CUBE queries containing windows and is required for correct global ranking; aggregate binding still deduplicates equivalent aggregate work. No concurrency, lifecycle, leak, hang, or unbounded-growth path is introduced.

Local verification on macOS:
- focused rollup-window planner tests: PASS
- focused tests with `-count=10`: PASS
- full `./pkg/sql/plan`: PASS
- `go vet ./pkg/sql/plan`: PASS
- `git diff --check`: PASS

The latest GitHub UT, SCA, pessimistic BVT, proxy multi-CN BVT, and coverage checks are green.

PR #25532, opened 2026-07-08
[4.1] fix publication table lookup for quoted database names

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue matrixorigin/matrixflow#11551

4.1-dev port of matrixorigin/matrixone#25518.

This ports the publication/subscription fix to 4.1-dev.

Root cause: table-level publication refresh builds `show tables from <db>` by raw string concatenation. Database names that require identifier quoting, such as Chinese database names or names containing backticks, produce invalid SQL or the wrong identifier.

Changes:
- Quote the database name in `showTablesFromDb` with `sqlquote.Ident`.
- Add a unit test covering Chinese database names and embedded backticks.
- Add a BVT case for table-level publication from a Chinese-named source database, including table filtering checks so unpublished tables remain hidden.

Validation:
- `make build`
- `go test -v -count=1 -timeout 120s ./pkg/frontend -run 'Test_showTablesFromDbQuotesDatabaseName'` with MatrixOne CGo flags
- `git diff --check upstream/4.1-dev...HEAD`

Note: local BVT execution against a temporary 4.1 MO did not complete because the local temporary MO process exited during mo-tester cleanup. The added BVT case itself already passed on the main PR CI/local run, and this PR includes it for 4.1 CI coverage.

### DIFF
--- a/pkg/frontend/publication_subscription.go
+++ b/pkg/frontend/publication_subscription.go
@@ -30,6 +30,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/cdc"
 	"github.com/matrixorigin/matrixone/pkg/common/moerr"
 	"github.com/matrixorigin/matrixone/pkg/common/pubsub"
+	"github.com/matrixorigin/matrixone/pkg/common/sqlquote"
 	"github.com/matrixorigin/matrixone/pkg/defines"
 	"github.com/matrixorigin/matrixone/pkg/pb/plan"
 	"github.com/matrixorigin/matrixone/pkg/publication"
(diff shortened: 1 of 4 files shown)


XuPeng-SH (approved):
Reviewed the 4.1-dev port. The fix is the right direction: the table-level publication path validates explicit table lists via showTablesFromDb, and that SQL needs identifier quoting rather than raw concatenation. The caller passes a database identifier value, not a prebuilt SQL fragment, so sqlquote.Ident(dbName) is appropriate and does not affect DATABASE * because genPubTablesStr returns before calling showTablesFromDb for TABLE *.

Coverage looks adequate for this bug: the unit test covers non-ASCII names and embedded backticks at the generated SQL boundary, and the BVT covers create publication, alter publication, subscriber visibility, and the negative case that an unpublished table remains hidden. CI is green for UT, BVT, coverage, upgrade, and benchmark jobs.

No blocking findings.

heni02 (approved):

Issue #25533, opened 2026-07-08
Move appendable object creation outside transactions


jiangxinmeng1:
## Summary

- Create appendable objects as committed catalog entries from the appender path using `TxnMgr.Now()`, without attaching them to the active transaction.
- Rebuild missing appendable objects during WAL replay from append command row IDs, then normalize their create timestamp after replay using the minimum observed commit timestamp.
- Preserve apply-table-data semantics for appendable stats and reject unexpected transactional appendable-object create entries while keeping old WAL replay compatible.
- Added/updated tests around appendable object replay and apply-table-data behavior.

PR #25534, opened 2026-07-08
 Move appendable object creation outside transactions

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25533 

  - Create appendable objects as committed catalog entries from the appender path using TxnMgr.Now(), without attaching them to the active transaction.
  - Rebuild missing appendable objects during WAL replay from append command row IDs, then normalize their create timestamp after replay using the minimum observed commit timestamp.
  - Preserve apply-table-data semantics for appendable stats and reject unexpected transactional appendable-object create entries while keeping old WAL replay compatible.
  - Added/updated tests around appendable object replay and apply-table-data behavior.


gouhongshen (changes requested):
## Codex automated review

Found a concurrent-flush correctness regression in tombstone range scanning.

### P1 - Do not discard the visible C entry while its drop is still active (pkg/vm/engine/tae/tables/table_scan.go:99)
This skips a C entry whenever a D counterpart exists, then line 110 also skips that D entry while its deletion is active (`DeletedAt == UncommitTS`). `WaitTombstoneObjectCommitted(end)` does not wait for active transactions (`NeedWaitCommitting` returns false unless the txn is already committing), so a concurrent flush/compaction can install the D entry and leave both versions excluded. The pre-change code scanned the C entry in this state. As a result, committed tombstones in that appendable object can be omitted from `collectDelsAndTransfer`, so deletes are not transferred to the newly flushed object and deleted rows can become visible again. Only suppress the C entry once its D counterpart is committed (and add a concurrent active-drop coverage case).

gouhongshen (changes requested):
## Codex automated review

A committed appendable-object creation can now be lost on crash before checkpoint because it no longer produces a WAL catalog entry or an append record to reconstruct it.

### P1 - Persist standalone appendable-object creation before returning commit success (pkg/vm/engine/tae/txn/txnimpl/table.go:745)
`CreateObject`/`CreateObjectWithOpt` now create a committed catalog entry directly, but deliberately do not add it to `txnEntries` (the new test asserts that length is zero). Therefore a transaction that only creates an appendable object successfully commits with no object catalog command in its WAL. Recovery only synthesizes such objects from `AppendCmd`/append-node records, so after a crash before the next checkpoint there is no record to recreate a standalone object (including a caller-supplied object ID). Before this change, `createObject` appended the metadata entry and its catalog command was WAL-replayed. Preserve a durable creation record, or explicitly make this API operation transactional/durable again.

jiangxinmeng1:
> ## Codex automated review
> A committed appendable-object creation can now be lost on crash before checkpoint because it no longer produces a WAL catalog entry or an append record to reconstruct it.
> 
> ### P1 - Persist standalone appendable-object creation before returning commit success (pkg/vm/engine/tae/txn/txnimpl/table.go:745)
> `CreateObject`/`CreateObjectWithOpt` now create a committed catalog entry directly, but deliberately do not add it to `txnEntries` (the new test asserts that length is zero). Therefore a transaction that only creates an appendable object successfully commits with no object catalog command in its WAL. Recovery only synthesizes such objects from `AppendCmd`/append-node records, so after a crash before the next checkpoint there is no record to recreate a standalone object (including a caller-supplied object ID). Before this change, `createObject` appended the metadata entry and its catalog command was WAL-replayed. Preserve a durable creation record, or explicitly make this API operation transactional/durable again.

@gouhongshen 
Thanks for pointing this out. Standalone appendable-object creation is intentionally
  not a durability boundary here.

  An appendable object with neither an append record nor a delete entry contains no
  durable table data. It is only an empty allocation artifact, so it is safe to discard
  it during recovery, even when the object ID was supplied by the caller.

  The cases that do carry durable state are recovered as follows:

  1. If the object has an append record, WAL replay reconstructs the missing appendable
  object with the original object ID before applying the append.
  2. If the object has a delete entry but no append record, object-command replay now
  detects the missing appendable C entry, reconstructs it from the delete command
  metadata, and then applies the D entry. This preserves the original object ID, create
  timestamp, object stats, and committed-drop state.

  `TestFlushEmptyAppendableObjectReplay` covers both WAL-only recovery and checkpoint-
  plus-WAL recovery for an empty appendable object that is subsequently dropped. Both
  paths verify that the dropped object is recovered correctly.

  Therefore, a crash before checkpoint does not lose committed rows or a committed object
  deletion. Only a live empty appendable object with no append and no delete is omitted,
  which is intentional because it carries no durable state.

gouhongshen (changes requested):
## Codex automated review

Found one blocking catalog timestamp publication race and one non-blocking flush-scan performance regression. Prior review findings were not repeated: the active-drop bug is fixed, and the empty-object durability behavior is intentional and covered.

### P1 - Publish committed object creation atomically with timestamp allocation (pkg/vm/engine/tae/txn/txnimpl/table.go:745)
`Now()` releases the timestamp lock before the catalog entry is inserted and `MaxCommittedTS` is advanced. Two concurrent creations can therefore allocate t1<t2, publish t2, allow a transaction to start at t2, and only then publish the t1 object. That transaction can omit the object from an iterator created before publication but see it from a later iterator, even though the object is marked committed at t1 and should already belong to its snapshot. The concurrent CAS fix only keeps `MaxCommittedTS` monotonic; it does not serialize timestamp allocation with catalog publication. Allocate, insert, and publish through one TxnManager operation so no later snapshot/prepare timestamp can pass an unpublished committed entry.

### P2 - Stop tombstone scans after the remaining lifetimes precede the range (pkg/vm/engine/tae/tables/table_scan.go:97)
The previous scan stopped after reaching the relevant appendable object below `start`; this loop now traverses the entire tombstone object history. `ObjectList` is ordered by max(CreatedAt, DeletedAt), so once that value is below `start`, every remaining entry is also outside the range and scanning can safely stop while still retaining D entries whose deletion overlaps the range. This matters because `collectDelsAndTransfer` invokes this full scan once per flushed data object, turning flush work into O(data appendable objects × historical tombstone versions).

jiangxinmeng1:
> ## Codex automated review
> Found one blocking catalog timestamp publication race and one non-blocking flush-scan performance regression. Prior review findings were not repeated: the active-drop bug is fixed, and the empty-object durability behavior is intentional and covered.
> 
> ### P1 - Publish committed object creation atomically with timestamp allocation (pkg/vm/engine/tae/txn/txnimpl/table.go:745)
> `Now()` releases the timestamp lock before the catalog entry is inserted and `MaxCommittedTS` is advanced. Two concurrent creations can therefore allocate t1<t2, publish t2, allow a transaction to start at t2, and only then publish the t1 object. That transaction can omit the object from an iterator created before publication but see it from a later iterator, even though the object is marked committed at t1 and should already belong to its snapshot. The concurrent CAS fix only keeps `MaxCommittedTS` monotonic; it does not serialize timestamp allocation with catalog publication. Allocate, insert, and publish through one TxnManager operation so no later snapshot/prepare timestamp can pass an unpublished committed entry.
> 
> ### P2 - Stop tombstone scans after the remaining lifetimes precede the range (pkg/vm/engine/tae/tables/table_scan.go:97)
> The previous scan stopped after reaching the relevant appendable object below `start`; this loop now traverses the entire tombstone object history. `ObjectList` is ordered by max(CreatedAt, DeletedAt), so once that value is below `start`, every remaining entry is also outside the range and scanning can safely stop while still retaining D entries whose deletion overlaps the range. This matters because `collectDelsAndTransfer` invokes this full scan once per flushed data object, turning flush work into O(data appendable objects × historical tombstone versions).

@gouhongshen 
• This performance issue cannot be safely fixed in this PR without changing the object-
  list ordering.

  The current list is ordered by catalog lifetime (max(CreatedAt, DeletedAt)), which is
  not monotonic with the commit timestamps of tombstones appended to each object. An
  earlier object may contain later committed tombstones, or still have an active
  transaction. Therefore, stopping the scan based on CreatedAt, DeletedAt, or an
  individual object’s MaxCommitTS could skip visible tombstones and cause a correctness
  regression.

  A safe early-termination condition requires ordering/indexing objects by their
  finalized maximum append commit timestamp, while handling active or unsealed objects
  separately. That object-list change is planned for the next PR. For this PR, the scan
  must traverse the full history to preserve correctness.

gouhongshen (approved):
## Codex automated review

No blocking or non-blocking findings remain. The latest head fixes the catalog timestamp publication race. The author’s newest explanation also disproves the prior tombstone early-stop suggestion because catalog lifetime does not order later append commits. Focused publication, replay, tombstone-scan, and empty-object recovery tests passed.

XuPeng-SH (changes requested):
Deep review verdict: Request Changes. Two blocking lifecycle/correctness issues remain on the latest head `24cebe70`.

1. `apply-table-data` rejects valid flushed appendable objects.

`pkg/vm/engine/tae/rpc/apply_table_data.go` treats `stats.Rows() > 0` as invalid for every appendable object. That invariant only holds for a newly created live in-memory aobject. A compacted/flushed appendable history object can legitimately carry Rows/BlkCnt/Extent. The existing `pkg/vm/engine/tae/db/test.Test_ApplyTableData` deterministically fails on both the PR head and PR+latest-main with:

```
internal error: appendable object ... has non-empty row count 2
```

The same test passes on clean latest main, so this is a PR regression. Please model live/non-persisted versus flushed/file-backed appendable objects separately; do not merely drop the validation and recreate a persisted object as an active memory appender. Add compacted dump/apply, rollback/retry, object-state, and restart/read coverage.

2. write -> replay can deadlock while draining transactions.

`handleToReplayCmd` stops `MergeScheduler` before `TxnMgr.SwitchToReadonly/WaitEmpty`, but leaves the catalog notifier attached until after `WaitEmpty`. Existing transactions can still synchronously send `OnCreateTableCommit` / `OnCreateNonAppendObject` into the stopped scheduler. `msgChan` is bounded at 4096 and has no consumer after `Stop`; once full, the transaction blocks in the notifier while `WaitEmpty` waits for that transaction. With a background context the mode switch can hang forever.

Please first prevent new work while keeping catalog events drainable, wait for in-flight transactions, then detach the notifier and stop the consumer (or implement an equivalent lossless lifecycle protocol). Add a test with an in-flight transaction producing more than 4096 events plus switch failure/rollback coverage.

The timestamp publication, WAL replay reconstruction, checkpoint range-recovery selection, tombstone lifetime scan, and empty-aobject paths otherwise look sound. CI must be rerun after these two blockers are fixed.

XuPeng-SH (changes requested):
Blocking: balance `txnWaiter` when readonly admission rejects a new CN transaction.

`SwitchToReadonly` sets all skip flags and then waits for the waiter to become empty. In `loadOrStoreTxn`, however, the waiter is incremented before the skip check; when the flag is skipped and the txn ID is not already present, the code still stores the new txn and returns it as offline without calling `Done`. Both `Txn.Commit` and `Txn.Rollback` return immediately for offline stores, so that map entry is never deleted and its waiter count can never reach zero.

This is reachable during the mode switch: the RPC server checks `TxnLocalHandle` before calling `req.exec`, so a request already dequeued before `TxnForwardWait` can enter `GetOrCreateTxnWithMeta` after readonly flags are installed. The current switch then waits until its 10-minute timeout; switching back to write mode does not remove the leaked entry, subsequent switches keep timing out, and that txn ID remains poisoned as offline.

I reproduced it with a focused test that sets `TxnFlag_Normal`, calls `loadOrStoreTxn` with an absent ID, and then waits with a 50 ms context: `WaitEmpty` returns `context deadline exceeded`. Please make the skipped/new-ID path leave neither a managed map entry nor a waiter reference (while preserving reuse of an already managed ID), and add this case beside `TestLoadOrStoreTxnBalancesWaiterOnHit`.

The two previously reported blockers are fixed on this head: persisted appendable stats now restore transactionally, and the merge notification queue is drained while its consumer is alive.

pkg/vm/engine/tae/db/test/db_test.go
```diff
@@ -11886,6 +12028,116 @@ func Test_BasicTxnModeSwitch(t *testing.T) {
...
 
+func prepareTxnModeSwitchWithInflightTxn(
+	t *testing.T,
+) (*testutil.TestEngine, txnif.AsyncTxn, handle.Relation, string) {
+	t.Helper()
+	ctx := context.Background()
+	opts := config.WithLongScanAndCKPOpts(nil)
+	tae := testutil.NewTestEngine(ctx, ModuleName, t, opts)
+	var err error
+	tae.TxnServer, err = rpc.NewTxnServer("localhost:12345", runtime.ServiceRuntime(""))
```
XuPeng-SH: `rpc.NewTxnServer` owns a listener, but `tae.Close()` does not close `tae.TxnServer`. Both new tests call this helper with the same fixed address, so the first leaves port 12345 bound and the second deterministically fails. I reproduced the pair failing while each test passes alone. Please use an ephemeral address and register `t.Cleanup` to close the server.

XuPeng-SH (changes requested):
The earlier correctness blockers are fixed on this head: persisted appendable stats are handled safely during apply, write-to-replay drains in-flight transactions before stopping the merge scheduler, and readonly admission no longer leaks the waiter/store entry.

One deterministic test-lifecycle blocker remains. The new mode-switch helper creates a TxnServer on a fixed address and never closes it. Running the two new tests in the same package process makes the second test fail with `listen tcp4 127.0.0.1:12345: bind: address already in use`, although each test passes alone.

Reproduced with:
`go test ./pkg/vm/engine/tae/db/test -count=1 -run 'TestTxnModeSwitch(DrainsInflightCatalogEvents|WaitCancelRollsBack)$'`

Please give each test an ephemeral listener (for example `127.0.0.1:0`) and register cleanup that calls `TxnServer.Close()`. This also prevents conflict with the existing basic mode-switch test.

pkg/vm/engine/tae/db/test/db_test.go
```diff
@@ -12833,13 +13093,31 @@ func Test_ApplyTableData(t *testing.T) {
...
+		objectCount := 0
+		for it.Next() {
+			objectCount++
+			meta := it.GetObject().GetMeta().(*catalog.ObjectEntry)
+			require.False(t, meta.IsAppendable())
+			require.True(t, meta.ObjectPersisted())
+			require.False(t, meta.GetObjectData().IsAppendable())
+		}
+		require.NoError(t, it.Close())
+		require.Equal(t, 1, objectCount)
```
XuPeng-SH, line 13119: [P1] Make the new apply-table-data regression pass on a clean head

This assertion currently fails deterministically on this exact commit: `go test -count=1 -run '^Test_ApplyTableData$' ./pkg/vm/engine/tae/db/test` reports `expected: 1, actual: 2`. In this setup the compacted appendable's latest D entry and the replacement N entry are both written to the dump (`DUMP-TABLE-END` reports `object_count: 2`), then apply recreates both as live non-appendable objects. `TableEntry.RecurLoop` yields the D entry, while `onObject` only filters deletes strictly before the dump start TS, so the compaction handoff at the snapshot boundary is retained. Please filter the dump by the table snapshot's object visibility (or otherwise preserve the delete state) so apply restores only the serving object, and keep this new assertion/restart check passing.

pkg/vm/engine/tae/db/controller.go
```diff
@@ -336,12 +346,28 @@ func (c *Controller) handleToReplayCmd(cmd *controlCmd) {
...
+	rollbackSteps.Add("switch txn manager to write mode", func() error {
+		c.db.TxnMgr.ToWriteMode()
+		return nil
+	})
+
+	// No transaction can publish another catalog notification after WaitEmpty.
+	// Detach the producer, process every event already queued, and only then stop
+	// the consumer.
+	c.db.Catalog.SetMergeNotifier(nil)
+	c.db.MergeScheduler.Query(nil)
```
XuPeng-SH: [P1] Keep the scheduler drain barrier bounded by the switch context

This new barrier can wait forever after `SwitchToReadonly(ctx)` has honored its deadline. `MergeScheduler.Query` performs an unconditional send and receive with no context. The scheduler main loop also performs unconditional `a.ioChan <- ...` sends (capacity 256), while its single I/O loop can block in `CalculateVacuumStats(context.Background())` on object-store scans. If that I/O blocks and pending vacuum checks fill `ioChan`, the main loop cannot process this query (or `stopCh`), so write→replay/failover hangs indefinitely and the controller never executes rollback even after `ctx` expires. Please make the barrier/send path context- or stop-aware (and cover stalled I/O plus a full queue) so the mode-switch deadline remains effective.

XuPeng-SH (changes requested):
Deep review on exact head `624a27306527a16c2c8a6924492235448d76f86a` found two blockers: the PR's new apply-table-data regression test fails deterministically and leaves an extra live object, and the new merge-scheduler drain barrier can bypass the mode-switch context under stalled I/O/backpressure. The timestamp publication, readonly admission, replay reconstruction, empty-object flush/recovery, active tombstone-drop scan, and mode-switch race tests otherwise passed, including targeted `-race` runs.

pkg/vm/engine/tae/db/merge/scheduler.go
```diff
@@ -470,13 +472,53 @@ type todoItem struct {
...
+
+func (a *MergeScheduler) sendIOForGeneration(
+	stopCh *chan struct{},
+	msg *MMsg,
+) bool {
+	if stopCh == nil {
+		return false
+	}
+	select {
+	case a.ioChan <- msg:
```
XuPeng-SH: This does not reliably reject work from a stopped generation. Once stopCh is closed and ioChan has capacity, both select cases are ready, so Go may choose the send and enqueue the old generation vacuum work. Start() can then launch a new I/O loop over the same ioChan, which consumes that stale work. I reproduced this by closing the captured stop channel and calling sendIOForGeneration repeatedly; it enqueued on attempt 4. Please make the closed-generation check deterministic (for example, a non-blocking stop pre-check before the blocking select, or explicit generation tagging) and add a regression test asserting that a closed generation can never enqueue.

pkg/vm/engine/tae/db/merge/scheduler.go
```diff
@@ -40,6 +41,8 @@ const (
 	objectOpsTriggerThreshold = 5
 )
 
+var ErrMergeSchedulerStopped = errors.New("merge scheduler stopped")
```
XuPeng-SH: The current exact-head SCA job fails here because MatrixOne static check forbids errors.New and requires moerr. Please use the repository error constructor; this is the reported failure in Matrixone CI / SCA Test on Linux/arm64.

XuPeng-SH (changes requested):
Requesting changes on exact head 97839bc3afc6748eb5a39bba7377322cc7c6bf00.

The earlier Apply Table Data snapshot issue and the mode-switch query hang are addressed, and the focused functional/race tests pass. Two blockers remain in the latest scheduler fix: stopped-generation I/O can still cross into a restarted scheduler, and the exact-head SCA check currently fails on the newly added errors.New.

gouhongshen (changes requested):
## Codex automated review

Two issues remain on the exact head: a scheduler generation race and a nil-options panic in the new API.

### P1 - Prevent stopped-generation I/O from entering a restarted scheduler (pkg/vm/engine/tae/db/merge/scheduler.go:521)
The preliminary stop check is racy. A delayed old-generation callback can pass lines 516–520, then Stop closes its channel before the second select. With ioChan capacity available, both cases are ready and Go may select the send, placing old vacuum work on the shared queue after Stop has returned. A restarted scheduler can consume it; alternatively, the old I/O loop can run it and block trying to publish triggers with no main loop. The test only invokes the function after the channel is already closed, so it misses this interleaving.

### P2 - Return an error instead of panicking for nil CreateObjectWithOpt options (pkg/vm/engine/tae/txn/txnimpl/txndb.go:356)
The newly exposed CreateObjectWithOpt path dereferences opt via opt.WithIsTombstone without validation; passing nil through Relation.CreateObjectWithOpt panics before reaching the helper. The sibling CreateNonAppendableObject explicitly normalizes nil options at the relation boundary. Reject nil options (or provide the documented default) and add coverage so an invalid caller input cannot crash the TN.

XuPeng-SH (changes requested):
Deep review verdict: Request Changes on exact head `84f0581254a3bb2bc01099c34d73edb4348f7285`.

### P1 - Do not discard old-generation merge completion accounting (`pkg/vm/engine/tae/db/merge/scheduler.go:215`)

`doSched` records every admitted task in scheduler-wide state by incrementing `supp.mergingTaskCnt` and calling `rc.Acquire`. The new observer now tags the matching `DoneTask` message with the old generation, and `dispatchMsg` drops that message after `Stop`/`Start`. The state being balanced is not generation-local: `supps` and `rc` survive the restart.

This is reachable during write -> replay: `TxnMgr.WaitEmpty` can finish before the task observer runs because the observer is invoked by `ops.Op.SetError` only after the task implementation returns (and its transaction has already left TxnMgr). If the scheduler stops in that gap and later restarts, the completion is lost. `PendingMergeCnt` remains positive, line 1189 postpones that table forever, and the matching memory reservation is never released; repeated occurrences can also exhaust the global merge budget.

I reproduced this deterministically by admitting one task in generation N, stopping/restarting the scheduler, then firing its captured completion callback. A query on generation N+1 returns `PendingMergeCnt == 1` instead of 0. The PR test `TestStoppedGenerationIOCannotCrossRestart` only covers stale I/O and misses this persistent accounting edge.

Please either wait/cancel all admitted merge work before closing a generation, or separate generation-bound scheduling messages from generation-independent exactly-once completion accounting. Add a deterministic Stop/restart test with a blocked in-flight merge and assert both `PendingMergeCnt == 0` and the resource reservation is restored after completion.

The replay/create timestamp, apply-table-data, empty-object, tombstone scan, readonly waiter, and scheduler I/O paths otherwise passed the focused and package tests, including targeted `-race` runs.

gouhongshen (changes requested):
## Codex automated review

Previously blocking active-drop scanning, timestamp publication, apply-table-data restoration, readonly admission, mode-switch draining, test cleanup, nil-option handling, and stopped-generation I/O are fixed. Empty-object durability is accepted as intentional; the tombstone early-stop suggestion is withdrawn. One new blocking scheduler lifecycle issue remains. Focused tests could not run because cgo/libmo.dylib is absent; git diff --check passed.

### P1 - Preserve merge completion accounting across scheduler restarts (pkg/vm/engine/tae/db/merge/scheduler.go:216)
`doSched` counts every accepted asynchronous compaction with `mergingTaskCnt++` and `rc.Acquire`, but this observer sends the only corresponding DoneTask/Release notification through the task's generation. A task may still be queued when SwitchToReadonly waits for transactions, or its transaction may be removed before Op.SetError invokes observers; therefore the Query barrier can finish and Stop can close the generation first. `sendMsgForGeneration` then returns false and discards the completion. Start reuses the same supporters and resource controller, so `mergingTaskCnt` remains positive, `doSched` postpones that table forever, and the reservation is never released. A write→replay→write cycle can thus permanently disable automatic merging and eventually exhaust storage. Stop must drain all accepted task observers before closing, or completion accounting must cross generations losslessly and exactly once; add coverage with a queued/in-flight merge task during the mode switch.

gouhongshen (approved):
## Codex automated review

Previously blocking active-drop scanning, timestamp publication, apply-table-data restoration, readonly admission, mode-switch draining, test listener cleanup, nil-option validation, and stale-generation I/O are fixed. Empty-object durability is accepted as intentional behavior; the tombstone early-stop concern was withdrawn. The prior scheduler completion-accounting blocker is fixed by dcce80062: completion now balances persistent accounting exactly once across scheduler restarts, with regression coverage. No new findings. `git diff --check` passed; focused Go tests could not run because the Go toolchain is absent.

XuPeng-SH (approved):
Deep-reviewed the complete 44-file diff at dcce8006220bf91a173b8c4ce90e6789bf1059e8. No blocking findings remain.

Verified functional closure:
- appendable creation is atomically published with its catalog timestamp; nil/invalid options are rejected; transactional create entries are guarded
- WAL replay reconstructs missing appendable objects, preserves object IDs, normalizes create timestamps, and remains compatible with explicit old-WAL catalog entries
- dump/apply-table-data preserves live vs persisted appendable semantics and rollback behavior
- tombstone range scans cover active/committed drops without duplicate versions; full-history scan is an intentional correctness tradeoff
- write/replay mode switches drain catalog notifications and bound readonly waits
- merge scheduler Stop/Start isolates I/O generations while completion observers balance persistent pending-task and memory accounting exactly once across restarts

Local exact-head validation:
- focused concurrency/replay/mode-switch tests under -race with stress counts
- full -race suites for objectio, disttae/logtailreplay, tae/catalog, tae/db/merge, tae/rpc, tae/tables, tae/txn/txnbase, and tae/txn/txnimpl
- go build and go vet for the same core packages
- git diff --check

Decision log: a live empty appendable with no append/delete remains intentionally non-durable because it carries no table state; the broader tombstone history scan is accepted until a safe commit-ordered index exists.

gouhongshen (approved):
## Codex automated review

Follow-up disposition: prior active-drop, timestamp-publication, apply-table-data, readonly-admission, listener-cleanup, nil-option, stale-I/O, and scheduler-accounting blockers are fixed. Empty appendable durability remains an accepted contract; tombstone early-stop was withdrawn as unsafe, with full-history scanning accepted as a performance trade-off. One new non-blocking scheduler lifecycle gap remains.

### P2 - Rebuild merge scheduler state when leaving replay mode (pkg/vm/engine/tae/db/controller.go:815)
Replay mode now leaves the scheduler stopped and without a catalog notifier. ReplayWal runs in ReplayForever, so a table committed after NewMergeScheduler's one-time InitSource snapshot is never delivered to the scheduler. Switching to write mode only installs the notifier and calls Start; Start does not rescan InitSource, leaving that table unscheduled for automatic merging. Merge show also now fails with ErrMergeSchedulerStopped in replay mode, while the previous implementation kept the scheduler running. Add a no-gap catalog rescan/catch-up when starting write mode, or explicitly reject merge controls in replay mode.

jiangxinmeng1:
@Mergifyio refresh

Issue #25104, opened 2026-06-23
[Bug]: UPDATE of an ENUM column triggers a server-side panic (nil pointer)


dengn:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

v4.0.0-rc3

### Commit ID

8.0.30-MatrixOne-v4.0.0-rc3

### Other Environment Information

- Hardware parameters: N/A
- OS type: N/A
- Others: Reproduced via cross-language ORM compatibility testing (PHP / Python / Java / Node.js) over the MySQL 8.0.30 wire protocol. The minimal repro below uses raw SQL so it is driver-independent. Originally surfaced under Sequelize-generated `UPDATE`s.

### Actual Behavior

Updating a row that has an `ENUM` column crashes the server with error `1105` (nil-pointer panic). A client statement must never panic the server (crash / DoS-class).

```sql
UPDATE t SET status='paid' WHERE id=1;   -- observed: 1105 server panic
```

### Expected Behavior

The `UPDATE` succeeds (or returns a clean, non-fatal error); the server must not panic.

### Steps to Reproduce

```sql
CREATE TABLE t (id INT PRIMARY KEY, status ENUM('new','paid','shipped'));
INSERT INTO t VALUES (1,'new');
UPDATE t SET status='paid' WHERE id=1;   -- observed: 1105 server panic
```

(Please confirm the exact trigger; in the suite it surfaced under Sequelize ENUM updates.)

### Additional information

**Background:** dengn/php-tester#5.

fengttt:
Not repro.

ck89119:
I tested the minimal repro on current `main` (commit `a0f1006fb`) with a freshly built local `mo-service`.

SQL tested:

```sql
CREATE TABLE t (id INT PRIMARY KEY, status ENUM('new','paid','shipped'));
INSERT INTO t VALUES (1,'new');
UPDATE t SET status='paid' WHERE id=1;
SELECT id, status FROM t;
```

Result:

```text
id  status
1   paid
```

I could not reproduce the reported server-side panic / error 1105 on this build.

Ariznawlll:
Verified on latest `main` at `bd2497354`.

Result: the original server-side panic did not reproduce, but the behavior is still not compatible with MySQL because the `ENUM` update does not take effect.

SQL tested:

```sql
create table t (
  id int primary key,
  status enum('new','paid','shipped'),
  note varchar(20),
  key idx_status(status)
);
[...]

PR #25538, opened 2026-07-08
fix(plan): UPDATE of ENUM column with index panics or fails to update (#25104)

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25104

UPDATE of an ENUM/SET/Geometry column triggers server-side panic (nil pointer) or silently fails to update the value when the column is indexed.

**Root cause:** the OLD value in the UPDATE projection keeps the raw column scan type (VARCHAR display value from `cast_index_to_value`), while index tables store the typed representation (T_enum uint16). This type mismatch breaks index-maintenance joins.

### Three symptoms, one root cause:

| Scenario | Join type | Result |
|----------|-----------|--------|
| Single-col UNIQUE index | LEFT join, `=` compare | VARCHAR vs T_enum → bind fails → nil pointer **panic** |
| Secondary index | INNER join, `serial_full` | serial_full(VARCHAR) ≠ stored serial_full(uint16) → row dropped → **no update** |
| Composite UNIQUE | LEFT join, `serial_full` | mismatch but LEFT join preserves row → main table updates, but **index maintenance stale** |

### Two fixes:

1. **`bind_update.go`**: cast the OLD value of updated ENUM/SET/Geometry columns via `funcCastForEnumType`/`funcCastForSetType`/`funcCastForGeometryType`, matching what INSERT already does for index key construction.

2. **`build_index_util.go`**: preserve `Enumvalues` in `indexTableKeyTypeForSinglePart` so single-column UNIQUE index tables carry the enum metadata needed for equality binding and UNIQUE conflict detection.

### Changes:

| File | Change |
|------|--------|
| `pkg/sql/plan/bind_update.go` | +23: cast OLD value for updated ENUM/SET/Geometry columns |
| `pkg/sql/plan/build_index_util.go` | +10/-3: preserve Enumvalues in indexTableKeyTypeForSinglePart |
| `pkg/sql/plan/build_index_util_test.go` | +14: UT for enum/set preserves enumvalues |
| `test/distributed/cases/dtype/enum_index_update.sql` | +31: BVT covering 3 index scenarios |
| `test/distributed/cases/dtype/enum_index_update.result` | +46: expected results |

### Verified:

- 5 reproduce scenarios all pass (MySQL-compatible)
- UNIQUE conflict detection works (ERROR 1062)
- Index maintenance correct (old entry deleted, new inserted)
- Existing BVT `enum.sql` 262/262 no regression
- `pkg/sql/plan` full UT suite passes


XuPeng-SH (changes requested):
I found one blocker before approval.\n\nMedium: the UNIQUE-conflict half of this fix is no longer covered. The code change in build_index_util.go preserves Enumvalues specifically so single-column UNIQUE ENUM index equality / duplicate binding works, and the PR body still claims UNIQUE conflict detection is verified. But the latest BVT in test/distributed/cases/dtype/enum_index_update.sql only checks non-conflicting updates plus index lookup; commit 46301b6 removed the duplicate-key assertion entirely because optimistic and pessimistic modes report different key names. That leaves the exact panic/failure mode for UNIQUE ENUM conflicts unguarded.\n\nPlease add mode-stable unhappy-path coverage instead of dropping the assertion. For example, after enum_idx07 has id=1 status='shipped' and id=2 status='paid', run a conflicting UPDATE of id=1 to 'paid' with a broad regex such as Duplicate entry and then verify the table is unchanged. Ideally also cover the composite UNIQUE case in enum_idx08, e.g. update row 1 to the existing (note,status) pair. Existing BVT supports -- @regex, so the key-name difference can be handled with a regex matching either the user key or __mo_index_idx_col.\n\nThe implementation direction otherwise looks reasonable: casting the OLD updated ENUM/SET/geometry value closes the index-maintenance old-key join mismatch, and preserving Enumvalues on single-part index table keys explains the UNIQUE ENUM bind issue.

heni02 (approved):

XuPeng-SH (approved):
Reviewed this from the two real risk surfaces in the patch:
- tuple enum encode/decode/stringify correctness (including the non-trailing enum fast path used by `serial_extract`)
- UPDATE index-maintenance typing for ENUM/SET/geometry old/new values and single-part enum metadata propagation

I don't see a blocking issue in this change set.

The tuple fix is consistent with `EncodeEnum`'s actual byte layout, and the added tests cover both the duplicate-key stringify path and the element-after-enum unpack path. On the plan side, casting the OLD projection for updated ENUM/SET/geometry columns matches the existing INSERT/index-key construction behavior, and preserving `Enumvalues` on single-part index key types closes the enum metadata hole for equality binding.

Approving.

aunjgr (approved):

PR #25540, opened 2026-07-08
perf(trace): reduce SQLFlusher inline SQL allocation

### LINKED ISSUE #25505: [Performance] SQLFlusher trace/ETL path allocates heavily under high-concurrency sysbench workloads
## Summary

High-concurrency sysbench workloads show very large allocations from the trace/ETL `SQLFlusher` path. CPU cost is tiny (~0.006%), so this is not a direct CPU hotspot — the concern is allocation pressure and downstream GC overhead under high concurrency.

## Workloads

1. `oltp_point_select` `--db-ps-mode=auto` 1000vu (prepared)
2. `oltp_read_write` 1000vu

Both on mo-srv-129 standalone (64-core, `GOMEMLIMIT=40000MiB`), main commit `2a008285e` + closeTxn fix (#25490).

## Profile Evidence

`alloc_space` top (30s sample):

| workload | SQLFlusher.FlushBuffer flat | total alloc |
|----------|----------------------------|-------------|
| point_select auto 1000vu | 151.07GB (16.68%) | 905.86GB |
| read_write 1000vu | 156.35GB (9.90%) | 1580.01GB |

Call chain (cpu, focus FlushBuffer):
```
motrace.(*batchETLHandler).NewItemBatchHandler...
  → export.(*MOCollector).doExport
  → export.(*bufferExportReq).handle
  → export/etl.(*SQLFlusher).FlushBuffer
```

CPU cost is ~0.006% — the path spends almost no CPU, but allocates heavily for trace/ETL buffers.

read_write 1000vu GC rises to ~16% CPU (scanSpan 7.56% + tryDeferToSpanScan 5.20% + fpTracebackPartialExpand 3.25%), vs point_select's ~10%. The heavy allocation is a contributing factor.

## Impact

- Large background allocations increase GC pressure.
- read_write already shows high GC time (~16%).
- This path is independent of the closeTxn lock issue fixed by #25490, and independent of plan-cache miss (DeepCopy) — it fires under both prepared and text protocols.

## Suggested Direction

- Reuse buffers in SQLFlusher / MOCollector path.
- Reduce per-statement trace allocation.
- Consider sampling or high-concurrency degradation.
- Verify with alloc_space and GC CPU before/after.

## Reproduction

Profiles at `mo-srv-129:/data/mo-prof/pprof/1000vu_auto/heap.pprof` and `mo-srv-129:/data/mo-prof/pprof_rw/1000vu/heap.pprof`. Repro scripts `/data/mo-prof/run_bench.sh` (point_select) and `/data/mo-prof/run_rw.sh` (read_write). Related profiling task: #25472.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25505

`SQLFlusher.FlushBuffer` used to allocate a fixed `table.DefaultWriterBufferSize` 10MB intermediate buffer for every trace/ETL SQL flush before building the final `LOAD DATA INLINE` statement. Under high-concurrency sysbench workloads this became a large allocation amplifier even when the actual CSV batch was much smaller than 10MB.

This PR builds the final inline SQL directly with `strings.Builder` and sizes it from the actual batch length plus the number of bytes that need SQL escaping. This removes the fixed 10MB intermediate buffer and avoids the extra `fmt.Sprintf` copy.

Changes:
- Replace the fixed 10MB `sqlCsv` buffer in `SQLFlusher.FlushBuffer` with an exactly-sized `strings.Builder` for the final SQL.
- Preserve the existing escape behavior for `\` and `'`.
- Add a regression test for inline SQL escaping.

Validation:
- `go test ./pkg/util/export/etl/...`
- mo-srv-129 clean point_select validation, 1000vu, `--db-ps-mode=auto`, 8GB SHARED cache, fresh `sysbench_ps_sqlflusher_clean` database:

| metric | clean baseline | patched run 1 | patched run 2 |
|---|---:|---:|---:|
| TPS | 150,941/s | 169,974/s | 161,550/s |
| avg latency | 6.61ms | 5.87ms | 6.18ms |
| p95 latency | 27.17ms | 31.94ms | 27.17ms |
| CPU samples / 30s | 1602.14s (~53.4 cores) | 1620.27s (~54.0 cores) | 1604.88s (~53.5 cores) |
| heap alloc_space total | 4.62TB | 506.39GB | 1019.38GB |
| `SQLFlusher.FlushBuffer` | 223.18GB flat / 225.06GB cum | 0 flat / 30.54MB cum | 0 flat / 77.48MB cum |

Note: Go heap `alloc_space` is cumulative since process start. Patched run 2 reused the same patched MO process after run 1, so its heap total includes prior allocations. It is included as a stability check that `SQLFlusher.FlushBuffer` remains MB-level after another 1000vu run.

Profile paths:
- baseline: `mo-srv-129:/data/mo-prof/pprof/1000vu_clean_auto/`
- patched run 1: `mo-srv-129:/data/mo-prof/pprof/sqlflusher_clean_auto_1000vu/`
- patched run 2: `mo-srv-129:/data/mo-prof/pprof/sqlflusher_clean_auto_1000vu_rerun/`


gouhongshen (approved):
Reviewed the allocation change in `SQLFlusher.FlushBuffer`.

No blocking issues found. The new builder path preserves the byte-level escaping behavior for `\` and `'`, removes the fixed 10MB intermediate buffer, and leaves the existing file-flush-on-SQL-error path unchanged. PR scope is clean (`content.go` / `content_test.go` only), and GitHub checks are green.

Local note: I could not rerun `go test ./pkg/util/export/etl/...` on this workstation because the MatrixOne third-party cgo headers (`xxhash.h`, `roaring.h`, `usearch.h`) are missing here; the corresponding GitHub UT check passed.

XuPeng-SH (approved):
Reviewed this from correctness, unhappy-path behavior, and the specific allocation goal of #25505.

I checked the main risk points:
- escaping semantics for `\` and `'` stayed identical while removing the fixed 10MB intermediate buffer
- the final SQL shape is unchanged except for the allocation strategy
- failure/backoff / reconnect paths in `FlushBuffer` are untouched
- the added regression test covers the inline escaping case that is easiest to break in this refactor

I don't see a blocking issue in this change set. Approving.

PR #25543, opened 2026-07-08
fix: guard regular index ordered limit pushdown

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23965

The regular secondary index ORDER BY primary-key LIMIT optimization can push a static limit down to the scan/index reader when the secondary-index prefix is fixed.
With an extra primary-key cursor predicate, for example `id < ...`, the scan candidates no longer exactly match the SQL filter. Applying the static limit before the residual filter can return too few rows, or even zero rows.

This PR keeps the hidden-key ordering rewrite and TopValue message path, but only enables static limit pushdown when the index scan filter list is exactly the full-prefix equality filter.
Residual filters continue to be applied before the final LIMIT semantics are observed.

Added a BVT regression case:
`test/distributed/cases/optimizer/regular_index_order_limit_cursor.sql`

Tested with:

`env GOCACHE=/private/tmp/mo-go-build-cache-41 CGO_CFLAGS='-I/Users/shenjiangwei/Work/code/view/matrixone/cgo -I/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/include' CGO_LDFLAGS='-Wl,-w,-rpath,/Users/shenjiangwei/Work/code/view/matrixone/cgo -Wl,-rpath,/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/lib -L/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/lib -L/Users/shenjiangwei/Work/code/view/matrixone/cgo -lmo' go test ./pkg/sql/plan -count=1`

### DIFF
--- a/pkg/sql/plan/apply_indices.go
+++ b/pkg/sql/plan/apply_indices.go
@@ -719,6 +719,15 @@ func canUseRegularIndexHiddenSortKey(scanNode *plan.Node, orderByCol *plan.ColRe
 	return isRegularIndexFullPrefixEquality(scanNode.FilterList[0], numKeyParts)
 }
 
+func canPushRegularIndexOrderedLimit(scanNode *plan.Node) bool {
+	if scanNode == nil || len(scanNode.IndexScanInfo.Parts) < 2 || len(scanNode.FilterList) != 1 {
+		return false
+	}
+	// Static reader limit is valid only when index scan candidates exactly match the SQL filter.
+	numKeyParts := len(scanNode.IndexScanInfo.Parts) - 1
+	return isRegularIndexFullPrefixEquality(scanNode.FilterList[0], numKeyParts)
+}
+
 func isRegularIndexFullPrefixEquality(expr *plan.Expr, numKeyParts int) bool {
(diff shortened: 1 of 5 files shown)


ouyuanning (approved):

XuPeng-SH (approved):
Reviewed the latest head (5f1ec03). The fix correctly separates the safe hidden-key ordering rewrite from the unsafe static IndexReaderParam limit pushdown: the reader limit is now only enabled when the regular index scan filter is exactly the full-prefix equality predicate, so cursor/residual PK predicates cannot truncate candidates before residual filtering. The added unit tests cover both optimizer paths, and the BVT covers DESC/ASC cursor pagination plus limit larger than remaining rows. No blocking correctness, robustness, or performance issues found in this diff. Coverage is still pending at review time, so final merge should still wait for required checks.

heni02 (approved):

PR #25544, opened 2026-07-08
fix: guard regular index ordered limit pushdown

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23965

The regular secondary index ORDER BY primary-key LIMIT optimization can push a static scan-side limit when the secondary-index prefix is fixed.
With an extra primary-key cursor predicate, for example `id < ...`, the scan candidates no longer exactly match the SQL filter. Applying the static limit before the residual filter can return too few rows, or even zero rows.

This PR keeps the hidden-key ordering rewrite and TopValue message path, but only enables BlockTop/static limit pushdown when the index scan filter list is exactly the full-prefix equality filter.
Residual filters continue to be applied before the final LIMIT semantics are observed.

Added a BVT regression case:
`test/distributed/cases/optimizer/regular_index_order_limit_cursor.sql`

Tested with:

`env GOCACHE=/private/tmp/mo-go-build-cache-30 CGO_CFLAGS='-I/Users/shenjiangwei/Work/code/view/matrixone/cgo -I/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/include' CGO_LDFLAGS='-Wl,-w,-rpath,/Users/shenjiangwei/Work/code/view/matrixone/cgo -Wl,-rpath,/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/lib -L/Users/shenjiangwei/Work/code/view/matrixone/thirdparties/install/lib -L/Users/shenjiangwei/Work/code/view/matrixone/cgo -lmo' go test ./pkg/sql/plan -count=1`

### DIFF
--- a/pkg/sql/plan/apply_indices.go
+++ b/pkg/sql/plan/apply_indices.go
@@ -522,6 +522,15 @@ func canUseRegularIndexHiddenSortKey(scanNode *plan.Node, orderByCol *plan.ColRe
 	return isRegularIndexFullPrefixEquality(scanNode.FilterList[0], numKeyParts)
 }
 
+func canPushRegularIndexOrderedLimit(scanNode *plan.Node) bool {
+	if scanNode == nil || len(scanNode.IndexScanInfo.Parts) < 2 || len(scanNode.FilterList) != 1 {
+		return false
+	}
+	// Static limit pushdown is valid only when index scan candidates exactly match the SQL filter.
+	numKeyParts := len(scanNode.IndexScanInfo.Parts) - 1
+	return isRegularIndexFullPrefixEquality(scanNode.FilterList[0], numKeyParts)
+}
+
 func isRegularIndexFullPrefixEquality(expr *plan.Expr, numKeyParts int) bool {
 	if numKeyParts <= 0 || expr == nil {
 		return false
@@ -567,7 +576,9 @@ func (builder *QueryBuilder) applyRegularIndexTopSort(ctx *regularIndexTopSortCo
 		Expr: scanHiddenKeyExpr,
 		Flag: ctx.sortNode.OrderBy[0].Flag,
 	})
-	applyRegularIndexBlockTop(ctx.scanNode, ctx.scanNode.OrderBy[len(ctx.scanNode.OrderBy)-1], ctx.sortNode.Limit)
+	if ctx.sortNode.Offset == nil && ctx.sortNode.RankOption == nil && canPushRegularIndexOrderedLimit(ctx.scanNode) {
+		applyRegularIndexBlockTop(ctx.scanNode, ctx.scanNode.OrderBy[len(ctx.scanNode.OrderBy)-1], ctx.sortNode.Limit)
+	}
 
 	if !hasTopValueMessage(ctx.sortNode) {
 		msgHeader := plan.MsgHeader{
--- a/pkg/sql/plan/apply_indices_test.go
+++ b/pkg/sql/plan/apply_indices_test.go
@@ -670,6 +670,23 @@ func makeTestRegularIndexPrefixEq(t *testing.T, numArgs int) *planpb.Expr {
 	return prefixExpr
 }
 
+func makeTestRegularIndexPKLessThan(t *testing.T, value int64) *planpb.Expr {
+	t.Helper()
+	expr, err := BindFuncExprImplByPlanExpr(context.Background(), "<", []*planpb.Expr{
+		GetColExpr(planpb.Type{Id: int32(types.T_int64)}, 100, 1),
+		{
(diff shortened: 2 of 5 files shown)


XuPeng-SH (changes requested):
I agree with the core direction: separating the safe hidden-key ORDER BY rewrite from the stricter static ordered LIMIT/BlockTop pushdown is the right first-principles fix. `applyRegularIndexTopSort` and `handleMessageFromTopToScan` now share the same `canPushRegularIndexOrderedLimit` gate, which is important because the bug is specifically caused by applying a fixed scan-side limit before residual filters such as the PK cursor predicate.

Requesting changes because the new BVT regression is currently not executable in 3.0-dev CI, so the end-to-end coverage is not closed yet.

High: `test/distributed/cases/optimizer/regular_index_order_limit_cursor.result` does not match the command format parsed by mo-tester. The failed CI job reports:

`ResultParser:111 - [Exceptional command] ... regular_index_order_limit_cursor.sql][16]: create table t (...) ;, it does not exist in result file`

and the file summary is `TOTAL:9, SUCCESS:0, ABNORAML:9`. So the newly added regression file never actually runs; it fails during result parsing before validating the cursor-query result. The result file also appears to use a newer/other output format (`➤ ... 𝄀`) while nearby 3.0-dev optimizer BVT result files use the plain format (`mo_ctl(...)`, `Sleep(1)`, etc.).

Please regenerate this `.result` with the 3.0-dev mo-tester format, or rewrite the SQL/result so the normalized multi-line commands match exactly. In practice, making the DDL/insert/query command text match existing optimizer BVT style, then rerunning at least the affected BVT plus `go test ./pkg/sql/plan -count=1`, should close this.

Non-blocking but recommended while touching the test: add the symmetric cursor shape (`ORDER BY id ASC` with `id > ...`) or another residual-filter case. The planner unit tests cover the two planner entry points well; the BVT should be the executable correctness closure for the SQL behavior.

LeftHandCold:
Updated the BVT regression case after the standalone BVT failure.

Root cause of the CI failure: the generated `.result` needed to match the 3.0-dev mo-tester parser's normalized multiline SQL format. The SQL itself was valid.

Changes:
- regenerated `regular_index_order_limit_cursor.result` with mo-tester
- added two more cursor pagination regressions:
  - `ORDER BY id ASC LIMIT` with `id > cursor`
  - `ORDER BY id DESC LIMIT` with cursor near the beginning and limit larger than remaining rows

Verified on `shanghai50` with the PR branch code in `/home/mo/matrixone`:

`/home/mo/projects/mo-tester/run.sh -p /home/mo/matrixone/test/distributed/cases/optimizer/regular_index_order_limit_cursor.sql -m run -n -g -r 100`

Result: `TOTAL :11, SUCCESS :11, FAILED :0, ABNORAML :0, SUCCESS RATE : 100%`

LeftHandCold:
Updated 3.0-dev PR to fix the BVT result-format mismatch found in run 28954868617. The SQL output in CI was correct, but 3.0 CI uses the old .result format while the file was regenerated with the newer mo-tester delimiters from shanghai50. New head: 78d59132b5.

heni02 (approved):

LeftHandCold:
Updated again for the remaining 3.0 BVT failure in run 28991952768. The only failing assertion was the dynamic JSON returned by select mo_ctl('dn','flush',...). The case now still executes flush, but asserts a stable wrapper result (flushed = 1), so the regression focuses on the three cursor ORDER BY/LIMIT queries instead of DN-specific flush JSON. Verified the updated SQL on shanghai50 via mysql; new head: 7bbe0f1b18.

ouyuanning (approved):

XuPeng-SH (approved):
Reviewed latest head 7bbe0f1. The previous blocker I raised is addressed: the new BVT result is now in the 3.0-dev plain mo-tester format, the dynamic mo_ctl flush JSON is wrapped behind a stable flushed=1 assertion, and the SQL keeps the flush while focusing the regression on the three cursor ORDER BY/LIMIT result checks.\n\nCode-wise, the backport uses the correct 3.0-dev mechanism (BlockOrderBy/BlockLimit) and gates only the unsafe static BlockTop pushdown behind exact full-prefix equality with no residual filters. The safe hidden-key ordering rewrite and TopValue message path are preserved, so cursor predicates no longer allow early scan-side truncation before residual filtering while the pure prefix-equality optimization remains intact.\n\nLocal check: targeted pkg/sql/plan regular-index/message tests passed on the PR head with the project CGo flags. Latest CI is still pending, so final merge should wait for required checks.

PR #25552, opened 2026-07-09
fix: ivf search MultiCN execution

### LINKED ISSUE #25454: [Tech Request]: Preserve Multi-CN for ivf_search entries scans under small estimated stats

ULookup:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25454

This PR fixes the real `ivf_search` execution path so IVFFLAT entries scans can be stably dispatched across multiple CNs.

The previous fix targeted an internal `TABLE_SCAN` shape, but the production `ivf_search` path is a `FUNCTION_SCAN`. As a result, Multi-CN scheduling did not reliably reach the actual entries loading path, and each CN could still load/cache the full entries data independently.

This PR updates the end-to-end path so:

- the real `ivf_search FUNCTION_SCAN` is recognized for Multi-CN scheduling
- compile creates one `ivf_search` scope per CN when Multi-CN execution is allowed
- `IndexReaderParam` carries `PartitionCnCnt` / `PartitionCnIdx` through pipeline and remote execution
- `ivf_search` cache keys are partition-aware
- IVFFLAT entries SQL applies a deterministic PK-based partition filter per CN

With this change, Multi-CN execution loads disjoint IVFFLAT entries subsets per CN instead of duplicating the full entries cache on every CN. This reduces redundant cache usage and improves cache utilization for vector search workloads.

The PR also adds UT and BVT coverage for the Multi-CN partition path, including normal search, exact PK filtering, integer/varchar/composite primary keys, and `EXPLAIN` / `EXPLAIN ANALYZE` verification.


heni02 (approved):

XuPeng-SH (changes requested):
I think this still needs changes before merge. There are two blocking issues in the end-to-end design.

1. **Correctness / rolling upgrade: the new Multi-CN path is not safe with an older remote CN.**

`compileIvfSearchParallel` now sends `ivf_search` scopes to every selected CN and correctness depends on `IndexReaderParam.PartitionCnCnt/PartitionCnIdx`, the original limit, and `RuntimeFilterProbeList` surviving remote-run serialization. Those values are carried by newly added fields 7/8 in `pipeline.TableFunction`.

Protobuf wire decoding is syntactically backward compatible, but the execution semantics are not: an older CN silently ignores these unknown fields. Its restored `TableFunction` therefore has a nil `IndexReaderParam` and no runtime-filter specs. `ivfSearchPrepare` then falls back to limit 1, and that CN searches the full, unpartitioned entries set. In a mixed-version cluster, a new coordinator can consequently merge a partitioned local result with an unpartitioned remote top-1 result, producing duplicate PKs, fewer than K distinct rows, or incorrect pre-filter results.

The green upgrade jobs do not exercise a vector query whose new coordinator remote-runs this pipeline on an old CN. Please gate this execution mode on a negotiated protocol/capability that guarantees every selected CN understands these fields, or keep `ivf_search` on the old single-CN path until the cluster has completed the upgrade. Add a mixed-version codec/execution regression that proves the fallback cannot return wrong results.

2. **Performance: every CN now scans the full entries relation, so total scan/cache work grows approximately with CN count.**

The SQL adds `mod(crc32(cast(pk as varchar)), N) = i`, but `partitionedIvfScanNodes` then deliberately forces every inner entries scan to `CNCNT=1/CNIDX=0`. The hash expression is a row-level, non-sargable predicate for the entries PK layout; it does not assign persisted objects/blocks to one CN. Therefore N outer `ivf_search` scopes each read the same candidate-centroid entries blocks and discard roughly `(N-1)/N` of the rows after reading them. This avoids the earlier correctness bug caused by applying two incompatible partitions, but it defeats the stated goal of reducing duplicate entries loading/cache usage and can turn a one-table scan into N full scans.

Please use one ownership model end to end. For example, propagate a physical partition to the inner reader and handle appendable objects explicitly so they have exactly one owner; or provide a storage-prunable partition key. Do not combine a full scan on every CN with a post-read hash filter. At minimum, add a multi-CN integration/performance assertion showing aggregate input blocks/read bytes and file-service cache population do not scale linearly with CN count.

The current BVT is not sufficient evidence for either property. All four result checks also pass on the single-CN path, and commit `187f53593` removed the `EXPLAIN` / `EXPLAIN ANALYZE` assertions even though the PR description still claims that coverage. Please add a stable assertion that the real `FUNCTION_SCAN` executes on multiple CNs, verify global Top-K/distinctness for uneven and empty partitions plus runtime-filter modes, and cover the mixed-version fallback above.

I ran the affected package suites locally on `1dce73e3be3e1d6ddfce06601be469d3530e78b7`:

```text
go test -count=1 ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/colexec/table_function ./pkg/vectorindex/ivfflat
```

They pass, but the current tests exercise field propagation and generated SQL shape rather than these end-to-end failure modes.

fengttt (approved):

XuPeng-SH (changes requested):
Blocking issue:

- The new Multi-CN `ivf_search` path suppresses **all** internal background-query plans, so `EXPLAIN` / `EXPLAIN ANALYZE` on the partitioned path loses the entries-scan subplan entirely.

Why this happens:

- `compileIvfSearchParallel()` partitions every `ivf_search` scope with `PartitionCnCnt > 1`.
- `runIvfSearchVector()` only appends `rt.BackgroundQueries` when `shouldRecordIvfSearchBackgroundQueries()` returns true.
- That helper now returns false for both partitioned params and remote-run contexts, which is exactly the new Multi-CN path this PR introduces.
- `analyze_module.go` only surfaces background plans via `opStats.BackgroundQueries`, and `explain_query.go` only renders `qry.BackgroundQueries`.

So the partitioned execution path can run correctly, but the explain/diagnostic surface becomes incomplete right where the PR claims EXPLAIN / EXPLAIN ANALYZE verification.

Suggestion:

- Keep one representative background query plan (for example coordinator-side or `CNIDX == 0`) instead of dropping all of them, or aggregate them explicitly.
- Add end-to-end `EXPLAIN` / `EXPLAIN ANALYZE` coverage for the partitioned path. The new BVT file currently exercises result correctness only, and the new unit test locks in the suppression behavior instead of validating explain output.

XuPeng-SH (approved):
Approved on head `c9881386ea3` after re-reviewing the complete Multi-CN execution chain.

The previously blocking issues are addressed:

- The entries scan is now physically partitioned by complete ObjectID, so persisted object/block reads have exactly one owner instead of every CN scanning the full relation.
- Appendable/uncommitted visibility stays on the coordinator path, and writable workspaces fall back to the single-CN table-function path.
- Mixed-commit discovery is propagated through both the context-aware and legacy `Engine.Nodes` scheduling paths; an IVF query conservatively falls back to one CN before any new protobuf-dependent remote execution.
- `IndexReaderParam` and runtime-filter specs survive protobuf marshal/unmarshal and connector-child reconstruction.
- Each partition computes a local Top-K over disjoint data and the existing outer Sort/Limit performs the global Top-K; exact-PK, uneven, and empty-partition cases are covered.
- EXPLAIN ANALYZE now retains one coordinator-side background entries plan. The coordinator-address scope is localized by `RemoteRun`, while remote decoded scopes suppress the protobuf-oneof plan that cannot safely pass through the JSON stats path.

Unhappy-path audit: scope/child-process and result-batch ownership terminate through existing pipeline cleanup; runtime-filter waits are cancellable by the query context; per-query scope/result growth is bounded by eligible CN count and K, and partitioned cache entries retain the existing TTL/housekeeping destruction path. I found no remaining correctness, performance, leak, hang, or unbounded-growth blocker.

Local validation passed:

- `go test -count=1 -timeout 600s ./pkg/sql/compile`
- `go test -count=1 -timeout 600s ./pkg/sql/plan`
- `go test -count=1 -timeout 600s ./pkg/sql/colexec/table_function ./pkg/vectorindex/ivfflat ./pkg/vectorindex/cache`
- `go test -count=1 -timeout 600s ./pkg/sql/schedule ./pkg/vm/engine/disttae`

GitHub UT/SCA/BVT/benchmark checks are still running at review time, so merge should remain gated on their final results.

gouhongshen (approved):

aunjgr (approved):

Issue #25551, opened 2026-07-09
[Bug] `BIGINT UNSIGNED / BIGINT` division rejected — `system.statement_info.duration / 1000000000` fails


heni02:
## Summary

Dividing a **`BIGINT UNSIGNED`** column by a numeric literal fails at runtime:

```text
invalid argument operator /, bad value [BIGINT UNSIGNED BIGINT]
```

This breaks the slow-query probe used by **issue-10313-storage-query-test** (`checkSlowQueries`) and any monitoring SQL that converts `system.statement_info.duration` (nanoseconds) to seconds without an explicit cast.

**Impact:** Low — no data corruption, no soak failure. The probe silently fails (test swallows the error), but CN logs show recurring ERRORs and slow-query dashboards return no rows.

**First observed:** `mo-stb-3-0-dev-18addf5-202607082233` (~12h stability soak, `commit-18addf5`).

## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-stb-3-0-dev-18addf5-202607082233` |
| Endpoint | `10.222.6.254:6001` |
| Build / image | `commit-18addf5` |
| Artifact | `/data1/stability-test/28950457513_18addf5` |
| Trigger workload | `issue-10313-storage-query-test` — `checkSlowQueries()` ticker (~30s) |
| Session | `sys:dump` via proxy `10.10.17.225` |
| CN log caller | `frontend/util.go:463` (`query trace status`), `function/function.go:186` |
| Error count (12h soak) | `crksx` ×18, `6l2l2` ×3 (before CN pod recycle) |

## Table DDL (builtin, not executed at incident time)

`system.statement_info` is a MatrixOne builtin table. The relevant column:

```sql
CREATE TABLE `statement_info` (
  ...
  `duration` bigint unsigned DEFAULT '0' COMMENT 'exec time, unit: ns',
  ...
) COMMENT='record each statement and stats info[mo_no_del_hint]'
CLUSTER BY (`account`, `request_at`);
```

Confirmed on cluster: `duration` → `BIGINT UNSIGNED(64)`.

**No runtime DDL is involved.** The bug is triggered purely by a SELECT against this existing schema.

## Failing query

From CN `query trace status` logs and motr source:

```sql
SELECT statement, duration / 1000000000 AS duration_sec
FROM system.statement_info
WHERE duration > 1000000000
  AND request_at > DATE_SUB(NOW(), INTERVAL 1 MINUTE)
ORDER BY duration DESC
[...]

PR #25555, opened 2026-07-09
fix uint64 div

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25551

新增 int64 和 uint64 混合除法                                                                                                                                                                                                                                                                    
     将 int64 <-> uint64 的二元操作公共类型改成 decimal128，避免 uint64 大值转 int64 时溢出。                                                                                                                                                                                                    
     当前值正确，除法 scale 是 6，是matrixone一直确定的行为. 不与和 MySQL 默认 4 位小数兼容.


XuPeng-SH (approved):
Approved from code-review side.

The fix is scoped correctly: `fixedTypeCastRule2` is only used by `/` and `DIV`, so this does not change `+`, `-`, `*`, or `%`. Promoting the `BIGINT`/`BIGINT UNSIGNED` pair to `decimal128` avoids the unsafe uint64-to-int64 path and the new function tests plus BVT cover the issue query shape, both operand orders, and `uint64` max division for `/`.

Local verification:

```text
git diff --check origin/main...pr/25555: clean
go test -v -count=1 -timeout 120s -run 'Test_fixedTypeCastRule2|Test_GetFunctionByName' ./pkg/sql/plan/function: pass
go test -v -count=1 -timeout 120s -run 'TestIntegerDivOperatorSupports|TestUint64DivConsistency|Test_IntegerDivFn' ./pkg/sql/plan/function: pass
go test -v -count=1 -timeout 120s ./pkg/sql/plan/function: pass
```

Non-blocking: because the same cast rule is also used by SQL `DIV`, it would be useful to add a small regression case for `BIGINT UNSIGNED DIV BIGINT` as follow-up coverage, especially around max uint64 and negative signed operands.

ouyuanning (approved):

heni02 (approved):

Issue #25554, opened 2026-07-09
LOAD DATA with flate compression hangs indefinitely — query stuck at planning/dispatch phase with no execution


Ariznawlll:
## Summary

`LOAD DATA` with `compression='flate'` on a large file (40M rows x 100 columns) hangs indefinitely for 2h44m until workflow timeout. The query never reaches the CN execution layer — it gets stuck at the planning/dispatch phase.

## Environment

- **Branch**: main (commit b9efcce38)
- **Cluster**: TKE, 3 CNs + 1 DN
- **CN memory limit**: 55 GiB
- **Job**: [LOAD DATA TEST](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28954031117/job/86037586626)
- **Namespace**: `mo-main-commit-b9efcce38-20260709`

## Reproduction

```sql
load data url s3option {
  'endpoint'='http://cos.ap-guangzhou.myqcloud.com',
  'bucket'='mo-load-guangzhou-1308875761',
  'filepath'='compressed_file/40000000_100_columns_load_data.flate',
  'compression'='flate'
} into table test.table_100_columns
fields terminated by ',' lines terminated by '\n' parallel 'true';
```

## Timeline

```
04:22:56  LOAD DATA started
04:22:56 - 07:07:24  COMPLETELY SILENT (2h 44m)
  - No CN logs about load/flate/compress
  - No memory growth on any CN (stable 27-30 GiB)
  - No goroutine growth (stable ~3100 per CN)
  - No GC pauses (p75 < 1ms)
  - DN shows 0 data written to table_100_columns
07:07:24  Job cancelled by workflow timeout
```

## Comparison with other loads on the same table

| File | Rows | Compression | Duration | Status |
|---|---|---|---|---|
| 10000_100_columns.csv | 10K | none | 1.4s | OK |
| 100000_100_columns.csv | 100K | none | 7.4s | OK |
| 1000000_100_columns.csv | 1M | none | 35s | OK |
| 100000000_100_columns.csv | 100M | none | 4.9 min | OK |
| 40000000_100_columns.csv | 40M | none | 2.0 min | OK |
| 1000000_100_columns.csv.bz2 | 1M | bz2 | 3.3 min | OK |
| **40000000_100_columns.flate** | **40M** | **flate** | **2h44m (hung)** | **HANG** |

Only flate compression causes the hang. Same data uncompressed or bz2-compressed loads fine.

### 1. Memory — No growth during hang

```
[...]

PR #25556, opened 2026-07-09
fix(compile): LOAD DATA with compressed file hangs due to empty dispatch source address

ck89119:
## What type of PR is this?

- [x] BUG

issue #25554

PR #25537 replaced `isSameCN` with `sameExecutionNode` in `constructDispatchLocalAndRemote`. The new function correctly treats empty-vs-non-empty addresses as different execution nodes, but `compileExternScanParallelWrite` still hardcoded an empty address for the source scope, causing:

1. dispatch Prepare sees RemoteRegs + FuncId=`SendToAnyLocalFunc` → raises `"SendToAnyLocalFunc should not send to remote"`
2. Runtime tries to connect to empty address via morpc → `"dial tcp4: missing address"` infinite retry
3. Query hangs forever until workflow timeout

**Fix**: Pass the current CN address (`c.addr`) to `constructScopeForExternal` in `compileExternScanParallelWrite`, so `sameExecutionNode` correctly identifies source and target as the same local node.

**Affected**: All LOAD DATA with compressed files (flate/bz2/gzip/lz4) where `mcpu > 1`. Not specific to any compression type.

## Verification

- `go vet ./pkg/sql/compile/` ✅
- `go test ./pkg/sql/compile/` ✅
- Added `TestSameExecutionNodeMatchesByAddressForLocalScope` regression test ✅

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


XuPeng-SH (changes requested):
Requesting changes. The one-line fix is in the right direction, but the PR does not yet have an effective regression guard for this production hang.

1. High: the new regression test does not exercise the changed behavior and would pass before this fix.

`pkg/sql/compile/max_dop_test.go:248` only checks `sameExecutionNode` with equal/non-equal addresses. That helper already behaves that way on `main`, so this test would not catch the actual regression from `pkg/sql/compile/compile.go:1892` where `compileExternScanParallelWrite` constructed the source scope with an empty address.

The contract that failed is in `constructDispatchLocalAndRemote`: for compressed LOAD DATA with `mcpu > 1`, the source external scope and local merge scopes must be classified as the same execution node, producing `SendToAnyLocalFunc` with `RemoteRegs == 0` and no empty `NodeAddr`. Please make the regression test assert that actual compile shape, for example by extending `TestCompileExternScanParallelWrite` to inspect `rs[0].PreScopes[0].RootOp.(*dispatch.Dispatch)` and require:

- source scope `NodeInfo.Addr == testCompile.addr`
- `dispatchOp.FuncId == dispatch.SendToAnyLocalFunc`
- `len(dispatchOp.RemoteRegs) == 0`
- `len(dispatchOp.LocalRegs) == expected mcpu`
- no generated remote receive info with empty addr

That test should fail on the pre-fix line `constructScopeForExternal("", false)` and pass with `constructScopeForExternal(c.addr, false)`. This is important because the unhappy path is a hang / infinite retry to an empty morpc address, so a helper-only test is too weak.

2. Low: the PR currently fails whitespace hygiene.

`git diff --check origin/main...pr/25556` reports:

```text
pkg/sql/compile/max_dop_test.go:241: trailing whitespace.
```

Please run gofmt or remove the stray tab/blank whitespace around the new test.

XuPeng-SH (approved):
Approved from code-review side.

The previous blocker is addressed: the regression test now exercises the actual `compileExternScanParallelWrite` shape, asserts the source scope uses the current CN address, and verifies `SendToAnyLocalFunc` has zero `RemoteRegs` with all merge scopes covered by `LocalRegs`. `git diff --check origin/main...pr/25556` is also clean.

I was not able to run the target `pkg/sql/compile` test locally because this checkout is missing CGo headers under `thirdparties/install/include/cppjieba` (`github.com/yanyiwu/gojieba: cppjieba/Jieba.hpp not found`). CI still needs to finish.

Non-blocking: the new test could be made even more environment-stable by setting `testCompile.ncpu = 4` explicitly, since explicit stats guarantee the requested parallelism but `getParallelSizeForExternalScan` still caps by `c.ncpu`.

PR #25562, opened 2026-07-09
fix(frontend): force recompile on each prepared execute to prevent hang

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25526

Fix prepared UPDATE...JOIN hanging on second execution through binary protocol
(COM_STMT_EXECUTE). First execution succeeds, second hangs forever.

**Root Cause**: The `initExecuteStmtParam` function caches and reuses a compiled
scope tree (`prepareStmt.compile`) across executions when no schema change is
detected. After the first execution, operators in the cached compile retain stale
state (e.g. hashbuild ctr, dispatch channels) from the `isPrepare=true` cleanup
path that skips `Free()`. On re-execution, hashbuild operators enter `Call` but
stall in `build()` → never reach `handleRuntimeFilter` → never send runtime
filter messages → scan receivers block forever in `waitForRuntimeFilters`
(confirmed via goroutine dump).

**Fix**: Restructure `initExecuteStmtParam` to always release the cached compile
(via `FreeOperator` + `SetIsPrepare(false)` + `Release`) and create a fresh one
with `isPrepare=true` on every execution. This follows the same pattern the
`change=true` path already used. The plan remains cached and reused; only the
compile (lightweight scope tree) is rebuilt. This avoids:

- Resource leaks (proper release before replacement)
- Execution path drift (`isPrepare=true` preserved for `GetExecType`)
- Side effects in read paths (consistent recreation for all callers)

**Verification**: Reproduced with Go driver (`interpolateParams=false`) on
single-node mo-service. Goroutine dump confirmed scan receivers blocked in
`MessageReceiver.ReceiveMessage` with hashbuild never sending runtime filters.
After fix: both executions complete normally.


XuPeng-SH (approved):
Reviewed the prepared-execute lifecycle, operator/pipeline cleanup path, binary-protocol coverage, and unhappy paths around failed recompile and non-cached prepare.

No blocking issues found. The direction is correct: prepared compiles intentionally survive normal Run()/Release(), so stale operator state can cross executions; explicitly freeing the old operator tree and rebuilding the cached compile before each prepared execute closes that lifecycle gap while keeping the prepared plan cache behavior.

Coverage looks appropriate: frontend unit tests cover cached recompile, no-cache skip, and recompile failure cleanup; the issue test covers the real binary protocol path with repeated COM_STMT_EXECUTE; BVT adds SQL PREPARE coverage. CI is green across UT, Coverage, BVT, multi-CN, upgrade, and benchmark checks.

Local targeted tests passed:
- go test ./pkg/frontend -run TestInitExecuteStmtParam...
- go test ./pkg/tests/issues -run TestIssue25526PreparedUpdateJoinSecondExecute

aptend:
Reviewed the prepared-compile lifecycle and the full failure closure. No blocking issues found.

The ownership path is consistent: the stale cached compile is fully destroyed with `FreeOperator` + `SetIsPrepare(false)` + `Release`, the rebuilt TP compile remains owned by `PrepareStmt` until the next execution or statement close, AP statements with a nil cache continue through the regular fresh-compile path, and real recompile errors do not leave a stale/partial compile cached. The change introduces no new unbounded wait or resource-growth path.

The regression coverage exercises both the real binary-protocol `COM_STMT_EXECUTE` path and SQL PREPARE/EXECUTE. Local verification passed:

```text
go test -count=1 -run ^TestInitExecuteStmtParam ./pkg/frontend
go test -count=1 -run ^TestIssue25526PreparedUpdateJoinSecondExecute$ ./pkg/tests/issues
go test -count=1 ./pkg/frontend
go build ./pkg/frontend
go vet ./pkg/frontend ./pkg/tests/issues
```

The remaining tradeoff is the intentional loss of prepared scope-tree reuse for cached TP statements; that is acceptable here because it removes stale operator state while retaining the cached logical plan.

aptend (approved):

heni02 (approved):

Issue #22115, opened 2025-07-07
[Bug]: data may be wrong if restart cn or dn


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

2.1-dev:7b9e6df 2.2-dev:44a3a2118

### Commit ID

2.1-dev:7b9e6df 2.2-dev:44a3a2118

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior
tke环境：kubectl get pod -n mo-upgrade-nightly-7b9e6df-20250707

job: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/16110822404/job/45455129754

<img width="1404" alt="Image" src="https://github.com/user-attachments/assets/88524306-19b9-4e47-bf25-c1ab4b326847" />

<img width="1045" alt="Image" src="https://github.com/user-attachments/assets/cc21fe40-d6d8-4422-8d05-aa0c0f6039b0" />

<img width="1012" alt="Image" src="https://github.com/user-attachments/assets/cfa1a0cc-7c42-422b-be35-ed9da7cbde62" />

log： https://grafana.ci.matrixorigin.cn/explore?panes=%7B%224pV%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-upgrade-nightly-7b9e6df-20250707%5C%22%7D%20%7C%3D%20%60w_id%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221751875331000%22,%22to%22:%221751878931000%22%7D%7D%7D&schemaVersion=1&orgId=1

### Steps to Reproduce

```Markdown
trigger upgrade workflow
```

XuPeng-SH:
This is a bug related to transaction processing during network partitioning. When the CN submits a transaction but receives no response, the transaction's status becomes unknown, and it cannot obtain a commit timestamp. After being unlocked, the waiting transaction fails to retrieve the latest value.

XuPeng-SH:
Downgrade to s0 due to workaround.

iamlinjunhong:
修复 https://github.com/matrixorigin/matrixone/pull/25722

[...]

PR #25563, opened 2026-07-09
fix(lockservice): fence orphan txn unlocks

### LINKED ISSUE #25553: [Bug] Chaos TPCC (4056142): durable consistency corruption — Payment ytd split + NewOrder duplicate o_id (1062)
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-chaos-4-1-dev-4056142-202607061227` |
| Namespace | `mo-chaos-4-1-dev-4056142-202607061227` |
| Build / image | `matrixone:commit-4056142` (4.1-dev) |
| Endpoint | `10.222.6.253:6001` |
| Artifact root | `/data1/chaos-test/28765371805_4056142` |
| Workload | TPCC 10 warehouses / 10 terminals + chaos (kill CN/DN/Log, mem/network stress) |
| Analysis window | **2026-07-06 16:00 CST** → **2026-07-09 ~15:00 CST** (~71h) |
| Soak control (same commit) | `mo-stb-4-1-dev-d497091-202607031843` — **278 pass / 0 fail**, **0×1062** |

**Scope:** **Permanent** TPCC `ConsistencyCheck` failures under chaos — warehouse/district `ytd` divergence and `D_NEXT_O_ID` vs `max(o_id)` desync, driven by **Payment cross-partition partial commit** during kill chaos and **NewOrder duplicate primary key (1062)** on `bmsql_oorder`.

**Out of scope:** Transient read-skew-only false positives without sustained ytd/1062 (Phase A, Jul 6–7 — see [4164cfb](matrixone-chaos-4164cfb-tpcc-consistency-read-skew.md) / [#25232](https://github.com/matrixorigin/matrixone/issues/25232)); chaos orchestrator `kubectl delete chaos` CRD typo; leak-txn session cleanup (306 unique txn-ids in full run, separate class).

## Summary

During chaos long-run on `commit-4056142`, TPCC `runVerify.sh` / `io.mo.ConsistencyCheck` reported **12682 failures** vs **3747 passes** (**77% fail rate** over full run). After **2026-07-08 03:08:02 CST**, checks **never recovered** (**12680 consecutive FAIL**; **Jul 9: 0 pass / 5591 fail**).

Dominant failing invariants:

```sql
-- warehouse vs district payment totals (permanent drift, w_id=1 example)
(SELECT d_w_id, sum(d_ytd) FROM bmsql_district GROUP BY d_w_id)
EXCEPT (SELECT w_id, w_ytd FROM bmsql_warehouse);

-- order ID counter vs max order (from ~07:59 CST)
(SELECT d_w_id, d_id, D_NEXT_O_ID - 1 FROM bmsql_district)
EXCEPT (SELECT o_w_id, o_d_id, max(o_id) FROM bmsql_oorder GROUP BY o_w_id, o_d_id);
```

**This IS permanent data corruption**, not transient LogTail skew:

| Check | Result |
|-------|--------|
| Fail → PASS after 03:08:02 | **0** (12680 consecutive FAIL) |
| `ErrorCode 1062` / `Duplicate entry` | **16588** (Jul 7: 1; Jul 8: 8354; Jul 9: 8233) |
| First sustained 1062 on `bmsql_oorder` | **2026-07-08 06:44:40** — `(9,6,21246)` |
| `ErrorCode 20638` (full run) | **60** (sparse; corruption onset did **not** require 20638 storm) |
| Same commit soak (`mo-stb-4-1-dev-d497091`) | **278 pass / 0 fail**, **0×1062** |
| commit rate during incident | **healthy** (~128/s; `wait-active=0`) |

**MatrixOne issue (same defect family as [7cbeacd](matrixone-chaos-7cbeacd-tpcc-consistency-durable-corruption.md)):**

1. **Payment (R1):** Under `task_kill_dn` + `task_kill_cn`, Payment txn cross-partition updates to `bmsql_warehouse.w_ytd` and `bmsql_district.d_ytd` become **permanently inconsistent** (`sum(d_ytd) ≠ w_ytd`).
2. **NewOrder (R2):** Cross-CN concurrent read of `bmsql_district.d_next_o_id` allocates the **same `o_id`** → **1062** on `(o_w_id,o_d_id,o_id)`; district counter advances ahead of committed orders → `D_NEXT_O_ID - 1 ≠ max(o_id)`.

**Secondary (Phase A, two events):** **2026-07-06 22:23:26** and **2026-07-07 12:13:50** single `o_ol_cnt` vs `order_line` EXCEPT fail, **PASS in 31–35s** — transient read skew (#25232 class); **not** the sustained failure mode.

## TPCC invariants violated

| Invariant | TPCC txn | Tables |
|-----------|----------|--------|
| `sum(d_ytd) per w_id = w_ytd` | Payment | `bmsql_district`, `bmsql_warehouse` |
| `D_NEXT_O_ID - 1 = max(o_id)` per district | NewOrder | `bmsql_district`, `bmsql_oorder`, `bmsql_new_order` |
| `sum(o_ol_cnt) = count(order_line)` per district | NewOrder | `bmsql_oorder`, `bmsql_order_line` |

Payment (single JDBC txn):

```text
UPDATE bmsql_warehouse SET w_ytd = w_ytd + :amount WHERE w_id = :w_id;
UPDATE bmsql_district   SET d_ytd = d_ytd + :amount WHERE d_w_id = :w AND d_id = :d;
UPDATE bmsql_customer   SET c_balance ...;
COMMIT;
```

NewOrder reads/increments `d_next_o_id`, inserts `bmsql_oorder` + `bmsql_order_line` in one txn.

## Timeline

| Time (CST) | Event | Evidence |
|------------|-------|----------|
| **07-06 16:00** | Analysis window start; consistency healthy | check.log |
| **07-06 22:23:26** | **Phase A:** ol_cnt EXCEPT fail ×1 | Self-heal **PASS @22:23:57** (31s) |
| **07-07 12:13:50** | **Phase A:** ol_cnt EXCEPT fail ×1 | Self-heal **PASS @12:14:25** (35s) |
| **07-07 16:53:01** | Isolated 1062 on `bmsql_new_order` `(1,2,13235)` | 1×; verify still PASS |
| **07-08 02:51** | `task_network_dn` | mo_chaos_test.log |
| **07-08 03:02:39** | **`task_kill_dn`** | mo_chaos_test.log |
| **07-08 03:08:02** | Last consistency **PASS** | check.log |
| **07-08 03:08:10** | **`task_kill_cn`** starts | mo_chaos_test.log |
| **07-08 03:08:49** | **Phase B begins:** ytd EXCEPT fail | `w_id=1`, `sum(d_ytd)=320385339.28`; never PASS again |
| **07-08 06:44:40** | First sustained 1062 on `bmsql_oorder` `(9,6,21246)` | benchmarksql-error-10-10.log |
| **07-08 07:59:55** | `D_NEXT_O_ID` EXCEPT fail joins | district counter desync |
| **07-08 08:00+** | 1062 storm on `(9,3,21369)` etc. | repeated dup o_id |
| **07-09** | **0 pass / 5591 fail** | no self-heal |

Nearby client errors during kill window (03:03–03:08): **20502** backend connection closed, **20702** lock table bind changed — Payment/NewOrder in flight during DN/CN kill.

### Consistency fail stats (full run)

| Metric | Value |
|--------|-------|
| pass / fail | 3747 / 12682 |
| fail→pass after 03:08:02 | **0** |
| max consecutive fail (post-incident) | **12680** |
| Failed SQL types | ytd ×6811; D_NEXT_O_ID ×5864; ol_cnt ×7 |

### ytd drift example (w_id=1, first FAIL)

| Time | Exception side | Value |
|------|----------------|-------|
| 03:08:49 | `sum(d_ytd)` | 320,385,339.28 |
| 03:09:22+ | both directions | persistent EXCEPT rows |

Drift **never clears** — not 30s self-healing snapshot skew.

### 1062 breakdown

| Key | Count / note |
|-----|--------------|
| `(o_w_id,o_d_id,o_id)` on `bmsql_oorder` | **16588** total (Jul 8–9 storm) |
| First isolated (Jul 7) | `(1,2,13235)` on `bmsql_new_order` — 1× |
| First sustained oorder dup | **06:44:40** `(9,6,21246)` |
| Recurring key (Jul 8 08:00+) | `(9,3,21369)` hammered repeatedly |

### 20638 (secondary in this run)

| Day | Count |
|-----|-------|
| Jul 6 | 23 |
| Jul 7 | 13 |
| Jul 8 | 22 |
| Jul 9 | 2 |
| **Total** | **60** |

Unlike [7cbeacd](matrixone-chaos-7cbeacd-tpcc-consistency-durable-corruption.md), corruption onset aligned with **kill_dn + kill_cn** and connection/bind errors, not a large 20638 burst at the same second as first 1062.

### Control: same commit soak

| Env | consistency | 1062 |
|-----|-------------|------|
| `mo-stb-4-1-dev-d497091-202607031843` | pass=278, fail=0 | 0 |
| `mo-chaos-4-1-dev-4056142-202607061227` | pass=3747, fail=12682 | 16588 |

Same TPCC config: `warehouses=10`, `terminals=10`.

## What this is NOT

| Ruled out | Evidence |
|-----------|----------|
| BenchmarkSQL / TPCC harness logic bug | stb same commit **0 fail**; standard NewOrder/Payment |
| Transient LogTail read skew only (#25232 model) | **12680** consecutive FAIL; **16588×1062**; ytd drift permanent |
| Phase A read-skew as root of Phase B | Jul 6–7 ol_cnt fails self-healed; **31h healthy** before kill_dn/cn |
| Txn scheduler stall | `wait-active=0`; commit rate healthy through incident |
| Client mis-config | Identical to passing stb env |

### Symptom

TPCC `ConsistencyCheck` permanently fails on warehouse/district ytd and (later) district order-id counter vs `bmsql_oorder`.

### Mechanism

**R1 — Payment partial / non-atomic commit (primary, ytd):**

1. **`task_kill_dn` (03:02:39)** then **`task_kill_cn` (03:08:10)** disrupt CN↔TN paths during in-flight **Payment** txns.
2. Client sees **20502** / **20702** / **1105**; MO may **partially apply** cross-partition Payment updates: `bmsql_warehouse` (partition by `w_id`) and `bmsql_district` (partition by `d_w_id`) **not atomically visible**.
3. `sum(d_ytd) ≠ w_ytd` becomes **permanently true** at **03:08:49**; EXCEPT never clears.

**R2 — NewOrder duplicate o_id (secondary, order-id + accelerator):**

1. Multiple CNs concurrently read the same `bmsql_district.d_next_o_id` without serialisation.
2. Two NewOrders insert the same `(o_w_id,o_d_id,o_id)` → **1062** (storm from **06:44**).
3. Failed txn may still have incremented `d_next_o_id` → `D_NEXT_O_ID - 1 > max(o_id)` from **07:59:55**.

**Phase A (Jul 6–7, separate):** two ol_cnt EXCEPT fails, PASS in ~31–35s, no sustained ytd/1062 — transient cross-partition read skew under chaos; harness has no retry.

### Earliest fixable defects

| # | Defect | Layer |
|---|--------|-------|
| **R1** | Distributed Payment commit must be **all-or-nothing** across warehouse + district partitions; kill/timeout must not leave split ytd | MO txn / 2PC |
| **R2** | `d_next_o_id` allocation must be **serialised per district**; failed insert must not leave counter ahead of data | MO txn / lock on district row |
| **R3** | Single-statement multi-table read snapshot (Phase A only) | MO disttae (#25232) |

## Expected vs actual

| Expected | Actual |
|----------|--------|
| Payment COMMIT atomic across warehouse + district | **Permanent ytd split** after kill_dn/cn |
| NewOrder unique `(o_w_id,o_d_id,o_id)` | **16588×1062** under chaos |
| Consistency FAIL self-heals within 30s | **12680** consecutive FAIL after 03:08:02 |
| Same commit soak stable | stb **0 fail**; chaos **77% fail** |

### P0 — MatrixOne: Payment cross-partition atomicity (R1)

1. **`pkg/txn/client/`** (and CN→TN commit path): on kill/timeout/disconnect, return definitive **rollback** or provide **txn status recovery** — never leave partial warehouse/district ytd visible.
2. **Cross-partition atomicity:** `bmsql_warehouse.w_ytd` and `bmsql_district.d_ytd` updates in one Payment txn must share one **CommitTimestamp** and become visible **together**.
3. **Regression:** inject `task_kill_dn` + `task_kill_cn` during Payment; assert `sum(d_ytd)=w_ytd` for all warehouses after recovery.

### P0 — MatrixOne: NewOrder order-id allocation (R2)

1. **Serialise** `bmsql_district` row read/update for `d_next_o_id` (e.g. `SELECT … FOR UPDATE`, or route district row to single CN leader).
2. On **1062** or txn abort, **do not** leave `d_next_o_id` incremented without a matching committed `bmsql_oorder` row.
3. **Files:** district update in SQL executor / lockservice; consider `pkg/lockservice` row lock on district PK.
4. **Regression:** 2+ CN, concurrent NewOrder + periodic `task_kill_cn`; **1062=0** on `bmsql_oorder`.

### P1 — MatrixOne: read snapshot (R3, Phase A)

Same as [#25232](https://github.com/matrixorigin/matrixone/issues/25232) / [4164cfb](matrixone-chaos-4164cfb-tpcc-consistency-read-skew.md): `pkg/vm/engine/disttae` — one consistent snapshot per statement.

### P1 — Chaos / TPCC harness (parallel)

1. **`ConsistencyCheck.java`:** retry ol_cnt EXCEPT **3× / 5s** before exit (Phase A only).
2. **Gate:** treat **1062** and sustained ytd/D_NEXT FAIL as **MO CRITICAL**; fail soak on **3 consecutive** consistency failures or ytd/D_NEXT type.

## Reproduce

1. Deploy `mo-chaos-4-1-dev-4056142-*` class cluster (`commit-4056142`, 4.1-dev).
2. Run chaos TPCC (`props_10.mo`: 10 wh, 10 terminals) + `runVerify.sh` every **30s** + full chaos loop (kill CN/DN/Log, mem/network).
3. Run **>48h** (failure observed ~39h in at 03:08 on Jul 8).
4. Observe:
   - ytd EXCEPT **permanent** FAIL after **`task_kill_dn` + `task_kill_cn`**
   - **16000+** `Duplicate entry` on `bmsql_oorder` PK (Jul 8–9)
   - **0** consistency PASS after first ytd FAIL
5. Control: same commit on `mo-stb-4-1-dev-d497091-*` without chaos → **0 fail**.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

fixes #25553

refs #22115
refs #22791

This PR prevents lockservice orphan cleanup from releasing a holder while its TN Commit can still finish, and ensures every exceptional orphan unlock carries a timestamp that fences stale waiters.

### Root cause

The incident follows the single-TN-shard 1PC branch and is a race between the TN Commit handler and lockservice orphan cleanup:

1. TN receives Commit and `LockTableAllocator.Valid` records the transaction as committing.
2. Commit remains in flight while CN/TN communication is abnormal.
3. `cleanCommitState` fails to obtain the CN active transaction list and deletes the service's commit-control state.
4. A waiter classifies the same holder as orphan. `CannotCommit` no longer sees the in-flight Commit and incorrectly reports that the holder is safe to unlock.
5. The holder lock is released, but the original TN Commit subsequently succeeds. A waiter can execute from state that did not include that Commit, breaking hotspot-row serialization.

The previous fence-only implementation also used the raw HLC upper bound. A timestamp such as `(110, 0)` does not dominate `(110, 7)`, so it was not a strict fence for all logical timestamps at the same physical time.

### Fix

1. Track the full in-flight Commit lifecycle in the allocator.
   - Every successful `Valid` increments a per-transaction reference count.
   - TN Commit defers exactly one `FinishCommit` for every successful `Valid`.
   - Duplicate or concurrent Commit requests remain protected until the final reference finishes.
   - Normal completed state is removed immediately and does not wait for the periodic cleaner.
2. Make commit-state cleanup conservative.
   - Cleaner never removes an in-flight Commit.
   - Before each `GetActiveTxn` call, cleaner captures that service's state-generation watermark and only applies the response to states at or below that watermark.
   - Creating or refreshing `cannotCommit`, and a late `Valid` rejected by that tombstone, advances the generation so an older active-txn snapshot cannot delete it.
   - A failed active-transaction query does not erase `cannotCommit` tombstones.
   - Permanently disconnected services still use the existing 24-hour retention boundary, so state cannot grow forever.
3. Return a strict allocator fence from `CannotCommit`.
   - The fence is `(upper.PhysicalTime + 1, 0)`, which dominates every logical timestamp at the current upper-bound physical time.
   - Clock absence or physical overflow returns no fence and fails closed.
4. Require both proofs before orphan unlock: the transaction is not committing and the allocator returned a non-empty fence.
   - The CN local-clock fallback is removed.
   - Waiter, timeout, stale-bind, local, and remote orphan paths all skip unlock when either proof is missing.

### Scope and risk

The normal Commit path adds only allocator-local state operations; it adds no RPC and removes state as soon as the final Commit attempt returns. Normal lock acquisition, isolation semantics, lock wait timeouts, and ordinary commit/unlock are unchanged. The fix does not introduce or depend on 2PC: `committing`/`cannotCommit` are lockservice allocator arbitration states around the current 1PC Commit handler.

The deployment invariant for this hotfix is one TN shard per transaction (`len(TNShards) == 1`). The retained multi-TN asynchronous Commit branch is outside this PR's supported scope; this PR does not claim to fix that branch's commit-reference lifetime.

If a TN Commit itself never returns, the holder remains protected. Unlocking it would be unsafe because that Commit may still become visible; TN recovery must resolve that condition. This PR does not turn an uncertain Commit into an early unlock.

### Main changes

| Area | Change |
|------|--------|
| `proto/lock.proto`, generated lock protobuf | Add `FenceTS` to `CannotCommitResponse` |
| `lock_table_allocator.go` | Reference-count in-flight Commit, generation-aware cleaner behavior, strict HLC fence |
| `service_cn_handler.go` | Pair successful `Valid` with deferred `FinishCommit` |
| orphan cleanup call sites | Remove local timestamp fallback and fail closed without allocator proof |
| tests | Cover duplicate Commit references, two-service stale cleaner snapshots, cleaner failures, strict HLC ordering, missing fence, and TN lifecycle pairing |

### Validation

- `go test ./pkg/lockservice -count=1`
- `go test -race ./pkg/lockservice -run 'Test(CannotCommitWaitsForAllCommitAttempts|CleanCommitStateKeepsInflightCommit|CleanerPreservesCannotCommitOnActiveTxnQueryError|CleanerDoesNotDeleteTombstoneCreatedAfterServiceSnapshot|CannotCommitGenerationRefresh|NewFenceTS)' -count=1`
- CGo-configured `go test ./pkg/txn/service -run TestCommitWithLockTables -count=1`
- `go vet ./pkg/lockservice`
- CGo-configured `go vet ./pkg/txn/service`


XuPeng-SH (approved):
Reviewed the lockservice orphan-cleanup path from first principles: CannotCommit ordering, remote/local timeout cleanup, stale bind heartbeat cleanup, waiter wake-up timestamp propagation, mixed empty-FenceTS fallback, and repeated CannotCommit races.

No blocking issues found. The change is directionally correct: orphan cleanup should not unlock with an empty commit timestamp after TN has only proven cannot-commit safety. Returning an allocator HLC upper-bound as a fence, and only unlocking when CannotCommit does not report the txn as committing, closes the unsafe stale-snapshot path while keeping normal commit/unlock behavior unchanged.

I also checked the unhappy paths: CannotCommit RPC errors keep the lock; committing txns are not unlocked; the second CannotCommit check in checkTxnTimeout avoids a race where the first check becomes stale; empty FenceTS falls back to a local clock value for compatibility instead of reintroducing empty unlocks.

This PR should be viewed as the lockservice orphan-unlock fix for #25553, not as standalone proof that the full TPCC chaos scenario is solved end-to-end. The code-level coverage is solid for this layer.

Local validation passed on the PR head:
- targeted fenced orphan/CannotCommit tests from the PR body
- broader orphan cleanup regressions including communication interruption, service restart, bind heartbeat missing, unstable network, and ValidTxn paths

CI is green across UT, Coverage, BVT, multi-CN, upgrade, and benchmark checks.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -965,6 +973,9 @@ func (l *lockTableAllocator) handleCannotCommit(
 	cs morpc.ClientSession) {
 	committingTxn := l.AddCannotCommit(req.CannotCommit.OrphanTxnList)
 	resp.CannotCommit.CommittingTxn = committingTxn
+	if l.clock != nil {
+		_, resp.CannotCommit.FenceTS = l.clock.Now()
```
XuPeng-SH: Using `Clock.Now()`'s upper-bound timestamp directly here still leaves a gap in HLC total ordering. `timestamp.Timestamp` compares `(PhysicalTime, LogicalTime)`, and `Now()` returns an upper bound with `LogicalTime == 0`. So a real commit that lands at the same `PhysicalTime` but with a higher `LogicalTime` is still greater than this fence.

Concrete example: fence `(110,0)` is `<` commitTS `(110,7)`. That means the orphan unlock can still publish a previous-commit timestamp that does not dominate the hidden commit we are fencing against.

I think the fence needs to be converted into a value that is guaranteed `>` every timestamp whose physical part is `<= upperBound.PhysicalTime` (for example, by moving to the next physical tick rather than reusing the raw upper bound). The same issue exists in the `ensureFenceTS()` fallback path too.

XuPeng-SH (changes requested):
I found one blocking correctness issue in the new fenced orphan-unlock path.

The new fence timestamp is derived from `Clock.Now()`'s upper-bound value, but `timestamp.Timestamp` ordering is `(PhysicalTime, LogicalTime)`. That upper-bound timestamp carries `LogicalTime == 0`, so it is not a total-order upper fence: a hidden commit at the same `PhysicalTime` with a larger `LogicalTime` still compares greater than the fence. In that case the orphan unlock can still publish a previous-commit timestamp that does not actually dominate the commit we are trying to fence, which reopens the stale-snapshot anomaly this PR is fixing.

This affects both the allocator-produced `FenceTS` and the local `ensureFenceTS()` fallback path. I'd fix it before merge by converting the bound into a timestamp that is guaranteed `>` every possible commit timestamp at or below that physical upper bound, then add a regression test for the equal-physical / higher-logical case.

XuPeng-SH (changes requested):
The previous HLC ordering issue is fixed on the latest commit: moving the allocator upper bound to the next physical tick does dominate timestamps with the same physical component and a higher logical component. However, I found another blocking 1PC correctness race in the new cleanup lifecycle.

`cleanCommitState` obtains active-transaction snapshots for all services first and only later acquires `ctlMu.Lock` to apply those snapshots. A `cannotCommit` tombstone created after one service's snapshot was obtained can therefore be deleted using that older snapshot.

One deterministic interleaving is:

1. The cleaner queries service A and receives an active-txn set that does not contain txn X.
2. The cleaner blocks while querying service B.
3. Orphan cleanup calls `AddCannotCommit` for A/X, installs the tombstone, receives a valid fence, and can proceed toward unlocking the holder.
4. Service B's query finishes. The cleaner acquires `ctlMu.Lock` and calls `cleanCtlLocked(A, oldSnapshot, false)`.
5. `commitCtl.clean` sees X as absent from the old snapshot and immediately deletes the newly created tombstone.
6. A delayed/retried 1PC Commit can now call `Valid(A, X, ...)` successfully even though the CN already received the fence and may have released X's lock.

This reopens the exact stale-waiter/hidden-commit safety violation this PR is intended to close. `ctlMu` prevents mutation while cleanup is being applied, but it does not make the earlier remote snapshots current and does not distinguish state created after a snapshot.

I reproduced this on `4e290c5f38a83ac7d0a202a63b19b25f0cf59ba3` with a temporary deterministic test using two services: return A's empty active snapshot, block B's query, create A's tombstone, then release B. The test fails because A's new tombstone is removed; the allocator logs `clean cannot commit txn` for it.

Please make cleanup generation-aware. For example, capture a per-service state generation/watermark before issuing `GetActiveTxn`, store a creation/update generation on each `commitState`, and after the RPC delete only records that both existed at or before that watermark and have not changed since. A state created or transitioned while the RPC round was in progress must survive that round. The regression should cover the two-service interleaving above, in addition to the existing RPC-error and in-flight-Commit tests.

Validation on the latest head:

```text
go test -count=1 ./pkg/lockservice
ok (134.494s)

go test -count=1 ./pkg/txn/service -run TestCommitWithLockTables
ok
```

The existing suites pass, but none places a new tombstone between an earlier service snapshot and the later cleanup application.

iamlinjunhong (approved):

XuPeng-SH (approved):
Re-reviewed the latest head `931748dd5` after the generation-aware cleanup changes.

The previous blocking stale-cleaner race is fixed correctly. The cleaner now captures a per-service generation watermark immediately before each `GetActiveTxn` request, and `commitState` creation/refresh/in-flight transitions advance the generation. Cleanup therefore only removes states that existed at or before the snapshot boundary and did not change while the RPC round was in flight. The new two-service interleaving test deterministically covers the original failure mode.

I also rechecked the surrounding state machine and unhappy paths:

- every successful lock-table `Valid` in the TN Commit handler is paired with deferred `FinishCommit`;
- duplicate/concurrent Commit attempts remain protected by the in-flight reference count;
- CannotCommit cannot replace a live committing state;
- failed active-txn queries preserve tombstones, while disconnected-service expiry remains fail-closed for newer generations;
- generation overflow intentionally preserves state;
- the strict fence still dominates equal-physical/higher-logical timestamps, and missing/overflowed clocks produce no fence and no orphan unlock;
- waiter, timeout, local, remote, and stale-bind cleanup paths require both allocator proofs before unlocking.

Performance is non-blocking: the normal Commit path adds no RPC or I/O, state is removed on completion, and the extra operations are small allocator-local map/mutex operations. `Valid` already serialized bind validation under `l.mu`; the new cost is a low constant factor plus a per-service `FinishCommit` critical section.

Compatibility remains fail-closed in the safety-critical direction: a new CN talking to an older allocator receives an empty fence and skips exceptional unlock. An old CN does not gain the fenced behavior until it is upgraded, which is an expected rollout limitation rather than a new regression.

Local validation on the latest head:

```text
go test ./pkg/lockservice -count=1
ok (124.063s)

go test -race ./pkg/lockservice -run generation/fence/in-flight tests -count=10
ok

go test ./pkg/txn/service -run ^TestCommitWithLockTables$ -count=10
ok

git diff --check and gofmt -l on touched Go files
clean
```

Approved. CI for the fresh head is still running and should remain the merge gate.

gouhongshen (approved):

aptend:
I found one remaining lifecycle gap in the latest head: the new commit reference is released too early on the 2PC path.

`Commit` registers the attempt through `Valid` and defers `FinishCommit`, but after prepare succeeds it only schedules `startAsyncCommitTask` and returns. The defer therefore runs as soon as the handler returns, while the background task may not yet have executed `storage.Committing`, sent `CommitTNShard` to the other shards, or completed the coordinator `storage.Commit`.

A concrete interleaving is:

1. `Valid` registers txn X as committing.
2. All prepares succeed and `startAsyncCommitTask` is scheduled, but the task has not completed.
3. The Commit handler returns and deferred `FinishCommit` removes X from allocator commit control.
4. Orphan cleanup now receives not-committing plus a fence and releases X's CN lock.
5. The async task subsequently makes X committed/visible.

This recreates the same unlock-before-commit-completion window that this PR is intended to close. The fence orders timestamps, but it does not make an asynchronously pending commit visible before the waiter proceeds.

Please transfer ownership of the successful `Valid` reference to the async commit task and call `FinishCommit` only when that task reaches its terminal point, while keeping exactly-once release on scheduling failure and prepare/rollback paths. A deterministic regression test can block the async commit task after `startAsyncCommitTask` returns and assert that `AddCannotCommit` continues to report X as committing until the task is released and finishes. The current `TestCommitWithLockTables` uses one TN shard and covers only the synchronous 1PC path.

Operational scope: this is not reachable in a deployment that strictly guarantees `len(Txn.TNShards) == 1` for every transaction. However, it remains a blocker for the general upstream 2PC-supported path unless that limitation is explicitly accepted and tracked as a known gap.

aptend (approved):

PR #25566, opened 2026-07-10
fix uint64 div

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25551

新增 int64 和 uint64 混合除法
将 int64 <-> uint64 的二元操作公共类型改成 decimal128，避免 uint64 大值转 int64 时溢出。
     当前值正确，除法 scale 是 6，是matrixone一直确定的行为. 不与和 MySQL 默认 4 位小数兼容.


XuPeng-SH (approved):
Approved from code-review side.\n\nThis is a clean 4.1-dev port of the uint64 division fix. The type coercion change is scoped to fixedTypeCastRule2, which is used by SQL / and DIV rather than the general arithmetic/comparison rules, so the blast radius is limited. Promoting the int64 <-> uint64 pair to decimal128 avoids the unsafe uint64-to-int64 path and preserves the existing MatrixOne division scale behavior.\n\nThe added coverage is appropriate for the reported issue: both operand orders are covered at the function/type-resolution level, and the BVT covers the real statement_info.duration / 1000000000 shape plus max uint64 division. CI is green for the latest run I checked.\n\nNon-blocking: a direct SQL regression for BIGINT UNSIGNED DIV BIGINT would still be useful because the same cast rule is shared by DIV. The PR is also currently behind 4.1-dev, so it may need an update before merge queue can proceed.

ouyuanning (approved):

heni02 (approved):

PR #25568, opened 2026-07-10
fix: support BLOB columns in data branch merge

### LINKED ISSUE #24927: [Bug] data branch merge fails on table with BLOB column
## Bug Report

`DATA BRANCH 100M` failed in the snapshot backup/restore nightly run because `data branch merge` cannot handle a table that contains a `BLOB` column.

Regression job:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27294308590/job/80660681490

The user originally shared the run/job around this workflow:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27294308590/job/80640171460

The linked workflow run also has a dedicated failed job named `DATA BRANCH 100M`.

## Version

- MatrixOne branch: `4.0-dev`
- MatrixOne commit: `b860e97918fce957f4a51f95d2115154eb808572`
- Image tag used by the run: `matrixone-dev/matrixone:commit-b860e9791`
- Regression branch checked out by the data branch job: `mo-nightly-regression:big_data`

## Failed Step

Job: `DATA BRANCH 100M`

Failed step:

```text
Data branch diff and merge on 100M table
```

The step ran:

```bash
cd $GITHUB_WORKSPACE/mo-nightly-regression/tools/mo-regression-test
python run.py -H 172.16.36.105 -u acc04:test_account -c ./cases/big_data_test/06_CLONE/nonsys_data_branch_100m.sql
```

## Reproduction SQL

The regression case is:

```sql
data branch create table big_data_test_db_clone.table_with_pk_index_for_load_100M_branch from big_data_test_db_clone.table_with_pk_index_for_load_100M;

update big_data_test_db_clone.table_with_pk_index_for_load_100M_branch
set col3 = col3 + 1
where id between 1 and 1000;

data branch diff big_data_test_db_clone.table_with_pk_index_for_load_100M_branch against big_data_test_db_clone.table_with_pk_index_for_load_100M output count;

data branch merge big_data_test_db_clone.table_with_pk_index_for_load_100M_branch into big_data_test_db_clone.table_with_pk_index_for_load_100M when conflict accept;

data branch diff big_data_test_db_clone.table_with_pk_index_for_load_100M_branch against big_data_test_db_clone.table_with_pk_index_for_load_100M output count;

data branch delete table big_data_test_db_clone.table_with_pk_index_for_load_100M_branch;
```

The source table includes a `BLOB` column:

```sql
create table if not exists big_data_test.table_with_pk_index_for_load_100M(
  id bigint primary key,
  col1 tinyint,
  col2 smallint,
  col3 int,
  col4 bigint,
  col5 tinyint unsigned,
  col6 smallint unsigned,
  col7 int unsigned,
  col8 bigint unsigned,
  col9 float,
  col10 double,
  col11 varchar(255),
  col12 Date,
  col13 DateTime,
  col14 timestamp,
  col15 bool,
  col16 decimal(16,6),
  col17 text,
  col18 json,
  col19 blob,
  col20 binary(255),
  col21 varbinary(255),
  col22 vecf32(3),
  col23 vecf32(3),
  col24 vecf64(3),
  col25 vecf64(3),
  key(col3),
  unique key(col4)
);
```

## Actual Result

Before merge, diff correctly reports 1000 changed rows:

```text
data branch diff ... output count
COUNT(*)
1000
```

The merge then fails:

```text
data branch merge big_data_test_db_clone.table_with_pk_index_for_load_100M_branch into big_data_test_db_clone.table_with_pk_index_for_load_100M when conflict accept
ERROR 20105: not supported: formatValIntoString: not support type BLOB
```

Because merge failed, the following diff still reports 1000 rows instead of 0:

```text
data branch diff ... output count
Actual Result:
  COUNT(*)
  1000
Expect Result:
  COUNT(*)
  0
```

## Expected Result

`data branch merge ... when conflict accept` should merge the branch back into the source table successfully, even when the table has a `BLOB` column.

Expected regression output:

```text
data branch merge ... when conflict accept
empty set

data branch diff ... output count
COUNT(*)
0
```

## Why This Looks Like A Product Bug

- The branch table is created successfully from the source table.
- Updating 1000 rows on the branch succeeds.
- `data branch diff ... output count` succeeds and sees the 1000 modified rows.
- The only failing operation is `data branch merge ... when conflict accept`.
- The failure is an internal unsupported-type error: `formatValIntoString: not support type BLOB`.
- The later `compare_failed` is only a consequence of the merge failure; since merge did not apply, diff remains 1000 instead of 0.

This suggests the merge implementation formats row values through `formatValIntoString`, but that path does not support `BLOB` values.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24927

`DATA BRANCH MERGE` failed when a table contained a `BLOB` column because the shared SQL literal formatter did not support `T_blob`.

This PR:

- adds `T_blob` to the shared data branch value formatter;
- serializes `BINARY`, `VARBINARY`, and `BLOB` values as hexadecimal SQL literals so arbitrary bytes are preserved exactly;
- keeps character and JSON values on their existing escaped-string paths;
- adds unit coverage for control bytes, empty values, and both byte-slice and string inputs;
- adds a data branch merge BVT covering binary padding, variable-length bytes, BLOB values, `NULL`, empty bytes, and post-merge diff convergence.

Using escaped SQL strings for binary values is not lossless: control bytes represented as `\\xNN` are parsed as text rather than converted back to the original byte. Hexadecimal literals avoid that corruption and work consistently for all binary-family types.

## Validation

- `go build ./pkg/frontend`
- `go vet ./pkg/frontend`
- full `go test -v -count=1 -timeout 300s ./pkg/frontend`
- `merge_9` BVT: 12/12 statements passed, 100% success rate
- independent SQL verification: diff count changed from 3 before merge to 0 after merge, with all binary values verified through `HEX()` and `LENGTH()`


gouhongshen:
## Review result

Reviewed commit `f1f9113777f33e1d096ef80d9e3cb777c29724a4`. No blocking issues found.

- `T_blob` now shares the byte-preserving hexadecimal SQL literal path used by `BINARY` and `VARBINARY`; character and JSON escaping are unchanged.
- `x'…'` is parser-supported and preserves NUL, quotes, backslashes, control bytes, and empty values exactly.
- The regression BVT covers BLOB, BINARY, VARBINARY, NULL, empty values, binary padding, and post-merge diff convergence.

Validated locally:

- `go vet ./pkg/frontend`
- `go test -count=1 -timeout 300s ./pkg/frontend`
- `go test -count=1 ./pkg/sql/parsers/dialect/mysql -run '^TestBinaryIntroducedHexLiteralHasDistinctType$'`

Required CI, including the regression BVT, is green.

I could not submit the approval because the active GitHub identity is the PR author, and GitHub prohibits self-approval. A non-author reviewer should approve this PR.

XuPeng-SH (approved):
Reviewed `f1f9113777f33e1d096ef80d9e3cb777c29724a4`. I found no blocking correctness, robustness, or performance regression in this change.

The fix uses a byte-preserving SQL representation for the complete binary family (`BINARY`, `VARBINARY`, and `BLOB`). The implementation handles both `[]byte` and `string`, preserves empty and arbitrary byte sequences, and avoids an intermediate encoded allocation by writing directly into the destination buffer. Character and JSON formatting remain on their existing paths.

The BVT closes the important execution loop: binary padding, variable-length bytes, BLOB data, `NULL`, empty bytes, online merge, value verification via `HEX()`/`LENGTH()`, and post-merge diff convergence. I also verified the formatter tests, the full `pkg/frontend` test suite, `go vet ./pkg/frontend`, and `git diff --check` locally; all pass. The required CI is green.

Non-blocking follow-up: the issue discussion identifies other user-visible types that the shared formatter still does not support (`BIT`, `UUID`, `ENUM`, geometry, and `DATALINK`). Please track those as a separate formatter capability-matrix issue rather than considering this helper generally exhaustive. It would also be useful to establish a documented/tested capacity boundary for large BLOBs: the maximum 64 MB value expands to roughly 128 MB as a hex SQL literal, and one row can exceed the current 32 MB batching threshold. That is an existing SQL-replay architecture limitation, not a regression caused by this targeted BLOB fix.

PR #25569, opened 2026-07-10
[4.1] fix: support BLOB columns in data branch merge

### LINKED ISSUE #24927: [Bug] data branch merge fails on table with BLOB column

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24927

Port of #25568 to `4.1-dev`.

`DATA BRANCH MERGE` failed when a table contained a `BLOB` column because the shared SQL literal formatter did not support `T_blob`.

This PR:

- adds `T_blob` to the shared data branch value formatter;
- serializes `BINARY`, `VARBINARY`, and `BLOB` values as hexadecimal SQL literals so arbitrary bytes are preserved exactly;
- keeps character and JSON values on their existing escaped-string paths;
- adds unit coverage for control bytes, empty values, and both byte-slice and string inputs;
- adds a data branch merge BVT covering fixed and variable-length binary values, BLOB values, `NULL`, empty bytes, and post-merge diff convergence.

Using escaped SQL strings for binary values is not lossless: control bytes represented as `\\xNN` are parsed as text rather than converted back to the original byte. Hexadecimal literals avoid that corruption and work consistently for all binary-family types.

The BVT result was generated and independently verified on `4.1-dev`; it intentionally reflects this branch's existing `BINARY` display behavior rather than copying the `main` result.

## Validation

- `make -j8 build`
- `go build ./pkg/frontend`
- `go vet ./pkg/frontend`
- full `go test -v -count=1 -timeout 300s ./pkg/frontend`
- `merge_9` BVT: 12/12 statements passed, 100% success rate
- independent SQL verification: diff count changed from 3 before merge to 0 after merge, with all binary values verified through `HEX()` and `LENGTH()`

### DIFF
--- a/pkg/frontend/data_branch_helpers.go
+++ b/pkg/frontend/data_branch_helpers.go
@@ -463,7 +463,7 @@ func formatValIntoString(ses *Session, val any, t types.Type, buf *bytes.Buffer)
 
 	switch t.Oid {
 	case types.T_varchar, types.T_text, types.T_json, types.T_char, types.
-		T_varbinary, types.T_binary:
+		T_varbinary, types.T_binary, types.T_blob:
 		if t.Oid == types.T_json {
 			var strVal string
 			switch x := val.(type) {
@@ -488,11 +488,15 @@ func formatValIntoString(ses *Session, val any, t types.Type, buf *bytes.Buffer)
 			writeEscapedSQLString(buf, jsonLiteral)
 			return nil
 		}
+		writeBytes := writeEscapedSQLString
+		if t.Oid == types.T_binary || t.Oid == types.T_varbinary || t.Oid == types.T_blob {
+			writeBytes = writeSQLHexLiteral
+		}
 		switch x := val.(type) {
 		case []byte:
-			writeEscapedSQLString(buf, x)
+			writeBytes(buf, x)
 		case string:
-			writeEscapedSQLString(buf, []byte(x))
+			writeBytes(buf, []byte(x))
 		default:
 			return moerr.NewInternalErrorNoCtxf("formatValIntoString: unexpected string type %T", val)
 		}
@@ -555,6 +559,19 @@ func formatValIntoString(ses *Session, val any, t types.Type, buf *bytes.Buffer)
 	}
 
 	return nil
+}
+
+func writeSQLHexLiteral(buf *bytes.Buffer, b []byte) {
+	const hexDigits = "0123456789abcdef"
+
+	// Hex literals preserve every byte; SQL string escapes are not reversible for all control bytes.
+	buf.Grow(2*len(b) + 3)
+	buf.WriteString("x'")
+	for _, c := range b {
+		buf.WriteByte(hexDigits[c>>4])
+		buf.WriteByte(hexDigits[c&0x0f])
+	}
+	buf.WriteByte('\'')
 }
 
 func extractDataBranchSQLRowValue(
--- a/pkg/frontend/data_branch_test.go
+++ b/pkg/frontend/data_branch_test.go
@@ -67,13 +67,23 @@ func TestFormatValIntoString_StringEscaping(t *testing.T) {
 	require.Equal(t, `'a\'b"c\\\n\t\r\Z\0'`, buf.String())
 }
 
-func TestFormatValIntoString_ByteEscaping(t *testing.T) {
-	var buf bytes.Buffer
-	ses := &Session{}
-
+func TestFormatValIntoString_BinaryHexLiteral(t *testing.T) {
 	val := []byte{'x', 0x00, '\\', 0x07, '\''}
-	require.NoError(t, formatValIntoString(ses, val, types.New(types.T_varbinary, 0, 0), &buf))
-	require.Equal(t, `'x\0\\\x07\''`, buf.String())
+	for _, oid := range []types.T{types.T_binary, types.T_varbinary, types.T_blob} {
+		t.Run(oid.String(), func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


XuPeng-SH (approved):
Approved from code-review side.\n\nThe fix addresses the actual data-branch merge failure by teaching the shared value formatter to handle BLOB and by switching the binary-family types (BINARY/VARBINARY/BLOB) to SQL hex literals. That is the right representation here: escaped string literals are not lossless for arbitrary bytes, while x'...' preserves control bytes, NUL, quotes, backslashes, and empty values.\n\nI checked the formatter call sites used by data branch diff/merge/PK value construction/conflict messages. Character and JSON values stay on the existing escaped-string paths, NULL remains handled before type dispatch, and the binary-family change is local to SQL literal generation. The new unit cases cover byte slices, string inputs, control bytes, and empty bytes; the BVT covers merge convergence plus BINARY, VARBINARY, BLOB, NULL, and empty-byte cases. CI is green.\n\nResidual non-blocking scope: this does not try to solve every currently unsupported visible type in formatValIntoString, but it closes the reported BLOB regression and also fixes the binary literal lossiness on the same path.

heni02 (approved):

Issue #25203, opened 2026-06-29
lockservice: LockTableAllocator stale Valid=false bind entries never garbage collected


XuPeng-SH:
## Summary

`LockTableAllocator.disableTableBinds()` in `lock_table_allocator.go` sets `bind.Valid = false` on table lock bindings but never removes them from the `lockTables` map. Over multiple CN restarts, stale entries accumulate with no garbage collection path.

## Root Cause

`lock_table_allocator.go:291-323`:
```go
func (l *LockTableAllocator) disableTableBinds(serviceID string) {
    old.Valid = false                                    // marked invalid
    l.getLockTablesLocked(serviceID)[table] = bind        // kept in map
    delete(l.mu.services, serviceID)                      // service removed
    // BUT: lockTables[serviceID] entry persists forever
}
```

No GC paths touch `lockTables` stale entries:
- `cleanCommitState`: cleans `commitCtl`, not `lockTables`
- `checkInvalidBinds`: operates on `lockTable.mu.bind`, not `allocator.lockTables`
- `Get()`: returns `Valid=false` entries, caller must check `Valid` field

## Impact

- Slow map growth in long-running LockTableAllocator (on TN nodes)
- Each CN restart adds one stale entry per table binding
- Each entry lightweight (map entries), but unbounded over months

## Suggested Fix

Option A: `delete` entries from `lockTables` map instead of setting `Valid=false`
Option B: Periodic sweep in `cleanCommitState` to remove `Valid=false` entries

## Related

Discovered during multi-CN transaction unhappy-path audit. See also: #25195, #25198

aptend:
Closing after re-checking the allocator data model and full rebind lifecycle.

The issue's central growth model is incorrect:

- lockTables is keyed by group ID and table ID, not by service ID.
- disableTableBinds marks the existing group/table entry invalid; it does not append a new per-service entry.
- A later bind for the same group/table reuses that entry, updates ServiceID, increments Version, and marks it valid.
- The invalid entry is a fencing tombstone used to preserve bind generation. Blind deletion or a periodic Valid=false sweep can reintroduce an ABA/stale-bind problem within the same allocator epoch.

Therefore repeated CN restarts do not add one lockTables entry per table per restart, and neither suggested option should be implemented as written.

[...]

PR #25571, opened 2026-07-10
fix(lockservice): ignore inactive waiters in deadlock detection

aptend:
## Summary

- exclude notified and completed waiter entries from distributed deadlock wait-for graphs
- preserve active remote wait-for edges by filtering them on the lock owner and representing the RPC response as snapshot edges on the caller
- keep the normal waiter notification lifecycle unchanged

## Related work

Follow-up to #25447, which added the owner-local remote deadlock path.

## Validation

- `go test ./pkg/lockservice -run '^TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots$' -count=20`
- `go test -race ./pkg/lockservice -run 'TestFetchWhoWaitingMe(UsesActiveRemoteWaiterSnapshots|SkipsInactiveWaiters)$|TestLockBlockedOnRemote' -count=5`
- `go test ./pkg/lockservice -count=1 -timeout=20m`
- `go vet ./pkg/lockservice`


XuPeng-SH (changes requested):
## Blocking finding

**[P1] Preserve active remote wait-for edges** (`pkg/lockservice/txn.go:430`).

The local waiter-state filter is correct, but this callback also receives locks reconstructed by `remoteLockTable.doGetLock`. Each `resp.GetTxnLock.WaitingList` entry is recreated with `acquireWaiter` and remains in the default `ready` state because the RPC carries no waiter status. The new `status == blocking` condition therefore drops every remote waiter, including genuinely blocking ones. A distributed cycle involving a remote-owned lock can become invisible to deadlock detection and remain blocked until transaction timeout.

Please make the active-edge contract explicit across the RPC boundary:

- Filter non-`blocking` waiters on the lock owner before serializing `GetTxnLock.WaitingList`.
- Preserve/represent returned entries as active snapshot edges on the caller without treating them as ordinary live waiters. Blindly setting synthetic waiters to `blocking` is not sufficient because synthetic lock cleanup then enters the notification lifecycle.
- Add an end-to-end regression where a local txn holds a remote-owned lock and the remote snapshot contains an active waiter plus inactive `notified`/`completed` entries, asserting that only the active edge reaches `fetchWhoWaitingMe`.

I reproduced the false negative with the existing remote RPC fixture: the response contained one active waiter, while `fetchWhoWaitingMe` returned an empty list.

Non-blocking metadata note: #25203 is closed and describes stale lock-table bindings/memory growth, not this deadlock-graph bug. Please remove or replace that relation.

XuPeng-SH (approved):
LGTM after reviewing this from the wait-for-graph, remote-snapshot, and unhappy-path angles.

- correctness: local deadlock traversal now ignores waiters that are already `notified` or `completed`, so stale queue residue no longer creates bogus edges
- remote-edge preservation: synthetic waiters reconstructed from `GetTxnLock` are explicitly marked as remote snapshots and stay visible to the caller-side wait-for graph, which keeps the active owner-local remote edge intact
- unhappy path coverage: the new tests cover both sides of the contract — inactive local waiters are skipped, while active remote waiter snapshots are still reported — without changing the normal notification lifecycle

Issue #25546, opened 2026-07-08
[Bug]: fulltext: single-keyword NL/DEFAULT top-k crashes on common terms (COUNT(*) OVER() window split), and is much slower than the heap path


cpegeric:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

890a3800c44932b4dd190ba59277aa73c502bd05

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

  Environment

  - Branch: main (introduced by #24027, commit 20d0ded07)
  - Single-node local deploy
  
  Reproduction

  Any table with a fulltext index where a single term appears in more than ~one batch (~8k) of documents:

  CREATE TABLE t (id BIGINT PRIMARY KEY, body TEXT);
  -- load enough rows that one term ('common') appears in > ~8k documents
  CREATE FULLTEXT INDEX ft ON t(body) WITH PARSER gojieba;  -- default parser also affected

  -- single keyword + NL/DEFAULT mode + LIMIT  → crashes
  SELECT id FROM t WHERE MATCH(body) AGAINST('common' IN NATURAL LANGUAGE MODE) LIMIT 10;
  -- ERROR 20101: the Window operator currently does not support sending split result of window function.

  In our test corpus (50k docs, ~1024 tokens each), the term 负责 matched 13,684 docs → crash. Terms matching fewer docs (≤ ~3.7k) return, but
  slowly.
  
  Unaffected: BOOLEAN mode and multi-term queries — they don't enter this path (singleKeywordPattern gates on FULLTEXT_NL/FULLTEXT_DEFAULT) and
  go through the heap path.
  
  Root cause

  runSingleKeywordTopK (pkg/sql/colexec/table_function/fulltext.go:806) → SingleKeywordTopKBM25SQL (pkg/fulltext/sql.go:187) generates:

  SELECT doc_id, score, nmatch FROM (
    SELECT a.doc_id, <bm25> AS score, COUNT(*) OVER() AS nmatch
    FROM (SELECT doc_id, tf … WHERE word = 'X' GROUP BY doc_id) a
    LEFT JOIN … dl … ) ranked
  ORDER BY score DESC LIMIT k
  
  The COUNT(*) OVER() (used to fetch nmatch for IDF) is a window function. When the match set is large, its aggregate flush produces more than
[...]

PR #24027, opened 2026-03-31
  optimize(fulltext): harden single-keyword fast path and fix BM25/runtime correctness

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24026 

### Summary

   This PR hardens MatrixOne fulltext execution while keeping only a behavior-safe optimization scope.

   The original optimization direction was useful, but review surfaced several correctness, lifecycle, and performance-shape risks.
  This change narrows the fast path to the safe subset, fixes the risky parts, and removes duplicated SQL work in the single-keyword
  limited path.

PR #25572, opened 2026-07-10
fix(fulltext): remove single-keyword window top-k path

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25546

The single-keyword NL/DEFAULT LIMIT fast path generated an internal SQL query with `COUNT(*) OVER()`. When more than 8192 matching documents reached the window operator, its aggregate result split into multiple vectors and the operator rejected the result. The same window implementation also evaluates the unpartitioned frame quadratically in the number of matches.

This change removes that fast path and routes single-keyword limited searches through the existing streaming evaluator and Top-K heap, which are already used by other fulltext query shapes. It preserves the #24027 correctness and lifecycle fixes while removing the generated window SQL and its unused helpers.

The regression coverage now:

- verifies TF-IDF/BM25 and NL/DEFAULT single-keyword LIMIT searches use streaming SQL;
- verifies membership-filter propagation on the streaming path; and
- verifies a zero-match LIMIT stream terminates cleanly and can be freed without blocking; and
- adds 8193-match TF-IDF and BM25 BVTs for both NL and DEFAULT syntax, with non-matching documents to keep IDF positive and result order deterministic.

Tested with:

```
go build ./pkg/fulltext ./pkg/sql/colexec/table_function
go vet ./pkg/fulltext ./pkg/sql/colexec/table_function
go test -v -count=1 ./pkg/fulltext ./pkg/sql/colexec/table_function
go test -tags matrixone_test -count=1 ./pkg/fulltext ./pkg/sql/colexec/table_function
go test -race -count=1 ./pkg/sql/colexec/table_function -run '^TestFullTextCallWithLimit(ZeroMatchStreams|SingleKeywordFallsBackToStreaming|PropagatesMembershipFilterToStreamingSQL)$'
```

The local worktree does not contain a running MatrixOne service, so the distributed BVT was added but not executed locally.


XuPeng-SH (approved):
Approved from code-review side.\n\nThe direction is sound and addresses both failure modes at their source: the single-keyword SQL fast path duplicates fulltext scoring/Top-K logic, depends on a window implementation that cannot emit split results, and performs a full ORDER BY over all matches. Removing it consolidates NL/DEFAULT single-keyword LIMIT execution onto the existing streaming evaluator and bounded heap, giving O(matches * log K) Top-K work and eliminating the COUNT(*) OVER() failure surface.\n\nI checked the complete execution path for TF-IDF and BM25, NL and DEFAULT modes, membership-filter propagation, score inputs (document frequency and BM25 document length), result ordering, binary document IDs, multi-row state reset, producer error/cancellation handling, and cleanup. The removed helpers have no remaining in-repository callers. The 8193-match BVT is well constructed: it crosses the old vector boundary, keeps IDF positive with non-matching rows, and uses distinct term frequencies for deterministic top-10 output.\n\nLocal verification on commit 7c368974ac992eb034b98db19310dd83ba59f572:\n- go test -count=1 ./pkg/fulltext ./pkg/sql/colexec/table_function\n- go test -race -count=1 ./pkg/sql/colexec/table_function -run '^TestFullTextCallWithLimit(SingleKeywordFallsBackToStreaming|PropagatesMembershipFilterToStreamingSQL)$'\n\nBoth passed.\n\nNon-blocking coverage improvements:\n- retain an explicit zero-match + LIMIT streaming regression after removing the old fast-path zero-match test;\n- add a large-match TF-IDF integration case as a symmetric counterpart to the BM25 BVT.\n\nSome required CI jobs are still running, so merge should continue to depend on those gates.

heni02 (approved):

Issue #25356, opened 2026-07-02
[Bug]: MySQL index hints return wrong rows or ignore missing composite indexes


Ariznawlll:
## Summary

On latest `main` (`0e8602b78`), MySQL index hints are accepted but can produce incorrect results with composite secondary indexes:

- `USE INDEX(...)` / `FORCE INDEX FOR ORDER BY(...)` / `IGNORE INDEX FOR ORDER BY(...)` return empty results for queries that return rows in MySQL.
- `FORCE INDEX(idx_missing)` does not report the missing index. MySQL returns `ERROR 1176`, while MO returns `0`.

## Duplicate check

Searched before filing:

- `FORCE INDEX`
- `USE INDEX`
- `IGNORE INDEX`
- `force index missing index`
- `USE INDEX returns zero rows`

No direct duplicate was found. Existing composite-index issues such as #25344/#25345/#25348/#25349/#25351 cover predicate/type/aggregate paths, not MySQL index-hint semantics.

## Reproduce

```sql
drop database if exists mysql_compat_horizontal_index_hint;
create database mysql_compat_horizontal_index_hint;
use mysql_compat_horizontal_index_hint;

create table t (
  id int primary key,
  a varchar(8) not null,
  b int not null,
  c int not null,
  status varchar(12),
  score int,
  dt datetime,
  key idx_a_b_c_id(a,b,c,id),
  key idx_b_a_id(b,a,id),
  key idx_status_dt(status,dt,id)
);

insert into t values
(2,'g02',2,14,'archived',22,'2024-03-03 02:00:00'),
(42,'g02',5,29,'archived',58,'2024-07-15 18:00:00'),
(81,'g01',7,37,'pending',83,'2024-10-26 09:00:00'),
(82,'g02',8,44,'archived',94,'2024-11-27 10:00:00'),
(112,'g02',1,42,'active',20,'2024-05-01 16:00:00'),
(152,'g02',4,4,'active',56,'2024-09-13 08:00:00'),
(155,'g05',7,25,'hold',89,'2024-12-16 11:00:00'),
(192,'g02',7,19,'active',92,'2024-01-25 00:00:00'),
(222,'g02',0,17,'archived',14,'2024-07-27 06:00:00'),
(232,'g02',10,34,'active',23,'2024-05-09 16:00:00');

select id,a,b,c
from t use index(idx_b_a_id)
where b=7 and a in ('g01','g03','g05')
order by a,id;

select id,a,b,c
[...]

PR #25573, opened 2026-07-10
fix(plan): honor mysql index hints

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25356

This PR adds planner support for MySQL index hints on regular secondary indexes. It records table-scan index hints during binding, validates hinted index names, filters regular-index candidates for scan-scope USE/FORCE/IGNORE hints, and returns the MySQL-compatible ER_KEY_DOES_NOT_EXIST / SQLSTATE 42000 error when a hinted index is missing.

It also adds regression coverage for composite secondary indexes, missing hinted indexes, and composite-primary-key tables with composite secondary indexes.

Validation:

- `go test -v -count=1 -run 'Test.*IndexHint|TestFilterRegularIndexesByScanHints|TestRecordIndexHintsValidatesNames' ./pkg/sql/plan`\n- `go test -v -count=1 ./pkg/sql/plan`\n- `go test -v -count=1 ./pkg/common/moerr`\n- `go build ./pkg/common/moerr ./pkg/sql/plan`\n- `go vet ./pkg/common/moerr ./pkg/sql/plan`\n- `mo-tester/run.sh -p test/distributed/cases/optimizer/index_hint.sql -g -n`\n- `git diff --check`


XuPeng-SH (changes requested):
[P1] `IGNORE/USE/FORCE INDEX FOR ORDER BY` is bypassed after a covering-index rewrite.

`indexHintsByScan` is keyed by node ID. `tryIndexOnlyScan` creates a replacement index-table scan but does not call `inheritIndexHints`, so the ORDER BY hint is lost. More importantly, even after adding that inheritance, the later `handleMessageFromTopToScan` pass independently turns the regular index into an ordered Top-N scan without calling `regularIndexScanAllowedByOrderHints`. The earlier gate in `applyRegularIndexTopSort` therefore does not close the optimizer pipeline.

Reproducer (fails on this head 10/10):
```sql
select id, a
from index_hint_t ignore index for order by (idx_a)
where a = 1
order by id
limit 10;
```

The resulting `idx_a` scan gets `OrderBy`, a `MsgTopValue`, and `IndexReaderParam`, so the optimizer uses the explicitly ignored index to provide ORDER BY / LIMIT ordering. This violates the PR's advertised scoped-hint contract.

Please (1) propagate the hint state when `tryIndexOnlyScan` replaces a scan node, and (2) gate the post-index `handleMessageFromTopToScan` ordered-read/Top-N transformation with the same ORDER BY eligibility predicate. Add a full planner regression (not only a direct candidate-filter test) that asserts the ignored index can still serve the `WHERE` access but has no ordered-read state: no `IndexReaderParam`, no scan `OrderBy`, and no Top-N receive message. I validated this combined fix locally: the new regression and the index-hint suite passed repeatedly, and `go test ./pkg/sql/plan -count=1` passed.

heni02 (approved):

XuPeng-SH (changes requested):
The previous ORDER BY-hint blocker still is not fully closed on the latest head.

1. `tryIndexOnlyScan()` still creates a replacement index scan without inheriting the original scan's hint state.

In the latest head, `pkg/sql/plan/apply_indices.go` still returns the new `idxTableNodeID` from `tryIndexOnlyScan()` without an `inheritIndexHints(...)` call. So for covering-index rewrites, the replacement scan can still lose `IGNORE/USE/FORCE INDEX FOR ORDER BY` state before the later optimizer passes run.

2. The later `handleMessageFromTopToScan()` pass still bypasses the ORDER BY hint gate.

The PR added `regularIndexScanAllowedByOrderHints()` and now checks it in `buildRegularIndexTopSortContext()`, but `pkg/sql/plan/message.go` still independently pushes `scanNode.OrderBy` / `IndexReaderParam` for the regular-index ordered-read path without consulting the same predicate. So even if you fix the hint inheritance problem above, an ignored regular index can still be turned into the ordered Top-N read by this later pass.

This is the same closure problem as before: the hint contract is not enforced end-to-end across the full optimizer pipeline. Please fix both sides together:
- inherit hint state when a scan node is replaced by `tryIndexOnlyScan()`;
- gate `handleMessageFromTopToScan()` with the same ORDER BY eligibility check before it adds ordered-read state.

The new tests also still miss this exact unhappy path. The added BVT covers positive `FORCE INDEX FOR ORDER BY` cases, but there is still no negative regression for `IGNORE INDEX FOR ORDER BY` across covering-index rewrite + later ordered-read rewrite.

XuPeng-SH (approved):
LGTM after reviewing this from the hint-binding, access-path selection, and unhappy-path angles.

- correctness: the planner now records scan-scope hint ownership, validates hinted names up front, and filters regular-index candidates consistently for USE/FORCE/IGNORE, which closes the empty-result and missing-index gaps from the issue
- path control: the new owner-tracking/inheritance keeps JOIN / covering-scan rewrites attached to the original hinted table scan instead of losing hint intent after the first rewrite
- unhappy path coverage: the new tests cover missing hinted indexes, empty USE, invalid hint combinations, composite secondary indexes, join/order/group scopes, and metadata-failure branches

PR #25574, opened 2026-07-10
fix: make remote dispatch registration race-safe

### LINKED ISSUE #25565: [Bug]: Refactor remote dispatch receiver registry to eliminate unstable registration/attach races
Title: [Bug]: Refactor remote dispatch receiver registry to eliminate unstable registration/attach races

Is there an existing issue for the same bug?

Branch Name
main

Commit ID
c600bf18ba003564955ee7962ec2bacead97d65d

Other Environment Information
- Hardware parameters: GitHub Actions nightly environment
- OS type: Linux
- Others:
  - Multi-CN execution path
  - External Hive partition queries
  - Reproduced in nightly run `29054645722`, job `86253126494`
  - Existing related issue: #25211
  - Existing direct fixes already merged before this commit: #25238, #25475

Actual Behavior
`#25211` is not fully closed at the root-cause level. On `main` at `c600bf18ba003564955ee7962ec2bacead97d65d`, the same workload still fails in the remote dispatch path with inconsistent lifecycle errors:

- `internal error: remote receiver already done`
- `dispatch process not registered within 30s`

The current remote receiver registry still relies on implicit map states and destructive transitions in `pkg/sql/colexec/server.go`. It conflates at least these different states:

- not registered yet
- registered and waiting to be attached
- already attached / already consumed
- canceled before registration
- cleanup tombstone

Because these states are encoded indirectly by `nil` fields and delete-on-read behavior, duplicate or reordered register/attach/cancel flows can report misleading terminal errors, remove valid entries, or turn a real upstream failure into a registry artifact.

This means the current fix set mitigates one race, but the lifecycle model is still fragile and not first-principles safe under retry, duplicate notification, cancellation, and concurrent cleanup.

Expected Behavior
The remote dispatch receiver lifecycle should be explicit, idempotent, and race-safe:

- registration, attach, cancel, and completion should use explicit state transitions instead of implicit `nil`/delete semantics
- duplicate register / duplicate notify / cleanup-after-cancel should not destroy a valid live entry
- sender retry should not be the primary correctness mechanism
- the first real root cause should be preserved instead of being overwritten by secondary registry errors
- multi-CN external queries should either succeed or fail with the true upstream reason, without `remote receiver already done` / `not registered within 30s` artifacts caused by registry state corruption

Steps to Reproduce
1. Run a multi-CN environment on `main` at commit `c600bf18ba003564955ee7962ec2bacead97d65d`.
2. Execute the external Hive partition workload associated with `#25211`.
3. Observe the latest nightly reproduction:
   - nightly run: `29054645722`
   - job: `86253126494`
4. In that run:
   - the `EXISTS` query succeeds after about 91 seconds
   - a fact/dim join then fails with `internal error: remote receiver already done`
   - a self-join then fails with `dispatch process not registered within 30s`
5. Inspect `pkg/sql/colexec/server.go` and related remote dispatch scheduling paths to see that the registry API is still destructive and state-ambiguous even after #25238 and #25475.

Additional information
Relevant code paths:

- `pkg/sql/colexec/server.go`
- `pkg/sql/compile/scopeRemoteRun.go`
- `pkg/sql/colexec/dispatch/`

Relevant observations from code review:

- `GetProcByUuid` consumes live entries and mutates registry state as part of lookup
- `PutProcIntoUuidMap` can delete an existing entry and return `remote receiver already done`, which is unsafe for duplicate registration or reordered cleanup
- `DeleteUuids` uses tombstones, but tombstones, live entries, attached entries, and canceled entries are not modeled explicitly
- current retry logic in #25475 improves timing sensitivity, but does not fix lifecycle ambiguity

Suggested fix direction:

1. Replace the implicit receiver map with an explicit state machine:
   - `pending`
   - `registered`
   - `attached`
   - `completed`
   - `canceled`
2. Split APIs into non-destructive operations such as:
   - `RegisterReceiver`
   - `AttachReceiver`
   - `CancelReceiver`
   - `CompleteReceiver`
3. Make duplicate operations idempotent and generation-safe.
4. Add a registration-ready acknowledgment between remote scope setup and sender notify, instead of relying primarily on polling/retry.
5. Preserve the first root error and improve observability with receiver UUID, query ID, source CN, target CN, and lifecycle transition logs/metrics.
6. Add focused tests for:
   - duplicate registration
   - duplicate notify
   - cancel-before-register
   - register-vs-cleanup races
   - attach-vs-timeout races
   - end-to-end multi-CN external join execution

This issue is intentionally separate from `#25211`: `#25211` tracks the user-visible failure, while this one tracks the deeper registry/lifecycle redesign still needed to make the fix set complete and robust.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25565
Related to #25211

This PR fixes two remote dispatch registration races found while investigating the recurring Hive JOIN failures in #25211:

1. A retry attempt no longer tombstones an absent receiver UUID when its RPC connection closes. The attempt connection does not own the query-wide receiver lifecycle, so late registration remains attachable by the next retry.
2. Coordinator/local dispatch receivers are published before `Compile.runOnce` starts any scope goroutine. This establishes the required happens-before relation between receiver registration and remote notification.

The implementation also:

- registers each dispatch with its owning `scope.Proc`;
- distinguishes local scope trees, coordinator-retained remote root dispatches, and remote-owned subtrees;
- transfers early-registration state to the retained local dispatch runner to avoid duplicate registration;
- cancels owner processes and performs owner-checked rollback on partial registration failure;
- releases pending attaches when `RemoteRun` fails before `Dispatch.Prepare`;
- includes receiver UUID and the previous `live`/`tombstone` state in registration conflict errors.

A full registry state-machine redesign remains possible follow-up hardening, but is not required for these two proven failure paths.

## Validation

- `go test -count=1 ./pkg/sql/colexec ./pkg/sql/colexec/dispatch`
- focused `go test -race` across `pkg/sql/colexec`, `pkg/sql/colexec/dispatch`, and `pkg/sql/compile`
- `go build ./pkg/sql/colexec ./pkg/sql/colexec/dispatch ./pkg/sql/compile`
- `go vet ./pkg/sql/colexec ./pkg/sql/colexec/dispatch ./pkg/sql/compile`
- `git diff --check`

The full `pkg/sql/compile` suite still has three parquet external-scan fanout failures; the same command and assertions fail on clean base commit `0fcd615a7f`, so they are unrelated to this change.


XuPeng-SH (changes requested):
Requesting changes on head `6a5784ebc`.

The two primary fixes are directionally correct: a typed retry attempt no longer tombstones an absent query-wide receiver, and local/coordinator receiver registration now happens before scope goroutines start. The scope ownership traversal, retained-root state transfer, cancellation on RemoteRun failure, and owner-token cleanup all look coherent under the covered paths.

## High: a failed duplicate registration still deletes the existing live/attached owner

`Server.PutProcIntoUuidMap` still mutates the registry destructively on conflict (`pkg/sql/colexec/server.go:105-110`): whenever the UUID already exists, it deletes that entry and then returns `remote receiver ... already done`.

That violates the atomicity promised by the new owner-checked rollback. `prepareRemote` can successfully register UUID A and then encounter an existing owner for UUID B. It rolls back A using the new channel owner token, but B has already been deleted by `PutProcIntoUuidMap` even though B was never owned by this registration attempt. The previous valid receiver can no longer attach, and subsequent notification falls back to NotRegistered/timeout behavior, recreating the exact registry artifact this PR is intended to remove.

I reproduced this deterministically with a temporary regression:

1. pre-register UUID B with `oldProc` / `oldCh`;
2. register a new Dispatch containing `[A, B]`;
3. registration of A succeeds and B conflicts;
4. the call returns an error, then `GetProcByUuid(B, false)` returns `ok=false` instead of the original owner.

The assertion fails on the current head with:

```text
Should be true: a failed duplicate registration deleted the existing owner
```

The current tests miss this because `TestPutProcIntoUuidMapConflictIncludesUuidAndState` checks only the error text, and `TestRegisterRemoteReceiversRollbackOnPartialFailure` checks only that this attempt removes A; neither verifies that the pre-existing B owner survives.

Please make conflict handling non-destructive for an existing live or attached registration. Note that `proc == nil` alone is not enough to classify an entry as a cancellation tombstone: an attached registration also has `proc/ch == nil` but retains a non-nil `ownerCh`. At minimum the state handling should distinguish:

- live owner: `proc != nil`, preserve it;
- attached owner: `proc == nil && ownerCh != nil`, preserve it;
- pre-registration cancellation tombstone: no owner token, apply the explicitly intended tombstone transition.

Then add regressions for both live-owner and attached-owner conflicts, including the `[A, conflicting B]` partial-registration rollback. Only UUIDs written by the failing attempt should be removed.

Local validation on the PR head:

```text
go test ./pkg/sql/colexec ./pkg/sql/colexec/dispatch -count=1
PASS

go test -race ./pkg/sql/compile -run focused-registration-tests -count=5
PASS

temporary existing-owner preservation regression
FAIL deterministically
```

CI currently has no failed checks, but this uncovered ownership case is a correctness blocker independent of CI.

aptend:
Addressed the change request in `b17ff9b33d`.

- Registration conflicts now preserve existing `live` and `attached` owners.
- Only an ownerless cancellation tombstone is consumed/deleted on a late registration attempt.
- Conflict diagnostics now distinguish `live`, `attached`, and `tombstone`.
- Added direct owner-preservation tests and `[A, conflicting B]` partial-registration tests for both live and attached B. The failing attempt rolls back only A.

Validation:

- `go test ./pkg/sql/colexec ./pkg/sql/colexec/dispatch -count=1`
- focused related tests under `-race`
- focused owner/rollback tests repeated with `-count=10`
- `git diff --check`

XuPeng-SH (approved):
Approved on head `b17ff9b33d897568134e111e919a375e0783ba01`.

The previous correctness blocker is resolved. Registration conflicts are now non-destructive for both live and attached owners, while only an ownerless cancellation tombstone is consumed. The new direct registry tests and `[A, conflicting B]` partial-registration regressions verify that rollback removes only entries owned by the failing attempt.

I also rechecked the broader lifecycle:

- local/coordinator receivers are published before any scope goroutine starts;
- remote-owned subtrees are skipped while retained remote roots are registered with the owning scope process;
- repeated traversal is idempotent and retained-root cleanup state is transferred back safely;
- partial registration failure cancels affected processes and performs owner-checked rollback;
- RemoteRun failure cancels the owner process, releasing a notify blocked before Dispatch.Prepare;
- normal/error cleanup removes live or attached entries without allowing delayed cleanup to delete a newer owner.

No production hot-path regression was found. The extra work is one pre-execution scope/operator walk plus the same O(receiver) registry inserts that previously occurred during Prepare. The explicit registry state-machine redesign remains useful follow-up hardening, but I did not find it necessary to close either proven race fixed here.

Local validation:

- `go test ./pkg/sql/colexec ./pkg/sql/colexec/dispatch -count=1`
- focused compile registration/interleaving tests under `-race -count=10`
- owner/conflict/rollback tests under `-race -count=20`
- `git diff --check`

All CI checks, including UT, multi-CN BVT, coverage, benchmark, and both upgrade jobs, are green.

PR #25575, opened 2026-07-10
fix varbinary equal

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25241

• - 将 BINARY x VARBINARY 和 VARBINARY x BINARY 的共同类型改为 VARBINARY，避免把短 VARBINARY 隐式补零到 BINARY(255)。
  - 增加类型规则单测，覆盖两个操作数方向。
  - 扩展 BVT：普通比较、补零字节、<=>、排序、IN、子查询 IN、索引前后比较及行构造器 IN。
  - 更新设计文档，说明普通比较和 IN/子查询最终复用等值比较类型规则。


XuPeng-SH (changes requested):
The targeted `BINARY`/`VARBINARY` change is correct, but the common-type fix is incomplete and leaves the same root correctness bug in adjacent rules.

1. **High: `BINARY` with `BLOB` (and `TEXT`) is still coerced to fixed-length `BINARY`, so equal byte strings compare unequal.**

   The changed rows make variable length dominate for `BINARY` x `VARBINARY`, but the immediately adjacent rules still map both directions of `BINARY` x `BLOB` and `BINARY` x `TEXT` to `BINARY` (`type_check.go:1078-1079` and `1127`, plus the symmetric `TEXT` rule). I reproduced the resulting wrong answers on this exact PR binary (`fdf8535970a6d78139628d7ab2b51772f733c3f6`):

   ```sql
   CREATE TABLE tb(id INT PRIMARY KEY, bl BLOB, b BINARY(4));
   INSERT INTO tb VALUES
     (1, 'a', 'a'),
     (2, X'61000000', 'a');

   SELECT id, HEX(bl), HEX(b),
          bl = BINARY 'a', BINARY 'a' = bl,
          bl = b, b = bl
   FROM tb ORDER BY id;
   ```

   MySQL 8.0.44 returns:

   ```text
   1  61        61000000  1  1  0  0
   2  61000000  61000000  0  0  1  1
   ```

   This PR returns:

   ```text
   1  61        61000000  0  0  0  0
   2  61000000  61000000  0  0  0  0
   ```

   Notice that row 2 is false even when both stored values are byte-identical. Replacing `BLOB` with `TEXT` reproduces the same mismatch against MySQL.

   Please fix this as a binary/string common-type lattice rather than another isolated pair. The invariant is that coercion for comparison must not turn a variable-length payload into a fixed-length `BINARY`, because implicit padding changes the value. Existing padding already stored in a `BINARY(N)` column must remain significant, while the other operand must not be padded merely for comparison. At minimum, audit and fix both directions of `BINARY` with `BLOB` and `TEXT`; add a symmetric matrix test so future entries cannot drift.

2. **Low: the index section does not assert that the target index path is exercised.**

   The new BVT creates `idx_binary_varbinary_subquery_b` and `idx_binary_varbinary_subquery_vb`, then repeats subquery `IN` queries. On the PR binary, `EXPLAIN` shows those queries remain table-scan SEMI joins on both sides, so this only proves that index *presence* does not alter the result. The important target predicate does perform well: I verified that both `vb = BINARY 'a'` and `BINARY 'a' = vb` use `idx_vb` on this PR. Please add an `EXPLAIN` regex assertion for those direct predicates to lock in the no-column-cast/index-usable property.

Validation performed on the PR head:

- full `go test ./pkg/sql/plan/function -count=1`: pass
- complete modified BVT SQL against a locally built PR `mo-service`: pass
- target `VARBINARY` results against MySQL 8.0.44: match
- target direct predicates use the secondary index in both operand directions
- `BLOB`/`TEXT` reproductions above: fail against the MySQL oracle

daviszhen:
> The targeted `BINARY`/`VARBINARY` change is correct, but the common-type fix is incomplete and leaves the same root correctness bug in adjacent rules.
> 
> 1. **High: `BINARY` with `BLOB` (and `TEXT`) is still coerced to fixed-length `BINARY`, so equal byte strings compare unequal.**
>    The changed rows make variable length dominate for `BINARY` x `VARBINARY`, but the immediately adjacent rules still map both directions of `BINARY` x `BLOB` and `BINARY` x `TEXT` to `BINARY` (`type_check.go:1078-1079` and `1127`, plus the symmetric `TEXT` rule). I reproduced the resulting wrong answers on this exact PR binary (`fdf8535970a6d78139628d7ab2b51772f733c3f6`):
>    ```sql
>    CREATE TABLE tb(id INT PRIMARY KEY, bl BLOB, b BINARY(4));
>    INSERT INTO tb VALUES
>      (1, 'a', 'a'),
>      (2, X'61000000', 'a');
>    
>    SELECT id, HEX(bl), HEX(b),
>           bl = BINARY 'a', BINARY 'a' = bl,
>           bl = b, b = bl
>    FROM tb ORDER BY id;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    MySQL 8.0.44 returns:
>    ```
>    1  61        61000000  1  1  0  0
>    2  61000000  61000000  0  0  1  1
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    This PR returns:
>    ```
>    1  61        61000000  0  0  0  0
>    2  61000000  61000000  0  0  0  0
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Notice that row 2 is false even when both stored values are byte-identical. Replacing `BLOB` with `TEXT` reproduces the same mismatch against MySQL.
>    Please fix this as a binary/string common-type lattice rather than another isolated pair. The invariant is that coercion for comparison must not turn a variable-length payload into a fixed-length `BINARY`, because implicit padding changes the value. Existing padding already stored in a `BINARY(N)` column must remain significant, while the other operand must not be padded merely for comparison. At minimum, audit and fix both directions of `BINARY` with `BLOB` and `TEXT`; add a symmetric matrix test so future entries cannot drift.
> 2. **Low: the index section does not assert that the target index path is exercised.**
>    The new BVT creates `idx_binary_varbinary_subquery_b` and `idx_binary_varbinary_subquery_vb`, then repeats subquery `IN` queries. On the PR binary, `EXPLAIN` shows those queries remain table-scan SEMI joins on both sides, so this only proves that index _presence_ does not alter the result. The important target predicate does perform well: I verified that both `vb = BINARY 'a'` and `BINARY 'a' = vb` use `idx_vb` on this PR. Please add an `EXPLAIN` regex assertion for those direct predicates to lock in the no-column-cast/index-usable property.
> 
> Validation performed on the PR head:
> 
> * full `go test ./pkg/sql/plan/function -count=1`: pass
> * complete modified BVT SQL against a locally built PR `mo-service`: pass
> * target `VARBINARY` results against MySQL 8.0.44: match
> * target direct predicates use the secondary index in both operand directions
> * `BLOB`/`TEXT` reproductions above: fail against the MySQL oracle

已经修复

heni02 (approved):

XuPeng-SH (approved):
最新 head 看下来，我上次提的两个点都已经补上了，可以 approve。

- `BINARY`/`VARBINARY` 不再只是修一个孤立 pair，`BINARY`/`BLOB`、`BINARY`/`TEXT` 这条相邻 common-type 路径也一起修到了，新的 `function_test` 和 BVT 已经把这组 byte-semantics 覆盖进来了。
- 之前缺的索引断言也补了：现在直接比较 `vb = binary x'41'` 和 `binary x'41' = vb` 都有 `EXPLAIN` regex，能锁住“不对列侧做错误 cast、索引仍可用”这个目标性质。

按当前 scope，我这里没有剩余 blocker 了。

aunjgr (approved):

PR #25577, opened 2026-07-10
perf(plan): push down regular index cursor limit

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23965

This completes the safe index-reader LIMIT pushdown for regular secondary-index cursor pagination when:

- every secondary-index prefix part is fixed by equality;
- the only residual predicate is `<`, `<=`, `>`, or `>=` on the base-table primary key;
- the query orders by that primary key and uses a positive literal LIMIT without OFFSET or rank semantics.

The planner combines the fixed prefix and cursor into a `prefix_in_range` predicate on the hidden serialized index key. The range is evaluated before the ordered reader LIMIT, so the reader can continue from the cursor boundary and return the requested rows without scanning and sorting the full prefix group.

The original primary-key cursor predicate remains in the scan filter list as a correctness recheck.

## Root Cause

PR #25545 correctly disabled static reader LIMIT pushdown whenever an index scan had residual filters. Before that guard, the reader could take the first N entries in hidden-key order and only then apply `id < cursor`, which could return too few rows or zero rows.

That guard restored correctness but also disabled early termination for cursor predicates that can be represented exactly as a range on the same hidden ordered key. As a result, the cursor query still read the complete prefix group before the final Top-N sort.

## Safety and compatibility

- No storage, index format, reader API, or SQL semantics are changed.
- The optimization is enabled only by the cursor-range rewrite performed in the two existing regular-index ORDER BY LIMIT planner paths.
- Static reader LIMIT is applied only when the SORT path reaches the table scan directly. A FILTER or cardinality-reducing INNER/SEMI JOIN on that path disables static LIMIT pushdown.
- TopValue messaging and ordered scanning remain enabled across those FILTER/JOIN paths, preserving the existing dynamic optimization without risking an incomplete result.
- Additional residual filters, incomplete index prefixes, non-primary-key predicates, computed cursor expressions, float primary keys, type mismatches, OFFSET, rank semantics, and non-literal LIMIT values retain the existing conservative behavior.
- Prepared cursor parameters and casts over stable literal/parameter values are supported.
- In-memory and unsorted-block behavior continues to use the existing fallback paths.

## Testing

- `go test -count=1 ./pkg/vm/engine/readutil`
- `go test -count=1 ./pkg/vm/engine/tae/blockio`
- Added planner regressions proving static reader LIMIT is not pushed through FILTER, INNER JOIN, or SEMI JOIN.
- Extended `regular_index_order_limit_cursor` BVT coverage for `<`, `<=`, `>`, and `>=` cursor boundaries.


heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes on head `5610e283ee71d090a7d9d51243556d2b9e2383cc`.

## High: the message path pushes the static reader LIMIT below unaccounted Filter and Join operators

The new cursor safety check proves only that `scanNode.FilterList` is exactly `[full-prefix equality, one PK range]`. It does not prove that the scan candidates are the rows observed by the SQL LIMIT.

`gatherLeavesForMessageFromTopToScan` explicitly walks through `Node_FILTER` and the left side of INNER/SEMI JOINs (`pkg/sql/plan/message.go:22-41`). After that traversal, `handleMessageFromTopToScan` enables `IndexReaderParam.Limit` based only on the scan-local filters (`message.go:80-86,110-112`). A Filter or Join between the reader and Sort can reject rows after the reader has already stopped at N, which recreates the same “too few rows / zero rows” correctness failure that #25545 fixed.

I reproduced both cases deterministically against the current head with temporary planner regressions:

1. `Sort DESC LIMIT 20 -> Filter(id < 4800) -> index scan(prefix equality, id < 4900)`
2. `Sort DESC LIMIT 20 -> INNER JOIN(index scan(prefix equality, id < 4900), ...)`

In both cases, the current code sets a non-nil `scanNode.IndexReaderParam`. In case 1, the reader can take 4899..4880, the Filter removes all 20 rows, and the query returns zero even though 4799 and below can satisfy the LIMIT. The Join case has the same failure whenever the first N index candidates have no match.

This is specific to the modified message path: the direct project/sort rewrite does not cross Filter or Join nodes.

Please separate the two different safety questions currently conflated by the traversal:

- TopValue messaging/ordered scan may continue to traverse the operators for which that protocol is valid.
- Static reader LIMIT pushdown must additionally require that **every** operator between the reader and SQL LIMIT preserves the candidate set. Any residual Filter or cardinality-reducing Join should disable `enableOrderedLimit` unless its effect has been represented exactly in the reader range (a Join generally cannot be).

A practical fix is for the traversal to return both the scan ID and a `staticLimitSafe` property, and require that property before calling `rewriteRegularIndexCursorRangeFilter` / `applyRegularIndexOrderedLimitParam`. Please add regressions for both Filter and INNER/SEMI JOIN paths, including a predicate/join that rejects the first N ordered candidates.

The range construction itself, reversed operands, open/closed flags, type/stability guards, original cursor recheck, and persisted/in-memory/unsorted reader ordering otherwise look sound. Local package tests pass. Non-blocking test note: the BVT validates result rows but not reader inputRows, so an execution-level assertion for the intended early-termination performance would strengthen the closure.

LeftHandCold:
The review is valid and has been addressed in `ae7314f2e0`.

- `gatherLeavesForMessageFromTopToScan` now returns both the scan ID and whether a static reader LIMIT is safe.
- Traversal through `FILTER`, `INNER JOIN`, or `SEMI JOIN` marks static LIMIT pushdown unsafe, preventing the reader from stopping before those operators reject rows.
- Existing TopValue messaging and ordered scanning are preserved on those paths.
- Added regressions for FILTER, INNER JOIN, and SEMI JOIN, while retaining the direct scan positive case.

Validation: `go test -count=1 ./pkg/sql/plan` passed after merging the latest `main`.

XuPeng-SH (approved):
Re-reviewed the latest head after the requested changes. The previous correctness blocker is resolved: static reader LIMIT pushdown is now explicitly disabled across FILTER and cardinality-reducing INNER/SEMI JOIN paths, while TopValue messaging and ordered scanning remain available. The added regressions cover those unsafe paths and the direct-scan positive path.

I also rechecked range-bound construction, reversed comparisons, open/closed flags, cursor type/stability guards, OFFSET/rank/non-literal LIMIT fallbacks, and the persisted/in-memory/unsorted reader behavior. No blocking findings remain.

Local validation on the latest main merge passed for `pkg/sql/plan`, `pkg/vm/engine/readutil`, and `pkg/vm/engine/tae/blockio`, including targeted race repetitions and `go vet`. CI is green.

Non-blocking follow-up: an execution-level assertion that the cursor case keeps index-scan `inputRows` near LIMIT would protect the performance objective in addition to the current result-correctness BVT.

PR #25582, opened 2026-07-10
fix(txn): finalize unknown commits without rollback GC

### LINKED ISSUE #25198: CN transaction: permanent workspace/offheap memory leak in FinalizeCommitWithUnknownResult
## Summary

`FinalizeCommitWithUnknownResult` never calls `delTransaction()`, causing permanent workspace, offheap batch vectors, quota, and cached metadata to leak indefinitely when a commit encounters `ErrTxnUnknown`. The leak persists until CN process restart.

## Root Cause

There is a clear asymmetry between the three finalization paths:

| Path | Calls `delTransaction()`? | Cleans workspace/batch/quota? |
|------|---------------------------|-------------------------------|
| `FinalizeCommit()` | ✅ Yes (line 2148) | ✅ |
| `Rollback()` | ✅ Yes (line 2236) | ✅ |
| `FinalizeCommitWithUnknownResult()` | ❌ **No** (line 2156–2158) | ❌ |

### Code comparison

```go
// disttae/txn.go - FinalizeCommit (line 2143-2148):
func (txn *Transaction) FinalizeCommit(context.Context) {
    if txn.isCCPRTxn... { ... }
    txn.delTransaction()  // ✅ cleans up
}

// disttae/txn.go - FinalizeCommitWithUnknownResult (line 2150-2158):
func (txn *Transaction) FinalizeCommitWithUnknownResult(context.Context) {
    if txn.isCCPRTxn... { ... }
    // Comment says: "Keep workspace/object-stat metadata intact"
    // txn.delTransaction() is NEVER called ❌
}
```

### What `delTransaction()` actually releases

| Resource | Type | Cleanup |
|----------|------|---------|
| `writes[i].bat` | batch + offheap C-malloc'd vectors | `bat.Clean(mp)` → `vec.Free(mp)` → `mpool.Free` → `C.free` |
| `tableCache` / `tableOps` / `databaseOps` | maps/slices | set to `nil` (GC) |
| `compactWorker` | goroutine pool | `Release()` |
| `extraWriteWorkspaceThreshold` | engine quota | `engine.ReleaseQuota()` |
| `deletedBlocks` | slice | set to `nil` |

None of these are freed in the `ErrTxnUnknown` path.

### Full call chain

```
Commit() → doWrite() → defer {
    if ErrTxnUnknown {
        FinalizeCommitWithUnknownResult()  // never cleans up
        return                              // Rollback is NOT called either
    }
}
→ closeLocked() → ClosedEvent → closeTxn()
  // Only removes txnOperator from txnClient.activeTxns
  // The disttae Transaction object and all its resources persist
```

## Trigger conditions

- Commit RPC times out (network partition, TN restart)
- TN returns `ErrTxnUnknown` (uncertain commit outcome)
- CN calls `FinalizeCommitWithUnknownResult` intending to preserve workspace for potential S3 GC

## Impact

- **Offheap (C heap) memory leaks**: batch vectors allocated via `mpool` with `offHeap=true` → `C.malloc` memory never freed until process exit
- **Quota leak**: `extraWriteWorkspaceThreshold` quota permanently consumed
- **Silent accumulation**: no observability or timeout-driven cleanup
- **Risk scenario**: long-running CN process with intermittent network issues → hundreds of MB to GB of leaked offheap memory over days/weeks

## Suggested Fix

Add either:

1. **A timeout-driven cleanup**: If `FinalizeCommitWithUnknownResult` has been unresolved for N minutes, forcibly call `delTransaction()` to release resources.

2. **An explicit cleanup trigger**: Allow the caller to signal that the unknown result has been resolved (either committed or aborted externally), at which point `delTransaction()` is called.

3. **Defensive partial cleanup**: At minimum, call `delTransaction()` on CN shutdown or after confirming the commit cannot be recovered.

## Related

- #25195 — previously reported Rollback resource management issues in `service_cn_handler.go`
- Double-check report confirmed this is the only real leak (two other suspected issues were disproven)

aptend:
## Summary

- make ErrTxnUnknown a terminal, non-destructive workspace finalization path
- classify every single-request Commit response/transport failure, including context deadline, as ErrTxnUnknown
- release CN-local batches, caches, worker pools, clone tracking, and workspace quota through delTransaction
- atomically fence any later Rollback from running workspace object GC
- invalidate the frontend transaction handle after returning ErrTxnUnknown
- keep the transaction outcome unknown instead of rewriting it to Aborted

## Safety invariant

A missing Commit response does not prove abort. This PR never deletes transaction objects on the unknown path because TN may already have committed metadata that references them.

## Known remaining gaps

If TN actually aborted and never checkpointed a CN-created object, existing TN checkpoint GC cannot discover it. Such objects can remain indefinitely and accumulate until a durable unknown-object candidate manifest and TN consumer are implemented. This PR is the immediate data-safety and CN-resource fix, not complete orphan-object lifecycle closure.

The multi-request WriteAndCommit sender path does not yet classify every response failure as unknown. No production caller was found. Fixing that API safely needs stream-generation or in-flight fencing so a late response cannot be delivered after a pooled stream is reused.

## Tests

- full pkg/txn/rpc, pkg/txn/client, pkg/frontend, and pkg/vm/engine/disttae
- repeated Commit deadline and EOF/reset classification tests
- race tests for response-deadline classification and unknown finalization followed by direct rollback

Fixes #25198

### DIFF
--- a/pkg/frontend/txn.go
+++ b/pkg/frontend/txn.go
@@ -592,6 +592,11 @@ func (th *TxnHandler) commitUnsafe(execCtx *ExecCtx) error {
 		}
 		execCtx.ses.updateLastCommitTS(commitTs)
 		if commitResultUnknown {
+			// ErrTxnUnknown is terminal for this frontend handle. The operator
+			// has already finalized its workspace non-destructively; retaining it
+			// lets later rollback/connection cleanup reach destructive cleanup.
+			th.invalidateTxnUnsafe()
+			execCtx.ses.SetTxnId(dumpUUID[:])
 			return err
 		}
 	}
--- a/pkg/frontend/txn_test.go
+++ b/pkg/frontend/txn_test.go
@@ -786,8 +786,8 @@ func Test_commit(t *testing.T) {
 	})
 }
 
-func TestCommitTxnUnknownPreservesTxnOperatorForLaterCleanup(t *testing.T) {
-	convey.Convey("commit ErrTxnUnknown keeps txn operator reachable", t, func() {
+func TestCommitTxnUnknownInvalidatesTxnOperator(t *testing.T) {
+	convey.Convey("commit ErrTxnUnknown invalidates the frontend txn operator", t, func() {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
@@ -818,7 +818,8 @@ func TestCommitTxnUnknownPreservesTxnOperatorForLaterCleanup(t *testing.T) {
 		convey.So(moerr.IsMoErrCode(err, moerr.ErrTxnUnknown), convey.ShouldBeTrue)
 		convey.So(txnOp.commitCalls, convey.ShouldEqual, 1)
 		convey.So(txnOp.rollbackCalls, convey.ShouldEqual, 0)
-		convey.So(ses.GetTxnHandler().GetTxn(), convey.ShouldEqual, txnOp)
+		convey.So(ses.GetTxnHandler().GetTxn(), convey.ShouldBeNil)
+		convey.So(ses.GetTxnHandler().InActiveTxn(), convey.ShouldBeFalse)
 	})
 }
 
--- a/pkg/txn/client/operator.go
+++ b/pkg/txn/client/operator.go
@@ -300,7 +300,7 @@ type txnOperator struct {
 		createAt             time.Time
 		commitAt             time.Time
 		createTs             timestamp.Timestamp
-		cannotCleanWorkspace bool
+		cannotCleanWorkspace atomic.Bool
 		workspace            Workspace
 		commitCounter        counter
 		rollbackCounter      counter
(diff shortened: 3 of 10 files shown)


XuPeng-SH (approved):
Reviewed the unknown-commit state machine end to end: remote send classification, operator finalization and locking, frontend handle invalidation, disttae/CCPR/clone cleanup, object-GC exclusion, direct later rollback, resource bounds, and production call sites. I found no blocking correctness, robustness, leak/hang/OOM, compatibility, or performance issue.

The safety boundary is sound for this PR: an ambiguous Commit never runs object GC; CN-local batches, caches, pools, clone tracking, segment tracking, and quota are released; the operator is closed and removed; the frontend handle is invalidated; and a retained direct handle cannot later run destructive workspace rollback. The declared durable orphan-object cleanup and unused multi-request WriteAndCommit path are real follow-up gaps, but they do not invalidate this immediate data-safety/CN-resource fix.

Non-blocking: the new comment in `pkg/frontend/txn.go` above `invalidateTxnUnsafe()` says that "retaining it lets later rollback/connection cleanup reach destructive cleanup", while the implementation intentionally invalidates the handle to prevent that behavior. Please reword it to match the new invariant.

Local validation passed: full tests for `pkg/txn/rpc`, `pkg/txn/client`, `pkg/frontend`, and `pkg/vm/engine/disttae`; race tests for txn/rpc+client and the frontend/disttae unknown paths; repeated deadline/finalization tests; `go vet`; and `git diff --check`. CI is fully green.

PR #25585, opened 2026-07-10
perf: eliminate avoidable heap escapes on hot paths (escape-analysis driven)

fengttt:
## What

A set of small, targeted fixes that remove **avoidable heap escapes** on hot paths, found by auditing `go build -gcflags=all=\"-m -m\"` escape-analysis output. Each fix is verified with `-gcflags=-m` (escape gone), builds, `go vet`, and tests.

## Commits

1. **`types.Decode(x, &local)` interface boxing (per-row/per-column)** — `types.Decode` takes an `encoding.BinaryUnmarshaler`; passing `&local` boxes it into the interface and heap-escapes the local every iteration. Call the concrete `local.Unmarshal(...)` instead (semantically identical — `Decode` just calls `v.UnmarshalBinary`). Fixed in 8 sites: the 6 `builtInInternalChar*/Numeric*/Datetime*/CharacterSet` functions (`func_builtin.go`, per-row when serving `information_schema.columns`), plus `catalog/tuplesParse.go` and `disttae/cache/catalog.go` (per-column on table-schema materialization).

2. **`vector.InplaceSort` — `sort.Slice` → generic `slices.Sort`** — `sort.Slice` boxes the slice into `any`, escapes the capturing closure, and sorts via runtime reflection (`reflect.Swapper`). Replaced all 27 type cases with `slices.Sort` (ordered types) / `slices.SortFunc` (custom comparators built from the same predicate the original used). Result: **0 allocs/op** (int64 & varchar, 8192 rows), reflection removed; 54 escapes → 0.

3. **`Rowid.ComparePrefix` — panic-path `fmt` leaked the prefix arg** — `panic(fmt.Sprintf(\"...%X\", to))` boxed `to` into `...any`, marking it a leaking param for the whole method, forcing every caller's prefix slice to the heap (callers are zone-map prefix comparison on the scan-filter path). Format `string(to)` (a copy) so `to` no longer leaks.

4. **`bitToStr` (bit→string/binary cast) — per-row encode escape** — the closure returned a slice aliasing its local `v` (via `EncodeUint64(&v)`), forcing `v` to the heap every row. Encode into a buffer captured once per call via `binary.LittleEndian.PutUint64` (takes `v` by value); byte-identical on little-endian targets, reuse is safe because every consumer copies.

## Notes

- All changes are behavior-preserving. The custom `slices.SortFunc` comparators reproduce `sort.Slice`'s exact ordering (`!less(a,b) && !less(b,a)` ⇒ equal). `slices.Sort` additionally gives floats deterministic NaN ordering.
- Deliberately **not** changed: escapes that are necessary (aliased/returned `unsafe` byte slices), one-time (init/setup), error-path-only (panic/Sprintf args), or per-batch/spill boundaries — these are not per-row and carry no GC benefit.


XuPeng-SH (approved):
Reviewed the latest head from correctness, ordering semantics, allocation ownership, cross-path compatibility, performance, and unhappy-path/resource-lifecycle angles. The direct Type.Unmarshal calls are equivalent to types.Decode; the generic comparators preserve the prior ordering predicates while giving floats a coherent NaN order; the reusable BIT cast buffer is copied by every result path before reuse; and no new unbounded state, wait, or resource lifetime is introduced. Local targeted tests, vet, race tests, and an additional multi-row BIT-to-binary buffer-reuse audit passed. No blocking findings.

PR #25587, opened 2026-07-10
fix(plan): preserve pagination across LIMIT optimizations

### LINKED ISSUE #25586: [Bug]: master index applies LIMIT/OFFSET twice and returns wrong pages

### Branch Name

`optimize_limit`

### Commit ID

`899e3a236e`

- Hardware parameters: local Linux/x86 development environment
- OS type: Linux
- Others: planner-level reproduction using the MatrixOne mock optimizer; the failure is independent of data distribution once at least `LIMIT + OFFSET` rows match.

### Actual Behavior

A query using a master index with `LIMIT ... OFFSET ...` receives the same pagination clause twice:

1. `applyIndicesForFiltersUsingMasterIndex` puts the original `Limit` and `Offset` on the final master-index scan/intersection node.
2. It then copies the same `Limit` and `Offset` to the outer `INDEX` join used for the base-table lookup.

Both nodes are compiled through `compileSort`, so the inner operation first skips `OFFSET` rows and limits the candidate set; the outer lookup then skips `OFFSET` rows again.

For a query with at least 15 matching rows, `LIMIT 10 OFFSET 5` can therefore return only 5 rows and starts from the wrong logical page.

The isolated plan reproduction produced both of these nodes:

```text
JOIN (master-index intersection)  limit=10 offset=5
JOIN (INDEX/base-table lookup)    limit=10 offset=5
```

Relevant code:

- `pkg/sql/plan/apply_indices_master.go:82-85`
- `pkg/sql/plan/apply_indices_master.go:113-120`
- `pkg/sql/compile/compile.go:1118-1131`

### Expected Behavior

`LIMIT 10 OFFSET 5` must be applied exactly once to the final result. When at least 15 rows match, the query must return 10 rows after skipping the first 5 rows.

The index path may fetch `LIMIT + OFFSET` candidates internally, but only the final layer may apply the user-visible OFFSET.

### Steps to Reproduce

```sql
CREATE TABLE t_master_limit (
  id INT PRIMARY KEY,
  a VARCHAR(20),
  b VARCHAR(20),
  INDEX idx_master USING master (a, b)
);

-- Insert at least 15 rows with the same indexed values.
INSERT INTO t_master_limit
SELECT result, 'same-a', 'same-b'
FROM generate_series(1, 30);

SELECT id
FROM t_master_limit
WHERE a = 'same-a' AND b = 'same-b'
LIMIT 10 OFFSET 5;
```

Inspect the logical/physical plan. The current master-index rewrite attaches `LIMIT 10 OFFSET 5` to both the internal master-index node and the outer INDEX join. The result can contain only 5 rows rather than 10.

### Additional information

Severity rationale: this is a stable wrong-result bug in an active secondary-index query path. It has a confirmed owner and a small deterministic regression-test shape.

Suggested fix direction:

- keep the original `LIMIT/OFFSET` only on the outer INDEX join; or
- fetch `LIMIT + OFFSET` from the inner index path and leave the user-visible OFFSET only on the outer layer, using overflow-safe arithmetic.

Add regression coverage for one and multiple master-index predicates, with and without OFFSET.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25586

This PR makes LIMIT/OFFSET ownership explicit across planner rewrites and ensures that internal candidate pruning never replaces user-visible pagination.

- folds constant LIMIT/OFFSET expressions early and normalizes OFFSET 0;
- uses an overflow-safe LIMIT + OFFSET candidate budget for master, ANN, and single-stream fulltext index paths;
- preserves original pagination through vector plugin dispatch and on the final result node;
- safely skips internal pruning for dynamic offsets, overflow, explicit fulltext ordering, and independent fulltext intersections;
- evaluates prepared LIMIT expressions once during table-function Prepare and makes hidden index-reader expressions visible to prepared-plan visitors;
- hardens stats, Top-to-scan messaging, range extraction, reader setup, ANN/fulltext allocation, and HNSW signaling against wrong results, nil dereferences, overflow, lost wakeups, and OOM amplification.

## CODEOWNERS update

The `CODEOWNERS` changes are intentional and are part of the requested maintainer ownership update; they are not generated files or incidental churn from the LIMIT work.

- SQL planning/execution, pipeline, and related protobuf ownership moves from `@ouyuanning` to `@XuPeng-SH`, while existing co-owners remain unchanged.
- `pSpool` and `testutil` remove the former owner while retaining their other current owners.
- `multi_update` is assigned to `@XuPeng`.
- proxy ownership moves from `@volgariver6` to `@XuPeng-SH`, while `@gouhongshen` remains a co-owner.

Keeping these entries in this PR makes GitHub review routing and future maintenance responsibility match the intended ownership immediately after merge.

## Multi-angle review

- Performance: removes eager O(LIMIT) allocations, bounds ANN backend/result capacities, avoids LIMIT 0 over-fetch, preserves Top/ordered/index paths, and delays TopValue shuffle mutations until the optimization is accepted.
- Accuracy: keeps final LIMIT/OFFSET separate from candidate budgets across plugin boundaries, preserves duplicate distance predicates as residual filters, supports one-sided ranges, and makes pagination stats idempotent.
- Unhappy paths: covers uint64 overflow, NULL/dynamic expressions, invalid reader metadata, fulltext 3x overflow, HNSW condition signaling, extreme GPU limits, and partial Prepare failures.

## Validation

- affected Go packages pass, including targeted race tests;
- `go vet` passes for all changed packages;
- `make build` passes after rebasing onto current main;
- mo-tester BVT: fulltext 16/16, master index 231/231, IVF mode 88/88, IVF advanced 41/41, vector edge cases 13/13;
- 100k-row operator benchmark: Top 9.90 ms and 2.27 MB versus full Order 20.68 ms and 6.55 MB, about 2.1x faster and 65% less memory.

GPU search implementations use shared, CPU-tested LIMIT bounds. GPU runtime validation was not available because the local NVIDIA driver and cuVS libraries are unavailable.


CODEOWNERS
```diff
@@ -84,7 +84,7 @@
 # pkg/container
 /pkg/container @XuPeng-SH
 /pkg/container/vector @aunjgr @XuPeng-SH
-/pkg/container/pSpool @aunjgr @ouyuanning @iamlinjunhong
+/pkg/container/pSpool @aunjgr @iamlinjunhong
```
gouhongshen: 这个 PR 的目标是分页修复，但这里以及同一文件的多处改动移除了 @ouyuanning，并把 SQL/plan/VM/pb/proto 等路径的 owner 转给作者。这是独立的治理/审批变更，PR 描述也没有说明。请从本 PR 移除；若确有需要，应拆成单独 PR 并按 CODEOWNERS 流程确认。

pkg/sql/plan/apply_indices.go
```diff
@@ -112,6 +113,31 @@ func calculateFilteredPostModeOverFetchFactor(originalLimit uint64) float64 {
...
+	}
+	if factor < 1 {
+		factor = 1
+	}
+	multiplied := originalLimit
+	if factor > 1 {
+		if float64(originalLimit) > float64(math.MaxUint64)/factor {
+			multiplied = math.MaxUint64
+		} else {
+			multiplied = uint64(float64(originalLimit) * factor)
```
gouhongshen: 这个 guard 不能保证转换安全。以 factor=1.2、originalLimit=15372286728091293696 为例，前一行比较为 false，但 float64(originalLimit)*factor 已舍入为 2^64；把它转换为 uint64 是 Go 规范定义的 implementation-dependent 行为。请在转换前检查乘积是否 >= float64(math.MaxUint64)（并显式饱和），并覆盖这个边界；当前只测 MaxUint64，覆盖不到此分支。

pkg/sql/plan/apply_indices_vector.go
```diff
@@ -513,17 +527,35 @@ func vectorSearchProviderChildren(vecCtx *vectorSortContext) []int32 {
...
+	if vecCtx.resultLimit != nil {
+		return DeepCopyExpr(vecCtx.resultLimit), DeepCopyExpr(vecCtx.resultOffset)
+	}
+	// Compatibility for algorithm-level callers that construct the context
+	// directly instead of going through buildVectorSortContext.
+	var offset *plan.Expr
+	if vecCtx.sortNode != nil {
+		offset = vecCtx.sortNode.Offset
+	}
+	return DeepCopyExpr(vecCtx.limit), DeepCopyExpr(offset)
```
gouhongshen: 这里的兼容分支重新混淆了两个已被刻意拆开的概念：vecCtx.limit 现在是内部 candidate budget（LIMIT+OFFSET），不是结果 LIMIT。若 ResultLimit 缺失而 limit=15、sortNode.Offset=5，最终节点会得到 LIMIT 15 OFFSET 5，而不是原始 LIMIT 10 OFFSET 5。请不要静默 fallback 到 candidate limit；要求所有构造/插件 round-trip 填充 result pagination，或在缺失时放弃 rewrite。

pkg/sql/plan/message.go
```diff
@@ -61,25 +74,30 @@ func (builder *QueryBuilder) handleMessageFromTopToScan(nodeID int32) {
...
+	}
 	scanNode := builder.qry.Nodes[scanID]
 	if len(scanNode.OrderBy) != 0 {
 		return
 	}
 	scanOrderBy := DeepCopyOrderBySpec(node.OrderBy[0])
+	scanOrderBy.Expr = scanOrderExpr
 	enableOrderedLimit := false
-	if canUseRegularIndexHiddenSortKey(scanNode, orderByCol) {
+	if orderByCol.RelPos == scanNode.BindingTags[0] && canUseRegularIndexHiddenSortKey(scanNode, scanOrderByCol) {
```
gouhongshen: gatherLeavesForMessageFromTopToScan 已把 ORDER BY 映射为 scanOrderByCol，但这个条件仍使用投影层的 orderByCol.RelPos。直接投影后两者的 binding tag 不同，因此会跳过常规索引的 hidden-sort-key / ordered-limit 优化。这里应使用 scanOrderByCol；请补一个带 regular index 的直接投影测试，断言 IndexReaderParam/ordered limit 被设置。

gouhongshen (changes requested):
这组修改修复了主索引分页，但当前还不能合入：包含未说明的 CODEOWNERS 所有权转移，并且仍有极值分页与向量重写的语义缺口。请处理以下行级问题后再请求 review。

heni02 (approved):

gouhongshen (approved):
Re-review complete. The overflow boundary, vector pagination ownership, and direct-projection regular-index pushdown issues are fixed with targeted regression coverage. Approved.

pkg/sql/plan/apply_indices_fulltext.go
```diff
@@ -61,15 +61,38 @@ func (builder *QueryBuilder) applyIndicesForProjectionUsingFullTextIndex(nodeID
...
 	}
+	// Pagination belongs to the result of the base-table/fulltext join, never
+	// to the base scan and fulltext stream independently. Both inputs are
+	// unordered, so paginating them separately can select disjoint subsets.
+	scanNode.Limit = nil
+	scanNode.Offset = nil
 
 	if sortNode != nil {
 		sortNode.Children[0] = idxID
+		sortNode.Limit = DeepCopyExpr(paginationLimit)
```
fengttt, line 94: All these deepcopy, nobody cares why

fengttt (approved):

fengttt (approved):

Issue #25588, opened 2026-07-11
[Performance]: Preserve early filtering for partition-state OR, large PREFIX_IN, and mixed ranges filters


XuPeng-SH:
## Is there an existing issue for performance?

- [x] I have checked the existing issues.

## Environment

- Version or commit-id: `7821d21c72`
- Branch: `optimize_filter_0711`
- OS: Linux
- Scope: partition-state/workspace reads and ranges block-filter compilation

## Details of Performance

Three confirmed paths preserve correctness but lose available early filtering and can scan substantially more in-memory rows, blocks, or object metadata than necessary.

1. **Partition-state OR/disjunct filters**
   - `BasePKFilter.Disjuncts` makes `NewMemPKFilter` invalid because the partition-state primary-key iterator accepts only one atomic predicate.
   - Reads therefore fall back to `NewRowsIter` and scan all visible in-memory rows even when every OR branch is individually primary-key-filterable.

2. **Large `PREFIX_IN` in workspace reads**
   - `MemPKFilter.FilterVector` intentionally returns without filtering when `PREFIX_IN` contains more than four prefixes.
   - This avoids repeated linear matching, but turns a selective prefix set into a full workspace scan.

3. **All-or-nothing ranges compilation**
   - `readutil.CompileFilterExprs` marks the entire filter list as non-compilable when any one expression lacks a fast compiler implementation.
   - The supported predicates then also lose object/block-level fast filtering, causing avoidable metadata loads and block scans.

These are performance gaps, not correctness bugs: residual SQL filters still produce correct query results.

## Expected Behavior

- OR/disjunct primary-key predicates should use a safe union-capable partition-state iterator, or an equivalent bounded batch filter, whenever all branches are supported.
- Large `PREFIX_IN` sets should use a scalable representation (for example, sorted prefixes with binary search/trie/range pruning) instead of disabling filtering.
- Ranges compilation should retain all independently compilable fast filters and route only unsupported predicates through the slow evaluator.

## Steps to Reproduce

1. Create a table with a primary key and leave selective rows in the transaction workspace/partition state.
2. Execute a predicate that produces `BasePKFilter.Disjuncts`, such as `pk = 1 OR pk = 1000000`; observe that the partition-state path cannot use the primary-key iterator.
3. Execute `PREFIX_IN` with more than four prefixes against workspace rows; observe that `MemPKFilter.FilterVector` skips early filtering.
[...]

PR #25596, opened 2026-07-11
fix: strengthen filter pushdown and PK pruning

### LINKED ISSUE #25592: [Bug]: Composite-key pruning loses constituent block filters after stats recalculation

## Branch Name

main

## Commit ID

7821d21c72

- Hardware parameters: N/A
- OS type: Linux
- Others: Optimizer block-filter planning for composite primary keys and composite cluster keys.

## Actual Behavior

When composite-key predicate rewriting replaces predicates on physical key parts with a predicate on the hidden compound key, later `ReCalcNodeStats` rebuilds `BlockFilterList` from the rewritten `FilterList`. The physical-part predicates are then lost.

This preserves query results but removes valid zonemap pruning opportunities for each constituent key column. A high-cost/AP stats pass can repeat the loss after an earlier pass had retained the part filters.

## Expected Behavior

The block-filter list should retain both:

- the compound-key predicate, for sorted-key/read-filter pruning; and
- every consumed, zonemappable predicate on a constituent key column, for independent zonemap pruning.

Repeated stats recalculation must neither lose nor duplicate these predicates.

## Steps to Reproduce

1. Create a table with a composite primary key or composite cluster key.
2. Run a query such as `WHERE c1 = 5 AND c2 > 3 AND c2 < 100`.
3. Inspect `EXPLAIN` before the fix: only the hidden compound-key predicate appears in `Block Filter Cond`.
4. Trigger a plan path with repeated stats recalculation; the part-column predicates remain absent.

## Additional Information

The fix preserves consumed constituent predicates across stats recalculation, retains the generated compound predicate, and deduplicates equivalent filters. Regression coverage includes unit tests and `test/distributed/cases/optimizer/blockfilter.test`.

### LINKED ISSUE #25593: [Bug]: Function-scan filter pushdown can silently drop non-pushable predicates

## Branch Name

main

## Commit ID

7821d21c72

- Hardware parameters: N/A
- OS type: Linux
- Others: SQL optimizer filter pushdown through `FUNCTION_SCAN`.

## Actual Behavior

`pushdownFilters` forwarded filters that did not belong to a `FUNCTION_SCAN` node to its child, but discarded the child’s returned `cantPushdown` filters. A mixed-tag predicate could therefore disappear from the plan rather than being kept above the function scan.

This is a correctness issue: a query can return rows that should have been filtered out.

## Expected Behavior

Every filter rejected by a `FUNCTION_SCAN` child must be returned to the parent pushdown caller and materialized as a normal filter at the correct plan level. A function scan with no child must also return, rather than discard, incoming non-self filters.

## Steps to Reproduce

1. Build a plan with a `FUNCTION_SCAN` above a table scan.
2. Use a predicate that references both the function-scan binding tag and the child binding tag.
3. Run filter pushdown before the fix.
4. Observe that the predicate is neither attached to the child nor returned as `cantPushdown`.

## Additional Information

The regression test constructs a mixed-tag equality predicate and verifies that it is returned intact, with neither scan receiving an invalid partial pushdown.

### LINKED ISSUE #25594: [Bug]: Primary-key read filters can miscompare typed values and prune valid rows

- [x] I have searched the existing issues.

### Branch Name

main

### Commit ID

N/A

### Actual Behavior

Primary-key read-filter construction and merge paths do not consistently compare values with their SQL type semantics. Scalar range merging can compare encoded bytes rather than typed values, which is not order-preserving for little-endian numeric encodings. IN-list intersection and range construction also rely on comparison helpers that are not safe for every typed value (notably signed decimal values and edge values), and may receive nullable or unsorted serialized vectors.

This can construct an overly restrictive read filter, prune valid rows, or fail unexpectedly on malformed/unsupported serialized filter data.

### Expected Behavior

All primary-key filter comparisons must use type-correct, overflow-safe ordering. IN vectors should be validated and normalized without mutating planner-owned serialized data. Unsupported or malformed read-filter data must fail open (disable pruning) rather than panic or cause false-negative pruning.

### Steps to Reproduce

1. Use a numeric primary key and a predicate whose OR terms merge scalar ranges containing values such as `10` and `256`; inspect the merged filter on a little-endian platform.
2. Use a Decimal64 primary key with negative and positive IN/range values and intersect or merge filters.
3. Supply an unsorted nullable IN vector, or malformed serialized IN-vector bytes, through the read-filter path.
4. Observe that the constructed bounds/set can be incorrect, or that the read path is not robust to invalid input.

### Additional information

The fix uses typed comparisons, validates and normalizes IN vectors on owned memory, expands supported primitive PK types, and adds regressions for numeric ordering, signed Decimal64 values, nullable/unsorted inputs, malformed inputs, and block-filter mismatches.

### LINKED ISSUE #25595: [Bug]: Partition-state filtering does not apply composite-key prefix open ranges

- [x] I have searched the existing issues.

### Branch Name

main

### Commit ID

N/A

### Actual Behavior

Composite-primary-key predicate rewriting can produce prefix open-range filters such as `(prefix, value)`, `(prefix, +∞)`, `(-∞, prefix)`, and closed/open variants. The partition-state in-memory filter and primary-key iterator do not represent all of these forms. They therefore fall back to scanning workspace/partition rows instead of filtering and terminating the iterator by primary-key bounds.

Results remain correct after residual filtering, but selective workspace queries lose expected early pruning and can become unnecessarily expensive.

### Expected Behavior

Partition-state filtering and the primary-key iterator should support all composite-key prefix range forms, apply them to workspace rows, and stop iteration as soon as an upper bound proves that no later key can match.

### Steps to Reproduce

1. Create a table with a composite primary key `(c1, c2)` and workspace/partition-state rows.
2. Query a selective predicate such as `c1 = 5 AND c2 > 3`, or the corresponding left/right/open/closed range variants.
3. Inspect the partition-state path: the predicate is not represented as an in-memory PK filter and rows are scanned rather than pruned by range bounds.

### Additional information

The fix adds prefix open-range support to the in-memory filter and iterator, includes workspace filtering and upper-bound early termination, and covers all boundary inclusivity combinations.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25592
Fixes #25593
Fixes #25594
Fixes #25595

Related to #25588

Makes filter pushdown both safer and more effective across planner, block-filter, primary-key read-filter, partition-state, and iterator paths.

- Retains compound and constituent block filters through query-statistics recalculation, so composite-key predicates can prune using both the compound zonemap and individual column zonemaps.
- Preserves non-pushable predicates above function scans instead of dropping them.
- Uses typed, overflow-safe primary-key comparisons; validates and normalizes IN vectors on owned memory; supports additional primitive PK types; and fails open for malformed/unsupported filter data.
- Adds partition-state and iterator support for composite-key prefix open ranges, including early upper-bound termination.
- Reduces hot-path allocation and repeated work by reusing retained block-filter expressions, removing map-based OR union construction, and avoiding unnecessary deep copies.

## Validation

- `go test ./pkg/container/types ./pkg/container/vector ./pkg/vm/engine/readutil ./pkg/vm/engine/disttae/logtailreplay ./pkg/sql/plan`
- `go test -race ./pkg/container/types ./pkg/container/vector ./pkg/vm/engine/readutil ./pkg/vm/engine/disttae/logtailreplay`
- `go vet ./pkg/container/types ./pkg/container/vector ./pkg/vm/engine/readutil ./pkg/vm/engine/disttae/logtailreplay ./pkg/sql/plan`
- `make build`
- Distributed BVT: `test/distributed/cases/optimizer/blockfilter.test` (55/55 passed)

## Performance notes

Selective composite-key and prefix-range queries can now prune more blocks and partition-state rows, while iterator upper bounds stop scans early. The implementation avoids per-row maps and unnecessary expression/vector copies in hot paths. Unresolved optimizer limitations that can still cause conservative fallback are tracked in #25588.


XuPeng-SH (commented):
Blocking findings:

1. **High - NaN breaks the new typed range merge and can prune valid rows.**

`compareBasePKValues` now obtains a total order for floats through `types.CompareValue`/`cmp.Compare`, but SQL predicates and the block-search functions use IEEE comparisons, where every ordered comparison with NaN is false. For a FLOAT64 PK containing `1` and `2`, `pk >= CAST('NaN' AS DOUBLE) OR pk >= 1` is semantically `pk >= 1`; the OR merge treats NaN as the minimum bound and produces `pk >= NaN`, whose sorted search returns no offsets. I reproduced the exact `[0,1] -> nil` false-negative locally.

Please reject NaN bounds/IN members from PK pruning and fail open (infinity remains ordered), or introduce comparison/search semantics that preserve SQL's unordered behavior end to end. Add scalar AND/OR and IN regressions for float32/float64, covering sorted block, unsorted block, and workspace/memory paths.

2. **High - malformed IN type mismatch fails open on persisted blocks but filters out workspace rows.**

`validBlockPKSearchFilter` accepts any non-null IN vector without checking `filter.Oid == filter.Vec.GetType().Oid`. `buildBlockPKSearchFuncs` later guards the block path by input type, but `NewMemPKFilter` uses the same weak validation and immediately encodes the mismatched vector. With an INT64 PK and a VARCHAR IN vector containing `"1"`, the block path keeps all rows while `MemPKFilter.FilterVector` marks the matching INT64 workspace row as skipped. I reproduced `skipMask.Count() == 1`; malformed/unsupported data therefore still causes false-negative pruning.

Use one shared validation contract before either filter is constructed: IN must match the PK OID; PREFIX_IN must match the explicitly supported compound/varlen domain. A mismatch must make both block and memory filters invalid. Add a regression that executes both paths from the same malformed BasePKFilter.

3. **Blocking CI - static analysis rejects the four new `fmt.Errorf` calls.**

`pkg/vm/engine/readutil/pk_filter_base.go:401,428,432,441` violates MatrixOne's `moerr` rule. The arm64 SCA job already fails with `Use moerr instead`. Replace all four errors with appropriate `moerr` constructors and remove the `fmt` import.

Verified locally on head `09aa3b78f17ce32d3870a7cb6490a23105ffc25c`:

- existing unit tests for `vector`, `readutil`, `logtailreplay`, and `sql/plan` pass (the standalone `types` test binary is blocked locally by the external `libusearch_c.dylib` runtime path);
- `go test -race` passes for `readutil`, `logtailreplay`, and `sql/plan`;
- the two focused negative regressions above both fail exactly as described;
- no additional leak, hang, or unbounded-growth issue was confirmed in the changed paths.

XuPeng-SH:
Addressed all blocking findings in `b185f2b7ca`:

1. Float32/Float64 NaN bounds and IN members now disable PK pruning before filter merge or search-function construction. Both operand orders and AND/OR merges are covered; infinities remain ordered and optimizable.
2. Persisted-block and workspace filters now share one validation contract. Exact IN requires the vector OID to match the PK OID; PREFIX_IN requires a matching explicitly supported prefix domain. Malformed layouts fail open consistently on both paths.
3. Replaced the four `fmt.Errorf` calls with `moerr` constructors. No recover whitelist is needed: serialized-vector headers, field lengths, null bitmaps, physical data, and varlena area bounds are validated before zero-copy unmarshal.

Additional unhappy-path closure:

- malformed/version-skewed serialized IN filters stay on the residual-expression path instead of failing the query;
- huge constant IN vectors collapse to one set element, and huge const-NULL vectors collapse to an empty set, preventing logical-length CPU/allocation amplification;
- common scalar, float-IN, and PREFIX_IN validation paths are covered by a zero-allocation regression.

Validation completed locally:

- `make err-check`
- `go vet -vettool=$(which molint) ./pkg/vm/engine/readutil`
- `golangci-lint run -c .golangci.yml ./pkg/vm/engine/readutil` (0 issues)
- full `readutil` tests and affected package tests
- `go test -race ./pkg/vm/engine/readutil`
- `go vet` for all affected packages
- `make build`

XuPeng-SH:
Fixed the remaining CI failures in `403d5ef947`:

- deduplicate block filters across folded `IN` list/vector representations;
- deduplicate associative flattened/nested `AND`/`OR` filters without quadratic matching;
- retain distinct predicates (different IN values are covered by a negative regression);
- update the partition-prune expectation for the corrected sorted-vector metadata;
- accept the intentional `in_domain` plan change that now exposes both the compound-key and constituent-column block filters.

Verified locally:

- full tests for `pkg/sql/plan` and `pkg/partitionprune`;
- focused race tests for both regressions;
- all branch-affected package tests (`types`, `vector`, `plan`, `logtailreplay`, `readutil`, `partitionprune`);
- `go vet`, `molint`, `make err-check`, and `golangci-lint` (0 issues);
- `explain_index.sql`: 69/69;
- `in_domain.test`: 40/40.

I also checked the completed optimistic, pessimistic, proxy, compose, Ubuntu UT, and coverage logs. They reduce to these same five BVT comparisons plus the same `TestConvertFoldExprToNormal` expectation; no additional failure signature was present.

pkg/container/vector/vector.go
```diff
@@ -4805,7 +4805,12 @@ func (v *Vector) InplaceSortAndCompact() {
 			cleanDataNotResetArea()
 			appendList(v, newCol, nil, nil)
 		}
+	default:
```
fengttt, line 4808: you sure this is a return, not an error?

XuPeng-SH: Yes, the `return` is intentional here. Before this change, unsupported OIDs fell through the switch and `InplaceSortAndCompact` was a no-op; the method has no error-return contract. This PR adds the shared `SetSorted(true)` after the switch, so the explicit default return preserves the existing behavior and, importantly, avoids falsely marking an unsupported vector as sorted. `TestInplaceSortAndCompactMarksUniqueVectorsSorted` covers this with `T_any` and asserts that it remains unsorted. Turning this into an error would require changing the public method signature and all callers, which is outside this metadata-invariant fix.

gouhongshen (changes requested):
Blocking findings on 6260242e5f4fc2e0182dee668698ddfbd7e1685a:

1. High — block-filter dedup can panic on a valid constant LiteralVec. blockFilterConstantSet accepts a vector whenever its logical Length matches LiteralVec.Len, then calls GetConstantValue once per logical row (expr_opt.go:420-428). A CONSTANT vector with logical length > 1 has exactly one physical value: MustFixedColNoTypeCheck intentionally returns a one-element slice for it (vector/tools.go:27-33). Therefore a list-vs-constant-Vec IN comparison indexes row 1 and panics during planning. Do not index a constant vector by its logical row: reduce it to its single value (or simply decline semantic dedup for constants), and add a regression using NewConstFixed(..., length=2) as LiteralVec.

2. High — the new global dedup pass regresses planner complexity from bounded linear passes to O(B^2 * K) for B block filters containing K-element IN predicates. Every pair first performs structural comparison; Expr_Vec takes the proto-marshalling fallback, then the semantic path unmarshals both vectors and rebuilds value maps (expr_opt.go:377-434, 510-520). This runs after index application for every scan. exprStructuralHash already exists specifically to bucket dedup without this repeated work (expr_hash.go:24-32), but is not used here. Precompute a canonical key once per filter and use hash buckets plus exact collision checking; add a large-IN / many-block-filter regression or benchmark.

3. P2 — sameGeneratedCompoundFilter treats two compound predicates as equal solely by operation and target column, without comparing RHS (expr_opt.go:475-492). It is used both while rebuilding stats block filters and in final dedup. The stated single-generated-predicate invariant is neither encoded nor checked, so two distinct compound IN/range predicates can silently lose one pruning constraint. Dedup must retain structural/semantic equality, or carry an explicit generation identity; please add a negative test with different compound RHS.

The PK-filter cleanup handoff in the latest commit was checked through reader reset/Close and looks sound. I am requesting changes for the planner panic and the avoidable planner-scale regression.

XuPeng-SH:
Addressed in 3a8fc3cd9.

- Fixed constant `LiteralVec` handling: semantic set extraction now reads the single physical value of a constant vector (and handles zero-length vectors), so a logical length greater than one cannot index past physical storage. Added regression coverage with `NewConstFixed(..., length=2)`.
- Replaced pairwise global dedup with a precomputed semantic hash and hash buckets. Hash collisions still use exact semantic comparison; the same helper is used while rebuilding stats block filters. Added a large-IN (`128 x 512`) vector benchmark.
- Removed the operation+compound-column shortcut. Compound predicates are removed only when structurally/semantically equal; added coverage proving distinct RHS values are retained.

Validation: `go test ./pkg/sql/plan`, targeted tests with `-race -count=5`, `go build ./pkg/sql/plan`, `go vet ./pkg/sql/plan`, and dependent `pkg/sql/compile` tests all pass.

PR #25608, opened 2026-07-11
fix: preserve workspace files after unknown commit

### LINKED ISSUE #25591: [Bug] AI-WALK (Txn workspace): lost commit response causes rollback GC of committed data

- [x] I searched open and closed issues for ambiguous commit, commit timeout, workspace rollback GC, missing object files, and the `AI-WALK` title prefix. I did not find the same data-loss path.

## Branch Name

`main`

## Commit ID

`efe2379669c61a0a7b1a34dd9bf94e59bcf61132` (2026-07-10 21:00:24 +0800)

- Hardware: 20 vCPU (Intel Core i9-10900K), 62 GiB RAM
- OS: Debian GNU/Linux 12 (bookworm), Linux 6.1.0-31-amd64, x86_64
- Docker server: 29.1.3
- MO image: source-built `matrixorigin/matrixone:workspace-attack-efe2379669`, image ID `sha256:5de7da05e3adb305b7cf1a35b8619c6e032c10f083af3a1fd2a1de0224b1c9e8`
- Deployment: isolated Docker multi-CN local-disk cluster generated from `etc/docker-multi-cn-local-disk`: 1 Log Service, 1 TN, 2 CNs, 1 proxy, shared `/mo-data`; CN1 port 26001, CN2 port 26002
- Fault injection: `tc netem` on the TN container drops only TN -> CN1 packets. CN1 -> TN commit requests and CN SQL traffic remain available.

## Actual Behavior

A transaction can commit successfully on TN, then CN receives a raw commit-response I/O timeout, treats it as an ordinary failure instead of an unknown commit result, calls workspace rollback, and asynchronously deletes the CN-produced object already referenced by the committed transaction.

The table is initially visible from CN2. After the rollback GC removes the object and CN2's file cache is cleared by restart, every read fails with `ERROR 20405 ... file ... is not found`. This is committed-data loss, not only a leaked object or an ambiguous client result.

The failure was reproduced twice with 100,000 and 20,000 inserted rows. The minimized 20,000-row run produced this sequence:

```text
01:46:19 CN-COMMIT-S3-Data txn=70f70f4866a9b62a18c0fd68ac9ab752 row-cnt=20000 obj-cnt=1
01:46:49 txn send requests failed: Commit ... read ... i/o timeout
01:46:59 Transaction.Rollback txn=70f70f4866a9b62a18c0fd68ac9ab752
01:46:59 GC-WORKSPACE-FILES names=[019f4d23-3066-7152-909b-60425ec5a4b0_00000]
01:46:59 COMMIT status=fail, same i/o timeout
01:49:01 CN2 cold read: file 019f4d23-3066-7152-909b-60425ec5a4b0_00000 is not found
```

Before cache clear, CN2 returned all 20,000 committed rows and accepted a marker insert. After restarting CN2, this query failed deterministically:

```sql
select count(*), sum(id), min(id), max(id)
from workspace_partition_attack.chaos_t;
-- ERROR 20405 (HY000): file 019f4d23-3066-7152-909b-60425ec5a4b0_00000 is not found
```

The same missing object also repeatedly breaks merge tasks. The file is absent from the shared local fileservice directory.

## Expected Behavior

Once a commit request may have reached TN, a transport timeout while waiting for the response must be classified as an unknown commit result. CN must preserve workspace object metadata/files until the commit outcome is authoritatively resolved. It must never run abort-only GC solely because the commit response was lost.

The client may receive an ambiguous commit error, but committed table data must remain readable and mergeable after cache eviction and CN restart.

## Steps to Reproduce

1. Start the repository multi-CN local-disk configuration with two CNs. The reproduced topology used separate Docker containers on one network and a shared `/mo-data` volume.
2. On CN1, create the table and prepare a large transaction so the workspace flushes one CN object:

```sql
drop database if exists workspace_partition_attack;
create database workspace_partition_attack;
use workspace_partition_attack;
create table chaos_t(id bigint primary key, pad varchar(1024));
begin;
insert into chaos_t
select *, rpad('x', 512, 'x') from generate_series(1, 20000) g;
```

3. Resolve CN1's container IP and install a TN egress filter that drops only packets destined for CN1:

```bash
CN1_IP=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' wa-cn1)
docker exec -u 0 wa-tn tc qdisc add dev eth0 root handle 1: prio
docker exec -u 0 wa-tn tc qdisc add dev eth0 parent 1:3 handle 30: netem loss 100%
docker exec -u 0 wa-tn tc filter add dev eth0 protocol ip parent 1:0 prio 3 u32 \
  match ip dst "$CN1_IP/32" flowid 1:3
```

4. Send `COMMIT` from CN1 using a client read timeout of 8 seconds. Keep the response partition for 30 seconds after the client timeout, then restore traffic:

```bash
docker exec -u 0 wa-tn tc qdisc del dev eth0 root
```

5. Query CN2. The 20,000 rows are initially visible, proving TN committed the transaction. Wait for CN1's error path to log `Transaction.Rollback` and `GC-WORKSPACE-FILES` for the transaction object.
6. Restart CN2 to clear the fileservice cache, wait for SQL readiness, and query the table again:

```bash
docker restart wa-cn2
until mysql --connect-timeout=2 -h127.0.0.1 -P26002 -uroot -p111 -e 'select 1'; do sleep 1; done
mysql -h127.0.0.1 -P26002 -uroot -p111 -e \
  'select count(*),sum(id),min(id),max(id) from workspace_partition_attack.chaos_t'
```

The final query fails with error 20405 because the committed object was deleted.

## Additional information

The source path explains the observed sequence:

- `pkg/txn/client/operator.go`, `txnOperator.doWrite`: after workspace preparation, any commit error other than `moerr.ErrTxnUnknown` calls `workspace.Rollback`. Only `ErrTxnUnknown` calls `FinalizeCommitWithUnknownResult`.
- In this case the RPC layer returns a raw TCP read I/O timeout, so the error does not satisfy `moerr.ErrTxnUnknown` even though TN committed.
- `pkg/vm/engine/disttae/txn.go`, `Transaction.Rollback`: calls `gcObjsByIdxRange(..., cloneGCTxnRollback)` for all workspace writes.
- `pkg/vm/engine/disttae/types.go`, `gcObjsByIdxRange`: collects the CN object names and deletes them asynchronously through `gcFiles`.

The critical invariant is: **an error after a commit request has been sent is not proof of abort**. Transport errors on that phase must enter the unknown-result state, and object GC must be gated by an authoritative abort result.

Suggested solution direction: normalize all post-send commit transport failures into the unknown-result path at the transaction client boundary, then resolve the final transaction status before either finalizing or running rollback GC. Add a deterministic two-CN fault-injection regression that performs a cold read after CN restart; a warm-cache read alone can hide the data loss.

gouhongshen:
## Summary

Fixes #25591.

- Classify every post-dispatch commit RPC failure as `ErrTxnUnknown`, including raw read timeouts, for both single-request and stream sends.
- Preserve prepared workspace files when a commit outcome is unknown, including the frontend's subsequent transaction cleanup rollback.
- Add sender coverage for timeout and stream response loss, and extend the operator unknown-commit test through cleanup rollback.

## Root cause

TN can commit a transaction after CN has prepared workspace files, while the commit response is lost. Raw timeout errors were not consistently classified as an unknown commit result. Even when classified, connection cleanup invoked `TxnOperator.Rollback`, which rolled back the workspace before recognizing that the operator had already closed, deleting files that a committed transaction still referenced.

## Validation

- `go test ./pkg/txn/rpc ./pkg/txn/client -count=1` on 50
- `go vet ./pkg/txn/rpc ./pkg/txn/client` on 50
- `go test -race ./pkg/txn/client -run 'TestCommitKeepsPreparedWorkspaceAfterTxnUnknown$' -count=1` on 50
- `go test -race ./pkg/txn/rpc -run 'Test(CommitResponseTimeoutReturnsTxnUnknown|StreamCommitResponseLossReturnsTxnUnknown)$' -count=1` on 50
- Isolated two-CN chaos regression on 50: after TN committed 20,000 rows and its response to CN1 was dropped, `syncCommit` plus a CN2 restart still cold-read all 20,000 rows (`sum(id)=200010000`), with no `GC-WORKSPACE-FILES` entry.

### DIFF
--- a/pkg/txn/client/operator.go
+++ b/pkg/txn/client/operator.go
@@ -912,7 +912,7 @@ func (tc *txnOperator) Rollback(ctx context.Context) (err error) {
 	txnMeta := tc.getTxnMeta(false)
 	util.LogTxnRollback(tc.logger, txnMeta)
 
-	if tc.reset.workspace != nil && !tc.reset.cannotCleanWorkspace {
+	if tc.reset.workspace != nil && tc.canCleanWorkspace() {
 		if err = tc.reset.workspace.Rollback(ctx); err != nil {
 			tc.logger.Error("rollback workspace failed",
 				util.TxnIDField(txnMeta), zap.Error(err))
@@ -1223,8 +1223,20 @@ func (tc *txnOperator) doWrite(
 	resp, err = tc.trimResponses(tc.handleError(ctx, result, err))
 	if err != nil && commit {
 		tc.reset.commitErr = err
+		if moerr.IsMoErrCode(err, moerr.ErrTxnUnknown) {
+			// The commit may have reached TN. Connection cleanup calls Rollback
+			// after Commit returns, so preserve prepared workspace files until an
+			// authoritative outcome is available.
+			tc.reset.cannotCleanWorkspace = true
+		}
 	}
 	return resp, err
+}
+
+func (tc *txnOperator) canCleanWorkspace() bool {
+	tc.mu.RLock()
+	defer tc.mu.RUnlock()
+	return !tc.reset.cannotCleanWorkspace
 }
 
 func (tc *txnOperator) updateWritePartitions(requests []txn.TxnRequest, locked bool) {
--- a/pkg/txn/client/operator_test.go
+++ b/pkg/txn/client/operator_test.go
@@ -316,6 +316,7 @@ func TestCommitKeepsPreparedWorkspaceAfterTxnUnknown(t *testing.T) {
 
 		err := tc.Commit(ctx)
 		require.True(t, moerr.IsMoErrCode(err, moerr.ErrTxnUnknown))
+		require.NoError(t, tc.Rollback(ctx))
 		require.Equal(t, 1, ws.commitCount)
 		require.Zero(t, ws.finalizeCount)
 		require.Equal(t, 1, ws.unknownCount)
--- a/pkg/txn/rpc/sender.go
+++ b/pkg/txn/rpc/sender.go
@@ -136,6 +136,7 @@ func (s *sender) Send(ctx context.Context, requests []txn.TxnRequest) (*SendResu
 	}
 
 	sr.reset(requests)
+	var commitRequest *txn.TxnRequest
 	for idx := range requests {
 		tn := requests[idx].GetTargetTN()
 		st := sr.getStream(tn.ShardID)
@@ -150,9 +151,15 @@ func (s *sender) Send(ctx context.Context, requests []txn.TxnRequest) (*SendResu
 		}
 
 		requests[idx].RequestID = st.ID()
+		if requests[idx].Method == txn.TxnMethod_Commit {
+			// Once Send is entered, the commit may reach TN even if the
+			// transport reports an error. Its result must be treated as unknown
+			// so the caller does not discard committed workspace files.
+			commitRequest = &requests[idx]
+		}
 		if err := st.Send(ctx, &requests[idx]); err != nil {
 			sr.Release()
-			return nil, err
+			return nil, unknownCommitResult(ctx, commitRequest, err)
 		}
 	}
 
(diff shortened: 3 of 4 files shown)


gouhongshen:
Closing as superseded by #25582, now merged into main. #25582 covers the same commit-response-loss invariant with the complete unknown-finalization lifecycle (atomic rollback-GC fence, retained unknown status, frontend invalidation, and local-resource release). Its documented stream-generation constraint is stricter than this draft's simplified multi-request handling, so rebasing this PR would duplicate or weaken the merged design.

Issue #25612, opened 2026-07-11
[Performance]: Complete special-operator pruning and use low-cost row carriers


XuPeng-SH:
## Is there an existing issue for performance?

- [x] I have checked the existing issues.

## Environment

- Version or commit-id: main at `7821d21c72`
- Hardware parameters: N/A
- OS type: Linux
- Others: logical-plan column pruning in `QueryBuilder.remapAllColRefs`

## Details of Performance

After closing the general `UNION ALL`, `AGG`, and unused-window gaps, several special operators and row-carrier fallbacks still retain avoidable work:

1. `SAMPLE` marks every sample expression as required even if only a subset of sampled outputs is consumed.
2. `TIME_WINDOW` retains every aggregate expression and input even when an outer query discards some aggregate outputs.
3. Zero-column fallbacks choose position/column/aggregate 0 for table/function scans, `UNION ALL`, and scalar aggregates. The first candidate can be a wide varlen column or an expensive aggregate such as `group_concat`, causing avoidable I/O, allocation, and CPU work.

The target is to minimize physical input width and operator state without changing row-count semantics. In particular, `SAMPLE` and scalar aggregation may still need a row-producing carrier when all values are discarded.

## Additional information

Proposed design constraints:

- Model row cardinality as a first-class requirement rather than implicitly retaining slot 0.
- Prefer zero-column batches if all affected execution operators support them; otherwise choose a deterministic low-cost carrier using type width and aggregate cost.
- Compact `SAMPLE`/`TIME_WINDOW` physical slots and remap `HAVING`, special `_wstart`/`_wend` outputs, merge-stage state, and parent projections together.
- Preserve side-effecting expressions and all group/order/partition dependencies.
- Add plan-shape tests, empty-input and 0/1/many-row result tests, wide-varlen performance harnesses, and allocation/peak-memory measurements.

Related architectural work: #14077.

Issue #14077, opened 2024-01-09
[Tech Request]: separate column pruning into a individual optimize rule


badboynt1:
### Is there an existing issue for the same tech request?

- [X] I have checked the existing issues.

### Does this tech request not affect user experience?

- [X] This tech request doesn't affect user experience.

### What would you like to be added ?

```Markdown
一些新的优化器规则依赖列裁剪。
需要将列裁剪独立出来成为一个单独的规则，并且移动到合适的位置
```

PR #25613, opened 2026-07-11
fix: prune unused set, aggregate and window columns

### LINKED ISSUE #25611: [Bug]: Column pruning retains unused UNION ALL, aggregate, and window work

## Branch Name

main

## Commit ID

7821d21c72

- Hardware parameters: N/A
- OS type: Linux
- Others: logical-plan column pruning in `QueryBuilder.remapAllColRefs`

## Actual Behavior

Column pruning misses three classes of removable work:

1. `UNION ALL` retains every aligned output position even when the parent references only a subset. Both branches therefore scan and materialize unused columns.
2. `AGG` retains every aggregate expression and its input columns even when an aggregate result is unused by both the parent and `HAVING`.
3. An unused window result keeps the `WINDOW` and its dedicated `PARTITION` operator reachable, including partition/order-only scan columns.

These gaps increase scan I/O, vector width, aggregate state, sort/partition work, and memory usage. They do not affect set operators with duplicate/set semantics (`UNION DISTINCT`, `INTERSECT`, and `MINUS`), whose complete keys must remain available.

## Expected Behavior

- Compact unreferenced `UNION ALL` positions symmetrically in both children while preserving every input row.
- Remove aggregate expressions not consumed by the parent or `HAVING`, and remap surviving physical aggregate slots correctly.
- Bypass an unused, side-effect-free window and its partition operator.
- Preserve semantic dependencies: all DISTINCT/set/grouping keys, `HAVING` aggregates, window post-filters, scalar-aggregate one-row behavior on empty input, and side-effecting expressions.

## Steps to Reproduce

On commit `7821d21c72`, build plans for:

```sql
select n_name
from (
  select n_name, n_comment from nation
  union all
  select r_name, r_comment from region
) u;

select c_custkey
from (
  select c_custkey, count(c_comment) as unused_count
  from customer
  group by c_custkey
) g;

select n_name
from (
  select n_name,
         row_number() over (partition by n_regionkey order by n_comment) as rn
  from nation
) w;
```

Observe that the first plan scans both columns on each branch, the second scans `c_comment` and computes `count`, and the third retains `WINDOW`/`PARTITION` plus `n_regionkey` and `n_comment`.

## Additional information

Related architectural work: #14077.

The regression harness should verify both scan columns and reachable operator shape, then execute result-based BVT cases for nonzero-slot remapping, duplicate/set semantics, `HAVING`, empty scalar aggregates, stacked windows, and side-effect preservation.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25611

Related to #25612 and #14077.

Column pruning previously retained three categories of unused physical work:

- every aligned `UNION ALL` position on both branches;
- every aggregate expression and input, including results unused by both the parent and `HAVING`;
- an unused window function together with its `PARTITION` operator and partition/order-only columns.

This PR:

- compacts only `UNION ALL` output positions while retaining complete keys for `UNION DISTINCT`, `INTERSECT`, and `MINUS`;
- compacts aggregate expressions and remaps `HAVING` plus parent-visible physical slots;
- preserves one scalar-aggregate carrier so an empty input still produces one row;
- bypasses unused window/partition operators and their exclusive scan inputs;
- retains otherwise-unused windows when `GROUP_CONCAT`, `JSON_ARRAYAGG`, or `JSON_OBJECTAGG` consumes their input order;
- preserves side-effecting expressions such as `sleep()` across all new pruning paths;
- validates malformed `HAVING` relation/slot mappings and returns an internal planner error instead of panicking;
- avoids new allocations on unchanged set/aggregate paths, reuses one bounded slice for the rare order-retention path, and clears stale slice-tail references after in-place compaction.

Correctness coverage includes DISTINCT/set/grouping keys, `HAVING`, empty scalar aggregates, window post-filters, stacked windows, order-sensitive aggregates, nonzero slot compaction, CTE consumers, ORDER BY below LIMIT, and side effects. Plan-shape tests verify that the intended physical operators and columns are actually removed; result-based BVT verifies execution semantics.

## Testing

- `make build`
- `go test ./pkg/sql/plan/... -count=1` with the repository CGo linker flags
- `go vet ./pkg/sql/plan/...` with the repository CGo linker flags
- focused `go test -race` for column-pruning and `HAVING` remapping tests
- mo-tester: `func_arrayagg.sql`, 57/57 passed
- mo-tester: `column_pruning.test`, 21/21 passed

## Review decision log

- DISTINCT/set operators and every grouping key remain unpruned because they determine row multiplicity or membership.
- Input ordering remains observable for MatrixOne's existing order-sensitive aggregate behavior, so an unused window is retained along order-transparent input paths when the aggregate result is consumed.
- Unreachable optimizer-history nodes remain in `Query.Nodes`; execution and shape assertions traverse only `Query.Steps`. Their lifetime is bounded by one plan.
- `SAMPLE`, `TIME_WINDOW`, and cost-based row-carrier selection are intentionally tracked separately in #25612 because their row-count and special-slot contracts require an execution-layer design.

### DIFF
--- a/pkg/sql/plan/explain/column_prune_test.go
+++ b/pkg/sql/plan/explain/column_prune_test.go
@@ -636,6 +636,13 @@ func TestColumnPruneSemanticBoundaries(t *testing.T) {
 			},
 		},
 		{
+			name: "order sensitive aggregate retains input window order",
+			sql:  "select n_regionkey, json_arrayagg(n_name) from (select n_regionkey, n_name, row_number() over (partition by n_regionkey order by n_comment) as rn from nation) w group by n_regionkey",
+			wantTableCol: []Entry[string, []string]{
+				{tableName: "nation", colNames: []string{"n_name", "n_regionkey", "n_comment"}},
+			},
+		},
+		{
 			name: "having keeps only its aggregate input",
 			sql:  "select c_custkey from (select c_custkey, count(c_comment) as cnt, sum(c_acctbal) as total from customer group by c_custkey having sum(c_acctbal) > 0) g",
 			wantTableCol: []Entry[string, []string]{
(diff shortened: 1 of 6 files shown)


XuPeng-SH:
CI failure root cause and fix:

The eight failing BVT/upgrade/coverage jobs all reached the same `func_arrayagg.sql` mismatch. Pruning the unused `ROW_NUMBER()` also removed the Window sort, changing the input sequence observed by `JSON_ARRAYAGG`.

The follow-up commit retains otherwise-unused Window nodes along order-transparent paths when a consumed `GROUP_CONCAT`, `JSON_ARRAYAGG`, or `JSON_OBJECTAGG` depends on that input order. It also covers discarded scalar carriers and stacked windows, and releases temporary planner references on error paths.

Validated locally on the final head:

- full `pkg/sql/plan/...` tests and `go vet`
- focused race tests
- `func_arrayagg.sql`: 57/57
- `column_pruning.test`: 21/21

Issue #25614, opened 2026-07-11
TPCC 10 warehouse performance regression: 25% tpmC drop after commit 522224799 (prepared statement recompile fix)


Ariznawlll:
## Summary

TPCC benchmark shows a **25% performance regression** in tpmC for 10-warehouse workload after commit [`522224799`](https://github.com/matrixorigin/matrixone/commit/5222247997ce2bfa9115bc7d778e4f80664bce09).

## Bisect Result

Binary search successfully identified the first bad commit:

| Field | Value |
|---|---|
| **First bad commit** | [`522224799`](https://github.com/matrixorigin/matrixone/commit/5222247997ce2bfa9115bc7d778e4f80664bce09) |
| **Last good commit** | [`3a8d66954`](https://github.com/matrixorigin/matrixone/commit/3a8d66954014b9de54e4736d32996c261bb1b298) |
| **Test** | TPCC: warehouse=10, terminals=10 |
| **Baseline tpmC** | 6097.17 |
| **Bad commit tpmC** | 4568.1 |
| **Regression** | **-25.08%** |
| **Threshold** | 25% |

### Iteration History

| Iteration | Commit | Verdict | tpmC | Delta vs Baseline |
|---|---|---|---:|---:|
| suspect | efe237966 | bad | 4261.7 | -30.10% |
| 1 | 25abedd7c | bad | 4405.96 | -27.74% |
| 2 | 0fcd615a7 | good | 6041.19 | -0.92% |
| 3 | 3a8d66954 | good | 6047.22 | -0.82% |
| 4 | 522224799 | bad | 4483.84 | -26.46% |
| 5 | 522224799 | bad | 4568.1 | -25.08% (confirmed) |

### Bad Commit Analysis

**Commit**: [`522224799`](https://github.com/matrixorigin/matrixone/commit/5222247997ce2bfa9115bc7d778e4f80664bce09)  
**PR**: [#25562](https://github.com/matrixorigin/matrixone/pull/25562)  
**Author**: Cao Kai  
**Date**: 2026-07-10 10:06:39 UTC  
**Message**: `fix(frontend): force recompile on each prepared execute to prevent hang`

**Original Fix**: Fixed prepared UPDATE...JOIN hanging on second execution through binary protocol (COM_STMT_EXECUTE). The fix forces recompilation of the scope tree on every prepared statement execution to avoid stale state in hashbuild operators.

[...]

PR #25623, opened 2026-07-11
fix(frontend/compile): reset PipelineEdge terminal state on compiled scope reuse to fix TPCC regression

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25614

Issue #25614: prepared UPDATE...JOIN hang fix (commit 522224799 / PR #25562) forced per-execution recompilation of the compiled scope tree, causing a 25% TPCC tpmC regression.

**Root cause of the hang**: not in hashbuild operator state but in PipelineEdge (WaitRegister) persistent terminal state (doneClosed/endDelivered/fatalTerminal) surviving into the next execution when the cached compile was reused — data and End signals were silently rejected by the done edge, leaving merge receivers waiting forever.

**Fix**: instead of recompiling on every execution (the band-aid from #25562), reset the pipeline edge terminal state so the cached compile can safely be reused:

1. Add `PipelineEdge.ResetTerminalStateForReuse()` that drains buffered stale signals and clears the terminal state while preserving the channel buffer and NilBatchCnt.
2. Call `ResetTerminalStateForReuse` on all MergeReceivers in `Scope.resetForReuse`, which runs via `Compile.Reset` on the prepared compile reuse path.
3. Restore the original schema-change-only recompilation in `initExecuteStmtParam` (revert the always-recompile logic from #25562).

**Verification**:
- Issue #25526 reproducer (binary protocol prepared UPDATE...JOIN): both executions complete with correct updated rows (no hang).
- `TestIssue25526PreparedUpdateJoinSecondExecute` (embedded cluster): PASS
- mo-tester prepare/ BVT: 1021/1021 pass (100%)
- PipelineEdge + Scope + frontend UT all pass (incl. `-race`)
- `git diff --check` + `make static-check`: 0 issues

### DIFF
--- a/pkg/frontend/computation_wrapper.go
+++ b/pkg/frontend/computation_wrapper.go
@@ -476,14 +476,16 @@ func initExecuteStmtParam(execCtx *ExecCtx, ses *Session, cwft *TxnComputationWr
 		prepareStmt.Ts = timestamp.Timestamp{PhysicalTime: time.Now().Unix()}
 	}
 
-	// Recreate the cached compile on each execution to avoid stale operator state
-	// (e.g. hashbuild ctr, dispatch channels) from previous executions when the
-	// plan is reused. A nil cache means the statement is not eligible for
-	// prepare-time compile (e.g. AP query); recompiling would fail with
-	// ErrCantCompileForPrepare on every execution, so leave it to the regular
-	// compile path (isPrepare=false).
-	// See: https://github.com/matrixorigin/matrixone/issues/25526
-	if prepareStmt.compile != nil {
+	// Recreate the cached compile only when the schema changed. Without a
+	// schema change the cached compile is reused as-is: Compile.Reset clears
+	// the per-execution state, including the pipeline edges' terminal state
+	// (see Scope.resetForReuse), so reuse is safe and avoids the
+	// per-execution recompilation overhead that regressed TPCC. A nil cache
+	// means the statement is not eligible for prepare-time compile (e.g. AP
+	// query); recompiling would fail with ErrCantCompileForPrepare on every
+	// execution, so leave it to the regular compile path (isPrepare=false).
+	// See: https://github.com/matrixorigin/matrixone/issues/25614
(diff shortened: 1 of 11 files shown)


XuPeng-SH (approved):
Reviewed the latest head from the prepared-compile, pipeline-protocol, resource-lifecycle, unhappy-path, and performance perspectives. I found no blocking issue.

The fix is placed at the correct lifecycle boundary: the previous execution's pipeline cleanup and MergeRun wait for their sender/receiver goroutines before the next COM_STMT_EXECUTE reaches Compile.Reset, so ResetTerminalStateForReuse does not race with live pipeline users. Scope.Reset walks the full cached scope tree and resets every MergeReceiver while preserving the channel and NilBatchCnt. The edge reset clears all terminal/error/abort counters and drains stale signals; direct batches are cleaned and stale spool references are released through the spool lifecycle.

I also traced the originally suspected hashbuild/runtime-filter path. HashBuild.Reset restores state/runtimeFilterIn/hashmapBuilder, Compile.Reset replaces the MessageBoard, and table-scan runtime-filter executors are recreated per execution. The observed hashbuild stall is consistent with the already-terminal edge rejecting second-execution data/End, rather than an additional persistent hashbuild state bug.

No new unbounded wait or accumulation is introduced. Reset cost is O(number of cached edges + buffered stale signals) once per prepared execution and removes full physical recompilation, with no row-processing hot-path change. Schema-change recompilation and nil-cache/AP behavior remain intact.

Local verification on this exact head passed:
- full pkg/vm/process tests
- full pkg/sql/compile tests
- full pkg/frontend tests
- PipelineEdge reset/drain tests under -race
- repeated fatal-terminal + two-sender reuse fault injection under -race (100 runs)
- TestIssue25526PreparedUpdateJoinSecondExecute (real binary protocol)
- git diff --check

The repository-wide CI and benchmark jobs are still in progress, but the code review itself has no blocker.

PR #25634, opened 2026-07-11
fix(disttae): strengthen reusable relation handle reset coverage and correctness

### LINKED ISSUE #25627: [Refactoring]: Separate transaction-scoped relation state from reusable compiled operators
## Why do you want to refactor this code?

Prepared-statement compile reuse exposed an ownership and lifecycle mismatch between reusable execution plans and transaction-scoped engine relations.

A compiled `Scope` and its operators can be reused across executions, but several operator containers retain `engine.Relation` handles across executions. At the same time, `Transaction.tableCache` can return the same underlying `*txnTable` to multiple operators, and `txnTable.Reset` mutates shared transaction-specific fields:

```go
tbl.db.op = op
tbl.proc.Store(txn.proc)
tbl.createdInTxn = false
tbl.lastTS = op.SnapshotTS()
```

This caused the Ubuntu UT race detector to report concurrent writes from `Insert.Prepare` and `Deletion.Prepare` during the second execution of a prepared `UPDATE ... JOIN`. The failure was observed in `TestIssue25526PreparedUpdateJoinSecondExecute` in Actions job 86539533869 for PR #25623.

The immediate race has been mitigated in PR #25623 by clearing `Insert.ctr.source` and `Deletion.ctr.source` from operator `Reset`, so those operators reacquire relations for the next execution. However, the underlying contract remains unclear and similar relation reuse still exists in other paths, including:

- `MultiUpdate.Prepare`
- `LockOp.Prepare`
- `LockMeta.initLockExe`

The current design combines three incompatible properties:

1. compiled operators are long-lived and reusable;
2. relation handles contain mutable transaction-specific state;
3. the same relation object can be shared by concurrently prepared or running operators.

The existing comment above `txnTable.Reset` already notes: `TODO: txnTable should be stateless`.

## Describe the solution you would like

Define and enforce explicit relation ownership and lifetime rules:

1. Treat relation handles stored in execution operators as execution/transaction scoped, not compile scoped.
2. Keep reusable compiled operators limited to immutable plan metadata and object references.
3. Reacquire or create per-execution relation handles when an execution begins.
4. Make shared `txnTable` state immutable/stateless where practical, or split immutable table metadata from a transaction-bound relation wrapper.
5. Audit all `engine.Relation.Reset` call sites for concurrent reuse and stale-transaction risks.
6. Add race-enabled lifecycle tests covering two operators targeting the same table across repeated prepared executions and retry paths.
7. Document whether `Relation.Reset` requires exclusive ownership and whether it may overlap with `Write`, `Delete`, locking, or reader construction.

Acceptance criteria:

- No reusable operator retains a relation handle from a previous transaction unless the handle has a documented exclusive-ownership contract.
- Shared table-cache objects do not store unsynchronized caller-specific transaction state.
- All remaining `Relation.Reset` call sites are proven safe or replaced.
- Prepared DML reuse tests pass under `-race`.
- The fix preserves the compile-reuse performance improvement tracked by #25614.

## Describe alternatives you have considered

- Clear relation fields in each operator `Reset`: appropriate as a short-term mitigation, but easy to miss when new operators or relation fields are added.
- Add a mutex only inside `txnTable.Reset`: insufficient because one scope can proceed into `Write` or `Delete` while another scope is still resetting the shared object.
- Serialize all scope preparation: avoids the race at a broad execution-performance cost.
- Recompile every prepared execution: avoids stale runtime state but reintroduces the 25% TPCC regression tracked by #25614.

## Additional information

Related items:

- Original prepared UPDATE JOIN hang: #25526
- Compile-reuse performance regression: #25614
- Fix under development: #25623
- Failing CI job: https://github.com/matrixorigin/matrixone/actions/runs/29150642450/job/86539533869
- Race location: `pkg/vm/engine/disttae/txn_table.go`, `txnTable.Reset`
- Immediate mitigation commit in PR #25623: `3c99de345ba29fd5cfa2885959ffe06d39f8b565`

XuPeng-SH:
## Why

This PR addresses correctness and unhappy-path coverage gaps in relation-handle reset behavior under reusable relation lifecycle, especially during cache misses and DDL race cases around table existence and deletion.

## What changed

- Add missing unhappy-path tests for `reusableRelationHandle.Reset` in `pkg/vm/engine/disttae/txn_table_test.go`.
  - `TestReusableRelationHandleResetFromCatalogCacheMiss` covers `tableOps` and `tableCache` misses and verifies rebind via catalog path.
  - `TestReusableRelationHandleResetFromCatalogMissingTable` covers catalog lookup miss after lookup and returns `NoSuchTable`.
  - `TestReusableRelationHandleResetRejectsDeletedTable` covers `existAndDeleted` returning true after DDL drop and returns `NoSuchTable`.
  - `combinedTxnTable.Reset` behavior and shared reset semantics are explicitly validated.
  - Added test helpers `newResetTxnForTest` and `insertCatalogTableForResetTest` for realistic catalog-backed reset scenarios.

## How it fixes the issue

- Avoids blind reliance on cache-hit paths by exercising catalog-rebind and failure paths that were previously untested.
- Makes table-drop race behavior explicit and deterministic in tests.
- Confirms shared/combined reset behavior remains constrained while operator-owned reusable handles rebind correctly.

## Performance

- No new hot-path overhead expected for cache-hit reset path.
- Cache-miss reset remains an exceptional path and can involve canonical reconstruction as designed.

## Related
- Fixes: #25627

### DIFF
--- a/pkg/sql/colexec/constraint_util.go
+++ b/pkg/sql/colexec/constraint_util.go
@@ -139,7 +139,7 @@ func getRelationByObjRef(ctx context.Context, proc *process.Process, eg engine.E
 		return nil, moerr.NewNoSuchTable(ctx, ref.SchemaName, objName)
 	}
 
-	return relation, nil
+	return engine.NewRelationHandle(relation), nil
 }
 
 func GetRelAndPartitionRelsByObjRef(
--- a/pkg/sql/colexec/deletion/deletion.go
+++ b/pkg/sql/colexec/deletion/deletion.go
@@ -73,11 +73,8 @@ func (deletion *Deletion) Prepare(proc *process.Process) error {
 				return err
 			}
 			deletion.ctr.source = rel
-		} else {
-			err := deletion.ctr.source.Reset(proc.GetTxnOperator())
-			if err != nil {
-				return err
-			}
+		} else if err := deletion.ctr.source.Reset(proc.GetTxnOperator()); err != nil {
+			return err
 		}
 
 	}
--- a/pkg/sql/colexec/deletion/deletion_partition_test.go
+++ b/pkg/sql/colexec/deletion/deletion_partition_test.go
@@ -32,7 +32,7 @@ func TestPartitionDeleteCallDoesNotMutateSharedObjectRef(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	proc, eng := prepareDeletionTest(t, ctrl)
+	proc, eng, _, _ := prepareDeletionTest(t, ctrl)
 
 	raw := &Deletion{
 		DeleteCtx: &DeleteCtx{
--- a/pkg/sql/colexec/deletion/deletion_test.go
+++ b/pkg/sql/colexec/deletion/deletion_test.go
@@ -17,6 +17,7 @@ package deletion
 import (
 	"bytes"
 	"context"
+	"errors"
 	"testing"
 	"time"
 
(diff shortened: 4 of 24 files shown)


XuPeng-SH (commented):
Reviewed the full relation lifecycle and all changed call sites. I found no blocking correctness, concurrency, lifecycle, compatibility, or performance issue.

Verified closures:
- shared disttae canonical relation -> exclusive reusable handle -> per-execution Reset/rebind
- tableCache hit, catalog cache miss, txn-local DROP, txn-local CREATE, and non-disttae workspace failure paths
- table scan, insert, delete, lock, multi-update, and LockMeta retained-relation call sites
- partial Prepare/initialization failures preserve a retryable state

Local verification:
- targeted affected-package tests passed
- reusable relation reset tests passed under `-race -count=5`
- changed operator/compile tests passed under `-race -count=3`

The cache-hit reset path remains allocation-free according to the added benchmark assertion; catalog reconstruction is confined to the cache-miss path. GitHub does not allow approving one's own PR, so this is recorded as a non-blocking review comment.

PR #25640, opened 2026-07-12
fix: close disk object storage readers on validation errors

### LINKED ISSUE #25633: [Bug]: diskObjectStorage.Read leaks a file descriptor on invalid ranges
## Summary

`diskObjectStorage.Read` leaks the file descriptor opened for an object whenever range validation fails after `os.Open`.

The deterministic case is an empty range (`max <= min`): the method returns `ErrEmptyRange` and no reader, but never closes the opened file. Seek errors and unexpected seek positions have the same cleanup problem.

## Version

Commit: `7f7de555c`

## Code path

`pkg/fileservice/disk_object_storage.go`:

```go
f, err := os.Open(path)
// ...
if min != nil {
    pos, err := f.Seek(*min, io.SeekStart)
    if err != nil {
        return nil, err
    }
    if pos != *min {
        return nil, moerr.NewEmptyRangeNoCtx(key)
    }
}

if max != nil {
    limit := *max
    if min != nil {
        limit -= *min
    }
    if limit <= 0 {
        return nil, moerr.NewEmptyRangeNoCtx(key)
    }
}
```

Only the successful return wraps `f` in a `readCloser`. All post-open error returns bypass `f.Close()`.

## White-box reproduction

1. Create a disk object containing `data`.
2. Disable GC/finalizers.
3. Count the process's descriptors using `fcntl(F_GETFD)`.
4. Call `Read(ctx, "object", &min, &max)` 20 times with `min=1`, `max=1`.
5. Verify every call returns an error and a nil reader, then count descriptors again.

## Actual result

```text
open descriptors before=4 after=24 delta=20
```

Each rejected range leaks exactly one descriptor.

## Expected result

Every path after `os.Open` must either transfer ownership to the returned `ReadCloser` or close the file before returning. Repeated invalid range requests must not change the process's open-descriptor count.

## Distinction from existing issue

#25603 covers failed `diskObjectStorage.Write` temporary files and swallowed `Delete` errors. This issue is an independent descriptor leak in the `Read` range-validation path.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

Fixes #25633

`diskObjectStorage.Read` opened a local object file before validating the requested range, but only transferred ownership to a returned `readCloser` on success. Any post-open validation failure therefore leaked the file descriptor until an eventual GC finalizer.

This PR defines one shared cleanup contract for the two file-backed object-storage readers, Disk and HDFS:

- after an open, an error return closes the underlying reader exactly once;
- a successful return transfers ownership exclusively to the returned `readCloser`;
- a successful cleanup returns the primary error object unchanged, preserving MatrixOne error-code classification;
- if both the operation and `Close` fail, the returned error joins both causes with the operation error first.

Using the same helper for HDFS removes its pre-existing successful-close error wrapping as well, so the two implementations now have identical ownership and error semantics. The refactor does not alter range validation, normal reads, metrics, semaphore release, or `fadviseDontNeed`.

The Unix regression test disables GC, repeatedly exercises equal bounds, inverted bounds, a non-positive upper bound, and a negative seek offset, and verifies both the expected MatrixOne error code and that the process FD count is unchanged. The platform-neutral helper test covers ownership transfer, successful cleanup with exact error identity, and a simultaneous operation/close failure. The descriptor test is intentionally limited to Linux and Darwin because it uses `fcntl(F_GETFD)` to observe descriptor ownership.

Validation:

- `go test -count=20 ./pkg/fileservice -run "^(TestCloseOnError|TestDiskObjectStorageReadInvalidRangeClosesFile)$"`
- `go test -race -count=10 ./pkg/fileservice -run "^(TestCloseOnError|TestDiskObjectStorageReadInvalidRangeClosesFile)$"`
- `go test ./pkg/fileservice`
- `go vet ./pkg/fileservice`
- `git diff --check`

The commands were run with the repository CGo include/library flags on macOS. A `CGO_ENABLED=0` Linux cross-build is not applicable to this repository because the existing `mpool` package requires MatrixOne's C allocator.


LeftHandCold (approved):
Reviewed current head `968c5b1584`. No blocking issue found.

Q1 resource lifecycle:
- open failure returns before ownership exists;
- every post-open error reaches deferred `closeOnError` exactly once;
- successful cleanup preserves the primary error object, while simultaneous operation/Close failures retain both causes in operation-first order;
- successful Read leaves the deferred cleanup inactive and transfers ownership to `readCloser`; Disk terminates at `fadviseDontNeed + os.File.Close`, HDFS at `FileReader.Close`;
- semaphore/metrics/trace wrappers propagate errors and preserve downstream Close/release paths.

Q2: no new wait/blocking lifecycle is introduced.
Q3: no new unbounded accumulation is introduced.

Local validation:
- `git diff --check`
- targeted tests `count=20`
- targeted race tests `count=10`
- full `go test ./pkg/fileservice` with repository CGo flags
- `go vet ./pkg/fileservice`

All local checks passed. GitHub CI was still in progress at review time.

PR #25641, opened 2026-07-12
fix(plan): harden folded IN vectors for composite keys

### LINKED ISSUE #25624: [Bug]: Composite-key filter rewrite panics on folded IN Expr_Vec

- [x] I searched existing issues and pull requests for `Expr_Vec`, `Expr_List`, `doMergeFiltersOnCompositeKey`, and compound/prefix IN.

### Branch Name

main

### Commit ID

### Actual Behavior

The composite-primary-key filter rewriter panics when an `IN` predicate has already been constant-folded from `Expr_List` to `Expr_Vec`.

`doMergeFiltersOnCompositeKey` recognizes the predicate as `in`, but later unconditionally dereferences the list representation:

```go
inArgs := make([]*plan.Expr, len(lastFilter.GetF().Args[1].GetList().List))
```

For a valid folded vector expression, `GetList()` is nil. The result is:

```text
github.com/matrixorigin/matrixone/pkg/sql/plan.(*QueryBuilder).doMergeFiltersOnCompositeKey
    pkg/sql/plan/expr_opt.go:1597
```

This is relevant to repeated optimization/prepared-plan paths because the same logical `IN` predicate can be represented as `Expr_List` before folding and `Expr_Vec` after folding.

### Expected Behavior

Composite-key rewriting must support both legal RHS representations:

- `Expr_List`: build serialized compound values from its expression elements.
- `Expr_Vec`: decode/copy the folded vector and build the equivalent compound `in` / `prefix_in` predicate, or conservatively keep the original filters.

It must never panic or mutate planner-owned folded-vector bytes.

### White-box Reproduction

Add this regression in `pkg/sql/plan`:

```go
func TestCompositeRewriteAcceptsFoldedInVector(t *testing.T) {
    ctx := NewMockCompilerContext(true)
    builder := NewQueryBuilder(planpb.Query_SELECT, ctx, false, false)
    tag := builder.genNewBindTag()
    tableDef := makeExprOptCompositeSortKeyTableDef()

    aEq := makeExprOptBinaryInt64Expr(
        t, ctx, "=", makeExprOptInt64Col(tag, 0, "a"), 1)
    bIn, err := BindFuncExprImplByPlanExpr(ctx.GetContext(), "in", []*planpb.Expr{
        makeExprOptInt64Col(tag, 1, "b"),
        MakePlan2Int64VecExprWithType(ctx.GetProcess().Mp(), 2, 3),
    })
    require.NoError(t, err)

    require.NotPanics(t, func() {
        ret := builder.doMergeFiltersOnCompositeKey(tableDef, tag, aEq, bIn)
        require.NotEmpty(t, ret)
    })
}
```

Run:

```text
go test ./pkg/sql/plan -run TestCompositeRewriteAcceptsFoldedInVector -count=1 -v
```

The test deterministically panics at `expr_opt.go:1597`.

### External-path observation

A normal single-pass literal query can still succeed because its optimizer phase ordering may keep/normalize the list before this branch. For example, `a=1 AND b IN (1,2)` returned correct rows in a direct smoke test, although `EXPLAIN` displayed the generated compound filter as `__mo_cpkey_col in ()`.

This issue is specifically about the already-folded internal plan shape used by repeated/reoptimization paths, not a claim that every literal composite-IN query currently panics.

### Acceptance Criteria

1. Cover `Expr_List` and `Expr_Vec` RHS forms for full and prefix composite keys.
2. Cover equality-prefix + IN, compound cluster keys, and prepared/repeated optimization.
3. Ensure repeated rewrite is idempotent and does not generate duplicate compound predicates.
4. Preserve original component predicates needed for per-column pruning.
5. Do not mutate or retain unsafe references to planner-owned serialized vector data.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25624

Composite-key predicate merging assumed every IN RHS remained an `Expr_List`. Prepared/re-optimized plans can fold the list into an `Expr_Vec`, which previously caused a nil-list panic.

This change decodes valid folded vectors into typed literals before composing serialized composite keys. It validates the encoded vector length, physical backing storage, and the embedded vector type against the compared column before replacing component predicates. Invalid, empty, unsupported, or inconsistent vectors now preserve the original predicate rather than producing a wrong compound key or panicking.

The composite-key rewrite now also uses one transactional failure rule across IN, BETWEEN, IN_RANGE, range, and equality branches: only replace original filters after all serial expressions and the final predicate bind successfully. Malformed function shapes are ignored by the optional optimization.

Coverage includes the real `ConstantFold(List -> Vec)` path, full and prefix composite primary keys, composite cluster keys, OR merge, NULL/empty/constant vectors, strings, decimal scale metadata, corrupted payloads, type-mismatched payloads, block-filter retention, repeated optimization, race execution, and malformed predicate shapes.

Validation:

- `go test -count=1 -timeout 180s ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `go test -race ... ./pkg/sql/plan` for the changed paths
- focused tests with `-shuffle=on -count=5`
- benchmarked List vs folded-Vec composite rewrite and block-filter dedup; no relative folded-Vec regression observed.


gouhongshen:
Review complete: no actionable findings. The folded Expr_Vec path now validates its payload, materializes independent literals for composite-key serialization, and retains original predicates on invalid input or binding failure. Local pkg/sql/plan tests could not start because this environment lacks roaring.h, xxhash.h, and usearch.h; GitHub UT and coverage checks were still pending at review time.

gouhongshen (approved):
Approved based on the review above.

gouhongshen (changes requested):
Blocking: the malformed-predicate hardening is incomplete. In pkg/sql/plan/expr_opt.go:2171, the pre-merge range scan bypasses hasNonNilFunctionArgs. A binary > or < predicate with Args: []*plan.Expr{nil, constant} reaches classifyRangeBound; canonicalRangeOp then calls isRuntimeConstExpr(nil), which dereferences expr.Expr in apply_indices.go:157. Please reject nil operands before the second range scan (or make the shared range helpers nil-safe), and add a regression for nil operands, not only short argument lists.

gouhongshen:
Blocking: malformed OR shapes still panic in the new retention path. For an `or` function with a single non-mergeable arm (for example, `Args: []*plan.Expr{MakePlan2BoolConstExprWithType(true)}`), flattening produces that one arm; it is appended to `newOrArgs`, while `firstEquiExpr` remains nil because no equality/IN candidate was found. The `len(newOrArgs) == 1` branch at `expr_opt.go:2136` then dereferences `firstEquiExpr.GetF()` and panics. Please only derive/update `col2filter` when a merge candidate exists (or otherwise return the retained arm without that dereference), and add a unary/non-mergeable OR regression. This is separate from the already-reported nil range-operand panic.

gouhongshen (approved):

PR #25642, opened 2026-07-12
fix: clean plan cache and stabilize scope test

### LINKED ISSUE #25597: [Bug] planCache.clean() loops forever when cache is non-empty
## Summary

`planCache.clean()` can loop forever whenever the LRU list is non-empty. The cleanup loop frees the front statement but never removes the list element, so `Len()` never decreases.

## Evidence

- Commit: `7821d21c723372e45a8f65da3bbb896a157b1036`
- Code: `pkg/frontend/plan_cache.go:94-105`
- The existing `Test_CleanCache` populates the cache and then calls `clean`; with the current loop this does not terminate.
- Session DDL/rollback paths call cache cleanup, so a populated cache can hang the session after DDL or rollback.

## Expected

Cleanup should remove each evicted LRU entry and return.

## Actual

The same front entry is processed repeatedly and the loop never terminates.

Please confirm and add a regression test that asserts the LRU length reaches zero.

### LINKED ISSUE #25628: [Bug]: TestScopeGetRelDataError panics because catalog_defines is not initialized
## Description

The existing unit test `TestScopeGetRelDataError` does not reach the error path it is intended to verify. It deterministically panics while constructing the test engine because the default service runtime has no `catalog_defines` value.

This makes the test fail when it is executed directly or selected by a focused `go test -run` invocation.

## Version

Commit: `7f7de555c`

## Reproduction

From the repository root, with the normal MatrixOne CGO environment:

```bash
go test ./pkg/sql/compile -run '^TestScopeGetRelDataError$' -count=1
```

## Actual result

The test panics before calling `Scope.getRelData`:

```text
--- FAIL: TestScopeGetRelDataError (0.00s)
panic: catalog_defines is not set:

github.com/matrixorigin/matrixone/pkg/catalog.GetDefines(...)
    pkg/catalog/catalog.go:57
github.com/matrixorigin/matrixone/pkg/txn/storage/memorystorage.NewCatalogHandler(...)
    pkg/txn/storage/memorystorage/catalog_handler.go:117
github.com/matrixorigin/matrixone/pkg/txn/storage/memorystorage.NewMemoryStorage(...)
    pkg/txn/storage/memorystorage/mem_storage.go:31
github.com/matrixorigin/matrixone/pkg/testutil/testengine.New(...)
    pkg/testutil/testengine/testengine.go:61
github.com/matrixorigin/matrixone/pkg/sql/compile.TestScopeGetRelDataError(...)
    pkg/sql/compile/scope_test.go:782
```

## Expected result

The test engine should be initialized successfully, `s.getRelData(c, nil)` should run, and the test should assert the intended returned error through `require.Error(t, err)` without a panic.

## Root cause

The test currently creates the engine with:

```go
e, _, _ := testengine.New(
    defines.AttachAccountId(context.Background(), catalog.System_Account),
)
```

`testengine.New` creates a `memorystorage.MemoryStorage` for service ID `""`. Its catalog handler calls `catalog.GetDefines("")`, which panics unless `catalog.SetupDefines("")` has initialized the service runtime.

Other tests using the same memory engine explicitly call `catalog.SetupDefines("")`, but `TestScopeGetRelDataError` does not.

## Suggested fix

Initialize catalog definitions before `testengine.New`, or make the shared test-engine setup guarantee the required runtime state. The test should also check the error returned by `testengine.New` instead of discarding all returned errors.

Please retain a focused invocation of this test in verification so it cannot depend on execution order or unrelated suite-wide initialization.

XuPeng-SH:
## Summary

- initialize `catalog_defines` before `TestScopeGetRelDataError` creates its memory test engine
- remove each plan-cache LRU element during cleanup so non-empty caches terminate
- release cached AST statements when a session closes
- add regressions for LRU removal, exactly-once statement release, multi-statement entries, and the session-close release path

Fixes #25597
Fixes #25628

## Testing

- `go test -race -count=1 -timeout 120s ./pkg/frontend -run '^(Test_CleanCache|Test_CleanCacheRemovesEntriesAndFreesStatements|TestSessionReleasePlanCache)$'`
- `go build ./pkg/frontend`
- `go vet ./pkg/frontend`
- `go test -count=1 -timeout 120s ./pkg/frontend`
- `go test -count=1 -timeout 120s ./pkg/sql/compile -run '^TestScopeGetRelDataError$'`


gouhongshen (approved):
Approved.

Non-blocking follow-up: `TestSessionReleasePlanCache` invokes `releasePlanCache` directly, so it does not verify the `Session.Close` integration point. Please add a close-path regression test when practical; the implementation itself is correct.

PR #25645, opened 2026-07-12
fix: harden logtail push lifecycle

### LINKED ISSUE #25644: [Bug]: Logtail push sessions can leak or block publication on disconnect and cancellation
## Branch Name
fix/tn-impl-lifecycle

## Commit ID
730ce216a

## Actual Behavior
The TN logtail push path has several lifecycle failures under ordinary unhappy paths: an idle disconnected client session is retained until a later data write, a canceled subscription can publish a stale response, a slow session can block global incremental publication, and queued events can lose their release callback during shutdown.

## Expected Behavior
Disconnects, cancellations, queue saturation, and shutdown must not leak sessions or batches, block healthy subscribers or the publish frontier, or emit stale subscription responses.

## Steps to Reproduce
1. Open a logtail stream and subscribe a table.
2. Close the client stream while the table is idle.
3. Observe the TN session remains without an explicit probe.

Additional deterministic reproductions cover cancellation during phase 1/2, a full session send queue, a server close with queued events, and startup before the first incremental event.

## Additional information
The fix adds deterministic in-memory and real morpc integration coverage, generation-based subscription completion, bounded slow-session isolation, queued-event draining, bootstrap waterline initialization, and low-frequency idle probing.

XuPeng-SH:
## What type of PR is this?

- [x] BUG
- [x] Improvement
- [x] Test and CI

Fixes #25644

Hardens the TN logtail push lifecycle under normal unhappy-path conditions.

- Makes subscription completion generation-aware so canceled or superseded phase results cannot publish stale responses.
- Honors caller cancellation in the logtail client request queue and in disttae readiness waiting.
- Isolates a congested session from global incremental publication.
- Releases queued logtail batches during server shutdown and establishes a bootstrap waterline before the first incremental event.
- Detects and removes idle disconnected morpc sessions with a low-frequency probe.
- Adds deterministic in-memory, real morpc integration, concurrent lifecycle, shutdown, ownership, bootstrap, boundary, and race coverage.

## Validation

- `go test -race ./pkg/vm/engine/tae/logtail/service ./pkg/vm/engine/tae/logtail ./pkg/vm/engine/disttae`
- Repeated race coverage for lifecycle and generation tests.
- `git diff --check`


XuPeng-SH (commented):
Re-reviewed the latest head `965e9c43f6` with a broader unhappy-path pass, not just the previously discussed ownership threads. The session/server/notifier closeCB ownership chain is now sealed on every enqueue/release path, phase-1 subscription tails and queued notifier/session responses are drained on close, `waitReady` is context-terminable, and the `safeQueue` / `TxnManager` / `WaitWalAndTail` lifecycle gaps are closed coherently. I re-ran the targeted package suites and race suites for `pkg/vm/engine/tae/logtail/service`, `pkg/vm/engine/disttae`, `pkg/vm/engine/tae/logstore/sm`, `pkg/vm/engine/tae/txn/txnbase`, and `pkg/vm/engine/tae/txn/txnimpl`, and I do not see a remaining high-confidence Q1/Q2/Q3 or correctness blocker on this head. GitHub does not allow me to APPROVE my own pull request, so I am leaving this as an approve-level review comment instead of an approval.

XuPeng-SH (commented):
Re-reviewed the latest head `be0fc6d627`. The only PR-local delta after my last pass is the mock logtail session lifecycle fix in `pkg/vm/engine/test/testutil`: `testClientSession` now has a real session context rooted in the transport context, `Close()` cancels it, and `Write()` terminates on either the caller context or the session/transport shutdown. That closes the remaining test-fixture gap where a mock session could keep looking alive and keep a send path blocked after transport teardown, so the unhappy-path tests now model the production morpc/session lifetime much more faithfully.

I reran the focused packages on the latest head (`./pkg/vm/engine/test/testutil`, `./pkg/vm/engine/tae/logtail/service`, `./pkg/vm/engine/disttae`) plus race for `./pkg/vm/engine/test/testutil` and `./pkg/vm/engine/tae/logtail/service`, and I do not see a new high-confidence correctness or Q1/Q2/Q3 blocker on this head. GitHub still will not let me APPROVE my own pull request, so I am leaving this as another approve-level review comment.

pkg/vm/engine/tae/logtail/service/session.go
```diff
@@ -357,59 +416,106 @@ func NewSession(
...
+func (ss *Session) sendHeartbeat() error {
+	ss.publishMu.Lock()
+	from := ss.exactFrom
+	ss.publishMu.Unlock()
+	resp := ss.responses.Acquire()
+	resp.Response = newUpdateResponse(from, from)
+	defer ss.responses.Release(resp)
+	ctx, cancel := context.WithTimeout(ss.sessionCtx, ss.sendTimeout)
+	defer cancel()
+	return ss.stream.write(ctx, resp)
```
gouhongshen, line 528: [P1] Serialize transport probes behind queued responses

Publish enqueues an update containing To=to and then advances ss.exactFrom (lines 592-597). If that response is still in sendChan when this timer fires, the sender's select has both timer.C and sendChan ready and may execute this direct stream.write first. The empty heartbeat is then put on the wire before the logtail response it summarizes.

The disttae client treats an empty update as progress: dispatchUpdateResponse immediately updates receivedLogTailTime from response.To and queues only time-update commands for routines. A transaction/read can therefore observe T before the queued logtail mutations through T are applied; the pending-update guard has not seen those tables yet. Please route probes through sendChan, or otherwise guarantee that a probe cannot overtake queued responses, and add a deterministic regression test for the ordering.

gouhongshen (changes requested):
Review result: REQUEST_CHANGES. I reviewed the lifecycle, ordering, cancellation, callback ownership, slow-session isolation, heartbeat/bootstrap protocol, and regression tests. I found one P1 blocker below. The issue is a protocol-ordering violation under a congested session, so this cannot be approved as-is. Validation: git diff --check passed; the focused logstore/sm unit and race tests passed. The service/disttae package tests could not compile in this checkout because the environment lacks required cgo headers (usearch.h/roaring.h/xxhash.h).

pkg/vm/engine/tae/logtail/service/session.go
XuPeng-SH: Fixed in 0e1bfe33e6.

The sender now gives an already queued response priority when the transport-probe timer fires. It also maintains a separate sentThrough frontier that advances only after a complete transport write succeeds; probes never use the newer enqueue frontier. Both update and subscription responses are included because the disttae client treats both as timestamp progress. Failed writes leave sentThrough unchanged and still release response ownership.

Added deterministic tests for queued-update-before-probe ordering, enqueue-vs-sent frontier separation, subscription progress, failed-write ownership/frontier behavior, repeated execution, and race. The full logtail service package, full service race suite, build/vet, and the dependent disttae progress/pending-update tests pass.

aptend:
Re-reviewed the latest head `d7f63f1bfd` using the full lifecycle/self-review pass. The previous P1 transport-ordering issue is fixed: an already queued response is sent before a probe, and probes advertise only the successfully written `sentThrough` frontier. I also rechecked subscription generations, callback ownership, shutdown admission/draining, slow-session isolation, safeQueue stop/enqueue races, transaction cleanup, and disttae cancellation. Q1/Q2/Q3 ownership, termination, and boundedness all close without a remaining high-confidence blocker.

Local validation passed: full logtail service race suite; lifecycle/probe/generation/close race stress (`-count=50`); safeQueue race stress (`-count=100`); txn cleanup race stress (`-count=50`); affected disttae/logtail/txn/testutil packages; `go vet`; and `git diff --check`. Approve-level on this head.

aptend (approved):

LeftHandCold (approved):
Re-reviewed the latest head `d7f63f1bfd` with a full lifecycle and regression pass. I traced callback/response ownership, subscription generation transitions, notifier/session/server shutdown drains, slow-session isolation, transport-probe ordering, pull-worker bounds, safeQueue Stop/Enqueue races, TxnManager waitgroup accounting, WAL entry cleanup, and disttae readiness cancellation down through the real morpc `ClientSession.SessionCtx` and stopper termination contracts. I do not see a remaining high-confidence correctness or Q1/Q2/Q3 blocker.

Validation on this head: all affected package tests passed for logtail/service, logtail, disttae, logstore/sm, txnbase, txnimpl, and test/testutil; full race passed; the critical concurrency/shutdown tests passed for 10 repeated race runs; build, vet, and diff-check passed.

The branch currently conflicts with latest main only where two adjacent tests were added in `server_test.go`; production changes auto-merge, and retaining both tests resolves that conflict. It still needs a rebase before merge, but that is not a code-review blocker for this head.

gouhongshen (approved):
## Codex automated review

No blocking or non-blocking findings identified in the PR diff.

XuPeng-SH:
Re-reviewed the latest head `39bf1a4b8`.

This PR确实把 logtail push lifecycle 加固了不少：closeCB / session 清理链路、stale subscription generation 回收、TxnManager waitgroup 泄漏、WAL entry 释放、`waitReady` 可取消、slow session 隔离、shutdown drain、bootstrap waterline 这些点现在大多是闭合的；我本地补跑的 `logtail/service`、`logtail`、`disttae`、`safeq`、`txnbase`、`txnimpl`、`testutil` 非 race 都过了，`logtail/service` / `logtail` / `disttae` 的 `-race` 也过了。

但 latest 还有一个 merge blocker：`Session.CompleteSubscription` 现在在持有 `ss.mu` 的情况下调用 `SendSubscriptionResponse`（`pkg/vm/engine/tae/logtail/service/session.go:725-745`），而这个发送在 `sendChan` 满时会阻塞到 poison timeout（`session.go:821-885`）。与此同时，全局增量发布路径 `publishEvent -> session.Publish -> FilterLogtail` 会去拿同一个 `ss.mu.RLock()`（`server.go:630-635`, `session.go:632-689`）。因为 `logtailSender` 是单 goroutine 串行处理 `event.C` 的，一个已经拥塞、又在处理新订阅完成的 session，就能把后续所有 session 的增量发布一起卡住，直到超时。现有 `TestPublishEventDoesNotWaitForSlowSessionQueue` 只保护了 incremental 的 `TrySendUpdateResponse` 路径，但这个新的持锁订阅完成路径绕过了那层保护。

建议把 generation / state 线性化和阻塞发送拆开：在锁内只 claim / validate 当前 generation，释放锁后再 enqueue subscription response，发送完成后再回锁 finalize（必要时重检 generation 并回滚 / closeCB）。不然“slow session 不影响全局发布”这个 unhappy-path 不变量在订阅完成路径上还是破的。

gouhongshen (changes requested):
## Codex automated review

The PR fixes the prior blocking incremental enqueue, but queue saturation still blocks the single global sender during subscription completion. The latest discussion’s mutex explanation is not the primary mechanism; the synchronous call itself is sufficient. Focused tests could not run locally because cgo/libmo.dylib is absent; current CI is green.

### P1 - Do not block the global sender while enqueueing a subscription response (pkg/vm/engine/tae/logtail/service/session.go:739)
On main, both incremental and subscription responses could block the single logtailSender. This PR makes incremental publication non-blocking, but CompleteSubscription still calls the blocking SendSubscriptionResponse inline from logtailSender (server.go:607-627). With a full sendChan, SendResponse waits until capacity, session cancellation, or the default 5-second poison timeout. During that wait logtailSender cannot consume event.C or publish to any healthy session, violating issue #25644's queue-saturation requirement. The new TestPublishEventDoesNotWaitForSlowSessionQueue exercises only TrySendUpdateResponse by calling publishEvent directly, so it misses this transition. Make subscription delivery non-blocking or move the potentially blocking handoff off the global sender while preserving generation validation and exactly-once callback ownership.

XuPeng-SH:
Fixed the remaining P1 and its adjacent global-sender error path on latest head `6cb50479f7` (`a301d4e6df` + `6cb50479f7`).

Root cause correction: the latest review is right that `ss.mu` was not the primary blocker. The decisive wait-for edge was synchronous response hand-off from the single `logtailSender`. A full per-session queue could stop global progress even without lock contention.

What changed:

- Successful subscription response hand-off now uses a bounded fail-fast enqueue. If the per-session queue is full, the newly acquired response is released immediately, its merged phase-1/phase-2 callback is consumed exactly once, the slow stream is closed, and that subscription generation is rolled back (`a301d4e6df`).
- The final caller-by-response-type audit found the adjacent phase-2 failure edge: `logtailSender -> getSubLogtailPhase -> SendErrorResponse` could wait on the same full queue. Error response delivery is now fail-fast too; the existing error cleanup releases phase callbacks and unregisters the generation (`6cb50479f7`).
- The existing `ss.mu` critical section is intentionally retained as the generation/unsubscribe linearization point. Since enqueue is now a non-blocking select, it no longer adds a wait edge to the global sender. This avoids introducing an asynchronous third owner or a claim/finalize race.

Regression coverage:

- `TestCompleteSubscriptionDoesNotWaitForFullSessionQueue`: queue-full response ownership, generation rollback, active-count stability, and stream termination.
- `TestCongestedSubscriptionDoesNotBlockHealthyIncrementalProgress`: real single `logtailSender`, congested successful phase-2 completion, healthy-session progress, and exactly-once phase/incremental callbacks.
- `TestCongestedSubscriptionErrorDoesNotBlockHealthyIncrementalProgress`: real single `logtailSender`, congested phase-2 error response, healthy-session progress, error-path generation rollback, stream termination, and exactly-once phase/incremental callbacks.

Historical comment closure:

- The earlier transport-probe ordering finding in `discussion_r3587288844` remains fixed: queued responses win over probes, and probes advertise only successfully written `sentThrough` (`0e1bfe33e6`).
- Subscription generation validation, cancellation cleanup, shutdown admission/draining, notifier/session queue ownership, bootstrap waterline, safeQueue lifecycle, TxnManager accounting, WAL entry cleanup, disttae cancellation, and the test MORPC session lifetime were re-audited on the complete current diff.
- The previous incremental-only saturation test remains useful; the three new tests close the successful-subscription and phase-2-error entry points missing from review `4740748546`.
- Production caller enumeration now leaves blocking unsubscription delivery only in its independent request handler; global incremental delivery, successful subscription completion, and phase-2 error delivery are all bounded fail-fast paths.

Q1/Q2/Q3 result:

| Check | Result |
| --- | --- |
| Q1 ownership | successful phase callbacks transfer into one response and are released exactly once; error-path phase callbacks are also released exactly once |
| Q2 termination | the global sender never waits for per-session queue capacity; only the congested stream is closed and its in-flight generation is rolled back |
| Q3 boundedness | session queues remain bounded; no new goroutine, retry loop, or unbounded ownership queue was introduced |

Validation on `6cb50479f7`: all three new tests; focused lifecycle/probe/generation tests under `-race -count=50`; full normal and race suites for logtail service, logtail, disttae, safeQueue, txnbase, txnimpl, and testutil; service `go build`; service `go vet`; and `git diff --check` all passed.

Reflection: my prior comment over-attributed the stall to the mutex and treated the incremental fail-fast path as if it covered every response class. Even after fixing successful subscription delivery, the final matrix exposed the error-response sibling. The durable review method is to enumerate each response entry point by caller (`global sender` versus per-request handler), queue behavior, callback owner, generation rollback, and termination edge.

@gouhongshen please re-review the latest head when convenient. I am leaving the prior changes-requested review intact for the reviewer to resolve.

gouhongshen (approved):
## Codex automated review

No blocking or non-blocking findings. The latest discussion’s queue-saturation blocker is addressed: global-sender subscription success and error hand-offs are fail-fast, release callback ownership, and roll back the matching generation. Existing probe-ordering discussion remains fixed.

jiangxinmeng1 (approved):
> Reviewed the latest head d3762449. I found no blocking issues.
  >
  > The previous global-sender congestion problem is addressed: incremental updates,
  > successful subscription completions, and subscription error responses now use fail-
  > fast handoffs, while callback ownership and subscription-generation rollback remain
  > correctly handled. The shutdown draining, session cleanup, transport probing,
  > bootstrap waterline, and cancellation paths also look consistent.
  >
  > The caller-context behavior in the client request queue is acceptable as best-effort
  > cancellation once enqueueing can proceed immediately.
  >
  > Approving.

pkg/vm/engine/tae/logtail/service/session.go
```diff
@@ -503,24 +711,57 @@ func (ss *Session) AdvanceState(id TableID) {
...
+	defer ss.mu.Unlock()
+
+	entry, ok := ss.tables[id]
+	if !ok || entry.state != TableOnSubscription || entry.generation != generation {
+		if closeCB != nil {
+			closeCB()
+		}
+		return false, nil
+	}
+	if err := ss.SendSubscriptionResponse(sendCtx, tail, closeCB); err != nil {
```
XuPeng-SH, line 751: [P1] Break the subscription-completion / cleanup / sender lock cycle

`CompleteSubscription` holds `ss.mu` and calls `SendSubscriptionResponse`, which needs `enqueueMu.RLock`. Concurrently, `PostClean` can acquire `enqueueMu.Lock` and wait for `ss.wg`. If cleanup closes the transport while the sender is inside an earlier write, the write returns an error and `sendMessage` evaluates `ss.RemoteAddress()` before the sender can return and call `wg.Done`; `RemoteAddress` needs `ss.mu`. The resulting wait-for graph is:

- global logtail sender: `ss.mu -> enqueueMu.RLock`
- session reaper: `enqueueMu.Lock -> ss.wg`
- session sender: outstanding `ss.wg` member -> `ss.mu`

This is reachable when a slow/in-flight transport is reaped (for example after queue saturation) while the global sender completes a subscription for the same session. It indefinitely blocks cleanup and the single global logtail sender, so healthy sessions stop receiving progress.

I reproduced it deterministically on this exact head by parking `ClientSession.Write`, queueing `PostClean` as the `enqueueMu` writer, starting `CompleteSubscription` until it owned `ss.mu`, then allowing the canceled write to return. Neither cleanup nor completion could finish, and the test failed with the graph above after 250 ms. The existing full-queue completion test constructs a Session without a sender goroutine, so it cannot cover this edge.

Please break the cycle while retaining generation linearization and exactly-once response ownership—for example, do not wait for the sender while holding `enqueueMu`, or avoid acquiring the enqueue lock while holding `ss.mu`—and add the three-party regression.

XuPeng-SH (commented):
## Codex deep review — blocking

GitHub does not allow the authenticated PR author to submit `REQUEST_CHANGES`, so this is the equivalent blocking review comment: one P1 lifecycle deadlock remains on the latest head `d37624493b`.

I re-audited the complete diff and earlier review threads across subscription generations, response/callback ownership, transport probing, queue saturation, server shutdown, upstream callback hand-off, safeQueue, transaction waitgroups, and WAL cleanup. The earlier congestion and probe-ordering findings are fixed, but the new cleanup serialization introduces the P1 lock cycle in the inline finding.

Validation on the exact PR head: the seven affected package suites passed normally and with `-race`; `go vet` and `git diff --check` passed. I also added a temporary barrier-based regression using a real `NewSession` sender and the `ClientSession` contract. It deterministically failed after 250 ms with the wait-for graph described below. The temporary test and CGo links were removed; the review worktree is clean.

pkg/vm/engine/tae/logtail/service/session.go
XuPeng-SH: Fixed in `179dbfe201`.

The cleanup protocol is now `cancel response admission -> close transport -> cross the admission barrier -> join the sender without holding the barrier -> drain queued responses`. This removes the `enqueueMu -> wg` lock edge while preserving the barrier guarantee: any hand-off that passed the pre-cancel context check must leave `enqueueMu.RLock` before cleanup crosses the writer barrier; later hand-offs observe the canceled session and cannot enqueue. `RemoteAddress` also reads the construction-sealed stream address without taking subscription-state `ss.mu`, so diagnostic error logging cannot re-enter the state lock.

Added deterministic regressions:
- `TestPostCleanCancelsBeforeClosingTransport`
- `TestPostCleanDoesNotHoldAdmissionBarrierWhileJoiningSender`

The second test uses a real `NewSession` sender and recreates the three-party wait-for graph with barriers. It deadlocks on the previous implementation and terminates on this one, while also asserting generation rollback, exactly-once callback release, zero active subscriptions, and no table residue.

Validation on the final diff: both new tests under `-race -count=100`; the 20-test historical probe/generation/congestion/shutdown/bootstrap matrix under `-race -count=50`; all seven affected packages normally and with race; build; vet; and `git diff --check` all passed.

XuPeng-SH:
Final first-principles closure pass on head `179dbfe201`.

| Historical finding / lifecycle area | Current closure |
| --- | --- |
| Probe can overtake a queued update (`discussion_r3587288844`) | Fixed by queued-response priority plus sender-owned `sentThrough`; update/subscription/failed-write regressions passed under race stress. |
| Full session queue blocks global sender on subscription success (`review 4740748546`) | Success hand-off is fail-fast; response/callback ownership is released exactly once and the matching generation rolls back. |
| Adjacent phase-2 error response can block the same global sender | Error hand-off is also fail-fast; healthy incremental progress and error cleanup regressions passed. |
| Cleanup / subscription completion / session sender lock cycle (`discussion_r3637569191`) | Fixed in `179dbfe201`: cancel before transport close, cross the admission barrier, release it before sender join, and keep immutable diagnostics outside `ss.mu`. Deterministic three-party regression added. |
| Stale subscription generations and cancel/failure races | Completion/failure carries and rechecks generation; stale attempts cannot publish or cancel a newer attempt. |
| closeCB / response / phase ownership | Queue rejection, write failure, stale completion, notifier drain, session drain, and server shutdown each consume ownership exactly once. |
| Shutdown and boundedness | Lifecycle admission closes before drain; sender/session/notifier/phase queues terminate; pull workers and history/queues remain bounded. |
| Bootstrap/progress correctness | Bootstrap waterline is ordered behind accepted events; probes expose only successfully sent progress. |
| safeQueue / TxnManager / WAL / disttae readiness | Stop-vs-enqueue, waitgroup rollback, WAL entry release, and context cancellation regressions all remain green. |

Final validation:
- 20 historical critical regressions: `-race -count=50`
- new cleanup regressions: `-race -count=100`
- affected seven packages: full normal and full race suites
- affected packages: build and vet
- `git diff --check`

No remaining high-confidence Q1 ownership, Q2 termination, Q3 boundedness, protocol-ordering, or generation-lifecycle blocker was found in the complete current diff. The PR remains mergeable but behind `main`; CI for the pushed head has started.

gouhongshen (approved):
## Codex automated review

The previously reported lifecycle blockers are closed on the current head. One non-blocking admission-latency regression remains.

### P2 - Preserve immediate wake-up when subscription queue capacity returns (pkg/vm/engine/tae/logtail/service/server.go:371)
Before this change, the retry select included `case s.subReqChan <- sub`, so a pull worker freeing a slot woke the blocked request immediately. The new non-blocking probe can miss a slot that opens just afterward, then waits only for the 1-second ticker; under load this adds up to 1 second per admission and can make short caller deadlines expire despite available capacity. Keep an event-driven channel-send wakeup while retaining the shutdown admission barrier.

XuPeng-SH:
Fixed the Linux/arm64 SCA failure in `3856a1dedb`.

The notification summary pointed at `pkg/common/mpool/mpool_test.go` importing `unsafe`, but the job log shows that as an existing `molint` baseline report; the same step also reports the existing vector test import and then continues through license checks. This PR has no diff in `mpool_test.go`. The command that actually exits non-zero reports one issue only: `session.go:554 SA2001: empty critical section` from staticcheck.

The fix does not suppress the linter. Cleanup now explicitly sets `enqueueClosed` while holding the writer barrier, and every response hand-off checks that state under the reader lock. This preserves the proven `cancel -> cross/seal admission -> unlocked sender join -> drain` lifecycle, makes the synchronization purpose concrete, and guarantees rejected post-clean responses release their callback ownership.

- CI-matching `golangci-lint 2.6.2` on `./pkg/vm/engine/tae/logtail/service`: `0 issues`
- cleanup regressions: `-race -count=100`
- 20 historical lifecycle/probe/generation/congestion/shutdown regressions: `-race -count=50`
- all seven affected packages: normal and race suites
- affected package build and vet
- `git diff --check`

The new CI run has started on head `3856a1dedb`.

gouhongshen (approved):
## Codex automated review

The lifecycle, ownership, ordering, cancellation, and shutdown fixes are otherwise sound. Prior blocking findings are closed on the current head. One non-blocking admission-latency regression remains.

### P2 - Restore event-driven subscription queue admission (pkg/vm/engine/tae/logtail/service/server.go:371)
After the non-blocking enqueue at lines 355–362 misses a full queue, this select waits only for cancellation or the one-second ticker. If a pull worker frees capacity immediately afterward, the request is not awakened and can wait nearly one second—or fail a shorter caller deadline—even though the queue has space. The previous select waited directly on `s.subReqChan <- sub`, so capacity restoration woke it immediately. Preserve that wake-up while retaining the shutdown admission barrier.

Issue #25598, opened 2026-07-11
[Bug] task runner retries all-due tasks indefinitely and breaks exactly-once


Ariznawlll:
## Summary

The task runner does not drain its retry queue when all retry entries are due. The same expired task remains in the queue and is re-executed approximately every 100 ms, violating exactly-once behavior.

## Evidence

- Commit: `7821d21c723372e45a8f65da3bbb896a157b1036`
- Code: `pkg/taskservice/task_runner.go:452-479`
- The loop only compacts `r.retryTasks.s` inside the `retryAt.After(now)` branch. If every entry is due, that branch is never reached and the slice is unchanged.

## Expected

Due retry entries should be removed from the queue before execution, and a task should not be re-enqueued after it has been consumed.

## Actual

An all-due retry queue is left intact; the same `runningTask` can be executed repeatedly on each retry tick.

Please add a regression test with an all-due queue and verify exactly-once execution.

jiangxinmeng1:
fixed by #25647

Ariznawlll:
已在最新 `main` 上验证 task runner 的 all-due retry queue 行为。

### 验证版本

- Commit：`0977fc07c51067ce76b769c7853d7374cb2fccec`
- 验证时间：2026-07-14

### 验证方法

```bash
go test ./pkg/taskservice \
  -run 'TestRetryDrainsAllDueTasks|TestRetryPreservesFutureTasks' \
  -count=1 -v
```

第一条用例将 retry queue 中所有任务设置为已到期，检查执行后队列是否被 drain；第二条混入 future task，检查 compact 时是否误删未到期项。

### 实际结果

all-due task 只执行一次，等待多个 retry tick 后没有再次执行；future task 保持在队列中。两个用例均 PASS。

### 结论

当前 main 已在执行前消费 due entries，不再每约 100ms 重复执行同一 expired task，exactly-once 破坏路径已修复。

![#25598 verification](https://gist.githubusercontent.com/Ariznawlll/a2432edd0696ba6684e345e80bf49320/raw/25598.png)

PR #25647, opened 2026-07-13
fix: drain due retry tasks before execution

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25598

  Fix task runner retry queue draining when every retry entry is already due.

  Previously, `retry()` only compacted `retryTasks.s` after finding the first
  future `retryAt`. If all entries were due, the queue was left unchanged, so
  the same `runningTask` could be re-executed on every retry tick.

  This change removes the due prefix from the retry queue before executing those
  tasks, including the all-due case.


XuPeng-SH (changes requested):
Request changes: retry queue draining is fixed, but the retry execution lifecycle still has a P1 parallelism-token ownership gap.

Concrete failure path:
1. The initial dispatch calls runTask and acquires one parallelismC token.
2. On executor failure, taskExecResult successfully adds the runningTask to retryTasks but does not release that token.
3. retry() drains the due entry and calls runTask again, which attempts to acquire a second token.
4. With parallelism=1, the retry worker blocks for the runner lifetime and the task never retries. With parallelism=N, each retry consumes another slot while terminal completion releases only one, permanently reducing capacity.

The new regression test manually appends a runningTask to retryTasks.s without modeling the token held by the real producer path, so it passes with parallelism=1 while missing this failure.

Please make token ownership explicit: either reuse the retained token by calling run(rt) for retry entries, or release before enqueue and reacquire at retry time. Add an end-to-end test using the real dispatch path with parallelism=1: first execution fails, second succeeds, execution count is 2, retry queue is empty, task completes, and len(parallelismC) returns to 0. A mixed due/future queue case would also verify that future entries are preserved.

XuPeng-SH (approved):
Latest head looks good to me.

The retry queue drain fix now closes the ownership gap I called out last time: due retry entries are removed from the queue before execution, and retry() reuses the already-held parallelism token by calling run(rt) directly instead of going back through runTask and trying to acquire a second token.

That means parallelism=1 no longer deadlocks on retry, terminal completion still releases exactly one token, and the new tests also cover the all-due and due-plus-future queue shapes. I don't see a remaining high-confidence correctness blocker at this head.

PR #25648, opened 2026-07-13
fix(proxy): cancel backend on client disconnect

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25599

 When the proxy detects a client-side disconnect, route backend cleanup through
  KillCurrentBackendConn instead of only closing the current proxy-to-CN
  connection.
  This sends KILL CONNECTION to the CN so blocked in-flight statements are
  canceled
  and cannot commit after the client has gone away.

  Keep the expected COM_QUIT/cache path unchanged by skipping cancellation when
  the tunnel has already marked an expected cache quit.

  Tests cover EOF, closed/reset connection errors, wrapped client disconnect
  errors, and the expected cache-quit skip path.


pkg/proxy/handler.go
```diff
@@ -350,20 +350,7 @@ func (h *handler) cleanupBackendOnClientDisconnect(err error, cc ClientConn, t *
 	if t != nil {
 		currentSC = t.getServerConn()
 	}
-	// For normal EOF / closed-connection exits, close the current backend
-	// session directly instead of creating an extra proxy->CN KILL connection.
-	if isEOFErr(err) || isConnEndErr(err) {
```
gouhongshen, line 355: [P1] Avoid unconditionally issuing KILL for every EOF/closed disconnect

This deletion makes the KILL path run for all codeClientDisconnect errors, including io.EOF, net.ErrClosed, and ECONNRESET. KillCurrentBackendConn opens a new proxy-to-CN connection, performs a handshake, and executes KILL CONNECTION. For a normal idle EOF—especially when conn-cache is disabled, where tunnel.Close has already closed the backend—this adds one unnecessary control connection per disconnect.

Merged PR #24157 documented this exact regression: phase-boundary high-concurrency workloads produced CN handshake storms, dial/auth timeouts, and new-client failures. Please distinguish the in-flight-statement cancellation case from normal EOF, or use a cancellation mechanism that does not create a new CN handshake for every normal disconnect.

pkg/proxy/handler.go
```diff
@@ -350,20 +350,7 @@ func (h *handler) cleanupBackendOnClientDisconnect(err error, cc ClientConn, t *
...
-		}
-		if currentSC != nil {
-			if err := currentSC.Close(); err != nil {
-				h.logger.Warn("failed to close backend connection after client disconnect",
-					zap.Uint32("Conn ID", cc.ConnID()),
-					zap.Error(err),
-				)
-			}
-		}
+	if t != nil && t.hasExpectedCacheQuit() {
```
gouhongshen, line 356: [P1] Define deterministic cleanup when KILL fails

KillCurrentBackendConn can fail during HAKeeper ID allocation, CN dial/handshake, or command execution, but this code only logs the error. With conn-cache enabled, tunnel.Close intentionally leaves the current backend connection open. The subsequent handleQuitEvent path then relies on connCache.Push/reset-session or ServerConn.Quit; the latter has no timeout and waits for a backend response.

If the KILL control path fails while the original statement is still blocked, this can leave cleanup dependent on a secondary path or hang the handler. The previous EOF path at least closed currentSC here. Please define and test a deterministic, bounded terminal state for this failure case; direct TCP close is not equivalent to KILL CONNECTION for the late-commit bug.

pkg/proxy/handler_test.go
```diff
@@ -413,12 +413,12 @@ func TestHandler_handleTunnelErrClosesBackendForEOFClientDisconnect(t *testing.T
 
 	ret := h.handleTunnelErr(withCode(io.EOF, codeClientDisconnect), cc, tun, 733923, 100)
 	require.NoError(t, ret)
-	require.Equal(t, 0, killCalled)
-	require.Equal(t, 1, closeCalled)
+	require.Equal(t, 1, killCalled)
```
gouhongshen, line 460: [P2] Add an end-to-end and failure-path regression test

These tests only assert that a mocked ClientConn receives one KillCurrentBackendConn call. They do not exercise the real blocked UPDATE + client hard-abort reproduction from issue #25599, a KILL failure, conn-cache enabled cleanup, or concurrent normal EOFs. The existing client_conn unit test checks the generated SQL separately, but not this handler lifecycle. Please add coverage for the behavior that this PR is intended to guarantee and for the failure path above.

gouhongshen (changes requested):
Request changes.

I found two blocking issues and one test-coverage gap:

1. P1 — The change unconditionally opens a CN control connection for every client-disconnect error, including normal EOF/closed/reset paths. This reintroduces the connection/handshake storm documented in merged PR #24157 under phase-boundary high-concurrency workloads.

2. P1 — If KillCurrentBackendConn fails, the new path only logs the error. With conn-cache enabled, tunnel.Close intentionally leaves the backend connection alive, so cleanup then relies on the cache/QUIT path, which is not a deterministic or bounded termination path for a blocked statement.

3. P2 — The new tests mock KillCurrentBackendConn and do not cover the actual blocked-statement reproduction, KILL failure, conn-cache path, or high-concurrency normal EOF behavior.

The PR scope is otherwise clean, and the existing GitHub CI is green, but these paths need to be resolved and covered before merge.

XuPeng-SH (changes requested):
Requesting changes.

Besides the existing review, I still see two cleanup-path blockers here.

1. The PR description says the expected COM_QUIT path stays unchanged, but that is only true for the conn-cache path.

`expectedCacheQuit` is only set when `makeEvent()` intercepts COM_QUIT while `connCacheEnabled` is true (`pkg/proxy/event.go`, `pkg/proxy/client_conn.go`, `pkg/proxy/tunnel.go`). When conn-cache is disabled, a normal graceful COM_QUIT is forwarded to CN and nothing marks it as expected. The follow-up client EOF can therefore still enter `cleanupBackendOnClientDisconnect()` and call `KillCurrentBackendConn(...)`, so ordinary closes now pay the extra proxy->CN control connection / KILL path as well. That does not match the stated intent. Please either narrow the KILL trigger to genuinely abnormal disconnects, or add a separate marker so non-cache COM_QUIT keeps the old path too.

2. The cache-enabled abnormal-disconnect path is not closed end-to-end yet.

`cleanupBackendOnClientDisconnect()` now sends KILL first, but `handleTunnelErr()` still unconditionally runs `clientConn.handleQuitEvent()` on EOF/conn-end when `h.connCache != nil` (`pkg/proxy/handler.go`). `handleQuitEvent()` then pauses the pipe and tries `connCache.Push(c.sc)` or `c.sc.Quit()` (`pkg/proxy/client_conn.go`) against the same backend session we just chose to cancel.

So there is still no single deterministic "cancel and discard" path for client-abort: the killed session is fed straight into cache/QUIT cleanup afterward. At best this is redundant control-plane work on every abort; at worst it races the cache-reset path. Once we decide to KILL on client-abort, the conn-cache cleanup path should be suppressed or explicitly switched into a non-cacheable discard mode.

Coverage is also still short on the unhappy paths above: the new tests stay on `mockClientConn` / nil-cache handler paths, so they do not cover normal COM_QUIT without cache, the cache-enabled `*clientConn` sequencing, or kill-failure behavior.

XuPeng-SH (approved):
LGTM after reviewing this from the disconnect-classification, backend-cancellation, and unhappy-path angles.

- correctness: client-side disconnects that can strand an in-flight backend statement now route through `KillCurrentBackendConn`, which is the missing path needed to turn a vanished client into backend cancellation instead of a later silent commit
- containment: the expected COM_QUIT / conn-cache lifecycle is still carved out explicitly via `expectedCacheQuit` / `expectedClientQuit`, so the new cancellation path does not interfere with the normal quit/cache behavior
- unhappy path coverage: the new tests cover EOF, closed/reset connection variants, wrapped client-disconnect errors, kill failure fallback, and the expected cache-quit / forwarded-QUIT skip paths

PR #25651, opened 2026-07-13
fix(lockservice): recover active txn checks after CN recreation

### LINKED ISSUE #25650: [Bug]: TN lockservice keeps fencing a CN after hot recreation

## Branch Name

main

## Commit ID

2865ed6cc8a39727b8b274bb52b88fe441d0b30c

- Deployment: Docker multi-CN local-disk setup
- Topology: one TN and two CN services
- Failure injection: force-recreate both CN containers while preserving their hostnames and CN UUIDs

## Actual Behavior

After both CN containers are hot-recreated, their Docker IPs can swap while the hostnames and CN UUIDs remain stable. The long-lived TN process can keep using a stale hostname/backend resolution for `GetActiveTxn`. Orphan-transaction cleanup then repeatedly treats the CN as unreachable and fences it. The recreated CN calls `Resume`, but a stale cleanup result can race with that request and fence the CN again. Commits eventually fail with error 20708 and long-running transactions are asynchronously aborted.

A second failure mode appears after recovery has pinned an IP: on a later IP swap, that stale IP can belong to another healthy CN. The RPC succeeds but returns `Valid=false` for the target service, which was incorrectly treated as an authoritative negative response.

## Expected Behavior

A CN hot recreation must not leave TN permanently fencing a healthy recreated CN. Cleanup must refresh the affected backend, re-resolve the endpoint without relying on stale CGO resolver state, and confirm a negative `GetActiveTxn` response before discarding transaction-control state. Cancellation, timeouts, and backend recreation races must remain bounded.

## Steps to Reproduce

1. Start the Docker multi-CN local-disk cluster.
2. Run a long transaction or continuous writes through both CNs.
3. Force-recreate both CN containers without restarting TN.
4. Observe the CNs call `lockservice.Resume()` after recovery.
5. Wait for TN orphan-transaction cleanup cycles.
6. Before the fix, TN can repeatedly return 20708 (`CannotCommitOnInvalidCN`) and abort the long-running transaction.

## Additional Information

Root cause combines three independent races:

1. MORPC pools by hostname, and an asynchronous pre-reset backend-create request can republish a stale backend after targeted reset.
2. In a long-lived CGO process, hostname resolution can remain stale across Docker endpoint recreation.
3. Cleanup fencing and `Resume` were not linearized, so an old cleanup could re-fence a resumed CN.

The fix adds epoch-protected targeted backend invalidation, a bounded pure-Go DNS recovery lookup for `GetActiveTxn`, one reset-and-retry confirmation for `Valid=false`, context-aware bounded retries, and deterministic regression/race tests.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25650

Fixes TN-side lockservice recovery after CN hot recreation without restarting TN.

- Adds MORPC targeted backend detachment with per-remote epochs, preventing queued or in-flight pre-reset backend creation from repopulating a stale pool.
- Uses a bounded pure-Go DNS lookup only on the `GetActiveTxn` recovery path, avoiding stale long-lived CGO hostname resolution while leaving normal lock/transaction hot paths unchanged.
- Confirms the first successful `Valid=false` response after resetting the endpoint, preventing a stale IP reassigned to another CN from being treated as an authoritative response.
- Makes cleanup retries bounded and context-aware; handles timeouts, EOF, network errors, unavailable backends, and cancellation through MORPC error classification.
- Linearizes `Valid` commit admission with invalid-CN fencing and protects `Resume` from stale cleanup re-fencing.
- Bounds the recovery endpoint cache and preserves public MORPC interface compatibility.

Validation:

- `go test ./pkg/common/morpc -count=1`
- `go test ./pkg/lockservice -count=1`
- Focused race tests (`-race`, repeated 10/20 times)
- `make dev-build` with typecheck
- `BenchmarkValidDuringRecoveryChanges`: 125-145 ns/op, 0 B/op, 0 allocs/op
- Docker hot-recreate reproduction: no post-ready 20708 in the observed cleanup window.

### DIFF
--- a/pkg/clusterservice/cluster.go
+++ b/pkg/clusterservice/cluster.go
@@ -162,7 +162,8 @@ func WithDisableRefresh() Option {
 type cluster struct {
 	logger          *log.MOLogger
 	stopper         *stopper.Stopper
-	mu              sync.Mutex
+	refreshGateOnce sync.Once
+	refreshGateC    chan struct{}
 	client          ClusterClient
 	refreshInterval time.Duration
 	forceRefreshC   chan struct{}
@@ -307,7 +308,7 @@ func (c *cluster) ForceRefresh(sync bool) {
 		return
 	}
 	if sync {
-		c.refresh()
+		c.refreshAndLog(context.Background())
 		return
 	}
 
@@ -315,6 +316,25 @@ func (c *cluster) ForceRefresh(sync bool) {
 	case c.forceRefreshC <- struct{}{}:
 	default:
 	}
+}
+
+// Refresh obtains and publishes a new HAKeeper snapshot, or returns the reason
+// why freshness could not be established. Unlike ForceRefresh, this method is
+// suitable for callers whose correctness depends on a successful refresh.
+func (c *cluster) Refresh(ctx context.Context) error {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if c.options.disableRefresh {
+		// WithDisableRefresh makes the supplied static snapshot authoritative.
+		return nil
+	}
+	ctx, cancel := context.WithTimeoutCause(ctx, c.refreshInterval, moerr.CauseRefresh)
+	defer cancel()
+	return c.refreshWithContext(ctx)
 }
 
 func (c *cluster) Close() {
@@ -430,32 +450,36 @@ func (c *cluster) refreshTask(ctx context.Context) {
 			c.logger.Info("refresh cluster details task stopped")
 			return
 		case <-timer.C:
-			c.refresh()
+			c.refreshAndLog(ctx)
 			timer.Reset(c.refreshInterval)
 		case <-c.forceRefreshC:
-			c.refresh()
+			c.refreshAndLog(ctx)
 		}
 	}
 }
 
-func (c *cluster) refresh() {
+func (c *cluster) refreshAndLog(parent context.Context) {
+	if err := c.Refresh(parent); err != nil {
+		c.logger.Error("failed to refresh cluster details from hakeeper",
+			zap.Error(err))
+	}
+}
+
+func (c *cluster) refreshWithContext(ctx context.Context) error {
 	defer c.logger.LogAction("refresh from hakeeper",
 		log.DefaultLogOptions().WithLevel(zap.DebugLevel))()
 
-	// There is data race as ForceRefresh and refreshTask may call this function
-	// at the same time, which will cause inconsistent CN services.
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
-	ctx, cancel := context.WithTimeoutCause(context.Background(), c.refreshInterval, moerr.CauseRefresh)
-	defer cancel()
+	// Serialize snapshots so an older concurrent HAKeeper response cannot
+	// overwrite a newer one. Admission is context-aware: freshness callers must
+	// not wait past their own deadline behind another refresh.
+	if err := c.acquireRefresh(ctx); err != nil {
+		return moerr.AttachCause(ctx, err)
+	}
+	defer c.releaseRefresh()
 
 	details, err := c.client.GetClusterDetails(ctx)
 	if err != nil {
-		err = moerr.AttachCause(ctx, err)
-		c.logger.Error("failed to refresh cluster details from hakeeper",
-			zap.Error(err))
-		return
+		return moerr.AttachCause(ctx, err)
 	}
 
 	c.logger.Debug("refresh cluster details from hakeeper",
@@ -491,6 +515,24 @@ func (c *cluster) refresh() {
 		c.ready.Store(true)
 		close(c.readyC)
 	})
+	return nil
+}
+
+func (c *cluster) acquireRefresh(ctx context.Context) error {
+	c.refreshGateOnce.Do(func() {
+		c.refreshGateC = make(chan struct{}, 1)
+		c.refreshGateC <- struct{}{}
+	})
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-c.refreshGateC:
+		return nil
+	}
+}
+
+func (c *cluster) releaseRefresh() {
+	c.refreshGateC <- struct{}{}
 }
 
 func (c *cluster) copyServices() *services {
--- a/pkg/clusterservice/cluster_test.go
+++ b/pkg/clusterservice/cluster_test.go
@@ -16,6 +16,7 @@ package clusterservice
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"testing"
 	"time"
@@ -87,6 +88,61 @@ func TestClusterForceRefresh(t *testing.T) {
 			c.GetCNService(NewServiceIDSelector("cn0"), apply)
 			assert.Equal(t, 1, cnt)
 		})
+}
+
+func TestClusterRefreshReportsFailureAndPreservesSnapshot(t *testing.T) {
+	runClusterTest(
+		time.Hour,
+		func(hc *testHAKeeperClient, c *cluster) {
+			hc.addCN("cn-old")
+			require.NoError(t, c.Refresh(context.Background()))
+
+			refreshErr := errors.New("injected hakeeper failure")
+			hc.Lock()
+			hc.err = refreshErr
+			hc.value.CNStores = nil
+			hc.Unlock()
+
+			require.ErrorIs(t, c.Refresh(context.Background()), refreshErr)
+			var ids []string
+			c.GetCNServiceWithoutWorkingState(
+				NewSelector(),
+				func(service metadata.CNService) bool {
+					ids = append(ids, service.ServiceID)
+					return true
+				},
+			)
+			require.Equal(t, []string{"cn-old"}, ids)
+		},
+	)
+}
+
+func TestClusterRefreshHonorsCanceledContext(t *testing.T) {
+	runClusterTest(
+		time.Hour,
+		func(_ *testHAKeeperClient, c *cluster) {
+			ctx, cancel := context.WithCancel(context.Background())
+			cancel()
+			require.ErrorIs(t, c.Refresh(ctx), context.Canceled)
+		},
+	)
+}
+
+func TestClusterRefreshWaitHonorsContext(t *testing.T) {
+	runClusterTest(
+		time.Hour,
+		func(_ *testHAKeeperClient, c *cluster) {
+			require.NoError(t, c.acquireRefresh(context.Background()))
+			defer c.releaseRefresh()
+
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
+			defer cancel()
+			started := time.Now()
+			err := c.Refresh(ctx)
+			require.ErrorIs(t, err, context.DeadlineExceeded)
+			require.Less(t, time.Since(started), time.Second)
+		},
+	)
 }
 
 func TestClusterRefresh(t *testing.T) {
--- a/pkg/clusterservice/types.go
+++ b/pkg/clusterservice/types.go
@@ -109,6 +109,14 @@ type MOCluster interface {
 	UpdateCN(metadata.CNService)
 }
 
+// AuthoritativeRefresher is an optional capability for callers that must know
+// whether a synchronous cluster snapshot refresh actually succeeded. The
+// legacy ForceRefresh API intentionally has no result and is unsuitable for
+// correctness decisions based on freshness.
+type AuthoritativeRefresher interface {
+	Refresh(context.Context) error
+}
+
 type ClusterClient interface {
 	GetClusterDetails(ctx context.Context) (logpb.ClusterDetails, error)
 }
--- a/pkg/common/moerr/cause.go
+++ b/pkg/common/moerr/cause.go
@@ -109,6 +109,7 @@ var (
 	CauseDoGetLock                 = NewInternalError(context.Background(), "doGetLock")
 	CauseInitRemote1               = NewInternalError(context.Background(), "initRemote 1")
 	CauseInitRemote2               = NewInternalError(context.Background(), "initRemote 2")
+	CauseResetLockServiceBackend   = NewInternalError(context.Background(), "resetLockServiceBackend")
 	CauseGetTxnWaitingListOnRemote = NewInternalError(context.Background(), "getTxnWaitingListOnRemote")
 	CauseGetLockTableBind          = NewInternalError(context.Background(), "getLockTableBind")
 	CauseAbortRemoteDeadlockTxn    = NewInternalError(context.Background(), "abortRemoteDeadlockTxn")
--- a/pkg/common/moerr/cause_test.go
+++ b/pkg/common/moerr/cause_test.go
@@ -101,6 +101,7 @@ var causeArray = []error{
 	CauseDoGetLock,
 	CauseInitRemote1,
 	CauseInitRemote2,
+	CauseResetLockServiceBackend,
 	CauseGetTxnWaitingListOnRemote,
 	CauseGetLockTableBind,
 
--- a/pkg/common/morpc/backend_test.go
+++ b/pkg/common/morpc/backend_test.go
@@ -1132,9 +1132,13 @@ func (b *testBackend) Locked() bool {
 }
 
 func (b *testBackend) active() {
+	b.setActiveTime(time.Now())
+}
+
+func (b *testBackend) setActiveTime(value time.Time) {
 	b.RWMutex.Lock()
 	defer b.RWMutex.Unlock()
-	b.activeTime = time.Now()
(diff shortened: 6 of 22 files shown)


XuPeng-SH (commented):
Self-review blocker before merge:

The allocator-side `GetActiveTxn` recovery work in this PR is much stronger now, but the functional closure is still open on the peer-CN active-txn validation path.

`pkg/lockservice/rpc.go` only routes `Method_GetActiveTxn` through the new recovery backend / targeted reset flow. `Method_CheckActiveTxn` still goes through the normal backend path. But `pkg/lockservice/service_remote.go` still uses `pb.Method_CheckActiveTxn` first for peer validation, and it treats `resp.CheckActiveTxn.Valid == false` as an authoritative "txn inactive" answer.

That means the hot-recreate/stale-IP problem this PR is fixing can still hit the peer path:
- stale MORPC backend or stale endpoint reaches the old/wrong CN incarnation;
- that CN legitimately returns `Valid=false` for the target service;
- `service_remote.go` returns `false, nil`;
- `pkg/lockservice/service.go:isValidRemoteTxn()` then falls through to `canUnlockRemoteTxn(...)`, which can fence/unlock a still-live remote txn.

So right now the PR hardens the allocator cleanup plane, but not the peer active-txn validation plane that feeds unlock/fencing decisions.

Suggested fix:
1. either teach `CheckActiveTxn` to use the same `activeTxnBackend` / targeted reset / confirmation flow as `GetActiveTxn`,
2. or, on `Valid=false` in `service_remote.go`, reset the backend and re-check before treating the txn as inactive.

Please also add at least one regression test for the peer path (`CheckActiveTxn` / `isValidRemoteTxn`) covering stale `Valid=false` after CN recreation or stale-IP reassignment; the new tests are strong on allocator recovery, but they currently do not cover this second closure.

aptend (changes requested):
Requesting changes for the following correctness and lifecycle blockers:

1. **The peer-CN `CheckActiveTxn` path still bypasses recovery.** The new recovery endpoint and targeted reset logic is only applied to `GetActiveTxn`. `service_remote.go` uses `CheckActiveTxn` first and treats `Valid=false` as authoritative, so a stale IP reassigned to another CN can still cause a live remote transaction to be fenced and its locks released. Please apply the same reset/re-resolution/negative-confirmation semantics to this path and add a peer-path regression test.

2. **A failed reset is currently counted as a successful negative-response confirmation.** `resetActiveTxnBackend` returns `true` whenever the client implements `ResetBackend`, even when `ResetBackend` returns an error. A DNS or discovery failure can therefore be followed by a second response from the same stale endpoint, which is then accepted and used to clean control state. Please distinguish unsupported, successful reset, and failed reset; a reset failure must leave the service state unknown and preserve control state.

3. **The queue-full create fallback can adopt a post-reset epoch.** `getBackend` captures the old epoch, but when `triggerCreateAtEpoch` reports a full queue it calls `createBackendWithBookkeeping`, which reads the epoch again. If `CloseBackendFor` runs between those operations, the pre-reset request adopts the new epoch and can repopulate the backend after reset. Please pass the originally captured epoch through the fallback path and reject it when the epoch changed. Add a deterministic regression test for this interleaving.

4. **`backendEpoch` has no reclamation path.** Every targeted reset inserts entries for discovered and resolved endpoints, while the map is never pruned. The recovery cache is bounded, but historical CN hostname/IP churn can still grow this map for the lifetime of TN. Please add safe reclamation or document and enforce a concrete bound without re-admitting stale create requests.

The existing focused race tests and full `pkg/common/morpc` / `pkg/lockservice` tests pass, but they do not cover these failure paths.

gouhongshen:
Deep re-review of the current head `35fc58b`. I verified that the four earlier review findings are fixed in this head, so I am not repeating them.

Conclusion: request changes before merge.

### 1. Blocker: SCA fails on a field introduced by this PR

The arm64 SCA job fails with:

`pkg/common/morpc/client.go:1090:2: field value is unused (unused)`

This is the `backendGeneration.value` field in [client.go](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client.go#L1087-L1091). It is intended only to make the struct non-zero-sized, but the current representation fails golangci-lint. Please use a valid non-zero blank field or make the generation value meaningful and rerun SCA.

[Failed SCA job](https://github.com/matrixorigin/matrixone/actions/runs/29228745258/job/86748521806)

### 2. P1: multi-A DNS can still accept a stale negative identity response

[resolveTCP4Endpoint](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/lockservice/rpc.go#L386-L404) deliberately returns the original hostname when DNS returns more than one IPv4 address. [ResetBackend](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/lockservice/rpc.go#L328-L343) then treats `endpoint == address` as a successful reset.

The next request can therefore still resolve to the old IP or another CN, return `Valid=false`, and [checkRemoteActiveTxn](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/lockservice/service_remote.go#L150-L159) accepts the second negative response as authoritative. That reopens the original live-transaction fencing bug for a multi-A CN hostname.

The reset should not be considered confirmed merely because the hostname was re-dialed. Either verify a concrete endpoint/service identity, or explicitly enforce a single-endpoint contract and preserve the state when that contract is not met. Please add a deterministic multi-A regression test.

### 3. P2: targeted reset can block behind global GC network creation

[runCreateLoop](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client_gc.go#L434-L450) holds `req.c.mu` while calling `createBackendLocked`; that path calls the backend factory and performs network connection work in [client.go](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client.go#L1202-L1226). [CloseBackendFor](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client.go#L846-L863) needs the same mutex before it can invalidate the generation.

As a result, ResetBackend has no way to honor the caller's context while waiting for an in-progress GC connection. This can stall peer active-txn validation and cleanup; a custom BackendFactory can make the wait unbounded. The current tests cover the out-of-lock synchronous creation path, but not a blocked queued GC creation racing with CloseBackendFor.

Please move GC factory I/O out of `c.mu` while retaining the generation checks before and after creation, or provide a genuinely context-aware reset path.

### 4. P2: active-txn recovery resets the shared lock RPC transport

The lockservice client has one underlying MORPC client for all methods ([rpc.go](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/lockservice/rpc.go#L69-L79), [AsyncSend](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/lockservice/rpc.go#L175-L275)). ResetBackend closes every backend for that address, not only the active-txn probe connection.

Therefore a single stale `CheckActiveTxn` response or connection error can interrupt concurrent `Lock`, `Unlock`, `GetTxnLock`, and `KeepRemoteLock` requests. The lock path has conservative bookkeeping for lost lock responses, so I am not claiming immediate data corruption, but this can cause response loss, retry storms, and avoidable latency. Please consider a dedicated active-txn RPC client/backend namespace, or add a concurrency regression test that proves the disruption is acceptable.

### 5. P2 when MORPC circuit breakers are enabled: old requests can reopen the new generation

[CloseBackendFor](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client.go#L857-L863) removes the circuit breaker, but requests already using the old backend can return later and call `RecordFailure` in [client.Send](https://github.com/matrixorigin/matrixone/blob/35fc58bcb9dd6f7cab0a6ae0f988974a2afb762a/pkg/common/morpc/client.go#L507-L547). That recreates a breaker for the same remote after reset; enough late failures can open it and reject the freshly created backend.

The default breaker is disabled, but this is a lifecycle bug for configurations that enable it. The breaker state needs generation-aware invalidation, draining, or an equivalent way to ignore old-generation results.

Local verification: `go test ./pkg/common/morpc ./pkg/lockservice`, targeted MORPC/lockservice race tests, and `git diff --check` passed. The GitHub SCA check is currently failed and the main UT/E2E checks were still pending.

XuPeng-SH:
Blocking issue:

- `ResetBackend()` is still not actually bounded by its caller context on the cluster-discovery path. It acquires the global recovery-reset slot and then does both pre/post-refresh address lookups via `MOCluster.GetCNServiceWithoutWorkingState()` (`pkg/lockservice/rpc.go`), but the built-in cluster implementation waits on `readyC` without any context check (`pkg/clusterservice/cluster.go`). So if the cluster has not finished its first refresh yet, or is stuck behind that wait, one `ResetBackend` call can block indefinitely while holding `recoveryResetC`. Every later recovery attempt then times out behind it, and cleanup/shutdown can wedge.

Why I think this is a real bug:

- The PR already added the context-aware helper `GetCNServiceWithoutWorkingStateWithContext(...)` plus a cancellation test for exactly this startup-wait case in `pkg/clusterservice`.
- But `ResetBackend()` still calls the old context-insensitive method twice.
- Because the reset slot is acquired *before* those lookups, this is not just one slow request: it can serialize and starve all later reset attempts too.

Suggestion:

- Switch both `lookupAddress()` calls in `ResetBackend()` to `clusterservice.GetCNServiceWithoutWorkingStateWithContext(ctx, ...)`, so the whole reset path really respects `defaultRPCTimeout`.
- Add a lockservice-side regression where the cluster is not ready (or the built-in wait is blocked), and assert `ResetBackend()` returns on context deadline instead of hanging while holding the reset slot.

XuPeng-SH:
合理采纳并已修复，提交 7b185853c8。

根因是 ResetBackend 虽然为 refresh/reset gate 使用了 bounded context，但 pre/post-refresh 两次 snapshot lookup 仍调用 context-insensitive API；built-in cluster 未 ready 时会阻塞在 readyC，并持续占用全局 reset slot。

修复：
- 两次 lookup 都改用 GetCNServiceWithoutWorkingStateWithContext，并贯穿同一个 reset context；
- lookup 超时立即返回，且不会执行 DNS、backend close 或 recovery cache mutation；
- 新增真实 built-in cluster 首次 HAKeeper refresh 被阻塞的回归，验证 deadline 返回、零 backend 副作用，以及失败后全局 reset slot 已释放。

验证：新回归 count=50；相关 race count=10；完整 clusterservice/common-morpc/lockservice 测试、go vet、go build、git diff --check 均通过。分支已 rebase 到最新 main。

XuPeng-SH:
已系统性核验并合理采纳，修复提交 308a27b3d8。

采纳：
- RPC queue admission：快路径无额外 timer；仅队列满时进入有界等待，同时监听 request context、session context 和 5s queue deadline。超时通过正常 RPC error response 拒绝单个请求，不返回 handler error，避免关闭共享 MORPC session。
- 补齐 admission/shutdown ownership：所有拒绝分支 cancel + release pooled request；RPC server 成功停止后 drain 两个 buffered queue，避免遗留 request/CancelFunc。
- Recovery observability：新增低基数 active_txn_recovery_total 与 rpc_queue_rejected_total；ResetBackend 记录成功/失败、旧/新地址、解析 endpoint、耗时。未把 txn/service ID 放入 metric label。
- incarnation 前提：补充注释说明完整 lockservice identifier 包含进程启动版本，并依赖编排层先终止旧 listener 再发布 replacement。

Decision log：
- 部分 CloseBackendFor 失败残留：生产 MORPC 实现先从所有 backend/generation/breaker map 摘除，再关闭 detached backend，且 CloseBackendFor 本身不返回底层 close error；不存在失败后复用旧连接。ResetBackend 失败现已有 warning。
- 二次 Valid=false：完整 service identifier 含 incarnation；authoritative refresh + 单 IPv4 endpoint + fresh transport 后的 mismatch 表示旧 incarnation 已被替代，保留现有语义。
- ErrNotSupported fallback：server 未注册方法统一返回该错误码，契约闭合。
- recovery cache：4096 硬上限且 entry 仅为 hint，随机驱逐只导致重新 discovery/reset，不影响正确性，暂不引入 LRU。

验证：新增 queue/session/drain 测试 count=50；相关 race count=10；完整 clusterservice/common-morpc/lockservice/metric 测试、go vet、go build、git diff --check 全部通过。

XuPeng-SH:
Another blocker on the latest head:

- `ResetBackend()` now uses the context-aware cluster lookup, so the previous unbounded wait bug is fixed. But it still keeps `recoveryBackends[sid]` pinned to the **old resolved recovery IP** until *after* authoritative refresh and re-lookup succeed. If `Refresh(ctx)` fails, the function returns early and the stale pinned endpoint is left in place.

Why this matters:

- The whole point of the reset path is to stop later `CheckActiveTxn` / `GetActiveTxn` probes from reusing a known-stale endpoint after CN hot recreation.
- Right now, a transient control-plane refresh failure preserves exactly that stale override.
- On the next probe, `activeTxnBackend("cn-id", discoveredHostname)` can still route to the old cached `10.0.0.1:18101` instead of falling back to the hostname path that may already resolve to the replacement CN.
- `cleanCommitStateOnce()` will then keep probing the dead IP and can fence a healthy recreated CN after bounded failures.

There is even a test locking in the stale-pin behavior today: `TestResetBackendRejectsFailedDiscoveryRefresh` still expects `activeTxnBackend(...)` to return the old cached endpoint after reset failure.

Suggestion:

- Clear the cached recovery override before any early return in `ResetBackend()` (or otherwise force post-failure probes to fall back to the discovered hostname path instead of the old pinned endpoint).
- Update the refresh-failure test to assert that reset failure preserves “unknown state” **without** preserving the stale endpoint override.

XuPeng-SH:
合理采纳并已修复，提交 5e7326feb7。

这个 blocker 的本质是：refresh 失败必须保持业务状态 unknown，但 route invalidation 必须是单调的；已被负响应证明可疑的 endpoint 不能因为 control-plane 失败而重新变成可用 hint。

修复：
- ResetBackend 获取 reset slot 后，先原子删除 recoveryBackends[sid]，再执行任何可能失败的 discovery/refresh；
- 立即 detach 旧 cached discovered address 和 pinned endpoint；
- pre-refresh lookup 成功后，在 authoritative refresh 之前 detach 当前 hostname backend；
- authoritative refresh、二次 lookup 或 DNS 失败时返回 unknown/error，但不会恢复旧 override；
- 成功路径同时 detach refreshed hostname 和 newly-resolved endpoint，保证确认请求创建新的 MORPC generation；
- close error 与 discovery/refresh error 使用 errors.Join 保留完整失败信息。

测试更新：
- refresh failure：断言旧 hostname/IP 均被 detach，cache miss 回退 hostname，且仍返回 refresh error；
- startup lookup timeout：即使首次 snapshot 尚未 ready，旧 pin 也先失效；
- slow refresh：并发 route lookup 立即回退 hostname，不再暴露旧 IP；
- 覆盖首次 pin、地址变化、close 部分失败、新 resolved endpoint。

验证：ResetBackend count=50；相关 recovery race count=10；完整 clusterservice/common-morpc/lockservice/metric 测试、go vet、go build、git diff --check 均通过。分支已 rebase 到最新 main 78f14f98d7。

gouhongshen (approved):

XuPeng-SH:
补充：main 在上一轮推送后立即新增提交，已再次 rebase 到 b6b91bcf1c；修复提交的新 SHA 为 64cb51df9b。rebase 后 ResetBackend count=20、相关 race count=5 再次通过。

aptend:
已复审当前 head `64cb51df9b317aaf5438b261dd25183ee3b5d098`。

此前提出的 correctness/lifecycle blockers 均已闭环：peer `CheckActiveTxn` recovery、reset 失败保持 unknown、旧 generation create 隔离与有界回收、独立 active-txn transport、circuit-breaker incarnation 隔离、context-aware discovery，以及 stale recovery route 的单调失效。未发现新的 blocker。

本地验证通过：
- `go test ./pkg/common/morpc ./pkg/lockservice ./pkg/clusterservice -count=1`
- 关键 recovery/generation/reset race tests，`-race -count=3`
- `git diff --check`

结论：可以合入。

aptend (approved):
Re-reviewed current head 64cb51df9b317aaf5438b261dd25183ee3b5d098. Previous blocking findings are resolved, focused tests and race tests pass, and no new blockers were found.

Issue #22258, opened 2025-07-25
[Tech Request]: Storage optimization - main theme


aptend:
### Is there an existing issue for the same tech request?

- [x] I have checked the existing issues.

### Does this tech request not affect user experience?

- [x] This tech request doesn't affect user experience.

### What would you like to be added ?

```Markdown
This is a parent issue for holding various storage optimization pull requests.
```

aptend:
Move merge from DN to CN

aptend:
Ref: Tetris: Lightweight Hyperparameter Auto-Tuning for Mitigating Performance Spikes in LSM-KVS

PR #25653, opened 2026-07-13
fix(merge): handle objects without sort-key zonemaps

aptend:
## What type of PR is this?

- [x] BUG

Related to #22258

Tables without an explicit primary key use a fake primary key that is not a sort key. After flush, their object stats therefore have no initialized sort-key zonemap. This is valid: L0 compaction still merges these objects by size/count, but overlap statistics previously returned the misleading internal error `no valid intervals with initialized zonemaps`.

Treat an all-uninitialized sort-key-zonemap input as having no overlap intervals: retain the scan/uninitialized counts, return no overlap merge tasks, and avoid division by zero when detailed statistics are requested. This does not alter L0 compaction or overlap scheduling for objects with initialized zonemaps.

## Tests

- `go test ./pkg/vm/engine/tae/db/merge -count=1`
- `go test -race ./pkg/vm/engine/tae/db/merge -run TestOverlapStatsWithoutInitializedZoneMap -count=1`


XuPeng-SH (approved):
最新 head 看下来这个修复方向是对的，我这里没有剩余 blocker，可以 approve。

关键点已经收住了：
- all-uninitialized sort-key zonemap 不再报内部错误，而是明确落到“无 overlap interval / 无 merge task”；
- `CalculateOverlapStats` 和 `GatherOverlapMergeTasks` 的 nil/empty 分支都补齐了；
- `processed <= 0` 的 further-stat 防护也到位了，避免再掉到除零。

非阻塞建议：后面可以再顺手补两组小覆盖——多对象 all-uninitialized，以及 `constant + uninitialized` 命中 `processed <= 0` 的路径。但这些我不作为 blocker。

Issue #25606, opened 2026-07-11
[Bug] HAKeeper client panics when context has no deadline and default service is reachable


Ariznawlll:
## Summary

HAKeeper client construction can panic instead of returning an error when the caller passes a context without a deadline and the default HAKeeper address is reachable.

## Reproduction

- Commit: `7821d21c723372e45a8f65da3bbb896a157b1036`
- Start a local MatrixOne cluster with HAKeeper at the default `127.0.0.1:32000`.
- Run the existing test:

```bash
env GOROOT=/opt/homebrew/opt/go/libexec \
  PATH=/opt/homebrew/opt/go/libexec/bin:$PATH \
  DYLD_LIBRARY_PATH=$PWD/thirdparties/install/lib \
  CGO_CFLAGS=-I$PWD/thirdparties/install/include \
  CGO_LDFLAGS='-L'$PWD/thirdparties/install/lib' -lusearch_c' \
  go test ./pkg/logservice -run '^TestHAKeeperClientConfigIsValidated$' -count=1
```

The test calls `NewCNHAKeeperClient(context.TODO(), "", HAKeeperClientConfig{})`. `HAKeeperClientConfig.Validate` fills the default service address, the client connects successfully, and `pkg/common/morpc/future.go:57` panics with `context deadline not set` while sending `CHECK_HAKEEPER`.

## Expected

The constructor should reject a context without a deadline or apply a bounded internal timeout and return a normal error. A public client constructor should not panic based on whether the default service happens to be reachable.

## Relevant code

- `pkg/logservice/config.go:571-578` silently fills the default address for an empty config.
- `pkg/logservice/hakeeper_client.go:958-1004` sends the initial health check using the caller context.
- `pkg/common/morpc/future.go:55-58` panics when a request context has no deadline.

jiangxinmeng1:
fixed by #25654

Ariznawlll:
已在最新 `main` 上验证 HAKeeper client 对“无 deadline context”的处理。

### 验证版本

- Commit：`0977fc07c51067ce76b769c7853d7374cb2fccec`
- 验证时间：2026-07-14

### 验证方法

```bash
go test ./pkg/logservice \
  -run 'TestHAKeeperClient.*WithoutDeadline' -count=1 -v
```

构造可达默认服务，同时向 cluster/CN/TN/log/proxy constructors 及 client methods 传入没有 deadline 的 context。

### 实际结果

[...]

PR #25654, opened 2026-07-13
fix logservice hakeeper client context validation

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25606

Fix/hakeeper client context deadline


XuPeng-SH (changes requested):
Requesting changes.

The constructor-side validation is directionally right, but it only fixes the very first `CHECK_HAKEEPER` send. The same no-deadline panic hazard still exists on all later public HAKeeper client methods.

`pkg/common/morpc/future.go` still panics on any non-oneway send whose context has no deadline. After construction, methods like:
- `GetClusterDetails`
- `GetClusterState`
- `AllocateID` / `AllocateIDByKeyWithBatch`
- `Send*Heartbeat`
- `GetCNState`, `UpdateCN*`, `PatchCNStore`, `DeleteCNStore`, `GetBackupData`

all still pass the caller's `ctx` straight through to `hakeeperClient.request()` and then `c.client.Send(ctx, ...)`.

So with the latest head:
1. `New*HAKeeperClient(ctxWithDeadline, ...)` succeeds,
2. a later call such as `GetClusterDetails(context.Background())` or `AllocateID(context.Background())` can still panic in `morpc.Future.init()` instead of returning an error.

This means the original crash class is still present for normal post-construction client use; the PR currently fixes only the constructor entrypoint.

Suggested fix:
- enforce the same deadline validation on every exported per-call HAKeeper client method, or
- centralize it in `request()` / `prepareClient()` before any morpc send happens.

I would also add at least one regression test that constructs the client with a valid timeout context, then calls a real public method with `context.Background()` and asserts an error return rather than a panic.

XuPeng-SH (approved):
Latest head looks good to me.

The original crash class is now covered at the right boundary: not just constructors, but every exported per-call HAKeeper client method validates the caller context for a deadline before any morpc send path can begin. That closes the remaining normal-use panic path from post-construction calls like GetClusterDetails / AllocateID / heartbeat methods.

Issue #25631, opened 2026-07-11
[Bug]: ISCP retry returns nil when context is already canceled


Ariznawlll:
## Summary

The shared ISCP `retry` helper returns a nil error when its context is already canceled. The callback is not executed, but callers are told that the operation succeeded.

This helper is used by ISCP job registration/unregistration, job-spec updates, watermark/status flushing, worker iteration, replay, and lease checks.

## Version

Commit: `7f7de555c`

## Code path

`pkg/iscp/executor.go`:

```go
for i := 0; i < retryTimes; i++ {
    select {
    case <-ctx.Done():
        return
    default:
    }
    // ...
}
```

The function has a named `err` return value. If cancellation is observed before the first callback attempt, `err` is still nil.

The retry backoff also uses an unconditional `time.Sleep(interval)`, so cancellation during backoff cannot interrupt the wait.

## White-box reproduction

Add this test in package `iscp`:

```go
func TestRetryReturnsCanceledContext(t *testing.T) {
    ctx, cancel := context.WithCancel(context.Background())
    cancel()

    called := false
    err := retry(ctx, func() error {
        called = true
        return nil
    }, 3, time.Millisecond, time.Second)

    require.False(t, called)
    require.ErrorIs(t, err, context.Canceled)
}
```

Run:

```bash
go test ./pkg/iscp -run '^TestRetryReturnsCanceledContext$' -count=1
```

## Actual result

```text
Expected error with "context canceled" in chain but got nil.
```

The callback is not called, confirming that no operation was performed.

For example, `RegisterJob` consequently returns `ok=false, err=nil`; `CreateIndexCdcTask` interprets that combination as "CDC task already exists" even though registration was never attempted. Whether an enclosing SQL transaction subsequently aborts does not repair the helper's false-success contract, and background ISCP callers use the same helper.

## Expected result

- Return `ctx.Err()` when the context is canceled.
- Make the exponential backoff interruptible with a timer/select on `ctx.Done()`.
- Callers must never interpret cancellation as a successful no-op.

## Suggested verification

Cover cancellation before the first attempt, during an attempt, and during each backoff interval for both a direct `retry` call and the `RegisterJob`/`UnregisterJob` wrappers.

jiangxinmeng1:
fixed by #25655

Ariznawlll:
已在最新 `main` 上验证 ISCP `retry` 的 context cancellation 行为。

### 验证版本

[...]

PR #25655, opened 2026-07-13
fix: return context error from ISCP retry

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25631

  - return `ctx.Err()` from the shared ISCP retry helper when the context is
  already canceled
  - make retry backoff interruptible via timer/select on `ctx.Done()`
  - add regression coverage for cancellation before retry, during retry, during
  backoff, zero retry attempts, and RegisterJob/UnregisterJob wrappers


XuPeng-SH (approved):
LGTM after reviewing this from the retry-contract, cancellation, and unhappy-path angles.

- correctness: the shared retry helper now returns `ctx.Err()` when cancellation is already visible before any attempt starts, which closes the false-success hole from the issue
- cancellation behavior: the backoff wait is now interruptible through `ctx.Done()`, so callers no longer sit through an unconditional sleep after cancellation
- unhappy path coverage: the new tests cover cancellation before the first attempt, zero retry attempts, cancellation during an attempt, cancellation during backoff, and the RegisterJob/UnregisterJob wrappers that rely on the shared helper

Issue #25483, opened 2026-07-06
[Bug]: JSON_ARRAYAGG cannot be used as a window aggregate


Ariznawlll:
### Branch Name

main

### Commit ID

d2617529a

### Other Environment Information

- MySQL baseline: 8.0.45
- MatrixOne SQL endpoint used for validation: local latest-main build on `127.0.0.1:26001`

### Actual Behavior

MatrixOne supports `JSON_ARRAYAGG` as an aggregate function, but does not parse it as a window function.

MO result:

```text
ERROR 1064 (HY000): SQL parser error: ... syntax error ... near " over (partition by customer_id order by id) ..."
```

### Expected Behavior

MySQL 8.0 accepts `JSON_ARRAYAGG(...) OVER (...)`.

Example MySQL result for customer `10`:

```text
["paid"]
["paid", "paid"]
["paid", "paid", "refund"]
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS mysql_compat_window_json_arrayagg;
CREATE DATABASE mysql_compat_window_json_arrayagg;
USE mysql_compat_window_json_arrayagg;

CREATE TABLE orders (
  id int primary key,
  customer_id int,
  status varchar(16)
);
INSERT INTO orders VALUES
(1,10,'paid'),(2,10,'paid'),(3,10,'refund'),
(4,20,'paid'),(5,20,'paid'),(6,20,'pending'),
(7,30,'paid'),(8,30,'refund');

SELECT 'json_arrayagg' q, id,
       json_arrayagg(status) over (partition by customer_id order by id) v
FROM orders ORDER BY id;
```

### Additional information

This differs from `GROUP_CONCAT(...) OVER (...)`: MySQL rejects `GROUP_CONCAT` as a window function, while MO currently accepts it. That existing behavior appears related to closed issue #23485 and should be handled separately to avoid duplicate tracking.

Ariznawlll:
已在最新 `main` 上完成验证，并补充 motr 回归测试。

**验证环境**

- MatrixOne：`main@52afd25fa63a3cb6037b7bf8ff0bf1328c4793ec`
- motr 用例：`issue_25483_json_arrayagg_window`
- 回归 PR：matrixorigin/motr#48

**测试过程**

1. 复测 `JSON_ARRAYAGG(status) OVER (PARTITION BY customer_id ORDER BY id)`。
2. 使用多个 partition 检查分区隔离。
[...]

Issue #23485, opened 2026-01-08
[Bug]: gruop_concat use new window framework


iamlinjunhong:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

waitting

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

1. gruop_concat use new window framework
2. clean old code of window framework

### Steps to Reproduce

```Markdown
none
```

PR #25656, opened 2026-07-13
fix: support JSON_ARRAYAGG and JSON_OBJECTAGG as window functions

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25483

MySQL 8.0 accepts `JSON_ARRAYAGG(...) OVER (...)`, but MatrixOne rejected it with a SQL parser error. `JSON_ARRAYAGG` / `JSON_OBJECTAGG` were lexed as generic identifiers and parsed through `ID '(' ... ')'`, a rule that has no `window_spec_opt`, so the trailing `OVER (...)` failed to reduce.

**Parser**

- Promote `JSON_ARRAYAGG` / `JSON_OBJECTAGG` to non-reserved keyword tokens and add dedicated `function_call_aggregate` rules that carry `window_spec_opt` (single arg for `json_arrayagg`, two args for `json_objectagg`). They remain usable as ordinary identifiers/column names.

**Executor**

- Fixing only the parser exposed a second gap: `json_objectagg(...) OVER (...)` then failed at runtime with `json_objectagg needs exactly two arguments`. The window operator carried a single argument type per aggregate (`Window.Types`), dropping the second argument of two-argument aggregates. It now derives one type per argument from `GetArgExpressions()` when building the aggregate executor, mirroring the group operator, and the redundant `Window.Types` field is removed (it is never serialized into the remote pipeline). This also fixes multi-column `GROUP_CONCAT(a, b) OVER (...)`.

Example (customer `10`), now matching MySQL:

```
["paid"]
["paid", "paid"]
["paid", "paid", "refund"]
```

## Tests

- Parser unit tests for the window and plain-aggregate forms, plus identifier usage.
- Window-operator unit test covering the two-argument `json_objectagg` path (including no mpool leak).
- BVT: `test/distributed/cases/window/mysql_compat_window_json_arrayagg.sql` (16/16).

Note: `GROUP_CONCAT(...) OVER (...)` acceptance differs from MySQL (tracked separately in #23485) and is intentionally out of scope here.


heni02 (approved):

XuPeng-SH (approved):
LGTM after reviewing this from the parser, executor, and unhappy-path angles.

- parser correctness: `JSON_ARRAYAGG` / `JSON_OBJECTAGG` now go through aggregate-function rules that carry `window_spec_opt`, while still remaining usable as ordinary identifiers
- runtime wiring: the window operator now derives one type per aggregate argument from `GetArgExpressions()`, which fixes the real multi-argument executor gap instead of only masking the parser error
- unhappy-path coverage: the new window tests cover the two-argument `json_objectagg` path and the NULL-key error cleanup path, and the BVT covers window + plain aggregate + identifier usage

Issue #25550, opened 2026-07-09
[Bug]: TestS3FSIOMerger flaky — ioMerger.Merge timing assertion fails on CI runner


matrix-meow:
## Problem

`TestS3FSIOMerger` in `pkg/fileservice/s3_fs_test.go:1428` fails intermittently on CI:

```
s3_fs_test.go:1428: Error: Should be true
```

The test verifies IO operations are properly merged by the ioMerger. On CI runners under load, `ioMerger.Merge` takes ~200ms (vs expected fast merge), causing the merge timing assertion to fail.

Slow event trace shows:
```
<204.03899ms ioMerger.Merge begin>
<204.041955ms ioMerger.Merge wait>
<428.383411ms ioMerger.Merge end>
```

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-09 | [PR #25549](https://github.com/matrixorigin/matrixone/actions/runs/28992177037/job/86034283759) | UT Test on Ubuntu/x86 |

## Root Cause

The test relies on timing-sensitive IO merge behavior. Under CI runner load, disk cache wait/update operations take longer (~200ms), causing concurrent IO requests to not merge as expected, or the merge assertion to evaluate before merge completes.

## Suggested Fix

1. Increase timing tolerance in the merge assertion
2. Or use synchronization primitives instead of timing-based assertions
3. Or skip the test on CI when runner load is high

LeftHandCold:
fixed

heni02:
confirm.closed

Issue #25603, opened 2026-07-11
[Bug] disk object storage leaks temp files and hides Delete errors


Ariznawlll:
## Summary

Disk object storage has two error-handling problems: failed writes can leave `.mofstemp` files behind, and `Delete` swallows removal errors.

## Evidence

- Commit: `7821d21c723372e45a8f65da3bbb896a157b1036`
- `pkg/fileservice/disk_object_storage.go:245-291` creates a `*.mofstemp` file and returns directly on copy, size-check, sync, close, mkdir, or rename errors without consistently removing the temporary file.
- A temporary test using a reader that returns an error left a `.mofstemp` file in the object directory.
- `pkg/fileservice/disk_object_storage.go:72-86` ignores every `os.Remove` error and returns nil. Deleting a non-empty directory therefore reports success while the object remains.

## Expected

Failed writes should close and remove temporary files, and `Delete` should preserve idempotent-not-found semantics while returning other removal errors such as `ENOTEMPTY` or permission failures.

LeftHandCold:
fixed

Ariznawlll:
## 最新 main 详细验证

- **源码基线**：`origin/main@f1c0a92d882dbdad469ffd2c96f607796207c797`。
- **执行命令**：

```bash
go test ./pkg/fileservice \
  -run '^Test(DiskObjectStorageWriteFailureRemovesTempFile|DiskObjectStorageSizeMismatchRemovesTempFile|DiskObjectStorageDeleteReturnsRemovalError|LocalFSStatFileReturnsNonNotExistError)$' \
  -count=3 -v
```

### 本 issue 对应测试过程

1. 用 reader 注入写入错误，断言原始错误透传、`.mofstemp` 被删除、目标对象不存在。
2. 再构造声明 size=1、实际 0 字节的 mismatch，断言写入失败且临时文件同样被删除。
3. 构造无法删除的非空目录，断言 `Delete` 返回 removal error；同时断言后续合法对象仍被处理、失败对象没有被误删。

### 结果

- 三条 DiskObjectStorage 错误路径均 **PASS 3/3**。
- package：`ok github.com/matrixorigin/matrixone/pkg/fileservice 0.461s`。
- **结论**：临时文件清理和删除错误传播在最新 main 定向回归通过。

PR #25659, opened 2026-07-13
fix(fileservice): preserve errors and cache merge ordering

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25550
issue #25602
issue #25603

This change closes three related fileservice error-path gaps:

- delay the S3 I/O-merger completion signal until deferred cache updates finish, so waiters observe the leader's cached data instead of issuing duplicate object reads;
- return non-`ENOENT` errors from `LocalFS.StatFile` instead of dereferencing a nil `FileInfo` after `os.Stat` fails;
- close and remove `.mofstemp` files on every failed disk-object write, propagate cleanup failures, and report real delete failures while keeping missing-object deletes idempotent.

The merger regression test now uses a blocked leader and an observed waiter handshake, making the original ordering race deterministic. Disk and local filesystem tests cover copy failure, size mismatch, delete failure, and `ENOTDIR`.

Tested with:

`go test -count=1 ./pkg/fileservice`

`go test -race -count=20 ./pkg/fileservice -run '^(TestLocalFSStatFileReturnsNonNotExistError|TestDiskObjectStorageWriteFailureRemovesTempFile|TestDiskObjectStorageSizeMismatchRemovesTempFile|TestDiskObjectStorageDeleteReturnsRemovalError|TestS3FSIOMerger)$'`

`go vet ./pkg/fileservice`


XuPeng-SH (changes requested):
The production fixes look sound, but the S3 merger regression test does not make the old ordering race deterministic as claimed. The waiter handshake only proves that the second read has entered ioMerger.Merge; after done() closes the channel, the waiter is merely runnable, and the leader can still execute the cache-update defers before the scheduler runs it. I kept the new test unchanged, reverted only finishMerge back to the old defer done(), and the test still passed 100/100 times with GOMAXPROCS=1 and 100/100 times with GOMAXPROCS=8. Therefore this test is green on the buggy ordering and does not protect the fix. Please add a controlled barrier/hook around cache update versus merger completion (or an equivalent deterministic mechanism), and verify that the test is reliably red with the old ordering and green with the new ordering. The disk-object and LocalFS error-path changes and tests otherwise look good.

fengttt (approved):

XuPeng-SH (approved):
Approved. The previous review blocker is resolved.

The rewritten S3 merger regression is now deterministic: on the current head it passed 20/20 under the race detector; after temporarily restoring only the old `defer done()` ordering, it failed 20/20 at the `IsMerging` assertion while the cache update was blocked. Restoring the PR implementation returned it to green. The production defer order keeps the merge active through synchronous cache updates and releases waiters on every return/error path.

The LocalFS non-ENOENT handling is correct. Disk-object writes now close and remove the temp file across copy, size, sync, close, mkdir, and rename failures while preserving cleanup errors; Delete remains idempotent for missing keys and aggregates real removal failures.

Verified locally: full `pkg/fileservice` tests, targeted race tests with `-count=20`, `go build`, `go vet`, and `git diff --check`. All GitHub CI checks are green.

Issue #25580, opened 2026-07-10
[Bug]: possible stale AUTO_INCREMENT allocation after bulk LOAD DATA in multi-CN TPCC


XuPeng-SH:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

Historical related issues:

- #9222 reported `bmsql_history.hist_id` duplicate keys after TPCC bulk loading and benchmark execution.
- #16493 fixed a previous multi-CN incrservice cache issue. This report may be a regression or a different bulk-load/high-water propagation path.

### Branch Name

`refactor_schedule_phase7_0710` (the symptom was observed while validating this branch, but the changed scheduler code is not currently suspected).

### Commit ID

`51c1de768d85ae3cbff49a3d04fa1f2eaec585f4`

### Other Environment Information

```Markdown
- MatrixOne version: 8.0.30-MatrixOne-v1.3.0 (local dev build)
- Topology: mo-cn1 + mo-cn2 + mo-tn + mo-proxy
- SQL endpoint: proxy at 127.0.0.1:6001; CN SQL endpoints at 127.0.0.1:16001 and 127.0.0.1:16002
- OS: Linux 6.12.88+deb13-amd64 x86_64
- TPCC workload: 50 warehouses, 120 terminals, `terminalWarehouseFixed=true`
- Chaos: not enabled when the duplicate-key errors occurred
```

### Actual Behavior

After creating the TPCC schema and importing the 50-warehouse CSV data with server-side `LOAD DATA`, `runBenchmark.sh props50.mo` immediately reported duplicate primary keys for `bmsql_history.hist_id`.

`bmsql_history.hist_id` is `INTEGER AUTO_INCREMENT PRIMARY KEY`, and runtime Payment inserts omit `hist_id`.

```text
2026-07-10 18:53:01 FATAL jTPCCTerminal - [UNEXPECTED][TT_PAYMENT][EXECUTION] ErrorCode : 1062, ErrorMessage : Duplicate entry '5413' for key 'hist_id'
2026-07-10 18:53:01 FATAL jTPCCTerminal - [UNEXPECTED][TT_PAYMENT][EXECUTION] ErrorCode : 1062, ErrorMessage : Duplicate entry '5414' for key 'hist_id'
```

The imported `cust-hist.csv` has explicit IDs `1..1500000`. The first benchmark run produced repeated duplicates before any network chaos was injected.

[...]

Issue #25626, opened 2026-07-11
[Bug]: UPDATE leaves composite CLUSTER BY hidden key stale


Ariznawlll:
### Existing issue check

- [x] Searched open/closed issues and pull requests; no duplicate found.
- Distinct from #25617, which covers primary-key changes leaving FULLTEXT/IVF origin keys stale. This issue needs no primary key or user index and corrupts the internal composite CLUSTER BY value itself.

### Branch / commit / environment

- Branch: `test/issues-24122-21685-4.0-dev`
- Commit: `7f7de555c49acc3a64995b078025f5526afc9c34`
- Local launch-with-proxy, two CNs, macOS arm64

### Summary

UPDATE changes the visible columns of a multi-column `CLUSTER BY`, but the persisted hidden composite key (`__mo_cbkey...`) retains the old encoded values.

This produces silent false positives and false negatives:

- the old equality predicate returns the row while projecting its new values;
- the correct new equality predicate returns no row;
- IN and BETWEEN on the new cluster values also miss the row;
- DELETE by either the new or old equality key affects 0 rows in the minimal case;
- an equivalent expression that prevents compound rewrite (`a+0=...`) finds the row;
- ORDER BY/LIMIT reads the visible columns and places the updated values correctly.

The mismatch survives TN flush and a complete CN/TN/logservice restart.

### Minimal reproduction

```sql
DROP DATABASE IF EXISTS cluster_by_update_stale;
CREATE DATABASE cluster_by_update_stale;
USE cluster_by_update_stale;

CREATE TABLE t(a INT,b INT,payload VARCHAR(20)) CLUSTER BY(a,b);
INSERT INTO t VALUES(1,10,'x'),(2,20,'y');

UPDATE t SET a=9,b=90 WHERE a=1 AND b=10;
-- Query OK, 1 row affected

SELECT * FROM t WHERE a=1 AND b=10;
-- actual: (9,90,'x'); expected: empty

SELECT * FROM t WHERE a=9 AND b=90;
-- actual: empty; expected: (9,90,'x')

SELECT * FROM t WHERE a+0=9 AND b+0=90;
-- correct control: (9,90,'x')

DELETE FROM t WHERE a=9 AND b=90;
SELECT ROW_COUNT();
-- actual: 0; expected: 1

DELETE FROM t WHERE a=1 AND b=10;
SELECT ROW_COUNT();
-- actual: 0; row remains present
```

[...]

PR #25661, opened 2026-07-13
fix(dml): keep generated keys consistent

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25580
issue #25626

This change fixes two generated-key consistency bugs in DML paths:

- auto-increment caches now retain the allocation timestamp for each cached range. Conflict detection uses the timestamp of the oldest range that can still issue a value and advances only when that range is exhausted. Prefetching a newer range can therefore no longer hide explicit inserts that conflict with values remaining in an older range;
- `UPDATE` now rebuilds the hidden composite `CLUSTER BY` key with `serial_full` whenever any component changes. The stored hidden key stays aligned with the updated visible columns, so equality/range pruning, delete, indexes, flush, and restart all address the updated row correctly.

Unit tests cover prefetch timestamp advancement and every range mutation that can discard or split cached ranges, plus first/last/all composite cluster-key component updates. A distributed BVT covers equality, `IN`, range and ordered reads, TN flush, delete, secondary/unique indexes, expression updates, and prepared updates.

Tested with:

`go test -count=1 ./pkg/incrservice ./pkg/sql/plan`

`go test -race -count=20 ./pkg/incrservice ./pkg/sql/plan -run '^(TestOldestAllocateAtFollowsConsumableRange|TestLastAllocateAtEmptyInitial|TestRangeAllocateTimestampTracksMutations|TestUpdateRecomputesCompositeClusterByKey)$'`

`go vet ./pkg/incrservice ./pkg/sql/plan ./pkg/sql/colexec/preinsert`

The composite `CLUSTER BY` regression was also verified against a patched quickstart service before and after TN flush and a full service restart.


XuPeng-SH (approved):

On the incrservice side, the important fix is that allocation timestamps now follow the consumable ranges themselves across every relevant mutation path (add, next, setManual, updateTo), so conflict detection starts from the oldest range that can still issue a value instead of being narrowed incorrectly by a newer prefetch.

On the UPDATE side, the hidden composite CLUSTER BY key is now rebuilt from the final projected component expressions before writeback, which closes the stale-old-key / missing-new-key failure mode from #25626. The planner and BVT coverage also hit first/last/all component updates plus equality/IN/range/delete/prepared-update paths.

pkg/incrservice/range_cache.go
```diff
@@ -74,48 +78,70 @@ func (r *ranges) left() int {
 func (r *ranges) setManual(
 	value uint64,
 	skipped *ranges) {
-	newValues := r.values[:0]
+	newValues := make([]uint64, 0, len(r.values))
```
gouhongshen: [P1/blocker] `insertAutoValues` 会对每个 distinct explicit value 调用一次 `setManual`（`pkg/incrservice/column_cache.go:541-543`）。这里把原来的 `r.values[:0]` 改成了每次 `make` 新 backing array；`newAllocatedAt` 又额外制造一次分配。对于 #25580 中 150 万条显式 ID 的 LOAD DATA，在预取范围覆盖输入时，这会产生百万级短命对象和 O(rows) 的 GC/内存压力，可能直接拖慢或耗尽 CN。请复用 `r.values`/`r.allocatedAt` 的 backing storage，或把重建延迟为每个 batch 一次，同时保持 range 与 timestamp 的一一对应。

gouhongshen (changes requested):
REQUEST_CHANGES

发现 1 个 blocker：这个改动会把 bulk LOAD DATA 的显式值路径从复用 backing array 改成逐值分配，可能造成严重的 GC/内存回归。详见 inline comment。除此之外，本次源码审查未发现另一个足以阻断合并的高置信度正确性问题。

验证说明：本机无法运行 CGo 依赖测试；仓库所需的 xxhash.h、roaring.h、usearch.h 与 thirdparties/install 未安装。该 blocker 基于调用链和分配行为直接成立，不依赖测试环境。

gouhongshen (approved):
Follow-up review on 3430c63a3bb1d48b187bb84aac4f569bddb5be9c:\n\nThe previous P1 is fixed: setManual now reuses both the range-value and allocation-timestamp backing storage, preserving the one-timestamp-per-range invariant. I rechecked the complete PR diff and the incrservice / UPDATE cluster-by call paths; I found no remaining high-confidence correctness, concurrency, resource, performance, or compatibility blocker.\n\nDiff check and gofmt are clean. The local CGo-transitive tests remain unavailable because this environment lacks the repository CGo prerequisites; GitHub CI run 29309221484 is still in progress.\n\nAPPROVE

aunjgr (changes requested):
I confirm the current-head allocation blocker already identified at `range_cache.go:81`. `insertAutoValues` calls `setManual` once per distinct explicit value, while `setManual` now allocates fresh `values` and `allocatedAt` backing arrays on every call. Across the 1.5M explicit IDs from #25580 this creates O(rows) short-lived allocations and severe avoidable GC pressure. Preserve the timestamp alignment while compacting both slices in place (normalize timestamps first if needed, then reuse `r.values[:0]` and `r.allocatedAt[:0]`) and add an allocation regression. I found no second high-confidence correctness blocker in the range-timestamp or composite cluster-key changes. The current SCA failure is checkout TLS infrastructure failure, not a source failure.

aunjgr (approved):
Follow-up review on current head `3430c63a`: the prior P1 allocation regression is fixed. `setManual` now compacts `values` and `allocatedAt` in place; since each input range emits at most one surviving range, the write cursor cannot overtake unread input, and the one-timestamp-per-range invariant remains intact. The new `AllocsPerRun` regression directly verifies backing-storage reuse. I rechecked the full range-timestamp and composite CLUSTER BY changes and found no remaining high-confidence correctness, concurrency, lifecycle, or performance blocker. Current UT and BVT checks pass. The SCA and Coverage failures are TLS/checkout/install infrastructure failures, not source failures; remaining checks are still running.

Issue #25604, opened 2026-07-11
[Bug] CompleteSQLTaskRun accepts stale runner completion


Ariznawlll:
## Summary

`CompleteSQLTaskRun` updates a task-run row using only `run_id`. A stale runner can therefore complete a run after that run has timed out/reallocated and overwrite the newer state.

## Evidence

- Commit: `7821d21c723372e45a8f65da3bbb896a157b1036`
- `pkg/taskservice/mysql_task_storage.go:264-279` executes the update with `WHERE run_id=?` only.
- `pkg/taskservice/sql_task_executor.go:224-237` calls `CompleteSQLTaskRun` after execution, without a lease/attempt/version predicate.
- The in-memory storage implementation has the same unconditional RunID overwrite behavior (`pkg/taskservice/mem_task_storage.go:421-429`).

## Expected

Completion should be conditional on the runner's current lease/attempt/version, so a stale runner cannot overwrite a newer allocation or completion.

Please add a test that reallocates/times out a run while the old runner is still executing and verifies that the old completion is rejected.

LeftHandCold:
fixed

Ariznawlll:
已在最新 `main` 上验证 SQL task stale completion fencing。

### 验证版本

- Commit：`0977fc07c51067ce76b769c7853d7374cb2fccec`
- 验证时间：2026-07-14

### 验证方法

```bash
go test ./pkg/taskservice \
  -run '^TestCompleteSQLTaskRunRejectsReapedRun$' -count=1 -v
```

测试先让旧 runner 的 task run 被 reaper 回收/重新分配，再模拟旧 runner 上报完成，检查 storage 是否会错误接受旧 run 的 completion。

### 实际结果

memory storage 与 refresh 路径均拒绝 stale/reaped run completion，两个子用例全部 PASS。

### 结论

当前 main 会根据 run identity/version 对完成请求进行 fencing，旧 runner 无法覆盖新 runner 的 task 状态。

![#25604 verification](https://gist.githubusercontent.com/Ariznawlll/a2432edd0696ba6684e345e80bf49320/raw/25604.png)

PR #25662, opened 2026-07-13
fix(taskservice): fence stale SQL task completions

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25604

A timed-out SQL task run can be reaped while its original worker is still finishing. The old completion path unconditionally updated the row by `run_id`, allowing the late worker to overwrite the terminal `TIMEOUT` state with `SUCCESS` or `FAILED`.

This change makes completion a fenced state transition. MySQL only updates a row that is still `RUNNING` and still belongs to the same runner and attempt; the memory storage follows the same contract. Reaper updates remain unconditional so they can continue to mark stale runs atomically under the existing task lock.

The shared storage regression test reaps a stale run, verifies that its late completion affects zero rows and preserves `TIMEOUT`, then verifies that the current attempt still completes normally.

Tested with:

`go test -count=1 ./pkg/taskservice`

`go test -race -count=10 ./pkg/taskservice -run '^(TestCompleteSQLTaskRunRejectsReapedRun|TestAcquireAndCompleteSQLTaskRun|TestAcquireAndCompleteSQLTaskRunInSqlMock|TestSQLTaskMySQLStorageErrorBranches)$'`

`go vet ./pkg/taskservice`


XuPeng-SH (approved):
LGTM after reviewing this from the state-machine, concurrency, and unhappy-path angles.

- correctness: completion is now fenced to the same `RUNNING` row plus the same runner/attempt, so a late stale worker can no longer overwrite a reaped terminal state
- symmetry: MySQL and memory storage now follow the same completion contract, while reaper updates remain intentionally unconditional under the existing task lock
- unhappy path coverage: the new regression proves stale completion preserves `TIMEOUT` and does not block the current attempt from completing normally

I did not find a remaining correctness, hang, or state-corruption issue in the latest diff.

PR #25664, opened 2026-07-13
fix: register local dispatches below remote scopes

### LINKED ISSUE #25211: Hive partition JOIN queries intermittently fail with dispatch timeout (dispatch process not registered within 30s)

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25211

Fixes #25211

The early dispatch-receiver registration pass stops at a Remote scope when that tree can execute independently on another CN and its root is not a Dispatch. Such a remote tree can still contain a child Remote scope that returns execution to the coordinator CN. Its local dispatch receiver is therefore never published, and the remote sender retries `TryGetProcByUuid` until the 30-second timeout.

Keep the outer remote operator tree remote-only, but continue walking its `PreScopes`. This publishes only dispatch receivers that will execute locally after a nested return RemoteRun. Retained remote Dispatch roots also continue to their child scopes.

Regression coverage exercises both a remote non-Dispatch root and a retained remote Dispatch root with a nested local-return scope.

## Tests:

- `go test -race ./pkg/sql/compile -run "^TestRegisterLocalDispatchReceiversTraversesRemoteAncestorForNestedLocalReturn$" -count=1`
- `go test ./pkg/sql/compile -count=1`

### DIFF
--- a/pkg/sql/compile/remoterunServer.go
+++ b/pkg/sql/compile/remoterunServer.go
@@ -292,7 +292,10 @@ func registerLocalDispatchReceivers(scopes []*Scope, localAddr string) (*remoteD
 		if _, ok := s.RootOp.(*dispatch.Dispatch); ok {
 			return registerDispatchReceiverRoot
 		}
-		return skipDispatchReceivers
+		// The tree runs on another CN, so its own operators must not be
+		// registered here. Its descendants can still RemoteRun back to this CN;
+		// visit them to publish those local return dispatches early.
+		return traverseDispatchReceiverChildren
 	})
 }
 
@@ -302,6 +305,7 @@ const (
 	skipDispatchReceivers dispatchReceiverRegistrationMode = iota
 	registerDispatchReceiverRoot
 	registerDispatchReceiverTree
+	traverseDispatchReceiverChildren
 )
 
 func registerDispatchReceivers(
@@ -348,10 +352,10 @@ func registerDispatchReceivers(
 		}
 
 		if mode == registerDispatchReceiverRoot {
-			return registerDispatch(s.RootOp.(*dispatch.Dispatch))
-		}
-
-		if s.RootOp != nil {
+			if err := registerDispatch(s.RootOp.(*dispatch.Dispatch)); err != nil {
+				return err
+			}
+		} else if mode == registerDispatchReceiverTree && s.RootOp != nil {
 			if err := vm.HandleAllOp(s.RootOp, func(_ vm.Operator, op vm.Operator) error {
 				d, ok := op.(*dispatch.Dispatch)
 				if !ok {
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
LGTM after reviewing this from the execution-topology, registration-lifecycle, and unhappy-path angles.

- correctness: the local registration pass now keeps walking `PreScopes` below a standalone remote non-Dispatch ancestor, which is the missing path behind the 30s `TryGetProcByUuid` timeout
- containment: the change still avoids registering the remote-only operator tree itself, and retained remote Dispatch roots keep the old root-only behavior while also reaching nested local-return children
- unhappy path coverage: the new regression covers both the remote non-Dispatch root and the retained remote Dispatch root cases, which are the two important nested-return shapes here

PR #25665, opened 2026-07-13
fix: retain table publication during alter copy

### LINKED ISSUE #25601: [Bug] table-level publication loses its table after source ADD COLUMN
## Summary

A table-level publication can lose its published table after a source `ADD COLUMN` DDL. The subscriber then reports that the table does not exist and `SHOW CREATE PUBLICATION` shows an empty table list.

## Reproduction

At commit `7821d21c723372e45a8f65da3bbb896a157b1036`:

1. Create source database `pubmin_src`, table `part_t`, account `pubmin_acc`, and a table-level publication for `part_t`.
2. Create a subscriber and verify the table is readable.
3. Execute `ALTER TABLE pubmin_src.part_t ADD COLUMN extra INT DEFAULT 9`.
4. Query the subscriber and inspect `SHOW CREATE PUBLICATION pubmin_part`.

Observed: the subscriber returns `table "part_t" does not exist`, and the publication definition contains an empty `TABLE` clause. This was reproduced with the running two-CN publication setup.

## Expected

The publication should continue to include the source table after compatible DDL, and the subscriber should retain the table with the new column.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25601

`ALTER TABLE` using the COPY algorithm internally drops and renames the source table. The internal drop previously updated table-level publication metadata as though the user had issued `DROP TABLE`, removing the published table and making existing subscriptions unusable.

Mark that internal drop with `WithIgnorePublish()` so publication metadata is preserved during the replacement. Add a distributed regression that verifies the publication definition and subscriber data after `ADD COLUMN`.

Validation:

- `go test ./pkg/sql/compile -run 'TestScope_AlterTableCopy|TestScope_AlterTableInplace' -count=1`
- Local two-CN reproduction: after `ADD COLUMN`, the publication retains its table and the subscriber reads the new column.


XuPeng-SH (approved):
LGTM after reviewing this from the internal-DDL semantics, publication-metadata retention, and unhappy-path angles.

- correctness: the fix is on the right boundary — only the internal DROP in the ALTER COPY replacement path gets `WithIgnorePublish()`, so the publication metadata is preserved for the user-visible table while the copy/rename swap still proceeds normally
- consistency: the same ignore-publish pattern is already used in other internal replacement flows (for example truncate-style drop/create), so this change aligns ALTER COPY with existing internal-DDL semantics instead of inventing a new path
- unhappy path coverage: the new distributed regression checks both the publication definition and the subscriber-visible table after `ADD COLUMN`, which is the exact broken behavior from the issue

PR #25666, opened 2026-07-13
Fix index consumer deadlock

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25632

 Fixes a permanent deadlock in `IndexConsumer.Consume` when the downstream
  index SQL executor returns an error.

  Previously, the consumer goroutine could report the executor error and exit
  while the producer continued generating CDC SQL and attempted to send to the
  unbuffered `sqlBufSendCh`. With no receiver left, the producer could block
  forever and the original executor error would never be returned.

  This change:
  - makes producer sends select on downstream errors and context cancellation
  - preserves final `flushCdc` errors from `sinkTail`
  - reports ISCP index consumer errors through a non-blocking helper
  - applies the same non-blocking error reporting pattern to the CUVS runner
  - adds ISCP index failpoints for executor/send/watermark/close/HNSW/CUVS paths
  - adds a dev-cluster chaos script for repeatable ISCP index failure testing

### Unit / Package Tests

   Area                Case                Expected Result     Status
  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━
   IndexConsumer       SQL executor        Consume returns     Passed
   executor failure    fails after         the original
                       first batch         executor error
                                           promptly;
                                           producer does
                                           not deadlock
  ──────────────────  ──────────────────  ──────────────────  ──────────────────
   IndexConsumer       flushCdc fails      final flush         Passed
   tail final flush    at end of           error is
                       sinkTail            returned instead
                                           of discarded
  ──────────────────  ──────────────────  ──────────────────  ──────────────────
   IndexConsumer       producer            sendSql exits       Passed
   blocked send        attempts to send    via ctx.Done() /
                       while context is    errch; no
                       canceled /          permanent block
                       consumer is gone
  ──────────────────  ──────────────────  ──────────────────  ──────────────────
   CUVS runner         CUVS factory/       errors are          Passed by
   error reporting     save/append         reported through    package tests /
                       errors              non-blocking        compile path
                                           reportIndexConsu
                                           merErr

  ### Dev Cluster Failpoint Tests

  Run:

  make dev-build TYPECHECK=0
  make dev-up
  TIMEOUT_SECONDS=120 CASE=all-cpu etc/docker-multi-cn-local-disk/iscp-chaos-
  test.sh

  | Flow | Fault Point | Workload | Status |
  |---|---|---|---|
  | IVF-FLAT snapshot first batch | `fj/iscp/index/exec/error` | async IVF index
  over existing rows | Passed |
  | IVF-FLAT snapshot second batch | `fj/iscp/index/exec/error` | async IVF
  index over 8300 rows to force multiple flushes | Passed |
  | IVF-FLAT tail first batch | `fj/iscp/index/exec/error` | create async IVF
  index, then insert tail rows | Passed |
  | IVF-FLAT tail second batch | `fj/iscp/index/exec/error` | create async IVF
  index, then insert 8300 tail rows | Passed |
  | FULLTEXT snapshot | `fj/iscp/index/exec/error` | async FULLTEXT index over
  existing rows | Passed |
  | FULLTEXT tail | `fj/iscp/index/exec/error` | create async FULLTEXT index,
  then insert/delete tail rows | Passed |
  | Producer send slow path | `fj/iscp/index/send/block` | injected sleep before
  channel send | Passed |
  | Channel close slow path | `fj/iscp/index/close/block` | injected sleep
  before closing `sqlBufSendCh` | Passed |
  | HNSW update | `fj/iscp/index/hnsw/update/error` | create async HNSW index,
  then insert tail row | Passed |


XuPeng-SH (changes requested):
Blocking issue:

- In the tail path, `sendSql` can now stop the producer via `errch` / `ctx.Done`, but the consumer side still handles `case e2 := <-errch: errch <- e2; return` as a successful return from the long-lived `RunTxnWithSqlContext` callback (`pkg/iscp/index_consumer.go`). `RunTxnWithSqlContext` commits on `nil` (`pkg/vectorindex/sqlexec/sqlexec.go:360-365`), so a producer-side failure after some tail SQL has already been consumed will commit partial hidden-table changes while skipping `UpdateWatermark()`. That trades the deadlock for a correctness hole: the next retry replays from the old watermark against already-committed partial index state.

Suggestion:

- On the tail runner's `errch` branch, propagate the error (or a rollback sentinel) instead of bare-returning, so the transaction rolls back together with the missing watermark update.
- Add a regression test that forces a later `sendSql` failure/cancel after at least one tail SQL has already been executed, then asserts rollback + no watermark advance.

XuPeng-SH:
Another blocker:

- `processISCPData` still reports a terminal `flushCdc` error and then unconditionally closes `sqlBufSendCh` in both the `data == nil` and `noMoreData` paths (`pkg/iscp/index_consumer.go`). That means the worker races between reading `errch` and observing a clean EOF. If EOF wins, `RunHnsw` / `RunCuvs` go down their normal close path and call `Save()` (and tail watermark finalization) as if the final CDC chunk had been delivered. A final flush/send failure can therefore persist an incomplete index state instead of aborting.

Suggestion:

- Only close `sqlBufSendCh` after the terminal `flushCdc` succeeds. If the final flush fails, signal abort and let the worker exit via the error path rather than treating it as clean EOF.
- Add a regression test that forces the last flush to fail in the snapshot path and asserts the runner does not execute `Save()` / finalization on the partial state.

XuPeng-SH (changes requested):
Thanks for the follow-up fixes. The earlier partial-tail-commit issue and the terminal-flush/clean-EOF race are fixed at this head. Two blocking lifecycle issues remain:

1. A downstream failure still cannot interrupt the producer once it is blocked in DataRetriever.Next(). Consume calls r.Next() synchronously and only observes errch from sendSql. A concrete sequence is: the producer sends one chunk, the worker receives it and fails, the producer finishes that batch and enters the next Next(), and the upstream then stalls. The worker error remains buffered but Consume cannot return. In the production wiring, DataRetriever.SetError is called only after Consume returns, and the iteration cancel happens only after waiting for all consumers, so neither path breaks the cycle. I reproduced this 10/10 with a retriever that returns one batch and then blocks: Consume returned only after manually releasing Next(). Please make Next context/terminal-state aware (or otherwise connect worker failure to retriever cancellation) and add a regression test proving a blocked Next exits promptly after the worker fails.

2. Context cancellation can still be reported as success. The snapshot, HNSW, and CUVS runners return silently on ctx.Done(); when the writer is empty, no sendSql call propagates ctx.Err(), errch stays empty, and Consume returns nil. A canceled-context counterexample returned nil 10/10. RunTxnWithSqlContext also derives its transaction context from context.Background() rather than the supplied ctx, so in-flight SQL/Save work is not canceled by the caller. Please propagate ctx.Err() from every runner and derive transaction contexts from the parent context, with tests for cancellation before data, while blocked, and during in-flight execution/finalization.

There is also an adjacent ownership leak that should be closed or tracked explicitly: iteration increments ISCPData refcount for every consumer, but DataRetrieverImpl.SetNextBatch returns without data.Done() when that consumer has failed or its context is canceled. With multiple consumers, one failed index can leave every subsequent shared AtomicBatch unclosed while the other consumers continue scanning, causing mpool growth.

The latest package itself passes go test -race ./pkg/iscp; the failures above require the missing unhappy-path counterexamples.

XuPeng-SH:
Another blocker on the latest head:

- The new parent-context cancellation wiring in `sqlexec.RunTxnWithSqlContext()` is not actually connected to the production iteration path. The updated tests prove that in-flight snapshot exec / tail finalization are cancelable **when `Consume()` receives a cancelable parent context**, but `runISCPTaskIteration` still calls each consumer with `context.Background()` plus a tenant value (`pkg/iscp/iteration.go`), not with the iteration's cancelable context.

That means iteration cancellation and sibling-consumer failure still do **not** propagate into an in-flight `ExecWithResult` / `UpdateWatermark` / commit path through the parent context. `DataRetrieverImpl.SetError()` can unblock `Next()`, and `errch` can stop the runner once it returns to its select, but if the worker is already inside SQL execution the current operation still runs until it finishes on its own. So the new `RunTxnWithSqlContext` fix is only partially wired: the unit tests pass on a direct `Consume(ctx, ...)` call, while the real task loop still uses a non-cancelable parent.

Suggestion:

- Pass a cancelable iteration context into `consumerEntry.Consume` (for example, derive the system-account context from `ctxWithCancel` instead of `context.Background()`).
- Add an integration-style regression around the iteration path that cancels while a consumer is inside snapshot exec / tail watermark finalization, and assert prompt return.

XuPeng-SH (approved):
Latest head looks good to me.

The two blocking lifecycle issues from my last review are fixed in this update:
- the consumer-side cancellation path is now wired through the shared iteration context instead of context.Background(), so sibling failure / explicit cancel can actually interrupt in-flight consumers;
- SQL/index runners now propagate ctx.Err() instead of returning success-shaped nil on cancellation, and RunTxnWithSqlContext derives its txn context from the parent ctx instead of context.Background().

The DataRetriever side also now drops refcounts on failed/canceled deliveries, so the earlier post-failure AtomicBatch ownership leak is closed as well. I don't see a remaining high-confidence correctness blocker at this head.

pkg/iscp/data_retriever.go
```diff
@@ -133,9 +132,15 @@ func (r *DataRetrieverImpl) Next() *ISCPData {
 	var data *ISCPData
 	select {
 	case <-r.ctx.Done():
+		r.mu.Lock()
```
fengttt: This code does not look correct.  I don't know what r.mu is guarding.   And after you unlock, what ctx.Err() is returning.   Need to be mu protected?    

Essentially this mu should protect the r, then it should protect r.ctx as well, I assume?   Shouldn't the mu is "private" and encapsulated in the methods of r?

fengttt (changes requested):
Please take another review/design at the mutex usage.

XuPeng-SH (changes requested):
Thanks for the latest fixes — the earlier deadlock / cancellation blockers I raised are addressed. But two high-signal issues still remain at this head:

1. Shared iterations now fail all sibling jobs on the first consumer error.

runISCPTaskIterationConsumers groups jobs together when their watermark window can be shared (pkg/iscp/table_entry.go and pkg/iscp/job_config.go), but on the first consumer failure it now calls cancel() for the whole shared iteration (pkg/iscp/iteration.go:472-487). Any healthy sibling in the same iteration then returns context.Canceled, gets statuses[i].SetError(err), and is flushed with watermark=status.From instead of status.To (pkg/iscp/iteration.go:420-428, 677-687). The new test TestRunISCPTaskIterationConsumersSiblingFailureCancelsInFlightConsumer locks in exactly that behavior. That means one broken index job can roll back progress for every healthy sibling that only happened to share the same source scan window.

Suggestion: shared iteration should stay an optimization, not an all-or-nothing failure domain. Either do not cancel healthy siblings on one consumer failure, or at minimum do not persist sibling context.Canceled as a job failure / watermark rollback.

2. Caller cancellation can now bypass explicit txn cleanup in RunTxnWithSqlContext.

RunTxnWithSqlContext now derives newctx from the parent ctx (pkg/vectorindex/sqlexec/sqlexec.go:350-365), which is correct for in-flight cancellation, but it also uses that same canceled context for txnOp.Rollback / Commit. On this PR head, txnOperator.Rollback/Commit first call CancelAndWaitRunningSQLWithSQL(ctx, ...) and close locally on ctx error before sending the rollback/commit RPC (pkg/txn/client/operator.go:847-852, 901-906, 950-955 in this head). So if the caller cancels mid-transaction, the ISCP helper can skip explicit TN rollback/commit cleanup entirely and rely on backend GC instead.

Suggestion: keep parent cancellation for body execution, but perform rollback/commit finalization with a fresh bounded cleanup context so cleanup still reaches TN deterministically.

pkg/iscp/data_retriever.go
jiangxinmeng1: I updated this to encapsulate the protected access in DataRetrieverImpl.terminalError(). The mutex is the existing DataRetrieverImpl mutex and it
  only protects r.err, which is written by SetError() and read by hasError()/terminalError(). r.ctx itself is assigned once during construction and
  context.Err() is safe to call concurrently, so it does not need to be protected by this mutex.

  The intended priority is:
  1. return the recorded r.err when SetError() canceled the retriever, because that is the root cause;
  2. otherwise return r.ctx.Err() for plain parent cancellation or timeout.

XuPeng-SH (approved):
Latest head looks good to me.

The two blockers from my last review are addressed in this update:
- shared ISCP iterations no longer cancel every healthy sibling job on the first consumer failure, and the new test now checks that a healthy sibling keeps a clean status while only the failing sibling is marked failed;
- RunTxnWithSqlContext now uses a fresh bounded cleanup context for rollback/commit finalization, so caller cancellation no longer bypasses explicit txn cleanup.

I don't see a remaining high-confidence correctness blocker at this head.

PR #25670, opened 2026-07-13
fix: invalidate prepared statements on newer table versions

### LINKED ISSUE #25600: [Bug] prepared statement execution panics after column ADD followed by RENAME
## Summary

Executing a SQL prepared statement after a column ADD followed by a column RENAME can panic with an unsafe slice cast.

## Reproduction

At commit `7821d21c723372e45a8f65da3bbb896a157b1036`:

```sql
drop database if exists pddl;
create database pddl;
use pddl;
create table t (id int primary key, payload varchar(20));
insert into t values (1, 'x');
prepare ps from 'select payload from t where id=?';
set @id = 1;
execute ps using @id;
alter table t add column extra int;
execute ps using @id;
alter table t rename column payload to payload2;
execute ps using @id;
```

The final `EXECUTE` returns `ERROR 20101 internal error: panic unsafe slice cast: from length 4 is not enough for length 24`. The stack includes `pkg/container/vector/vector.go:2530` and `pkg/vm/engine/disttae/local_disttae_datasource.go:887`.

The same failure is reproducible on a normal table (no publication), so this appears to be a prepared-plan/DDL invalidation issue rather than a publication-only problem.

## Expected

The prepared statement should be invalidated/replanned or return a normal schema error; it must not panic the server.

### LINKED ISSUE #25622: [Bug]: Prepared SELECT returns a different column after RENAME COLUMN

- [x] I searched existing issues and pull requests.
- Related: #25600 reports an unsafe-slice panic for a VARCHAR projection after `ADD COLUMN` followed by `RENAME COLUMN`. This issue has a smaller trigger and silent correctness failures for fixed-width prepared SELECT projection, filtering, sorting, and indexed access after `RENAME COLUMN` alone.

### Branch Name

main

### Commit ID

- Local launch-with-proxy deployment with two CNs
- Reproduced through Proxy `6001` and direct CN ports `16001` / `16002`
- OS: macOS arm64

### Actual Behavior

Prepared SELECT plans retain stale column-position state after a referenced INT column is renamed. The statement succeeds but can read a different column, filter out valid rows, or sort by the wrong value.

Confirmed variants:

| Prepared query shape | Before rename | After rename | Fresh/non-prepared result |
|---|---|---|---|
| `SELECT payload WHERE id=?` | `42` | `1` (the primary-key value) | `42` |
| `SELECT id WHERE payload=42 ORDER BY id` | `1,3` | empty | `1,3` |
| `SELECT id ORDER BY payload LIMIT 2` | `2,3` | `1,2` | `2,3` |
| Indexed `SELECT id WHERE payload=42` | `1,3` | empty | `1,3` |

All failures are silent wrong results. No `ADD COLUMN` is required.

The projection case is deterministic through Proxy and independently on both CNs:

```text
initial EXECUTE:     42
post-rename EXECUTE: 1
normal SELECT:       42
```

Reconnecting and preparing a fresh statement against `payload2` returns correct results.

### Expected Behavior

After a schema change that renames a referenced column, a prepared statement must either:

1. invalidate and safely rebind/reprepare while preserving intended column identity; or
2. return a normal stale-schema / unknown-column error.

It must never resolve a stale ordinal to another column or silently change filter/order semantics.

### Minimal Reproduction

```sql
DROP DATABASE IF EXISTS prepared_rename_wrong_result;
CREATE DATABASE prepared_rename_wrong_result;
USE prepared_rename_wrong_result;

CREATE TABLE t (id INT PRIMARY KEY, payload INT);
INSERT INTO t VALUES (1, 42);

PREPARE ps FROM 'SELECT payload FROM t WHERE id=?';
SET @id = 1;
EXECUTE ps USING @id; -- 42

ALTER TABLE t RENAME COLUMN payload TO payload2;

EXECUTE ps USING @id;                 -- actual: 1
SELECT payload2 FROM t WHERE id = 1;  -- 42
DEALLOCATE PREPARE ps;
```

### Filter / ORDER BY / Index Reproductions

```sql
CREATE TABLE q (
    id INT PRIMARY KEY,
    payload INT,
    guard_col INT,
    KEY idx_payload(payload)
);
INSERT INTO q VALUES (1,42,7),(2,99,42),(3,42,9);

PREPARE pf FROM 'SELECT id FROM q WHERE payload=? ORDER BY id';
SET @v=42;
EXECUTE pf USING @v; -- 1,3
ALTER TABLE q RENAME COLUMN payload TO payload2;
EXECUTE pf USING @v; -- actual: empty; correct: 1,3

CREATE TABLE s(id INT PRIMARY KEY,payload INT,guard_col INT);
INSERT INTO s VALUES(1,30,300),(2,10,100),(3,20,200);
PREPARE po FROM 'SELECT id FROM s ORDER BY payload LIMIT 2';
EXECUTE po; -- 2,3
ALTER TABLE s RENAME COLUMN payload TO payload2;
EXECUTE po; -- actual: 1,2; correct: 2,3
```

### Differential Evidence

| Prepared operation | DDL | Result |
|---|---|---|
| SELECT fixed-width projection/filter/order/index | `RENAME` only | silent wrong result |
| SELECT VARCHAR projection | `ADD` then `RENAME` | unsafe slice cast panic (#25600) |
| UPDATE target column | `ADD` then `RENAME` | renamed column updated correctly in smoke test |
| DELETE predicate column | `ADD` then `RENAME` | normal `column ... does not exist` error |

This isolates the defect to prepared SELECT plan/projection/filter state rather than stored data corruption.

### Acceptance Criteria

1. Invalidate or safely rebind prepared SELECT plans after ADD, DROP, RENAME, and CHANGE COLUMN.
2. Cover fixed-width and varlen columns in projection, WHERE, ORDER BY, LIMIT, and secondary-index scans.
3. Cover column-position changes before and after referenced columns.
4. Verify repeated execution through Proxy and independently on multiple CNs.
5. Ensure both result metadata and values match the rebound schema.
6. Never turn a stale projection/filter into a type-compatible wrong-column read.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25600
Fixes #25622

Correct the table-version comparison used to invalidate prepared statements after schema changes. A catalog entry is newer when its version is greater than the version captured during prepare.

Add regression coverage for newer, equal, and timestamp-ineligible catalog versions.


XuPeng-SH (approved):
LGTM after reviewing this from the catalog-version, prepared-invalidation, and unhappy-path angles.

- correctness: `HasNewerVersion` now treats a table entry as newer only when its captured version is greater than the prepared-plan version, which matches the invalidation contract described by the issues
- scope: the change is tight and preserves the existing timestamp gating, table-ID mismatch handling, and deletion invalidation behavior
- unhappy path coverage: the new regression covers newer, equal, and timestamp-ineligible catalog versions, which are the three comparison cases that matter here

gouhongshen (approved):

PR #25675, opened 2026-07-13
fix: preserve expression defaults in alter table

### LINKED ISSUE #25384: [Bug]: ALTER TABLE ADD COLUMN with expression default fails

### Branch Name

main

### Commit ID

51656deb6

```Markdown
- MO version: 8.0.30-MatrixOne-v
- MySQL version: 8.0.45
- Verify date: 2026-07-02
```

### Actual Behavior

`ALTER TABLE ... ADD COLUMN ... DEFAULT (expression)` fails in MatrixOne, while the same statement succeeds in MySQL and backfills existing rows with the expression default.

MatrixOne error:

```text
ERROR 1105 (HY000): syntax error ... near "x'', ''y'')), ..."
```

### Expected Behavior

The column should be added successfully, and existing rows should receive the expression default value, consistent with MySQL.

### Steps to Reproduce

```sql
drop database if exists mysql_compat_model11_min;
create database mysql_compat_model11_min;
use mysql_compat_model11_min;

create table t_expr (id int primary key);
insert into t_expr values (1);

alter table t_expr add column c varchar(10) default (concat('x','y')) first;
select * from t_expr;
show create table t_expr;
```

MySQL 8.0.45 result:

```text
c   id
xy  1

CREATE TABLE `t_expr` (
  `c` varchar(10) DEFAULT (concat(_utf8mb3'x',_utf8mb3'y')),
  `id` int NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3
```

MatrixOne main `51656deb6` result:

```text
ERROR 1105 (HY000): syntax error ...

id
1

CREATE TABLE `t_expr` (
  `id` int NOT NULL,
  PRIMARY KEY (`id`)
)
```

### Additional information

`CREATE TABLE` with expression defaults is accepted, so this appears specific to the `ALTER TABLE ADD COLUMN` path.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25384

Preserve parenthesized expression defaults when reconstructing CREATE TABLE SQL instead of escaping quotes inside the expression. Also preserve the expression in DESC/SHOW COLUMNS output.

Add planner and distributed regression coverage for ADD COLUMN and ALTER COLUMN SET DEFAULT expression defaults, including backfill, inserts, DESC, and SHOW CREATE TABLE.


XuPeng-SH (approved):

The fix is appropriately narrow: show/create and DESC now preserve parenthesized expression defaults verbatim, while plain literal defaults still flow through the existing string-formatting path. That matches the actual metadata distinction the ALTER TABLE bug was tripping over.

The added coverage also exercises both affected surfaces: ADD COLUMN with expression default backfill plus ALTER COLUMN SET DEFAULT, and it checks the stored value, DESC output, and SHOW CREATE TABLE reconstruction.

heni02 (approved):

aunjgr (approved):

Issue #25567, opened 2026-07-10
[Bug]: RESTORE ACCOUNT intermittently fails to clean cross-database foreign keys


matrix-meow:
## Problem

`Compatibility Test With Target` intermittently fails while executing:

```sql
-- test/distributed/cases/snapshot/snapshot_restore_to_table_fk_2.sql:872
restore account sys{snapshot="snapshot_01"};
```

The first failure is:

```text
internal error: can not drop database. It has been referenced by foreign keys
```

The test creates `Company`, `Projects`, and `Payroll` after taking the account snapshot. These databases contain cross-database foreign keys. `RESTORE ACCOUNT` should remove them and restore the account to the snapshot state, but it occasionally fails during the cleanup phase.

## Evidence

Occurrence: [PR #25566, run attempt 1, job 86275758678](https://github.com/matrixorigin/matrixone/actions/runs/29065408149/job/86275758678)

The report contains 27 failed assertions, but they are not independent failures:

1. The root failure is `snapshot_restore_to_table_fk_2.sql:872`, where `RESTORE ACCOUNT` fails to drop a database because of an FK reference.
2. The failed restore transaction leaves `Company`, `Projects`, and `Payroll` in the account.
3. The remaining 26 failures are downstream state pollution: 21 `SHOW DATABASES` mismatches, 1 system table count mismatch, and 4 duplicate-name checks in `information_schema`.

The same workflow was rerun with the same head SHA (`c3c9048`) and the same target commit (`5fc3e50`):

- Attempt 1: failed, job `86275758678`
- Attempt 2: passed, job `86283655932`

This makes the failure an intermittent product/runtime issue rather than a deterministic golden-file or per-file test-isolation issue.

## Failure Timeline

The failed restore uses transaction `70f70f4866a9b62a18c0ce4b7e070559`.

1. Its original snapshot is approximately `1783653914973800021-1`.
2. `deleteCurFkTables` submits deletion of all 16 FK-related tables, in reverse topological order, before any database is dropped. This includes all FK tables in `acc_test02`, `company`, `projects`, and `payroll`.
3. At `03:25:15.182917`, `mo_foreign_keys` starts a flush/merge. It replaces appendable object `428d269050df_0` and tombstone object `ba9cd301b9e9_0`.
4. At `03:25:15.195159`, that flush ends with `aobj-deletes:181`, `aobj-merge-rows:20`, and new objects `35f8d5dfccf8_0` / `56fa2c616489_0`.
[...]

PR #25676, opened 2026-07-13
fix: preserve tombstones when advancing RC snapshot

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25567

`DROP DATABASE` advances the snapshot of pessimistic RC transactions after acquiring the `mo_database` lock. If the transaction already contains tombstones and TN rewrites the referenced objects, advancing only `TxnOperator.SnapshotTS` leaves those tombstones pointing at obsolete row IDs and can resurrect deleted catalog rows.

This PR:

- adds `Workspace.AdvanceSnapshot` to advance the RC snapshot and transfer in-memory and persisted tombstones to newly visible objects as one workspace operation
- uses it in `DropDatabase` and keeps the snapshot advanced instead of rewinding only `SnapshotTS`
- leaves clone timestamp minting and lock-conflict refresh semantics unchanged
- adds deterministic coverage for in-memory and persisted tombstones, commit, transaction rollback, subsequent statements, and statement rollback followed by commit/rollback
- strengthens the restore-cluster BVT with two databases that must be removed in the same restore transaction

Validation:

- `go test -count=1 -timeout 10m ./pkg/sql/compile ./pkg/txn/client ./pkg/vm/engine/disttae`
- focused RC snapshot advance harness tests
- statement rollback harness with `-race`
- `mo-tester test/distributed/cases/snapshot/cluster/restore_cluster_simple.sql`: 429/429 passed
- `go vet ./pkg/sql/compile ./pkg/txn/client ./pkg/vm/engine/disttae`
- `git diff --check`


XuPeng-SH (approved):
No new high-confidence blocker found in the latest head.

I re-checked the RC snapshot-advance path from first principles: `DropDatabase` now advances through `Workspace.AdvanceSnapshot`, so the RC workspace transfers both in-memory and persisted tombstones before reading catalog/FK metadata at the newer snapshot, instead of leaving old rowid tombstones behind. The snapshot also correctly stays advanced afterward, which avoids the old rewind-only inconsistency.

The added coverage hits the main unhappy paths that matter here: direct `AdvanceSnapshot` branch behavior, in-memory vs persisted tombstone transfer, subsequent statements, statement rollback followed by commit/rollback, and the restore-cluster regression with multiple transient databases removed in one restore transaction.

LGTM.

PR #25679, opened 2026-07-13
fix(disttae): prevent sync.Mutex reentrant deadlock in dumpBatch (cherry-pick #25561)

### LINKED ISSUE #25557: [Bug]: sync.Mutex reentrant deadlock in disttae Transaction causes all subsequent queries to block indefinitely

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25557

Cherry-pick of #25561 (base: 3.0-dev) onto main, squashed to the branch's net diff since the source PR is still open and its history contains superseded intermediate approaches.

Internal SQL executed from workspace-dump paths (`getTable` → `Engine.Database` → `loadDatabaseFromStorage` → `execReadSql` → `NewCompile`) used to re-enter `txn.Lock()` and store to the shared `snapshotWriteOffset`, corrupting the positional statement boundary of `txn.writes` (reentrant deadlock, mid-statement visibility corruption, and an `index out of range` panic under many-to-one workspace compaction).

### Fix

- Internal sub-sql compiles (`DisableIncrStatement`, marked on the process) capture a transient `Workspace.WriteOffset()` as their own `TxnOffset` and never advance the shared statement boundary.
- `snapshotWriteOffset` becomes an `atomic.Int64`: reads are lock-free, stores stay under `txn.Lock` at statement boundaries only.
- Workspace dumps resolve target tables before mutating `txn.writes`, releasing the lock around `getTable` and re-validating the offset after relock.
- `compactDeletionOnObjsLocked` pre-resolves tables so workers never call `getTable` while the parent holds the lock (commit deadlock).
- `Transaction.Commit` runs `mergeTxnWorkspaceLocked` + the final dump under one lock block.
- Fault-injection hooks + regressions (unit and e2e) cover multi-entry compaction, resolution-window append, object-compaction deadlock, and rogue boundary advance.

### Main-specific adaptations

- `UpdateSnapshotWriteOffset` keeps main's `adjustWriteOffset` in sync (`= len(txn.writes)`).
- Kept main's `txn.engine == nil` guard at the top of `getTable`.
- The `CNCommitAfterWorkspaceDumpFailed` injection now runs after the new `Commit` lock block.

Verified the net diff of this branch vs `main` matches the net diff of #25561 vs `3.0-dev` except for the adaptations above.

### DIFF
--- a/pkg/frontend/test/txn_mock.go
+++ b/pkg/frontend/test/txn_mock.go
@@ -1254,6 +1254,20 @@ func (mr *MockWorkspaceMockRecorder) GetSyncProtectionJobID() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetSyncProtectionJobID", reflect.TypeOf((*MockWorkspace)(nil).GetSyncProtectionJobID))
 }
 
+// WriteOffset mocks base method.
+func (m *MockWorkspace) WriteOffset() uint64 {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "WriteOffset")
+	ret0, _ := ret[0].(uint64)
+	return ret0
+}
+
+// WriteOffset indicates an expected call of WriteOffset.
+func (mr *MockWorkspaceMockRecorder) WriteOffset() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "WriteOffset", reflect.TypeOf((*MockWorkspace)(nil).WriteOffset))
+}
+
 // IncrSQLCount mocks base method.
 func (m *MockWorkspace) IncrSQLCount() {
 	m.ctrl.T.Helper()
--- a/pkg/objectio/injects.go
+++ b/pkg/objectio/injects.go
@@ -76,6 +76,20 @@ const (
 	FJ_CNCLONEFailed                    = "fj/cn/clone_fails"
 	FJ_CNCommitAfterWorkspaceDumpFailed = "fj/cn/commit_after_workspace_dump_fails"
 	FJ_CNNeedRetryError                 = "fj/cn/need_retry_error"
+
+	// FJ_CNReenterSnapshotOffsetOnGetTable is a test-only fault that makes
+	// Transaction.getTable reenter UpdateSnapshotWriteOffset, deterministically
+	// simulating the internal-SQL leg of the issue #25557 deadlock:
+	// getTable -> Engine.Database -> loadDatabaseFromStorage -> execReadSql
+	// -> NewCompile -> UpdateSnapshotWriteOffset.
+	FJ_CNReenterSnapshotOffsetOnGetTable = "fj/cn/reenter_snapshot_offset_on_get_table"
+
+	// FJ_CNDumpResolveWindowWait is a test-only orchestration point inside the
+	// unlocked table-resolution window of a workspace dump. Armed with the
+	// WAIT action, it parks the dumping goroutine right before the transaction
+	// lock is re-acquired, so a test can deterministically mutate the
+	// workspace from another goroutine while the window is open.
+	FJ_CNDumpResolveWindowWait = "fj/cn/dump_resolve_window_wait"
 )
 
 const (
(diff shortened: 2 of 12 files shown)


XuPeng-SH (approved):

The fix addresses the issue at the right boundaries: internal sub-SQL compiles now capture a transient workspace end offset instead of advancing the shared statement boundary, dump/compaction resolves tables only with the transaction lock released, and commit keeps the final workspace merge plus dump inside one lock block so the boundary and writes stay coherent.

The injected regression coverage also matches the real failure modes from #25557: many-to-one dump compaction, append during the dump-resolution window, blocked reentry into snapshot offset capture, and post-dump failure handling.

Issue #25680, opened 2026-07-13
[Bug]: drop account FK cleanup can panic and expose cross-tenant DB names to UDF checks


LeftHandCold:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main, 4.1-dev

### Commit ID

- main: `118bb865d037c0c11113786db98999284945d1c4`
- 4.1-dev: `33bc7a22726f94e6dcbd0fcd73d4ebbfe7d177ba`

### Other Environment Information

- GitHub Actions BVT: https://github.com/matrixorigin/matrixone/actions/runs/29235449949/job/86769554358
- Mode: standalone, pessimistic transactions

### Actual Behavior

The BVT has two causally linked failures:

1. `test/distributed/cases/tenant/drop_account.sql` fails while dropping an account that owns a cross-database foreign key. `DROP ACCOUNT` chooses the database order from a Go map. When it drops referenced database `db1` before child database `db2`, FK cleanup rewrites `db2.t2` in the same transaction. The later drop of `db2.t2` reads both catalog versions and panics because `deleteTable` receives two row IDs:

   ```text
   internal error: panic delete table failed: query failed
   FIND_TABLE deleteTableError
   select __mo_rowid from mo_catalog.mo_tables
     where account_id = 345 and reldatabase = "db2" and relname = "t2"
   ```

2. The failed account drop rolls back and leaves tenant 345's `db2`. Later, sys executes `drop function db2.helloworld5()`. `checkDatabaseExistsOrNot` queries `mo_database` by database name only, so sys sees another tenant's `db2` and returns `function helloworld5 doesn't exist` instead of `Unknown database db2`.

The failures are unrelated to PR #25660's one-line TAE test timeout change.

### Expected Behavior

- Dropping an account must succeed regardless of database iteration order when its databases contain cross-database foreign keys.
- UDF database existence checks must only consider the current account.

### Steps to Reproduce

```sql
drop account if exists account_test;
create account account_test admin_name = 'test_account' identified by '111';

-- Run as account_test
create database db1;
create table db1.t1(a int primary key);
create database db2;
create table db2.t2(
  a int primary key,
  b int,
  foreign key (b) references db1.t1(a)
);

-- Run as sys. The failure depends on the Go map yielding db1 before db2.
drop account account_test;

-- If the previous statement rolled back, this also demonstrates the
-- cross-tenant database-name leak in the UDF existence check.
drop function db2.helloworld5();
```

[...]

PR #25681, opened 2026-07-13
fix: isolate account FK cleanup and UDF database checks

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25680

`DROP ACCOUNT` disables foreign-key enforcement but still rewrites cross-database FK metadata while dropping each database. Because the databases are collected in a Go map, a referenced database can be dropped before its child database. That rewrite leaves a delete+insert catalog pair for the child table in the same transaction, and the later table drop can read both versions and panic in `disttae.deleteTable`.

This change marks account-wide cleanup with `IgnoreForeignKey` and makes `DropDatabase` skip FK relationship rewrites for that internal path. All databases in the account are being removed, so no surviving relation needs the metadata update.

The failed account drop also exposed a second tenant-isolation issue: UDF database existence checks queried `mo_database` by name only, so sys could mistake another tenant's database for its own. The UDF-specific check now includes the current `account_id`.

Regression tests verify that the flag is propagated from `DROP ACCOUNT` into database drops, that `DropDatabase` skips FK cleanup only when requested, and that UDF database checks include the tenant ID.

Tested with:

`go test -count=1 -timeout 180s ./pkg/frontend`

`go test -count=1 -timeout 180s ./pkg/sql/compile`

`go build ./pkg/frontend ./pkg/sql/compile`

`go vet ./pkg/frontend ./pkg/sql/compile`

### DIFF
--- a/pkg/frontend/authenticate.go
+++ b/pkg/frontend/authenticate.go
@@ -1344,7 +1344,8 @@ const (
        									    and privilege_id = %d 
        									    and privilege_level = "%s";`
 
-	checkDatabaseFormat = `select dat_id from mo_catalog.mo_database where datname = "%s";`
+	checkDatabaseFormat          = `select dat_id from mo_catalog.mo_database where datname = "%s";`
+	checkDatabaseByAccountFormat = `select dat_id from mo_catalog.mo_database where datname = "%s" and account_id = %d;`
 
 	checkDatabaseWithOwnerFormat = `select dat_id, owner from mo_catalog.mo_database where datname = "%s" and account_id = %d;`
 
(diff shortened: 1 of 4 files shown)


XuPeng-SH (approved):

The DROP ACCOUNT fix is scoped the right way: IgnoreForeignKey is propagated only for the internal account-wide cleanup path, and DropDatabase now skips cross-database FK metadata rewrites only for that internal case. That closes the delete+insert catalog-version hazard without weakening normal user DROP DATABASE behavior.

The UDF follow-up also looks correct: checkDatabaseExistsOrNot now scopes the mo_database lookup by the current account_id, so sys no longer mistakes another tenant's database for its own when validating CREATE/DROP FUNCTION database qualifiers.

The focused regression tests and full pkg/frontend plus pkg/sql/compile package tests cover the changed paths cleanly.

PR #25682, opened 2026-07-13
fix: isolate colexec server by CN service

aptend:
## What type of PR is this?

- [x] Bug

`colexec.Server` was a process-global singleton. Standalone Multi-CN runs multiple CN services in one process, so dispatch UUID registrations, remote pipeline state, and CN transaction segment state were shared across CNs. Cleanup from one CN could therefore remove another CN's registration and make remote notify wait until the 30-second timeout.

This PR:

- creates one `colexec.Server` for every CN service;
- stores it in the service-scoped runtime and removes the process-global singleton;
- pins dispatch registration/cleanup and remote RPC handling to the owning CN server;
- scopes transaction segment tracking by CN service;
- removes the hidden global fallback in segment helpers;
- adds a regression test proving that the same receiver UUID can exist on two CN services and cleanup remains isolated.

Related issue: #25211

Failure evidence and the standalone BVT reproduction are documented in [this issue comment](https://github.com/matrixorigin/matrixone/issues/25211#issuecomment-4958352927).

Related to #25211.

## Testing

- `go test -count=1 -timeout 180s ./pkg/sql/colexec ./pkg/sql/colexec/dispatch ./pkg/sql/compile ./pkg/cnservice ./pkg/vm/engine/disttae`
- `go test -count=1 -timeout 180s ./pkg/vm/engine/test`
- `go test -race -count=1 -run '^TestServerIsIsolatedByCNService$' ./pkg/sql/colexec`
- `go build ./pkg/sql/colexec/... ./pkg/sql/compile ./pkg/vm/engine/disttae ./pkg/cnservice ./pkg/vm/engine/test/testutil`
- `go vet ./pkg/sql/colexec ./pkg/sql/colexec/dispatch ./pkg/sql/colexec/multi_update ./pkg/sql/compile ./pkg/vm/engine/disttae ./pkg/cnservice ./pkg/vm/engine/test/testutil`


XuPeng-SH (approved):

The core fix is in the right place: colexec server ownership is moved from a process-global singleton to the CN service runtime, and the affected call sites now consistently resolve the owning server by service ID instead of silently falling back to a shared global. That matches the actual failure mode from standalone multi-CN-in-one-process runs.

I also like that the regression coverage checks the real isolation property: the same receiver UUID can exist on two CN services at once, and cleanup from one service does not delete the other service's registration.

Issue #24823, opened 2026-06-03
[Bug] CREATE TABLE AS SELECT with MATCH...AGAINST drops string-literal quotes during rewrite


dengn:
## Summary
`CREATE TABLE ... AS SELECT` containing `MATCH()...AGAINST('...')` mangles the statement during rewrite — the string-literal single quotes are dropped, producing a parse error.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test case
```sql
CREATE TABLE ctas_t AS
SELECT id, MATCH(body) AGAINST('防水' IN BOOLEAN MODE) sc
FROM ft WHERE MATCH(body) AGAINST('防水' IN BOOLEAN MODE);
```

## Actual result
```
ERROR 1105 (HY000): ... syntax error near "防水 IN BOOLEAN MODE) as sc from ft where MATCH (body) AGAINST (防水 IN BOOLEAN MODE))"
```
Note the quotes around `'防水'` are gone in the rewritten SQL — a rewrite defect.

## Version status
- v3.0.11 (Cloud): reproduces
- main / 4.0-dev (commit 0bfda60): reproduces

heni02:
pr还没合入，待验证

heni02:
@cpegeric cc pr一直没合入

Ariznawlll:
最新 main  复测 3/3：CTAS 中 MATCH 投影+WHERE、WHERE-only、无命中和不同词项控制均正确，连接后续查询正常。现有 BVT  已覆盖原始路径。无需 big-data、PITR/snapshot、stability 或 chaos。结论：关闭。

Ariznawlll:
更正补充：验证基线为 MatrixOne main 28769ffe4439001a012dea74adfc31e188426df9。MATCH CTAS 投影+WHERE、WHERE-only、无命中和多词项控制均连续三轮正确。现有 test/distributed/cases/fulltext/fulltext_ctas.sql 覆盖原始路径，无需专项回归。

PR #25683, opened 2026-07-13
FIX: CREATE TABLE AS SELECT with MATCH...AGAINST drops string-literal quotes during rewrite

cpegeric:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24823

fixed. CREATE TABLE AS SELECT with MATCH...AGAINST drops string-literal quotes during rewrite


XuPeng-SH (changes requested):
This still needs changes.

The CTAS path is only fixed because its formatter context enables quoteString, but the underlying FullTextMatchExpr formatter is still invalid under the default tree.String path.

Current code writes the AGAINST pattern through WriteValue(P_char, FormatString(node.Pattern)). That only adds quotes when the FmtCtx has quoteString or singleQuoteString enabled. On the default formatter path, it still emits the pattern bare.

I verified this directly on the latest head with a tiny parser round-trip program:
- tree.String(...) still prints: MATCH (body) AGAINST (防水 IN BOOLEAN MODE)
- only tree.StringWithOpts(..., WithSingleQuoteString()) prints: MATCH (body) AGAINST ('防水' IN BOOLEAN MODE)

So the PR does not actually fix the node's default deparse behavior; it only fixes contexts that happen to opt into string quoting. Because tree.String(...) is used in many other places across the codebase, any other current or future caller that serializes MATCH ... AGAINST through the default path can still generate broken SQL.

Suggestion: make FullTextMatchExpr.Format emit a properly quoted SQL string literal unconditionally (or otherwise ensure the default formatter path round-trips), and add a regression in the default tree.String / TestValid path, not only in the WithSingleQuoteString test path.

gouhongshen (approved):
## Codex automated review

The formatter now emits a syntactically valid, escaped string literal for every MATCH...AGAINST pattern, including the CTAS reconstruction path. Focused parser tests passed; no blocking regressions found.

iamlinjunhong (changes requested):
Findings
[P1] NO_BACKSLASH_ESCAPES 下反序列化会不断增加反斜杠
[expr.go:1940](https://github.com/matrixorigin/matrixone/blob/e414129124b0e301e59220715e5f5d1599d2f462/pkg/sql/parsers/tree/expr.go#L1940)
FormatString(node.Pattern) 总是按默认 MySQL 模式转义反斜杠。但最新 main 已支持 NO_BACKSLASH_ESCAPES；该模式解析字符串时把反斜杠视为普通字符。
在最新 main 与 PR head 合并后的代码上，我验证了：
select match(body) against('a\nb' in boolean mode) from t
使用 NO_BACKSLASH_ESCAPES：
第一次 format:  AGAINST ('a\\nb' ...)
第二次 format:  AGAINST ('a\\\\nb' ...)
每次 parse → format 都会翻倍，搜索 pattern 的语义随 CTAS、view 或其他重写而改变。PR 修复了默认模式的掉引号问题，但没有在当前支持的 SQL mode 下闭合 formatter → parser 契约。
建议让字符串转义感知 SQL mode，并由 CTAS formatter 传入当前 session parser mode；同时增加：
ParseOneWithSQLMode(NO_BACKSLASH_ESCAPES)
  → tree.String/StringWithOpts
  → ParseOneWithSQLMode(NO_BACKSLASH_ESCAPES)
  → AST pattern 相等且二次 format 幂等
测试至少覆盖普通反斜杠、尾随反斜杠、\n 字面量和反斜杠紧邻单引号。

heni02 (approved):

gouhongshen (changes requested):
## Codex automated review

The default quoting regression is fixed, but CTAS still changes MATCH patterns containing backslashes under NO_BACKSLASH_ESCAPES.

### P1 - CTAS does not propagate NO_BACKSLASH_ESCAPES to the formatter (pkg/sql/parsers/tree/expr.go:1954)
When FmtCtx.NoBackslashEscape is false, this calls FormatString, doubling stored backslashes. However, CTAS formats stmt.AsSource with only WithQuoteString(true) in build_ddl.go:1504, then reparses the generated SQL under the session mode. Under NO_BACKSLASH_ESCAPES, the original backslash is stored literally, so the CTAS follow-up query searches for a different pattern and can copy incorrect rows. The new test avoids the bug by explicitly passing WithNoBackslashEscape; no production caller passes this option. Propagate the session mode into the CTAS formatter and add an end-to-end regression.

fengttt:
Review follow-up (pushed via maintainer edit — @cpegeric please take a look):

**Assessment of the three CHANGES_REQUESTED reviews against the head:**
- @XuPeng-SH's default-path finding: already addressed by the current head (unconditional quoting in `FullTextMatchExpr.Format` + default-path regressions in `TestFullTextMatchDeparseRoundTrip`/`TestValid`).
- @iamlinjunhong's NO_BACKSLASH_ESCAPES doubling: mechanism addressed by the head (`WithNoBackslashEscape` + the four requested round-trip cases), **but** —
- @gouhongshen's follow-up was correct: **no production caller passed the option**. CTAS formatted `AsSource` with only quote options while the generated `INSERT ... SELECT` is re-parsed by the internal executor under the session's sql_mode — so the doubling bug persisted in production and only the test opted in.

**Fixed in e3eca1a2fa:**
- `buildTable` now resolves the session `sql_mode` (same `parserSQLModeFromContext` helper the view path uses) and passes `WithNoBackslashEscape` to the CTAS formatter when the mode is active.
- Regression `TestCTASFullTextNoBackslashEscapes`: builds the CTAS plan under `NO_BACKSLASH_ESCAPES` (mock `sql_mode` override, mode-aware parse) and asserts the generated `CreateAsSelectSql` keeps the literal single backslash — plus the default mode still re-escapes. **Verified it fails without the `build_ddl.go` change and passes with it.** I chose this plan-level regression over a BVT because fulltext tokenizers split on `\`, so a doubled pattern produces identical match results end-to-end — the BVT would be blind to exactly the bug it's meant to catch; the generated-SQL assertion is the observable contract.
- Also merged current main (was DIRTY): kept both `FmtCtx` additions (`noBackslashEscape` + main's `paramExprOffset`) and both test insertions.

One known adjacent gap, deliberately out of scope: ordinary (non-MATCH) string literals in CTAS go through `FmtCtx.WriteValue`'s `%q` path, which also backslash-escapes regardless of mode — a pre-existing issue affecting all quoted formatting, not introduced here. Happy to file it separately.

fengttt:
Filed the adjacent `WriteValue` `%q` gap as #26300, with an empirical probe of the three failure classes (backslash doubling under NO_BACKSLASH_ESCAPES, outright syntax errors for double-quote-containing literals under NBE/ANSI_QUOTES, and silent control-char drift) plus the list of affected `WithQuoteString` production callers. Out of scope here as discussed.

gouhongshen (changes requested):
## Codex automated review

Default-path MATCH literal quoting is fixed. The prior NO_BACKSLASH_ESCAPES CTAS blocker is still blocking with corrected evidence: the new formatter option assumes session-mode reparsing, but CTAS’s internal executor parses in default mode. The generic WriteValue gap is pre-existing and tracked separately (#26300), so not a finding. Merge-resolution changes introduced no separate issue. Focused tests could not run because `go` is unavailable in this workspace.

### P1 - CTAS now changes NO_BACKSLASH_ESCAPES MATCH patterns during its default-mode internal reparse (pkg/sql/plan/build_ddl.go:1560)
Author response: this condition adds WithNoBackslashEscape because the CTAS follow-up INSERT is said to be reparsed under the session SQL mode. Why this remains blocking: Compile executes that INSERT through txnExecutor.Exec, which calls parsers.Parse at pkg/sql/compile/sql_executor.go:351; parsers.Parse unconditionally supplies an empty SQL mode (pkg/sql/parsers/sqlparse.go:30-32). Thus an original NBE pattern `a\nb` is stored as literal backslash+n, but the new branch emits one backslash and the internal default parser consumes `\n` as a newline. Before this change, the normal formatter emitted `\\`, which the default parser reduced back to the original literal backslash. CTAS can therefore use a different fulltext pattern and copy different rows. Either keep CTAS formatting compatible with its enforced default parser, or propagate SQL mode into the internal parser and make all CTAS literal formatting mode-aware; add a regression that reparses the generated CTAS SQL through the actual executor path and compares the resulting pattern.

fengttt:
@gouhongshen's P1 is correct — thanks, the automated review caught a wrong premise in my previous commit. Verified all three links: the CTAS follow-up INSERT runs through `runSqlWithResultAndOptions` → internal `SQLExecutor`, whose `Exec` calls `parsers.Parse` (pkg/sql/compile/sql_executor.go:351) with an **empty** SQL mode, and `ParseSQLModeFlags("")` leaves backslash-escape processing ON in the scanner. So the reparse is always default-mode: under an NBE session my conditional emitted a single backslash that the executor's parse consumed as an escape (`a\nb` → `a<newline>b`), while the original unconditional doubling round-tripped correctly. My earlier regression asserted the generated SQL text instead of the reparse result — the wrong contract.

**Fixed in 2224671345** (reviewer's option A — keep CTAS formatting compatible with the enforced default parser):
- Reverted the `WithNoBackslashEscape` wiring in `buildTable`; the comment there now states the real contract (CTAS literals must stay default-escaped because the internal executor always parses in default mode — do not make this session-mode-aware unless the executor's parse becomes so too). The formatter option itself stays: it is still correct for same-mode round-trips and keeps the parser-level `TestFullTextMatchPatternRoundTrip` cases valid.
- Replaced the test with `TestCTASFullTextPatternSurvivesInternalReparse`, which asserts the requested contract: build the CTAS plan under a session sql_mode, reparse `CreateAsSelectSql` exactly as the executor does (default mode), and require the resulting MATCH pattern to equal what the user wrote — for both `NO_BACKSLASH_ESCAPES` and default sessions. Verified it **fails** against the previous conditional and **passes** with this change; related parser/plan tests and vet are green.

I chose option A over propagating session sql_mode into the internal parser because that would change parsing for *all* internal SQL (much of it system-generated assuming default escaping) — too risky for this PR's scope.

@cpegeric FYI — pushed via maintainer edit again.

gouhongshen (approved):
## Codex automated review

Previous blockers are resolved: default MATCH literal quoting remains fixed, and the CTAS NO_BACKSLASH_ESCAPES issue is fixed by retaining default escaping for the internal executor’s enforced default-mode reparse, with a regression covering both modes. The broader pre-existing WriteValue escaping gap is tracked in #26300 and is outside this review. Focused Go tests were unavailable because no Go executable is installed locally.

XuPeng-SH (approved):
Approved on exact head `2224671345568b12d6e3c84d76c348a1719b6d8e`.

The earlier blockers are closed. `FullTextMatchExpr` now always emits a valid quoted literal on the default formatter path, while the opt-in NO_BACKSLASH_ESCAPES path preserves same-mode AST round trips. More importantly, the latest CTAS correction matches the actual execution contract: the follow-up INSERT is reparsed by the internal executor in default mode, so CTAS must keep default escaping even when the user statement was parsed under NBE. The new plan regression checks the post-reparse pattern rather than only generated text.

I additionally exercised 20,000 deterministic randomized literal round trips across default and NO_BACKSLASH_ESCAPES modes (quotes, backslashes, controls, wildcard escapes, comments/delimiters, Unicode) with no value drift. The three exact new tests passed; focused race repetition passed 67x/67x for parser tests and 42x for the plan test; both complete owning packages passed under `-race`; vet is clean. Latest `origin/main` (`159742db6672ff5dadb617880b831da026f0a3f4`) merges cleanly and focused race tests pass on the prospective merge.

The generic `FmtCtx.WriteValue` mode gap is pre-existing, separately tracked in #26300, and is not a blocker for this MATCH-specific fix.

fengttt:
@Mergifyio queue

fengttt:
@iamlinjunhong could you take another look? Your NO_BACKSLASH_ESCAPES round-trip concern was addressed (WithNoBackslashEscape formatter option + the four requested round-trip cases in TestFullTextMatchPatternRoundTrip), and the follow-up review iterations settled the CTAS wiring: the internal executor always reparses in default mode, so CTAS formatting stays default-escaped with TestCTASFullTextPatternSurvivesInternalReparse asserting the reparse contract. gouhongshen and XuPeng-SH have both approved the current head; your CHANGES_REQUESTED is the last gate before the merge queue.

PR #25688, opened 2026-07-13
fix: wait for remote receivers on empty dispatch

### LINKED ISSUE #25687: [Bug]: Empty remote dispatch can lose receiver registration and time out
## Have you checked whether this issue has already been reported?

- [x] I have searched the existing issues. #25211 tracks the broader symptom, but this report has a distinct deterministic root cause and reproducer.

## Branch Name

main

## Commit ID

Reproduced on PR #25135 commit 28dd648512526d68488eecdfcc4a874dae7ad5c6. The same dispatch lifecycle is present on main commit 3728f38726f209b20129fcf4578ff0095f824893.

- Multi-CN Docker deployment started with make dev-up
- Two CN services in separate containers/processes, one TN, one log service, and one proxy
- The failure was reproduced through the proxy on port 6001

## Actual Behavior

A distributed query whose producer returns no batch can intermittently fail after 30 seconds with:

Both CN processes remain healthy. The error is caused by a registration race rather than a failed CN.

The observed lifecycle is:

1. A remote notify reaches the producer CN a few milliseconds before the dispatch UUID is registered and receives RemoteDispatchNotRegistered.
2. The producer registers the UUID.
3. The producer returns an empty result. Dispatch.Call returns ExecStop without waiting for remote receivers because no send function is called.
4. Pipeline cleanup removes the registration before the client's 200 ms retry.
5. Every later retry sees an absent UUID until the 30-second registration timeout.

One captured failure registered the UUID about 2 ms after the first lookup, but cleaned it about 140 ms later; the next retry could only occur at 200 ms.

## Expected Behavior

A successful remote dispatch must establish all of its remote receivers before it reaches terminal completion, including when its child produces zero batches. An early NotRegistered response must not outlive the producer pipeline. Cancellation and an unreachable remote CN must still terminate through the existing bounded wait.

## Steps to Reproduce

Start a two-CN development cluster and create the following data:

    create database d1;
    use d1;
    create table t1(c1 int not null, c2 int not null, c3 int not null) cluster by c1;
    create table t2(c1 int not null, c2 int not null, c3 int not null) cluster by c1;
    create table t3(c1 int not null, c2 int not null) cluster by c1;
    insert into t1 select *,*,* from generate_series(5000000) g;
    insert into t2 select *,*,* from generate_series(4000000) g;
    insert into t3 select *,* from generate_series(1,1000000) g;

Run this query concurrently through the proxy:

    select count(*)
    from t1, t2, t3
    where t1.c1 = t2.c1
      and t1.c2 = t3.c2
      and t2.c2 < 900000
      and t3.c1 < 500000;

The correct result is 499999. On the unfixed code, 4 of 200 executions failed at concurrency 8, and 78 of 400 executions failed at concurrency 16 with the 30-second dispatch registration error.

The original CI occurrence is in PR #25135:
https://github.com/matrixorigin/matrixone/actions/runs/29256681241/job/86838789949?pr=25135

## Additional Information

PR #25475 made an early missing registration explicit and added bounded client retries. That exposed an older producer-side lifecycle gap: receiver readiness is currently checked only by data send functions, so the zero-batch fast path skips the barrier entirely.

PR #25682 addresses process-global colexec server state in standalone multi-CN. This issue also reproduces when CNs run in separate containers/processes, so that change does not close this race.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25687

A remote dispatch currently waits for receiver attachment only when a data batch reaches its send function. If the child produces no batches, Dispatch.Call returns ExecStop immediately. An early RemoteDispatchNotRegistered response can then outlive the producer: cleanup removes the newly published UUID before the client's 200 ms retry, and all later retries fail until the 30-second timeout.

This PR restores the remote-dispatch lifecycle invariant at the producer boundary: successful terminal completion waits for all remote receivers even when the input is empty. It reuses the existing receiver barrier, context cancellation, and 30-second upper bound.

The production change is intentionally limited to the zero-batch remote path:

- non-empty dispatch is unchanged;
- local-only dispatch is unchanged;
- an already attached empty dispatch adds only two boolean checks;
- a racing registration waits for the existing notify retry instead of failing after 30 seconds;
- cancellation and unreachable receivers return through the existing bounded error paths;
- no protocol, lock, goroutine, retry interval, or retained state is added.

The regression tests cover direct empty-input attachment, cancellation-cause propagation, and the exact end-to-end interleaving where the first notify misses registration, the producer registers and becomes empty, and the retry must attach before terminal cleanup. The direct test was also run with the production fix removed and failed immediately, proving that it detects the original bug.

## Testing

- make -j12
- make dev-build
- go test ./pkg/sql/colexec/dispatch
- go test ./pkg/sql/compile
- go test -race for the empty-dispatch tests
- go test -race for TestRemoteNotifyRetryAttachesToEmptyDispatchBeforeCompletion
- go vet ./pkg/sql/colexec/dispatch ./pkg/sql/compile
- go vet with molint for both affected packages
- golangci-lint run -c .golangci.yml ./pkg/sql/colexec/dispatch ./pkg/sql/compile
- Multi-CN Docker reproduction with 5M, 4M, and 1M-row tables: expected result 499999
  - concurrency 8: 200/200 passed; unfixed code failed 4/200
  - concurrency 16: 100/100 passed

### DIFF
--- a/pkg/sql/colexec/dispatch/dispatch.go
+++ b/pkg/sql/colexec/dispatch/dispatch.go
@@ -136,6 +136,15 @@ func (dispatch *Dispatch) Call(proc *process.Process) (vm.CallResult, error) {
 
 	whichToSend := result.Batch
 	if result.Batch == nil {
+		// A remote dispatch must attach every receiver even when its child
+		// produces no batches. Otherwise an early NotRegistered response can
+		// outlive this pipeline: cleanup removes the registration before the
+		// client's next retry, which then retries until the query timeout.
+		if dispatch.ctr.isRemote && !dispatch.ctr.prepared {
+			if _, err = dispatch.waitRemoteRegsReady(proc); err != nil {
+				return result, err
+			}
+		}
 		result.Status = vm.ExecStop
 		printShuffleResult(dispatch)
 		return result, nil
--- a/pkg/sql/colexec/dispatch/dispatch_test.go
+++ b/pkg/sql/colexec/dispatch/dispatch_test.go
@@ -30,9 +30,21 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/container/types"
 	"github.com/matrixorigin/matrixone/pkg/container/vector"
 	"github.com/matrixorigin/matrixone/pkg/sql/colexec"
+	"github.com/matrixorigin/matrixone/pkg/sql/colexec/value_scan"
 	"github.com/matrixorigin/matrixone/pkg/testutil"
+	"github.com/matrixorigin/matrixone/pkg/vm"
 	"github.com/matrixorigin/matrixone/pkg/vm/process"
 )
+
+type emptyDispatchChild struct {
+	*value_scan.ValueScan
+	called chan struct{}
+}
+
+func (child *emptyDispatchChild) Call(*process.Process) (vm.CallResult, error) {
+	close(child.called)
+	return vm.NewCallResult(), nil
+}
 
 func TestPrepareRemote(t *testing.T) {
 	_ = colexec.NewServer(nil)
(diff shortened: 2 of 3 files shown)


LeftHandCold (approved):
Reviewed head 8ff7e4f63c. No blocking issues found. The zero-batch remote path reuses the existing receiver-registration barrier and preserves the Call/Reset lifecycle: Call only waits for attachment, while terminal delivery remains in Reset. Cancellation and the 30s bound continue through the existing paths, and the regression tests cover both the direct empty-input case and the retry-before-registration interleaving.

Local validation:
- full pkg/sql/colexec/dispatch test suite
- the end-to-end compile regression test, 10 consecutive runs
- race runs for all three added regression tests
- git diff --check

PR #25689, opened 2026-07-13
fix(data-branch): keep snapshot scan state consistent during flush

### LINKED ISSUE #25686: [Bug]: data branch snapshot scan can miss catalog rows during flush
## Branch Name

main

## Commit ID

343b1a7b4bf350b27638c52442b972d70f4a9a7b (PR #25527 merge result; the failing path is unchanged from main)

## Environment

- GitHub Actions standalone multi-CN BVT with proxy
- Run: https://github.com/matrixorigin/matrixone/actions/runs/29258549940/job/86845336838?pr=25527
- Test: `test/distributed/cases/git4data/branch/diff/diff_12.sql`, line 249

## Actual Behavior

`data branch diff t2 against t0 output summary` intermittently fails with:

```text
internal error: cannot find database 282762 (test_chain_diff) created_time at snapshot 1783954047824351071-1
```

The database was committed at `1783954047186529290-0`, well before the query snapshot. In the same CN, transaction, and snapshot, the first two `mo_database` snapshot scans each read 5 rows and found database 282762. A third scan read only 1 row and reported the database missing.

## Expected Behavior

Snapshot scans must return a consistent view while an appendable object is being flushed to a persisted object. The data branch query must not intermittently lose committed catalog rows.

## Steps to Reproduce

Run `test/distributed/cases/git4data/branch/diff/diff_12.sql` in standalone multi-CN + proxy BVT while catalog flush is active. The failure is timing dependent.

## Root Cause

`scanSnapshotRelationByID` obtains current ranges from `Relation.Ranges()`. The returned `relData` embeds the immutable `PartitionState` used to build those ranges. However, `ScanSnapshotWithCurrentRanges` discards that state and calls `tbl.getPartitionState()` again for in-memory rows; `CollectTombstones()` also resolves the latest state independently.

A flush uses copy-on-write to swap the partition state. This permits a mixed view: ranges from the pre-flush state, but rows and tombstones from the post-flush state. The failing timeline proves this exact window:

- 14:47:27.821903: `[FT-11499]1-mo_database` flush starts.
- 14:47:27.829672: ranges contain only the in-memory marker; `disk-block-cnt=0`.
- 14:47:27.829905 and 14:47:27.830842: scans on the pre-flush state read 5 rows and find the database.
- 14:47:27.830680: third scan starts with the same pre-flush ranges.
- 14:47:27.830758: `mo_database` flush ends and swaps state.
- 14:47:27.834885: third scan uses the newer state, reads only 1 row, and misses database 282762.

The fix should use the `PartitionState` embedded in `relData` consistently for persisted ranges, committed in-memory rows, and committed tombstones.

## Additional Information

This is unrelated to PR #25527. That PR only changes prepared-expression JSON comparison paths in frontend/planner/colexec and does not touch data branch, catalog, logtail, snapshot scan, or storage code.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25686

A data-branch diff BVT intermittently failed with `cannot find database ... created_time at snapshot ...` while scanning `mo_database`.

### Root cause

`Ranges` captures an immutable partition-state pointer together with the persisted block ranges. During a concurrent catalog flush, `ScanSnapshotWithCurrentRanges` previously fetched the latest partition state again for in-memory rows and tombstones. If the flush swapped the state between those two steps, the scan combined pre-flush ranges with post-flush rows/tombstones and could miss a committed catalog row.

### Fix

- Use the partition state embedded in `relData` consistently for persisted ranges, committed in-memory rows, and committed tombstones.
- Preserve the existing latest-state fallback when callers pass nil `relData`.
- Add focused regression tests for state consistency and tombstone collection.

### Validation

- `go test -mod=mod ./pkg/vm/engine/disttae`
- `go test -race -mod=mod ./pkg/vm/engine/disttae -run "TestScanSnapshotWithCurrentRangesUsesRangePartitionState|TestCollectSnapshotScanTombstonesUsesProvidedState"`
- Coverage run: 27.5% package statement coverage; new helper 63.2%; changed scan function 66.7%.

Observed in PR #25527 BVT: https://github.com/matrixorigin/matrixone/actions/runs/29258549940/job/86845336838?pr=25527


XuPeng-SH (approved):
Approved after a first-principles review of the complete diff and the snapshot/flush state transition.\n\nThe fix restores the required atomic view: persisted ranges, committed in-memory rows, and committed tombstones are all derived from the immutable PartitionState captured by Ranges(). This closes the pre-flush ranges / post-flush state mismatch without adding synchronization to the scan hot path.\n\nI checked:\n- block-list and object-list state propagation, nil-relData fallback, empty ranges, serial/parallel paths\n- in-memory and persisted tombstone visibility/error propagation\n- iterator/source ownership and cleanup\n- cancellation/wait behavior and bounded memory/CPU cost\n- latest-main compatibility (one unrelated objectio change; merge-tree is clean)\n\nValidation:\n- focused regression and unhappy-path tests: pass\n- focused tests with -race: pass\n- full ./pkg/vm/engine/disttae package: pass\n- go vet ./pkg/vm/engine/disttae: pass\n- diff --check: pass\n\nThe regression test meaningfully pins the invariant: its table cannot fetch a replacement latest state, so it succeeds only when the range-embedded state is consumed.

XuPeng-SH (changes requested):
Requesting changes because the Coverage check completed after my initial review and deterministically fails the repository gate.\n\nThe changed file pkg/vm/engine/disttae/snapshot_scan.go is reported at 68.42%, below the required 75% (job 87137695243). This is not an infrastructure failure: UT and BVT execution succeeded, coverage artifacts were generated, and parse_coverage.py rejected this changed file.\n\nPlease close the missing test/implementation branches before merge:\n- Add a focused test for the documented relData == nil latest-state fallback.\n- Exercise the meaningful tombstone collection/error behavior that remains uncovered.\n- Simplify impossible error branches where appropriate: collectSnapshotScanTombstones constructs readutil.tombstoneData directly, whose AppendInMemory and AppendFiles currently always return nil, and rowsIter.Close also always returns nil. Carrying unreachable generic-error branches increases complexity without adding a real failure mode.\n\nAfter the coverage gate is green, my correctness/concurrency assessment remains positive: the range-embedded PartitionState approach is the right fix, and focused/full/race tests passed locally. This review supersedes my earlier approval, which was submitted while Coverage was still in progress.

Issue #25638, opened 2026-07-12
[Bug]: FULLTEXT top-K retains all matched document state outside the spillable pool


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

`de64386657a07eb3bcb2b411cf7989023b1e16c0`

### Other Environment Information

- Hardware parameters: impact grows with the number of distinct documents matching a term.
- OS type: any supported platform.
- Others: this is independent of the outer SQL `LIMIT`; it is most visible for a high-frequency token and a small top-K query.

### Actual Behavior

The FULLTEXT table function consumes and aggregates the complete internal word-statistics stream before applying its top-K limit. `FixedBytePool` can spill the fixed per-document term vector, but the per-document Go maps remain unbounded and are not spillable:

- `agghtab map[docID]address`
- `docLenMap map[docID]length`
- `docIDMap map[normalizedDocID]originalDocID`

For every distinct matching document, `groupby` adds an entry to these structures. `runWordStats` invokes `PatternToSql` without a candidate limit, and `fulltextIndexMatch` drains the stream to closure before `call` invokes `sort_topk`. Thus a query such as `MATCH(content) AGAINST('common') ... LIMIT 10` still accumulates state for all documents containing `common`.

The pool spill limit does not bound those Go maps. It also cannot guarantee the final scoring phase remains within the pool limit: `sort_topk` iterates the complete aggregate map, and `FixedBytePool.GetItem` unspills a partition without applying the pool's spill admission check.

Relevant code on main:

- `pkg/sql/colexec/table_function/fulltext.go`: initializes the spillable pool and the document maps.
- `pkg/sql/colexec/table_function/fulltext.go`: `groupby` inserts one map entry per distinct document.
- `pkg/sql/colexec/table_function/fulltext.go`: `fulltextIndexMatch` drains all streamed results before returning.
- `pkg/sql/colexec/table_function/fulltext.go`: top-K is evaluated only later in `call`/`sort_topk`.
- `pkg/fulltext/fixedbytepool.go`: the 1 GiB default applies to pool partitions, not Go maps, and `GetItem` can unspill partitions during evaluation.

This is related to previous fulltext large-query incidents (#20213, #21414, #24684), but is a distinct current accounting gap: the implemented spillable byte pool does not include the maps that determine the number of live documents.

### Expected Behavior

[...]

PR #25692, opened 2026-07-13
fix(fulltext): bound top-K aggregation memory to prevent CN OOM (#25638)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25638

The fulltext table function drains its whole postings stream into non-spillable Go maps (agghtab/docLenMap/docIDMap), one entry per matched doc, before top-K. A query matching far more docs than its LIMIT grows the CN heap unbounded and OOMs.

Gate growth in FixedBytePool.NewItem: before creating a new partition (never the first), estimate the partition's added heap (docvec block + per-doc map entries) and refuse to grow if live Go heap would exceed a fraction (default 80%) of total node memory. Gating on live Go heap (which includes the offending maps) against MemoryTotal, rather than on MemoryAvailable, avoids failing open at the cgroup limit and false-aborting under reclaimable page cache. Default partition sized to ~1M items so the common case is a single, ungated partition.


XuPeng-SH (changes requested):
This still needs changes.

The new admission check only guards growth during NewItem, but the final top-K path can still re-materialize spilled partitions without any budget check.

In fixedbytepool.go:
- NewItem gates only when creating a new partition.
- GetItem unspills a spilled partition and unconditionally does pool.mem_in_use += pool.partition_cap.
- There is no corresponding heap-budget check or spill policy on that unspill path.

And in the fulltext table function, sort_topk iterates the whole agghtab and calls u.mpool.GetItem(addr) for every matched document. So after the build phase has already spilled older partitions, the scoring phase can still pull them all back into memory one by one and grow resident pool memory well past the intended limit. That means the PR still does not close the issue's OOM path; it only shifts it later in the query.

A second, smaller hole is that the first partition is explicitly ungated (len(pool.partitions) > 0), so a CN that is already near its heap budget can still fail open on the very first partition plus side maps.

Please either make unspill obey the same admission / eviction policy, or prove that top-K never needs more than a bounded resident working set. As written, the new unit test only covers the second-partition NewItem gate and does not cover the unspill path that sort_topk actually exercises.

fengttt (approved):

gouhongshen (changes requested):
## Codex automated review

The new admission check still permits an unbounded first 1M matched documents, so it does not reliably prevent the CN OOM this PR targets.

### P1 - Enforce the heap budget while filling the first partition (pkg/fulltext/fixedbytepool.go:348)
After the first partition is allocated, every subsequent new document takes this fast path and never executes the budget check. The default partition holds 1<<20 items (lines 321-325), while the check deliberately sets est=0 when there are no existing partitions (lines 364-377). In the actual fulltext aggregation path, each of those documents then adds entries to agghtab/docLenMap/docIDMap after NewItem returns (pkg/sql/colexec/table_function/fulltext.go:611-620). Thus a CN below the 80% budget can accept the first document and grow the non-spillable maps for up to one million distinct matches without any further admission check. At the PR's own 128-byte estimate this is already about 128 MiB of side-map memory, excluding document-ID strings/copies and map overhead, so a constrained CN can still exceed its heap budget and OOM before the second-partition check is reached. Check growth at a bounded interval (or reserve/enforce the first partition's projected side-map cost) and add a test where the budget is crossed while filling partition zero.

XuPeng-SH (changes requested):
This still needs changes on head `1d0cc08c145dd2c2833f2d4f6100e0e8d58d5ac4`.

### P1 — Account for non-Go resident memory in the admission budget

`heapBudget` compares `runtime.MemStats.Alloc` with total node memory, but pool blocks and other mpools are allocated through the C allocator. For example, on a 16-GiB cgroup with 12 GiB of non-Go anonymous/C-allocated memory and a 2-GiB Go heap, this check permits another 10.8 GiB even though only 2 GiB remains. The CN can therefore still be OOM-killed long before the 80% Go-heap threshold. Concurrent queries also have no shared reservation and can multiply the overshoot.

Please gate or reserve combined process/anonymous and Go-map growth, while handling reclaimable page cache explicitly.

### P1 — Bound growth by the actual document-ID size

The claimed `HeapCheckInterval` overshoot bound assumes `MapMemPerItem == 128`, but fulltext preserves the source primary-key type and varchar keys can be 65,535 bytes. Before `NewItem` performs its periodic check, `normalizeDocID` retains both a string copy and a byte-slice copy. With the 16,384-item interval, that permits roughly 2 GiB of persistent key data between checks, excluding map overhead, so a constrained CN can OOM before the next check.

Please account for the actual key size before retaining it and add a high-cardinality varchar-PK regression.

### P1 — Avoid whole-partition I/O per document during top-K scoring

`sort_topk` traverses `agghtab` in randomized map order, while `GetItem` spills and rereads entire partitions on a miss. A diagnostic with 40 partitions and an approximately three-partition resident set produced 120 whole-partition reloads for 120 item reads. The added `TestPoolUnspillBounded` reads addresses sequentially, partition by partition, so it exercises ideal locality and misses the production access pattern.

Please score in partition order, or otherwise guarantee each spilled partition is materialized only a bounded number of times, and test an interleaved/hash-order traversal.

The earlier first-partition, unbounded-unspill, and spilled-`FreeItem` accounting gaps are closed in the normal path. Focused `pkg/fulltext` tests, `-race`, build/vet, and dependent table-function tests pass; the blockers above are remaining Q3 memory/I/O-bound failures.

gouhongshen (changes requested):
## Codex automated review

The prior first-partition and unspill-residency issues are fixed, but the remaining accounting and scoring paths can still OOM or make large searches effectively uninterruptible.

### P1 - Account for variable-length document IDs before the interval check (pkg/fulltext/fixedbytepool.go:70)
The 128-byte estimate is only valid for integer IDs. FULLTEXT supports varchar/composite primary keys (up to 65,535 bytes), and `normalizeDocID` retains both a string map key and an `append`ed byte copy before `groupby` calls `NewItem`. The fast-path check then permits 16,384 such new documents before re-reading heap usage, i.e. roughly 2 GiB of retained key bytes alone at the supported maximum width. A constrained CN can therefore be OOM-killed before this guard runs. Account for the actual key allocation before retaining it (or check/reserve at that point) and add a high-cardinality large-varchar-PK regression.

### P1 - Budget total resident memory, not only runtime.Alloc (pkg/fulltext/fixedbytepool.go:129)
`heapBudget` compares only Go `runtime.MemStats.Alloc` with the cgroup/node limit, but the partition data is deliberately allocated through the C allocator and is excluded from that value. The pool may retain its default 1 GiB off-heap resident set while this check allows Go maps to approach 80% of the same cgroup; on a 4 GiB CN that admits roughly 3 GiB Go plus 1 GiB pool memory, before other C allocations. The default global mpool cap is 1 TiB, so it does not provide a cgroup-sized combined reservation either. Concurrent queries widen the same gap. This does not establish the claimed CN OOM bound; admission needs a shared/process-level budget that accounts for both resident off-heap memory and Go growth.

### P1 - Avoid partition-sized I/O on every randomized top-K document access (pkg/fulltext/fixedbytepool.go:475)
`sort_topk` ranges `agghtab`, whose key order has no relationship to partition ID. For each spilled address, this new path can spill up to 16 whole resident partitions and then unspill the requested whole partition. With more partitions than the resident set, an interleaved map traversal can therefore write and reread a partition for each document rather than once per partition. `sort_topk` also has no context-cancellation check, so cancellation cannot stop that I/O loop promptly. `TestPoolUnspillBounded` reads addresses sequentially by partition and asserts only memory residency, so it misses this production order. Score/group by partition (and check cancellation), with an interleaved-address regression that bounds reloads.

fengttt:
@XuPeng-SH @gouhongshen all three remaining P1s are closed in 2a3550a5e1 (pushed via maintainer edit — @cpegeric please review), and the branch is merged with current main:

**1. Resident-memory budget (not only `runtime.Alloc`).** `heapBudget` now adds the process-wide mpool counter (`mpool.GlobalStats().NumCurrBytes`) to the live Go heap. That accounts the off-heap partition blocks, *concurrent queries'* pools, and every other mpool consumer — closing the shared-reservation gap — while still excluding reclaimable page cache by construction (mpool bytes are anonymous memory), so the original page-cache false-abort concern stays addressed. Scope note: non-mpool C allocations remain uncounted; mpool is MO's dominant query allocator, and the hook is stubable if a broader signal lands later.

**2. Actual doc-ID size accounting.** `normalizeDocID` now charges the retained bytes (string key + `[]byte` copy) into the pool via `ChargeSideBytes`; the fast-path budget re-check triggers on **either** 16384 items **or** 16 MiB of accumulated side bytes (`HeapCheckBytes`). Worst-case overshoot is bounded by bytes, independent of key width — the ~2 GiB-between-checks hole at the 65,535-byte varchar maximum is gone.

**3. Partition-ordered top-K scoring + cancellation.** Both `evaluate` and `sort_topk` now bucket `agghtab` keys by partition id (one O(n) pass) and score partition-by-partition, so each spilled partition is materialized **at most once per pass**; both loops check `proc.Ctx` every 1024 documents so KILL/timeout stops the I/O promptly.

**Regressions — each verified to fail without its fix:**
- `TestPoolBudgetIncludesMpool` — off-heap bytes alone push admission over budget.
- `TestPoolSideBytesCrossBudget` — wide-key byte accumulation forces the re-check long before the item interval.
- `TestSortTopKBoundedUnspills` — 64 docs / 16 partitions / 2-partition resident set: hash-order traversal produced **60 unspills**, the partition-ordered pass is asserted **≤ 16**. This is the interleaved/hash-order test you asked for (the pool grew an `Unspills()` counter to make reload counts observable).

Both packages pass with `-race` ×2; existing budget tests were made deterministic against the new mpool term.

gouhongshen (changes requested):
## Codex automated review

Previously blocking resident-mpool accounting and wide document-ID interval accounting are fixed. Partition-ordered scoring fixes the prior map-order unspill thrash, but its new key-materialization path introduces a distinct blocking memory/complexity regression for unlimited-result searches.

### P1 - Avoid rebuilding a full aggregation-key copy for every 8K result batch (pkg/sql/colexec/table_function/fulltext.go:413)
`partitionOrderedKeys` appends every remaining `agghtab` key into partition buckets, then copies them all again into `keys`. `evaluate` consumes at most 8,192 scored results before returning (lines 476-478), while the zero-LIMIT path invokes `evaluate` again on each output batch (lines 274-282). Thus an N-document no-LIMIT search allocates O(N) temporary key storage on every batch and performs O(N²/8192) map traversal/copy work overall. At its peak the two interface-slice copies add roughly 32 bytes per remaining document, outside the pool's heap-budget accounting; after aggregation has legitimately approached the 80% admission threshold this extra allocation can exhaust the remaining headroom and OOM the CN. The cancellation check is also reached only after this full allocation/traversal. Keep partition ordering with bounded, budgeted traversal state/workspace (and cancellation during its construction), and add a regression covering multiple `evaluate` batches rather than only the one-pass `sort_topk` case.

fengttt:
The new P1 is correct — `partitionOrderedKeys` was rebuilt on every `evaluate` call while the zero-LIMIT path re-enters `evaluate` per 8K output batch: O(N) unaccounted workspace per batch (~32 B/remaining doc across the two interface-slice copies), O(N²/8192) traversal work overall, and the cancellation check only reachable after the full rebuild. Since aggregation completes in `start()` before scoring, the ordering can be snapshot once.

**Fixed in be0b1108a6:**
- `partitionOrderedBuckets` replaces `partitionOrderedKeys`: returns the per-partition buckets directly (no flatten copy — halves the peak), **gates the ~16 B/doc workspace on the pool's heap budget** via the new `FixedBytePool.CheckBudget` (same admission error path as partition growth), and **honors cancellation every 1024 keys during construction**.
- `evaluate` builds the buckets **once per scoring phase** (`scoreBuckets`/`scoreOrdered` on `fulltextState`, cleared in `resetRowState`) and drains them across batches, releasing each bucket as it empties. Spilled partitions are now materialized at most once across **all** batches, not once per batch.
- `sort_topk` uses the same budgeted, cancellable constructor for its single pass.

**Regressions — the requested multi-batch coverage, each verified to fail without the fix:**
- `TestEvaluateMultiBatchBoundedWork` — 20,000 docs, ~10 partitions, 2-partition resident set, 3+ `evaluate` batches. Asserts: every doc scored exactly once across batches; a key injected into `agghtab` *after* the first batch is **not** discovered (a per-batch rebuild scores it — this is the assertion that fails on the old code); total unspills bounded by the partition count across all batches.
- `TestEvaluateOrderingBudgetGated` — with `HeapBudgetPct` forced to 0, `evaluate` refuses to allocate the ordering workspace with the standard over-budget error instead of allocating first.

Both packages pass `-race` ×2, vet clean. @cpegeric pushed via maintainer edit again.

gouhongshen (approved):
## Codex automated review

Previously blocking resident-mpool accounting, wide document-ID accounting, and partition I/O thrashing were fixed before this update. The latest update fixes the remaining multi-batch blocker by building the budgeted, cancellable partition traversal once and draining it across calls. No blocking or non-blocking defects remain. Focused local tests could not run because the required cgo/libmo.dylib artifact is absent.

XuPeng-SH (changes requested):
Requesting changes on exact head `be0b1108a6c64451eb5e8f033b9f0db0b1b53cde`.

The latest update does fix the previously reported O(N²/8192) traversal rebuild, and the earlier mpool, wide-key, and partition-thrashing blockers are materially improved. Three memory-bound paths still prevent this from closing the CN OOM issue:

### P1 — Make the workspace admission estimate match the allocation

`partitionOrderedBuckets` gates only `len(agghtab) * 16`, then appends into zero-capacity per-partition slices (`pkg/sql/colexec/table_function/fulltext.go:442-458`). That is not the peak allocation: every slice growth allocates a new backing array, prior arrays remain live until a GC, capacity growth adds slack, and the outer slice/headers are not counted.

On this exact head, a diagnostic with 200,000 keys in one production-shaped partition admitted 3,200,000 bytes but the constructor allocated 17,942,920 bytes before an intervening GC (5.61× the estimate). Aggregation is explicitly allowed to approach the 80% ceiling before this constructor runs, so this unreserved gap can cross the ceiling and OOM in the phase the PR claims is budgeted.

Please pre-count and preallocate exact bucket capacities (or use one exact flat key buffer with partition offsets), include the outer metadata/capacity in the estimate, and add a regression that measures/bounds the actual constructed workspace rather than only forcing `HeapBudgetPct = 0`.

### P1 — Bound candidate work when scoring produces few results

`evaluate` stops on `len(scoremap) == 8192`, but appends every processed candidate to the local `keys` slice (`fulltext.go:483-525`). A valid boolean query such as `+apple -banana` can aggregate a very large union and then produce no score for most or all candidates. In that case one call consumes all N candidates and grows `keys` to N, adding a second ungated O(N) interface buffer even though the new persistent traversal is already O(N).

A 200,000-document diagnostic on this head allocated 38,996,024 bytes for the all-filtered call versus 18,576,440 bytes for a normal dense first batch: 20,419,584 extra bytes, growing linearly with N. The added multi-batch test makes every document score and misses this shape.

Because traversal no longer ranges the map directly, each processed document can be freed/deleted immediately; also bound each call by candidates processed, not only results produced, and nil consumed bucket slots so variable-width IDs can be reclaimed incrementally. Add a sparse/empty-score regression.

### P1 — Use the effective cgroup ceiling, not host total

`memTotal` is still `system.MemoryTotal` (`pkg/fulltext/fixedbytepool.go:57-58`). `MemoryTotal` only records the cgroup limit when the process is PID 1; for a CN under an init process, nested container, or systemd scope it is host RAM. The current base already provides `system.CgroupMemoryLimit()` specifically for non-PID-1 processes, but this budget ignores it.

For example, on a 256-GiB host with a 4-GiB process cgroup, this computes an approximately 205-GiB ceiling instead of 3.2 GiB and therefore fails open until the kernel OOM-kills the CN. Resolve the ceiling from the minimum finite cgroup/host (and any stricter process/global cap), as the current hash-build budget does, and cover the non-PID-1/cgroup-smaller-than-host case.

Validation performed: all 11 new exact tests passed normally; focused race repetition passed 47× for `pkg/fulltext` and 43× for `pkg/sql/colexec/table_function`; both complete owning packages passed under `-race`; both packages are vet-clean. The exact head also merges cleanly into latest `origin/main` (`dce208ba8cfdb32a2b4fee108824d28819416b80`) and the focused tests pass on that prospective merge. These blockers are missing memory-shape coverage, not existing CI failures.

fengttt:
@XuPeng-SH thanks — the first two P1s are verified and fixed in ffe48f4179 (branch also merged with current main):

**P1 — estimate == allocation.** Confirmed: an emulation of the append-grown buckets reproduced your diagnostic almost exactly (21.1 MB constructed vs 3.2 MB admitted at 200k keys). `partitionOrderedKeys` now builds ONE flat `[]any` of exactly `len(agghtab)` plus a single per-partition offset array in two counted O(n) passes (both cancellation-checked) — no append growth, no per-bucket slack. The admission estimate lives in `scoreTraversalEstimate`, one definition shared by the budget gate and the new regression. `TestScoreTraversalWorkspaceExact` measures the constructor's real `TotalAlloc` and requires it ≤ the admitted estimate (+256 KiB fixed slack), plus a constant-allocation-count structural check; it fails at 21.1 MB vs 3.2 MB against the old shape.

**P1 — sparse results.** Confirmed: the all-filtered emulation allocated 39.0 MB at 200k candidates, matching your measurement. `evaluate` now frees each candidate the moment it is scored (`FreeItem` + side-map deletes), nils consumed traversal slots so wide doc IDs reclaim incrementally, and no longer keeps any per-candidate accumulation — per-call memory is bounded by the returned scoremap. `TestEvaluateSparseScoreBounded` drives 200k zero-score candidates through one call and asserts empty scoremap, fully-drained side maps, and total allocation ≤ traversal estimate + 2 MiB (fails at 39 MB on the old shape). On the per-call candidate cap specifically: the table-function protocol treats an empty batch as end-of-stream, so evaluate cannot return early with zero results without truncating the query — with immediate release the all-filtered pass now carries no per-candidate memory, which I believe was the substance of the concern.

**P1 — cgroup ceiling: deliberately not taken in this PR** (maintainer decision, @fengttt). The ceiling source is a stubable seam (`memTotal`), and swapping the budget's memory-topology signal (host vs `CgroupMemoryLimit` vs nested v1/v2 scopes) changes admission behavior for every deployment shape — a mis-detected small limit would turn the budget into false query aborts, which is worse than the conservative host ceiling for this fix. Happy to track the ceiling-source question as a separate issue.

Both packages pass `-race` ×2, vet clean. @cpegeric pushed via maintainer edit.

fengttt:
UT CI failure triaged: `TestGlobalCheckpoint2` timed out in `WaitAllCheckpointsFinished` with the last forced checkpoint still in flight ("checkpointed LSN: 8, target: 9, pending: 1") — the 10s `DefaultCheckpointWaitTimeoutMS` expired on a loaded runner (the same job logged >1000 GC cron iterations inside that window; the test passes locally in 0.24s). Unrelated to the fulltext change, and the same flake class as #26245, which raised `TestEngine.ForceCheckpoint`'s deadline to two minutes.

Fixed in f8ead8b034 by aligning `WaitAllCheckpointsFinished` with that two-minute safety net — `WaitExpect` polls, so green runs still return as soon as the condition holds; only a genuinely stuck checkpoint waits longer.

PR #25701, opened 2026-07-14
perf(frontend): reuse SELECT result columns

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25472

The ordinary `SELECT` response path built result column definitions twice: once
to produce MySQL protocol columns and again to populate `Session.rs`. Under the
prepared point-select workload, the duplicate `GetResultColumnsFromPlan` call
was a recurring allocation hotspot.

This PR builds the result column definitions once for a `SELECT` handled by
`TxnComputationWrapper`, converts that same snapshot into MySQL columns, and
reuses it for `Session.rs`. Other computation wrappers and statement types keep
the existing path, including the special `SHOW COLUMNS` metadata handling.

Changes:

- split MySQL column conversion from result column construction;
- return MySQL columns and result column definitions together for ordinary
  `SELECT` statements;
- cover metadata equivalence, result-column mutation isolation, `SHOW COLUMNS`
  behavior, prepared compile reuse, and allocation counts.

Validation:

- `go test -race ./pkg/frontend -run 'TestGetSelectColumns' -count=1`
- `go test ./pkg/frontend -run '^$' -bench '^BenchmarkGetSelectColumns$' -benchmem -count=3`
  - duplicate construction: 736 B/op, 6 allocs/op
  - reused construction: 440 B/op, 4 allocs/op
- Remote 1000-VU point-select A/B/A on `mo-srv-129`, based on `2865ed6cc`:
  - baseline mean: 177,857 TPS, 25.0 KB and 317.1 mallocs per query
  - patched mean: 178,904 TPS, 24.7 KB and 314.8 mallocs per query
  - `GetResultColumnsFromPlan.func1` alloc space: 3.02 GB/30s to 1.44 GB/30.7s

Throughput and average latency remained neutral; the intended benefit is lower
allocation churn on the SELECT response path.


XuPeng-SH (approved):
Reviewed the current head (054802d8). The SELECT-only fast path preserves metadata semantics and ownership isolation while eliminating the duplicate result-column construction. SHOW COLUMNS and non-Txn wrappers retain the legacy path, and prepared execution reuse is covered. I found no blocking correctness, lifecycle, unhappy-path, or performance issue. Local validation passed: focused tests x20, focused race test, full frontend package, build, vet, and the benchmark consistently reproduced 736 B/6 allocs down to 440 B/4 allocs.

PR #25706, opened 2026-07-14
fix: honor UPDATE IGNORE on duplicate keys

### LINKED ISSUE #25347: [Bug]: UPDATE IGNORE still errors on duplicate unique-key updates

Existing issue #15345 covers `INSERT IGNORE` reporting errors for several incompatible rows. This issue is for `UPDATE IGNORE`: MySQL skips the duplicate-key update and keeps the row unchanged, while MatrixOne returns a duplicate-key error.

### Branch Name

main

### Commit ID

`0e8602b78`

- MySQL oracle: MySQL 8.0.45

### Actual Behavior

`UPDATE IGNORE` still reports a duplicate unique-key error when the update would conflict with an existing unique key.

Observed on MatrixOne latest `main`:

```text
ERROR 1062 (HY000): Duplicate entry '(b,x)' for key '(a,b)'
```

The data remains unchanged after the failed statement, but the statement should not error in MySQL-compatible `IGNORE` mode.

### Expected Behavior

MySQL skips the conflicting update row and keeps it unchanged:

```text
id  a     b
1   b     x
2   NULL  x
3   a     x
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS mysql_compat_update_ignore_min;
CREATE DATABASE mysql_compat_update_ignore_min;
USE mysql_compat_update_ignore_min;

CREATE TABLE t (
  id int primary key,
  a varchar(16) null,
  b varchar(16) null,
  unique key uq_ab(a,b),
  index idx_ab(a,b,id)
);

INSERT INTO t VALUES (1,null,'x'),(2,null,'x'),(3,'a','x');

UPDATE t SET a='b', b='x' WHERE id=1;

UPDATE IGNORE t SET a='b', b='x' WHERE id=2;

SELECT id, a, b FROM t ORDER BY id;
```

### Additional information

MySQL 8.0.45 result:

```text
id  a     b
1   b     x
2   NULL  x
3   a     x
```

MatrixOne latest `main` at `0e8602b78`:

```text
ERROR 1062 (HY000): Duplicate entry '(b,x)' for key '(a,b)'
```

This was found while backfilling Model 3 NULL / unique composite-index compatibility coverage. The case also exercises MySQL's rule that unique indexes permit multiple `NULL` values, then checks `UPDATE IGNORE` behavior when a later update would create a non-NULL duplicate key.

aptend:
## Summary

- preserve the `IGNORE` modifier in the UPDATE AST and formatter
- use the existing dedup `Node_IGNORE` action for primary-key and unique-key conflicts
- filter conflicting rows before `MULTI_UPDATE`, keeping their original data unchanged
- add parser and planner regression coverage

`SHOW WARNINGS` emission is unchanged; this PR fixes the reported skip-without-error behavior.

## Validation

- `go test -count=1 ./pkg/sql/parsers/dialect/mysql`
- `go test -count=1 -run ^TestUpdateIgnoreUsesIgnoreDedupAction$ ./pkg/sql/plan`
- `go vet ./pkg/sql/parsers/dialect/mysql ./pkg/sql/plan`
- `make build`
- reproduced #25347 against the rebuilt local server: the conflicting row is skipped, `ROW_COUNT()` is 0, and the row remains unchanged

Fixes #25347


XuPeng-SH (changes requested):
The original #25347 reproduction now passes, but this change introduces a correctness regression: UPDATE IGNORE drops a non-conflicting row whenever an assigned primary/unique key keeps its current value.

Reproduced on head 948c3a52aa472d64fb26f3af15977da885c12ef3:

    CREATE TABLE t(id INT PRIMARY KEY, a INT UNIQUE, v INT);
    INSERT INTO t VALUES (1, 10, 0), (2, 20, 0);
    UPDATE IGNORE t SET id = id, v = v + 1 WHERE id = 1;
    SELECT * FROM t WHERE id = 1;

Expected v = 1; actual v = 0. The same happens with SET a = a, v = v + 1.

The planner still populates DedupJoinCtx.OldColList, but constructDedupJoin/constructRightDedupJoin only propagate it as DelColIdx when OnDuplicateAction == Node_FAIL. After this PR switches the action to Node_IGNORE, the target rows own old key is no longer excluded, and the IGNORE probe path treats the self-match as a duplicate and filters the whole update. The Node_IGNORE probe branch also does not honor IsDeleted as the FAIL branch does.

Please preserve old-row exclusion for IGNORE and add execution-level regression coverage for both the issue case and a no-op key assignment combined with a real non-key update. The current plan-only assertion verifies the enum but cannot catch this data-loss-of-update behavior.

XuPeng-SH (approved):
Re-reviewed the latest head d242c9e9.

The previous blocker is fixed: UPDATE IGNORE now propagates the old-row identity, excludes its own old-key match, preserves an existing key owner independent of input order, and rebuilds the survivor/group mappings after filtering. I also checked multi-unique-key interactions, nullable unique keys, shuffle and join orientation, cleanup, and bounded-memory behavior.

Local validation passed:
- focused hashbuild and dedupjoin regressions repeated 20x
- planner regressions repeated 10x and parser package repeated 5x
- full relevant package tests
- focused race tests repeated 5x
- go vet on all relevant packages

The current CI is green. I found no remaining correctness, performance, or unhappy-path merge blocker.

PR #25707, opened 2026-07-14
fix(tae): return open errors from storage init

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25677

  - change rpc.NewTAEHandle to return TAE open errors instead of panicking
  - propagate TAE open failures through taestorage.NewTAEStorage
  - close an already-opened TAE handle if later storage initialization fails
  - add unhappy-path tests for open failures and cleanup after logtail server creation errors


XuPeng-SH (changes requested):
The panic-to-error direction is correct, but the initialization failure path is not ownership- or generation-safe yet. I found two blockers.

1. A transient logtail-server creation failure permanently closes the QueryClient reused by the retry loop.

`newTAEStorage` rolls back by calling `taeHandler.HandleClose(ctx)`, and `HandleClose` closes both the TAE DB and `h.client`. However, that QueryClient is created and owned by `tnservice.store`; every `createReplica` retry passes the same `s.queryClient`. After the first post-open initialization failure, the retry therefore receives an already-closed client. The storage may subsequently start, but operations using that client will return client-closed errors.

Please make ownership explicit: the handle should close resources it creates (the DB), while the store should close its shared QueryClient exactly once during store shutdown. Add a regression test proving that a failed storage attempt closes the DB without closing the caller-owned client, and that the same client remains usable by the retry.

2. The failed TAE generation is published globally before storage initialization commits.

`NewTAEHandle` calls `RegisterManifestHTTP(tae)` immediately after opening the DB. `RegisterManifestHTTP` uses `sync.Once` and permanently captures that first DB pointer. If logtail-server creation then fails, this PR closes the DB, but the HTTP handler still points to the closed generation and a later successful retry cannot replace it.

I reproduced this deterministically: inject the logtail-server factory failure, let `newTAEStorage` return, then serve `GET /debug/tae/manifest?table=db.tbl` through the default mux. It panics with a nil-pointer dereference because `parseTableTarget` ignores the failed `StartTxn` result and dereferences the nil transaction.

Please publish the active DB only after the whole storage constructor succeeds, and make the handler generation-aware (atomically replace/clear the active DB without allowing an old close to clear a newer generation). `parseTableTarget` should also handle `StartTxn` errors rather than panic. Tests should cover failed initialization, retry to a new generation, endpoint access between/after generations, and stale-generation cleanup.

The changed package tests and `go vet` pass locally, but the current tests use a nil QueryClient and only prove that the database directory can be reopened, so they do not exercise either ownership or generation boundary.

XuPeng-SH (changes requested):
The previous QueryClient premature-close and stale manifest-generation blockers are fixed, but the ownership transfer is still incomplete and now leaks the QueryClient on store shutdown.

`HandleClose` correctly no longer closes `h.client`, because the handle does not own the client. However, that client is created by `tnservice.store` via `client.NewQueryClient`, and `store.Close()` still never calls `s.queryClient.Close()`. `queryService.Close()` is unrelated—it only closes the server-side handler. Consequently, the underlying morpc client remains registered in `globalClientGC.clients`, its backends are not closed, its stopper is not stopped, and the active-client gauge is not decremented. Repeated store lifecycles therefore retain clients/backends in the global manager.

Please complete the ownership move by closing the shared QueryClient exactly once from the store shutdown path, after replica users have stopped, and include its error in the aggregate close result. Add a lifecycle regression proving that closing/retrying an individual TAE storage does not close the shared client, while closing the owning store closes it once.

The new TAE open-error, cleanup, and manifest-generation tests pass locally on this head. I found no additional correctness or performance blocker: manifest dispatch adds only an atomic load, and the reviewed wait paths introduce no new hang risk.

XuPeng-SH (changes requested):
The previous QueryClient ownership/leak and manifest-generation blockers are fixed, and the production `NewTAEStorage` path now propagates open errors correctly. One blocking contract issue remains: the canonical exported `rpc.NewTAEHandle` still converts `newTAEHandle` errors into a panic.

This is the exact API and failure mode reported by #25677, and it also contradicts this PR's description (“change rpc.NewTAEHandle to return TAE open errors instead of panicking”). The new regression only calls the sibling `NewTAEHandleWithError`, so it bypasses the original reproducer. I reran that reproducer against the current head with `/dev/null/cannot-create-tae-store`; `NewTAEHandle` still panics 3/3. Any current or future caller choosing the canonical exported constructor therefore retains the process-crash behavior this PR claims to remove.

Please make `NewTAEHandle` itself return `(*Handle, error)` and update its callers/tests. If a panic convenience is genuinely required, expose it explicitly as a `MustNewTAEHandle`-style API rather than leaving the canonical constructor unsafe. The regression should invoke `NewTAEHandle` on an invalid path and assert error return plus no panic.

Non-blocking cleanup: after the merge from main, `queryClientFactory` has no assignment/use outside `NewService`; the stronger branch-local shutdown test was replaced by main's existing test. Please remove this now-dead production test seam unless it is used by a retained lifecycle test.

Local validation otherwise passed: focused unhappy-path tests x5, focused race tests, all four changed packages, build, vet, diff check, and a clean virtual merge with latest main.

XuPeng-SH (approved):
The latest head closes the remaining constructor-contract blocker, and I found no additional correctness, lifecycle, performance, or unhappy-path blocker.

The fix is now complete at the API boundary: `rpc.NewTAEHandle` returns `(*Handle, error)`, all in-repository callers handle the error, the production `NewTAEStorage` path propagates it, and panic semantics are isolated behind the explicit `MustNewTAEHandle` name. The regression invokes the canonical constructor on the original invalid-path reproducer and verifies error return without panic.

The earlier ownership and generation findings also remain correctly resolved: a Handle owns and closes its TAE DB but not the caller-owned shared QueryClient; `tnservice.store` closes that client once after replica users stop; failed post-open initialization closes only the failed DB; the manifest DB is published only after constructor success; atomic replacement plus generation-specific CAS cleanup prevents an old close from clearing a newer generation; and closed/unavailable endpoint paths return errors instead of dereferencing a nil transaction.

First-principles audit: Q1 has one effective cleanup owner for DB, logtail server, and QueryClient; Q2 introduces no unbounded wait dependency and all failure paths return or clean up; Q3 retains only one atomic manifest pointer and no new unbounded accumulation. The only request-path overhead is one atomic pointer load on a debug HTTP endpoint; initialization/cleanup changes are off the data hot path.

Local validation on `00a87aa75195363b7ce39f197d33cee9944e0fc0` passed: focused semantic matrix across open failure, canceled open, partial-init rollback, shared-client ownership, generation replacement/stale cleanup, closed endpoint, and store shutdown (`-count=3`); the same focused matrix under `-race`; complete tests for `pkg/vm/engine/tae/rpc`, `pkg/txn/storage/tae`, and `pkg/tnservice`; complete race tests for those packages; testutil compilation; `go vet` for all changed packages; PR diff check; and a clean virtual merge with latest main followed by the complete changed-package test set. CI is still in progress and should remain enforced by the merge gate.

Issue #24915, opened 2026-06-10
[Bug] LockService row-lock holder blocks sysbench1000w read_write for 3h49m — 32k+ `wait too long`, lock released only on client broken pipe


heni02:
## Environment

| Field | Value |
|-------|--------|
| GitHub Actions run | [27181418714](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/27181418714) |
| CI step | `SYSBENCH OLTP TEST 1000W` → `Start OLTP Read-Write Test With 1000 Threads` |
| Namespace | `mo-main-commit-024703548-20260609` |
| MO commit | `024703548` (main) |
| Cluster | Guangzhou nightly TKE |
| Workload | sysbench `oltp_read_write.lua`, 10 tables × 10M rows, **1000 threads**, `--time=300` |
| Job timeout | `timeout-minutes: 280` |
| Incident window | **2026-06-09 17:30 CST → 21:20 CST** (~3h50m) |

**Evidence source:** Loki queried from **Guangzhou cluster** via `nightly-devops` jump host → `kubectl exec -n mo-ob loki-read-b959847bc-5wpbg` (in-cluster read, port 3100). Time range matches Grafana Explore: `2026-06-09 11:35:19` – `22:15:41 CST`.

## Summary

During sysbench1000w **Read-Write 1000 threads**, a **single row-lock holder txn** on CN pod `nightly-regression-dis-tp-cn-pgb6g` holds `Exclusive(row)` for **3h49m38s**. LockService logs **`wait too long`** continuously but **does not abort waiters or break the stale holder**. All **32,610+** warnings (Grafana `count_over_time`) come from this one pod. The lock is released only when the client connection breaks (`broken pipe`) and CN rolls back the holder txn — not via LockService timeout.

This is a **MatrixOne LockService defect**, not txn-scheduler stall (`wait active = 0`) and not CN-roll orphan holder (no OOM/restart/bind-loss on `pgb6g`).

## Timeline

| Time (CST) | Time (UTC) | Event | Source |
|------------|------------|-------|--------|
| 14:52 | 06:52 | COS PUT `connection reset` on **`tp-cn-pgb6g`** during LOAD (upstream, separate issue) | Cluster Loki |
| 16:40 | 08:40 | Second COS PUT reset burst on **`tp-cn-pgb6g`** | Cluster Loki |
| **17:30:48** | **09:30:48** | Lock wait starts (`21:20:26 − 3h49m38s`) | Derived |
[...]

PR #25708, opened 2026-07-14
fix: bound default lock wait timeout

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24915 

fix: bound default lock wait timeout


XuPeng-SH (approved):
Re-reviewed latest head `8e246e18` from first principles. I didn't find a new merge blocker.

The change is internally consistent across the full closure it touches:
- the default session/global `lock_wait_timeout` is now bounded to 60s instead of effectively unbounded,
- `lockservice.ErrLockTimeout` now surfaces as a dedicated `moerr.ErrLockWaitTimeout` with MySQL `ER_LOCK_WAIT_TIMEOUT`,
- frontend still treats that as a statement-level failure rather than a whole-txn rollback, and
- the explicit/zero timeout semantics remain intact (`LockWaitTimeout=0` still falls back to caller context / remote-owner timeout instead of inheriting the new default).

The added coverage is also on point: variable defaulting, dedicated moerr mapping, local/remote lockservice timeout behavior, lockop retry classification, and the remote BVT path that exercises `SET SESSION lock_wait_timeout = 1` after `BEGIN`. LGTM.

heni02 (approved):

aunjgr (approved):

PR #25712, opened 2026-07-14
perf(plan): choose low-cost row carriers

### LINKED ISSUE #25697: [SubTask]: Low-cost row carriers and zero-column scan/UNION/scalar-agg fallbacks
## Brief Description

Part of #25612. Replace the implicit "keep slot 0" fallback with a first-class row-cardinality requirement and a deterministic low-cost carrier selector.

- Model row cardinality as a first-class requirement in `remapAllColRefs` (`pkg/sql/plan/query_builder.go`): distinguish "needs a column value" from "needs only to preserve row count".
- Add a carrier selector: prefer zero-column batches when all affected operators support them; otherwise pick a deterministic minimum-cost carrier by type width and aggregate cost (avoid wide varlen columns and expensive aggregates such as `group_concat`).
- Apply to the three slot-0 fallbacks: table/function scan, `UNION ALL`, and scalar aggregate.
- Preserve side-effecting expressions and all group/order/partition dependencies. Scalar aggregation with all outputs discarded must still keep a row-producing carrier.

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25697

- Models row-cardinality preservation explicitly during column remapping.
- Replaces slot-0 fallbacks for scans, `UNION ALL`, and scalar aggregates with deterministic low-cost carrier selection.
- Scores fixed-width and varlen values, aggregate state cost, and aggregate input width while preserving side effects and row-count semantics.
- Adds selector unit tests, plan-shape coverage, and column-pruning BVT cases for wide columns, `UNION ALL`, and empty/non-empty scalar aggregates.

## Testing

- `go test ./pkg/sql/plan ./pkg/sql/plan/explain -count=1`
- `go test -race ./pkg/sql/plan -run '^TestChooseRowCarrier$' -count=1`
- `go vet ./pkg/sql/plan ./pkg/sql/plan/explain`
- `make build`
- `make static-check`
- `column_pruning.test`: 29/29 passed with a clean data directory and freshly built service
- `git diff --check`


XuPeng-SH (changes requested):
Blocking correctness regression in the row-carrier selection for external scans.

`chooseTableRowCarrier` considers every `TableDef.Cols` entry, including hidden columns that an external reader does not materialize. For an external CSV table whose visible columns are varlen, a fixed-width hidden column can win the cost comparison. Column pruning then leaves only that hidden carrier, while `buildExternalAttrs` drops hidden columns. `External.Prepare` consequently creates a batch with zero vectors, and `CsvReader.makeBatchRows` panics at `bat.Vecs[0]`.

This is reproduced deterministically on this exact head by all three failed BVT jobs (standalone pessimistic, standalone proxy, and compose pessimistic). Each fails in `external_csv_comment.sql` on:

```sql
select count(*) from t_default;
select count(*) from t_hash;
select count(*) from t_rem;
```

with `index out of range [0] with length 0` at `reader_csv.go:190`. The ordinary projection tests in this PR do not cover an external scan with hidden/non-materializable columns.

Please make carrier selection node-aware so it only chooses columns the scan implementation can actually produce (at minimum for `EXTERNAL_SCAN`), and add a regression test that executes `COUNT(*)` against an external CSV table with varlen visible columns. A defensive empty-vector handling check in the external reader would also prevent this failure mode from becoming a panic.

heni02 (approved):

XuPeng-SH (approved):
Re-reviewed the latest head `70e22c782c`. The external-scan fix now keeps hidden/non-materialized carriers out of `chooseTableRowCarrier` for `EXTERNAL_SCAN`, and the defensive zero-vector guard in `CsvReader.makeBatchRows` closes the previous panic mode cleanly. I reran `go test ./pkg/sql/plan ./pkg/sql/colexec/external -count=1` and `go test -race ./pkg/sql/plan ./pkg/sql/colexec/external -count=1`; I do not see a remaining high-confidence correctness or concurrency blocker on this head.

aunjgr (approved):

Issue #25691, opened 2026-07-13
[Refactoring]: Decouple remote pipeline streams from CN-to-CN TCP backend lifetime


XuPeng-SH:
### Why do you want to refactor this code?

The remote pipeline transport currently conflates three different lifecycles:

1. a logical remote pipeline stream;
2. a morpc backend;
3. a physical CN-to-CN TCP connection.

PipelineClient.NewStream requests a locked backend, so one active pipeline stream exclusively owns one backend/TCP connection. messageSenderOnClient.close then calls Stream.Close(true), which closes the entire backend connection when that logical stream finishes.

This coupling was introduced to protect correctness. Historical issue #9545 notes that blindly sharing a TCP connection between cyclic pipelines can fill TCP buffers and deadlock. However, closing one TCP connection per logical stream creates a different systemic failure mode:

- complex distributed queries create hundreds of logical streams;
- concurrent queries require thousands of simultaneous TCP connections;
- completed streams generate a high rate of closed sockets and TIME_WAIT tuples;
- the default 100000-backend-per-host limit is higher than a typical Linux ephemeral-port budget;
- there is no context-aware admission point at the actual constrained resource;
- sustained load eventually fails with EADDRNOTAVAIL, as reproduced in #25690.

The compile layer also carries transport-specific flags such as safeToClose and alreadyClose. Normal terminal completion, query cancellation, protocol failure, and connection poisoning are not represented as an explicit ownership/state model. This makes the lifecycle difficult to reason about, test independently, and evolve safely.

### Describe the solution you'd like

Refactor remote pipeline transport around explicit and independently testable lifecycle contracts.

1. Define explicit ownership and states

   - Model logical stream states such as created, attached, active, terminal-confirmed, canceled, and closed.
   - Model backend/TCP health separately from logical stream completion.
   - Specify exactly which events release stream state, return a reusable backend, or poison and close the physical connection.
   - Make close and cleanup idempotent on every normal and unhappy path.

2. Introduce bounded, context-aware admission per remote CN

   - Bound active physical connections using a configurable budget that is lower than the effective socket and file-descriptor limits.
   - Queue or backpressure new pipeline streams when the budget is reached.
   - Ensure waiting observes query cancellation, deadline, CN shutdown, and remote health changes.
   - Use fair scheduling so one large query cannot starve unrelated queries.

3. Separate logical stream close from physical connection close

   - A normally acknowledged terminal stream should release only its logical state when reuse is proven safe.
[...]

PR #25713, opened 2026-07-14
fix: reuse CN pipeline connections after clean teardown

### LINKED ISSUE #25690: [Bug]: CN pipeline backend churn exhausts ephemeral ports under sustained distributed queries

This is a current reproducible recurrence of the connection-exhaustion class previously tracked by #9545. That issue was closed in 2023 and there is no matching open issue.

### Branch Name

main

### Commit ID

3728f38726f209b20129fcf4578ff0095f824893

- Deployment: make dev-up, two CNs in separate Docker containers/processes, one TN, one log service, and one proxy
- Client shape: 16 persistent MySQL sessions; the failure is not caused by repeatedly opening frontend connections
- CN ephemeral port range: 32768-60999
- Pipeline client default MaxSenderNumber per remote host: 100000
- OS: Linux x86_64

### Actual Behavior

A sustained distributed shuffle query causes CN-to-CN morpc connections to grow and churn until new connections fail with EADDRNOTAVAIL:

    ERROR 20505 (HY000): cannot connect to backend: dial tcp4 172.23.0.5:18100: connect: cannot assign requested address

Downstream remote pipelines can then fail with:

    ERROR 20101 (HY000): internal error: remote receiver not ready within 30s, remote CN may have failed

Both CN processes remain alive.

Using 16 persistent MySQL sessions, each scheduled to execute the query 25 times, 320 statements completed with the correct result before multiple sessions aborted around statements 19-24 due to CN-to-CN connection failures. This excludes frontend short-connection churn as the cause.

A socket snapshot taken after the failures showed approximately 5.8K established connections per CN plus 4K-5K TIME_WAIT sockets. At peak, new connections to the same remote CN could no longer obtain a local ephemeral port.

The application-level limit does not protect the real resource: PipelineConfig defaults MaxSenderNumber to 100000 backends per remote host, while the observed Linux ephemeral range contains only about 28K ports.

### Expected Behavior

Sustained distributed queries should not exhaust CN ephemeral ports. CN-to-CN pipeline concurrency should be bounded by a realistic socket budget and apply context-aware backpressure before connection creation fails.

The solution must preserve cancellation and timeout behavior and must not introduce a deadlock when pipelines have cyclic dependencies.

### Steps to Reproduce

1. Build and start a two-CN development cluster:

       make dev-build
       make dev-up

2. Create the distributed test data:

       create database d1;
       use d1;
       create table t1(c1 int not null, c2 int not null, c3 int not null) cluster by c1;
       create table t2(c1 int not null, c2 int not null, c3 int not null) cluster by c1;
       create table t3(c1 int not null, c2 int not null) cluster by c1;
       insert into t1 select *,*,* from generate_series(5000000) g;
       insert into t2 select *,*,* from generate_series(4000000) g;
       insert into t3 select *,* from generate_series(1,1000000) g;

3. Open 16 persistent MySQL sessions through the proxy. In every session, execute the following statement 25 times:

       use d1;
       select count(*)
       from t1, t2, t3
       where t1.c1 = t2.c1
         and t1.c2 = t3.c2
         and t2.c2 < 900000
         and t3.c1 < 500000;

   The expected result of every statement is 499999.

4. Monitor CN socket state and logs:

       docker exec mo-cn1 cat /proc/sys/net/ipv4/ip_local_port_range
       docker exec mo-cn1 ss -s
       docker exec mo-cn2 ss -s

5. Observe repeated CN-to-CN dial failures with cannot assign requested address, followed by query failures or remote receiver timeouts.

### Additional information

The current lifecycle is:

1. PipelineClient.NewStream calls morpc NewStream with lock=true, so a pipeline stream exclusively locks one backend/TCP connection.
2. messageSenderOnClient.close calls stream.Close(true), which closes the entire backend connection when the stream ends.
3. Closed backends are removed from the morpc pool by inactive GC, but the kernel keeps the closed TCP tuples in TIME_WAIT.
4. New distributed pipelines continue creating backends up to the configured 100000-per-host application limit, which is higher than the available ephemeral-port budget.

This is not a classic permanent connection leak: destruction is reached. It is a lifecycle and capacity-control bug in which the connection creation/close rate exceeds kernel reclamation and there is no admission control at the true resource boundary.

Historical issue #9545 documents the same failure signature and notes that blindly reusing a single TCP connection is unsafe because cyclic pipeline dependencies can fill TCP buffers and deadlock. Therefore simply changing Close(true) to Close(false) is not a safe fix. The repair needs bounded, context-aware admission/backpressure and a deliberate reuse strategy that preserves pipeline progress.

This issue is independent of #25687/#25688. The empty-dispatch change restores protocol correctness; it does not create the underlying backend limit mismatch or TCP lifecycle behavior.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [x] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Closes #25690

Related to #25691.

Remote pipeline completion currently couples a logical stream teardown to destruction of its morpc backend and CN-to-CN TCP connection. High-fanout distributed queries therefore create enough connection churn and `TIME_WAIT` tuples to exhaust Linux ephemeral ports even though the CN processes remain healthy.

This PR adds a negotiated FIN/FIN_ACK lifecycle for successfully drained pipeline streams:

- the client requests FIN/ACK teardown per stream and falls back to the legacy close-connection path when the server does not accept it;
- the server acknowledges FIN only after pipeline and dispatch cleanup crosses a completion barrier;
- morpc atomically retires sent/received sequence state before unlocking the still-exclusive backend for reuse;
- uncertain FIN delivery, timeout, malformed sequencing, unread data, cancellation, and ambiguous errors poison and close the backend;
- lifecycle, FIN/ACK, reuse, close-reason, and sequence-container metrics make teardown outcomes observable;
- `disable-stream-reuse` provides a rollback switch.

The connection remains exclusively locked while a logical stream is active. This PR does not multiplex active streams on one TCP connection, add query-boundary admission, or change the existing maximum-concurrency policy. Those require separate deadlock/fairness design under #25691.

### Validation

- `go test -count=1 ./pkg/common/morpc ./pkg/cnservice ./pkg/sql/compile`
- targeted FIN/ACK, lifecycle, and negotiated teardown tests with `-race`
- `make -j8 build`
- macOS two-CN reproduction: two consecutive 400-statement waves completed 400/400 with result `499999`; CN pipeline `TIME_WAIT` remained zero
- Linux amd64 with independent log/TN/CN1/CN2 processes: two consecutive 400-statement waves completed 400/400; established socket endpoints plateaued at 2,778 then 2,814; `TIME_WAIT` remained zero during load; idle GC returned established pipeline sockets to zero and CN file descriptors from about 2,895 to 63
- both CNs remained healthy, and the exact join still returned `499999` after idle GC

The design, failure matrix, rollout/rollback plan, and detailed acceptance records are included in `docs/design/cn_pipeline_connection_lifecycle.md`.


XuPeng-SH (approved):
Re-reviewed the latest head `f92db6739b`. The FIN/FIN_ACK lifecycle looks coherent end-to-end: server-side stream retirement is fenced behind `pipelineStreamLifecycle`, `FinishStream` flushes the final response before retiring stream sequence state, and the client only reuses the backend after receiving a validated `PipelineStreamFinishAck`. The negotiated cleanup paths also look bounded on unhappy exits (message cancel, connection close, timeout, malformed FIN/ACK). I reran focused `pkg/sql/compile`, `pkg/cnservice`, and full `pkg/common/morpc` tests on this head and do not see a remaining high-confidence correctness, leak, hang, or OOM blocker.

aptend:
Coverage gap fixed in `ea7271bffa`. The previous artifact showed `remoterun_lifecycle.go` at 29.2% (19/65 changed statements), and the existing “without token” test was actually stopped earlier by teardown-mode validation. The new tests separately cover malformed FIN, missing validated token, missing lifecycle, non-finisher session, invalid response, successful FIN_ACK, FinishStream failure, cleanup timeout, message cancellation, connection close, and idempotent cleanup. Focused coverage now reports 91.3% for `remoterun_lifecycle.go` (63/69 statements), projecting overall patch coverage from 67.3% to about 90.9%. Full `pkg/common/morpc`, `pkg/cnservice`, and `pkg/sql/compile` tests, focused race tests, build, and vet pass. The branch was also merged with current main `ddbff5b400` before the coverage change.

PR #25714, opened 2026-07-14
fix(parser): quote FULLTEXT MATCH...AGAINST pattern on the default deparse path 

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #https://github.com/matrixorigin/matrixone/issues/24823

FullTextMatchExpr.Format wrote the AGAINST pattern bare (ctx.WriteString(node.Pattern)), so any re-serialization through tree.String (CREATE TABLE AS SELECT, view expansion, etc.) produced invalid, un-reparseable SQL (AGAINST (防水 ...)).

The pattern is ALWAYS a string literal (unlike the polymorphic NumVal), so emit it single-quoted and escaped unconditionally. The node now round-trips on every formatter path.

Regression coverage on the DEFAULT path: correct the six validSQL expectations in TestValid (which run through tree.String) from bare to quoted, and add TestFullTextMatchDeparseRoundTrip (default-path deparse -> re-parse -> idempotent), covering CJK and an embedded single quote.

### DIFF
--- a/pkg/sql/parsers/dialect/mysql/mysql_sql_test.go
+++ b/pkg/sql/parsers/dialect/mysql/mysql_sql_test.go
@@ -3530,27 +3530,27 @@ var (
 		},
 		{
 			input:  "select * from t1 where MATCH (body, title) AGAINST ('abc dfc ghc')",
-			output: "select * from t1 where MATCH (body, title) AGAINST (abc dfc ghc)",
+			output: "select * from t1 where MATCH (body, title) AGAINST ('abc dfc ghc')",
 		},
 		{
 			input:  "select * from t1 where MATCH (body, title) AGAINST ('abc- +abc' IN BOOLEAN MODE)",
-			output: "select * from t1 where MATCH (body, title) AGAINST (abc- +abc IN BOOLEAN MODE)",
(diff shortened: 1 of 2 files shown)


gouhongshen (changes requested):
## Codex automated review

The parser unit tests pass, but this formatter change also changes fulltext projection column labels and leaves checked-in distributed-test expectations stale.

### P1 - Update fulltext BVT result snapshots for the changed projection headings (pkg/sql/parsers/tree/expr.go:1937)
`Format` is used to derive unaliased expression headings, so this change turns headings such as `MATCH (body) AGAINST (red)` into `MATCH (body) AGAINST ('red')`. The checked-in exact expectations still contain the old headings in `test/distributed/cases/fulltext/fulltext.result` (for example lines 95 and 131), `fulltext_bm25.result`, and `gojieba.result` (20 occurrences total). Those cases project MATCH expressions without aliases, so their output necessarily changes with this patch; the `.result` files must be regenerated/updated or the distributed BVT comparisons will fail.

Issue #25637, opened 2026-07-12
[Bug]: HNSW multi-index merge still materializes shard_count × LIMIT candidates


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

`de64386657a07eb3bcb2b411cf7989023b1e16c0`

### Other Environment Information

- Hardware parameters: not required; the defect is deterministic from the in-process merge path.
- OS type: any supported platform.
- Others: #25587 (`df6a34119e`) updated the existing merge path but did not complete its intended bound.

### Actual Behavior

HNSW searches over multiple index files still retain every per-file candidate in the shared merge heap. The recent mitigation caps only the heap's *initial allocation*, not its length:

1. `boundedHnswSearchLimits` requests `min(LIMIT, fileCardinality)` candidates from every index file.
2. Every returned candidate is unconditionally appended via `SearchResultSafeHeap.Push`.
3. `SearchResultSafeHeap` is an ordinary unbounded min-heap. Its `size` constructor argument is capacity only.
4. Only after all workers finish does the code pop and return the global `LIMIT` rows.

Therefore the live candidate set is `sum(min(LIMIT, cardinality_i))`, which is `index_file_count × LIMIT` for sufficiently populated files, while the user-visible result is at most `LIMIT`. This is an avoidable memory and CPU amplification on fragmented HNSW indexes. The comment above `boundedHnswSearchLimits` says it prevents this multiplication, but the current `Push` path does not enforce that bound.

Relevant code on main:

- `pkg/vectorindex/hnsw/search.go`: `boundedHnswSearchLimits` computes per-file K and a preallocation capacity.
- `pkg/vectorindex/hnsw/search.go`: workers unconditionally push every candidate into `SearchResultSafeHeap`.
- `pkg/vectorindex/index.go`: `SearchResultSafeHeap.Push` has no maximum-size or eviction logic.

This is related to the earlier HNSW benchmark OOM reports (#24041, #23904), but is a precise current root cause in the merge path and a follow-up to #25587 rather than a duplicate of their environment-level symptoms.

### Expected Behavior

Each index file may search its local top K, because that is necessary for exact global top-K merging. The merge structure itself must retain only the best global K candidates. Peak merge memory should be `O(min(LIMIT, total_cardinality))`, independent of index-file count.

### Steps to Reproduce

1. Build or mock an `HnswSearch` with at least two non-empty index files.
[...]

Issue #24041, opened 2026-04-02
[Bug]: [main tke regression] Hnsw Index Gist960 Vector 1M Benchmark Test(L2_distance) oom.


Ariznawlll:
### Branch Name

main

### Commit ID

79cca0649490247ff816e7c8c0bbc57ea1469697

### Other Environment Information

- Nightly regression test
- Test suite: main TKE测试

### Actual Behavior

Nightly test step failed:
- Failed step: Main Nightly Regression Test / HNSW AND IVF VECTOR INDEX Benchmark Test / Hnsw Index Gist960 Vector 1M Benchmark Test(L2_distance)
- Time: 2026-04-02T02:45:26.000Z
- Run link: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/23854959086
log：https://grafana.ci.matrixorigin.cn/explore?panes=%7B%22Q4q%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-main-commit-79cca0649-20260401%5C%22%7D%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221775097926000%22,%22to%22:%221775099126000%22%7D%7D%7D&schemaVersion=1&orgId=1

metric：https://grafana.ci.matrixorigin.cn/explore?panes=%7B%221vf%22:%7B%22datasource%22:%22pyroscope%22,%22queries%22:%5B%7B%22groupBy%22:%5B%5D,%22labelSelector%22:%22%7Bnamespace%3D%27mo-main-commit-79cca0649-20260401%27%7D%22,%22queryType%22:%22both%22,%22refId%22:%22A%22,%22profileTypeId%22:%22memory:inuse_space:bytes:space:bytes%22,%22datasource%22:%7B%22type%22:%22grafana-pyroscope-datasource%22,%22uid%22:%22pyroscope%22%7D%7D%5D,%22range%22:%7B%22from%22:%221775097952000%22,%22to%22:%221775099152000%22%7D%7D%7D&schemaVersion=1&orgId=1

### Expected Behavior

The test should pass.

### Steps to Reproduce

Triggered by nightly regression workflow.
[...]

Issue #23904, opened 2026-03-19
[Bug]: [date 3.18]main tke regression: hnsw index query mo oom


heni02:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

d1e615c143be24457a78b73a616f925cf18c9ab3

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

job: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/23250842454/job/67686641599

<img width="1597" height="576" alt="Image" src="https://github.com/user-attachments/assets/6e588424-dfcc-47c8-a575-650f10efa0db" />

<img width="1277" height="371" alt="Image" src="https://github.com/user-attachments/assets/cd8285b6-1635-4a5b-b5c2-840956f6b060" />
heap:
https://grafana.ci.matrixorigin.cn/explore?panes=%7B%22AVP%22:%7B%22datasource%22:%22pyroscope%22,%22queries%22:%5B%7B%22groupBy%22:%5B%5D,%22labelSelector%22:%22%7Bnamespace%3D%5C%22mo-main-commit-d1e615c14-20260318%5C%22,pod%3D%5C%22nightly-regression-dis-tp-cn-75tzk%5C%22%7D%22,%22queryType%22:%22both%22,%22refId%22:%22A%22,%22profileTypeId%22:%22memory:alloc_objects:count:space:bytes%22,%22datasource%22:%7B%22type%22:%22grafana-pyroscope-datasource%22,%22uid%22:%22pyroscope%22%7D%7D%5D,%22range%22:%7B%22from%22:%221773892518601%22,%22to%22:%221773893359852%22%7D%7D%7D&schemaVersion=1&orgId=1

<img width="1857" height="816" alt="Image" src="https://github.com/user-attachments/assets/d63f8ac3-3b5b-4876-b68b-e7292e790fee" />

commit 范围：

<img width="1537" height="806" alt="Image" src="https://github.com/user-attachments/assets/c75fe8ec-f34f-4308-b96e-4c6240e24eeb" />

### Steps to Reproduce

```Markdown
tke regression hnsw 100万benchmark
```

heni02:
d1e615c1 单跑不能复现，无法二分 @aunjgr 
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/23333432640/job/67870496406

heni02:
[...]

PR #25716, opened 2026-07-14
fix(vectorindex): bound HNSW multi-index merge heap to LIMIT (#25637)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25637

The HNSW multi-index search pushed every per-file candidate (sum of min(LIMIT, cardinality) over files ~= shard_count * LIMIT) into an unbounded min-heap and truncated to LIMIT only at the end, so peak merge memory scaled with the number of index files. On fragmented indexes this is an OOM vector (follow-up to #25587; relates to #24041, #23904).

Make SearchResultSafeHeap a concurrency-safe BOUNDED top-K max-heap: it retains only the `limit` smallest-distance results, evicting the current worst (heap root) in O(log k) once full; rejects are O(1). Peak merge memory is now O(min(LIMIT, total_cardinality)), independent of file count. A max-heap (worst at root) is required so the worst retained item can be evicted cheaply; the search drains it back-to-front to restore ascending (nearest-first) order. boundedHnswSearchLimits no longer needs to return the sum-based heap capacity; the heap bound is simply the result limit. The retained set is the exact global top-K regardless of push order (order-independent invariant), so parallel workers can keep pushing concurrently.

Tests: TestSafeHeapBounded (4 shards push 40 candidates concurrently into a bound-10 heap -> retains exactly the 10 smallest); TestBoundedHnswSearchLimits updated for the 2-value signature.


XuPeng-SH (approved):
Reviewed the bounded HNSW merge implementation across correctness, performance, concurrency, and unhappy paths. The max-heap invariant retains the exact global smallest K independently of push order; reverse draining preserves nearest-first output; zero/large limits and preallocation are bounded; and the existing worker cancellation/resource lifecycle remains intact. CI is green, and local race/property tests, vectorindex build, and vet all passed. No merge-blocking issues found.

Issue #25630, opened 2026-07-11
[Bug]: HnswModel.SaveToFile leaks one file descriptor per save


Ariznawlll:
## Summary

Each successful `HnswModel.SaveToFile` call leaks the file descriptor returned by `os.CreateTemp`. The model records only `f.Name()` and never closes `f`; later `Destroy` removes the pathname but cannot close the lost descriptor.

Repeated HNSW CDC saves, builds, unloads, or reindex operations can therefore accumulate descriptors until the process hits its open-file limit.

## Version

Commit: `7f7de555c`

## Code path

`pkg/vectorindex/hnsw/model.go`:

```go
f, err := os.CreateTemp("", "hnsw")
if err != nil {
    return err
}

err = idx.Index.Save(f.Name())
// checksum and model destruction
idx.Path = f.Name()
return nil
```

There is no `f.Close()` or deferred close on the success path. Several error paths after creation also return without closing the descriptor.

## White-box reproduction

Disable GC/finalizers, count the process's open descriptors with `fcntl(F_GETFD)`, and run 20 model save/destroy cycles:

```go
oldGC := debug.SetGCPercent(-1)
defer debug.SetGCPercent(oldGC)

before := openFDCount()
for i := 0; i < 20; i++ {
    model, _ := NewHnswModelForBuild[float32]("fd-test", cfg, 1, 64)
    require.NoError(t, model.Add(int64(i), []float32{1, 2, 3}))
    require.NoError(t, model.SaveToFile())
    require.NoError(t, model.Destroy())
}
after := openFDCount()
```

## Actual result

```text
open descriptors before=4 after=24 delta=20
```

The descriptor increase exactly matches the number of successful saves. `Destroy()` does not reduce it.

## Expected result

Every descriptor created by `SaveToFile` must be closed on all success and failure paths. Saving by pathname through usearch does not require MatrixOne to keep the original `*os.File` open.

## Suggested fix and verification

- Close `f` immediately after creation if only its pathname is needed, or use a deferred close with explicit error handling.
- Remove the temporary pathname on checksum/save failures.
- Add an FD-count regression test with GC disabled and tests for save/checksum failure cleanup.

Ariznawlll:
## 最新 main 详细验证

- **源码基线**：`origin/main@f1c0a92d882dbdad469ffd2c96f607796207c797`。
- **执行命令**：

```bash
go test ./pkg/vectorindex/hnsw \
[...]

PR #25717, opened 2026-07-14
fix(vectorindex): close temp file fd in HnswModel.SaveToFile (#25630)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25630

SaveToFile created a temp file via os.CreateTemp but never closed the returned *os.File on any path, so every save leaked one file descriptor until the process hit its fd limit. Only f.Name() is ever used — usearch's Index.Save and CheckSum reopen the path themselves — so close the handle immediately after capturing the name. Also remove the temp file on every error path (close/Save/CheckSum/Destroy failures); previously the CheckSum and Destroy failures orphaned the file on disk (idx.Path is not set until success). The other two os.CreateTemp sites in model.go already close via defer, so the leak was isolated to SaveToFile.

Test: TestSaveToFileNoFDLeak runs 20 save cycles with GC disabled and counts open fds via /dev/fd (portable across Linux/macOS); pre-fix the delta is exactly 20, post-fix ~0.


XuPeng-SH (approved):
Reviewed the complete diff and the SaveToFile resource lifecycle. The temp-file descriptor is closed before pathname-based Save/CheckSum, failure paths retain a single local cleanup owner, and ownership transfers to idx.Path only after Save, checksum, and index destruction succeed. Targeted tests, targeted race tests, the full HNSW package, go vet, and go build pass on the current head. No correctness, leak, hang, unbounded-growth, or compatibility blocker found.

Issue #25629, opened 2026-07-11
[Bug]: HNSW parallel CDC insert drops Add errors and corrupts model length


Ariznawlll:
## Summary

The pure-INSERT HNSW CDC path discards every error returned by `HnswModel.AddWithoutIncr`. The CDC update returns success even when usearch rejects a row, while the pre-reserved logical `Len` counter remains incremented.

This can persist an HNSW index with missing rows and an incorrect model length, and allows the CDC task/watermark to advance past data that was not indexed.

## Version

Commit: `7f7de555c`

## Code path

`pkg/vectorindex/hnsw/sync.go` in `insertAllInParallel`:

```go
last, _, err := s.getLastModelAndIncrForSync(sqlproc, maxcap, &mu)
if err != nil {
    err_chan <- err
    return
}

// Len counter already incremented. Just add to last model
last.AddWithoutIncr(row.PKey, row.Vec)
```

The return value of `AddWithoutIncr` is ignored. `getLastModelAndIncrForSync` has already incremented `last.Len` before the physical add.

The sequential UPDATE/DELETE path checks `Add` and `Remove` errors correctly; the defect is specific to the parallel pure-INSERT path.

## White-box reproduction

Create an empty 3-dimensional HNSW model and pass two pure CDC INSERT entries with the same primary key to `insertAllInParallel`:

```go
cdc := &vectorindex.VectorIndexCdc[float32]{Data: []vectorindex.VectorIndexCdcEntry[float32]{
    {Type: vectorindex.CDC_INSERT, PKey: 1, Vec: []float32{1, 2, 3}},
    {Type: vectorindex.CDC_INSERT, PKey: 1, Vec: []float32{3, 2, 1}},
}}

err := syncer.insertAllInParallel(nil, 64, []int{-1, -1}, cdc)
physicalLen, _ := model.Index.Len()
```

## Actual result

```text
returned error=<nil> logical Len=2 physical Len=1
```

usearch rejects the duplicate physical key, but the HNSW CDC layer reports success and retains the extra logical reservation. A nil/invalid vector produces the same false-success behavior with `logical Len=1`, `physical Len=0`.

## Expected result

- Propagate every `AddWithoutIncr` error through `err_chan` and fail the CDC iteration.
- Roll back the pre-reserved `Len` value when the physical add fails.
- Do not save the model or advance CDC state after any row fails.
- Make duplicate/replayed CDC entries idempotent or fail explicitly without corrupting model counters.

[...]

PR #25719, opened 2026-07-14
fix(vectorindex): propagate Add error and roll back Len in parallel CDC insert(#25629) 

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25629

insertAllInParallel discarded the return value of HnswModel.AddWithoutIncr, but getLastModelAndIncrForSync had already pre-incremented last.Len before the physical add. When usearch rejects the add (e.g. a duplicate primary key), the reserved Len slot was never released, so the logical Len diverged from the physical index size and the sync reported success on a partially applied batch.

On add failure, roll back the reserved slot (last.Len.Add(-1)) and send the error to err_chan. The error then propagates out of insertAllInParallel and Update returns it, so the batch is not saved and CDC state is not advanced. Duplicate/replayed entries now fail explicitly instead of silently corrupting the length counter.

Test: TestSyncInsertDuplicateKeyFailsCleanly feeds two INSERT entries with the same PK into an empty model and asserts Update fails and logical Len equals the physical Index.Len(). Pre-fix Update returned nil.


XuPeng-SH (approved):
Approved on `4c88195e`.

I traced the failure path from `HnswSync.Update` through the parallel workers, native `AddUnsafe`, and the ISCP consumer lifecycle. The change correctly propagates `AddWithoutIncr` failures and rolls back the exact model slot reserved by `getLastModelAndIncrForSync`.

I also verified the unhappy paths:
- each worker can publish at most one error, matching the bounded error-channel capacity, so concurrent failures cannot deadlock;
- an Update error skips Save and watermark advancement;
- the ISCP HNSW runner destroys the partially modified in-memory/native model before the iteration exits, so partial success cannot leak into a retry;
- the fast path adds only an error check, with no material performance regression.

Validation:
- the PR regression test passed 10 times under `-race`;
- additional 8-worker duplicate-key and invalid-vector stress cases kept logical `Len` equal to physical `Index.Len()`;
- the full `pkg/vectorindex/hnsw` package passed under `-race`.

Issue #25621, opened 2026-07-11
[Bug]: CBloomFilter returns false negatives or panics for constant vectors


Ariznawlll:
### Is there an existing issue for this?

- [x] I have searched the existing issues and pull requests.

### Branch Name

main

### Commit ID

7821d21c723372e45a8f65da3bbb896a157b1036

### Other Environment Information

- OS: macOS arm64
- Go: repository toolchain
- The same implementation is present on current `origin/main`.

### Actual Behavior

`CBloomFilter.TestVector` mishandles constant vectors because their logical row count is greater than their physical element count.

Confirmed failures:

| Input | Logical rows | Physical elements | Actual | Expected |
|---|---:|---:|---|---|
| Constant INT64 `42` | 3 | 1 | `[1,0,0]` | `[1,1,1]` |
| Constant VARCHAR `key` | 3 | 1 | `[1,0,0]` | `[1,1,1]` |
| `CONST NULL` INT64 | 3 | 0 | index-out-of-range panic | `[0,0,0]` |
| `CONST NULL` VARCHAR | 3 | 0 | index-out-of-range panic | `[0,0,0]` |

The Bloom filter contains `42` and `key`, but rows 1 and 2 of the corresponding constant vectors are reported as definitely absent. Any caller that uses this output for pruning can drop valid rows.

This is distinct from the zero-length fixed-vector panic in #25618: these vectors have a non-zero logical length, affect both fixed and varlen types, and require constant-vector expansion rather than only an empty-input guard.

### Root Cause

The implementation passes the vector's logical length to C while passing a physical buffer containing only one value (or no value for `CONST NULL`):

```go
fixedData := v.GetData() // one physical value for CONST
length := v.Length()     // e.g. three logical rows
results := make([]uint8, length)
C.bloomfilter_test_fixed(
    bf.ptr,
    unsafe.Pointer(&fixedData[0]),
    C.size_t(len(fixedData)),
    C.size_t(typeSize),
    C.size_t(length),
    ...,
    unsafe.Pointer(&results[0]),
)
```

The varlen path has the same logical/physical cardinality mismatch. `AddVector` and `TestAndAddVector` share the same assumption and need equivalent constant handling.

### Expected Behavior

- A non-NULL constant vector tests/adds its one physical value once and expands the result/callback semantics to all `v.Length()` logical rows.
- A `CONST NULL` vector returns a length-matched all-zero result, adds nothing, and reports NULL for every callback row.
[...]

PR #25720, opened 2026-07-14
fix(filter): handle constant vectors in CBloomFilter/CRoaring/Cbitmap

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25621 #25618

The three cgo membership filters confused a vector's logical row count with its physical element count for CONSTANT vectors. A constant stores a single physical value (none for const-null) but reports Length() = row count, so passing the logical length to C over the 1-element (or empty) physical buffer read past the buffer — false negatives ([1,0,0] instead of [1,1,1]) — or panicked on the empty const-null buffer (&data[0]). This affects PK/partition filtering and IVF_FLAT bloom ops, where a single-value predicate folds to a constant vector.

Process only the PHYSICAL element(s) and broadcast the result to all logical rows:
  - bloomfilter/cbloomfilter.go: vecPhysCount (1 for non-null const, 0 for const-null, Length() otherwise), emitConstNull (all-zero result, every row NULL), broadcastResults (expand the single result across rows). All six fixed/varlena test/add/testAndAdd helpers short-circuit const-null, guard nitem==0 (also fixes a latent panic on regular length-0 fixed vectors), call C with the physical count, then broadcast.
  - docfilter/croaring.go: vecFixedArgs computes the physical nitem; new finalizeVecResults broadcasts for constants. cbitmap.go shares both (its TestVector rewired). Build-from-constant now contains the value.

Non-const vectors take a byte-identical path. Tests: TestCBloomFilterConstVector (const int64/varchar, non-null + null, across Test/Add/TestAndAdd), TestCRoaringConstVector + TestCbitmapConstVector (present/absent/null probes, build-from-const). All fail on the original code (verified via git-revert) and pass under -race.


XuPeng-SH (changes requested):
The stateless TestVector/AddVector fixes look correct, but TestAndAddVector does not preserve the logical per-row semantics for a constant vector.

Reproduced on head 1a3664c1058965d00c573037e3c7184c7bd5d528 with both INT64 and VARCHAR. On a fresh filter, a regular vector containing [99, 99, 99] reports callback states [false, true, true]: the first logical row inserts the value, so subsequent duplicate rows observe it. The equivalent constant vector (value 99, logical length 3) reports [false, false, false].

The C test-and-add implementation is intentionally stateful and processes elements sequentially. This PR calls it once for the single physical constant value, then uses broadcastResults to copy the pre-insert result to every logical row. Broadcasting is valid for TestVector, but not for TestAndAddVector because the filter state changes after the first logical row. This can let repeated logical values all appear new to a deduplication callback and also makes constant and flat representations of the same rows observably different.

Please give TestAndAddVector state-aware constant handling and add fixed-width plus varlena equivalence tests against a flat repeated-value vector. The current test only asserts [false, false, false], which bakes in the incorrect behavior.

gouhongshen (changes requested):
## Codex automated review

Constant-vector handling fixes TestVector, but makes TestAndAddVector report incorrect per-row membership for repeated logical rows.

### P1 - Preserve per-row TestAndAdd semantics for constant vectors (pkg/common/bloomfilter/cbloomfilter.go:351)
`TestAndAddVector` promises to test and add every logical element. For a previously unseen constant vector of length 3, the equivalent regular vector `[99,99,99]` reports false for row 0 and true for rows 1 and 2, since each call mutates the bloom filter. This path sends only one physical element to `bloomfilter_test_and_add_fixed` and then `broadcastResults` copies that pre-add result to all logical rows, reporting `[false,false,false]`. The same problem exists in the varlena path at line 379, and the added test explicitly locks in the incorrect all-false result. Handle constant TestAndAdd rows sequentially (or otherwise preserve its stateful per-row results) rather than broadcasting a single result.

XuPeng-SH (changes requested):
One blocking cross-product case remains in the new physical-count abstraction.

### P2 - Keep zero-logical-row constants as a no-op (`pkg/common/bloomfilter/cbloomfilter.go:276`)

`vecPhysCount` returns one physical item for every non-null constant without first checking `v.Length()`. A non-null constant can legally retain its physical value while its logical length becomes zero (`NewConstFixed(..., 3)` followed by the public `SetLength(0)`, as occurs with reused/shrunk vectors). Before this PR every vector helper began with `if v.Length() == 0 { return }`; now `TestAndAddVector` and `AddVector` pass the retained value to C and mutate the bloom filter even though the input has no logical rows. The same helper used by the doc filters builds a non-empty filter from that logically empty vector.

I reproduced the observable state corruption on current head `fed419cf56`: on a fresh filter, shrink constant 99 to length 0, call `TestAndAddVector`, then test-and-add a real one-row 99. The real row is reported `present=true`; without the empty-vector call it is correctly `false`. The focused CGo regression exits 1 at that assertion.

Please make the shared physical-count rule return zero whenever logical length is zero, before classifying constant/flat, and add the missing matrix cells for zero-length constants (fixed and varlena; Test/Add/TestAndAdd, plus CRoaring/Cbitmap build/test as applicable). This preserves the original no-op contract and fixes the whole family rather than special-casing one method.

The positive-length stateful constant fix is otherwise correct. Latest-main synthetic merge is conflict-free; build, vet, complete bloomfilter/docfilter tests, and focused race tests pass. I found no other correctness, CGo safety, lifecycle, leak, hang, OOM, or material performance blocker.

gouhongshen (changes requested):
## Codex automated review

The prior TestAndAdd finding is resolved, but a zero-length non-null constant regression remains. Focused CGo tests could not run because the environment has no Go executable.

### P2 - Keep zero-logical-row constant vectors as no-ops (pkg/common/bloomfilter/cbloomfilter.go:277)
`Vector.SetLength(0)` is public and leaves a non-null constant's physical value intact. Before this change every bloom helper returned on `v.Length() == 0`; now `vecPhysCount` returns 1 for that vector, so `AddVector` and `TestAndAddVector` call C and insert a value despite there being no logical row. The same rule in `pkg/common/docfilter/croaring.go:59` makes `BuildCRoaringBytes` and `BuildCbitmapBytes` construct a non-empty filter from the empty vector. A later real row with that value is then reported present. Return zero before constant classification when `Length() == 0`, and cover zero-length non-null constants across the bloom and docfilter paths.

XuPeng-SH (approved):
Latest head resolves the earlier blockers.

The constant-vector TestAndAdd path is now state-aware, the shared physical-count rule keeps zero-logical-row constants as no-ops, and the bloom/docfilter paths consistently handle const-null / empty inputs without out-of-bounds reads or empty-vector state corruption. I do not see additional correctness, CGo safety, lifecycle, or unhappy-path blockers on this head.

Issue #25695, opened 2026-07-14
[Bug]: TN RPC handlers ignore request deadlines while waiting for replica startup


Ariznawlll:
## Environment

- Branch: `main`
- Commit: `cb94d94954a098ee8be2a33d5c0e37ebbc8ca489`
- OS: macOS arm64
- Component: `pkg/tnservice`

## Description

TN RPC handlers block in `replica.waitStarted()` without observing the request context. If a request reaches a replica whose storage/TxnService is still starting, cancellation or deadline expiry does not release the handler. It remains blocked until `replica.startedC` is closed.

This affects read, write, debug, commit, rollback, prepare, commit/rollback-TN-shard, get-status, and local-dispatch paths.

## Code path

```go
func (s *store) doRead(ctx context.Context, ...) error {
    r := s.validTNShard(ctx, request, response)
    if r == nil {
        return nil
    }
    r.waitStarted() // <- only waits on startedC
    ...
}

func (r *replica) waitStarted() {
    <-r.startedC
}
```

## Deterministic white-box reproduction

1. Install a replica object without starting its TxnService.
2. Call `handleRead` with a 20 ms context deadline.
3. Verify that the handler is still blocked after 100 ms.
4. Start the replica and observe that only then does the handler return.

```go
ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
defer cancel()

done := make(chan error, 1)
go func() {
    done <- s.handleRead(ctx, &req, &txn.TxnResponse{})
}()

select {
case err := <-done:
    t.Fatalf("returned before replica start: %v", err)
case <-time.After(100 * time.Millisecond):
    // Reproduced: deadline did not release the handler.
}

require.NoError(t, r.start(testTxnService))
<-done
```

Command:

```bash
go test ./pkg/tnservice \
  -run '^TestCodexRPCWaitStartedIgnoresDeadline$' -count=3 -v
```

Result: reproduced 3/3 on the latest `main`. Every run remains blocked for at least 100 ms despite a 20 ms request deadline.

## Expected behavior

Waiting for replica readiness must select on both `startedC` and `ctx.Done()`. Once the request is canceled, the handler should return promptly and must not dispatch work to a service that starts later.

## Actual behavior

The handler outlives its request deadline and only resumes when the replica eventually starts. If storage creation retries indefinitely, request goroutines can also wait indefinitely.

## Impact

- Transaction RPC latency can exceed caller deadlines by an unbounded duration.
[...]

PR #25721, opened 2026-07-14
fix: harden TN replica lifecycle

### LINKED ISSUE #25668: [Bug]: TN RPC retry ignores context cancellation during RetryInterval and dispatches again
## Environment

- Branch: `main`
- Component: `pkg/tnservice/store_rpc_handler.go`

## Description

TN RPC retry waiting does not observe the request context. If the context is canceled or reaches its deadline during `RetryInterval`, `maybeRetry` still sleeps for the complete interval and returns `true`. `handleWithRetry` then invokes the delegate one more time with an already-expired context.

This makes handler latency exceed the caller's deadline by up to the configured retry interval and performs work after cancellation.

## Reproduction

A deterministic package-level white-box test uses:

- request context deadline: 20 ms;
- retry interval: 250 ms;
- retry code: `moerr.ErrTNShardNotFound`;
- delegate: counts invocations and always returns the retryable error in `TxnResponse`.

Core test logic:

```go
ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
defer cancel()

req := &txn.TxnRequest{Options: &txn.TxnRequestOptions{
    RetryCodes:    []int32{int32(moerr.ErrTNShardNotFound)},
    RetryInterval: int64(250 * time.Millisecond),
}}

var calls atomic.Int32
start := time.Now()
err := (&store{}).handleWithRetry(ctx, req, &txn.TxnResponse{},
    func(ctx context.Context, _ *txn.TxnRequest, resp *txn.TxnResponse) error {
        calls.Add(1)
        resp.TxnError = txn.WrapError(
            moerr.NewTNShardNotFound(ctx, "test", 1), 0)
        return nil
    })
elapsed := time.Since(start)
```

Command:

```bash
go test -race ./pkg/tnservice -run '^TestCodexRetrySleepIgnoresCancellation$' -count=3 -v
```

## Actual result

All three runs reproduce the same behavior:

```text
=== RUN   TestCodexRetrySleepIgnoresCancellation
--- PASS: TestCodexRetrySleepIgnoresCancellation (0.25s)
=== RUN   TestCodexRetrySleepIgnoresCancellation
--- PASS: TestCodexRetrySleepIgnoresCancellation (0.25s)
=== RUN   TestCodexRetrySleepIgnoresCancellation
--- PASS: TestCodexRetrySleepIgnoresCancellation (0.25s)
```

The test passes because it asserts the incorrect current behavior:

- elapsed time is at least 200 ms even though the context deadline is 20 ms;
- the delegate is invoked twice;
- the second invocation receives an expired context.

## Expected result

Retry waiting should return promptly when `ctx.Done()` is closed, and `handleWithRetry` must not call the delegate again after cancellation.

## Root cause

`maybeRetry` checks the context only before sleeping:

```go
select {
case <-ctx.Done():
    return false
default:
    // ... match retry code ...
    time.Sleep(wait)
    return true
}
```

Cancellation during `time.Sleep(wait)` cannot wake the wait. After sleep, the function returns `true` without checking the context again.

## Impact

- TN handlers exceed request deadlines by up to `RetryInterval`.
- Canceled requests can be dispatched again and consume CPU/storage/RPC resources.
- A large configured interval can retain handler goroutines and request state for a long time after client disconnect or upstream timeout.
- Under retry storms, delayed cancellation can amplify TN load.

## Suggested fix direction

Use a timer controlled by the context:

```go
timer := time.NewTimer(wait)
defer timer.Stop()
select {
case <-ctx.Done():
    return false
case <-timer.C:
    return true
}
```

Also consider returning `ctx.Err()` from the handler instead of returning a retryable response after the request has already expired.

### LINKED ISSUE #25678: [Bug]: Small positive logtail rpc-max-message-size passes validation and panics TN
## Environment

- Branch: `main`
- OS: macOS arm64

## Description

The TN and TAE logtail configuration validators accept any positive `rpc-max-message-size`, but `NewLogtailServer` requires the value to be larger than the encoded `LogtailResponseSegment` header. A small positive value therefore passes configuration validation and later panics while a TN replica creates its TAE/logtail storage.

For example, `rpc-max-message-size = 1` passes both validation paths and deterministically reaches:

```go
if s.maxChunkSize <= 0 {
	panic("rpc max message size isn't enough")
}
```

## White-box reproduction

```go
func TestValidatedSmallRPCMessageSizePanics(t *testing.T) {
	cfg := &options.LogtailServerCfg{RpcMaxMessageSize: 1}
	cfg.Validate()
	require.Equal(t, int64(1), cfg.RpcMaxMessageSize)

	require.PanicsWithValue(t, "rpc max message size isn't enough", func() {
		_, _ = NewLogtailServer("127.0.0.1:0", cfg, nil, mockRuntime(), nil)
	})
}
```

Run:

```text
go test ./pkg/vm/engine/tae/logtail/service \
  -run '^TestCodexValidatedSmallRPCMessageSizePanics$' -count=3 -v
```

Result: reproduced 3/3 on the latest `main`.

## Production path

```text
TN Config.Validate
  -> accepts every RpcMaxMessageSize > 0
TN create replica
  -> taestorage.NewTAEStorage
    -> service.NewLogtailServer
      -> NewLogtailServerSegmentPool(configured size)
      -> LeastEffectiveCapacity() <= 0
      -> panic("rpc max message size isn't enough")
```

The relevant validators only replace values `<= 0`; they do not enforce the minimum protocol-header size.

## Expected behavior

Invalid positive values should be rejected by configuration validation with a descriptive error, or `NewLogtailServer` should return an error. User configuration must not cause a TN process panic during replica creation.

## Actual behavior

The configuration is accepted and the TN panics later when the logtail server is constructed.

## Suggested direction

- Define and validate a minimum `RpcMaxMessageSize` that can hold the maximum segment header plus at least one payload byte.
- Replace the constructor panic with a returned error as defense in depth.

### LINKED ISSUE #25694: [Bug]: TN store.Close leaks QueryService and QueryClient resources
## Environment

- Branch: `main`
- Commit: `cb94d94954a098ee8be2a33d5c0e37ebbc8ca489`
- OS: macOS arm64
- Component: `pkg/tnservice`

## Description

`tnservice.store.Close()` does not close the QueryService or QueryClient created and owned by the TN store.

`NewService` always creates a QueryClient, and non-standalone TN creates and starts a QueryService. `store.Close` closes the stopper, cluster service, shard server, HAKeeper client, transaction sender/server, lock allocator, replicas, task holder, and I/O pipeline, but never invokes either query RPC resource's `Close` method.

The test replaces the store-owned query resources with tracking implementations immediately before `store.Close()`:

```go
func TestCloseSkipsQueryRPCResources(t *testing.T) {
    serviceTracker := &trackingQueryService{}
    clientTracker := &trackingQueryClient{}

    runTNStoreTest(t, func(s *store) {
        require.NotNil(t, s.queryService)
        require.NotNil(t, s.queryClient)
        s.queryService = serviceTracker
        s.queryClient = clientTracker
    })

    require.Equal(t, int32(0), serviceTracker.closes.Load())
    require.Equal(t, int32(0), clientTracker.closes.Load())
}
```

Command:

```bash
go test ./pkg/tnservice \
  -run '^TestCodexCloseSkipsQueryRPCResources$' -count=3 -v
```

Result: reproduced 3/3 on the latest `main`.

## Actual behavior

After the complete TN store shutdown path returns:

- `QueryService.Close()` was called 0 times.
- `QueryClient.Close()` was called 0 times.

## Expected behavior

`store.Close()` should close every resource created and owned by `NewService`, including QueryService and QueryClient, and join any errors with the existing close result.

## Impact

- Query RPC listeners, client sessions, goroutines, and buffers can outlive the TN store.
- Same-process TN recreation can fail to bind the old query address.
- Repeated service lifecycle operations can accumulate leaked RPC resources.

## Suggested fix direction

Close `s.queryService` when non-nil and close `s.queryClient`, preferably before shared runtime resources are torn down. Add lifecycle tests that assert both close calls and that the query listen address can be reused after shutdown.

## Duplicate search

Searched open/closed issues and PRs for `queryService Close TN`, `queryClient Close TN`, `TN store resource leak query service`, and broader query-service lifecycle terms. No matching issue or PR was found.

### LINKED ISSUE #25696: [Bug]: TN heartbeat advertises replicas before storage and TxnService start
## Environment

- Branch: `main`
- Commit: `cb94d94954a098ee8be2a33d5c0e37ebbc8ca489`
- OS: macOS arm64
- Component: `pkg/tnservice`

## Description

TN heartbeat reports every replica present in `store.replicas` as a launched TN shard, even when its storage creation is still retrying and its TxnService has never started.

`createReplica` puts the replica into `store.replicas` before the asynchronous `createTxnStorage` loop succeeds. `getTNShardInfo` checks only map membership and does not check `replica.startedC` or `replica.mu.starting`.

The protobuf contract describes `TNShardInfo` as information for a **launched TN shard**, but the heartbeat cannot distinguish a launched shard from a placeholder that may be retrying initialization indefinitely.

```go
func TestHeartbeatAdvertisesUnstartedReplica(t *testing.T) {
    r := newReplica(newTestTNShard(7, 8, 9), runtime.DefaultRuntime())
    s := &store{replicas: &sync.Map{}}
    s.replicas.Store(uint64(7), r)

    select {
    case <-r.startedC:
        t.Fatal("replica unexpectedly started")
    default:
    }

    shards := s.getTNShardInfo()
    require.Len(t, shards, 1)
    require.Equal(t, uint64(7), shards[0].ShardID)
    require.Equal(t, uint64(8), shards[0].ReplicaID)
}
```

Command:

```bash
go test ./pkg/tnservice \
  -run '^TestCodexHeartbeatAdvertisesUnstartedReplica$' -count=3 -v
```

Result: reproduced 3/3 on the latest `main`.

## Actual behavior

An unstarted replica with an open `startedC` and no TxnService appears in `TNStoreHeartbeat.Shards` exactly like a launched and serving replica.

## Expected behavior

Heartbeat should either:

- report only replicas whose TxnService has successfully started; or
- include an explicit readiness/state field so HAKeeper can distinguish `creating`, `ready`, and `failed` replicas.

## Impact

- The control plane can record a TN shard as present while every transaction request to it blocks in `waitStarted`.
- A permanent storage-creation failure is hidden behind a normal-looking shard heartbeat.
- Placement/recovery decisions have no readiness signal with which to distinguish an operational replica from one that has never started.

## Duplicate search

Searched open/closed issues and PRs for `TN heartbeat unstarted replica`, `DN heartbeat shard starting`, `shard heartbeat readiness TN`, `heartbeat replica`, and `shard readiness`. No matching issue or PR was found.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25668
Fixes #25678
Fixes #25694
Fixes #25696

`#25695` is no longer part of this PR: its replica-start wait cancellation path was implemented by #25724.

This PR was rebuilt on the current `main` after #25724 replaced the TN replica creation/cancellation model. It keeps and completes the lifecycle gaps that remain after that merge:

- TN RPC retry waits observe request cancellation and re-check cancellation when the timer and context become ready together.
- Each RPC/local dispatch now acquires a replica service lease. Replica removal first rejects new calls and cancels active call contexts, then drains active calls before closing or destroying storage. Service calls do not hold the lifecycle lock, so local 2PC nested dispatch remains reentrant.
- Caller cancellation is checked before and after startup readiness and again while acquiring the lease, so an already-canceled request is never dispatched through the ready-channel double-ready window.
- TN shutdown cancels replica lifecycles before draining the RPC server, then closes replica storage, the shared sender, and the remaining clients. This prevents both request-vs-destroy races and server-drain/lifecycle-cancel deadlocks.
- Clock-SI reads waiting for a future snapshot and async `Committing` retries now stop on their request/service context, so replica close cannot wait forever on those loops.
- Heartbeats report a replica only after service startup succeeds and stop reporting it once removal begins.
- Logtail RPC message-size validation rejects values too small for the protobuf header and values above the 100 MiB RPC safety bound. Final option values are validated before TAE resources or segment buffers are created, without allocating the configured payload size.

The ownership review also found related concrete gaps on the same paths:

- `TxnService` borrows the store-wide sender; only the store now closes it, after all replicas are closed.
- The store closes both QueryService and QueryClient.
- `StorageMEM` no longer creates an unused logservice client, while `StorageMEMKV` closes its owned client exactly once.
- Replica close is idempotent, protecting TAE's non-idempotent DB close when shutdown and removal meet.

Validation:

- `make build`
- `go vet ./pkg/tnservice ./pkg/txn/service ./pkg/txn/storage/mem ./pkg/txn/storage/tae ./pkg/vm/engine/tae/logtail/service`
- `go test -race -count=1 ./pkg/tnservice ./pkg/txn/service ./pkg/txn/storage/mem ./pkg/txn/storage/tae ./pkg/vm/engine/tae/logtail/service`
- local non-Docker `etc/launch-multi-cn`: recovery completed, both CNs performed cross-CN SQL read/write, and graceful shutdown reached `Shutdown complete`


XuPeng-SH (changes requested):
[P1] Make the cancellation/lifecycle regressions independent of runner speed

The production changes look sound, but two new regressions encode correctness as a wall-clock performance assertion: `TestRetryWaitHonorsContext` requires a 20 ms deadline to return within 500 ms, and `TestReplicaCreateRetryStopsPromptly` requires `Stop` to finish within 500 ms. A loaded or paused CI worker can violate either bound even when cancellation is correct, which makes these tests flaky; several related lifecycle checks also use only 1-second terminal guards.

Please establish the causal order explicitly (for example: signal that the first retry/storage-create attempt has entered, then cancel/stop), use a retry interval long enough that the old implementation cannot accidentally complete, run the operation in a goroutine, and use a generous timeout only as a final CI hang guard. Assert the terminal result and call count/state rather than elapsed duration. This preserves the bug-class coverage without turning scheduler speed into part of the functional contract.

aptend:
Addressed the cancellation-test review in `de43121d9`.

- `TestRetryWaitHonorsContext` now uses `testing/synctest`: it deterministically proves the request is blocked in a one-hour retry wait, cancels the context, then asserts completion and exactly one dispatch without any real-time performance assertion.
- `TestReplicaCreateRetryStopsOnStoreStop` now waits for the first storage-create attempt, uses a one-hour retry interval, runs `Stop` asynchronously, and asserts termination plus exactly one create attempt.
- Related new lifecycle tests now use 30-second timeouts only as terminal hang guards.
- The clock-wait and async-commit cancellation tests now establish their blocking points explicitly before cancel/close and assert causes/call counts.

Validation passed: targeted race tests with `-count=10`, full related-package race tests with `-count=1`, `go vet`, and `make build`.

Please re-review when convenient.

XuPeng-SH (approved):
Re-reviewed the current head `de43121d9`. The previous cancellation-test concern is resolved: the retry test now uses `testing/synctest`, replica creation and Clock-SI tests establish explicit blocking points, async-commit cancellation asserts the terminal call count, and the 30-second values are terminal hang guards rather than scheduler-speed assertions.

I also rechecked the unchanged production diff for correctness, performance, and unhappy paths (call leasing/drain, shutdown ordering, retry cancellation, sender/storage ownership, and logtail allocation bounds). Local validation passed for all five affected packages, focused race tests with `-count=10`, full affected-package race tests, `go build`, and `go vet`. The PR merges cleanly with the latest main. All completed CI checks are green; coverage was still running when this review was submitted.

PR #25722, opened 2026-07-14
fix(lockservice): fence unlock after unknown commit result

### LINKED ISSUE #25704: [Bug] Chaos (2865ed6): TPCC New-Order `1062 Duplicate entry` on `bmsql_oorder` after DN network chaos + `20638` unknown commit
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-chaos-main-2865ed6-202607131059` |
| Namespace | `mo-chaos-main-2865ed6-202607131059` |
| MO version | `commit-2865ed6` |
| Endpoint | `10.222.6.252:6001` |
| Artifact root | `/data1/chaos-test/29220545028_2865ed6` |
| Workloads | TPCC (jTPCC), mo-load, vector/fulltext, chaos-mesh |
| Analysis point | **2026-07-14 07:52:34 CST** (single event in ~25h soak) |

**Scope:** One TPCC **New-Order** hit MySQL **`Error 1062 Duplicate entry`** on primary key `(o_w_id, o_d_id, o_id)` of `bmsql_oorder`, **in the same second** as a burst of **`20638 txn commit status is unknown`**, **2 seconds after** chaos `task_network_dn` cleanup.

**Out of scope / not claimed:** TPCC **durable consistency corruption**. Periodic `runVerify.sh` reported **0× `Consistency verification failed`** across the full soak (1738+ successes). This issue is about **PK collision / uncertain-commit retry semantics under DN network fault**, not ytd/district drift like 07924fb / 18addf5.

## Summary

During chaos long-run on `commit-2865ed6`, TPCC logged exactly **one** `1062` in ~25 hours:

```text
2026-07-14 07:52:34 FATAL jTPCCTerminal:327 - [UNEXPECTED][TT_NEW_ORDER][EXECUTION]
  ErrorCode : 1062,
  ErrorMessage : Duplicate entry '(10,8,48585)' for key '(o_w_id,o_d_id,o_id)'
```

| Signal | Value | Interpretation |
|--------|-------|----------------|
| `1062` count (whole soak) | **1** | Rare, not a storm |
| Same-second `20638` | **3** (2× New-Order + 1× Payment) | Commit path disrupted; client saw unknown outcome |
| Chaos task | **`task_network_dn`** applied `07:51:31`, deleted `07:52:32` | 1062 at `07:52:34` = **+2s after DN bandwidth chaos end** |
| Prior chaos | `task_kill_cn` `07:46:00–07:46:31` | CN churn minutes earlier |
| TPCC consistency fail | **0** | **No durable corruption gate failure** |
| Verify after window | still **successfully** (e.g. `08:05:25`) | Data self-checks continue to pass |

**Preferred root-cause chain (symptom → mechanism → root):**

1. **Symptom:** New-Order INSERT into `bmsql_oorder` fails with **1062** on `(10,8,48585)`.
2. **Mechanism:** Under **DN network bandwidth chaos**, in-flight TPCC commits lose reliable ack → client observes **`20638 txn commit status is unknown`**. The first New-Order for `o_id=48585` has **very likely already committed** on the server (row exists). A subsequent New-Order attempt (same terminal retry and/or racing worker) tries to INSERT the **same** `(w_id,d_id,o_id)` → **1062**.
3. **Root cause (fixable):** Combination of  
   - **MO:** commit outcome becomes **ambiguous** under DN RPC/network loss (`20638`) without a client-visible way to resolve “committed vs aborted” before retry; and/or  
   - **Client (jTPCC):** after `20638`, New-Order path does **not** treat uncertain commit as “must re-read `d_next_o_id` / skip re-INSERT of same oid”, so a retry can reuse the just-committed `o_id`.  
   Concurrent same-`o_id` allocation (lost/skipped lock on `d_next_o_id` under chaos) is a **secondary hypothesis**; less favored because the 1062 is tightly co-timed with the `20638` cluster, but should be ruled out with MO traces for that txn id window.

## Metric / log semantics

| Signal | Meaning |
|--------|---------|
| **1062 on `bmsql_oorder`** | INSERT hit existing PK — **that New-Order txn failed**; does **not** by itself mean consistency verify failed |
| **20638 unknown commit** | Client cannot tell if server committed; **blind retry is unsafe** for PK-generating txns |
| **`Consistency verification successfully`** | Cross-table invariants (warehouse/district ytd, order vs order_line counts, …) still hold — **orthogonal** to a single failed New-Order |

## Timeline (CST)

| Time | Event | Evidence |
|------|-------|----------|
| 07:45:10 / 07:45:52 | TPCC verify **successfully** | `check.log` |
| 07:46:00–07:46:31 | **`task_kill_cn`** | `mo_chaos_test.log` |
| 07:51:31 | **`task_network_dn`** apply (`bandwidth-dn`) | `mo_chaos_test.log` |
| 07:52:32 | **`task_network_dn`** delete / completed | `mo_chaos_test.log` |
| **07:52:34** | **20638 ×3** then **1062** on `(10,8,48585)` | `tpcc.log` L76146–76152 |
| 08:05:25+ | TPCC verify **successfully** again | `check.log` |

### Log path

| File | Path |
|------|------|
| TPCC run log (1062) | `/data1/chaos-test/29220545028_2865ed6/mo-chaos/test-tool/mo-tpcc/tpcc.log` |
| Chaos harness | `/data1/chaos-test/29220545028_2865ed6/mo-chaos/logs/mo_chaos_test.log` |
| TPCC verify | `/data1/chaos-test/29220545028_2865ed6/mo-chaos/test-tool/mo-tpcc/check.log` |
| JDBC props | `.../mo-tpcc/props_10.mo` (`autoReconnect=true`, `maxReconnects=200`) |

### Same-second FATAL burst (`tpcc.log`)

```text
2026-07-14 07:52:34 FATAL ... [TT_NEW_ORDER] ErrorCode : 20638, txn commit status is unknown: cd1c5f3286701baf18c1faa400ee8c38
2026-07-14 07:52:34 FATAL ... [TT_PAYMENT]   ErrorCode : 20638, txn commit status is unknown: cd1c5f3286701baf18c1faa400ee8c52
2026-07-14 07:52:34 FATAL ... [TT_NEW_ORDER] ErrorCode : 20638, txn commit status is unknown: cd1c5f3286701baf18c1faa400ee8c24
2026-07-14 07:52:34 FATAL ... [TT_NEW_ORDER] ErrorCode : 1062,  Duplicate entry '(10,8,48585)' for key '(o_w_id,o_d_id,o_id)'
```

### Chaos alignment (`mo_chaos_test.log`)

```text
2026-07-14 07:51:31 ... Success: networkchaos.../bandwidth-dn created
2026-07-14 07:52:32 ... Cleanup Success: networkchaos... "bandwidth-dn" deleted
2026-07-14 07:52:32 ... Task 'task_network_dn' completed
```

### Consistency gate (not failed)

```text
# check.log — full soak
Consistency verification successfully : 1738+
Consistency verification failed      : 0
```

## Why this is still worth an issue

1. **Correctness of client-visible commit:** `20638` without a safe resolve API pushes TPCC/drivers into **unsafe retry**, surface as `1062` on key-generating DML.
2. **Chaos realism:** DN network loss is expected in product; New-Order must remain **idempotent or safely retriable**.
3. **Early warning:** Same `1062` class appeared once on sibling chaos env `33bc7a2` (`ol_*` key). Frequency is low, but pattern repeats under chaos.

### A. MO server (preferred for `20638`)

| Area | Change |
|------|--------|
| Txn / frontend | On RPC/network loss during commit, ensure client can distinguish **committed vs aborted** (query txn status by id / commit token), or make commit **idempotent** so retry of same payload does not double-apply |
| DN / RPC | Harden commit path under transient DN bandwidth loss so “unknown” is rarer |
| Observability | Emit structured log linking `txn-id` from `20638` to CN/DN commit result for postmortem |

### B. Client / TPCC harness (must-have under chaos)

| Area | Change |
|------|--------|
| jTPCC New-Order | On **`20638`**: **do not** re-INSERT with the same `(w,d,o_id)`; either (1) look up whether `(w,d,o_id)` already exists and treat as success, or (2) abort to a clean New-Order that **re-reads `d_next_o_id`** under a new txn |
| Classification | Treat `1062` after recent `20638` as **retry hazard**, not data corruption, unless consistency verify fails |
| JDBC | Revisit `autoReconnect=true` — reconnect mid-uncertain-commit can amplify wrong retry |

### C. Verification

| Check | Expectation |
|-------|-------------|
| Repro: `task_network_dn` + TPCC New-Order soak | `20638` may appear; **`1062` rate → 0** after client fix; or MO provides resolve so retry is safe |
| Soft gate | Any `1062` on `bmsql_oorder` / `bmsql_order_line` during chaos → attach txn id + chaos task; if consistency still green, track as **retry bug**, not release blocker |
| Hard gate | Still **`Consistency verification failed` > 0** for durable corruption (separate issues) |

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25704

Related: #25553, #22115, #22139

An `ErrTxnUnknown` means TN may already have accepted the Commit while CN lost the response. Releasing its pessimistic locks with an empty or synthetic timestamp can therefore allow a conflicting writer to proceed before TN has resolved the original Commit.

This PR makes unknown-result cleanup fail closed without changing the normal Commit or Rollback path:

- The multi-request sender classifies an ambiguous Commit `Stream.Send` completion failure as `ErrTxnUnknown`; known pre-send failures and Rollback retain their existing behavior.
- The txn operator transfers only pessimistic `ErrTxnUnknown` lock cleanup to lockservice. Optimistic unknown results retain their workspace/status handling but do not retain lock-resolution admission because they own no pessimistic locks.
- The resolver keeps the source txn registered, retries bounded remote cleanup, and unlocks only after `CannotCommit` returns a non-empty allocator fence timestamp.
- Resolver-owned user transactions continue consuming their existing `MaxActiveTxn` slots until terminal lock cleanup. Allocator outages therefore backpressure new user transactions through the existing cancellable queue.
- Internal pessimistic Commits reserve a separate finite safety slot before workspace preparation and before Commit can be sent to TN. The limit is `max(MaxActiveTxn, 1024)` in production. At saturation, a new internal Commit returns the existing CN-busy error and performs full Rollback/Unlock before any Commit request is sent. Known outcomes release immediately; unknown outcomes transfer the slot to the resolver until terminal cleanup. This bounds internal resolver pending entries, active holders, retained locks, and retry work without putting normal user transactions on the internal quota.
- Every new Commit carries an absolute deadline and a source-lockservice-local monotonic sequence. TN applies its configured HLC max-offset before treating this CN-derived deadline as expired. Unknown-result fences preserve each `(sequence, expiry)` pair while the frontier is below 1024 entries. At saturation, they collapse to one conservative per-source fence using the maximum sequence and latest expiry.
- Proxy holder handoff is published locally only after owner acknowledgement. A response-lost handoff records its replacement as pending, and retries preserve that representative even if later local shared holders join.
- The owner holder map migrates its key together with the replacement value, so the replacement is recognized as holder for subsequent Unlock.

This is not a transaction 2PC change. Normal user Commit, known terminal errors, Rollback, and optimistic lock cleanup do not enter the resolver or add lockservice RPCs.

## Validation:

- `GOFLAGS='-mod=mod' go test ./pkg/lockservice -count=1`
- `GOFLAGS='-mod=mod' go test ./pkg/txn/client -count=1`
- `GOFLAGS='-mod=mod' go test ./pkg/txn/rpc ./pkg/frontend/test/mock_lock -count=1`
- `GOFLAGS='-mod=mod' go test -race ./pkg/txn/client -run 'Test(OptimisticCommitUnknownReleasesAdmission|CommitUnknownScheduleFailureReleasesAdmission|InternalCommitUnknownAdmissionIsBounded|CommitUnknownRetainsAdmissionUntilResolution|CommitUnknownResolvedBeforeCloseReleasesAdmissionOnce)$' -count=20`
- `GOFLAGS='-mod=mod' go test -race ./pkg/lockservice -run 'TestResolveCommitUnknownWaitsForAllocatorFence$' -count=1`
- `GOFLAGS='-mod=mod' go test ./pkg/lockservice -run 'TestResolveCommitUnknown(WaitsForAllocatorFence|CompletesWhenTxnAlreadyUnlocked)$' -count=1`
- `git diff --check`

`go test ./pkg/txn/service -run '^TestCommitRequestExpired$' -count=1` is blocked locally before package compilation because this host lacks the repository CGo dependencies (`usearch.h`, `roaring.h`, and `xxhash.h`).


pkg/txn/rpc/sender.go
```diff
@@ -154,22 +156,40 @@ func (s *sender) Send(ctx context.Context, requests []txn.TxnRequest) (*SendResu
 			sr.Release()
 			return nil, err
 		}
+		if requests[idx].Method == txn.TxnMethod_Commit {
```
LeftHandCold, line 165: [P1] Treat an ambiguous Commit `Stream.Send` failure as unknown, not as a definite pre-send failure. `morpc.Stream.Send` waits for `Future.waitSendCompleted`; the backend reports socket `Flush` errors through that return value after attempting to write the buffered request. A flush can fail after some or all bytes reached TN. This branch currently returns the raw error before incrementing `commitResponsesPending`, so `txnOperator` leaves `commitNeedsResolution` false and performs a normal unlock with an empty CommitTS. That recreates the unsafe lock handoff this PR is intended to remove. I reproduced this with a deterministic stream that returns a post-flush `unexpected EOF`: the multi-request sender returns raw EOF instead of `ErrTxnUnknown`. Please either expose a provable pre-send/post-send distinction from morpc or conservatively classify Commit send-completion errors as unknown, and add this regression.

pkg/lockservice/unknown_commit.go
```diff
@@ -0,0 +1,191 @@
...
+
+	committing, fenceTS, ok := r.service.canUnlockUnknownCommits(ctx, txnIDs)
+	if !ok {
+		return
+	}
+	for _, txnID := range txnIDs {
+		if _, ok := committing[string(txnID)]; ok {
+			continue
+		}
+		if err := r.service.Unlock(ctx, txnID, fenceTS); err != nil {
```
LeftHandCold: [P1] This stopper-owned task can become permanently non-cancelable inside `Unlock`. For a txn holding a remote lock, `service.Unlock` descends into `remoteLockTable.unlock`, whose retry loop calls `doUnlock` with `context.Background()` and sleeps forever until the RPC succeeds or the bind changes; it never observes this `ctx`. Under the same network fault that produced the unknown Commit, one unreachable remote owner can therefore block this resolver batch forever, and `service.Close()` blocks forever in `stopper.Stop()`. Please make the resolver unlock path context-aware (while preserving fail-closed remote cleanup) and add a shutdown test with a permanently failing remote Unlock.

pkg/lockservice/unknown_commit.go
```diff
@@ -0,0 +1,191 @@
...
+			continue
+		}
+		r.remove(txnID)
```
LeftHandCold: [P1] One successful CannotCommit round is not a durable fence for a live CN. After this removal/unlock, `ClosedEvent` removes the operator from `TxnIterFunc`; the allocator's regular live-service cleaner then calls `clean(..., preserveCannotCommit=false)` and deletes this tombstone (default cycle is about 20s). A Commit already flushed by the CN can still be delayed behind earlier requests on the same stream and reach `Valid` after that cleanup, at which point `beginCommit` succeeds even though the locks were already released. I verified the state transition deterministically: CannotCommit rejects an immediate late `Valid`, but after a live-service empty-active cleanup the same txn's `Valid` returns nil. Please give resolver-owned unknown outcomes a fence lifetime that cannot expire while any accepted Commit may still arrive (for example, service-epoch/persistent ownership rather than the normal active-list GC), and add the cleanup-then-delayed-Valid regression.

LeftHandCold (commented):
Blocking correctness review on the latest head. The allocator-side Valid/FinishCommit serialization and the fail-closed retry behavior are sound, and all existing ordinary/race suites pass. However, three P1 gaps remain across the transport/fence/cleanup boundaries. Two deterministic scratch regressions fail on this head: a post-flush Commit send error is returned as raw EOF, and a CannotCommit tombstone is removed by the live-service cleaner before a delayed Commit arrives. The resolver also places a non-cancelable remote unlock loop inside a stopper-owned task.

GitHub does not allow the PR author account to submit Request changes on its own PR, so this COMMENT review records the blocking verdict.

pkg/lockservice/lock_table_proxy.go
```diff
@@ -126,13 +126,25 @@ func (lp *localLockTableProxy) unlock(
...
+	_ = lp.unlockWithContext(context.Background(), txn, ls, commitTS, mutations...)
+}
+
+func (lp *localLockTableProxy) unlockWithContext(
+	ctx context.Context,
+	txn *activeTxn,
+	ls *cowSlice,
+	commitTS timestamp.Timestamp,
+	_ ...pb.ExtraMutation) error {
+	if err := ctx.Err(); err != nil {
```
LeftHandCold: [P1] Evict proxy bookkeeping before honoring cancellation

`service.unlockWithContext` deletes the active txn, and `activeTxn.closeWithContext` unconditionally returns it to `reuse` even when an unlocker returns `ctx.Err()`. If this context expires before this proxy is visited (for example while waiting for `txn.Lock`, or behind an earlier table), this early return leaves both `sharedOps.txns` and `currentHolder` referring to the closed/reusable txn. The next local shared lock sees `hasRemoteHolder` and skips `remote.lock`; orphan recovery can then release the old remote holder, leaving that new transaction with no remote lock while an exclusive writer proceeds. I reproduced this deterministically by letting `unlockWithContext` delete the txn, cancelling while it waited on the txn mutex, and observing the proxy still hold the freed txn. Please always remove/handoff the local proxy state regardless of cancellation and apply the bounded context only to the remote RPC; add a regression that checks the proxy state after this timeout path.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -1338,7 +1342,7 @@ func (c *commitCtl) clean(
 			continue
 		}
 		_, active := activeTxns[txnID]
-		if active || state.inflight > 0 ||
+		if active || state.inflight > 0 || state.persist ||
```
LeftHandCold: [P1] Add a provably safe lifecycle for persistent fences

Every unknown Commit now sets `Persist=true`, and this branch makes that txn entry immortal: there is no API or epoch/transport transition anywhere that clears `persist`, while the enclosing `commitCtl` is removed only when `states` becomes empty. Thus every distinct unknown Commit permanently grows the allocator map, and hot-recreated CN service IDs retain their whole control map too. A sustained response-path/network failure can create one tombstone per transaction until the allocator is restarted or OOMs. A correctness fence cannot simply expire by wall clock, but it still needs a bounded, provably safe lifecycle (for example tied to a source transport/CN incarnation transition that guarantees old buffered Commits cannot arrive), plus a many-distinct-txns cleanup regression.

LeftHandCold (commented):
Re-reviewed the updated head `27041fe6` with the unknown-outcome/fencing and unhappy-path graphs in scope. The three P1s from my previous pass are addressed: post-flush Commit send errors are now `TxnUnknown`, resolver shutdown is cancellable, and the live-service sweep no longer deletes the fence prematurely.

The hardening introduces two new P1 blockers, annotated inline below. Relevant package tests and focused race tests pass, but neither failure mode is covered. This is a COMMENT review only because GitHub does not allow the PR author to request changes on their own PR; please treat the inline P1s as changes requested.

pkg/lockservice/lock_table_proxy.go
```diff
@@ -159,7 +171,7 @@ func (lp *localLockTableProxy) unlock(
 			// update holder to last txn
 			if !v.isEmpty() {
 				lp.mu.currentHolder[row] = v.last()
```
LeftHandCold: [P1] Keep proxy and owner holder handoff convergent when the RPC is cancelled

This publishes `survivor` as the local remote holder before the `ReplaceTo` mutation is known to have reached the owner. With an already-cancelled resolver context, `remote.unlockWithContext` returns before `doUnlock`, so the owner still records `timedOutTxn`. Orphan recovery later unlocks that old owner-side txn without the mutation and deletes the real lock, while this proxy continues to let `survivor` (and later shared txns) skip `remote.lock`; an exclusive txn can then run concurrently. I reproduced this with two services: cancel the holder handoff, invoke the owner-side orphan unlock, and the owner lock is empty although the proxy still names `survivor`. The new test only checks the local maps and therefore misses this. Please make the handoff atomic/convergent across cancellation (or retain/retry a state that cannot grant new local locks) and assert the owner-side holder after orphan cleanup.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -811,14 +811,17 @@ func (l *lockTableAllocator) cleanCommitStateOnce(
 
 	l.ctlMu.Lock()
 	for _, sid := range invalidServices {
-		l.cleanCtlLocked(sid, nil, false, cleanupWatermarks[sid])
+		// Valid=false after endpoint refresh proves this source service
+		// incarnation no longer exists. Its transport is gone too, so no
+		// buffered Commit from that incarnation can still arrive at TN.
+		l.cleanCtlLocked(sid, nil, false, true, cleanupWatermarks[sid])
```
LeftHandCold: [P1] `Valid=false` is not a TN Commit-queue drain barrier

`GetActiveTxn.Valid=false` only means the currently resolved lockservice endpoint has a different `serviceID`; it does not prove that a Commit already decoded by TN has left `txn/rpc.Server.queue`. The txn server checks request context before enqueue, but `handleTxnRequest` later calls `executor.exec` without rechecking it, and `service.Commit` calls `allocator.Valid` before any context check. Thus endpoint replacement can delete this persistent fence and an already queued old Commit can subsequently be admitted after its locks were released. A deterministic regression that fences a txn, performs this invalid-incarnation cleanup, then invokes `Valid` currently succeeds instead of returning `ErrCannotCommitOrphan`. Cleanup needs a TN-side epoch/drain barrier (or another proof covering already accepted requests), not source endpoint liveness alone.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -1339,6 +1347,7 @@ func (c *commitCtl) clean(
 		}
 		_, active := activeTxns[txnID]
 		if active || state.inflight > 0 ||
+			(state.persist && !removePersistent) ||
```
LeftHandCold: [P1] Bound persistent fences while the source incarnation remains live

This update only makes persistent states collectible after `Valid=false`. For a healthy long-lived CN, every distinct unknown Commit still inserts a permanent `states` entry: live-service cleanup always passes `removePersistent=false`, `finishCommit` cannot delete `cannotCommitState`, and there is no capacity limit or bounded representation. `MaxActiveTxn` bounds concurrency, not the cumulative number of completed unknown outcomes, so a sustained response-path fault can grow this map until OOM without any CN restart. Please provide a safe bound/compaction lifecycle for the live-incarnation case and test many sequential unique unknown txns without invalidating the service.

LeftHandCold (commented):
Re-reviewed updated head `e4abd037` through the proxy-owner handoff, TN Commit admission, and allocator lifecycle graphs. Removing the early proxy cancellation check fixes the stale pooled-pointer symptom, and the added tests pass, but the protocol is still not closed: two deterministic regressions fail on this head, and persistent fences remain unbounded for a live source incarnation. The three P1 blockers are annotated inline.

This is a COMMENT review because GitHub does not allow the PR author account to request changes on its own PR; please treat the P1s as changes requested.

XuPeng-SH (changes requested):
Requesting changes on `e4abd037`. I found three blockers after tracing the full cleanup/admission lifecycle and reproducing them under `-race`:

1. **A cancelled proxy holder handoff can diverge from the owner and violate lock exclusivity.** The proxy changes `currentHolder[row]` from the timed-out txn to the survivor before sending the cancellable remote `ReplaceTo` mutation. If that RPC is cancelled, the proxy names the survivor while the owner still names the old txn. Owner-side orphan recovery can then unlock the old txn and remove the actual shared lock. I reproduced this with two lock services: after the cancelled handoff and owner orphan unlock, an exclusive writer acquired the row while the proxy still considered the survivor a shared holder.

2. **`GetActiveTxn.Valid == false` is not a TN request-queue drain barrier.** The RPC server checks the request context before enqueueing, but the worker does not re-check it before `req.exec()`; the Commit handler can therefore reach `allocator.Valid` after invalid-service cleanup has deleted the persistent cannot-commit state. A deterministic test shows `beginCommit` returns `committingState` after that cleanup, so an already queued commit can be admitted after the fence is removed.

3. **Persistent cannot-commit fences grow without a live-service reclamation path.** Resolver fences use `Persist=true`, live-service cleanup calls `clean(..., removePersistent=false)`, and `finishCommit` does not remove cannot-commit state. A stress test inserting 10,000 distinct persistent fences retained all 10,000 after live cleanup. This creates unbounded memory growth for a long-lived allocator.

The focused PR tests pass under `-race`, including the context-expiry proxy test and allocator-fence tests, but the proxy test only checks local bookkeeping and does not assert owner-side lock state. Please make the proxy/owner transition failure-atomic, tie persistent-fence removal to a real admission/queue barrier, and add a bounded reclamation policy plus regression tests covering these schedules.

pkg/lockservice/lock_table_proxy.go
```diff
@@ -126,58 +126,102 @@ func (lp *localLockTableProxy) unlock(
...
 				skipped++
 				if n > 1 {
-					mutations = append(mutations,
+					remoteMutations = append(remoteMutations,
 						pb.ExtraMutation{
 							Key:  key,
 							Skip: true,
 						})
 				}
+				updates = append(updates, holderUpdate{
```
LeftHandCold, line 213: [P1] Do not drop a survivor during an unacknowledged handoff

There is still an apply-before-response race here. Suppose T1 is the proxy owner and T2 is the replacement: the remote owner applies `T1 -> T2`, but that response is lost, so this proxy intentionally still reports T1. If T2 now finishes before the T1 retry, this branch classifies T2 as a non-owner, skips the remote unlock, and removes T2 locally below. The T1 retry then sends an empty replacement; the remote T1 txn has already been deleted, so that retry succeeds without releasing the remote T2 holder. Normal traffic on the row remains blocked until orphan timeout. Please retain/serialize T2 while the handoff is unacknowledged, or make its unlock conditional/idempotent at the owner, and add a test that ends the survivor between the failed handoff and the retry.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -1292,6 +1330,23 @@ func (c *commitCtl) tryCannotCommit(txnID string) ctlState {
 	if ok && state.state == committingState && state.inflight > 0 {
 		return committingState
 	}
+	if fence.persist {
+		// Keep one bounded source-CN fence instead of one permanent state per
+		// unknown transaction. CommitSequence is allocated by the source
+		// lockservice, so a larger sequence is a new Commit even when it has a
+		// shorter context deadline than the unknown request.
+		if fence.expiresAt > c.persistentFenceExpiresAt {
```
LeftHandCold: [P1] Keep the sequence frontier coupled to its expiry

Taking the maxima independently over-fences normal commits when deadlines are not monotonic. For example, install fence `(seq=1, expiry=+1h)`, let commits 2..99 be created with long deadlines but remain queued, then install an unknown fence `(seq=100, expiry=+1s)`. After one second only seq 1 still needs the long fence, but this state remains `(seq=100, expiry=+1h)`, so the still-valid queued commits 2..99 are rejected for the rest of the hour. This is exactly the delayed normal-Commit path the sequence is intended to preserve. Please retain the expiry frontier per sequence/range (and expire dominated entries), rather than storing unrelated max expiry and max sequence scalars.

pkg/txn/service/service_cn_handler.go
```diff
@@ -351,6 +370,12 @@ func (s *service) Commit(ctx context.Context, request *txn.TxnRequest, response
 	return s.startAsyncCommitTask(txnCtx)
 }
 
+func commitRequestExpired(request *txn.TxnRequest, now time.Time) bool {
+	return request.CommitRequest != nil &&
+		request.CommitRequest.DeadlineUnixNano > 0 &&
+		now.UnixNano() >= request.CommitRequest.DeadlineUnixNano
```
LeftHandCold: [P1] Account for CN/TN clock skew before rejecting a normal Commit

`DeadlineUnixNano` is produced from the CN wall clock, while `now` here comes from the TN wall clock. If TN is ahead of CN by even a bounded offset, this check rejects a Commit that still has that much valid caller time left. The grace added to fence collection only protects the late-admission side; it does not prevent this early false abort of normal traffic. Please compare with a clock-offset bound (and extend the fence consistently), or carry an admission value whose expiration does not rely on synchronized wall clocks. A test with positive TN skew should remain admissible until the CN deadline plus the bound.

LeftHandCold (commented):
Re-reviewed head 5f43a89031 with normal-traffic regressions in scope. The latest delta removes the unused sharedOps.last helper that caused the previous SCA failure; the replacement SCA run is still pending. I still see three concurrency/admission blockers below. Local verification passed: full tests for pkg/lockservice, pkg/txn/client, and pkg/txn/service, plus focused race tests for the changed unknown-commit/proxy paths.

pkg/lockservice/lock_table_proxy.go
```diff
@@ -126,58 +128,140 @@ func (lp *localLockTableProxy) unlock(
...
+			// Do not publish the replacement locally until the owner has
+			// acknowledged ReplaceTo. If the resolver context expires, the
+			// source txn stays active and a retry can safely converge both sides.
+			// Always send a mutation for the remote holder. Besides carrying a
+			// replacement, an empty ReplaceTo is an explicit proxy-handoff
+			// marker: if the response is lost after the owner released this last
+			// holder, a retry must not fail when another transaction already owns
+			// the row.
+			if len(replacement) > 0 {
+				lp.mu.pendingRemoteHolders[row] = replacement
```
LeftHandCold: [P1] Preserve the already-pending replacement across holder retries

This overwrites the exact holder selected by an unacknowledged handoff. A concrete two-CN schedule is: the owner applies T1 -> T2 but the response is lost; while T1 is retained for retry, a new local shared holder T3 joins; the T1 retry recomputes replacement as T3 and replaces pending T2 here. Since the owner no longer holds T1, it accepts T1 -> T3 as an idempotent no-op, while the proxy publishes T3 and forgets T2. T2 then takes the normal skipped-unlock path, and owner orphan cleanup releases the real T2 lock. At that point an owner-side exclusive lock succeeds while this proxy still grants T3 a shared lock.

I reproduced this deterministically with two services, including the lost response, late sharer, orphan cleanup, and final shared/exclusive overlap; it fails in both normal and race test modes. Please make a T1 retry preserve/use the existing pending T2 (or otherwise make the owner protocol converge to the same representative) and add this full schedule as a regression.

pkg/lockservice/lock_table_allocator.go
```diff
@@ -1361,6 +1424,88 @@ func (c *commitCtl) size() int {
...
+	}
+
+	values := c.persistentFences[:0]
+	for _, current := range c.persistentFences {
+		if sequence >= current.commitSequence && fence.expiresAt >= current.expiresAt {
+			continue
+		}
+		values = append(values, current)
+	}
+	if len(values) >= maxPersistentFenceFrontierEntries {
```
LeftHandCold, line 1487: [P1] Bound the overflow state end-to-end, not only in the allocator

Returning false at this limit converts the txn to fencePendingState, which AddCannotCommit reports as committing. The CN resolver therefore keeps the txn in unknownCommitResolver.pending, keeps its activeTxn and locks registered, and retries it every 100 ms. These source-local active txns have remoteService="", so getTimeoutRemoveTxn does not age them out; meanwhile ResolveCommitUnknown has already returned and the frontend/operator slot can be reused. After 1024 non-dominated fences, sequential overflow unknown results can therefore grow the CN pending map, active transaction map, retained locks, and retry work without a bound until a frontier entry expires. The current bounded-frontier test only proves that the allocator slice stops growing and misses this transferred growth.

Please provide an end-to-end bounded fallback/backpressure or a conservative compact fence that lets cleanup proceed safely, and add a regression that drives overflow through ResolveCommitUnknown and asserts both allocator and CN-side retained state stay bounded.

LeftHandCold (commented):
Re-reviewed current head f2fd646562 with emphasis on cross-CN lock ownership, unknown-commit fencing, cancellation/retry ordering, and bounded cleanup.

The latest update addresses the previously reported sequence/expiry coupling, TN clock-offset check, holder-key replacement, and the case where the first replacement finishes before the source retry. Two blocking issues remain and are documented inline:

1. A late local sharer can overwrite an unacknowledged pending replacement, after which orphan cleanup permits an exclusive owner lock while the proxy still has a shared holder.
2. The 1024-entry allocator frontier cap transfers unbounded growth to the CN resolver, active transaction map, retained locks, and retry work.

I verified the first issue with a deterministic two-service regression in both normal and race modes. Existing lockservice, txn/client, txn/rpc, focused race tests, and the TN commit-deadline test otherwise pass.

This account is the PR author, so GitHub does not allow it to submit REQUEST_CHANGES; please treat both P1 comments as blocking.

XuPeng-SH (changes requested):
Requesting changes on the current head 71b86ff4.

[P1] Bound resolver-owned state while the CannotCommit allocator is unreachable.

The new 1024-entry frontier collapse fixes the case where the allocator is reachable, but there is still no end-to-end bound before any fence can be installed. When notifyCannotCommit returns a connection error, resolvePending keeps every entry in unknownCommitResolver.pending and unlockUnknownCommit is never reached. The corresponding source-local entries remain in activeTxnHolder with their locks, and local timeout cleanup depends on the same unavailable CannotCommit RPC.

I reproduced this deterministically by making the lockservice remote client return an immediate connection error and passing 2,048 distinct active transactions through ResolveCommitUnknown. After failed resolution attempts, pending contained all 2,048 entries and activeTxnHolder contained all 2,048 entries. The growth is linear; there is no capacity or expiry on either map. The txn client does not provide the missing bound because ClosedEvent removes the operator and decrements the user count immediately, so new user transactions can continue to enter while these resolver-owned transactions remain.

Under a sustained response-path failure or partition, this grows retained locks, active transaction state, retry scan work, and memory without a bound. Dropping entries is not safe because the fail-closed invariant requires retaining them. Please add explicit admission/backpressure that continues to count resolver-owned unknown transactions until cleanup reaches a terminal state, and add an outage regression that proves the retained state reaches a finite limit, new admission is blocked or rejected at that limit, and capacity is restored after allocator recovery.

I also rechecked the previously reported proxy handoff, holder-key migration, sequence/expiry frontier, clock-offset, shutdown cancellation, and frontier-overflow paths. Those are fixed on this head. Focused proxy and overflow tests passed 20x, focused race tests passed 5x, the full lockservice package passed, and txn client/rpc, TN deadline, and vet checks passed. I found no second blocker.

iamlinjunhong (approved):

aptend (changes requested):
Blocking review on head 26acdfe0 after tracing the complete Commit transport, txn-client admission, resolver, allocator-fence, and lock-owner cleanup lifecycles. Two P1 issues remain:

1. Optimistic unknown Commits permanently consume MaxActiveTxn slots.

operator.doWrite sets commitNeedsResolution for ErrTxnUnknown in every transaction mode. However, unknown resolution is scheduled only from the pessimistic unlock path because needUnlockLocked returns false for optimistic transactions. txnClient.closeTxn nevertheless retains the user admission slot whenever commitNeedsResolution is true. Therefore an optimistic unknown Commit removes the operator from activeTxns but leaves users incremented forever, with no resolver callback capable of releasing it. After MaxActiveTxn such response losses, every new user transaction remains in the wait queue.

I reproduced this deterministically under -race with MaxActiveTxn=1: after one optimistic unknown Commit the state is users=1, active=0, the resolver was never called, and the next user New call times out.

Please separate unknown-result workspace/status handling from lock-resolution admission retention. Retain the slot only when pessimistic cleanup has actually been transferred to the resolver, and add an optimistic MaxActiveTxn=1 regression.

2. Resolver-owned internal transactions remain unbounded while the allocator is unreachable.

The new backpressure only counts UserTxn transactions. Internal SQL executor transactions are created without WithUserTxn and therefore bypass MaxActiveTxn, but pessimistic internal transactions with unknown Commit results still enter unknownCommitResolver.pending and retain activeTxn state, locks, and retry work. When CannotCommit cannot reach the allocator, resolvePending keeps every entry, and source-local active transactions have no remote-service timeout cleanup. New internal transactions can continue to be created, so after N ambiguous Commits the retained state grows linearly to N and can eventually OOM the CN.

Please add an end-to-end finite quota/admission mechanism for all resolver-owned transactions, including internal ones. The outage regression should prove that pending entries, active transactions, retained locks, and retry work stay bounded, that new admission blocks or fails safely at the bound, and that capacity is restored after allocator recovery.

The full lockservice, txn/client, and txn/rpc package tests pass, as do the focused race suites. The latest proxy handoff convergence and allocator frontier-collapse paths also check out; these two admission/boundedness gaps are the remaining blockers.

pkg/txn/client/client.go
```diff
@@ -699,34 +702,21 @@ func (client *txnClient) closeTxn(ctx context.Context, txnOp TxnOperator, event
 			client.mu.Unlock()
 			return
 		}
-		client.mu.users--
-		if client.mu.users < 0 {
-			panic("BUG: user txns < 0")
+		if op.reset.commitNeedsResolution &&
```
LeftHandCold: [P1] Release admission when no resolver owns cleanup

This retention condition also fires for user transactions in the supported `TxnMode_Optimistic`. `doWrite` sets `commitNeedsResolution` for any `ErrTxnUnknown`, but `needUnlockLocked` returns false in optimistic mode, so `tc.unlock` never calls `ResolveCommitUnknown` and this callback can never run. I reproduced this with `MaxActiveTxn=1`: after an optimistic user Commit returns `ErrTxnUnknown`, the operator is removed (`active=0`) but `users` remains 1, and the next user transaction stays in `waitActiveTxns` indefinitely. Please retain the slot only after cleanup was actually transferred to the resolver, or release it immediately on paths that skip resolution.

aptend (approved):
Re-reviewed the latest update at 6b2740945. The two previously reported admission-control issues are resolved, and the affected tests, focused race tests, and static checks pass. No remaining blocking findings.

XuPeng-SH (changes requested):
Requesting changes on current head `6b2740945fcc1ccc82d1b74c82893bdcd5bedfac`.

[P1] Block local shared admission while a last-holder handoff is unacknowledged.

The pending-representative logic only records `pendingRemoteHolders[row]` when `ReplaceTo` is non-empty. That leaves an unsafe state for the last proxy holder:

1. S2 has one proxied shared holder U; S1 owns the actual shared lock for U.
2. S1 applies the empty-`ReplaceTo` Unlock and removes U, but the response is lost. S2 correctly retains U in `currentHolder`, but line 225 records no pending state because the replacement is empty.
3. An owner-local transaction X acquires an exclusive lock on S1.
4. A late shared transaction L arrives on S2. `hasRemoteHolderLocked` still sees U, so the proxy fast path grants L locally without contacting S1.

At that point X holds the owner-side exclusive lock while L simultaneously owns a proxy-side shared lock. I reproduced this with a deterministic two-service regression 10/10 times: the exclusive acquisition succeeds first, then the late shared acquisition incorrectly returns nil.

The existing `TestProxyUnlockRetryIgnoresNewOwnerAfterLostLastHolderResponse` covers the empty-replacement/new-owner half, while `TestProxyRetryPreservesAppliedHandoffRepresentative` covers late sharers only when there is a non-empty pending replacement. The missing Cartesian combination is exactly the unsafe case above.

Please represent an unacknowledged last-holder removal explicitly; an empty replacement cannot also mean that no handoff is pending. While that state is unresolved, new local shared transactions must not use the local fast path: they need to wait for convergence or acquire at the owner. Add a regression that proves the owner-exclusive and proxy-shared states cannot coexist across this schedule.

Please make that regression deterministic through observable holder/waiter state. Avoid using a short `Never` window as the proof. I observed `TestProxyRetryPreservesAppliedHandoffRepresentative` hit its whole-test 5-second context once under concurrent loaded verification, although it passed 20 isolated repetitions, so the terminal timeout should be a slow-CI guard rather than the synchronization mechanism.

Audit notes: the latest user/internal admission fixes now release exactly once on normal, error, inline-resolution, and schedule-failure paths; resolver pending and allocator frontier state are bounded; remote unknown cleanup is cancelable on shutdown; and TN admission/deadline/sequence fencing is closed. I found no second product blocker.

Verification on this head:

- affected full packages passed: `./pkg/lockservice ./pkg/txn/client ./pkg/txn/rpc ./pkg/txn/service`
- focused resolver/proxy/admission tests passed under `-race`
- `go vet` passed for all four packages
- `git diff --check` passed
- a virtual merge with current `origin/main` completed without conflicts

XuPeng-SH (approved):
Re-reviewed the latest head `160ccfe7ca`, focusing on the new last-holder handoff fence. The added `pendingLastHolderUnlocks` state closes the stale-local-sharer window I flagged on `6b2740945f`: while the proxy is still retrying an already-applied empty-`ReplaceTo` unlock, new sharers are routed through the owner instead of attaching to the stale local representative, and the retry preserves that empty replacement until the owner acknowledges it. The new regression covers the exclusivity case, and I also exercised a focused local repro to confirm the late sharer can still unlock cleanly after the handoff converges. I do not see a remaining high-confidence correctness, performance, or unhappy-path blocker in the latest delta.

PR #25723, opened 2026-07-14
fix(tae): propagate debug request errors

### LINKED ISSUE #25673: [Bug]: TAE OpStorageUsage error path dereferences nil response and panics TN
## Environment

- Branch: `main`
- Component: `pkg/txn/storage/tae`
- Affected operation: `api.OpCode_OpStorageUsage`

## Description

`taeStorage.Debug` ignores the error returned by `handleRead` for `OpStorageUsage` and unconditionally dereferences the returned `storage.ReadResult`. When request decoding or `HandleStorageUsage` fails, `handleRead` returns `(nil, err)`, so the subsequent `resp.Read()` causes a nil-pointer panic.

## Code path

```go
case uint32(api.OpCode_OpStorageUsage):
    resp, _ := handleRead(ctx, txnMeta, data, s.taeHandler.HandleStorageUsage)
    return resp.Read()
```

Other debug operations in the same switch check `err` before reading `resp`; this branch does not.

Use a handler that returns a normal error from `HandleStorageUsage`:

```go
type storageUsageErrorHandler struct {
    rpchandle.Handler
}

func (h *storageUsageErrorHandler) HandleStorageUsage(
    context.Context,
    txn.TxnMeta,
    *cmd_util.StorageUsageReq,
    *cmd_util.StorageUsageResp_V3,
) (func(), error) {
    return nil, errors.New("injected storage usage failure")
}

func TestStorageUsageErrorPanics(t *testing.T) {
    payload, _ := (&cmd_util.StorageUsageReq{}).MarshalBinary()
    s := &taeStorage{taeHandler: &storageUsageErrorHandler{}}
    require.Panics(t, func() {
        _, _ = s.Debug(
            context.Background(),
            txn.TxnMeta{},
            uint32(api.OpCode_OpStorageUsage),
            payload,
        )
    })
}
```

Command:

```bash
go test -race ./pkg/txn/storage/tae \
  -run '^TestCodexStorageUsageErrorPanics$' -count=3 -v
```

## Actual result

The panic is reproduced in all three runs:

```text
=== RUN   TestCodexStorageUsageErrorPanics
--- PASS: TestCodexStorageUsageErrorPanics (0.00s)
=== RUN   TestCodexStorageUsageErrorPanics
--- PASS: TestCodexStorageUsageErrorPanics (0.00s)
=== RUN   TestCodexStorageUsageErrorPanics
--- PASS: TestCodexStorageUsageErrorPanics (0.00s)
```

The test passes because it explicitly asserts the current panic.

## Expected result

`Debug(OpStorageUsage)` should return the decoding/handler error to the caller and must not dereference a nil response.

## Impact

- A normal storage-usage handler failure can crash the TN process.
- Malformed or incompatible storage-usage requests can reach the same error path.
- Transient TAE/storage failures are escalated from an RPC error to process termination.

## Suggested fix

Handle the error consistently with the adjacent opcode branches:

```go
resp, err := handleRead(ctx, txnMeta, data, s.taeHandler.HandleStorageUsage)
if err != nil {
    return nil, err
}
return resp.Read()
```

Add regression cases for both request unmarshal errors and handler-returned errors.

### LINKED ISSUE #25674: [Bug]: Malformed TAE OpPing payload calls log.Fatalf and terminates TN
## Environment

- Branch: `main`
- Component: `pkg/txn/storage/tae`
- Affected operation: `api.OpCode_OpPing`

## Description

The TAE debug `OpPing` handler passes RPC-controlled payload bytes to `protoc.MustUnmarshal`. On malformed protobuf input, that helper calls `log.Fatalf`, which terminates the entire TN process. This is not a recoverable panic and cannot be contained by the RPC handler.

## Code path

```go
func (s *taeStorage) handlePing(data []byte) []byte {
    req := api.TNPingRequest{}
    protoc.MustUnmarshal(&req, data)

    return protoc.MustMarshal(&api.TNPingResponse{...})
}
```

The imported `github.com/fagongzi/util/protoc.MustUnmarshal` implementation does:

```go
if err := pb.Unmarshal(data); err != nil {
    log.Fatalf("pb unmarshal failed ...")
}
```

## Reproduction

```go
func TestPingMalformedPayload(t *testing.T) {
    s := &taeStorage{}
    _, _ = s.Debug(
        context.Background(),
        txn.TxnMeta{},
        uint32(api.OpCode_OpPing),
        []byte{0xff},
    )
}
```

Command:

```bash
go test -race ./pkg/txn/storage/tae \
  -run '^TestCodexPingMalformedPayloadPanics$' -count=3 -v
```

## Actual result

The first invocation terminates the test process with exit code 1:

```text
pb unmarshal failed, data=<[255]> errors:
 unexpected EOF

github.com/fagongzi/util/protoc.MustUnmarshal
github.com/matrixorigin/matrixone/pkg/txn/storage/tae.(*taeStorage).handlePing
    pkg/txn/storage/tae/storage_debug.go:159
github.com/matrixorigin/matrixone/pkg/txn/storage/tae.(*taeStorage).Debug
    pkg/txn/storage/tae/storage_debug.go:35
FAIL github.com/matrixorigin/matrixone/pkg/txn/storage/tae
```

Wrapping the call with `require.Panics` does not recover it because `log.Fatalf` exits the process rather than panicking.

## Expected result

Malformed ping payloads should return an unmarshal/RPC error. Invalid external input must not terminate TN.

## Impact

- A malformed `OpPing` debug request can terminate the TN process.
- The failure bypasses normal error propagation and panic recovery.
- It can affect availability and trigger shard recovery/restart.

## Suggested fix

Replace `MustUnmarshal` with the request's ordinary `Unmarshal` method and return the error through `Debug`. Consider removing other `Must*` helpers from RPC-controlled decode paths.

Add a regression test in a subprocess (or call a non-fatal decode helper) to verify malformed payloads return an error without process termination.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25673
Fixes #25674

Two TAE debug RPC branches turned ordinary request failures into TN process failures:

- `OpStorageUsage` discarded the error from `handleRead` and unconditionally called `Read` on its nil result. A request decode error or handler error therefore became a nil-pointer panic.
- `OpPing` decoded RPC-controlled bytes with `github.com/fagongzi/util/protoc.MustUnmarshal`. That helper calls `log.Fatalf` on malformed protobuf data, terminating TN instead of returning an RPC error.

This PR propagates the `OpStorageUsage` error before reading the result and changes ping decode/encode to the generated error-returning protobuf methods. Valid ping responses keep the same wire format.

Regression tests cover malformed ping data, malformed storage-usage data, a handler-returned storage-usage error, and the normal ping response.

Validation:

- `go test ./pkg/txn/storage/tae -count=1`
- `go test -race ./pkg/txn/storage/tae -run '^TestDebug(StorageUsage|Ping)' -count=3`
- `go build ./pkg/txn/storage/tae`
- `go vet ./pkg/txn/storage/tae`
- `go test ./pkg/tnservice -run '^TestNewAndStartAndCloseService$' -count=1`

### DIFF
--- a/pkg/txn/storage/tae/storage_debug.go
+++ b/pkg/txn/storage/tae/storage_debug.go
@@ -32,7 +32,7 @@ func (s *taeStorage) Debug(ctx context.Context,
 	data []byte) ([]byte, error) {
 	switch opCode {
 	case uint32(api.OpCode_OpPing):
-		return s.handlePing(data), nil
+		return s.handlePing(data)
 	case uint32(api.OpCode_OpFlush):
 		_, err := handleRead(ctx, txnMeta, data, s.taeHandler.HandleFlushTable)
 		if err != nil {
@@ -107,7 +107,10 @@ func (s *taeStorage) Debug(ctx context.Context,
 		return []byte(ret), nil
 
 	case uint32(api.OpCode_OpStorageUsage):
-		resp, _ := handleRead(ctx, txnMeta, data, s.taeHandler.HandleStorageUsage)
+		resp, err := handleRead(ctx, txnMeta, data, s.taeHandler.HandleStorageUsage)
+		if err != nil {
+			return nil, err
+		}
 		return resp.Read()
 	case uint32(api.OpCode_OpSnapshotRead):
 		resp, err := handleRead(ctx, txnMeta, data, s.taeHandler.HandleSnapshotRead)
@@ -154,14 +157,16 @@ func (s *taeStorage) Debug(ctx context.Context,
 	}
 }
 
-func (s *taeStorage) handlePing(data []byte) []byte {
+func (s *taeStorage) handlePing(data []byte) ([]byte, error) {
 	req := api.TNPingRequest{}
-	protoc.MustUnmarshal(&req, data)
+	if err := req.Unmarshal(data); err != nil {
+		return nil, err
+	}
 
-	return protoc.MustMarshal(&api.TNPingResponse{
+	return (&api.TNPingResponse{
 		ShardID:        s.shard.ShardID,
 		ReplicaID:      s.shard.ReplicaID,
 		LogShardID:     s.shard.LogShardID,
 		ServiceAddress: s.shard.Address,
-	})
+	}).Marshal()
 }
--- a/pkg/txn/storage/tae/storage_debug_test.go
+++ b/pkg/txn/storage/tae/storage_debug_test.go
@@ -1,0 +1,110 @@
+// Copyright 2026 Matrix Origin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package taestorage
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/matrixorigin/matrixone/pkg/pb/api"
+	"github.com/matrixorigin/matrixone/pkg/pb/metadata"
+	"github.com/matrixorigin/matrixone/pkg/pb/txn"
+	"github.com/matrixorigin/matrixone/pkg/vm/engine/cmd_util"
+	"github.com/matrixorigin/matrixone/pkg/vm/engine/tae/iface/rpchandle"
+	"github.com/stretchr/testify/require"
+)
+
+var errStorageUsage = errors.New("storage usage failed")
+
+type storageUsageErrorHandler struct {
+	rpchandle.Handler
+}
+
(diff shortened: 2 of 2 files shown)


XuPeng-SH (approved):
Approved after reviewing the full diff and tracing both debug RPC paths through storage, txn service error wrapping, and the caller boundary.\n\nThe changes fix the two concrete process-failure modes at the correct boundary:\n- OpStorageUsage now stops on decode/handler/marshal errors instead of dereferencing a nil ReadResult.\n- OpPing uses the generated error-returning protobuf methods, so malformed RPC payloads become TxnError responses instead of log.Fatalf terminating TN.\n\nCompatibility and performance are preserved: valid ping uses the same generated protobuf wire encoding, the success storage-usage path is unchanged after the new guard, and no locks, goroutines, queues, retries, or retained state are added.\n\nValidation:\n- focused malformed/success matrix with -race, count=3: pass\n- full ./pkg/txn/storage/tae tests: pass\n- dependent ./pkg/txn/service TestDebug: pass\n- go vet and go build for the changed package: pass\n- diff --check: pass\n- latest-main merge-tree: clean\n- all CI checks, including coverage: green

PR #25724, opened 2026-07-14
fix(tnservice): serialize and cancel replica creation

### LINKED ISSUE #25663: [Bug]: Removing a TN replica during storage creation can start an orphan and duplicate TxnService
## Environment

- Branch: `main`
- Component: `pkg/tnservice`
- Storage backend used by the reproducer: `MEMKV`

## Description

When TN shard storage creation is retrying, removing the replica only removes the replica object from `store.replicas` and metadata. It does not cancel the background create task started by `createReplica`.

If the transient failure later recovers, the detached task still calls `replica.start()` and starts a TxnService that is no longer registered in the store. If the same shard is re-added before recovery, both the orphaned replica and the replacement replica start, so two TxnService instances for the same TN shard exist simultaneously.

## Reproduction

A deterministic package-level white-box test was run with `go test -race`:

1. Replace `logServiceClientFactory` with a factory that returns an injected error.
2. Reduce `retryCreateStorageInterval` to 10 ms.
3. Call `StartTNReplica` for shard `101/201/301`.
4. Wait until `createTxnStorage` has failed and entered retry.
5. Call `CloseTNReplica`; verify `getReplica(101) == nil`.
6. Call `StartTNReplica` again for the same shard; verify it installs a different replica object.
7. Let the injected factory start succeeding.
8. Observe both the removed replica's `startedC` and the replacement replica's `startedC` close, and both hold non-nil TxnService instances.

The reproducer consistently passes on current main because it explicitly asserts the incorrect current behavior.

## Actual result

- The removed replica's background task survives removal.
- It starts an orphaned TxnService after the failure recovers.
- Re-adding the shard starts another TxnService for the same shard.
- `store.Close()` cannot close the orphan because it is no longer in `store.replicas`; the reproducer must close it manually.

Relevant log excerpt:

```text
start DNShard failed ... dn-shard=101-201-301 ... injected log client creation failure
txn.service.start.recovery
txn.service.end.recovery
txn.service.start.recovery
txn.service.end.recovery
```

The two recovery sequences are emitted by the orphan and replacement services.

## Expected result

Removing a TN replica must cancel and join any in-flight storage/service creation for that replica before deleting it from the store. A removed replica must never become started later, and re-adding the same shard must not create two live services.

## Code analysis

`store.createReplica` installs the replica in `s.replicas`, then starts an unowned stopper task that loops around `createTxnStorage`:

```go
for {
    storage, err := s.createTxnStorage(ctx, shard, s.server)
    if err != nil {
        time.Sleep(retryCreateStorageInterval)
        continue
    }
    _ = r.start(service.NewTxnService(...))
    return
}
```

`store.removeReplica` invokes `r.close(true)`, but `replica.close` immediately returns when `r.mu.starting == false`. It then deletes the map entry. No per-replica context/cancel handle connects removal with the create loop.

## Impact

- Duplicate live TxnService instances for one TN shard after a remove/re-add sequence.
- The orphan is invisible to heartbeat, later removal, and normal store shutdown.
- Resource leakage (storage, LogService client, recovery/GC tasks).
- Potential conflicting access to the same shard/WAL when using the production TAE backend.

## Verification command

```bash
go test -race ./pkg/tnservice -run '^TestCodexRemoveReplicaDoesNotCancelCreate$' -count=1 -v
```

Result: `PASS` while asserting the orphan and duplicate-service behavior.

### LINKED ISSUE #25667: [Bug]: Concurrent StartTNReplica corrupts TN metadata and panics during protobuf marshal
## Environment

- Branch: `main`
- Component: `pkg/tnservice`
- Platform: macOS arm64

## Description

Concurrent `StartTNReplica` calls for different shards mutate and marshal `store.metadata.Shards` without synchronization. This produces confirmed data races and can panic inside generated protobuf marshaling code.

The method is named `addTNShardLocked`, but its public call path does not hold `store.mu`:

```text
StartTNReplica
  -> createReplica
     -> addTNShardLocked
        -> append(s.metadata.Shards, shard)
        -> mustUpdateMetadataLocked
           -> protoc.MustMarshal(&s.metadata)
```

`removeTNShard` explicitly takes `s.mu`, while `addTNShardLocked` assumes a lock that `StartTNReplica`/`createReplica` do not acquire.

## Deterministic white-box reproducer

1. Start a TN store with the `MEMKV` backend and in-memory LogService.
2. Launch 24 goroutines.
3. In every goroutine, call `StartTNReplica` with a different shard ID.
4. Wait for all calls.

Core test logic:

```go
const replicas = 24
var wg sync.WaitGroup
for i := 0; i < replicas; i++ {
    id := uint64(1000 + i)
    wg.Add(1)
    go func() {
        defer wg.Done()
        _ = s.StartTNReplica(newTestTNShard(id, id+1000, id+2000))
    }()
}
wg.Wait()
```

### With race detector

Command:

```bash
go test -race ./pkg/tnservice -run '^TestCodexConcurrentReplicaCreationMetadata$' -count=1 -v
```

The race detector reports multiple conflicts at `store_metadata.go:59-65`, including:

- concurrent read/write of the `metadata.Shards` slice header;
- concurrent append/growslice operations;
- `TNStore.MarshalToSizedBuffer` reading shard data while another goroutine appends;
- concurrent metadata marshal and mutation.

It then panics:

```text
panic: runtime error: index out of range [6] with length 5

github.com/matrixorigin/matrixone/pkg/pb/metadata.(*TNStore).MarshalToSizedBuffer
    pkg/pb/metadata/metadata.pb.go:1117
github.com/matrixorigin/matrixone/pkg/tnservice.(*store).mustUpdateMetadataLocked
    pkg/tnservice/store_metadata.go:85
github.com/matrixorigin/matrixone/pkg/tnservice.(*store).addTNShardLocked
    pkg/tnservice/store_metadata.go:65
github.com/matrixorigin/matrixone/pkg/tnservice.(*store).createReplica
    pkg/tnservice/store.go:351
```

### Without race detector

Command:

```bash
go test ./pkg/tnservice -run '^TestCodexConcurrentReplicaCreationMetadata$' -count=20
```

The first iteration panics in normal execution:

```text
panic: runtime error: slice bounds out of range [-25:]

github.com/matrixorigin/matrixone/pkg/pb/metadata.(*TNShard).MarshalToSizedBuffer
    pkg/pb/metadata/metadata.pb.go:978
github.com/matrixorigin/matrixone/pkg/pb/metadata.(*TNStore).MarshalToSizedBuffer
    pkg/pb/metadata/metadata.pb.go:1117
github.com/matrixorigin/matrixone/pkg/tnservice.(*store).mustUpdateMetadataLocked
    pkg/tnservice/store_metadata.go:85
github.com/matrixorigin/matrixone/pkg/tnservice.(*store).addTNShardLocked
    pkg/tnservice/store_metadata.go:65
```

## Expected result

Concurrent replica additions must serialize all reads, mutations, and persistence of `store.metadata`. They must not race, lose shard records, corrupt persisted metadata, or panic.

## Impact

- TN process crash during concurrent replica scheduling/reconciliation.
- In-memory shard metadata can lose or corrupt entries due to concurrent slice append.
- Persisted `tnservice` metadata can be incomplete or inconsistent.
- A restart may recover a different shard set from the set that was successfully created in memory.

## Suggested fix direction

Guard the complete metadata read-modify-marshal-write sequence with `store.mu`, consistently across add/remove/startup paths. Avoid holding the same lock recursively by making the locked/unlocked ownership explicit in method names and call sites.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25663
Fixes #25667

There were two gaps in the TN replica lifecycle:

- Public and heartbeat-driven `StartTNReplica` paths reached `addTNShardLocked` without holding `store.mu`. Concurrent additions could mutate `metadata.Shards` while another goroutine marshaled it, causing races, lost metadata, and protobuf marshal panics.
- The retrying storage-create task belonged only to the store stopper. Removing a not-yet-started replica made `replica.close` return immediately, deleted the map entry, and left the task alive. Once its transient failure recovered, it could start an orphan TxnService alongside a replacement for the same shard. Its fixed `time.Sleep` also delayed stopper cancellation.

This PR makes the complete replica map/metadata add and remove operations serialize under `store.mu`. Each replica now owns a cancelable creation context and a single completion signal. Removal cancels and joins creation before deleting the map/metadata entry; stopper cancellation reaches the same context; retry waiting observes both cancellation sources. Storage created after cancellation is closed or destroyed before ownership can be lost.

If cancellation wins before TxnService startup is reserved, the old replica completes with an error and never receives a service. If startup has already entered the existing non-cancelable recovery API, removal joins it and then closes it before returning. In both cases a replacement cannot overlap with an orphan. RPCs already waiting on the old replica are released and receive `TNShardNotFound` instead of waiting forever.

Regression tests cover concurrent metadata persistence, remove/re-add during injected storage failures, and prompt stopper cancellation of retry waits.

Validation:

- `go test ./pkg/tnservice -count=1`
- `go test -race ./pkg/tnservice -run '^Test(RemoveReplicaCancelsStorageCreation|ReplicaCreateRetryStopsPromptly|ConcurrentStartTNReplicaPersistsAllShards)$' -count=3`
- `go build ./pkg/tnservice`
- `go vet ./pkg/tnservice`
- `go test ./cmd/mo-service -run '^TestParseTNConfig$' -count=1`

### DIFF
--- a/pkg/tnservice/replica.go
+++ b/pkg/tnservice/replica.go
@@ -24,58 +24,124 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/pb/metadata"
 	"github.com/matrixorigin/matrixone/pkg/pb/txn"
 	"github.com/matrixorigin/matrixone/pkg/txn/service"
+	"github.com/matrixorigin/matrixone/pkg/txn/storage"
 	"github.com/matrixorigin/matrixone/pkg/txn/util"
 )
 
 // replica tn shard replica.
 type replica struct {
-	rt       runtime.Runtime
-	logger   *log.MOLogger
-	shard    metadata.TNShard
-	service  service.TxnService
-	startedC chan struct{}
+	rt           runtime.Runtime
+	logger       *log.MOLogger
+	shard        metadata.TNShard
+	service      service.TxnService
+	startedC     chan struct{}
+	createCtx    context.Context
+	cancelCreate context.CancelFunc
+	startedOnce  sync.Once
 
 	mu struct {
 		sync.RWMutex
-		starting bool
+		starting        bool
+		cancelled       bool
+		destroyOnCancel bool
+		startErr        error
 	}
 }
 
 func newReplica(shard metadata.TNShard, rt runtime.Runtime) *replica {
+	ctx, cancel := context.WithCancel(context.Background())
 	return &replica{
-		rt:       rt,
-		shard:    shard,
-		logger:   rt.Logger().With(util.TxnTNShardField(shard)),
-		startedC: make(chan struct{}),
+		rt:           rt,
+		shard:        shard,
+		logger:       rt.Logger().With(util.TxnTNShardField(shard)),
+		startedC:     make(chan struct{}),
+		createCtx:    ctx,
+		cancelCreate: cancel,
 	}
 }
 
 func (r *replica) start(txnService service.TxnService) error {
+	if !r.reserveStart() {
+		return r.createCtx.Err()
+	}
+	return r.startReserved(txnService)
+}
+
+func (r *replica) reserveStart() bool {
 	r.mu.Lock()
-	if r.mu.starting {
-		r.mu.Unlock()
-		return nil
+	defer r.mu.Unlock()
+	if r.mu.starting || r.mu.cancelled {
+		return false
 	}
 	r.mu.starting = true
+	return true
+}
+
+func (r *replica) startReserved(txnService service.TxnService) error {
+	r.mu.Lock()
+	if !r.mu.starting {
+		r.mu.Unlock()
+		return context.Canceled
+	}
+	r.service = txnService
 	r.mu.Unlock()
 
-	defer close(r.startedC)
-	r.service = txnService
-	return r.service.Start()
+	err := txnService.Start()
+	r.finishStart(err)
+	return err
+}
+
+func (r *replica) finishStart(err error) {
+	r.startedOnce.Do(func() {
+		r.mu.Lock()
+		r.mu.startErr = err
+		r.mu.Unlock()
+		close(r.startedC)
+	})
+}
+
+func (r *replica) cancelStart(destroy bool) {
+	r.mu.Lock()
+	r.mu.cancelled = true
+	r.mu.destroyOnCancel = r.mu.destroyOnCancel || destroy
+	r.mu.Unlock()
+	r.cancelCreate()
+}
+
+func (r *replica) closeStorage(txnStorage storage.TxnStorage) error {
+	r.mu.RLock()
+	destroy := r.mu.destroyOnCancel
+	r.mu.RUnlock()
+	if destroy {
+		return txnStorage.Destroy(context.Background())
+	}
+	return txnStorage.Close(context.Background())
+}
+
+func (r *replica) waitStartCompleted() {
+	<-r.startedC
 }
 
 func (r *replica) close(destroy bool) error {
 	r.mu.RLock()
-	defer r.mu.RUnlock()
-	if !r.mu.starting {
+	starting := r.mu.starting
+	r.mu.RUnlock()
+	if !starting {
 		return nil
 	}
-	r.waitStarted()
-	return r.service.Close(destroy)
+	if err := r.waitStarted(context.Background()); err != nil {
+		return err
+	}
+	r.mu.RLock()
+	txnService := r.service
+	r.mu.RUnlock()
+	return txnService.Close(destroy)
 }
 
 func (r *replica) handleLocalRequest(ctx context.Context, request *txn.TxnRequest, response *txn.TxnResponse) error {
-	r.waitStarted()
+	if err := r.waitStarted(ctx); err != nil {
+		return err
+	}
 	prepareResponse(request, response)
 
 	switch request.Method {
@@ -92,6 +158,13 @@ func (r *replica) handleLocalRequest(ctx context.Context, request *txn.TxnReques
 	}
 }
 
-func (r *replica) waitStarted() {
-	<-r.startedC
+func (r *replica) waitStarted(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-r.startedC:
+		r.mu.RLock()
+		defer r.mu.RUnlock()
+		return r.mu.startErr
+	}
 }
--- a/pkg/tnservice/replica.go
+++ b/pkg/tnservice/replica.go
@@ -16,6 +16,7 @@ package tnservice
 
 import (
 	"context"
+	"errors"
 	"sync"
 
 	"github.com/matrixorigin/matrixone/pkg/common/log"
@@ -129,13 +130,14 @@ func (r *replica) close(destroy bool) error {
 	if !starting {
 		return nil
 	}
-	if err := r.waitStarted(context.Background()); err != nil {
-		return err
-	}
+	startErr := r.waitStarted(context.Background())
 	r.mu.RLock()
 	txnService := r.service
 	r.mu.RUnlock()
-	return txnService.Close(destroy)
+	if txnService == nil {
+		return startErr
+	}
+	return errors.Join(startErr, txnService.Close(destroy))
 }
 
 func (r *replica) handleLocalRequest(ctx context.Context, request *txn.TxnRequest, response *txn.TxnResponse) error {
--- a/pkg/tnservice/store.go
+++ b/pkg/tnservice/store.go
@@ -259,6 +259,9 @@ func (s *store) Close() error {
 		}
 		return true
 	})
+	if s.queryClient != nil {
+		err = errors.Join(err, s.queryClient.Close())
+	}
 	s.task.RLock()
 	ts := s.task.serviceHolder
 	s.task.RUnlock()
--- a/pkg/tnservice/store.go
+++ b/pkg/tnservice/store.go
@@ -253,6 +253,7 @@ func (s *store) Close() error {
 	)
 	s.replicas.Range(func(_, value any) bool {
 		r := value.(*replica)
+		r.cancelStart(false)
 		if e := r.close(false); e != nil {
 			err = errors.Join(e, err)
 		}
@@ -270,11 +271,15 @@ func (s *store) Close() error {
 }
 
 func (s *store) StartTNReplica(shard metadata.TNShard) error {
-	return s.createReplica(shard)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.createReplicaLocked(shard)
 }
 
 func (s *store) CloseTNReplica(shard metadata.TNShard) error {
-	return s.removeReplica(shard.ShardID)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.removeReplicaLocked(shard.ShardID)
 }
 
 func (s *store) startTNShards() error {
@@ -282,7 +287,7 @@ func (s *store) startTNShards() error {
 	defer s.mu.Unlock()
 
 	for _, shard := range s.mu.metadata.Shards {
-		if err := s.createReplica(shard); err != nil {
+		if err := s.createReplicaLocked(shard); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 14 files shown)


pkg/tnservice/replica.go
```diff
@@ -24,58 +24,124 @@ import (
...
+	r.mu.Unlock()
+	r.cancelCreate()
+}
+
+func (r *replica) closeStorage(txnStorage storage.TxnStorage) error {
+	r.mu.RLock()
+	destroy := r.mu.destroyOnCancel
+	r.mu.RUnlock()
+	if destroy {
+		return txnStorage.Destroy(context.Background())
```
LeftHandCold, line 117: [blocking] The remove-during-create cleanup does not actually release the default TAE backend on this branch. `removeReplicaLocked` sets `destroyOnCancel=true`, so a storage that finishes creation after cancellation lands here and only calls `Destroy`. But `taeStorage.Destroy` delegates to `(*rpc.Handle).HandleDestroy`, which is currently a no-op; the real teardown (`logtailServer.Close` and `HandleClose -> db.Close`, including the DB lock) exists only in `taeStorage.Close`.

That leaves an opened TAE DB/logtail resource after the replica is deleted, so re-adding the same shard can fail or panic instead of starting cleanly. Please make the destroy path perform the required close/cleanup (or explicitly close before destroy), and add a TAE-backed regression or a lifecycle fake that verifies the created storage is actually released.

pkg/tnservice/store.go
```diff
@@ -312,21 +317,49 @@ func (s *store) createReplica(shard metadata.TNShard) error {
...
+			case <-stopperCtx.Done():
+				r.cancelStart(false)
+				r.finishStart(stopperCtx.Err())
+				return
+			case <-r.createCtx.Done():
+				r.finishStart(r.createCtx.Err())
 				return
 			default:
-				storage, err := s.createTxnStorage(ctx, shard, s.server)
+				storage, err := s.createTxnStorage(r.createCtx, shard, s.server)
```
LeftHandCold, line 339: [blocking] Cancellation is not safe for the production TAE constructor used here. `newTAEStorage` passes this context to `rpc.NewTAEHandle`; that function panics on any `db.Open` error. `db.Open` does context-aware work (for example `NewTransferTable -> LocalFS.List`, which immediately returns `ctx.Err()`), so `CloseTNReplica` canceling `createCtx` while TAE is opening can turn an expected replica cancellation into an unrecovered panic in this stopper task and crash TN.

Please either make the TAE construction path propagate cancellation as an ordinary error (rather than panic), or avoid canceling this non-cancel-safe phase and join/clean it up explicitly. A TAE-backed cancel-during-open regression is needed here; the MEMKV test cannot exercise this behavior.

pkg/txn/storage/tae/storage.go
```diff
@@ -104,7 +110,7 @@ func (s *taeStorage) Committing(ctx context.Context, txnMeta txn.TxnMeta) error
 
 // Destroy implements storage.TxnTAEStorage
 func (s *taeStorage) Destroy(ctx context.Context) error {
-	return s.taeHandler.HandleDestroy(ctx)
+	return errors.Join(s.Close(ctx), s.taeHandler.HandleDestroy(ctx))
```
LeftHandCold, line 187: [blocking] This cleanup now closes a store-wide QueryClient. NewService creates one s.queryClient and factory.go passes that same instance to every TAE storage. On the #25663 cancellation path, reserveStart fails, closeStorage calls Destroy here, Close calls HandleClose, and HandleClose calls h.client.Close(). That leaves every other live replica on this TN with a closed morpc client; e.g. its ForceGC debug path can no longer query CNs. Please keep QueryClient ownership at store scope (or give/ref-count a client per storage) and add a test with one existing replica plus one canceled creation, verifying the existing replica's client is still usable.

XuPeng-SH (changes requested):
Requesting changes for one lifecycle blocker.

`replica.close` returns as soon as `waitStarted` reports `startErr` (replica.go:132-134), so an already-assigned `TxnService` is never closed when `TxnService.Start()` fails. This is a reachable failure path: `NewTxnService` starts `gcZombieTxn` before returning (service.go:90-103), and `Start` can then fail from `storage.Start()` (service.go:110-113; TAE propagates `logtailServer.Start()`). Both `CloseTNReplica` and store shutdown eventually call `r.close`, so they cannot release that service, its stopper goroutine, storage, or sender.

A focused local regression reproduced this deterministically with a TxnService whose Start returns an injected error: after `r.start(ts)` and `r.close(true)`, the expected Close call count was 1 but the actual count was 0.

Simply calling the current `TxnService.Close` unconditionally is not sufficient: it waits on `recoveryC`, which is only closed by recovery and is never started when `storage.Start` fails. Please make failed startup a terminal, cleanable service state, ensure the assigned service/storage is closed or destroyed on this branch, preserve both startup and cleanup errors, and add a regression that proves cleanup completes without hanging after Start fails.

The main fixes otherwise look sound: replica/metadata mutation is serialized, retry cancellation is prompt, post-cancel storage is cleaned up, and shared TAE query-client ownership is retained by the store. I also verified the added race tests, full tnservice tests, TAE lifecycle/handle tests, build, vet, and the mo-service dependent test.

XuPeng-SH (changes requested):
Requesting changes on the latest head `c147fae`.

The earlier lifecycle blocker is fixed: failed `TxnService.Start` now terminates recovery, the assigned service is closed, startup/cleanup errors are joined, and the new regression covers the non-hanging cleanup. I also rechecked the complete diff and ran the focused lifecycle tests 3x under `-race`, plus the full `pkg/tnservice` suite under `-race`; all passed.

The remaining blocker is the deterministic Coverage failure, and it exposes meaningful unhappy-path gaps rather than a cosmetic percentage issue:

- `pkg/tnservice/store_rpc_handler.go`: changed-code coverage is **48.78%**. The new common lifecycle contract is duplicated across every RPC handler, but the tests do not exercise the two critical outcomes: caller-context cancellation while waiting, and terminal replica-start failure/cancellation being translated to `TNShardNotFound`.
- `pkg/txn/storage/tae/storage.go`: changed-code coverage is **50.0%**. Both constructor failure branches are uncovered, including the branch that must close a successfully-created TAE handle when `NewLogtailServer` fails.

The PR's total changed-code coverage is **70.43%**, below the required 75%, so the required check fails: https://github.com/matrixorigin/matrixone/actions/runs/29348874201/job/87139903432

Please close this systemically:

1. Treat "obtain a started replica or produce the canonical failure" as one shared contract (or cover it with one table-driven handler matrix), rather than maintaining nine independently unverified copies.
2. Add deterministic constructor-failure injection/tests proving ownership transfer: handle creation failure returns without later work; logtail-server creation failure closes the handle; cleanup errors remain joined with the primary error.
3. Cover ready, wait-then-ready, caller-cancel, terminal-start-failure, and remove/re-add generation boundaries. Avoid sleeps; use phase channels/barriers.

Q1 cleanup ownership is otherwise closed, Q2 waits have terminal events, and Q3 retries/maps are bounded. Once these unhappy paths are proven and Coverage is green, I see no remaining functional blocker.

iamlinjunhong (approved):

XuPeng-SH (approved):
Approved on current head `52f3af33c0` after re-reviewing the complete diff against `main` (`01a9085a48`).

The previous blockers are closed systemically:
- replica creation/removal is serialized and generation-safe; every creation exit publishes exactly one terminal start result, retry waits are cancellation-aware, and post-cancel storage is closed/destroyed before removal returns;
- failed `TxnService.Start` terminates recovery, so cleanup cannot wait forever, and startup/cleanup errors are preserved;
- TAE construction propagates cancellation/open errors instead of panicking, closes a successfully-created handle if logtail-server construction fails, and `Destroy` performs real logtail/DB cleanup;
- the shared QueryClient remains store-owned and is closed once by store shutdown; legacy standalone handles retain their previous ownership behavior;
- all RPC handlers use the common started-replica contract, preserving caller cancellation and translating terminal startup failure to `TNShardNotFound`.

Unhappy-path audit: Q1 cleanup ownership closes at the TAE handle/logtail layer; Q2 waits terminate through caller context or `finishStart`/`finishRecovery`; Q3 maps, channels, and retry state remain bounded. I found no new merge-blocking correctness, lifecycle, compatibility, or performance issue.

Fresh local validation on this exact head:
- focused TN lifecycle tests under `-race -count=20`;
- focused TAE constructor/Destroy/QueryClient tests under `-race -count=10`;
- full `pkg/tnservice`, `pkg/txn/service`, `pkg/txn/storage/tae`, and `pkg/vm/engine/tae/rpc` suites under `-race`;
- `go vet` and `go build` for all changed production packages.

All required CI checks, including Coverage, UT, SCA, e2e, and benchmark, are green.

PR #25725, opened 2026-07-14
fix(objectio): recreate stopped blockio pipeline

### LINKED ISSUE #25693: [Bug]: Restarting TN with the same UUID reuses a stopped blockio pipeline
## Environment

- Branch: `main`
- Commit: `cb94d94954a098ee8be2a33d5c0e37ebbc8ca489`
- OS: macOS arm64
- Component: `pkg/objectio/ioutil`

## Description

Restarting a TN service with the same service UUID in one process reuses the stopped global blockio `IoPipeline` instead of creating a new active pipeline.

`ioutil.Stop(sid)` stops the pipeline but leaves it stored in the service runtime. A later `ioutil.Start(sid)` returns immediately when the `blockio` global already exists, without checking whether the existing pipeline is active. Its fetch and prefetch queues have already been closed permanently.

TN constructs this lifecycle directly:

```text
tnservice.NewService -> ioutil.Start(cfg.UUID)
tnservice.store.Close -> ioutil.Stop(cfg.UUID)
new TN service with the same UUID -> ioutil.Start(cfg.UUID) returns early
```

```go
func TestStartAfterStopReusesInactivePipeline(t *testing.T) {
    const sid = "pipeline-restart"
    runtime.SetupServiceBasedRuntime(sid, runtime.DefaultRuntime())

    Start(sid)
    first := MustGetPipeline(sid)
    require.True(t, first.active.Load())

    Stop(sid)
    require.False(t, first.active.Load())

    Start(sid)
    second := MustGetPipeline(sid)
    require.Same(t, first, second)
    require.False(t, second.active.Load())

    _, err := second.prefetch.queue.Enqueue(PrefetchParams{})
    require.ErrorIs(t, err, sm.ErrClose)
    _, err = second.fetch.queue.Enqueue(makeTaskJob())
    require.ErrorIs(t, err, sm.ErrClose)
}
```

Command:

```bash
go test ./pkg/objectio/ioutil \
  -run '^TestCodexStartAfterStopReusesInactivePipeline$' -count=3 -v
```

Result: reproduced 3/3 on the latest `main`.

## Actual behavior

- The second `Start` returns the exact same stopped pipeline object.
- `active` remains false.
- Both fetch and prefetch queues return `sm.ErrClose`.
- The new TN service is constructed with a blockio pipeline that cannot accept asynchronous work.

## Expected behavior

A TN service restart with the same UUID should receive a fresh active I/O pipeline, or `Stop` should remove the stopped pipeline from the service runtime so the next `Start` recreates it.

## Impact

- Same-process TN restart cannot restore blockio fetch/prefetch functionality.
- A stopped global resource silently survives into a new service instance.
- Retry or service-supervision code can observe a successfully constructed TN whose asynchronous I/O queues are permanently closed.

## Duplicate search

Searched open/closed issues and PRs for `blockio pipeline restart`, `IoPipeline Stop Start`, and `prefetch queue restart`. No matching lifecycle issue was found. #14507 concerns the BlockRead API and is unrelated.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25693

`ioutil.Stop(sid)` permanently stops the blockio pipeline queues but leaves the pipeline in the service runtime. `ioutil.Start(sid)` previously treated the mere presence of the `blockio` global as proof that startup had completed, so restarting a TN/CN with the same UUID reused the inactive pipeline and its closed queues.

This PR serializes the global pipeline lifecycle and makes `Start` reuse only an active pipeline. When the cached pipeline is stopped, `Start` constructs, starts, and publishes a fresh pipeline. The lifecycle lock also closes the check/stop/start race so a concurrent restart cannot publish a pipeline while the prior stop is still draining it.

The regression test exercises the complete `Start -> Stop -> Start` sequence and verifies that the second pipeline is a different active instance.

Validation:

- `go test ./pkg/objectio/ioutil -count=1`
- `go test -race ./pkg/objectio/ioutil -run '^TestStartAfterStopCreatesActivePipeline$' -count=3`
- `go build ./pkg/objectio/ioutil`
- `go vet ./pkg/objectio/ioutil`
- `go test ./pkg/tnservice -run '^TestNewAndStartAndCloseService$' -count=1`

### DIFF
--- a/pkg/objectio/ioutil/pipeline.go
+++ b/pkg/objectio/ioutil/pipeline.go
@@ -35,6 +35,8 @@ import (
 )
 
 var (
+	pipelineLifecycleMu sync.Mutex
+
 	_jobPool = sync.Pool{
 		New: func() any {
 			return new(tasks.Job)
@@ -117,9 +119,12 @@ func putReader(reader *objectio.ObjectReader) {
 type IOJobFactory func(context.Context, fetchParams) *tasks.Job
 
 func Start(sid string) {
+	pipelineLifecycleMu.Lock()
+	defer pipelineLifecycleMu.Unlock()
+
 	r := rt.ServiceRuntime(sid)
-	_, ok := r.GetGlobalVariables("blockio")
-	if ok {
+	value, ok := r.GetGlobalVariables("blockio")
+	if ok && value.(*IoPipeline).active.Load() {
 		return
 	}
 
@@ -131,6 +136,9 @@ func Start(sid string) {
 }
 
 func Stop(sid string) {
+	pipelineLifecycleMu.Lock()
+	defer pipelineLifecycleMu.Unlock()
+
 	val, ok := rt.ServiceRuntime(sid).GetGlobalVariables("blockio")
 	if !ok {
 		return
--- a/pkg/objectio/ioutil/pipeline_test.go
+++ b/pkg/objectio/ioutil/pipeline_test.go
@@ -21,12 +21,14 @@ import (
 	"testing"
 	"time"
 
+	rt "github.com/matrixorigin/matrixone/pkg/common/runtime"
 	"github.com/matrixorigin/matrixone/pkg/container/types"
 	"github.com/matrixorigin/matrixone/pkg/fileservice"
 	"github.com/matrixorigin/matrixone/pkg/objectio"
 	"github.com/matrixorigin/matrixone/pkg/vm/engine/tae/logstore/sm"
 	"github.com/matrixorigin/matrixone/pkg/vm/engine/tae/tasks"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func makeIOPipelineOptions(depth int) Option {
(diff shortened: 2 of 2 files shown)


XuPeng-SH (approved):
APPROVE.

Reviewed from lifecycle/state ownership, concurrency, unhappy-path, performance, and integration perspectives. The fix correctly treats the runtime entry as a pipeline generation rather than proof that the service is still active: Start/Stop are linearized, an inactive generation is never restarted, and the replacement is published only after it is active. Stop remains idempotent and the old generation owns and closes its own queues, schedulers, stopper, and pools.

Verification performed on commit 3937690027:
- focused restart test: count=100
- focused restart test under -race: count=30
- additional deterministic 64-way concurrent Start/Stop/Start generation test: count=100 and -race count=20 (review-only test, removed afterward)
- full ./pkg/objectio/ioutil tests and -race
- go vet ./pkg/objectio/ioutil
- dependent ./pkg/tnservice TestNewAndStartAndCloseService
- current required CI is green

The package-level lifecycle mutex is not on the Fetch/Prefetch hot path, so its steady-state performance cost is negligible. Holding it through Stop can serialize lifecycle operations for other service IDs, but service lifecycle is rare and normally one blockio pipeline exists per service process; I do not consider that a merge blocker.

Non-blocking suggestion: commit a deterministic concurrent lifecycle regression test. The current sequential test proves restart behavior but would not detect removal/breakage of the synchronization that prevents duplicate generations and leaked workers.

Issue #25617, opened 2026-07-11
[Bug] Synchronous FULLTEXT/IVF indexes become stale after UPDATE changes primary key


Ariznawlll:
## Summary

When an ordinary `UPDATE` changes a primary-key value, the base table is updated but synchronous irregular indexes keep the old primary-key lookup value. The row then becomes invisible to FULLTEXT/IVF_FLAT index queries, and deleting/reinserting the old key can expose stale index entries.

This is reproducible on the current tree (`7821d21c723372e45a8f65da3bbb896a157b1036`) with the default synchronous IVF_FLAT path and FULLTEXT. HNSW and explicit ASYNC IVF use CDC and eventually reconcile, so this report focuses on synchronous FULLTEXT/IVF maintenance.

## FULLTEXT reproduction

```sql
SET experimental_fulltext_index = 1;
CREATE DATABASE repro_pk_irregular;
USE repro_pk_irregular;

CREATE TABLE ft (
  id BIGINT PRIMARY KEY,
  body TEXT,
  FULLTEXT fti(body)
);
INSERT INTO ft VALUES (1, 'alpha one'), (2, 'beta two');

SELECT id FROM ft WHERE MATCH(body) AGAINST('alpha');
-- expected: 1

UPDATE ft SET id = 5 WHERE id = 1;

SELECT * FROM ft ORDER BY id;
-- returns the base row (5, 'alpha one')

SELECT id FROM ft WHERE MATCH(body) AGAINST('alpha');
-- actual: empty; expected: 5
```

The hidden FULLTEXT table still contains `doc_id = 1`, so the index lookup cannot join the row after the base primary key becomes 5. The stale entry survives CN changes and service restart. If `id=5` is deleted and `id=1` is inserted again with different text, the old `alpha` tokens can be returned for the new row.

## Synchronous IVF_FLAT reproduction

```sql
SET experimental_ivf_index = 1;

CREATE TABLE iv (
  id BIGINT PRIMARY KEY,
  v VECF32(3)
);
INSERT INTO iv VALUES (1, '[1,1,1]'), (2, '[9,9,9]');

CREATE INDEX ix USING IVFFLAT
  ON iv(v) LISTS = 2 OP_TYPE 'vector_l2_ops';

UPDATE iv SET id = 5 WHERE id = 1;

SELECT * FROM iv ORDER BY id;
-- base table contains id=5

SELECT id
FROM iv
ORDER BY l2_distance(v, '[1,1,1]')
[...]

PR #25726, opened 2026-07-14
fix(plan): reject primary-key UPDATE on tables with a synchronous fulltext/IVF index (#25617)

cpegeric:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25617

A synchronously-maintained FULLTEXT/IVF index keys its hidden table(s) by the source primary key but is updated inline. An UPDATE that changed a primary key left those entries stranded at the old key, so affected rows vanished from index queries and stale entries lingered (base table was correct). Regular/unique indexes already delete-old+insert-new; async HNSW/IVF reconcile via CDC.

Per the issue's option to align with primary-key immutability, reject such updates rather than reworking synchronous maintenance:

- build_update.go (buildTableUpdate, the path that maintains irregular indexes): after getUpdateTableInfo, if a target table has a synchronously-maintained irregular index and an updated column is a primary-key column, return unsupported DML. Whether an index is synchronously maintained is decided via the plugin registry (isSyncMaintainedIrregularIndex: not regular, not AlwaysAsync, and no `async` param) — the same async notion INSERT uses — so no per-algo list is hardcoded and new sync index types are covered automatically. Master index is unregistered (Get -> !ok) and keeps working (it does delete-old+insert-new); HNSW/CAGRA/IVFPQ are AlwaysAsync and are not blocked.
- bind_update.go (bindUpdate, the modern binder that cannot maintain irregular indexes): also bail to the table-update path when a primary-key column of an irregular-index table is updated, so the guard above runs (a pure PK update was otherwise handled by the modern path and skipped it).

Non-PK updates, plain-table PK updates, and async indexes are unaffected.

Test: fulltext_update_pk.sql — sync fulltext and IVF PK updates are rejected and rows unchanged; non-PK and plain-table PK updates still work. Verified on a live cluster across fulltext/IVF (rejected) and HNSW/master/plain/non-PK (allowed).


XuPeng-SH (approved):
Approved.

I verified the primary-key UPDATE guard end to end on head `c7d9756224`:

- The modern binder falls back to the table-update path for PK updates on irregular-index tables; the final rejection occurs only for synchronously maintained registered plugins.
- Classification is registry-driven through `SyncDescriptor().AlwaysAsync` plus the persisted `async` parameter. Regular and unregistered master indexes remain allowed, while HNSW/CAGRA/IVFPQ and explicit async FULLTEXT/IVFFLAT remain allowed.
- Single and composite PK parts are matched through normalized update-column maps, including aliased targets and multi-table fallback.
- The guard runs before QueryBuilder construction or DML side effects, so rejected statements cannot partially update the base or hidden tables.
- No new per-algorithm dispatch was introduced in the SQL/catalog layer, and the index-plugin review greps are clean.

I found no blocking correctness, performance, or unhappy-path issue. The added planning work is bounded by table/index/PK metadata and has no execution-time impact.

Local verification passed: full `pkg/sql/plan` tests, targeted UPDATE/async-index race tests, FULLTEXT/IVFFLAT/HNSW descriptor tests, build, vet, and gofmt. Current CI, coverage, BVT, and the basic benchmark are green.

PR #25728, opened 2026-07-14
add json_merge_patch, json_merge_preserve

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24738

增加 json_merge_patch, json_merge_preserve


XuPeng-SH (changes requested):
I found two blocking issues on the current head (`24648796487dbec13dbcf9659b16b0517212776d`).

1. **Correctness: `JSON_MERGE_PATCH` does not propagate SQL NULL.**

   In `buildJsonMergePatch`, an SQL NULL argument sets `known = false`, but a later non-object argument sets it back to true. Consequently, cases such as `JSON_MERGE_PATCH(NULL, "[1,2]")` return `[1,2]`, and `JSON_MERGE_PATCH("{\"a\":1}", NULL, "2")` returns `2`.

   MySQL expressions propagate SQL NULL unless a function explicitly documents an exception, and `JSON_MERGE_PATCH` has no such exception. This is distinct from the JSON literal `null`. `JSON_MERGE_PRESERVE` already short-circuits the row correctly.

   Please make PATCH return SQL NULL when any argument is SQL NULL, and replace the UT/BVT expectations that currently describe NULL as an overridable intermediate state. Add first/middle/last SQL NULL cases, including later object, scalar, and array arguments, and distinguish SQL NULL from JSON literal null with `IS NULL` / `JSON_TYPE` assertions.

   References:
   - https://dev.mysql.com/doc/refman/8.4/en/functions.html
   - https://dev.mysql.com/doc/refman/8.0/en/json-modification-functions.html

2. **Performance/unhappy path: both recursive merge implementations rebuild every ancestor object, causing O(depth^2) allocation.**

   Each recursive collision returns a newly serialized `ByteJson`, then `buildJsonObject` copies that entire child again at its parent. A local one-iteration benchmark with two approximately 30 KB, depth-5000 objects measured:

   - `MergePatch`: 94.6 ms and 266,338,960 B/op
   - `MergePreserve`: 95.5 ms and 266,350,160 B/op

   At depth 1000, each already allocates about 10.7 MB/op. MatrixOne currently accepts the depth-5000 inputs, so a small SQL input can create hundreds of MB of transient allocation per row and can OOM a CN across a vector batch.

   Please either use a merge representation/builder that serializes the final tree once (linear in input plus output size), or enforce a documented compatible nesting limit before recursion. MySQL itself limits JSON document nesting to 100, so a compatibility guard is also a viable direction: https://dev.mysql.com/doc/dev/mysql-server/8.4.7/json__syntax__check_8cc.html

   Add deterministic boundary/deep-shape functional coverage. Avoid wall-clock or allocation thresholds in ordinary CI tests; an informational benchmark is appropriate for tracking complexity without making CI flaky.

I also checked the binary-object key-order invariant, function IDs against current `main`, vector/select-list handling, invalid JSON propagation, and object/array/scalar merge rules. Those paths look consistent. Local affected-package tests pass with the repository CGo environment:

`go test -count=1 -timeout 180s ./pkg/container/bytejson ./pkg/sql/plan/function`

pkg/sql/plan/function/func_builtin_json.go
```diff
@@ -1505,6 +1536,127 @@ rowLoop:
...
+		}
+
+		var merged bytejson.ByteJson
+		known := false
+		for j, parameter := range parameters {
+			document, isSQLNull, err := jsonMergeDocument(parameter, i)
+			if err != nil {
+				return err
+			}
+			if isSQLNull {
```
XuPeng-SH, line 1574: This row still revives after an SQL NULL. `known = false` here is not terminal: a later non-object argument hits `j == 0 || document.Type != TpCodeObject`, overwrites `merged`, flips `known` back to true, and the row returns a non-NULL JSON value. A minimal repro on this head is `JSON_MERGE_PATCH(NULL, '[1,2]')`, which currently returns `[1,2]` instead of SQL NULL. Please make SQL NULL sticky for the whole row, not just until the next non-object argument.

pkg/container/bytejson/bytejson_merge.go
```diff
@@ -0,0 +1,172 @@
...
+						return Null, err
+					}
+					keys = append(keys, patchKey)
+					values = append(values, value)
+				}
+				patchIdx++
+			default:
+				patchValue := patch.GetObjectVal(patchIdx)
+				if !isJSONNull(patchValue) {
+					value, err := target.GetObjectVal(targetIdx).MergePatch(patchValue)
```
XuPeng-SH: This recursive collision path still creates quadratic copy volume for deep objects. `MergePatch` returns a fully serialized child subtree here, then the parent `buildJsonObject` copies that entire `value.Data` again into a fresh buffer; the same pattern exists in `mergePreserveObjects`. Because I do not see a nesting-depth guard elsewhere in `bytejson`, a small but deeply nested SQL JSON document can still drive O(depth^2) transient allocation per row.

test/distributed/cases/function/func_json_merge.test
```diff
@@ -0,0 +1,66 @@
...
+select json_merge_patch('1', 'true');
+select json_merge_patch('{"a":1}', '{"a":2}', '{"a":3,"b":4}');
+select json_merge_patch('null', '{"a":1}');
+select json_merge_patch('{"a":["b"]}', '{"a":"c"}');
+select json_merge_patch('{"a":"c"}', '{"a":["b"]}');
+select json_merge_patch('{"a":[{"b":"c"}]}', '{"a":[1]}');
+select json_merge_patch('{"a":"foo"}', 'null') is null, json_type(json_merge_patch('{"a":"foo"}', 'null'));
+select json_merge_patch('[1,2]', '{"a":"b","c":null}');
+
+-- PATCH SQL NULL is an unknown intermediate state that a later non-object patch can replace.
```
XuPeng-SH, line 17: These BVT expectations are asserting the current bug, not the SQL contract. SQL NULL should be terminal for `JSON_MERGE_PATCH`, so cases like `JSON_MERGE_PATCH(NULL, '[1,2]')` and `JSON_MERGE_PATCH('{"a":"b"}', NULL, '2')` should be checked with `IS NULL` / `JSON_TYPE(...)` rather than expecting later arguments to override the NULL.

XuPeng-SH (changes requested):
Re-reviewed the latest head `24648796487dbec13dbcf9659b16b0517212776d` from first principles, including the unhappy-path/resource angle instead of only the previous repros. I still see two blocking issues.

1. **Correctness:** `JSON_MERGE_PATCH` still does not propagate SQL NULL. `buildJsonMergePatch` clears `known` on a NULL argument, but a later non-object argument sets it back to true and overwrites the row. A minimal package-local repro still fails on this head for `JSON_MERGE_PATCH(NULL, '[1,2]')`, and the new UT/BVT expectations currently codify that wrong behavior. Please make SQL NULL terminal for the row, and add first/middle/last SQL NULL cases with later object/scalar/array arguments, distinguishing SQL NULL from JSON literal `null` via `IS NULL` / `JSON_TYPE`.

2. **Performance / unhappy path (Q3):** both recursive merge implementations still rebuild every ancestor object, so deeply nested conflicting objects incur quadratic transient allocation. `mergePatchObjects` / `mergePreserveObjects` recurse into the child, get back a fully serialized `ByteJson`, then `buildJsonObject` copies that entire child again at the parent; `buildJsonObject` copies every non-literal child payload into a new buffer. I also rechecked the package and do not see any nesting-limit guard in `pkg/container/bytejson` or `pkg/container/types`, so this path is still available to SQL input. Please either switch to a builder that serializes once, or enforce a documented depth limit before recursion and add a deterministic deep-shape regression.

The affected package tests still pass locally on this head, which matches the problem: current tests exercise the implemented behavior, not the SQL NULL contract or deep nesting unhappy path.

pkg/sql/plan/function/func_builtin_json.go
daviszhen: 这两个 comment 的前提不对，不能按其建议修改。

  JSON_MERGE_PATCH 不是普通“任一 SQL NULL 都 terminal NULL”的 strict 函数。它遵循 RFC 7396 的 patch 语义：后续非 object patch 的结果就是 patch 自身，不依赖此前 target，因此可以把“未知 target”恢复为已知结果。

  MySQL 8.0.46 源码 Documents/mysql-server-mysql-8.0.46/sql/item_json_func.cc:3431 就是这样实现：

  if (args[i]->null_value) {
    // Cannot return NULL immediately, since a later patch can give
    // a known result.
    null_value = true;
    continue;
  }

  if (patch_wr.type() != enum_json_type::J_OBJECT) {
    *wr = std::move(patch_wr);
    null_value = false;
    continue;
  }

  if (null_value) continue;

  所以 MatrixOne 里的：

  if isSQLNull {
      known = false
      continue
  }
  if j == 0 || document.Type != object {
      merged = document
      known = true
  }

  与 MySQL 的状态机对应，不是“revive bug”。

  MySQL 官方回归测试也明确断言：

  JSON_MERGE_PATCH(NULL, '[1,2,3]')                  -- [1, 2, 3]
  JSON_MERGE_PATCH('{"a":"b"}', NULL, '{"c":"d"}')  -- NULL
  JSON_MERGE_PATCH('{"a":"b"}', NULL, '[1,2,3]',
                   '{"c":null,"d":"e"})             -- {"d": "e}

  见 Documents/mysql-server-mysql-8.0.46/mysql-test/suite/json/r/json_functions_innodb.result:10888。

  因此：

  - NULL + array/scalar/JSON literal null：后续 non-object patch 可确定结果，应返回该 JSON 值。
  - object + NULL + object：后续 object 不能在未知 target 上独立确定结果，应返回 SQL NULL。
  - JSON 'null' 是 JSON literal，不是 SQL NULL。
  - JSON_MERGE_PRESERVE 才是任一 SQL NULL 立即返回 SQL NULL；它没有“non-object patch 覆盖 target”的 RFC 7396 规则。

  建议回复 reviewer：当前 BVT 的 “unknown intermediate state” 描述与 MySQL 源码、官方测试一致，应保留；不应改为 sticky SQL NULL。

test/distributed/cases/function/func_json_merge.test
daviszhen: 这两个 comment 的前提不对，不能按其建议修改。

  JSON_MERGE_PATCH 不是普通“任一 SQL NULL 都 terminal NULL”的 strict 函数。它遵循 RFC 7396 的 patch 语义：后续非 object patch 的结果就是 patch 自身，不依赖此前 target，因此可以把“未知 target”恢复为已知结果。

  MySQL 8.0.46 源码 Documents/mysql-server-mysql-8.0.46/sql/item_json_func.cc:3431 就是这样实现：

  if (args[i]->null_value) {
    // Cannot return NULL immediately, since a later patch can give
    // a known result.
    null_value = true;
    continue;
  }

  if (patch_wr.type() != enum_json_type::J_OBJECT) {
    *wr = std::move(patch_wr);
    null_value = false;
    continue;
  }

  if (null_value) continue;

  所以 MatrixOne 里的：

  if isSQLNull {
      known = false
      continue
  }
  if j == 0 || document.Type != object {
      merged = document
      known = true
  }

  与 MySQL 的状态机对应，不是“revive bug”。

  MySQL 官方回归测试也明确断言：

  JSON_MERGE_PATCH(NULL, '[1,2,3]')                  -- [1, 2, 3]
  JSON_MERGE_PATCH('{"a":"b"}', NULL, '{"c":"d"}')  -- NULL
  JSON_MERGE_PATCH('{"a":"b"}', NULL, '[1,2,3]',
                   '{"c":null,"d":"e"})             -- {"d": "e}

  见 Documents/mysql-server-mysql-8.0.46/mysql-test/suite/json/r/json_functions_innodb.result:10888。

  因此：

  - NULL + array/scalar/JSON literal null：后续 non-object patch 可确定结果，应返回该 JSON 值。
  - object + NULL + object：后续 object 不能在未知 target 上独立确定结果，应返回 SQL NULL。
  - JSON 'null' 是 JSON literal，不是 SQL NULL。
  - JSON_MERGE_PRESERVE 才是任一 SQL NULL 立即返回 SQL NULL；它没有“non-object patch 覆盖 target”的 RFC 7396 规则。

  建议回复 reviewer：当前 BVT 的 “unknown intermediate state” 描述与 MySQL 源码、官方测试一致，应保留；不应改为 sticky SQL NULL。

daviszhen:
> I found two blocking issues on the current head (`24648796487dbec13dbcf9659b16b0517212776d`).
> 
> 1. **Correctness: `JSON_MERGE_PATCH` does not propagate SQL NULL.**
>    In `buildJsonMergePatch`, an SQL NULL argument sets `known = false`, but a later non-object argument sets it back to true. Consequently, cases such as `JSON_MERGE_PATCH(NULL, "[1,2]")` return `[1,2]`, and `JSON_MERGE_PATCH("{\"a\":1}", NULL, "2")` returns `2`.
>    MySQL expressions propagate SQL NULL unless a function explicitly documents an exception, and `JSON_MERGE_PATCH` has no such exception. This is distinct from the JSON literal `null`. `JSON_MERGE_PRESERVE` already short-circuits the row correctly.
>    Please make PATCH return SQL NULL when any argument is SQL NULL, and replace the UT/BVT expectations that currently describe NULL as an overridable intermediate state. Add first/middle/last SQL NULL cases, including later object, scalar, and array arguments, and distinguish SQL NULL from JSON literal null with `IS NULL` / `JSON_TYPE` assertions.
>    References:
>    
>    * https://dev.mysql.com/doc/refman/8.4/en/functions.html
>    * https://dev.mysql.com/doc/refman/8.0/en/json-modification-functions.html
> 2. **Performance/unhappy path: both recursive merge implementations rebuild every ancestor object, causing O(depth^2) allocation.**
>    Each recursive collision returns a newly serialized `ByteJson`, then `buildJsonObject` copies that entire child again at its parent. A local one-iteration benchmark with two approximately 30 KB, depth-5000 objects measured:
>    
>    * `MergePatch`: 94.6 ms and 266,338,960 B/op
>    * `MergePreserve`: 95.5 ms and 266,350,160 B/op
>    
>    At depth 1000, each already allocates about 10.7 MB/op. MatrixOne currently accepts the depth-5000 inputs, so a small SQL input can create hundreds of MB of transient allocation per row and can OOM a CN across a vector batch.
>    Please either use a merge representation/builder that serializes the final tree once (linear in input plus output size), or enforce a documented compatible nesting limit before recursion. MySQL itself limits JSON document nesting to 100, so a compatibility guard is also a viable direction: https://dev.mysql.com/doc/dev/mysql-server/8.4.7/json__syntax__check_8cc.html
>    Add deterministic boundary/deep-shape functional coverage. Avoid wall-clock or allocation thresholds in ordinary CI tests; an informational benchmark is appropriate for tracking complexity without making CI flaky.
> 
> I also checked the binary-object key-order invariant, function IDs against current `main`, vector/select-list handling, invalid JSON propagation, and object/array/scalar merge rules. Those paths look consistent. Local affected-package tests pass with the repository CGo environment:
> 
> `go test -count=1 -timeout 180s ./pkg/container/bytejson ./pkg/sql/plan/function`


  1. SQL NULL 评论不成立，不应修改。
     当前 /mnt/fastdata/temp4/matrixone/pkg/sql/plan/function/func_builtin_json.go:1565 的 known 状态机与 MySQL 8.0.46 源码一致。MySQL 在 SQL NULL 后不会立即返回；后续非对象 patch 会覆盖未知状态并恢复非 NULL，见 Documents/mysql-server-mysql-8.0.46/sql/item_json_func.cc:3445。
     因此：
      - JSON_MERGE_PATCH(NULL, '[1,2]') 返回 [1,2]。
      - JSON_MERGE_PATCH('{"a":"b"}', NULL, '2') 返回 2。
      - 后续仅有对象 patch 时才保持 SQL NULL。
        MatrixOne 的 BVT 注释和结果正确，MySQL 回归结果也明确覆盖这些行为。

  2. 深层递归分配问题已通过兼容性深度限制解决。
     递归序列化复制模式仍存在于 /mnt/fastdata/temp4/matrixone/pkg/container/bytejson/bytejson_merge.go:152 和 /mnt/fastdata/temp4/matrixone/pkg/container/bytejson/bytejson_merge.go:194，但不再能接受无界深度输入：
      - 两个输入在递归前检查。
      - 合并结果也检查，覆盖 PRESERVE 自动增加一层的场景。
      - 最大层数为 100，见 /mnt/fastdata/temp4/matrixone/pkg/container/bytejson/bytejson_merge.go:23 和 /mnt/fastdata/temp4/matrixone/pkg/container/bytejson/bytejson_merge.go:216。
        对应 UT 和 BVT 已覆盖 100 层成功、101 层输入拒绝、PRESERVE 合并后超深拒绝。

  结论：该评论的第 1 条应回复为 MySQL 语义误判；第 2 条已由当前 100 层 MySQL 兼容保护解决。无需再按该评论修改代码。

daviszhen:
> Re-reviewed the latest head `24648796487dbec13dbcf9659b16b0517212776d` from first principles, including the unhappy-path/resource angle instead of only the previous repros. I still see two blocking issues.
> 
> 1. **Correctness:** `JSON_MERGE_PATCH` still does not propagate SQL NULL. `buildJsonMergePatch` clears `known` on a NULL argument, but a later non-object argument sets it back to true and overwrites the row. A minimal package-local repro still fails on this head for `JSON_MERGE_PATCH(NULL, '[1,2]')`, and the new UT/BVT expectations currently codify that wrong behavior. Please make SQL NULL terminal for the row, and add first/middle/last SQL NULL cases with later object/scalar/array arguments, distinguishing SQL NULL from JSON literal `null` via `IS NULL` / `JSON_TYPE`.
> 2. **Performance / unhappy path (Q3):** both recursive merge implementations still rebuild every ancestor object, so deeply nested conflicting objects incur quadratic transient allocation. `mergePatchObjects` / `mergePreserveObjects` recurse into the child, get back a fully serialized `ByteJson`, then `buildJsonObject` copies that entire child again at the parent; `buildJsonObject` copies every non-literal child payload into a new buffer. I also rechecked the package and do not see any nesting-limit guard in `pkg/container/bytejson` or `pkg/container/types`, so this path is still available to SQL input. Please either switch to a builder that serializes once, or enforce a documented depth limit before recursion and add a deterministic deep-shape regression.
> 
> The affected package tests still pass locally on this head, which matches the problem: current tests exercise the implemented behavior, not the SQL NULL contract or deep nesting unhappy path.

如上答复

pkg/container/bytejson/bytejson_merge.go
daviszhen: 当前修改已经加了合并入口保护：

  - 输入对象超过 100 层，在递归前由 /mnt/fastdata/temp4/matrixone/pkg/container/bytejson/bytejson_merge.go:216 拒绝。
  - PRESERVE 合并两个合法对象后可能多出一层，也在合并结果返回前检查。
  - 内部递归调用的是私有 mergePatch / mergePreserve，不会每层重复深度扫描。

  因此，该评论所说的“无界深度导致的 O(depth²) 分配”对当前代码已不成立；剩余成本被限制在最多 100 层。对于超大叶子，仍可能有最多约 100 次的复制，但这属于当前二进制 JSON 序列化模型的有界成本。要彻底消除它，需要改成 MySQL DOM 式可变节点/所有权转移，改动面较大，不应作为这个深度保护修复的前置条
  件。

heni02 (approved):

XuPeng-SH (changes requested):
Correction to my earlier review: the author is right about `JSON_MERGE_PATCH` SQL-NULL semantics. MySQL intentionally allows a later non-object patch to recover a known result, so I withdraw the sticky-NULL finding. The current state machine and its SQL-NULL/JSON-null tests are correct.

Two independent performance/Q3 blockers remain on current head `f6baf202901b0cbd756a01d7484f493757dd23bf`:

1. **[P1] The depth limit does not bound allocation amplification (`pkg/container/bytejson/bytejson_merge.go:112-202`).** Each recursive level serializes and copies the complete merged child again. The new depth-100 check bounds recursion count, but a legal wide leaf is still copied about 100 times. On this head, two valid depth-100 documents with 256 KiB string leaves measured:
   - PATCH: 27,362,096 B/op and succeeds
   - PRESERVE: 53,841,016 B/op, then returns the expected depth-101 error
   Thus a roughly 256 KiB row can allocate tens of MB, and the PRESERVE unhappy path performs the entire expensive construction before rejecting the result. Across a vector batch this remains an OOM/GC-amplification path. Please serialize the final tree once/use a mutable merge representation, or enforce a safe complexity/memory budget derived from both depth and payload size before allocating. Detect result-depth overflow during the merge rather than after constructing the full result. Add an informational deep-wide benchmark and deterministic early-rejection coverage.

2. **[P1] Build string accessors once per parameter (`pkg/sql/plan/function/func_builtin_json.go:1645-1647`).** `jsonMergeDocument` calls `vector.GenerateFunctionStrParameter(parameter)` from the row × argument inner loop. A scalar-only probe (no object merge allocation) with 8,192 rows and two JSON arguments measured exactly 16,384 allocations per execution. Prebuild/reuse the wrappers before the row loop so accessor allocations scale with argument count per batch, not rows × arguments, and add an allocation/benchmark regression.

The functional matrix otherwise closes correctly: RFC 7396 object/scalar/array behavior, PRESERVE autowrap, SQL NULL versus JSON null, invalid JSON/type/arity, select-list materialization, function IDs, and depth boundaries are covered. Full `pkg/container/bytejson` and `pkg/sql/plan/function` tests plus focused race tests pass under the repository CGo wrapper. Q1/Q2 found no retained-resource leak or blocking/wait cycle.

daviszhen:
> Correction to my earlier review: the author is right about `JSON_MERGE_PATCH` SQL-NULL semantics. MySQL intentionally allows a later non-object patch to recover a known result, so I withdraw the sticky-NULL finding. The current state machine and its SQL-NULL/JSON-null tests are correct.
> 
> Two independent performance/Q3 blockers remain on current head `f6baf202901b0cbd756a01d7484f493757dd23bf`:
> 
> 1. **[P1] The depth limit does not bound allocation amplification (`pkg/container/bytejson/bytejson_merge.go:112-202`).** Each recursive level serializes and copies the complete merged child again. The new depth-100 check bounds recursion count, but a legal wide leaf is still copied about 100 times. On this head, two valid depth-100 documents with 256 KiB string leaves measured:
>    
>    * PATCH: 27,362,096 B/op and succeeds
>    * PRESERVE: 53,841,016 B/op, then returns the expected depth-101 error
>      Thus a roughly 256 KiB row can allocate tens of MB, and the PRESERVE unhappy path performs the entire expensive construction before rejecting the result. Across a vector batch this remains an OOM/GC-amplification path. Please serialize the final tree once/use a mutable merge representation, or enforce a safe complexity/memory budget derived from both depth and payload size before allocating. Detect result-depth overflow during the merge rather than after constructing the full result. Add an informational deep-wide benchmark and deterministic early-rejection coverage.
> 2. **[P1] Build string accessors once per parameter (`pkg/sql/plan/function/func_builtin_json.go:1645-1647`).** `jsonMergeDocument` calls `vector.GenerateFunctionStrParameter(parameter)` from the row × argument inner loop. A scalar-only probe (no object merge allocation) with 8,192 rows and two JSON arguments measured exactly 16,384 allocations per execution. Prebuild/reuse the wrappers before the row loop so accessor allocations scale with argument count per batch, not rows × arguments, and add an allocation/benchmark regression.
> 
> The functional matrix otherwise closes correctly: RFC 7396 object/scalar/array behavior, PRESERVE autowrap, SQL NULL versus JSON null, invalid JSON/type/arity, select-list materialization, function IDs, and depth boundaries are covered. Full `pkg/container/bytejson` and `pkg/sql/plan/function` tests plus focused race tests pass under the repository CGo wrapper. Q1/Q2 found no retained-resource leak or blocking/wait cycle.

 本次修改主要解决两个性能问题：

  - 新增 row-scoped mutable JSON merge builder，PATCH/PRESERVE 合并期间不再反复序列化祖先节点，最终只编码一次，并在合并阶段提前拒绝 depth overflow。
  - SQL 参数 wrapper 移到 row loop 外复用，消除 row × argument 分配。

  同时新增：

  - Vector exact-size direct sink，直接写最终 Varlena，并完整处理 Grow2 失败回滚。
  - Binary differential、随机语义、深宽输入、many-argument、allocation、depth 和 race 回归测试。

  效果：256 KiB、depth-100 场景从约 27–54 MB/op 降至约 70–340 KB/op。相关包全量测试和 focused race 均通过。

XuPeng-SH (changes requested):
Re-reviewed current head `6f933aaf51bf99f597fa8366b281a19bba3afe40`. The two previous Q3 blockers are fixed: the mutable plan encodes once, the PRESERVE depth overflow is rejected before output construction, and parameter accessors are now batch-scoped. One correctness blocker remains.

**[P1] Validate every non-NULL JSON argument before the merge state machine can discard it.**

`MergeBuilder.Reset` deliberately skips depth validation (`pkg/container/bytejson/bytejson_merge_plan.go:132-141`), and the SQL loops also bypass `Merge` for a PATCH object while the target is unknown and return early for PRESERVE on SQL NULL (`pkg/sql/plan/function/func_builtin_json.go:1574-1593,1649-1657`). Therefore current head accepts inputs that violate the new 100-level contract:

- `JSON_MERGE_PATCH(depth_101_object, '1')` returns `1`.
- `JSON_MERGE_PATCH(NULL, depth_101_object)` returns SQL NULL.
- `JSON_MERGE_PRESERVE(depth_101_object, NULL)` returns SQL NULL.

MySQL parses each argument through `get_json_wrapper` with `JsonDocumentDefaultDepthHandler` before applying either the PATCH unknown-target skip or the PRESERVE NULL return, so all three cases raise the nesting-depth error. See the upstream parse path and merge loops: https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L315-L349, https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L3133-L3160, https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L3431-L3477.

Please validate every non-SQL-NULL document immediately after parsing, independently of whether it will later be merged, skipped, replaced, or followed by NULL. Add the three regressions above; retaining a separate final-result depth check is still necessary for PRESERVE autowrap.

Local verification otherwise passes: affected package tests, focused `-race -count=10`, and `go vet`. The depth-100/256 KiB benchmark is now about 340 KB/op for PATCH and 71 KB/op for the early PRESERVE error, versus the previous 27-54 MB/op.

daviszhen:
> Re-reviewed current head `6f933aaf51bf99f597fa8366b281a19bba3afe40`. The two previous Q3 blockers are fixed: the mutable plan encodes once, the PRESERVE depth overflow is rejected before output construction, and parameter accessors are now batch-scoped. One correctness blocker remains.
> 
> **[P1] Validate every non-NULL JSON argument before the merge state machine can discard it.**
> 
> `MergeBuilder.Reset` deliberately skips depth validation (`pkg/container/bytejson/bytejson_merge_plan.go:132-141`), and the SQL loops also bypass `Merge` for a PATCH object while the target is unknown and return early for PRESERVE on SQL NULL (`pkg/sql/plan/function/func_builtin_json.go:1574-1593,1649-1657`). Therefore current head accepts inputs that violate the new 100-level contract:
> 
> * `JSON_MERGE_PATCH(depth_101_object, '1')` returns `1`.
> * `JSON_MERGE_PATCH(NULL, depth_101_object)` returns SQL NULL.
> * `JSON_MERGE_PRESERVE(depth_101_object, NULL)` returns SQL NULL.
> 
> MySQL parses each argument through `get_json_wrapper` with `JsonDocumentDefaultDepthHandler` before applying either the PATCH unknown-target skip or the PRESERVE NULL return, so all three cases raise the nesting-depth error. See the upstream parse path and merge loops: https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L315-L349, https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L3133-L3160, https://github.com/mysql/mysql-server/blob/8.0/sql/item_json_func.cc#L3431-L3477.
> 
> Please validate every non-SQL-NULL document immediately after parsing, independently of whether it will later be merged, skipped, replaced, or followed by NULL. Add the three regressions above; retaining a separate final-result depth check is still necessary for PRESERVE autowrap.
> 
> Local verification otherwise passes: affected package tests, focused `-race -count=10`, and `go vet`. The depth-100/256 KiB benchmark is now about 340 KB/op for PATCH and 71 KB/op for the early PRESERVE error, versus the previous 27-54 MB/op.

• 修改记录：

  - 修复所有非 SQL-NULL JSON 参数的深度校验遗漏。
  - Reset 现在校验输入深度，并维护 rootValid 状态。
  - PATCH 对 unknown 状态下被跳过的对象单独校验。
  - PRESERVE 保证参数在 NULL 短路前已通过 Reset/Merge 校验。
  - 保留最终结果深度校验，覆盖 PRESERVE autowrap。
  - 深度遍历改为固定栈，实现零堆分配。
  - 新增三个 SQL 回归用例及 builder/零分配测试。
  - 全量测试、race -count=10、针对性 go vet 均通过。

XuPeng-SH (approved):
Re-reviewed current head 99bb9fa53 after the depth-validation fix. The previous blocker is closed.

- Every non-SQL-NULL input is now depth-validated before Reset, scalar replacement, an unknown PATCH state, or a PRESERVE NULL short-circuit can discard it.
- Reset records the validated root, so Merge does not rescan it; the final materialized result remains validated.
- The fixed 100-frame traversal rejects depth 101 before exceeding its bound and keeps validation allocation-free.
- Q1: builder cleanup and sink rollback retain a single owner. Q2: this path introduces no waits. Q3: nesting, encoded size, and row-scoped memory remain bounded.

Verified locally:
- full tests for bytejson, vector, and SQL plan/function
- focused race tests, 10 repetitions
- go vet for all affected packages
- deep/wide merge benchmark

The added regressions directly cover all three previously reported discard/short-circuit cases. No remaining correctness, lifecycle, or performance blocker found.

Issue #25731, opened 2026-07-14
[Bug]: asynchronous task workers can start after Stop and panic


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

6b2f87c7ff381515d868c2afea1f4061c6998f23

### Other Environment Information

- CI: GitHub Actions, Linux/amd64
- Local reproduction: macOS/arm64 with GOMAXPROCS=1
- Failing CI job: https://github.com/matrixorigin/matrixone/actions/runs/29329920632/job/87075303997?pr=25725
- The failing PR only changes pkg/objectio/ioutil; the failure reproduces on a clean main branch.

### Actual Behavior

TestISCPExecutorStartError intermittently panics after a failed Resume:

    panic: runtime error: invalid memory address or nil pointer dereference
    github.com/matrixorigin/matrixone/pkg/iscp.(*worker).onItem(..., 0x0)
        pkg/iscp/worker.go:124
    created by github.com/matrixorigin/matrixone/pkg/iscp.(*worker).Run

The same lifecycle protocol exists in the publication worker and its subordinate pools. A deterministic single-P test on clean main also panics in filterObjectWorker.Run after an immediate NewWorker/Stop sequence.

Root cause:

1. Constructors publish workers before their asynchronous Run methods register owned goroutines in the WaitGroup.
2. An immediate Stop can observe a zero WaitGroup, close the task channel, and return.
3. The delayed Run method then starts goroutines. Both cancellation and the closed channel are ready; receiving from the closed channel yields a nil task/job, which is dereferenced.
4. Submit checks closed separately from a blocking channel send, so concurrent shutdown can also leave a sender blocked or panic on a closed channel.
5. ISCP marks the executor running and publishes its worker before replay/state initialization succeeds. A failed Resume therefore leaves a false running state and prevents a later Resume from rebuilding.
6. Factory registration errors are ignored, and the ISCP singleton admission flag is not released on exit.

### Expected Behavior

- A worker returned by its constructor is fully sealed: Stop knows every goroutine it owns.
- Stop is exactly-once, cancellation-based, and concurrent Stop callers observe completed cleanup.
- Sender admission has a linearization point: submissions admitted before closure may enqueue and be discarded by cancel-and-drain shutdown; submissions admitted afterward return a closed error.
- No admitted sender can arrive after Stop drains queue references and pending metrics.
[...]

PR #25732, opened 2026-07-14
fix(task): make worker shutdown race-free

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25731

The CI failure in TestISCPExecutorStartError is a pre-existing worker lifecycle race, not a regression from PR #25725. NewWorker returned before Run registered its goroutines in the WaitGroup. An immediate Stop could therefore wait for zero goroutines, close the task channel, and return; the delayed Run then received a nil value from that closed channel and dereferenced it.

This PR fixes the lifecycle protocol instead of guarding the observed nil dereference:

- registers every owned worker goroutine before publishing the worker;
- uses context cancellation instead of closing work queues that have racing senders;
- seals sender admission before cancellation, waits for admitted senders, and drains queued references only after no late sender can arrive;
- makes concurrent Stop calls wait for one shared, exactly-once cleanup;
- defines shutdown as cancel-and-drain: calls admitted before the close linearization point may enqueue successfully, while calls admitted afterward return a closed error; queued work need not execute;
- rejects nil work and rolls back pending counters/gauges for cancellation and drained jobs;
- clears and seals sync-protection state during shutdown;
- publishes an ISCP executor generation only after replay and state initialization succeed;
- binds each run loop to immutable context/worker generation snapshots;
- propagates daemon-task parent cancellation, releases singleton admission, and returns attachment failures;
- applies the same invariants to the publication worker and its subordinate pools, where a deterministic clean-main reproduction found the same panic;
- stops all executor and keepalive tickers through their owning goroutines.

The original failing engine test now verifies that a failed Resume leaves the executor stopped and that the next Resume creates a healthy generation.

Performance validation on Apple M4:

- buffered send: about 14 ns/op;
- cancellation-aware send: about 22 ns/op;
- full admission-gated send: about 28 ns/op;
- all variants: 0 allocations;
- publication worker construction: about 0.10-0.13 ms.

The additional nanoseconds are negligible relative to the SQL/RPC/fileservice work dispatched by these queues.

Test matrix after the final edit and latest-main merge:

- go test ./pkg/iscp ./pkg/publication
- go test -race ./pkg/iscp ./pkg/publication
- go test ./pkg/vm/engine/test -run '^TestISCPExecutorStartError$' -count=20
- go test -race ./pkg/vm/engine/test -run '^TestISCPExecutorStartError$' -count=1
- go test ./pkg/vm/engine/test
- go vet ./pkg/iscp ./pkg/publication
- go build ./pkg/iscp ./pkg/publication

All commands passed with branch head fa20b0d37c based on main 0c21589ddd.

### DIFF
--- a/pkg/iscp/executor.go
+++ b/pkg/iscp/executor.go
@@ -108,6 +108,7 @@ func ISCPTaskExecutorFactory(
 		if err != nil {
 			return
 		}
+		defer exec.terminateLifetime()
 		if err = attachToTask(ctx, task.GetID(), exec); err != nil {
 			return
 		}
@@ -172,10 +173,11 @@ func NewISCPTaskExecutor(
 		)
 	}()
 	option = fillDefaultOption(option)
+	ownerCtx, terminate := context.WithCancel(ctx)
 	exec = &ISCPTaskExecutor{
-		ownerCtx:    ctx,
-		terminated:  make(chan struct{}),
-		ctx:         ctx,
+		ownerCtx:    ownerCtx,
+		terminate:   terminate,
+		ctx:         ownerCtx,
 		packer:      types.NewPacker(),
 		tables:      newISCPTableTree(),
 		cnUUID:      cdUUID,
@@ -238,9 +240,12 @@ func (exec *ISCPTaskExecutor) Pause() error {
 	return nil
 }
 func (exec *ISCPTaskExecutor) Cancel() error {
+	// Broadcast lifetime termination before waiting for the state lock. Start
+	// may hold the lock while doing recovery I/O; its generation context is
+	// derived from ownerCtx and must remain cancelable during that wait.
+	exec.terminateLifetime()
 	exec.runningMu.Lock()
 	defer exec.runningMu.Unlock()
-	exec.terminate()
 	exec.stopLocked()
 	return nil
 }
@@ -274,28 +279,24 @@ func (exec *ISCPTaskExecutor) Start() error {
 }
 
 func (exec *ISCPTaskExecutor) waitForTermination(ctx context.Context) {
+	var ownerDone <-chan struct{}
+	if exec.ownerCtx != nil {
+		ownerDone = exec.ownerCtx.Done()
+	}
 	select {
 	case <-ctx.Done():
-	case <-exec.terminated:
+	case <-ownerDone:
 	}
 }
 
-func (exec *ISCPTaskExecutor) terminate() {
-	if exec.terminated != nil {
-		exec.terminateOnce.Do(func() { close(exec.terminated) })
+func (exec *ISCPTaskExecutor) terminateLifetime() {
+	if exec.terminate != nil {
+		exec.terminate()
 	}
 }
 
 func (exec *ISCPTaskExecutor) isTerminated() bool {
-	if exec.terminated == nil {
-		return false
-	}
-	select {
-	case <-exec.terminated:
-		return true
-	default:
-		return false
-	}
+	return exec.ownerCtx != nil && exec.ownerCtx.Err() != nil
 }
 
 func (exec *ISCPTaskExecutor) initStateLocked(parent context.Context) (err error) {
@@ -562,6 +563,21 @@ func (exec *ISCPTaskExecutor) GetWatermark(accountID uint32, srcTableID uint64, 
 	}
 	watermark, ok = table.GetWatermark(jobName)
 	return
+}
+
+// GetJobState returns the scheduler's in-memory LSN and state. It is used by
+// diagnostics and lifecycle tests to distinguish admitted work from durable
+// progress.
+func (exec *ISCPTaskExecutor) GetJobState(
+	accountID uint32,
+	srcTableID uint64,
+	jobName string,
+) (lsn uint64, state int8, ok bool) {
+	table, ok := exec.getTable(accountID, srcTableID)
+	if !ok {
+		return 0, ISCPJobState_Invalid, false
+	}
+	return table.getJobState(jobName)
 }
 
 // For UT
--- a/pkg/iscp/executor.go
+++ b/pkg/iscp/executor.go
@@ -16,10 +16,10 @@ package iscp
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"sync/atomic"
-
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/zap"
@@ -73,6 +73,10 @@ type ISCPExecutorOption struct {
 	FlushWatermarkInterval time.Duration
 	FlushWatermarkTTL      time.Duration
 	RetryTimes             int
+}
+
+func newISCPTableTree() *btree.BTreeG[*TableEntry] {
+	return btree.NewBTreeGOptions(tableInfoLess, btree.Options{NoLocks: true})
 }
 
 func ISCPTaskExecutorFactory(
@@ -173,7 +177,7 @@ func NewISCPTaskExecutor(
 		terminated:  make(chan struct{}),
 		ctx:         ctx,
 		packer:      types.NewPacker(),
-		tables:      btree.NewBTreeGOptions(tableInfoLess, btree.Options{NoLocks: true}),
+		tables:      newISCPTableTree(),
 		cnUUID:      cdUUID,
 		txnEngine:   txnEngine,
 		cnTxnClient: cnTxnClient,
@@ -305,10 +309,14 @@ func (exec *ISCPTaskExecutor) initStateLocked(parent context.Context) (err error
 		}
 	}()
 
+	// Repair abandoned durable states before loading the next generation.
+	// A stopped generation may have persisted Running, or may only have
+	// advanced its in-memory LSN before cancellation. The database is the
+	// generation boundary's source of truth.
 	err = retry(
 		ctx,
 		func() error {
-			return exec.replay(ctx)
+			return exec.repairAbandonedJobs(ctx)
 		},
 		exec.option.RetryTimes,
 		DefaultRetryInterval,
@@ -320,7 +328,7 @@ func (exec *ISCPTaskExecutor) initStateLocked(parent context.Context) (err error
 	err = retry(
 		ctx,
 		func() error {
-			return exec.initState(ctx)
+			return exec.rebuildState(ctx)
 		},
 		exec.option.RetryTimes,
 		DefaultRetryInterval,
@@ -362,7 +370,7 @@ func (exec *ISCPTaskExecutor) stopLocked() {
 	exec.worker = nil
 }
 
-func (exec *ISCPTaskExecutor) initState(ctx context.Context) (err error) {
+func (exec *ISCPTaskExecutor) repairAbandonedJobs(ctx context.Context) (err error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, time.Minute*5)
 	defer cancel()
 	sql := fmt.Sprintf(
@@ -382,7 +390,13 @@ func (exec *ISCPTaskExecutor) initState(ctx context.Context) (err error) {
 		0)
 	txnOp, err := exec.cnTxnClient.New(ctxWithTimeout, nowTs, createByOpt)
 	if txnOp != nil {
-		defer txnOp.Commit(ctxWithTimeout)
+		defer func() {
+			if err != nil {
+				err = errors.Join(err, txnOp.Rollback(ctxWithTimeout))
+				return
+			}
+			err = txnOp.Commit(ctxWithTimeout)
+		}()
 	}
 	if err != nil {
 		return
@@ -482,35 +496,19 @@ func (exec *ISCPTaskExecutor) run(ctx context.Context, worker Worker) {
 				} else {
 					_, ok = tables[iter.tableID]
 				}
-				table, ok2 := exec.getTable(iter.accountID, iter.tableID)
+				table, tableExists := exec.getTable(iter.accountID, iter.tableID)
+				if !tableExists {
+					logutil.Error(
+						"ISCP-Task get table failed",
+						zap.Uint32("accountID", iter.accountID),
+						zap.Uint64("tableID", iter.tableID),
+					)
+					continue
+				}
 				if ok {
-					// The update on mo_iscp_log may not be available in the next applyISCPLog,
-					// so update the in-memory state directly to prevent repeated triggering of the iteration.
-					for i, jobName := range iter.jobNames {
-						job := table.jobs[JobKey{
-							JobName: jobName,
-							JobID:   iter.jobIDs[i],
-						}]
-						job.currentLSN++
-						job.state = ISCPJobState_Pending
-					}
-					onErrorFn := func(err error) {
-						for i, jobName := range iter.jobNames {
-							job := table.jobs[JobKey{
-								JobName: jobName,
-								JobID:   iter.jobIDs[i],
-							}]
-							job.currentLSN--
-							job.state = ISCPJobState_Completed
-						}
-						logutil.Error(
-							"ISCP-Task submit iteration failed",
-							zap.Error(err),
-						)
-					}
 					ok, err := CheckLeaseWithRetry(ctx, exec.cnUUID, exec.txnEngine, exec.cnTxnClient)
 					if err != nil {
-						onErrorFn(err)
+						logutil.Error("ISCP-Task check lease failed", zap.Error(err))
 						continue
 					}
 					if !ok {
@@ -519,18 +517,19 @@ func (exec *ISCPTaskExecutor) run(ctx context.Context, worker Worker) {
 					}
 					err = worker.Submit(iter)
 					if err != nil {
-						onErrorFn(err)
+						logutil.Error("ISCP-Task submit iteration failed", zap.Error(err))
 						continue
+					}
+					// Admission is the linearization point. The worker now owns the
+					// iteration, so hide it from the next candidate scan.
+					if err = table.markIterationPending(iter); err != nil {
+						logutil.Error("ISCP-Task mark admitted iteration failed", zap.Error(err))
+						// The accepted iteration cannot be retracted safely. Cancel this
+						// generation; the next Start repairs and rebuilds its state.
+						go exec.Cancel()
+						return
 					}
 				} else {
-					if !ok2 {
-						logutil.Error(
-							"ISCP-Task get table failed",
-							zap.Uint32("accountID", iter.accountID),
-							zap.Uint64("tableID", iter.tableID),
-						)
-						continue
-					}
 					table.UpdateWatermark(iter)
 				}
 			}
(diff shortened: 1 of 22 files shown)


pkg/iscp/executor.go
```diff
@@ -103,19 +104,22 @@ func ISCPTaskExecutorFactory(
...
 		}
-		err = exec.initStateLocked()
+		err = exec.initStateLocked(ctx)
+		exec.runningMu.Unlock()
 		if err != nil {
 			return
 		}
-		exec.run(ctx)
-		<-ctx.Done()
+		exec.run(exec.ctx, exec.worker)
```
LeftHandCold: [P1] Keep the daemon-task executor alive across Pause/Resume. Pause calls exec.Stop(), which cancels exec.ctx; this synchronous run therefore returns, and the factory now returns nil immediately afterward. taskservice.startTask removes every non-CDC daemonTask when its executor returns (ISCP and Publication have their own task types). A later Resume creates a fresh daemonTask with no activeRoutine; resumeTask writes TaskStatus_Running first and then fails with active routine not set, leaving the task reported as running with no executor. The old parent-context wait kept the attached routine alive across a paused generation. Please supervise generations until the daemon parent ctx is canceled, while still stopping the current generation on parent cancellation, and add taskservice-level Pause-to-Resume regressions for both ISCP and Publication. The same return path exists at pkg/publication/executor.go:138.

pkg/publication/worker.go
```diff
@@ -817,66 +948,91 @@ func (w *getChunkWorker) SubmitGetChunk(job Job) error {
 	default:
 		globalJobStats.IncrementGetChunkPending()
 	}
-	w.jobChan <- job
-	return nil
+	select {
+	case w.jobChan <- job:
+		return nil
+	case <-doneChan(w.ctx):
```
LeftHandCold, line 987: [P1] Propagate rejected admission before waiting on the job. This branch returns an error without executing the job, but production callers ignore SubmitGetChunk and immediately call WaitDone (for example filter_object.go:494-500; the chunk path at 438-449 does the same). WaitDone is a bare receive from the result channel, so no context can release it when admission was rejected. On this head I reproduced it with a real GetChunkWorker: Stop it, call getMetaWithRetry with a 20 ms context, and the call was still blocked in WaitDone after 250 ms. SubmitFilterObject and SubmitWriteObject have the same ignored-error pattern. Please return the admission error or complete the job with cancellation at every caller, and add stopped/canceled-worker regressions through the real helper paths.

LeftHandCold (changes requested):
Requesting changes for two confirmed P1 lifecycle gaps on the latest head fa20b0d37c. First, pausing either daemon executor now lets its factory return and causes taskservice to drop the attached active routine, so a later Resume marks the task Running and then fails with no executor. Second, the new cancellation-aware worker admission errors are ignored by publication job callers, which then block forever in WaitDone on jobs that were never executed. I reproduced the second path deterministically with a real stopped GetChunkWorker. The normal, race, engine regression, vet, and build checks pass because they do not cover these termination paths.

pkg/iscp/executor.go
XuPeng-SH: Fixed in 12644b7fee. The daemon lifetime is now distinct from each pausable execution generation: the factory remains attached while Pause stops only the current generation, and it returns only on owner-context cancellation, explicit Cancel, or fatal lease loss. Cancel/Start/Restart are linearized by the executor state mutex and canceled executors cannot reopen. Added lifetime regressions for both executors and taskservice PauseRequested -> Paused -> ResumeRequested -> Running coverage for both TaskCode_ISCPExecutor and TaskCode_PublicationExecutor.

pkg/publication/worker.go
XuPeng-SH: Fixed in 12644b7fee. All production SubmitFilterObject/SubmitGetChunk/SubmitWriteObject callers now propagate admission errors before WaitDone. In addition, rejected and drained jobs are completed with an error through an exactly-once terminal gate, so future callers cannot recreate the hang by forgetting an error check. Stop now closes worker pools in reverse dependency order (get-chunk/write -> filter -> top-level iteration), removing the wait-for inversion. Added real stopped-worker helper regression plus deterministic rejected/drained coverage for all three job worker classes under race.

pkg/iscp/worker.go
```diff
@@ -164,8 +211,16 @@ func (w *worker) onItem(iterCtx *IterationContext) {
...
-	w.cancel()
-	w.wg.Wait()
-	close(w.taskChan)
+	w.stopOnce.Do(func() {
+		w.closed.Store(true)
+		w.admissions.seal(w.cancel)
+		w.wg.Wait()
+		for {
+			select {
+			case <-w.taskChan:
```
XuPeng-SH, line 222: This drain path is safe for publication because those jobs have terminal result channels, but it is not safe for ISCP. By the time an iteration reaches `taskChan`, `executor.run()` has already advanced each job's `currentLSN` and moved it to `Pending`. If Stop() drops that queued iteration here, nothing restores the optimistic advance, and `replay()` cannot repair it because `JobEntry.update()` ignores lower persisted LSN/state from the DB. After a pause/resume, the job can stay stuck ahead of storage and never become a candidate again. Please either let accepted ISCP iterations run to a terminal flush, or feed drained iterations back through a rollback path that restores the pre-submit LSN/state before Stop() returns.

XuPeng-SH (commented):
GitHub does not allow me to send REQUEST_CHANGES on my own pull request, but this is still a request-changes level blocker on the latest head. The earlier daemon-lifetime and publication job-result issues look fixed. The remaining problem is in ISCP: the new cancel-and-drain shutdown path can permanently lose an accepted-but-not-started iteration because ISCP advances currentLSN/state before enqueue, worker.Stop() now discards queued iterations, and replay only applies monotonic DB advances. I reproduced this with a package-local repro where an optimistic 6/pending state stayed stuck after replaying the persisted 5/completed row. Please either stop discarding accepted ISCP iterations, or roll back every drained iteration before returning; clearing/rebuilding the in-memory table state on replay would also make recovery robust.

pkg/iscp/worker.go
XuPeng-SH: Confirmed and fixed in 686669cbae. The drain-only rollback suggested here would still miss iterations that had started and were canceled before reaching a durable terminal state, so the fix closes the generation boundary instead: (1) worker admission is now the linearization point and in-memory Pending/LSN advances happen only after Submit succeeds; (2) Stop still cancels and drains for bounded shutdown; (3) before publishing the next generation, Start repairs durable Pending/Running rows, rebuilds a fresh in-memory tree from storage, and restores currentLSN from persisted JobStatus; and (4) recovery normalizes a pre-repair snapshot so correctness does not depend on asynchronous logtail visibility. The analogous top-level publication TaskContext path is fixed with the same repair-and-rebuild rule, while live replay preserves admitted in-flight state. Transaction commit/rollback errors are propagated and engine-registration failures now roll back the created operator. Added regressions for abandoned-generation replacement, persisted LSN restoration, atomic shared-iteration admission, stale pre-repair snapshots, and publication live-replay monotonicity. Validation: full pkg/iscp + pkg/publication tests, full race tests, focused tests x100, go vet, and TestISCPExecutorStartError x10.

XuPeng-SH:
Follow-up hardening is pushed in 75a55d4134.

This replaces the weak replay smoke test with a deterministic engine-level recovery regression. The test uses the real catalog, transaction engine, scheduler, and worker to create the exact generation-boundary divergence: an iteration is accepted and visible as in-memory Pending at LSN 1 while durable state remains Completed at LSN 0; Stop cancels/drains that generation; Resume must rebuild from storage, retry the iteration, advance the watermark, and persist Completed at LSN 1. It uses state-based synchronization and passed 10 consecutive runs plus race.

The full lifecycle audit also found that Cancel previously waited for runningMu before signaling termination, so a Start/Resume blocked in repair/rebuild I/O could delay cancellation. Both ISCP and Publication now derive generation contexts from a separately cancelable lifetime context. Cancel broadcasts lifetime termination before joining under the state lock, and factory error/exit paths always release that context. Added deterministic lock-contention regressions for both executors (50x under race).

Validation on the final head:
- full pkg/iscp + pkg/publication under race
- engine-level accepted-iteration Stop/Resume recovery x10
- engine-level recovery under race
- taskservice Pause -> Resume lifecycle x10 under race
- lifecycle cancellation regressions x50 under race
- go vet for pkg/iscp, pkg/publication, and pkg/taskservice
- git diff --check

XuPeng-SH (commented):
Broader latest-head re-review on `ff97a144b4` (not just the prior thread):

1. Publication executor/worker shutdown paths now satisfy the unhappy-path terminal chain. Submission rejection completes jobs, Stop drains/fails leaf queues, and replay overwrites in-memory task state from `mo_ccpr_log`, so I did not confirm a Q1/Q2/Q3 blocker there.
2. ISCP still has one request-changes-level Q1 break. The executor advances `job.currentLSN`/`job.state=Pending` before enqueue (`pkg/iscp/executor.go`), `worker.Stop()` now intentionally discards queued iterations after cancel (`pkg/iscp/worker.go`), `replay()` cannot repair that dropped work because `JobEntry.update()` only accepts monotonic DB progress (`pkg/iscp/consumer_entry.go`), and `getCandidate()` only schedules `Completed` jobs (`pkg/iscp/table_entry.go`). A pause/resume can therefore strand a job at the higher in-memory LSN/state even though storage still says the previous completed LSN.

I reproduced this on the latest head with a package-local repro, so after the broader unhappy-path audit I still land on request-changes for the ISCP path.

XuPeng-SH:
The review at https://github.com/matrixorigin/matrixone/pull/25732#pullrequestreview-4699940019 is stale and superseded: it explicitly reviews `ff97a144b4`, while the current head is `75a55d4134` and contains the generation-boundary fix from `686669cbae`.

The concrete failure chain described by that review no longer exists on the current head:

1. The scheduler now calls `worker.Submit(iter)` first and advances `currentLSN` / marks `Pending` only after admission succeeds (`pkg/iscp/executor.go`). Rejected work therefore remains a candidate.
2. More importantly, correctness does not depend on rolling back only queued work. After Stop has canceled and joined the old worker generation, the next Start repairs durable Pending/Running rows and calls `rebuildState`, which constructs a fresh tree from `mo_iscp_log` and replaces the abandoned in-memory tree. The monotonic live `replay()` path cited by the review is intentionally not the generation-recovery path.
3. `TestISCPResumeRecoversAcceptedIteration` now reproduces the exact reported window with the real engine: memory reaches Pending/LSN 1 while storage remains Completed/LSN 0; Stop cancels/drains; Resume rebuilds, retries, advances the watermark, and persists Completed/LSN 1. It passed 10 consecutive runs and race on the current head.

No additional code change is appropriate for this review because its finding was valid for `ff97a144b4` and is already fixed and behavior-tested on the current head.

pkg/publication/executor.go
```diff
@@ -691,10 +750,86 @@ func (exec *PublicationTaskExecutor) replay(ctx context.Context) (err error) {
...
+	taskID string,
+	lsn uint64,
+	state int8,
+	subscriptionState int8,
+	dropAt *time.Time,
+) error {
+	current, ok := exec.getTask(taskID)
+	if ok &&
+		(current.State == IterationStatePending || current.State == IterationStateRunning) &&
+		lsn <= current.LSN {
```
gouhongshen: P1: do not mask an equal-LSN terminal transition. This branch also matches a durable terminal row when lsn == current.LSN. For example, after admission the in-memory entry can be Pending/LSN=6; the worker can then persist a retryable Completed or non-retryable Error at LSN 6 before a stale-read replay runs. loadReplay reads that row, but this condition keeps the entry Pending. replay then advances ccprLogWm to the replay transaction snapshot, so the LSN-6 terminal update will not be delivered by the next logtail pass; the task can remain permanently stuck until restart. Preserve only strictly older snapshots (or only Pending/Running rows at equal LSN), and apply equal-LSN Completed/Error/Canceled states. Please add a regression test for equal-LSN terminal replay.

gouhongshen (changes requested):
Requesting changes on the current head 75a55d413441fae0f49aab5da7f7ff9ea64e000a.

Deep review covered lifecycle owner and execution contexts, Start/Pause/Resume/Cancel/Restart serialization, generation rebuild and durable-state repair, worker admission gates and cancel-and-drain behavior, ISCP admission and recovery, Publication replay and error propagation, API compatibility, and the changed tests. Current CI is green, including Ubuntu UT, arm64 SCA, e2e suites, CodeQL, and coverage.

The inline P1 is a correctness blocker: equal-LSN terminal Publication states can be hidden by mergeReplayedTask, and the replay watermark then makes the missed terminal transition unrecoverable without a process restart.

Non-blocking compatibility note: this PR changes exported APIs (publication.Job adds Fail(error), publication.NewWorker adds an argument, and exported ISCP constructors or methods add jobStatus parameters). Existing external implementers or callers will stop compiling. If these pkg APIs are intentionally not compatibility-stable, please document that; otherwise use additive interfaces or wrappers.

I did not find another confirmed correctness blocker in the current head. The earlier daemon-lifetime, ignored-publication-submit, and ISCP accepted-but-not-started findings are addressed by the current implementation and tests.

pkg/publication/filter_object.go
```diff
@@ -438,7 +440,9 @@ var GetObjectFromUpstreamWithWorker = func(
 		chunkJob := NewGetChunkJob(ctx, upstreamExecutor, objectName, i, subscriptionAccountName, pubName)
 		chunkJobs[i-1] = chunkJob
 		if getChunkWorker != nil {
-			getChunkWorker.SubmitGetChunk(chunkJob)
+			if err := getChunkWorker.SubmitGetChunk(chunkJob); err != nil {
+				return nil, err
```
LeftHandCold, line 450: [P1] Join the already accepted chunk jobs before returning here. If shutdown seals this worker between two submissions, chunks 1..k-1 can still be executing while chunk k is rejected. This function then unwinds its FilterObjectJob, and ExecuteIteration can close the shared upstream executor/transaction even though those earlier chunk jobs still use it. The worker eventually drains them, but that happens concurrently with the caller's cleanup. Please cancel the batch and wait for every successfully submitted job to publish its terminal result (or submit/wait sequentially) before returning; add a regression where an accepted prefix is followed by a rejected submission.

pkg/publication/filter_object_submit.go
```diff
@@ -455,7 +455,9 @@ func ApplyObjects(
 			// Data objects don't need aobjectMap for rewriting, pass nil
 			filterJob := NewFilterObjectJob(ctx, statsBytes, currentTS, upstreamExecutor, false, fs, mp, getChunkWorker, writeObjectWorker, subscriptionAccountName, pubName, ccprCache, txnID, nil, ttlChecker)
 			if filterObjectWorker != nil {
-				filterObjectWorker.SubmitFilterObject(filterJob)
+				if err = filterObjectWorker.SubmitFilterObject(filterJob); err != nil {
+					return
```
LeftHandCold, line 467: [P1] This has the same partial-admission lifetime hole at the filter layer. If several FilterObjectJobs were accepted and a later SubmitFilterObject is rejected during Stop, returning immediately skips Phase 2 and leaves the accepted jobs running. ExecuteIteration's defer then calls IterationContext.Close(), closing LocalTxn and UpstreamExecutor while those jobs may still access them or their leaf workers. Please cancel/drain and join every previously accepted filter job before unwinding, and cover partial success followed by shutdown (for both data and tombstone phases).

LeftHandCold (changes requested):
The previous two P1s are fixed, and the new generation-recovery tests pass. Two shutdown lifecycle gaps remain in partial-batch admission: completing the newly rejected job is not enough; jobs accepted earlier in the same batch must reach terminal state before iteration-owned resources are closed. Please join/drain the accepted prefix and add partial-success-then-shutdown regressions.

pkg/publication/executor.go
XuPeng-SH: Fixed in bb26e16463. mergeReplayedTask now preserves local in-flight ownership only for a strictly older replay LSN, or for the same LSN when the durable row is still Pending/Running. Equal-LSN Completed/Error/Canceled states now win before the replay watermark advances. The regression covers both local Pending/Running states crossed with all three durable terminal states, plus older, equal-in-flight, and newer-LSN behavior.

pkg/publication/filter_object.go
XuPeng-SH: Fixed in bb26e16463 with a batch ownership protocol rather than a submit-site special case. acceptedJobBatch records every successful admission; chunk jobs use a child context; every exit cancels the batch and joins every unconsumed accepted job before the shared executor can close. Deterministic channel-gated tests cover accepted-prefix-then-reject and first-result-failure followed by retry rejection while a sibling is still running. Both pass under -race -count=100.

pkg/publication/filter_object_submit.go
XuPeng-SH: Fixed in bb26e16463 with the same accepted-job ownership protocol in ApplyObjects. Both data and tombstone phases add jobs only after successful admission, consume each terminal result exactly once, and cancel+join every remaining accepted job on submission failure, result failure, context cancellation, or any other return before iteration-owned transactions/executors/workers can close. Deterministic tests cover partial admission and early result failure for both phases under race.

XuPeng-SH:
Addressed the latest review in `bb26e16463`, rebased onto current main `01a9085a48`.

The three P1s are fixed by two explicit invariants:

1. A replay may preserve local in-flight state only when the durable snapshot is strictly older, or equal-LSN and still in-flight; equal-LSN terminal state is authoritative.
2. Successful async admission transfers ownership to a batch; every exit cancels and joins all accepted children before iteration-owned resources are released. This applies to chunk, data-filter, and tombstone-filter phases, including both partial submission and early result failure.

I also addressed the non-blocking compatibility note where it was real: `Job` remains source-compatible and the shutdown requirement is an additive `TerminalJob` admission contract; the original `NewJobEntry` signature remains and `NewJobEntryWithStatus` carries durable recovery state. `publication.NewWorker` already matches current main and needed no compatibility shim.

Validation on the final rebased tree:
- full `pkg/publication` + `pkg/iscp` normal tests
- full `pkg/publication` + `pkg/iscp` race tests
- new lifecycle/state regressions under `-race -count=100`
- `TestPauseResumeDaemonTask` under `-race -count=20`
- real-engine `TestISCPExecutorStartError` with `-count=10`
- `go vet` and `go build` for both affected packages

All passed.

aptend (approved):
Re-reviewed current head bb26e16463. The previously reported lifecycle, equal-LSN replay, and partial-batch ownership issues are addressed. ISCP/publication normal and race tests, focused lifecycle stress tests, taskservice pause/resume, vet, build, and diff checks passed. No remaining correctness blocker found.

LeftHandCold (approved):
Re-reviewed the latest head. The acceptedJobBatch ownership protocol closes both partial-admission P1s: every successfully admitted chunk/filter job is canceled and joined on all early-return paths before iteration-owned resources close, and worker admission now guarantees a terminal Fail path. Equal-LSN terminal replay and ISCP persisted-LSN compatibility also look correct. Local package tests, full race tests, build/vet, the deterministic race regressions (-count=100), and the engine recovery test pass.

gouhongshen (approved):

Issue #25711, opened 2026-07-14
[Bug]: DROP HNSW INDEX can deadlock with an in-flight ISCP update


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

`fix/25211-per-cn-colexec-server` (based on `main`; the affected HNSW/ISCP code is also present on `main`)

### Commit ID

`66a9db25f16b4961319cf96bd8098511198bf2d0`

### Other Environment Information

```Markdown
- Hardware parameters: GitHub-hosted ubuntu-22.04 x64 runner
- OS type: Linux/x64
- Others: Standalone launch mode, pessimistic transactions, BVT job 86991930850
```

### Actual Behavior

The standalone pessimistic BVT intermittently fails in `test/distributed/cases/pessimistic_transaction/vector/vector_hnsw.sql` when dropping an HNSW index immediately after inserts:

```sql
alter table vector_index_09 drop index idx01;
```

The statement returns:

```text
deadlock detected
```

The following `SHOW CREATE TABLE` assertion also fails because the failed DDL is rolled back and `idx01` remains present. This is a secondary failure, not a separate root cause.

The lockservice log records a real two-transaction cycle:

```text
70f70f4866a9b62a18c20a4f89053227
  <= 70f70f4866a9b62a18c20a4f89053216
  <= 70f70f4866a9b62a18c20a4f89053227
```

The participants are:

- `...53227`: an ISCP HNSW consumer transaction saving an update for `vector_index_09.idx01`.
- `...53216`: the foreground `ALTER TABLE ... DROP INDEX` transaction, failing while dropping the HNSW metadata hidden table.

The observed timeline is:

1. `03:45:43.092`: the ISCP job for table ID `290263`, job `index_idx01`, is registered.
2. `03:45:43.265`: ISCP consumes three inserted rows.
3. `03:45:43.281`: transaction `...53227` executes a DELETE against the HNSW storage hidden table (`290267`) and then enters lock wait.
4. `03:45:43.287`: transaction `...53216` enters lock wait while dropping the HNSW metadata hidden table (`290266`).
5. `03:45:49.298`: lockservice detects the cycle and both the ISCP sink and foreground DDL report `deadlock detected`.

The lifecycle race is also visible in the code:

- `Scope.DropIndex` deletes the hidden index tables before calling `DropIndexCdcTask`, so an already-running ISCP consumer is not quiesced before destructive DDL:
[...]

PR #25733, opened 2026-07-15
Fix ISCP index consumer cancellation during DROP INDEX

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25711

 This PR fixes DROP INDEX and other destructive ISCP-backed index DDL paths by
  canceling and draining the active index consumer before index metadata and hidden
  tables are removed.

  Changes:

  - Add ISCP runtime tracking for running job consumers.

  - Fence dropped index job generations so new iterations skip them while DDL is
    waiting for quiescence.

  - Cancel the consumer context and data retriever, then wait for the running
    consumer to exit.

  - Route DROP INDEX, ALTER TABLE DROP INDEX, DROP TABLE, and DROP DATABASE
    through the ISCP drain path before deleting metadata or hidden tables.

  - Use the existing global ISCP daemon task runner UUID from sys_daemon_task as
    the single ISCP executor owner.

  - Add a queryservice-based cross-CN drain request so a DDL CN can ask the actual
    ISCP runner CN to cancel and drain the local executor.

  - Keep the local executor fast path for single-CN and same-runner cases.

  - Fail closed when quiescence cannot be confirmed, including missing query client,
    missing runner query address, or remote drain RPC failure.

  - Clean remote generation fences through transaction commit/rollback callbacks.

  - Make data retrievers reject/cleanup queued data after cancellation.

  - Add ctx-aware fault injection for WAIT and SLEEP.

  - Fix WAIT fault removal so blocked waiters are notified, including the
    TriggerFaultWithContext path.

  - Add ISCP cancel failpoints for different consumer phases, including
    per-index-name injection.

  - Add unit tests for runtime cancel, retriever cleanup, fault cancellation,
    local drain, remote queryservice drain, remote fence cleanup, and DDL drain
    behavior.

  Test Matrix:

  | Area | Scenario | Result |
  | --- | --- | --- |
  | Original reproduction | Reproduced the documented ISCP DROP INDEX cancel issue. |
  Fixed: DROP INDEX no longer leaves a background consumer running. |
  | Single-process multi-CN | Started local multi-CN launch and issued DROP INDEX from
  different CNs. | Passed for both directions. |
  | Docker multi-process CN | Started `make dev-up` with separate `mo-cn1` and `mo-cn2`
  containers. ISCP runner was on `mo-cn2`; DDL was issued from non-runner `mo-cn1`. |
  Passed. |
  | Cross-CN DROP INDEX | Created async HNSW index from CN1 and dropped it from CN1 while
  runner was CN2. | DROP INDEX succeeded through queryservice remote drain. |
  | Cross-CN DROP TABLE | Created ISCP-backed HNSW table and dropped table from non-
  runner CN. | DROP TABLE succeeded. |
  | Cross-CN ALTER TABLE DROP INDEX | Created ISCP-backed HNSW table and ran ALTER TABLE
  DROP INDEX from non-runner CN. | ALTER succeeded and index metadata was removed. |
  | Cross-CN DROP DATABASE | Created ISCP-backed tables and dropped database from non-
  runner CN. | DROP DATABASE succeeded. |
  | In-flight remote consumer | Injected `fj/iscp/cancel/after-register-consumer` WAIT on
  the runner CN, then issued DROP INDEX from another CN. | DROP waited before fault
  removal, then completed after the runner consumer exited. |
  | Cancel after consumer registration | `fj/iscp/cancel/after-register-consumer`. | DROP
  waits for drain, then succeeds. |
  | Cancel after worker submit | `fj/iscp/cancel/after-submit`. | Fenced job is skipped,
  DROP succeeds. |
  | Cancel during fanout | `fj/iscp/cancel/fanout-before-send`. | Canceled retriever
  rejects data, DROP succeeds. |
  | Cancel before HNSW update | `fj/iscp/cancel/long/hnsw-before-update`. | DROP returns
  quickly. |
  | Cancel before HNSW save | `fj/iscp/cancel/long/hnsw-before-save`. | DROP returns
  quickly. |
  | Cancel before watermark update | `fj/iscp/cancel/long/before-watermark`. | DROP
  returns quickly. |
  | Cancel before SQL execution | `fj/iscp/cancel/long/before-exec` with fulltext ASYNC.
  | DROP returns quickly. |
  | Long wait via sleep | `fj/iscp/cancel/long/before-send:idx01` with sleep 30. | DROP
  returns quickly after canceling the consumer context. |
  | Iteration already failed before cancel | Injected iteration error before DROP. | Job
  records the error path correctly; DROP still succeeds and sets `drop_at`. |
  | Multiple jobs in one iteration | Two HNSW indexes on different columns, cancel only
  `idx01`. | `idx01` removed; `idx02` preserved. |
  | Fault cleanup | Listed fault points after tests. | No remaining fault points. |
  | Goroutine cleanup | Checked pprof goroutines after fault/ISCP tests. | No stuck
  cancel/scan/ISCP goroutines. |
  | Unit tests | `go test -count=1 ./pkg/iscp ./pkg/sql/compile ./pkg/cnservice ./pkg/
  queryservice/client` with local CGo flags. | Passed. |
  | Build/package graph | `make build`. | Passed, including Go package graph check. |
  | Static check | `git diff --check`. | Passed. |


XuPeng-SH (changes requested):
The single-CN happy path is covered, but the cancellation invariant is not established for the actual lifecycle. Requesting changes for the following blockers:

1. **[High][correctness] Drain is process-local and silently succeeds on the wrong CN.** `iscpExecutors` is a process-local `sync.Map` (`pkg/iscp/runtime_cancel.go:24`), while the singleton ISCP daemon task is leased to one runner CN (`pkg/iscp/util.go:443-459`). A DROP executed on another CN calls `GetExecutorRuntime(c.proc.GetService())`; that process has no runner registered, and `DrainIndexCdcTaskConsumer` logs and returns nil (`pkg/sql/compile/iscp_util.go:205-208`). The foreground DDL then deletes the hidden tables while the remote consumer is still running, so the original deadlock/write-after-drop race remains in multi-CN deployments. Please route the drain to the lease holder (or use a durable/shared quiescence handshake), fail closed when quiescence cannot be confirmed, and add a test where DDL CN != ISCP runner CN.

2. **[High][correctness/unhappy path] The fence is permanent, non-transactional, and not generation-scoped.** `CancelAndDrainJobConsumer` inserts `(accountID, tableID, jobName)` into `fencedJobs` before any DDL mutation (`pkg/iscp/runtime_cancel.go:100-103`), and there is no removal path. If a later `UpdateConstraint`, hidden-table delete, unregister, or transaction commit fails/rolls back, the still-valid original job remains permanently skipped. If DROP succeeds and the same index name is created again, `RegisterJob` creates a new internal job ID (`pkg/iscp/watermark_updater.go:269-303`) but the key is unchanged, so `TableEntry.getCandidate` keeps skipping the new job (`pkg/iscp/table_entry.go:142-143`). The map also grows without bound. Please make the fence generation-aware and align its publish/cleanup with DDL commit/rollback; add deterministic rollback and `DROP INDEX; CREATE INDEX same_name` regressions.

3. **[Medium][hang/CI flakiness] WAIT removal still has a lost-wakeup window.** TRIGGER publishes the `faultEntry` at `pkg/util/fault/fault.go:136-146`, then the caller separately enters `do()` and registers itself as a waiter at lines 169-172. REMOVE can run between those steps, broadcast when no waiter exists, and delete the entry (lines 116-132); the trigger then waits forever with no future notifier. The new test explicitly waits for `nWaiters == 1` before REMOVE, so it cannot exercise this schedule. Please add a removed/predicate state protected by the same mutex (or an equivalent registration handshake) and a deterministic test for remove-after-dispatch-before-wait.

4. **[Functional closure] Other destructive ISCP paths still unregister without draining.** `DROP TABLE` and restore/clone call `DropAllIndexCdcTasks` (`pkg/sql/compile/ddl.go:3172`, `3540`) and then mutate/delete source or hidden tables, but no running consumer is canceled/drained. That leaves the same lifecycle race through sibling DDL paths. Please wire a drain-all primitive into those paths or document and test a concrete exclusion guarantee.

Validation performed on head `fcce6f3f8ae2104032037d1da5ef79bd427ed1e3`: focused `pkg/iscp`, `pkg/sql/compile`, and repeated/race `pkg/util/fault` tests pass. Those tests currently cover only the local executor, successful DDL, and already-registered waiter shapes described above.

XuPeng-SH:
@jiangxinmeng1 conflict

XuPeng-SH (changes requested):
conflict

pkg/util/fault/fault.go
```diff
@@ -192,6 +209,38 @@ func (e *faultEntry) do() (int64, string) {
...
+		select {
+		case <-timer.C:
+		case <-ctx.Done():
+		}
+		return 0, ""
+	}
+	if e.action != WAIT {
+		return e.do()
+	}
+	e.mutex.Lock()
```
XuPeng-SH: `do()` was fixed to skip waiting when `removed` is already true, but the context-aware WAIT path still doesn't have that guard. If `REMOVE` runs after `TRIGGER` publishes `out` but before this waiter registers, the earlier broadcast is lost, `removed` is ignored here, and the waiter blocks until its context is canceled. `objectio.WaitInjectedCtx` and `pkg/iscp/index_consumer.go` make this production-reachable for injected waits. Please mirror the same removed-state handshake here instead of only in `do()`.

pkg/util/fault/fault_test.go
```diff
@@ -182,6 +183,126 @@ func TestWait(t *testing.T) {
...
+		select {
+		case <-done:
+			return true
+		default:
+			return false
+		}
+	}, time.Second, 10*time.Millisecond)
+}
+
+func TestRemoveWaitFaultPointBeforeWaitRegistrationDoesNotHang(t *testing.T) {
```
XuPeng-SH, line 219: This regression only exercises `entry.do()` directly. The remaining bug is on the `TriggerFaultWithContext` path (`doWithContext`), where REMOVE can still happen after TRIGGER dispatch but before waiter registration. Please add the same interleaving for the context-aware entrypoint; otherwise this test stays green while `objectio.WaitInjectedCtx(...)` can still hang until context cancellation.

XuPeng-SH (changes requested):
Re-reviewed the latest head `4c546a95994f95023d458249634c14008446c728`. The earlier correctness blockers around wrong-CN silent success, permanent name-scoped fences, and drop-all paths look materially improved: wrong-CN drain now fails closed, fences are cleaned on commit/rollback and keyed by jobID, and `DropAllIndexCdcTasks` now routes through drain. The focused `pkg/iscp`, `pkg/sql/compile`, and `pkg/util/fault` tests also pass on this head.

I still see one remaining blocker in the fault-injection wait path. `faultEntry.do()` now checks `removed` before registering a waiter, but the context-aware path used by `TriggerFaultWithContext` still does not. `doWithContext(WAIT)` always increments `nWaiters` and waits without first consulting `removed`, so `REMOVE` can still run after TRIGGER dispatch but before waiter registration, broadcast while nobody is waiting, delete the entry, and leave the context-aware waiter blocked until its context is canceled. This path is not just theoretical: `objectio.WaitInjectedCtx` uses `TriggerFaultWithContext`, and `pkg/iscp/index_consumer.go` calls it in the injected wait hooks. The new tests cover `entry.do()` and the cancel-after-registration shape, but they still do not exercise remove-after-dispatch-before-wait on the context-aware trigger path.

Please apply the same removed/registration handshake to `doWithContext(WAIT)` and add a deterministic regression around `TriggerFaultWithContext` + `RemoveFaultPoint` for that interleaving.

XuPeng-SH (approved):
Re-reviewed the latest head `5264206547`, focusing on the context-aware WAIT path I blocked on `4c546a9599`. The added `removed` check in `doWithContext(WAIT)` closes the remove-after-dispatch-before-wait window, and the new regression covers exactly that interleaving. I reran the focused `pkg/util/fault` tests on this head and do not see a remaining high-confidence correctness or unhappy-path blocker in the latest delta.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +196,87 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+	if !ok || exec == nil {
+		if !failIfNoLocalExecutor {
+			logutil.Infof(
+				"skip local drain for index cdc task consumer, iscp executor not found: cn=%s tableID=%d jobName=%s jobID=%d",
+				c.proc.GetService(),
+				tableID,
+				jobName,
+				jobID,
+			)
+			return nil
```
LeftHandCold: [P1] Do not continue destructive DDL without quiescing the runner CN

This is still the original deadlock when the ISCP task runs on a different CN. DropIndexCdcTask writes Canceling/drop_at in the DROP transaction, so the runner cannot observe or enqueue that cancellation until this transaction commits; however the hidden tables are dropped later in this same transaction. If an existing consumer already holds a hidden-table write lock and then performs its normal mo_iscp_log result update, the DDL owns/waits on the log row and then waits on the hidden table while the consumer waits on the log row. The transaction therefore cannot reach the commit that would trigger the async cancellation. Returning success here makes the two-CN path fail open. Please route cancel/drain to the actual runner CN, or provide an out-of-transaction durable quiescence handshake and wait for it before dropping the hidden tables. The regression test needs an already-running remote consumer paused after touching the hidden table, not merely a two-CN DROP with no in-flight consumer.

pkg/iscp/runtime_cancel.go
```diff
@@ -0,0 +1,361 @@
...
+	if exec == nil {
+		return nil
+	}
+	key := NewJobRuntimeKey(accountID, tableID, jobName, jobID)
+	groupKey := key.consumerGroupKey()
+	cancelErr := moerr.NewInternalErrorNoCtx("iscp job consumer canceled")
+
+	exec.runtimeMu.Lock()
+	exec.ensureRuntimeMapsLocked()
+	exec.fencedJobs[key] = struct{}{}
```
LeftHandCold: [P2] Reclaim generation fences after terminal state is installed

Every async cancellation reaches finishCanceledJob, which calls this method and inserts a generation key into fencedJobs; after markJobCanceled succeeds it only clears cancelingJobs. The cross-CN path has no compile callback that removes this fence, and the local path can re-add it after its commit callback already ran. GCInMemoryJob also does not remove these keys. Since every drop/recreate gets a new JobID, a long-lived runner retains one map entry and job-name string per historical generation without bound. Please remove the fence at a safe point after the canceled/drop state has been applied to in-memory scheduling state, or from the corresponding GC path, so cleanup does not reopen the pre-logtail admission window.

LeftHandCold (changes requested):
Reviewed the current head (76ddc868). The local cancellation path is substantially safer now, and the targeted fault/ISCP/compile tests pass, but the cross-CN path still does not establish quiescence before destructive DDL. That leaves the original lock cycle reachable. I also found a deterministic lifetime leak in the generation-scoped fence map. Please address the two inline findings and add a two-CN regression that pauses an already-running consumer after it has touched the hidden table, then verifies DROP completes without waiting for the DDL transaction to commit first.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +197,83 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+		tableID = tableDef.TblId
+	}
+	runnerCN, err := iscpGetTaskRunnerFunc(c.proc.Ctx, c.proc.GetService(), c.proc.GetTxnOperator())
+	if err != nil {
+		return err
+	}
+	if runnerCN == "" {
+		runnerCN = c.proc.GetService()
+	}
+	exec, ok := iscpGetExecutorFunc(runnerCN)
```
LeftHandCold: [P1] The runner UUID does not make this process-local lookup remote

GetExecutorRuntime reads the package-level iscpExecutors sync.Map in the DDL CN process. RegisterExecutorRuntime is only called by the executor that actually started in the runner process, so a different CN process cannot have an entry for runnerCN. Consequently every normal DROP INDEX/DROP TABLE issued on a non-runner CN now reaches the error below instead of draining the runner; the new unit test hides this boundary by stubbing iscpGetExecutorFunc to return an in-process runnerExec for an arbitrary UUID. This still needs an actual RPC to the lease holder or a durable shared quiescence handshake. Please cover it with separate CN processes (and an already-running consumer), not a function stub that models a cross-process pointer lookup.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +197,83 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+	}
+	if txnOp := c.proc.GetTxnOperator(); txnOp != nil {
+		cleanup := client.NewTxnEventCallback(func(_ context.Context, _ client.TxnOperator, event client.TxnEvent, _ any) error {
+			if !event.CostEvent {
+				return nil
+			}
+			exec.RemoveJobFence(key)
+			return nil
+		})
+		txnOp.AppendEventCallback(client.CommitEvent, cleanup)
```
LeftHandCold: [P1] Do not remove the successful-drop fence before logtail applies

This same cleanup is registered for both CommitEvent and RollbackEvent, so a successful commit removes the fence as soon as the commit response returns. At that point the runner is not guaranteed to have applied the Canceling/drop_at logtail yet; executor.go explicitly handles the fact that a transaction commit can precede logtail visibility. If a scheduler tick lands in that gap, the old in-memory job is still Completed with dropAt == 0 and can be admitted again after the hidden tables were dropped. Rollback must remove the fence, but success should retain it until addOrUpdateJob has installed dropAt and finishCanceledJob performs the new safe RemoveJobFence cleanup. Please register this callback only for rollback and add a regression where commit completes before the cancellation logtail is applied.

LeftHandCold (changes requested):
Re-reviewed the latest head (4ad3702d). The previous unbounded generation-fence retention is materially improved: async finalization now removes the exact fence, and table GC has a fallback cleanup. The focused ISCP and compile tests pass. Two correctness blockers remain, though: the new runner lookup still dereferences a process-local registry and therefore cannot drain a separate CN process, and the successful-commit callback removes the fence before the cancel record is guaranteed to be visible in the runner in-memory state. Please address the inline findings and replace the stubbed runner-routing test with a real cross-process/RPC or durable-handshake test.

pkg/sql/compile/iscp_util.go
jiangxinmeng1: This review is addressed by the latest changes.

  The destructive DDL path no longer depends on a process-local executor lookup on
  the DDL CN. We still use the global ISCP daemon task runner UUID from
  sys_daemon_task/GetTaskRunner, but when that runner is remote, the DDL CN now
  routes an ISCPDrainConsumer request through the existing queryservice to the
  runner CN. The runner CN then performs CancelAndDrainJobConsumer against its
  local executor before the DDL proceeds to drop metadata or hidden tables.

  The generation fence lifetime is also handled for the remote path. The drain
  request installs the fence on the runner CN, and the DDL transaction registers
  commit/rollback callbacks that send a RemoveFenceOnly request back to the same
  runner CN, so the remote fence is reclaimed after the DDL transaction reaches a
  terminal outcome.

jiangxinmeng1:
This review is addressed by the latest changes.

  The destructive DDL path no longer depends on a process-local executor lookup on
  the DDL CN. We still use the global ISCP daemon task runner UUID from
  sys_daemon_task/GetTaskRunner, but when that runner is remote, the DDL CN now
  routes an ISCPDrainConsumer request through the existing queryservice to the
  runner CN. The runner CN then performs CancelAndDrainJobConsumer against its
  local executor before the DDL proceeds to drop metadata or hidden tables.

  The generation fence lifetime is also handled for the remote path. The drain
  request installs the fence on the runner CN, and the DDL transaction registers
  commit/rollback callbacks that send a RemoveFenceOnly request back to the same
  runner CN, so the remote fence is reclaimed after the DDL transaction reaches a
  terminal outcome.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +202,158 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+		}
+		if txnOp := c.proc.GetTxnOperator(); txnOp != nil {
+			cleanup := client.NewTxnEventCallback(func(_ context.Context, _ client.TxnOperator, event client.TxnEvent, _ any) error {
+				if !event.CostEvent {
+					return nil
+				}
+				exec.RemoveJobFence(key)
+				return nil
+			})
+			txnOp.AppendEventCallback(client.CommitEvent, cleanup)
```
LeftHandCold: [P1] Keep the success fence until this drop is visible to the runner

The queryservice routing fixes the previous cross-process lookup problem, but registering this cleanup on `CommitEvent` preserves the commit-to-logtail race. The CostEvent runs after the commit RPC returns, not after the runner has applied the `drop_at` logtail. A scheduler tick in that window sees the old in-memory job with `dropAt == 0`, admits it again, and can write to hidden tables that this transaction has already dropped. The remote branch repeats the same cleanup at line 299; if that post-commit RPC fails, its error is also joined into `Commit()`, so the client can be told the DDL failed even though it committed. Only rollback should clear here. On success, clear the exact generation fence on the runner when `addOrUpdateJob` has installed the nonzero `dropAt` (including the no-table case), and cover commit-before-logtail with a regression test.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +202,158 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+	}
+	if txnOp := c.proc.GetTxnOperator(); txnOp != nil {
+		cleanup := client.NewTxnEventCallback(func(ctx context.Context, _ client.TxnOperator, event client.TxnEvent, _ any) error {
+			if !event.CostEvent {
+				return nil
+			}
+			if ctx == nil {
+				ctx = c.proc.Ctx
+			}
+			return sendISCPDrainConsumerRequest(ctx, qc, queryAddress, accountID, tableID, jobName, jobID, true)
```
LeftHandCold: [P1] A failed rollback cleanup permanently disables a live job

After the first RPC succeeds, the runner has fenced this generation. If the DDL rolls back, no `drop_at` will ever become visible, so this RPC is the only release path. It is sent once with the transaction callback context; cancellation or a transient CN-to-CN failure skips the release, and `txnOperator.Rollback` ignores callback errors. There is no retry, fence expiry, or durable recovery, so `getCandidate`/consumer registration will keep skipping an otherwise valid job indefinitely, leaving its index stale until the executor restarts. Please make rollback release survive a canceled caller and transient delivery failure (or give the fence a recoverable lease/durable owner), and test a failed cleanup followed by rollback/recovery.

LeftHandCold (changes requested):
Re-reviewed current head dce285f1. The queryservice RPC fixes the process-local cross-CN routing blocker. Two lifecycle blockers remain: the success fence is still removed before the drop logtail is visible, and remote rollback cleanup is a one-shot RPC whose failure can permanently fence a live job.

pkg/sql/compile/iscp_util.go
LeftHandCold: Clarification: a definitely failed commit also needs rollback-style cleanup; the unsafe case is removing the fence after a successful commit before `drop_at` is applied. An `ErrTxnUnknown` outcome cannot safely choose either immediate removal or an infinite fence, so it needs the same recoverable lease/outcome mechanism. My “only rollback should clear here” wording means the successful-commit path must not use this callback.

aptend (changes requested):
Re-reviewed the current head `1083581c61`. The cross-CN drain routing and rollback recovery are materially improved, but the latest fence-lifetime change still leaves two generation-safety gaps, and the context-aware fault wait has an independent cross-waiter cancellation bug. Requesting changes for the following blockers:

1. **[P1][generation safety] The success fence is removed before `dropAt` is installed in memory.** `addOrUpdateJob` calls `RemoveJobFence` at `pkg/iscp/executor.go:975-977`, then parses the row, looks up the table, and only installs `dropAt` through `AddOrUpdateSinker` at line 1020. During that interval, `TableEntry.getCandidate` can observe the old entry with `dropAt == 0` and no fence; both worker-side fence checks also see the fence removed, so the old generation can be admitted against hidden tables that the committed DDL has already dropped. I reproduced this deterministically by pausing `addOrUpdateJob` after fence removal but before the table lookup: `getCandidate` returned the dropped job. Please install the terminal in-memory state before removing the exact-generation fence, preserve the fence on parse/update failure, and add a regression covering this handoff.

2. **[P1][generation safety] The fixed TTL reopens successful drops when logtail catch-up exceeds five minutes.** `CancelAndDrainJobConsumer` assigns every fence `DefaultRollbackFenceTTL`, and `isJobFencedLocked` deletes it after expiration regardless of transaction outcome. The TTL solves permanent retention after failed rollback cleanup, but it also applies to a successfully committed DROP. If the runner cannot apply the `drop_at` logtail within five minutes, the old in-memory job becomes schedulable again and the original write-after-drop race returns. Please distinguish rollback/unknown-outcome recovery from a successful-drop fence, or renew/reconcile the lease against durable state so success remains fenced until `dropAt` is actually installed. Add a success-with-stalled-logtail regression.

3. **[P2][fault correctness] Canceling one context-aware WAIT releases unrelated waiters.** In `pkg/util/fault/fault.go:232-244`, each waiter's context watcher calls the shared condition's `Broadcast()`. With two `TriggerFaultWithContext` calls waiting on the same fault point, canceling only one context wakes both calls even though the second context is still live. A deterministic two-waiter regression fails immediately on this head. Please use per-waiter cancellation/registration or an equivalent predicate that does not turn one caller's cancellation into a global notify.

Validation on `1083581c61`: focused tests for `pkg/iscp`, `pkg/sql/compile`, `pkg/cnservice`, `pkg/queryservice/client`, and `pkg/util/fault` pass; `pkg/iscp` and `pkg/util/fault` pass under `-race`, including focused `-count=20` stress. The two deterministic regressions described above fail as expected and expose gaps not covered by the current tests. `git diff --check` passes.

jiangxinmeng1:
> Re-reviewed the current head `1083581c61`. The cross-CN drain routing and rollback recovery are materially improved, but the latest fence-lifetime change still leaves two generation-safety gaps, and the context-aware fault wait has an independent cross-waiter cancellation bug. Requesting changes for the following blockers:
> 
> 1. **[P1][generation safety] The success fence is removed before `dropAt` is installed in memory.** `addOrUpdateJob` calls `RemoveJobFence` at `pkg/iscp/executor.go:975-977`, then parses the row, looks up the table, and only installs `dropAt` through `AddOrUpdateSinker` at line 1020. During that interval, `TableEntry.getCandidate` can observe the old entry with `dropAt == 0` and no fence; both worker-side fence checks also see the fence removed, so the old generation can be admitted against hidden tables that the committed DDL has already dropped. I reproduced this deterministically by pausing `addOrUpdateJob` after fence removal but before the table lookup: `getCandidate` returned the dropped job. Please install the terminal in-memory state before removing the exact-generation fence, preserve the fence on parse/update failure, and add a regression covering this handoff.
> 2. **[P1][generation safety] The fixed TTL reopens successful drops when logtail catch-up exceeds five minutes.** `CancelAndDrainJobConsumer` assigns every fence `DefaultRollbackFenceTTL`, and `isJobFencedLocked` deletes it after expiration regardless of transaction outcome. The TTL solves permanent retention after failed rollback cleanup, but it also applies to a successfully committed DROP. If the runner cannot apply the `drop_at` logtail within five minutes, the old in-memory job becomes schedulable again and the original write-after-drop race returns. Please distinguish rollback/unknown-outcome recovery from a successful-drop fence, or renew/reconcile the lease against durable state so success remains fenced until `dropAt` is actually installed. Add a success-with-stalled-logtail regression.
> 3. **[P2][fault correctness] Canceling one context-aware WAIT releases unrelated waiters.** In `pkg/util/fault/fault.go:232-244`, each waiter's context watcher calls the shared condition's `Broadcast()`. With two `TriggerFaultWithContext` calls waiting on the same fault point, canceling only one context wakes both calls even though the second context is still live. A deterministic two-waiter regression fails immediately on this head. Please use per-waiter cancellation/registration or an equivalent predicate that does not turn one caller's cancellation into a global notify.
> 
> Validation on `1083581c61`: focused tests for `pkg/iscp`, `pkg/sql/compile`, `pkg/cnservice`, `pkg/queryservice/client`, and `pkg/util/fault` pass; `pkg/iscp` and `pkg/util/fault` pass under `-race`, including focused `-count=20` stress. The two deterministic regressions described above fail as expected and expose gaps not covered by the current tests. `git diff --check` passes.

  Thanks for calling this out. I intentionally did not change the 5-minute fence TTL
  semantics in this revision.

  The fence is installed only after the foreground DROP has synchronously routed to the
  ISCP runner and drained the currently running consumer. After that point, the remaining
  window is the runner applying the committed `drop_at` logtail into its in-memory job
  state. Under normal operation that catch-up should complete well within the existing
  rollback fence TTL; the TTL is mainly a recovery guard so a rolled-back DROP whose
  rollback cleanup RPC is lost does not leave a live job fenced forever.

  Making successful DROP fences non-expiring until `drop_at` is applied would require
  introducing an additional commit-outcome/generation state, or another reliable cross-CN
  handoff, so rollback cleanup remains recoverable without permanently fencing a live
  job. That is a larger lifecycle change than I want to add in this PR. For this patch I
  kept the existing TTL behavior, and fixed the concrete generation handoff bug by moving
  fence removal until after `drop_at` has been installed in memory and preserving the
  fence on parse/update failure.

  The added regression covers this handoff: after `addOrUpdateJob` installs `drop_at` and
  removes the fence, `getCandidate` still does not return the dropped job; if parsing/
  updating the dropped job fails, the generation fence is preserved.

pkg/iscp/runtime_cancel.go
```diff
@@ -0,0 +1,252 @@
...
+	if exec == nil {
+		return nil
+	}
+	key := NewJobRuntimeKey(accountID, tableID, jobName, jobID)
+	groupKey := key.consumerGroupKey()
+	cancelErr := moerr.NewInternalErrorNoCtx("iscp job consumer canceled")
+
+	exec.runtimeMu.Lock()
+	exec.ensureRuntimeMapsLocked()
+	exec.fencedJobs[key] = JobFence{ExpireAt: time.Now().Add(rollbackFenceTTL())}
```
LeftHandCold: [P1] Do not expire an in-flight DDL fence before the transaction resolves

This assigns the rollback fallback TTL to every newly drained fence, before the DDL transaction has committed or rolled back. MatrixOne explicitly permits `DROP INDEX`, `DROP TABLE`, and `DROP DATABASE` inside an uncommitted transaction (`statementCanBeExecutedInUncommittedTransaction` returns true), and these statements do not force a commit via `NeedToBeCommittedInActiveTransaction`. Thus `BEGIN; DROP INDEX ...;` followed by more than five minutes of work/idle time deterministically reaches `isJobFencedLocked`, deletes this still-live fence, and lets `getCandidate` admit the old generation while `drop_at` remains uncommitted. That reopens the consumer/hidden-table race this PR is fixing. The expiry should only become eligible after rollback cleanup delivery has failed (or the DDL side should renew a lease until transaction resolution); it cannot be an unconditional wall-clock deadline from the initial drain. Please add a regression that keeps the transaction unresolved past an injected short TTL and verifies the job remains fenced until commit-logtail handoff or rollback.

LeftHandCold (changes requested):
Re-reviewed current head a08c234d. The previous commit-to-logtail cleanup race and permanent rollback-fence leak are addressed, but the new fixed fence TTL can expire while a valid explicit DDL transaction is still open, reopening the consumer/hidden-table race this PR is intended to close.

aptend (changes requested):
Reviewed current head `70aba40f90`. The latest change only extends the in-memory fence TTL from 5 to 30 minutes; it does not close the generation/terminal handoff gaps below. Requesting changes for three blockers:

1. **[P1][generation safety] The fence is lost when the ISCP runner executor is replaced.** `fencedJobs` is process-local and every new `ISCPTaskExecutor` starts with an empty map. If the runner CN restarts or the daemon-task lease migrates while the DROP transaction is still open, the replacement executor replays the still-live durable job with no fence and can admit it against the hidden tables again. The lease cannot repair this: renewal is pinned to the originally resolved query address, `RenewJobFence` silently does nothing when the key is absent, and the RPC handler still returns success. I reproduced this with a deterministic executor-replacement regression: the old executor is fenced, the replacement receives the current renewal call, and `replacement.IsJobFenced(key)` is false. Please make the fence durable/reconstructible across runner generations (or explicitly hand it off/re-resolve and install it), make renewal fail when the fence is missing, and add a runner restart/lease-migration regression while the DDL transaction is unresolved.

2. **[P1][correctness] Renewal stops at commit entry, before either commit or the `drop_at` logtail handoff is complete.** The non-cost `CommitEvent` invokes `lease.Stop()`, but `txnOperator.Commit` emits that event before `doWrite`. Therefore a slow/unknown commit or delayed logtail can outlive the remaining TTL; `IsJobFenced` then deletes the fence while the in-memory job still has `dropAt == 0`, reopening the original consumer/hidden-table race. Increasing the TTL to 30 minutes only delays this failure. The new test explicitly expects the fence to expire after commit starts, which codifies the unsafe contract. Remote renewal failures have the same problem: they are only logged, so an unresolved DDL silently becomes unfenced after the TTL. Please keep a successful-drop fence until `addOrUpdateJob` installs `dropAt`, remove/recover it only for rollback or a definitely failed commit, reconcile unknown outcomes, and make renewal failure fail closed. Add short injected-TTL regressions for commit-before-logtail, stalled logtail, and failed renewal.

3. **[P2][build reproducibility] `proto/query.proto` and `query.pb.go` disagree.** The generated Go request contains and the implementation uses `RenewFenceOnly` field 6, but `ISCPDrainConsumerRequest` in `proto/query.proto` ends at `RemoveFenceOnly = 5`. Regenerating protobufs removes the field and breaks the two call sites. Please add `bool RenewFenceOnly = 6;` to the source proto and regenerate `query.pb.go` with the repository target.

Validation: existing focused tests for `pkg/iscp`, the drain-related `pkg/sql/compile` and `pkg/cnservice` cases, and `pkg/util/fault -race` pass; `git diff --check` passes. The deterministic runner-replacement regression described above fails on this head.

jiangxinmeng1:
I fixed the proto source mismatch by adding `RenewFenceOnly = 6` to `proto/
  query.proto`.

  For the durable fence suggestion, I agree that a purely in-memory fence cannot survive
  runner restart or lease migration. However, I do not think making this fence durable is
  the right fix for this PR. A durable fence needs a full lifecycle protocol for commit,
  rollback, unknown commit outcome, executor recovery, and cleanup. If any cleanup path
  is missed, the stale durable fence can permanently suppress a still-valid ISCP job and
  leave the index stale. In this case, a leftover fence is more dangerous than the
  transient in-memory fence loss, because it can silently stop future maintenance for a
  live index until manual intervention.

  The current patch keeps the fence process-local and transaction-scoped, with renewal
  while the DDL transaction is unresolved. This avoids introducing a new durable state
  machine with incomplete cleanup guarantees. I agree this means runner restart or lease
  migration during an unresolved DDL transaction is not fully covered by the in-memory
  fence design; covering that correctly should be handled with a separate, explicit
  cross-runner lifecycle design rather than by persisting this map opportunistically

fengttt (approved):

aunjgr (approved):

aptend (changes requested):
There is still a generation-safety gap in the cancellation fence.

The fence is stored only in the current executor's in-memory `fencedJobs` map. If the ISCP executor restarts or the daemon-task lease migrates while the DROP transaction is unresolved, the replacement executor starts with an empty fence map and can rebuild/admit the still-visible job (`dropAt` is not committed yet). The local lease also keeps renewing the old executor pointer.

The remote path has the same safety hole after restart or expiry: `RenewJobFence` silently does nothing when the exact fence is absent, while `handleISCPDrainConsumer` returns success. The DDL side therefore believes the fence is renewed even though the consumer can run again, reopening the DROP INDEX/consumer race this PR is intended to close.

Please make the fence durable or executor-generation/lease-owner aware, and require renewal to prove that the exact fence still exists (otherwise re-resolve the runner and re-drain/reinstall it before acknowledging success). The DDL must not commit if quiescence can no longer be proven. Please also add coverage for executor restart/task migration during an unresolved DDL transaction and renewal after the fence has expired or disappeared.

pkg/sql/compile/iscp_util.go
```diff
@@ -192,6 +204,231 @@ func DropIndexCdcTask(c *Compile, tableDef *plan.TableDef, dbname string, tablen
...
+type iscpJobFenceLease struct {
+	stop func()
+}
+
+func startISCPJobFenceLease(renew func() error) iscpJobFenceLease {
+	ttl := iscp.RollbackFenceTTL()
+	if ttl <= 0 || renew == nil {
+		return iscpJobFenceLease{stop: func() {}}
+	}
+	interval := ttl / 2
```
LeftHandCold, line 353: [P2] Do not let a renewal run past the current lease

The remote renew call is allowed to run for the full TTL (lines 303-306), but the first attempt does not start until `ttl/2` here. A slow or partitioned runner can therefore keep that call in flight beyond the current fence expiry. Once `IsJobFenced` observes the expiry it deletes the key; the late RPC then cannot restore it because `RenewJobFence` only updates an existing key, while `handleISCPDrainConsumer` still returns `Success: true`. Every later renewal silently no-ops, so a long explicit DROP transaction can re-admit this consumer before the transaction outcome and recreate the drop/write race. I reproduced this with the 1-second injected TTL and a renew blocked for more than the remaining 500 ms. Please bound/retry renewal within the remaining lease with safety margin, and make a missing/expired renewal fail rather than acknowledging lost protection.

LeftHandCold (changes requested):
Found one remaining correctness issue in the remote fence renewal. The lease can still expire while a renewal RPC is in flight, and a late renewal is then silently acknowledged after the key has already been deleted, permanently losing the protection during an open DDL transaction. Requesting changes for the inline issue. I rechecked the consumer cancellation/drain, fan-out refcounting, rollback cleanup, and committed drop-logtail handoff; those paths look closed. The affected package tests pass.

pkg/sql/compile/iscp_util.go
jiangxinmeng1: The default fence TTL is now 30 minutes, so the specific renewal-overrun schedule
  requires a renewal RPC to remain blocked for longer than the remaining half-lease
  window, roughly 15 minutes with the current interval. That is not impossible, but it is
  a very conservative failure window for this in-memory guard.

  Also, this fence is a best-effort transaction-scoped quiescence guard rather than
  durable job state. If renewal cannot be proven, the intended behavior should be fail-
  closed on the DDL side, which degrades back to the previous “cannot confirm ISCP
  consumer quiescence” failure instead of allowing hidden-table mutation to proceed. I
  agree we should make missing/expired renewal return an explicit error instead of
  silently acknowledging it, but I do not think this requires making the fence durable.

jiangxinmeng1:
> There is still a generation-safety gap in the cancellation fence.
> 
> The fence is stored only in the current executor's in-memory `fencedJobs` map. If the ISCP executor restarts or the daemon-task lease migrates while the DROP transaction is unresolved, the replacement executor starts with an empty fence map and can rebuild/admit the still-visible job (`dropAt` is not committed yet). The local lease also keeps renewing the old executor pointer.
> 
> The remote path has the same safety hole after restart or expiry: `RenewJobFence` silently does nothing when the exact fence is absent, while `handleISCPDrainConsumer` returns success. The DDL side therefore believes the fence is renewed even though the consumer can run again, reopening the DROP INDEX/consumer race this PR is intended to close.
> 
> Please make the fence durable or executor-generation/lease-owner aware, and require renewal to prove that the exact fence still exists (otherwise re-resolve the runner and re-drain/reinstall it before acknowledging success). The DDL must not commit if quiescence can no longer be proven. Please also add coverage for executor restart/task migration during an unresolved DDL transaction and renewal after the fence has expired or disappeared.

I agree that an in-memory fence cannot provide a hard guarantee across runner restart
or daemon-task lease migration while the DDL transaction is still unresolved. I also
agree that a non-durable re-resolve/re-drain scheme has its own risks: it adds more
cross-CN retry logic inside an open DDL transaction, has to reason about unknown commit
outcomes, and can still fail to prove quiescence under partitions or runner churn.

For this PR, I intentionally kept the fence process-local and transaction-scoped. The
default TTL is now 30 minutes and the lease is renewed while the DDL transaction
remains unresolved, so the normal long-transaction case is covered without adding a new
durable state machine or a complex cross-runner handoff protocol. Persisting the fence
is also risky: a stale durable fence can permanently suppress a still-valid ISCP job
and leave the index stale until manual intervention.

Given that tradeoff, I would prefer to keep this PR scoped to the active-runner
lifecycle: cancel and drain the current consumer, keep the in-memory generation fence
alive during the unresolved DDL transaction, remove it on rollback, and clear it after
the committed drop_at is installed. Runner crash or daemon-task migration during an
unresolved destructive DDL is a narrower availability/failure-domain case that should
be handled by a separate lifecycle design rather than by opportunistically persisting
this fence map in this PR.

@aptend

aptend:
@jiangxinmeng1 Thanks, I agree that persisting the fence is not necessarily the right scope for this PR, and runner restart/daemon-task lease migration can be tracked as a separate lifecycle design.

However, the current implementation still does not guarantee the active-runner lifecycle described here. No restart or migration is required: a renewal RPC starts at TTL/2 but may run for a full TTL. The existing fence can therefore expire while renewal is still in flight. Once `IsJobFenced` deletes the expired key, `RenewJobFence` silently no-ops and `handleISCPDrainConsumer` still returns success. The DDL side consequently does not fail closed, and the consumer can be admitted again while the transaction remains unresolved.

This is the same active-runner case described in https://github.com/matrixorigin/matrixone/pull/25733#discussion_r3601489108, and does not require a durable fence. Please make renewal finish within the remaining lease with a safety margin, return an explicit error for a missing/expired fence, and ensure loss of the lease prevents the DDL from committing. With that closed, I am fine with documenting restart/migration as a separate follow-up.

LeftHandCold (approved):
Re-reviewed the latest head `fc114288d7`. The remaining fence-renewal failure mode can re-expose an intermittent DDL deadlock/retry window, but it does not establish a committed data-correctness failure: the initial consumer drain closes the active race, the destructive DDL changes are transactional, and consumer result updates do not clear `drop_at`. Treating renewal as best-effort avoids turning a transient control-plane failure into a guaranteed rollback of an otherwise valid transaction. I do not see a remaining blocker on this head.

jiangxinmeng1:
> @jiangxinmeng1 Thanks, I agree that persisting the fence is not necessarily the right scope for this PR, and runner restart/daemon-task lease migration can be tracked as a separate lifecycle design.
> 
> However, the current implementation still does not guarantee the active-runner lifecycle described here. No restart or migration is required: a renewal RPC starts at TTL/2 but may run for a full TTL. The existing fence can therefore expire while renewal is still in flight. Once `IsJobFenced` deletes the expired key, `RenewJobFence` silently no-ops and `handleISCPDrainConsumer` still returns success. The DDL side consequently does not fail closed, and the consumer can be admitted again while the transaction remains unresolved.
> 
> This is the same active-runner case described in [#25733 (comment)](https://github.com/matrixorigin/matrixone/pull/25733#discussion_r3601489108), and does not require a durable fence. Please make renewal finish within the remaining lease with a safety margin, return an explicit error for a missing/expired fence, and ensure loss of the lease prevents the DDL from committing. With that closed, I am fine with documenting restart/migration as a separate follow-up.

@aptend 
The original issue already demonstrates that, without a confirmed ISCP quiescence
  fence, the DDL path does not silently succeed in the observed race. It fails with
  `cannot confirm ISCP consumer quiescence` / DDL conflict instead of completing hidden-
  table removal under an active consumer. So losing the renewal does not necessarily mean
  the DDL must be force-aborted by the lease layer itself; the underlying DDL/lock path
  can still fail the statement.

  That said, I agree the renewal ACK semantics must not be misleading. I have changed
  `RenewJobFence` / `RenewFenceOnly` so a missing or expired exact fence returns an
  explicit error, and I bounded the renewal RPC timeout to complete within the lease
  window. With that change, renewal loss is visible instead of being acknowledged as
  success.

  I would prefer not to add another pre-commit abort path in this PR unless we can show
  that the DDL can silently commit hidden-table removal after the active-runner fence is
  lost. The currently observed failure mode is an explicit DDL error, which is acceptable
  for this best-effort in-memory fence.

gouhongshen (changes requested):
## Codex automated review

The fence-renewal failure path is not enforced, so a runner restart during an open DROP INDEX transaction can re-enable the consumer before the drop commits.

### P1 - Abort or otherwise preserve quiescence when fence renewal fails (pkg/sql/compile/iscp_util.go:365)
Fence state is in-memory only. If the ISCP runner restarts (or the fence expires) while the DDL transaction remains open, `RenewJobFence` correctly returns false, but this goroutine only logs the error and continues. The replacement executor has no fence and `RegisterRunningConsumer` accepts the still-active, uncommitted job, allowing it to resume writes to the index being dropped. The new handler error for a missing fence is therefore never propagated to the DDL transaction. The renewal failure must keep the job fenced or cause the DROP transaction to fail/abort; logging is insufficient.

jiangxinmeng1:
> ## Codex automated review
> The fence-renewal failure path is not enforced, so a runner restart during an open DROP INDEX transaction can re-enable the consumer before the drop commits.
> 
> ### P1 - Abort or otherwise preserve quiescence when fence renewal fails (pkg/sql/compile/iscp_util.go:365)
> Fence state is in-memory only. If the ISCP runner restarts (or the fence expires) while the DDL transaction remains open, `RenewJobFence` correctly returns false, but this goroutine only logs the error and continues. The replacement executor has no fence and `RegisterRunningConsumer` accepts the still-active, uncommitted job, allowing it to resume writes to the index being dropped. The new handler error for a missing fence is therefore never propagated to the DDL transaction. The renewal failure must keep the job fenced or cause the DROP transaction to fail/abort; logging is insufficient.

@gouhongshen 
I do not think renewal loss should automatically abort the DDL transaction.

  The cancellation fence is an in-memory guard used to improve the chance that the DROP
  can complete cleanly by keeping the active ISCP consumer out of the hidden-table path
  while the DDL transaction is unresolved. If that fence is lost, the system falls back
  to the pre-fence behavior: the DDL and the consumer may contend on the hidden tables /
  ISCP log rows, and the normal transaction, lock, deadlock-detection, and table-deletion
  paths decide the outcome.

  That can increase the probability that the DROP statement fails or is rolled back by
  the underlying DDL/lock conflict handling, but it does not by itself create a
  correctness violation. A consumer error after or during hidden-table removal is
  acceptable for a job whose index is being dropped; the index no longer needs to be
  maintained. The important part is that renewal loss is no longer acknowledged as
  success: `RenewJobFence` / `RenewFenceOnly` now return an explicit error when the exact
  fence is missing or expired, and the renewal RPC is bounded within the lease window.

  Given that, I would prefer not to add a separate pre-commit abort path in this PR. The
  fence remains a best-effort active-runner quiescence guard, not durable DDL state.

gouhongshen (changes requested):
## Codex automated review

Two blocking lifecycle gaps can let DROP INDEX race with ISCP work after cancellation.

### P1 - Executor handoff can silently lose the DROP INDEX fence (pkg/sql/compile/iscp_util.go:365)
Fence renewal failures are only logged. The fence is stored in the selected executor instance; if that runner stops/restarts, the DDL lease continues using the old executor or fixed query address, while the replacement executor starts with an empty fence map and rebuilds the still-uncommitted job as active. Its candidate scan then admits the job. Renewal failure or generation replacement must be treated as a failed quiescence operation or transferred to the current executor generation.

### P1 - InitSQL index builds are not drained or cancellable (pkg/iscp/runtime_cancel.go:120)
CancelAndDrainJobConsumer only cancels entries in runningConsumers. The InitSQL branch in ExecuteIterationWithRuntime calls ProcessInitSQL and returns before runISCPTaskIterationConsumers registers a RunningJobConsumer, so an async CAGRA/IVF-PQ build can pass the one-time fence check, continue in its own transaction, and never be awaited by DROP INDEX before hidden index tables are dropped. The InitSQL execution needs the same runtime registration/cancellation lifecycle or a fence recheck plus join.

aptend (changes requested):
Reviewed head `9677012a68`.

1. **[P1][correctness] Do not mark InitSQL complete when registration loses to a fence.** `runInitSQLWithRuntime` returns `nil` when `RegisterRunningConsumer` rejects an already-fenced generation. The caller treats that as successful execution, changes the status stage to `JobStage_Running`, and flushes it as completed. A concrete sequence is: the scheduler passes its earlier fence check; DROP installs the fence before this registration; registration returns false; the DROP later rolls back; the independent status flush then succeeds, but the InitSQL was never executed. The live job will permanently skip its initial CAGRA/IVF-PQ build (or restore rebuild), leaving the index incomplete. Return a cancellation/fenced sentinel and exit before the stage/status transition; add a caller-level regression for fence-before-register followed by DROP rollback. (`pkg/iscp/iteration.go`, `runInitSQLWithRuntime` and the `needInit` branch.)

2. **[P2][fault correctness] Canceling a context-aware WAIT still releases a plain WAIT on the same fault point.** The context watcher calls the shared condition variable’s `Broadcast`, while the non-context `do(WAIT)` performs one unconditional `cond.Wait()` with no predicate. With one `TriggerFault` waiter and one `TriggerFaultWithContext` waiter on the same entry, canceling only the latter makes both return. I reproduced this deterministically on this head. Use per-waiter cancellation/notification, or make every wait path use a predicate that does not interpret another waiter’s cancellation broadcast as NOTIFY. (`pkg/util/fault/fault.go:177-184, 241-247`.)

3. **[P2][index-plugin gate] Keep HNSW-specific drop ordering out of the generic SQL layer.** `makeNewDropConstraint` now checks the literal `hnsw` and `Hnsw_TblType_Storage`. This reintroduces per-algorithm dispatch in `pkg/sql/compile`, which the index-plugin contract explicitly forbids. Expose hidden-table drop priority/order through the plugin hooks and keep the SQL layer algorithm-agnostic. (`pkg/sql/compile/ddl.go:2584-2589`.)

The previously reported fence-renewal failure also remains open; I did not duplicate that finding here.

aptend (changes requested):
Re-reviewed head `ba9bf8a8eadf5b35d6a72c80964a2669da483e5f`. The latest update fixes the previous fenced-registration status transition, moves HNSW drop ordering behind the plugin hook, and closes the direct plain-WAIT cancellation repro. Two correctness gaps remain, and the previously reported renewal blocker is unchanged:

1. **[P1][InitSQL lifecycle] Return the terminal transaction error before advancing the job stage.** `ProcessInitSQL` defers `txnOp.Commit(ctx)` and discards its result. The context is now the cancelable runtime context installed by `runInitSQLWithRuntime`. If DROP cancels after `RunSql` has returned success but before the deferred commit completes, `Commit` can fail on the canceled context while `ProcessInitSQL` still returns nil; the caller then writes `JobStage_Running` even though the initial index-build transaction did not commit. Error paths also attempt commit rather than rollback. Please perform explicit commit/rollback with an independent bounded cleanup context, propagate the terminal error, and add a regression where the SQL body succeeds but commit fails/cancellation races with commit. (`pkg/iscp/iteration.go:1037-1040`, caller at lines 158-185.)

2. **[P2][fault correctness] A prior NOTIFY plus a later cancellation releases unrelated waiters.** All waiters capture the same global `notifySeq`. `NOTIFY` increments it and signals one waiter; if one remaining context waiter is then canceled, its watcher broadcasts the shared condition. Every other waiter observes the sequence change caused by the earlier single NOTIFY and exits too. I reproduced this deterministically with three waiters: one NOTIFY followed by cancellation of one remaining context released all three. A notification needs per-waiter ownership/token accounting (or per-waiter channels); a global generation cannot distinguish which waiter consumed a single Signal. (`pkg/util/fault/fault.go:177-206, 240-255`.)

3. **[P1][existing generation-safety blocker] Fence-renewal failure is still only logged.** `startISCPJobFenceLease` continues after renewal failure, so an unresolved DDL transaction cannot enforce or observe loss of quiescence. The previously requested fail-closed/handoff behavior remains absent. (`pkg/sql/compile/iscp_util.go:348-368`.)

Validation: the existing `pkg/util/fault` suite passes, the supplemental three-waiter regression fails with “one NOTIFY plus one cancellation released 3 waiters”, and `git diff --check` passes. CI for this head is still in progress.

aptend (changes requested):
Re-reviewed head `65c05b3969ab95d938e219d7ca34b44ea4bfb6ee`. The latest update correctly propagates the InitSQL transaction terminal result and replaces the shared WAIT condition with per-waiter channels, closing the two findings from the previous round. One generation-safety blocker remains.

### P1 - Transfer or fail the open-DDL fence when the executor generation changes

The local drain path captures the current `exec` pointer in its lease-renewal closure (`pkg/sql/compile/iscp_util.go:255-272`). If the ISCP executor restarts or is replaced on the same CN while the DROP transaction is unresolved, `RegisterExecutorRuntime` publishes a new executor with an empty `fencedJobs` map, but the lease continues renewing the captured old executor successfully. The replacement executor can therefore admit the still-visible, uncommitted job generation through `RegisterRunningConsumer` and start a new consumer after the original one was drained. DROP can then remove the hidden tables while that new consumer is active, reopening the race this PR is intended to close.

The remote form is also not fail-closed: a missing/replaced executor makes renewal return an error, but `startISCPJobFenceLease` only logs it and continues (`pkg/sql/compile/iscp_util.go:348-368`), so the DDL transaction cannot observe that quiescence was lost.

Please bind the lease to the current executor generation/runner and re-resolve plus re-drain/reinstall the exact fence on replacement, or propagate loss of quiescence so the DDL cannot commit. Add a deterministic regression that replaces the executor while a DROP transaction remains open and verifies the replacement cannot admit that job generation.

Fresh validation on this exact head passed the focused InitSQL/fault lifecycle tests under `-race -count=10`, the full `pkg/iscp`, `pkg/sql/compile`, `pkg/cnservice`, `pkg/queryservice/client`, and `pkg/util/fault` suites, plugin guard greps, and `git diff --check`. A supplemental executor-replacement regression fails deterministically because the new executor admits the fenced job.

jiangxinmeng1:
> Re-reviewed head `65c05b3969ab95d938e219d7ca34b44ea4bfb6ee`. The latest update correctly propagates the InitSQL transaction terminal result and replaces the shared WAIT condition with per-waiter channels, closing the two findings from the previous round. One generation-safety blocker remains.
> 
> ### P1 - Transfer or fail the open-DDL fence when the executor generation changes
> The local drain path captures the current `exec` pointer in its lease-renewal closure (`pkg/sql/compile/iscp_util.go:255-272`). If the ISCP executor restarts or is replaced on the same CN while the DROP transaction is unresolved, `RegisterExecutorRuntime` publishes a new executor with an empty `fencedJobs` map, but the lease continues renewing the captured old executor successfully. The replacement executor can therefore admit the still-visible, uncommitted job generation through `RegisterRunningConsumer` and start a new consumer after the original one was drained. DROP can then remove the hidden tables while that new consumer is active, reopening the race this PR is intended to close.
> 
> The remote form is also not fail-closed: a missing/replaced executor makes renewal return an error, but `startISCPJobFenceLease` only logs it and continues (`pkg/sql/compile/iscp_util.go:348-368`), so the DDL transaction cannot observe that quiescence was lost.
> 
> Please bind the lease to the current executor generation/runner and re-resolve plus re-drain/reinstall the exact fence on replacement, or propagate loss of quiescence so the DDL cannot commit. Add a deterministic regression that replaces the executor while a DROP transaction remains open and verifies the replacement cannot admit that job generation.
> 
> Fresh validation on this exact head passed the focused InitSQL/fault lifecycle tests under `-race -count=10`, the full `pkg/iscp`, `pkg/sql/compile`, `pkg/cnservice`, `pkg/queryservice/client`, and `pkg/util/fault` suites, plugin guard greps, and `git diff --check`. A supplemental executor-replacement regression fails deterministically because the new executor admits the fenced job.

@aptend 
Thanks for the re-review. I agree this is a remaining limitation of the current in-
  memory fence design, but I do not plan to address executor restart / lease migration in
  this PR.

  The fence here is intentionally a best-effort quiescence aid for improving DROP success
  and avoiding the known active-consumer conflict window. It is not the source of DDL
  correctness. If the fence is lost, renewal failure or a restarted executor may increase
  the chance that the foreground DDL hits the existing lock/conflict path and fails
  explicitly, but it does not make the DDL silently correct/incorrect by itself.

  Making the fence survive executor replacement would require a separate lifecycle
  protocol: generation ownership, handoff, recovery, rollback cleanup, unknown commit
  outcome handling, and stale-fence GC. Persisting or transferring this opportunistically
  in this PR would carry a higher risk of leaving a stale fence that suppresses a still-
  valid ISCP job indefinitely, which is worse than the current best-effort in-memory
  behavior.

  So for this PR I am keeping the fence process-local and transaction-scoped, and I am
  treating executor restart / daemon-task lease migration during an unresolved DDL as a
  known follow-up design item rather than extending this fix further.

aptend:
@jiangxinmeng1 Thanks for the clarification. I rechecked the generation boundary and agree the previous review over-calibrated this as a P1 correctness blocker.

The remaining factual gap is limited to executor replacement / daemon-task lease migration while a remote DDL transaction is still unresolved: the new executor can start without the old in-memory fence and may reintroduce the hidden-table lock/deadlock window. A same-instance `ISCPTaskExecutor.Restart()` preserves `fencedJobs`, and a full local CN restart also terminates that CNs DDL transaction. More importantly, the supplemental regression proved re-admission on a replacement executor, but did not prove a silent partial DDL commit or data-corruption outcome; the established failure mode is an explicit DDL conflict/deadlock and rollback.

Given that, I accept the best-effort in-memory fence scope for this PR and treat runner migration during an unresolved DDL as a non-blocking follow-up lifecycle item. The InitSQL transaction outcome, per-waiter fault cancellation, plugin dispatch, and active-runner drain paths are closed on the current head. I am updating my review accordingly.

aptend (approved):
Approved after re-review. The current implementation closes the active-runner cancellation/drain lifecycle, InitSQL transaction outcome, fault WAIT ownership, and plugin-dispatch issues. I accept executor replacement or daemon-task migration during an unresolved DDL as a documented non-blocking availability follow-up; no remaining merge blocker found. Focused race tests, affected package suites, plugin guards, diff checks, and CI pass.

gouhongshen (changes requested):
## Codex automated review

Two P1 fence-lifecycle failures can reopen the DROP INDEX race or leave a rolled-back index stale for 30 minutes.

### P1 - Fail the DDL when a fence renewal cannot be maintained (pkg/sql/compile/iscp_util.go:365)
The remote fence lives for 30 minutes, but renewal runs only every TTL/2 and a failed renewal is merely logged. A first renewal can consume its TTL/4 RPC timeout, after which the next attempt is scheduled 15 minutes later—past the original fence expiry. IsJobFenced then deletes the expired fence and the scheduler admits the consumer while the same DROP transaction may still be open, allowing it to write a hidden table that DROP is about to remove. Renewal loss must abort/rollback the DDL or retry before the current fence can expire.

### P1 - Retry or otherwise guarantee remote rollback fence removal (pkg/sql/compile/iscp_util.go:331)
On rollback the callback stops the lease and makes exactly one RemoveFenceOnly RPC. Its error is discarded by txnOperator.rollbackOwned, which invokes rollback callbacks without checking triggerEventLocked's returned error. A transient RPC or handler failure therefore leaves the job fenced even though the drop rolled back; getCandidate skips it until the 30-minute TTL expires, so subsequent writes are not propagated to the still-existing index. Add a reliable retry or compensation path rather than returning an error from a callback whose caller ignores it.

jiangxinmeng1:
> ## Codex automated review
> Two P1 fence-lifecycle failures can reopen the DROP INDEX race or leave a rolled-back index stale for 30 minutes.
> 
> ### P1 - Fail the DDL when a fence renewal cannot be maintained (pkg/sql/compile/iscp_util.go:365)
> The remote fence lives for 30 minutes, but renewal runs only every TTL/2 and a failed renewal is merely logged. A first renewal can consume its TTL/4 RPC timeout, after which the next attempt is scheduled 15 minutes later—past the original fence expiry. IsJobFenced then deletes the expired fence and the scheduler admits the consumer while the same DROP transaction may still be open, allowing it to write a hidden table that DROP is about to remove. Renewal loss must abort/rollback the DDL or retry before the current fence can expire.
> 
> ### P1 - Retry or otherwise guarantee remote rollback fence removal (pkg/sql/compile/iscp_util.go:331)
> On rollback the callback stops the lease and makes exactly one RemoveFenceOnly RPC. Its error is discarded by txnOperator.rollbackOwned, which invokes rollback callbacks without checking triggerEventLocked's returned error. A transient RPC or handler failure therefore leaves the job fenced even though the drop rolled back; getCandidate skips it until the 30-minute TTL expires, so subsequent writes are not propagated to the still-existing index. Add a reliable retry or compensation path rather than returning an error from a callback whose caller ignores it.

@gouhongshen 

 Thanks for the review. I understand both points, but I do not plan to make the in-
  memory fence lifecycle fully fail-closed in this PR.

  The cancellation fence in this patch is intentionally a best-effort quiescence
  mechanism. Its purpose is to improve DROP success by stopping the active ISCP consumer
  during the destructive DDL window. It is not intended to become a durable or
  transaction-authoritative state machine.

  For renewal failure: if the fence cannot be maintained, the system falls back to the
  pre-existing DDL/lock conflict behavior. That can increase the chance of an explicit
  DDL failure, but it does not by itself make the DDL silently commit incorrectly. Making
  renewal failure abort the DDL would require wiring an asynchronous lease goroutine into
  transaction commit/rollback semantics, which is a larger lifecycle contract than this
  PR is trying to introduce.

  For rollback cleanup: the fence has a bounded TTL specifically to avoid permanent
  suppression of a live job if rollback cleanup delivery fails. A transient stale fence
  may delay the job, but it self-recovers without manual intervention. Making cleanup
  reliable across transient RPC failures, executor replacement, unknown transaction
  outcome, and retry ownership is a separate cross-CN lifecycle design.

  So I would prefer to keep this PR scoped to cancel/drain of active consumers and
  bounded in-memory fences. A fully fail-closed or durable fence protocol should be
  handled as a follow-up design rather than added opportunistically here.

gouhongshen (approved):

PR #25741, opened 2026-07-15
fix: preserve binary semantics for prepared user variables

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25421

Preserves binary-literal metadata for user variables passed through SQL `EXECUTE ... USING`. The metadata is tracked per parameter, propagated to remote execution through `ProcessInfo`, restored by parameter expression evaluation, and retained by `EXPLAIN EXECUTE` plan substitution.

This keeps indexed and non-indexed `BINARY` predicates consistent with direct hex literals without leaking binary semantics to text parameters or binary-protocol executions. It also releases the temporary parameter vector when argument resolution fails.

Tests cover mixed binary/text parameters, repeated binary-to-text-to-binary execution, remote process codec propagation, plan substitution, error cleanup, and indexed/non-indexed BVT cases.


heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes after tracing the full frontend -> plan -> colexec -> remote-process path.

The binary metadata propagation itself is coherent and the targeted tests pass, but two production-path lifecycle problems remain:

1. EXPLAIN EXECUTE mutates the cached prepared plan in place. FillValuesOfParamsInPlan assigns copied := preparePlan and then replaces ParamRefs on that same object. A second substitution with a different value still returns the first literal; binary -> text therefore retains the old value and IsBin flag. Please deep-copy the plan before substitution and add a repeated binary -> text -> binary EXPLAIN regression test.

2. SQL EXECUTE USING creates a temporary parameter vector and transfers it to the session-level Process, but the next SetPrepareParams call overwrites the pointer without freeing the previous owned vector, and Process.Free does not release it. The new defer closes only the error path; successful executions remain unbounded over the session lifetime. Please make owned vs borrowed prepare-parameter lifetime explicit and test repeated successful replacement.

The cleanup test based on MPool.CurrNB is also not an effective proof for default heap-backed vectors because offHeap is false and those bytes are not included in CurrNB.

Both underlying weaknesses predate this PR, but this PR explicitly extends and claims closure over these paths, so they need to be made safe or explicitly split into tracked blocking follow-ups before merge.

XuPeng-SH (changes requested):
Re-reviewed current head `77d293318`. The two previous blockers are fixed: `FillValuesOfParamsInPlan` now substitutes into a deep copy, and owned prepare-parameter vectors are released on successful replacement and `Process.Free`; the remote decode path also copies protobuf backing into owned mpool memory. The affected package tests, targeted race tests, vet, and diff-check all pass.

One blocking scope regression remains:

1. **[P1] Resolve parameter metadata through the same stored-procedure scope as the value.** `buildExecuteUserParams` obtains the value with `proc.GetResolveVariableFunc()`, which intentionally checks `VarScopeKey` first, but then obtains `IsBin` with `ses.GetUserDefinedVar`. Inside a stored procedure, `EXECUTE ... USING @local_only` can therefore resolve the local value successfully and then fail with `the user variable local_only does not exist`. If a same-named session binary variable exists, the local value instead receives the wrong binary flag.

I verified this with a temporary regression test using a real `VarScopeKey`/`InSp` context; `buildExecuteUserParams` failed exactly at the Session-only metadata lookup. The existing `TestResolveVariableIsBinHonorsStoredProcedureScope` only tests the resolver in isolation and does not exercise this production function. Please use `proc.GetResolveVariableIsBinFunc()` (with the same system/global flags) when building parameters and add a `buildExecuteUserParams` or end-to-end stored-procedure EXECUTE regression covering a local-only variable and same-name shadowing.

XuPeng-SH (approved):
Re-reviewed current head `c13c0b5d95` end-to-end. The previous blockers are resolved:

- SQL EXECUTE now resolves the parameter value and its binary metadata through the same stored-procedure-aware scope, including local-only variables and session-variable shadowing.
- EXPLAIN EXECUTE substitutes into a deep-copied plan, so repeated binary/text executions cannot contaminate the cached prepared plan.
- Prepare-parameter ownership is explicit: SQL and remote-decoded vectors are owned and released on replacement/Process.Free, binary-protocol vectors remain borrowed, and protobuf backing is copied before asynchronous use.
- The binary flag is propagated across frontend, plan literals, expression evaluation, ProcessInfo protobuf, and remote CN reconstruction with backward-compatible default-false semantics.

I also checked the snapshot/binder consumers for retained aliases and the cached-compile reset path; no dangling-reference or cross-execution state issue remains. Q1-Q3 audit found no unbounded allocation, double-free, new wait edge, or unbounded accumulation. Targeted tests for vector ownership, process codec/ownership, plan immutability, colexec binary propagation, remote compile cleanup, and stored-procedure scope all pass; the focused matrix also passes under `-race -count=3`. Required CI, including Ubuntu UT, SCA, standalone multi-CN, pessimistic BVT, and multi-CN proxy BVT, is green.

Issue #25761, opened 2026-07-15
fileservice: clear IOEntry buffer ownership after disk-cache read errors


LeftHandCold:
## Summary

`IOEntry.ReadFromOSFile` deallocates a newly allocated I/O buffer when a disk-cache read or conversion fails, but leaves `IOEntry.Data` and `IOEntry.releaseData` pointing at the released allocation. A `ReadCloserForRead` created before a later conversion error can also retain a reader over the released bytes.

## Root cause

`IOEntry.prepareData` installs the allocation in the entry and sets `releaseData` (`pkg/fileservice/io_entry.go:115-139`). Its deferred error cleanup only invokes `dec.Deallocate()` (`io_entry.go:130-134`); it does not clear the owner fields.

`DiskCache.Read` treats an entry read failure as a cache miss: it logs the error and continues, then returns nil to the caller (`pkg/fileservice/disk_cache.go:393-418`). The same `IOEntry` is subsequently reused by the origin-read path with stale ownership state.

This violates the ownership contract: after deallocation, no reachable owner or reader may still reference the allocation.

## Impact

- a later `releaseData` call can deallocate the same allocation twice;
- fallback code can read or reuse released memory;
- a retained `ReadCloserForRead` can expose bytes backed by the released buffer;
- allocator-dependent outcomes include crashes, memory corruption, or silent bad data.

## Reproduction

Use a truncated range-cache file or force `WriterForRead` / `ToCacheData` to fail after `prepareData` allocates. After `DiskCache.Read` returns, inspect the entry: the current code leaves non-nil ownership fields for memory that has already been released.

Deterministic regression tests are available on `LeftHandCold:agent/fileservice-disk-cache-read-safety`.

## Proposed fix

Make error cleanup local and idempotent:

1. deallocate at most once;
2. clear `Data` and `releaseData` when the allocation is released;
3. clear any `ReadCloserForRead` installed over the failed buffer;
4. preserve the successful-read ownership path unchanged.

## Acceptance criteria

- short read, writer failure, and conversion failure leave no released owner reachable from `IOEntry`;
- repeated cleanup is safe;
- origin fallback still succeeds;
- targeted tests pass under `-race`.

Issue #25762, opened 2026-07-15
fileservice: seek reused range-cache files from file offset zero


LeftHandCold:
## Summary

When `DiskCache.Read` reuses an already opened range-cache file, it seeks to the logical object offset instead of file offset zero. Range-cache files contain only the cached interval, so their first byte is always at file offset zero.

## Root cause

Range entries are written from `entry.Data` into a standalone file named with the logical offset and size (`pkg/fileservice/disk_cache.go:445-461`, `619-626`). On the first open, reading starts at file offset zero. When the same range file is reused from `openedFiles`, the code executes `f.Seek(entry.Offset, io.SeekStart)` (`disk_cache.go:305-314`).

That seek is valid for a full-object cache file, but invalid for a range-cache file. For any non-zero logical offset it skips past the beginning of the standalone interval and commonly produces a short read.

## Impact

- deterministic cache-read errors for repeated entries sharing the same range-cache file;
- unnecessary origin/S3 fallback and degraded cache hit rate;
- interaction with error-path ownership bugs can make the failure unsafe rather than merely slow.

## Reproduction

Cache a range with a non-zero object offset, then read two equivalent entries in one vector so the second read reuses the opened range file. The second entry reads from the wrong physical offset on the current code.

A regression test is available on `LeftHandCold:agent/fileservice-disk-cache-read-safety`.

## Proposed fix

Seek reused range-cache files to physical offset zero. Continue seeking full-object cache files to `entry.Offset`.

## Acceptance criteria

- repeated reads from the same non-zero logical range succeed from DiskCache;
- full-object cache reads retain their existing offset behavior;
- regression test passes under `-race`.

PR #25767, opened 2026-07-15
fix(fileservice): make disk cache reads ownership-safe

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25761
issue #25762

`IOEntry.prepareData` released a newly allocated I/O buffer when a disk-cache read or conversion failed, but kept `Data`, `releaseData`, and potentially `ReadCloserForRead` pointing at that released allocation. Because `DiskCache.Read` treats an entry error as a cache miss and lets origin reading continue with the same entry, the stale owner could cause a double release or use-after-free.

This change makes the error cleanup idempotent, clears all ownership fields after release, and clears a reader whose backing buffer became invalid.

The PR also fixes reused standalone range-cache files being seeked to the logical object offset. A range-cache file contains only the interval bytes and must be read from physical file offset zero; full-object cache files continue to seek to `entry.Offset`.

Regression tests cover short reads, repeated cleanup, cache conversion failure, reader cleanup, and reuse of a non-zero-offset range file.

Tested with:

`go test -race -mod=mod ./pkg/fileservice -run 'TestIOEntryReadErrorClearsAllocatedDataOwner|TestIOEntryAllocatedDataErrorCleanupIsIdempotent|TestIOEntryCacheConversionErrorClearsAllocatedReader|TestDiskCacheReusesRangeFileFromStart' -count=1`

`go test -mod=mod ./pkg/fileservice/... -count=1`


XuPeng-SH (approved):
Re-reviewed the latest head `1514d65e2d`. The two ownership fixes look coherent: range-cache files are now always re-read from offset 0, which matches how `DiskCache.Update` materializes per-entry cache files, and the `IOEntry.prepareData` cleanup path now releases internally allocated buffers/readers only on error and does so idempotently. I also reran the focused fileservice lifetime tests added here and do not see a remaining high-confidence correctness, leak, hang, or OOM blocker on this head.

Issue #25763, opened 2026-07-15
fileservice: canonicalize cache keys across full writes, reads, and deletes


LeftHandCold:
## Summary

Full-object cache reads, writes, and deletes do not derive the same canonical key. Service-qualified and leading-slash paths can therefore be written under a key that later reads cannot find, or deleted under a key that does not remove the cached entry.

## Root cause

Range read/update paths call `ParsePath` and use `path.File`. `ParsePath` strips the service qualifier and leading slash (`pkg/fileservice/path.go:46-83`).

In contrast:

- S3 full-object paths pass raw `vector.FilePath` to `DiskCache.SetFile` (`pkg/fileservice/s3_fs.go:1003-1013`, `1043-1061`);
- `DiskCache.SetFile` derives the physical key directly from its input (`pkg/fileservice/disk_cache.go:757-768`);
- `DiskCache.DeletePaths` and `MemCache.DeletePaths` also consume raw input paths (`disk_cache.go:770-779`, `mem_cache.go:374-379`).

For example, `shared:/foo` and `/foo` canonicalize to `foo` on the read path, while the current full-object write/delete paths use the original strings.

## Impact

- a successful full-object cache write can be unreachable by subsequent reads;
- delete can leave MemCache or DiskCache entries reachable under the canonical key;
- unreachable disk files consume quota until eviction;
- mutable or reused paths can return stale data after delete.

## Proposed fix

Canonicalize at every public cache boundary using `ParsePath(...).File`:

1. canonicalize `DiskCache.SetFile` input;
2. canonicalize the complete path list before mutating DiskCache or MemCache during delete;
3. reject invalid paths before partial deletion begins;
4. keep range and full-object keys in the same namespace.

This does not solve deletion of range-cache variants; that requires a separate reverse index or bounded scan.

## Acceptance criteria

- `shared:/foo`, `/foo`, and `foo` address the same full-object cache entry;
- all three forms delete the same MemCache and DiskCache entry;
- invalid path lists fail before any cache mutation;
- service-qualified and leading-slash regression tests pass under `-race`.

An implementation and tests are available on `LeftHandCold:agent/fileservice-canonical-cache-keys`.

PR #25768, opened 2026-07-15
fix(fileservice): canonicalize cache keys at cache boundaries

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25763

Full-object cache writes, reads, and deletes used different keys for service-qualified or leading-slash paths. Read/update paths already use `ParsePath(...).File`, while S3 full-object writes passed raw `vector.FilePath` to `DiskCache.SetFile` and cache deletion consumed raw paths. For example, a cache entry written as `shared:/foo` could not be found by a later read using canonical key `foo`, and deleting `/foo` could leave the canonical entry behind.

This change canonicalizes `DiskCache.SetFile`, `DiskCache.DeletePaths`, and `MemCache.DeletePaths` at their public cache boundaries. Delete canonicalizes the complete input list before mutation, so an invalid path cannot cause a partially applied deletion.

Regression tests cover cross-form full-object reads/deletes and memory-cache deletion for service-qualified and leading-slash paths. They also verify that invalid full-file writes do not invoke the reader and that mixed valid/invalid delete lists fail without removing an existing entry. Deleting standalone range-cache variants remains separate work because it needs a reverse index or bounded scan.

Tested with:

`go test -race -mod=mod -count=10 -timeout 180s -run "TestDiskCacheCanonicalizesFullFileKeys|TestDiskCacheRejectsInvalidFullFilePath|TestDiskCacheDeletePathsRejectsInvalidListAtomically|TestMemCacheCanonicalizesDeletePaths|TestMemCacheDeletePathsRejectsInvalidListAtomically" ./pkg/fileservice`

`go test -race -mod=mod -count=1 -timeout 15m ./pkg/fileservice/...`

`go build -mod=mod ./pkg/fileservice/...`

`go vet -mod=mod ./pkg/fileservice/...`


XuPeng-SH (changes requested):
The production fix looks correct, but the required Coverage check is failing for a real, in-scope gap rather than infrastructure noise.

The uploaded diff coverage artifact shows that the new parse-error branches are uncovered: `DiskCache.SetFile` is 80%, `DiskCache.DeletePaths` is 80%, `MemCache.DeletePaths` is 80%, and `canonicalFilePaths` is 85.7% (four uncovered statements out of 22, about 81.8% overall). These are exactly the unhappy paths behind the PR's stated guarantee that an invalid path list is rejected before any cache mutation. The current tests exercise only valid service-qualified and leading-slash paths, so that atomicity contract is not verified.

Please add public-boundary regressions that prove at least: (1) an invalid `SetFile` path returns an invalid-path error without invoking `openReader`; and (2) `DiskCache.DeletePaths` and `MemCache.DeletePaths` given a list containing a valid cached path followed by an invalid path return an error while leaving the valid entry intact. This covers the behavior rather than testing the helper implementation and should restore the required coverage gate. A cross-form assertion (for example, set with `shared:/foo`, then read/delete with `foo`) would also make the canonical-key equivalence contract more explicit.

I found no blocking production-code issue otherwise. Canonicalization happens before mutation; disk update ownership, readers, temporary files, and path locks retain their existing guaranteed cleanup; no new unbounded wait or accumulation is introduced; and the extra parsing/allocation is bounded by the caller's path list and negligible relative to cache I/O. Targeted race tests repeated 20 times, the full fileservice test suite, `go build`, and `go vet` all pass locally. The latest-main disk-cache change is in a different hunk and merge-tree combines the two cleanly.

XuPeng-SH (approved):
Re-reviewed the latest head `b03b739167`. The prior unhappy-path/coverage request is addressed at the public cache boundaries:

- invalid `SetFile` input returns before `openReader` or cache mutation;
- DiskCache and MemCache validate the complete delete list before deleting anything;
- the tests prove atomic rejection by reading the valid cached entry after a valid+invalid batch;
- service-qualified and leading-slash forms are verified against the canonical key used by read/update paths.

The production change keeps key generation consistent with existing `Read`/`Update` (`ParsePath(...).File`), preserves disk update locks and reader/temp-file cleanup, and introduces no new unbounded work. The O(n) pre-validation allocation is bounded by the caller list and is appropriate for the atomicity guarantee. Standalone range-cache deletion remains explicitly out of scope and is not regressed.

Local validation on this head:
- focused canonicalization/error/atomicity tests under `-race -count=10`: PASS
- `go test -mod=mod -count=1 ./pkg/fileservice/...`: PASS
- `git diff --check`: PASS

No remaining blocking correctness, robustness, or performance issue found. Remote CI is still running and should remain a merge prerequisite.

Issue #25765, opened 2026-07-15
gossip: avoid empty entries in cache-key broadcast batches


LeftHandCold:
## Summary

`delegate.broadcastData` preallocates a slice with `len(items)` elements and then appends the same number of real elements. Every cache-key batch therefore starts with empty `GossipData` entries.

## Root cause

The function uses:

```go
data := make([]gossip.GossipData, len(items))
for _, item := range items {
    data = append(data, ...)
}
```

at `pkg/gossip/delegate.go:116-145`. The first `len(items)` entries remain zero values. The intended construction is a zero-length slice with capacity `len(items)`.

## Impact

- cache-key broadcasts contain empty protobuf messages;
- batching limits and serialized network work no longer correspond to the number of actual updates;
- receivers spend resources on invalid/no-op entries;
- when the cache-key batch consumes the limit, the returned slice is exactly twice the intended length.

The queue itself has a hard item-count limit (`pkg/gossip/base_item.go:23-80`), so this issue is not claiming unbounded growth. Queue byte accounting/coalescing is a separate design concern.

## Proposed fix

Construct the slice as `make([]gossip.GossipData, 0, len(items))` and retain the existing append logic.

## Acceptance criteria

- returned batch length equals the number of real cache-key plus stats updates;
- no element has a nil/empty payload;
- mixed cache-key/stats and limit-boundary tests pass under `-race`.

An implementation and tests are available on `LeftHandCold:agent/gossip-broadcast-lock-fixes`.

Issue #25766, opened 2026-07-15
gossip: lock statsInfoKey while building stats state


LeftHandCold:
## Summary

`delegate.statsInfoState` locks the cache-key store mutex while iterating the stats-info store map. Concurrent stats updates therefore access the map without a common lock and can race or panic.

## Root cause

`statsInfoState` locks `d.dataCacheKey.mu` but iterates `d.statsInfoKey.mu.keyTarget` (`pkg/gossip/delegate.go:183-193`). Stats updates correctly mutate the map while holding `d.statsInfoKey.mu` (`delegate.go:103-112`). The reader and writer consequently synchronize on different mutexes.

This is on the normal CN path, not only the optional remote file-cache path: standard `mo-service` creates a gossip node, CN query setup calls `Create`, and disttae uses the stats key router.

## Impact

- Go map read/write data race;
- possible `concurrent map iteration and map write` process panic during gossip LocalState generation;
- inconsistent stats routing snapshots.

## Proposed fix

Lock `d.statsInfoKey.mu` for the complete map traversal. Keep cache-key state protected by `d.dataCacheKey.mu`.

## Acceptance criteria

- `statsInfoState` blocks on the stats store mutex, not the cache-key mutex;
- concurrent stats update/LocalState tests pass under `-race`;
- cache-key and stats snapshots retain their current wire format.

An implementation and tests are available on `LeftHandCold:agent/gossip-broadcast-lock-fixes`.

PR #25769, opened 2026-07-15
fix(gossip): correct broadcast batches and stats locking

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25765
issue #25766

`delegate.broadcastData` created its output slice with `len(items)` elements and then appended the real cache-key updates. Each batch was therefore prefixed with empty `GossipData` messages, doubled the expected cache-key batch length, and distorted batching and serialization work. This change creates a zero-length slice with the required capacity and adds exact-length assertions.

`delegate.statsInfoState` also locked `dataCacheKey.mu` while iterating `statsInfoKey.mu.keyTarget`. Concurrent stats updates use `statsInfoKey.mu`, so the previous reader and writer synchronized on different mutexes and could race or panic with concurrent map iteration/write. This change locks the stats store mutex and adds a synchronization regression test.

The PR does not redesign the bounded gossip queue; byte-based limits, key coalescing, and dropped/lag metrics remain follow-up design work.

Tested with:

`go test -race -mod=mod ./pkg/gossip/... -count=1`


XuPeng-SH (changes requested):
The production changes look correct, but the concurrency regression test is not deterministic enough to protect this severity/s0 fix.

`TestDelegateStatsInfoStateUsesStatsMutex` closes `started` before calling `statsInfoState`, then treats 20 ms without completion as proof that the goroutine is blocked on `statsInfoKey.mu`. If the goroutine is descheduled immediately after closing `started`, the old implementation can survive that 20 ms window; the test then unlocks the stats mutex, the old implementation runs without touching it, and the test passes. Conversely, the one-second completion bound can fail a correct implementation on a sufficiently slow CI worker. The timeouts are establishing the causal order instead of merely bounding a deterministically constructed test.

Please make the lock discrimination deterministic without changing production code. For example, hold `dataCacheKey.mu` continuously, also hold `statsInfoKey.mu` while starting the call, then release only `statsInfoKey.mu`: the fixed implementation must complete regardless of when the goroutine is scheduled, while the old implementation remains blocked on the still-held cache mutex. Use a generous timeout only as a terminal CI guard, and ensure the cache mutex is released and the helper goroutine is collected before any fatal assertion. This directly proves the independence contract and is stable on slow CI.

I otherwise found no blocking production issue: the broadcast slice fix restores exact payload cardinality and reduces allocation/serialization work; all stats map readers and writers now use the same mutex; no nested lock ordering, unbounded wait, or new accumulation is introduced. Targeted race tests repeated 100 times, full `go test -race ./pkg/gossip/...`, `go build`, and `go vet` all pass on the current head.

XuPeng-SH (approved):
The latest test change resolves the only blocking review finding, and I found no remaining correctness, concurrency, performance, or unhappy-path blocker.

The mutex regression is now causally deterministic: it holds both store mutexes, releases only `statsInfoKey.mu`, and keeps `dataCacheKey.mu` held. The fixed implementation completes under every goroutine scheduling order, while the old implementation remains blocked on the cache mutex until the terminal guard releases it. The timeout now bounds CI failure rather than establishing the tested phase, and both paths release the held lock and collect the helper goroutine. I also restored the old production lock locally and confirmed that this test fails with `statsInfoState blocked on dataCacheKey`; the PR implementation passes the focused set under `-race -count=100`.

The production changes are correct and minimal: `make([]gossip.GossipData, 0, len(items))` restores one output per real queue item and removes empty-message allocation/serialization amplification, while `statsInfoState` now uses the same mutex as every stats map writer and reader. There is no new lock nesting or wait cycle. Q1 introduces no new resource owner, Q2 makes all stats map access synchronize on one lock with no unbounded dependency, and Q3 preserves the bounded queue while reducing output amplification. The changes are off the SQL/data hot path except for less gossip allocation and serialization work, so the performance effect is non-negative.

Validation on `677931445599b6561580fca6836890e9957cc131`: focused race stress x100; an adversarial old-implementation negative control; a temporary mixed cache/stats + exact-limit + non-nil payload matrix under race x20; complete `pkg/gossip/...` tests and race tests; `go build`; `go vet`; PR diff check; and a clean virtual merge with latest main followed by complete gossip tests, race tests, and vet. All CI checks are complete and green.

PR #25771, opened 2026-07-15
fix(morpc): wake waiters when backend creation completes

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #25650 #25770

The MORPC recovery changes for CN hot recreation moved backend factory I/O out of the client mutex and deduplicated creation per remote generation. This preserved reset responsiveness and stale-generation safety, but callers waiting for that asynchronous result could only poll with exponential backoff. A backend that became ready just after a retry timer started remained unused until the timer expired. Multi-CN queries amplify this delay because remote pipelines repeatedly acquire locked streams.

This PR makes backend creation completion event-driven while preserving the recovery guarantees:

- associates each queued or in-flight create with a generation-scoped completion signal;
- makes `Send`, `NewStream`, and `Ping` wait for the actual create state transition instead of sleeping past a ready backend;
- publishes a new backend before waking waiters;
- wakes waiters when reset, generation eviction, or client close invalidates their create;
- preserves exponential backoff after factory failures to avoid a retry storm;
- retains context cancellation, the bounded auto-create timeout, create deduplication, and stale-generation rejection.

Validation:

- Deterministic regression: clean `origin/main` slept on retry backoff and failed the 1-second assertion; this change completed immediately.
- `go test -mod=mod -count=1 -timeout 120s ./pkg/common/morpc`
- Race-enabled create/reset/dedup and new regression tests.
- Local TPCH 10G with one TN and two CN processes, six warm runs:

| Query | `origin/main` | This PR | Speedup |
| --- | ---: | ---: | ---: |
| q2 | 1.567 s | 0.342 s | 4.58x |
| q5 | 2.812 s | 0.975 s | 2.88x |

The benchmark used the same LOG/TN processes and persisted data while switching the two CN binaries. Switching back to `origin/main` reproduced the slowdown, ruling out benchmark order and warm-cache effects.


XuPeng-SH (commented):
Re-reviewed the latest head `8ba0cdb7c1`. The new `backendCreateState` wakeup path looks coherent: waiters now block on a real creation-completion signal instead of sleeping through long retry backoff, failed creates still retain throttling via the `failed` bit, and targeted backend resets invalidate both the generation token and any queued create state before replacement. I also reran the focused backend-creation regressions in `pkg/common/morpc`, and I do not see a remaining high-confidence correctness, leak, hang, or OOM blocker on this head. GitHub does not allow me to approve my own pull request, so I am leaving this as an approve-level review comment instead of an approval.

LeftHandCold (approved):
Reviewed head `8ba0cdb7c1`. I traced the generation-scoped create state through queue admission, factory success/failure, queue-full fallback, targeted reset, generation eviction, client close, and the unregistered-GC path. Each admitted state is removed and its completion channel is closed; waits remain bounded by caller context / auto-create timeout, failed creates retain retry backoff, and successful backends are published before waiters are woken. Full `pkg/common/morpc` tests, the full race suite, and 50 repeated runs of the new success/failure regressions passed locally. I do not see a high-confidence correctness, leak, hang, OOM, or normal-business regression blocker on this head.

XuPeng-SH:
性能回来了：https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29400935065/job/87309039831

XuPeng-SH:
Fixed the `TestCloseIdleBackends` failure in `ffce45e49b`.

This was not introduced by the first PR commit. The same race reproduces on clean `origin/main@2d01d204b2` with `-race -count=5000` (both `ErrBackendClosed` and `ErrBackendCreating`). The failure exposed an existing lookup/create-admission TOCTOU:

1. a lookup observed no available backend under `c.mu`;
2. it released `c.mu` before submitting creation;
3. the first asynchronous create published an available backend in that gap;
4. the stale lookup still submitted a second create;
5. the test's explicit second create then raced with the speculative create and saw either an in-flight claim or a full pool.

The fix makes backend selection and create admission one atomic state transition under `c.mu`. The create-channel send remains non-blocking, and factory/DNS/connection I/O remains outside the lock. Queue-full fallback now claims the same observed demand before unlocking, so publication cannot slip between observation and ownership. The obsolete post-unlock fallback-state lookup was removed.

Validation:
- clean main reproducer: failed with the same two errors;
- fixed branch: `TestCloseIdleBackends -race -count=5000` passed;
- focused success/failure/queue-full/reset/idle matrix: normal `count=100`, race `count=50`, shuffled;
- full `pkg/common/morpc` ordinary and race suites;
- `go vet` and `go build` for `pkg/common/morpc`;
- full downstream `pkg/lockservice` suite;
- hot available-backend benchmark versus prior PR head: median approximately 102 ns/op → 103 ns/op, unchanged 128 B/op and 1 alloc/op (noise-level difference);
- `git diff --check`.

New CI run: https://github.com/matrixorigin/matrixone/actions/runs/29405085709

XuPeng-SH (commented):
Re-reviewed the latest head `ffce45e49b`, focusing on the new atomic create-admission change. The revised `getBackendForOperation` now claims either the async queue slot or the synchronous `creating[backend]` slot while `c.mu` still protects the observed pool state, so the queue-full fallback no longer races a successful publication between demand observation and local create admission. I reran the focused `pkg/common/morpc` regression that exercises `CloseBackendFor` against a concurrent synchronous create, and I do not see a new high-confidence correctness, leak, hang, or OOM blocker in the latest delta. GitHub still does not allow me to approve my own pull request, so I am leaving this as an approve-level review comment rather than an approval.

PR #25772, opened 2026-07-15
fix: scope INNER JOIN ON subquery binding

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24996

Follow-up to #25339 and the review findings on #25359.

#25339 enabled `TableBinder.BindSubquery` globally, while subquery flattening exists only for `INNER JOIN ... ON`. This also allowed unsupported subqueries through outer-join ON conditions, table-function arguments, and `result_scan`, where an executable `Expr_Sub` could survive into compilation.

This PR adds a narrowly enabled `NewJoinOnBinder` used only for INNER JOIN ON predicates. Other `TableBinder` callers continue rejecting subqueries. The existing INNER JOIN flattening path and aggregate/window binding semantics are preserved.

Planner coverage now includes correlated, uncorrelated, and mixed ON predicates; asserts that no `Expr_Sub` survives in JOIN/FILTER expressions; and verifies that outer-join ON subqueries remain rejected.

Based directly on upstream main `2d01d204b`.

Verification:

- `make build`
- `join.sql` BVT: 202/202, 100%

The equivalent correction is included in the 4.1 backport PR #25359.


XuPeng-SH (approved):
Reviewed current head `0cbe5fdf` against the latest main and the original #25339/#25359 failure modes. This delta correctly restores the default fail-closed behavior for ordinary `TableBinder` callers and enables subquery binding only for INNER JOIN ON predicates, where the existing lowering path removes executable `Expr_Sub` values before compilation. Outer joins, table functions, `result_scan`, and the DML helper call sites continue using the rejecting binder.

I also exercised correlated IN, correlated EXISTS, correlated scalar, and uncorrelated scalar ON predicates against the optimized plan; they planned successfully and no `Expr_Sub` remained in JOIN/FILTER expressions. `TestSubqueryInJoinOn` passed 20 consecutive runs, the full `pkg/sql/plan` package passed, the merge with current main is clean, and all required CI checks are green.

Non-blocking follow-up: the already-merged #25339 lowering still leaves an empty-OnList INNER JOIN for a scalar-only ON predicate and filters after subquery flattening, so the common “join to per-group max” shape can form a Cartesian product. That behavior exists in the base branch and is neither introduced nor worsened by this scope-correction PR, so it should be tracked and optimized separately rather than delaying this safety fix.

PR #25775, opened 2026-07-15
fix(proxy): report current connections with a gauge

### LINKED ISSUE #25738: [Bug] Proxy current-connection metric counts closed connections cumulatively
## Summary

The Prometheus metric labeled `current` does not report active Proxy connections. It is a Counter that is incremented when a connection handler exits and is never updated when a connection enters.

## Affected revision

## Root cause

`pkg/proxy/handler.go` correctly updates the internal gauge-like counter:

```go
h.counterSet.connTotal.Add(1)
defer func() {
    v2.ProxyConnectCurrentCounter.Inc()
    h.counterSet.connTotal.Add(-1)
}()
```

But `ProxyConnectCurrentCounter` is created from a Prometheus `CounterVec` in `pkg/util/metric/v2/proxy.go`. It cannot decrement, and there is no matching update at connection acceptance.

## Actual behavior

`mo_proxy_connect_counter{type="current"}` increases whenever a connection closes. It reports cumulative handler exits rather than current connections, producing incorrect dashboards and alerts during connection churn.

## Expected behavior

Expose active connections as a Gauge, incrementing on entry and decrementing on every exit path. If the existing Counter is intended to count closed connections, rename it and add a separate active-connection Gauge.

## Suggested regression test

Open N handlers, verify the gauge is N, close them one by one, and verify it returns to zero. Include early failures before backend establishment.

## Relevant code

- `pkg/proxy/handler.go:155-163`
- `pkg/util/metric/v2/proxy.go:20-32`

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25738

`mo_proxy_connect_counter{type="current"}` was a Counter incremented when a Proxy connection handler exited, so it reported cumulative closes rather than active connections. Dashboards applying `rate()` to that series therefore displayed the wrong meaning.

Changes:
- add `mo_proxy_connections_current` as a Gauge, incremented when a handler starts and decremented from its top-level defer on every exit path
- preserve cumulative close events as `mo_proxy_connect_counter{type="closed"}`
- register the new Gauge and update the Proxy Grafana dashboard to query the Gauge directly
- query both the legacy `current` close-event label and the new `closed` label in the close-rate panel during rolling upgrades
- add handler lifecycle tests for `3 -> 2 -> 1 -> 0`, early backend-establishment failure, and cumulative close counting
- add metric scrape and dashboard query tests that lock the metric types, names, registration, and migration selector

Compatibility and migration:
- existing dashboards and alerts that use `mo_proxy_connect_counter{type="current"}` as an active-connection metric must migrate to `mo_proxy_connections_current`
- the Proxy close-rate panel queries `type=~"current|closed"` so old and upgraded Proxy instances remain visible during a rolling upgrade
- during a rolling upgrade, the new Gauge only covers upgraded Proxy instances, so aggregate active-connection visibility is temporarily partial

Non-goals:
- do not dual-write the misleading legacy `type="current"` series from upgraded Proxy instances
- do not infer active connections from the legacy cumulative Counter

Validation:
- `go test ./pkg/util/metric/v2/dashboard -short -count=1`
- `go test ./pkg/util/metric/v2 -run TestProxyConnectionMetricTypes -count=1`
- `go test ./pkg/proxy -run 'TestHandler(CurrentConnections|_Handle|_HandleErr)$' -count=1`
- `go test -race ./pkg/util/metric/v2/dashboard -run TestProxyConnectionDashboardSupportsRollingUpgrade -count=1`
- `go test -race ./pkg/proxy -run 'TestHandler(CurrentConnections|_Handle|_HandleErr)$' -count=1`
- `mo-self-review` full diff gate, including Q1-Q3 lifecycle checks


XuPeng-SH (approved):
Reviewed the current head (6ca2eafd). The active-connection Gauge is incremented only after handler entry and decremented by the top-level defer on every normal and early-error exit, while cumulative closes retain Counter semantics under the new closed label. Metric registration and dashboard queries are consistent, and the rolling-upgrade selector does not double-count because old and new proxy binaries emit disjoint labels. I found no blocking correctness, lifecycle, cardinality, performance, or test-stability issue. Local validation passed: focused tests x10, focused race tests, full proxy/metric packages, short dashboard package, build, vet, diff check, and a clean merge against current main.

PR #25777, opened 2026-07-15
fix(proxy): reject short backend handshake responses

### LINKED ISSUE #25736: [Bug] Proxy panics on a short backend handshake response
## Summary

Proxy detects a backend CN handshake response shorter than the required MySQL packet prefix, closes the backend connection, but then continues slicing the response at `r[4:]`. A response with fewer than four bytes causes a runtime slice-bounds panic.

## Affected revision

## Root cause

In `pkg/proxy/client_conn.go`, the malformed-response branch does not return or retry:

```go
if len(r) < 5 {
    // log and close sc
}
sc.SetConnResponse(r[4:])
if err := c.sendPacketToClient(r[4:], sc); err != nil {
    return nil, err
}
```

## Actual behavior

- `len(r) == 0..3`: `r[4:]` panics.
- `len(r) == 4`: processing continues with an empty payload even though the response was rejected and the backend was closed.
- A truncated CN handshake can therefore turn a backend failure into a Proxy failure.

## Expected behavior

Reject the malformed response with a normal error, close the backend exactly once, and either retry another CN or return the connection error. Malformed backend input must not panic Proxy.

## Suggested regression test

Make the router return handshake responses of lengths 0 through 4 and assert that `connectToBackend` returns an error without panic, closes the failed backend, and forwards no packet.

## Relevant code

- `pkg/proxy/client_conn.go:672-688`

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25736

When a backend CN returned a handshake response shorter than the MySQL packet prefix, Proxy closed the backend connection but continued to slice and forward `r[4:]`. Responses of length 0–3 panicked at the slice, while length 4 passed an empty payload into the frontend write path and also panicked.

Changes:
- return a controlled error immediately after closing a backend that returned a response shorter than five bytes
- count the malformed handshake as an existing `common-fail` connection event
- prevent the invalid response from being cached or forwarded to the client
- add regression coverage for response lengths 0 through 4, including no panic, one backend close, no cached response, no client write, and the failure metric

Non-goals:
- do not change the normal handshake path
- do not change the existing retry and health-ejection semantics for transport connection errors
- do not add an end-to-end faulty-CN deployment test in this PR

Validation:
- `go test ./pkg/proxy -run '^Test_connectToBackend_ShortHandshakeResponse$' -count=1`
- `go test -race ./pkg/proxy -run '^Test_connectToBackend_ShortHandshakeResponse$' -count=1`
- `go vet ./pkg/proxy`
- gofmt and `git diff --check`
- `mo-self-review` full diff gate, including connection-close Q1 and failure-metric closure


XuPeng-SH (approved):
Reviewed the current head (cea2b8b6). The length check is placed before every slice, response-cache mutation, and client write, and the malformed backend connection is closed and untracked before returning a controlled error. The 0–4 byte table covers the full invalid boundary and asserts the observable contract: no panic, exactly one close, no cached response, no client write, and one failure event. I found no blocking correctness, resource-lifecycle, retry-policy, performance, or test-stability issue. Local validation passed: focused test x20, focused race test, full proxy package, build, vet, diff check, and a clean virtual merge with current main.

PR #25778, opened 2026-07-15
fix(proxy): retry HAKeeper errors during bootstrap

### LINKED ISSUE #25737: [Bug] Proxy bootstrap panics instead of retrying transient HAKeeper errors
## Summary

Proxy bootstrap has a five-minute retry loop, but any single `GetClusterState` error panics immediately. A temporary HAKeeper timeout or network interruption can therefore terminate Proxy instead of being retried.

## Affected revision

## Root cause

`pkg/proxy/server.go` starts bootstrap asynchronously with `go h.bootstrap(ctx)`. Inside `pkg/proxy/bootstrap.go`:

```go
for retry < int(BootstrapTimeout/BootstrapInterval) {
    state, err = h.haKeeperClient.GetClusterState(ctx)
    if err != nil {
        panic(moerr.AttachCause(ctx, err))
    }
}
```

The panic makes the surrounding retry budget ineffective. An unrecovered panic in this goroutine terminates the process.

## Actual behavior

The first transient HAKeeper error during task-table credential discovery causes a panic. The configured five-minute bootstrap retry window is not used.

## Expected behavior

Retry transient errors until bootstrap succeeds, the parent context is canceled, or the bootstrap timeout expires. Report a controlled startup failure at timeout rather than panicking from an unmanaged goroutine.

## Suggested regression test

Use a mock HAKeeper client that fails the first request and succeeds on the second. Assert bootstrap completes without panic. Add cancellation and permanent-failure cases.

## Relevant code

- `pkg/proxy/bootstrap.go:32-64`
- `pkg/proxy/server.go:108`

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25737

Proxy bootstrap had a five-minute retry window, but a single HAKeeper `GetClusterState` error panicked from an unmanaged goroutine and bypassed that window.

Changes:
- Retry returned `GetClusterState` errors within the existing five-minute bootstrap window.
- Return controlled cancellation/timeout errors from bootstrap instead of panicking.
- Run bootstrap as a Stopper-managed task so caller and server shutdown cancellation terminate it.
- Add regression coverage for a transient failure followed by success, parent cancellation, actual Stopper shutdown cancellation, and persistent failure/timeout.
- Use generous deadlock guards in asynchronous tests instead of machine-speed assertions.

Semantics:
- Success: credentials are present, the SQL writer and worker configuration is installed, and bootstrap returns nil.
- Cancellation: parent or Stopper cancellation stops the current request and returns the cancellation error without logging a startup failure.
- Transient error: the error is retained for diagnostics and the next interval retries.
- Persistent failure/timeout: the task returns a deadline error joined with the last HAKeeper error, and the managed task logs the controlled failure.

Decision / Non-goals / Risks:
- Proxy startup remains asynchronous. On bootstrap timeout, Proxy remains running, the SQL worker stays unready, and bootstrap does not retry past the existing five-minute window.
- This preserves the current Server/Stopper contract: Stopper tasks do not return errors to `NewServer`, and changing process/readiness policy is outside this fix.
- `MOCluster.Close` waiting for an independent first `GetClusterDetails` refresh is a pre-existing clusterservice lifecycle concern and is not changed here.

Validation:
- `go test ./pkg/proxy -run '^TestBootstrap' -count=20`
- `go test -race ./pkg/proxy -run '^TestBootstrap' -count=1`
- `go vet ./pkg/proxy`
- `mo-self-review` full diff gate, including Q1-Q3 lifecycle checks


XuPeng-SH (changes requested):
The production change looks sound: transient HAKeeper failures are retried within a real wall-clock bootstrap deadline, request/parent cancellation is propagated, and the task/ticker/error state all have bounded lifecycles. The focused tests, race run, full proxy package, and CI are green.

One blocking regression-coverage gap remains. The lifecycle behavior added in server.go is not tested: TestBootstrapReturnsContextCancellation calls bootstrapWithTimeout with a directly canceled parent context, so it bypasses the new Stopper taskCtx -> context.AfterFunc -> bootstrapCtx chain. If that wiring is removed or broken later, all new tests still pass while Server.Close can fail to terminate an in-flight bootstrap through the intended managed-task path. Please add a deterministic test around the actual Stopper task (or Server.Close): wait until the mock GetClusterState request has started, invoke Stopper.Stop/Close, then prove the request context is canceled, shutdown returns, and the task reaches its terminal state.

Please also make the new tests slow-CI safe while updating them. TestBootstrapReturnsTimeoutAfterPersistentHAKeeperError asserts completion in less than one second, which is a scheduler/performance assertion unrelated to the functional contract and can flake on a loaded runner. Remove that wall-clock assertion and use only a generous timeout as a deadlock guard; likewise, the one-second channel guards in the cancellation test should be generous guards rather than tight scheduling requirements. The 50 ms bootstrap timeout can remain the semantic stimulus—the test should assert DeadlineExceeded plus the retained HAKeeper error, not machine speed.

VioletQwQ-0:
@XuPeng-SH Addressed the requested regression coverage and slow-CI safety changes in `e4d9edb63`.

Changes:
- extracted `runBootstrapTask` so production startup and the regression test execute the same `Stopper taskCtx -> context.AfterFunc -> bootstrapCtx` cancellation chain
- added a deterministic Stopper shutdown test that waits for `GetClusterState` to start, calls `Stopper.Stop`, verifies the in-flight request context is canceled, and verifies `Stop` returns after the task reaches its terminal state
- replaced the one-second channel guards with ten-second deadlock guards
- removed the elapsed-time performance assertion from the persistent-error test while retaining the 50 ms semantic timeout and assertions for `DeadlineExceeded` plus the retained HAKeeper error

Validation:
- `go test ./pkg/proxy -run '^TestBootstrap' -count=20`
- `go test -race ./pkg/proxy -run '^TestBootstrap' -count=1`
- `go vet ./pkg/proxy`
- `mo-self-review` Q1-Q3 lifecycle gate: no open blockers

XuPeng-SH (approved):
Re-reviewed head e4d9edb637. Both requested changes are fully addressed: production and regression now share runBootstrapTask, and TestBootstrapStopsWithStopper deterministically reaches an in-flight GetClusterState call before exercising the actual Stopper taskCtx -> AfterFunc -> bootstrapCtx cancellation chain and proving Stop waits for task termination. The asynchronous guards are now generous, the wall-clock performance assertion is removed, and the persistent-error test still verifies DeadlineExceeded plus preservation of the last HAKeeper error.

I also rechecked the extracted helper for semantic drift and the test failure path, and found no new correctness, performance, or Q1/Q2/Q3 lifecycle issue. Focused tests x20, focused -race, and the full proxy package pass locally. Approving; the newly triggered CI run is still in progress with no failures at review time.

gouhongshen (commented):
Deep review of head `e4d9edb637faeba545eba61fdf97574b783b6ed3`.

### Finding

**[P2] Clean up startup tasks when `NewServer` fails after scheduling them**

In `pkg/proxy/server.go`, `runBootstrapTask` (L108) and the heartbeat task (L112) are started before `goetty.NewApplication` (L117). If listener parsing or binding fails, the function returns at L126 without calling `s.stopper.Stop()` or `h.Close()`. At that point the returned `*Server` is nil, so there is no owner left to stop these tasks: the new bootstrap task retains the handler/HAKeeper client and continues polling until the five-minute bootstrap timeout, while the heartbeat task also remains alive. Please add one failure-cleanup path after the stopper/handler are created, and cancel the tasks before closing their dependencies.

### Validation

- `CGO_CFLAGS="-I$(pwd)/cgo -I$(pwd)/thirdparties/install/include" go test ./pkg/proxy -run '^TestBootstrap' -count=20`
- `go test -race ./pkg/proxy -count=1`
- `go vet ./pkg/proxy`
- `git diff --check base...head`

All passed locally; the finding above is an untested constructor-error path.

PR #25643, opened 2026-07-12
fix: harden txn client cancellation and cleanup

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [x] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

## Which issue(s) this PR fixes:

N/A

> This PR does not fix #25206. That lockservice retry issue was fixed by #25651.

## What this PR does / why we need it:

This PR hardens the txn client lifecycle under cancellation, client shutdown, CN failure, concurrent terminal calls, and operator restart. The core contract is generation ownership: a txn operator can be reused only after the previous generation has closed normally and every workspace, TN, lock, admission, callback, and running-SQL responsibility has reached a proven safe outcome.

### Main changes

1. **Cancellation-aware client gates**
   - Replaces uncancelable condition waits with channel-based pause/close signaling.
   - Makes snapshot acquisition, active admission, and abort-marking waits observe context cancellation and client close.

2. **Linearized admission ownership**
   - Gives queued and active transactions one explicit owner across cancellation, promotion, client shutdown, and post-open creation failures.
   - Ensures every admitted failure crosses the ownership barrier and emits exactly one close transition.

3. **Generation-safe SQL and terminal ownership**
   - Tracks running SQL with generation-scoped tokens and seals new work before terminal cleanup.
   - Serializes public `Commit`, `Rollback`, and `WriteAndCommit` ownership.
   - Uses explicit terminal outcomes: `None`, `Pending`, `Succeeded`, and `Failed`.
   - Admits `RestartTxn` only after `Succeeded`; failed workspace, TN rollback, unlock, callback, panic, and `runtime.Goexit` paths remain sealed.

4. **Complete cleanup instead of first-error short circuiting**
   - Attempts workspace rollback, TN rollback, and unlock even when an earlier layer fails.
   - Preserves unknown-commit safety so later rollback cannot delete objects that TN may already have committed.
   - Keeps a durable Commit result successful when only unlock fails, while marking that operator generation non-reusable.

5. **Defensive protocol and waiter handling**
   - Converts malformed TN response panics into errors and publishes response state only after full validation.
   - Removes canceled timestamp waiters immediately and keeps waiter indexing consistent.
   - Coalesces abort wakeups while retaining the latest monotonic observation.

6. **Deterministic CGo test support**
   - Adds a repository test wrapper that derives the native dependency closure from the CGo package and sets compile, link, and runtime-loader paths consistently.

[...]

PR #25780, opened 2026-07-15
fix(txn): wait for remote snapshot visibility

### LINKED ISSUE #25776: [Bug]: Multi-CN IVFFLAT query can fail with missing __mo_index_key instead of vector dimension error
## Description

A Multi-CN compose BVT run produced an unexpected IVFFLAT internal metadata error for an invalid-dimension vector query. The query should fail with a vector dimension mismatch, but one run instead failed while reading the IVFFLAT metadata table.

Observed in PR #25713 after merging main commit f1c0a92d88 (#25552).

CI job: https://github.com/matrixorigin/matrixone/actions/runs/29404607349/job/87316934761

## Failing SQL

```sql
create table vector_index_07(a int primary key, b vecf32(128), c int, key c_k(c));
-- insert rows and create IVFFLAT index
create index idx01 using ivfflat on vector_index_07(b) lists=5 op_type "vector_l2_ops";
select * from vector_index_07
order by L2_DISTANCE(b, "[16, 15, 3, 0, 0, 4, 3, 5, 51]") asc limit 2;
```

Expected:

```text
invalid input: vector ops between different dimensions (128, 9) is not permitted.
```

Actual in CI:

```text
invalid input: column __mo_index_key does not exist
```

`__mo_index_key` is a column of the hidden IVFFLAT metadata table and is queried by `ivfflat.GetVersion()` before the search runs.

## Evidence and current status

- CI ran with PR head `024afed532225d5b5cfc76a5e78db4514c3997d4` and base `f1c0a92d882dbdad469ffd2c96f607796207c797`.
- The only failed SQL was `test/distributed/cases/vector/vector_index.sql:94`; 169/170 statements passed.
- The new `vector_ivf_multicn_search.sql` case from #25552 passed 26/26 in the same run.
- The same compose Multi-CN job passed on #25552 itself.
- The same job passed on #25713 before merging f1c0a92d88.
- A Linux Multi-CN launch using the failing PR head confirmed `AP QUERY PLAN ON MULTICN`; 50 create-index/immediate-query iterations all returned the expected dimension error and did not reproduce the metadata error.

This currently looks intermittent or sequence-dependent rather than a deterministic planner failure.

## Investigation hypotheses

1. Catalog/schema visibility race on a remote CN immediately after hidden IVFFLAT tables are created.
2. A sequence-dependent issue in remote `IndexReaderParam` / table-function scope serialization introduced or exposed by #25552.
3. Hidden metadata table identity/configuration is stale or reused across remote pipeline requests.
4. Error precedence is fragile: invalid vector dimensions are validated only after `ivfflat.GetVersion()` reads metadata, allowing an internal error to mask the user-facing validation error.

## Acceptance criteria

- Identify the CN and execution stage that emits the missing-column error.
- Provide a deterministic or stress reproduction, or enough instrumentation to prove the catalog/serialization race.
- Invalid-dimension IVFFLAT queries consistently return the dimension mismatch error in standalone and Multi-CN modes.
- Valid IVFFLAT searches continue to work across all CN partitions.
- Add a Multi-CN regression/stress test that covers create-index followed immediately by invalid-dimension search.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25776

Remote SQL pipelines and table sharding reconstruct a transaction snapshot on another CN. Previously, `TxnClient.NewWithSnapshot` created that mirror transaction before the receiving CN logtail necessarily reached the snapshot visibility boundary. Remote catalog or partition lookup could therefore observe stale state and fail with errors such as `column __mo_index_key does not exist`.

This PR:

- waits until the receiving CN has applied logtail through `SnapshotTS.Prev()` before constructing the mirror transaction;
- covers both remote SQL execution and table-sharding process decoding;
- propagates snapshot construction failures through the pipeline response instead of panicking;
- binds the wait to query cancellation, connection close, and transaction-client close;
- linearizes final mirror admission with `TxnClient.Close()`, so a successful timestamp notification cannot publish an operator after shutdown.

`SnapshotTS.Prev()` is intentional because a transaction snapshot at `SnapshotTS` observes commits strictly before that timestamp.

### Testing

- Focused `NewWithSnapshot` tests passed.
- Deterministic concurrent close/successful-notification regression passed under `-race -count=100`.
- Full `pkg/txn/client` tests and full-package race tests passed.
- Tests passed for `pkg/txn/storage/memorystorage`, `pkg/vm/process`, `pkg/sql/compile`, `pkg/frontend`, and `pkg/frontend/test`.
- Build and vet passed for all affected packages.
- Linux multi-CN validation on `10.222.1.52` covered immediate create-index/query visibility: 100 iterations produced zero metadata-visibility errors and zero valid-query failures.

### Rebase

Rebased onto current `main` at `c1843634bf`. The branch preserves the timestamp-waiter ownership and deterministic race coverage already merged in main, plus the CN pipeline FIN/FIN_ACK lifecycle from #25713.


XuPeng-SH (changes requested):
The visibility barrier, cancellation propagation, waiter removal, and normal-path cost look sound. One blocking regression-test issue remains in pkg/txn/client/timestamp_waiter_test.go: TestGetTimestampCancelRacesWithNotify can pass without ever registering a waiter. If the spawned goroutine has not reached addToWait before cancel and notify run, latestTS advances first and GetTimestamp returns on its fast path, so removeWaiter and the cancel-vs-notify refcount race are never exercised. Repeating this 100 times relies on scheduler luck and still does not prove either ordering.

Please synchronize without sleeps until the waiter is registered, then deterministically cover both cancel-wins and notify-wins, assert the returned error/value and an empty waiter list, and ideally verify a subsequent waiter succeeds to cover pooled-object reuse. A simultaneous-start stress loop can supplement those ordered cases, but should not replace them.

aptend:
@XuPeng-SH Thanks, the regression-test finding is valid. I rebased onto current main (`ddbff5b400`), which includes #25643, and removed both our overlapping waiter implementation and `TestGetTimestampCancelRacesWithNotify`. The PR now preserves main’s stronger O(1) waiter ownership/cleanup and its tests that synchronize until registration before cancel/notify/close races. The remaining diff is limited to the remote snapshot visibility barrier and its call-chain/error/cancellation tests. Fresh txn-client tests and race tests, including main’s deterministic waiter race cases, pass. Please take another look.

aptend:
Rebased onto current main `71f8d076a8`; conflict came from merged #25713 touching the same remote-run receiver initialization. Resolution preserves #25713’s requested/accepted teardown negotiation and FIN/FIN_ACK lifecycle, while using the single session context to cancel the remote snapshot logtail wait on disconnect. Snapshot construction still finishes before lifecycle registration, so construction failures cannot leak a lifecycle entry. Fresh txn-client tests/race, full morpc/process/memorystorage/compile/frontend tests, combined remote-receiver + FIN lifecycle race tests, build, and vet pass.

XuPeng-SH (changes requested):
Blocking lifecycle race in `TxnClient.NewWithSnapshot`: the client-closed check happens only before the new logtail wait. If timestamp notification wins against close cancellation, the waiter is allowed to complete successfully; `Close()` can then finish before `GetTimestamp` returns, and lines 776-782 still construct and return a mirror operator backed by an already-closed client/sender. `WaitLogTailAppliedAt` only translates the close when the waiter returns an error, so the successful-notification branch bypasses close admission entirely.

I reproduced this deterministically with a waiter that (1) enters the wait, (2) selects a successful notification, (3) lets `TxnClient.Close()` complete, and then (4) returns that already-selected success: `NewWithSnapshot` returns a non-nil operator and nil error. This ordering is valid for the real waiter too; its notify-vs-cancel race explicitly permits notification to win.

Please make the final admission linearizable with `Close` (for example, acquire `lifecycle.gate.RLock()` after the wait, re-check `isClosed()`, and hold it through mirror construction), and add the concurrent-close/successful-wait regression test. A plain pre-wait check is insufficient because this PR materially lengthens that window.

aptend:
@XuPeng-SH Thanks, this lifecycle race is valid and is now fixed. I rebased the PR onto current main (`d8ca64575d`). `NewWithSnapshot` still performs the logtail wait without holding the lifecycle gate, then acquires `lifecycle.gate.RLock()`, rechecks `isClosed()`, and holds the gate through mirror construction. This makes final admission linearizable with `Close()` without blocking close-driven cancellation. I added a deterministic regression waiter that selects successful notification, allows `Close()` to complete, and only then returns the selected success; the call now returns nil plus `ErrClientClosed`. The test also asserts `Close()` is not blocked by the in-flight wait. Focused race stress (`-race -count=100`), full `pkg/txn/client` tests and race, all affected package tests, build, and vet pass.

XuPeng-SH (approved):
Reviewed current head e31eb3bdc4 across snapshot semantics, cancellation, client-close linearization, remote error propagation, ownership, and normal-path cost. The previously reported lifecycle race is fixed: the logtail wait remains outside lifecycle.gate, then final mirror admission acquires RLock, rechecks closed state, and holds the gate through construction. Close can cancel an in-flight wait without deadlock, while a success selected before shutdown cannot publish a mirror after Close wins.

SnapshotTS.Prev() matches the exclusive snapshot contract and timestamp-waiter semantics. Both reconstruction entry points pass their caller context; connection close cancels construction; failures are returned before stream lifecycle registration and sent as terminal pipeline errors. I found no unpaired owner, unterminated wait, unbounded accumulation, compatibility break, or material hot-path regression.

Local evidence: focused snapshot/close and remote receiver/helper tests passed under -race -count=20; full pkg/txn/client race and affected package tests passed. A dry merge with latest origin/main 6b1ad0c32b was clean, and focused tests, build, and vet passed on the combined tree. Approved. The PR is behind main, so update it and let CI revalidate the delivered head before merge.

aptend:
Rebased onto current main at c1843634bf. The only conflict was in pkg/sql/compile/remoterunServer_test.go imports: current main added prepared-parameter/affected-row coverage while this PR added the morpc session mock; the resolution keeps both. Range-diff confirms the lifecycle-admission commit is unchanged. Fresh focused race tests, the full txn/client race suite, and tests/build/vet for all affected packages pass locally.

aunjgr (approved):

Issue #25328, opened 2026-07-01
[Bug]: SELECT DISTINCT rejects ORDER BY expression derived from selected columns


Ariznawlll:
### Branch

main at `200bfa7e0efb`.

### Summary

For `SELECT DISTINCT`, MatrixOne rejects an `ORDER BY` expression that is derived from selected columns. MySQL accepts this because all columns referenced by the `ORDER BY` expression are present in the select list.

MatrixOne already correctly rejects `SELECT DISTINCT grp FROM t_order ORDER BY label` when `label` is not selected. This issue is about the valid derived-expression case.

### Reproduction

```sql
drop database if exists mysql_compat_module9_distinct_order;
create database mysql_compat_module9_distinct_order;
use mysql_compat_module9_distinct_order;

create table t_order(id int, grp int, label varchar(10));
insert into t_order values
  (1,1,'b'),
  (2,1,'a'),
  (3,2,'c'),
  (4,2,'a'),
  (5,3,'b'),
  (6,3,'d'),
  (7,3,'d');

select distinct grp, label
from t_order
order by concat(label, 'x'), grp;
```

### MySQL result

```text
+------+-------+
| grp  | label |
+------+-------+
|    1 | a     |
|    2 | a     |
|    1 | b     |
|    3 | b     |
|    2 | c     |
|    3 | d     |
+------+-------+
```

### MatrixOne result

```text
ERROR 1149 (HY000): SQL syntax error: for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

### Expected behavior

MatrixOne should allow `ORDER BY` expressions whose referenced columns are already in the `SELECT DISTINCT` list, matching MySQL.

Ariznawlll:
Additional Module 9 exploration found a smaller reproduction for the same rule:

```sql
drop database if exists mysql_compat_module9_distinct_order_expr;
create database mysql_compat_module9_distinct_order_expr;
use mysql_compat_module9_distinct_order_expr;

create table t(id int, grp int, label varchar(10));
insert into t values
  (1,1,'b'),
  (2,1,'a'),
  (3,2,'c'),
  (4,2,'a'),
  (5,3,'b');

select distinct grp from t order by grp + 0 desc;
```

MySQL accepts this because `grp + 0` is derived only from the selected `grp` column:

```text
+------+
| grp  |
+------+
|    3 |
|    2 |
|    1 |
+------+
```

MatrixOne rejects it:

```text
[...]

PR #25783, opened 2026-07-15
fix: support expressions over DISTINCT select columns in ORDER BY

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25328

`SELECT DISTINCT` currently rejects `ORDER BY` scalar expressions even when every referenced value is already present in the select output, for example `SELECT DISTINCT b FROM t ORDER BY abs(b)`.

This change:

- binds derived `ORDER BY` expressions only against selected aliases and directly selected columns;
- keeps the DISTINCT tuple unchanged and materializes hidden sort keys in a projection after duplicate elimination;
- keeps sort keys as column references for optimizer and spill compatibility;
- continues to reject expressions that introduce unselected input columns, aggregates, windows, subqueries, or time-window functions;
- adds planner unit tests and BVT coverage for selected columns, aliases, rejection paths, duplicate hidden keys, and final projection shape.

Validation:

- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `go test -count=1 -timeout=120s ./pkg/sql/plan`
- DISTINCT BVT: 133/133 passed
- ABS BVT: 40 passed, 2 expected ignored
- added executable-line coverage: 81.45% (statement-weighted: 88.76%)


heni02 (approved):

XuPeng-SH (changes requested):
Blocking correctness issue: `distinctOrderBinder.BindColRef` applies alias-first resolution unconditionally and ignores its `isRoot` argument. MySQL uses different resolution for a bare/root ORDER BY name and for a name nested inside an expression.

Concrete reproductions against MySQL 8.0:

```sql
CREATE TABLE t(a INT, b INT);
SELECT DISTINCT a AS b FROM t ORDER BY ABS(b);
```

MySQL rejects this with error 3065 because `b` inside `ABS(b)` resolves to the FROM column `t.b`, which is not selected. This PR instead binds the select alias at output position 0 and accepts the query.

```sql
SELECT DISTINCT a AS b, b FROM t ORDER BY ABS(b);
```

MySQL resolves the nested `b` to the selected FROM column at output position 1. A planner probe on this PR shows `actual ColPos=0`, so the query can return rows in the wrong order.

There is also a root-expression regression:

```sql
SELECT DISTINCT a AS b, b FROM t ORDER BY (b);
```

MySQL reports column `b` as ambiguous, while this PR accepts it because parentheses bypass the existing top-level ambiguity check and the new binder chooses the alias.

Please make name resolution root-aware instead of always alias-first: preserve the existing select-list/ambiguity rules for bare root names (including parenthesized roots); inside derived expressions, resolve FROM columns first and require the resolved column to be directly projected, falling back to a select alias only when no FROM column with that name exists. Add regression tests for (1) alias colliding with a selected FROM column, (2) alias colliding with an unselected FROM column, (3) parenthesized root ambiguity, (4) alias with no same-named FROM column, and (5) qualified-column control.

The DISTINCT -> hidden PROJECT -> SORT -> result PROJECT shape otherwise looks sound, and the added targeted tests plus `-race` pass.

aunjgr:
@XuPeng-SH Fixed the blocking name-resolution issues in `0e4e41d11`.

- Bare/root names keep the existing select-list ambiguity rules; parenthesized roots are now handled the same way.
- Names nested in expressions resolve against FROM columns first and must map to a directly projected column.
- Alias fallback is used only when no same-named FROM column exists.
- Qualified selected/unselected column behavior is covered as a control.

Added unit regressions for all five requested cases plus BVT cases. Validation passed for the targeted DISTINCT tests under `-race`, the full `pkg/sql/plan` suite, `go build`, and `go vet`. PR-added production-line coverage is 79.9% (107/134).

Please take another look.

XuPeng-SH (approved):
Re-reviewed head 0e4e41d11. The previous DISTINCT ORDER BY name-resolution blockers are resolved.

Verified:
- bare and parenthesized root names preserve output ambiguity rules;
- names nested in expressions prefer FROM columns and reject unselected inputs;
- alias fallback occurs only when no same-named FROM column exists;
- qualified selected and unselected controls behave correctly;
- exact selected expressions remain orderable under alias collision;
- DISTINCT -> hidden PROJECT -> SORT -> result PROJECT keeps the DISTINCT tuple and final output shape intact.

Fresh local validation passed: full pkg/sql/plan test suite, go build, go vet, targeted race tests, and an additional adversarial collision probe. No new blocking findings. Remaining in-progress CI is left to branch protection.

Issue #25751, opened 2026-07-15
[Subtask]: add numeric resolver and prepared expression context


ck89119:
### Parent Issue

Parent issue: #25705

### Detail of Subtask

Introduce an arithmetic-specific numeric type resolver and prepared-expression context propagation for `+`, `-`, `*`, `/`,
`DIV`, `MOD`, and unary `+/-`.

The resolver must select parameter and result types from typed operands and optional outer context with this precedence:

1. approximate operand;
2. decimal operand;
3. exact outer context;
4. integer operand;
5. context-free parameters default to `DOUBLE`.

Cover explicit casts, literals, read-only column type lookup, nested arithmetic, function-form `MOD()`, and strict boundaries
that prevent numeric context from leaking into unrelated functions or comparisons.

### Describe implementation you've considered

Implement a pure resolver in `pkg/sql/plan/function` that returns operand coercion and result types. During PREPARE, scan one
continuous arithmetic AST without mutating binder state, then bind the subtree once with the selected parameter type. Keep
non-prepared queries and non-numeric coercion on their existing paths.

### Additional information

Acceptance requires table-driven resolver tests, prepared plan tests, ordinary-query regressions, decimal precision/scale
coverage, and no mutation of GROUP BY, projection, or alias bookkeeping during type inspection.

PR #25788, opened 2026-07-16
fix(plan): resolve prepared numeric expression types

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

issue #25751

- Add an arithmetic-specific numeric resolver for prepared expressions.
- Infer parameter types from approximate operands, decimal operands, exact outer context, integer operands, or the default
  `DOUBLE` domain.
- Propagate numeric context through nested arithmetic, casts, unary operators, columns, and `MOD` while isolating unrelated
  functions, comparisons, and temporal expressions.
- Keep `ParamRef` transport as `TEXT` and materialize the computation type through explicit casts.
- Cover all supported arithmetic operators, decimal result metadata, context boundaries, and `GROUP BY`/alias state
  preservation.

Validation:

- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- `go build ./pkg/sql/plan/function ./pkg/sql/plan`
- Modified resolver coverage: 75%-100%; binder integration: 85.7%.
- `git diff --check`


pkg/sql/plan/base_binder.go
```diff
@@ -668,6 +693,13 @@ func (b *baseBinder) bindUnaryExpr(astExpr *tree.UnaryExpr, depth int32, isRoot
 }
 
 func (b *baseBinder) bindBinaryExpr(astExpr *tree.BinaryExpr, depth int32, isRoot bool) (*Expr, error) {
+	if isNumericBinaryOp(astExpr.Op) && b.numericParamType == nil {
```
XuPeng-SH, line 696: This admits every prepared `+`/`-` subtree based only on the AST operator. A DATE sibling is collected by `numericAstTypes`, but `InferNumericParameterType` ignores that non-numeric type and falls back to DOUBLE. Consequently `SELECT O_ORDERDATE - ? FROM orders` now fails as `[DATE DOUBLE]`, and `SELECT ? - O_ORDERDATE FROM orders` fails as `[DOUBLE DATE]`; both build successfully on clean latest main because the parameter remains TEXT and the existing temporal cast rule handles it. Please bypass the numeric resolver when the scan encounters a temporal/non-numeric typed operand (and cover DATE/DATETIME in both operand orders).

XuPeng-SH (changes requested):
Requesting changes for one confirmed compatibility regression.\n\nThe numeric resolver's priority, explicit-cast transport, decimal metadata, nested arithmetic, and binder-state isolation look sound, and the submitted unit suites plus full plan/function tests pass. However, admission is currently based on the AST operator alone, so prepared temporal subtraction is incorrectly routed through the numeric resolver.\n\nClean latest-main comparison:\n- main: SELECT O_ORDERDATE - ? FROM orders builds successfully\n- this PR: fails with invalid argument operator -, bad value [DATE DOUBLE]\n- main: SELECT ? - O_ORDERDATE FROM orders builds successfully\n- this PR: fails with invalid argument operator -, bad value [DOUBLE DATE]\n\nThis violates the stated boundary that temporal arithmetic stays on its existing coercion path. Please make numeric scanning/admission reject temporal or otherwise non-numeric typed operands and add regression coverage for DATE/DATETIME with the bare parameter on both sides.

heni02 (approved):

XuPeng-SH (approved):
Approved after re-reviewing current head fffa28e4 on latest main.

The previously confirmed temporal-subtraction blocker is closed. Numeric AST admission now rejects non-numeric typed operands and falls back to the legacy coercion path; DATE and DATETIME are covered with the prepared parameter on both sides.

Reviewed closure:
- numeric precedence: approximate > decimal > exact outer context > integer > default DOUBLE
- explicit parameter casts preserve the TEXT transport contract
- comparison, unrelated-function, direct-cast, and temporal boundaries remain isolated
- MOD, unary operators, all binary arithmetic operators, BIT capacity, decimal width/scale, correlated lookup, and binder group/alias state are covered
- temporary binder numeric context is restored on success/error/boundary paths; no new resource, wait, or unbounded-retention edge

Local deterministic validation:
- focused resolver and binder regression suites: count=5
- full pkg/sql/plan/function and pkg/sql/plan tests
- dependent pkg/sql/compile tests
- compile-only checks and go vet for all three packages

All passed; no blocking correctness or unhappy-path finding remains.

PR #25789, opened 2026-07-16
perf(plan): prune unused SAMPLE outputs

### LINKED ISSUE #25698: [SubTask]: Prune unused SAMPLE outputs and compact physical slots
## Brief Description

Part of #25612. `SAMPLE` currently marks every sample expression as required even when an outer query consumes only a subset of the sampled outputs.

- In the `Node_SAMPLE` branch of `remapAllColRefs` (`pkg/sql/plan/query_builder.go`), stop unconditionally ref-counting the whole `AggList`; prune sample outputs down to those actually consumed.
- Compact `SAMPLE` physical slots and remap `HAVING` and parent projections accordingly.
- Keep a row-producing carrier when all sample values are discarded (depends on the low-cost row carrier work).
- Preserve group/sample column dependencies and row-count semantics.

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25698

- Prunes unused SAMPLE expressions and their exclusive input dependencies during column remapping.
- Compacts physical SAMPLE slots and remaps HAVING filters, parent projections, and passthrough offsets.
- Preserves side-effecting expressions and uses the deterministic low-cost row carrier from #25697 when every SAMPLE value is discarded.
- Adds plan-shape coverage and BVT cases for partial pruning, HAVING-only consumers, side effects, row and percent sampling, and 0/1/many-row cardinality.

## Testing

- `go test ./pkg/sql/plan ./pkg/sql/plan/explain -count=1`
- `go test -race ./pkg/sql/plan/explain -run '^TestColumnPruneOperatorShape$/sample' -count=1`
- `go vet ./pkg/sql/plan ./pkg/sql/plan/explain`
- `go build ./pkg/sql/plan ./pkg/sql/plan/explain`
- `make build`
- `make static-check`
- `column_pruning.test`: 36/36 passed with a clean data directory and freshly built service
- `git diff --check`


XuPeng-SH (changes requested):
Blocking: SAMPLE pruning drops volatile side effects other than sleep.\n\nThe new pruning checks at pkg/sql/plan/query_builder.go:1329-1371 rely on exprCanRemoveProject. That helper only special-cases sleep; it does not consult the function overload volatility metadata. As a result, sequence functions are considered removable even though nextval/setval mutate sequence and session state.\n\nA concrete query is:\n\nselect n_regionkey\nfrom (\n  select sample(nextval('review_seq'), n_regionkey, 2 rows)\n  from nation\n) s;\n\nnextval is registered with volatile=true and realTimeRelated=true, and its executor advances the sequence for every evaluated row. On current PR head, because the parent consumes only n_regionkey and both expressions are statically non-null, column pruning compacts the SAMPLE AggList from [nextval, n_regionkey] to [n_regionkey]. constructSample therefore never creates the nextval executor, silently removing the sequence advancement.\n\nI reproduced this with a focused plan-shape regression. It passes on the exact base ddbff5b40 with AggList length 2, but fails on e1a025be3 and on the merge with current main because AggList contains only n_regionkey.\n\nPlease make expression removability reject volatile/real-time-related functions recursively, using the existing function overload metadata rather than a sleep-only name check, and add sequence-function coverage (nextval, and ideally setval).

heni02 (approved):

XuPeng-SH (approved):
Re-reviewed current head `6180f33c0a`. The previous P1 blocker is fixed.

The closure is now sound for this PR's SAMPLE pruning path:

- removability is derived recursively from function overload metadata (`volatile` / real-time-related), rather than a `sleep` name special case, so `nextval`, `setval`, nested volatile calls, and unknown overloads are retained;
- SAMPLE slot compaction is permitted only when every multi-column candidate is a direct, strictly `NOT NULL` table column; nullable and runtime-null-producing expressions conservatively retain the original multi-column semantics;
- HAVING-only consumers, parent projections, passthrough offsets, and compacted physical SAMPLE positions are remapped consistently;
- when every value is discarded, one low-cost carrier remains, preserving 0/1/many-row cardinality and avoiding a zero-column SAMPLE;
- consumed volatile projections are not inlined across Project/Sort/Join boundaries, preserving evaluation count and sibling order;
- the accompanying `mo_table_col_min/max` state fix makes failure state row-local and defaults an unmatched column to NULL, which is required once the volatile project is retained as a batch evaluation boundary.

Unhappy-path audit: the change introduces no goroutine, wait, or external-resource ownership edge (Q1/Q2); temporary remap slices are bounded by the plan's expression count and function-local (Q3).

Fresh validation on the exact PR head:

- focused column-pruning/volatility tests passed;
- full `pkg/sql/plan` and `pkg/sql/plan/explain` tests passed;
- focused tests passed with `-race -count=5`;
- affected-package build, vet, and `git diff --check` passed.

Approved.

Issue #25735, opened 2026-07-15
[Bug]: TestManyRangeLockInManyGoroutines data race in lockservice


matrix-meow:
## Problem

`TestManyRangeLockInManyGoroutines/local` in `pkg/lockservice` fails with a data race:

```
testing.go:1712: race detected during execution of test
```

The test exercises range lock operations across multiple goroutines. The Go race detector found concurrent unsynchronized access in the lockservice code path.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-15 | [PR #25728](https://github.com/matrixorigin/matrixone/actions/runs/29389731923/job/87270295458) | UT Test on Ubuntu/x86 |

## Root Cause

Data race in lockservice range lock operations under concurrent goroutine access. The race is intermittent and depends on goroutine scheduling.

## Suggested Fix

1. Run the test locally with `-race` flag to get the full race report and identify the exact shared variable
2. Add proper synchronization (mutex/atomic) to the racing code path
3. Or fix the test if the race is in test setup rather than production code

daviszhen:
fixed

Issue #25734, opened 2026-07-15
[Bug]: TestRequestMultipleCn_ConcurrentSafety flaky — one node fails with closed connection


matrix-meow:
## Problem

`TestRequestMultipleCn_ConcurrentSafety` in `pkg/queryservice` fails intermittently:

```
actual: 2 (expected 3)
Messages: All nodes should succeed
ERROR: close unix @->/tmp/cn-*-test_concurrent.sock: use of closed network connection
```

The test sends concurrent requests to 3 mock CN nodes, but one node's connection is closed before the request completes, causing only 2/3 nodes to succeed.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-15 | [PR #25715](https://github.com/matrixorigin/matrixone/actions/runs/29389515987/job/87269719259) | Coverage |

## Root Cause

Race condition in concurrent connection handling — one CN's Unix socket connection is closed before the request-response cycle completes, likely due to goroutine scheduling variability on CI runners.

## Suggested Fix

1. Add retry logic for transient connection failures in the test
2. Or use synchronization to ensure all connections are established before sending requests
3. Or increase connection timeout tolerance

daviszhen:
fixed

Issue #25729, opened 2026-07-14
[Bug]: TestDispatchEmptyInputWaitsForRemoteReceiver flaky — dispatch returns before remote receiver attaches


matrix-meow:
## Problem

`TestDispatchEmptyInputWaitsForRemoteReceiver` in `pkg/sql/colexec/dispatch/dispatch_test.go:336` fails intermittently:

```
dispatch_test.go:336: empty dispatch returned before its remote receiver attached
```

The test verifies that the dispatch operator waits for a remote receiver to attach before returning when input is empty. The dispatch returns prematurely (before receiver attachment), indicating a race condition in the wait logic.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-14 | [PR #25728](https://github.com/matrixorigin/matrixone/actions/runs/29332591357/job/87083996547) | UT Test on Ubuntu/x86 |

## Root Cause

Race condition between dispatch returning and remote receiver attachment. Under CI runner scheduling variability, the dispatch goroutine may complete before the receiver goroutine attaches.

## Suggested Fix

1. Add synchronization (channel/WaitGroup) to ensure receiver attachment happens before dispatch checks
2. Or increase the wait tolerance in the dispatch wait logic
3. Or make the test deterministic by using explicit synchronization instead of timing

daviszhen:
fixed

LeftHandCold:
Rechecked against current main 1a14737560. The originally reported post-send completion assertion is absent. The dispatch test now directly asserts the authoritative state: prepared is true and remoteReceivers contains one entry; both are possible only after Dispatch.Call receives the remote handoff.

The removed completion signal was closed after an unbuffered channel send. Its receiver could therefore return and inspect the signal before the sending goroutine was rescheduled to close it. That was a test scheduling race, not evidence that Dispatch returned before attachment.

The PR #25742 failure record was generated against a merge ref, but the reported old assertion does not exist in the current-main test source. No duplicate fix is proposed: #25820 already superseded this issue code and test path.

PR #25792, opened 2026-07-16
 fix ut

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25735

https://github.com/matrixorigin/matrixone/issues/25734

https://github.com/matrixorigin/matrixone/issues/25729

  - 修复同步 lock waiter 返回后仍被 waiterEvents 持有的问题，及时移除并释放 events 引用，避免 waiter 复用时发生 data race。
  - waiterEvents.close() 清空 waiter 列表，防止关闭后重复释放引用。
  - 增加 waiter 生命周期、关闭和 range-lock 并发回归测试。
  - 修正 QueryService 并发 UT：预热 MORPC backend，消除首次建连竞争导致的偶现失败。
  - 删除 Dispatch UT 中不成立的异步时序断言，保留能够直接证明 receiver 已注册的断言。
  - UT 使用非池化 waiter，避免污染全局 reuse checker。


XuPeng-SH (approved):
Reviewed the waiter ownership graph and the two flaky-test adjustments on head 509404b529aae2c721883b175be41606d9d7cf53.

The waiterEvents reference is now removed before a completed waiter can be reset/reused; checker, explicit removal, and service close are serialized by the same mutex, so the events-owned ref is released exactly once. Range-lock requeue refs remain symmetric across the lock queue, active transaction, events list, and caller ownership. Clearing the list on close also removes stale backing-array references.

The QueryService handler is invoked serially by the response-consumer loop, so removing its test-only mutex is safe; backend warmup isolates the intended concurrent-send behavior and releases the warmup response. The removed Dispatch assertion depended on sender scheduling after an unbuffered-channel handoff, while `remoteReceivers` directly proves registration.

Validation: lockservice race tests ×20, reuse-checker race tests ×10, QueryService race test ×20, and Dispatch race test ×50 all passed. No blocking findings.

pkg/queryservice/multi_cn_bug_test.go
```diff
@@ -340,19 +339,27 @@ func TestRequestMultipleCn_EmptyNodeAddress(t *testing.T) {
...
+func TestRequestMultipleCnConcurrentSend(t *testing.T) {
 	defer leaktest.AfterTest(t)()
 
 	cn := metadata.CNService{ServiceID: "test_concurrent"}
 	runTestWithQueryService(t, cn, nil, func(cli client.QueryClient, addr string) {
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer cancel()
 
-		// Multiple nodes to increase concurrency
+		warmupReq := cli.NewRequest(pb.CmdMethod_GetCacheInfo)
```
XuPeng-SH: [P1] Do not pre-create the backend here. The current base already includes #25771's fix for concurrent cold backend creation, and the original no-warmup test passes `-race -count=100` on this head. Warming up changes the test from the #25734 failure scenario into only testing sends against an existing backend, so a regression in creation/waiter wakeup would pass unnoticed. Keep the cold-start setup; the callback mutex and artificial sleep can still be removed.

XuPeng-SH (changes requested):
Double-check found one blocking test regression; the lockservice waiter fix itself remains sound.

The PR base already contains `cfca2ec32 fix(morpc): wake waiters when backend creation completes (#25771)`, which fixed the cold-start backend-creation wakeup race reported by #25734. I restored the original no-warmup scenario on this head and ran it with `-race -count=100`; all runs passed. Therefore the warmup is not needed for stability on the current base. It removes coverage of the exact cold concurrent backend-creation path that #25771 fixed and changes this test into the materially weaker case of concurrent sends after a backend already exists.

Please keep the requests cold so this remains a regression test for #25734/#25771. Removing the callback mutex and scheduling sleep is fine because `RequestMultipleCn` invokes response handlers serially.

fengttt (commented):
Do we have a BVT test on these lock functions?

daviszhen:
> Do we have a BVT test on these lock functions?

悲观事务测试case 间接测试的. 没有直接的测试case.
从目前的issue看, 跟rpc,lock相关的issue 比较多.
还是就是sql语义,兼容性相关的多.

pkg/queryservice/multi_cn_bug_test.go
daviszhen: - 修复 lockservice 同步 waiter 复用时与后台检查的 data race，等待结束后及时移出事件列表并释放引用。
  - 修正 dispatch UT 中错误的时序断言，避免偶现失败。
  - 恢复 queryservice 冷启动并发测试，覆盖 MORPC backend 首次创建及 waiter 唤醒，删除无必要的 mutex 和 sleep。
  - 补充 waiter 生命周期、引用释放和关闭清理 UT。

daviszhen:
> Double-check found one blocking test regression; the lockservice waiter fix itself remains sound.
> 
> The PR base already contains `cfca2ec32 fix(morpc): wake waiters when backend creation completes (#25771)`, which fixed the cold-start backend-creation wakeup race reported by #25734. I restored the original no-warmup scenario on this head and ran it with `-race -count=100`; all runs passed. Therefore the warmup is not needed for stability on the current base. It removes coverage of the exact cold concurrent backend-creation path that #25771 fixed and changes this test into the materially weaker case of concurrent sends after a backend already exists.
> 
> Please keep the requests cold so this remains a regression test for #25734/#25771. Removing the callback mutex and scheduling sleep is fine because `RequestMultipleCn` invokes response handlers serially.

已经修改.

  - 修复 lockservice 同步 waiter 复用时与后台检查的 data race，等待结束后及时移出事件列表并释放引用。
  - 修正 dispatch UT 中错误的时序断言，避免偶现失败。
  - 恢复 queryservice 冷启动并发测试，覆盖 MORPC backend 首次创建及 waiter 唤醒，删除无必要的 mutex 和 sleep。
  - 补充 waiter 生命周期、引用释放和关闭清理 UT。

XuPeng-SH (approved):
Re-reviewed current head e0613dc56 after the cold-start QueryService regression was restored. The previous blocker is resolved.

The lockservice ownership closure is sound: waiterEvents takes one reference per blocked-list entry; checker removal, synchronous removal, and service close are serialized by the same mutex, so each events-owned reference is released exactly once. The synchronous path now removes that old-generation reference after wait completion and before reset/requeue, preventing the checker from observing a reused waiter. Lock-queue, active-transaction, events, and caller references remain independent and symmetric across success, timeout/cancellation, range-lock requeue, and close. Clearing the slice also drops stale backing-array references.

The QueryService test again keeps the backend cold and therefore covers concurrent backend creation and waiter wakeup. Response handlers are invoked serially by the draining loop, so removing the test mutex and sleep is valid. The Dispatch assertion that was removed depended on sender scheduling after an unbuffered-channel handoff; the retained remoteReceivers assertion directly proves registration.

Validation on the clean synthetic merge with current main: focused race tests passed for lockservice x20, cold QueryService x50, and Dispatch x50. Full owning-package race suites also passed for lockservice, queryservice, and dispatch. CI is green and the PR merges cleanly. No blocking findings.

iamlinjunhong (approved):

PR #25793, opened 2026-07-16
fix(parser): support Unicode select aliases

### LINKED ISSUE #25014: Non-ASCII (Arabic/Chinese) column identifiers rejected even when backtick-quoted (error 1064) — v4.0.0-rc2
## Version

`8.0.30-OmniFabric-v4.0.0-rc2` (output of `SELECT version()`)

## Description

MatrixOne's SQL parser rejects non-ASCII column identifiers / aliases (Arabic, Chinese, …) with `error 1064`, **even when they are backtick-quoted**. MySQL 8 accepts backtick-quoted non-ASCII identifiers, so portable SQL that uses them fails on MatrixOne.

## Reproduction

```sql
SELECT 1 AS qty;          -- OK
SELECT 1 AS الكمية;       -- ERROR 1064
SELECT 1 AS `الكمية`;     -- ERROR 1064  (backtick-quoted Arabic, still rejected)
SELECT 1 AS `数量`;       -- ERROR 1064  (backtick-quoted Chinese, still rejected)
```

## Actual

```
ERROR 1064 (HY000): SQL parser error: You have an error in your SQL syntax; check the manual that corresponds to your MatrixOne server version for the right syntax ...
```

## Expected

Backtick-quoted non-ASCII identifiers are accepted (as in MySQL 8), so aliases like `` `数量` `` / `` `الكمية` `` are usable.

## Impact

Hit by a natural-language-to-SQL assistant answering Arabic / Chinese questions: the model sometimes names an alias in the user's language (e.g. `SELECT COUNT(*) AS القيمة FROM dim_site`), and the whole query fails. The practical workaround is to force ASCII aliases, but accepting backtick-quoted non-ASCII identifiers would match MySQL behavior and avoid the failure.

ULookup:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25014

The MySQL scanner only treated ASCII letters as the start and continuation of unquoted identifiers, so select aliases containing Arabic, Chinese, or other BMP Unicode characters failed to parse.

This change enables valid BMP Unicode identifier scanning only when the lexer is reading the token after `AS`. Other grammar contexts retain their original tokenization and parser error positions. Invalid UTF-8, NUL, and supplementary characters remain rejected, including numeric, hexadecimal, and binary prefix paths.

Regression coverage includes Arabic and Chinese aliases, numeric prefixes, quoted aliases, invalid UTF-8, NUL, supplementary characters, bind variables, and exact legacy errors for non-alias DDL and account paths.

Validation:

- `make build`
- `go test -race -count=1 -run 'TestUnicodeIdentifier|TestInvalidUnicodeIdentifiers' ./pkg/sql/parsers/dialect/mysql`
- Verified through the MySQL protocol against a locally built MatrixOne service using an 8-row Chinese/Arabic dataset with grouping, NULL aggregation, and computed amounts; Unicode result-column headers were returned correctly.
- Verified the affected DDL and account cases still return parser error 1064 with the original error columns.

### DIFF
--- a/pkg/sql/parsers/dialect/mysql/mysql_lexer.go
+++ b/pkg/sql/parsers/dialect/mysql/mysql_lexer.go
@@ -166,6 +166,9 @@ func (l *Lexer) Lex(lval *yySymType) int {
 	if reservedKeywordsAfterAS[typ] && l.lastToken == AS {
 		typ = ID
 	}
+	if typ == unicodeID && l.lastToken == AS {
+		typ = ID
+	}
 
 	l.lastToken = typ
 	lval.str = str
--- a/pkg/sql/parsers/dialect/mysql/mysql_lexer.go
+++ b/pkg/sql/parsers/dialect/mysql/mysql_lexer.go
@@ -150,6 +150,7 @@ func (l *Lexer) GetParamIndex() int {
 }
 
 func (l *Lexer) Lex(lval *yySymType) int {
+	l.scanner.allowUnicodeIdentifier = l.lastToken == AS
 	typ, str := l.scanner.Scan()
 	l.scanner.LastToken = str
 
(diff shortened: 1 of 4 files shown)


VioletQwQ-0:
Superseded by merged PR #25879. The final takeover preserved the compatible behavior with regression coverage and removed the scanner-side charset validation that lacked COM_QUERY client-charset context.

PR #25796, opened 2026-07-16
perf(hashmap): allocate int iterator buffers lazily

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25472

`IntHashMap.NewIterator` eagerly allocated five `UnitLimit`-sized buffers for every iterator, even when a call processed only a few rows. In the clean 1000-VU insert profile this was the largest flat allocation site: 19,271.57 MB, or 4.56% of total allocation space.

This change:

- allocates iterator buffers lazily on the first non-empty `Find` or `Insert`, sized to the requested row count;
- reuses existing capacity when the iterator processes a smaller or equal batch later;
- preserves the nullable 8-byte-key guard slot at `count == UnitLimit` while keeping all logical slice lengths equal to `count`;
- keeps the existing `UnitLimit` contract and returns empty slices without allocating for `count == 0`.

There is no public API change. Tests cover real NULL values, `int32`/`int64`, counts `0/1/255/256`, grow/shrink/reuse, and changing an iterator owner from a non-nullable map to a nullable map at `UnitLimit`.

### Validation

```text
go test -race -count=1 ./pkg/common/hashmap
go test -count=1 ./pkg/sql/colexec/hashbuild ./pkg/sql/colexec/dedupjoin ./pkg/sql/colexec/rightdedupjoin ./pkg/sql/colexec/group
go test -run '^$' -bench '^BenchmarkIntHashMapIteratorFirstInsert$' -benchmem -count=5 ./pkg/common/hashmap
```

All passed. Allocation results were stable across five benchmark runs:

| Rows | B/op | allocs/op |
|---:|---:|---:|
| 1 | 168 | 6 |
| 255 | 9,344 | 6 |
| 256 | 9,344 | 6 |

### Fresh-data insert A/B/A

Environment: mo-129, 1000 VUs, 120 seconds, `max-active=640`, `disable-trace=true` in log/TN/CN, 8 GB SHARED memory cache and 50 GB SHARED disk cache. Every arm used an independent fresh data, cache, config, and result directory; prepare/warmup/run contamination logs were all empty.

| Arm | TPS | Avg | P95 | CPU ms/query | alloc kB/query | malloc/query | GC count |
|---|---:|---:|---:|---:|---:|---:|---:|
| Baseline A | 20,188.98 | 49.45 ms | 71.83 ms | 2.162 | 198.698 | 1,902.307 | 192 |
| Patched | 21,585.35 | 46.26 ms | 69.29 ms | 2.061 | 188.904 | 1,901.542 | 187 |
| Baseline A2 | 20,227.90 | 49.34 ms | 71.83 ms | 2.154 | 198.629 | 1,901.761 | 187 |

Compared with the mean of both baseline arms:

- TPS: +6.814%
- average latency: -6.347%
- P95: -3.536%
- CPU/query: -4.505%
- alloc/query: -4.913%
- malloc/query: -0.026% (neutral, as expected; this change reduces buffer sizes rather than allocation count)

Baseline binary SHA256: `b3c6c0d834110db64402bddceb2dd534ca532c412ac4e84d8cc5310342aa423c`  
Patched binary SHA256: `9c9cdbad7b3f8bd27a8f85dc55f6cbacc9b743d6cda82507cbf7aedaa7e4a83a`

### Exact change-commit allocation profile

The change commit was rebuilt and profiled separately with the same workload and configuration using an independent fresh dataset:

- profiled change commit: `b9876a9548c14ee59f02d056204811da7e393a88`
- current PR head: `d53b8bbe2cdef0d3c3eb2161bcc446b267ef0224` (Mergify's automatic merge of current `main`; the hashmap diff is unchanged)
- binary SHA256: `8b81bbf605a0e6a8c9031c28e9ead6c685090aec45a9e38404c60973c304c498`
- result: 2,430,963 transactions, 20,228.51 TPS, 49.36 ms average, 110.66 ms P95
- normalized runtime: 2.120 ms CPU/query, 191.071 kB allocated/query, 1,930.619 mallocs/query, 183 GCs
- prepare/warmup/run contamination logs: empty

This is a single-arm run for exact change-commit hotspot validation; cross-base throughput and latency are not used as an A/B claim.

The original clean baseline profile attributed 19,271.57 MB / 4.56% of 423,012.13 MB to `IntHashMap.NewIterator`. In the final 30-second delta profile:

- `IntHashMap.NewIterator`: 73.51 MB / 0.065%
- `intHashMapIterator.ensureCapacity`: 27.00 MB / 0.024%
- combined: 100.51 MB / 0.089%

After normalizing by total allocation space, the iterator allocation share fell by 98.04% (51.0x). The removed fixed-buffer allocation did not migrate to `ensureCapacity` or another new hotspot.

### DIFF
--- a/pkg/common/hashmap/inthashmap.go
+++ b/pkg/common/hashmap/inthashmap.go
@@ -44,12 +44,7 @@ func NewIntHashMap(hasNull bool, memPool *mpool.MPool) (*IntHashMap, error) {
 
 func (m *IntHashMap) NewIterator() *intHashMapIterator {
 	return &intHashMapIterator{
-		mp:      m,
-		keys:    make([]uint64, UnitLimit),
-		keyOffs: make([]uint32, UnitLimit),
-		values:  make([]uint64, UnitLimit),
-		zValues: make([]int64, UnitLimit),
-		hashes:  make([]uint64, UnitLimit),
+		mp: m,
 	}
 }
 
--- a/pkg/common/hashmap/inthashmap_lazy_test.go
+++ b/pkg/common/hashmap/inthashmap_lazy_test.go
@@ -1,0 +1,172 @@
+// Copyright 2026 Matrix Origin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hashmap
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/matrixorigin/matrixone/pkg/common/mpool"
+	"github.com/matrixorigin/matrixone/pkg/container/types"
+	"github.com/matrixorigin/matrixone/pkg/container/vector"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIntHashMapIteratorLazyBuffers(t *testing.T) {
+	for _, hasNull := range []bool{false, true} {
+		t.Run(fmt.Sprintf("has-null-%t", hasNull), func(t *testing.T) {
+			for _, typ := range []types.Type{types.T_int32.ToType(), types.T_int64.ToType()} {
+				t.Run(typ.String(), func(t *testing.T) {
+					for _, count := range []int{0, 1, 255, 256} {
+						t.Run(fmt.Sprintf("rows-%d", count), func(t *testing.T) {
+							m := mpool.MustNewZero()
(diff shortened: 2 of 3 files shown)


XuPeng-SH (approved):
Reviewed the complete diff and the iterator call-site/lifecycle closure. Approved.

Key checks:

- Every non-empty IntHashMap `Find` / `Insert` initializes all scratch buffers before encoding or calling the hash table; `count == 0` avoids the unsafe `&keys[0]` path.
- Grow, shrink, reuse, and iterator owner changes preserve the synchronized buffer-capacity invariant.
- Nullable 8-byte keys retain the required guard backing storage at `UnitLimit`, including a non-nullable → nullable owner change.
- Scratch growth is bounded by `UnitLimit` (plus one guard `uint64`), buffers are iterator-private and GC-managed, and no wait/resource lifecycle is introduced.
- The PR head passes `go test -race ./pkg/common/hashmap` and dependent hashbuild/dedup/group package tests. Existing CI is green.

I found no merge-blocking correctness, lifecycle, or compatibility issue.

PR #25801, opened 2026-07-16
fix: preserve extended TIME function semantics

### LINKED ISSUE #24784: [Bug]: time functions mishandle large hours and fractional seconds

I have checked the existing issues. #24481 is about unsupported functions on 3.0-dev; this issue is about incorrect results for already-supported time functions on 4.0-dev.

### Branch Name

4.0-dev

### Commit ID

47459ce08a

- Hardware parameters: N/A
- OS type: N/A
- Others: verified by source inspection on 4.0-dev

### Actual Behavior

Several MySQL-compatible time functions return incorrect results for TIME values whose hour part is greater than 23, or for MAKETIME seconds with fractional precision:

1. `HOUR(TIME)` folds the hour with `% 24`, so a TIME interval such as `272:59:59` returns `8` instead of the full hour value.
2. `TIME_FORMAT(time, "%T")` folds the hour with `% 24`, so `123:45:06` formats as `03:45:06` instead of preserving the full hour part.
3. `MAKETIME(hour, minute, second)` truncates fractional seconds for floating second arguments and always constructs the result with microseconds set to `0`.

### Expected Behavior

MatrixOne should match MySQL TIME semantics:

- `HOUR("272:59:59")` should return `272`.
- `TIME_FORMAT("123:45:06", "%T")` should return `123:45:06`.
- `MAKETIME(12, 34, 56.789012)` should preserve the fractional seconds and return `12:34:56.789012`.

### Steps to Reproduce

```sql
SELECT HOUR(CAST("272:59:59" AS TIME));
SELECT TIME_FORMAT(CAST("123:45:06" AS TIME), "%T");
SELECT MAKETIME(12, 34, 56.789012);
```

### Additional information

The source currently contains these incorrect paths on 4.0-dev:

- `pkg/sql/plan/function/func_unary.go`: `TimeToHour` returns `hour % 24`.
- `pkg/sql/plan/function/func_binary.go`: `TIME_FORMAT` `%T` uses `hour%24`.
- `pkg/sql/plan/function/func_binary.go`: `MakeTime` truncates floating seconds and passes `0` microseconds to `types.TimeFromClock`.

ULookup:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24784

MatrixOne's TIME function implementations did not preserve MySQL-compatible interval semantics:

- `HOUR(TIME)` reduced hours modulo 24 and returned `uint8`, so extended TIME values such as `272:59:59` returned `8`.
- `TIME_FORMAT` reduced `%T` hours modulo 24 and dropped the sign of negative TIME values.
- `MAKETIME` rejected negative hours, truncated fractional seconds, always returned scale 0, and could mishandle rounding and unsigned overflow boundaries.

This change:

- returns the complete absolute hour component for positive and negative TIME values;
- formats extended hours without modulo reduction and prefixes a negative sign once for the complete formatted result;
- preserves fractional seconds and result scale in `MAKETIME`;
- accepts valid negative hours and rejects invalid minute/second, normalized overflow, NaN/Inf, and unsigned overflow inputs;
- adds unit and distributed regression coverage for extended hours, negative values, fractional precision, rounding carry, and boundary cases.

## Validation

- `go test -count=1 -timeout 120s ./pkg/sql/plan/function`
- `go test -count=1 -timeout 120s ./pkg/sql/plan`
- Started a local single-process LOG/TN/CN deployment and executed the three modified distributed SQL cases through the MySQL protocol.
- Loaded 10,000 generated TIME rows covering positive/negative hours up to 838; observed 0 NULL rows and 0 `HOUR` mismatches.


heni02 (approved):

pkg/sql/plan/function/list_builtIn.go
```diff
@@ -10737,6 +10737,16 @@ var supportedDateAndTimeBuiltIns = []FuncNew{
 	},
 }
 
+func makeTimeReturnType(parameters []types.Type) types.Type {
+	scale := parameters[2].Scale
+	if scale < 0 {
+		scale = 0
```
XuPeng-SH: [P1] Preserve fractional precision for default DOUBLE inputs. A DOUBLE column with no declared scale is represented here as scale -1, but mapping that to TIME(0) makes MAKETIME(h, m, double_col) lose its fractional result metadata. MySQL resolves this as TIME(min(second.decimals, 6)); unspecified floating precision therefore becomes TIME(6), not TIME(0). I reproduced this through GetFunctionByName with INT, INT, FLOAT64(scale=-1): the selected second target remains float64, but retType is TIME(0). Since timeToStr formats with the source TIME scale, the microseconds calculated by MakeTime are then omitted. Map unknown floating scale to 6 and add a real default-DOUBLE column/planner regression test.

pkg/sql/plan/function/func_binary.go
```diff
@@ -7270,29 +7273,31 @@ func MakeDateString(
...
+	isNegative := hour < 0
+	if isNegative {
+		hour = -hour
 	}
-	if !types.ValidTime(uint64(h), 0, 0) {
+	timeValue := types.TimeFromClock(isNegative, uint64(hour), uint8(minute), uint8(second), microsecond)
+
+	normalizedHour, _, _, _, _ := timeValue.ClockFormat()
+	if normalizedHour > 838 {
 		return rs.Append(types.Time(0), true)
```
XuPeng-SH: [P1] Do not turn rounding overflow at the valid TIME endpoint into NULL. MySQL's Item_func_maketime calls adjust_time_range_with_warn and clamps otherwise-valid out-of-range results to the nearest TIME endpoint. Consequently MAKETIME(838, 59, 59.9999996) rounds/clamps to 838:59:59 (with a warning); a local test against this head gets NULL here. The same endpoint policy applies to larger hour inputs. Preserve the endpoint result and add this exact carry case to the regression suite.

XuPeng-SH (changes requested):
The HOUR and TIME_FORMAT changes are sound, and the existing targeted tests pass under the race detector. However, the MAKETIME path still has two blocking MySQL-compatibility/data-correctness regressions.

First, unknown/default DOUBLE scale is converted to TIME(0). MatrixOne represents an unspecified DOUBLE column with scale -1, while MySQL resolves MAKETIME to TIME(min(second.decimals, 6)); unspecified floating precision therefore retains up to six fractional digits. I added the missing planner case locally: MAKETIME(INT, INT, DOUBLE(scale=-1)) resolves to TIME(0) on this head instead of TIME(6). Because TIME-to-string formatting uses the result vector's scale, the execution layer can compute microseconds and then silently hide/drop them at the type boundary.

Second, rounded values crossing the upper TIME boundary are returned as NULL. MySQL adjusts otherwise-valid out-of-range TIME results to the nearest endpoint with a warning. A local regression test for MAKETIME(838, 59, 59.9999996) expects 838:59:59 but gets NULL on this head.

Please fix both type inference and endpoint normalization and add column/boundary regression coverage. Validation of the current intended paths: targeted function tests passed with -race -count=20; CI is green.

XuPeng-SH (changes requested):
Thanks for fixing the two blockers from the previous review: default DOUBLE scale now resolves to TIME(6), and fractional carry at the TIME endpoint is clamped correctly. Re-reviewing the current head found two new blocking correctness issues in MAKETIME.

1. The exact uint32 overload panics on valid input. list_builtIn.go registers (uint32, uint32, uint32), but each unsigned branch in MakeTime groups uint8/16/32/64 and constructs GenerateFunctionFixedTypeParameter[uint64]. That wrapper is a strict typed view; it does not widen the underlying vector. I reproduced this on the synthetic merge with a one-row uint32 call and got: panic: type mismatch: casting INT UNSIGNED vector to []uint64, originating at func_binary.go:7352. The same mismatch exists for minute and second at lines 7407 and 7456. Please either read each concrete width with its matching wrapper or ensure the binder always casts this overload to uint64, and add execution coverage for the registered uint32 overload.

2. Floating hour/minute conversion incorrectly truncates toward zero. MySQL MAKETIME calls val_int() for hour and minute, and Item_float::val_int() uses rint(): https://github.com/mysql/mysql-server/blob/trunk/sql/item_timefunc.cc#L3743-L3751 and https://github.com/mysql/mysql-server/blob/trunk/sql/item.h#L5602-L5609. MatrixOne float-to-integer cast also rounds rather than truncates. The math.Trunc calls at lines 7367 and 7419 therefore change both values and NULL/range behavior: 12.7/15.8 become 12/15 instead of 13/16; minute 59.9 becomes valid 59 instead of rounded 60 and NULL; minute -0.9 becomes valid 0 instead of rounded -1 and NULL; hour 838.9 becomes 838:00:00 instead of overflowing/clamping from 839. Please use MySQL-compatible integer conversion with explicit NaN/Inf/range guards, update the current truncation golden results, and cover positive/negative half boundaries plus minute values that cross 0 or 60 after rounding.

Validation: the seven current targeted tests pass on the synthetic merge, but they do not exercise uint32 and currently encode the truncation behavior. The uint32 probe fails deterministically with the panic above. CI is green, and the PR merges cleanly with current main.

pkg/sql/plan/function/list_builtIn.go
```diff
@@ -10737,6 +10737,50 @@ var supportedDateAndTimeBuiltIns = []FuncNew{
...
+	}
+	return types.T_time.ToTypeWithScale(scale)
+}
+
+func makeTimeCheck(overloads []overload, inputs []types.Type) checkResult {
+	if len(inputs) != 3 || (!inputs[0].Oid.IsMySQLString() && !inputs[1].Oid.IsMySQLString() && !inputs[2].Oid.IsMySQLString()) {
+		return fixedTypeMatch(overloads, inputs)
+	}
+
+	targetOids := []types.T{types.T_float64, types.T_float64, types.T_float64}
```
XuPeng-SH, line 10755: [P1] Keep the seconds operand exact until nanosecond-to-microsecond rounding. MySQL resolves argument 3 as NEWDECIMAL and calls `val_decimal`; routing VARCHAR (and, through the ordinary matcher, DECIMAL) through FLOAT64 creates a double-rounding error. Concrete repro: `MAKETIME(12, 59, '59.99999949999999999')`. This head parses the value as float64 `59.999999500000001` and `MakeTime` returns `13:00:00.000000`; MySQL keeps the first nine fractional digits (`999999499`), does not carry at 499 ns, and returns `12:59:59.999999`. The same failure applies to an exact DECIMAL seconds expression near this boundary. Please preserve VARCHAR/DECIMAL seconds through an exact decimal/nanosecond path and add below/at/above-500ns regression cases. MySQL reference: https://github.com/mysql/mysql-server/blob/trunk/sql/item_timefunc.cc#L3735-L3808

pkg/sql/plan/function/func_binary.go
```diff
@@ -4043,7 +4043,10 @@ func TimeFormat(ivecs []*vector.Vector, result vector.FunctionResultWrapper, pro
 // timeFormat: Get the format string corresponding to the time according to format specifiers
 // Only supports time-related format specifiers: %H, %h, %I, %i, %k, %l, %S, %s, %f, %p, %r, %T
 func timeFormat(ctx context.Context, t types.Time, format string, buf *bytes.Buffer) error {
-	hour, minute, sec, msec, _ := t.ClockFormat()
+	hour, minute, sec, msec, isNeg := t.ClockFormat()
+	if isNeg {
```
XuPeng-SH, line 4047: [P2] Preserve the empty-format NULL contract before writing the sign. MySQL returns NULL when the TIME_FORMAT format string has length zero, before `make_date_time` prefixes a negative sign. On this head `TIME_FORMAT(CAST('-01:00:00' AS TIME), '')` reaches this branch and returns the non-NULL string `"-"` (main returned an empty string), so the sign fix introduces a new observable wrong result at this boundary. Add the empty-format guard in `TimeFormat` and cover both positive and negative TIME values. MySQL reference: https://github.com/mysql/mysql-server/blob/trunk/sql/item_timefunc.cc#L2713-L2738

XuPeng-SH (changes requested):
Requesting changes on current head 8f007842 after a full re-review and latest-main synthetic merge.

The previous blockers are closed: default DOUBLE scale, endpoint clamping, exact uint32 vector getters, and MySQL-compatible float hour/minute rounding now behave correctly. HOUR and ordinary TIME_FORMAT sign/extended-hour paths are also sound.

Two correctness gaps remain:

1. MAKETIME still routes exact VARCHAR/DECIMAL seconds through FLOAT64. At the microsecond rounding boundary, `59.99999949999999999` becomes float64 `59.999999500000001`, so this head returns `13:00:00.000000` for hour=12/minute=59 while MySQL's decimal/nanosecond path returns `12:59:59.999999`. Preserve an exact seconds path and add below/at/above-boundary tests.
2. TIME_FORMAT does not reject an empty format before the new negative-sign prefix. A negative TIME with `''` now returns `"-"`; MySQL returns NULL. Add the empty-format guard and positive/negative regressions.

Validation on the synthetic merge with latest main:
- focused changed-path suite under `-race -count=20`: passed
- full `pkg/sql/plan/function`: passed
- dependent `pkg/sql/plan`: passed
- compile-only checks and `go vet`: passed
- dedicated exact-second probe: reproduced the wrong carry deterministically

Q1-Q3 audit found no resource ownership, wait-chain, or unbounded-growth issue; these findings are result-correctness blockers.

gouhongshen (changes requested):
## Codex automated review

MAKETIME's float-to-integer conversion is incompatible with MySQL at .5 boundaries.

### P1 - Use MySQL-compatible half-away-from-zero rounding for float hour/minute arguments (pkg/sql/plan/function/func_binary.go:7411)
Both float conversion paths use math.RoundToEven (hour at line 7411; minute at line 7441), but MySQL rounds these arguments half away from zero. For example, MySQL 8.4 returns MAKETIME(12.5,0,0) = '13:00:00', MAKETIME(-12.5,0,0) = '-13:00:00', and MAKETIME(12,58.5,0) = '12:59:00'; this implementation returns 12:00:00, -12:00:00, and 12:58:00 respectively. It also incorrectly accepts MAKETIME(12,-0.5,0) as 12:00:00 because banker’s rounding produces zero, whereas MySQL returns NULL. The newly added tests codify these incorrect results, so compatibility coverage will not catch the regression.

aunjgr (changes requested):
MAKETIME still uses banker rounding for floating hour and minute arguments, so half values such as 12.5 and -12.5 do not follow the required half-away-from-zero behavior. PR #25878 fixes this and supersedes this implementation. Please close this PR in favor of #25878.

aunjgr (changes requested):
MAKETIME still uses banker rounding for floating hour and minute arguments, so half values such as 12.5 and -12.5 do not follow the required half-away-from-zero behavior. PR #25878 fixes this and supersedes this implementation. Please close this PR in favor of #25878.

PR #25804, opened 2026-07-16
fix: make lock wait timeout handling safer

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24915 

fix: make lock wait timeout handling safer


aunjgr (approved):
Reviewed the timeout and cancellation lifecycle at the current head. Local and remote waiter ownership, timer notification, waiter removal, retry deadline propagation, response cleanup, and bounded worker queues close correctly. I found no lifecycle blocker.

aunjgr (approved):
Reviewed the timeout and cancellation lifecycle at the current head. Local and remote waiter ownership, timer notification, waiter removal, retry deadline propagation, response cleanup, and bounded worker queues close correctly. I found no lifecycle blocker.

gouhongshen (commented):
## Codex automated review

Found one non-blocking API correctness gap. The lock timeout, rollback, waiter cleanup, and remote lifecycle paths otherwise close correctly; current-head CI is green. Focused local tests were blocked by the missing prebuilt cgo/libmo.dylib.

### P2 - WithLockWaitTimeout(0) cannot clear an existing transaction's budget (pkg/util/executor/options.go:252)
This documented clearing path sets hasLockWaitTimeout=false. For Options.WithTxn(existingTxn), adjustOptions does not apply txnOpts, and applyExecutorLockWaitTimeout returns early; lockWaitTimeout then prefers the existing transaction's old timeout. Consequently, clearing an override while reusing a transaction retains the stale budget and can time out too early or wait too long. Track whether the option was explicitly supplied separately from whether its value is positive, and test the existing-transaction clearing path.

iamlinjunhong:
> ## Codex automated review
> Found one non-blocking API correctness gap. The lock timeout, rollback, waiter cleanup, and remote lifecycle paths otherwise close correctly; current-head CI is green. Focused local tests were blocked by the missing prebuilt cgo/libmo.dylib.
> 
> ### P2 - WithLockWaitTimeout(0) cannot clear an existing transaction's budget (pkg/util/executor/options.go:252)
> This documented clearing path sets hasLockWaitTimeout=false. For Options.WithTxn(existingTxn), adjustOptions does not apply txnOpts, and applyExecutorLockWaitTimeout returns early; lockWaitTimeout then prefers the existing transaction's old timeout. Consequently, clearing an override while reusing a transaction retains the stale budget and can time out too early or wait too long. Track whether the option was explicitly supplied separately from whether its value is positive, and test the existing-transaction clearing path.

fixed

gouhongshen (changes requested):
## Codex automated review

The timeout ceiling is not applied to ForwardTo RPC contexts, allowing mirrored lock calls to panic or exceed the configured budget. There is also a rolling-upgrade gap for clearing an existing timeout. Focused Go tests could not run because `go` is unavailable.

### P1 - Bound the ForwardTo RPC with the effective lock deadline (pkg/lockservice/service_forward.go:48)
`service.Lock` now injects `LockWaitTimeout`/`LockWaitDeadline`, and the regular remote-lock path derives a timeout context, but `forwardLock` still sends using the original `ctx`. A deadline-less context can trigger `morpc.Future.init`'s `context deadline not set` panic; a long-lived context is not bounded by the lockservice safety ceiling. This path is used for mirrored transactions via `options.ForwardTo`. Derive the RPC context from the effective deadline while preserving any earlier parent deadline.

### P2 - Preserve explicit zero timeout across older pipeline peers (proto/pipeline.proto:579)
`lock_wait_timeout_set` is required to distinguish an explicit zero from an omitted value, but older pipeline peers ignore field 11. When a background sender has no resolver, `applyExecutorLockWaitTimeout` serializes an explicit clear as field 10 = 0; an older peer then falls back to `client.LockWaitTimeoutFromTxn(txnOp)` and revives the existing positive transaction timeout. The wire format is syntactically compatible but the new clearing semantics are not during rolling upgrades. Add a version-compatible capability/default mechanism and test this mixed-version case.

pkg/lockservice/service.go
```diff
@@ -248,6 +253,62 @@ func (s *service) Lock(
...
+	effectiveDeadline := now.Add(time.Duration(effectiveSeconds) * time.Second)
+	if options.LockWaitDeadline > 0 {
+		callerDeadline := time.Unix(0, options.LockWaitDeadline)
+		if callerDeadline.Before(effectiveDeadline) {
+			effectiveDeadline = callerDeadline
+			effectiveSeconds = int64(effectiveDeadline.Sub(now) / time.Second)
+			if effectiveDeadline.Sub(now)%time.Second != 0 {
+				effectiveSeconds++
+			}
+			if effectiveSeconds <= 0 {
```
XuPeng-SH: **[P1] Do not restart an expired absolute lock-wait budget**

When `callerDeadline <= now`, this branch forces `effectiveSeconds` to 1 while preserving an already-expired `LockWaitDeadline`. The consumers do not enforce that absolute timestamp: the synchronous path creates a fresh `context.WithTimeoutCause(..., 1s)`, and `getAsyncLockWaitDeadline` reconstructs another deadline from `LockWaitTimeout`. Therefore an RPC/forward hop that used up the budget still enters the waiter queue with a new one-second budget (and can interact with the 30-second RPC slack).

I reproduced this through the real `service.Lock` conflict path: with a held row, `LockWaitTimeout=60`, and `LockWaitDeadline=now-1s`, the call returned `ErrLockWaitTimeout` after ~1.000s rather than immediately. Reject an expired deadline at service entry, or make both sync and async lock contexts consume the absolute deadline directly. Please add an expired-deadline/delayed-hop regression test.

XuPeng-SH (changes requested):
## Deep review — changes requested

I reviewed the exact current head across the frontend → executor → process codec → local/remote lockservice lifecycle.

The previous `WithLockWaitTimeout(0)` finding is fixed: explicit presence now survives existing-transaction reuse and current-version pipeline encoding. I also independently confirmed the already-reported `ForwardTo` RPC deadline gap.

I found one additional blocking deadline bug: an absolute `LockWaitDeadline` that is already expired at a service/forward hop is converted into a fresh one-second relative timeout. A deterministic local lock-conflict test with `LockWaitDeadline = now-1s` returned after ~1.000s instead of failing before queue admission. This breaks the stated single-budget-across-hops invariant and is particularly relevant when RPC admission/forwarding consumes the original budget.

Validation on this head:
- all newly added focused tests passed;
- complete `pkg/lockservice` suite passed (122s);
- focused local/remote timeout tests passed under `-race`;
- `go vet` passed for all affected handwritten packages;
- current-head CI is green.

Separate follow-up risk (present in the base, but made reachable for timeout-less async callers by the new ceiling): `waiterEvents.check` holds `mw.mu` while synchronously feeding its own bounded `eventC`. A diagnostic batch of `cap(eventC)+1` expired waiters self-blocked. This should be tracked/fixed, but the inline comment below is the direct PR regression blocker.

Issue: #24915.

XuPeng-SH (approved):
Reviewed latest head 8b5fe8ed after re-reading the complete diff and all historical review threads.

All previously blocking paths are now closed:
- ForwardTo derives its RPC context from the effective lock deadline while preserving an earlier parent deadline.
- an already-expired absolute deadline fails before queue admission instead of becoming a fresh one-second budget.
- explicit WithLockWaitTimeout(0) survives existing-txn reuse and new/old pipeline peers without reviving stale state.
- the full-eventC timeout path no longer self-blocks while holding the waiter mutex.

Q1-Q3 lifecycle audit: pooled RPC responses and waiter references/timers have symmetric cleanup; cancellation, timeout, retry, remote forwarding, and rolling-upgrade waits all have bounded release paths; reused waiters/processes reset generation-local timeout state; queues and admission are bounded. No merge-blocking leak, double cleanup, hang, or OOM risk found.

- focused new timeout/deadline/codec/rollback tests passed;
- concurrency-sensitive lockservice and lockop cases passed under -race;
- complete pkg/lockservice and pkg/sql/colexec/lockop suites passed;
- git diff --check passed; current required CI is green.

Approved.

gouhongshen (changes requested):
## Codex automated review

The new absolute lock-wait budget is still bypassed during lock-table binding/allocation.

### P1 - Recheck the deadline after lock-table binding (pkg/lockservice/service.go:198)
The new check at lines 181-187 only covers `s.wait()`. `getLockTableWithCreate` can then block on an in-flight allocation or call `getLockTableBind` with `context.Background()` (its own 10s RPC timeout), without consuming or checking the caller's deadline. Once it returns, `l.lock` can succeed on an uncontended table: `localLockTable.doLock` acquires and completes at lines 114-142 without testing `lockWaitDeadline`. Thus a 1s lock budget can wait through a slow bind lookup and succeed well after expiry, contrary to the new admission-expiry behavior; the same lookup-before-dispatch gap exists on forwarding and remote handling. Propagate the effective deadline into binding/allocation waits or recheck immediately afterward, with a delayed-bind regression test.

iamlinjunhong:
> ## Codex automated review
> The new absolute lock-wait budget is still bypassed during lock-table binding/allocation.
> 
> ### P1 - Recheck the deadline after lock-table binding (pkg/lockservice/service.go:198)
> The new check at lines 181-187 only covers `s.wait()`. `getLockTableWithCreate` can then block on an in-flight allocation or call `getLockTableBind` with `context.Background()` (its own 10s RPC timeout), without consuming or checking the caller's deadline. Once it returns, `l.lock` can succeed on an uncontended table: `localLockTable.doLock` acquires and completes at lines 114-142 without testing `lockWaitDeadline`. Thus a 1s lock budget can wait through a slow bind lookup and succeed well after expiry, contrary to the new admission-expiry behavior; the same lookup-before-dispatch gap exists on forwarding and remote handling. Propagate the effective deadline into binding/allocation waits or recheck immediately afterward, with a delayed-bind regression test.

fixed

heni02 (approved):

gouhongshen (approved):
## Codex automated review

One non-blocking configuration-precision issue.

### P2 - Honor sub-second max-lock-wait-duration ceilings (pkg/lockservice/service.go:294)
`MaxLockWaitDuration` accepts any positive `toml.Duration`, but a valid `500ms` ceiling is rounded up to one second here and the authoritative deadline is then set to `now + 1s` at line 308. Local and remote waiters consume that absolute deadline, so a contended request can wait almost twice the configured ceiling. Preserve the whole-second value only for legacy consumers, while deriving the absolute deadline from the unrounded configured duration (or reject sub-second configurations), and add a fractional-ceiling test.

Issue #25802, opened 2026-07-16
[Bug]: hashjoin panic: nil pointer in (*container).finalize (join.go:716) on multi-CTE LEFT JOIN query


heni02:
## Summary

Executing a multi-CTE query with several `LEFT JOIN`s (jinpan `bill_collection_p4` template) causes an internal panic:

```text
(20101) internal error: panic runtime error: invalid memory address or nil pointer dereference
  hashjoin.(*container).finalize  join.go:716
  hashjoin.(*HashJoin).Call       join.go:237
```

Observed twice in the same AP Regression join suite (`q0503`, `q0751`). Same stack; SQL differs only by outer `LIMIT` (300 vs 100). Same template has ~33 other variants that succeed, so this looks plan/data-path specific rather than a pure syntax issue.

## Environment

| Item | Value |
|------|-------|
| MO commit | [`f8aaa25e670d0936b14602becd70d5966b1a87a1`](https://github.com/matrixorigin/matrixone/commit/f8aaa25e670d0936b14602becd70d5966b1a87a1) |
| Image tag | `jinpan-f8aaa25` |
| CI | [mo-nightly-regression AP Regression #29470315932](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29470315932/job/87539496633) (Fact+Dim Join Query Tests) |
| Case branch | `big_data_jinpan` |
| Dataset | jinpan_001 (S3-loaded fact/dim + flat/receipts tables) |

## Repro SQL (q0503)

```sql
WITH perf AS (
  SELECT VBELN, HTZJE FROM jst_receipts_tables.t_s4_performance
  WHERE VBELN IS NOT NULL AND VBELN <> ''
    AND HTZJE > 50000
  LIMIT 5000
),
vat_agg AS (
  SELECT v.VBELN, SUM(v.ZSAPAMT) AS total_sap_amt, MAX(v.ZGTCD) AS zgtcd, MAX(v.ZGTID) AS zgtid
  FROM jst_flat_table.vat_sales_invoice v
  INNER JOIN perf p ON v.VBELN = p.VBELN
  GROUP BY v.VBELN
),
pc_agg AS (
  SELECT pc.zuonr AS VBELN, SUM(pc.wsl) AS total_wsl, COUNT(*) AS pc_cnt
  FROM jst_flat_table.payment_collection pc
  INNER JOIN perf p ON pc.zuonr = p.VBELN
[...]

PR #25809, opened 2026-07-16
fix: avoid nil rightMatchedIter panic in hashjoin finalize on early teardown

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25802

Fixes a nil pointer panic in `hashjoin.(*container).finalize` (join.go:716) hit by multi-CTE LEFT JOIN queries with an outer LIMIT.

**Root cause.** In a parallel (NumCPU > 1) right-flavored hash join (`EmitUnmatchedBuild() == true`), the merger's `syncBitmap` can receive a nil bitmap from the shared channel and returns early **without initializing `rightMatchedIter`**. The nil arrives when:

1. a non-merger worker is Reset before syncing its bitmap — its `Reset` intentionally sends nil so the merger doesn't block forever. This happens whenever the query is torn down early, e.g. the outer `LIMIT 300`/`LIMIT 100` in the failing q0503/q0751 is satisfied mid-probe; or
2. `proc.Ctx` is canceled while the merger waits in `ReceiveBitmapFromChannel`.

The `Call` state machine then advanced from `SyncBitmap` to `Finalize` because it only checked `rightRowsMatched != nil` (the merger's own bitmap always exists), and `finalize` dereferenced the nil iterator. This explains why only the two LIMIT variants of the 35-query template panicked: it is a race between early LIMIT termination and bitmap synchronization.

**Fix.** Gate the `SyncBitmap -> Finalize` transition on `ctr.rightMatchedIter != nil` — exactly the invariant `finalize` depends on, and only true when `syncBitmap` ran to completion. All early-exit paths (non-merger worker, no bitmap, merge aborted by teardown/cancellation) now go to `End`. When the merge is aborted the query is being torn down anyway, so no unmatched-build rows need to be emitted.

**Tests.** Two new unit tests:
- `TestHashJoinMergerSyncBitmapAborted` reproduces the exact panic stack from the issue without the fix (merger receives nil from a torn-down worker) and asserts a clean `ExecStop` with the fix;
- `TestHashJoinMergerFinalizeEmitsUnmatchedBuildRows` guards the healthy path: the merger merges a real worker bitmap and still emits the unmatched build rows.

No BVT case is added: the bug is a teardown race between parallel pipelines that cannot be triggered deterministically from SQL.

**Note (out of scope).** While investigating, we noticed the parallel + spill + right-join combination has deeper pre-existing issues (exclusive `TakeSpillBuildFds` on a shared JoinMap; workers reach `End` after the first bucket). Those are unrelated to this panic and left for a separate issue.

### DIFF
--- a/pkg/sql/colexec/hashjoin/join.go
+++ b/pkg/sql/colexec/hashjoin/join.go
@@ -225,7 +225,12 @@ func (hashJoin *HashJoin) Call(proc *process.Process) (vm.CallResult, error) {
 				return result, err
 			}
 
-			if ctr.rightRowsMatched == nil || (hashJoin.NumCPU > 1 && !hashJoin.IsMerger) {
+			// Only enter Finalize when syncBitmap ran to completion and set
+			// the iterator. It stays nil for non-merger workers, when there
+			// is no bitmap at all, or when the merger observed teardown (a
+			// worker sent a nil bitmap on Reset, or the context was
+			// canceled) — in all these cases there is nothing to finalize.
+			if ctr.rightMatchedIter == nil {
 				ctr.state = End
 			} else {
 				ctr.state = Finalize
--- a/pkg/sql/colexec/hashjoin/join.go
+++ b/pkg/sql/colexec/hashjoin/join.go
@@ -689,12 +689,20 @@ func (ctr *container) syncBitmap(hashJoin *HashJoin, proc *process.Process) erro
 
 			for cnt := 1; cnt < int(hashJoin.NumCPU); cnt++ {
 				v := colexec.ReceiveBitmapFromChannel(proc.Ctx, hashJoin.Channel)
-				if v != nil {
-					matchedCnt += v.Count()
-					ctr.rightRowsMatched.Or(v)
-				} else {
+				if v == nil {
+					// A worker was torn down before syncing (its Reset sends
+					// nil) or the context was canceled. The merge is aborted,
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
Reviewed head 4ad70a31265764c347db443840a99d4954ded972 across the HashJoin state machine, Call/Reset teardown protocol, channel synchronization, cancellation, spill transitions, and cleanup paths.

Gating Finalize on rightMatchedIter directly enforces the invariant finalize dereferences and correctly routes aborted bitmap synchronization to End. For the affected non-spill parallel right-flavored join path, each non-merger contributes exactly one generation signal: its bitmap after reaching syncBitmap, or nil from Reset when torn down earlier. The channel is DOP-sized, the merger consumes at most NumCPU-1 messages, and context cancellation terminates receive waits. Draining the remainder after an abort prevents buffered messages from contaminating a subsequent generation without changing healthy unmatched-build emission.

Q1: no new resource ownership or cleanup gap. Q2: no unbounded wait found in the affected protocol. Q3: channel/message accumulation is bounded by DOP. The healthy finalize path and abort path are both covered.

Validation: full hashjoin package test passed; full package race test passed; the two new tests passed under -race with count=50; go vet passed with the repository CGo environment. CI is green. No blocking findings.

Issue #25794, opened 2026-07-16
[Bug]: CDC pause race leaves daemon Paused while mo_cdc_task remains running


Ariznawlll:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Related but not duplicate: #24797 / #24804 addressed the inverse inconsistency (`mo_cdc_task=paused` while daemon task was still `PauseRequested`). This report observes daemon task `Paused` while `mo_cdc_task` remains `running`.

## Branch Name

`main`

## Commit ID

`30df79c47bff5a783658ec9df1d6980e64f4384f`

## Other Environment Information

- Deployment: TKE regression cluster, 3 CN
- Workflow: `Binary Search On TKE New`, run attempt 5
- MOTR branch: `codex/add-motr-issue-regressions`
- MOTR commit: `987659dd969b74ff1d780c28f7e0eb0b9f50018a`
- mo-nightly-regression branch: `agent/test-motr-pr39-main`
- mo-nightly-regression commit: `8a5516a4ce5b267d044b3ca7de048fe0b5b391a2`

## Actual Behavior

The MOTR case `03_cdc.gc-4` creates CDC task `cdc_gc_4` and immediately pauses it. For 60 seconds:

- `select state from mo_catalog.mo_cdc_task where task_name = 'cdc_gc_4'` continued to return `running`.
- Repeated `pause cdc task cdc_gc_4` returned error 20101 saying the internal daemon task was already `Paused`.

The test aborted with:

```text
pause cdc task cdc_gc_4 did not reach paused within 60s
(last state: running; last action: ERROR 20101: internal error:
Task cdc_gc_4 status can not be change, now it is Paused)
```

Failed job:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29411310354/job/87540670604

Run:
https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29411310354

The parallel MOTR run returned `MOTR_PARALLEL_RC=1`; the serial phase passed with `MOTR_SERIAL_RC=0`. This was the only failed test.

The same MatrixOne and MOTR commits passed `GC-4` in the previous run attempt, where it completed in about 22 seconds. This suggests a timing-sensitive race rather than a deterministic syntax/configuration failure.

## Expected Behavior

After `pause cdc task` is accepted, the internal daemon task and `mo_catalog.mo_cdc_task.state` should converge to `Paused`/`paused` within the timeout.

[...]

PR #25810, opened 2026-07-16
fix(cdc): guard task state transitions

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25794

- Guard CDC catalog state transitions with current-state predicates and affected-row checks.
- Change `updateErrMsg` to update `mo_catalog.mo_cdc_task` only when the catalog row is still `running`, so startup clear-error and failed-error updates cannot overwrite a concurrent pause.
- Make affected-row validation state-aware: when an update affects 0 rows, read back the catalog state and treat “already at target state” as idempotent success, while conflicting states and missing rows return explicit errors.
- Keep the `pausing -> paused` pause-complete hook durable/idempotent across task-runner restart or crash-after-SQL-commit retries.
- Add a CDC state-transition failpoint before catalog updates so the race window can be exercised deterministically.

## Validation
- Built `mo-service` with `make build`, including thirdparties and `cgo/libmo.so`.
- Started a local MO service from the current binary with an isolated temporary launch config:
  - config: `/tmp/mo-25794-launch/launch.toml`
  - data: `/tmp/mo-25794-data`
  - SQL: `127.0.0.1:26001`
  - status: `27001`
  - logservice: `42000/42001/42002`
- Verified live SQL connectivity:
  `mysql -h127.0.0.1 -P26001 -uroot -p111 -e "select 1 as ok;"`
- Replayed the issue-style CDC create/pause flow on the live service with the required PITR prerequisite:
  - created `case_gc_4` and `case_gc_4_sink`
  - created `pitr_case_gc_4 for database case_gc_4 range 2 'h'`
  - created `cdc_gc_4`
  - paused after the task reached `running`
  - observed catalog states `pausing -> paused`; final row was `state = 'paused'`, `err_msg = ''`
- Replayed a tighter immediate-pause loop on the live service:
  - created `case_gc_4_loop`, `case_gc_4_loop_sink`, `pitr_case_gc_4_loop`, and `cdc_gc_4_loop`
  - sent `pause cdc task cdc_gc_4_loop` repeatedly immediately after create
  - observed early daemon `Created` rejections, then one accepted pause, catalog `running -> pausing -> paused`
  - final row was `state = 'paused'`, `err_msg = ''`; it did not remain `running`
- Covered deterministic concurrent state changes with failpoint injection in `TestCDCStateTransitionDoesNotOverwriteConcurrentCatalogState`:
  - `running -> failed`: blocked before catalog update, concurrently changed catalog state to `pausing`; final state stayed `pausing`, read-back classified it as a conflict
  - `running -> running`: blocked startup clear-error update before catalog update, concurrently changed catalog state to `pausing`; final state stayed `pausing`, read-back classified it as a conflict. This covers the issue #25794 suspected overwrite path.
  - `pausing -> paused`: blocked pause-complete hook before catalog update, concurrently changed catalog state to `failed`; final state stayed `failed`, read-back classified it as a conflict
- Added idempotency regressions for the review request:
  - `TestCDCPauseTaskCompleteHookAlreadyPausedIsIdempotent`: invokes the pause-completion hook once for `pausing -> paused`, then invokes it again as if the task-runner in-memory completion marker was lost; the retry succeeds when catalog is already `paused`
  - `TestCDCTaskUpdateErrMsgAlreadyInTargetStateAllowsZeroAffectedRows`: verifies `running -> running` with 0 affected rows succeeds when read-back state is already `running`
  - conflict and missing-row tests verify 0 affected rows is not blindly accepted
- Ran targeted tests after the final change:
  - `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/cdc -run 'TestUpdateTaskStateAndErrMsg|TestGetTaskStateSQL'`
  - `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/frontend -run 'TestRegisterCdcExecutor|TestCDCPauseTaskCompleteHook|TestCDCTaskUpdateErrMsg|TestCDCStateTransitionDoesNotOverwriteConcurrentCatalogState'`
  - `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/taskservice -run 'TestPauseTasksRetriesPausedCDCFinalize'`


XuPeng-SH (changes requested):
Blocking correctness/liveness issue: the new affected-row check makes pause completion non-idempotent and creates a permanent retry loop after a task-runner restart.

The task-runner contract explicitly retries CDC tasks already persisted as `Paused` when their completion marker is absent. That marker lives only in `pauseCompletedTasks.m`, so it is lost on every runner restart. A normal sequence is therefore:

1. The first hook commits `mo_cdc_task: pausing -> paused`.
2. The runner restarts, or crashes after the SQL commit; the in-memory completion marker is absent.
3. `pauseTasks` polls the persisted `Paused` daemon task and invokes the hook again.
4. The retry executes `UPDATE ... WHERE state = 'pausing'`; the catalog is already `paused`, so it affects zero rows.
5. `execCDCSQLWithAffectedRows` returns an error for every result other than exactly one row. `pauseTaskCompleted` therefore never marks completion, and the same task is re-enqueued forever.

I reproduced this on head `e635c36eaa` with a focused test that invokes the pause-completion hook for the already-applied state: it deterministically returns `internal error: cdc task state transition affected 0 rows ... current_state=pausing, target_state=paused`. This conflicts with the existing task-runner invariant that pause is idempotent and with issue #25794's requirement that repeated pause converge.

An affected-row count of zero is ambiguous; it cannot by itself be classified as a state mismatch. At minimum, distinguish:

- current state already equals the target: idempotent success;
- current state is a conflicting state: preserve it and return the intended conflict result;
- task row is absent because it was canceled/deleted: classify that terminal race explicitly.

Please make the completion transition durable and idempotent, and add a regression that clears/recreates the runner completion state after a successful `pausing -> paused` transition, then proves the retry succeeds and stops being polled. The same helper is also used for `running -> running`; an unchanged row may legitimately report zero affected rows, so a blanket `AffectedRows == 1` rule should not be used as the state-machine oracle.

XuPeng-SH (approved):
Re-reviewed head 842ba9ad1d against the latest main. The previous restart/idempotency blocker is resolved: zero affected rows are now classified by durable catalog state, so an already-applied pausing-to-paused transition succeeds while conflicts and missing rows remain explicit errors. I also verified the production executor supports affected-row status, traced pause retry termination across task-runner restart/cancel/resume races, and found no new correctness, liveness, resource-growth, or material performance blocker. The virtual merge is clean; full pkg/cdc, pkg/frontend, and pkg/taskservice tests pass, and the deterministic state-transition concurrency regression passes with -race -count=10.

PR #25811, opened 2026-07-16
fix(morpc): avoid stream sequence gaps for skipped responses

### LINKED ISSUE #25803: [Bug]: ErrStreamClosed / context canceled on WITH RECURSIVE + bkpf/bseg join (jinpan payment_details_tree)
## Summary

In jinpan_001 Fact+Dim join regression, several `payment_details_tree` queries fail within **1–3 seconds** with:

- `(20503) rpc stream closed` (`ErrStreamClosed`) — 4 hits
- `(1105) context canceled` — 1 hit on the same template

Same template succeeds in **24/34** runs in the same suite, so this is intermittent and not a SQL syntax/schema error. Client timeout is 1800s; these are not client timeouts.

`20503` and `1105` here look like **two client-visible faces of the same failure** (pipeline/stream torn down → query context canceled).

## Environment

| Item | Value |
|------|-------|
| MO commit | [`f8aaa25e670d0936b14602becd70d5966b1a87a1`](https://github.com/matrixorigin/matrixone/commit/f8aaa25e670d0936b14602becd70d5966b1a87a1) (`jinpan-f8aaa25`) |
| Note | This commit is the CN pipeline FIN/ACK / connection-reuse change (#25713) — possibly related |
| CI | [AP Regression #29470315932 / Fact+Dim Join](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29470315932/job/87539496633) |
| Case branch | `big_data_jinpan` |
| Dataset | jinpan_001 |

## Hits (same template family)

| Case | Error | Wall clock to fail |
|------|-------|--------------------|
| `q0385_payment_details_tree` | `(20503) rpc stream closed` | ~3s |
| `q0447_payment_details_tree` | `(20503) rpc stream closed` | ~2s |
| `q0664_payment_details_tree` | `(20503) rpc stream closed` | ~2s |
| `q0911_payment_details_tree` | `(20503) rpc stream closed` | ~1s |
| `q0695_payment_details_tree` | `(1105) context canceled` | ~2s |

Case files under `mo-nightly-regression`:
`tools/mo-regression-test/cases/jinpan_001/join/j_batch_{020,023,034,035,046}.sql`

After each failure, the **next** SQL in the same session often succeeds immediately (cluster not fully dead).

## Repro SQL (q0385)

```sql
WITH w AS (
WITH RECURSIVE acct_tree AS (
  SELECT cpmb_kgd4b76 AS node_id, parenth1 AS parent_id, 0 AS lvl
  FROM dwd_dcp.dwd_bw_1cpmb_bkgd4b76
  WHERE cpmb_kgd4b76 = '200000'
  UNION ALL
  SELECT c.cpmb_kgd4b76, c.parenth1, t.lvl + 1
  FROM dwd_dcp.dwd_bw_1cpmb_bkgd4b76 c
  INNER JOIN acct_tree t ON c.parenth1 = t.node_id
  WHERE t.lvl < 3
)
SELECT
  CONCAT(b.belnr, '-', b.bukrs, '-', b.gjahr) AS voucher_header_id,
  b.belnr, b.bukrs, b.gjahr, b.bldat,
  s.buzei, s.hkont, s.dmbtr, s.zuonr,
  t.node_id AS acct_node, t.lvl,
  pd.payment_type
FROM dwd_dcp.dwd_s4_bkpf b
INNER JOIN dwd_dcp.dwd_s4_bseg s ON b.bukrs = s.bukrs AND b.belnr = s.belnr AND b.gjahr = s.gjahr
INNER JOIN acct_tree t ON s.hkont = t.node_id
LEFT JOIN jst_receipts_tables.payment_details pd
  ON pd.belnr = b.belnr AND pd.bukrs = b.bukrs AND pd.gjahr = b.gjahr
WHERE b.gjahr = '2024' AND s.dmbtr <> 0
  AND b.bukrs IN ('2000', '3000')
  AND b.belnr >= '8' AND b.belnr < '9'
LIMIT 2000
)
SELECT * FROM w
WHERE 1 = 1
LIMIT 50;
```

Other hits are the same shape; only recursive root (`100000`/`200000`/`600000`), `lvl` bound, company-code / belnr filters, and outer `LIMIT` vary.

## Why this looks like an MO bug (not case)

1. Error is engine `ErrStreamClosed` / canceled context, not parser/schema.
2. Same template succeeds most of the time in the same run → intermittent engine/pipeline path.
3. Failures cluster on `WITH RECURSIVE` + large fact joins (`bkpf`/`bseg`).
4. Fail in 1–3s under 1800s client timeout → not harness timeout.
5. Tested MO build centers on pipeline stream lifecycle (#25713); worth regression-checking with `disable-stream-reuse` / prior commit.

## Related but uncertain (appendix)

One additional hit in the same job:

| Case | Error |
|------|-------|
| `q0190_invoice_acdoca_link` | `(1105) context canceled` (<1s) |

Different SQL (multi-CTE invoice ↔ acdoca/bseg, **no** recursive CTE). Same error class; **unclear if same root cause** — include for correlation only.

## Also seen in same suite (separate)

HashJoin nil panic on `bill_collection_p4`: https://github.com/matrixorigin/matrixone/issues/25802 — different stack/template; do not merge unless investigation proves shared cause.

## Expected

Query completes (rows or empty set) without `ErrStreamClosed` / unexpected `context canceled`. If the CN aborts, the client error should surface the underlying panic/error, not only stream closed.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25803

MORPC previously assigned stream response sequence numbers when responses entered the server send queue. If a queued response expired or was canceled before the write loop processed it, that response was skipped after already consuming its sequence number. The next valid response then reached the client with a gap, causing `sequence out of order` followed by `rpc stream closed`.

This PR:

- assigns stream sequence numbers in the single server write loop after filter, context, and timeout checks;
- synchronizes write-time assignment with session close so retired sequence state cannot be missed or resurrected;
- adds regression coverage for canceled responses and the close/assignment boundary;
- suppresses automatic ERROR reporting for the expected, typed `RemoteDispatchNotRegistered` retry response while preserving its wire code and terminal timeout behavior.

Validation:

- `go test -race -count=100 -run "TestCanceledStreamResponseDoesNotCreateSequenceGap|TestAssignStreamSequenceRejectsClosedSession" ./pkg/common/morpc`
- `go test -race -count=1 ./pkg/common/morpc`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 ./pkg/sql/compile`
- focused compile race tests
- `go build ./pkg/common/morpc ./pkg/sql/compile`
- `go vet ./pkg/common/morpc ./pkg/sql/compile`

### DIFF
--- a/pkg/common/morpc/server.go
+++ b/pkg/common/morpc/server.go
@@ -415,6 +415,11 @@ func (s *server) startWriteLoop(cs *clientSession) error {
 						continue
 					}
 
+					if !cs.assignStreamSequence(&f.send) {
+						cs.releaseMessage(f.send)
+						f.messageSent(backendClosed)
+						continue
+					}
 					timeout += v
 					// Record the information of some responses in advance, because after flush,
 					// these responses will be released, thus avoiding causing data race.
@@ -718,14 +723,6 @@ func (cs *clientSession) send(msg RPCMessage) (*Future, error) {
 		return nil, moerr.NewClientClosedNoCtx()
 	}
 
-	id := response.GetID()
-	if v, ok := cs.sentStreamSequences.Load(id); ok {
-		seq := v.(uint32) + 1
-		cs.sentStreamSequences.Store(id, seq)
-		msg.stream = true
-		msg.streamSequence = seq
-	}
-
 	f := cs.newFutureFunc()
 	f.init(msg)
 	if !f.oneWay {
@@ -743,6 +740,27 @@ func (cs *clientSession) send(msg RPCMessage) (*Future, error) {
 	}
 	cs.metrics.sendingQueueSizeGauge.Set(float64(len(cs.c)))
 	return f, nil
+}
+
+// assignStreamSequence runs in the single server write loop after a response
+// has passed the filter and context checks. Assigning the sequence at enqueue
+// time leaves a permanent gap when the queued response expires before it is
+// written, causing the client to tear down an otherwise healthy stream.
+func (cs *clientSession) assignStreamSequence(msg *RPCMessage) bool {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+	if cs.mu.closed {
+		return false
+	}
+
+	id := msg.Message.GetID()
+	if v, ok := cs.sentStreamSequences.Load(id); ok {
+		seq := v.(uint32) + 1
+		cs.sentStreamSequences.Store(id, seq)
+		msg.stream = true
+		msg.streamSequence = seq
+	}
+	return true
 }
 
 func (cs *clientSession) releaseMessage(msg RPCMessage) {
--- a/pkg/common/morpc/server.go
+++ b/pkg/common/morpc/server.go
@@ -561,6 +561,71 @@ func (s *server) getSessionCount() int {
 	return n
 }
 
+// sentStreamState owns the complete lifecycle of server-side stream response
+// sequences. Its lock is never held across queue operations, network I/O, or
+// future waits.
+type sentStreamState struct {
+	mu        sync.Mutex
+	closed    bool
+	sequences map[uint64]uint32
+}
+
+func (s *sentStreamState) start(id uint64) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.closed {
+		return false
+	}
+	if s.sequences == nil {
+		s.sequences = make(map[uint64]uint32)
+	}
+	if _, ok := s.sequences[id]; ok {
+		return false
+	}
+	s.sequences[id] = 0
+	return true
+}
+
+func (s *sentStreamState) next(id uint64) (uint32, bool, bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.closed {
+		return 0, false, false
+	}
+	seq, ok := s.sequences[id]
+	if !ok {
+		return 0, false, true
+	}
+	seq++
+	s.sequences[id] = seq
+	return seq, true, true
+}
+
+func (s *sentStreamState) finish(id uint64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	delete(s.sequences, id)
+}
+
+func (s *sentStreamState) close() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.closed {
+		return 0
+	}
+	s.closed = true
+	count := len(s.sequences)
+	clear(s.sequences)
+	return count
+}
+
+func (s *sentStreamState) contains(id uint64) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, ok := s.sequences[id]
+	return ok
+}
+
 type clientSession struct {
 	metrics       *serverMetrics
 	codec         Codec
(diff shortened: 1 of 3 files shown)


XuPeng-SH (changes requested):
Request changes on current head `cb05aff850b3fee9fa4b7a68aa1cba1e7c75af86`. Moving sequence assignment after the skip/timeout checks fixes the reported gap, but the new `cs.mu.RLock()` in the write loop introduces a queue-saturation close deadlock.

### [P1] Avoid the `send` / `Close` / write-loop RWMutex cycle

`clientSession.send` holds `cs.mu.RLock()` while it may block sending to the bounded 1024-entry `cs.c` (`server.go:715-740`). `Close` needs `cs.mu.Lock()` (`server.go:620-623`). This PR makes the write loop call `assignStreamSequence`, which also needs `cs.mu.RLock()` (`server.go:418`, `server.go:749-763`).

The following schedule forms a closed wait cycle:

1. The write loop has already fetched a response batch. Producers refill `cs.c`; one more sender blocks on the full channel while still holding `cs.mu.RLock`.
2. A concurrent session `Close` waits for `cs.mu.Lock`. Go `RWMutex` blocks new readers once that writer is waiting.
3. The write loop reaches the fetched response and blocks in `assignStreamSequence` waiting for `cs.mu.RLock`.
4. The write loop cannot finish the batch and drain `cs.c`; the sender cannot enqueue and release its old read lock; `Close` cannot acquire the write lock.

There is no guaranteed timeout escape: `AsyncWrite` calls `send` with `context.Background()`, and `ClientSession.Write` also accepts contexts without deadlines. This can permanently hang response writers and session cleanup under a slow connection / response burst plus concurrent close.

Please remove this lock cycle rather than adding a timeout. In particular, the write loop must not need a lock that `Close` waits for while `send` can hold that lock across a blocking queue operation. Add a deterministic regression that models an already-fetched response, refills `cs.c`, blocks another background-context sender, starts `Close`, and verifies sequence assignment/write progress and close both terminate.

Local checks: the PR's two focused race tests pass at `-count=100`, and the full `pkg/common/morpc` suite passes; neither currently exercises this saturated-queue schedule.

aptend:
Addressed the queue-saturation deadlock on head `245e60673`.

The write loop no longer acquires `cs.mu`. Sent-stream sequence membership and allocation are now owned by a small terminal `sentStreamState`; its private lock is held only for short map/state transitions and never across queue operations, network I/O, or Future waits.

Added `TestAssignStreamSequenceProgressesWhileCloseWaitsOnFullQueue`, which deterministically:

1. fills `cs.c` and blocks an `AsyncWrite` while it holds `cs.mu.RLock`;
2. starts `Close` and observes the pending writer via `TryRLock`;
3. verifies sequence assignment still progresses;
4. releases queue capacity and verifies both sender and Close terminate.

Validation passed: focused concurrency tests with `-race -count=100`, full `pkg/common/morpc` under `-race`, compile package tests and focused race tests, build, vet, and an independent concurrency review.

XuPeng-SH (approved):
Re-reviewed current head `26036f41b0c038ac5fa47c3cfb38d5044b181d8f`. The previous P1 queue-saturation deadlock is fixed.

The write loop no longer acquires `cs.mu`; sent-stream membership/allocation now lives behind the independent `sentStreamState` lock, which is held only for bounded map transitions. The old cycle (`send` blocked on full queue while holding `cs.mu.RLock` -> `Close` waiting for `cs.mu.Lock` -> write loop waiting for a new `cs.mu.RLock`) is therefore broken. Close permanently seals and clears sequence state, and later `start`/`next` calls cannot resurrect it. Normal `FinishStream` retirement and session-close cleanup remain symmetric, while sequence assignment occurs only after filter/context/timeout checks.

The typed RemoteDispatchNotRegistered retry still preserves its wire code and terminal timeout; NoReportContext only suppresses expected retry log noise.

Local verification passed: saturated-queue/canceled-response race tests at `-count=100`, full `pkg/common/morpc` under `-race`, focused compile tests and race tests, plus build/vet for the changed packages. No blocking findings remain. GitHub CI for the latest merge is still running.

PR #25819, opened 2026-07-16
fix(lockservice): fence stale binds by allocator epoch (#24906)

iamlinjunhong:

Approved by: @aptend, @XuPeng-SH, @aunjgr

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24896 

fix(lockservice): fence stale binds by allocator epoch (#24906)


pkg/lockservice/lock_table_remote.go
```diff
@@ -503,8 +497,40 @@ func (l *remoteLockTable) maybeHandleBindChanged(resp *pb.Response) error {
...
+			"remote-new-bind",
+			l.bind,
+			refreshedBind,
+			allocator,
+			requestAllocator); err != nil {
+			return err
+		}
+		return ErrLockTableBindChanged
+	}
+	l.bindChangedHandler(newBind)
```
XuPeng-SH, line 543: [P1] Validate same-ID `NewBind` against the current allocator state. This branch compares the response only with `l.bind`. If a request was sent under A0, A1 is observed and purges the table while the RPC is in flight, then A0 returns a `NewBind` carrying the same allocator ID as `l.bind`, this direct call reinstalls the superseded A0 bind. I reproduced exactly that sequence: the A1 observation removed one table, and `maybeHandleBindChanged` recreated it with `AllocatorID=A0`. Route this result through allocator-aware validation/current-allocator refresh before publishing.

pkg/lockservice/lock_table_keeper.go
```diff
@@ -186,7 +186,13 @@ func (k *lockTableKeeper) doKeepRemoteLock(
 					maybeHandleRemoteBindChanged(binds[idx])
 				}
 			} else if resp.NewBind != nil {
-				k.service.handleBindChanged(*resp.NewBind)
+				if binds[idx].AllocatorID != "" &&
+					resp.NewBind.AllocatorID != "" &&
+					binds[idx].AllocatorID != resp.NewBind.AllocatorID {
+					maybeHandleRemoteBindChanged(binds[idx])
+				} else {
+					k.service.handleBindChanged(*resp.NewBind)
```
XuPeng-SH: The keeper has the same stale-response window: `binds[idx]` and a delayed response can both carry A0, so the equality check takes this direct publication path even after another keepalive/GetBind has observed A1 and purged A0. Please validate against the current allocator snapshot or refresh every exceptional `NewBind` through the allocator.

XuPeng-SH (changes requested):
Request changes: late `NewBind` responses can republish a bind from a superseded allocator after the new allocator has already purged it.

Reproduced locally with this ordering: cache remote bind from A0 → observe A1 (version may regress) and purge the table → deliver an in-flight `NewBind` from A0 whose allocator ID matches the request bind. The table is recreated with `AllocatorID=A0`, violating the core fence invariant and allowing new transactions to use a bind the current TN allocator no longer recognizes. The same bypass exists in the keeper response path.

Please route these asynchronous `NewBind` publications through allocator-state validation (or refresh from the current allocator), and add race regressions for both paths.

gouhongshen (approved):
## Codex automated review

No blocking correctness, concurrency, compatibility, or resource-handling regressions found. Verified the allocator-epoch fencing paths and ran `go test ./pkg/lockservice -count=1`.

aunjgr (changes requested):
The current 4.1-dev backport is not reviewable or mergeable in this shape: it is conflicting and contains roughly 1,315 files with about 154k additions, far beyond the intended lockservice allocator-epoch fix. Please recreate or rebase it from current 4.1-dev and cherry-pick only the intended lockservice commits, then request review on the clean diff.

aunjgr (changes requested):
The current 4.1-dev backport is not reviewable or mergeable in this shape: it is conflicting and contains roughly 1,315 files with about 154k additions, far beyond the intended lockservice allocator-epoch fix. Please recreate or rebase it from current 4.1-dev and cherry-pick only the intended lockservice commits, then request review on the clean diff.

aunjgr (approved):
Reviewed head efa61615c. This is a clean 4.1-dev backport of the already-merged allocator-epoch fencing and late-bind rejection fixes. The ownership, cancellation, refresh-failure, and exact-bind invalidation paths retain the upstream behavior. No remaining blockers found.

gouhongshen (changes requested):
## Codex automated review

The allocator fencing works for fully upgraded peers, and both the full lockservice suite and focused race tests pass. However, the original stale-bind-until-COMMIT failure remains reachable during a mixed-version rolling upgrade.

### P1 - Fence legacy binds when allocator identity first appears (pkg/lockservice/service.go:795)
A new CN can cache binds from an old TN whose responses contain no allocator metadata, leaving `lastAllocatorID/Version` and each bind's `AllocatorID` empty. If that TN is then upgraded/restarted with a regressed or equal epoch, this condition sets `allocatorChanged=false` because no previous allocator state was recorded. The purge consequently falls back to `bind.Version < observed.version`, retains the legacy binds, and then records the new allocator ID; subsequent keepalives hit the same-allocator fast path and never reconsider them. If another table re-registers the CN first, keepalive is `OK=true`, reproducing the original behavior where the stale bind survives until commit validation fails. Treat the first non-empty allocator ID as an identity boundary for already-cached empty-ID binds, and cover old-TN → new-TN transition with a regressed epoch.

iamlinjunhong:
> ## Codex automated review
> The allocator fencing works for fully upgraded peers, and both the full lockservice suite and focused race tests pass. However, the original stale-bind-until-COMMIT failure remains reachable during a mixed-version rolling upgrade.
> 
> ### P1 - Fence legacy binds when allocator identity first appears (pkg/lockservice/service.go:795)
> A new CN can cache binds from an old TN whose responses contain no allocator metadata, leaving `lastAllocatorID/Version` and each bind's `AllocatorID` empty. If that TN is then upgraded/restarted with a regressed or equal epoch, this condition sets `allocatorChanged=false` because no previous allocator state was recorded. The purge consequently falls back to `bind.Version < observed.version`, retains the legacy binds, and then records the new allocator ID; subsequent keepalives hit the same-allocator fast path and never reconsider them. If another table re-registers the CN first, keepalive is `OK=true`, reproducing the original behavior where the stale bind survives until commit validation fails. Treat the first non-empty allocator ID as an identity boundary for already-cached empty-ID binds, and cover old-TN → new-TN transition with a regressed epoch.

fixed

gouhongshen (changes requested):
## Codex automated review

One P1 mixed-version fencing gap remains.

### P1 - Observe allocator identity before filtering an otherwise equal legacy bind (pkg/lockservice/lock_table_keeper.go:236)
A successful GetBind response is passed to allocator fencing only when newBind.Changed(bind). Changed intentionally treats an empty AllocatorID as compatible with a non-empty one, so an old-TN bind and an upgraded allocator's equal-version, same-owner bind compare unchanged. The new allocator identity is then discarded: legacy tables are not purged and active transactions are not fenced. Because allocator validation uses the same compatibility comparison, such a transaction can also pass commit validation before the next keepalive. The identical gate exists in remoteLockTable refresh paths. Validate allocator state on every successful GetBind before using structural equality to decide publication, and test equal-epoch old-TN→new-TN refreshes with an active transaction.

gouhongshen (approved):
## Codex automated review

No actionable regressions found at verified HEAD 15bf3091094341401e22be4668af6fc0a8d38136. Prior review blockers are resolved: stale responses are fenced before publication, and first allocator identity observation purges legacy empty-ID binds. Protocol changes are additive and rolling compatibility is handled explicitly. Concurrency, cleanup, unhappy paths, and focused coverage are coherent. `go test ./pkg/lockservice -count=1` passed in 584.972s; current CI checks are green and `git diff --check` is clean.

PR #25820, opened 2026-07-16
fix(pipeline): make remote receiver readiness lifecycle-driven

### LINKED ISSUE #25816: [Bug] 30s remote dispatch watchdog can falsely fail slow receiver registration
## Summary

Remote dispatch can falsely fail a valid query when receiver registration takes longer than 30 seconds.

The regression was introduced by commit [`36016d54a8a5ea4d1bda127e11a402c828538bd7`](https://github.com/matrixorigin/matrixone/commit/36016d54a8a5ea4d1bda127e11a402c828538bd7) / #24269. That change replaced the previous 3600-second wait with fixed 30-second watchdogs and introduced the current error:

```text
```

The timeout detects slow registration as if the remote CN had failed. This is timing-sensitive and therefore appears intermittently under slow external scans, JOIN preparation, high concurrency, GC, or storage/network pressure.

This issue tracks the timeout/readiness design regression separately from workload-specific symptom issue #25211.

## Regression evidence

A history search for both the error string and timeout constant points to the same first main commit:

```text
36016d54a8a5ea4d1bda127e11a402c828538bd7
fix(pipeline): replace busy-wait polling with channel-based notification (#24269)
```

Before #24269:

```go
HandleNotifyTimeout = 3600 * time.Second
```

After #24269:

```go
const waitRegistrationTimeout = 30 * time.Second
const waitRemoteRegTimeout = 30 * time.Second
```

It also introduced:

```go
"dispatch process not registered within %s, remote CN may have failed"
```

The PR stated that registration normally happens within seconds. That assumption does not hold for all execution topologies: remote/pre scopes can begin notify processing before the coordinator/local dispatch reaches `Dispatch.Prepare` and publishes its receiver UUID. Slow Hive external-scan/JOIN preparation can expose this ordering window for longer than 30 seconds.

The later fixes also corroborate incomplete registration coverage:

- #25238: early registration for decoded remote scopes
- #25574: early/race-safe registration for coordinator/local dispatch receivers
- #25664: registration below nested remote scopes
- #25682: per-CN `colexec.Server` isolation for standalone Multi-CN
- #25688: wait for receiver attachment on empty dispatch

These fixes cover specific paths, but the fixed watchdog still conflates receiver readiness with remote CN failure.

## Intermittent pass/fail evidence

The same Hive Partition job alternates between success and failure, so a one-shot bisect is not reliable:

| Run | Hive job result |
|---|---|
| [28061543580](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28061543580), job [83097381717](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28061543580/job/83097381717) | success |
| [28133872754](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28133872754), job [83335832278](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28133872754/job/83335832278) | failed with dispatch registration timeout |
| [28721408964](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28721408964), job [85236481992](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28721408964/job/85236481992) | success |
| [28756800333](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28756800333), job [85273758086](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28756800333/job/85273758086) | failure |
| [28902823580](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28902823580), job [85780232473](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28902823580/job/85780232473) | success |
| [29054645722](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29054645722), job [86253126494](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29054645722/job/86253126494) | failed with receiver lifecycle/registration errors |
| [29170188051](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29170188051), job [86594634090](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29170188051/job/86594634090) | success |
| [29211194235](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29211194235), job [86704634878](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29211194235/job/86704634878) | failed with dispatch registration timeout |

## Latest main occurrence

The latest affected path is documented in #25211:

- Run: [29428464735](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29428464735)
- Job: [87416236626](https://github.com/matrixorigin/mo-nightly-regression/actions/runs/29428464735/job/87416236626)
- MatrixOne commit: [`0d8111aef18ed25aa4eb35113f8bea9e6ab66587`](https://github.com/matrixorigin/matrixone/commit/0d8111aef18ed25aa4eb35113f8bea9e6ab66587)

That commit already contains #25682. The same Hive fact/dimension JOIN stopped making progress and hit the client-side 1800-second timeout.

The latest runner log does not expose the inner 30-second dispatch error, so it proves recurrence on the same SQL path but does **not** by itself prove the latest hang has the identical internal cause. CN logs and a query profile are still required for that run.

Also, #25682 primarily fixes multiple CN services sharing one process in standalone Multi-CN. The TKE reproduction uses separate CN pods/processes, so per-service server isolation does not remove the TKE registration-ordering problem.

## Affected SQL

```sql
  on f.part_id = d.part_id
group by d.part_name
order by d.part_name;
```

## Expected behavior

A valid slow query must not be classified as a remote CN failure merely because receiver registration has not happened within a fixed 30-second wall-clock window.

Remote dispatch should either:

1. establish an explicit receiver-readiness/registration handshake before remote notify begins; or
2. start a bounded failure timeout only after the relevant scope is expected to be registration-ready.

The original/root error should be preserved when registration does fail.

## Suggested fix direction

1. Define an explicit lifecycle for receiver creation, registration, attachment, cancellation, completion, and retry.
2. Publish all coordinator/local receivers before starting any scope goroutine that can notify them.
3. Do not use a fixed pre-readiness timeout as proof that a remote CN failed.
4. Add receiver UUID, query/statement ID, source/target CN service, scope topology, and registration/attachment timestamps to logs and metrics.
5. Add deterministic tests with an injected delay longer than 30 seconds before `Dispatch.Prepare`.
6. Run the affected Hive JOIN repeatedly per commit; classify a commit only after multiple attempts because the failure is flaky.
7. On client timeout, cancel the server query and recreate the connection before executing the next statement.

## Related issues and PRs

- #25211
- #24930
- #24269
- #25238
- #25475
- #25574
- #25664
- #25682
- #25688

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

Fixes #25816

Remote dispatch readiness is a lifecycle event, not a wall-clock duration. The fixed 30-second watchdog can classify a valid slow registration as a failed CN, while simply extending the timeout leaves the same correctness bug and delays real cancellation.

This PR closes the receiver lifecycle and its resource consequences as one system:

- replaces fixed pre-readiness watchdogs with query/connection cancellation-driven waits;
- models receiver state explicitly as ready, attached, closed, or terminal, with generation-scoped waiters and per-UUID wakeups;
- makes publication, attachment, cancellation, completion, duplicate notify, and owner cleanup race-safe;
- cancels the parent pipeline on early local/remote errors before waiting for sibling scopes, preventing error-path wait cycles;
- keeps new-client/old-server compatibility through a query-bounded, jittered legacy retry only for typed `NotRegistered` responses;
- records bounded receiver-wait outcomes and legacy retry counts without high-cardinality labels;
- routes large (>=128 KiB) mpool buffers through direct mmap/munmap while retaining libc for small allocations, so short-lived multi-CN fan-out buffers return RSS deterministically instead of accumulating in libc arenas.

The allocator change follows a causal 2-CN control test: with identical 2380 MiB CN limits, the affected baseline completed the first 400-query wave and was OOM-killed early in the second wave, while the implementation in this PR completed two full 400-query waves without OOM. Connection counts had already plateaued and live Go heap grew only slightly, isolating retained off-heap arena memory rather than an active connection leak.

### Compatibility and failure semantics

- cancellation before registration does not poison a future registration for the same UUID;
- cancellation after attachment cancels the owning process;
- a terminal owner cannot be resurrected by a late publish/attach race;
- legacy retry is never used for unrelated errors and is bounded by the query context;
- no polling or hot-path sleep is reintroduced, preserving the event-driven backend creation behavior from #25771;
- FIN/FIN_ACK connection reuse and response sequence semantics from recent MORPC fixes remain unchanged.

### Validation completed

Package-level validation:

- `mo-cgo-test -count=1` on `pkg/sql/colexec`, `pkg/sql/colexec/dispatch`, `pkg/sql/compile`, `pkg/util/metric/v2`, `pkg/common/morpc`, `pkg/common/malloc`, and `pkg/common/mpool`;
- 30 repeated race runs for the new receiver lifecycle/cancellation/retry tests;
- 10 repeated race runs for allocator threshold and realloc transitions;
- full `pkg/common/morpc` race run;
- `go vet` on all changed/adjacent packages;
- allocator matrix covers below/at/above threshold, small-to-small, small-to-large, large-to-large, large-to-small, logical length smaller than capacity, zero size, mismatch rejection, and concurrent transitions.

Fresh Linux image and resource-constrained 2-CN validation, with no `MALLOC_*` environment override and a 2380 MiB cgroup limit per CN:

- cold `make dev-build` completed successfully;
- wave 1: 400/400 queries, 0 failures, 21.289 s wall time, 0.756 s median, 2.481 s max;
- wave 2: 400/400 queries, 0 failures, 18.180 s wall time, 0.695 s median, 1.334 s max;
- peak cgroup memory: CN1 2,206,769,152 bytes, CN2 1,221,582,848 bytes;
- both CNs reported `memory.events`: `max=0`, `oom=0`, `oom_kill=0`;
- receiver lifecycle reached zero active entries after completion, terminal receiver errors remained zero, and `TIME_WAIT` remained one per CN.

Cancellation unhappy path:

- 16 clients deliberately hit a 1-second client read timeout;
- an immediate recovery query returned the correct result in 0.136 s;
- the temporary connection/lifecycle increase returned to the pre-test baseline within 10 seconds and remained stable through 50 seconds;
- no stuck receiver owner, connection leak, or OOM event was observed.

Mixed-version compatibility (new CN plus affected old CN):

- 160/160 queries succeeded, with 0 failures;
- the new CN recorded 601 typed legacy retries, proving that the compatibility path was exercised rather than bypassed;
- no protocol sequence, unsupported-method, or OOM errors were observed.


aptend (changes requested):
Blocking allocator ownership issue: SimpleCAllocator.ReallocZero chooses mmap vs libc provenance from cap(old), but callers can legally pass a reduced-cap view of an existing mpool allocation. A concrete existing path is: allocate 128 KiB off-heap (mmap), PtrLen.Replace shrinks only its logical length, PtrLen.ToByteSlice reconstructs len/cap from that shorter length, then AppendRawBytes calls MPool.ReallocZero. sane_allocator.go then sees cap(old) below 128 KiB and calls C.realloc on a unix.Mmap pointer, which is an invalid-pointer abort/undefined behavior. A reduced-cap mmap slice that still stays above the threshold instead reaches unix.Munmap with the wrong mapping length and panics with EINVAL. The current logical-length test preserves the original cap, so it misses this case. Please derive allocation size/provenance from stable mpool ownership metadata (not slice capacity) and add a reduced-cap regression covering threshold transitions.

XuPeng-SH:
Addressed the requested allocator ownership fix in `50c25fdd7b`.

The review finding was valid: a slice's `cap` is a mutable view property and cannot identify the backing allocator or its actual size. The fix now uses the stable `MPool` allocation header (`allocSz`, `offHeap`, and `poolId`) as the source of truth whenever growth exceeds the caller-visible capacity. `SimpleCAllocator.ReallocZero` receives the recorded old allocation size explicitly, so mmap/libc selection, unmap length, accounting, and copy/free ownership no longer depend on a reduced-capacity view.

The common capacity-local resize path remains lock-free. The metadata path also handles the related closure cases rather than special-casing `PtrLen`:

- reduced-cap and zero-cap views;
- libc -> mmap, mmap -> libc, and mmap -> mmap transitions;
- on-heap -> off-heap and off-heap -> on-heap transitions;
- growth within the recorded backing allocation without an unnecessary realloc;
- growth beyond the recorded allocation with exactly-once old ownership release;
- allocation failure preserving the old allocation/header.

Added regressions include the concrete `PtrLen.Replace` shrink followed by `AppendRawBytes`, reduced-cap transitions across the mmap threshold, zero-cap tracked views, and source-provenance transitions.

The branch was then rebased onto latest `main` (`757cacf0a5`). `git range-diff` reports both commits semantically unchanged by the rebase.

Fresh post-rebase validation:

- all affected packages passed: colexec, dispatch, compile, metrics, MORPC, malloc, and mpool;
- allocator/provenance focused race stress: `-race -count=10`;
- receiver lifecycle/cancellation focused race stress: `-race -count=30`;
- full MORPC race suite passed;
- full race suites for all other owning packages passed;
- build, vet, and `git diff --check` passed.

XuPeng-SH:
> Blocking allocator ownership issue: SimpleCAllocator.ReallocZero chooses mmap vs libc provenance from cap(old), but callers can legally pass a reduced-cap view of an existing mpool allocation. A concrete existing path is: allocate 128 KiB off-heap (mmap), PtrLen.Replace shrinks only its logical length, PtrLen.ToByteSlice reconstructs len/cap from that shorter length, then AppendRawBytes calls MPool.ReallocZero. sane_allocator.go then sees cap(old) below 128 KiB and calls C.realloc on a unix.Mmap pointer, which is an invalid-pointer abort/undefined behavior. A reduced-cap mmap slice that still stays above the threshold instead reaches unix.Munmap with the wrong mapping length and panics with EINVAL. The current logical-length test preserves the original cap, so it misses this case. Please derive allocation size/provenance from stable mpool ownership metadata (not slice capacity) and add a reduced-cap regression covering threshold transitions.

fixed

XuPeng-SH:
Added the liveness closure in `b9710c5436` and rebased the branch onto latest `main` (`734fa4e50b`).

The 120s limit is admission-only: it bounds the period in which a remote dispatch receiver has not registered/attached. It does not become the stream context and does not limit an attached HTAP query, so queries running longer than 120s remain valid.

The fix closes all registration wait chains:

- local dispatch waiting for all remote receivers;
- new server waiting for a receiver UUID to be published;
- new client retrying an old server that returns typed `RemoteDispatchNotRegistered`.

Boundary and performance details:

- timeout races perform a final registry lookup while the waiter reference is still held, so a receiver published or closed at the boundary wins with its real terminal state;
- timeout/cancel paths release generation-scoped waiter references and do not poison later registration of the same UUID;
- the compatibility deadline is captured from notify start but its timer is created lazily only after an old peer returns `NotRegistered`; the normal new-to-new path gets no extra client timer;
- only startup/admission paths allocate timers; there is no per-batch hot-path cost;
- the wait histogram now includes a `timeout` outcome and a finite bucket range beyond 120s.

Added coverage for no registration, partial registration, timeout cleanup followed by later registration, legacy retry exhaustion, and an already-expired admission budget after successful attach (proving the successful stream still uses the query context).

Fresh post-rebase validation:

- affected packages passed: colexec, dispatch, compile, metrics, MORPC, malloc, and mpool;
- full race suites passed for colexec/dispatch/compile/metrics;
- full race suites also passed for MORPC/malloc/mpool before the final lazy-timer-only refinement;
- focused timeout/compatibility tests passed with `-race -count=10` after the refinement;
- `go build`, `go vet`, `git diff --check`, and clean-worktree checks passed;
- `git range-diff` shows the two prior commits are semantically unchanged by the rebase.

aptend (approved):
Re-reviewed the current head. The allocator provenance blocker is fixed using stable MPool allocation metadata, and the receiver admission lifecycle now has bounded cancellation/timeout closure without limiting attached query streams. Generation-scoped waiter cleanup, timeout-boundary lookup, legacy typed-error retry, and sibling fail-fast cancellation all look sound. Focused race tests and the affected package suites pass. LGTM.

aunjgr (changes requested):
Requesting changes for one P1 correctness blocker. The receiver state/generation model and allocator provenance fix look sound, including traversal of nested and subquery-produced scope graphs. However, pkg/sql/colexec/types.go still defines a shared 120-second admission cutoff. Its three consumers—local dispatch waiting for attachment, the new server waiting for UUID publication, and new-client/old-server retry—will cancel a valid query if registration occurs at 121 seconds while the query, stream, connection, and CN remain alive. Mixed-version and subquery-produced remote branches inherit this failure. This directly contradicts the PR description and #25816: extending the old 30-second watchdog leaves the same wall-clock correctness bug. Please make these waits terminate on lifecycle evidence; any watchdog should be diagnostic-only. Add a positive test where the former timeout elapses, all lifecycle contexts remain live, publication then occurs, and attachment succeeds.

XuPeng-SH:
Addressed the P1 blocker in `594186b252` after merging the latest `main` (`c1843634bf`). The review finding was valid.

The 120-second change was the wrong abstraction: elapsed wall-clock time is not evidence that a receiver can no longer be published. It merely moved the false-negative boundary from 30 seconds to 120 seconds. The corrected termination model now uses only lifecycle evidence:

- local dispatch waits for attachment, query cancellation, or terminal registration-channel closure;
- the new server waits for UUID publication/terminal state, message cancellation, or connection closure;
- new-client/old-server compatibility retry uses the query context directly, with no derived admission deadline.

The shared cutoff, all three timeout branches, the timeout error, and the timeout metric outcome were removed. This also removes admission timer/deadline allocations; retry backoff remains capped and jittered. Generation-scoped waiter ownership and terminal-state retention are unchanged.

Regression coverage now includes delayed publication while every lifecycle context remains live followed by successful attachment, delayed legacy retry success with proof that the exact query context has no derived deadline, partial attachment followed by query cancellation, message/connection cancellation, terminal registry states, generation reuse, and duplicate attachment.

Fresh validation on the updated branch:

- focused lifecycle tests passed;
- focused concurrency/lifecycle stress passed with `-race -count=10`;
- full tests and full race suites passed for `malloc`, `mpool`, `colexec`, `dispatch`, `compile`, and pipeline metrics;
- affected-package `go build`, `go vet`, and `git diff --check` passed.

The process correction is to separate correctness from diagnostics: a watchdog may report a long wait, but it must never manufacture a terminal state. In this path no independent watchdog was needed, so the semantic timer was deleted rather than renamed or widened.

XuPeng-SH (commented):
Withdrawn after tracing the complete distributed cancellation and ownership graph.

The local interleaving described in the original review is reproducible, but it does not establish a production hang. A notify-stream error returns to sendNotifyMessage.closeWithError, which cancels the merge parent before waiting for siblings. That cancellation reaches the remote pipeline sender; its StopSending or poisoned-backend close reaches the RecordBuiltPipeline cleanup owner and cancels the dispatch process. A physical session failure also runs session cleanup for recorded pipelines, while deadline expiry is inherited by the remotely built process through its own pipeline message context.

More importantly, a missing waiter has not acquired dispatch-process ownership. Multiple waiters for one UUID are explicitly supported and share a generation. Making a canceled pre-attach waiter consume and cancel a newly published ready receiver would let a stale or duplicate notify attempt kill the valid query before another waiter attaches. The ready-to-attached transition is the correct ownership transfer point; cancellation after that point already cancels the consumed process.

Therefore this is not a P1 blocker and the proposed change should not be made. The original isolated reproduction omitted the upstream and built-pipeline cleanup owners.

Issue #12212, opened 2023-10-19
[Tech Request]: Reimplement hash MARK join


aunjgr:
### Is there an existing issue for the same tech request?

- [X] I have checked the existing issues.

### Does this tech request not affect user experience?

- [X] This tech request doesn't affect user experience.

### What would you like to be added ?

```Markdown
The former hash MARK join implementation is broken and abandoned. All MARK join plans are executed by nested-loop join.

We still need a working hash MARK join.
```

PR #25822, opened 2026-07-16
fix(sql): reimplement hash MARK join for IN subqueries

### LINKED ISSUE #25821: [Bug]: IN/NOT IN MARK joins fall back to nested loops and large expressions are left-deep

### Branch Name

main

### Commit ID

757cacf0a552025d928118da8ce52a524f99d2e9

- Hardware parameters: N/A
- OS type: Linux
- Others: MatrixOne main; reproduced through compiler/operator unit tests.

### Actual Behavior

Scalar and correlated-subquery `IN` / `NOT IN` cases that require a MARK join are planned as a nested-loop join because the former hash MARK implementation is unavailable. This makes projected `IN` and nullable `NOT IN` evaluate the build side row-by-row, with O(P × B) probe/build work.

In addition, large tuple/mixed-type `IN` lists and row comparisons are lowered into left-deep boolean trees. Their depth grows linearly with the number of values or columns, unnecessarily increasing binder/optimizer recursion and stack growth.

The older technical request #12212 records the missing hash MARK implementation but does not cover the null-aware execution contract, distributed pipeline propagation, or the large-expression-tree failure mode.

### Expected Behavior

For a safe equality MARK join, MatrixOne should use a hash join while preserving SQL three-valued semantics:

- a matching row returns TRUE;
- a non-matching row returns FALSE only when the build side has no NULL key;
- a non-matching or NULL probe key returns UNKNOWN when a build NULL makes the comparison unknown;
- an empty build returns FALSE, including for a NULL probe key.

Unsafe composite nullable keys and non-equality predicates must retain the row-aware nested-loop implementation.

Large `IN` expansions and row comparisons should preserve order and semantics while using a logarithmic-depth boolean tree.

### Steps to Reproduce

1. On the listed main commit, create a plan for a projected `IN` subquery or a nullable `NOT IN` subquery. It requires MARK semantics but is instantiated as `LoopJoin`.
2. Use a build side containing both a matching value and NULL, then a non-matching value and NULL, then an empty build. The MARK result must distinguish TRUE, UNKNOWN, and FALSE as above.
3. Bind a tuple `IN` list or row comparison with thousands of members/columns. Inspect the generated boolean expression: it is left-deep with linear depth.

### Additional information

This issue is fixed by a hash MARK implementation that tracks whether the build keys contain NULL, carries that fact through distributed pipeline instructions, emits one nullable boolean marker per probe row, and uses balanced boolean-tree construction for large IN/row-comparison expansions.

Related: #12212.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25821

Related: #12212

Reimplements the safe subset of hash MARK joins for `IN` / `NOT IN` subqueries and removes linear-depth expression construction for large `IN` and row-comparison expansions.

### Correctness

- Uses hash MARK only for a single equality key, or for composite equality keys whose arguments are statically NOT NULL.
- Preserves SQL three-valued semantics: match → TRUE; no match with a build NULL or a NULL probe → UNKNOWN; empty build → FALSE, including for a NULL probe.
- Retains LoopJoin for non-equality predicates, same-side equality, mixed-side correlated expressions, and nullable composite keys, where hashing cannot preserve the full row-aware predicate.
- Carries the build-side NULL-key fact through HashBuild, JoinMap, and distributed pipeline serialization.
- Ensures MARK joins emit every probe row even when the build side is empty, and clears null-key state on operator reuse.
- Rejects any residual condition on a hash MARK operator during `Prepare`, preventing unsupported plans from silently dropping predicates.

### Scalability

- Replaces left-deep AND/OR construction for tuple/mixed-type `IN` and tuple subquery comparisons with order-preserving balanced trees.
- The number of predicate nodes remains linear, while expression depth becomes O(log N), avoiding recursive stack amplification for large valid inputs.
- This PR does not claim a percentage benchmark. The hash MARK path removes nested-loop O(P × B) work for the safe MARK-join subset; ordinary filter-only `IN` already uses SEMI join planning.

### Validation

- Hash MARK truth table: matching value with build NULL, non-match with/without build NULL, all-NULL build, NULL probe, and empty build.
- Batch boundary and operator Reset/reuse state; tests assert zero remaining mpool bytes.
- Planner selection/fallback matrix, including a hashable IN key plus mixed-side correlated equality and ordinary residual predicates, and distributed pipeline `TrackNullKeys` round trip.
- Tuple/mixed-type IN and row-comparison trees at up to `TableColumnCountLimit` (4096); depth is asserted at most 14.
- Race test for hash MARK join.

Commands run after the final semantic edit:

```text
.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s \
  ./pkg/pb/pipeline ./pkg/vm/message ./pkg/sql/colexec/hashbuild \
  ./pkg/sql/colexec/hashjoin ./pkg/sql/plan ./pkg/sql/compile

.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=300s \
  ./pkg/sql/colexec/hashjoin -run 'TestHashMarkJoin'
```

Both commands passed. `go build` and `go vet` for the modified plan/compile packages also passed, and `git diff --check` is clean.


XuPeng-SH:
**Review: request changes**

I still see one correctness blocker before this can be approved.

`canUseHashMarkJoin` admits MARK joins by checking `exprRelPos` on each equality term. That accepts predicates where one equality argument mixes probe/build columns, for example `t2.z + t1.c = t1.d`. `constructHashJoin` later rejects the same predicate as a hash key via `HasColExpr` and moves it into `NonEqCond`, but the new `probeMark` fast path in `pkg/sql/colexec/hashjoin/join.go` never evaluates `NonEqCond` once MARK uses the hash path.

That means a query like `... where t1.a in (select t2.x from t2 where t2.y = t1.b and t2.z + t1.c = t1.d)` can now return TRUE based only on `t2.x = t1.a`, silently ignoring the residual correlated equality.

Please derive hash-MARK eligibility from the same criterion used by `extraJoinConditions`/`HasColExpr` (or explicitly require `NonEqCond == nil`) and add a regression test for a hashable IN key plus a mixed-side correlated equality.

XuPeng-SH:
Fixed in 1f2c74c98.

The reported correctness case now falls back to LoopJoin:

- `canUseHashMarkJoin` first requires the same condition split used by `constructHashJoin` to produce no residual condition.
- It additionally walks every equality operand and requires all referenced columns to belong to exactly one relation. This extra check is necessary because the current recursive `HasColExpr` can return a representative relation for nested mixed-side expressions instead of rejecting them.
- `HashJoin.Prepare` now fails fast if a MARK operator contains `NonEqCond`, so an unsupported operator cannot silently skip a residual predicate.

Regression coverage includes one hashable IN key plus `t2.z + t1.c = t1.d`, one hashable key plus an ordinary non-equality residual, and the runtime fail-fast invariant.

Post-rebase validation passed:

```text
mo-cgo-test ./pkg/pb/pipeline ./pkg/vm/message ./pkg/sql/colexec/hashbuild ./pkg/sql/colexec/hashjoin ./pkg/sql/plan ./pkg/sql/compile
mo-cgo-test -race ./pkg/sql/colexec/hashjoin -run 'TestHashMarkJoin'
go build ./pkg/sql/colexec/hashjoin ./pkg/sql/compile
go vet ./pkg/sql/colexec/hashjoin ./pkg/sql/compile
```

XuPeng-SH:
**Review: request changes**

I found two correctness blockers in the state/ownership boundaries of the new hash MARK path.

1. **Spilled MARK joins lose probe rows and cannot preserve global three-valued semantics.** In `pkg/sql/colexec/hashjoin/join.go`, `SpillEngineConfig.NeedsProbeForEmptyBuild` is still derived only from `EmitUnmatchedProbe()`. MARK returns false there, so `ScatterProbeTable` disables writers for build-empty buckets and those probe rows are discarded. A review-only reproducer with a spilled JoinMap containing 32 empty build buckets and probe values `[1,2]` expected two FALSE marks; `Exec` returned a nil batch.

   Adding `IsMark()` to that flag alone is insufficient. `emptyProbe` currently treats a locally empty spill bucket as a globally empty build and emits constant FALSE. Those are different states: with a globally non-empty build, a NULL probe is UNKNOWN; if any build key is NULL, every non-matching probe is UNKNOWN even when that NULL lives in another bucket. The initial spilled JoinMap already carries global row count, but hashjoin drops it before bucket probing, and the spill build path skips `BuildHashmap`, so `HasNullKey` is never discovered/published there.

   Please carry the global build facts (`has rows`, `has NULL key`) across the initial spill handoff, retain probe files for MARK empty buckets, and make the bucket-empty path evaluate the mark from those global facts plus probe-key nullness. Cover global-empty, locally-empty/globally-nonempty, NULL-in-another-bucket, match-with-global-NULL, and recursive re-spill.

2. **Operator duplication drops `TrackNullKeys`.** The `vm.HashBuild` branch in `pkg/sql/compile/operator.go:134-160` copies the existing HashBuild configuration but not the new field. A duplicated/parallelized MARK build therefore silently stops tracking NULL keys, while the remote serialization round-trip test still passes. Please copy the field and add a `dupOperator` regression test. More generally, the test matrix needs every state-transfer boundary: construct, duplicate, remote encode/decode, reset/reuse, normal build, and spill/re-spill.

The non-spill focused implementation and balanced-expression changes otherwise look sound, and the affected package tests plus focused race runs pass. These two paths can still return wrong SQL results, so I cannot approve the PR yet.

XuPeng-SH:
Resolved systematically in a3cf2f4bb8 (rebased on origin/main 824090c2e6).

The fix is based on the MARK truth-table invariant, not bucket-local special cases:

- global empty build => FALSE, including a NULL probe
- exact match => TRUE
- otherwise a probe/build NULL => UNKNOWN
- otherwise => FALSE

Implementation closure:

- preserves global build row count and global NULL-key state across the initial spilled JoinMap handoff
- retains MARK probe rows for empty build buckets in both initial spill and recursive re-spill
- distinguishes a locally empty bucket from a globally empty build and derives the marker from global facts plus probe-key nullness
- discovers NULL keys in both normal and spill builds using batch-level vector metadata, then publishes the fact on spilled JoinMaps
- copies TrackNullKeys through operator duplication and both broadcast/shuffle construction; the existing protobuf remote round trip remains covered
- resets all new per-execution facts and fails fast for unsupported MARK operator contracts, including residual predicates, mismatched keys, nullable composite keys, and build-side result columns
- uses bitmap union for empty-bucket probe NULLs; no per-row NULL scan or new FD/wait ownership was introduced

Coverage added/verified:

- normal single-key TRUE/FALSE/UNKNOWN truth table
- globally empty build, locally empty/globally nonempty bucket, global NULL in another bucket, and exact match with global NULL
- real forced spill and forced recursive re-spill
- composite NOT NULL keys and nullable-composite rejection/fallback contract
- reset/reuse without stale row/NULL state
- duplicate, broadcast/shuffle construction, remote serialization, and batch boundary
- lost/duplicated probe-row checks and mpool cleanup

Validation passed:

- deterministic CGo tests for pipeline, message, hashbuild, hashjoin, plan, and compile packages
- focused tests repeated 20 times
- focused race tests
- go build, go vet, and git diff --check

The existing mysql_compat_null_subquery BVT already supplies end-to-end SQL coverage for NULL/non-NULL sets, empty sets, IN, and NOT IN; CI will rerun it on the new head.

LeftHandCold (approved):
Reviewed the latest head a3cf2f4bb8 with a correctness/regression focus. I checked hash-MARK eligibility and LoopJoin fallback for residual, correlated, mixed-side, and nullable composite predicates; SQL three-valued behavior for matches, NULL probes/build keys, and empty builds; state transfer across duplicate/remote/reset; spill global-vs-local bucket facts; and impact on existing non-MARK join paths. I found no blocking issue. Targeted tests for hashbuild, hashjoin, compile, and plan passed locally, as did the focused TestHashMarkJoin race run.

aptend (approved):
Reviewed the current head (`cd7bdfb9`) across planner selection, MARK three-valued semantics, normal/spill execution, operator duplication, remote serialization, and Reset/reuse lifecycle.

No merge-blocking findings. The affected package suite and focused hash MARK race tests pass locally; required CI checks are green.

LeftHandCold (approved):
Re-reviewed the latest head (ada8a8083895983820667679b1fd6439df0350da). The MARK join planner eligibility, three-valued semantics, spill/recursive-spill paths, state transfer, and reset handling are consistent. Targeted package tests, race test, build, and vet passed locally.

XuPeng-SH:
**Self-review conclusion: APPROVE (code) — exact head `ada8a8083895983820667679b1fd6439df0350da`.**

GitHub does not allow the PR author to submit a formal approval review, so I am recording the exact-head decision here. Independent approval for this head is already present.

I re-reviewed the complete diff from first principles across planner admission/fallback, SQL three-valued MARK semantics, normal build, forced spill and recursive re-spill, reset/reuse, duplicate and remote state transfer, FD ownership, mixed-version CN scheduling, and expression-depth/complexity bounds. I found no merge-blocking issue.

Local validation on this head passed:

- affected packages: `pipeline`, `message`, `hashbuild`, `hashjoin`, `plan`, `compile`
- focused `TestHashMarkJoin*` race run, 3 repetitions
- planner/compile admission, fallback, duplication, and balanced-tree tests, 20 repetitions
- `go vet`, `make err-check`, and `git diff --check`

The first local run exposed only a temporary-worktree CGo rpath issue; rerunning with deterministic absolute rpaths passed. Required CI jobs are still running and remain the delivery gate, separate from the code-review conclusion.

PR #25823, opened 2026-07-16
perf(lockservice): shard active transaction holder

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25472

### Summary

- Splits the active transaction registry into 16 fixed FNV-1a shards, each protected by its own RWMutex.
- Keeps remote-service liveness metadata under the existing mutex and preserves current main's remote-liveness, unknown-commit, and cancellation behavior.
- Stores the transaction and its remote service in one shard entry, removing the second centralized map.
- Uses a conservative atomic active count so a zero count always means every shard is empty during service drain.
- Scans one shard at a time for timeout, bind-fencing, reference, and diagnostic paths; service close locks shards in a fixed order.

### Why

TPCC mutex profiles showed the former holder-wide mutex serializing transaction creation and deletion on the Lock/Unlock path. Sharding removes that unrelated cross-transaction serialization while retaining per-transaction locking.

### Compatibility and non-goals

This changes only in-memory lockservice synchronization. It does not change SQL, wire or persisted formats, configuration defaults, remote transaction liveness decisions, or unknown-commit fencing.

### Validation

- `go vet ./pkg/lockservice`
- `go test -count=1 -timeout=5m ./pkg/lockservice`
- `go test -race -count=1 -timeout=10m ./pkg/lockservice`
- Relevant create/delete, remote, timeout, and validity race tests repeated 10 times
- Combined P0-1/P0-3 branch: full test and race for lockservice, vector, sql/util, and disttae

Current-main local benchmark, 10 samples at each concurrency:

| GOMAXPROCS | main median | candidate median | change |
|---:|---:|---:|---:|
| 8 | 423.8 ns/op | 181.6 ns/op | -57.1% |
| 16 | 440.6 ns/op | 216.6 ns/op | -50.8% |
| 32 | 438.6 ns/op | 268.1 ns/op | -38.9% |
| 64 | 437.1 ns/op | 281.6 ns/op | -35.6% |

All cases remain at 40 B/op and 3 allocs/op.

### Current-main integration evidence

The fixed current-main A/B/A used base `1eeb912ada6f5eb0a29ac20503a8d7f50c3d3ebe` and combined candidate `e7568b1568ab8d74e499e1ff3c4c6f44f6b85f26`:

| Run | Commit | tpmC | tpmTOTAL |
|---|---|---:|---:|
| A1 [29550222344](https://github.com/matrixorigin/mo-auto-test/actions/runs/29550222344) | base | 44,260.60 | 98,325.07 |
| B [29550787803](https://github.com/matrixorigin/mo-auto-test/actions/runs/29550787803) | P0-1 + P0-3 | 49,727.19 | 110,314.98 |
| A2 [29551321777](https://github.com/matrixorigin/mo-auto-test/actions/runs/29551321777) | base | 42,935.42 | 95,533.01 |

Against the A1/A2 mean, B improved tpmC by **14.0584%** and tpmTOTAL by **13.8100%**. Baseline drift was 2.994%; all runs had zero transaction errors and passed consistency verification.

The focused P0-1 profile also moved in the intended direction: `activeTxnHolder.getActiveTxn` CPU was 6.18 s / 2.03 s / 6.12 s (A1 / B / A2), a 66.99% reduction against the baseline mean. Its sampled mutex delay was 21.19 s / 0.09 s / 44.01 s. `activeTxnHolder.delete` was 0.32 s / below sampling resolution / 42.55 s. The absolute mutex samples are noisy, but the CPU reduction and disappearance of the candidate's sampled delete contention verify that the target holder-wide serialization was removed.

The combined A/B/A is integration and compatibility evidence, not a standalone attribution to P0-1. Historical isolated TPCC run 29390166051 on commit `da21392c` reported +1.43% tpmC versus adjacent baselines and remains directional evidence only.


XuPeng-SH (approved):
No blocking issues found. APPROVE.

I reviewed the complete diff from the ownership, lock-order, publication-order, lifecycle, and unhappy-path perspectives.

Key points checked:
- Each active transaction has exactly one owning shard; duplicate creation is serialized by the shard lock and duplicate deletion does not decrement the count.
- The active count is intentionally conservative: creation increments before map publication and deletion decrements after removal, so empty() cannot report empty while a published transaction still exists.
- close() locks shards in a fixed ascending order and unlocks in reverse order. Other paths hold at most one shard at a time, and I found no new inverse holder.mu -> shard dependency.
- Cross-shard scans are weak snapshots, but their callers either have bind/status synchronization or periodic retry semantics. A transaction created in an already-scanned shard may be handled on the next cycle, but this does not create an unsafe false-negative for bind fencing or service restart.
- Remote-service metadata remains bounded by active transaction lifecycle, and cleanup/count behavior stays idempotent.

Validation:
- ./pkg/lockservice: PASS
- Full ./pkg/lockservice with -race: PASS
- New holder concurrency/shard tests with -race -count=20: PASS
- Local benchmark (Apple M4, GOMAXPROCS=10): 125.0-131.4 ns/op, 40 B/op, 3 allocs/op across 5 runs

The PR is currently behind main, but current main has no overlapping changes in the modified files and I found no merge conflict.

PR #25824, opened 2026-07-16
perf(vector): compact varlen data in CopyBatch

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25472

### Summary

- Adds `Vector.CloneToFlatCompact`, which returns a deep flat copy containing only varlen payload referenced by logical rows.
- Precomputes live varlen bytes, allocates the destination area once, and rewrites offsets instead of repeatedly growing the area.
- Preserves fixed/varlen values, NULL and grouping bitmaps, CONST/CONST NULL flattening, attributes, row count, and source/destination independence.
- Frees all partial mpool allocations on every error path.
- Switches `sql/util.CopyBatch` to the compact clone while leaving ownership and caller cleanup unchanged.

### Why

After row selection, a vector can retain a much larger stale varlen area. CopyBatch previously rebuilt it through repeated UnionAll growth, producing avoidable MPool Grow/reAlloc traffic, mutex contention, allocation bytes, and GC work.

### Compatibility and non-goals

This introduces one Go API and changes only in-memory copying. It does not change SQL behavior, wire or persisted formats, configuration, `Dup`/`DupOffHeap` semantics, or batch ownership.

### Validation

- `go build ./pkg/container/vector ./pkg/sql/util ./pkg/vm/engine/disttae`
- `go vet ./pkg/container/vector ./pkg/sql/util ./pkg/vm/engine/disttae`
- Full tests for vector, sql/util, and disttae
- Clone and CopyBatch race tests repeated 10 times
- Allocation-failure test verifies the destination mpool returns to zero
- Combined P0-1/P0-3 branch: full test and race for lockservice, vector, sql/util, and disttae

Current-main local benchmark, 10 samples:

| Input | old median | compact median | time | bytes/op | allocs/op |
|---|---:|---:|---:|---:|---:|
| stale varlen area | 52.98 us | 19.48 us | -63.2% | 933,936 -> 141,456 (-84.9%) | 42 -> 32 |
| already compact | 20.60 us | 18.14 us | -11.9% | 141,616 -> 141,456 | 42 -> 32 |

### Current-main integration evidence

The fixed current-main A/B/A used base `1eeb912ada6f5eb0a29ac20503a8d7f50c3d3ebe` and combined candidate `e7568b1568ab8d74e499e1ff3c4c6f44f6b85f26`:

| Run | Commit | tpmC | tpmTOTAL |
|---|---|---:|---:|
| A1 [29550222344](https://github.com/matrixorigin/mo-auto-test/actions/runs/29550222344) | base | 44,260.60 | 98,325.07 |
| B [29550787803](https://github.com/matrixorigin/mo-auto-test/actions/runs/29550787803) | P0-1 + P0-3 | 49,727.19 | 110,314.98 |
| A2 [29551321777](https://github.com/matrixorigin/mo-auto-test/actions/runs/29551321777) | base | 42,935.42 | 95,533.01 |

Against the A1/A2 mean, B improved tpmC by **14.0584%** and tpmTOTAL by **13.8100%**. Baseline drift was 2.994%; all runs had zero transaction errors and passed consistency verification.

The focused P0-3 profile matches the implementation mechanism. Against the A1/A2 mean, `CopyBatch` CPU fell 75.36%, its allocation space fell 95.76%, `MPool.Grow` CPU fell 62.69%, and GC cycles fell 28.98%.

This combined A/B/A is integration evidence. The same P0-3 patch previously produced +13.90% tpmC and +14.14% tpmTOTAL in isolated run 29392593786. The profiled implementation commit was `b290103c8`; current PR head `6de31a8a` adds only the bounded grouping-bitmap copy on the fallback path for correctness, so it was not followed by another TPCC run.


pkg/container/vector/vector.go
```diff
@@ -1022,6 +1022,88 @@ func (v *Vector) dup(mp *mpool.MPool, offHeap, areaOffHeap bool) (*Vector, error
...
+// retains varlen payload referenced by the vector's logical rows, so stale or
+// unreferenced bytes in area are not propagated into batch memory accounting.
+func (v *Vector) CloneToFlatCompact(mp *mpool.MPool) (*Vector, error) {
+	w := NewVec(v.typ)
+	if v.class != FLAT || (!v.typ.IsFixedLen() && !v.typ.IsVarlen()) {
+		if err := GetUnionAllFunction(v.typ, mp)(w, v); err != nil {
+			w.Free(mp)
+			return nil, err
+		}
+		return w, nil
```
XuPeng-SH, line 1036: [blocker] This early return drops `gsp` for CONSTANT vectors. Repro: `src := NewRollupConst(types.T_int8.ToType(), 3, mp)`; after `CloneToFlatCompact`, the destination is flat and length 3, but grouping rows 0, 1, and 2 are all absent. `GetUnionAllFunction` expands const NULL/value data without copying grouping metadata, while the FLAT path below explicitly copies `gsp`. Please copy the bounded grouping bitmap on this path (or handle CONST explicitly) and cover rollup CONST fixed/varlen cases.

XuPeng-SH (changes requested):
The compact-copy performance improvement is real, the branch merges cleanly with current `main`, and the package tests/race/build/vet checks pass. I found one correctness gap in the advertised clone contract: CONST rollup vectors lose their grouping bitmap.

Concrete repro: clone `vector.NewRollupConst(types.T_int8.ToType(), 3, mp)`. The result is correctly flattened to three NULL rows, but `dst.GetGrouping().Contains(0/1/2)` is false for all rows. The non-FLAT branch delegates to `GetUnionAllFunction` and returns before `copyBitmapWithinLength`; the const paths in `GetUnionAllFunction` expand NULL/value data but do not propagate `gsp`. This contradicts the PR's explicit guarantee that grouping bitmaps and CONST flattening are preserved, and makes `CloneToFlatCompact` inconsistent between FLAT and CONSTANT inputs.

Please preserve `[0, length)` grouping bits on the fallback/CONST path and add regressions for rollup CONST fixed and varlen vectors (including CONST NULL).

pkg/container/vector/vector.go
VioletQwQ-0: Fixed in `6de31a8a8`. `CloneToFlatCompact` now copies the source grouping bitmap within `[0, v.length)` after the fallback/CONST union succeeds, preserving CONST grouping while excluding stale out-of-range bits. Added table-driven regressions for fixed and varlen CONST values plus fixed and varlen rollup CONST NULL vectors; all cases also verify flattening, NULL/value semantics, and the logical-length bound.

Validation:
- targeted clone tests, `-count=10`
- targeted clone race test, `-count=10`
- full `pkg/container/vector`, `pkg/sql/util`, and `pkg/vm/engine/disttae` tests
- CopyBatch race test, `-count=10`
- vet, build, and `git diff --check`

XuPeng-SH (approved):
Re-reviewed current head `6de31a8a80` against the complete CopyBatch/vector ownership and semantics closure. The previous blocker is fixed: the fallback/CONSTANT path now copies grouping bits only within `[0, length)`, and the regressions cover fixed and varlen CONST values plus fixed and varlen rollup CONST NULL vectors.

The fast path remains semantically equivalent to the prior UnionAll copy for logical rows while avoiding stale varlen payload: fixed/varlen data, NULLs, grouping, CONST flattening, batch attrs/row count, source independence, and partial-allocation cleanup are covered. Allocation ownership is unchanged (`CopyBatch` owns the cloned vectors until workspace/batch cleanup); every failure path frees the partially built vector and previously cloned columns. No waits, goroutines, persistent cache, or unbounded cross-call accumulation are introduced.

Fresh validation passed: targeted tests `-count=10`, targeted race tests `-race -count=5`, full `pkg/container/vector`, `pkg/sql/util`, and `pkg/vm/engine/disttae` tests, and a local benchmark confirming lower time/bytes/allocs for stale-area inputs with no compact-input regression. I also merged the latest `origin/main` (`e2c5dc00bb`) into a temporary worktree without conflicts and reran the focused tests successfully.

PR #25825, opened 2026-07-17
fix: stop TN read clock wait on context cancel

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25808

fix: stop TN read clock wait on context cancel


XuPeng-SH (changes requested):
Blocking protocol correctness issue: the new response error cannot be consumed by the CN transaction client.

`waitClockTo` returns `context.Canceled` or another Go context cause. This change passes that value through `moerr.ConvertGoError`, which converts it to `moerr.ErrInternal`, then stores it in `TxnResponse.TxnError`. However, `pkg/txn/client/operator.go` only accepts the defined Read protocol errors in `readTxnErrors`; `ErrInternal` is not one of them. When the response is delivered, `txnOperator.handleErrorResponse` replaces the cancellation with:

`invalid txn error, code 20101`

I reproduced this on head `abbc914c22` with a focused consumer-side test that builds exactly the response produced here and passes it to `txnOperator.handleErrorResponse`; it fails deterministically because the CN rejects the TN error code.

This is reachable beyond an already-canceled outer request. Replica removal cancels the acquired service lease context while the outer RPC context can still be alive. In that case `Read` creates this `ErrInternal` response and the server can write it successfully, so the client observes the invalid protocol error instead of a cancellation or retryable shard-lifecycle result. Local dispatch can consume the response directly as well.

Please classify the cancellation source and preserve the end-to-end contract:

- caller cancellation should propagate as the caller cancellation path rather than an unsupported TxnError code;
- replica lifecycle cancellation while the caller is still active should use a supported retryable result such as the existing TN-shard-not-found path, if that is the intended retry semantics;
- if a new wire-level cancellation code is introduced, update both producer and consumer explicitly.

Add a test through the actual sender/operator consumer that asserts the exact resulting error, not merely `resp.TxnError != nil`. Also make the mid-wait test deterministic: `time.Sleep(20ms)` does not prove that the goroutine entered the clock wait before cancellation and can false-pass under scheduler delay. Reuse a channel/injected clock phase barrier similar to `TestWaitClockToHonorsContext`.

XuPeng-SH (approved):
Approved on current head `6d04781b96`.

The latest changes close the previously reported wire-protocol gap:

- caller-context cancellation is returned through the sender error path instead of being encoded as an unsupported `TxnError`;
- replica-lease cancellation is mapped to the existing retryable `ErrTNShardNotFound` protocol result;
- unrelated `TxnService.Read` errors remain unchanged;
- the CN transaction client is covered consuming the exact `TNShardNotFound` response.

I traced the full lifecycle and wait graph (`store.doRead` -> replica lease -> `TxnService.Read` / clock timer -> RPC/local sender -> `txnOperator.handleErrorResponse`). Lease release remains exactly once, replica close cancels in-flight reads and waits for `activeCalls` to drain, caller cancellation wins when both contexts are canceled, and the success hot path adds no allocation, lock, or retry work.

- focused cancellation/protocol tests passed 50 times;
- focused race tests passed 20 times;
- full ordinary and race tests passed for `pkg/tnservice`, `pkg/txn/service`, and `pkg/txn/client`;
- `go build`, `go vet`, `make err-check`, and `git diff --check` passed;
- synthetic merge with current `origin/main@826a3e3d2` passed.

No blocking correctness, performance, leak, hang, or unbounded-growth issue found. GitHub UT/BVT/coverage jobs are still running and remain the merge gate.

PR #25831, opened 2026-07-17
fix: release create source options during reset

### LINKED ISSUE #25075: [Bug]: CREATE SOURCE WITH options panics in CreateSource.reset

### Branch Name

3.0-dev / nightly regression head

### Commit ID

Unknown. Found while verifying SQL coverage.

```Markdown
- Server: MatrixOne, MySQL protocol
- Statement type: CREATE SOURCE ... WITH (...)
```

### Actual Behavior

`CREATE SOURCE ... WITH (...)` can return an internal panic during statement cleanup:

```text
ERROR 1105 (HY000): internal error: panic miss Free for &{{<nil>} type kafka}:
github.com/matrixorigin/matrixone/pkg/sql/parsers/tree.(*CreateSource).reset
    /data1/runners/action-runner/_work/mo-nightly-regression/mo-nightly-regression/head/pkg/sql/parsers/tree/source.go:200
github.com/matrixorigin/matrixone/pkg/sql/parsers/tree.init.19.func2
    /data1/runners/action-runner/_work/mo-nightly-regression/mo-nightly-regression/head/pkg/sql/parsers/tree/source.go:26
github.com/matrixorigin/matrixone/pkg/common/reuse.(*syncPoolBased
```

This appears to be independent of protected database behavior. The protected database test was only where the issue was first noticed.

From code inspection, `CREATE SOURCE ... WITH (...)` parses each option as `*CreateSourceWithOption`, while `CreateSource.reset()` frees many `TableOption*` variants but does not handle `CreateSourceWithOption`, so cleanup can hit:

```go
panic(fmt.Sprintf("miss Free for %v", item))
```

### Expected Behavior

`CREATE SOURCE ... WITH (...)` should not panic during parser/AST cleanup.

The statement should either execute normally or return a deterministic user-facing error if the source configuration/runtime environment is invalid.

### Steps to Reproduce

Minimal standalone reproduction, no protected database setup required:

```sql
drop database if exists source_panic_repro;
create database source_panic_repro;
use source_panic_repro;

create source src1(a varchar) with (
  "type"='kafka',
  "topic"='t1',
  "partition"='0',
  "value"='json',
  "bootstrap.servers"='127.0.0.1:9092'
);
```

### Additional information

This was originally observed while extending protected database privilege coverage, but the panic is not a protected database bug. It is a standalone `CREATE SOURCE ... WITH (...)` cleanup issue.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25075

`CREATE SOURCE ... WITH (...)` stores each parsed option as a
`CreateSourceWithOption`. `CreateSource.reset` did not release that option type,
so statement cleanup panicked after otherwise successful parsing.

Changes:
- release `CreateSourceWithOption` entries during `CreateSource.reset`;
- add a parser regression test that parses the reported statement and frees the AST.

Validation:
- `go test ./pkg/sql/parsers/dialect/mysql -run '^TestCreateSourceWithOptionsFree$' -count=1`
- `go test ./pkg/sql/parsers/tree ./pkg/sql/parsers/dialect/mysql -count=1`
- `git diff --check`


XuPeng-SH (approved):
Reviewed current head `3504ebe612` across the complete parser AST ownership path. The fix closes the exact missing ownership edge: every `CreateSourceWithOption` allocated by both source-option grammar branches is owned by `CreateSource.Options` and is now returned once during the parent `Free -> reset` path. The option value is a non-reuse `Expr` object and remains GC-managed after the wrapper reset, consistent with neighboring AST nodes.

No double-free path, wait dependency, persistent accumulation, SQL behavior change, or compatibility boundary is introduced. Empty/no-option statements are unchanged; one or many options follow the same deterministic cleanup loop.

Fresh validation passed: regression `-count=100`, focused `-race -count=20`, full `pkg/sql/parsers/tree` and `pkg/sql/parsers/dialect/mysql` tests, `git diff --check`, and the focused test after a clean temporary merge of latest `origin/main` (`e2c5dc00bb`).

PR #25834, opened 2026-07-17
fix(proxy): close backend sessions on connect errors

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25826

After `CNServer.Connect` establishes a backend TCP session, salt validation or initial metadata encoding/write can still fail. Those error paths returned `nil, err` without closing the session, so the caller could not release it.

This change keeps ownership local until each construction step succeeds:

- `CNServer.Connect` closes the goetty session on every pre-handoff error;
- `newServerConn` closes that session if frontend IO-session construction fails;
- successful connection and handshake ownership semantics remain unchanged.

The regression test establishes a real TCP connection, triggers invalid-salt validation, and verifies that the backend immediately observes EOF.

Tested with MatrixOne's macOS CGo header/link settings:

- `go test ./pkg/proxy -run '^TestCNServerConnectClosesSessionOnInvalidSalt$' -count=10`
- `go test -race ./pkg/proxy -run '^TestCNServerConnectClosesSessionOnInvalidSalt$' -count=5`
- `go build ./pkg/proxy`
- `go vet ./pkg/proxy`


XuPeng-SH (approved):
Reviewed current head 6b91a532d on a clean synthetic merge with current main. No blocking findings.

The ownership transfer is correct at both layers. CNServer.Connect retains the goetty session until connect, salt validation, ExtraInfo encoding, and the initial flushed write have all succeeded; every earlier return closes the session. newServerConn then owns that returned session until frontend IO-session construction and protocol initialization complete. The owned flags are cleared only at the final handoff points.

I also checked the lower-level contracts: goetty Connect and Write return errors without internally releasing the IOSession reference, while frontend NewIOSession releases its own partial buffer allocation on failure but does not close the supplied raw connection. The new defers are therefore the unique cleanup owners rather than double-close paths. goetty Close performs local socket/ref/buffer cleanup without introducing a wait-for dependency.

Validation: the real-TCP invalid-salt EOF regression passed under the race detector x20; the full proxy package race suite passed; proxy build and vet passed. CI is green and the PR merges cleanly with current main.

LeftHandCold:
The failed `TestDiskCacheQuotaExceeded` UT is unrelated to this PR's proxy changes.

Root cause: the preceding `TestDiskCacheReadEnsuresMemoryCacheCapacity` test allocates a 10-byte `CachedData` result but does not call the caller-owned `IOVector.Release()`. The checked allocator reports that missing free asynchronously from its GC finalizer, so `TestDiskCacheQuotaExceeded` is only the test running when the panic surfaces. The allocation size and full stack in the job match that earlier test exactly.

I reproduced it on current `main` with `MO_MALLOC_DEBUG=1` and repeated execution. Tracking issue: #25840. Independent ready fix: #25842.

I kept the fix separate because it is a one-line FileService test-lifecycle correction and does not belong in this already-reviewed proxy change.

PR #25835, opened 2026-07-17
fix(logservice): make HAKeeper retry backoff cancellable

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25828

NewLogHAKeeperClientWithRetry checked the caller context at the top of its retry loop, but a failed attempt then entered an unconditional three-second sleep. Cancellation during that backoff was not observed until the sleep completed.

This change replaces the sleep with a retry timer selected together with the caller context. The existing outer retry-loop check continues to enforce the two-minute failure limit; successful client creation and the three-second retry interval remain unchanged.

The regression test injects deterministic creation failures, verifies that one backoff expires and causes a second attempt, then cancels during the next backoff and verifies immediate return.

Tested with MatrixOne macOS CGo header/link settings:

- go test ./pkg/logservice -run Test_NewLogHAKeeperClientWithRetry -count=10
- go test -race ./pkg/logservice -run Test_NewLogHAKeeperClientWithRetry -count=5
- go build ./pkg/logservice
- go vet ./pkg/logservice
- go test ./pkg/logservice -count=1


volgariver6 (approved):

XuPeng-SH (approved):
Approved after a first-principles lifecycle and unhappy-path review.

The change closes the reported cancellation gap correctly:
- each failed create attempt owns at most one retry timer;
- cancellation stops that timer and returns immediately;
- the fired path consumes the timer channel;
- cancellation racing with timer expiry is safe because the outer loop observes the canceled context on the next iteration;
- the production caller handles the nil result as an error, so the return path is closed.

Validation on the PR head merged with current main:
- focused cancellation test: normal count=5, race count=3;
- full ./pkg/logservice: normal and race passed;
- ./pkg/datasync passed;
- go build and go vet passed for both affected and dependent packages.

No blocking correctness, leak, hang, reuse, or unbounded-state issue found.

PR #25836, opened 2026-07-17
fix(proxy): move conn cache I/O out of global lock

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25827

`connCache.Push` and `Pop` performed reset-session RPC and backend MySQL I/O while holding the one mutex shared by every tenant. One unresponsive cached CN connection could therefore block unrelated cache operations and prevent cache shutdown.

This change separates cache state transitions from backend I/O:

- `Push` checks capacity before reset-session and revalidates it before insertion;
- `Pop` reserves a connection by removing it from the available queue, while retaining cache ownership in `allConns` until successful handoff;
- `Close` can hard-close both available and in-flight connections, interrupting a blocked `SET CONNECTION ID`;
- authentication failure keeps the previous behavior by requeueing the connection when capacity still permits;
- discarded and shutdown connections use local close instead of a potentially blocking COM_QUIT round trip.

No cache key, authentication rule, capacity limit, FIFO selection rule, or successful handoff behavior is changed.

Tested with MatrixOne's macOS CGo header/link settings:

- `go test ./pkg/proxy -run '^TestConnCache(BlockedPopDoesNotBlockOtherTenantsOrClose|CloseDoesNotWaitForPushReset)$' -count=10`
- `go test -race ./pkg/proxy -run '^TestConnCache' -count=1`
- `go build ./pkg/proxy`
- `go vet ./pkg/proxy`


XuPeng-SH (approved):
Reviewed head 136ffde273bc6f9f590915d09b456655c4b399cc against the latest main (096940bd1b4edda487d3406aac08b4a483c817fc).\n\nThe concurrency/ownership model is sound:\n- Push performs reset I/O outside the global mutex and revalidates closed, duplicate, total-capacity, and per-tenant-capacity state before insertion.\n- Pop reserves an entry outside the available queue while retaining cache ownership in allConns, so Close can interrupt blocked SET CONNECTION ID I/O.\n- Successful Pop handoff and Close are linearized by the same mutex; Close owns the connection only when it wins that race.\n- Authentication-failure requeue rechecks tenant capacity, and closeOnce makes eviction vs shutdown cleanup idempotent.\n\nVerification on the synthetic latest-main merge:\n- go test ./pkg/proxy\n- go test -race ./pkg/proxy\n- go build ./pkg/proxy\n- go vet ./pkg/proxy\n\nI also stress-checked the new focused tests (20 normal iterations and 5 race iterations) and exercised concurrent capacity revalidation, auth-failure requeue with a filled tenant slot, and Pop-handoff-before-Close. No blocking correctness issue found.

PR #25837, opened 2026-07-17
fix: bound HashBuild memory and spill under query budgets

### LINKED ISSUE #25782: [Bug]: HashBuild can bypass spill and trigger system OOM during large hash table growth
## Description

A long-running `INSERT ... SELECT` with several grouped subqueries and `LEFT JOIN`s caused the CN process to be killed by the node OOM killer. The failing query's pipeline mpool retained `23.78 GB`; immediately before the kill, hash-table-like allocations increased it by another `7.52 GB` in about 1 ms. HashBuild did not enter spill mode.

This is not only a sizing/tuning problem. The current spill decision excludes broadcast hash builds, and eligible shuffle hash builds check retained input-batch memory before `BuildHashmap`, with no admission check during hash table growth.

## Incident context

All times below are UTC on `2026-07-15`.

| Field | Value |
|---|---|
| Environment | cn-dev, Kubernetes namespace `freetier-01` |
| Pod | `freetier-01-s16c64g-854877c9c-cn-bnxq8` |
| Pod UID | `0338817b-437f-4681-b339-6dff5aed6c64` |
| Node | `cn-hangzhou.10.3.93.228` |
| Image | `matrixone:v4.1.0-c58a0e62c-2026-07-11` |
| MO commit | `c58a0e62c` |
| Session/pipeline | `019f6571-fccd-7735-ac92-4f15a51967de` |
| Statement | `019f6571-fd12-7263-ac05-61e875ec0c01` |
| Transaction | `4f352d3cf2ca125118c1424e75ff854a` |
| Transaction created | `11:03:17.010486` |
| Statement build completed | `11:03:21.854903` (`Exec.Build` took `4.844 s`) |
| Process killed | `11:32:19` |
| Replacement process started | `11:32:22` |

The statement shape was:

```sql
INSERT INTO <target>.voucher_line_details__stg
SELECT ...
FROM ...
LEFT JOIN (... GROUP BY ...) ...
LEFT JOIN (... GROUP BY ...) ...
...
GROUP BY ...;
```

The account name, user name, client address, and business columns are omitted from this public issue. The IDs above are sufficient to locate the complete statement and trace in the dev observability data.

## Failure evidence

The current pod status preserves the previous container termination:

```text
lastState.terminated.reason     = OOMKilled
lastState.terminated.exitCode   = 137
lastState.terminated.finishedAt = 2026-07-15T11:32:19Z
state.running.startedAt         = 2026-07-15T11:32:22Z
restartCount                    = 2
```

The Kubernetes event reported a node-level OOM and selected `mo-service` as the victim:

```text
SystemOOM encountered, victim process: mo-service
```

This was not a normal query error or a Go runtime OOM: the process stopped without a statement completion/error log and the container was restarted after exit 137.

## Memory state at the scene

Historical Prometheus data for the old container, queried for `11:20:00-11:33:00`:

| Signal | Observation |
|---|---|
| Node `MemTotal` | `61.54 GiB` |
| Node `MemAvailable` | `2.63 GiB` at `11:20`; recovered to `45.19 GiB` after restart |
| Container resource limit metric | `85 GiB` from `11:20:10` through the OOM |
| Container RSS | `42.87 GiB` at `11:20`; last/max sample `44.17 GiB` at `11:30:45` |
| Container working set | last/max sample `44.43 GiB` at `11:30:45` |
| Container usage | max sample `44.74 GiB` at `11:29:00` |

The configured container limit exceeded the node's physical memory, so it could not protect the node. The current pod spec later shows `21 GiB`; that is post-incident state and must not be used as the limit at the time of failure.

At `11:30:40.066771`, the failing pipeline reported:

```text
MPool pipeline-019f6571-fccd-7735-ac92-4f15a51967de new high watermark
allocations     : 163571
frees           : 131264
alloc bytes     : 24656403965
free bytes      : 8394263181
current bytes   : 16262140784
high water mark : 16262140784
```

By `11:30:40.067843`, about `1.07 ms` later:

```text
allocations     : 163599
frees           : 131264
alloc bytes     : 32172596733
free bytes      : 8394263181
current bytes   : 23778333552
high water mark : 23778333552
```

Delta:

```text
allocations:  +28
frees:          0
bytes:        +7,516,192,768
average:       268,435,456 bytes/allocation (exactly 256 MiB)
```

The global mpool moved in lockstep from `17,323,812,368` to `24,840,005,136` bytes. This strongly correlates the final memory step with one pipeline rather than gradual accumulation across unrelated sessions. The repeated 256 MiB allocation size is consistent with hash table block growth; this attribution is an inference from allocation shape plus the execution/code path, not a symbolized allocation stack.

Earlier pipeline high-watermarks show sustained growth before the final step:

```text
11:03:25   1.07 GB
11:03:31   5.42 GB
11:06:09   7.66 GB
11:12:03  10.86 GB
11:18:35  15.17 GB
11:30:40  16.26 GB -> 23.78 GB in ~1 ms
11:32:19  process killed
```

The transaction leak detector began reporting this transaction at `11:18:46` and continued until the process died. Those messages establish that the same transaction remained active; they do **not** by themselves prove a transaction leak.

## Spill evidence

A Loki search on the affected pod for `entering spill mode` over `10:50:00-11:35:00` returned no matching line for this execution. HashBuild emits that exact log when `shouldSpillBatches()` succeeds, so the observed execution did not enter that spill path.

### Broadcast hash builds cannot spill

[`pkg/sql/colexec/hashbuild/spill.go#L273-L285`](https://github.com/matrixorigin/matrixone/blob/c58a0e62c/pkg/sql/colexec/hashbuild/spill.go#L273-L285):

```go
func (hashBuild *HashBuild) shouldSpillBatches() bool {
    if !hashBuild.CanSpill || !hashBuild.IsShuffle || !hashBuild.NeedHashMap {
        return false
    }
    // ...
}
```

[`constructBroadcastHashBuild`](https://github.com/matrixorigin/matrixone/blob/c58a0e62c/pkg/sql/compile/operator.go#L1824-L1851) sets both `IsShuffle = false` and `CanSpill = true`. The `IsShuffle` condition therefore makes spilling unreachable for a broadcast hash join even when runtime cardinality is much larger than planned.

### Spill is checked before hash map construction

[`pkg/sql/colexec/hashbuild/build.go#L181-L216`](https://github.com/matrixorigin/matrixone/blob/c58a0e62c/pkg/sql/colexec/hashbuild/build.go#L181-L216) copies each input batch and checks spill based on retained batch memory. If that check never fires, [`BuildHashmap`](https://github.com/matrixorigin/matrixone/blob/c58a0e62c/pkg/sql/colexec/hashbuild/build.go#L245-L255) runs afterward.

There is no spill/budget decision while the hash map is being built. `Int64HashMap.ResizeOnDemand` and `StringHashMap.ResizeOnDemand` return allocation errors, but do not receive a query/hash-build budget or a way to return `NeedSpill` before committing the next large growth step.

### Existing limits do not provide a reliable fallback

- `join_spill_mem = 0` selects an automatic threshold; it does not disable spill, but the threshold cannot help on the excluded/runtime-growth paths above.
- The session mpool uses `GuestMmuLimitation`, whose default is 1 TiB.
- The global mpool limit also defaults to 1 TiB when `[limit].memory` is not configured.
- `ProcessLimitationSize` is assigned to the process but is not enforced on this allocation path.
- In this incident, the Kubernetes resource-limit metric (`85 GiB`) was itself larger than node physical memory (`61.54 GiB`).

## Expected behavior

A hash build must not cause node-level OOM because planner estimates were low or hash table overhead grew after the batch-memory spill check.

Before the next hash table growth exceeds the query/hash-build budget, execution should spill, repartition, fall back from broadcast to a partitioned strategy, or return a controlled query error.

## Proposed direction

1. Introduce an enforceable per-query/hash-build memory budget.
2. Estimate projected hash table memory before `BuildHashmap` and enter spill mode when it exceeds the budget.
3. Add admission control to hash table `ResizeOnDemand` before allocating additional blocks.
4. Return a typed `NeedSpill` result when growth cannot be admitted, allowing HashBuild to partition and retry.
5. Support runtime fallback for broadcast hash joins, or reject broadcast plans whose build side cannot be bounded safely.
6. Derive automatic budgets from `min(cgroup limit, node allocatable memory)` with reserve for Go heap, cache, other operators, and other containers.
7. Independently prevent workload/container limits from exceeding usable node memory.

## Acceptance criteria

- A high-cardinality build side with underestimated statistics does not cause process/node OOM.
- Broadcast hash build either spills/falls back or terminates the query with a controlled error.
- Shuffle hash build can spill based on projected hash table size, not only retained input batch size.
- Hash table expansion cannot exceed its query budget by more than one bounded allocation block.
- Tests cover both broadcast and shuffle joins with a deliberately small memory budget.
- Metrics/logs identify the join operator, query/statement ID, threshold, projected hash size, spill reason, and spilled bytes.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25782

HashBuild could grow past spill and process-memory safeguards, allowing large joins to exhaust a CN and trigger a system OOM. This PR makes memory admission finite and hierarchical, and gives shuffle joins a bounded spill/re-spill path when hard admission rejects growth.

The commits are separated so the correctness, performance, diagnostics, and validation infrastructure can be reviewed independently. Together they:

- make hashmap resize transactional so a rejected allocation leaves the old table usable;
- account retained batches, hashmap growth peaks, expressions, spill scratch/readers, disk bytes, and file descriptors;
- propagate process spill limits to remote pipelines;
- deliver one terminal build error to all broadcast consumers instead of exposing a partial build;
- avoid double-charging an already-accounted retained batch when emergency spill scratch grows;
- preserve expression join keys in shuffle spill/re-spill by admitting the full evaluator-tree peak before `Eval`, returning a bounded budget error before allocation when that peak cannot be admitted;
- bound group spill-reload preallocation by proven cardinality and the active spill-memory limit;
- compact and reuse spilled aggregate state and remove avoidable decimal aggregation/hash-probe work;
- emit sparse spill/budget diagnostics to statement profiles and durable logs;
- add native `DUMP TABLE` / `LOAD TABLE` fixture support and reusable image-build caches for repeatable TPCH-scale validation;
- cover budget compatibility, ownership transfer, overflow, provider failure, spill-record merge, and cleanup/error paths.

### `LOAD TABLE` trust and integrity boundary

`DUMP TABLE` / `LOAD TABLE` is an administrator-level fixture transport, not an untrusted file-ingestion API. The manifest and the configured `ObjectCopier` are trusted inputs. The streaming fallback computes and verifies SHA-256, while a provider-native server-side copy deliberately verifies object existence and size without reading the copied object back through CN to recompute its hash or physical metadata.

This is an intentional performance contract: re-reading and hashing every copied object would turn a metadata/server-side-copy operation back into a full 100 GB / 1 TB data scan, consume CN bandwidth and CPU, and erase the measured fixture-load benefit (about 9 seconds for 100 GB and 44 seconds for 1 TB). Same-size corruption introduced by a trusted storage provider is therefore outside this SQL command contract and belongs to the provider checksum/integrity layer.

## Validation

After rebasing onto current `main`:

- `go build` and `go vet` pass for `process`, `group`, `hashbuild`, `hashjoin`, and `spillutil`;
- full tests pass for the same packages;
- focused `-race -count=3` tests pass for retained emergency scratch growth, bounded group reload preallocation, and expression-key spill admission;
- a real shuffle HashBuild test verifies an expression join key spills successfully, while a low-budget test verifies rejection happens before evaluator allocation;
- the latest coverage-only commit passes `go test ./pkg/vm/process ./pkg/sql/colexec/spillutil ./pkg/sql/colexec/hashbuild`;
- the earlier two-CN hard-budget reproduction returned exactly `132096`, with spill on both CNs and no OOM, OOM kill, or swap;
- TPCH fixture load was measured at about 9 seconds for 100 GB and 44 seconds for 1 TB.

Performance run: https://github.com/matrixorigin/mo-auto-test/actions/runs/30081071473

- TPCH 100 GB: four measured turns average **87.592 s**, versus **98.679 s** on the latest `main` daily run (**11.2% faster**). All queries succeeded and no spill occurred.
- TPCH 1 TB: two measured turns completed in **969.523 s** and **973.105 s**, averaging **971.314 s**, versus **1113.188 s** on the latest `main` daily run (**12.7% faster**). The two turns differ by only **0.37%**.
- Both measured 1 TB turns exercised the real spill path, spilling **60.14 GB** and **77.57 GB** respectively in Q9 and Q18. All queries succeeded without OOM or spill-budget errors, so the improvement is not obtained by avoiding spill.
- The latest `main` run also spilled in Q18 through the existing group-aggregation spill path, but it did **not** enter HashBuild spill for Q9. With this PR, Q9 reached the 40 GiB query-wide HashBuild budget and spilled **10.96 GB** and **28.39 GB** in the two measured turns. Q9 still completed in **92.33 s** and **92.78 s**, effectively flat against the `main` average of **92.39 s**, demonstrating added OOM protection without a material Q9 regression.


aptend:
Fixed the two BVT regressions on top of latest `main` in two commits:

- spill record merge now reserves the allocator-rounded destination capacity (`O + N + D`) before `UnionBatch`, then reconciles to actual retained bytes and cleans every failure path;
- prepared parameters are charged at their physical const cardinality (1), with actual payload accounting for values above the declared type bound.

Local business validation on a fresh service:
- low-threshold dedup spill ODKU completed without OOM/budget-invalid and produced 600000 rows (the expected result was corrected to match current `generate_series` semantics);
- prepared modulo join returned 5000000 without admission rejection.

Also passed focused package tests, race tests, `go vet`, full binary build, and an independent final diff review.

XuPeng-SH (changes requested):
发现至少 1 个明确的 correctness blocker，另外还有 1 个完整性风险：

1. `LOAD TABLE` 现在会直接把 manifest 里的 `ObjectStats` 当成权威元数据写进 catalog，而不再回读目标对象做物理元数据校验。`handleLoadTable` 在 `pkg/frontend/table_dump.go:1199-1253` 这里只检查了 `len(item.Stats)`、`GetAppendable()` 和 `ObjectName()`，随后 `submitTableDumpObjects` 在 `table_dump.go:990-1045` 里直接把这些 stats 展开并提交给 relation。相比 main 分支原先会通过 `loadPhysicalTableDumpStats/validatePhysicalTableDumpObjects` 重新读取 extent/blocks/rows/size/sort-key zone map 并覆盖 manifest stats，这里少了最后一道 fail-closed 校验。结果是：只要 manifest 里的对象名对得上，一个 stale/forged manifest 就能把错误的 extent、block count、row count、origin size、sort-key zone map 写进 catalog；后续 block location 展开和 zone-map pruning 都会基于这些错误 stats 工作，直接变成错误结果风险，而不仅仅是装载失败。

2. provider-side `CopyObject` 路径现在也把 manifest 的 SHA-256 检查降成了 size-only。`copyTableDumpFile` 在 `pkg/frontend/table_dump.go:348-382` 对 `providerCopied` 返回空 hash，`installTableDumpObject` 在 `table_dump.go:921-929` 里对 `providerCopied` 只看 size 相等就放行；`verifyTableDumpObject` 对空 hash 也是 size-only（`table_dump.go:886-896`）。现有测试甚至显式锁定了这个行为：`pkg/frontend/table_dump_test.go:838-843` 允许 same-size 的 `"corrupt object byte"` 成功安装，`pkg/frontend/table_dump_test.go:964-972` 也断言 provider-copy 返回空 hash 并只做 size 校验。

如果你坚持当前 PR 描述里的“trusted manifest / trusted provider”边界，第 2 点可以算风险接受；但第 1 点我认为仍然是 blocker，因为它把未经物理校验的 stats 直接写成了后续读路径的权威 catalog 元数据。

aptend:
关于 physical `ObjectStats` 校验，我不同意把它作为当前 PR 的 correctness blocker，也不会把逐 object 物理 metadata 回读重新放回默认 `LOAD TABLE` 路径。这里不是遗漏校验，而是已经通过性能实验确定的设计边界。

`DUMP TABLE` / `LOAD TABLE` 的定位是管理员级 fixture/restore 通道，不是接受任意 manifest 的文件导入接口。实际闭环是：

1. `DUMP TABLE` 从 catalog 生成 manifest stats，同时复制对应的 immutable object；
2. 所有 object 完成后才写 ready marker；
3. `LOAD TABLE` 要求 relation topology/schema hash、object name、non-appendable 标志、fixture path、size、对象数和 block 数边界全部匹配；
4. provider-side `CopyObject` 复制 manifest 指向的同名 immutable object，随后确认目标对象存在且 size 一致；
5. manifest 和配置的 provider 都是该管理员操作的 trusted input。

在这个契约下，manifest stats 就是备份元数据的权威来源。stale/forged manifest 与 trusted provider 产生 same-size corruption 属于同一个被明确排除的威胁模型。若认为 manifest 可被不可信方修改，那么仅恢复 extent/blocks/rows/zone-map 回读也不是完整的安全方案：对象 payload 仍可能发生同尺寸篡改，应该另行设计 authenticated manifest 和 provider checksum，而不是只在 catalog stats 这一侧增加昂贵的部分防御。

这里也需要区分两类成本：SHA-256 回读会扫描完整 100 GB / 1 TB；物理 stats 校验虽只读取 header/meta range，但仍会为每个 object 增加远端 GET、metadata 解析和 CN 路径，成本按 object 数增长，直接破坏 server-side copy 后的快速 catalog install。我们之前正是因为实际大规模验证显示这些逐对象防御让 LOAD 接近重新导入的慢路径，才删除它们。

当前真实结果是：

- TPCH 100 GB fixture load 约 **9 秒**；
- TPCH 1 TB fixture load 约 **44 秒**；
- 性能验证 run：https://github.com/matrixorigin/mo-auto-test/actions/runs/30081071473
- load 后 100 GB 四轮、1 TB 两轮 TPCH 查询全部成功；
- 100 GB 平均 **87.592 秒**，比 main **快 11.2%**；
- 1 TB 平均 **971.314 秒**，比 main **快 12.7%**，并且两轮实际产生 **60.14 GB / 77.57 GB spill**，没有错误结果、OOM 或 spill-budget error。

这些数据验证的是我们真正支持的闭环：由 `DUMP TABLE` 生成、存放在受信 stage、通过受信 provider 复制并由 `LOAD TABLE` 安装。它并不声称支持人工构造或不可信 manifest。

因此本 PR 保持当前 fast trusted-restore 路径，不恢复默认 physical metadata reread。若产品后续需要接收不可信备份，可以单独设计显式 strict validation/authenticated manifest 模式，并单独承担和验收其性能成本。

aptend:
补充精确的逐提交 A/B 数据。刚才重新核对了当时三个 TKE job 的原始日志；使用的是同一个 COS fixture `tpch100g/v1` 和同一条 TPCH benchmark load 路径：

| MatrixOne commit | 校验路径 | TPCH 100G fixture load |
|---|---|---:|
| `38873779232a` | provider copy 后 SHA-256 回读 + physical metadata 校验 | **511s** |
| `51b71138bfa5` | 去掉 provider SHA-256 回读，保留 physical metadata 校验 | **453s** |
| `27a74f04f50d` | 同时去掉默认 physical metadata 回读，采用 trusted manifest stats | **11s** |

对应原始 run/job：

- 511s：https://github.com/matrixorigin/mo-auto-test/actions/runs/30069256247/job/89409077135
- 453s：https://github.com/matrixorigin/mo-auto-test/actions/runs/30071510434/job/89414526094
- 11s：https://github.com/matrixorigin/mo-auto-test/actions/runs/30081071473/job/89446382317

三个日志分别明确打印：

- `Loaded tpch100g fixture ... in 511s`
- `Loaded tpch100g fixture ... in 453s`
- `Loaded tpch100g fixture ... in 11s`

两个 MatrixOne 变更也是连续且隔离的：`51b71138` 只移除 provider-copy 后的重新哈希；`27a74f04` 再移除逐 object physical stats 回读。auto-test 在第一和后两次之间只有删除未使用 standalone native-fixture jobs 的变化，benchmark/load 脚本没有变化。

因此这不是笼统的性能担忧：

- 仅去掉 SHA 回读：511s → 453s，减少 **58s**；
- 再去掉 physical metadata 回读：453s → 11s，减少 **442s**，load 快约 **41.2 倍**；
- 两类重复校验全部去掉：511s → 11s，减少 **500s**，快约 **46.5 倍**。

同一 final run 的 1T fixture load 为 **49s**；另一次独立 load 记录为约 **44s**。所以恢复默认 physical metadata 校验会直接否定该功能的规模化价值。

pkg/sql/colexec/spillutil/join_spill.go
```diff
@@ -659,20 +1660,35 @@ func (e *SpillEngine) RebuildHashmap(proc *process.Process, analyzer process.Ana
 	// Pop the head bucket and open probe reader.
 	e.buckets = e.buckets[1:]
 	if bucket.ProbeFd != nil {
-		e.probeReader.ResetForFd(bucket.ProbeFd)
+		if err := e.probeReader.EnsureBuffer(e.cfg.Budget); err != nil {
```
XuPeng-SH, line 1667: This probe-buffer admission still sees stale build-batch reservations when `NeedBatches=false`: the physical batches were cleaned just above, but `GetJoinMap` transferred their tokens. A focused 8,192-row int32 reproduction left `generation.Used()` unchanged at 1,769,472 bytes across that clean. Please release batch ownership before this admission through a HashmapBuilder lifecycle API.

pkg/sql/colexec/spillutil/join_spill.go
```diff
@@ -722,13 +1766,26 @@ func (e *SpillEngine) reSpillBucket(proc *process.Process, analyzer process.Anal
 			}
 		}
 	}()
+	var buildRows int64
 	for i, b := range buildBatches {
-		if err := evalAndScatter(b, buildWriters, buildBuffers, e.keyExecs); err != nil {
+		if b != nil {
+			buildRows += int64(b.RowCount())
+		}
+		if err := evalAndScatter(b, buildWriters, nil, e.keyExecs); err != nil {
```
XuPeng-SH, line 1766: The re-spill loop cleans these detached physical batches, but their reservations remain in the builder until the caller frees it after `reSpillBucket` returns. Since this is the admission-recovery path, those phantom charges can reject the expression/scatter/read scratch needed to progress. Please drain `Batches` and reservation ownership together rather than detaching `Batches.Buf` directly.

XuPeng-SH (changes requested):
Request changes.

I withdraw my previous physical ObjectStats blocker after re-checking the enforced tenant-admin boundary, DUMP publication order, and the supplied A/B data. Same-size corruption is outside the explicitly trusted fixture/provider contract and is not a blocker for this PR.

There is, however, a blocking ownership gap in the new HashBuild budget implementation: SpillEngine can destroy retained build batches without releasing the reservations that own those bytes.

1. In RebuildHashmap with NeedBatches=false, builder.Batches.Clean releases the physical mpool allocations, but GetJoinMap immediately transfers builder.batchReservations into the JoinMap. Semi/anti/mark-style joins therefore carry phantom batch charges throughout probing. A focused 8,192-row int32 reproduction left generation.Used unchanged at 1,769,472 bytes before and after the physical clean.

2. reSpillBucket detaches builder.Batches.Buf and cleans each batch, but the corresponding reservations remain charged until reSpillBucket returns and the caller later frees the builder. This is specifically the recovery path entered after an admission rejection; retaining already-freed batch charges while reserving expression/scatter/read scratch can turn a recoverable spill into another false admission failure.

Please fix this at the ownership boundary: provide a HashmapBuilder API that drains physical batches and their reservations together (respecting CopyIntoBatches coalescing, so releasing all ingress tokens after the last physical batch is fine), and use it in both paths instead of manipulating Batches directly. Add budgeted tests proving that NeedBatches=false drops the batch charge before probe admission and that re-spill drops drained-batch charges before subsequent scratch/read admission.

The current Ubuntu UT failure is TestDelete3 in untouched TAE code; it does not appear related to this diff, but the check is still red.

aptend:
Addressed the HashBuild budget ownership blocker in `7088293ab`.

The fix introduces one ownership boundary, `HashmapBuilder.DrainCopiedBatches`:

- a successful visit destroys each physical build batch and releases all coalesced ingress reservations after the final physical batch;
- a failed visit retains the current/remaining physical batches and all conservative reservations for the existing builder cleanup path;
- the implementation is a single linear pass, avoiding repeated slice shifts/recalculation on large re-spills.

Both affected SpillEngine paths now use this boundary:

1. `RebuildHashmap` with `NeedBatches=false` drains batch memory and batch reservations before `GetJoinMap`, so the JoinMap no longer inherits phantom batch charges during probe;
2. `reSpillBucket` drains each accumulated batch after successful scatter and releases the coalesced reservations before subsequent pending/read admission.

Added budgeted regressions covering:

- coalesced ingress reservation release after the last physical batch;
- partial visitor failure retaining cleanup ownership;
- `NeedBatches=false` carrying less live budget than `NeedBatches=true` while entering probe;
- direct re-spill returning without a stale builder batch charge, followed by successful full-cap admission.

Fresh local validation:

- focused regressions, `-count=10`: pass;
- focused `-race`: pass;
- full `hashbuild` and `spillutil` package tests: pass;
- package build and vet: pass.

The unrelated `TestDelete3` failure is being handled by the existing separate PR, as noted.

XuPeng-SH (approved):
我重新做了一轮 deep review，并且把作者在评论里对 trusted-manifest / trusted-provider 边界的说明也纳入考虑了。当前结论是：在这个明确声明的管理员级 fixture/restore 契约下，`LOAD TABLE` 不回读物理 metadata、provider-side copy 只做 size 校验，是有意识的性能/信任边界选择，不再构成我这边的 correctness blocker。

除此之外，我又把 HashBuild budget / spill / remote pipeline 终态发布主链过了一遍，没有再确认出新的高置信 correctness、robustness 或 unhappy-path 问题。LGTM.

Issue #23073, opened 2025-11-24
Feature Not Supported: Nested aggregate functions in SELECT expressions


dengn:
## Description

MatrixOne does not support nested aggregate functions in SELECT expressions, specifically when using aggregate functions (like AVG, SUM, COUNT) with correlated subqueries that contain aggregate functions. The error indicates that the expression executor for this type of nested aggregation is not yet implemented.

## Error Message

```
ERROR 20102 (HY000) at line 1: unsupported expression executor for typ:<id:23 notNullable:true > sub:<node_id:5 row_size:1 >  now is not yet implemented
```

## Affected Cases

This issue affects **4 test cases** from our MySQL compatibility testing on the `industry_radar_test` database.

### Table: company

```sql
CREATE TABLE `company` (
  `id` int NOT NULL,
  `full_name` varchar(200) DEFAULT NULL,
  `short_name` varchar(128) DEFAULT NULL,
  `update_time` datetime NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp() COMMENT '最后修改时间',
  `address` varchar(255) DEFAULT NULL,
  `board_id` int DEFAULT NULL,
  `business_scope` text DEFAULT NULL,
  `city` varchar(32) DEFAULT NULL,
  `company_profile` text DEFAULT NULL,
  `company_type_tags` varchar(255) DEFAULT NULL,
  `country` varchar(20) DEFAULT NULL,
  `directory_label` varchar(512) DEFAULT NULL,
  `district` varchar(32) DEFAULT NULL,
  `email` varchar(255) DEFAULT NULL,
  `english_name` varchar(255) DEFAULT NULL,
  `enterprises_type` varchar(50) DEFAULT NULL,
  `founded_time` datetime(6) DEFAULT NULL,
  `industry` varchar(32) DEFAULT NULL,
  `industry_paqu` varchar(1000) DEFAULT NULL,
  `insured_num` varchar(32) DEFAULT NULL,
  `product_label` varchar(512) DEFAULT NULL,
  `province` varchar(32) DEFAULT NULL,
  `registered_capital` varchar(50) DEFAULT NULL,
  `representative` varchar(32) DEFAULT NULL,
  `taxpayer_num` varchar(50) DEFAULT NULL,
  `telephone` varchar(255) DEFAULT NULL,
  `website` varchar(500) DEFAULT NULL,
  `industry_web_label` varchar(512) DEFAULT NULL COMMENT '国标行业标签',
[...]

Issue #23108, opened 2025-11-24
Feature Not Supported: Nested aggregate functions in HAVING clause expressions not supported


dengn:
## Description

MatrixOne does not support nested aggregate functions in HAVING clause expressions. When attempting to use nested aggregates (e.g., `AVG((SELECT COUNT(*) ...))`) in HAVING clauses, MatrixOne returns an error indicating that the expression executor is not yet implemented.

## Error Message

```
ERROR 20102 (HY000): unsupported expression executor for typ:<id:23 notNullable:true > sub:<node_id:3 row_size:1 >  now is not yet implemented
```

## Affected Cases

This issue affects **2 test cases** from our boundary testing on the `industry_radar_test` database.

### Example 1: HAVING with Nested Aggregate in Subquery

```sql
SELECT province, COUNT(*) FROM company 
GROUP BY province 
HAVING COUNT(*) > AVG((SELECT COUNT(*) FROM company GROUP BY province)) 
LIMIT 10;
```

**Error**: `ERROR 20102 (HY000): unsupported expression executor for typ:<id:23 notNullable:true > sub:<node_id:3 row_size:1 >  now is not yet implemented`

### Example 2: HAVING with Subquery Aggregate

```sql
SELECT province, COUNT(*) FROM company 
GROUP BY province 
HAVING COUNT(*) > (SELECT COUNT(*) FROM company) 
LIMIT 10;
```

**Error**: `ERROR 20102 (HY000): unsupported expression executor for typ:<id:23 notNullable:true > sub:<node_id:3 row_size:1 >  now is not yet implemented`

## Expected Behavior

In MySQL, HAVING clauses can contain subqueries with aggregate functions. The subquery is evaluated and the result is used in the HAVING condition.

Example MySQL behavior:
```sql
-- MySQL: Fully supported
SELECT province, COUNT(*) 
FROM company 
GROUP BY province 
HAVING COUNT(*) > (SELECT AVG(cnt) FROM (SELECT COUNT(*) as cnt FROM company GROUP BY province) as subq);
```

## Impact

- **Functionality**: Complex HAVING conditions with subqueries are not supported
- **Query Flexibility**: Users cannot write advanced filtering conditions in HAVING clauses
- **Migration**: Applications using nested aggregates in HAVING will fail when migrating to MatrixOne

## Use Cases

HAVING with nested aggregates is used for:
1. **Comparative Aggregation**: Comparing group aggregates against overall aggregates
[...]

PR #25838, opened 2026-07-17
fix(plan): support scalar subqueries in aggregate arguments

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23073
issue #23076

Aggregate arguments such as `AVG((SELECT COUNT(*) ...))` retained an executable `Expr_Sub` in `AGG.AggList`. The expression executor cannot evaluate that node and returned `unsupported expression executor`.

Changes:
- Treat correlated references from a scalar subquery inside an outer aggregate as aggregate inputs during `ONLY_FULL_GROUP_BY` validation.
- Flatten scalar subqueries in aggregate expressions before constructing the `AGG` node.
- Flatten aggregate subqueries before the ordered-aggregate Sort, so a JOIN (including its spill path) cannot invalidate `GROUP_CONCAT(... ORDER BY ...)` input order.
- Recursively reject direct, positional, and wrapped subquery expressions in `GROUP_CONCAT(... ORDER BY ...)` during binding, before an `Expr_Sub` can reach the Sort executor.
- Add planner regression coverage for plain SELECT, grouped SELECT, CTE forms, the required `AGG -> SORT -> JOIN` plan shape, subquery-free Sort expressions, and direct/positional/wrapped ordered-subquery rejection.
- Add BVT coverage for `AVG`, `SUM`, ordered `GROUP_CONCAT`, direct/positional/wrapped ordered-subquery rejection, and a correlated `COUNT(*)` result of zero.

Non-goal:
- Subqueries in HAVING expressions tracked separately by #23108 are not changed here.

Validation:
- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `git diff --check`
- Built `cmd/mo-service` and started an isolated local cluster. The nested aggregate SELECT, ordered `GROUP_CONCAT`, and CTE results matched their checked-in results; direct, positional, wrapped-positional, and wrapped-explicit `GROUP_CONCAT ORDER BY` subqueries returned `not supported: subquery in group_concat ORDER BY` during binding.


heni02 (approved):

gouhongshen (approved):
## Codex automated review

No blocking or non-blocking regressions found in the PR diff.

XuPeng-SH (changes requested):
Request changes: the new flattening invariant is not closed for ordered aggregate inputs.

A reproducible case is:

```sql
SELECT n.N_REGIONKEY,
       GROUP_CONCAT(
           (SELECT r.R_NAME
              FROM REGION r
             WHERE r.R_REGIONKEY = n.N_NATIONKEY)
           ORDER BY 1)
  FROM NATION n
 GROUP BY n.N_REGIONKEY;
```

`processForceWindows` resolves `ORDER BY 1` to the aggregate argument and binds a separate copy into `ctx.groupConcatOrderBys`. The new loop flattens only `ctx.aggregates`, so the resulting plan has a subquery-free `AGG.AggList` but still contains an executable `Expr_Sub` in `SORT.OrderBy`. I confirmed this with a planner test against the PR head. `mergeorder.Prepare` then passes that expression to `colexec.NewExpressionExecutor`, which returns the same unsupported-expression error this PR is intended to eliminate.

Please either flatten/canonicalize the positional ORDER BY expression before creating the Sort, or reject this form during binding (consistently with an explicit subquery in GROUP_CONCAT ORDER BY). Add a regression that checks every executor-bound expression introduced by this path, not only `AGG.AggList`.

I also confirmed the main AVG/SUM/CTE cases pass, and `pkg/sql/plan` tests plus vet pass after a clean merge with the latest main.

VioletQwQ-0:
@XuPeng-SH Addressed the ordered-aggregate change request in `e4e37bee3`.

Changes:
- Validate the resolved `orderExpr` after positional `ORDER BY` expansion, so `GROUP_CONCAT((SELECT ...) ORDER BY 1)` is rejected consistently with an explicit subquery in `GROUP_CONCAT ORDER BY`.
- Added a planner regression using the reported SQL and requiring `subquery in group_concat ORDER BY`.
- Added matching BVT SQL/result coverage.

Validation:
- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `git diff --check`
- Built `cmd/mo-service` and ran the reported shape against an isolated local cluster; it now returns `ERROR 20105: not supported: subquery in group_concat ORDER BY` before expression executor preparation.

Re-requesting review on the updated head.

gouhongshen (changes requested):
## Codex automated review

An ordered GROUP_CONCAT path still permits an executable scalar-subquery expression to reach SORT and fail at runtime.

### P1 - Reject or flatten scalar subqueries nested inside positional GROUP_CONCAT ORDER BY expressions (pkg/sql/plan/having_binder.go:369)
The new check only recognizes orderExpr when its outer AST node is *tree.Subquery. For example, GROUP_CONCAT(COALESCE((SELECT r.R_NAME FROM REGION r WHERE r.R_REGIONKEY = n.N_NATIONKEY), '') ORDER BY 1) resolves 1 to a FuncExpr, bypasses this check, and processForceWindows stores a separately bound Expr_F containing Expr_Sub in ctx.groupConcatOrderBys. The new flattening loop only rewrites ctx.aggregates, then query_builder.go appends the untouched order specs to the SORT node. SORT prepares an expression executor for that expression, but colexec.NewExpressionExecutor has no Expr_Sub implementation, so this still fails with the unsupported-expression error. Recursively detect/reject subqueries after ordinal expansion, or flatten/canonicalize the stored ordering expression, and add coverage for a wrapped scalar subquery.

XuPeng-SH (changes requested):
The latest fix still allows a nested scalar subquery to reach an executor-bound GROUP_CONCAT sort expression.

`having_binder.go:369` only rejects `orderExpr` when the top-level AST node is `*tree.Subquery`. A wrapped expression bypasses that check, for example:

```sql
SELECT n.N_REGIONKEY,
       GROUP_CONCAT(
         COALESCE((SELECT r.R_NAME
                     FROM REGION r
                    WHERE r.R_REGIONKEY = n.N_NATIONKEY), "")
         ORDER BY 1)
  FROM NATION n
 GROUP BY n.N_REGIONKEY;
```

For this positional ORDER BY, the resolved AST node is a `*tree.FuncExpr`, so it is accepted. The aggregate-argument copy is flattened by the new loop, but the independently bound `ctx.groupConcatOrderBys` copy still contains `Expr_Sub` and is attached to `SORT.OrderBy`. The sort expression executor cannot execute that node.

I reproduced this deterministically on current head `e4e37bee3`: planning succeeds, but a plan invariant test finds an executable `Expr_Sub` under `SORT.OrderBy`.

Please reject subqueries recursively on the exact bound expression after `BindExpr` (the existing `hasSubquery` already traverses function/list children), or use equivalent recursive AST validation, before appending the order specification. Add regression coverage for both explicit and positional wrapped forms, and assert that every executor-bound `SORT.OrderBy` expression is subquery-free.

VioletQwQ-0:
@gouhongshen @XuPeng-SH Addressed the wrapped ordered-subquery change requests in `bb85ad8d4`.

Changes:
- After binding each resolved `GROUP_CONCAT ORDER BY` expression, recursively call the existing `hasSubquery` helper before appending the order spec. This catches `Expr_Sub` nested under function/list expressions, not only a top-level AST subquery.
- Added planner regressions for direct positional, wrapped positional, and wrapped explicit forms.
- Added a plan invariant assertion that every executor-bound `SORT.OrderBy` expression is subquery-free.
- Added matching BVT SQL/result coverage for both wrapped forms.

Validation:
- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan`
- `git diff --check`
- Built `cmd/mo-service` and ran both reported wrapped forms against an isolated local cluster; each now returns `ERROR 20105: not supported: subquery in group_concat ORDER BY` during binding.

Re-requesting review from both reviewers on the updated head.

gouhongshen (approved):
## Codex automated review

No concrete regressions found in the PR diff.

aunjgr (approved):
Re-reviewed current head. Aggregate argument subqueries are flattened before ordered-aggregate sorting, and GROUP_CONCAT ORDER BY now recursively rejects wrapped and positional subqueries. The previous nested-subquery blocker is closed.

aunjgr (approved):
Re-reviewed current head. Aggregate argument subqueries are flattened before ordered-aggregate sorting, and GROUP_CONCAT ORDER BY now recursively rejects wrapped and positional subqueries. The previous nested-subquery blocker is closed.

XuPeng-SH (approved):

Issue #25839, opened 2026-07-17
[Performance]: retire Span recording to eliminate IsMOCtledSpan contention


LeftHandCold:
## Summary

MO no longer consumes internal Span records, but the old request path still entered the Span implementation and queried a global controlled-span map protected by a mutex. This issue tracks retiring that unused recording path while preserving the observability data that MO still depends on.

## Correction to the original profile conclusion

The flame graphs and the summary image do **not** prove that "trace consumes 60% CPU" or that trace is the main performance bottleneck.

The reported `sync.Mutex.Unlock` / `IsMOCtledSpan` percentages are attribution within a **mutex blocking-delay profile**. They mean that this lock accounted for a large share of sampled waiting time in those captures; they are not CPU percentages and cannot be added to frontend percentages as total CPU usage.

The supported conclusion is narrower: `IsMOCtledSpan` was a real, avoidable concurrency hotspot in the tested high-concurrency point-select/read-only/insert workloads. Removing it should reduce contention, but SQL execution, storage reads, compilation/planning, transaction work, syscalls, and GC remain the main workload-dependent costs.

## Final decision

Retire the Span recording implementation, not the whole trace/observability subsystem.

Removed by PR #25841:

- `MOTracer`, `MOSpan`, `MOHungSpan`
- Span batch processor/exporter and Span-only buffering
- controlled-span global maps/mutex
- Span-triggered timers and profile capture
- Span-specific tests
- runtime ability to re-enable Span through configuration

Preserved:

- `StatementInfo`, statement status/error information, and CU accounting
- structured/raw log export
- error export
- `SpanContext`, context helpers, and RPC wire layout
- trace/span ID fields in retained log/error records
- existing `span_info` schema/view for upgrade compatibility
- shared collector lifecycle and graceful flush

`StatementInfo` does not depend on a recorded `MOSpan`; it has its own statement, transaction, session, status, error, and CU fields. Therefore it is preserved directly rather than filled with artificial Span values.

## Compatibility

- Code default remains `disable-span=true`.
- Explicit `disable-span=false` is still accepted but cannot re-enable the retired implementation.
- `enable-span-profile`, `long-span-time`, and `enable-trace-debug` remain accepted compatibility no-ops.
- CN/TN TraceSpan control calls return `span tracing is no longer supported`.
- No new `span_info` rows are created.
- No schema migration is required; rollback is a binary rollback.

## Evidence

Controlled ABBA test on the same machine/configuration/data/workload:

[...]

PR #25841, opened 2026-07-17
refactor(trace): retire span recording pipeline

LeftHandCold:
## What type of PR is this?

- [x] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

issue #25839

MO no longer consumes internal Span records, but every hot-path `trace.Start` still entered the Span recording implementation and queried global controlled-span state. Under concurrency, that query appeared prominently in mutex-delay profiles.

This PR retires the Span recording pipeline while preserving the observability data and trace-context contracts that MO still uses:

- removes `MOTracer`, `MOSpan`, `MOHungSpan`, the Span batch processor/exporter, controlled-span maps, timers, profile hooks, and Span-only tests;
- defaults `disable-span` to `true` in code, independent of deployment configuration files;
- keeps the fully disabled path on the zero-allocation `NoopTracer`;
- uses a lightweight `NonRecordingTracer` when observability is enabled, generating and propagating TraceID/SpanID without recording, profiling, queuing, or exporting Spans;
- makes `mo_ctl` TraceSpan requests return the deterministic response `span tracing is no longer supported`;
- retains legacy Span configuration fields and public APIs as parse/source-compatible no-ops.

This is deliberately **not** a deletion of MO observability:

- `StatementInfo` export, statement status/error fields, and CU accounting are unchanged;
- structured/raw logs and `MOErrorHolder` export remain enabled and retain per-request TraceID/SpanID correlation;
- `SpanContext`, context helpers, and the 24-byte MORPC wire layout are retained;
- existing `span_info` schema/view definitions remain for upgrade compatibility, but no new Span rows are recorded;
- the shared statement/log/error collector is still stopped and flushed during service shutdown.

### Preserved non-recording behavior

Three behaviors were independent of Span export and are preserved explicitly:

1. `WithHungThreshold` previously introduced real context deadlines. The 1-minute storage-usage query deadline and 30-minute frontend cleanup guard are now explicit `context.WithTimeoutCause` scopes, with the original `moerr.CauseNewMOHungSpan` cause and former cancellation lifetime.
2. `trace.Generate` previously created per-connection IDs. The non-recording tracer preserves root/child/new-root ID semantics and context propagation while ordinary `NoopTracer.Start` remains zero-allocation.
3. The four former `mo_ctl` controlled diagnostic kinds (`statement`, `remoteFSOperation`, `localFSOperation`, and `tnRPCHandle`) were disabled by default. They remain default-off and preserve the parent context; the removed runtime enable/disable state, mutex, and map are not restored.

Regression tests prove that enabled `Generate` returns non-zero, distinct root IDs; the context reaches frontend routines, structured log rows, and error rows; MORPC encode/decode preserves the IDs; formerly controlled diagnostic starts do not overwrite the parent log context; and the fully disabled/ordinary no-op paths remain no-op.

### Compatibility and behavior change

This intentionally removes the diagnostic Span opt-in. `disable-span=false`, `enable-span-profile`, `long-span-time`, and `enable-trace-debug` are still accepted to avoid configuration breakage, but cannot restore Span recording or Span-triggered profiles.

`StatementInfo` retains its existing statement, transaction, session, error, and CU fields. No fake Span rows or placeholder IDs are injected. Logs and errors use real non-recording correlation IDs generated for their request context.

### BVT regression closure

The failed `Matrixone Utils CI / Coverage` job on head `cb5e3217ce` was caused by this PR, not by an unrelated `main` change. The initial `NonRecordingTracer` applied every requested kind, so a formerly default-off diagnostic start could change a log row's `span_kind` to `localFSOperation`, `remoteFSOperation`, or `tnRPCHandle`. `MOZapLog.FillRow` persists that kind verbatim, while `zz_statement_query_type/query_stmt.sql` accepts only `internal`, `statement`, `session`, or `remote`, producing the observed `expected 1, actual 0`.

Commit `6dc3383ce0` restores the old default behavior with a private static four-kind filter. It does not broaden the BVT allowlist, restore controlled-span shared state, or affect StatementInfo/CU/log/error collection. A regression test follows a controlled LocalFS start through `ReportZap` and raw-log row filling and verifies that the parent `internal` kind is retained.

### Performance evidence

A controlled ABBA comparison on the same machine, configuration, data, workload, and concurrency (128 clients, 10 measured iterations per side) measured:

- baseline: 79,393 QPS
- Span-disabled candidate: 81,463 QPS
- delta: +2.61%
- baseline `IsMOCtledSpan` mutex-delay attribution: 43.77% / 40.58%
- candidate exact `IsMOCtledSpan` focus: no samples

These percentages are **mutex blocking-delay attribution, not CPU percentages**. The result proves removal of that contention path; it does not show that Span was the primary CPU bottleneck or guarantee a 2.61% gain for every workload. Physical deletion has the same no-recording hot-path intent as the measured disabled candidate, so no additional performance gain is claimed for deletion itself. The lightweight context generator adds only ID/context propagation and does not restore the removed locks or exporter pipeline.

### Validation

Validated at final head `6dc3383ce0`:

- focused regression tests for non-recording ID propagation and controlled-kind raw-log compatibility;
- normal CGo-aware tests: `./pkg/util/trace/...`, `./pkg/fileservice`, `./pkg/frontend`, `./pkg/sql/compile`, `./pkg/common/log`, and `./pkg/common/morpc`;
- race tests: `./pkg/util/trace/...`, `./pkg/fileservice`, `./pkg/frontend`, and `./pkg/common/morpc`;
- CGo-aware `go vet` for the affected dependency chain;
- `make build`;
- `git diff --check`;
- first-principles correctness, compatibility, lifecycle, wait-for, and Q1-Q3 unhappy-path review;
- built binary contains `NonRecordingTracer`/ID-generation symbols but no executable `MOTracer`, `MOSpan`, `MOHungSpan`, controlled-span-state, or Span batch-processor symbols.

Machine `10.222.1.50` validation of the retired recording runtime:

- rebuilt and started with an explicit `disable-span=false` configuration; Span recording still could not be re-enabled;
- statement count increased and successful/failed marker statements retained status, error code, and CU;
- raw-log and error counts increased; the injected SQL error appeared in `system.error_info`;
- `span_info` remained unchanged at 3 -> 3;
- CN and TN TraceSpan control calls returned `span tracing is no longer supported`;
- 1,000,000 point selects at concurrency 128 completed successfully;
- mutex-profile exact focus for `IsMOCtledSpan`, `MOTracer`, `MOSpan`, and `MOHungSpan`: no samples;
- graceful shutdown flushed the last statement buffer and logged `trace.shutdown.complete`;
- the pre-existing system MO process was not modified.

The final diff is limited to the trace runtime/configuration, its TraceSpan compatibility endpoint, explicit preservation of the two former hung-context deadlines, non-recording trace-context correlation, and preservation of the former default-off controlled kinds. SQL execution, storage, fileservice, `StatementInfo`, and CU calculation semantics are not changed.


LeftHandCold:
Final scope expansion and self-review completed at `62c1c608ba`.

The PR now retires the Span recording implementation instead of only disabling it by default. Review-relevant boundaries:

- StatementInfo, status/error fields, CU, logs, and error export remain active.
- SpanContext/public helpers/RPC layout and `span_info` schema remain compatible.
- legacy Span configuration parses but is a no-op; TraceSpan control returns an explicit unsupported response.
- shared telemetry collector shutdown/flush ownership was migrated out of the deleted Span processor and is covered by normal/canceled-context tests.
- no SQL execution, frontend request, storage, fileservice, or CU calculation semantics changed.

Final validation passed locally (race, focused/dependent tests, vet, full build) and on machine 50 with explicit `disable-span=false`. The 50-machine run retained Statement/log/error/CU records, created no new Span rows, completed a 1,000,000-query concurrency-128 workload, showed no `IsMOCtledSpan`/Span implementation samples in the mutex profile, and shut down with a successful final telemetry flush.

Detailed final implementation, all three validation rounds, and before/after evidence: https://github.com/matrixorigin/matrixone/issues/25839#issuecomment-5001848388

gouhongshen (changes requested):
## Blocking finding

`WithHungThreshold` is not only a Span-recording option; it currently provides a real cancellation deadline. Before this PR, `MOTracer.Start` applied the option and `newMOHungSpan` wrapped the returned context with `context.WithTimeoutCause(..., threshold, moerr.CauseNewMOHungSpan)`.

There are production call sites that consume that returned context:

- `pkg/util/metric/mometric/cron_task.go:253-255` passes it to `sqlExecutor().Query`, relying on the 1-minute bound.
- `pkg/frontend/routine.go:268` later checks `routineCtx.Done()` at line 345, relying on the 30-minute hung-request guard.

After this change, `NoopTracer.Start` returns the original context and never applies `WithHungThreshold`, so both deadlines disappear. A hung storage-size query/request can therefore outlive the previous bound and change shutdown/cancellation behavior. Please preserve this timeout semantics with an explicit context-timeout helper or at the call sites before making the Span runtime a no-op.

Other reviewed areas (retained StatementInfo/log/error registration, schema compatibility, CN/TN TraceSpan response path, and collector shutdown ownership) did not reveal an additional blocker.

Validation: `go test -race ./pkg/util/trace -count=1` and `go vet ./pkg/util/trace` pass. The broader affected-package test command was blocked by missing local CGo headers (`xxhash.h`, `roaring.h`, `usearch.h`).

fengttt (approved):

LeftHandCold:
@gouhongshen This blocking finding is valid and has been fixed in `513a9fab8f`.

`WithHungThreshold` mixed Span diagnostics with a real context deadline. I audited every production use and separated the business timeout semantics from the retired Span runtime:

- `pkg/util/metric/mometric/cron_task.go`: the storage-usage query now receives an explicit 1-minute `context.WithTimeoutCause` context.
- `pkg/frontend/routine.go`: the existing `routineCtx.Done()` cleanup guard now receives an explicit 30-minute `context.WithTimeoutCause` context.
- Both preserve `moerr.CauseNewMOHungSpan` and use `defer cancel()` at the same lifetime boundary where `MOHungSpan.End()` previously canceled its context.
- The third production use in DDL execution discarded the context and existed only to trigger Span profiles. That now-no-op block, plus its unreachable system-status profile throttling state, was removed.

I also re-audited all Span start/end options and `trace.IsEnable` consumers. No other option has a non-recording business side effect: the remaining options only enriched exported Span fields or triggered the retired profiles/backoff. StatementInfo/log/error/CU collection is independent and remains enabled.

Final head is rebased on current `main@1a14737560`. Fresh validation after the final edit:

- normal and race tests passed for `./pkg/util/trace/...`, `./pkg/util/metric/mometric`, and `./pkg/frontend`
- config, CTL, CN, ETL, mo-service, log, and morpc tests passed
- TAE storage, TAE RPC, and fileservice dependent test binaries compiled
- CGo-aware vet passed for all affected packages
- `make build` passed
- full correctness/compatibility/lifecycle and Q1-Q3 unhappy-path review found no remaining blocker

Please re-review the updated head.

gouhongshen (changes requested):
## Codex automated review

Retiring span export also removes the non-recording trace context that remaining log/error telemetry relies on.

### P1 - Preserve trace context generation when spans are non-recording (pkg/util/trace/noop_trace.go:48)
`motrace.Init` now installs `NoopTracer`, whose `Start` returns the original context with an empty `NoopSpan`. Consequently `trace.Generate` no longer generates a root TraceID. `frontend.NewRoutine` calls `trace.Generate` specifically to attach a per-connection trace ID, so request contexts now remain empty. Session/process log helpers omit the context field when it is empty, and `ReportZap` falls back to the node-only default context; error rows likewise get zero IDs. Thus log/error collection remains enabled but can no longer be correlated to a client request or propagated over RPC. Retire recording/export only; `Start`/`Generate` still need to install a non-recording SpanContext with generated IDs.

XuPeng-SH (changes requested):
Re-reviewed the complete diff at `513a9fab8f`. The explicit 1-minute metric-query deadline and 30-minute frontend cleanup guard correctly fix the earlier `WithHungThreshold` regression. Statement/log/error collector registration and shutdown ownership also look closed.

One independent blocker remains:

### Preserve non-recording IDs for explicit `trace.Generate`

`motrace.Init` permanently installs `NoopTracer`, while `trace.Generate` delegates to `DefaultTracer().Start(..., WithNewRoot(true))`. Since `NoopTracer.Start` returns the input context unchanged, `Generate` now returns a context with no generated TraceID/SpanID.

This is observable production behavior, not Span export:

- `frontend.NewRoutine` explicitly calls `trace.Generate` to attach a per-connection trace ID, then stores that context in the protocol/routine.
- session/process logging only adds the trace context when it is non-empty;
- error reporting reads the same context IDs;
- morpc serializes and propagates the existing `SpanContext`.

After this PR, locally-created connection contexts therefore lose request correlation even though log/error telemetry remains enabled. Falling back to the node-only default context does not preserve per-connection identity.

This does **not** require restoring the Span recording pipeline or its hot-path locks. Keep ordinary `NoopTracer.Start` zero-allocation, but preserve the explicit `Generate` contract with a cheap non-recording root context generator (respecting the intended enabled/disabled configuration semantics). Please add tests that prove:

1. enabled `Generate` installs non-zero IDs;
2. independent roots get distinct IDs;
3. the generated context survives log/error extraction and morpc encode/decode;
4. the disabled path and ordinary `NoopTracer.Start` retain their intended no-op behavior.

Validation performed on this head:

- CGo-aware `-race` tests: `./pkg/util/trace/...` — pass
- CGo-aware tests: `./pkg/frontend` — pass
- CGo-aware trace tests in the broader matrix — pass
- full `mometric` package could not run locally because an existing Docker listener owns `:7001`; CI for this head is green

Requesting changes until explicit non-recording correlation is preserved.

LeftHandCold:
@gouhongshen @XuPeng-SH The P1 finding is valid and fixed in `7748b5706e`.

The fix preserves correlation without restoring the retired Span pipeline:

- when tracing is fully disabled, the existing `NoopTracer` remains zero-allocation and `Start`/`Generate` remain no-op;
- when observability is enabled, `NonRecordingTracer.Start` creates root/child/new-root TraceID/SpanID contexts, but does not create recorded Span data, profile, enqueue, export, or consult controlled-span state;
- `frontend.NewRoutine -> trace.Generate` again installs a non-empty, per-connection context;
- structured log extraction, error-row extraction, and MORPC encode/decode all preserve the generated IDs.

Added regression coverage for the four requested boundaries:

1. enabled `Generate` creates non-zero IDs;
2. independent roots receive distinct TraceID and SpanID values;
3. generated IDs survive frontend routine setup, log/error extraction, and MORPC encoding/decoding;
4. fully disabled providers and ordinary `NoopTracer.Start` retain no-op behavior.

The branch was rebased onto `main@2d01285f8b` and revalidated:

- normal CGo-aware tests passed for all affected trace, metric, frontend, config, control, service, exporter, log, and MORPC packages;
- race tests passed for `./pkg/util/trace/...`, `./pkg/util/metric/mometric`, `./pkg/frontend`, and `./pkg/common/morpc`;
- CGo-aware `go vet`, `make build`, and `git diff --check` passed;
- the built binary contains the lightweight context generator but no executable `MOTracer`, `MOSpan`, `MOHungSpan`, controlled-span-state, or Span batch-processor symbols.

I also re-reviewed the full diff for retained StatementInfo/log/error collection, collector shutdown ownership, deadline/cancellation behavior, wire/schema compatibility, and resource/wait lifecycle. No additional blocker was found. Please re-review the updated head.

XuPeng-SH (approved):
Re-reviewed the complete diff at `7748b5706e`. The previous blockers are closed:

- the former `WithHungThreshold` business side effects are now explicit 1-minute and 30-minute contexts at the two real consumers; the DDL-only profiling call did not propagate its returned context and is correctly removed;
- enabled observability now uses a stateless non-recording tracer that preserves root/child/new-root IDs, frontend connection correlation, log/error extraction, and MORPC propagation; fully disabled tracing and `NoopTracer.Start` remain no-op/zero-allocation;
- Span recording, profile hooks, controlled-span maps/locks, queue/export paths, and runtime re-enable behavior remain retired; StatementInfo, log/error export, CU, schema, and wire compatibility remain intact;
- shared collector shutdown has explicit ownership and a bounded cancellation path.

Performance was checked independently because there are 73 production `trace.Start` call sites, including SQL/fileservice hot paths. On the same machine, a focused latest-head non-recording Start+End benchmark measured about 55 ns/op, 256 B/op, 3 allocs/op versus about 178 ns/op, 384 B/op, 3 allocs/op for the main-branch `MOTracer` Start+End path. The new path therefore does not increase allocation count and removes the global controlled-span mutex and recording/export work.

Fresh reviewer validation passed:

- CGo-aware race tests: `./pkg/util/trace/...`
- CGo-aware race test: frontend generated routine context
- CGo-aware race test: MORPC generated-context codec round trip
- `git diff --check`

GitHub Ubuntu UT, arm64 SCA, pessimistic multi-CN BVT, proxy BVT, and standalone pessimistic BVT are green; coverage is still running and should remain a merge gate. No remaining code blocker found.

gouhongshen (approved):
## Codex automated review

No correctness regressions found in the PR diff. Verified trace-context propagation, collector shutdown ownership, retained telemetry registration, timeout preservation, and compatibility paths. `pkg/util/trace` tests, vet, and build passed; CGo-dependent focused packages could not run because this checkout lacks `cgo/libmo.dylib` and third-party headers.

LeftHandCold:
### BVT failure: root cause confirmed and fixed

The failure in `zz_statement_query_type/query_stmt.sql` is caused by this PR; it is not an unrelated `main` failure.

Evidence chain:

1. On `main`, the four `mo_ctl` diagnostic span kinds (`statement`, `remoteFSOperation`, `localFSOperation`, and `tnRPCHandle`) are all disabled by default. Their starts return a no-op span without replacing the parent context.
2. The initial `NonRecordingTracer` removed the old controlled-state check but applied every requested kind. Therefore a formerly disabled LocalFS/RemoteFS/TN-RPC start could replace the parent kind.
3. `MOZapLog.FillRow` writes `SpanContext.Kind.String()` directly to `system.rawlog.span_kind`.
4. The failing BVT accepts only `internal`, `statement`, `session`, or `remote`; a leaked diagnostic kind makes its expression return the observed `0`.

Fixed in `6dc3383ce0` by keeping exactly those four retired controlled kinds default-off with a private static filter. This preserves the pre-PR default semantics without restoring the controlled-span mutex/map/runtime switch and without changing the BVT allowlist.

Regression coverage now verifies the complete failing path: an `internal` root context -> controlled LocalFS start -> `ReportZap` -> `MOZapLog.FillRow`, with the raw-log kind remaining `internal`. StatementInfo, CU, logs, errors, and non-recording trace correlation are unchanged.

Validation after the fix:

- focused regression tests passed;
- normal CGo-aware tests passed for trace, fileservice, frontend, SQL compile, log, and MORPC;
- race tests passed for trace, fileservice, frontend, and MORPC;
- affected-package `go vet`, `make build`, `git diff --check`, and the complete lifecycle/unhappy-path review passed.

Fresh CI is running at https://github.com/matrixorigin/matrixone/actions/runs/29671757688; the same `Matrixone Utils CI / Coverage` job is included.

Issue #25840, opened 2026-07-17
[Bug]: TestDiskCacheReadEnsuresMemoryCacheCapacity leaks CachedData and flakes unrelated UTs


LeftHandCold:
## Summary

`TestDiskCacheReadEnsuresMemoryCacheCapacity` leaves an allocator-backed `CachedData` object unreleased. The checked allocator reports the leak asynchronously from its finalizer, so a later and unrelated test such as `TestDiskCacheQuotaExceeded` can be reported as the failing test.

This occurred in the Ubuntu/x86 UT job for PR #25834: https://github.com/matrixorigin/matrixone/actions/runs/29564751079/job/87834783504?pr=25834

## Root cause

`pkg/fileservice/disk_cache_test.go:1121` allocates a 10-byte `fscache.Data` through `AllocateCacheDataWithHint`. `DiskCache.Read` stores that object in `vec.Entries[0].CachedData` and transfers result ownership to the caller's `IOVector`.

The caller-side release contract is `IOVector.Release()`, which releases both `IOEntry.CachedData` and any raw read buffer. `TestDiskCacheReadEnsuresMemoryCacheCapacity` verifies that `CachedData` was populated but returns without calling `vec.Release()`.

The checked allocator attaches a finalizer to sampled allocations. When GC later collects the abandoned deallocator, the finalizer panics. This explains why the job labels `TestDiskCacheQuotaExceeded` as failed even though the allocation stack ends in `TestDiskCacheReadEnsuresMemoryCacheCapacity`.

The missing release was introduced with the test in #24387 (`bdb848db22b468b98deedef227b948bbace68ffa`). This is a test cleanup defect; the production `DiskCache.Read` ownership semantics are working as designed.

## Impact

Confirmed impact:

- FileService UT is flaky under the default sampled checked allocator.
- The asynchronous finalizer can blame whichever test happens to run when GC occurs, obscuring the real source and causing unrelated PR CI failures.

No production behavior or cache ownership semantic is affected by the proposed fix.

## Reproduction / evidence

On current `main`, run:

```shell
MO_MALLOC_DEBUG=1 GOFLAGS=-mod=mod go test ./pkg/fileservice \
  -run '^TestDiskCache(ReadEnsuresMemoryCacheCapacity|QuotaExceeded)$' -count=100
```

[...]

PR #25842, opened 2026-07-17
test(fileservice): release disk cache read result

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25840

`TestDiskCacheReadEnsuresMemoryCacheCapacity` allocates a 10-byte allocator-backed `CachedData` result through `DiskCache.Read`, but did not release the caller-owned `IOVector`. The checked allocator detects that missing free from a GC finalizer, which can asynchronously fail a later unrelated test such as `TestDiskCacheQuotaExceeded`.

This PR registers `vec.Release()` immediately after constructing the read vector. That closes the existing caller-side ownership lifecycle on success, read errors, and assertion failures.

Scope is intentionally limited to test cleanup. It does not change `DiskCache.Read`, allocator behavior, cache capacity accounting, or production FileService semantics. This is therefore kept separate from the proxy changes in #25834.

Tested with:

```shell
MO_MALLOC_DEBUG=1 GOFLAGS=-mod=mod go test ./pkg/fileservice -run '^TestDiskCache(ReadEnsuresMemoryCacheCapacity|QuotaExceeded)$' -count=100
GOFLAGS=-mod=mod go test ./pkg/fileservice -run '^TestDiskCache(ReadEnsuresMemoryCacheCapacity|QuotaExceeded)$' -race -count=10
GOFLAGS=-mod=mod go test ./pkg/fileservice -count=1
```


XuPeng-SH (approved):
Reviewed current head 79501326f. This correctly closes the test-owned CachedData lifecycle without changing production FileService behavior.

- DiskCache.Read only uses the injected memory cache as a capacity guard here; it does not take ownership of the allocated result.
- The sole caller reference remains in the IOVector, and the defer is registered before Read, so success, read-error, and assertion-failure exits all release it exactly once.
- Releasing an unpopulated vector is a safe no-op; no double-release edge is introduced.
- Q1 is closed by IOVector.Release; Q2 has no new wait dependency; Q3 only shortens the lifetime of the fixed-size allocation.

Verified with the checked allocator reproduction at count 100, focused race tests at count 10, the full fileservice package, go build, go vet, and a clean virtual merge against current main. All remote CI checks are green.

PR #25843, opened 2026-07-17
refactor: unify SQL resource accounting semantics

### LINKED ISSUE #18449: [Feature Request]: Resource statistics when compile retry execute
### Is there an existing issue for the same feature request?

### Is your feature request related to a problem?

```Markdown
由于compile物理计划失败重试时，不同CU资源维度，统计策略并不一样， 主要涉及如下三个方面：

1. 对于物理计划的Pipeline执行， 其资源消耗（内存Mem， CPU时间， 网络IO等），会将每次重试执行的资源消耗累加到原始的Plan Node中， 但是如果重试执行时，逻辑计划重新BuildPlan，那么之前执行失败的Pipeline资源消耗会被丢弃

2. 对于S3 空间资源的统计，在每次重试执行时，会用 `NewCompile` 方法重新创建一个 `Compile` 对象，重新实例化`counterSet`作为观察者， 因此，S3空间资源的统计总是来自于最后一次物理计划的重试执行

3. 对于S3 和磁盘IO等待时间 ，在前端接收到请求后在Context中埋入StatsInfo, 物理计划执行期间，由`FileService`磁盘IO, 网络IO，缓存IO和 锁等待的时间统计，该部分时间资源的统计贯穿整个sql执行阶段，包括了多次失败重试

基于以上现状，在发生compile发生重试时，资源统计策略并不一致，而且系统没有重试次数的限制，这样可能会导致，CU的时间统计结果为负
```

### Describe the feature you'd like

1. 当sql执行compile出现重试执行， 对每次重试执行的物理资源单独统计，不遗漏
2. 当发生重试执行时， 要做执行的资源统计模块重置， 每次执行的资源互补干涉

### Describe implementation you've considered

_No response_

### Documentation, Adoption, Use Case, Migration Strategy

_No response_

### Additional information

_No response_

### LINKED ISSUE #22263: [Bug]: negative cpu reported

### Branch Name

main

### Commit ID

latest

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

Image

### Expected Behavior

_No response_

### Steps to Reproduce

```Markdown
Daily Run

https://grafana.ci.matrixorigin.cn/explore?panes=%7B%225nk%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-main-nightly-e0c9a88cf-20250724%5C%22%7D%20%7C%3D%20%60negative%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221753409546514%22,%22to%22:%221753412236678%22%7D%7D%7D&schemaVersion=1&orgId=1
```

### Additional information

_No response_

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

Fixes #22263
Fixes #18449

This PR refactors SQL resource accounting into one authoritative statement-level aggregation path.

- gives retry attempts independent resource generations and merges every attempt without producing negative CPU;
- unifies local, remote, nested-remote, and local-fallback accounting without double counting;
- records meaningful active time, classified wait time, S3/network/spill usage, exact statement memory-domain totals, quality flags, and terminal outcomes;
- projects the resulting summary consistently to statement_info.stats and EXPLAIN ANALYZE while keeping the persisted schema unchanged;
- adds focused unit, race, BVT, and local acceptance coverage, and documents the semantics and known limits in the RFC.

### Observability and AI-analysis benefits

The result is a stable, machine-readable statement-level resource summary that downstream diagnostics and AI analysis can use without reconstructing semantics from unrelated counters.

- active time and classified wait time make CPU, lock, filesystem, remote, and network bottlenecks distinguishable;
- attempt count and retry wall time expose retry amplification instead of folding retries into ambiguous or negative values;
- allocated/freed bytes, maximum memory-domain peak, bounded sum of domain peaks, spill bytes, and S3/network facts provide useful evidence for OOM and resource-pressure investigations;
- outcome, quality flags, and missing fragment/domain counts let analysis reject or qualify incomplete samples instead of treating them as exact;
- the same semantics in statement_info.stats and EXPLAIN ANALYZE make workload comparison and query-level diagnosis consistent.

This PR intentionally provides statement and plan-level facts, not exact per-operator or per-CN memory attribution. A process killed before terminal publication may still have no final statement record; these limits are documented in the RFC.

Validation completed locally:

- pkg/sql/compile and pkg/vm/process tests;
- pkg/sql/compile with the race detector;
- go vet for the affected packages;
- make build.


XuPeng-SH (approved):
Deep-reviewed exact head `ca97a095a54d6eabab851cb6eeb55f8aaba6b4f1`; no blocking correctness issue found.

I traced the complete ownership/lifecycle chain across statement root, retry generations, local/remote/nested/fallback fragments, operator analyzers, MPool epochs, protocol output, StatsArray v6, CU, and EXPLAIN persistence. In particular:

- attempt finish/publish is exactly once on success, error, retry-build failure, and panic; child executions cannot steal the top-level attempt owner;
- remote terminal reports are sender-deduplicated, missing direct reports propagate once across nested hops, and remote fallback is counted locally;
- remote cleanup captures operator facts before clear and allocator facts after clear; stale/wrong MPool tokens cannot terminate a newer epoch;
- statement completion clears the live preview, ends the exact epoch, records protocol bytes/packets, and seals before projection/export;
- merge operations saturate on overflow, report invariant/partial quality explicitly, and retain O(1) state in retry/fragment count;
- no mutex critical section performs RPC, I/O, channel wait, or goroutine join.

Two suspicious theoretical paths were checked and are not reachable as accounting loss in the current call graph: `newCompile` can fail only before an execution MPool domain exists, while all later remote errors are inside terminal capture; `PrepareDoneNotifyMessage` uses a sender with `AnalyzeModule=nil`, so its empty v1 envelope cannot inflate direct fragment reports.

Validation:

- targeted tests for `pkg/util/resource`, `pkg/common/mpool`, `pkg/vm/process`, `pkg/util/trace/impl/motrace/statistic`, `pkg/sql/models`, `pkg/sql/compile`, and `pkg/frontend`
- race tests for resource, mpool, operator analyzer, statistic, and compile packages
- clean merge with current `origin/main`, followed by the same targeted package tests
- required PR checks are green, including UT, SCA, multi-CN/proxy BVT, coverage, and Summary

gouhongshen (changes requested):
## Codex automated review

Found three blocking accounting regressions. The existing approval does not address or disprove these paths; there are no inline review comments. Focused resource, statistic, and fileservice tests passed, and CI is green.

### P1 - Account for writable external-table I/O (pkg/sql/colexec/insert/insert.go:281)
The external writer receives proc.Ctx directly for both WriteBatch and Close, without an analyzer-owned CounterSet or filesystem-wait boundary. Its asynchronous FileService.Write therefore updates only the execution-wide counter, which collectScopeResourceDelta never merges. Before this PR, v5 still estimated S3 input from analyzer.WrittenRows; typed v6 projection deliberately ignores those plan statistics, so an external-table write now reports zero PUT/write bytes and treats blocking output time as active, all with quality=0. Accounting must cover the writer's entire asynchronous lifetime.

### P1 - Retain bytes written before an object-store failure (pkg/fileservice/s3_fs.go:483)
Both error branches return before publishing n to S3WriteSize. A multipart upload can successfully write earlier parts and then fail on a later part or completion; those physical bytes are discarded from the failed attempt even though the RFC requires failed and retried work to be retained. Terminal analyzer harvesting cannot recover them, and no partial-quality flag is set. Successful physical parts need to be accounted at their ownership boundary rather than only after the whole upload succeeds.

### P1 - Mark unavailable root-phase active work as partial (pkg/util/trace/impl/motrace/statistic/resource_stats.go:74)
Plan and compile I/O counters aggregate parallel waits and can legitimately exceed their coordinator wall interval. In that case this helper silently omits the entire phase's active contribution while still publishing its wait, and TestRootPhaseResourceKeepsAggregateWaitWithoutInvalidActive explicitly expects quality=0. StatsArray v6 and CU consequently treat an undercounted active value as complete. Use a producer-local interval or set QualityPartial when active work cannot be derived.

heni02 (approved):

aptend:
Addressed two of the three findings in cc6d53750:

- Writable external-table writes now own a dedicated counter for the complete async writer lifetime. WriteBatch, Close, and Abort waits are classified as filesystem wait; counters are harvested only after the writer goroutine is joined. The success and abort paths both have focused tests.
- Plan/compile phases now set QualityPartial when aggregate wait exceeds coordinator wall and active time therefore cannot be derived.

I do not think publishing s3_fs n on the error branches is a correct fix for the multipart point. n measures bytes consumed from the source reader, not bytes successfully accepted by object storage. SDK buffering/read-ahead and a rejected request can make n larger than successful physical writes, so moving the current publication into a defer would replace an undercount with an overcount while still claiming exact semantics.

Exact failed-multipart accounting has to be emitted after each successful PutObject/UploadPart ownership boundary, with retry semantics defined and implemented consistently across the object-storage backends and parallel writers. That is a different producer-level change; the generic S3FS layer cannot reconstruct it from n. My proposal is to keep S3WriteBytes strict in this PR and handle successful-part accounting at those SDK boundaries separately. If you believe part bytes from an eventually aborted multipart upload must be included in this PR, please clarify whether S3WriteBytes should mean accepted transfer bytes including later-aborted parts, rather than bytes belonging to a successfully committed object.

gouhongshen (changes requested):
## Codex automated review

Two blocking accounting regressions remain. Focused pure-Go resource/statistics tests passed; CGo-transitive tests could not run because the expected cgo/libmo.dylib is absent.

### P1 - Keep result encoding out of output wait time (pkg/sql/colexec/output/output.go:87)
MeasureOutputWaitErr times the entire output.Func callback. For normal SQL this reaches getDataFromPipeline → RespResult → MysqlProtocolImpl.Write, which allocates result rows, converts every vector, and encodes every row in addition to performing blocking writes. operatorAnalyzer.Stop subtracts this whole duration from the Call wall before publishing ExclusiveActiveNS, which CU prices as CPU. Consequently, result conversion and serialization that origin/main counted as active work are now classified as wait, systematically underbilling result-heavy queries. Only actual blocking protocol operations should be measured as output wait.

### P1 - Account multipart writes at physical upload boundaries (pkg/fileservice/s3_fs.go:483)
The new write-byte publication occurs only after the whole multipart operation succeeds and uses bytes consumed from the source reader. AWS and COS multipart implementations call UploadPart directly without incrementing FileService.S3.Put, so even successful large writes report zero physical PUTs. They can also accept some parts before another part, read, or completion fails; this return then publishes neither those accepted bytes nor a partial-quality flag. The new discussion correctly explains why publishing the source-reader count on error would overcount, but it does not resolve the omission: PUT counts and accepted job.n bytes must be recorded at each successful PutObject/UploadPart boundary, replacing the logical whole-write counter.

aunjgr (approved):
Reviewed independently with a full ownership/wait-for/retry-generation audit. The statement root has one terminal seal; top-level retry attempts are single-owned and folded into fixed-size state; success, non-retry error, retry transition/build failure, and panic paths converge on publication. Remote fragments emit the same terminal aggregate on success and error, nested reports are reduced once per hop, and absent reports become explicit partial-quality counts. Protocol output is measured after terminal writes, mpool epochs use identity-safe closure, and operator counter/wait intervals are harvested once across normal, error, and reset paths. Overflow and incomplete domains fail visibly through quality flags instead of fabricating usage. No actionable issues found.

aptend:
Addressed both remaining findings in c622018a3.

Output wait:
- removed callback-wide output wait measurement;
- result conversion and row encoding remain active work;
- only the physical net.Conn.Write boundary records WaitOutput, through the analyzer-owned per-Call counter;
- failed and partial socket writes retain their actual wait, and the scoped counter is restored on every exit.

Object-store writes:
- removed the logical whole-write S3FS byte publication;
- PUT attempts are counted at each real PutObject/UploadPart invocation, including failures and retries;
- accepted bytes are published only after a successful PutObject/UploadPart response;
- successful multipart parts remain recorded if a later part or CompleteMultipartUpload fails;
- AWS serial/parallel multipart, COS parallel multipart, and the ordinary AWS/COS/Aliyun/MinIO PutObject paths now publish at their physical ownership boundaries.

Validation on the exact head:
- full tests for perfcounter, fileservice, frontend, process, output, insert, compile, and statistic packages;
- full race tests for fileservice, frontend, and process;
- focused race stress for successful and failed AWS/COS multipart accounting and physical socket output wait;
- go build, go vet, make build, and git diff --check.

gouhongshen (changes requested):
## Codex automated review

The latest author discussion is supported by the exact HEAD: socket output wait is now limited to physical writes, and object-store attempts/accepted bytes are recorded at physical SDK boundaries, so those prior findings are withdrawn. Three blocking accounting regressions remain. The focused pure-Go resource tests passed; frontend tests could not run because cgo/libmo.dylib is absent.

### P1 - Merge ANALYZE's derived query into the client statement resource root (pkg/frontend/mysql_cmd_executor.go:236)
executeAnalyzeDerivedQuery recursively calls doComQuery with outerExecCtx.reqCtx while the session is marked derived, so the context already contains the ANALYZE statement's root. RecordStatement unconditionally replaces it with a new root, and resourceAttemptOwnerEligible excludes only backSession, allowing the generated SELECT to claim its own attempt. Consequently, the table-scan CPU, waits, S3 usage, and attempt count are exported under a separate generated statement while the client-visible ANALYZE retains only outer phases/memory. This violates the stated one-root model and misattributes billable work; same-session derived execution must merge into the parent root without owning attempts.

### P1 - Do not defer completion of nested same-session statements (pkg/frontend/util.go:469)
responseAccounting is enabled for the entire client request, so a failing ANALYZE-derived SELECT also satisfies deferStatementCompletion and skips finishStatementAccounting for its inner StatementInfo. The nested doComQuery then restores the outer ses.tStmt, making the inner StatementInfo unreachable; finishResponseAccounting later closes only the outer statement. The derived failure therefore never reaches EndStatement, never seals its resource root, and never emits its terminal failure/accounting record. Deferral must be scoped to the StatementInfo that owns the client response, or nested derived work must share the outer statement lifecycle.

### P1 - Keep external-table row encoding in active time (pkg/sql/colexec/insert/insert.go:288)
The new wait wrapper measures the entire ExternalWriter.WriteBatch call as WaitFilesystem. WriteBatch performs encodeCSV or encodeJSONLine before calling fw.Write, so potentially substantial row conversion and serialization CPU is included in the wait interval. operatorAnalyzer.Stop subtracts that interval from the Call wall, causing writable external-table encoding to disappear from ExclusiveActiveNS and CU accounting. origin/main treated this work as active. Measure only the actual blocking file-service write/open boundaries and keep encoding outside the wait scope.

aptend:
Addressed the three findings in 7fe0a6319 after rebasing onto current main.

- Same-session derived SQL now keeps its own StatsInfo but reuses the parent resource root, creates no nested StatementInfo, and cannot claim an attempt. The outer request remains the sole terminal owner on success and failure, including protocol counters.
- Writable external-table encoding is active work again. Only FileServiceWriter open/write/header-write and terminal close/abort boundaries contribute filesystem wait; the existing dedicated async counter lifetime and post-join harvest remain unchanged.
- Added focused lifecycle/ownership tests, including preservation of the parent response counters.

Validation on the new head: related frontend/externalwrite/insert package tests, race tests (including repeated focused lifecycle tests), go vet, make build, and git diff --check all pass.

aptend:
Rebased onto current main (`1ac64a870`) and resolved the new conflicts in `c7654f08a`.

Conflict handling preserved the upstream changes rather than reviving removed paths:
- adapted insert accounting to the removal of the legacy memory-table branch;
- kept the new per-part goroutine and dual-semaphore multipart upload lifecycle, adding physical PUT/accepted-byte accounting at its actual UploadPart boundaries;
- adapted frontend tests to the removed memory-engine scaffolding.

Fresh validation after the rebase: relevant package tests, full race tests for fileservice/frontend/insert, go vet with the required CGo environment, make build, range-diff inspection, and git diff --check. The PR is MERGEABLE and the existing pending review requests remain present.

gouhongshen (changes requested):
## Codex automated review

Two accounting correctness regressions remain in the current head.

### P1 - Response finalization drops root-phase accounting on errors (pkg/frontend/routine.go:321)
ExecRequest/doComQuery replaces execCtx.reqCtx with contexts carrying StatsInfo and the resource root. On parse, rewrite, authorization, or compile errors, logStatementStatus now defers EndStatement, but this finalizer passes the original tenantCtx, which has neither value. StatementInfo.EndStatement therefore cannot ClaimRootPhaseResource, silently dropping parse/plan/compile time, waits, and S3 requests without marking the result partial. The finalizer must retain and use the enriched execution context.

### P2 - Output wait is measured only while rows are buffered (pkg/frontend/mysql_protocol.go:3558)
withOutputCounter is restored as soon as writeResultSetRow2 returns. FinishedPacket flushes only when the buffer reaches maxBytesToFlush, so normal small results are sent later by Write/Flush after the counter is cleared; metadata, EOF/OK, errors, WriteResultSetRow, and sendResultSet likewise have no counter scope. Conn.WriteToConn consequently records client bytes but misses the actual socket-blocking time in ProtocolOutputWaitNS for common responses.

aptend:
Addressed both findings in f8e1a7663 after rebasing onto current main (0d7eeb38b).

- Response finalization now uses the enriched ExecCtx request context, including panic/error paths, so parse/plan/compile root phases and their StatsInfo are retained through deferred statement completion.
- Output wait now spans the complete statement response at the physical net.Conn.Write boundary, covering buffered flushes, metadata, terminal OK/EOF, and errors.
- Immediate writes executed inside Output.Call continue to be published through the operator-owned counter and subtracted from active time. A lightweight per-response tracker records total physical-write wait, and finalization merges only the remainder not already owned by the operator, preventing double counting.
- Trackers rotate between multi-statements and detach at request completion; same-session derived SQL remains owned by the outer response lifecycle.

Validation on the rebased head:
- focused frontend accounting tests repeated 10x;
- full pkg/frontend, pkg/sql/colexec/output, and pkg/vm/process tests;
- race tests for those packages;
- go vet with the required CGo environment;
- make build, range-diff inspection, and git diff --check.

gouhongshen (approved):

PR #25844, opened 2026-07-17
fix(txn): make RPC queue admission cancellable

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25786

Make txn RPC queue admission observe both request cancellation and server shutdown when the bounded queue is full.

The shutdown path now:

1. closes producer admission and wakes blocked producers;
2. stops the RPC entry point and waits for admitted producers;
3. stops queue consumers;
4. cancels and releases any residual queued requests.

Concurrent and repeated `Close` calls are idempotent and return the same RPC close error. All pre-enqueue exits retain cleanup ownership; ownership transfers to the worker/drain path only after a successful queue send.

Testing:

- focused white-box regression tests, `-count=20`
- full `pkg/txn/rpc` tests
- full `pkg/txn/rpc` tests with `-race`
- `go build ./pkg/txn/rpc`
- `go vet ./pkg/txn/rpc`
- dependent `pkg/txn/service` regression test


XuPeng-SH (approved):
Reviewed the complete change from first principles, including cancellation, shutdown ordering, request/response ownership, concurrent and repeated Close, bounded queue behavior, and worker/producer wait-for dependencies. I found no correctness or lifecycle blocker.

Local verification on head 918399ba5df7bec93f6b6c4ec12087c5f142e72a:
- focused cancellation/close race tests, `-race -count=50`
- full `pkg/txn/rpc` test and race test
- `pkg/txn/service` regression test
- `go vet` and `go build` for `pkg/txn/rpc`
- `git diff --check`

The branch is currently behind main; please update it and require green CI before merge.

Issue #25845, opened 2026-07-17
[Bug] Chaos (33bc7a2): TPCC durable ytd corruption on warehouse 8 after `task_kill_dn` — Payment district/warehouse split never heals


heni02:
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-chaos-4-1-dev-33bc7a2-202607131556` |
| Namespace | `mo-chaos-4-1-dev-33bc7a2-202607131556` |
| MO version | env name `33bc7a2`; running image at incident **`commit-ce99d329c`** (S3 prefix still `…/commit-33bc7a2/data`) |
| Endpoint | `10.222.6.253:6001` |
| Artifact root | `/data1/chaos-test/29230227851_33bc7a2` |
| Workloads | TPCC (jTPCC 10 wh / 10 terminals), mo-load, vector/fulltext, chaos-mesh |
| Incident start | **2026-07-17 15:52:14 CST** |
| Analysis point | **2026-07-17 17:20 CST** (still failing; **0** PASS after incident) |

**Scope:** **Permanent** TPCC `ConsistencyCheck` failure on **`w_ytd` ↔ `sum(d_ytd)`** for **warehouse id = 8** after chaos **`task_kill_dn`**, preceded by **20702 / 20502** and **1062** bursts. Drift **does not self-heal** and **continues to widen** as Payments run.

**Out of scope / do not merge:**

- Transient **o_ol_cnt** read-skew (self-heal in ≤1 min) — earlier same env (e.g. Jul 17 01:28) and 2865ed6 / 2a00828 class.
- DN merge `file … is not found` ([#25718](https://github.com/matrixorigin/matrixone/issues/25718)) — **same soak, different defect**; do not treat ytd as caused by rawlog/statement_info missing aobj without proof.
- Single New-Order **1062** without sustained ytd fail ([#25704](https://github.com/matrixorigin/matrixone/issues/25704) / mem_cn 1062 class).

## Summary

On `commit-ce99d329c` chaos long-run, TPCC verify was healthy through **15:51:40 CST**, then starting **15:52:14** every cycle fails **ytd** invariants for **w_id=8**. Through **17:20+** (**≥1.5h**, **290+** `Consistency verification failed` lines, **0** `Consistency verification successfully`):

```sql
(SELECT d_w_id, sum(d_ytd) FROM bmsql_district GROUP BY d_w_id)
EXCEPT (SELECT w_id, w_ytd FROM bmsql_warehouse);

[...]

PR #25847, opened 2026-07-17
fix(lockservice): reject late binds from stale allocators

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25845

`#24906` fences cached lock-table binds by allocator epoch, but two asynchronous
`NewBind` response paths could still bypass that fence:

1. a remote lock or keep-remote request is sent using allocator A0;
2. while the RPC is in flight, the CN observes allocator A1 and purges A0 binds;
3. the delayed A0 response carries a `NewBind` whose allocator ID matches the
   request bind;
4. the response was published directly and recreated an A0 lock table after the
   A1 purge.

This ordering is reachable on the normal CN path during TN/allocator restart and
network delay. The keeper always runs on a lock service and sends asynchronous
keep-remote RPCs, while ordinary remote lock operations can receive `NewBind`
concurrently with allocator keepalive processing.

The change treats an owner response as a point-in-time hint. It refreshes the
bind from the current allocator and publishes it only through the existing
allocator-epoch validation. This covers both the keeper response path and the
normal remote lock-table response path.

If a foreground remote operation cannot refresh the bind, it returns
`ErrLockTableBindChanged` and never publishes the response bind. The keeper has
no caller to receive that error, so a refresh failure after a proven bind-change
signal (`NewBind`, `ErrLockTableBindChanged`, or `ErrLockTableNotFound`)
atomically removes the exact old bind and fences transactions that use it.
Generic transport failures remain fail-neutral: a failed refresh does not
invalidate a bind without proof that it changed.

All added allocator refreshes inherit the caller or keeper-round context. A
burst of changed tables therefore remains bounded by the existing round
deadline instead of waiting one full RPC timeout per table.

The additional `GetBind` RPC runs only after an exceptional bind-change signal,
not on normal lock or keepalive success. There are no wire-format changes.
Same-allocator legitimate rebinds remain supported after allocator validation.

Regression tests deterministically cover:

- a delayed A0 owner response arriving after the CN observed and purged A0;
- allocator refresh failure after `NewBind` and explicit bind-not-found;
- exact fencing without affecting another table on the same owner and version;
- foreground and keeper refresh cancellation;
- allocator refresh failure never publishing the response bind.

Tested with:

`go test ./pkg/lockservice -coverprofile=/tmp/pr25847-final.out -count=1`

`go test -race ./pkg/lockservice -count=1`

`go test ./pkg/lockservice -run '^(TestKeepRemoteLockBindRefreshHonorsContext|TestRemoteNewBindRefreshHonorsContext)$' -count=50`

`go test ./pkg/lockservice -run '^TestKeepRemoteLockBindChangedRefreshFailureInvalidatesOldBind$' -count=10`

`go build ./pkg/lockservice`

`go vet ./pkg/lockservice`


XuPeng-SH (changes requested):
The allocator-epoch validation is directionally correct, but one High-severity fail-open path remains in the keeper.

When `KeepRemoteLock` returns `resp.NewBind`, the old owner has not renewed the requested bind. `doKeepRemoteLock` now calls `maybeHandleRemoteBindChanged`; if the follow-up `GetBind` fails, that helper only logs and returns (lines 188-192 and 210-222). It correctly avoids publishing the unvalidated response bind, but it also leaves active transactions holding the old bind unfenced. Unlike `remoteLockTable.maybeHandleBindChanged`, the background keeper has no caller to receive `ErrLockTableBindChanged`.

A concrete sequence is:

1. An active transaction holds an A0 remote bind.
2. The owner replies with `NewBind`; the old bind is not renewed.
3. The allocator refresh times out or otherwise fails.
4. `txn.bindChanged` remains false, so the transaction can continue and commit with an unmaintained remote lock until a later keeper iteration.

That is a fail-open window in the same correctness class as #25845. The new remote refresh-failure test does not exercise this keeper path.

Please distinguish proven bind-change signals from generic transport failures. For `NewBind` (and, if appropriate, explicit bind-changed/not-found errors), a refresh failure should conservatively fence transactions holding the old bind without publishing the unvalidated response bind. Please also add a deterministic keeper regression test: `KeepRemoteLock` returns `NewBind` + `GetBind` returns an error => the active transaction is fenced and no stale response bind is published.

gouhongshen (approved):
## Codex automated review

No blocking regressions found. The allocator refresh is correctly fenced through existing epoch validation, keeper refresh failures fence only the exact stale bind, and concurrent bind changes are serialized. Focused local tests could not run because the checkout lacks cgo/libmo.dylib; PR CI is green.

XuPeng-SH (approved):
Deep re-review completed on 033602dc7. No blocking correctness, performance, concurrency, resource-leak, or liveness regressions found. The stale NewBind response is now validated through the current allocator epoch; keeper refresh failures after proven bind-change signals fence only the exact old bind. Focused stress/race tests, full pkg/lockservice tests and race tests, go vet, diff check, and merge-tree validation passed locally.

PR #25850, opened 2026-07-17
fix(pSpool): close pooled resource ownership on all exits

### LINKED ISSUE #25759: [Bug]: pSpool SendBatch leaks capacity after batch copy failure

## Branch Name

main

## Commit ID

e538e203ad47ebf0506a424fe93aed61290c4061

## Actual Behavior

`PipelineSpool.SendBatch` acquires a message slot from `freeShardPool` before copying the batch. If `cachedBatch.GetCopiedBatch` fails, it returns the error without returning that message slot (`pkg/container/pSpool/sender.go:72-80`).

`GetCopiedBatch` also pops a cache ID/batch from `spoolBuffer.readyToUse` before copying and does not put it back on its error paths (`pkg/container/pSpool/copy.go:73-120`).

For a one-receiver spool, the shared and cache capacities are both two. Two deterministic mpool allocation failures consume both capacities. A third send no longer reaches the copy operation: it blocks waiting for a free slot until its context expires and returns `(done=true, err=nil)`, masking the original failure and permanently exhausting that spool.

## Expected Behavior

A failed copy must roll back every resource acquired by that `SendBatch` call. Repeated copy failures should continue returning the copy/allocation error and must not reduce future spool capacity.

## Steps to Reproduce

1. Create a source batch containing a 2 MiB varchar using an unlimited mpool.
2. Create `InitMyPipelineSpool` with one receiver and a destination mpool limited to 1 MiB.
3. Call `SendBatch` three times, each with a 100 ms context.

Observed:

- calls 1 and 2 return the mpool allocation error;
- call 3 waits for the context, then returns `done=true, err=nil`.

Invariant:

```go
for i := 0; i < 3; i++ {
    done, err := spool.SendBatch(ctx, 0, oversizedBatch, nil)
    require.Error(t, err)
    require.False(t, done)
}
```

The third iteration fails both assertions on current code.

## Additional information

The rollback needs to restore both ownership layers: the `messageIdx` acquired from `freeShardPool` and the cache ID/batch removed from `readyToUse`. Restoring only one still leaves the other pool permanently depleted.

### LINKED ISSUE #25854: [Bug]: pSpool Abort can clear a live batch reused by another slot

### Branch Name

main

### Commit ID

- Hardware parameters: N/A; deterministic unit reproduction
- OS type: Platform independent
- Others: Pipeline spool with one or multiple local receivers

### Actual Behavior

After a receiver releases a batch, pSpool returns the batch to its cache and the slot to its free pool but leaves the old slot metadata intact. Because the cache and slot pools reuse objects independently, that batch can next be owned by another slot while the old free slot still aliases it. Abort scans and cleans every non-current slot, so cleaning the stale slot clears the batch that a receiver is currently reading. The receiver observes an empty batch and data loss.

### Expected Behavior

A slot must relinquish all data, error, cache, and reference metadata before it is published to the free pool. Abort must preserve every batch currently owned by a receiver until the final ReleaseCurrent call.

### Steps to Reproduce

1. Initialize a spool with one receiver.
2. Send, receive, and release a batch so its batch object and slot are returned to their independent pools.
3. Send and receive a second batch. The cache reuses the first batch object in the other slot.
4. Call Abort before ReleaseCurrent.
5. Observe that the current batch row count changes from 1024 to 0.

The same failure reproduces with a broadcast batch and two receivers.

### Additional information

The stale metadata originates at the normal release boundary. Abort cleanup introduced in #25273 makes the stale alias observable, but adding an Abort-specific exception would not restore ownership correctness. The systemic fix is to detach metadata before returning either pooled object, with single-receiver, broadcast, late-release, cleanup, and race coverage.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25759
Fixes #25854

This PR supersedes #25855 and closes two missing transitions in the same pSpool ownership state machine.

PipelineSpool independently pools message slots and cached batches. Every checkout must either be published to a receiver or rolled back, and every final release must detach the old owner before either pooled object becomes reusable.

This PR restores that lifecycle invariant on both paths:

- Failed admission: when batch copying fails, return the partially built batch and cache ID through the existing cache path, then return the unpublished message slot before propagating the original error.
- Successful final release: clear all slot data, error, cache, and reference metadata before returning the batch to the cache and publishing the slot to the free pool.
- Abort and late cleanup: reuse the same slot-clear operation so pending, current, broadcast, and cleanup-after-terminal paths cannot double-own or prematurely free a batch.

The successful release path adds no locks, goroutines, allocations, scans, atomic operations, or extra helper calls. It only performs the fixed-size slot clear required to transfer ownership. Copy-failure cleanup runs only on the error path.

Testing:

- pSpool full package, count 100
- pSpool full race suite, count 20
- focused race matrix, count 100, including 256 concurrent Abort and broadcast-release interleavings per run
- flat-vector and const-vector copy failures, repeated rollback, restored capacity, subsequent successful send, and final MPool cleanup
- connector, dispatch, and process full package tests
- connector, dispatch, and process terminal/abort race tests, count 10
- go build on pSpool and all direct consumers
- go vet on pSpool
- git diff check

### DIFF
--- a/pkg/container/pSpool/copy.go
+++ b/pkg/container/pSpool/copy.go
@@ -105,7 +105,7 @@ func (cb *cachedBatch) GetCopiedBatch(
 
 		if vec.IsConst() {
 			if err = vector.GetConstSetFunction(typ, cb.mp)(dst.Vecs[i], vec, 0, vec.Length()); err != nil {
-				dst.Clean(cb.mp)
+				cb.CacheBatch(true, cacheID, dst)
 				return nil, false, 0, err
 			}
 
@@ -116,7 +116,7 @@ func (cb *cachedBatch) GetCopiedBatch(
 			if err = vector.GetUnionAllFunction(typ, cb.mp)(
 				dst.Vecs[i],
 				vec); err != nil {
-				dst.Clean(cb.mp)
+				cb.CacheBatch(true, cacheID, dst)
 				return nil, false, 0, err
 			}
 
--- a/pkg/container/pSpool/sender.go
+++ b/pkg/container/pSpool/sender.go
@@ -108,6 +108,7 @@ func (ps *PipelineSpool) SendBatch(
 
 	dst, useCache, cacheID, err := ps.cache.GetCopiedBatch(data)
 	if err != nil {
+		ps.freeShardPool <- messageIdx
 		return false, err
 	}
 	ps.updateSpoolMessage(messageIdx, dst, info, useCache, cacheID)
--- a/pkg/container/pSpool/sender.go
+++ b/pkg/container/pSpool/sender.go
@@ -335,8 +335,16 @@ func (ps *PipelineSpool) releaseCurrentLocked(idx int) {
 			if ps.cleanupDone {
 				ps.cleanSlotLocked(last)
 			} else {
-				ps.cache.CacheBatch(
-					ps.shardPool[last].useCache, ps.shardPool[last].cacheID, ps.shardPool[last].dataContent)
+				// The slot and batch pools have independent reuse orders, so a
+				// returned batch may be assigned to a different slot immediately.
+				// Detach the batch before either object is republished; otherwise
+				// this slot can retain an alias that later cleanup frees prematurely.
+				msg := &ps.shardPool[last]
+				data := msg.dataContent
+				useCache := msg.useCache
+				cacheID := msg.cacheID
+				ps.clearSlotLocked(last)
+				ps.cache.CacheBatch(useCache, cacheID, data)
 				ps.freeShardPool <- last
 			}
 		}
@@ -383,8 +391,10 @@ func (ps *PipelineSpool) cleanSlotLocked(idx uint32) {
 	if msg.useCache && msg.dataContent != nil {
 		msg.dataContent.Clean(ps.cache.mp)
 	}
-	msg.dataContent = nil
-	msg.errContent = nil
-	msg.useCache = false
-	msg.cacheID = 0
+	ps.clearSlotLocked(idx)
+}
+
+func (ps *PipelineSpool) clearSlotLocked(idx uint32) {
+	ps.shardPool[idx] = pipelineSpoolMessage{}
+	ps.doRefCheck[idx] = false
 }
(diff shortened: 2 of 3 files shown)


LeftHandCold (approved):
Reviewed the complete diff at `d8e0e655a3d3d32f76a1dde7f80a8cb14ad97fd6`. No blocking issues found.

Ownership and lifecycle audit:

| Check | Transition | Result |
|---|---|---|
| Q1 / exactly-one owner | free slot + cache ID -> claimed copy -> published receiver slot -> final release -> cleared slot + cached/freed batch | PASS |
| Q1 / failed admission | copy failure -> partial batch rollback -> cache ID restored -> unpublished slot restored | PASS |
| Q1 / abort and late release | queued slots cleaned immediately; current single/broadcast references survive until final `ReleaseCurrent`; cleanupDone releases directly | PASS |
| Q2 / termination | free-slot wait observes context or abortDone; terminal cleanup is bounded; rollback sends cannot block because each send restores its own checkout | PASS |
| Q3 / boundedness | slot pool, cache IDs, and receiver rings remain fixed by `getBufferLength(receiverCnt)` | PASS |

I also traced the closure through connector/dispatch Reset, PipelineSignalReceiver release/timeout, and deferred spool cleanup. Clearing the slot before either pool republishes its object removes the stale alias without changing the refcount linearization point. The stale `shardRefs` value is intentionally harmless because `doRefCheck` gates it and every broadcast/Abort claim overwrites it before use.

Local validation:
- full tests for pSpool, connector, dispatch, and process: PASS;
- full pSpool race suite with `-race -count=20`: PASS;
- connector/dispatch/process terminal paths with `-race -count=10`: PASS;
- `go vet` and `go build` for the owning package and direct consumers: PASS;
- `git diff --check`: PASS.

gouhongshen (approved):
## Codex automated review

No correctness, lifecycle, concurrency, resource, compatibility, performance, or coverage findings in the PR diff.

PR #25851, opened 2026-07-17
perf(plan): avoid TableDef deep copies in TP planning

### LINKED ISSUE #24097: perf: query execution performance improvements
## Summary

Umbrella issue for tracking query execution performance improvements in MatrixOne, covering both AP and TP workloads.

### Computation
- Branchless arithmetic kernels for hot paths
- Type promotion avoidance (keep narrow types when possible)
- Batch operation kernels with pre-computed constants
- SIMD intrinsics for vectorized computation (currently batch-level only)

### Memory
- Zero-copy data paths (reduce unnecessary row copying / shuffling)
- Reduce memmove/memclr in operator pipelines

### Aggregation
- Lightweight null tracking (count-based instead of per-row bitmap checks)
- Specialized hashmap paths for small fixed-width keys
- Reduce varlena key construction overhead

### I/O
- I/O and computation interleaving (prefetch / async decompression)
- Decompression overhead reduction
- Syscall batching

### Filter & Scan
- Short-circuit evaluation for selective predicates
- Skip row copying when selectivity is high (in-place bitmap filtering)

### Optimizer Rules
- Cost model accuracy improvements
- Statistics estimation for complex predicates
- Group join optimization (combine join + group-by into single operator)
- Delimitation join pushdown
- Redundant cast elimination

### Point Lookup
- Short-path execution for simple point queries (bypass full optimizer)
- Index-only scans for covering indexes

### Write Path
- WAL write coalescing
- Lock-free conflict detection for non-overlapping writes

### Concurrency
- Reduce transaction coordination overhead for single-shard operations
- Optimistic concurrency control improvements

### Latency
- Reduce plan compilation time for simple queries
- Fast-path for single-table DML without cross-node coordination

aunjgr:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24097

`TableDef` deep copies account for a significant share of plan-compilation time in TP workloads. The engine already caches a canonical table definition, but each resolve and several scan/DML planning paths recursively copied its columns, indexes, keys, constraints, and expressions.

This change introduces a planner-owned `TableDef` shell. The planner copies the top-level definition and the `Cols` pointer slice, while sharing immutable nested schema metadata. It applies that clone to table resolution, scan-column pruning, constraint planning, pre-insert/update/delete contexts, and hidden-index maintenance. DDL and generic deep-copy boundaries retain full copies because they may mutate nested schema objects.

For a representative 64-column, 8-index definition, the included benchmark improves cloning from about 6.4 us, 149 allocations, and 25.6 KB to about 164 ns, 2 allocations, and 928 B.

Validation:

- `go build ./pkg/sql/plan ./pkg/frontend ./pkg/sql/compile` with MatrixOne CGo flags
- `go vet ./pkg/sql/plan ./pkg/frontend ./pkg/sql/compile` with MatrixOne CGo flags
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/sql/plan`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/frontend ./pkg/sql/compile`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=240s -run TestCloneTableDefForPlan ./pkg/sql/plan`


aunjgr:
@XuPeng-SH Updated the PR and merged the latest `main`.

The reported `TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots` failure is unrelated to this change and was confirmed as flaky. The `table_func_generate_series.test` failure at line 37 was a real pre-existing binder/executor schema mismatch exposed by the branch, not an expected-result problem:

- the binder advertised `VARCHAR` for string endpoints;
- the executor later rewrote only its local schema to `DATETIME`;
- MariaDB/JDBC then observed inconsistent result metadata and formatting.

The fix now binds temporal endpoints to a stable `DATETIME(fsp)` representation, preserves the string overload public `VARCHAR` result, and makes the executor emit vectors matching the prepared schema without mutating plan metadata. Typed temporal inputs continue to return `DATETIME`. I also added binder, schema-ownership, and executor regression tests.

Validation:
- `./pkg/sql/plan` and `./pkg/sql/colexec/table_function`: passed
- targeted `generate_series` BVT: 56/56 passed, 0 ignored
- added executable-line coverage for the fix: 88.1% (74/84)
- targeted `go vet` and `make build`: passed

Please take another look.

gouhongshen (changes requested):
## Codex automated review

One blocking schema-cache aliasing regression was found.

### P1 - Shallow Resolve clone exposes cached IndexDef to mutating DDL formatting (pkg/frontend/compiler_context.go:517)
Previously Resolve used CopyTableDef, which deep-copied IndexDef values. This clone only copies the outer TableDef and Cols slice, while GetTableDef returns the cached schema, so IndexDef pointers remain shared. ConstructCreateTableSQL mutates indexdef.Comment in place when escaping apostrophes (pkg/sql/plan/build_show_util.go:321); SHOW CREATE TABLE and CREATE TABLE ... LIKE both call it. A comment such as O'Reilly is therefore rewritten in the cached schema, escaped again on each subsequent call, and concurrently accessed planners can race. Preserve a deep-copy boundary for these formatting paths or make the formatter non-mutating.

aunjgr:
@gouhongshen Fixed in 9bfde4f5d.

ConstructCreateTableSQL now formats IndexDef.Comment into a local value and never mutates shared schema metadata. The same formatted value is used for both displayed and rewritten SQL, with standard doubled-quote escaping.

Added TestConstructCreateTableSQLDoesNotMutateIndexComments, which calls the formatter twice, checks identical output, verifies the source comment remains unchanged, and covers a comment containing an apostrophe.

Also merged the latest upstream/main at 826a3e3d2; the branch head is b0d8daf28.

Validation after the merge:
- pkg/sql/plan, pkg/sql/colexec/table_function, pkg/frontend, and pkg/sql/compile tests passed
- focused planner and table-function race tests passed
- targeted build and vet passed

Please take another look.

gouhongshen (approved):
## Codex automated review

No correctness, ownership, concurrency, compatibility, or resource-handling regressions found. Focused and affected package tests passed.

XuPeng-SH (approved):
Deep review completed: no blocking correctness, concurrency, lifecycle, compatibility, or performance issues found.

The planner now owns only the shallow TableDef shell and Cols slice while nested schema metadata remains shared and immutable; all audited mutation sites either replace top-level fields, operate on locally-created definitions, or retain the explicit DeepCopyTableDef boundary for ALTER/DDL. The SHOW CREATE comment path is now non-mutating and idempotent. generate_series binds a stable result schema and the PREPARE parameter path is closed by explicit TEXT-to-DATETIME casts rather than runtime plan mutation.

Verified against a synthetic merge with current origin/main (9bc6f2a647): pkg/sql/plan and pkg/sql/colexec/table_function passed, focused race tests passed, and the clone benchmark measured ~92-96 ns / 2 allocs / 928 B versus ~3.5-3.6 us / 149 allocs / 25.6 KB for deep copy. Please sync the latest main before merge.

PR #25855, opened 2026-07-17
fix(pSpool): clear ownership before recycling slots

### LINKED ISSUE #25854: [Bug]: pSpool Abort can clear a live batch reused by another slot

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25854

Fixes #25854.

pSpool manages message slots and cached batches in two independent pools. The normal release path returned both objects but retained the old batch, error, and cache metadata in the released slot. When the cached batch was reused by another slot, Abort could clean the stale free slot and thereby clear a batch still owned by a receiver.

This PR restores the ownership boundary:

- Detach all message metadata before returning the batch to the cache or publishing the slot to the free pool.
- Reuse the same clear operation for direct cleanup so normal, aborted, and late-release paths have one invariant.
- Cover the free-slot invariant and the stale-alias sequence for both a single receiver and a two-receiver broadcast.
- Cover concurrent Abort and final broadcast releases over all lock orderings.
- Verify current data remains readable until the final ReleaseCurrent and all MPool memory is reclaimed afterward.

The release path adds no locks, goroutines, allocations, scans, atomic operations, or extra helper calls. It only clears a fixed-size slot at the existing final-reference release point.

Testing:

- pSpool full package, count 100
- pSpool focused race matrix, count 100, including 256 Abort/release interleavings per run
- connector, dispatch, and process full package tests
- connector, dispatch, and process terminal/abort race tests, count 10
- pSpool plus PR #25850 combined race suite, count 20
- go build on pSpool and all direct consumers
- go vet on pSpool
- git diff check

### DIFF
--- a/pkg/container/pSpool/sender.go
+++ b/pkg/container/pSpool/sender.go
@@ -334,8 +334,16 @@ func (ps *PipelineSpool) releaseCurrentLocked(idx int) {
 			if ps.cleanupDone {
 				ps.cleanSlotLocked(last)
 			} else {
-				ps.cache.CacheBatch(
-					ps.shardPool[last].useCache, ps.shardPool[last].cacheID, ps.shardPool[last].dataContent)
+				// The slot and batch pools have independent reuse orders, so a
+				// returned batch may be assigned to a different slot immediately.
+				// Detach the batch before either object is republished; otherwise
+				// this slot can retain an alias that later cleanup frees prematurely.
+				msg := &ps.shardPool[last]
+				data := msg.dataContent
+				useCache := msg.useCache
+				cacheID := msg.cacheID
+				ps.clearSlotLocked(last)
+				ps.cache.CacheBatch(useCache, cacheID, data)
 				ps.freeShardPool <- last
 			}
 		}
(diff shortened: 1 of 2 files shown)


XuPeng-SH:
Superseded by #25850. The two fixes are now kept as separate commits in one PR because they close the failed-admission and successful-release transitions of the same pSpool pooled-resource ownership lifecycle. The combined diff and race matrix were verified before consolidation.

Issue #25856, opened 2026-07-17
[Bug]: Proxy deterministically runs out of memory as idle connections grow without bounds


XuPeng-SH:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

# Branch Name

main

# Commit ID

1a1473756016e2cebe7a39a76b441b4fc7a663f5

# Other Environment Information

- Hardware parameters: Kubernetes deployment with two Proxy instances and shared CNs
- OS type: Linux
- Others: Proxy main container has a finite memory limit; the pod also contains a sidecar

All deployment identifiers, network addresses, tenant/user identifiers, and credentials are intentionally omitted.

# Actual Behavior

Proxy memory grows almost linearly with the number of active client connections and eventually causes node memory pressure/OOM.

In the investigated incident:

- Each Proxy grew from roughly 250 MiB RSS to about 5.9 GiB.
- Each Proxy had roughly 3,000 active connections before OOM.
- The two Proxies had 6,054 active connections in total.
- One tenant accounted for 5,799 of those connections; all sampled connections for that tenant came from one client endpoint and one database user.
- During normal periods, that client established about 241 connections/hour but cleaned up only about 121 connections/hour, producing a stable net increase of about 120 connections/hour.
- Sampled retained sessions completed authentication but emitted no subsequent SQL activity before the incident.
- The process had accepted more than two million connections cumulatively, but only about 6,100 session allocations remained live. This rules out a leak proportional to historical connection count.

The strongest allocation correlation was:

```text
active proxy connections:       ~3,055
session in-use objects:         ~6,112
session off-heap allocation:    ~6,112 MiB

6,112 ~= 2 * 3,055
```

Each proxied connection owns two frontend `IOSession` objects:

1. client -> Proxy: `pkg/proxy/client_conn.go`
2. Proxy -> CN: `pkg/proxy/server_conn.go`

Each `frontend.NewIOSession` eagerly allocates a fixed 1 MiB buffer in `pkg/frontend/mysql_buffer.go`.

The sidecar used only tens of MiB. The OOM was node-level; the sidecar was selected as a victim because it lacked resource protection, while the Proxy main container was the dominant memory consumer.

# Expected Behavior

A sustained increase in idle client connections must not lead to deterministic Proxy/node OOM.

Proxy should have all of the following properties:

- A bounded global and per-tenant active-connection budget.
- Configurable idle-connection reclamation for the tunnel phase.
[...]

PR #25857, opened 2026-07-17
refactor(frontend): make protocol buffers configurable

XuPeng-SH:
## What type of PR is this?

- [x] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

Part of #25856.

This PR adds backward-compatible, opt-in controls for MySQL protocol memory:

- configure the retained `IOSession` buffer size without changing the existing 1 MiB default;
- inject a concurrency-safe shared allocator so multiple sessions can share one bounded memory budget;
- configure a logical packet limit for `IOSession` readers;
- reject oversized packets in `sqlCodec` from the 4-byte header, before waiting for or decoding the payload;
- make fixed/dynamic buffer growth use the actual configured allocation unit;
- make `err-check` diagnostics shell-safe by selecting the first match in Make instead of executing grep output as shell syntax.

Existing `NewIOSession` and `NewSqlCodec` callers retain their historical behavior. Tests cover exact-buffer boundaries, packets larger than the retained buffer, dynamic-buffer cleanup, aggregate allocator exhaustion/recovery, header-time rejection, boundary acceptance, and default compatibility.

### DIFF
--- a/pkg/frontend/mysql_buffer.go
+++ b/pkg/frontend/mysql_buffer.go
@@ -56,6 +56,43 @@ type BufferAllocator struct {
 	allocator Allocator
 }
 
+// IOSessionOption customizes the memory retained by an IOSession. Options are
+// intentionally applied only by NewIOSessionWithOptions so existing frontend
+// callers keep the historical defaults.
+type IOSessionOption func(*ioSessionOptions)
+
+type ioSessionOptions struct {
+	bufferSize        int
+	allowedPacketSize int
+	allocator         Allocator
+}
+
+// WithIOSessionBufferSize sets the size of the buffer retained for the whole
+// session lifetime. Packets larger than this buffer continue to use the
+// existing dynamic read/write paths.
+func WithIOSessionBufferSize(size int) IOSessionOption {
+	return func(opts *ioSessionOptions) {
+		opts.bufferSize = size
+	}
+}
+
+// WithIOSessionAllowedPacketSize bounds a single logical MySQL packet. This is
+// useful for protocol endpoints, such as a Proxy login path, that never need
+// the frontend server's general-purpose packet maximum.
+func WithIOSessionAllowedPacketSize(size int) IOSessionOption {
+	return func(opts *ioSessionOptions) {
+		opts.allowedPacketSize = size
+	}
+}
+
+// WithIOSessionAllocator uses allocator for all buffers owned by the session.
+// The allocator must be safe for concurrent use when shared by sessions.
+func WithIOSessionAllocator(allocator Allocator) IOSessionOption {
+	return func(opts *ioSessionOptions) {
+		opts.allocator = allocator
+	}
+}
+
 func (ba *BufferAllocator) Alloc(size int) ([]byte, error) {
 	if size == 0 {
 		return nil, nil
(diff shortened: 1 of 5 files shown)


pkg/frontend/mysql_buffer.go
```diff
@@ -204,20 +245,58 @@ func (c *Conn) SetTimeout(d time.Duration) {
...
-		allocator:             &BufferAllocator{allocator: getSessionAlloc(service)},
+		allocator:             &BufferAllocator{allocator: opts.allocator},
 		timeout:               pu.SV.SessionTimeout.Duration,
 		readTimeout:           pu.SV.NetReadTimeout.Duration,
 		writeTimeout:          pu.SV.NetWriteTimeout.Duration,
 		loadLocalReadTimeout:  pu.SV.LoadLocalReadTimeout.Duration,
 		loadLocalWriteTimeout: pu.SV.LoadLocalWriteTimeout.Duration,
 		maxBytesToFlush:       int(pu.SV.MaxBytesInOutbufToFlush * 1024),
-		allowedPacketSize:     int(MaxPayloadSize),
+		allowedPacketSize:     opts.allowedPacketSize,
```
LeftHandCold, line 299: [P1] `WithIOSessionAllowedPacketSize` does not currently bound all IOSession reads. `ReadLoadLocalPacket` parses `packetLength` from the header and immediately allocates/reads that many bytes without calling `CheckAllowedPacketSize`, so a session configured with (for example) a 16-byte limit still accepts a 17-byte LOAD DATA packet. I reproduced this with `WithIOSessionBufferSize(16)` and `WithIOSessionAllowedPacketSize(16)`: `ReadLoadLocalPacket()` returned the full payload with `err == nil`. This bypasses the advertised logical-packet bound and can consume the shared allocator budget before rejection. Please check the header length before the allocation/read and cover this reader in the option tests.

LeftHandCold (changes requested):
The fixed/dynamic buffer ownership paths and the sqlCodec header-time rejection look sound, but the new IOSession packet limit is not enforced by every IOSession reader. `ReadLoadLocalPacket` still trusts the 3-byte header and allocates/reads that size directly. Please enforce the configured limit there before allocation and add the corresponding regression test.

pkg/frontend/mysql_buffer.go
XuPeng-SH: Fixed in 6788d7c78f. `ReadLoadLocalPacket` now applies the configured packet limit immediately after parsing the header and before any payload allocation/read. The regression matrix covers the exact boundary, limit+1 rejection with the payload left unread and no new allocation, and limit+1 after a reusable LOAD LOCAL buffer already exists to verify that error cleanup releases the prior allocation. I also enumerated the IOSession logical read entry points: `Read` and `ReadLoadLocalPacket` now both enforce the same bound before dynamic payload acquisition.

LeftHandCold (approved):
Re-reviewed current head 6788d7c78f. The previous blocking issue is fixed: ReadLoadLocalPacket now applies the configured limit immediately after parsing the header and before payload allocation/read, with regression coverage for the exact boundary, limit+1 leaving the payload unread without a new allocation, and cleanup of an existing reusable LOAD LOCAL buffer. I also rechecked ordinary Read multi-packet accumulation, sqlCodec header-time rejection, fixed/dynamic/LOAD LOCAL buffer ownership, shared allocator exhaustion/recovery, and default compatibility. Local validation passed: full pkg/frontend tests, focused tests under -race -count=10, go build, go vet, and make err-check. No remaining blockers from me.

PR #25858, opened 2026-07-17
fix(proxy): bound connection memory and admission

### LINKED ISSUE #25856: [Bug]: Proxy deterministically runs out of memory as idle connections grow without bounds

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25856.

Depends on #25857. This is a stacked PR and should be retargeted to `main` after #25857 merges.

The Proxy previously paid two eager 1 MiB `frontend.IOSession` buffers per live proxied connection, created one managed allocator per connection, retained an unbounded login packet for migration, and had no global or per-tenant admission boundary. Raising a memory limit alone would only delay exhaustion.

This PR closes the resource lifecycle at multiple layers:

- reduces the two retained Proxy protocol buffers from about 2 MiB to 32 KiB per active connection (16 KiB on each side);
- shares one size-bounded protocol allocator across client, backend, migration, and cached backend sessions;
- adds global and per-tenant live-connection admission with idempotent leases;
- rejects global overload before per-connection protocol allocation with MySQL error 1040;
- caps retained client login packets at 64 KiB and rejects them from the packet header before decoding;
- bounds the complete unauthenticated handshake lifecycle, including TLS, and clears the read deadline before entering the raw tunnel;
- rejects duplicate PROXY protocol headers so they cannot extend the handshake or grow recursion;
- validates connection, protocol-memory, handshake-size, and handshake-timeout configuration as one coherent budget;
- records expected overload as reject metrics without error-log amplification.

The SQL/tunnel data path is unchanged: the limiter lock is touched only during acquire, tenant bind, and release, and no query/transaction idle deadline is introduced. This avoids regressions for long HTAP queries.

## Test matrix

- full `./pkg/frontend` and `./pkg/proxy` unit suites;
- `go build ./pkg/frontend ./pkg/proxy`;
- `go vet ./pkg/frontend ./pkg/proxy`;
- new boundary/concurrency tests repeated 100 times;
- full `./pkg/proxy` race suite;
- focused frontend race tests.

Covered unhappy paths include allocator exhaustion and recovery, exact-fill and oversized packets, global and tenant saturation, concurrent bind/release, double release, constructor/TLS failure cleanup, silent-client timeout, deadline clearing after successful login, standard client-visible errors, and bounded rejection writes.


pkg/proxy/client_conn.go
```diff
@@ -776,22 +864,36 @@ func (c *clientConn) sendPacketToClient(r []byte, sc ServerConn) error {
...
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return nil, context.DeadlineExceeded
 		}
-		return c.readPacket()
-	}
-	packet, ok := msg.(*frontend.Packet)
-	if !ok {
-		return nil, moerr.NewInternalError(c.ctx, "message is not a Packet")
+		msg, err := c.conn.Read(goetty.ReadOptions{Timeout: remaining})
```
LeftHandCold, line 1244: [P1] Enforce an absolute deadline across fragmented handshake reads. goetty treats ReadOptions.Timeout as a relative timeout and re-applies time.Now().Add(timeout) for each internal socket read. Because this call passes the same remaining value throughout one c.conn.Read, a client can send one byte before each timeout and renew the deadline indefinitely. I reproduced this with a 30 ms handshake timeout: sending one byte every 20 ms was still blocked after 75 ms. Such an unauthenticated client can therefore retain the global lease indefinitely and eventually deny all new connections. Please enforce the lifecycle deadline independently of goetty per-read timeout renewal, and add a fragmented slow-client test.

pkg/proxy/client_conn.go
```diff
@@ -230,11 +266,31 @@ func newClientConn(
 	fp.SetDefaultValues()
 	pu := config.NewParameterUnit(&fp, nil, nil, nil)
 	frontend.InitServerLevelVars(cfg.UUID)
-	frontend.SetSessionAlloc(cfg.UUID, frontend.NewSessionAllocator(pu))
-	ios, err := frontend.NewIOSession(c.RawConn(), pu, cfg.UUID)
+	allocator := frontend.Allocator(frontend.NewSessionAllocator(pu))
```
LeftHandCold: [P2] Avoid constructing the fallback allocator when the shared allocator is present. The right-hand side here is evaluated unconditionally, so every production client connection still creates a NewSessionAllocator and then immediately discards it. NewSessionAllocator initializes a ManagedAllocator with 256 maps; the same pattern in server_conn.go creates another one for every backend connection. This removes retention but preserves substantial allocation and GC churn during connection storms. Initialize allocator from c.sessionAllocator first and only call NewSessionAllocator in the nil fallback; apply the same change to newServerConn.

pkg/proxy/handler.go
```diff
@@ -224,6 +270,12 @@ func (h *handler) handle(c goetty.IOSession) error {
 		if isConnEndErr(err) {
 			return nil
 		}
+		if isProxyAdmissionError(err) {
+			v2.ProxyConnectRejectCounter.Inc()
+			h.logger.Debug("connection rejected during handshake", zap.Error(err))
+			cc.SendErrToClient(err)
+			return err
```
LeftHandCold: [P2] Do not return expected admission failures to the goetty application after the client response is sent. goetty logs every non-nil handler result as ERROR with handle session failed, so tenant-limit and oversized-packet rejections generate one error log per attempt despite the local Debug log. The global-limit return at line 221 has the same issue. After writing the bounded MySQL error, return nil so overload does not create error-log amplification.

pkg/proxy/handshake.go
```diff
@@ -34,8 +35,12 @@ func (c *clientConn) writeInitialHandshake() error {
 // handleHandshakeResp receives login information from client and saves it
 // in proxy end.
 func (c *clientConn) handleHandshakeResp() error {
+	return c.handleHandshakeRespBefore(time.Now().Add(c.clientHandshakeTimeout))
+}
+
+func (c *clientConn) handleHandshakeRespBefore(deadline time.Time) error {
 	// The proxy reads login request from client.
-	pack, err := c.readPacket()
+	pack, err := c.readPacketBefore(deadline)
```
LeftHandCold: [P2] Preserve the MySQL packet sequence when rejecting an oversized login header. readPacketBefore returns ErrPacketTooLarge before a Packet is produced, so AddSequenceId(1) below is never reached. After the server handshake uses sequence 0 and the client sends login sequence 1, SendErrToClient consequently emits ERR sequence 1 instead of 2. I reproduced the exact sequence mismatch. Please advance the sequence for this header-time rejection, including the post-TLS login case, and cover it with a wire-level test.

LeftHandCold (changes requested):
Requesting changes for one availability bypass and three concrete overload/protocol regressions. I reviewed only the stacked delta from da2e56cf9a3f82c6bda114c201bb100a34e5903b to f259642ceb5dbe7d814edeadc1daf7688491a7e2. The full proxy test suite, focused race tests, build, and vet pass; additional focused diagnostics reproduced the absolute-deadline bypass and the incorrect ERR packet sequence.

pkg/proxy/handler.go
XuPeng-SH: Fixed in 2c57d308c7. Both global admission rejection and post-login tenant or packet rejection now return nil after the bounded MySQL error has been written. Cleanup is unchanged because goetty closes the session in its unconditional defer; this only prevents expected load shedding from being logged again as handle session failed. Added a handler-level regression for the global rejection result.

pkg/proxy/client_conn.go
XuPeng-SH: Fixed in 2c57d308c7. The handshake now installs an absolute-read-deadline wrapper that caps every downstream SetReadDeadline call at the lifecycle deadline, so goetty cannot renew it while decoding fragmented input. The cap spans PROXY-header and TLS/post-TLS login reads, is disabled after the handshake, and the wrapper is removed from non-TLS connections to avoid per-connection retention. Added a slow-byte regression that would outlive the outer bound with the previous implementation.

pkg/proxy/handshake.go
XuPeng-SH: Fixed in 2c57d308c7. The proxy codec now retains the header sequence when the frontend codec rejects an oversized payload before producing a Packet; readPacketBefore advances the protocol to header sequence plus one before the ERR response. Wire-level tests cover sequence 1 (initial login to ERR 2) and sequence 2 (post-TLS login to ERR 3).

pkg/proxy/client_conn.go
XuPeng-SH: Fixed in 2c57d308c7. Both newClientConn and newServerConn now select the supplied shared allocator first and call NewSessionAllocator only in the nil fallback. Production connections therefore no longer construct and discard the per-connection ManagedAllocator.

pkg/proxy/ppv2.go
```diff
@@ -80,7 +98,19 @@ func (c *proxyProtocolCodec) Decode(in *buf.ByteBuf) (interface{}, bool, error)
...
+	if hasPacketHeader {
+		sequenceID = in.PeekN(0, frontend.PacketHeaderLength)[3]
+	}
+	value, complete, err := c.Codec.Decode(in)
+	if hasPacketHeader && errors.Is(err, frontend.ErrPacketTooLarge) {
+		err = &mysqlPacketDecodeError{
+			cause:      err,
+			sequenceID: sequenceID,
+		}
+	}
```
LeftHandCold, line 133: [P1] Wait for a fragmented PROXY v2 header before applying the MySQL limit

With the new default 64 KiB cap, a valid PROXY v2 header is rejected whenever the first socket read contains only 4-15 bytes of its signature. In that state parseProxyHeaderV2 returns incomplete/non-PROXY and this delegates to SQLCodec; the first four signature bytes (0d 0a 0d 0a) are then decoded as a 0x0d0a0d (854541-byte) MySQL payload and immediately return ErrPacketTooLarge. I reproduced this by feeding only the first four signature bytes. Goetty returns after a single net.Conn.Read, so the decoder cannot assume that all 16 header bytes arrive together. Before this PR, the same bytes remained incomplete because 854541 was below the old limit, allowing the remaining PROXY header to arrive. Please keep a buffered prefix of ProxyProtocolV2Signature in the incomplete state until there are enough bytes to distinguish it, and add a regression test that feeds the prefix and the rest in separate reads. Ideally the test should also fragment the declared PROXY body, which must likewise be awaited before consuming the header.

LeftHandCold (changes requested):
Re-reviewed the latest head. The four findings from my previous review are fixed. Requesting changes for one newly confirmed PROXY v2 protocol regression. The full proxy/frontend test suites, the full proxy race suite, build, vet, and diff checks pass; a focused diagnostic reproduces this failure deterministically.

pkg/proxy/ppv2.go
XuPeng-SH: Fixed in a15e8e11c6. The PROXY v2 decoder now distinguishes definite non-PROXY input from an incomplete candidate. Matching signature prefixes, partial 16-byte headers, and partial declared bodies return incomplete without consuming ByteBuf data; the SQL codec is called only after a signature mismatch, and a complete frame is consumed once. The regression feeds the signature, fixed header, and IPv4 body in six separate fragments, checks zero consumption at every incomplete boundary, covers all 1-11 byte signature prefixes, and verifies mismatched prefixes still delegate to the bounded MySQL codec. Full proxy tests, full proxy race tests, go vet/build, and make static-check pass.

gouhongshen (changes requested):
## Codex automated review

The new admission and memory limits have a concrete OOM gap, and duplicate PROXY v2 detection is incomplete.

### P1 - Include retained handshake storage in the memory budget (pkg/proxy/config.go:348)
Validation accounts only for the shared 16 KiB client/backend session buffers. The goetty input buffer and the copied handshakePack.Payload are outside that allocator but remain retained for authenticated connections. With the defaults, 10,000 connections may each retain a valid 64 KiB login twice, already exceeding 1 GiB before the fixed session buffers and tunnel state, so ProtocolMemoryLimit does not prevent OOM.

### P2 - Mark consumed LOCAL/UNSPEC PROXY frames for duplicate detection (pkg/proxy/ppv2.go:194)
For valid unspec/unix PROXY v2 frames, the parser consumes the frame but returns ok=false. The wrapper then delegates to the SQL codec, and readPacketBefore never sets proxyHeaderReceived. A LOCAL/UNSPEC frame followed by an address-bearing frame in a later read therefore bypasses duplicate rejection and can overwrite origin metadata used for internal classification and routing.

LeftHandCold (approved):
Re-reviewed the latest head and approving. The previously reported fragmented PROXY v2 regression is fixed: matching signature prefixes, partial fixed headers, and partial declared bodies now remain buffered without consumption, while definite mismatches still delegate to the bounded MySQL codec. I found no new blocker in the updated delta. The focused fragmentation test passed 100 iterations; the full proxy suite, full proxy race suite, build, vet, and diff checks also pass.

XuPeng-SH:
Addressed both findings in c9d4ae3559.

- Retained handshake memory now has an explicit phase/ownership boundary. After the final login is accepted, the handshake-only goetty input buffer is released instead of remaining attached to the established tunnel. The login payload needed for migration is copied into the Proxy's shared size-bounded allocator, owned by `clientConn`, and released exactly once on `Close`. Configuration validation now includes one retained login payload per admitted connection (plus client/backend and cached session buffers), with overflow-safe arithmetic; the default 10,000-connection / 1 GiB budget remains valid.
- Addressless PROXY v2 frames (UNSPEC/UNIX) are now emitted as consumed protocol messages. `readPacketBefore` therefore records the first frame even without an address and deterministically rejects a later PROXY header before it can replace origin metadata.

Added regressions for a near-limit 64 KiB login lifecycle, allocator exhaustion, double close, exact memory-budget boundaries/overflow, and LOCAL/UNSPEC followed by an address-bearing frame. Local validation passed full frontend/proxy tests, 100x focused stress, full proxy race plus focused race, build, vet, and `make static-check`.

gouhongshen (changes requested):
## Codex automated review

TLS connections can panic the proxy when data follows the login packet in the same TLS record.

### P1 - Do not close a TLS transport buffer that may contain the first tunneled packet (pkg/proxy/handshake.go:63)
After TLS upgrade, `c.RawConn()` is the `tls.Conn` backed by goetty's `BufferedConn`, whose reader first consumes `session.InBuf()`. A client can write its final login packet and a following MySQL command in one TLS application record; goetty decodes the login but leaves the command readable in `InBuf`. This code frees the ByteBuf while its reader/writer indices still describe that data. When `t.run` starts and the client pipe reads the TLS connection, `BufferedConn` reads that closed ByteBuf; `ByteBuf.Read` slices its now-nil backing slice at non-zero indices and panics. This is a process-crash path reachable by an authenticated client. Only release the input buffer after proving it is empty, or preserve/hand off its unread bytes to the tunnel. Add a TLS test that coalesces a login and a subsequent packet.

XuPeng-SH:
@gouhongshen Fixed in 6084083a5b. The P1 finding is valid.

The root problem was phase ownership, not TLS alone: the handshake decoder can read beyond the final login packet, while the old code freed its ByteBuf before transferring unread bytes to the tunnel. TLS could later touch stale indices and panic; plaintext silently lost the first pipelined command.

The fix now performs an explicit ownership handoff:
- empty handshake input is reset and released immediately, so the normal client path gets no wrapper or tunnel overhead;
- unread bytes are copied into the existing shared size-bounded allocator before the goetty ByteBuf is reset and closed;
- both the application session and MySQL protocol switch atomically to the handoff connection;
- the copied prefix is freed as soon as it is drained, or exactly once on Close;
- allocator exhaustion leaves the original owner intact so normal session cleanup remains balanced.

Regression coverage includes coalesced plaintext login + command, coalesced TLS login + command in one TLS write, handoff allocation failure, partial-read/Close ownership, repeated Close, and concurrent Read/Close. Validation passed:
- focused tests with -race
- full pkg/proxy suite
- full pkg/proxy suite with -race
- go build ./pkg/proxy
- go vet ./pkg/proxy
- make static-check
- git diff --check

XuPeng-SH:
Deep self-review found and fixed one additional P1 ownership race in 7032ba939a.

Root cause: this PR made the retained client login payload explicitly allocator-owned and freed it on `clientConn.Close`, but a concurrent backend connect/migration could still be using that payload. In particular, `serverConn.HandleHandshake` previously returned immediately on timeout while its worker remained alive, so Close could return the allocation to the shared allocator and another connection could reuse it while the worker was still authenticating. That is a use-after-free/data-race/auth-corruption path introduced by active memory reclamation.

The fix closes the lifecycle systemically:
- every synchronous backend connect now holds a read lease on the retained handshake packet;
- Close takes the exclusive side, marks the packet terminal, releases it exactly once, and cannot race an active reader;
- `GetHandshakePack` returns a protected snapshot rather than exposing allocator-backed storage past the lock;
- backend handshake timeout closes only the transport first, joins the worker, and returns only after all payload readers have stopped; the full `serverConn.Close` still owns protocol-buffer cleanup afterward.

This does not copy the login payload per migration and does not touch the established query tunnel hot path. The lock is used only for backend connection/authentication and connection teardown.

Fresh validation on the final code:
- full `pkg/proxy` suite
- full `pkg/proxy` suite with `-race`
- focused lifecycle/timeout/compatibility race tests
- 50x focused lifecycle/timeout race stress
- `go build -mod=readonly ./pkg/proxy`
- `go vet -mod=readonly ./pkg/proxy`
- `make static-check`
- `git diff --check`

XuPeng-SH:
Deep-review blocker (GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR):

[P1] Reject handshake packet limits that cannot admit a final login

`minimumClientHandshakePacketLimit` is currently 32 because a protocol 4.1 SSLRequest is 32 bytes, and `Config.Validate` therefore accepts limits of 32 or 33. However, 32 bytes is only the fixed protocol-4.1 prefix. A final authentication response must additionally contain at least the NUL-terminated username and an auth length/terminator. On TLS connections the 32-byte SSLRequest is followed by that larger final login, and the same SQL codec limit applies to both phases.

I reproduced this at the wire level on the current head: a config with `client-handshake-packet-limit = 32` passes `FillDefault + Validate`, then a valid login is rejected with `frontend.ErrPacketTooLarge`. Thus an explicitly accepted configuration deterministically makes the proxy unable to authenticate any real client.

Please model the minimum for a complete login rather than the SSLRequest-only size (or otherwise reject unusable limits), and add a boundary test that the smallest configuration accepted by `Validate` can pass the packet-size gate for a final protocol-4.1 login.

XuPeng-SH:
@gouhongshen Addressed the latest handshake-limit blocker and the related phase-model gaps in a569dfaf22; the branch is also merged with the latest main at 80668d99da.

Root-cause classification:
- The 32-byte minimum was a configuration-contract bug: 32 bytes describes only the intermediate protocol-4.1 SSLRequest, not a usable terminal login. The accepted minimum is now derived as 32-byte fixed prefix + one-byte username + NUL + one auth length/terminator byte (35 bytes), with exact min/max wire and validation boundaries.
- The recursive TLS path exposed a broader protocol-state-machine gap. Handshake processing is now an explicit bounded phase loop: pre-TLS -> at most one TLS upgrade -> final login. A repeated SSLRequest is rejected instead of recursively layering TLS/stack/CPU work.
- Validation now mirrors runtime composition: plugin routing disables the connection cache, so that mode no longer reserves 1,024 cache sessions that can never exist.

Unhappy-path ownership coverage now proves:
- the intermediate SSLRequest allocation is released before both successful and failed TLS negotiation;
- a partially parsed final login transfers cleanup ownership and is released exactly once by Close;
- duplicate TLS requests are rejected promptly without recursion;
- plaintext and TLS logins at the exact accepted minimum complete parsing;
- allocator, admission, timeout, pipelined-prefix, double-close, and concurrent reader/Close paths remain covered.

Performance: the production delta is limited to startup validation and the unauthenticated handshake. The recursion-to-loop change adds only a stack boolean and performs the same packet parse/allocation work for the normal one- or two-packet flow. It does not touch the established tunnel/query hot path.

Fresh validation after merging latest main:
- focused phase/config tests: 20x
- full pkg/proxy
- full pkg/proxy with -race
- go build -mod=readonly ./pkg/proxy
- go vet -mod=readonly ./pkg/proxy
- make static-check
- git diff --check

During the unhappy-path test addition, it also exposed and fixed the existing "TSL handshake error" typo so production diagnostics now consistently say TLS.

XuPeng-SH:
Deep self-review found and fixed one additional admission trust-boundary issue in 584ff055c5.

Root cause:
- The global slot is intentionally acquired before expensive per-connection state, but the per-tenant slot was previously bound as soon as the login packet revealed a tenant name.
- Before backend authentication that name is only an untrusted client claim. An unauthenticated client could repeatedly claim another tenant and occupy the target tenant quota for the backend-auth window, causing targeted admission denial without owning any credentials.

The lifecycle is now:

`global admission -> parse claimed tenant -> authenticate via fresh backend or cache authenticator -> bind tenant admission -> expose backend OK to client`

This preserves the global memory/work bound during authentication while making tenant fairness depend on an authenticated identity. On tenant saturation, a freshly authenticated backend or cache hit is closed before any OK packet reaches the client. Backend authentication failures consume no tenant quota. Expected admission rejection remains debug-level and uses the existing reject metric, avoiding error-log amplification.

Added regression coverage for:
- two unauthenticated connections claiming the same tenant without consuming its authoritative quota;
- fresh-backend success binding before client OK;
- fresh-backend tenant rejection closing the backend before client OK;
- backend auth failure leaving tenant quota untouched;
- authenticated cache-hit rejection and cleanup.

Validation on the pushed commit:
- focused admission/lifecycle tests: 100x;
- focused race tests: 20x;
- full `pkg/proxy`;
- full `pkg/proxy` with `-race`;
- `go build -mod=readonly ./pkg/proxy`;
- `go vet -mod=readonly ./pkg/proxy`;
- `make static-check`;
- `git diff --check`.

Performance remains outside the established tunnel/query hot path: one existing limiter mutex operation moves from pre-auth parsing to the successful authentication boundary; rejected unauthenticated claims no longer contend on per-tenant counters.

LeftHandCold (approved):
Re-reviewed the latest head  carefully and approving.\n\nI re-audited the new handshake lifecycle and admission boundary from first principles:\n- retained auth payload ownership is closed: backend connects hold a read lease, timeout closes the transport and joins the worker before returning, and client Close frees the allocator-backed payload exactly once;\n- coalesced plaintext/TLS bytes are handed to the tunnel without loss or stale ByteBuf access;\n- SSLRequest is phase-local, duplicate TLS requests are bounded/rejected, and the final login remains subject to the absolute handshake deadline and packet limit;\n- per-tenant admission is bound only after fresh-backend or cache authentication and before the backend OK is exposed; rejection closes the accepted backend and does not consume unauthenticated tenant quota;\n- PROXY v2 fragmentation/LOCAL/UNSPEC handling, oversized ERR sequencing, shared-memory accounting, cache and migration paths have no new reachable regression found.\n\nFresh validation on this exact head:\n- focused lifecycle/admission/TLS/PROXY/timeout tests: 10x;\n- focused ownership tests with : 10x;\n- full FAIL	github.com/matrixorigin/matrixone/pkg/proxy [build failed]
FAIL;\n- full FAIL	github.com/matrixorigin/matrixone/pkg/proxy [build failed]
FAIL;\n- ;\n- ;\n- dependent frontend SQL codec test;\n- .\n\nNo new realistic blocker found.

LeftHandCold (approved):
Correction to my immediately preceding approval: its Markdown command names were expanded by the local shell while submitting the review body. The two displayed build-failed lines were submission-formatting artifacts, not validation results.

This approval is for head 584ff055c5753694f65a9b2b9ec3a886007c208b. The actual fresh validation all passed:
- focused lifecycle, admission, TLS, PROXY, timeout, and packet-sequence tests, 10 iterations;
- focused ownership tests under the race detector, 10 iterations;
- full pkg/proxy test suite;
- full pkg/proxy suite under the race detector;
- pkg/proxy build and vet;
- dependent frontend SQL codec test;
- diff check.

The code-review conclusion is unchanged: retained handshake ownership, timeout worker termination, pipelined plaintext and TLS handoff, authenticated tenant binding, packet bounds, cache and migration interactions form a closed lifecycle. I found no new realistic blocker on this head.

gouhongshen (changes requested):
## Codex automated review

Valid configurations can fail normal migration/control-plane operations because peak shared allocations exceed the validated budget. The fragmented PROXY-v2 path also bypasses the configured handshake memory limit.

### P1 - Account for the transient third protocol session (pkg/proxy/config.go:357)
Validation reserves only `maxConnections*2` fixed sessions, but migration allocates the replacement backend session before `replaceServerConn` closes the old one (`tunnel.go:491-496`, `tunnel.go:371-374`). The same third-session peak occurs for `connAndExec` during upgrade/kill. A configuration accepted by the exact-boundary test can therefore reach the allocator limit and return `out of space` during legitimate operations. Reserve transient headroom or release/coordinate the old session before allocating its replacement.

### P2 - Bound fragmented PROXY-v2 bodies by the handshake budget (pkg/proxy/ppv2.go:158)
The parser now waits for the complete declared body before delegating to the SQL codec. `Length` is a uint16, so an unauthenticated client can send an UNSPEC frame with a 65535-byte body and make goetty retain roughly 64 KiB outside the shared allocator; line 175 also creates a second temporary copy once complete. `ClientHandshakePacketLimit` can legally be as low as 35, but it is never applied to this frame, so the new fragmented-header path bypasses the configured handshake memory bound until timeout.

aptend (changes requested):
Re-review of head 584ff055c5: the protocol-memory bound is still not closed, and the fragmented PROXY-v2 path still bypasses it.

[P1] Account for all peak shared protocol allocations

Config.Validate accepts an exact budget of two 16 KiB IOSessions plus one retained login payload per admitted connection. That is below several reachable peaks. A valid 32 KiB login fills the accepted client + retained payload + backend budget, then forwarding that login to the backend needs an additional dynamic write block and can fail with out of space. Established connections also need a third IOSession while migration creates the replacement backend before closing the old one; connAndExec has the same peak for upgrade/kill operations. Pipelined handshake read-ahead adds another allocator-backed prefix that is also omitted. Consequently, configurations accepted by validation can reject normal login, migration, and control-plane work. Please reserve the complete peak or add explicit admission/reservation that guarantees these operations can allocate.

[P2] Bound PROXY-v2 bodies within the handshake memory budget

parseProxyHeaderV2 waits for the complete uint16-declared body before the SQL codec runs, so an unauthenticated client can retain up to 65,535 bytes in goetty input memory that is outside the shared SessionAllocator and unaffected by ClientHandshakePacketLimit. Once complete, the parser also allocates and copies another full body. This violates ProtocolMemoryLimit even when the configured handshake limit is only 35 bytes. Please reject bodies above an explicit accounted limit, or bring the goetty input allocation and parser copy under the same shared bound.

XuPeng-SH:
@gouhongshen @aptend Both findings were valid and are addressed in 15d8f921c and 7e619b7a5. The branch is also merged with latest main at 5b6234704.

Root cause and reflection

The previous validation modeled only long-lived allocations: client IOSession + backend IOSession + retained login. That is necessary but not sufficient. It did not model phase transitions, where a replacement or temporary backend coexists with the steady backend, login forwarding can allocate a rounded dynamic write block, and pipelined read-ahead can outlive authentication. The PROXY v2 decoder was also assumed to inherit the SQL codec limit, but it waits before delegating, so that assumption was false.

The corrected model is phase-based:

- steady = max-connections * (2 * 16 KiB IOSession + handshake limit) + reachable cache sessions * 16 KiB;
- initial transient = bounded pipelined prefix + exact rounded dynamic login write;
- migration/control transient = third IOSession + the larger sequential dynamic write reachable from login forwarding or the fixed event capture window;
- validation requires steady plus at least one complete transient operation;
- remaining headroom is enforced at runtime with byte-weighted admission, so concurrent valid transitions cannot race for the same allocator capacity. Small normal operations reserve only their real phase upper bound instead of a coarse worst-case token.

Ownership and cancellation are explicit:

- initial authentication owns the lease through backend auth; a pipelined prefix retains it until drained or Close;
- migration transfers the lease to the new ServerConn, and the tunnel releases it only after the old backend is closed and all pointers switch;
- connect/auth/migrate/control failures release through one cleanup edge;
- transfer context now propagates into admission, so cancellation interrupts a saturated-headroom wait instead of adding another independent 10 second wait.

PROXY v2 now rejects a declared body above ClientHandshakePacketLimit immediately after the 16-byte fixed header, before waiting for the body. Complete accepted bodies are parsed directly from the goetty ByteBuf, removing the second full-body allocation/copy. Pipelined handshake read-ahead is bounded by the same configured limit before entering the shared allocator.

I also checked the tempting 16 MiB theoretical control-statement case. It is not reachable in this architecture: only complete statements inside the fixed 8 KiB + 2 KiB msgBuf event window become migration/control events; larger packets are forwarded and never enter replay state. The budget ties to that actual window, avoiding a configuration and concurrency regression from over-reserving a theoretical path.

Regression coverage now includes exact budget minus-one/exact boundaries, weighted saturation and reuse, cancellation while saturated, prefix lease retention until drain, failed/double Close, migration promotion, oversized PROXY rejection from a fragmented fixed header, accepted boundary bodies, no body copy path, and oversized pipelined read-ahead cleanup.

Fresh validation on final head 7e619b7a5:

- full pkg/proxy suite;
- full pkg/proxy suite with race detector;
- focused lifecycle, cancellation, handshake, and PROXY tests, 100 iterations;
- CGo-contract build/compile;
- scoped go vet with MatrixOne CGo flags;
- full make static-check: 0 final issues;
- git diff --check.

Established query forwarding remains unchanged: admission runs only during initial connection setup, migration, and temporary control connections. The normal no-read-ahead path creates no wrapper; PROXY parsing removes a copy.

XuPeng-SH:
Deep re-review found and fixed one additional P1 availability/ownership gap in f81373fe5.\n\nRoot cause and reachable shape\n\nThe weighted protocol-memory limiter previously used the transfer context only while waiting for admission. After a transient lease was acquired, route/dial/authentication, cache validation, and migration replay fell back to the long-lived client context. A replacement CN can accept authentication and then stall on the first migration statement, SET transferred=1. frontend ExecStmt refreshes its normal session read timeout, whose default is 24 hours. With a valid exact-headroom configuration that admits one transient backend, this single stalled replay can retain all transient capacity for 24 hours and make later login, migration, and control operations queue or fail admission.\n\nThe lifecycle is now closed end to end:\n\n- BuildConnWithServer propagates its context through routing, dial, backend authentication, migration RPC, and every replay statement.\n- Backend dial uses DialContext. Authentication and simple control statements register cancellation against the phase-owned transport; cancellation closes that not-yet-delivered backend and joins the callback before ownership can move into the tunnel.\n- A cancellation racing with successful connect cannot publish the backend: the connection is closed and the transient lease is released.\n- Cached SET CONNECTION ID validation has the same bounded context. Optional context methods are explicitly forwarded through cache and protocol-memory wrappers, avoiding Go interface-method hiding.\n- Migration and kill are bounded, while UPGRADE ACCOUNT ALL retains its legitimate long-running semantics and is canceled only by its event/tunnel lifecycle.\n- No transport wrapper, deadline shim, atomic check, or allocation is added to established query forwarding. The returned steady-state backend is the original net.Conn.\n\nNew unhappy-path coverage proves post-admission cancellation and immediate limiter reuse, cancellation winning the connect/handoff race, migration replay cancellation, cached validation through stacked wrappers, interruption of frontend default 24-hour reads, callback ownership handoff, allocator balance, and backend close behavior.\n\nFresh validation on exact head f81373fe5:\n\n- focused cancellation/ownership tests under race: 50 iterations;\n- focused handler/upgrade/kill tests under race: 20 iterations;\n- full pkg/proxy suite;\n- full pkg/proxy suite under race;\n- full make static-check: 0 issues;\n- git diff --check.

gouhongshen (approved):
## Codex automated review

No correctness regressions found. The ownership, admission, cancellation, memory-budget, migration, cache, TLS, and PROXY-v2 paths form a coherent lifecycle. Focused tests could not compile locally because required CGo headers (`usearch.h`, `roaring.h`, `xxhash.h`) are unavailable.

aptend (changes requested):
Deep review of head f81373fe5e8cae93c0409d2f9b0e5282a4ab0627 found two memory/availability blockers and one protocol/configuration regression.

### P1 - Bound the outer backend Goetty protocol buffer

`CNServer.ConnectContext` creates the outer backend Goetty session with an uncapped `frontend.NewSqlCodec()` and the default Go heap allocator (`pkg/proxy/server_conn.go:545-549`). Backend handshake packets are read through that session (`server_conn.go:466-476`), while `calculateProtocolMemoryBudget` accounts only for the inner frontend session and transient dynamic writes.

A CN can send a wire-valid MySQL packet near the 16 MiB payload maximum. The Goetty `ByteBuf` grows to that size and retains the capacity for the connection or cache lifetime, entirely outside `SessionAllocator`, `GuestMmuLimitation`, and the weighted admission semaphore. Repeating this over active connections can exceed `ProtocolMemoryLimit` by orders of magnitude and OOM the Proxy.

Please cap and account the outer backend read buffer, for example by using a bounded/shared allocator plus an explicit backend-handshake payload limit, and add an oversized backend handshake regression.

### P1 - Do not let an unbounded UPGRADE hold global transient admission indefinitely

The event handler passes its connection-lifetime stopper context into `HandleEvent` (`pkg/proxy/handler.go:332-337`). `handleUpgradeEvent` forwards that context to `connAndExec` (`client_conn.go:782-818`), which acquires a global `backendBytes` lease (`client_conn.go:584-591`) and holds it through `ExecStmtContext` (`client_conn.go:614`). That execution has no operation deadline; it only stops when the long-lived event context is canceled.

If a CN authenticates the temporary connection but never returns a terminal packet for `UPGRADE ACCOUNT ALL`, the lease remains held until the client disconnects. With a valid exact-headroom configuration, one such operation consumes all transient capacity, so later logins, migrations, and control operations wait and then fail admission despite available connection slots. This starvation is introduced by the new global lease.

If upgrade must remain long-running, reserve independent login/migration capacity or split the limiter; otherwise bound the upgrade operation. Add a stalled-CN test that proves a second admission can still make progress.

### P2 - The accepted minimum packet limit rejects standard IPv6 PROXY-v2 frames

`minimumClientHandshakePacketLimit` is 35 and validation accepts that value (`pkg/proxy/config.go:68-74,336-339`). `NewServer` also uses it as `WithProxyProtocolMaxBodySize` (`server.go:121-123`). A standard TCP/IPv6 PROXY-v2 address body is 36 bytes: 16-byte source, 16-byte destination, and two 2-byte ports. Therefore a configuration explicitly accepted by validation deterministically rejects every such frame at `ppv2.go:172-175` before login.

Please separate the login and PROXY body limits, or make the accepted minimum cover the supported PROXY families, with a server-level IPv6 boundary test.

Fresh validation on this exact head passed the full `pkg/proxy` suite, the full race suite, focused lifecycle tests under `-race -count=20`, build, vet, the frontend codec regression, `make static-check`, and `git diff --check`. These tests do not cover the three failure paths above.

XuPeng-SH:
Addressed the three findings in 79e014bb36 and merged the latest main in 163f0ae858.

1. Backend buffer ownership: removed the redundant backend Goetty protocol session. ExtraInfo is written directly to the raw transport and all backend MySQL reads now use the existing frontend session with a 64 KiB packet cap and the shared allocator. Authentication packets that can become connection-lifetime cache state have a separate 1 KiB payload cap. The budget includes the retained header and both temporary control-response copies. Added an oversized backend packet regression.
2. UPGRADE starvation: split transient admission into bounded critical and background lanes whose capacities still sum to the configured headroom. Login, migration, and kill use the critical lane; long-running UPGRADE uses a one-operation background lane. A stalled upgrade can no longer consume critical admission. Added both limiter lifecycle coverage and a stalled contextual execution test.
3. IPv6 PROXY v2 boundary: introduced an independent proxy-protocol-body-limit with an IPv6-capable 36-byte minimum and separate MySQL login validation. The default is 4 KiB, the wire maximum is 65535 bytes, and pre-auth PROXY plus login memory is included in the per-connection budget. Added a server-codec IPv6 minimum-boundary test.

Fresh validation after the latest-main merge:
- full pkg/proxy suite
- full pkg/proxy race suite
- focused regressions repeated 10 times, including race coverage
- dependent frontend SQL codec test
- proxy and frontend build and vet
- make static-check
- git diff --check

Reflection: the prior model followed the shared allocator but not every wire-buffer owner and lifetime class, and it treated all transient work as one business priority. The correction is to derive admission from protocol state, ownership duration, peak overlap, and progress priority rather than patching individual packet sizes or adding an arbitrary UPGRADE timeout.

aptend (changes requested):
Deep re-review of head 163f0ae858270ae90a8988726c3152a8beb8874d found one remaining memory-bound blocker.

### P1 - Acquire protocol-memory admission before materializing the login copies

`calculateProtocolMemoryBudget` assumes that the retained login copy replaces the Goetty input and therefore reserves only one pre-auth/login-sized state per admitted connection (`pkg/proxy/protocol_memory.go:95-110`). The actual ownership overlap is larger: `SQLCodec.Decode` calls `ReadMarkedData`, which copies the payload out of Goetty input (`pkg/frontend/codec.go:91-99`); then `handleHandshakeRespBefore` allocates and copies another retained payload (`pkg/proxy/handshake.go:130-150`), while the Goetty input backing buffer remains allocated until `handoffHandshakeBuffer` closes it.

The transient lease is acquired only after `handleHandshakeRespBefore` returns (`pkg/proxy/handshake.go:54-70`), so every admitted client can reach this three-owner peak concurrently without weighted admission. With the defaults, the steady calculation already consumes about 1.034 GB of the 1 GiB limit; just one omitted 64 KiB copy across 10,000 connections adds about 625 MiB, and this path has two additional login-sized owners at peak. The new limit can therefore still be exceeded enough to OOM the Proxy under synchronized near-limit logins.

Please acquire a weighted phase lease before reading/copying the terminal login and hold it through release of the Goetty input, or include the complete concurrent overlap in the budget. Add a synchronized near-limit login boundary test that covers all live owners.

Fresh validation on this exact head passed `make thirdparties`, `make cgo`, the full `pkg/proxy` suite, focused ownership/memory tests under `-race -count=10`, and `git diff --check`. Current CI is green but does not exercise this concurrent peak.

gouhongshen (changes requested):
## Codex automated review

The configured protocol-memory limit can be exceeded by steady tunnel allocations, and initial-handshake cancellation does not honor the supplied context. Focused tests were blocked by the available Go 1.23 toolchain versus go.mod’s Go 1.26.4 requirement.

### P1 - Account for tunnel buffers in the protocol-memory budget (pkg/proxy/protocol_memory.go:91)
The budget counts only frontend IOSession buffers, but every established tunnel allocates two 10,240-byte msgBufs and a 65,536-byte bufio writer outside the shared allocator and semaphore (pkg/proxy/mysql_conn_buf.go:132-142 and pkg/proxy/tunnel.go:697-701). At the default 10,000 connections this adds about 860 MB beyond the 1 GiB configured limit, so the claimed global memory bound can still allow OOM.

### P2 - Honor cancellation during the initial client handshake (pkg/proxy/handshake.go:54)
BuildConnWithServer now accepts a context, but handleHandshakeResp runs handleHandshakeRespBefore before checking it. readPacketBefore uses only the absolute clientHandshakeTimeout, and TLS upgrade uses a separate background context, so a canceled caller cannot interrupt a silent client or TLS handshake and retains the connection/admission slot until the handshake timeout.

XuPeng-SH:
Fixed the latest review findings in b330c8e5c0.

What changed:
- Protocol admission is now acquired before the first client-handshake read, so the Goetty input, SQLCodec-decoded payload, retained login copy, and pipelined prefix are covered by one phase reservation. A synchronized exact-capacity regression proves that a second buffered login cannot be decoded before admission.
- Initial plain and TLS handshakes now use the caller context as a terminal lifecycle signal. Cancellation closes the phase-owned transport to interrupt I/O, releases the lease, and joins/stops the callback before successful handoff so a late cancellation cannot close the tunnel.
- The steady budget now includes the concrete per-tunnel Go-heap cost: two 10 KiB message buffers plus the 64 KiB server-to-client writer. The shared allocator ceiling subtracts this external steady reservation, so the total limit is not counted once for tunnels and again for managed buffers.
- The default total protocol budget is 2 GiB, derived to preserve the existing 10K connection target after the newly accounted ~860 MiB tunnel memory; configurations below steady + required transient peak fail validation.
- Migration accounts for the replacement backend message buffer and now transfers the existing writer in both sync and non-sync modes, removing a 64 KiB allocation/GC cycle per non-sync migration. Phase formulas use the maximum simultaneous ownership peak rather than summing non-overlapping lifetimes.

Reflection: the missed cases came from accounting by allocator/component and treating timeout as cancellation. The corrected model is phase × concurrently-live owners × ownership-transfer boundary: acquire before materialization, interrupt every blocking dependency from the caller context, and stop/join cancellation before handoff.

Validation on the final diff:
- full `pkg/proxy` suite
- full `pkg/proxy` suite with `-race`
- focused lifecycle/memory/migration tests with `-race -count=10`
- `make static-check`
- `git diff --check`

XuPeng-SH:
Reviewed the latest head again. I did not find a new merge blocker beyond the issues already raised in earlier review rounds.

The latest delta looks to have closed the previously reported proxy lifecycle and admission problems: the protocol-memory / handshake ownership model is coherent on the current head, duplicate PROXY/TLS/handshake boundaries are bounded, and I do not see a new correctness, memory-lifecycle, or unhappy-path regression on top of the earlier review history.

I cannot mark this PR approved because GitHub does not allow approving your own pull request.

aptend (changes requested):
Deep re-review of head 14fefd36dc found two remaining memory/resource lifecycle blockers.

### P1 - Do not publish a replacement backend after the tunnel has closed

pkg/proxy/tunnel.go:356-397 installs newSC without checking whether Close has already completed, while Close at lines 615-640 records no terminal state under t.mu. A transfer can finish building and authenticating a replacement, then race with a client disconnect: Close cancels the tunnel, closes the current backend, and completes closeOnce; afterward replaceServerConn installs the replacement into the already-closed tunnel. Future Close calls are no-ops, and the non-cache handler cleanup only closes the original sc. The replacement socket and its connManager entry therefore lose their cleanup owner. Repeated disconnects during migration can exhaust backend connections and file descriptors.

Please linearize terminal close under t.mu and make replacement publication reject and close newSC when the tunnel is already closed.

### P1 - Acquire global admission before constructing the Goetty application session

pkg/proxy/server.go:133-149 constructs a Goetty session for every accepted socket before pkg/proxy/handler.go:221 acquires the global connection slot. Each over-limit connection has already allocated the application input/output/copy buffers and a handler goroutine. These allocations are not included in the MaxConnections-based ProtocolMemoryLimit calculation, and the rejection write may retain them until its one-second deadline. Because the accept loop has no configured bound on these pre-admission sessions, a connection flood can accumulate protocol memory outside both new limits.

Please move global admission ahead of NewIOSession, rejecting the raw connection at the accept/listener boundary, or add a separately bounded pre-admission pool whose complete memory is included in the budget.

Fresh validation on this exact head passed the full pkg/proxy suite, the full race suite, focused lifecycle tests under -race -count=10, go build, go vet, the frontend IOSession regression, and git diff --check. The previously reported login-peak admission gap and late PROXY-header gap are fixed; current tests do not cover the two interleavings above.

LeftHandCold (changes requested):
Deep re-review of head 14fefd36dc found two concrete resource-bound blockers.

### P1 - Linearize tunnel close with replacement publication

pkg/proxy/tunnel.go:356-397 publishes newSC under t.mu, but pkg/proxy/tunnel.go:615-640 does not record a closed state under that lock. The following reachable ordering loses the replacement cleanup owner: a transfer finishes BuildConnWithServer; a pipe error calls Close, which cancels the tunnel, closes the current backend, and completes closeOnce; then replaceServerConn publishes the already-authenticated replacement. The handler can meanwhile finish its defers, and in the non-cache path it closes only the original sc. A later Close is a no-op, leaving the replacement socket, connManager registration, and protocol-memory ownership live.

Please make terminal close and replacement publication mutually exclusive under t.mu. If the tunnel is already closed, replacement must reject and close both the new ServerConn and its MySQLConn wrapper. A deterministic test should pause between replacement creation and publication, close the tunnel, then verify the new backend is closed exactly once.

### P1 - Apply global admission before allocating the Goetty session

pkg/proxy/server.go:133-149 configures Goetty to create an IOSession for every accepted socket, while the MaxConnections admission is acquired only in pkg/proxy/handler.go:221. Goetty NewIOSession has already allocated both copy buffers and the input/output ByteBufs, inserted the session into its unbounded session map, and started a handler goroutine before that check. Over-limit connections can retain this state through the one-second rejection write, but these pre-admission sessions are not included in the MaxConnections-based ProtocolMemoryLimit budget. A connection flood can therefore accumulate protocol memory, goroutines, and file descriptors outside both configured limits.

Please acquire global admission at the listener/accept boundary before NewIOSession, or introduce a separately bounded pre-admission pool and include its complete peak in the memory budget. Add a test that blocks rejection completion and proves the number of materialized sessions remains bounded.

The remaining handshake, TLS, PROXY-v2, cache/migration, packet-bound, and protocol-memory accounting paths did not reveal another blocker on this head.

gouhongshen (changes requested):
## Codex automated review

Three blocking resource-boundary holes remain at 14fefd36: admission happens after session allocation, tunnel replacement is not linearized with Close, and cached tunnels are omitted from steady-memory accounting. Existing discussions and resolved threads were reviewed; earlier corrected findings are not repeated. The diff passes git diff --check, but focused Go tests could not run because this checkout lacks the Go/CGo toolchain assets.

### P1 - Acquire global admission before Goetty materializes the session (pkg/proxy/handler.go:221)
This is the first global admission check, but the pinned Goetty Application has already accepted the socket, constructed NewIOSession, inserted it into the application session map, and started its handler goroutine. Construction allocates the 256-byte input/output buffers and two 1024-byte copy buffers, in addition to retaining the FD and session. Rejected connections can then remain for the one-second error-write deadline. Consequently, overload is bounded only by accept rate and OS resources—not MaxConnections or ProtocolMemoryLimit. Admission must be moved to, or a separate bound added around, the accept/session-creation boundary; the current handler test cannot detect this because it constructs the session before calling handle.

### P1 - Linearize replacement publication with terminal tunnel closure (pkg/proxy/tunnel.go:395)
getNewServerConn returns a connected, connManager-tracked replacement before replaceServerConn takes t.mu. If a pipe failure makes Close win during that interval, Close consumes closeOnce and closes the old current backend. replaceServerConn can then publish the replacement and promoteProtocolMemory releases its transient accounting, even though no future Close can own that new connection; the non-cache handler defer only closes the originally captured sc. This leaks the backend socket, MySQLConn buffer, and manager registration across completed clients. Record terminal state under the same lock as publication and close the unpublishable replacement, including its lease, when Close has already linearized.

### P1 - Account for protocol buffers retained by cached tunnels (pkg/proxy/protocol_memory.go:162)
cachedSessions contributes only backend IOSession and retained-response bytes, while tunnelBytes is multiplied solely by active connections. A cached serverConn retains its originating tunnel through serverConn.tun; after a successful cache Push, the handler and tunnel intentionally leave it alive. That tunnel still owns both 10 KiB MySQL message buffers and the 64 KiB writer, or proxyTunnelBufferSize (86,016 bytes). The 1024-entry cache can therefore retain about 84 MiB beyond the validated limit, independently of active tunnels; with a small MaxConnections setting this can make actual steady protocol memory several times the configured minimum. Include cached tunnel buffers in the budget or sever and release the tunnel ownership when caching.

XuPeng-SH:
Addressed the latest requested changes in `deb52ea29c`.

Latest review findings:

1. **Admission now happens before Goetty session materialization.** The listener acquires the global lease before returning a raw connection to Goetty. An over-limit socket is rejected and closed inside `Accept`, so it never reaches `NewIOSession`, the session map, or a handler goroutine. Lease ownership is explicit: listener wrapper -> handler, with close-before-handler and double-close covered.
2. **Tunnel close and backend publication now form one terminal state machine.** `closed` and replacement publication share `t.mu`. If Close wins, the unpublished `ServerConn` is closed immediately, including conn-manager cleanup and transient-memory release. If publication wins, Close snapshots and closes that installed generation. A deterministic close-before-publish regression verifies exactly one cleanup owner.
3. **Cached tunnel buffers are included in the protocol-memory budget.** Cached backend sessions retain their originating tunnel, so steady tunnel owners are now `active connections + cached sessions`; the external bytes are also deducted from the managed allocator ceiling.
4. **Additional Q2 issue found during first-principles review:** replacement previously flushed the client writer while holding `t.mu`; a non-reading client could block Close behind that write forever. The writer and its ordered pending bytes are now transferred directly to the replacement, with a no-reader regression proving replacement/Close control flow stays non-blocking.

I also re-audited all historical comments against the current complete diff, not only the latest three:

- framing/deadlines: fragmented PROXY v2 header/body, fixed-header body rejection, LOCAL/UNSPEC consumption, IPv4/IPv6 minimums, duplicate/late PROXY rejection, absolute client-handshake deadline;
- TLS/login ownership: duplicate TLS rejection without recursion, coalesced TLS login/command handoff, retained allocator payload ownership, cancellation by transport close plus callback/worker join;
- admission/memory: tenant binding only after authentication, phase admission before the first read, login three-owner peak, critical/background lanes, bounded backend packet/retained response, pipelined prefix lifetime, steady tunnel and cached-tunnel accounting;
- backend/control lifecycle: context propagated through route, dial, backend auth, cache reuse, control SQL and migration RPC; every transient lease has a terminal release edge.

Verification on the exact pushed tree / semantics:

- focused lifecycle/admission/budget tests: PASS
- `./pkg/proxy`, full package: PASS
- focused race stress, `-race -count=50`: PASS (including blocked rejection and close/publication regressions)
- `./pkg/proxy`, full race suite: PASS
- `go build -mod=readonly ./pkg/proxy`: PASS
- `go vet -mod=readonly ./pkg/proxy`: PASS
- dependent frontend codec/session tests: PASS
- `make static-check`: PASS, 0 issues; license check valid 3626 / invalid 0

@aptend @LeftHandCold @gouhongshen please take another look.

aptend:
@XuPeng-SH Thanks, I re-reviewed the current head `deb52ea29c` and verified the implementation rather than relying on the summary. Items 1, 2, and 4 look addressed, and the focused race tests pass.

There is still one P1 gap in item 3. `serverConn.tun` continues to reference the originating tunnel after a cached backend is returned by `Pop`, while the newly admitted client creates another current tunnel around that same backend. Therefore an active reused backend retains both its originating tunnel and its current tunnel.

The actual owner count can be:

`active current tunnels + active reused-backend origin tunnels + cached origin tunnels`

but `calculateProtocolMemoryBudget` only reserves `connections + cachedSessions`. The active reused-origin count is not bounded by `defaultMaxNumTotal`: the cache can be popped into long-lived active connections and replenished repeatedly. With the defaults, this can undercount up to `MaxConnections * proxyTunnelBufferSize`, approximately 820 MiB, so `ProtocolMemoryLimit` is still not a hard bound.

Please either safely detach/rebind the originating tunnel ownership when a cached backend is reused, or include the additional active-origin generation in the budget. A regression should cover cached backends being popped into active use while the cache is replenished, rather than only comparing the static cache-enabled and cache-disabled budgets.

gouhongshen (changes requested):
## Codex automated review

The protocol-memory budget is still not a hard bound when cached backend connections are reused.

### P1 - Account for originating tunnels of reused cached backends (pkg/proxy/protocol_memory.go:166)
A cached ServerConn retains its original `serverConn.tun`, including that tunnel’s two message buffers and writer. `connCache.PopContext` returns the backend without rebinding this owner, while the new client creates another tunnel around the same backend. Thus active reused sessions retain both current and originating tunnel buffers. The budget counts only `connections + cachedSessions`; the active-origin count can grow to MaxConnections as cached sessions are repeatedly popped and replenished, exceeding the configured ProtocolMemoryLimit by roughly MaxConnections*proxyTunnelBufferSize (about 820 MiB at the defaults). Detach/rebind the originating tunnel or account for this additional live generation.

LeftHandCold (approved):
Deep re-review of head 9e607b75cc completed. The previous resource-bound blockers are closed:

- global admission now occurs in the listener before Goetty materializes IOSession buffers, the session map entry, or the handler goroutine; listener-to-handler lease transfer has one idempotent cleanup owner;
- tunnel Close and replacement publication are linearized under the same mutex, so an authenticated replacement built after terminal close is rejected and closed with its transient memory lease;
- cached backend reuse waits for all work from the origin tunnel generation to finish, then atomically rebinds serverConn and connManager ownership to the next tunnel;
- steady protocol-memory accounting now includes cached tunnel buffers, and the listener/address/socket behavior remains equivalent to the pinned Goetty implementation.

I also rechecked normal handshake, TLS, PROXY-v2, migration, cache failure/cancellation, and shutdown paths and found no new correctness or business-regression blocker.

- focused admission, Close/replacement, cache barrier/rebind, and memory-budget tests under the race detector, 10 iterations;
- full pkg/proxy suite;
- full pkg/proxy suite under the race detector;
- go build and go vet for pkg/proxy;
- git diff --check.

Approving.

aptend (changes requested):
Re-reviewed the latest head 9e607b75cc. The previous originating-tunnel memory-accounting issue is fixed, but one cache generation-safety blocker remains.

### P1 - Seal the old client-to-server pipe before publishing a backend to the cache

In pkg/proxy/client_conn.go:801-809, quit handling pauses only the server-to-client pipe and then publishes the backend with connCache.Push. The client-to-server pipe is still running. The new reuse barrier delays Pop until the old handler exits, but it does not prevent the old client from writing into the already-cached backend before that exit.

A reachable ordering is:

1. The old client sends COM_QUIT; the proxy consumes it locally, pauses s2c, resets the backend, and pushes it into the cache.
2. The old client keeps its socket open and sends another command. The still-running c2s pipe forwards it into the cached backend.
3. Because s2c is stopped, that command response remains unread. The client then disconnects and the handler marks the reuse barrier ready while retaining the successfully cached backend.
4. A new client pops the backend and executes SET CONNECTION ID. That operation first reads the stale response from the old command and treats it as its own result. The new tunnel then starts with the SET CONNECTION ID response still pending and with session state potentially dirtied after reset.

This crosses MySQL response ordering and session state between connection generations. Please stop and join the old c2s pipe, or atomically detach/seal the backend from the old generation, before reset/Push makes it visible in the cache. Add a deterministic regression for COM_QUIT -> cache Push -> post-QUIT command -> old-client disconnect -> cache reuse.

Fresh verification on this exact head passed the new focused tests under -race -count=10, the full pkg/proxy race suite, and git diff --check. The current tests do not cover post-QUIT writes after cache publication.

gouhongshen (changes requested):
## Codex automated review

One blocking cache-reuse protocol isolation issue remains. Focused tests could not run because the checkout has no Go toolchain.

### P1 - Seal the client-to-server pipe before caching the backend (pkg/proxy/client_conn.go:801)
The new reuse barrier only delays Pop until the old handler exits; it does not prevent the old c2s pipe from writing after Push. After COM_QUIT, this code pauses only scp and publishes c.sc. The old client can then send another command, which csp forwards to the cached backend while scp remains paused. After the old client disconnects, the barrier opens but that command's response is still unread. Pop then issues SET CONNECTION ID and consumes the stale response as its own, leaving the real SET response/session state to cross into the new client generation. Pause/join csp (or otherwise detach the backend) before Push, and add a COM_QUIT → post-QUIT command → reuse regression.

XuPeng-SH (commented):
Deep review summary: I did not find a blocking correctness / lifecycle / unhappy-path bug in this patch.

What I checked:
- listener-level admission and handler fallback ownership
- handshake/login memory ownership and protocol-memory lease release paths
- cache reuse / tunnel replacement / connManager ownership transfer
- PROXY v2 / TLS / oversized-packet unhappy paths

Non-blocking follow-ups only:
- add one real server-level admission rejection test that exercises the listener boundary end-to-end;
- add an end-to-end oversized client handshake test through the actual proxy server wiring;
- consider whether the 1 KiB backend auth-response cap should remain strict for verbose ERR packets from future auth plugins.

aptend:
Re-reviewed current head `0d38a090ec`. The previous cache-generation blocker is closed: COM_QUIT now seals the client-to-server pipe before reset/cache publication, keeps the pipe synchronously blocked until publication finishes, and terminates it before any buffered post-QUIT command can reach the cached backend. Fragmented command detection is covered as well.

Fresh validation on this exact head passed the focused cache-quit regressions under `-race -count=10`, the full `pkg/proxy` race suite, and `git diff --check`. I found no remaining code-correctness blocker.

The Coverage job is still red because the reported aggregate coverage is 73.526% against the 75% threshold; that is a CI/process item rather than a correctness finding from this review.

aptend (approved):
Re-reviewed head `0d38a090ec`. The previously reported cache-generation isolation blocker is fixed, the focused and full proxy race tests pass, and I found no remaining code-correctness blocker. Approving.

XuPeng-SH:
Final deep-review update for head `ca8d782d12` (latest `main` merged by `6b279ac22e`).

I revalidated every historical review theme, then found and fixed additional reachable cache-generation holes:

- Cache publication now requires a proven clean MySQL request/response boundary. A pipelined `QUERY -> COM_QUIT` with an outstanding response discards the backend instead of exposing that response to the next cached generation.
- The old boolean/timestamp guess is replaced by one bounded sequential-protocol state machine. It handles ordinary OK/ERR, multi-results, legacy column/final EOF pairs, `CLIENT_DEPRECATE_EOF`, prepared-statement metadata, LOAD DATA LOCAL, fragmented terminal responses, and coalesced packet framing. Unsupported command pipelining makes only that tunnel generation non-cacheable; it does not retain an unbounded command queue.
- Intermediate `SERVER_MORE_RESULTS_EXISTS` responses and the first legacy result-set EOF cannot release response ownership.
- Cache rejection, unsafe cache discard, and handler terminal cleanup now close the transport directly. They no longer send protocol COM_QUIT and wait without a context for an unhealthy CN to return EOF.
- Plugin context-forwarding tests close the prior changed-line coverage hole.

Historical closure rechecked on the final tree:

- absolute handshake deadline, TLS read-ahead ownership, oversized ERR sequence, fragmented/duplicate PROXY v2 and LOCAL/UNSPEC behavior;
- listener-before-IOSession global admission, tenant admission, steady/cached/migration buffer accounting, transient critical/background lanes, allocator fallback;
- cancellation across dial/auth/replay/cache validation, replacement-after-close linearization, cache reuse barrier/rebind, post-QUIT generation isolation;
- overload logging, connection-manager ownership, migration and client-disconnect cleanup.

Bounded/non-blocking design points:

- `UPGRADE ACCOUNT ALL` intentionally keeps long-running business semantics. It uses the separate one-operation background lane, so it cannot consume critical admission capacity or make protocol memory unbounded; imposing an arbitrary query timeout here would change behavior.
- This PR fixes the OOM deterministically through hard global/per-tenant cardinality and protocol-memory admission. A general idle-session deadline is deliberately not introduced because it would terminate legitimate long HTAP queries/transactions. Idle-age reclamation and age/idle observability remain an opt-in policy/metrics follow-up, not a missing hard-safety edge.

Fresh exact-head validation:

- full `./pkg/proxy`: pass, 83.6% statement coverage;
- full `./pkg/proxy` with race detector: pass;
- focused cache-boundary/fragmentation/context tests: 20 iterations, pass;
- `go build` for proxy/frontend; standard vet; MO molint; golangci-lint: pass, 0 issues;
- exact CI `parse_coverage.py` against the final diff: `741/971 = 76.313%`, above the 75% gate;
- `git diff --check`: pass; all 5 inline review threads are resolved.

CI for this head is now running. @aptend @gouhongshen please re-review `ca8d782d12`; GitHub still shows an older `CHANGES_REQUESTED` review state even though the inline threads are resolved and the previous head had approvals.

XuPeng-SH (commented):
I found one blocking shutdown-path bug on a second deep pass.

`connCache.Close()` can concurrently destroy a real `serverConn` while `PopContext()` is still validating that same cached backend with `SET CONNECTION ID`.

Concrete path:
- `connCache.PopContext()` intentionally leaves the reserved cached connection in `allConns` while it runs `execStmtWithContext(... SET CONNECTION ID ...)` (`pkg/proxy/conn_cache.go:441-444, 483-503`).
- `handler.Close()` now calls `connCache.Close()` **before** `app.Stop()` (`pkg/proxy/handler.go:480-490`, `pkg/proxy/server.go:160-165`), so active handlers / cache pops can still be running when shutdown starts.
- `connCache.Close()` iterates `allConns` and calls `serverConnAuth.close()` → `ServerConn.Close()` (`pkg/proxy/conn_cache.go:561-576, 147-151`).
- Real `serverConn.Close()` closes the TCP conn **and** frees the protocol object immediately via `s.mysqlProto.Close()` / `frontend.Conn.Close()` (`pkg/proxy/server_conn.go:550-563`, `pkg/frontend/mysql_protocol.go:639-654`, `pkg/frontend/mysql_buffer.go:360-377`).
- Meanwhile `ExecStmtContext()` / `ExecStmt()` is still in `WritePacket()` / `readPacket()` on that same `mysqlProto` / `frontend.Conn` (`pkg/proxy/server_conn.go:488-515, 458-485, 567-582`; `pkg/frontend/mysql_protocol.go:3830-3845`; `pkg/frontend/mysql_buffer.go:471-558, 887-926`).

So the shutdown path currently reintroduces the exact class of race that `HandleHandshakeContext()` already avoided by closing only the raw transport and joining the worker before freeing protocol buffers.

Suggested fix:
1. Split “interrupt transport” from “final free” for cached backends under validation.
2. During `connCache.Close()`, only close the raw connection (or add an interrupt-only method) so blocked validation returns.
3. Call full `ServerConn.Close()` only after the validating goroutine exits.

I also re-checked the QUIT unhappy-path candidate from the previous pass: the current code no longer falls back to protocol QUIT on pause timeout (`handleQuitEventInternal()` now goes through `discardBackend()` and direct `Close()`), so I do **not** think that older concern is still valid in this revision.

Issue #25859, opened 2026-07-18
[Bug]: TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots  error


LeftHandCold:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

lasted

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

[TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots](https://github.com/matrixorigin/matrixone/actions/runs/29629432967/job/88040123784?pr=25841)

### Steps to Reproduce

```Markdown
run ut
```

LeftHandCold:
Root cause confirmed on current main: this is a test-liveness failure, not a demonstrated lockservice production deadlock. The test starts its background waiter with the same 10-second context used by setup, then waits forever for one queue entry. If that context expires before remote queue admission, Lock returns without enqueueing and test_helper.go:waitLocalWaiters spins until the package 40-minute timeout. The reproducing stack is in waitLocalWaiters -> WaitWaiters -> TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots. PR #25860 gives the waiter its own lifecycle and bounds admission, post-release, and cleanup waits.

PR #25860, opened 2026-07-18
test(lockservice): bound remote waiter snapshot setup

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25859

TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots used its 10-second setup context for the background waiter and then called an unbounded test helper while waiting for that waiter to reach the owner queue. If setup or remote routing consumed the deadline, Lock returned without entering the queue and the helper spun until the package-level 40-minute timeout.

The test now gives the background waiter an explicit lifecycle, bounds both admission and post-release waits, and uses independent cleanup contexts. On every assertion-failure path it releases the holder, cancels the waiter, and unlocks a waiter that acquired the lock before cleanup.

A follow-up CI run exposed that the independent waiter initially used a context without a deadline. Remote Lock crosses MORPC, whose Future contract requires every non-one-way send to carry a deadline. The waiter now has an independent 30-second timeout: it remains decoupled from setup while satisfying the RPC contract and retaining a hard liveness bound. The later RWMutex fatal in that run was cascading panic cleanup after the deadline-contract panic, not the original lockservice hang.

This is a test-liveness repair; it does not change lockservice production behavior.

Related CI failure #25729 is already covered by the receiver lifecycle fix merged in #25820. Current main dispatch regression no longer contains the racy post-send assertion, so no duplicate issue or code change is needed here.

Tested with:

mo-cgo-test -v -count=1 -timeout=90s ./pkg/lockservice -run ^TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots$

20 independent process repetitions of the focused test

mo-cgo-test -race -count=20 -timeout=180s ./pkg/lockservice -run ^TestFetchWhoWaitingMeUsesActiveRemoteWaiterSnapshots$

mo-cgo-test -count=1 -timeout=300s ./pkg/lockservice

go build ./pkg/lockservice

go vet ./pkg/lockservice


LeftHandCold:
Follow-up on the failed CI run: the first panic was caused by the test waiter using a context without a deadline. Remote Lock crosses MORPC, and Future.init requires a deadline for non-one-way sends. Commit acb41d7f3f changes the waiter to an independent 30-second timeout, preserving separation from the setup budget while satisfying the RPC contract. The subsequent RWMutex fatal was cascading panic cleanup after that first panic, not evidence of the original lockservice hang. Focused, 20 independent-process repetitions, race (-count=20), full lockservice package, build, and vet all pass locally.

LeftHandCold:
Latest main is now merged at PR head 812338843b. The conflict was limited to this test cleanup and was resolved by retaining the failure-safe lifecycle cleanup while incorporating the new main changes (including TxnIterFunc liveness and bounded helper coverage). Revalidation on the final merged head passed: focused test, race focused test with -count=5, full ./pkg/lockservice, go build, go vet, and git diff --check. New CI run: https://github.com/matrixorigin/matrixone/actions/runs/29639445063

gouhongshen (changes requested):
## Codex automated review

The new test still has an unbounded remote-unlock path, so it can hang despite the stated bounded setup/cleanup intent. Focused execution was unavailable because Go is not installed in the environment.

### P1 - The release timeout does not bound remote unlock (pkg/lockservice/service_remote_test.go:198)
`service.Unlock` deliberately discards its caller context and invokes `unlockWithContext(context.Background())` (pkg/lockservice/service.go:251-259). A remote lock table then retries unlock until success (pkg/lockservice/lock_table_remote.go:237-272). Therefore an RPC/routing failure at this new release call blocks before the subsequent `select` can observe `releaseCtx.Done()`; the cleanup call at line 117 has the same problem. This leaves the test able to hang until the package-level timeout, contradicting the new hard-bound guarantee. Use the context-aware internal unlock path for this test (or otherwise ensure the unlock attempt itself is bounded).

gouhongshen (commented):
## Codex automated review

The bounded happy path is sound. One failure-path cleanup issue remains; the focused test could not run locally because cgo/libmo.dylib is missing.

### P2 - Waiter cleanup skips remote unlock on Lock errors (pkg/lockservice/service_remote_test.go:126)
Cancellation or an RPC timeout can make Lock return an error after the owner has acquired the lock. remoteLockTable.lock deliberately records the lock on Send errors for this case, but this cleanup only calls unlockWithContext when err == nil. The waiter transaction/remote lock can therefore remain held, and waiterDone prevents a later retry. Attempt bounded unlock regardless of the Lock error; it is safe when no lock was recorded.

gouhongshen (approved):
## Codex automated review

No concrete regressions found in the actual diff; the focused test passes in CI.

XuPeng-SH (approved):
Deep review completed. The final head closes the waiter lifecycle and all newly introduced waits are bounded. Focused stress, race, full lockservice package, build, vet, latest-main merge validation, and CI passed. The pre-existing production retry issue is tracked separately in #25870 and does not block this test-only fix.

Issue #25752, opened 2026-07-15
[Subtask]: propagate numeric assignment context across DML paths


ck89119:
### Parent Issue

Parent issue: #25705

### Detail of Subtask

Propagate numeric assignment context through every DML statement shape after the arithmetic resolver is available.

Cover `VALUES`, direct and parenthesized `INSERT ... SELECT`, set operations, star expansion, legacy and modern insert paths,
`UPDATE`, and `ON DUPLICATE KEY UPDATE`. All paths must use one context API and must preserve the rule that an explicit
`DOUBLE` sibling overrides an exact target column.

### Describe implementation you've considered

Establish the target-column/projection mapping before binding the affected arithmetic expression. Pass the target type as an
outer numeric context instead of rebuilding an already-bound AST or plan expression. Keep final assignment casts responsible
only for validating and storing the result.

### Additional information

Add focused plan tests for each statement shape, values above `2^53`, decimal and integer targets, approximate siblings,
metadata, and both legacy and modern planner paths.

PR #25861, opened 2026-07-18
feat(plan): propagate numeric assignment context across DML paths

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25752

Propagates the numeric assignment context (the target column type) through every
DML statement shape after the arithmetic resolver from #25751, so a prepared
numeric parameter resolves against the column it is assigned to.

Covered paths, all through one context API (`bindNumericExprWithContext`):

- `INSERT ... VALUES`
- direct, parenthesized, and `UNION` `INSERT ... SELECT`
- legacy and modern `UPDATE`
- `ON DUPLICATE KEY UPDATE`

An explicit approximate sibling (e.g. `DOUBLE`) still overrides an exact target
column. The context is seeded only for numeric target columns: a non-numeric
target (`varchar`, `date`, ...) keeps the default binding path, because
`InferNumericParameterType` falls back to `float64` for a non-numeric outer type
and would otherwise mistype a bare parameter (e.g. `UPDATE t SET status = ?`
would store `0`).

The numeric target is applied only to a projection that is bound fresh: it is
checked after the group-by / aggregate / window / time-window / sample lookups so
a grouped arithmetic projection assigned to a numeric column resolves to its
grouped column instead of being re-bound against the raw scan columns.

Tests: focused plan tests for each statement shape (VALUES above `2^53`, decimal
and integer targets, approximate siblings, non-numeric targets, grouped
projections, legacy and modern paths). Full `pkg/sql/plan` unit tests and the
`prepare`, `dml/insert`, and `dml/update` BVT suites pass.


gouhongshen (changes requested):
## Codex automated review

Seven P1 correctness gaps leave prepared parameters on the default numeric path in valid DML shapes. Focused Go tests could not run because `go` is unavailable in the environment.

### P1 - Preserve parameter identity when matching GROUP BY expressions (pkg/sql/plan/group_binder.go:102)
`tree.String` formats every `ParamExpr` as `?`, ignoring its distinct parameter offset. Thus `SELECT ? + ? ... GROUP BY ? + ?` is treated as the same expression, and the GROUP BY markers receive the INSERT target type at line 75. In the base behavior GROUP BY was bound without assignment context; these independently supplied parameters can have different values and should not inherit the target type.

### P1 - Propagate targets through star-expanded derived projections (pkg/sql/plan/query_builder.go:3802)
`seedNumericTableProjectionTypes` runs before star expansion and explicitly skips `name.Star`. For `INSERT INTO constraint_test.emp(sal) SELECT d.* FROM (SELECT ? + ? AS x) d`, no target mapping reaches the derived table, so its parameters are bound on the default path and only the outer result is cast afterward, risking precision loss.

### P1 - Seed the legacy INSERT ... SELECT fallback (pkg/sql/plan/bind_insert.go:2698)
The target vector is installed only in the modern INSERT/REPLACE path. Plain INSERT into a non-ordinary/non-index table returns `ErrUnsupportedDML` and falls back through `buildInsert`; its `initInsertStmt` creates a fresh context and calls `bindSelect` without `numericProjectionTypes`. A prepared `INSERT ... SELECT ? + ?` into a writable external or index-related table therefore loses the destination numeric context.

### P1 - Propagate assignment context into ODKU scalar subqueries (pkg/sql/plan/ondup_update_binder.go:83)
`BindAssignmentExpr` enters numeric binding for numeric targets but never sets `numericSubqueryTarget`, unlike `ProjectionBinder`. For `ON DUPLICATE KEY UPDATE sal = (SELECT ? + ?)`, the numeric AST scan does not descend into the subquery, `baseBindSubquery` creates a context with no target vector, and the parameters remain default-typed before the final assignment cast.

### P1 - Handle UNION-derived sources in target lineage (pkg/sql/plan/query_builder.go:3866)
`numericProjectionOutputNames` accepts only a `SelectClause`. A derived source whose SELECT is a `UnionClause` therefore returns no output names, so `INSERT INTO constraint_test.emp(sal) SELECT d.x FROM (SELECT ? + ? AS x UNION ALL SELECT ? + ?) d` binds both UNION branches without the target context. The direct UNION path copies the vector, but the nested derived-UNION path does not.

### P1 - Carry target lineage across chained CTEs (pkg/sql/plan/query_builder.go:3886)
The CTE lookup only inspects the current SELECT's `With` clause. In `WITH c AS (SELECT ? + ? AS x), d AS (SELECT x FROM c) INSERT INTO constraint_test.emp(sal) SELECT x FROM d`, the outer query can seed `d`, but `d`'s body has no local `With` containing sibling `c`; `c` is consequently bound without the propagated target and its parameters keep the default type.

### P1 - Propagate through UPDATE ... FROM derived sources (pkg/sql/plan/query_builder.go:3841)
Both modern and legacy UPDATE builders merge the target table and `FROM` sources into a real `JoinTableExpr`. `numericProjectionSource` rejects any join with a non-nil right side, so `UPDATE constraint_test.emp SET sal = d.x FROM (SELECT ? + ? AS x) d WHERE emp.empno = 1` never seeds the derived source, leaving its parameters default-typed despite the numeric assignment target.

gouhongshen (changes requested):
## Codex automated review

Two blocking planner correctness defects remain in supported DML shapes.

### P1 - Flattened ODKU scalar subquery breaks unique-key conflict resolution (pkg/sql/plan/bind_insert.go:1352)
This newly flattens every ODKU RHS before `buildOnDupTargetPkResolution`. A scalar subquery with a FROM clause turns `lastNodeID` into a JOIN (which has no ProjectList), but the resolution helper subsequently treats that node as the incoming PROJECT and indexes `selectNode.ProjectList` for each unique index (lines 883 and 957). Thus a normal table with a PK and unique index panics while planning, e.g. `INSERT INTO constraint_test.emp (empno,ename,job,sal) VALUES (1,'e','j',1) ON DUPLICATE KEY UPDATE sal=(SELECT ? + ? FROM nation LIMIT 1)`. The existing scalar-subquery test uses a FROM-less subquery, which is optimized away and misses this path.

### P1 - Star target propagation is misaligned for JOIN ... USING (pkg/sql/plan/query_builder.go:3808)
The new propagation treats every JOIN as concatenated source outputs and advances through every column of both sides. Actual `SELECT *` expansion emits a USING column only once (`BindContext.doUnfoldStar` records it as visited before traversing both children), so all targets after a USING join are shifted. For example, in `INSERT INTO constraint_test.emp (empno,ename,job,mgr,hiredate,comm,deptno,sal) SELECT * FROM nation n1 JOIN nation n2 USING(n_nationkey) CROSS JOIN (SELECT ? + ? AS x) d`, the real output has 7 nation columns followed by `d.x`, but this helper counts 8 nation columns. It consequently never assigns `sal`'s DECIMAL context to `d`, leaving the prepared parameters as the default floating type and reintroducing the precision loss this PR is meant to prevent.

aunjgr (approved):
Re-reviewed current head. The previous assignment-context gaps are closed across GROUP BY parameter identity, derived stars and UNIONs, CTEs, legacy INSERT, ODKU scalar subqueries, and UPDATE FROM. Current CI was green and I found no new blocker.

aunjgr (approved):
Re-reviewed current head. The previous assignment-context gaps are closed across GROUP BY parameter identity, derived stars and UNIONs, CTEs, legacy INSERT, ODKU scalar subqueries, and UPDATE FROM. Current CI was green and I found no new blocker.

gouhongshen (changes requested):
## Codex automated review

Three P1 planner-correctness gaps remain. Focused tests could not run locally because the worktree lacks cgo/libmo.dylib; CI is green.

### P1 - Re-project the ODKU input after flattening a scalar RHS (pkg/sql/plan/bind_insert.go:1352)
A scalar RHS containing a FROM clause makes flattenSubqueries replace lastNodeID with a JOIN. buildOnDupTargetPkResolution then indexes that node's ProjectList for every unique key, but JOIN nodes have no ProjectList, causing a planning panic. For example: INSERT ... ON DUPLICATE KEY UPDATE sal = (SELECT ? + ? FROM nation LIMIT 1) on a normal real-PK table with a unique key. The added test only covers a FROM-less scalar subquery, which flattening removes.

### P1 - Propagate assignment type into scalar subqueries below arithmetic (pkg/sql/plan/projection_binder.go:101)
numericSubqueryTarget is set only when the whole assigned expression is a scalar subquery. For INSERT INTO constraint_test.emp (sal) SELECT (SELECT ? + ?) + 0, numericAstTypes does not traverse the nested Subquery, so the child binder receives no target type and resolves its parameters/arithmetic as FLOAT64. The final destination DECIMAL cast occurs only after the precision-lossing arithmetic.

### P1 - Apply propagated targets to recursive CTE branches (pkg/sql/plan/query_builder.go:3145)
bindNoRecursiveCte copies numericTableProjectionTypes into its child context, but bindRecursiveCte creates initCtx and recursive subCtx without doing so. Consequently, in INSERT INTO ... (sal) WITH RECURSIVE c(x) AS (SELECT ? + ? AS x UNION ALL SELECT x FROM c WHERE x < 1) SELECT x FROM c, the seed arithmetic binds as FLOAT64 rather than the destination numeric type. The new coverage exercises only non-recursive CTEs.

XuPeng-SH (changes requested):
Requesting changes on the current head `97d042972`. Four planner-correctness gaps remain in valid prepared DML shapes.

### P1 - Preserve parameter identity for GROUP BY aliases

At `group_binder.go:83`, `registerAst := isOrdinal || !containsDynamicParam(expr)` assumes that only an ordinal can refer to the SELECT expression itself. However, `bindGroupBy` expands a SELECT alias through `qualifyColumnNames(..., AliasAfterColumn)`, so this valid statement also reuses the SELECT AST rather than introducing new parameters:

```sql
INSERT INTO constraint_test.emp (sal)
SELECT ? + ? AS x FROM nation GROUP BY x;
```

On this head, parameter 1 (and 2) occurs as `FLOAT64` in the GROUP expression and `DECIMAL64(7,2)` in the projection. The same statement with `GROUP BY 1` is consistently DECIMAL. Track whether alias expansion refers to a select-list item and preserve its target/registration identity, while continuing to keep independently written `GROUP BY ? + ?` markers separate.

### P1 - Propagate the target into scalar subqueries nested inside ODKU arithmetic

`OndupUpdateBinder.BindAssignmentExpr` sets `numericSubqueryTarget` only when the entire RHS is a scalar subquery. The nested form remains on the default path:

```sql
INSERT INTO constraint_test.emp (empno, sal) VALUES (1, 1)
ON DUPLICATE KEY UPDATE sal = (SELECT ? + ?) + 0;
```

Both parameters are `FLOAT64` instead of destination `DECIMAL64(7,2)`, so values above 2^53 can lose precision before the final assignment cast. The target scope must cover scalar subqueries encountered anywhere in the numeric assignment tree, not only a parenthesized whole-RHS subquery.

### P1 - Expand derived/CTE stars before deriving output lineage

`numericProjectionOutputNames` treats every raw select-list expression as one output and assigns no name to `*`. That breaks an otherwise pure forwarding chain when a nested derived table or CTE expands a star:

```sql
INSERT INTO constraint_test.emp (sal)
SELECT d.x FROM (SELECT * FROM (SELECT ? + ? AS x) s) d;

INSERT INTO constraint_test.emp (sal)
WITH c AS (SELECT * FROM (SELECT ? + ? AS x) s)
SELECT x FROM c;
```

Both cases plan successfully but leave both parameters as `FLOAT64`. Lineage needs the actual expanded headings/width (including alias-column lists), or an equivalent recursive star propagation, before mapping destination positions.

### P1 - Propagate the resolved arithmetic domain across scalar-subquery boundaries

`ProjectionBinder` currently assigns the raw destination target to `numericSubqueryTarget` before resolving outer siblings:

```sql
INSERT INTO constraint_test.emp (sal)
SELECT (SELECT ? + ?) + CAST(1 AS DOUBLE);

UPDATE constraint_test.emp
SET sal = (SELECT ? + ?) + CAST(1 AS DOUBLE)
WHERE empno = 1;
```

The inner parameters become `DECIMAL64(7,2)` on this head, although the explicit `DOUBLE` sibling must override the exact assignment target. Large opposite inputs can therefore overflow in the inner DECIMAL operation before they get the chance to cancel in the approximate outer domain—the exact failure the resolver precedence is intended to avoid. Pass the domain resolved for the containing arithmetic expression into scalar projections, not the raw assignment type.

Validation performed with the deterministic CGo environment: PR-focused tests, full `pkg/sql/plan`, build, vet, formatting, and five race runs all pass. Independent tests above reproduce on the latest head. No goroutine/resource/leak/hang issue was found; statement-local maps/slices are bounded by AST/column/CTE counts. Please add these four semantic axes to the regression matrix, preferably also with a real PREPARE/EXECUTE case above 2^53 so plan-type checks cannot mask execution behavior.

gouhongshen (changes requested):
## Codex automated review

One blocking correctness issue remains. Focused tests could not run because required CGo headers are unavailable; diff checks passed and the worktree is clean.

### P1 - Cluster-table star expansion uses the hidden account_id column for target-position mapping (pkg/sql/plan/query_builder.go:4025)
numericPhysicalTableOutputNames includes every non-hidden TableDef column, but non-system cluster-table star expansion explicitly omits account_id in bind_context.go. Since account_id is physically added to cluster tables, a query such as INSERT INTO dst SELECT * FROM cluster_src CROSS JOIN (SELECT ? + ? AS x) d, executed by a tenant, is mapped as [visible columns, account_id, x] by this helper while the actual projection is [visible columns, x]. The numeric target for x is therefore assigned to the hidden account_id slot and the derived expression receives no target context, so its parameters fall back to FLOAT64 and can lose decimal precision. The mapping must use the same account-specific visibility rules as actual star expansion.

gouhongshen (changes requested):
## Codex automated review

Two P1 prepared-statement type-resolution defects remain. Local focused tests could not run because cgo/libmo.dylib is absent; current CI is green.

### P1 - Preserve explicit DOUBLE precedence through conditional operands (pkg/sql/plan/base_binder.go:951)
`numericAstTypesInternal` falls through to the default case for `CASE`, so it does not discover a typed approximate operand nested in a conditional. For `INSERT INTO constraint_test.emp (sal) SELECT ? + CASE WHEN 1=1 THEN CAST(1 AS DOUBLE) ELSE 0 END`, the scan sees no strong type and infers the parameter as the destination `DECIMAL(7,2)`. `baseBindParam` therefore installs a text→DECIMAL cast before the later arithmetic resolver sees the CASE result is DOUBLE and adds a second cast. Inputs outside DECIMAL(7,2) consequently overflow/round before the DOUBLE operation, violating the required explicit-DOUBLE-overrides-target rule. Resolve the output domain of conditional/wrapper expressions before selecting `numericParamType`, and add this regression case.

### P1 - Do not overwrite assignment context when one derived output feeds incompatible targets (pkg/sql/plan/query_builder.go:4324)
A derived/CTE output is evaluated once, but this assignment unconditionally replaces its previously seeded target. In `INSERT INTO constraint_test.emp (sal, empno) SELECT d.x, d.x FROM (SELECT ? + ? AS x) d`, the first reference seeds `d.x` with `DECIMAL(7,2)` and the second overwrites it with `UINT32`. The derived expression and both parameters are then bound as UINT32 before `sal` is cast, so fractional or large decimal inputs are lost/rejected; reversing the output order changes the computation type. `storeNumericTableProjectionTargets` cannot detect this because it receives only the already-overwritten slice. Treat distinct target types for the same source position as ambiguous and withhold propagation (including CTE/star paths), then add a regression test.

XuPeng-SH (changes requested):
Requesting changes on current head `1f9f54df`.

The blockers from the earlier rounds appear closed, but two P1 prepared-statement correctness gaps remain.

### P1 - Preserve explicit approximate precedence through numeric function results (`pkg/sql/plan/base_binder.go:931`)

`numericAstTypesInternal` treats every `FuncExpr` except `MOD` as if it contributed no typed operand. Consequently, a numeric function whose resolved result is `DOUBLE` is invisible when the outer arithmetic parameter type is selected. For example:

```sql
INSERT INTO dst_decimal(v)
SELECT ? + IF(1 = 1, CAST(-100000 AS DOUBLE), 0);
```

The explicit `DOUBLE` operand must override the exact destination context, but this head binds `?` as the destination `DECIMAL` before the outer addition. With a narrow destination and `? = 100000`, that early cast can overflow even though the `DOUBLE` computation cancels to `0` and the final value fits. `COALESCE(CAST(... AS DOUBLE), 0)` reproduces the same type error. The newly added `CASE` traversal fixes only one AST form; numeric operand-result domains need to be visible without blindly pushing assignment context into nonnumeric functions.

### P1 - Propagate the resolved target through derived/CTE columns embedded in arithmetic (`pkg/sql/plan/query_builder.go:3835`)

`seedNumericTableProjectionTypes` seeds only a bare column reference or a star. If the forwarded column is an operand of an outer arithmetic expression, the default branch drops the destination context:

```sql
INSERT INTO dst_decimal(v)
SELECT d.x + 0 FROM (SELECT ? + ? AS x) d;

INSERT INTO dst_decimal(v)
WITH c AS (SELECT ? + ? AS x)
SELECT x + 0 FROM c;
```

Both inner parameter pairs are `FLOAT64` on this head instead of the exact destination domain. On a wide decimal destination, inputs `9007199254740993` and `-9007199254740992` therefore compute `0` in the derived node rather than exact `1`; the outer arithmetic/final cast cannot recover the lost bit. Resolve the containing arithmetic domain first and seed referenced derived/CTE outputs with that domain, while retaining the existing ambiguity fallback when one source output feeds incompatible targets.

Validation:

- full `pkg/sql/plan` tests, build, and vet pass on the exact PR head;
- the PR-focused numeric tests pass 10x after a clean merge with latest `main`;
- both independent reproductions fail on the exact head and remain failing after that latest-`main` merge;
- latest `main` merges automatically with no conflicts, `git diff --check` passes, and required PR CI is green;
- Q1-Q3 audit found no resource, wait, or unbounded-growth blocker; the new state is statement-scoped and AST/column/CTE bounded.

gouhongshen (changes requested):
## Codex automated review

Three blocking planner correctness defects remain despite green CI. Focused tests could not compile locally because this checkout lacks required CGo artifacts and third-party headers; the worktree remains clean.

### P1 - Preserve numeric context inside CASE result branches (pkg/sql/plan/base_binder.go:944)
The scanner now descends into CASE result arms, but `isNumericContextNode` still excludes `CaseExpr`. Consequently, after `bindNumericExprWithContext` infers the assignment type, `baseBindExpr` clears `numericParamType` before `bindCaseExpr` binds those arms. For `INSERT INTO constraint_test.emp(sal) SELECT CASE WHEN 1=1 THEN ? + ? ELSE 0 END`, the inner arithmetic therefore starts a new context with no outer type and both markers remain FLOAT64; values above 2^53 lose precision before the final DECIMAL assignment cast. The added CASE tests place the marker outside CASE and do not cover this path.

### P1 - Keep CTE target conflicts permanently ambiguous (pkg/sql/plan/query_builder.go:3990)
`storeNumericTableProjectionTargets` represents both “unset” and “conflicted” with `Type{}`. After one incompatible merge clears a position, the next target refills it here. For `INSERT INTO constraint_test.emp(sal,empno,mgr) WITH c AS (SELECT ? + ? AS x) SELECT a.x,b.x,d.x FROM c a CROSS JOIN c b CROSS JOIN c d`, the shared CTE target transitions DECIMAL -> empty -> UINT32. The CTE parameters are then bound as UINT32, so fractional arithmetic is lost before reaching `sal`; changing consumer order changes the computation domain. The per-source `targetAmbiguous` flag added in this commit does not protect this map-level merge across aliases/references.

### P1 - Retain parameterized GROUP BY positions for grouping sets (pkg/sql/plan/group_binder.go:95)
Skipping `groupByAst` registration for every independently written expression containing a parameter breaks the later grouping-set lookup at line 111, which reads a missing map key as position 0. In `SELECT count(*) FROM nation GROUP BY GROUPING SETS ((n_regionkey + ?), (n_nationkey + ?))`, both expressions are appended to `ctx.groups` but neither is registered; the second UNION branch therefore marks group 0 active instead of group 1 and groups by `n_regionkey` rather than `n_nationkey`. Parameter identity needs a representation that still records each exact grouping expression’s position.

XuPeng-SH (changes requested):
Deep review on `3cf258a982`: two numeric-domain closure gaps still produce observable prepared-DML failures.

[P1] Numeric-result functions outside the five-name whitelist are opaque to domain inference. `numericAstTypesInternal` returns an empty scan for every function rejected by `numericFunctionResultArgs` (`pkg/sql/plan/base_binder.go:933-938`, `:1309-1346`). Repro:

```sql
INSERT INTO constraint_test.emp(sal)
SELECT ? + ABS(CAST(-100000 AS DOUBLE));
```

`sal` is `DECIMAL(7,2)`. Expected parameter #1: `FLOAT64`, because `ABS(DOUBLE)` is the explicit approximate sibling. Actual: `DECIMAL64(7,2)`. Binding `-100000` therefore overflows before the DOUBLE addition even though the DOUBLE-domain result is `0`. This is not ABS-specific; a hard-coded IF/COALESCE/IFNULL/NULLIF/MOD list leaves other numeric-result functions with the same hole. Please derive numeric result/argument behavior from common function metadata or otherwise cover the full supported numeric-function family, and add a plan-type regression for this query.

[P1] Assignment targets still stop at IF/CASE/COALESCE roots before reaching derived tables or CTEs. `numericProjectionExprTarget` rejects anything for which `isNumericArithmeticRoot` is false (`pkg/sql/plan/query_builder.go:3870-3877`), even though `seedNumericExprColumnTargets` immediately below already knows how to recurse through those result arms. All of these bind the inner parameters as default `FLOAT64`; they must be `DECIMAL64(7,2)`:

```sql
INSERT INTO constraint_test.emp(sal)
SELECT IF(1=1, d.x, 0) FROM (SELECT ? + ? AS x) d;

INSERT INTO constraint_test.emp(sal)
SELECT CASE WHEN 1=1 THEN d.x ELSE 0 END FROM (SELECT ? + ? AS x) d;

INSERT INTO constraint_test.emp(sal)
WITH c AS (SELECT ? + ? AS x)
SELECT COALESCE(c.x, 0) FROM c;
```

With values `9007199254740993` and `-9007199254740992`, the current FLOAT64 inner expression becomes `0` instead of the exact `1`; the final assignment cast cannot recover the lost bit. Please make the root eligibility consistent with the recursive result-arm logic and cover function/CASE wrappers over both derived and CTE sources.

Validation: existing focused numeric-context tests, full `./pkg/sql/plan`, and the focused race run pass; the four added review-only reproductions above fail deterministically with the actual/expected types described. Latest `main` merge-tree is conflict-free and does not close either gap.

gouhongshen (changes requested):
## Codex automated review

Two P1 prepared-DML numeric-context correctness gaps remain. CI is green; local focused tests could not start because `go` is unavailable.

### P1 - Numeric-result functions outside the whitelist lose explicit FLOAT/DOUBLE precedence (pkg/sql/plan/base_binder.go:935)
`numericAstTypesInternal` returns an empty scan for any function absent from `numericFunctionResultArgs`; `ABS` is absent even though its FLOAT64 overload returns FLOAT64. In `INSERT INTO constraint_test.emp(sal) SELECT ? + ABS(CAST(-100000 AS DOUBLE))`, the scan sees no typed sibling, so it seeds `?` from `sal` as DECIMAL(7,2) and `baseBindParam` installs that cast before ABS later resolves to FLOAT64. Binding `-100000` then overflows in the premature decimal cast although the intended DOUBLE calculation produces 0, which fits the target. The new tests cover the five whitelisted function names but not this supported numeric function family.

### P1 - Conditional wrappers do not propagate an assignment target into derived/CTE outputs (pkg/sql/plan/query_builder.go:3876)
`seedNumericExprColumnTargets` already descends through IF/COALESCE/CASE result arms, but `numericProjectionExprTarget` rejects those roots because `isNumericArithmeticRoot` accepts only binary/unary operators and MOD. Thus `INSERT INTO constraint_test.emp(sal) SELECT IF(1=1,d.x,0) FROM (SELECT ? + ? AS x) d` never seeds `d.x`; the derived query binds both markers as default FLOAT64 instead of DECIMAL(7,2). Inputs 9007199254740993 and -9007199254740992 consequently compute 0 in the derived expression rather than exact 1, and the outer assignment cast cannot recover it. CASE and COALESCE have the same gap; existing coverage only exercises bare and arithmetic-wrapped forwarded columns.

gouhongshen (changes requested):
## Codex automated review

One prepared-DML numeric-context propagation gap remains.

### P1 - Propagate target context through derived VALUES sources (pkg/sql/plan/query_builder.go:4258)
`numericProjectionOutputNames` only recognizes SelectClause/UNION roots, so a valid derived value table has unknown output names and `seedNumericTableProjectionTypes` never records a target for it. Consequently `INSERT INTO dst(v) SELECT d.x FROM (VALUES ROW(? + ?)) AS d(x)` binds the derived `? + ?` through `bindValues`' WhereBinder (line 5904) with no assignment context, leaving it on the default FLOAT64 path before the final destination cast. For a wide DECIMAL target, inputs 9007199254740993 and -9007199254740992 therefore compute 0 instead of exact 1. Derived VALUES is supported by the parser/planner; add its `column_N` output lineage and apply `numericProjectionTypes` while binding its cells.

gouhongshen (approved):
## Codex automated review

Reviewed final diff, all existing reviews, PR comments, and inline-thread state. No remaining findings. Local focused tests could not run because `go` is unavailable; head CI is green and diff check passes.

XuPeng-SH (approved):
Re-reviewed exact head 036ee26a1505d6974ed9391cb1b4237965606285. No blocking findings remain.

The latest main merge correctly keeps CTE name resolution on the declaration context while injecting numeric assignment targets from each use site into a fresh bind context. I revalidated the prior review gaps across numeric function domains, IF/CASE/COALESCE, scalar/derived/VALUES lineage, GROUP BY parameter identity, legacy INSERT, UPDATE FROM, ODKU subqueries, chained CTEs, conflicting targets, and recursive CTE members. The merge does not reintroduce target leakage or declaration-scope capture.

Unhappy-path check: this change creates no external resources or waits; the temporary planner maps and CTE propagation queue are scoped to one build and bounded by AST projections/CTE definitions plus monotonic target-state changes.

Fresh local validation on this head:
- focused numeric-assignment and CTE lazy-binding tests: PASS
- full pkg/sql/plan test suite: PASS
- focused race tests for INSERT/UPDATE/CTE paths: PASS
- go vet ./pkg/sql/plan: PASS
- go build ./pkg/sql/plan: PASS
- diff check: PASS

Current required CI checks are green.

PR #25863, opened 2026-07-18
feat(data-branch): materialize diff output as table

### LINKED ISSUE #25871: bug: DATA BRANCH DIFF OUTPUT AS cannot materialize durable diff tables
## Problem

`DATA BRANCH DIFF ... [COLUMNS (...)] OUTPUT AS <table>` is accepted by the parser but did not materialize a durable, queryable table. This blocks users from retaining and reusing DIFF results.

The affected execution paths also failed for historical source schemas after ALTER/DROP/recreate, did not preserve all serialized value families, and could lose the caller-visible error in an explicit transaction.

## Impact and S0 rationale

This is a confirmed broken Data Branch output workflow, including snapshot Diffs and normal persistent-table use. It has a stable local reproduction and an assigned owner with a complete fix and BVT coverage, so it is tracked as active `severity/s0` work.

## Root cause

The executor handled count/file outputs but had no physical-table materialization path. Schema generation depended on name-based source lookup instead of the relation already resolved at the requested snapshot. The internal result path also exposed enum representation and timezone/transaction cleanup gaps.

## Resolution

Draft PR #25863 implements physical-table materialization, snapshot-correct schema resolution, transaction-safe cleanup, complete serialization for the affected type families, and BVT/unit coverage.

## Validation

- `diff_17` BVT: 138/138
- `branch_privilege.sql` BVT: 257/257
- Affected Go package tests, focused race tests, and `make build` pass.

## AI effort

`ai-light` was explicitly requested for this issue.

gouhongshen:
## What type of PR is this?

- [x] Feature
- [x] BUG
- [x] Test and CI

Fixes #25871

Implements `DATA BRANCH DIFF ... [COLUMNS (...)] OUTPUT AS <table>` as a durable ordinary physical table rather than a branch node or session temporary table.

- Preserves requested visible-column order and types, adds collision-safe `__mo_diff_source` / `__mo_diff_flag` metadata columns, and returns the materialized table name.
- Requires read privileges on both DIFF inputs and `CREATE TABLE` or database-wide equivalent on the destination database; rejects destination snapshots and never overwrites an existing table.
- Resolves output DDL from the already-resolved base relation. Current/current, current/snapshot, both-snapshot, ALTERed, dropped, and recreated source schemas use the exact historical columns.
- Resolves equivalent columns between target, base, and LCA schemas by stable `ColId` and `Seqnum`; a one-sided rename retains the target-side output name while ancestor reads use the physical ancestor name.
- Rejects a DIFF before materialization when target/base logical types differ in OID, width, scale, ENUM metadata, nullability, or auto-increment metadata.
- Completes output serialization for BIT, UUID, ENUM, DATALINK, GEOMETRY, and GEOMETRY32; preserves session timezone, SRID-constrained geometry, and all character control bytes.
- Preserves explicit-transaction ownership, atomic rollback, and partial-destination cleanup after materialization failures.
- Establishes a strict producer/materialization phase boundary for `OUTPUT AS`: producer batches are written to a private unlinked spill file and released immediately; only after `diffOnBase`, channel drain, and all producer waits complete can destination CREATE/INSERT execute through `BackgroundExec`. This prevents output SQL from advancing the shared transaction statement state during DIFF reads without retaining the result set in memory.
- Bounds the spool to one producer serialization buffer and one replay batch, validates private-format metadata before allocation, and closes the spill file on every exit path.
- Fixes DATA BRANCH LCA ENUM reconstruction and enum tuple/hash encoding, including snapshot range lookup and geometry logtail copy handling.

## User impact

Diff output can be retained and queried across sessions. The result table has no `mo_branch_metadata` row and is never treated as a data branch. Historical snapshot DIFFs remain materializable after the live source table changes or is recreated. Unsupported cross-schema DIFFs fail before a destination side effect. Compatible one-sided renames retain the renamed output column. SRID-constrained GEOMETRY and GEOMETRY32 values and VARCHAR/TEXT control-byte values round trip exactly. OUTPUT AS no longer mutates the transaction workspace while DIFF reads are in progress.

## Validation

- `make build`
- `go vet ./pkg/frontend` with repository CGo include and library paths
- `mo-cgo-test -count=1 -timeout=120s ./pkg/frontend`
- `mo-cgo-test -race -count=1 -timeout=240s ./pkg/frontend -run Test(DataBranchOutputAsTablePhase_BlocksOutputSQLUntilProducerCompletes|DataBranchOutputSpool_RoundTrip|DataBranchOutputSpoolRejectsOversizedMetadata|MaterializeDiffOutputAsTable_InsertFailureDropsDestination)`
- Formatter-to-MySQL-scanner unit round trip for every byte `0x00..0x1f` and `0x7f`.
- BVT `git4data/branch/diff/diff_17.sql`: 154/154, including non-empty explicit-transaction OUTPUT AS and historical-snapshot OUTPUT AS; generated result manually checked.
- BVT `git4data/branch/edge/branch_rename_output.sql`: 13/13; manually checked renamed output column, type, and changed row.
- BVT `git4data/branch/edge/branch_output_geometry.sql`: 12/12; manually checked SRID-constrained GEOMETRY and GEOMETRY32 output.
- BVT `git4data/branch/edge/branch_output_control_bytes.sql`: 11/11; manually checked VARCHAR and TEXT `HEX()` values.
- BVT `git4data/branch/privilege/branch_privilege.sql`: 257/257.
- `git diff --check`


XuPeng-SH (changes requested):
I found three correctness gaps that need to be closed before this can merge.

1. [Blocking] The destination schema is resolved from the live table name, not from the relation/snapshot that was actually diffed. diffOutputTable.createSQL builds CREATE TABLE AS SELECT ... FROM baseTableDef.DbName/baseTableDef.Name without the source snapshot. A valid diff such as t{snapshot=s2} AGAINST t{snapshot=s1} therefore depends on the current object named t: if it was dropped, the materialization fails; if it was recreated or altered, the output can fail on a missing projected column or inherit the wrong width/scale/nullability. The output DDL must be derived from the already-resolved relation schema, or the exact snapshot must be carried into the schema-only query. Please cover current/current, current/snapshot, both-snapshots, and a snapshot whose live name was dropped/recreated/altered.

2. [Blocking] Errors from OUTPUT AS are swallowed inside an explicit transaction. getBackExecutor returns a shared-transaction cleanup callback that always returns nil, while diffMergeAgency assigns err = deferred(err). Thus BEGIN; DATA BRANCH DIFF ... OUTPUT AS existing_table can report success even though CREATE TABLE failed. Preserve the incoming error in the shared-transaction cleanup path and add explicit-transaction success/failure/rollback tests, including an existing destination and an injected insert failure.

3. [Blocking] OUTPUT AS does not support the full set of user-visible column types accepted by data-branch diff. Every projected value is passed through formatValIntoString, whose switch rejects valid types such as BIT, UUID, ENUM, DATALINK, and GEOMETRY. A changed row containing one of these columns aborts materialization and removes the destination, despite the feature claiming to preserve visible source types. Please either make serialization complete for the supported table-type matrix (with round-trip BVT coverage) or explicitly validate and reject unsupported schemas before creating the destination.

The lifecycle itself is otherwise well bounded: the producer is canceled on the first consumer error, remaining batches are drained/released, SQL batches are capped by row/byte thresholds, and autocommit execution is wrapped in one transaction. The missing tests are semantic boundaries rather than more happy-path variants: snapshot/live-name drift, explicit transaction errors, supported type families, cancellation/insertion failure, and cleanup failure.

gouhongshen (commented):
## Codex automated review

OUTPUT AS has correctness failures for historical base snapshots, session-specific TIMESTAMP values, and several supported visible column types. Focused parser tests passed; frontend integration tests were blocked by missing CGo headers.

### P1 - Preserve the base snapshot when deriving the output schema (pkg/frontend/data_branch_output.go:416)
getRelations resolves a base `{snapshot=...}` through a cloned historical transaction, but this CTAS references only the current `db.table` and omits the snapshot timestamp. If the current table was dropped, renamed, or recreated after the snapshot, materialization fails or derives the wrong schema even though the historical relation is valid.

### P1 - Keep the user session timezone for generated TIMESTAMP literals (pkg/frontend/data_branch_output.go:517)
The formatter renders TIMESTAMP values using the user session timezone, but the generated INSERT is executed through a newly initialized backSession whose timezone defaults to `time.Local`. Since the literal has no offset and parsing applies the executor timezone, `SET time_zone` sessions silently store shifted timestamp instants.

### P2 - Serialize all user-visible column types before materialization (pkg/frontend/data_branch_output.go:517)
extractDataBranchSQLRowValue supports visible types such as BIT, UUID, ENUM, DATALINK, and GEOMETRY, but formatValIntoString has no cases for them and returns not-supported from its default branch. A non-NULL diff row with any of these columns creates the destination and then fails; the feature does not preflight or document this restriction.

XuPeng-SH (changes requested):
The previous snapshot, transaction-error, timezone, and special-type blockers are fixed on cb626795. One correctness blocker remains.

[Blocking] OUTPUT AS accepts width/scale-incompatible schemas and then materializes rows into the base-side type.

getTableStuff relies on isSchemaEquivalent before starting the diff, but that helper compares only Typ.Id and ignores Width, Scale, Enumvalues, and Typ.NotNullable. The new output path then creates projected columns from baseTableDef while formatting values with tblStuff.def.colTypes derived from tarTblDef. Those assumptions are inconsistent.

This is reachable through ordinary DDL, not corrupted metadata. VARCHAR widening is explicitly an inplace ALTER and preserves the column identity. For example:

create table base_t(id int primary key, payload varchar(20));
data branch create table branch_t from base_t;
alter table branch_t modify payload varchar(80);
insert into branch_t values (1, repeat(char(120), 40));
data branch diff branch_t against base_t output as diff_out;

The schemas currently pass isSchemaEquivalent, diff_out is created with payload VARCHAR(20), and a target row valid for VARCHAR(80) then fails during the generated INSERT (followed by cleanup). A focused unit assertion that these two TableDefs are not equivalent fails on this head because the helper returns true.

Please make the preflight schema contract match the downstream operations: at minimum compare the complete logical type metadata used by FormatColType and value formatting, including width, scale, enum metadata, and effective nullability, or explicitly implement compatible cross-schema materialization. Add a regression covering an inplace widened branch with a value longer than the base width; precision/scale and enum-definition mismatches should also be covered by the equivalence test matrix. The command must reject before CREATE when the schemas are unsupported.

XuPeng-SH (changes requested):
One correctness blocker remains on the latest head.

[Blocking] `OUTPUT AS` breaks the already-supported one-sided column-rename case. `getTableStuff` deliberately treats a rename as schema-equivalent (column identity/order/type are unchanged), and the existing BVT `branch_edge_cases.sql` verifies that `ALTER TABLE rename_left RENAME COLUMN b TO bb; DATA BRANCH DIFF rename_left AGAINST rename_right;` succeeds and exposes `bb`. However, `diffOutputTable.createSQL` takes the target-side output name from `tblStuff.def.colNames[idx]` and then looks that name up in `baseTableDef`. After the rename it searches the base schema for `bb`, where the corresponding column is still named `b`, and returns `DATA BRANCH DIFF OUTPUT AS base column "bb" is unavailable` before creating the destination.

I reproduced this deterministically with a focused unit test that changes the target descriptor name from `name` to `display_name` while retaining the base logical column/type; `output.createSQL` fails with exactly that error. This is ordinary supported DDL, not corrupted metadata.

Please resolve the base type by the schema-equivalence identity/ordinal (preferably stable column identity such as ColId/Seqnum), while using the target-side name for the destination DDL. Add an `OUTPUT AS` regression beside the existing rename BVT and verify the destination column name, type, and rows. A name-based base lookup is incompatible with the current DIFF schema contract.

heni02 (approved):

XuPeng-SH (changes requested):
The previous one-sided rename blocker is fixed on the latest head, but one materialization correctness gap remains.

[Blocking] SRID-constrained geometry values lose their SRID in the generated INSERT.

`diffOutputTable.createSQL` preserves the complete destination type through `plan.FormatColType`, so a source `POINT SRID 4326` column creates a `POINT SRID 4326` output column. However, `formatValIntoString` at `data_branch_helpers.go:512` always emits either `cast(st_geomfromtext(<WKT>) as geometry)` or the corresponding generic `geometry32` cast. The single-argument constructor carries no SRID, and the generic cast does not restore the declared SRID.

In MatrixOne the SRID is type metadata, not part of the stored WKB payload. The existing geometry contract explicitly rejects a value with no SRID when inserting into a SRID-constrained column (`geo_srid.sql` covers this exact rule). Therefore this ordinary supported shape fails after CREATE and the cleanup removes the destination:

```sql
create table base_geo(id int primary key, g point srid 4326);
insert into base_geo values (1, st_geomfromtext("POINT(1 1)", 4326));
data branch create table branch_geo from base_geo;
update branch_geo set g = st_geomfromtext("POINT(2 2)", 4326) where id = 1;
data branch diff branch_geo against base_geo output as diff_geo;
```

I reproduced the serialization side deterministically on current head `cdb381685`: a geometry type with `Width=4327` (SRID 4326) produces exactly `cast(st_geomfromtext("POINT(1 2)") as geometry)`, with no `4326`. The existing planner test `TestFuncCastForGeometrySRID` proves that this resulting no-SRID value is rejected by the SRID-4326 destination.

Please preserve the exact geometry expression type when serializing: use the explicit-SRID constructor when the source type declares an SRID, and do not erase it with a generic cast (or cast to the exact subtype/precision/SRID type). Add an OUTPUT AS BVT with a changed `POINT SRID 4326` row and assert both `ST_AsText` and `ST_SRID`; cover the same metadata path for GEOMETRY32 as well.

XuPeng-SH (changes requested):
The previous SRID blocker is fixed on the latest head. One silent-data-corruption blocker remains.

[Blocking] `OUTPUT AS` does not round-trip control bytes in VARCHAR/TEXT values.

`formatValIntoString` routes character values through `writeEscapedSQLString`. For control bytes other than the few named MySQL escapes, lines 669-671 emit `\\xNN`. However, the MatrixOne MySQL scanner `handleEscape` has no hex-escape case: it removes the backslash and leaves the three ASCII bytes `xNN`.

I verified the exact generated form against the current scanner: parsing the SQL literal produced for byte `0x01`, `\\x01`, returns `x01` with bytes `78 30 31`, while a literal byte `0x01` returns byte `01`. This is reachable with ordinary valid data, for example a VARCHAR value produced by `CHAR(1)`. The materialization succeeds and commits, so this is not a safe insert error followed by cleanup; the destination silently contains different data.

A regression shape is:

```sql
create table control_base(id int primary key, v varchar(10));
insert into control_base values (1, char(1));
data branch create table control_branch from control_base;
update control_branch set v = concat(char(1), char(65)) where id = 1;
data branch diff control_branch against control_base output as control_out;
select hex(v) from control_out;
```

The target UPDATE row must retain `0141`, not become ASCII `78303141` (or another escaped spelling). Please make the SQL literal encoding match the scanner contract, or use a byte-exact typed representation, and add a parser round-trip unit plus OUTPUT AS BVT assertions with `HEX()` for the affected control-byte range. The same helper is old, but this new materialization path directly relies on it and currently publishes the corrupted table as success.

Other reviewed closure properties look sound: historical relation/schema identity, one-sided rename, complete logical-type preflight, explicit-transaction error propagation, SRID/ENUM handling, cancellation/drain/pool cleanup, bounded channel and SQL batching. Focused packages, full frontend, race, and 50 repeated boundary tests passed on head `3dde08859f8dde750b1754615e47b6ba3846b22a`.

gouhongshen (commented):
## Codex automated review

Found one blocking concurrency/transaction-consistency regression. Focused tests were blocked by missing CGo headers.

### P1 - [P1] Serialize OUTPUT AS writes with diff execution (pkg/frontend/data_branch_output.go:552)
The output consumer runs before `diffOnBase` and flushes batches through `bh.Exec`. In the normal path this is a non-derived background executor, so each INSERT opens a workspace statement and advances the shared transaction state. The producer uses the same `bh` transaction, while its internal SQL deliberately uses `WithDisableIncrStatement()` to avoid statement boundaries. These operations therefore mutate the same transaction workspace concurrently, while source/LCA reads are still running, violating the transaction's statement-boundary assumptions and allowing definition-change retries or inconsistent visibility/results. Materialize after the producer completes, or otherwise serialize the output SQL with the diff execution.

XuPeng-SH (changes requested):
Re-reviewed the complete latest head. The previous schema, rename, type, SRID, control-byte, timezone, and cleanup issues are fixed, but one P1 correctness blocker remains.

[Blocking][P1] DATA BRANCH DIFF OUTPUT AS performs destination DDL/DML concurrently with the diff producer on the same BackgroundExec and TxnOperator.

In diffMergeAgency, the output consumer goroutine starts at pkg/frontend/data_branch.go:797 and then diffOnBase runs on the caller goroutine at line 826. materializeDiffOutputAsTable immediately executes CREATE TABLE and later flushes INSERT statements through bh.Exec at pkg/frontend/data_branch_output.go:539 and :552 while diffOnBase is still resolving and reading target, base, and LCA state.

For the normal non-explicit-transaction path, getBackExecutor returns a non-derived background executor. Each output bh.Exec therefore enters StartStatement and IncrStatementID, while producer runSql calls reuse that same transaction with WithDisableIncrStatement. The output path can advance the RC statement snapshot and workspace offsets in the middle of one logical DIFF. For explicit transactions, the executor is derived, but snapshot/name fallback paths can still invoke the same mutable backExec/backSession concurrently. backSession result state, compile context, and stats are not a concurrent-execution boundary.

This violates the statement-view invariant even if current BVT results happen to be stable. The new unit test preloads and closes retCh before calling materializeDiffOutputAsTable, so it never exercises the production overlap; the passing BVTs only assert final rows.

Please establish an explicit phase boundary so destination CREATE/INSERT work cannot overlap producer execution or advance its statement boundary mid-DIFF. For example, prepare CREATE before producer admission and materialize buffered or spooled rows only after producer completion, or use another design that proves the same stable-snapshot and transaction-atomicity contract without unbounded memory. Add deterministic phase tests for both autocommit and explicit transactions, including a snapshot/fallback path: block the producer at a known phase and assert that unsafe output SQL cannot enter, then run the focused transition under -race.

gouhongshen (commented):
## Codex automated review

No blocking or non-blocking findings remain on e8cc570bade6ec43e45c4fa475368165fad5c659. Earlier review issues are fixed, including snapshot-derived schemas, explicit-transaction error propagation, type serialization, renamed columns, geometry SRIDs, control bytes, and concurrent transaction use. The new spool establishes a bounded-memory producer/materialization phase boundary with clear batch and file cleanup ownership. Focused tuple tests and current CI pass; local frontend race tests could not run because libmo.dylib is absent.

fengttt (approved):

XuPeng-SH (approved):
Final deep review of e8cc570bade6ec43e45c4fa475368165fad5c659: no blocking or non-blocking findings remain.

Revalidated every prior blocker: historical snapshot schema resolution (including dropped/recreated live names), explicit-transaction error propagation/commit/rollback, complete logical-type equivalence, one-sided rename identity mapping, session timezone, BIT/UUID/ENUM/DATALINK/GEOMETRY serialization, geometry/geometry32 SRID preservation, control-byte round trips, destination cleanup, and the shared-transaction concurrency bug.

The new spool establishes a real producer -> materializer phase boundary: the producer and drainer both finish before markProducerDone permits CREATE/INSERT; the drainer releases every emitted batch on success/cancel/write failure; the phase owns and closes the unlinked spill file and replay batch; memory remains bounded by the 10-batch channel, one serialized/replay batch, and capped SQL buffers. Cancellation and errors drain/wait before pool cleanup, and partial destination creation is dropped.

Fresh validation:
- merge-tree with current origin/main c0486a2446 is clean; merged-tree tests pass
- full pkg/frontend tests and full -race pass; focused phase/spool/cleanup tests pass under -race -count=50
- pkg/container/types, pkg/frontend/databranchutils, pkg/vm/engine/disttae/logtailreplay, and MySQL parser packages pass
- go vet for all touched Go packages and make build pass
- real mo-service: generate_series-backed 20,000-row diff materialized and committed in an explicit transaction; rollback removed the destination; existing-table failure returned ERROR 1050 without modifying the sentinel; snapshot drop/recreate, one-sided rename, control byte 0x01, ENUM/UUID, and SRID 4326 values all round-tripped exactly

The current CI matrix is green and there are no unresolved review threads.

PR #25864, opened 2026-07-18
fix: guard empty fixed-width vectors in CBloomFilter to avoid index-out-of-range panic

SAY-5:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25618

The three fixed-width `CBloomFilter` methods (`testAndAddFixedVector`, `testFixedVector`, `addFixedVector`) take `&fixedData[0]` and `&results[0]` unconditionally, which panics with `index out of range [0] with length 0` when the input vector is empty. Their varlena counterparts (`testAndAddVarlenaVector`, `testVarlenaVector`, `addVarlenaVector`) already return early on `v.Length() == 0`, so behavior on empty input was inconsistent by type.

This adds the same zero-length early return to the fixed-width path so `TestVector`/`AddVector`/`TestAndAddVector` are no-ops on an empty vector instead of panicking, matching the varlena behavior. A regression test covers all three fixed-width APIs.

### DIFF
--- a/pkg/common/bloomfilter/cbloomfilter.go
+++ b/pkg/common/bloomfilter/cbloomfilter.go
@@ -266,6 +266,9 @@ func (bf *CBloomFilter) AddVector(v *vector.Vector) {
 }
 
 func (bf *CBloomFilter) testAndAddFixedVector(v *vector.Vector, callBack func(bool, bool, int)) {
+	if v.Length() == 0 {
+		return
+	}
 	fixedData := v.GetData()
 	typeSize := v.GetType().TypeSize()
 	length := v.Length()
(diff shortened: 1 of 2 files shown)


gouhongshen (approved):
## Codex automated review

No concrete regressions found. The guards correctly make empty fixed-width vectors no-ops, matching the existing varlena behavior and issue contract. Local focused tests could not run because Go is unavailable; PR CI reported success.

XuPeng-SH (approved):
Deep review completed against f58f0260854383f19ad6a8757764d12ac4515f56. No merge-blocking findings.

The zero-length guard is placed before both fixedData[0] and results[0] in all three fixed-width vector paths. TestVector returns an empty result, while AddVector and TestAndAddVector are no-ops and invoke no callbacks, matching the existing varlena semantics. The non-empty path is unchanged apart from an O(1) length check, and the change introduces no new ownership, allocation-growth, concurrency, or wait-for edges.

Local verification passed:
- focused regression test
- full pkg/common/bloomfilter test suite
- full package with -race
- go vet

Non-blocking follow-up: the regression test could pass a counting callback and assert zero invocations instead of passing nil. Fixed const-null vectors with logical length greater than zero and empty backing data are a separate pre-existing API boundary and should be handled independently; this PR neither introduces nor claims to fix it.

Approved.

Issue #25862, opened 2026-07-18
[Performance]: Avoid full probe-table scans for right SINGLE joins without breaking multi-CN execution


XuPeng-SH:
### Is there an existing issue for performance?

- [x] I have checked the existing issues.

### Environment

- Version or commit-id: `main` at `946427b877c1b12c497743eff264047e60a00be2`
- Deployment: two-CN `make dev-up` cluster
- Hardware parameters: 24 CPU cores; 6 GiB file-service memory cache per CN
- OS type: Linux

### Details of Performance

A correlated scalar subquery can be flattened into a right `SINGLE` hash join when the preserved outer relation is much smaller than the inner table:

```sql
CREATE DATABASE rsingle_perf;
USE rsingle_perf;

CREATE TABLE big(id BIGINT PRIMARY KEY, v BIGINT);
INSERT INTO big
SELECT result, result * 10
FROM generate_series(1, 1000000, 1);

CREATE TABLE small(id BIGINT PRIMARY KEY);
INSERT INTO small VALUES (1), (5000000), (10000000);

ANALYZE TABLE big;
ANALYZE TABLE small;

SELECT s.id,
       (SELECT b.v FROM big b WHERE b.id = s.id) AS v
FROM small s
ORDER BY s.id;
```

The optimizer physically swaps the scalar `SINGLE` join so that the small preserved relation is the hash-build side and `big` is the probe table:

```text
Join Type: RIGHT SINGLE
Join Cond: (b.id = s.id)
├── Table Scan: big    -- probe; unmatched rows are discarded
└── Table Scan: small  -- build and preserved side
```

On `main`, `generateRuntimeFilters` rejects every `Node_SINGLE` join. As a result, the query scans and probes all rows from `big`, even though the build side contains only three keys.

A one-CN prototype that generates a runtime filter only for right `SINGLE` reduced the work as follows:

| Metric from `EXPLAIN ANALYZE` | Main-equivalent path (RF disabled) | Right-SINGLE RF prototype |
| --- | ---: | ---: |
| `big` input blocks | 123 | 1 |
| `big` input rows | 1,000,000 | 1 |
| Join input rows | 1,000,003 | 4 |
| Scan read size | 7.68 MiB | 96.33 KiB |
| Warm-cache client latency | 0.21 s | 0.21 s |

[...]

PR #25865, opened 2026-07-19
perf(plan): prune right SINGLE probe scans with runtime filters

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25862

Right SINGLE joins produced by correlated scalar-subquery flattening preserve the build side, but previously could scan the entire discardable probe table because SINGLE joins could not build runtime filters.

This PR adds a conservative, exact-IN runtime-filter path for eligible physical right SINGLE joins.

- Centralizes runtime-filter eligibility so planning, one-CN placement, statistics, and compile-time validation use the same policy.
- Restricts the optimization to direct table scans, supported key types, a leading primary/cluster-key prefix, no probe LIMIT/OFFSET, and build/probe cardinalities that fit the exact-IN limit.
- Colocates the producer and all consumers on one CN instead of extending the current-CN runtime-filter protocol across CNs.
- Validates the final physical topology before pipelines start: exactly one producer, every consumer colocated, and no ambiguous partial metadata.
- Records right-SINGLE nodes actually visited by `compilePlanScope`, so logically retained but physically pruned subtrees (for example branch-level `LIMIT 0`) are excluded without weakening validation for reachable subtrees.
- Sends DROP for empty or aborted builds and preserves query-context cancellation while consumers wait.
- Corrects right SINGLE cardinality after physical child swapping and propagates the corrected estimate to parent nodes.
- Preserves the existing fallback for ineligible, expensive, non-leading-key, or uncertain shapes.
- Keeps `disableRightSingleRF=1` as a diagnostic rollback switch.

### Correctness and unhappy-path coverage

- Match, miss, NULL, empty build, residual predicates, duplicate scalar results, no-PK fallback, composite-key full/prefix cases, and UPDATE/INSERT/DELETE consumers.
- Missing/duplicate producers, fully missing reachable topology, physically pruned topology, remote consumers, ambiguous metadata, standalone execution, shared scope graphs, and cancellation while waiting.
- Branch-level `LIMIT 0` + `UNION ALL` + right-SINGLE regression from review.
- Single-CN and 2-CN BVT execution without assuming a fixed deployment topology.
- Race testing covers plan, compile, hashbuild, hashjoin, and runtime-filter message packages.

### Performance evidence

For an eligible 100,000-row probe table, `EXPLAIN ANALYZE` showed:

| mode | probe rows | probe blocks |
| --- | ---: | ---: |
| right SINGLE runtime filter enabled | 5,000 | 1 |
| disabled fallback | 100,000 | 13 |

The compile-time no-filter topology path remains allocation-free at approximately 2.26 ns/op. Queries without a right-SINGLE join allocate no compiled-node tracking state.

### Phase boundary and follow-up

Phase 1 intentionally keeps the cardinality correction in a post-swap-only pass. Under the current frozen physical orientation, a right SINGLE uses physical child 1 as the semantic preserved side; the downstream-join regression proves that consumers see the preserved cardinality instead of the discardable probe cardinality.

This positional rule is a tested bridge, not the long-term optimizer model. A follow-up should introduce an explicit semantic join descriptor with stable preserved/discardable relation identities, keep physical probe/build orientation separate, migrate stats/RF/compile/explain consumers, and then remove the `IsRightJoin + child index` inference.

### Validation

- Full tests: `pkg/sql/plan`, `pkg/sql/compile`, `pkg/sql/colexec/hashbuild`, `pkg/sql/colexec/hashjoin`, and `pkg/vm/message`
- The same package set with `-race -p=1`
- `go vet` for the same package set and a full `make -j8 build`
- 34/34 semantic BVT statements from both CN entry points in a 2-CN deployment, then 34/34 again after membership converged to single-CN
- Shuffled compile/plan package tests, 3 repetitions
- 31,000 repeated standalone BVT statements
- 50,000 successful 2-CN stress executions
- 50 cancellation cycles while a consumer waited for its runtime filter; no stuck waiters or goroutine growth
- Heap-profile review after stress found no runtime-filter-specific retained ownership chain


pkg/sql/compile/runtime_filter_topology.go
```diff
@@ -0,0 +1,217 @@
...
+	consumers map[int32][]*Scope
+}
+
+func rightSingleLocalRuntimeFilterTags(qry *plan.Query) []int32 {
+	if qry == nil {
+		return nil
+	}
+
+	var tags []int32
+	for _, node := range qry.Nodes {
```
LeftHandCold: [P1] Do not validate tags from compile-pruned subtrees

This walks every node in the flat query, including subtrees that compilePlanScope intentionally does not compile. A reachable case is a right-SINGLE scalar subquery under a LIMIT 0 branch of UNION ALL: because the outer query itself is not LIMIT 0, planning uses real stats and generates the right-SINGLE RF; then compilePlanScope sees the branch-level literal LIMIT 0 and returns an empty scope without compiling that join. This function still returns the join tag, so validation deterministically fails with `tag ... has no scan consumer` even though the query should simply return the other UNION branch. I reproduced the planned shape with a right-SINGLE RF below the branch LIMIT 0. Please derive the tags from plan subtrees that were actually compiled, or explicitly exclude compiler-pruned subtrees, and add a nested LIMIT 0 / UNION regression test while keeping the missing-producer protection for executed subtrees.

LeftHandCold (changes requested):
Deep review found one reproducible compile-time regression in the new topology validation. The right-SINGLE semantics, physical side swap, exact-IN bounds, message delivery/cancellation, and resource lifecycle otherwise look consistent. Local full package tests, focused race tests, build, vet, and diff checks passed.

aptend (changes requested):
I confirmed the compile-time regression identified in the existing review. `rightSingleLocalRuntimeFilterTags` scans every node in the flat query, while `compilePlanScope` prunes literal `LIMIT 0` subtrees before producing scopes. A right-SINGLE scalar subquery below a branch-level `LIMIT 0` in `UNION ALL` therefore leaves a planned RF tag with no compiled consumer, and topology validation rejects the otherwise valid query with `tag ... has no scan consumer`.

Please derive the validated tags from subtrees that were actually compiled (or explicitly track compiled node IDs), while retaining missing/duplicate-producer validation for reachable subtrees. Add a regression covering branch-level `LIMIT 0` + `UNION ALL` + right-SINGLE.

pkg/sql/compile/runtime_filter_topology.go
XuPeng-SH: Fixed in 5a4407e64. compilePlanScope now records the right-SINGLE node IDs it actually visits, after the literal LIMIT 0 shortcut. Topology validation derives tags only from that materialized set, so a pruned subtree is ignored while an executed subtree with zero, one-sided, duplicate, or remote endpoints still fails validation. Added unit coverage for pruned versus fully missing reachable topology and the requested branch-level LIMIT 0 + UNION ALL + right-SINGLE BVT. Final validation on current main: focused/full tests, race, vet, build, 34/34 on both 2-CN entry points, and 34/34 after convergence to single-CN.

XuPeng-SH:
@aptend The confirmed compile-pruning regression is fixed in 5a4407e64 and pushed. The compiler now tracks actually visited right-SINGLE node IDs after pruning; validation remains strict for every reachable topology, including fully missing endpoints. The requested LIMIT 0 + UNION ALL regression and focused unhappy-path unit matrix are included. Full/race/vet/build and final single-/2-CN BVT validation pass; ready for re-review.

aptend (approved):
Re-reviewed head 595093fc9. The compile-pruning regression is fixed: right-SINGLE nodes are recorded only after the literal LIMIT 0 shortcut, so pruned branches are excluded while materialized topologies still fail closed for missing, duplicate, or non-colocated endpoints. The mixed pruned/materialized unit matrix and LIMIT 0 + UNION ALL regression cover the reported failure. Full compile/plan tests, focused race tests, vet, and diff checks passed locally. No remaining merge blocker found.

LeftHandCold (approved):
Re-reviewed head 595093fc91cdcc79c79d87c07cdfadd429f4539b. The previous LIMIT 0 plus UNION ALL compile-pruning regression is fixed: compilePlanScope records right-SINGLE nodes only after the pruning shortcut, so pruned branches are excluded while every materialized topology still fails closed for missing, duplicate, or non-colocated endpoints. The tracking state is statement-local and cleared on all exits. Full compile and plan package tests, focused race tests, build, vet, and diff checks passed locally. No remaining merge blocker found.

gouhongshen (changes requested):
## Codex automated review

发现一个会导致错误结果或运行时 panic 的 right-SINGLE 并行 producer 问题；另有未知统计信息导致的明显性能回退。

### P1 - Validate parallel HashBuild producers, not only producer scopes (pkg/sql/compile/runtime_filter_topology.go:126)
拓扑校验按 Scope 计 producer，但 ForceOneCN 只限制 CN 数，不限制 Stats.Dop。对足够大的直接 build scan，buildScanParallelRun 会通过 newParallelScope 复制 HashBuild；dupOperator 又复制同一个 RuntimeFilterSpec，因此每个副本都会发送同一 tag 的 IN/PASS/DROP。消费者只等待首批消息：收到单个分片 IN 会丢失其他分片的匹配行，收到 DROP 会错误清空扫描；同时多个 IN 会让 handleRuntimeFilters 用 rfSpecs[i] 越界并 panic。必须强制该 producer DOP=1，或实现带完成信号的跨分片 filter 聚合。

### P2 - Do not treat default statistics as a safe build-size bound (pkg/sql/plan/runtime_filter.go:141)
此检查把 build.Stats.TableCnt/Outcnt 当作 exact-IN 的硬上限，但 calcScanStats 在 skip/error/nil stats 时返回 DefaultStats（TableCnt/Outcnt=1000、BlockNum=1），且这里没有排除 IsDefaultStats。若 probe 侧统计足以通过 RF 启发式而 build 统计不可用，实际大型 build 仍会被强制放到一个 CN；HashBuild 读完整个 build、发现实际基数超过上限后发送 PASS，既没有剪枝收益又引入串行扫描和额外内存压力。

XuPeng-SH:
Addressed both findings in f630a08b08 and merged latest main in d2d748ebe5.

P1: CalcQueryDOP now serializes only the direct build scan that owns a phase-1 right-SINGLE exact-IN producer. The probe scan and unrelated ForceOneCN plans retain their normal intra-CN DOP. Physical topology validation also rejects a pending HashBuild expansion when producer scope Mcpu is greater than one, so scheduling drift fails before execution instead of producing partial IN/DROP/PASS messages.

P2: right-SINGLE eligibility now rejects DefaultStats. The 1,000-row sentinel is no longer treated as a real cardinality bound, so unavailable stats preserve the existing plan instead of forcing an unknown build onto one CN.

Added positive and negative DOP tests, default-stats coverage, and parallel-scope topology coverage. On latest main: full pkg/sql/plan and pkg/sql/compile tests, focused race tests, build, vet, and make static-check all pass.

XuPeng-SH:
Deep re-review of the latest changes-requested is complete and the final tree is pushed in c48017a6d (semantic changes in b02e5b7a4).

- P1 premise: the current physical compiler does not clone HashBuild with a parallel build scan. compileBuildSideForBroadcastJoin first wraps an Mcpu>1 build scan in a single-thread Merge, then attaches one HashBuild above that Merge. I added TestParallelBuildScanIsMergedBeforeRuntimeFilterHashBuild to lock down the actual TableScan -> Connector -> Merge -> HashBuild shape. I did not retain the proposed build Dop=1 serialization because it removes safe scan parallelism. The topology collector is nevertheless hardened to count post-ParallelRun HashBuild multiplicity if a future physical plan ever places HashBuild directly in a parallel scan scope, and fails closed in that case.
- P2 is valid and fixed: DefaultStats is now rejected as an exact build-size bound, so unavailable stats cannot force an arbitrarily large scan onto one CN only to produce PASS.
- An independent reachable correctness bug was also fixed: when multiple RF specs contain PASS before a later IN, the compacted expression list previously lost its originating spec and could route a PK filter as non-PK or vice versa. Received IN expressions now retain their exact RuntimeFilterSpec; both orderings have regression coverage.
- Added leading cluster-key coverage for block pruning plus exact non-PK row filtering.

Validation:
- relevant compile/plan/hashbuild/hashjoin/message packages pass normally and under race; vet and make build pass
- standalone and 2-CN BVT: 37/37
- no-filter topology benchmark: ~2.05 ns/op, 0 B/op, 0 allocs/op
- real 2-CN run: 3M-row probe + 200K-row build; build scan remained Dop=2, probe Dop=24, one HashBuild producer, RF reduced probe input to 200K rows / 26 blocks, and the result was 200000 rows with 0 missing

This preserves correctness, scan parallelism, cancellation/topology fail-closed behavior, and adds no goroutine, lock, channel, FD, or retained resource.

LeftHandCold (approved):
Re-reviewed head 70a366a5172644fbb57c162e10384d280fa5cdd7 with a focus on data correctness and normal-query regressions. The parallel build scan remains parallel below a single Merge and therefore has exactly one HashBuild runtime-filter producer; topology validation also fails closed if that physical invariant changes. DefaultStats is no longer accepted as a hard build-size bound, compile-pruned LIMIT 0 branches remain excluded, and each received IN payload now retains its originating spec across preceding PASS messages. Right-SINGLE preserved-side stats, exact-IN eligibility, cancellation/reset termination, and existing join/runtime-filter policies remain consistent. Full plan and compile package tests, focused race tests, dependent hashbuild/hashjoin/message tests, build, vet, diff checks, and current CI all pass. No remaining blocker found.

heni02 (approved):

gouhongshen (approved):
## Codex automated review

No concrete regressions found in the reviewed diff.

aptend (approved):
Re-reviewed current head ea10dc4bd. The parallel build scan is merged before the single HashBuild runtime-filter producer; the DefaultStats fallback and PASS/IN spec association are fixed. Relevant package tests, focused race tests, go vet, and go build all pass. No blocking issues found.

PR #25867, opened 2026-07-19
logservice: support offline WAL recovery

### LINKED ISSUE #24966: LogService HAKeeper can regress to stale shard membership after Dragonboat imported snapshot metadata rewrite

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24966.

Adds an explicit offline LogService recovery path for rebuilding a three-replica LogService cluster from artifacts generated by `mo_br`.

The recovery path is enabled only when `logservice.BootstrapConfig.restore.wal-data-path` is configured. Normal LogService startup and heartbeat behavior are unchanged when the option is absent.

Key behavior:

- Requires a valid HAKeeper backup and WAL data file and fails closed for missing or malformed recovery input.
- Restores `NextID` and `NextIDByKey` through replicated, max-only HAKeeper updates, including when another LogService is already the HAKeeper leader and the cluster has already entered bootstrap.
- Keeps HAKeeper out of the running state while WAL recovery is pending.
- Selects one recovery coordinator from replicated LogStore heartbeat state and sends replay requests to the current Log shard leader, so recovery does not depend on the restore pod being leader.
- Validates WAL ordering, bounds, and digest, and persists replay progress for crash-safe resume without duplicating recovered records.
- Writes the restore marker only after HAKeeper ID watermarks have been verified.

Operational constraint: all three LogService pods must run this version during recovery. Only one pod should be configured with the recovery files.

Validation:

- `go test -count=1 -timeout=15m ./pkg/hakeeper/... ./pkg/logservice ./pkg/pb/logservice ./pkg/util`
- `go test -race -count=1 -timeout=20m ./pkg/hakeeper ./pkg/logservice`
- `go vet ./pkg/hakeeper/... ./pkg/logservice ./pkg/pb/logservice ./pkg/util`
- `go test ./cmd/mo-service ./pkg/embed -run '^$' -count=1`
- `go build ./cmd/mo-service`
- `make err-check`

End-to-end validation on an isolated local kind cluster:

- Seeded an old cluster with 1,001 rows and 10 databases.
- Copied the three live LogService data directories and generated fresh recovery artifacts with `mo_br`.
- Recreated all three LogService PVCs and copied only `hakeeper_backup.data` and `wal_data.bin` to `log-0`.
- Started `log-1` and `log-2` first: `log-2` became HAKeeper leader and `log-1` initialized the cluster before `log-0` started.
- Started `log-0`: all HAKeeper replicas applied `NextID=50000000` and the complete `NextIDByKey`; 115/115 WAL entries were replayed.
- Verified all original rows/databases and created a new database/table with IDs `50000001` and `50000002`.
- Removed recovery configuration and restarted all three LogService pods. No WAL recovery ran; data remained intact and subsequent IDs advanced to `50010341` and `50010342`.


gouhongshen (changes requested):
## Codex automated review

Found four blocking recovery-safety bugs and two material limit issues. Focused tests were blocked locally by missing CGo headers.

### P1 - Require a clean destination before initializing replay state (pkg/logservice/wal_recovery.go:800)
When no state file exists, BaseLSN and LastLSN are initialized to the destination's current latest LSN. Existing records are therefore treated as baseline, and the old WAL is appended after them and marked complete. The bootstrap path can run after HAKeeper and the Log shard are already initialized, so a restart or late recovery can silently duplicate WAL history instead of failing closed.

### P1 - Reject recovery after HAKeeper has entered Running (pkg/hakeeper/rsm.go:808)
RestoreIDWatermarkUpdate sets LogServiceRecoveryPending even when the current state is HAKeeperRunning. The Running transition guard only applies while the current state is BootstrapCommandsReceived, and Running has no transition back to a gated state. With staggered startup, HAKeeper can reach Running from a quorum before the recovery pod heartbeats; the late pod then starts replay while the cluster is serving.

### P1 - Invalidate the cached ID allocator after restoring watermarks (pkg/logservice/store_hakeeper_check.go:142)
This updates the replicated HAKeeper watermark but leaves l.alloc unchanged. updateIDAlloc preallocates a 10,240-ID range, and getScheduleCommand reuses any remaining range with capacity >=1024. If bootstrap scheduling allocates a low range before a late recovery pod applies a higher backup watermark, those cached low IDs are handed out after recovery and can collide with IDs referenced by restored data.

### P1 - Fence user appends during WAL replay (pkg/logservice/wal_recovery.go:878)
Replay sends APPEND requests via doAppend, while the recovery status only gates HAKeeper state and ID allocation. LogService's normal append handler has no recovery-pending fence, so another valid writer can insert a user record between replay entries. The reconciliation loop explicitly rejects such records as unexpected or mismatched, leaving recovery failed and HAKeeper pending.

### P2 - Bound accepted WAL entries by the actual RPC message limit (pkg/logservice/wal_recovery.go:51)
The parser accepts RawData up to defaultMaxMessageSize (100 MiB), but the recovery client uses cfg.RPC.MaxMessageSize and morpc validates RPCRequest.ProtoSize(), which includes the protobuf request plus the APPEND command header and payload. Thus an entry at the parser limit, and entries near it under the default RPC limit, are accepted from disk but rejected before transmission.

### P2 - Avoid loading the full recovery artifact into memory (pkg/logservice/wal_recovery.go:339)
The code permits a 1 GiB file while retaining every RawData slice and WALEntry in memory. A valid artifact with millions of entries adds hundreds of megabytes of slice/struct/allocation overhead on top of the file contents, potentially exhausting a LogService before replay begins.

XuPeng-SH (changes requested):
Re-reviewed current head 4af6aed83b64b4f3803cab5e12566c02bdbb036e. The previous fence, streaming, RPC-limit, and local allocator fixes are meaningful, but the recovery protocol still has four fail-closed gaps.

1. [P1] Bind durable completion to the destination shard incarnation before skipping replay (pkg/logservice/wal_recovery.go:119-125, 948-959).

The sidecar and legacy tag are keyed only by the WAL path/digest, not by the destination Log shard. A normal operational retry is: recover successfully, recreate the LogService data/PVC again, and reuse the same recovery-artifact directory. The sidecar still says Complete (or the legacy tag exists), while the new shard is empty. Both branches return success without inspecting the destination, BootstrapHAKeeper then clears LogServiceRecoveryPending, and the empty cluster can enter Running. Force cannot recover this because it is explicitly ignored. Please bind progress/completion to a durable destination incarnation or recovery marker in the shard and verify the completed prefix before returning; an empty replacement shard must replay, while a non-empty unrelated shard must fail closed.

2. [P1] Make watermark restore a global barrier before any old-generation ID can escape (pkg/hakeeper/rsm.go:788-823, pkg/logservice/store_hakeeper_check.go:465-480).

LogService recovery is accepted in HAKeeperBootstrapCommandsReceived, but reaching that state means bootstrap schedule commands may already contain replica IDs allocated below the backup watermark. Incrementing IDWatermarkRestoreGeneration only discards the local checker's remaining cache on a future call; it cannot retract already replicated commands, and an in-flight checker that read the old generation can still propose an old-ID batch after the restore command. The non-WAL path is even allowed in HAKeeperRunning, while managed HAKeeper clients retain their own shared/keyed ID batches and never observe this generation. In either sequence, IDs already present in restored data can be emitted after the watermark update. The restore intent/watermarks need to be established before scheduling/client allocation is possible, or command proposals and every cached allocator need a replicated generation check that rejects stale work.

3. [P1] Verify the bytes actually replayed, not only the first scan of the file (pkg/logservice/wal_recovery.go:344-449, 905-912, 463-480).

The first pass hashes the file and retains offsets, then closes it. Replay can wait for leader/coordinator election and later reopens the path; each payload is read without comparing it to the bytes covered by WALDigest. Replacing or modifying the artifact during that window therefore appends different data, but completion is persisted against the old digest and recovery reports success. Keep an immutable source snapshot/file identity and validate each streamed payload (for example, hashes collected in the metadata pass), with a final full-source verification before Complete as a backstop.

4. [P1] Validate the semantic LogEntry envelope before making HAKeeper Running (pkg/logservice/wal_recovery.go:402-444, 478-503).

Current validation checks only the outer recovery header's raft index and DSN overflow. RawData may be one byte, have an invalid type/version/footer, or encode a start DSN / entry count different from the outer DSN / EntryCount. Such input is appended and counted as recovered; the integration fixture itself uses an incomplete synthetic envelope, so it cannot catch the TN replay failure. Validate the minimum LogEntry layout, header/version/command, footer bounds, and equality of the embedded DSN/count before fencing/replay, then add a TN replay/e2e assertion rather than stopping at LogService record equality.

Validation on this head:
- focused WAL recovery, state, coordinator, fence, and allocator tests: pass
- same focused set with the race detector: pass
- HAKeeper package matrix: pass
- GitHub UT/SCA/BVT checks: green

The full local pkg/logservice run was not usable as PR evidence because existing local MO/Docker listeners on fixed test ports 32001/26001 caused two unrelated tests to connect to the external service; the targeted current-head tests above completed cleanly.

gouhongshen (changes requested):
## Codex automated review

Four P1 correctness blockers can produce incomplete or corrupt recovery.

### P1 - Completion state can skip replay on a new destination (pkg/logservice/wal_recovery.go:948)
The durable state is not bound to a destination shard incarnation. When Complete is true, this branch returns without checking the current destination prefix. Rebuilding the Log shard while retaining the WAL artifact and sidecar therefore skips replay and can clear HAKeeper's pending gate with an empty shard. The legacy-tag path has the same issue.

### P1 - Restore generation does not fence old ID allocations (pkg/hakeeper/rsm.go:856)
This guard still permits GetIDUpdate while HAKeeper is Bootstrapping even when recovery is pending, and schedule-command updates are also accepted. Only one local allocator is reset; queued bootstrap commands and managed-client caches retain old ranges. If bootstrap allocates IDs before the restore watermark is applied, those IDs can collide with restored IDs.

### P1 - Replay trusts a second, unverified WAL read (pkg/logservice/wal_recovery.go:473)
The initial scan hashes the WAL, but replay reopens the path and reads payloads here without verifying them against that digest. A same-size replacement or in-place mutation is accepted; reconciliation compares against the same altered read, and completion is recorded with the old digest. Recovery can therefore commit data different from the validated artifact.

### P1 - Malformed raw payloads are accepted as WAL entries (pkg/logservice/wal_recovery.go:410)
Parsing only bounds dataLen and validates outer metadata; zero-length or arbitrary payloads pass. They are appended as UserRecord entries, while the downstream WAL decoder slices the LogEntry header and can panic or fatal on malformed type/version/footer data. Validate the LogEntry envelope and its embedded DSN/count before appending or marking recovery complete.

gouhongshen (changes requested):
## Codex automated review

The documented per-member recovery configuration fails when the backup artifact is available only on the WAL-recovery member and uses a custom path.

### P1 - Do not make a missing non-coordinator backup file fatal (pkg/logservice/service_bootstrap.go:337)
Every initial member enters recovery mode, while the new config contract says only the WALDataPath owner needs recovery artifacts. BootstrapHAKeeper nevertheless calls getBackupData before joining bootstrap on every member. If a non-coordinator has the shared explicit restore.file-path but no local backup mount, this branch returns ErrFileNotFound (and absolute paths fail similarly at os.ReadFile), aborting bootstrap before it can participate in the three-replica recovery. Before this PR, a missing backup file was ignored for every path. The new integration test uses one service with the backup present, so it misses this topology. Treat an absent backup as expected on non-WAL members and retain the existing coordinator-side requirement for a valid backup.

gouhongshen (changes requested):
## Codex automated review

Four blocking correctness issues remain. Pure-Go protobuf/bootstrap tests passed; CGo-transitive LogService tests could not run locally because cgo/libmo.dylib is absent.

### P1 - Keep TN bootstrap behind the WAL recovery barrier (pkg/logservice/store_hakeeper_check.go:355)
Once ID watermarks are prepared, this guard falls through while recovery is still pending. Bootstrap then emits LogService and TN StartReplica commands together, but the coordinator cannot import WAL until the Log replicas report and a leader is confirmed. A TN can therefore finish its one-shot write-mode WAL replay against the empty destination first; it does not tail records imported afterward. HAKeeper can later enter Running with that TN missing recovered transactions. Stage TN startup after replicated recovery completion, or make TN replay wait through the recovery barrier, and test this interleaving.

### P1 - Do not advance SafeDSN through the current batch (pkg/logservice/wal_recovery.go:788)
SafeDSN zero is legitimate before any prior group commits, but contiguousDSN already includes the current entry when this branch replaces it. For a batched range [10,20], this produces SafeDSN=20. TN replay initializes its scheduled watermark to safe-1 and indexes each LogEntry only by its start DSN, so it looks for DSN 20, misses the mapping at 10, treats the miss as a safe gap, and never schedules DSNs 10-20. Recovery still marks the outer record complete. Preserve a watermark preceding the current range and add an actual TN replay test with EntryCount greater than one and SafeDSN zero.

### P1 - Let replicated completion override stale heartbeat status (pkg/logservice/wal_recovery_status.go:81)
The coordinator publishes replicated HAKeeper completion before clearing its local heartbeat flag. If it crashes in that window, LogState retains coordinator-pending indefinitely; expired LogStores are not removed from this map. This loop therefore keeps checkBootstrap from entering Running even though LogServiceRecoveryCompleted is durable and a Log quorum exists. The same problem applies to a non-coordinator lost before its completion heartbeat. Replicated completion must override stale per-store statuses, and the crash window needs coverage.

### P1 - Validate embedded TN DSNs against the enclosing range (pkg/logservice/wal_recovery.go:631)
validateWALTNEntry skips the first eight bytes containing the serialized driver DSN. An artifact whose outer LogEntry declares start=10/count=2 but embeds DSNs 10 and 999 passes validation, is appended, and can mark recovery complete. TN replay later treats the embedded values as authoritative for WAL/checkpoint truncation bookkeeping, so this can corrupt mappings and truncate unrelated recovered or subsequent records. Require each embedded DSN to equal startDSN+i before replay.

jiangxinmeng1 (changes requested):
pkg/logservice/config.go:407

  I think this validation should also cover restore.enabled itself, not only the wal-
  data-path case. NewService marks any node with restore.enabled as walRecovery.pending
  and publishes that in LogStore heartbeats, and checkBootstrap keeps HAKeeper out of
  Running while walRecoveryPending(state) is true. So a non-initial LogService, or a
  misconfigured initial member with restore.enabled=true but no wal-data-path, can pass
  local validation and then block bootstrap indefinitely. Can we fail fast when
  restore.enabled is set on a node that is not one of the initial HAKeeper members, and
  add a test for that configuration?

gouhongshen (changes requested):
## Codex automated review

Reviewed head a9450016ed3a35d4f9888560f353f7b1c55d388a against origin/main after loading all existing reviews, issue comments, and inline comments. Three blocking data-integrity paths remain. Focused tests could not compile because this worktree lacks the required cgo/libmo.dylib prerequisite.

### P1 - Do not infer SafeDSN from the first surviving record (pkg/logservice/wal_recovery.go:793)
This clamp assumes `firstNormalDSN-1` is already durable and raises any smaller source SafeDSN. That is invalid because asynchronous commits allow PSN order to differ from DSN order: a surviving WAL can begin with DSN 10/SafeDSN 3 while DSNs 4-9 occur at later PSNs. Rewriting the first payload to SafeDSN 9 makes TN replay treat those delayed DSNs as already safe, potentially skipping them or hitting replay logic errors. Preserve the lower source watermark or derive the lower bound from authoritative checkpoint metadata, with an end-to-end test covering an out-of-order prefix across read batches.

### P1 - Gate persisted TN startup, not only bootstrap commands (pkg/logservice/store_hakeeper_check.go:510)
Replacing `state.TNState` with an empty state only suppresses new HAKeeper StartReplica commands. TN `Start()` independently reloads persisted shard metadata and calls `startTNShards()` before heartbeats; its normal ReplayForWrite stops at the current Log tail and cannot be rerun. A retained or restarting TN can therefore replay the newly created empty Log shard before WAL import, after which HAKeeper completes recovery and enters Running with the TN missing the recovered tail. The replicated barrier must be observed by TN startup/replay itself, and the integration test must retain TN metadata rather than using an empty heartbeat.

### P1 - Fence truncation throughout WAL replay (pkg/logservice/rsm.go:250)
The recovery lease rejects ordinary appends and lease changes, but `handleTruncateLsn` still accepts truncation commands. The coordinator checks TruncatedLsn only once after installing the fence, so an existing TN can truncate after that check while entries are being imported. Reconciliation ignores non-user records and the completion marker only counts imported user entries, allowing recovery to report success with a raised truncation watermark that makes restored records unreadable or eligible for compaction. Reject TruncateLSNUpdate while the recovery lease is active and test a truncate interleaved with replay.

LeftHandCold:
> ## Codex automated review
> Reviewed head [a945001](https://github.com/matrixorigin/matrixone/commit/a9450016ed3a35d4f9888560f353f7b1c55d388a) against origin/main after loading all existing reviews, issue comments, and inline comments. Three blocking data-integrity paths remain. Focused tests could not compile because this worktree lacks the required cgo/libmo.dylib prerequisite.
> 
> ### P1 - Do not infer SafeDSN from the first surviving record (pkg/logservice/wal_recovery.go:793)
> This clamp assumes `firstNormalDSN-1` is already durable and raises any smaller source SafeDSN. That is invalid because asynchronous commits allow PSN order to differ from DSN order: a surviving WAL can begin with DSN 10/SafeDSN 3 while DSNs 4-9 occur at later PSNs. Rewriting the first payload to SafeDSN 9 makes TN replay treat those delayed DSNs as already safe, potentially skipping them or hitting replay logic errors. Preserve the lower source watermark or derive the lower bound from authoritative checkpoint metadata, with an end-to-end test covering an out-of-order prefix across read batches.
> 
> ### P1 - Gate persisted TN startup, not only bootstrap commands (pkg/logservice/store_hakeeper_check.go:510)
> Replacing `state.TNState` with an empty state only suppresses new HAKeeper StartReplica commands. TN `Start()` independently reloads persisted shard metadata and calls `startTNShards()` before heartbeats; its normal ReplayForWrite stops at the current Log tail and cannot be rerun. A retained or restarting TN can therefore replay the newly created empty Log shard before WAL import, after which HAKeeper completes recovery and enters Running with the TN missing the recovered tail. The replicated barrier must be observed by TN startup/replay itself, and the integration test must retain TN metadata rather than using an empty heartbeat.
> 
> ### P1 - Fence truncation throughout WAL replay (pkg/logservice/rsm.go:250)
> The recovery lease rejects ordinary appends and lease changes, but `handleTruncateLsn` still accepts truncation commands. The coordinator checks TruncatedLsn only once after installing the fence, so an existing TN can truncate after that check while entries are being imported. Reconciliation ignores non-user records and the completion marker only counts imported user entries, allowing recovery to report success with a raised truncation watermark that makes restored records unreadable or eligible for compaction. Reject TruncateLSNUpdate while the recovery lease is active and test a truncate interleaved with replay.

TN 持久化启动：不采纳。 离线恢复明确要求 TN 在 WAL 恢复完成前保持停止。修改 TN 通用启动流程会增加正常集群启动风险。
恢复期间 truncate：不采纳。 严格离线流程下没有 TN 或外部客户端发送 truncate，LogService truncation worker 自身不会生成该命令。

jiangxinmeng1 (changes requested):
pkg/logservice/wal_recovery_status.go:113
  
This does not actually wait for every voting replica to report its recovery status. A
  store may already exist in LogState.Stores from an earlier heartbeat while ConfigData
  is nil or does not contain wal-recovery-status; continuing here allows election from a
  partial candidate set.

  For example, if store-b reports coordinator-pending first, it can elect itself while a
  lexicographically smaller artifact-owning store-a has not published its recovery status
  yet. Once store-a reports, it will also elect itself. The destination fence does not
  serialize these coordinators when they use the same WAL digest and recovery lease
  holder, so both can append the same source entry and corrupt the recovery sequence.

  Please return not-ready unless every voting replica has a valid recovery status item,
  and add a test where one replica has no status while another replica is coordinator-
  pending.

gouhongshen (changes requested):
## Codex automated review

Found three blocking correctness issues. Focused tests could not build because required C headers are unavailable.

### P1 - Existing backup-only restore is rejected without the new recovery flag (pkg/logservice/service_bootstrap.go:178)
With the existing `restore.file-path` configuration and `restore.enabled=false`, the initial request carries no recovery intent. If another initial member wins with default IDs, this unconditional watermark restore sends `RestoreIDWatermarkUpdate`, which HAKeeper rejects whenever `LogServiceRecoveryPending` is false. Bootstrap then retries and fails, regressing the pre-existing backup restore path.

### P1 - Malformed skip records can panic during replay (pkg/logservice/wal_recovery.go:639)
Skip payloads are accepted whenever they are nonempty and a multiple of 16 bytes; DSN/PSN semantics are not checked. The downstream replayer panics when a skip DSN is absent from `dsn2PSNMap` or appears twice. A structurally valid but corrupt recovery artifact therefore crashes replay instead of failing closed.

### P1 - Overlapping DSN ranges are not rejected (pkg/logservice/wal_recovery.go:546)
Cross-entry validation checks Raft indexes and arithmetic overflow but not overlap or duplicate normal DSN intervals. Two normal entries with distinct Raft indexes and the same DSN start pass validation; replay stores only one PSN per start DSN and overwrites the earlier record, allowing recovered WAL data to be silently omitted or applied incorrectly.

jiangxinmeng1 (approved):
Re-reviewed the latest update. The coordinator election now correctly waits for every
  voting replica to publish a valid recovery status, which addresses the partial-election
  race. I also reviewed the new WAL sequence and skip DSN/PSN validation and did not find
  any further blocking issues. The targeted recovery tests pass.

gouhongshen (approved):
## Codex automated review

Loaded all existing reviews, issue comments, and inline comments (none). Earlier blocking recovery findings are addressed; the retained-TN and truncate concerns are conditional on the explicit offline-recovery precondition. `git diff --check` passes. Focused Go tests could not run because the Go toolchain is unavailable.

### P2 - Bound retained SkipDSN references during WAL scanning (pkg/logservice/wal_recovery.go:480)
Although recovery accepts files up to 10 GiB and streams normal payloads, every SkipDSN payload is decoded into a reference slice and retained in `entry.skip` until the entire file has been parsed, sorted, and validated. Valid skip-heavy artifacts can therefore retain nearly the full 10 GiB as Go allocations (plus slice overhead), causing the LogService to OOM rather than fail cleanly. Cap cumulative references/memory, or retain offsets and re-read the skip payload only after sorting.

### P2 - Preserve missing custom backup behavior outside WAL recovery (pkg/logservice/service_bootstrap.go:343)
Before this change, `getBackupData` treated a missing restore file as no backup for every path. This branch now fails normal bootstrap for any missing custom path, even with `restore.enabled=false` and no WAL recovery configured; only the default `hk_data` path remains optional. That regresses existing backup-only/normal configurations and contradicts the stated unchanged normal startup behavior. Keep the coordinator-side missing-artifact requirement in `getBackupDataForBootstrap`, but preserve the prior optional behavior outside recovery.

volgariver6 (approved):

XuPeng-SH (changes requested):
Reviewed exact head c3a6a415676d065aa0dc718e21145e1b8ced55ad against current origin/main. The earlier recovery-integrity blockers are addressed, and fresh package plus race-focused tests pass, but two material gaps remain.

### P1 - Preserve optional missing custom backup behavior outside WAL recovery (pkg/logservice/service_bootstrap.go:314-345)

When wal-data-path is absent and restore.enabled is false, getBackupDataForBootstrap does not take the non-coordinator exception. A missing non-default relative file now returns ErrFileNotFound here, and a missing absolute file returns os.ErrNotExist above. Before this PR, every missing restore path returned nil. An existing deployment can retain a custom restore.file-path after the one-time artifact is removed or temporarily unavailable; upgrading to this change then aborts BootstrapHAKeeper even though WAL recovery is not configured. This contradicts the PR contract that normal startup is unchanged without wal-data-path. The new TestGetBackupDataForBootstrap currently codifies the regression. Keep strict missing-artifact failure for the actual WAL recovery coordinator, but preserve the previous optional behavior outside recovery.

### P2 - Bound retained SkipDSN references before semantic validation (pkg/logservice/wal_recovery.go:475-480)

The scanner permits a 10 GiB file and streams normal payloads, but each SkipDSN body is copied into a walRecoverySkipReference slice and retained on WALEntry until the entire file has been scanned and sorted. Unknown, duplicate, and wrong-PSN references are rejected only afterward by validateWALRecoveryEntries. Consequently, a structurally valid but semantically malformed skip-heavy artifact can make the process retain roughly the artifact size as Go heap and OOM before returning the promised validation error. Normal mo_br output is constrained by the number of live WAL entries, so this finding is specifically about the PR stated fail-closed behavior for malformed recovery input. Retain offsets and re-read skip bodies after sorting, validate incrementally where possible, or impose a safe cumulative reference bound.

XuPeng-SH (approved):
Re-reviewed exact head fafe6f879267643520776da5a2682586a06f7d5e against base dae28bed7bbe00de18754bf9ab3b9f255b3c604f. No remaining blocking issue found.

I revalidated every previously raised recovery gap. The current implementation now preserves optional missing custom backups outside WAL recovery, bounds retained SkipDSN references before allocation, streams payloads, derives the actual RPC payload ceiling, binds completion to the destination RSM marker, rejects dirty destinations, persists/reconciles crash progress, fences normal appends and lease changes, preserves SafeDSN semantics, waits for complete voting membership before coordinator election, and holds HAKeeper scheduling/ID allocation behind replicated recovery preparation with allocator-generation invalidation.

- `go build -mod=readonly ./pkg/hakeeper ./pkg/logservice`
- `go vet -mod=readonly ./pkg/hakeeper ./pkg/logservice`
- full `go test -count=1 -timeout=15m ./pkg/hakeeper ./pkg/logservice`
- focused `go test -race -count=1 -timeout=15m -run 'WAL|Recovery|Restore|Recovered|IDAllocatorDiscards' ./pkg/hakeeper ./pkg/logservice`
- `git diff --check`

All passed. The full package run also covers the real TAE LogEntry writer/decoder replay test, not only synthetic byte validation.

The current Ubuntu UT rerun reports `pkg/logservice` PASS (387.81s). Its sole failure is a data race in `pkg/vm/engine/tae/rpc/handle_debug.go` triggered by `pkg/tests/dml/TestDataBranchPickRetestCompositePKChunkProbe`; neither path is modified by this PR.

Issue #25868, opened 2026-07-19
[Bug]: caller timeout resets the shared HAKeeper client and cascades connection failures


LeftHandCold:
## Summary

A caller-scoped `context.Canceled` or `context.DeadlineExceeded` returned by a managed HAKeeper request currently closes the shared HAKeeper client.

The client is shared by connection-ID allocation, CN heartbeat, cluster refresh, task service, and other CN control-plane requests. One request reaching its own deadline therefore forces unrelated requests through HAKeeper rediscovery and can turn a single timeout into a connection-admission failure cascade.

## Root cause

Every managed HAKeeper method currently resets the shared client on any request error:

```go
result, err := c.getClient().sendCNHeartbeat(ctx, hb)
if err != nil {
    c.resetClient()
}
```

`AllocateIDByKeyWithBatch`, used by frontend connection admission, does the same while holding the managed-client mutex.

A caller context error only states that this request can no longer wait. It is not evidence that the shared MORPC transport is unusable. Closing that transport also interrupts concurrent users and makes the next user synchronously discover and reconnect to HAKeeper.

## Impact

This caused the SSB phase in [TKE mergerun job 88168675381](https://github.com/matrixorigin/matrixone/actions/runs/29675783924/job/88168675381) to fail while opening the fifth `q1.3` connection:

- `07:27:36.464 UTC`: the established HAKeeper TCP connection was closed.
- `07:27:36.469 UTC`: one frontend connection failed with `failed to get connection ID from HAKeeper: context deadline exceeded`.
- The next connection was forced into synchronous HAKeeper rediscovery; dials to log-0 and log-1 each reached the one-second connect timeout.
- `07:27:38.884 UTC`: the SSB `q1.3` connection failed with the same connection-ID error.
- In the same cascade, CN heartbeat, task service, and cluster refresh reported HAKeeper deadline errors.
- HAKeeper discovery succeeded again immediately afterward, and the next SSB query succeeded. The LogService pods had no restart, election, OOM, or raft failure in the incident window.

This is a pre-existing availability bug. The failure occurs during frontend routine creation, before SQL execution and before the trace changes in PR #25841.

## Reproduction / evidence

On current `main` at `4d78bc66670418fbdeac2b18b07b2ef5b08b1d72`, a deterministic package-level fault injection makes `sendCNAllocateID` return each caller context error and then inspects the managed client.

Expected: the request returns its context error and the shared client remains available.

Actual on unmodified production code:

```text
[...]

PR #25869, opened 2026-07-19
fix(logservice): isolate HAKeeper client from caller timeout

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25868

Managed HAKeeper operations share one MORPC client. Before this change, every request error reset and closed that shared client, including `context.Canceled` and `context.DeadlineExceeded` from one caller.

In TKE mergerun job 88168675381, a connection-ID allocation reached its frontend deadline and closed the established HAKeeper connection. Unrelated connection admission, CN heartbeat, task service, and cluster refresh requests were then forced through synchronous HAKeeper rediscovery. The SSB `q1.3` connection failed during that cascade even though HAKeeper recovered immediately and the next query succeeded.

A caller context error means only that request can no longer wait; it does not establish that the shared transport is broken. MORPC's backend read loop independently detects EOF, connection errors, and read timeouts and reconnects the transport.

This change:

- returns caller cancellation/deadline errors unchanged without closing the shared HAKeeper client;
- applies that rule consistently to all managed HAKeeper operations;
- preserves existing reset/reconnect/retry behavior for EOF, connection-close, `NoHAKeeper`, backend timeout, and every other non-context error;
- completes MORPC cleanup when `Send` fails before queue admission, so the Future is removed from the backend map instead of retaining its request indefinitely;
- recycles payload-bearing logservice responses that arrive after their Future has already been removed;
- does not change connection IDs, allocation batches, public APIs, configuration, timeouts, retry counts, or normal success paths.

The regression coverage proves:

- caller cancellation/deadline keeps the exact managed HAKeeper client reusable;
- EOF send errors still reset, rebuild, and retry the client;
- an admission-stage context error leaves no Future in MORPC's backend map (the test fails on unmodified code with one retained Future);
- the orphan-response callback releases the `RPCResponse` payload back through its existing pool lifecycle.

## Review follow-up and safety boundary

The requested review at `pullrequestreview-4730633356` identified two valid lifecycle gaps. Commit `ab3fdb94d4` addresses both with focused cleanup changes:

1. A Future is registered before `doSend`. If enqueue fails, no write-loop owner exists, so the synchronous send-error path now performs the single matching send-completion/unref before closing it.
2. A late payload response whose Future/Stream is already absent now uses `WithBackendFreeOrphansResponse` and the existing `RPCResponse.Release` implementation.

Normal queued sends are unchanged: only the write loop calls send completion. Normal responses are unchanged: the caller still owns `RPCResponse.Release`. The orphan callback runs only when MORPC finds neither a Future nor a Stream, so it does not overlap normal response ownership.

## Testing

Tested with MatrixOne macOS CGo header/link settings:

- `go test ./pkg/common/morpc -run '^TestSendContextErrorReleasesFuture$' -count=20`
- `go test -race ./pkg/common/morpc -run '^TestSendContextErrorReleasesFuture$' -count=20`
- focused HAKeeper/orphan-response tests: normal `count=10`, race `count=10`
- `go test ./pkg/common/morpc -count=1`
- `go test -race ./pkg/common/morpc -count=1`
- `go test ./pkg/logservice -count=1`
- `go test -race ./pkg/logservice -count=1`
- `go test ./pkg/frontend ./pkg/cnservice ./pkg/proxy -run '^$' -count=1`
- `go build ./pkg/common/morpc ./pkg/logservice ./pkg/frontend ./pkg/cnservice ./pkg/proxy`
- `go vet ./pkg/common/morpc ./pkg/logservice`
- `git diff --check`

### DIFF
--- a/pkg/common/morpc/backend.go
+++ b/pkg/common/morpc/backend.go
@@ -332,6 +332,7 @@ func (rb *remoteBackend) SendInternal(ctx context.Context, request Message) (*Fu
 func (rb *remoteBackend) send(ctx context.Context, request Message, internal bool) (*Future, error) {
 	f := rb.getFuture(ctx, request, internal)
 	if err := rb.doSend(f); err != nil {
+		f.messageSent(err)
 		f.Close()
 		return nil, err
 	}
--- a/pkg/common/morpc/backend_test.go
+++ b/pkg/common/morpc/backend_test.go
@@ -68,6 +68,36 @@ func TestSend(t *testing.T) {
 	)
 }
 
+func TestSendContextErrorReleasesFuture(t *testing.T) {
+	rb := &remoteBackend{
+		codec:      newTestCodec(),
+		metrics:    newMetrics(""),
+		waitWriteC: make(chan struct{}, 1),
+		writeC:     make(chan *Future, 1),
+	}
+	rb.stateMu.state = stateRunning
+	rb.mu.futures = make(map[uint64]*Future)
+	rb.pool.futures = &sync.Pool{
+		New: func() any {
+			return newFuture(rb.releaseFuture)
+		},
+	}
+	// Keep the write queue full so Send exits through the caller-context path
+	// before the future is handed to the write loop.
+	rb.writeC <- nil
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	cancel()
+	future, err := rb.Send(ctx, newTestMessage(1))
+	require.ErrorIs(t, err, context.Canceled)
+	require.Nil(t, future)
+
+	rb.mu.RLock()
+	defer rb.mu.RUnlock()
+	require.Empty(t, rb.mu.futures,
+		"a future that was never enqueued must be removed on Send failure")
+}
+
 func TestReadTimeoutWithNormalMessageMissed(t *testing.T) {
 	testBackendSend(t,
 		func(conn goetty.IOSession, msg interface{}, _ uint64) error {
--- a/pkg/logservice/client.go
+++ b/pkg/logservice/client.go
@@ -926,6 +926,7 @@ func getRPCClient(
 	backendOpts := []morpc.BackendOption{
 		morpc.WithBackendConnectTimeout(time.Second),
 		morpc.WithBackendHasPayloadResponse(),
+		morpc.WithBackendFreeOrphansResponse(releaseOrphanRPCResponse),
 		morpc.WithBackendLogger(logutil.GetGlobalLogger().Named("hakeeper-client-backend")),
 		morpc.WithBackendReadTimeout(readTimeout),
 	}
@@ -954,4 +955,8 @@ func getRPCClient(
 	codec := morpc.NewMessageCodec(sid, mf, codecOpts...)
 	bf := morpc.NewGoettyBasedBackendFactory(codec, backendOpts...)
 	return morpc.NewClient("logservice-client", bf, clientOpts...)
+}
+
+func releaseOrphanRPCResponse(message morpc.Message) {
+	message.(*RPCResponse).Release()
 }
--- a/pkg/logservice/client_test.go
+++ b/pkg/logservice/client_test.go
@@ -37,6 +37,17 @@ import (
 
 var testLogger = logutil.GetGlobalLogger().Named("logservice-test")
 
+func TestReleaseOrphanRPCResponse(t *testing.T) {
+	pool := &sync.Pool{}
+	response := &RPCResponse{
+		payload: []byte("payload"),
+		pool:    pool,
+	}
+
+	releaseOrphanRPCResponse(response)
+	require.Nil(t, response.payload)
+}
+
 func TestClientConfigIsValidated(t *testing.T) {
 	cfg := ClientConfig{}
 	cc, err := NewClient(context.TODO(), "", cfg)
--- a/pkg/logservice/hakeeper_client.go
+++ b/pkg/logservice/hakeeper_client.go
@@ -378,7 +378,7 @@ func (c *managedHAKeeperClient) GetClusterDetails(ctx context.Context) (pb.Clust
 			return pb.ClusterDetails{}, err
 		}
 		cd, err := c.getClient().getClusterDetails(ctx)
-		if err != nil {
+		if shouldResetHAKeeperClient(err) {
 			c.resetClient()
 		}
 		if c.isRetryableError(err) {
@@ -406,7 +406,7 @@ func (c *managedHAKeeperClient) GetClusterState(ctx context.Context) (pb.Checker
 			return pb.CheckerState{}, err
 		}
 		s, err := c.getClient().getClusterState(ctx)
-		if err != nil {
+		if shouldResetHAKeeperClient(err) {
 			c.resetClient()
 		}
 		if c.isRetryableError(err) {
@@ -451,7 +451,9 @@ func (c *managedHAKeeperClient) AllocateID(ctx context.Context) (uint64, error) 
 		firstID, err := sendCNAllocateIDFunc(c.mu.client, ctx, "", batchSize)
 
 		if err != nil {
-			c.resetClientLocked()
+			if shouldResetHAKeeperClient(err) {
+				c.resetClientLocked()
+			}
 			if c.isRetryableError(err) {
 				if err := c.waitRetryLocked(ctx); err != nil {
 					return 0, err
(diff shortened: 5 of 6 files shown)


LeftHandCold:
Completed a first-principles pre-push review of the complete PR diff at `d55ab8706c45b7bab96a5f335a8cae60d7670d4a`.

### Scope and semantic boundary

- Diff is limited to `pkg/logservice/hakeeper_client.go` and its package test.
- No frontend, trace, StatementInfo, timeout, retry-count, allocation-batch, ID-value, API, or configuration behavior changes.
- The original caller still receives `context.Canceled` / `context.DeadlineExceeded`. This PR isolates that caller-scoped failure; it does not claim the original caller can never time out.

### Error-path review

- `AllocateID` and `AllocateIDByKeyWithBatch` keep the outer `err != nil` branch, so context errors return before any ID-cache mutation.
- EOF, UnexpectedEOF, expected connection-close, `NoHAKeeper`, backend timeout, and all other non-context errors still reset the shared client.
- Existing EOF tests prove reset, reconstruction, and retry still occur.
- The new test proves both context error classes preserve the exact client instance and a following allocation uses it successfully.

### Concurrency and lifecycle review

- No goroutine, channel, timer, lock, state field, generation, or wait edge is added.
- Avoiding `close()` on a caller deadline prevents one request from invalidating concurrent heartbeat, cluster, task, and connection-ID users.
- MORPC closes/removes the timed-out request future independently. Its backend read loop still detects transport EOF/read errors and the configured read timeout and schedules reconnect, so a genuinely failed transport is not made permanent by this change.
- Managed-client ownership and shutdown ownership are unchanged.

### Validation

- Focused normal tests: count=10, PASS.
- Focused race tests: count=20, PASS.
- Full `./pkg/logservice`: normal and race, PASS.
- frontend/CN/proxy test compilation, PASS.
- logservice/frontend/CN/proxy build, PASS.
- `go vet ./pkg/logservice`, PASS.
- `git diff --check`, PASS.

Remote verification:
- PR is ready, targets `main`, and is mergeable.
- Remote head equals the locally tested commit.
- Only the two intended logservice files are present.

No blocking correctness, leak, hang, ownership, retry-regression, cache-corruption, or unintended cross-module semantic issue found. Full GitHub CI is still in progress.

volgariver6 (approved):

gouhongshen (changes requested):
## Codex automated review

The timeout-isolation change is incomplete for MORPC send-stage context failures and also introduces response-pool churn for late replies.

### P1 - Clean up futures when Send returns a caller context error (pkg/logservice/hakeeper_client.go:920)
This predicate skips client reset for errors from both `c.client.Send` and `future.Get`. In MORPC, `Send` creates and inserts a Future before attempting to enqueue it. If the write queue is full and the caller context expires, `doSend` returns the context error before enqueueing; `remoteBackend.send` only calls `f.Close`, without `messageSent`/`unRef`. Because `Future.Close` releases only after its reference reaches zero, the Future remains in the backend's futures map and retains the request. Keeping the shared client alive means repeated admission-time cancellations can grow this map and memory without bound. The new test stubs the send function and does not exercise this path. The reset exemption must distinguish send-stage failures, or MORPC must release the Future on this path.

### P2 - Recycle orphan responses when retaining timed-out clients (pkg/logservice/hakeeper_client.go:920)
Once the client is retained after a caller timeout, late HAKeeper replies can reach MORPC's orphan-response path. `getRPCClient` enables payload responses but does not configure `WithBackendFreeOrphansResponse`; MORPC therefore does not call `RPCResponse.Release()` for these late replies. Under repeated timeouts this bypasses the response pool and causes avoidable allocations and GC pressure. Configure the orphan-response release callback when constructing this client.

LeftHandCold:
Addressed both findings from review `4730633356` in `ab3fdb94d459fe33d5f8fbe024d85d3ef575b9b8`.

### P1 — pre-enqueue Future retention

Confirmed. `getFuture` inserts the Future and takes one reference before `doSend`. When queue admission returned the caller context error, the Future had never reached the write loop, so the old `Close` could not drop that reference. The synchronous failure path now calls `messageSent(err)` before `Close`, matching the ownership transition that the write loop performs only for successfully queued Futures.

Added a deterministic full-queue/canceled-context regression test. It fails before the fix with one Future left in `rb.mu.futures`; it passes after the fix. Normal and race runs both passed with `count=20`, and full MORPC normal/race suites passed.

### P2 — orphan payload-response recycling

Confirmed. `getRPCClient` now configures `WithBackendFreeOrphansResponse`; the callback uses the existing `RPCResponse.Release`, which clears the protobuf/payload and returns the object to its originating pool. It is invoked only in MORPC's no-Future/no-Stream orphan branch, so normal response ownership remains with the caller and there is no double-release path.

Added focused callback coverage. Full `pkg/logservice` normal and race suites passed.

### Final safety review

- no API, configuration, timeout, retry-count, ID-allocation, or success-path changes;
- non-context HAKeeper errors still reset/reconnect exactly as before;
- no new goroutine, lock, channel, timer, state field, or wait dependency;
- frontend/CN/proxy compile checks, target-package build, vet, and `git diff --check` all passed.

The PR description has also been updated with the two lifecycle gaps, ownership boundaries, and complete validation record.

XuPeng-SH (approved):
Reviewed the current head `ab3fdb94d4` from the transport, lifecycle, concurrency, failure-classification, and performance angles. No blocking issue found.

The previous findings are closed:

- A Future registered before queue admission now gets the matching send completion/unref only on synchronous `doSend` failure. Successfully queued Futures remain owned by the write loop, and the existing CAS prevents duplicate send completion.
- Late payload responses with neither a Future nor a Stream now follow the existing `RPCResponse.Release` pool lifecycle. Normal response ownership is unchanged.
- Caller `context.Canceled` and `context.DeadlineExceeded` no longer invalidate the shared managed HAKeeper client, while EOF, UnexpectedEOF, connection-close, NoHAKeeper, backend timeout, and other non-context errors preserve reset/reconnect behavior.

Focused regression tests passed repeatedly in normal and race modes; full MORPC normal/race passed; both modified packages build and vet cleanly; `git diff --check` is clean; GitHub CI is green. The change adds no success-path allocation or wait and only adds constant-time error-path cleanup/classification.

gouhongshen (commented):
## Codex automated review

The timeout isolation and synchronous Future cleanup are sound, but payload-response recycling misses a timeout race.

### P2 - Release responses discarded while the Future still exists (pkg/logservice/client.go:929)
This callback only handles the `requestDone` branch where no Future/Stream is found. If a response arrives after the caller context expires but before `future.Close`, `requestDone` still finds the Future and `Future.done` drops the response without invoking `freeResponse`. Conversely, if the response was queued just before cancellation, `Future.Get` may select `ctx.Done` and `Future.Close` drains it without calling `RPCResponse.Release`. Retaining the shared client makes these late-response races reachable, so repeated timeouts bypass `respPool` and increase allocation/GC pressure.

LeftHandCold:
Evaluated review `4731897565` against the production HAKeeper call path. The two timing windows described by the review are technically reachable, but I am intentionally not changing this PR for the following reasons.

### Actual impact

- In both windows, the response is removed from all live references after `requestDone` returns or `Future.clear` drains the channel. It is therefore reclaimed by Go GC.
- The Future itself is still removed and recycled; there is no backend-map growth.
- The payload read-loop callback is still invoked (`Future.done` on an expired context, or `Future.Close` for a queued response), so the MORPC read loop does not hang.
- The caller still receives the expected context error. There is no incorrect ID, retry, connection, or response result.

The only effect is that this particular `RPCResponse` is not returned to `sync.Pool`, causing a pool miss and a later allocation. `sync.Pool` is explicitly a best-effort allocation optimization and can be cleared by the runtime at any GC cycle; missing one return is not a resource leak.

### Production reachability and scale

The HAKeeper response must complete in the narrow interval around the caller deadline: either after context expiry but before the deferred `Future.Close`, or be queued at the same time that `Future.Get` selects `ctx.Done`. Responses arriving after `Future.Close` take the normal no-Future orphan branch and are already released by the callback added in `ab3fdb94d4`.

For the original connection-ID path, the RPC deadline is two seconds and IDs are fetched in batches (default 100), so this boundary race is not on every connection. Repeated boundary-aligned timeouts could add allocation/GC pressure, but do not create retained growth or a correctness/stability failure.

### Why no change in this PR

A complete fix is not local to the logservice callback. It requires changing generic MORPC `Future` ownership for responses rejected while the Future still exists and responses drained from its channel, or changing response-vs-context selection semantics. That affects all MORPC clients and introduces double-release/reuse risks that are disproportionate to this narrow pool-efficiency gap.

Decision: treat this as optional MORPC allocation hardening, not a blocker for the HAKeeper timeout-isolation correctness fix. Revisit separately only if profiling shows material orphan pool misses or GC pressure under timeout storms.

gouhongshen (approved):

PR #25878, opened 2026-07-20
fix: preserve extended TIME function semantics

### LINKED ISSUE #24784: [Bug]: time functions mishandle large hours and fractional seconds

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24784

Continues and completes the work from #25801 while preserving the original commits and authorship.

Changes:

- return complete extended TIME hours from `HOUR`;
- preserve extended hours and negative signs in `TIME_FORMAT`, and return NULL for an empty format;
- preserve `MAKETIME` fractional precision and negative/extended-hour behavior;
- use MySQL-compatible half-away-from-zero conversion for floating hour/minute arguments;
- keep VARCHAR/DECIMAL seconds exact through the microsecond rounding boundary instead of routing them through `FLOAT64`;
- bound exact-second parsing input/exponents to prevent unbounded big-integer allocation;
- cover integer widths, NULL/invalid inputs, endpoint clamping, exact rounding boundaries, and planner result scales.

Validation:

- `go test -count=1 -run 'Test(TimeFormat|MakeTime|MakeTimeReturnScale)' ./pkg/sql/plan/function`
- `go test -race -count=10 -run 'Test(TimeFormat|MakeTime)' ./pkg/sql/plan/function`
- `go test -count=1 ./pkg/sql/plan/function`
- `git diff --check`

Distributed SQL expectations were updated for the exact VARCHAR/DECIMAL boundary, `.5` hour/minute rounding, and positive/negative empty formats; CI is the execution source for those BVT cases.


gouhongshen (changes requested):
## Codex automated review

Binary MAKETIME arguments lose their binary numeric semantics during the new VARCHAR overload selection.

### P1 - Preserve IsBin when routing binary MAKETIME arguments through VARCHAR (pkg/sql/plan/function/list_builtIn.go:10800)
`T_binary`/`T_varbinary`/`T_blob` satisfy `IsMySQLString()`, so this checker forces them to a `VARCHAR` target (and does the same for exact second at line 10806). That inserts a cast; function-expression result setup unconditionally clears `Vector.IsBin` (`pkg/sql/colexec/evalExpression.go:550-557`). Consequently the binary branches in `makeTimeStringIntegerGetter`/`makeTimeStringSecondGetter` are never reached and raw bytes are parsed as text instead. For example, MySQL 8 returns `258:00:00` for `MAKETIME(X'0102',0,0)` and `12:00:01` for `MAKETIME(12,0,X'01')`; after this cast MatrixOne parses the first as zero and rejects the second as NULL. Before this change the normal numeric cast retained `IsBin` and used the binary conversion path. Add binary argument coverage and preserve the flag or avoid the VARCHAR cast for binary inputs.

aunjgr (approved):
Reviewed current head. The replacement correctly uses half-away-from-zero rounding for MAKETIME hour/minute inputs, preserves exact decimal and string seconds, and covers extended TIME signs and ranges. I found no remaining code blocker; merge after current CI is green.

aunjgr (approved):
Reviewed current head. The replacement correctly uses half-away-from-zero rounding for MAKETIME hour/minute inputs, preserves exact decimal and string seconds, and covers extended TIME signs and ranges. I found no remaining code blocker; merge after current CI is green.

VioletQwQ-0:
Addressed the binary MAKETIME regression in 3f329d10f.

Changes:
- binary/varbinary/blob arguments now stay on the numeric cast path, preserving hex-literal byte semantics instead of being routed through VARCHAR;
- added overload-selection coverage for all three binary types in every MAKETIME argument position;
- added BVT coverage for binary hour and second literals.

Validation:
- go test -count=1 ./pkg/sql/plan/function
- go vet ./pkg/sql/plan/function
- go build ./pkg/sql/plan/function
- focused race tests passed 10 consecutive runs
- git diff --check

VioletQwQ-0:
Fixed the remaining binary-second MAKETIME regression in bb599c6a8.

Root cause:
- exact VARCHAR seconds use target scale -1 to derive TIME(6);
- for X/B literals, that metadata-only VARCHAR-to-VARCHAR implicit cast cleared Literal.IsBin during function evaluation;
- MAKETIME then parsed the raw byte as ordinary text and returned NULL.

Changes:
- skip only the MAKETIME arg3 identity VARCHAR cast when width/type are unchanged, preserving IsBin;
- keep hex/bit seconds as TIME(0) metadata because they are integral numeric values;
- add parser -> binder -> executor regression coverage for hex literals in all three positions, bit seconds, binary strings, plain strings, decimal seconds, and large hex hours.

Validation:
- reproduced the previous CI failure with the exact SELECT CAST(MAKETIME(12, 0, X'01') AS VARCHAR) path before the fix;
- go test -count=1 ./pkg/sql/plan ./pkg/sql/plan/function
- go vet ./pkg/sql/plan ./pkg/sql/plan/function
- go test -race -count=1 ./pkg/sql/plan -run '^TestMakeTimeBinaryLiteralBindAndExecute$'
- git diff --check

The PR remains Ready for review.

VioletQwQ-0:
Fixed the binary `MAKETIME` hour overflow paths in `50073bfb3`.

Changes:
- decode binary hour/minute values as unsigned big-endian integers before any signed conversion;
- ignore arbitrary leading zero bytes, preserving numeric equivalence for width-extended encodings;
- saturate values above `math.MaxInt64`, including 8-byte `X'FFFFFFFFFFFFFFFF'` and non-zero encodings wider than 8 bytes, so the existing `MAKETIME` endpoint clamp returns `838:59:59` instead of wrapping negative or becoming zero;
- keep overflowing binary minute values invalid rather than silently treating them as minute zero.

Coverage:
- parser -> binder -> expression executor regressions for empty/zero, `MaxInt64`, `MaxInt64+1`, `MaxUint64`, >8-byte overflow, and wide leading-zero hour/minute encodings;
- existing hex/bit/binary-string/ordinary-string/decimal adjacent cases remain covered;
- direct decoder boundary table for empty, zero, both signed-conversion boundaries, 8-byte maximum, true width overflow, and leading-zero width extension.

Validation:
- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- `go test -race ./pkg/sql/plan/function -run '^(TestMakeTimeBinaryIntegerBoundaries|TestMakeTimeUnsignedHourOverflow|TestMakeTimeStringHourMinuteSemantics|TestMakeTimeIntegerSecondRange)$' -count=1`
- `go test -race ./pkg/sql/plan -run '^TestMakeTimeBinaryLiteralBindAndExecute$' -count=1`
- `git diff --check`

I also reran the full `mo-self-review` gate over the PR diff versus current `upstream/main`: all six lenses, the binder-to-executor functional closure, and Q1-Q3 passed with no remaining blocker.

heni02 (approved):

gouhongshen (changes requested):
## Codex automated review

One blocking MySQL-compatibility regression found. Existing reviews/comments were loaded; the earlier binary-MAKETIME finding is fixed at this head and is not repeated. Focused Go tests could not run because the environment lacks the Go toolchain.

### P1 - Do not turn an empty TIME_FORMAT format into NULL (pkg/sql/plan/function/func_binary.go:4027)
This newly treats `len(fmt) == 0` as a NULL condition, so `TIME_FORMAT(CAST('15:30:45' AS TIME), '')` now returns NULL. Before this change, the empty format reached `timeFormat`, appended an empty non-NULL byte slice, and returned `''`. MySQL specifies NULL only when the time or format argument is NULL; an empty string is a valid non-NULL format. The added BVT expectations codify the regression, so compatible callers that distinguish empty strings from NULL will break.

VioletQwQ-0:
Addressed the empty `TIME_FORMAT` compatibility blocker in `57b33f785`.

Changes:
- removed the `len(fmt) == 0` NULL branch, so only an actual NULL time or format produces NULL;
- an empty format now appends a zero-length, non-NULL result;
- guarded the new negative-time prefix for non-empty formats, so `TIME_FORMAT` with an empty format returns `` for both positive and negative TIME values;
- corrected the unit and BVT expectations from NULL to an empty row value while retaining the column headers.

Reproduction:
- before the fix, the focused unit case reported `expected empty string, but get NULL`;
- after removing the NULL branch, the negative case exposed and fixed the adjacent `expected empty string, but get -` path.

Validation:
- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- `go test -race ./pkg/sql/plan/function -run '^(TestTimeFormat|TestMakeTimeBinaryIntegerBoundaries|TestMakeTimeUnsignedHourOverflow)$' -count=1`
- `git diff --check`
- `mo-self-review`: zero open blockers

gouhongshen (changes requested):
## Codex automated review

Two MySQL-compatibility defects remain. Earlier binary-cast and empty-format findings are fixed at this head. Focused Go tests could not compile because this checkout lacks MatrixOne’s required CGo libraries/headers.

### P1 - Preserve zero-extended binary seconds (pkg/sql/plan/function/func_binary.go:7476)
The new `IsBin` second path rejects any byte string longer than eight bytes before considering its value. Thus `MAKETIME(12, 0, X'000000000000000001')` returns NULL even though the nine-byte literal is numerically 1; MySQL 8.4.10 returns `12:00:01` (and likewise accepts zero-extended 59). The new hour decoder already strips leading zeros, but seconds do not. Strip leading zero bytes before the width/range check and add a binder-to-executor regression test.

### P1 - Coerce nonnumeric VARCHAR seconds to zero (pkg/sql/plan/function/func_binary.go:7399)
The new exact-VARCHAR overload sends empty or nonnumeric seconds through `makeTimeExactSecond`, which reports them as NULL (`len == 0` here and `digits == 0` at line 7419). Consequently `MAKETIME(12, 34, '')` and `MAKETIME(12, 34, 'foo')` return NULL. MySQL 8.4.10 returns the non-NULL `12:34:00.000000` with a truncation warning. This changes result nullability for ordinary VARCHAR data; preserve the exact parser for numeric input but use MySQL’s zero coercion for invalid text, with regression coverage.

VioletQwQ-0:
Fixed both remaining MAKETIME compatibility findings in `8f1ca3ac3`.

Changes:
- binary seconds now use the same unsigned big-endian decoder as binary hours/minutes, so arbitrary leading-zero width extension is accepted while genuinely overflowing or out-of-range second values remain NULL;
- exact VARCHAR seconds now coerce empty and nonnumeric text to non-NULL zero, matching MySQL numeric coercion, while SQL NULL, negative/out-of-range numeric values, and the existing bounded-exponent guard retain their prior behavior.

Coverage:
- parser -> binder -> expression executor regressions for zero-extended binary seconds `1` and `59`, true wider overflow, and empty/nonnumeric VARCHAR seconds;
- function-level vector coverage for empty/nonnumeric VARCHAR coercion;
- distributed SQL expectations for the reported forms.

Validation:
- `go test -count=1 ./pkg/sql/plan/...`
- `go test -race -count=1 -run '^(TestMakeTimeExactStringSecondRounding|TestMakeTimeBinaryIntegerBoundaries|TestMakeTimeIntegerSecondRange)$' ./pkg/sql/plan/function`
- `go test -race -count=1 -run '^TestMakeTimeBinaryLiteralBindAndExecute$' ./pkg/sql/plan`
- `go build ./pkg/sql/plan/function ./pkg/sql/plan`
- `git diff --check`

I also reran the complete `mo-self-review` gate over the full PR diff against current `upstream/main`: all six lenses, both binder-to-executor closures, and Q1-Q3 passed with zero remaining blocker. The PR remains Ready for review.

gouhongshen (changes requested):
## Codex automated review

One blocking DECIMAL precision regression remains. Earlier reported binary, empty-format, and VARCHAR coercion defects are fixed at this head. Focused Go tests could not complete because required CGo headers are unavailable.

### P1 - Preserve exact DECIMAL hour/minute values through rounding (pkg/sql/plan/function/func_binary.go:7539)
When the seconds argument is DECIMAL, makeTimeCheck routes hour and minute DECIMAL arguments through FLOAT64. Values just below a half boundary can therefore become exactly `.5` before this new math.Round call. For example, `MAKETIME(12, CAST('59.49999999999999999999' AS DECIMAL(30,20)), CAST('0' AS DECIMAL(2,1)))` converts the minute to float64 `59.5`, rounds it to 60, and returns NULL. MySQL 8.4.10 returns `12:59:00`; the base implementation selected the same float overload but truncated it to 59, so this is a concrete regression. Preserve DECIMAL hour/minute precision until half-away-from-zero conversion and add boundary coverage.

VioletQwQ-0:
Fixed the remaining DECIMAL hour/minute precision issue in `3e1d50ae9`.

What changed:
- DECIMAL hour/minute arguments now stay on an exact `DECIMAL128` overload instead of being converted to `FLOAT64` before rounding.
- Exact DECIMAL rounding is half away from zero and clamps only after the exact integer result is known.
- The overload matrix covers DECIMAL hour/minute combined with float, VARCHAR, DECIMAL, and exact-second inputs without changing VARCHAR truncation semantics.

Coverage added:
- the reported `59.49999999999999999999` minute case;
- DECIMAL64 and DECIMAL128 values below/at the half boundary;
- negative half-hour, minute rounding to 60, and positive/negative `DECIMAL(38,0)` overflow;
- parser -> binder -> expression-executor probes for an 8K-digit exponent and a 4097-digit mantissa, including terminal executor/process cleanup.

I also rechecked every earlier change request at the new head: binary `IsBin`, empty `TIME_FORMAT`, zero-extended binary seconds, and nonnumeric VARCHAR-second coercion remain covered and passing.

Validation:
- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- focused `-race` tests in both packages, 3 runs each
- `go vet` and `go build` for both packages
- `git diff --check`
- full exact-head preflight manifest and `mo-self-review`: PASS, no pending evidence or open blocker

XuPeng-SH (changes requested):
## Codex deep review — changes requested

Reviewed exact head `3e1d50ae90d748ed3d6e771c2e13d294e1536916`. The earlier binary `IsBin`, zero-extended binary-second, nonnumeric VARCHAR-second, and DECIMAL128 boundary findings are fixed. Three reachable MySQL-compatibility blockers remain. Existing package tests and CI pass because these cases are not covered.

### P1 — Keep DECIMAL256 hour/minute exact instead of narrowing to DECIMAL128(38,0) (`pkg/sql/plan/function/list_builtIn.go:10815`)

`exactHour` and `exactMinute` accept every `Oid.IsDecimal()`, including SQL `DECIMAL(65,...)` backed by DECIMAL256, but the checker unconditionally targets DECIMAL128. `SetTargetScaleFromSource` does not preserve a DECIMAL256 source scale for a DECIMAL128 target, so `DECIMAL(65,30)` is bound as `DECIMAL128(38,0)` before the new exact getter runs.

A planner-to-expression-executor regression test reproduces both failures:

- `MAKETIME(CAST("12.499999999999999999999999999999" AS DECIMAL(65,30)),0,0)` returns `13:00:00`; MySQL 8.0.44 and 8.4.10 return `12:00:00`.
- A 65-digit `DECIMAL(65,0)` hour errors while converting to DECIMAL128; both MySQL versions return the saturated `838:59:59`.

Add DECIMAL256 overload/getter support or another bounded exact path, and cover below/at-half plus positive/negative overflow through binder and executor.

### P1 — Do not reject safe exponent underflow or a zero mantissa (`pkg/sql/plan/function/func_binary.go:7471`)

The allocation guard returns NULL solely when the exponent magnitude exceeds 4096. This changes the value semantics of short, safely classifiable inputs: `MAKETIME(12,34,"1e-5000")` and `MAKETIME(12,34,"0e5000")` both become NULL here, while MySQL 8.0.44 and 8.4.10 return `12:34:00.000000` for both. The current extreme test covers only a nonzero huge positive exponent.

Keep the bounded parser, but classify exponent direction and a zero mantissa before rejecting: sufficiently negative exponents underflow to zero, and a zero mantissa remains zero for any exponent. Add direct and binder-to-executor regressions.

### P1 — Empty TIME_FORMAT must remain NULL (`test/distributed/cases/function/func_datetime_time_format.test:43`)

The newly added BVT expectations codify an empty non-NULL string for both signs, and `TimeFormat` appends the empty buffer with `isNull=false`. Direct checks against official MySQL 8.0.44 and 8.4.10 both report `TIME_FORMAT(CAST("15:30:45" AS TIME), "") IS NULL = 1` (same for negative TIME). This also matches the PR description, which says an empty format returns NULL. The earlier review claiming an empty string is contradicted by both official runtime versions.

Restore the NULL result for an empty format and update both unit/BVT expectations.

gouhongshen (changes requested):
## Codex automated review

The latest DECIMAL128 precision fix is valid, but it introduces a blocking DECIMAL256 regression. Earlier review findings remain fixed. Focused tests could not run locally because cgo/libmo.dylib is absent; exact-head CI is green.

### P1 - Preserve DECIMAL256 hour/minute inputs instead of narrowing to DECIMAL128 (pkg/sql/plan/function/list_builtIn.go:10816)
`IsDecimal()` includes DECIMAL256, but the new exact path unconditionally targets DECIMAL128 for hour/minute arguments. Thus a valid value such as `MAKETIME(CAST('999999999999999999999999999999999999999' AS DECIMAL(39,0)),0,0)` is implicitly narrowed to DECIMAL128(38,0); `decimal256ToDecimal128` calls `ParseDecimal128`, which rejects the 39-digit value and aborts evaluation instead of letting MAKETIME clamp it to `838:59:59`. Before this head, DECIMAL256 followed the FLOAT64 overload and reached the existing endpoint clamp. DECIMAL256 is SQL-reachable for precision above 38 (`build_util.go:205-206`), but the new tests cover only DECIMAL64/128. Add an exact DECIMAL256 getter/overloads, or otherwise preserve its range and scale without narrowing.

VioletQwQ-0:
Addressed the latest change requests in `3513b2343`.

@XuPeng-SH:
- DECIMAL256 hour/minute values now stay on DECIMAL256 overloads instead of narrowing through DECIMAL128. The planner/executor tests cover below-half, at-half, and positive/negative overflow, and the overload test covers all 14 mixed DECIMAL256 combinations.
- The exact-second parser now distinguishes value semantics from its allocation guard: large negative exponents underflow to zero, and a zero mantissa stays zero even with a large exponent or a 4097-digit zero representation. Large positive nonzero exponents and over-limit nonzero mantissas remain bounded/rejected.
- Empty `TIME_FORMAT` is NULL again for both positive and negative TIME. I also checked the official MySQL 8.4.6 source independently: `Item_func_date_format::val_str` takes the NULL path when the format length is zero.

@gouhongshen:
- The duplicate DECIMAL256 narrowing finding is fixed by the same concrete DECIMAL256 getter/overload path above; it no longer calls the DECIMAL256-to-DECIMAL128 cast.

Validation:
- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- focused planner/executor MAKETIME and TIME_FORMAT tests
- `git diff --check`
- full exact-head preflight/self-review manifest PASS on `3513b23431994225264db1ec1ec7c0d6983b679c`

XuPeng-SH (changes requested):
## Codex deep review — changes requested

Reviewed exact head `3513b23431994225264db1ec1ec7c0d6983b679c`. The earlier DECIMAL256, exponent-underflow, zero-mantissa, and empty-TIME_FORMAT blockers are fixed. Full package tests, focused race tests, `go vet`, formatting, and current CI all pass. One reachable MySQL-compatibility blocker remains.

### P1 — Apply the exact-second cap to significant digits, not raw mantissa width (`pkg/sql/plan/function/func_binary.go:7478`)

The allocation guard counts every mantissa digit before removing insignificant zeros. A VARCHAR second consisting of 4096 leading zeroes followed by `1` has 4097 digits and a nonzero mantissa, so this branch returns NULL even though its numeric value is exactly 1. Both official MySQL 8.0.44 and 8.4.10 return `12:34:01.000000` for `MAKETIME(12,34,<that literal>)`. The new 4097-zero test misses this because the all-zero early return runs before the cap; analogous over-limit trailing-zero forms such as `1.` followed by zeroes have the same problem.

Keep the allocation bound, but normalize/skip insignificant leading integer zeroes and trailing fractional zeroes (with the decimal exponent adjusted as needed), or otherwise cap significant digits rather than representation length. Add direct parser and binder-to-executor regressions for an over-limit zero-extended nonzero value.

gouhongshen (changes requested):
## Codex automated review

One blocking compatibility defect remains in the bounded exact-second parser.

### P1 - Cap significant digits rather than raw mantissa length (pkg/sql/plan/function/func_binary.go:7478)
The exact VARCHAR-second path counts every mantissa digit before parsing and returns NULL when that count exceeds 4096. Therefore a value such as `MAKETIME(12, 34, '000...001')` with 4096 leading zeroes is rejected even though its exact numeric value is 1 and is within the valid second range. The existing all-zero test does not catch this because it returns through the earlier zero-mantissa branch. Normalize insignificant leading/trailing zeroes or otherwise cap significant digits, and add direct plus binder-to-executor coverage.

VioletQwQ-0:
Addressed both latest change requests in `3998fb267`.

@XuPeng-SH @gouhongshen: the exact-second parser now removes insignificant leading integer zeroes and trailing fractional zeroes before applying the 4096-digit allocation bound. The removed fractional zeroes are folded into the effective decimal exponent, so padded values and exponent-cancelled forms keep their exact value without passing an unbounded mantissa to `big.Rat`.

Added direct parser and planner-to-expression-executor regressions for:
- 4096 leading zeroes followed by `1`;
- `1.` followed by 4097 zeroes;
- wide fractional/integer padding cancelled by an exponent.

Validation:
- `go test ./pkg/sql/plan/function ./pkg/sql/plan -count=1`
- focused `go test -race` for the direct parser and binder/executor cases
- `git diff --check`
- full exact-head preflight/self-review PASS on `3998fb2679d808a1a136abce3b2a1cc682141466`

gouhongshen (approved):
## Codex automated review

No findings at 3998fb2679d808a1a136abce3b2a1cc682141466. The latest significant-digit fix correctly normalizes insignificant zero padding while preserving exponent semantics and bounded allocation. Earlier binary, DECIMAL256, underflow, empty-format, rounding, and range findings remain resolved. Focused tests could not run because the environment lacks the Go toolchain.

XuPeng-SH (changes requested):
## Codex deep review — changes requested

Reviewed exact head 6e7fde2cc248cfaa5996bb9f14e9616d27bc340a. The prior binary, DECIMAL256, exponent-underflow, empty-format, rounding, range, and insignificant-zero findings remain fixed; the latest exponent-assignment cleanup is behaviorally safe. One reachable MySQL-compatibility blocker remains.

### P1 — Do not return NULL for bounded, in-range seconds with more than 4096 significant digits (pkg/sql/plan/function/func_binary.go:7499)

The allocation guard rejects every value with more than 4096 significant digits before classifying or rounding it. A roughly 4 KB ordinary VARCHAR/TEXT value such as 1. followed by 4096 copies of 1 is finite, positive, and well below 60, but MAKETIME(12,34,<value>) now returns NULL. MySQL 8.4.10 returns 12:34:01.111111 for the same value. A temporary parser -> binder -> expression-executor regression on this exact head also fails because the result vector is NULL.

This is distinct from the previous insignificant-zero issue: all digits here are significant, and the input is small enough to be routinely stored in VARCHAR/TEXT. Keep memory bounded without changing value semantics, for example by parsing only the digits needed for microsecond rounding plus a sticky remainder bit, or otherwise classifying and rounding long in-range values without constructing an unbounded big.Int. Add direct-parser and binder-to-executor coverage for this case.

Verification on the unmodified head: all existing focused TIME_FORMAT/HOUR/MAKETIME tests pass; the existing planner-to-executor tests pass; both focused package sets pass with -race -count=3. The added counterexample alone fails.

VioletQwQ-0:
Fixed the long-significant-mantissa case on `8698f4830`.

The exact VARCHAR-second path no longer rejects input by significant-digit count or builds a `big.Rat` proportional to the mantissa. It now classifies the exact decimal against 60 and performs six-digit half-away rounding with bounded state (at most eight kept digits plus the rounding digit). `1.` followed by 4096 `1`s now evaluates through the real binder/executor path as `12:34:01.111111`; huge positive exponents and values at/above 60 still return NULL, while safe underflow remains zero. I also closed the adjacent negative huge-exponent bypass so nonzero negative seconds cannot return zero through the overflow shortcut.

Validation: full `pkg/sql/plan/function` and `pkg/sql/plan`, focused `-race -count=3`, `go vet`, `git diff --check`, exact-head preflight, and complete self-review.

XuPeng-SH (approved):
Deep-reviewed exact head 8698f4830f48e49ff9d650a0ed33d641ac9dd514; no blocking findings. The bounded exact-second parser fixes the prior long-significant-mantissa defect while preserving sign, exponent, <60 classification, and half-away microsecond rounding semantics. Verified against an arbitrary-precision oracle over 10,000 randomized cases plus 8K-digit boundaries; full affected-package tests, focused race tests, build, vet, diff checks, and current CI pass.

PR #25879, opened 2026-07-20
test(parser): cover quoted Unicode and client charset bytes

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR relates to:

issue #25014

This PR continues the unfinished work from #25793 while preserving its original commits and authorship.

Review established that current `main` already accepts the quoted Arabic and Chinese aliases reported in #25014. It also established that the parser receives raw COM_QUERY bytes without the negotiated client character set, so enforcing UTF-8/BMP rules inside the scanner would reject valid non-UTF-8 client input such as latin1 `0xE9`.

The production scanner changes have therefore been removed. The final diff is deliberately test-only and records the existing compatibility contract:

- quoted Arabic and Chinese aliases remain accepted;
- escaped delimiters preserve Unicode contents;
- raw latin1 bytes remain accepted on both the fast and escaped-delimiter paths;
- raw client bytes are not reclassified by interpreting them as unconditional UTF-8.

Character-set-aware identifier validation must happen after a future COM_QUERY normalization boundary has decoded `character_set_client`; this PR does not introduce that broader protocol change.

## Validation

- `go test -count=1 -run "Test(QuotedUnicodeIdentifier|QuotedUnicodeIdentifierAliases)$" ./pkg/sql/parsers/dialect/mysql`
- `go test -race -count=10 -run "Test(QuotedUnicodeIdentifier|QuotedUnicodeIdentifierAliases)$" ./pkg/sql/parsers/dialect/mysql`
- `git diff --check`


pkg/sql/parsers/dialect/mysql/scanner.go
```diff
@@ -939,6 +948,17 @@ func (s *Scanner) peek(dist int) uint16 {
 	return uint16(s.buf[s.Pos+dist])
 }
 
+func validIdentifier(s string) bool {
+	for len(s) > 0 {
+		r, size := utf8.DecodeRuneInString(s)
+		if r == 0 || r > '\uFFFF' || r == utf8.RuneError && size == 1 {
```
XuPeng-SH: This BMP-only check is not safe for every caller. `Scan` routes backtick-quoted user variables through this same function (`@` -> `scanLiteralIdentifier`), but a user-variable name is not a schema identifier. Reproducer: clean base parses `` SELECT @`😀` ``; this head rejects it. MySQL permits otherwise-special characters in quoted user-variable names and its lexer accepts valid multibyte characters in that branch. Please split the policy by call site (or pass an explicit semantic mode) and add a regression for the user-variable case. The object-identifier path can retain the BMP restriction.

XuPeng-SH (changes requested):
Requesting changes for one correctness regression. `scanLiteralIdentifier` is also used for backtick-quoted user-variable names after `@`. User variables are not schema identifiers, so the new BMP-only predicate changes pre-existing valid parser behavior: clean base accepts `` SELECT @`😀` ``, while this head returns a syntax error. MySQL documents that quoted user-variable names may contain otherwise-unavailable characters, and its lexer consumes that branch using the client multibyte character set rather than applying the schema-identifier BMP restriction. Please make validation context-aware: schema/object/alias identifiers may reject invalid UTF-8, NUL, and code points above U+FFFF, but quoted user variables must preserve valid utf8mb4 names while still rejecting malformed input as appropriate. Add regressions for `` SELECT @`😀` `` and the escaped-delimiter slow path. The rest of the review is clean: build, vet, full parser tests, and race tests pass; the validator adds no allocations and there are no resource/lifecycle risks. Also note that the exact quoted Arabic/Chinese cases from #25014 already pass on the clean base, so the actual delta here is validation hardening rather than enabling those valid cases.

gouhongshen (changes requested):
## Codex automated review

The shared validation introduces a parser compatibility regression for quoted user variables.

### P1 - Do not apply the BMP restriction to quoted user variables (pkg/sql/parsers/dialect/mysql/scanner.go:580)
`Scan` routes `@` followed by a backtick through `scanLiteralIdentifier`, then emits `AT_ID`; the grammar accepts that as a user variable. On `origin/main`, `SELECT @`😀`` was accepted because quoted contents were returned unchanged. This change rejects it via `r > '\uFFFF'`, producing a syntax error. Escaped-backtick names use the same validator. Apply the restriction only to schema/object identifiers and add fast- and slow-path user-variable regressions.

VioletQwQ-0:
Addressed the quoted user-variable regression in 6f2112a88.

Changes:
- schema/object/alias identifiers retain the BMP, NUL, and malformed UTF-8 restrictions;
- single-@ quoted user variables use a separate validator that accepts valid utf8mb4 supplementary characters;
- @@ system-variable handling remains unchanged;
- added fast-path, escaped-backtick slow-path, malformed UTF-8, NUL, and parser-level regressions.

Validation:
- go test -count=1 ./pkg/sql/parsers/dialect/mysql
- go vet ./pkg/sql/parsers/dialect/mysql
- go build ./pkg/sql/parsers/dialect/mysql
- focused race tests passed 10 consecutive runs
- git diff --check

XuPeng-SH (changes requested):
Requesting changes for a normal-path parser performance regression.

The correctness fix is now context-correct: schema/object/alias identifiers keep the MySQL BMP/NUL/malformed-UTF-8 restrictions, single-`@` quoted user variables preserve valid utf8mb4 supplementary names, `@@` remains strict, and both fast and escaped-delimiter paths are covered.

However, every quoted identifier is currently scanned twice: `scanLiteralIdentifierWithValidator` first walks every byte to find the closing delimiter, then `validQuotedName` decodes the complete string again. This affects ordinary ORM/generated SQL, where every database/table/column name may be backtick-quoted.

Reproducible microbenchmarks on the clean base versus this head (Ryzen 9 7900X, 8 runs, 500 ms, unchanged allocations):

- 64-byte ASCII quoted identifier: median ~76.6 ns -> ~114.8 ns (**+50%**)
- 64-rune BMP quoted identifier: median ~200.7 ns -> ~328 ns (**+63%**)
- escaped-delimiter identifier: median ~160 ns -> ~221 ns (**+38%**)

A `ParseOne` benchmark also shows the same direction: a short statement with three quoted ASCII identifiers moved from roughly 1.63 us to 2.02 us median (~24%, noisier than the scanner benchmark).

Please validate during the existing delimiter scan instead of doing a full second pass: track invalid/NUL state while advancing ASCII bytes and decode only when a non-ASCII byte is encountered (while retaining the separate supplementary policy for schema identifiers versus single-`@` user variables). The escaped slow path can carry the same state or use its existing builder result because it is uncommon. A focused benchmark would prevent this from returning.

All correctness, malformed-input, ANSI_QUOTES, replacement-rune, fast/slow-path, parser-package, vet, and race checks otherwise pass; there are no resource/lifecycle concerns. Also, the exact quoted Arabic/Chinese cases from #25014 already pass on the clean base, so this PR's actual delta is validation hardening and should not materially slow the established valid path.

VioletQwQ-0:
Addressed the quoted-identifier hot-path regression in `23081459c`.

Changes:
- validate quoted names while scanning instead of decoding the completed identifier a second time;
- scan ordinary ASCII eight bytes at a time while still detecting delimiters, NUL, newlines, and non-ASCII bytes before advancing;
- decode only non-ASCII input, preserving the BMP-only policy for schema/object/alias identifiers and `@@`, while single-`@` quoted user variables continue to accept valid utf8mb4 supplementary characters;
- carry validation state through the escaped-delimiter slow path;
- add chunk-boundary, fast/slow-path, ANSI_QUOTES, malformed UTF-8, NUL, supplementary-character, allocation, scanner benchmark, and ParseOne benchmark coverage.

Benchmark comparison on the same Apple M1 (`8 x 500ms`, medians, allocations unchanged):
- old head -> this head: ASCII64 `229.2 -> 29.1 ns/op`, BMP64 `557.7 -> 387.4 ns/op`, escaped delimiter `356.2 -> 185.8 ns/op`, ParseOne with three quoted identifiers `1784 -> 1533 ns/op`;
- clean main -> this head: ASCII64 `97.8 -> 29.1 ns/op`, escaped delimiter `193.7 -> 185.8 ns/op`, and representative ParseOne `1490 -> 1533 ns/op` (`+2.9%`, versus the prior roughly `+20%` direction on this machine).

BMP input still performs the required UTF-8/BMP validation that clean main does not; the normal quoted-ASCII path no longer performs a second scan and remains `0 alloc/op`.

Validation:
- `go test -count=1 ./pkg/sql/parsers/dialect/mysql`
- focused quoted-identifier tests under `-race`
- `go vet ./pkg/sql/parsers/dialect/mysql`
- `go build ./pkg/sql/parsers/dialect/mysql`
- `git diff --check`

Self-review covered ordinary/boundary inputs, fast and escaped paths, ANSI_QUOTES, single-`@` versus `@@`, malformed UTF-8, line tracking, premature EOF, allocations, and Q1-Q3. No remaining blocker was found.

gouhongshen (approved):
## Codex automated review

No correctness, compatibility, lifecycle, concurrency, performance, or test-coverage regressions found. Verified with package tests, focused race tests, vet, and diff check.

XuPeng-SH (approved):

XuPeng-SH (approved):
Approved on 23081459cc5ba216f7a56ef8928e0f7ebd241815.

The two prior blockers are resolved: quoted single-@ user variables preserve valid utf8mb4 supplementary names without weakening schema-identifier validation, and quoted identifiers are validated during the delimiter scan so the normal ASCII path no longer pays a second full pass.

I found no remaining correctness, compatibility, lifecycle, concurrency, performance, or coverage blocker. Verified with the full parser test suite, focused race tests x10, go vet, go build, git diff --check, a deterministic 10,000-case byte-level reference-model test, and base/head benchmarks.

XuPeng-SH (changes requested):
[P1] Validate identifiers in the negotiated client character set, not as unconditional UTF-8

The new `utf8.DecodeRuneInString` check runs inside the parser without any client-charset context. However, MatrixOne accepts non-UTF-8 MySQL client character sets, and `ExecRequest` passes COM_QUERY packet bytes directly to the parser via `UnsafeBytesToString` without converting them first. A byte sequence that is invalid UTF-8 can therefore be a perfectly valid identifier in the active client character set.

Verified end to end with a real server on clean current main `4fc49f69c` and this exact head `23081459cc`:

```sh
# Send byte 0xE9, which is é in latin1, inside a quoted alias.
printf "set @@session.character_set_client='latin1';\nselect 1 as \`\351\`;\n" |
  mysql --protocol=tcp -h127.0.0.1 -P6001 -udump -p111 --batch --raw
```

- clean main: succeeds; the result metadata contains byte `E9` and the row is `1`;
- this head: `ERROR 1064` near the quoted alias.
- a direct `ParseOne("SELECT 1 AS \`\xe9\`")` A/B likewise passes on main and fails here.

This is required by the MySQL connection contract: `character_set_client` defines the encoding of statements sent by the client, and the server converts from it to `character_set_connection` ([MySQL 8.0 connection character sets](https://dev.mysql.com/doc/refman/8.0/en/charset-connection.html)). The BMP/NUL policy is correct after decoding, but raw bytes cannot be classified as malformed UTF-8 before that decoding boundary.

Please either normalize COM_QUERY from the active client charset before parsing, or make validation charset-aware while preserving the established non-UTF-8 path. Apply the same rule to the escaped-delimiter slow path and add a latin1 protocol/parser regression. Merely accepting isolated invalid UTF-8 bytes is not a complete fix because valid byte sequences in another client charset can coincidentally decode as a disallowed UTF-8 code point.

The two previously requested changes are resolved: quoted user variables preserve supplementary utf8mb4, and the single-pass fast path removes the quoted-ASCII performance regression. Full parser tests, focused race tests, vet, 20k×4-context randomized scanner differential, and same-parameter benchmarks otherwise pass.

VioletQwQ-0:
Thanks for catching the client-charset boundary issue. I confirmed on current `main` that the exact Arabic/Chinese cases and raw latin1 `0xE9` already parse successfully. Because `ParseOne` has no negotiated-client-charset context, classifying raw bytes as unconditional UTF-8 in the scanner cannot be made correct with another byte exception.

Commit `62735e73b` removes the UTF-8/BMP validation and its fast-path implementation entirely, restoring the production scanner exactly to the merge-base behavior. The final PR diff is test-only. It adds parser and scanner regressions for Arabic, Chinese, escaped delimiters, latin1 `0xE9` on both paths, and a multi-byte sequence that demonstrates raw client bytes are not classified as UTF-8 before charset normalization.

Validation passed: focused tests, full parser tests, focused race tests x10, `go vet ./pkg/sql/parsers/...`, and `git diff --check`.

XuPeng-SH (approved):
Reviewed exact head 62735e73be92a74c0b5af65078f96ecb707c981f. The prior client-character-set blocker is resolved by removing the parser-side UTF-8/BMP production changes; the final diff is test-only and accurately preserves the current raw COM_QUERY byte contract across Arabic/Chinese aliases, escaped delimiters, latin1 bytes, and byte sequences that must not be reclassified before charset normalization. No blocking correctness, compatibility, lifecycle, or coverage issues found. Full parser tests, focused race tests x10, go vet, and git diff --check passed; GitHub CI is green.

Issue #24737, opened 2026-05-31
[Bug]: Prisma relationJoins (relationLoadStrategy:'join') queries fail (v3.0.11 & 4.0-dev)


dengn:
### Summary
Prisma's JOIN-based relation loading (preview feature `relationJoins`, `relationLoadStrategy: 'join'`) builds nested derived tables that wrap rows with `JSON_OBJECT(...)` and aggregate them with `JSON_ARRAYAGG(...)` / `CAST(... AS JSON)`. These queries fail on MatrixOne.

- **3.0.11**: `not supported: function or operator 'json_object'` (and `'json_arrayagg'`).
- **4.0-dev (`nightly-0bfda604`)**: the building blocks all work standalone — `JSON_OBJECT(...)`, `JSON_ARRAYAGG(...)`, `CAST('[]' AS JSON)`, `COALESCE(JSON_ARRAYAGG(...), CAST('[]' AS JSON))` — but the **full nested query** still fails with a parser error `near " JSON))"`. To-**one** relation loads work on 4.0-dev; to-**many** fail.

### Generated SQL that fails (4.0-dev, to-many `author → posts`, trimmed)
```sql
SELECT `t1`.`id`, `t1`.`name`,
  COALESCE(`posts`.`__prisma_data__`, CAST('[]' AS JSON)) AS `__prisma_data__`
FROM `Author` AS `t1`
LEFT JOIN ( SELECT JSON_ARRAYAGG(`t5`.`__prisma_data__`) ...
  ( SELECT JSON_OBJECT('id', `t3`.`id`, 'title', `t3`.`title`, ...) AS `__prisma_data__`
    FROM ( SELECT `t2`.* FROM `Post` AS `t2` WHERE `t1`.`id` = `t2`.`authorId` ) AS `t3`
  ) ... ) AS `posts`
WHERE `t1`.`id` = ?;
-- ERROR 1064 (HY000): SQL parser error ... syntax error near " JSON)) AS `__prisma_data__` ..."
```

### Prisma impact
Any query using `relationLoadStrategy: 'join'` fails. Workaround: use the default `'query'` strategy (separate queries), which works fine.

### Expected Behavior
The generated JSON-aggregation relation query parses and executes (to-many relations included).

---
*Found during deep Prisma ORM (v5.22, `relationJoins` preview) compatibility testing; reproduced on prod `8.0.30-MatrixOne-v3.0.11` and `nightly-0bfda604`.*

[...]

PR #24772, opened 2026-06-02
fix(parser): allow LEFT/RIGHT/FULL JOIN without ON/USING clause

ULookup:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24737

## What this PR does / why we need it:

Fixes a SQL parser bug where LEFT/RIGHT/FULL JOIN without ON/USING clause fails to parse when followed by WHERE or other clauses. The fix changes the outer_join grammar production from mandatory `join_condition` to optional `join_condition_opt`, matching the behavior already used for INNER JOIN.

This unblocks Prisma ORM's `relationLoadStrategy: 'join'` preview featur for to-many relations, which generates LEFT JOIN without ON and uses the outer WHERE for filtering.

PR #25880, opened 2026-07-20
fix(parser): report missing outer join conditions

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24737

This PR continues the unfinished work from #24772 while preserving its original commits and authorship.

The inherited change allowed LEFT/RIGHT/FULL OUTER JOIN without `ON` or `USING`. Review established that this is not valid MySQL outer-join syntax and that conditionless JOIN semantics should remain limited to inner/cross-product forms.

The implementation now:

- keeps `ON` or `USING` mandatory for LEFT, RIGHT, and FULL OUTER JOIN;
- returns the explicit parser error `outer join requires ON/USING clause` when the condition is missing;
- continues accepting ordinary JOIN, INNER JOIN, CROSS JOIN, STRAIGHT_JOIN, and NATURAL OUTER JOIN without an explicit condition where their grammar permits it;
- removes the inherited positive BVT cases that executed conditionless outer joins;
- covers LEFT/RIGHT/FULL, derived-table, trailing-clause, chained-join, valid `ON/USING`, and adjacent join semantics.

## Validation

- regenerated `mysql_sql.go` from `mysql_sql.y` with zero parser conflicts
- verified parser generation is deterministic (`make -B -C pkg/sql/parsers/dialect/mysql mysql_sql.go` leaves no diff)
- `go test -count=1 -run '^TestOuterJoinRequiresCondition$' ./pkg/sql/parsers/dialect/mysql`
- `go test -race -count=10 -run '^TestOuterJoinRequiresCondition$' ./pkg/sql/parsers/dialect/mysql`
- `git diff --check`


XuPeng-SH (approved):
Approved on head `8429c29b49b7e55dbe831ebefd142ed4992ae811`.

The grammar keeps conditionless INNER/CROSS/STRAIGHT and NATURAL OUTER joins valid while LEFT/RIGHT/FULL OUTER joins without ON/USING terminate with the explicit parser error. Valid ON/USING forms remain unchanged. I also verified condition ownership across chained and parenthesized joins, comma boundaries, and the same table-reference grammar in UPDATE/DELETE.

Validation: deterministic goyacc regeneration with zero diff and zero conflicts, `git diff --check`, full `go test ./pkg/sql/parsers/...`, `go vet ./pkg/sql/parsers/...`, target race test repeated 20 times, an independent join-reduction matrix, clean merge-tree against latest main, and all required CI checks green. No concurrency, lifecycle, leak, wait, or unbounded-growth surface is introduced.

The open predecessor PR #24772 still needs to be closed for Mergify to resume evaluating this PR; that is procedural and not a code blocker.

Issue #24550, opened 2026-05-25
[Feature Request]: Percentile / quantile aggregate functions (p95 / p99)


dengn:
**Version:** `8.0.30-MatrixOne-v3.0.11` (MatrixOne v3.0.11)

Only `median` (p50) is available; there is no general percentile/quantile function. p95/p99 are core metrics for latency/SLA analytics and observability.

**Reproduce**
```sql
SELECT median(v) FROM t;                                          -- OK (p50 only)
SELECT quantile(0.95)(v) FROM t;                                  -- SQL parser error
SELECT approx_percentile(v, 0.95) FROM t;                         -- not supported
SELECT percentile_cont(0.95) WITHIN GROUP (ORDER BY v) FROM t;    -- SQL parser error
```

**Request**
Add percentile/quantile aggregates (ideally both exact and an approximate / t-digest variant) so p95/p99 can be computed.

---
_Found while evaluating MatrixOne git4data / SQL as a data-versioning & agent-trace backend. Version: `8.0.30-MatrixOne-v3.0.11` (MatrixOne v3.0.11)._

## Testing background
Found while building **observability rollups over agent traces** (p95 / p99 latency per model / per tool) and while benchmarking MatrixOne against ClickHouse as an OTel trace backend. Only `median` (p50) is available.

## Why it matters (expected purpose)
p95/p99 latency are **core SLA / observability metrics** and standard analytics. Without percentile/quantile (ideally an approximate / t-digest variant for scale) MatrixOne can't serve these dashboards, which is a key gap versus OLAP engines used for tracing.

VioletQwQ-0:
## Post-merge QA handoff

Merged implementation: #25881, merge commit `ed1a70e2d756b895cec3f23254dafa98e200f27d` (final PR head `1f9949481a13743457ed8221173ae48bffd3f870`).

QA Decision
- required: yes
- reason: this adds a user-visible aggregate with approximate numerical semantics, distributed partial-state merging, bounded memory/state behavior, and exact DECIMAL interpolation requirements
- production_entrypoint: execute `approx_percentile(value, percentile)` through SQL in grouped and ungrouped aggregation on standalone and multi-CN deployments
[...]

PR #24667, opened 2026-05-28
feat(aggexec): add approx_percentile aggregate function for p95/p99

ULookup:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24550

## What this PR does / why we need it:

Adds the approx_percentile(v, percentile) aggregate function that computes
the approximate percentile of a numeric column using O(n) quickselect
with linear interpolation. This enables p95/p99 latency analytics and
observability dashboards previously limited to median (p50) only.

The implementation wires together the parser grammar (already present),
plan-layer function registration, execution-layer executor (reusing the
median infrastructure), and compile-layer config extraction for the
percentile constant.

PR #25881, opened 2026-07-20
feat(aggexec): add approx_percentile aggregate function for p95/p99

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24550

This PR continues the unfinished work from #24667 while preserving its original commits and authorship.

It adds approx_percentile(value, percentile) with:

- a bounded, mergeable KLL-style sketch instead of retaining every input row;
- bounded serialized partial state for distributed aggregation;
- exact rational/integer DECIMAL64 and DECIMAL128 interpolation without float64 conversion;
- exact min/max results for percentile 0 and 1;
- planner and executor validation for non-null constant percentiles in [0,1];
- current main merged with function IDs reconciled.

Review-blocker regressions cover 10K rows, duplicate values, NULLs, four-way partial merge, intermediate-state round trips, replacement of reused executors with empty and multi-chunk state, state-size bounds, mismatched merge configuration, and DECIMAL64/128 values above 2^53.

## Validation

- go test -count=1 -timeout 180s ./pkg/sql/colexec/aggexec ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function
- go test -race -count=10 -run TestApproxPercentileExec_... ./pkg/sql/colexec/aggexec
- go vet ./pkg/sql/colexec/aggexec ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function
- go build ./pkg/sql/colexec/aggexec ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function
- git diff upstream/main --check


aunjgr (changes requested):
[P1] The retained quantile samples are allocated in ordinary Go slices (`quantileSketch.levels`), but group spilling measures only `ctr.mp.CurrNB()`. The new aggregate `Size()` is not included in `group.memUsed()`. High-cardinality grouping can therefore accumulate O(groups * retained samples) on the Go heap without triggering spill or the managed-memory limit. Please move the retained sample buffers to mpool-backed off-heap storage and close allocation, compaction, merge, unmarshal, reset, and free ownership paths. At minimum spill accounting must include aggregate state, but that alone would still leave these value buffers outside mpool.

aunjgr (changes requested):
[P1] The retained quantile samples are allocated in ordinary Go slices (`quantileSketch.levels`), but group spilling measures only `ctr.mp.CurrNB()`. The new aggregate `Size()` is not included in `group.memUsed()`. High-cardinality grouping can therefore accumulate O(groups * retained samples) on the Go heap without triggering spill or the managed-memory limit. Please move the retained sample buffers to mpool-backed off-heap storage and close allocation, compaction, merge, unmarshal, reset, and free ownership paths. At minimum spill accounting must include aggregate state, but that alone would still leave these value buffers outside mpool.

heni02 (approved):

VioletQwQ-0:
Addressed the requested P1 memory-accounting fix in 7f6cc8f55.

- Retained KLL level backing buffers and quantile scratch space now use mpool-backed off-heap allocations, so the existing group spill metric (`ctr.mp.CurrNB()`) accounts for them directly. I intentionally did not add `Agg.Size()` on top of `CurrNB()`, because that would double-count these same buffers.
- Add, compaction, and merge grow replacement buffers through mpool and release the previous allocation only after the replacement succeeds.
- Unmarshal decodes into a temporary mpool-owned sketch, frees partial state on failure, and atomically replaces the prior state on success. Custom aggregate states now have an optional Free lifecycle that `aggState.free` invokes, including failed aggregate-state unmarshal cleanup.
- Added regression coverage for mpool visibility, merge/compaction, round-trip restore, quantile scratch cleanup, trailing-data rejection, truncated decode cleanup, and final zero `CurrNB()`.

Validation:
- `go test ./pkg/sql/colexec/aggexec -count=1`
- `go test ./pkg/sql/colexec/group -count=1`
- `go test -race ./pkg/sql/colexec/aggexec -run "TestApproxPercentile|TestQuantileSketch" -count=1`
- `go vet ./pkg/sql/colexec/aggexec ./pkg/sql/colexec/group`
- `git diff --check`

VioletQwQ-0:
Addressed the TIME_WINDOW `approx_percentile` blocker in `8bd441101`.

- GROUP, WINDOW, and TIME_WINDOW now share one special-aggregate construction path, so `approx_percentile` validates and evaluates its constant percentile once, stores it as executor config, and removes the configuration argument from row inputs in all three modes.
- The same helper preserves the existing `group_concat` separator and `cluster_centers` configuration behavior.
- Added compile regressions for percentile endpoints `0`/`1`, below/above-range values, non-constant arguments, and both other special aggregates.
- Added TIME_WINDOW executor regressions proving `0 -> min`, `1 -> max`, invalid executor configuration rejection, and zero remaining mpool bytes on both success and failure.
- Added a SQL BVT case covering both percentile endpoints through TIME_WINDOW. The BVT fixture was not run locally because this worktree has no running MO/mo-tester; it will be exercised by CI.

Validation:
- `go test ./pkg/sql/colexec/aggexec ./pkg/sql/colexec/timewin ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function -count=1`
- `go test ./pkg/sql/colexec/group ./pkg/sql/colexec/window -count=1`
- `go test -race ./pkg/sql/colexec/timewin -run 'TestTimeWinApproxPercentile' -count=1`
- `go vet ./pkg/sql/colexec/aggexec ./pkg/sql/colexec/timewin ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function`
- `git diff --check`

Self-review also rechecked the aggregate lifecycle: all new temporary/retained mpool allocations have success and error cleanup, there are no waits or goroutines in this path, and KLL levels/retained samples remain bounded.

VioletQwQ-0:
Addressed the new float64 interpolation overflow in `95260cf74`.

Changes:
- added one shared `interpolateFloat64` path for both KLL executor flush and the exact numeric helper;
- finite opposite-sign endpoints now use an overflow-safe FMA weighted sum, so `[-MaxFloat64, +MaxFloat64]` at `p=0.5` returns `0` instead of `+Inf`;
- `p=0` and `p=1` return the finite endpoints exactly;
- same-sign interpolation keeps the existing formula, and NaN/Inf input values keep the prior IEEE-754 propagation behavior.

Coverage:
- both KLL executor and exact-helper extreme-value paths;
- opposite-sign and same-sign extrema, negative values, subnormal values, endpoints, and existing NaN/Inf behavior.

Validation:
- `go test ./pkg/sql/colexec/aggexec -count=1`
- focused regression under `go test -race`
- `go test ./pkg/sql/colexec/timewin ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function ./pkg/sql/colexec/group ./pkg/sql/colexec/window -count=1`
- `go vet ./pkg/sql/colexec/aggexec`
- `git diff --check`
- full `mo-self-review` vs main: zero open blockers

aunjgr (approved):
Reviewed head 95260cf74. The retained KLL buffers and quantile scratch storage are now mpool-backed and fully released across merge, unmarshal, failure, and final cleanup. GROUP, WINDOW, and TIME_WINDOW share correct percentile configuration handling, and extreme float interpolation no longer overflows. No remaining blockers found.

gouhongshen (changes requested):
## Codex automated review

The bounded sketch and lifecycle handling are coherent, but integer interpolation has a blocking correctness bug.

### P1 - Preserve int64 precision through interpolation (pkg/sql/colexec/aggexec/approx_percentile.go:745)
Both integer rank endpoints are converted to float64 before interpolation. For the valid two-row input `[-9007199254740993, 9007199254740992]` at p=0.5, no sketch compaction occurs and the exact result is -0.5, but the conversions round the endpoints to -2^53 and 2^53, so this code returns 0. The same failure occurs for MinInt64/MaxInt64. Interpolate int64 values without rounding the endpoints first and add an opposite-sign large-integer regression.

VioletQwQ-0:
Addressed the integer interpolation precision blocker in `ee2cb37d2`.

Changes:
- Numeric executor flush and the exact helper now share `interpolateNumeric`.
- Integer endpoints are converted to exact `big.Int` values and interpolated with the existing rational rank fraction before the final `float64` rounding.
- Float inputs still use the existing overflow-safe IEEE-754 path; decimal behavior is unchanged.

Coverage:
- `[-9007199254740993, 9007199254740992]` at `p=0.5` returns `-0.5` through both the exact helper and the real executor.
- `MinInt64/MaxInt64` and adjacent large `uint64` regression cases.
- Existing float64 extreme, endpoint, NaN, and Inf cases remain covered.

Validation:
- `go test ./pkg/sql/colexec/aggexec -count=1`
- focused regression under `go test -race`
- `go test ./pkg/sql/colexec/timewin ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/plan/function ./pkg/sql/colexec/group ./pkg/sql/colexec/window -count=1`
- `go vet ./pkg/sql/colexec/aggexec`
- `git diff --check`
- `mo-self-review`: zero open blockers

XuPeng-SH (changes requested):
[P1] Preserve valid DECIMAL(38, s) endpoints instead of overflowing during interpolation.

The decimal return contract is not closed at maximum precision. `MedianReturnType` always declares `DECIMAL128(38, inputScale+1)`, while `interpolateDecimal` always multiplies the unscaled result by 10. A valid case such as:

```sql
CREATE TABLE t(v DECIMAL(38,0));
INSERT INTO t VALUES (99999999999999999999999999999999999999);
SELECT approx_percentile(v, 0) FROM t;
```

must return the input exactly, as promised for percentile 0, but the executor multiplies that 38-digit unscaled value by 10 and `decimal128FromBigInt` returns `approx_percentile: decimal interpolation overflow`. `DECIMAL(38,38)` also produces an invalid/unrepresentable result scale of 39. This affects both endpoints and ordinary interpolation near the precision limit.

Please define an approx-percentile decimal return rule that increases scale only when the declared precision has room (or otherwise chooses a representable result scale), and make interpolation use the actual input-to-output scale delta instead of an unconditional factor of 10. Add executor and SQL regressions for DECIMAL(38,0) maximum-magnitude p=0/p=1, DECIMAL(38,38), and a fractional percentile that exercises rounding at the selected result scale.

I rechecked the complete current head: the prior mpool ownership, failed-unmarshal cleanup, GROUP/WINDOW/TIME_WINDOW configuration, float64 overflow, and integer precision blockers are closed. The KLL retained state and serialized state are bounded, merge preserves represented weight, and no wait/concurrency paths are introduced. Focused tests across aggexec/compile/plan/function passed on this head.

gouhongshen (changes requested):
## Codex automated review

One blocking DECIMAL(38, s) correctness issue remains.

### P1 - Preserve valid DECIMAL(38, s) endpoint values (pkg/sql/colexec/aggexec/approx_percentile.go:848)
Decimal inputs receive `MedianReturnType`, i.e. DECIMAL128(38, inputScale+1), but this unconditional `* 10` assumes that extra scale is always representable. For `DECIMAL(38,0)` value `99999999999999999999999999999999999999` at percentile 0 or 1, it produces an unscaled 39-digit value and `decimal128FromBigInt` returns `approx_percentile: decimal interpolation overflow` instead of the required endpoint. `DECIMAL(38,38)` likewise declares scale 39. Choose a representable output scale when precision is already 38, apply the actual input/output scale delta rather than always multiplying by 10, and add executor and SQL coverage for max-magnitude endpoints and fractional rounding.

VioletQwQ-0:
Addressed the DECIMAL(38, s) P1 in f7c5a0281.

Changes:
- add an approx-percentile-specific decimal return rule: increase scale only while input precision has room; DECIMAL(38, s) retains its input scale;
- interpolate using the actual input-to-output scale delta instead of unconditional multiplication by 10;
- cover DECIMAL(38,0) maximum positive/negative p=0/p=1 endpoints, DECIMAL(38,38), positive/negative midpoint rounding at retained scale, and DECIMAL(37) scale expansion in executor, planner, and SQL BVT tests.

Validation:
- go test ./pkg/sql/colexec/aggexec -count=1
- focused race test
- go vet and go build for aggexec/function/compile
- git diff --check

pkg/sql/colexec/aggexec/register.go
```diff
@@ -129,6 +129,11 @@ func RegisterMedian(id int64) {
 	AggIdOfMedian = id
 }
 
+func RegisterApproxPercentile(id int64) {
+	specialAgg[id] = true
```
fengttt: I thought we have, or at least able to, get rid of these Register stuff.    They are totally useless from an extremely bad design/organization of code.

Can you double check if ALL these RegisterXXX code are dead code?    We should remove them all.

pkg/sql/colexec/aggexec/register.go
```diff
@@ -244,3 +249,5 @@ var (
 		return types.T_text.ToType()
 	}
 )
+
+var AggIdOfApproxPercentile = int64(-37)
```
fengttt: And this code, they are just horrible.   Can we get rid of all these stuff?

fengttt (changes requested):
We should get rid of ALL those RegisterXXX stuff.    We should either do this in this PR, or, if turns out to be tedious, open another Issue and clean everything up in another PR.

pkg/sql/colexec/aggexec/register.go
VioletQwQ-0: Handled in f93528020. I removed every aggregate/window RegisterXXX function, the specialAgg map, mutable runtime IDs, and the planner init callbacks. MakeAgg now dispatches directly on immutable encoded IDs, and planner metadata keeps only aggName. TestAggregateExecutorIDs pins all 38 aggregate/window catalog entries to their executor IDs; focused tests, race tests, go vet, and a full pkg/sql/... compile all pass.

pkg/sql/colexec/aggexec/register.go
VioletQwQ-0: Handled in f93528020 as part of the same cleanup: register.go is deleted, JSON registration helpers are removed, and the one-field aggregation wrapper is flattened to aggName. No init-time registration remains.

gouhongshen (approved):
## Codex automated review

No actionable regressions found. Prior blockers are resolved at the current head; aggregate/window ID dispatch, bounded mpool-backed sketch lifecycle, merge/serialization, numeric and DECIMAL interpolation, constant-percentile validation, and GROUP/WINDOW/TIME_WINDOW integration are coherent. Current-head unit, BVT, multi-CN, SCA, and coverage CI checks pass.

aunjgr (approved):
Reviewed the current head independently. The sketch state is bounded and its retained value buffers are mpool-backed, with free/reset/unmarshal error ownership closed. Integer interpolation preserves endpoint precision with exact big-integer arithmetic, decimal return scale is capped at precision 38 and interpolation applies the actual scale delta, and aggregate configuration is propagated through group/window/time-window execution. I found no remaining merge-blocking correctness or lifecycle issue.

XuPeng-SH (approved):
Re-reviewed exact head 6bbd7d70b6d178f3d184ee92ea6445cd8897a311 after the main merge. No blocking findings.

The prior blockers are closed: retained KLL sample and scratch buffers are mpool-backed and released across grow, merge, decode failure, restore, spill, and final cleanup; integer and DECIMAL interpolation preserve endpoint precision and the DECIMAL(38,s) return contract; all aggregate/window RegisterXXX runtime registration has been removed and the immutable catalog-to-executor ID map is pinned by tests. The only semantic merge conflict was in TIME_WINDOW construction, and its resolution correctly combines the current layout/AggIdx logic with the shared percentile configuration path. GROUP, WINDOW, TIME_WINDOW, and remote pipeline config propagation remain coherent.

Fresh validation on this head:
- full tests for aggexec, group, window, timewin, compile, plan, and plan/function: PASS
- focused lifecycle, partial-merge, decimal, TIME_WINDOW, binder, and ID tests under race, count 10: PASS
- build and vet for all seven packages: PASS
- merge-tree against current origin/main and diff-check: clean
- required CI: green

pkg/sql/colexec/aggexec/aggState.go
```diff
@@ -781,9 +791,15 @@ func checkAggStateMagic(reader io.Reader) {
...
 	defer checkAggStateMagic(reader)
+	defer func() {
+		if retErr != nil {
+			ae.Free()
+			ae.state = nil
+		}
+	}()
 
 	// Always unmarshal from a clean state.
 	ae.Free()
```
XuPeng-SH: Free releases every vec and freeable mob but does not clear the outer state slice or reset each state length/capacity. Consequently, UnmarshalFromReader on an already-populated executor followed by a valid zero-chunk encoding returns success at the cnt == 0 branch while GetNumGroups still reports the old groups and Flush indexes the now-nil mobs/vecs (panic). A multi-chunk replacement likewise starts preallocation/appending from freed stale states. I reproduced this with an approx_percentile executor: GroupGrow(1), then unmarshal SaveIntermediateResult(0, nil, ...); UnmarshalFromReader returns nil but GetNumGroups is 1 instead of 0. Clear ae.state immediately after this initial Free (as the error cleanup already does), and add empty-state plus reused multi-chunk regressions.

XuPeng-SH (changes requested):
One lifecycle correctness blocker remains in the generic aggregate-state unmarshal path. The approx_percentile implementation, registry cleanup, numeric/DECIMAL behavior, configuration propagation, bounded state, and current CI otherwise look good. Local related-package tests, race tests, build, and vet pass; the focused reuse regression below fails on this head.

pkg/sql/colexec/aggexec/aggState.go
VioletQwQ-0: Fixed in `8cbe5feb4`.

`aggExec.UnmarshalFromReader` now clears the outer `state` slice immediately after freeing its previous chunks. A successful zero-chunk decode therefore leaves zero groups, and multi-chunk decode now allocates from a genuinely empty state instead of appending through freed chunk metadata.

Added `TestAggExecUnmarshalReplacesExistingState` with two independent reuse regressions:
- populated executor -> valid zero-chunk state, asserting `GetNumGroups() == 0`;
- populated executor -> `AggBatchSize+1` groups encoded as two chunks, asserting the exact replacement group count and a successful two-vector `Flush`.

The empty-state regression failed on `6bbd7d70b` with one stale group before the fix. Validation on `8cbe5feb4`: focused test `-count=20`, focused `-race -count=10` before the final test split and the full lifecycle matrix under `-race -count=5` after it, full tests/build/vet for aggexec, group, window, timewin, compile, plan, and plan/function, plus `git diff --check`.

XuPeng-SH (approved):
Re-reviewed exact head `8cbe5feb459639af642b31abb29c617d9ad10b70`. The generic aggregate-state reuse blocker is fixed: `UnmarshalFromReader` now drops the freed outer chunk metadata before decoding, so an empty replacement has zero groups and multi-chunk replacement cannot append through stale freed state.

Independent lifecycle validation covered the PR regression 20x; real freeable approx_percentile sketch reuse from populated -> empty, populated -> two chunks with live mobs across the chunk boundary, and malformed replacement -> clean empty state/mpool zero 10x; the combined matrix under `-race` 10x; full `aggexec` tests and full-package race; related batch/timewin/compile/plan/function tests and vet; `git diff --check`; and a conflict-free merge tree against current `origin/main`. The previously reviewed bounded-state/mpool accounting, config propagation, numeric/DECIMAL correctness, and registration cleanup remain intact.

Current-head GitHub CI has no failures at review time, though several jobs are still in progress; this approval is based on the exact-head code and independent validation above.

gouhongshen (approved):
## Codex automated review

No blocking regression found. The author’s latest lifecycle explanation is supported by the exact-head code, and current-head CI is green. One helper-only DECIMAL inconsistency remains non-blocking because the production SQL/executor path does not call it.

### P2 - Apply the capped decimal scale rule to the direct helper (pkg/sql/colexec/aggexec/approx_percentile.go:1034)
The exported `PercentileDecimal128` helper still unconditionally passes `scaleDelta=1` and ignores `argScale`/the vector’s precision. For a DECIMAL(38,0) vector containing `99999999999999999999999999999999999999`, even `p=0` multiplies the unscaled endpoint by 10 and returns `decimal interpolation overflow`—the same max-precision defect already fixed in the executor path. There are no production callers in this repository, so this is non-blocking, but the direct-caller contract and regression coverage should be aligned with `ApproxPercentileReturnType`.

VioletQwQ-0:
@fengttt The aggregate/window registration layer was removed in f93528020b.
register.go, all RegisterXXX functions, specialAgg, mutable runtime IDs, and planner init callbacks are gone. Aggregate construction now dispatches directly from fixed catalog IDs, with tests covering the full ID mapping.
The remaining proto.RegisterType is generated protobuf code and unrelated to this registration layer. Please re-review the current head.

gouhongshen (approved):
## Codex automated review

Reviewed exact head 8cbe5feb459639af642b31abb29c617d9ad10b70 and refreshed all PR discussion. The author’s latest registration-cleanup statement is supported: runtime RegisterXXX machinery is gone and all 38 fixed executor IDs are pinned against catalog IDs. The aggregate-state reuse fix is correct, current-head CI is green, and no new actionable regression remains. Local focused tests could not start because the worktree lacks prebuilt cgo/libmo.dylib; the worktree remains clean.

VioletQwQ-0:
Fixed the direct DECIMAL128 helper inconsistency in `12c0bcdd5`.

`PercentileDecimal128` now reads the input precision from the vector type and combines it with the existing `argScale` argument, then uses the same `ApproxPercentileReturnType` rule as the production executor. It no longer unconditionally adds one scale digit when the input is already at precision 38.

Added direct-helper regressions for:
- DECIMAL(38,0) maximum positive/negative endpoints at p=0 and p=1;
- DECIMAL(38,38);
- retained-scale midpoint rounding;
- DECIMAL width 37 continuing to gain one result scale digit.

Validation on the complete final diff:
- related package tests for aggexec, timewin, compile, plan, plan/function, group, and window;
- full aggexec race test plus focused decimal boundary tests under `-race -count=10`;
- vet and build for all seven related packages;
- `git diff --check`;
- full preflight manifest and `mo-self-review`: exact-snapshot PASS, no remaining blocker or pending probe.

@gouhongshen @fengttt please re-review the current head.

gouhongshen (approved):
## Codex automated review

Reviewed exact head 12c0bcdd5 against origin/main after loading all existing reviews, issue comments, inline discussions, and the latest author reply. The new DECIMAL128 helper fix is consistent with aggregate return-type handling and has targeted boundary tests. Previously reported correctness, memory-accounting, registration, and state-reuse issues are resolved. Exact-head CI is green. Focused local tests could not start because the required prebuilt cgo/libmo.dylib is absent; no dependencies were modified.

gouhongshen (changes requested):
## Codex automated review

One blocking partial-aggregation merge panic remains; the previously reviewed lifecycle and DECIMAL issues are addressed.

### P1 - Avoid indexing a missing parity level during partial sketch merge (pkg/sql/colexec/aggexec/approx_percentile.go:115)
A 400-row source sketch compacts level 0 to an empty slice and toggles `other.parity[0]`. When it is merged into a newly created destination group, `appendValues(0, empty)` returns before `ensureLevel` (lines 148-152), so `s.parity` is still empty and line 116 panics. This occurs in distributed/partial aggregation whenever the merge-side group has no local values. The merge tests populate the destination before merging, so they do not cover this shape.

VioletQwQ-0:
Addressed the current-head partial-merge panic in `21026c8a0`.

- `quantileSketch.Merge` now creates the destination level/parity slot before copying values, including an empty compacted level whose parity bit still carries state.
- Added a regression that compacts exactly `2*k` source rows, verifies level 0 is empty with parity set, and merges into a destination with no local values.

Validation:
- `go test ./pkg/sql/colexec/aggexec -count=1`
- focused `go test -race ... -count=10`
- `go vet ./pkg/sql/colexec/aggexec`
- `git diff --check`
- exact-head preflight PASS on `21026c8a0cb9d5552a9dc77e75a3252c78c9f809`

gouhongshen (approved):
## Codex automated review

The upstream merge preserves the prior correctness fixes and resolves the new function-ID collision consistently. No blocking regression remains; one non-blocking memory-efficiency issue persists.

### P2 - Allocate sketches lazily for groups containing values (pkg/sql/colexec/aggexec/approx_percentile.go:570)
GroupGrow eagerly creates a quantileSketch for every group, including high-cardinality groups whose aggregate argument is always NULL. Fill, BulkFill, BatchFill, and merge already allocate lazily, while Flush explicitly handles nil state, so these empty sketches are unnecessary. Their Go-managed objects and slice headers are also excluded from the group spill calculation, which only reads ctr.mp.CurrNB(); consequently an all-NULL high-cardinality aggregation can accumulate substantial untracked heap memory. Remove this loop and retain ensureSketch at the existing value/merge sites.

gouhongshen (approved):
## Codex automated review

Prior blocking review findings are resolved at fa762ce. One non-blocking memory-efficiency issue remains.

### P2 - Allocate sketches lazily for NULL-only groups (pkg/sql/colexec/aggexec/approx_percentile.go:570)
GroupGrow eagerly creates a quantileSketch for every group. Fill/BatchFill skip NULL inputs before ensureSketch, merge skips nil source state, and Flush already emits NULL for a nil sketch, so groups whose value is always NULL never need one. These per-group Go heap objects and slice headers are outside mpool accounting; high-cardinality NULL-only aggregation therefore adds avoidable untracked heap pressure. Remove this eager loop and retain the existing lazy creation at value/merge sites.

gouhongshen (changes requested):
## Codex automated review

Prior blockers for mpool lifecycle, configuration propagation, numeric/DECIMAL precision, registration removal, state replacement, and partial merges are fixed. The prior lazy NULL-only sketch allocation remains unchanged and non-blocking. A non-finite float endpoint failure is blocking.

### P1 - Return extrema for infinite float inputs (pkg/sql/colexec/aggexec/approx_percentile.go:925)
Author response: non-finite inputs were intentionally left with the prior IEEE-754 propagation behavior. Why this remains blocking: non-finite float values are not excluded—`CAST('Inf' AS DOUBLE)` is accepted by the float cast path—and this aggregate accepts float64. For `approx_percentile(CAST('Inf' AS DOUBLE), 1)`, `Quantile` returns `lo=hi=+Inf` with zero fraction, but `interpolateFloat64` checks for Inf before the zero-fraction case and evaluates `+Inf + (+Inf - +Inf) * 0`, producing NaN. Thus p=1 fails to return the maximum; p=0 and singleton inputs have the same defect. Move the zero-fraction/equal-endpoint handling before the non-finite branch, and add executor/SQL coverage for +/-Inf endpoints and singleton inputs.

VioletQwQ-0:
Fixed the non-finite endpoint issue in 07d1efd67.

`interpolateFloat64` now returns exact rank endpoints and equal endpoints before evaluating non-finite arithmetic, so singleton/equal `+Inf` and `-Inf` no longer become `NaN`. Distinct `-Inf/+Inf` interpolation still keeps IEEE propagation.

Added coverage for:
- `p=0`, interior, and `p=1` across finite, equal-infinity, and opposite-infinity endpoints
- the real `Fill`, `BulkFill`, and `BatchFill` executor paths, including const-row and NULL handling
- `Merge -> SaveIntermediateResult -> UnmarshalFromReader -> Flush` with infinite values
- distributed SQL `+Inf`/`-Inf` singleton and endpoint results

Validation passed: full `aggexec`, `compile`, `plan`, `plan/function`, and `timewin` package tests; focused `-race`; `go vet`; `git diff --check`; exact-head preflight review. The previous-head Ubuntu UT failure was an unrelated `TestReplay1` port collision (`0.0.0.0:32002 already in use`); this new head will rerun CI.

gouhongshen (approved):
## Codex automated review

The previously blocking non-finite endpoint issue is fixed: exact ranks and equal infinite endpoints now bypass invalid infinity arithmetic, with executor, merge/serialization, and SQL regressions. Earlier blockers remain resolved. The unchanged eager allocation for NULL-only groups remains accepted as non-blocking and is not repeated in this follow-up.

VioletQwQ-0:
Merged the latest `upstream/main` in 68be31e63 and resolved the `pkg/sql/compile/operator.go` conflict. The resolution keeps the new unified `constructAggregateConfig` and `group_concat_max_len` behavior, while routing APPROX_PERCENTILE through the same GROUP/WINDOW/TIME_WINDOW config path. Focused compile tests and the exact-head preflight pass; new CI is running.

fengttt (approved):

fengttt (approved):

Issue #24549, opened 2026-05-25
[Feature Request]: Support schema evolution in DATA BRANCH DIFF / MERGE


dengn:
**Version:** `8.0.30-MatrixOne-v3.0.11` (MatrixOne v3.0.11)

`DATA BRANCH DIFF/MERGE` require the two tables to have an **identical schema**. Adding a column on a branch (a normal data-evolution step, e.g. a new feature column) breaks diff/merge.

**Reproduce**
```sql
CREATE TABLE a(id INT PRIMARY KEY, f0 DOUBLE);
DATA BRANCH CREATE TABLE b FROM a;
ALTER TABLE b ADD COLUMN f1 DOUBLE DEFAULT 0.0;
DATA BRANCH DIFF b AGAINST a;
```

**Actual**
```
ERROR 20101 (HY000): internal error: the target table schema is not equivalent to the base table.
```

**Request**
Support diff/merge across compatible schema changes (added / nullable columns), which is essential for evolving feature/label tables across versions.

---
_Found while evaluating MatrixOne git4data / SQL as a data-versioning & agent-trace backend. Version: `8.0.30-MatrixOne-v3.0.11` (MatrixOne v3.0.11)._

## Testing background
Found in a **feature-store experiment**: a feature definition changed on a branch (we added a feature column via `ALTER TABLE ... ADD COLUMN`), then we tried to `DATA BRANCH DIFF`/`MERGE` the new feature values against the base.

## Why it matters (expected purpose)
Feature/label/training tables **evolve over time** (new features get added). Requiring identical schemas means any column addition breaks diff/merge, blocking continuous feature engineering on branches. Supporting compatible schema evolution (added / nullable columns) is needed for iterative dataset work.

VioletQwQ-0:
## QA handoff

Merged implementation: #25882, merge commit `2ddf1d4c37c64e9b074aee9cab27a3234a667bc5` (final PR head `c86eb01c76582c26971b535caad91c8191f26227`).
Current `main` at handoff: `41332062a6740978f269088a92ee362a888c7bd7`.

QA Decision
- required: yes
- reason: this is user-visible DATA BRANCH DIFF/MERGE schema-evolution functionality involving persisted lineage, historical ownership, snapshot/PITR, concurrency, and GC
- production_entrypoint: DIFF/MERGE across supported schema evolution, plus COPY ALTER while branch/snapshot/PITR historical owners are created or removed
[...]

PR #24666, opened 2026-05-28
feat(frontend): support schema evolution in data branch diff/merge

ULookup:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #24549

## What this PR does / why we need it:

This PR supports schema evolution in DATA BRANCH DIFF/MERGE by replacing the strict schema equivalence check with a named-column compatibility check. When a branch table has extra columns (e.g., added via ALTER TABLE ADD COLUMN), DIFF and MERGE now work correctly by projecting base-side batches to match the target layout, comparing only common columns, and preserving extra-column data in output.

### Changes

- **`pkg/frontend/data_branch_types.go`**: Add `commonIdxes`, `tarOnlyIdxes`, and `baseColToTarIdx` fields to the internal branch table struct.
- **`pkg/frontend/data_branch.go`**: Replace `isSchemaEquivalent` with `checkSchemaCompatibility`; update `getTableStuff` to compute column index mappings.
- **`pkg/frontend/data_branch_hashdiff.go`**: Add `projectBaseBatchToTarget` helper; apply base batch projection before hashmap encoding; limit row comparison to common columns.
- **`pkg/frontend/data_branch_test.go`**: 6 unit tests for `checkSchemaCompatibility`.
- **`pkg/frontend/data_branch_hashdiff_test.go`**: 2 unit tests for projection and comparison.
- **`test/distributed/cases/git4data/branch/diff/diff_schema_evolve.sql`**: BVT for DIFF with schema evolution.
- **`test/distributed/cases/git4data/branch/merge/merge_schema_evolve.sql`**: BVT for MERGE with schema evolution.

PR #25882, opened 2026-07-20
feat(frontend): support schema evolution in data branch diff/merge

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24549

This PR continues the unfinished work from #24666 while preserving its original commits and authorship.

It supports schema evolution in DATA BRANCH DIFF/MERGE by matching columns with stable identity, separating physical and visible common-column indexes, projecting compatible historical batches, comparing common columns, preserving target-only data, and maintaining ALTER lineage ownership.

### Safety contract

DIFF/MERGE is allowed only when the endpoint schemas can be mapped without guessing:

| Schema relation | Behavior |
| --- | --- |
| Same stable column identity with equal type attributes and nullability | Treat as common; compare it and apply MERGE changes. |
| Pure reorder or lineage-proven rename | Treat as common using the resolved historical identity, not the mutable endpoint name. |
| Target-only visible column with an explicit primary key | Keep it in target-layout DIFF/OUTPUT, but do not write it into a base table that does not own the column. |
| Base-visible column missing from the target | Fail closed. |
| Same logical name reused by DROP/ADD, including rename/drop/add variants | Fail closed because the identities differ. |
| Common column with different type attributes, nullability, or primary-key identity | Fail closed. |
| Target-only column on a fake-PK table | Fail closed because the hidden key depends on the complete row layout. |

Historical batches are projected through the lineage-resolved layout. A target-only column whose historical representation changed is omitted from the old batch and reconstructed from the current endpoint instead of being cast or matched by name.

COPY ALTER preserves the replaced physical generation only while a live data branch or a covering snapshot/PITR owns it. Snapshot/PITR publication and COPY ALTER update the same bootstrap-created `mo_feature_registry` row before choosing/probing the history boundary, while explicitly preserving the row's metadata: pessimistic transactions serialize on the row lock, while optimistic transactions serialize through write-write conflict and whole-statement retry. Snapshot/PITR creation now returns commit conflicts instead of reporting false success. After a wait, ALTER probes owners in a fresh transaction so it does not reuse an old SI snapshot. If identity, ownership, or generation connectivity cannot be proven, the operation returns an error.

Both PITR execution locations follow this contract. The frontend path and `CREATE PITR ... INTERNAL` engine path cross the same stable barrier; the INTERNAL path refreshes a pessimistic-RC snapshot after waiting and re-resolves the current database/table ID before selecting its timestamp and publishing `mo_pitr`, rather than reusing the ID captured during planning.

Review fixes on the latest head:

- project reordered and target-only historical layouts without moving incompatible vectors;
- reject same-name DROP/ADD column redefinitions by `Seqnum`, while preserving pure `FIRST` / `AFTER` reorderings;
- retain ALTER snapshot/metadata pairs while table-, database-, or account-level historical owners still cover the component, including current-table-first deletion;
- reclaim deleted intermediate ALTER generations after the final logical or historical owner disappears;
- preserve explicit-transaction ALTER semantics for ordinary snapshot/PITR-covered tables;
- serialize snapshot/PITR owner publication with COPY ALTER even when the initial owner query is empty;
- serialize engine-executed INTERNAL PITR publication and refresh its target generation after waiting;
- rename the merge BVT tables by scenario and role so each case is readable in isolation.

## Validation

- `go test ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile ./pkg/vm/engine/test -count=1`
- full `go test -race` for `pkg/frontend/databranchutils`, `pkg/frontend`, and `pkg/sql/compile`, plus focused repeated race tests for lineage-owner publication
- `TestPreparePitrPublicationSerializesCopyAlter` and `TestResolveCurrentPitrObjectIDRefreshesPlannedTableID`, each with `-race -count=100`, plus production `TestPitrCases`
- authenticated cluster regressions `TestIssue26087ConcurrentDataBranchQuota` and `TestIssue26114CrossAccountBranchUsesTargetQuotaAndOwnership`
- `git diff upstream/main --check`
- full `mo-pr-preflight-review` and `mo-self-review`

The BVT fixtures are included for DIFF/MERGE identity redefinition and table/account/database historical-owner cleanup; they are executed by CI because this local worktree has no standalone `mo-tester` runner.


heni02 (approved):

gouhongshen (changes requested):
## Codex automated review

One blocking schema-evolution regression found. Focused frontend tests could not run because the worktree lacks cgo/libmo.dylib.

### P1 - Allow target-only historical type changes or reject them consistently (pkg/frontend/data_branch_hashdiff.go:2625)
`checkSchemaCompatibility` accepts a target-only column regardless of its type, so `t0(a,b)` and branch `t1(a,b,c varchar)` are compatible. However, after `t1` first adds `c int` and then modifies it to `varchar`, every DIFF/MERGE walks the old physical generation and this mapping rejects old `c int` against endpoint `c varchar` before even opening CollectChanges. The column is target-only and is deliberately omitted from MERGE apply (`commonVisibleIdxes`), so its historical representation cannot affect `t0`; nevertheless all operations fail, including when there were no writes to `c`. Skip/project such non-common historical columns as NULL, or make the endpoint compatibility check reject this evolution up front. Add an end-to-end regression covering add-target-only-column -> modify its type -> DIFF/MERGE.

VioletQwQ-0:
Addressed the blocking historical-generation schema mapping regression.

Changes:
- Thread the endpoint target-only column indexes into historical generation mapping.
- When a historical physical column has a different type from a target-only endpoint column, leave it unmapped so projection creates a NULL vector with the endpoint type; existing endpoint hydration restores the current value when the row still exists.
- Keep historical type drift for common columns and primary-key incompatibility fail-closed.
- Preserve compatible target-only mappings, source-only/rename behavior, and reordered columns.

Coverage:
- Added focused unit coverage for the old failure, typed-NULL projection and vector ownership, common-column rejection, compatible target-only columns, source-only/rename, and reorder mappings.
- Added DIFF BVT cases for no writes and historical target-only values across `ADD c INT -> MODIFY c VARCHAR`.
- Added a MERGE BVT case proving common-column changes apply while the historical target-only value is ignored by destination apply.

Validation:
- focused `go test -race ./pkg/frontend ... -count=1`
- `go vet ./pkg/frontend`
- `go build ./pkg/frontend`
- `git diff --check`
- full mo-self-review across correctness, concurrency, lifecycle, compatibility, failure modes, API closure, and Q1-Q3: zero open blockers

The BVT fixtures were not executed locally because this worktree has no mo-tester runner; the new-head CI will validate them.

aunjgr (approved):
Reviewed updated head c515a1456. The target-only historical type-drift blocker is closed: incompatible non-common columns are projected as NULL and hydrated from the endpoint, while common-column and primary-key drift remain rejected. DIFF and MERGE regressions cover both unchanged and updated rows. No remaining blockers found.

gouhongshen (changes requested):
## Codex automated review

The prior historical target-only type-drift blocker is fixed, but one blocking ALTER compatibility regression remains. GitHub checks pass on c515a145; local focused tests could not start because cgo/libmo.dylib is absent.

### P1 - Preserve explicit ALTER semantics when historical sources exist (pkg/sql/compile/alter.go:235)
This rejects every copy-based ALTER executed under BEGIN or autocommit=0 whenever prepareAlterDataBranchLineage finds a covering user snapshot/PITR. That search also matches account- and cluster-level history, so a single cluster snapshot can make explicit ALTER fail for ordinary, non-branch tables across all tenants. Before this patch AlterTableCopy had no such restriction, and existing DDL BVTs establish that ADD/DROP COLUMN inside explicit transactions is supported. Preserve the lineage transactionally without rejecting the existing ALTER workflow, or otherwise avoid applying this restriction to ordinary tables covered only by snapshots/PITR.

VioletQwQ-0:
Addressed the P1 explicit-transaction regression in d4ec9f818.

Root cause: prepareAlterDataBranchLineage performed the transaction restriction twice. The second unconditional check also rejected ordinary tables that had no live data-branch lineage and were covered only by a snapshot/PITR.

Changes:
- keep the explicit-transaction restriction for tables that actually participate in data-branch lineage;
- allow historical-source-only tables to preserve their ALTER lineage using the current transaction;
- add unit regressions for both snapshot and PITR history;
- add BVT coverage for BEGIN and autocommit=0 ALTERs on ordinary snapshot-covered tables.

Validation:
- go test ./pkg/frontend -count=1
- go test -race ./pkg/sql/compile -run "^(TestPrepareAlterDataBranchLineageAllowsHistoricalSourceTxn|TestValidateAlterDataBranchLineageTxn)$" -count=1
- go vet ./pkg/sql/compile ./pkg/frontend
- go build ./pkg/sql/compile ./pkg/frontend
- git diff --check

XuPeng-SH (changes requested):
Requesting changes on `d4ec9f818`. Two lifecycle blockers remain.

1. **[P1] Deleted intermediate ALTER generations are permanently excluded from compaction.**

   `ComputeAlterLineageCompactionPlan` skips every node with `meta.Deleted` (`pkg/frontend/databranchutils/branch_protect_snapshot.go`, around the per-component planning loop). A normal two-ALTER sequence creates exactly such a node:

   ```sql
   create snapshot s for table db t;
   alter table t add column c int; -- T2 -> T1
   alter table t add column d int; -- T3 -> T2; internal DROP marks T2 deleted
   drop snapshot s;
   ```

   After the last historical owner disappears, the plan contains T3 only. The T2 metadata row and `__mo_branch_T2` snapshot remain forever. Repeating covered ALTERs grows both catalogs linearly. The current tests only exercise one historical ALTER before dropping the source, and the inherited-owner test asserts snapshot cleanup but never asserts metadata cleanup.

   Please reclaim identity-matched deleted ALTER rows once the whole component has no logical owner or covering historical source. Add a unit test with `{T2 deleted, T3 live}` plus both edges, and a BVT that performs two covered ALTERs and verifies that both `mo_branch_metadata` and branch snapshots are empty after the last owner is dropped. Also cover dropping the current table before dropping the owner.

2. **[P1] Snapshot/PITR-triggered compaction is not transactional, so a mid-cleanup failure destroys the metadata/snapshot pair.**

   `doDropSnapshot`, `doDropPitr`, and `doAlterPitr` use `GetBackgroundExec` without `BEGIN`. That executor is an independent autocommit session, so the metadata `FOR UPDATE` read, owner/source reads, branch-snapshot DELETE, and metadata DELETE in `compactHistoricalAlterLineageWithBH` are separate transactions. For example, if the branch-snapshot DELETE commits and the following metadata DELETE fails, the metadata row loses the edge identity that the planner requires and can never be reclaimed. The user-visible snapshot/PITR mutation has already committed as well, so retrying it does not re-enter this cleanup. The `FOR UPDATE` lock also does not protect the later planning/deletion window.

   Please run the owner mutation, ownership validation, and paired snapshot/metadata deletion in one explicit transaction (with rollback on every error), or otherwise provide an equivalent atomic/revalidated cleanup protocol. Add a failure-injection test between the two DELETEs that proves both rows and the owner mutation roll back, plus a concurrent owner/lineage creation test for the plan-to-delete window.

The schema projection/hash-diff paths otherwise look consistent, and the focused UT/race/vet checks pass; those checks do not exercise either lifecycle sequence above.

gouhongshen (changes requested):
## Codex automated review

One blocking ALTER-lineage lifecycle leak remains. The prior transaction-atomicity concern is disproven at the current head because snapshot/PITR owner mutation and compaction share an explicit transaction.

### P1 - Reclaim deleted intermediate ALTER generations (pkg/frontend/databranchutils/branch_protect_snapshot.go:281)
The planner unconditionally skips `meta.Deleted` rows. After two snapshot-covered copy-based ALTERs, the second ALTER's internal DROP marks the intermediate generation deleted while its live successor keeps its protect snapshot. Once the user snapshot is dropped, compaction selects only the live successor, permanently leaving the intermediate metadata row and `__mo_branch_<tid>` snapshot. If the current table is dropped first, ordinary reclaim removes the snapshot but the deleted metadata row still becomes unreclaimable because later compaction requires a matching edge. Repeated covered ALTERs therefore grow catalog state linearly. Include identity-matched deleted ALTER rows when the entire component has no logical or historical owner, and cover both two-ALTER cleanup and current-table-before-owner deletion.

VioletQwQ-0:
Addressed the deleted intermediate ALTER-generation lifecycle blocker in `8df839ed5`.

Changes:
- Compaction now considers deleted `alter` / `alter:*` generations after their connected component has no live logical owner or covering historical source.
- Existing snapshot rows still must match `(child table ID, parent table ID, clone TS)` before deletion.
- Live metadata without a matching snapshot remains conservatively retained.
- Deleted metadata whose snapshot was already reclaimed by plain `DROP TABLE` remains reclaimable by its deterministic branch-snapshot name once the final historical owner disappears.

Coverage:
- Added the requested `{T2 deleted, T3 live}` unit case with both identity-matched edges.
- Added the current-table-first case where both snapshots are already absent, including retention while the table snapshot still owns the component and cleanup after owner deletion.
- Added frontend execution coverage for paired metadata/snapshot deletion without remaining edge rows.
- Added BVT flows for two covered ALTERs in both orders: owner drop before current-table drop, and current-table drop before owner drop. Both assert that `mo_branch_metadata` and internal branch snapshots reach zero.

Validation:
- `go test ./pkg/frontend/databranchutils ./pkg/frontend -count=1`
- focused `go test -race` for planner and frontend compaction paths, 10 consecutive runs
- `go vet ./pkg/frontend/databranchutils ./pkg/frontend`
- `go build ./pkg/frontend/databranchutils ./pkg/frontend`
- `git diff --check`
- full `mo-self-review`: correctness, concurrency, lifecycle, compatibility, failure modes, API closure, and Q1-Q3 all converged with zero open blockers

The transaction-atomicity concern required no code change: as the latest review confirmed, snapshot/PITR owner mutation and compaction already share the caller's explicit transaction. The distributed BVT fixture was not executed locally because this worktree has no `mo-tester`; the new-head CI will validate it.

XuPeng-SH (changes requested):
I still see one blocker on the latest head.

`dataBranchDeleteTable` now runs `compactHistoricalAlterLineageWithBH` immediately after `reclaimBranchSnapshotsWithBH` on the same DROP path (`pkg/frontend/data_branch.go:549-556`). The latest compaction fix correctly keeps current-table-first orphan metadata alive for a **table-level** owner because `historicalSourceOwnsComponent` can still match `source.ObjectID` against the component node IDs (`pkg/frontend/databranchutils/branch_protect_snapshot.go:203-207`). But **account** and **database** owners are matched only through `componentEdges` (`branch_protect_snapshot.go:190-201`), and those edges are exactly what plain DROP has already reclaimed before compaction runs.

So for a chain like `T1 -> T2 -> T3`, if the current table is dropped first while an account-level or database-level snapshot/PITR is still live, `componentEdges` is empty, `componentSources` stays empty (`branch_protect_snapshot.go:265-277`), and `ComputeAlterLineageCompactionPlan` immediately returns `T2/T3` for deletion (`branch_protect_snapshot.go:279-303`) even though the historical owner still exists. I reduced this to a pure planner case on the latest head: with deleted rows `{2<-1, 3<-2}`, no edges, and a live account-level source `{Level:"account", AccountName:"acc", OldestTS:50}`, the planner still returns `[2,3]`; the same logic applies to database-level owners.

Please preserve enough scope identity for orphaned deleted ALTER generations to remain owned by active account/database historical sources after current-table-first DROP, and add regressions for both owner scopes (snapshot or PITR) in that current-table-first path.

gouhongshen (changes requested):
## Codex automated review

Two blocking regressions found. Focused Go tests could not start because this worktree lacks the required CGo libraries/headers.

### P1 - Reject DROP/ADD redefinitions of a column name (pkg/frontend/data_branch.go:1212)
Compatibility now treats same-name, same-type columns as common without checking Seqnum, although Seqnum is the stable column identity across ALTERs (the previous isSchemaEquivalent check rejected differing Seqnum). Thus `update t1 set b=5; alter table t1 drop column b; alter table t1 add column b int default 0` leaves t1's current `b` as 0, but DIFF accepts t1 against unchanged t0. The old generation's `[a,b]` mapping is identity, so no historical projection/hydration occurs and its stale `b=5` update is emitted; MERGE then writes 5 into t0. Reject differing Seqnum for common columns (or model the redefinition as an unmapped column) and add DIFF/MERGE coverage.

### P1 - Keep account/database historical owners after branch snapshots are reclaimed (pkg/frontend/databranchutils/branch_protect_snapshot.go:190)
After `DATA BRANCH DELETE TABLE` drops the current generation, reclamation deletes its branch snapshots before compaction. For an ALTER chain with all nodes deleted, this leaves `componentEdges` empty. Account- and database-level sources are recognized only by iterating those edges, so an active account/database snapshot or PITR is no longer considered an owner and compaction deletes the `alter:*` metadata immediately. This destroys lineage still covered by the live historical source. Preserve scope identity for orphaned components, and cover current-table-first deletion with both account- and database-level snapshot/PITR owners.

XuPeng-SH (changes requested):
Requesting changes on 8df839ed589b8f37b9e51e0392dce95ad65086ae. Two blocking correctness regressions remain.\n\n1. P1 - Reject DROP/ADD redefinitions of the same column name.\n\ncheckSchemaCompatibility (pkg/frontend/data_branch.go:1211-1233) treats same-name, same-type columns as common without checking Seqnum/physical identity, and dataBranchSourceColToTargetIdx repeats the name-based mapping for historical generations (pkg/frontend/data_branch_hashdiff.go:2620-2632). A review-only unit case with base b Seqnum=1 and target b Seqnum=2 returns nil instead of rejecting the pair. This is the DROP b; ADD b case: the old generation's b values are then projected into the newly created logical b, so DIFF can emit stale values and MERGE can write them into the base. Preserve pure FIRST/AFTER reordering, but reject or explicitly unmap same-name columns whose stable identity changed. Add DIFF/MERGE coverage for update old b -> drop b -> add b with the same type.\n\n2. P1 - Retain orphaned ALTER lineage for live account/database historical owners.\n\nDATA BRANCH DELETE first reclaims branch snapshots and then compacts lineage (pkg/frontend/data_branch.go:553-556). Once all component rows are deleted, those snapshots were the only HistoricalLineageEdge records carrying account/database scope. historicalSourceOwnsComponent matches account and database owners only through edges (pkg/frontend/databranchutils/branch_protect_snapshot.go:190-201), so with no edges componentSources is empty and the planner deletes the metadata. A review-only planner case with deleted {2<-1, 3<-2}, edges=nil, and a live account source {AccountName:"acc", OldestTS:50} returns [2,3]; database scope behaves identically. That destroys lineage still covered by the owner. Preserve scope identity independently of the reclaimed edge (or otherwise keep enough edge state through compaction), and cover current-table-first deletion for both account- and database-level snapshot/PITR owners.\n\nThe PR's package tests, focused race x10, go vet, go build, and GitHub CI pass; the two failing review-only cases demonstrate gaps in the current matrix.

VioletQwQ-0:
Addressed all blocking review feedback on the current review head and resolved the `main` conflicts in `08cd75df6`.

Changes:
- Reject same-name endpoint columns when their stable `Seqnum` identity differs, while preserving pure `FIRST` / `AFTER` reorderings whose identity is unchanged.
- Enforce the same identity rule while mapping historical generations: common-column identity drift fails closed, while target-only drift is explicitly projected as `NULL` for endpoint hydration.
- Preserve ALTER-owned branch snapshots while a table-, database-, or account-level historical source still owns the lineage component. This keeps the scope evidence available after current-table-first `DROP TABLE` and `DATA BRANCH DELETE TABLE`; the snapshot/metadata pair becomes reclaimable after the last owner is removed.
- Merged the latest `upstream/main` without rebasing or rewriting the inherited history, preserving the new DATA BRANCH `OUTPUT AS` stable-identity and snapshot-range semantics.

Coverage:
- Unit regressions for endpoint and historical `Seqnum` drift, pure reorder compatibility, target-only unmapping, account/database owner retention, owner removal, LCA identity mapping, and snapshot-range fallback.
- DIFF and MERGE BVT regressions for `UPDATE old b -> DROP b -> ADD b`.
- Current-table-first BVT coverage for table/account/database historical owners, including ordinary `DROP TABLE` and `DATA BRANCH DELETE TABLE`.

Validation on `08cd75df6`:
- `go test ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile ./pkg/vm/engine/test -count=1`
- focused `go test -race` for identity mapping and historical-owner lifecycle paths
- `go vet ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile ./pkg/vm/engine/test`
- `go build ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile ./pkg/vm/engine/test`
- `git diff upstream/main --check`
- full `mo-pr-preflight-review` and `mo-self-review`: zero open merge blockers

The new BVT fixtures were not executed locally because this worktree has no standalone `mo-tester` runner; the new-head CI will validate them.

XuPeng-SH (changes requested):
Requesting changes on 08cd75df68132f1edadc631404e7e80c30bfa76f. One blocking historical-schema path remains.

[P1] Project incompatible target-only columns out of LCA probes — pkg/frontend/data_branch_hashdiff.go:114

The latest fix correctly maps an incompatible target-only historical column to -1 in dataBranchSourceColToTargetIdx, but lcaProbeColumnLayout independently selects every LCA column whose ColId and Seqnum match the target, without checking its logical type or targetOnlyIdxes. handleDelsOnLCA then appends that old LCA vector into the endpoint-typed dBat; appendLCAProbeValue rejects every non-ENUM type mismatch with unexpected LCA probe type conversion.

This is reachable when the fork point contains c INT, one descendant changes c to VARCHAR, and the sibling drops c. The endpoint compatibility check accepts c as target-only, but an UPDATE of an LCA row forces the probe to read c INT and append it into target c VARCHAR. A review-only unit with LCA c INT and target c VARCHAR sharing the same ColId/Seqnum expected the probe layout to contain only a; the current head returned [a,c]. DIFF/MERGE therefore fail before comparison even though c is excluded from comparison and MERGE apply.

Please thread the endpoint target-only/type-compatibility rule into LCA layout construction so such columns are omitted and reconstructed as NULL, matching the historical CollectChanges projection. Add DIFF/MERGE coverage with a common ancestor containing c INT, one endpoint MODIFY c VARCHAR, the sibling DROP c, and an updated ancestor row that exercises handleDelsOnLCA.

The prior Seqnum-redefinition and account/database-owner blockers are closed. The four affected package test suites and focused race x10 pass; the failing review-only case exposes the missing LCA branch in the current matrix.

VioletQwQ-0:
Fixed in 7a0fb11b0. The LCA probe now excludes endpoint target-only columns on both the SQL and reader paths, including the no-column-metadata fallback. Common columns still fail closed on identity/type mismatches, and full-layout reader results now resolve ENUM metadata by target index. This also prevents hidden RowID identity collisions from being appended as user column values, which was the source of the ROWID-to-INT/VARCHAR CI failures.

Validation:
- go test ./pkg/frontend -count=1
- go test -race ./pkg/frontend -run "TestHashDiff_HasLCAUpdateIgnoresTargetOnlyColumns|TestLCAProbeColumnLayout|TestLCAProbeResultTargetIndexes|TestHandleDelsOnLCA_SQLPaths" -count=1
- git diff --check

Added DIFF and MERGE BVT regressions for the common-ancestor c INT -> target c VARCHAR / sibling DROP c path. CI is rerunning on the new head.

VioletQwQ-0:
Updated head: `5b57d3fa0`.

I rechecked the complete PR diff and all previous change-request clusters. The earlier target-only projection, explicit ALTER semantics, deleted-generation compaction, owner-scope cleanup, DROP/ADD identity, and LCA target-only type fixes remain intact.

This update fixes the remaining composed historical-schema case found during preflight:
- treat a rebuilt hidden `__mo_cpkey_col` as the same derived composite key only when its ordered component names, Seqnums, types, and nullability still match;
- prefer that canonical identity over an accidental old RowID numeric-ID collision in the LCA path;
- quote reserved/spaced composite-key names in the LCA VALUES join;
- reject ALTER lineage timestamp overflow before `UpdateSnapshot`;
- verify `doDropSnapshot` rolls back if branch-snapshot compaction fails after the user snapshot delete.

Validation:
- `go test ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile -count=1`
- focused `-race` tests x10 for composite-key mapping, drop rollback, and timestamp overflow
- `go vet` on the same packages
- isolated live MO DIFF/MERGE with quoted composite PK, pre-ALTER UPDATE, COPY ALTER, and post-ALTER INSERT
- `git diff --check`
- full 255-symbol preflight + mo-self-review: PASS, no PENDING findings

Please re-review this head.

XuPeng-SH (approved):
Approved on exact head `5b57d3fa0ec18283cb15397262e5b163d04c00a6`. No blocking findings remain.

I revalidated the complete PR diff and every prior P1 cluster: target-only historical type projection, explicit ALTER transaction semantics, deleted intermediate-generation reclamation, account/database historical ownership after current-table-first deletion, DROP/ADD Seqnum identity, and incompatible target-only LCA probes. The latest derived composite-key fix is fail-closed on ordered component names, component Seqnums/types/nullability, handles accidental RowID identity collisions, and quotes PK identifiers in both VALUES aliases and join expressions.

| Audit area | Q1 trigger/ownership | Q2 cleanup/failure path | Q3 completion/reuse |
|---|---|---|---|
| LCA SQL/reader probe | Input PK batch and snapshot reader callbacks have explicit ownership; concurrent callbacks are serialized | temporary/filter/output batches close on success and error; target-only columns are reconstructed as NULL | full/projected layouts map by target index; duplicate/missing keys complete deterministically |
| Historical generation projection | vectors are transferred once from source batches | hydration/probe errors clean projected data and tombstones | Seqnum/type mismatches fail closed except the verified derived composite key identity |
| Snapshot/PITR lineage compaction | owner mutation and compaction share one explicit transaction | any compaction DELETE failure rolls back the owner mutation and paired metadata work | deleted generations remain retained while owned and are reclaimed after the final owner disappears |

Validation on this head:
- `go test ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile -count=1`
- focused `go test -race` for LCA layout/reader/SQL, rebuilt composite PK mapping, DROP snapshot rollback, and timestamp overflow
- `go vet ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile`

GitHub SCA and completed checks are green; UT/BVT/coverage jobs were still running with no failures at review submission.

VioletQwQ-0:
Fixed the current-head CI failures in bb8fb55af.

- Preserve a renamed column by its stable nonzero (ColId, Seqnum) when the endpoint no longer carries OriginName.
- Keep DROP/ADD fail-closed by validating every historical generation; added a colliding-identity regression.
- Do not treat unset (0,0) test/catalog identity as a match.
- Updated diff_17 to expect the newer precise type-attribute error.

Validation:
- go test ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile ./pkg/vm/engine/test -count=1
- focused pkg/frontend go test -race -count=3
- go vet on the four affected packages
- git diff --check
- exact-head preflight PASS on bb8fb55af

gouhongshen (changes requested):
## Codex automated review

One blocking regression remains in the newly added stable-identity rename path.

### P1 - Preserve identity-only renames when probing the LCA (pkg/frontend/data_branch_hashdiff.go:125)
`dataBranchLineageEndpointColumns` now accepts a rename without `OriginName` when `(ColId, Seqnum)` proves both endpoints reach the same LCA column (for example LCA/base `b`, target `bb`). However, `lcaProbeColumnLayout` still resolves `bb` through `dataBranchColumnDefByLogicalName`, which only checks names and `OriginName`; it therefore omits the LCA `b` column. Both SQL and reader probes use this layout, and `handleDelsOnLCA` fills the omitted, common target column with NULL. An update/delete of an ancestor row can consequently be compared against a fabricated NULL value and produce an incorrect DIFF/conflict/MERGE result. Thread the lineage-resolved endpoint mapping into LCA layout construction, and cover a no-`OriginName` rename with an updated ancestor row.

VioletQwQ-0:
Fixed in `86f1b9542`.

- Threaded separate lineage-resolved target-to-LCA and target-to-base column-name mappings into both SQL and reader probes. Target hydration uses the target layout; base hydration uses the resolved base endpoint layout.
- Added a regression for a rename without `OriginName` followed by an update to an inherited row.
- Restored early primary-key compatibility validation so the dropped-PK BVT keeps the documented schema error.
- Deferred ALTER identity-snapshot deletion to the historical-owner-aware compactor, preserving account/database scope until the owner is dropped.

Validation:

- `go test ./pkg/frontend/databranchutils ./pkg/frontend ./pkg/sql/compile`
- focused `go test -race ./pkg/frontend ...`
- `go vet ./pkg/frontend/... ./pkg/sql/compile`
- `git diff --check`
- exact-head preflight: PASS

gouhongshen (changes requested):
## Codex automated review

Two blocking lifecycle/concurrency gaps remain: optimistic clone/ALTER races can create unusable branches, and rolling PITR expiry does not trigger lineage cleanup.

### P1 - Serialize optimistic DATA BRANCH creation with copy-based ALTER (pkg/frontend/clone.go:106)
This path explicitly skips source-catalog locking for optimistic transactions. A clone can therefore snapshot the old physical table after ALTER's initial participation check but commit its branch metadata after ALTER's final lineage check and copy-and-swap. No ALTER edge connects the new current generation to that branch; the guard in data_branch.go then makes every later DIFF/MERGE fail with “lineage metadata is missing.” Thus a successful CREATE DATA BRANCH can become permanently unusable. Implement an optimistic validation/retry protocol that atomically verifies the source generation, or reject this mode before creating the branch.

### P1 - Reclaim lineage when a rolling PITR window expires without another DDL (pkg/sql/compile/alter.go:449)
PITR ownership changes merely as time advances: once `OldestTS > meta.CloneTS`, the ALTER snapshot is reclaimable. However, compaction is only invoked by later ALTER/DROP, explicit PITR mutation, snapshot deletion, or branch deletion; no timer/background expiry path invokes it. If the workload performs one PITR-covered copy ALTER and then only DML, the `kind='branch'` snapshot and metadata remain indefinitely after the configured PITR window passes, pinning the old physical generation and defeating retention. Repeated tables can accumulate unbounded retained storage. Integrate this cleanup with periodic PITR/GC expiry processing or otherwise make protection expire automatically.

VioletQwQ-0:
Addressed both findings on `597fa0cba`:

- DATA BRANCH creation now fails before opening the clone transaction when the caller is in optimistic mode. Both table/database entry points are guarded, and the source-lock path repeats the check so alternate callers cannot bypass it.
- Rolling PITR expiry no longer depends on a later DDL. A five-minute system cron task runs the existing lineage compaction logic in one transaction, locking metadata before atomically deleting the managed snapshot and ALTER metadata. The task is registered idempotently for upgrades and included in fresh-cluster bootstrap.
- The UT/Coverage failure was the stale lifecycle expectation: ordinary reclaim leaves the final ALTER identity snapshot for the historical compactor. The regression now exercises that terminal compaction.

Validation: full `pkg/frontend`, `pkg/frontend/databranchutils`, `pkg/sql/compile`, `pkg/cnservice`, `pkg/predefine`, and `pkg/vm/engine/test`; focused `-race`; affected-package `go vet`; protobuf regeneration; `git diff --check`; exact-head preflight and self-review.

gouhongshen (changes requested):
## Codex automated review

One blocking timestamp-source race remains despite the new current-source lock.

### P1 - Serialize timestamp-based data branch creation with ALTER (pkg/frontend/clone.go:461)
The new shared source lock is skipped whenever snapshot.TS is set. That includes `{MO_TS=...}` / timestamp sources (ResolveTsHint constructs a Snapshot with TS), not just named snapshots. Thus a timestamp-based CREATE DATA BRANCH can clone old generation A while a concurrent COPY ALTER acquires the exclusive catalog lock, sees neither branch_metadata nor a `kind='user'` snapshot/PITR, and omits the A→C lineage edge. Once the create commits B→A, B and current C are disconnected. DIFF/MERGE then takes the no-LCA path, whose endpoint resolver matches only name/type; after a same-name drop/re-add, it can accept and merge B's old column values into C's unrelated replacement column. Acquire the same shared lock for timestamp sources (or otherwise atomically publish the lineage before ALTER can probe), and add a concurrent `{MO_TS=...}` CREATE/ALTER regression test.

VioletQwQ-0:
Addressed the timestamp-source publication race in `21cc25707`, with the deterministic concurrency regression completed in `c04f9e350`.

- Clone source classification now distinguishes durable named snapshots (`ExtraInfo != nil`) from timestamp hints (`TS`/`Tenant` only).
- Live and `{MO_TS=...}` table sources acquire the same shared `mo_tables` row lock used to serialize with COPY ALTER; named snapshots retain the existing skip.
- Database/account source-set cloning uses the same classification, and nested table clones do not reacquire locks already held in sorted order.
- Added a deterministic shared/exclusive publication-gate regression proving a timestamp clone waits while ALTER holds the exclusive lock and resumes after release.

Validation:
- focused tests and `go test -race ... -count=10`
- `go vet ./pkg/frontend`
- `git diff --check`
- exact-head preflight PASS on `c04f9e3501330b67cf419540fac210b769059f4b`

gouhongshen (changes requested):
## Codex automated review

Two blocking correctness gaps remain on c04f9e350. Prior review findings were not repeated where the current code disproves or closes them. Focused tests could not run because this environment has no Go executable.

### P1 - Revalidate timestamp sources when ALTER acquires the publication lock first (pkg/frontend/clone.go:478)
The timestamp is resolved before this shared lock is acquired, and the code continues unconditionally after waiting. If COPY ALTER acquires the exclusive lock first, it can replace A with C and commit without seeing the unpublished branch; the clone then resumes and restores its already-selected historical A, publishing only B→A. No A→C lineage exists, so the successful branch later fails DIFF/MERGE as missing legacy ALTER lineage (or reaches the unrelated/no-LCA path). The new concurrency test explicitly exercises ALTER-first ordering but only asserts that the clone resumes; it does not verify generation identity or lineage closure. After acquiring the lock, reject/retry when the selected timestamp generation is no longer connected to the current source, or publish through an atomic protocol that closes A→C.

### P1 - Detect same-statement DROP/ADD column redefinitions (pkg/frontend/data_branch.go:1889)
Redefinition detection only notices that a name disappears in an intermediate physical generation. MatrixOne supports multi-clause COPY ALTER, so `ALTER TABLE t1 DROP COLUMN b, ADD COLUMN b INT DEFAULT 0` creates one old→new edge where both endpoint schemas contain `b`; this name lookup therefore treats the unrelated columns as continuous. If `b` was updated before the ALTER, historical mapping also maps old `b` into new `b`, allowing DIFF/MERGE to replay the stale value. Existing regressions use two separate ALTER statements and therefore do not cover this case. Preserve operation-level identity across a COPY ALTER or otherwise reject a same-name replacement even when no intermediate schema exists.

pkg/frontend/clone.go
```diff
@@ -317,6 +475,17 @@ func handleCloneTable(
 		err = moerr.NewInternalErrorNoCtxf("only sys can clone table to another account")
 		return
 	}
+	if err = withDataBranchCloneSourceLock(snapshot, func() error {
+		return lockDataBranchCloneSource(
```
XuPeng-SH, line 721: [P1] Make historical timestamp publication close the generation DAG. `getOpAndToAccountId` resolves the timestamp before this shared lock, and the lock does not revalidate or connect the physical source generation afterwards. If COPY ALTER wins the exclusive lock first, it sees no unpublished branch and replaces A with C without recording A→C; the waiting clone then restores B from historical A and only publishes B→A. `decideLCABranchTSFromBranchDAG` therefore finds no LCA between current C and B and treats them as unrelated tables, producing incorrect DIFF/MERGE semantics and a full-history scan. The same gap exists for an MO_TS clone started just after an untracked ALTER. Please either reject/retry when the resolved historical generation is disconnected or atomically publish/recover the missing ALTER edge, and make the regression assert physical IDs plus the final DAG/LCA—not only that the clone waited.

pkg/frontend/data_branch.go
```diff
@@ -1459,13 +1824,189 @@ func constructChangeHandle(
...
+	pathDefs []*plan.TableDef,
+	lineageOnly []bool,
+	endpointCol *plan.ColDef,
+) (reachesLCA bool, lcaCol *plan.ColDef, redefined bool) {
+	if len(pathDefs) == 0 || len(pathDefs) != len(lineageOnly) || endpointCol == nil {
+		return false, nil, false
+	}
+	current := endpointCol
+	for i := len(pathDefs) - 2; i >= 0; i-- {
+		previous := dataBranchColumnDefByName(pathDefs[i], current.Name)
```
XuPeng-SH, line 1927: [P1] Track a same-statement DROP/ADD as a column-identity break. MatrixOne supports `ALTER TABLE t DROP COLUMN b, ADD COLUMN b INT DEFAULT 0`; the copy planner applies both actions to one CopyTableDef and publishes only one old→new physical generation. Both endpoint definitions therefore contain `b`, so this unconditional by-name lookup reports continuity even though the new `b` must not inherit the old one. The existing regression uses two ALTER statements (old b → absent → new b), which is why it passes. As written, DIFF/MERGE can pair unrelated old/new values. Please persist action-level column continuity (or conservatively reject this ALTER while lineage exists) and add the single-statement case.

XuPeng-SH (changes requested):
Requesting changes for two P1 correctness gaps in schema-evolution lineage. The current CI and unit/race tests pass, but the concurrency test only checks wait/resume and the DROP/ADD test uses two separate ALTER generations, so neither failure below is covered.

pkg/frontend/data_branch.go
VioletQwQ-0: Fixed in 36723ddec. The COPY planner now carries original DROP column identity and ADD column name through plan deep-copy. When prepareAlterDataBranchLineage reports active lineage, compile rejects a same-statement DROP/ADD of the same logical name before the copy ALTER; ordinary tables and non-replacement DROP/ADD remain unchanged. Planner/deep-copy and compile regressions cover the single-statement case. Full four-package tests, focused x10/race, vet, diff check and exact-head preflight pass.

pkg/frontend/clone.go
VioletQwQ-0: Fixed in 36723ddec. After acquiring the shared source-row lock, timestamp/MO_TS clones now advance the RC snapshot, re-resolve both the selected historical physical table and the current physical table, and require a DAG LCA before ExecRestore or branch metadata publication. If ALTER won first and did not preserve A→C, CREATE DATA BRANCH fails closed. The altered concurrency regression now crosses the ALTER-first barrier and asserts rejection; connected A→C and same-generation cases are covered. Full frontend/plan/compile/databranchutils tests and focused race x10 pass.

VioletQwQ-0:
Addressed both current blocking review clusters on exact head `36723ddec0` (reported independently by gouhongshen and XuPeng-SH):

- Timestamp/MO_TS clone publication now revalidates after acquiring the shared source-row lock: it advances the RC snapshot, re-resolves the historical and current physical generations, and requires a DAG LCA before restore or metadata publication. ALTER-first disconnected generations fail closed; connected and same-generation cases remain valid.
- A same-statement `DROP b, ADD b` COPY ALTER now carries action-level DROP/ADD identity through planner deep-copy and is rejected only while data-branch lineage is active. Ordinary tables and non-replacement multi-clause ALTERs remain unchanged.

Validation on this exact head: full frontend/databranchutils/plan/compile tests; focused x10 and race x10 regressions; vet; `git diff --check`; exact-head preflight and complete self-review PASS. Both XuPeng-SH inline threads were answered and resolved. Re-requesting both reviewers on this head.

pkg/frontend/clone.go
```diff
@@ -317,6 +559,37 @@ func handleCloneTable(
...
+			reqCtx,
+			ses,
+			fromAccountId,
+			stmt.SrcTable.SchemaName.String(),
+			stmt.SrcTable.ObjectName.String(),
+		)
+	}); err != nil {
+		return
+	}
+	if isTimestampDataBranchCloneSource(snapshot) {
```
XuPeng-SH: [P1] Scope this validation to DATA BRANCH creation. `handleCloneTable` is also the normal `CloneTable` path (`self_handle.go`) and those callers do not set `dataBranchCloneLockCtxKey`; snapshot restore also emits `CREATE TABLE ... CLONE ... {MO_TS=...}`. After an ordinary copy ALTER from table ID A to C, with no branch/history owner and therefore no A→C metadata edge, a previously valid normal timestamp clone from pre-ALTER A now reaches this unconditional block and fails with the data-branch “generation is not connected” error. The same can break restore. Guard revalidation with the explicit data-branch mode/context (not just `snapshot.TS != nil`) and add ordinary timestamp-clone-after-ALTER plus restore regressions.

pkg/frontend/clone.go
```diff
@@ -70,6 +79,239 @@ type cloneReceipt struct {
...
+			if err != nil {
+				return 0, err
+			}
+			if tableDef == nil {
+				return 0, moerr.NewNoSuchTable(sourceCtx, databaseName, tableName)
+			}
+			return tableDef.TblId, nil
+		},
+		func() (*databranchutils.DataBranchDAG, error) {
+			return constructBranchDAG(ctx, ses, bh)
```
XuPeng-SH: [P1] Make DAG revalidation atomic with branch publication. This loader uses `constructBranchDAG`, whose metadata query is a plain SELECT, while scheduled/front-end lineage compaction locks eligible rows `FOR UPDATE` and deletes them. A timestamp clone can read C→A, accept A, then concurrent GC can delete C→A and its protect snapshot before this transaction publishes B→A; because the transactions touch different branch-metadata rows, no conflict guarantees rollback. The committed graph is then disconnected again. Lock the lineage rows used for validation through metadata/snapshot publication, or atomically register the new durable owner, and cover the revalidation-vs-GC interleaving deterministically.

XuPeng-SH (changes requested):
Requesting changes on the latest head. The same-statement DROP+ADD replacement is now rejected for data-branch lineages, which closes the previous identity-reuse issue. However, the timestamp-race fix introduces one ordinary CLONE/restore compatibility regression and still leaves revalidation non-atomic with lineage GC. Both can affect correctness, so this is not safe to approve yet.

aptend (changes requested):
[P1] Track same-statement column replacement by stable identity, not raw action names

The new guard can still be bypassed by a multi-clause COPY ALTER such as:

```sql
ALTER TABLE t
  RENAME COLUMN b TO tmp,
  DROP COLUMN tmp,
  ADD COLUMN b INT DEFAULT 0;
```

`buildAlterTableCopy` resolves the recorded drop against the original `tableDef` by the action name. After the rename, `tmp` exists only in the evolving `CopyTableDef`, so `DROP tmp` produces no `DropColumn` action. The detector then sees only `ADD b` and does not reject the replacement. The lineage walk still matches the old and new `b` by name, so DIFF/MERGE can project an old branch value into the semantically new column.

Please track the dropped/added column through the evolving schema using stable identity (including later renames), or validate the final old-to-new identity mapping. Add DIFF/MERGE coverage for `rename -> drop -> add` and `drop -> add -> rename` variants.

gouhongshen (changes requested):
## Codex automated review

The latest changes close the previously reported direct DROP+ADD case and post-lock generation re-resolution, but four blocking compatibility/correctness gaps remain.

### P1 - Scope timestamp lineage validation to DATA BRANCH clones (pkg/frontend/clone.go:573)
This block is keyed only on the timestamp snapshot, whereas the preceding lock correctly checks dataBranchCloneLockCtxKey. It therefore also runs for ordinary CREATE TABLE ... CLONE ... {MO_TS=...} and restore-generated clones. After an ordinary copy ALTER A→C with no historical owner, no A→C branch edge is retained; a still-valid pre-ALTER timestamp resolves A while current resolves C, so normal clone/restore now fails with “generation is not connected”. Gate this revalidation on explicit data-branch mode.

### P1 - Hold the validated lineage through branch publication (pkg/frontend/clone.go:169)
constructBranchDAG performs a plain SELECT and may use an independent internal executor, so the validated rows are not locked by the transaction that later publishes branch metadata. A clone can read C→A and accept A, concurrent lineage GC can delete C→A and its snapshot, and then the clone can insert B→A and its distinct snapshot without a write conflict. The committed branch B is disconnected from current C again. Lock the required lineage rows in the publication transaction or atomically register the new owner.

### P1 - Detect rename-then-name-reuse replacements in one ALTER (pkg/sql/compile/alter.go:685)
The new guard intersects only explicit DropColumn and AddColumn actions. ALTER TABLE t RENAME COLUMN b TO old_b, ADD COLUMN b INT DEFAULT 0 records ADD b but no drop, so it passes with active lineage. Across the resulting lineage-only edge, dataBranchColumnReachesLCA matches the new b to ancestor b by name while treating old_b as target-only; DIFF/MERGE can therefore compare or apply the replacement column and historical projection can move old b values into it. Track rename/change identity transitions as well as drops.

### P1 - Allow OUTPUT AS to materialize target-only columns (pkg/frontend/data_branch_output.go:420)
Schema reconciliation now permits target-only columns, and an unqualified DIFF includes them in its target-layout output. OUTPUT AS defaults to all target-visible columns, but createSQL maps every projected target column back to the base by ColId/Seqnum. A newly added target column has no base identity, so the otherwise valid schema-evolved DIFF fails before creating the output table. Build target-only definitions from the target schema while retaining the mapped base definition for common/renamed columns.

pkg/frontend/clone.go
VioletQwQ-0: Fixed in f47a3571a. Timestamp revalidation now loads mo_branch_metadata with SELECT ... FOR UPDATE using the same BackgroundExec transaction that subsequently inserts B-to-A metadata and the branch-protect snapshot. The accepted lineage rows therefore remain locked until the whole branch transaction commits or rolls back, serializing this path with both frontend and scheduled compaction. The locked-DAG SQL and a deterministic validation-vs-GC publication interleaving are covered, including focused race x10.

pkg/frontend/clone.go
VioletQwQ-0: Fixed in f47a3571a. Timestamp generation revalidation is now gated by the explicit DATA BRANCH context in addition to the timestamp-hint source kind. Ordinary CLONE/restore callers do not set that context, so they retain the existing timestamp restore behavior; named snapshots remain exempt as before. TestShouldRevalidateTimestampDataBranchCloneSource covers ordinary context, explicit DATA BRANCH context, and named snapshots. Full affected-package tests and focused race x10 pass.

VioletQwQ-0:
Addressed the latest exact-head CRs in `f47a3571a`.

For aptend’s multi-clause identity case, the lineage guard no longer depends on recorded action names. It compares each original visible logical name against the final `CopyTableDef` using stable `(ColId, Seqnum)` identity. This rejects direct `DROP b, ADD b`, `RENAME b TO tmp, DROP tmp, ADD b`, and `DROP b, ADD tmp, RENAME tmp TO b`, while preserving ordinary MODIFY, pure RENAME/reorder, and non-replacement DROP/ADD behavior.

For XuPeng-SH’s two timestamp findings, revalidation is now limited to explicit DATA BRANCH context so normal timestamp CLONE/restore is unchanged, and accepted DAG rows are read `FOR UPDATE` in the same transaction that publishes branch metadata and its protect snapshot, preventing concurrent lineage GC from deleting the validated path before publication.

Validation on exact head: full frontend/databranchutils/plan/compile tests; focused race x10; planner identity matrix x10; vet; `git diff --check`; CR Learning Gate; exact-head preflight validator and complete self-review PASS. XuPeng-SH’s two threads were answered and are being resolved; re-requesting aptend, XuPeng-SH, and gouhongshen on this head.

VioletQwQ-0:
Closed the remaining item from gouhongshen’s `36723ddec` formal CR in `b93a70ed5`.

The first three findings are covered by `f47a3571a` (DATA BRANCH-only timestamp validation, lineage `FOR UPDATE` through publication, and final-schema stable identity covering rename/name-reuse). The additional OUTPUT AS blocker is now fixed as follows:

- projected columns explicitly classified in `tarOnlyIdxes` use their target-side type when no base stable identity exists;
- those output columns are nullable because base-side DIFF rows legitimately project target-only values as NULL;
- common and renamed columns still require a stable base `(ColId, Seqnum)` mapping and remain fail-closed if it is missing.

Validation on exact head `b93a70ed5`: target-only OUTPUT AS regression x10; full frontend/databranchutils/plan/compile tests; vet; `git diff --check`; CR Learning Gate; refreshed exact-head preflight validator and complete self-review PASS. Re-requesting gouhongshen on the complete current head.

gouhongshen (changes requested):
## Codex automated review

Two concurrency races can leave data-branch lineage disconnected or delete history needed by a newly created PITR.

### P1 - Revalidate timestamp sources after database-clone locks (pkg/frontend/clone.go:777)
Database cloning resolves its source before acquiring the sorted shared locks at lines 721-725. Nested table clones then set dataBranchCloneLockCtxKey to false here; that disables both duplicate locking and shouldRevalidateTimestampDataBranchCloneSource. If COPY ALTER publishes a new generation between source collection and lock acquisition, the nested clone still restores the old generation at the MO_TS and updateBranchMetaTable records that old ID as the parent. When no prior branch or historical owner exists, ALTER creates no connecting lineage edge, so the new branch is published against a disconnected generation and later DIFF/MERGE cannot safely establish its LCA. Revalidate the complete database source set once after all locks are held, while retaining the no-reacquire behavior for nested clones.

### P1 - Synchronize PITR creation with lineage compaction (pkg/sql/compile/alter.go:367)
The compaction source query reads active PITR rows without locking them, while compaction only locks mo_branch_metadata. CREATE/activate PITR does not acquire that metadata lock, so a PITR can become active after this read but before the subsequent snapshot and metadata deletes. If its retention lower bound covers an ALTER generation, the stale plan still deletes that generation and its branch snapshot, making the newly created PITR unable to restore it. The source set must be synchronized with compaction (or revalidated while holding the deletion lock) before deleting lineage.

VioletQwQ-0:
已处理当前 head `7a3e913c1` 上的两项 CR，推送为 `c35f1cf5c`：

1. 数据库级 `{MO_TS=...}` DATA BRANCH 在拿到全部源表共享锁后，会先推进 RC snapshot，再逐个重校验所有非 view 源表的 selected/current generation；nested table clone 仍不重复拿锁，但不再跳过整库重校验。任何断开的 generation 会在创建目标数据库或 restore 之前失败。
2. CREATE PITR 事务现在会先获取与 lineage compaction 相同的 `mo_branch_metadata FOR UPDATE` 行锁，并持有到 PITR publication commit/rollback。这样 compaction 要么看到新 PITR owner，要么先完成删除后 PITR 才线性化创建，不再存在 owner read 与 delete 之间插入 PITR 的窗口。

验证：

- `go test ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile -count=1`
- 相关并发回归 `go test -race ... -count=10`
- `go vet ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile`
- `git diff --check`
- exact-head preflight PASS：`c35f1cf5c06abca8aebbc9abe90fa1df159e90f4`

已重新请求所有仍 blocking 的 reviewer 一起 review 当前 head。

gouhongshen (changes requested):
## Codex automated review

The latest database timestamp-source and PITR-publication fixes close the two re-review findings, but named-snapshot branch creation still has a blocking lineage-publication race.

### P1 - Make named-snapshot lineage ownership atomic with snapshot deletion (pkg/frontend/clone.go:620)
Revalidation and the `mo_branch_metadata FOR UPDATE` lock only run for timestamp hints; named snapshots (`ExtraInfo != nil`) skip both paths. Consider snapshot S created on physical generation A, followed by a snapshot-covered COPY ALTER A→C that records lineage C→A. A DATA BRANCH clone can resolve S, then concurrent DROP SNAPSHOT can delete S and run `compactHistoricalAlterLineageWithBH`, removing C→A before the clone publishes its independent B→A metadata/snapshot rows. The clone uses the already-resolved timestamp, so it can still succeed, but B is now disconnected from current C and later DIFF/MERGE cannot establish the correct LCA. This affects table and database clones. Lock/revalidate the named-snapshot lineage through branch publication, or atomically transfer ownership from the snapshot to the new branch.

VioletQwQ-0:
Addressed the named-snapshot lineage publication race in `5dda1174540b`.

The DATA BRANCH transaction now locks the selected named user-snapshot row with `FOR UPDATE` and revalidates its full identity (name, physical timestamp, level, object ID, and kind) before any target mutation. The lock remains owned by the same background transaction through branch metadata and protect-snapshot publication.

This gives the two race outcomes explicit semantics:

- If branch creation acquires the lock first, `DROP SNAPSHOT` waits until the new lineage owner is durable.
- If `DROP SNAPSHOT` (or same-name recreation) wins first, branch creation observes a missing or mismatched snapshot after the wait and returns `ErrTxnNeedRetryWithDefChanged` before creating the target.

The guard is restricted to DATA BRANCH named-snapshot sources. Ordinary CLONE/restore paths are unchanged, and both table-level and database-level DATA BRANCH entry points use the same lock contract.

Validation:

- `go test ./pkg/frontend -run 'Test(ShouldLockNamedDataBranchCloneSnapshot|LockNamedDataBranchCloneSnapshot)$' -count=1`
- `go test -race ./pkg/frontend -run 'Test(ShouldLockNamedDataBranchCloneSnapshot|LockNamedDataBranchCloneSnapshot|TimestampDataBranch)' -count=10`
- `go test ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile -count=1`
- `go vet ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile`
- `git diff --check`
- exact-head `mo-pr-preflight-review`: PASS for `5dda1174540b` (`4944a05d...`)

Re-requesting all currently blocking reviewers together.

gouhongshen (approved):
## Codex automated review

Reviewed 5dda1174540b against origin/main and the full existing review discussion. The named-snapshot locking and revalidation fix closes the prior blocker; no new actionable regressions found.

VioletQwQ-0:
The failed Coverage check on `5dda1174540b` was caused by incomplete coverage collection after the runner killed `mo-service` for cgroup OOM (about 7.8 GiB RSS). The resulting modified-line coverage was 74.7302%, just below the 75% gate; the unit, SCA, Proxy, and pessimistic BVT jobs on the same head passed.

Per the CI workflow, I first updated the branch to latest `upstream/main` instead of rerunning the superseded job. The 44 PR commits rebased without conflicts, and `git range-diff` reports every commit as semantically equivalent.

New head: `6fa4ee479ef4`

Validation before the force-with-lease update:

- `go test ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile -count=1`
- `go test -race ./pkg/frontend -run 'Test(ShouldLockNamedDataBranchCloneSnapshot|LockNamedDataBranchCloneSnapshot|TimestampDataBranch)' -count=10`
- `go vet ./pkg/frontend ./pkg/frontend/databranchutils ./pkg/sql/compile`
- exact-head `mo-pr-preflight-review`: PASS (`c7f34bbd...`)

Fresh CI is now running on the rebased head; the failed old-head Coverage job was not rerun.

gouhongshen (changes requested):
## Codex automated review

Earlier blockers around target-only drift, ALTER transactions, lineage GC/ownership, identity validation, timestamp/PITR publication, and named-snapshot locking are fixed or closed; the compaction-atomicity concern was disproven and withdrawn. The previously reported no-OriginName LCA issue was claimed fixed, but remains blocking on ALTER lineage edges.

### P1 - Preserve rename identity across ALTER lineage edges (pkg/frontend/data_branch.go:1890)
Author response: commit 86f1b9542 was said to thread lineage-resolved target-to-LCA/base mappings through SQL and reader probes and add a no-OriginName rename regression. Why this remains blocking: an ALTER copy-and-swap edge is marked lineage-only, so this condition skips logical-name and (ColId, Seqnum) identity fallback. For `base(a,b)`, two branches, then `ALTER ... RENAME COLUMN b TO bb`, the persisted endpoint has `Name=bb` and no distinct OriginName; `dataBranchColumnReachesLCA` therefore returns false for `bb`. `dataBranchLineageEndpointColumns` cannot map it to base `b`, `lcaColNames` remains empty, and schema compatibility treats `bb` as target-only while reporting base `b` as missing. The PR’s own `branch_edge_cases.sql` expects this exact case to emit an UPDATE, but the current code rejects it before LCA probing. Permit and validate stable-identity fallback for a pure ALTER rename, and add a regression covering `lineageOnly=true` with no OriginName.

gouhongshen (approved):
## Codex automated review

Follow-up delta is only the merge conflict resolution in `pkg/frontend/clone_test.go`, cleanly retaining both test sets. Earlier schema/lineage, publication, PITR, and snapshot-locking blockers are fixed; compaction atomicity was withdrawn. The no-OriginName ALTER-rename P1 remains an unchanged prior-head issue with no author response, so it is not duplicated in this round. Focused tests could not run because `go` is unavailable.

VioletQwQ-0:
Fixed the remaining ALTER-lineage rename blocker in f9177b351.

The lineage walk now uses stable `(ColId, Seqnum)` identity fallback on COPY ALTER edges as well as ordinary edges, so a pure `b -> bb` rename still reaches the LCA when `OriginName` is absent. Same-statement replacements remain rejected before lineage publication.

Added an `OriginName present/absent × ordinary/lineageOnly` matrix and changed the endpoint compatibility regression to exercise `lineageOnly=true`.

Validation:
- `go test ./pkg/frontend ./pkg/frontend/databranchutils -count=1`
- focused frontend `go test -race`
- `go vet ./pkg/frontend ./pkg/frontend/databranchutils`
- `git diff --check upstream/main`
- exact-head preflight: PASS (`validation=PENDING` for the new CI run)

gouhongshen (approved):
## Codex automated review

All prior blockers are fixed or withdrawn, including projection/LCA handling, lineage cleanup and ownership, column replacement, clone/PITR/snapshot races, and OUTPUT AS. The remaining no-OriginName rename blocker is fixed in f9177b3 via stable (ColId, Seqnum) fallback on COPY ALTER edges, with coverage for ordinary and lineage-only paths. No new findings.

VioletQwQ-0:
Fixed the current-head CI regressions in 3da72687c.

- DATA BRANCH source locking now uses the same forced pessimistic-RC background transaction that owns quota checking and metadata publication, without changing the outer SI transaction. The exact `TestIssue26087ConcurrentDataBranchQuota` regression now passes.
- COPY ALTER lineage fallback now requires a complete visible-column identity mapping, and the planner/compile path rejects same-statement DROP+ADD replacements before publishing lineage. Focused frontend, planner, and compile tests pass.

The PROXY `pipeline end signal delivery failed` was unrelated; the coverage summary failure was downstream of the failed UT producer. New CI is running on this head.

test/distributed/cases/git4data/branch/merge/merge_schema_evolve.sql
```diff
@@ -0,0 +1,233 @@
+drop database if exists test;
+create database test;
+use test;
+
+-- =====================================================
+-- Case 1: basic schema evolution merge - target has extra column
+--   base:   [a, b]
+--   target: [a, b, c]
+-- =====================================================
+create table t0(a int, b int, primary key(a));
```
fengttt: For multi test cases in one file, choose the name of the table carefully/meaningfully.   Do not use t0, t1, etc.

For example, create table merge_case1, or merge_case1_addcol, so that later when this test fails, we can immediately tell which cases failed and we can check with expected result.

fengttt (commented):
Please cleanup test cases, with better names.

Actually there need to be a solid/clean doc on the behavior of the merge.  I am not sure are all the cases listed/handled in reasonable way, and I want to read that doc and review the design.  This is a pretty tricky thing.  We should do the safe route -- unless we are exactly sure what to do, raise an error.     The "we are exactly sure what to do" part should be designed/documented/reviewed.

aptend (changes requested):
Requesting changes for one P1 concurrency-correctness issue.

[P1] Serialize historical-owner publication with copy ALTER

The historical snapshot/PITR probes in pkg/sql/compile/alter.go use SELECT ... FOR UPDATE, which only protects owner rows that already exist. A concurrent publisher can choose its snapshot timestamp or PITR create time before copy ALTER commits, while its catalog row is not visible yet. ALTER then observes no historical owner, skips lineage preservation, performs the copy-and-swap, and the owner commits afterward with a time before the ALTER boundary.

A later data branch created from that historical owner can therefore resolve the old physical generation without an ALTER edge to the current generation, so DIFF/MERGE cannot reliably reconcile schema identity across the copy ALTER.

The new PITR mo_branch_metadata lock only serializes PITR publication with compaction: ALTER does not acquire that lock before its early owner probe, and CREATE SNAPSHOT does not acquire it at all. Please add a stable serialization point shared by CREATE SNAPSHOT, CREATE PITR, and copy ALTER before timestamp selection/owner probing, or revalidate the owner set under such a lock before the swap. Please also add a barrier-based concurrent regression test for this interleaving.

Current head also has a failing PROXY BVT: pitr_basic.sql times out while selecting mo_pitr. That should be investigated or rerun before merge.

XuPeng-SH (changes requested):
Deep-reviewed exact head c955ed65a50ef517befe3a7e0cccb40e758b477a. One P1 correctness race remains.

doCreateSnapshot chooses the snapshot timestamp before publishing the owner row, and does not acquire a stable lock shared with COPY ALTER. prepareAlterDataBranchLineage probes owner rows with SELECT ... FOR UPDATE, but an empty result does not serialize against a concurrent later insert. The same issue exists for PITR because ALTER does not take the PITR-side stable lock before its early owner probe.

Interleaving: snapshot/PITR chooses time t; COPY ALTER sees no owner and skips the A-to-C lineage handoff, then swaps A to C; the owner row commits with timestamp t. A later data branch resolves old A at t, while current C has no LCA lineage, leaving the branch histories disconnected.

Please introduce synchronization on a stable key shared by owner publication and COPY ALTER before timestamp selection/owner probing, and add a deterministic barrier test covering this interleaving.

VioletQwQ-0:
Fixed the empty-owner publication race in 775574996. CREATE SNAPSHOT, CREATE PITR, and COPY ALTER now cross the same bootstrap-created mo_feature_registry write barrier before timestamp selection/owner probing. The write preserves the registry metadata, serializes pessimistic transactions by row lock and optimistic transactions by write conflict, and snapshot/PITR commit conflicts are now returned instead of being reported as success.

I initially used mo_database as the stable key, but an authenticated cluster regression exposed that as invalid because it enters special catalog DDL tuple parsing. I replaced it before pushing; both issue cluster tests now pass, along with the full frontend/databranchutils/compile normal and race suites and pkg/vm/engine/test. Please re-review the current head.

test/distributed/cases/git4data/branch/merge/merge_schema_evolve.sql
VioletQwQ-0: Renamed the generic t0/t1/t2 tables by scenario and role in merge_schema_evolve.sql, so a failing case can be identified directly from the table name. The thread is addressed on the current head.

aptend (changes requested):
Requesting changes for one remaining P1 correctness gap.

### [P1] Serialize the engine/internal PITR publication path with COPY ALTER

The new stable publication barrier closes the frontend `doCreatePitr` path, but `CREATE PITR ... INTERNAL` is deliberately routed to engine execution (`CreatePitr.StmtKind` returns `defaultStatusTyp`) and reaches `Scope.CreatePitr` in `pkg/sql/compile/ddl.go`. That path resolves and stores the target object ID during planning, then selects `now` and inserts into `mo_pitr` without crossing `LineageOwnerPublicationLockSQL`.

Concrete interleaving: an INTERNAL PITR resolves old generation A; concurrent COPY ALTER crosses the barrier, observes no published historical owner, skips lineage preservation, and swaps A to C; the PITR then inserts its old timestamp/object ID. A later data branch restored from that PITR can resolve A with no lineage edge to current C, reproducing the disconnected-generation failure this barrier is intended to prevent. INTERNAL PITR is a real supported path and is exercised by `pkg/tests/ddl/ddl_test.go`.

Please make the INTERNAL/engine path share the barrier across object resolution, timestamp selection, and publication. Adding the write only immediately before the insert is insufficient if the plan already captured a stale object ID; either route this path through the frontend creation flow or refresh/re-resolve the object under the barrier. Add a deterministic regression for `CREATE PITR ... INTERNAL` racing COPY ALTER.

VioletQwQ-0:
Addressed the INTERNAL PITR publication race in db2107d3f.

`Scope.CreatePitr` now crosses the same stable publication barrier as the frontend snapshot/PITR paths before choosing the PITR timestamp. After a pessimistic-RC wait it refreshes the transaction snapshot and re-resolves the current database/table ID, so the insert no longer uses the generation captured during planning. The barrier write and `mo_pitr` insert remain in the same statement transaction through commit/rollback.

I added a deterministic COPY ALTER interleaving and a stale planned-ID regression. I also fixed the current PESSIMISTIC BVT failure: `merge_schema_evolve.sql` had been renamed by scenario, but its `.result` commands still used `t0/t1`; all 139 SQL/result statements now match. The Utils Coverage failure was only the downstream prerequisite failure from that BVT job.

Validation:
- `TestPreparePitrPublicationSerializesCopyAlter` with `-race -count=100`
- `TestResolveCurrentPitrObjectIDRefreshesPlannedTableID` with `-race -count=100`
- full `pkg/sql/compile` with `-race -count=1`
- production `TestPitrCases`
- `TestIssue26087ConcurrentDataBranchQuota` and `TestIssue26114CrossAccountBranchUsesTargetQuotaAndOwnership`
- 139/139 BVT SQL/result statement alignment
- `git diff --check`
- exact-head preflight: `PASS review=PASS validation=PENDING`

VioletQwQ-0:
Synced with current `upstream/main` in 73e0cdb14. The only content conflict was in `clone_database_source_test.go`; I retained both this PR’s empty-source lock regression and main’s cyclic-FK ordering regression. Focused race tests (100 iterations each), full `pkg/frontend` and `pkg/sql/compile` race suites, production `TestPitrCases`, BVT statement alignment, diff check, self-review, and exact-head preflight all pass locally. Fresh CI is pending.

XuPeng-SH (approved):
Deep review complete on exact head `dba14dfe0c`. The prior historical-owner/COPY ALTER race is closed by a real stable-row write barrier shared by frontend snapshot/PITR publication, INTERNAL PITR publication, and COPY ALTER. The INTERNAL path now refreshes its RC snapshot and re-resolves the database/table object ID under that barrier instead of using the generation captured during planning. I traced all production owner-write entry points, the optimistic WW-conflict and pessimistic lock paths, transaction cleanup, lineage preservation, and the latest merge resolutions; no remaining blocker found. Six focused race regressions passed 100 iterations each, full race suites for frontend/databranchutils, frontend, and sql/compile passed, all eight owning packages passed test/build/vet, and embedded-cluster TestPitrCases passed. Required CI is green and this exact head merges cleanly with current main.

XuPeng-SH (approved):

PR #25894, opened 2026-07-20
fix: resolve table_stats in requested account

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25782

Fixes the three-part `table_stats` path (`db.table.account_id`) so relation lookup and statistics access run in the requested account context.

Previously, `table_stats` parsed `account_id` and placed it in `StatsInfoKey`, but resolved the database and table with the caller account context. A sys query for a tenant-only table therefore returned `database not found`; identical names could also combine the tenant account ID with sys relation IDs.

The fix:

- keeps cross-account authorization before engine access
- derives a local target-account context without mutating the process context
- reuses the current transaction operator
- uses the target context for database/relation IDs and GlobalStats operations
- strictly validates `account_id` as `uint32`
- adds unit and distributed cross-account regression coverage

Validation:

- `make thirdparties`
- `make build`
- `go test ./pkg/sql/colexec/table_function`
- `go test -race ./pkg/sql/colexec/table_function`
- local disttae integration: tenant same-account lookup passed; tenant cross-account lookup was rejected; sys dynamic cross-account lookup returned the tenant table


aunjgr (approved):
Reviewed head 74ac69ef1. No blocking defects found. The requested account context is used consistently for relation and GlobalStats lookup without mutating the shared process context, and account parsing and cross-account authorization have focused coverage.

heni02 (approved):

gouhongshen (approved):
## Codex automated review

No blocking findings. The target-account context is applied consistently to relation resolution and GlobalStats operations while preserving authorization and process-context isolation.

XuPeng-SH (approved):
No blocking findings after a full correctness / closure / unhappy-path pass.

The fix now carries the requested account context consistently through relation resolution and GlobalStats access without mutating `proc.Ctx`, and the stricter `account_id` parsing closes the old loose-parse edge cases. The added unit coverage plus distributed cross-account case cover the important same-account, forbidden tenant cross-account, and sys-to-tenant lookup paths.

PR #25898, opened 2026-07-20
fix: align explicit CAST overflow behavior with MySQL

### LINKED ISSUE #25252: [Bug]: CAST overflow handling differs from MySQL clamp/wrap behavior
### MatrixOne Version

Latest `origin/main` verified locally at `33e96877604ad4b847ea7d5c55eb09429de1a5b6`.

### Reproduce Steps

```sql
select cast('999999999999999999999999' as signed);
select cast('999999999999999999999999' as unsigned);
select cast('9999999.99' as decimal(6,2));
```

### Expected Behavior

MySQL returns saturated/wrapped cast results with warnings:

```text
-1
18446744073709551615
9999.99
```

### Actual Behavior

MatrixOne returns errors:

```text
ERROR 1690 (HY000): data out of range: data type int64, value '999999999999999999999999'
ERROR 1690 (HY000): data out of range: data type uint64, value '999999999999999999999999'
ERROR 20301 (HY000): invalid input: 9999999.99 beyond the range, can't be converted to Decimal64(6,2).
```

### Notes

Found during Batch A of the MySQL compatibility cast matrix. Overflow rows were excluded from passing BVT.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25252

Explicit SQL `CAST` and `CONVERT` expressions previously reused the ordinary internal cast overload, whose strict overflow errors differ from MySQL's explicit-cast behavior.

This change:

- routes only MySQL `SIGNED`, `UNSIGNED`, and `DECIMAL` cast targets through a dedicated overload, keeping MatrixOne extension integer targets and planner-inserted casts unchanged;
- applies MySQL-compatible signed/unsigned overflow handling across string, integer, float, Decimal64, Decimal128, and Decimal256 sources;
- clamps explicit decimal casts across Decimal64, Decimal128, and Decimal256 target precisions;
- preserves errors for invalid numeric strings;
- adds planner/function unit tests, column-vector coverage, extension-target regression tests, and BVT coverage for the reported cases and boundaries.

User impact: overflow cases such as large integer strings, negative-to-unsigned casts, and oversized decimals now produce MySQL-compatible values without changing existing strict semantics for implicit casts or MatrixOne extension integer targets.

Validation:

- `go test ./pkg/sql/plan/function ./pkg/sql/plan`
- `go test -race ./pkg/sql/plan/function ./pkg/sql/plan`
- `make build`
- `func_cast.test` BVT: 226/226 passed
- `boundary_comprehensive.sql` BVT: 241/241 passed
- changed helper coverage: 75.0%-100%
- `git diff --check`


pkg/sql/plan/function/func_cast.go
```diff
@@ -448,22 +617,89 @@ func IfTypeCastSupported(sourceType, targetType types.T) bool {
...
+				func(v types.Decimal64) string { return v.Round(fromType.Scale, 0, true).Format(0) })
+		case types.T_decimal128:
+			return decimalToInt64Explicit(vector.GenerateFunctionFixedTypeParameter[types.Decimal128](from), result, length, selectList,
+				func(v types.Decimal128) string { return v.Round(fromType.Scale, 0, true).Format(0) })
+		case types.T_decimal256:
+			return decimalToInt64Explicit(vector.GenerateFunctionFixedTypeParameter[types.Decimal256](from), result, length, selectList,
+				func(v types.Decimal256) string { return v.Round(fromType.Scale, 0, true).Format(0) })
+		}
+	}
+	strictStringWidth := mode.strictStringWidth()
```
XuPeng-SH, line 921: [P1] Clamp non-string inputs for explicit DECIMAL casts

Explicit mode reaches clamping only through `strToDecimal*`; integer, float, and Decimal64/128/256 sources still fall through the ordinary conversions. At this head, direct execution of the outer cast in `CAST(CAST(10000 AS SIGNED) AS DECIMAL(6,2))` produced raw Decimal64 `1000000` (`10000.00`) instead of the MySQL endpoint `999999` (`9999.99`). MySQL checks target precision after obtaining a decimal from any source ([reference implementation](https://github.com/mysql/mysql-server/blob/d229bb760c49b65e19ec28342236961ad961d7fe/sql/item_func.cc#L2094-L2126)). Please route every explicit DECIMAL target/source combination through the clamp and add nested-cast plus column-vector coverage.

pkg/sql/plan/function/func_cast.go
```diff
@@ -5493,6 +5804,39 @@ func strToDecimal64(
...
+		return false, err
+	}
+	if token.base != 10 {
+		var value big.Int
+		if _, ok := value.SetString(token.digits, token.base); !ok {
+			return false, strconv.ErrSyntax
+		}
+		return token.negative, nil
+	}
+	if _, _, err = big.ParseFloat(token.digits, 10, 256, big.ToNearestEven); err != nil {
```
XuPeng-SH: [P1] Do not classify `Inf` as decimal overflow

`big.ParseFloat` accepts `Inf`, so this fallback mistakes a non-finite/invalid token for a valid overflowing finite decimal. Exact-head execution of casting the string `Inf` to `DECIMAL(6,2)` succeeded with `9999.99`; before this overload the parse error propagated, and the PR contract says invalid numeric strings remain errors. Reject non-finite spellings (and clamp only proven range failures), with `Inf`, `+Inf`, and `-Inf` tests across Decimal64/128/256.

pkg/sql/plan/function/func_cast.go
```diff
@@ -40,6 +41,174 @@ import (
...
+	for i := uint64(0); i < uint64(length); i++ {
+		value, null := from.GetValue(i)
+		if null {
+			if err := to.Append(0, true); err != nil {
+				return err
+			}
+			continue
+		}
+		floatValue := float64(value)
+		rounded := math.Round(floatValue)
```
XuPeng-SH: [P1] Preserve nearest-even rounding for approximate values

Both new float helpers use `math.Round`, which rounds halves away from zero. The MySQL DOUBLE/FLOAT-to-integer path uses `rint` ([reference implementation](https://github.com/mysql/mysql-server/blob/d229bb760c49b65e19ec28342236961ad961d7fe/sql/item_func.cc#L2177-L2191)), so on Linux `CAST(2.5E0 AS SIGNED)` and the `-2.5E0` counterpart produce `2` and `-2`; exact-head `NewExplicitCast` produced `3` and `-3`. Use nearest-even for approximate inputs in both the signed and unsigned helpers while preserving half-away behavior for DECIMAL inputs, and cover `0.5`/`1.5`/`2.5` with both signs.

XuPeng-SH (changes requested):
Requesting changes for three independently reproduced correctness gaps at exact head 246ce82d. The PR-added function and planner tests pass, but they do not cover non-string DECIMAL overflow, non-finite decimal text, or approximate half-to-even rounding.

XuPeng-SH (approved):
Re-reviewed at 2920108b4a. The three requested correctness gaps are closed: explicit numeric-to-DECIMAL casts now clamp across numeric source vectors, non-finite decimal strings remain errors, and FLOAT/DOUBLE integer casts use nearest-even while DECIMAL keeps half-away rounding. Also covered the asymmetric negative DECIMAL endpoint and preserved large finite/in-range FLOAT-to-DECIMAL behavior. Local validation: full pkg/sql/plan/function + pkg/sql/plan tests, build, vet, and targeted race tests all pass.

heni02 (approved):

PR #25902, opened 2026-07-20
fix(proxy): track in-flight requests on disconnect

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25599 

  - Track client requests explicitly after the proxy forwards them to the backend.
  - Clear the in-flight marker once the proxy observes a terminal OK/ERR backend
  response.
  - Use the marker during abnormal client disconnect cleanup so blocked backend
  statements are killed instead of being allowed to continue and commit.
  - Add a regression test for disconnect cleanup that does not rely on the previous c2s/
  s2c timestamp heuristic.


XuPeng-SH (approved):

This PR improves the disconnect unhappy path by replacing the old c2s/s2c timestamp heuristic with an explicit “client request forwarded but no terminal backend response observed yet” marker. That makes abnormal client disconnect cleanup much more direct and robust: blocked backend work is killed when the proxy knows a request is still in flight, while idle/disconnect-after-completion paths still just close the backend side. The added regression test also exercises that explicit lifecycle instead of relying on timing side effects.

I did not find a new correctness or unhappy-path blocker in the updated diff.

PR #25909, opened 2026-07-20
fix(txn): prevent stale responses from pooled local streams

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25785

### Background and root cause

`TxnSender` supports local dispatch so callers can invoke a TN handler in the same process without loopback RPC.

The first version of this optimization (#3826) called the local handler synchronously. Later, #3959 unified multi-request handling around the remote stream model: send every request first, then receive every response. To fit that model, local dispatch gained a pooled `localStream` containing a persistent worker goroutine and reusable `in` / `out` channels. Response pooling was subsequently layered on top.

That lifecycle allows one `Send` generation to escape into the next:

1. a mixed batch queues a local request;
2. creation or sending of a later remote stream fails;
3. `SendResult.Release` closes the local stream and immediately returns it to `sync.Pool`;
4. the old handler can still finish and write to the reused `out` channel;
5. a later `Send` reuses the same stream and consumes the old response as its own.

This is an object-generation isolation bug rather than a Go data race.

### Design choice

Two approaches were considered:

1. retain `localStream` and add generation tokens, isolated channels, cancellation, draining, and exact late-response ownership;
2. remove the asynchronous adapter and execute local handlers directly, while retaining the existing remote `morpc.Stream` path.

This PR uses the second approach. The stale response is caused by the asynchronous pooled adapter, not by local dispatch itself. Removing that adapter restores the original synchronous local-dispatch semantics and eliminates the complete stale-generation state space instead of adding lifecycle machinery to manage it.

### Changes

- Execute local handlers synchronously in the multi-request `Send` path and place each response at its original request index.
- Resolve and cache `LocalDispatch` once per shard within a `SendResult`, preserving the previous per-shard handler semantics, including cached remote (`nil`) resolution.
- Preserve local handler error conversion to `TxnError` and existing `TxnResponse` pooling.
- Clear the per-send handler cache before returning `SendResult` to its pool.
- Keep remote requests on the existing `morpc.Stream` send/receive path.
- Remove `localStreamPool`, `localStream`, `sendMessage`, the worker goroutine, reusable channels, finalizer, and their setup/close lifecycle.

The production change is a net deletion of roughly 80 lines.

### Ownership and failure behavior

A local response now has one synchronous ownership chain:

`responsePool.Get -> local handler -> copy into SendResult -> Reset/Put`

There is no background owner that can outlive `Send`, and therefore no old generation that can publish into a reused stream. On send/setup failures, `SendResult.Release` closes remote streams and clears both response and local-handler state before reuse.

### Tradeoff

Mixed local/remote batches no longer overlap local handler execution with later remote stream setup or sending. A local handler that blocks forever and ignores its context will block the caller before it reaches a later remote failure. Single-request local dispatch was already synchronous, and production handlers propagate the request context, so this is accepted in exchange for the substantially simpler lifecycle. If mixed-path overlap is needed later, remote requests can be sent first without reintroducing a pooled asynchronous local stream.

### Validation

- Added a deterministic regression covering a completed old local response left behind after a mixed local/remote stream failure. The old implementation returns `old` where the next call expects `new-1`.
- Added coverage for multiple local requests, response ordering, per-shard handler resolution, handler-cache cleanup across pooled `SendResult` reuse, and local handler error wrapping.
- `go test -count=20 ./pkg/txn/rpc -run '^(TestSendWithMultiTNAndLocal|TestSenderDoesNotReturnStaleLocalResponseAfterMixedStreamFailure|TestSendWithMultiLocalRequestWrapsHandlerError)$'`
- `go test ./pkg/txn/rpc`
- `go test -race ./pkg/txn/rpc`
- `go test ./pkg/txn/client`
- `go vet ./pkg/txn/rpc`
- `go build ./pkg/txn/rpc`

`BenchmarkLocalSend` improved from approximately 2.1 us/op to 0.79 us/op on the same machine, with steady-state allocation remaining at `0 B/op, 0 allocs/op`.


gouhongshen (approved):
## Codex automated review

No concrete correctness, compatibility, concurrency, resource, or performance regression found. RPC/client tests, race checks, vet, and diff validation pass.

XuPeng-SH (approved):
Deep review at d9b0f198217d8711b3b29b7e08d382b5126b2910.\n\nNo merge-blocking correctness, concurrency, lifecycle, compatibility, or performance issue found.\n\nThe change removes the pooled asynchronous local-stream generation entirely, so an old handler can no longer publish into a reused channel. Local response ownership is synchronous and exact; remote streams retain their existing cleanup path; the per-send handler cache is cleared before result reuse.\n\nValidated:\n- focused local/mixed/error regressions: 20 iterations\n- focused race regressions: 10 iterations\n- full pkg/txn/rpc tests and full race suite\n- pkg/txn/client\n- targeted TN local-dispatch, 2PC, and recovery closure tests\n- go vet and go build\n- same-machine benchmark: 0 B/op, 0 allocs/op and materially faster than current main's pooled local stream

Issue #25892, opened 2026-07-20
[Bug]: Temporary proxy admin connection removes the active tunnel from the CN draining index


Ariznawlll:
## Summary

Proxy temporary admin connections created for `KILL` and `UPGRADE` reuse the current client session's `*tunnel` pointer when they are registered in `connManager`.

When the temporary connection targets the same CN as the active session, both connections use the same key in the global `CN UUID -> tunnelSet` map. Closing the temporary connection then deletes that key and silently removes the still-active session from the CN draining index.

## Affected version

Latest `main` tested at `862d9d30c2cd7a55995fd0bae0951af0809f5752`.

## Production path

`clientConn.connAndExec`, used by `handleKill`, `handleUpgradeEvent`, and backend cleanup, currently calls:

```go
c.router.Connect(cn, c.handshakePack, c.tun)
```

`router.connect` registers the temporary connection through `connManager.connect(cn, t)`. The global index is a set keyed only by the tunnel pointer:

```go
m.cnTunnels[cn.uuid][t] = struct{}{}
```

If the real session and temporary connection are on the same CN, the second registration is deduplicated. The temporary `ServerConn.Close` later calls `disconnect(cn, t)`, which deletes the one shared entry.

`scaling.doScaling` relies exclusively on `getTunnelsByCNID` to find sessions that must be moved away from a draining CN. The lost live tunnel is therefore not delivered for migration during scale-in/draining.

## Expected behavior

Closing a temporary admin connection must not change tracking ownership of the active client tunnel. Temporary connections should not be registered as the active session tunnel.

No matching open issue was found for temporary proxy connections corrupting the CN draining tunnel index.

Ariznawlll:
## 最新 main 白盒验证

- 测试基线：`main@862d9d30c2cd7a55995fd0bae0951af0809f5752`
- 临时测试文件：`pkg/proxy/conn_manager_temp_audit_test.go`
- 测试对象：`clientConn.connAndExec`、`router.Connect` 和 `connManager.cnTunnels` 的所有权关系。

### 测试用例

`TestAuditTemporaryAdminConnectionKeepsActiveTunnelTracked` 执行以下步骤：

1. 创建一个真实活动 tunnel，并把它注册到 `cn-1` 的全局 draining 索引。
2. 模拟 `KILL`/`UPGRADE` 的 `connAndExec` 路径，在同一个 `cn-1` 上建立临时连接。
[...]

PR #25912, opened 2026-07-20
fix(proxy): isolate temporary admin connection tracking

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25892

`clientConn.connAndExec` creates temporary backend connections for KILL,
UPGRADE, and backend cleanup. It previously registered each temporary connection
with the active client session's `*tunnel`.

The global draining index is a set keyed by tunnel pointer. When a temporary
connection targeted the same CN as the active session, its registration reused
the same key. Closing the temporary connection then deleted the only entry for
the still-active session, so `scaling.doScaling` could no longer find and
migrate that tunnel from a draining CN.

This change makes temporary admin connections pass a nil tunnel. The connection
manager treats nil as explicit temporary ownership: it keeps the
`connection ID -> CN server` mapping needed by admin routing, but never inserts
or deletes entries in the tenant-balancing or CN-draining tunnel maps.

Active session registration, placeholder accounting, connection-ID lookup,
normal tunnel cleanup, and scaling policy are unchanged. Regression tests cover
the `connAndExec` handoff and prove that connecting and closing a temporary
same-CN connection preserves the active draining-index entry while cleaning up
the temporary connection ID.

## Test

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=100 -timeout=5m -run '^(TestClientConn_KillCurrentBackendConn|TestConnManagerTemporaryConnectionDoesNotTrackNilTunnel)$' ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -timeout=5m -run '^(TestClientConn_KillCurrentBackendConn|TestConnManagerTemporaryConnectionDoesNotTrackNilTunnel)$' ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=10m ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=15m ./pkg/proxy`

`git diff --check`


XuPeng-SH (approved):
Approved after tracing the complete temporary-connection lifecycle.

The ownership split is correct: admin connections retain connection-ID routing but no longer participate in tenant balancing or CN-draining tunnel indexes. All connAndExec terminal paths close the temporary ServerConn, nil-tunnel disconnect is idempotent, and active-session tracking remains unchanged. Ordinary sessions and transfer paths still pass a non-nil tunnel.

No correctness, concurrency, lifecycle, compatibility, or performance blocker found. The extra nil branch is negligible and temporary connections avoid unnecessary tunnel-map mutations.

Validated current head 8a9aa4f09c9aa0f4fcf902edaeccf031ac5ba2bb with focused tests x20, focused race tests x5, full pkg/proxy tests, full pkg/proxy race tests, go vet, go build, and git diff --check. All passed.

PR #25913, opened 2026-07-20
fix(proxy): reject cached sessions on draining CNs

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25895

Normal proxy routing obtains candidates from `MOCluster.GetCNService`, which
excludes draining and removed CNs. Cached backend reuse bypassed that candidate
selection and `router.CanReuseCachedCN` checked only the CN health circuit
breaker.

A draining CN can remain breaker-healthy, so a fresh client login could continue
to reuse a cached backend session on a CN that normal routing had already made
ineligible. This could delay or defeat graceful draining and scale-in.

This change requires cached reuse to pass both existing gates:

- the health breaker must allow the CN; and
- the CN UUID must still be present in the current `GetCNService` routable
  cluster view.

The existing nil-entry behavior is preserved. Missing and draining CNs are
rejected, while a healthy routable CN remains reusable. No cache capacity,
authentication, FIFO, timeout, routing-label, or circuit-breaker state semantics
change. Regression tests cover nil, healthy, breaker-unhealthy, recovered,
draining, and removed states.

## Test

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=100 -timeout=5m -run '^TestCanReuseCachedCNRequiresCurrentRouteEligibility$' ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -timeout=5m -run '^TestCanReuseCachedCNRequiresCurrentRouteEligibility$' ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=10m ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=15m ./pkg/proxy`

`git diff --check`

### DIFF
--- a/pkg/proxy/client_conn.go
+++ b/pkg/proxy/client_conn.go
@@ -999,6 +999,7 @@ func (c *clientConn) connectToBackendContext(
 				c.connID,
 				c.mysqlProto.GetSalt(),
 				c.mysqlProto.GetAuthResponse(),
+				c.clientInfo,
 			)
 			cancel()
 		} else {
@@ -1007,6 +1008,7 @@ func (c *clientConn) connectToBackendContext(
 				c.connID,
 				c.mysqlProto.GetSalt(),
 				c.mysqlProto.GetAuthResponse(),
+				c.clientInfo,
 			)
 		}
 		if sc != nil {
--- a/pkg/proxy/client_conn_test.go
+++ b/pkg/proxy/client_conn_test.go
@@ -2416,6 +2416,63 @@ func Test_connectToBackend_SkipCacheOnMigration(t *testing.T) {
 	require.Equal(t, 0, cache.popCount)
 }
 
+func Test_connectToBackend_PassesClientInfoToCache(t *testing.T) {
+	cache := &popCountConnCache{}
+	client := clientInfo{labelInfo: labelInfo{
+		Tenant: "tenant1",
+		Labels: map[string]string{"region": "cn"},
+	}}
+	cConn := &clientConn{
+		ctx:        context.Background(),
+		router:     &routeErrRouter{},
+		mysqlProto: &frontend.MysqlProtocolImpl{},
+		connCache:  cache,
+		clientInfo: client,
+		log:        runtime.DefaultRuntime().Logger(),
+	}
+
+	sConn, err := cConn.connectToBackend("")
+	require.Error(t, err)
+	require.Nil(t, sConn)
+	require.Equal(t, 1, cache.popCount)
+	require.Equal(t, client, cache.lastClient)
+}
+
+type contextPopCountConnCache struct {
+	popCountConnCache
+	popContextCount int
+}
+
+func (c *contextPopCountConnCache) PopContext(
+	_ context.Context, _ cacheKey, _ uint32, _ []byte, _ []byte, client clientInfo,
+) ServerConn {
+	c.popContextCount++
+	c.lastClient = client
+	return nil
+}
+
+func Test_connectToBackend_PassesClientInfoToContextCache(t *testing.T) {
+	cache := &contextPopCountConnCache{}
+	client := clientInfo{labelInfo: labelInfo{
+		Tenant: "tenant1",
+		Labels: map[string]string{"region": "cn"},
+	}}
+	cConn := &clientConn{
+		ctx:        context.Background(),
+		router:     &routeErrRouter{},
+		mysqlProto: &frontend.MysqlProtocolImpl{},
+		connCache:  cache,
+		clientInfo: client,
+		log:        runtime.DefaultRuntime().Logger(),
+	}
+
+	sConn, err := cConn.connectToBackend("")
+	require.Error(t, err)
+	require.Nil(t, sConn)
+	require.Equal(t, 1, cache.popContextCount)
+	require.Equal(t, client, cache.lastClient)
+}
+
 func Test_connectToBackend_SkipCacheWhenPluginRouterEnabled(t *testing.T) {
 	cache := &popCountConnCache{}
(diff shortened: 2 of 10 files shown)


gouhongshen (approved):
## Codex automated review

No findings. Reviewed all PR/issue discussion and inline feedback, diff, routing/cache semantics, and `git diff --check`. Focused tests could not run because `cgo/libmo.dylib` is absent locally.

XuPeng-SH (changes requested):
This fix closes the specific draining-state gap, but cached reuse still does not follow the same routing predicate as a fresh login, so I do not think the closure is complete yet.

1. `pkg/proxy/router.go:349-365` only checks `GetCNService(NewServiceIDSelector(cn.uuid))`. `Selector.filter()` short-circuits on `byServiceID` (`pkg/clusterservice/selector.go:87-91`), so this skips the request-label matching that `Route()` still enforces (`pkg/proxy/label_info.go:112-117`, `pkg/vm/engine/disttae/route/route.go:198-255`). A concrete failure is a cached `t1 + k1=v1` session whose CN later changes labels to `k2=v2`: fresh `Route()` rejects that CN, but `CanReuseCachedCN()` stays true as long as the UUID is still present and Working. So cached reuse still bypasses normal routing after label drift; it just no longer bypasses draining/removed state.

2. More fundamentally, the reuse hook never receives the current login context: `connCache` is keyed only by `LabelHash` (`pkg/proxy/conn_cache.go:173-183`), and `withCanReuseCN` passes only `func(*CNServer) bool` (`pkg/proxy/conn_cache.go:222-225,273-276`). That means it cannot honor `RouteForSuperTenant`'s username-sensitive `dump/root` fallback (`pkg/vm/engine/disttae/route/route.go:25-35`, see the existing expectation in `pkg/proxy/router_test.go:173-200`). A cache entry created under that fallback can still be reused by a later login that `Route()` would reject.

Can we refactor the reuse check to run the same routing predicate as `Route()` (or a shared helper over a one-element candidate snapshot) with the current login context, and extend the regression coverage to at least label drift plus the sys-tenant fallback path?

gouhongshen (changes requested):
## Codex automated review

The draining/removed, label-drift, and sys-user cases are addressed, but cached reuse still does not apply the complete fresh-routing policy.

### P1 - Evaluate eligibility against the full CN snapshot (pkg/proxy/router.go:380)
The routing helpers are given only the cached CN, but their result depends on the other available CNs. For example, cache an empty-label fallback for tenant t1, then add a working CN labeled account=t1. Fresh Route() selects the labeled CN and excludes the empty fallback, whereas this one-element snapshot sets found=false and returns the empty CN through RouteForCommonTenantCandidates' fallback path. A new login therefore still reuses a backend that current routing rejects, preventing routing configuration changes from taking effect. Apply the routing policy to the full current working-CN snapshot and then check whether the cached UUID is among the selected candidates; add coverage for a fallback CN becoming lower priority after a matching CN appears.

gouhongshen (approved):
## Codex automated review

No findings. The current head closes the earlier routing-policy gaps by evaluating cached CNs against the full working-CN snapshot with the current login context, while preserving health and cleanup semantics. Current-head CI is green; local proxy tests could not start because cgo/libmo.dylib is absent.

XuPeng-SH (changes requested):
The routing-policy fix on this head is logically sound, but the PR is currently not mergeable against main and the conflicts combine two required cache-reuse contracts. Please resolve and revalidate before merge.

[P1] Preserve both context/generation safety and current-login eligibility while resolving the cache API conflicts

Current main added context-aware PopContext, bounded SET CONNECTION ID I/O, waitServerConnCacheReuseReady, and old/new tunnel-generation handoff. This PR independently changes Pop/CanReuseCachedCN to carry clientInfo so cached reuse can evaluate the exact current login policy. The merge conflicts in client_conn.go, client_conn_test.go, conn_cache.go, and plugin_test.go cannot safely be resolved by taking either side:

- taking this branch drops main’s cancellation and cache-generation readiness protections;
- taking main drops this PR’s username/label-aware eligibility check and reintroduces the routing bug.

The resolved API needs to carry clientInfo through both legacy Pop and context-aware PopContext. connectToBackendContext must pass c.clientInfo on both paths, while retaining the timeout context, waitServerConnCacheReuseReady, and context-aware ExecStmt. Keep both plugin context-propagation coverage and cache-reuse client-context coverage.

Fresh evidence on current head 9d1fb8a5d4: full pkg/proxy UT, full package race, build, and vet pass. I also traced draining/removed CN, label drift, sys root/dump fallback, and fallback-priority changes; no additional blocker was found in this head’s standalone logic. After resolving against current main, please rerun the full proxy package and race suite because that merged integration is the actual delivery boundary.

XuPeng-SH (approved):
No blocking findings on head `30f6ba8`.

The conflict resolution preserves main's context-bounded `PopContext`, cache-generation readiness barrier, context-aware `SET CONNECTION ID`, and tunnel rebind, while carrying the current `clientInfo` through both cache APIs. Cached eligibility now applies the current login's full fresh-routing policy over the complete working-CN snapshot, including label drift, sys root/dump fallback, empty-label fallback priority, draining/removed CNs, and the existing health gate. Cleanup ownership and cancellation paths remain bounded and generation-safe.

Fresh local evidence on this head: focused routing/context regressions passed 10x, the same set passed under `-race` 5x, full `pkg/proxy` tests passed, and `go build`, `go vet`, plus `git diff --check` passed. CI is still running and remains subject to normal merge protections.

PR #25911, opened 2026-07-20
fix(logservice): preserve HAKeeper ID batch boundaries

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #25887

## What this PR does / why we need it:

HAKeeper returns an allocated batch as the inclusive range
`[firstID, firstID+batchSize-1]`, and the managed client stores `lastID`
with that same inclusive meaning. The cache-consumption check previously treated
`nextID == lastID` as exhausted.

That boundary mismatch had three concrete effects:

- batches larger than one skipped the final allocated ID;
- a batch of one left `nextID > lastID`, so the client generated IDs locally
  without another HAKeeper allocation and could collide with another client;
- `AllocateIDByKeyWithBatch(..., 0)` reached HAKeeper, where `batch-1`
  underflowed.

This change uses one inclusive cache invariant for shared and keyed allocation:
a cached ID is available only when `nextID != 0 && nextID <= lastID`. It also
rejects a zero keyed batch before preparing or calling the HAKeeper client.

No wire format, persisted HAKeeper state, allocation key, default batch size, or
public API signature changes. Regression tests cover complete batch consumption,
batch-one uniqueness across independent clients, and zero-batch rejection before
RPC.

## Test

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=100 -timeout=5m -run '^(TestAllocateIDConsumesEntireBatch|TestAllocateBatchOneIDsRemainUniqueAcrossClients|TestAllocateIDByKeyRejectsZeroBatchBeforeRPC)$' ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -timeout=5m -run '^(TestAllocateIDConsumesEntireBatch|TestAllocateBatchOneIDsRemainUniqueAcrossClients|TestAllocateIDByKeyRejectsZeroBatchBeforeRPC)$' ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=10m ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=15m ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -run '^$' -count=1 -timeout=5m ./pkg/cnservice`

[...]

PR #25915, opened 2026-07-20
test(function): exclude JSON allocation check under race

### LINKED ISSUE #25914: [Bug]: JSON merge allocation test flakes under race instrumentation
## Summary

`TestJsonMergePatchWrapperAllocationsDoNotScaleWithRows` can fail in the Ubuntu/x86 race-enabled UT even when the JSON merge implementation and its dependencies are unchanged. The allocation-only assertion therefore blocks unrelated pull requests as a false regression.

## Root cause

The test in `pkg/sql/plan/function/func_json_valid_test.go` uses `testing.AllocsPerRun(3, ...)` around the complete JSON merge execution for 1 row and 8192 rows. That closure includes result-vector reset/growth, every merge-builder operation, every result append, and race-build-only vector checks; it does not isolate wrapper construction.

`testing.AllocsPerRun` derives its result from the process-wide `runtime.MemStats.Mallocs` delta. In the much longer 8192-row instrumented path, unrelated/instrumentation-dependent Go allocations are therefore attributed to the wrapper assertion. This is outside the production wrapper-allocation invariant that the test intends to protect.

The test was introduced by #25728 and passed that pull request's Ubuntu/x86 race UT. The affected package source remained unchanged, but the same test later failed in #25911 with `one=1, many=45`. On current main, repeated Darwin/arm64 runs report `one=1, many=1` both with and without `-race`. This pass/fail history with unchanged source pins the defect to the test boundary, not to JSON merge or #25911.

## Impact

Confirmed impact: unrelated pull requests can fail the required Ubuntu/x86 UT and require a rerun.

There is no confirmed production correctness or allocation regression in `json_merge_patch`.

## Reproduction / evidence

- Failing run: https://github.com/matrixorigin/matrixone/actions/runs/29736219928/job/88332146244?pr=25911
- Failure: `wrapper allocations must not grow with rows: one=1.000000 many=45.000000`
- Original implementation PR #25728 passed the same Ubuntu/x86 race UT.
- #25911 only changes `pkg/logservice/hakeeper_client.go` and its test; the failing test runs in the separate `pkg/sql/plan/function` test binary.
- Current main passes 100 repeated non-race runs and 20 repeated race runs on Darwin/arm64 with `one=1, many=1`.

## Proposed fix

Keep all JSON merge functional tests in the race-enabled suite, but compile this allocation-statistics-only test only for non-race builds. Go defines the `race` build tag specifically so tests whose timing or instrumentation behavior is incompatible with the detector can be excluded: https://go.dev/doc/articles/race_detector#Excluding_Tests

No production code should change.

## Acceptance criteria

- The Ubuntu/x86 `-race` package test no longer executes the allocation-statistics-only assertion.
- Non-race builds still execute `TestJsonMergePatchWrapperAllocationsDoNotScaleWithRows` and retain its 1-row versus 8192-row invariant.
- JSON merge functional tests continue to run in both race and non-race builds.
- `pkg/sql/plan/function` passes targeted race, non-race, and package-level tests.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25914

### Root cause

`TestJsonMergePatchWrapperAllocationsDoNotScaleWithRows` measures the complete 8192-row JSON merge path with `testing.AllocsPerRun(3)`, even though its asserted invariant is wrapper allocation behavior. Under `-race`, that path also executes instrumented and race-build-only vector operations, so process-wide allocation deltas outside the intended invariant can be charged to the test.

The same package source passed the Ubuntu/x86 race UT in #25728, then failed an unrelated PR in #25911 with `one=1, many=45`. #25911 only changes the HAKeeper client and its tests, which run in a separate package test binary.

### Fix

Move the allocation-statistics-only test unchanged into a `!race` test file.

This keeps:

- the 1-row versus 8192-row allocation regression assertion in non-race builds;
- all JSON merge functional tests in both race and non-race builds;
- production code and JSON merge behavior unchanged.

### Validation

```text
.agents/skills/mo-dev/scripts/mo-cgo-test -v -count=100 -timeout=10m \
  -run '^TestJsonMergePatchWrapperAllocationsDoNotScaleWithRows$' \
  ./pkg/sql/plan/function

.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -timeout=10m \
  -run '^TestJsonMerge($|DepthValidation$)' \
  ./pkg/sql/plan/function

.agents/skills/mo-dev/scripts/mo-cgo-test -count=20 -timeout=10m \
  -run '^TestJsonMerge($|DepthValidation$|PatchWrapperAllocationsDoNotScaleWithRows$)' \
  ./pkg/sql/plan/function

.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=20m \
  ./pkg/sql/plan/function

.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=30m \
  ./pkg/sql/plan/function
```

The race-built test list was also checked to confirm that only the allocation-statistics test is excluded.


XuPeng-SH (approved):

This keeps the allocation-regression assertion in non-race builds, leaves the JSON merge functional coverage in both race and non-race configurations, and does not change production code or JSON merge behavior. I do not see a correctness, closure, or unhappy-path issue with moving this allocator-sensitivity check behind `!race`.

PR #25916, opened 2026-07-20
fix(frontend): quote prepared statement names during migration

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25903

Prepared statement migration sends the normalized statement name and inner SQL
to the target CN. The target reconstructed textual PREPARE statements by
interpolating the normalized name without identifier quoting. A valid source
statement such as:

```sql
prepare `a-b` from select 1
```

was therefore replayed as invalid SQL:

```sql
prepare a-b from select 1
```

Quote the normalized name with the shared `sqlquote.Ident` helper before
replay. This also doubles embedded backticks, preserving the identifier and the
SQL statement boundary without changing the migration protobuf.

The regression test follows the real source parser and target migration path
for:

- ordinary names;
- names containing `-`;
- reserved words;
- names containing an embedded backtick.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s -run '^TestSession_Migrate$' ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=240s ./pkg/frontend`
- `go test -count=1 ./pkg/common/sqlquote`
- `go build ./pkg/frontend` (with the repository CGo environment)
- `go vet ./pkg/frontend` (with the repository CGo environment)

### DIFF
--- a/pkg/frontend/session.go
+++ b/pkg/frontend/session.go
@@ -2059,6 +2059,10 @@ type prepareStmtMigration struct {
 	commitFn   func(*Session, error) error
 }
 
+func quotePrepareStmtName(name string) string {
+	return "`" + strings.ReplaceAll(name, "`", "``") + "`"
+}
+
 func newPrepareStmtMigration(name string, sql string, paramTypes []byte) *prepareStmtMigration {
 	return &prepareStmtMigration{
 		name:       name,
@@ -2072,7 +2076,7 @@ func (p *prepareStmtMigration) Migrate(ctx context.Context, ses *Session) error 
 	ses.EnterFPrint(FPMigratePrepareStmt)
 	defer ses.ExitFPrint(FPMigratePrepareStmt)
 	if !strings.HasPrefix(strings.ToLower(p.sql), "prepare") {
-		p.sql = fmt.Sprintf("prepare %s from %s", p.name, p.sql)
+		p.sql = fmt.Sprintf("prepare %s from %s", quotePrepareStmtName(p.name), p.sql)
 	}
 
 	tempExecCtx := &ExecCtx{
--- a/pkg/frontend/session_test.go
+++ b/pkg/frontend/session_test.go
@@ -698,11 +698,18 @@ func TestSession_Migrate(t *testing.T) {
 			PrepareStmts: []*query.PrepareStmt{
 				{Name: "p1", SQL: `select ?`},
 				{Name: "p2", SQL: `select ?`},
+				{Name: "a-b", SQL: `select ?`},
+				{Name: "select", SQL: `select ?`},
+				{Name: "a`b", SQL: `select ?`},
+				{Name: "from", SQL: `select ?`},
 			},
 		})
 		assert.NoError(t, err)
 		assert.Equal(t, "d1", s.GetDatabaseName())
-		assert.Equal(t, 2, len(s.prepareStmts))
+		assert.Len(t, s.prepareStmts, 6)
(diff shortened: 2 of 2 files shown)


pkg/frontend/session.go
```diff
@@ -2072,7 +2073,7 @@ func (p *prepareStmtMigration) Migrate(ctx context.Context, ses *Session) error
 	ses.EnterFPrint(FPMigratePrepareStmt)
 	defer ses.ExitFPrint(FPMigratePrepareStmt)
 	if !strings.HasPrefix(strings.ToLower(p.sql), "prepare") {
-		p.sql = fmt.Sprintf("prepare %s from %s", p.name, p.sql)
+		p.sql = fmt.Sprintf("prepare %s from %s", sqlquote.Ident(p.name), p.sql)
```
XuPeng-SH: [P2] Preserve the inner SQL for names containing `from`

This branch still has the substring-based delimiter extraction in `doPrepareStmt`. For a valid source name such as ``prepare `from` from select ?``, this reconstructed SQL parses and the first migration returns success, but `doPrepareStmt` matches the `from` inside the quoted identifier and stores `PrepareStmt.Sql` as `from select ?` instead of `select ?`. `migrateConnectionFrom` then serializes that corrupted value, so the next CN migration replays ``prepare `from` from from select ?`` and fails to parse. I reproduced this by adding the `from` case to the existing full `TestSession_Migrate/ok` path and asserting the stored SQL: expected `select ?`, actual `from select ?`. Please close the repeated-migration path (for example by integrating the token-based delimiter fix) and add a regression that verifies the stored SQL or performs two migrations.

XuPeng-SH (changes requested):
Quoting fixes the first replay for ordinary quoted names, but the recreated statement is not migration-safe for a valid reserved-word name. Details are attached inline.

gouhongshen (changes requested):
## Codex automated review

Valid quoted names containing `from` cannot survive a second CN migration.

### P1 - Preserve inner SQL for names containing `from` (pkg/frontend/session.go:2076)
For Name `from` and SQL `select ?`, this builds `prepare `from` from select ?`. The first replay succeeds, but `doPrepareStmt` (pkg/frontend/mysql_cmd_executor.go:1545) finds the first `from` inside the quoted identifier and stores `Sql` as `from select ?`. `migrateConnectionFrom` serializes that corrupted field, so the next migration builds `prepare `from` from from select ?`, which cannot parse as PREPARE requires a prepareable statement after its delimiter. Fix delimiter extraction and add a two-hop migration or stored-SQL regression; the current `select` case misses this.

gouhongshen (approved):
## Codex automated review

No findings. The prior `from`-in-quoted-name regression is resolved by token-based delimiter extraction, and the migration tests cover reserved, punctuated, and embedded-backtick names. Focused tests could not run because `go` is unavailable in this environment.

XuPeng-SH (approved):
Re-reviewed at bebfd88f9efd491792d148d23442f96d1e07ea9d: no blocking findings. The prepared-statement name is quoted with sqlquote.Ident, and token-based SQL extraction preserves the inner SQL across repeated migrations, including the prior from-name regression. Verified with focused migration/extraction tests, a review-only two-hop migration test, race x20, full pkg/frontend tests, go vet, and pkg/common/sqlquote tests. CI is green.

PR #25918, opened 2026-07-20
fix(window): prevent overflow in value window offsets

### LINKED ISSUE #25906: [Bug]: LEAD with INT64_MAX offset returns wrong rows due to integer overflow
### Description

`LEAD` returns values from earlier rows instead of `NULL` when its offset is `INT64_MAX` and the current row index is greater than zero.

On latest `main` (`234a93f81ce52b4457c7595e455eca4bba3bc680`), the window operator computes the source row as:

```go
srcRow := j + int(offset)
if srcRow >= end {
    // append default or NULL
} else {
    localResult.UnionOne(srcVec, int64(srcRow), ...)
}
```

For `offset = math.MaxInt64`, `j + int(offset)` overflows when `j > 0`. The result becomes negative, bypasses the `srcRow >= end` check, and is passed to `UnionOne`. The vector access then silently reads earlier rows.

### Reproduction

Equivalent SQL shape:

```sql
select id,
       lead(id, 9223372036854775807) over (order by id)
from t;
```

With input `[10, 20, 30]`, the operator currently produces null bitmap `[0]` and physical values `[0, 10, 20]`: only row 0 is NULL, while rows 1 and 2 incorrectly read prior rows.

### Expected behavior

An offset larger than the remaining partition length is out of range for every row, so every `LEAD` result must be `NULL` (or the provided default value).

### Actual behavior

Rows after the first return data from earlier rows because of signed integer overflow.

### Impact

This is silent query-result corruption for a valid BIGINT offset. It does not return an error or panic, so callers can consume incorrect analytical results without noticing.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25906

`LEAD` previously added a BIGINT offset to the current row before checking the partition boundary. `INT64_MAX` therefore overflowed on later rows and silently read earlier values.

This PR:

- compares LAG/LEAD offsets with the available partition distance before converting or adding/subtracting them;
- applies the same bounded-index rule to NTH_VALUE positions;
- adds focused unit/race coverage for maximum int64 values;
- adds a distributed SQL regression for `LEAD(..., 9223372036854775807)`.

Validation:

- `go test ./pkg/sql/colexec/window -count=1`
- `go test -race ./pkg/sql/colexec/window -run '^(TestProcessValueFunc_LeadWithMaxInt64Offset|TestProcessValueFunc_NthValueWithMaxInt64Position)$' -count=1`
- `go vet ./pkg/sql/colexec/window`
- `go build ./pkg/sql/colexec/window`
- `git diff --check`


pkg/sql/colexec/window/window.go
```diff
@@ -545,12 +545,12 @@ func (ctr *container) processValueFunc(idx int, ap *Window, proc *process.Proces
 			if right > end {
 				right = end
 			}
-			targetRow := left + int(nthVal) - 1
-			if left >= right || targetRow >= right {
+			if left >= right || nthVal > int64(right-left) {
```
XuPeng-SH, line 548: [P2] Bound ROWS frame offsets before trusting the frame length

`buildInterval` runs before this comparison, and `buildRowsInterval` still computes a finite FOLLOWING bound as `rowIdx + int(fol)`. For `nth_value(id, 1) over (order by id rows between 9223372036854775807 following and unbounded following)` on `[10,20,30]`, the frame should be empty for every row. Instead, row 1 overflows its left bound negative, gets clamped to the partition start, and this branch returns `10` rather than `NULL`. I reproduced this with an operator-level regression: row 0 was NULL, while row 1 was non-NULL with value 10. Since this PR explicitly extends the overflow hardening to NTH_VALUE, please saturate/reject ROWS frame-bound arithmetic before `right-left` is trusted and add this case to the regression coverage.

XuPeng-SH (changes requested):
The LAG/LEAD distance checks are sound, but the NTH_VALUE branch still trusts overflowed ROWS frame bounds before applying its new bounded-index check. A concrete reproducer is attached inline.

gouhongshen (changes requested):
## Codex automated review

NTH_VALUE still has a silent wrong-result overflow path in ROWS frames.

### P1 - Bound ROWS frame offsets before trusting the NTH_VALUE frame length (pkg/sql/colexec/window/window.go:548)
`buildInterval` can return overflowed ROWS bounds before this check: `ROWS BETWEEN 9223372036854775807 FOLLOWING AND UNBOUNDED FOLLOWING` on three rows computes `rowIdx + int(following)`, which overflows for rows 1 and 2. Clamping then changes the negative bound to partition start 0, so this check sees a three-row frame and returns row 0 (10), although the frame must be empty and the result NULL. Saturate/validate ROWS-bound arithmetic before clamping and add this regression case.

pkg/sql/colexec/window/window.go
VioletQwQ-0: Fixed in 8d65d20c0. ROWS frame bounds now compare the uint64 offset with the remaining partition distance before converting to int, and saturate at the partition boundary. Added the exact NTH_VALUE regression plus table-driven coverage for large PRECEDING/FOLLOWING start/end offsets with a non-zero partition start. Validation: full window package UT, focused race test, go vet, go build, and git diff --check.

gouhongshen (commented):
## Codex automated review

The prior blocking ROWS-frame overflow is fixed on the current head, and no functional regression remains. Current CI is green; local tests were blocked by missing prebuilt CGo artifacts. One non-blocking regression-test gap remains.

### P2 - Make the NTH_VALUE test reach the overflow path (pkg/sql/colexec/window/value_window_test.go:566)
With three rows and `ROWS BETWEEN 1 PRECEDING AND CURRENT ROW`, `left` is at most 1, so the old `left + MaxInt64 - 1` calculation never overflows and this test passes before the fix. Using four rows makes `left == 2` on the last row: the old target wraps to MinInt64 and `UnionOne` can incorrectly read row 0, while the new length check returns NULL. The implementation is correct, so this is non-blocking, but the regression should exercise the bug it is intended to prevent.

gouhongshen (approved):
## Codex automated review

The earlier review on this head contains only non-blocking findings. Approving.

gouhongshen (approved):
## Codex automated review

No findings. The prior ROWS-frame overflow review thread is resolved and independently validated against the current head; bounds are checked before conversion/arithmetic. Regression coverage reaches the original LEAD, NTH_VALUE, and frame-bound cases. Local Go tests could not run because the environment has no Go toolchain; current PR CI is green.

XuPeng-SH (approved):
I checked the overflow fix from correctness / robustness / unhappy-path angles and I did not find a blocker.

The core arithmetic changes look right:
- `lead` now checks the partition-local distance before doing `j + offset`, so the overflowed wraparound path is gone.
- `nth_value` now checks `n > right-left` before computing `left + n - 1`, which removes the target-row overflow.
- `buildRowsInterval` now saturates huge `ROWS ... PRECEDING/FOLLOWING` bounds to the partition edges instead of overflowing signed `int` arithmetic.

The new unit + SQL regression coverage is enough for the bug being fixed.

Optional follow-up only: it would still be nice to add a couple more regressions for non-const huge offsets / huge `nth_value`, but I do not see that as a merge blocker here.

aunjgr (approved):
Reviewed the current head independently. The LAG/LEAD/NTH_VALUE index checks compare offsets before converting them to int, and ROWS frame bounds saturate against partition limits before arithmetic. Boundary tests cover MaxInt64 offsets and overflowing FOLLOWING frames. I found no merge-blocking correctness, lifecycle, or compatibility issue.

XuPeng-SH (approved):
Deep re-review on exact head 7ea630b83b: no blocking findings.

I rechecked the full arithmetic closure rather than relying on the prior approval. LAG/LEAD compare partition-local distance before int conversion/arithmetic; NTH_VALUE checks frame length before computing the target row; all four finite ROWS-bound directions saturate at partition edges before signed arithmetic.

Fresh validation:
- full pkg/sql/colexec/window tests: pass
- focused overflow tests under race, count=10: pass
- full window package under race: pass
- go build and go vet: pass
- exact-head SQL with constant and non-constant offsets, defaults, non-zero partition starts, MaxInt64 and MaxUint64, plus huge PRECEDING/FOLLOWING frames: correct

I also ran the same SQL shapes on the exact PR base. The base returned earlier rows for huge LEAD, FIRST_VALUE empty-frame, and four-row NTH_VALUE cases, while the PR head returned NULL/default as required. This confirms the regressions exercise real pre-fix failures. Q1-Q3 review found no new resource, wait, or growth risks.

heni02 (approved):

PR #25919, opened 2026-07-20
fix(frontend): locate PREPARE delimiter by token

### LINKED ISSUE #25900: [Bug]: PREPARE stores invalid inner SQL when statement name contains from
### Description

`doPrepareStmt` locates the `FROM` delimiter with a substring search over the original SQL. When the prepared statement name itself contains `from`, the search matches the identifier instead of the delimiter and stores an invalid inner SQL string.

On the latest `main` commit `234a93f81ce52b4457c7595e455eca4bba3bc680`, this valid statement:

```sql
prepare fromx from select 1;
```

is parsed successfully, but the resulting frontend `PrepareStmt.Sql` is `from select 1` instead of `select 1`.

### Root cause

`pkg/frontend/mysql_cmd_executor.go` calculates the inner-SQL offset using:

```go
idx := strings.Index(strings.ToLower(sql[:(len(st.Name)+20)]), "from") + 5
originSql := strings.TrimLeft(sql[idx:], " ")
```

The first `from` in `prepare fromx from select 1` belongs to the statement name `fromx`.

### Expected behavior

The prepared statement stores and compiles the inner SQL `select 1`, regardless of whether the statement name contains the characters `from`.

### Actual behavior

The prepared statement stores `from select 1`.

### Impact

Valid SQL can produce a corrupted prepared-statement SQL string. That string is passed to compile and retained for later prepared-statement reuse/rebuild paths, making the stored SQL inconsistent with the already parsed AST.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25900

`doPrepareStmt` previously searched for the first `from` substring in the original SQL. A statement name such as `fromx` was mistaken for the delimiter, so the cached inner SQL became `from select 1`.

This PR:

- locates the PREPARE delimiter using the MySQL scanner and the session SQL mode;
- preserves the original inner SQL text, including comments, instead of rebuilding it from the AST;
- covers substring-containing names, quoted names, ANSI_QUOTES, comments, whitespace, and invalid helper inputs;
- adds an end-to-end distributed PREPARE/EXECUTE regression.

Validation:

- `go test -race ./pkg/frontend -run '^(TestHandlePrepareStmtNameContainingFrom|TestExtractPrepareStmtSQL|TestExtractPrepareStmtSQLRejectsInvalidInput)$' -count=1`
- `go vet ./pkg/frontend`
- `go build ./pkg/frontend`
- `git diff --check`


pkg/frontend/mysql_cmd_executor.go
```diff
@@ -1541,10 +1542,28 @@ func handleExplainStmt(ses FeSession, execCtx *ExecCtx, stmt *tree.ExplainStmt)
...
+		return "", moerr.NewInvalidInput(ctx, "invalid PREPARE statement")
+	}
+	if token, _ := scanner.Scan(); token == mysql.EofChar() || token == mysql.LEX_ERROR {
+		return "", moerr.NewInvalidInput(ctx, "invalid PREPARE statement name")
+	}
+	if token, _ := scanner.Scan(); token != mysql.FROM {
+		return "", moerr.NewInvalidInput(ctx, "invalid PREPARE statement delimiter")
+	}
+
+	return strings.TrimLeftFunc(sql[scanner.Pos:], unicode.IsSpace), nil
```
XuPeng-SH: [P2] Handle a delimiter inside an executable comment

`Scanner.Scan` treats tokens inside MySQL executable comments as SQL. For `prepare fromx /*! from */ select 1`, the parser returns a `*tree.PrepareStmt`, and the third scan returns `FROM` while `scanner.Pos` is still before the closing `*/`. This slice therefore yields `*/ select 1`; a focused full `handlePrepareStmt` test stores that value instead of `select 1`. That leaves the reported `fromx` failure unresolved for executable-comment syntax already supported by this parser, and the malformed text is then passed to compile and SQL-recording paths. Please account for the executable-comment boundary when deriving the inner SQL and add this case as a regression test.

XuPeng-SH (changes requested):
The token-based approach fixes the reported `fromx` case, but one supported token layout still stores malformed inner SQL. Details are attached inline.

gouhongshen (changes requested):
## Codex automated review

The token-based extraction fixes the reported basic case, but remains incorrect for valid MySQL executable-comment syntax.

### P2 - Account for executable-comment boundaries after FROM (pkg/frontend/mysql_cmd_executor.go:1559)
For valid input `prepare fromx /*! from */ select 1`, `Scan` returns the `FROM` token inside the executable comment with `scanner.Pos` immediately after `from`, before the closing `*/`. The parser subsequently consumes that terminator and builds the expected `select 1` AST, but this slice produces `*/ select 1`. Consequently the prepared statement stores malformed SQL that can reach recording and rebuild paths. Derive the source offset past the executable-comment boundary and add an end-to-end regression for this layout.

pkg/frontend/mysql_cmd_executor.go
VioletQwQ-0: Fixed in e33ff0b0a. When FROM is tokenized inside a MySQL executable comment, extraction now removes that comment boundary and preserves executable SQL on either side. Added the exact full-handler regression, helper coverage for SQL inside the executable comment, and the unterminated-comment error path. Validation: full frontend package UT, go vet, go build, and git diff --check.

gouhongshen (changes requested):
## Codex automated review

The reported delimiter bug and simple executable-comment case are fixed, but executable-comment boundary detection still corrupts valid inner SQL containing `*/` inside a quoted string.

### P2 - Find the executable-comment terminator lexically (pkg/frontend/mysql_cmd_executor.go:1561)
When FROM is tokenized inside `/*! ... */`, this uses `strings.Index` to find the first `*/` bytes. The parser's scanner instead consumes quoted strings while `CommentFlag` is set, so `prepare fromx /*! from select 'x*/y' */` is accepted with the real terminator at the end. This code stops at the `*/` inside the string and constructs malformed SQL such as `select 'x y' */`, causing the stored/precompiled statement to diverge from the parsed AST or fail preparation. Derive the boundary using scanner state and add a regression test.

VioletQwQ-0:
@gouhongshen @XuPeng-SH Addressed the executable-comment boundary CR in 25226fb1b. The frontend no longer searches raw bytes for `*/`; the MySQL scanner now records the first executable-comment terminator it actually crosses, so quoted `*/` bytes remain part of the prepared SQL. Added the exact `prepare fromx /*! from select 'x*/y' */` full-handler regression, helper coverage that preserves a following ordinary comment, scanner boundary/pool-reset tests, and retained the unterminated-comment rejection. Validation: full frontend UT, full parser UT, focused scanner race, go vet/build for frontend and mysql parser, git diff --check, historical-CR preflight, and mo-self-review.

gouhongshen (approved):
## Codex automated review

No findings. The token-based extraction fixes the prior substring bug and correctly handles executable-comment boundaries, including quoted `*/`; existing review findings are addressed. Focused tests could not run locally because `go` is unavailable, but CI is passing.

heni02 (approved):

XuPeng-SH (approved):
Re-reviewed the complete diff on `25226fb1`. The original delimiter bug and both executable-comment follow-ups are fixed. Scanner event state is reset across pool reuse, extraction terminates on every invalid path, and extracted SQL remains AST-equivalent across versioned comments, comment-boundary splits, quoted `*/`, ANSI_QUOTES, and multiple executable comments. Focused and full package tests, race stress, vet/build, and required CI checks pass.

Issue #25739, opened 2026-07-15
[Bug] CN Close stops cleanup after the first subsystem error


Ariznawlll:
## Summary

CN service shutdown returns immediately when any individual close step fails. Resources later in the sequence are left unclosed, which can leak listeners, clients, locks, goroutines, and local service state during graceful shutdown or test teardown.

## Affected revision

- Branch: `test/issues-24122-21685-4.0-dev`
- Commit: `7f7de555c49acc3a64995b078025f5526afc9c34`

## Root cause

`pkg/cnservice/server.go:319-360` performs sequential cleanup with immediate `return err` after each operation: BootstrapService, Frontend, Task service, internal RPCs, I/O pipeline, Gossip, pipeline server, LockService, ShardService, and PipelineClient.

For example, if `bootstrapService.Close()` fails, `stopFrontend()` and every subsequent cleanup step are skipped. If Gossip leave fails, the pipeline server and lock service remain open.

TN shutdown already attempts independent cleanup operations and aggregates errors with `errors.Join`.

## Actual behavior

One close failure prevents unrelated resources from being released. Shutdown and restart behavior depends on which closer fails first.

## Expected behavior

Attempt all applicable cleanup steps while preserving required ordering, and aggregate returned errors. One subsystem failure must not suppress the remaining cleanup.

## Suggested regression test

Inject failures into early and middle close operations and assert all later closers are invoked exactly once, with the returned error containing every failure.

## Relevant code

- `pkg/cnservice/server.go:319-360`
- Reference pattern: `pkg/tnservice/store.go:240-269`

jiangxinmeng1:
fixed by #25924

Ariznawlll:
## 最新 main 验证结果：通过

**测试版本**

- MatrixOne：`main@ebd05ba52f0d08afb4225e7b4b2e97e8030c057d`
- 修复合入 commit：`e6f5587cfbe7794136a45408c6e91682dafd950d`

**测试步骤**

执行以下关闭流程定向用例：

1. `TestCloseCNServiceStepsAttemptsAllAndAggregatesErrors`
2. `TestServiceCloseDoesNotHangOnNeverReadyClusterAfterEarlyError`
3. `TestStopTaskStopsRunnerAfterHolderCloseFailure`
4. `TestMOServerStopCompletesCleanupAfterListenerCloseError`

重点检查前置关闭步骤报错后，后续资源清理是否仍会执行，以及多项错误是否被聚合返回。

[...]

PR #25924, opened 2026-07-21
fix cn shutdown cleanup aggregation

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25739

  - service.Close() now attempts all shutdown steps in the original order and aggregates
    errors with errors.Join.

  - stopFrontend() now still calls cancelMoServerFunc() even if frontend shutdown fails.
  - stopRPCs() now closes all applicable RPC-related resources and aggregates errors.
  - stopTask() now stops the task runner even if task holder close fails, and aggregates
    both errors.


gouhongshen (approved):
## Codex automated review

No actionable regressions found. The change preserves shutdown order, attempts all cleanup steps, aggregates errors correctly, and improves nil/error handling. Existing PR discussion contained no substantive reviews or inline feedback. Local focused tests were blocked by missing cgo/libmo.dylib, but the PR head's Ubuntu unit tests, coverage, SCA, and exercised E2E checks pass.

XuPeng-SH (changes requested):
Requesting changes for two shutdown-closure blockers:

1. `pkg/cnservice/server.go:424`: `stopRPCs` calls `moCluster.Close()`, whose implementation waits on `readyC` before stopping its refresh task. If the first HAKeeper refresh keeps failing, `readyC` is never closed. With this PR, an earlier frontend/task close error no longer returns immediately and instead reaches this wait, so `service.Close()` hangs forever and no later cleanup or aggregated error is produced. A minimal failing-refresh test on head 6506597501 confirmed that `Close` was still blocked after 200 ms while the refresh loop kept retrying. Please make cluster shutdown release/cancel the pre-ready wait and add a regression covering early closer failure plus a never-ready cluster.

2. `pkg/cnservice/server.go:411-415`: calling `cancelMoServerFunc()` after `MOServer.Stop()` fails is not sufficient to finish frontend cleanup. `MOServer.Stop` sets `running=false`, returns immediately on any listener-close error, and therefore skips `rm.cancelCtx()`, `wg.Wait()`, and `rm.killNetConns()`. The outer cancellation does not interrupt a socket already blocked in `Accept` or `rs.Read`; an idle client connection can therefore retain its goroutine/socket until a long or disabled read timeout, and a later `Stop` is a no-op because `running` is already false. Please make `MOServer.Stop` attempt its full cleanup sequence while aggregating listener errors, with a regression for a listener-close error and an active idle connection.

gouhongshen (approved):
## Codex automated review

No actionable regressions found. The current head addresses both prior blockers, preserves shutdown ordering, completes later cleanup after errors, and aggregates failures correctly. Cluster lifecycle race tests and vet passed; CGo-dependent local tests were unavailable because libmo.dylib is absent, while current-head CI is green.

aptend (changes requested):
Request changes on exact head 35e1e1b0ad: one deterministic test-isolation failure remains.

### P2 — Initialize the frontend ParameterUnit before constructing RoutineManager

TestMOServerStopCompletesCleanupAfterListenerCloseError calls NewRoutineManager(ctx, "") before registering a ParameterUnit. NewRoutineManager immediately reads the service-scoped global PU through getPu, so the test panics when run by itself:

panic: interface conversion: interface {} is nil, not *config.ParameterUnit

Reproduction:

  go test ./pkg/frontend -run "^TestMOServerStopCompletesCleanupAfterListenerCloseError$" -count=1

The same failure reproduces under -race. The full frontend package happens to pass because another test initializes the shared global first, making this regression test order-dependent and unusable in focused or shuffled validation.

Please construct the ParameterUnit, call SetDefaultValues, and register it with setPu before NewRoutineManager (ideally using a test-specific service key).

Relevant code: https://github.com/matrixorigin/matrixone/blob/35e1e1b0ad2857a1775ab09ee0c7f27954444fd1/pkg/frontend/server_test.go#L41-L53

No production correctness blocker was found: the affected package suites and go vet pass, and the other new lifecycle tests pass under -race -count=5.

aptend (approved):
Re-reviewed exact head `c7c337f797`. The previous frontend test-isolation issue is fixed, and the shutdown cleanup closure is complete at the CN level: all phases are attempted in order, errors are aggregated, and listener/cluster/task failures no longer prevent downstream cleanup.

Validation passed: all affected package tests, focused shutdown regressions under `-race -count=10 -shuffle=on`, `go build`, `go vet`, and `git diff --check`. No remaining code-review findings. The unrelated TAE `TestAppendAndGC2` CI failure should still be rerun or resolved before merge.

XuPeng-SH (approved):
Re-reviewed exact head `b75e2f5b4b0cbd10dccdebf6030a9cccb44490c8`. No remaining actionable findings.

The prior blockers are closed: cluster shutdown releases pre-ready waiters after stopping refresh tasks; frontend listener-close errors no longer skip context cancellation, waiter completion, or connection cleanup; task holder failure no longer suppresses runner shutdown; and the frontend regression test is service-scoped and standalone-safe. CN-level shutdown order is preserved, independent cleanup phases are all attempted, and returned errors are aggregated.

Local validation passed on this exact head: focused shutdown regressions, the complete `clusterservice`, `cnservice`, and `frontend` package suites, targeted `-race`, `go build`, `go vet`, and `git diff --check`. Current-head required CI is green.

Issue #25921, opened 2026-07-21
[Bug]: CDC watermark INSERT failures are reported as success and cached as committed


Ariznawlll:
### Component

CDC watermark updater (`pkg/cdc/watermark_updater.go`)

### Description

When a new CDC watermark row is absent and the INSERT fails, `execAddWM` still marks the watermark as committed in memory and completes the waiting job with a successful result.

Consequently, `GetOrAddCommitted` observes `nil` error even though the catalog INSERT failed, and later calls can hit `cacheCommitted` as if the watermark were durable.

### Reproduction

Use an internal executor that:

1. returns a valid empty result for the watermark SELECT;
2. returns an injected error for the watermark INSERT.

Submit a `JT_CDC_GetOrAddCommittedWM` job through the production `onJobs` path and assert that the job returns the INSERT error and that the key is absent from `cacheCommitted`.

```bash
go test ./pkg/cdc \
  -run '^TestAuditAddWatermarkFailureIsReturnedAndNotCached$' \
  -count=3 -v
```

All three runs fail both assertions:

```text
cdc.watermark.read_error: injected watermark insert failure
An error is expected but got nil
Should be false  // key is present in cacheCommitted
```

### Expected behavior

- The waiting `GetOrAddCommitted` job receives the catalog INSERT error.
- A watermark is added to `cacheCommitted` only after its INSERT succeeds.
- CDC table-stream initialization stops/retries when the initial watermark cannot be persisted.

### Actual behavior

- The INSERT error is only logged by `onJobs`.
- The waiting job is completed with `DoneWithResult`, so the caller receives success.
- The failed watermark is written to `cacheCommitted` and becomes a successful fast-path result.

### Root cause

`execAddWM` performs the cache update and successful job completion unconditionally after `u.ie.Exec`, even when `err != nil`:

```go
err = u.ie.Exec(ctx, addSql, ie.SessionOverrideOptions{})
...
for i, job := range u.addCommittedBuffer {
    u.cacheCommitted[*job.Key] = job.Watermark
    job.DoneWithResult(job.Watermark)
    ...
}
```

The buffer is then cleared, so the deferred `resetJobs(err)` in `onJobs` can no longer propagate the failure to those jobs.

### Impact

CDC can continue initialization while its initial watermark is not durable. Catalog failures are hidden from the caller, subsequent reads incorrectly report the watermark as committed, and a restart can reinitialize from an older or missing row and duplicate processing.

[...]

PR #25925, opened 2026-07-21
fix cdc watermark add failure handling

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25921

- Updated pkg/cdc/watermark_updater.go: execAddWM now returns immediately when the
    catalog INSERT fails, so it no longer writes the watermark into cacheCommitted. The
    existing onJobs cleanup path then propagates the error to the waiting job.

  - Updated pkg/cdc/watermark_updater_test.go: added
    TestAuditAddWatermarkFailureIsReturnedAndNotCached, covering the case where SELECT
    returns no watermark row, INSERT fails with an injected error, the job receives that
    error, and the failed watermark is not cached.


gouhongshen (approved):
## Codex automated review

The INSERT error now returns before cache mutation or success notification; deferred cleanup propagates the original error and clears affected jobs. The regression test covers both error propagation and cache integrity. Existing PR discussions contained no substantive prior findings. CI is green; local tests were blocked by missing prebuilt CGo dependencies.

Issue #22151, opened 2025-07-11
[Feature]:  implement RIGHT DEDUP join to reduce memory usage for big insert


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

2.2-dev

### Commit ID

6f28646defdbfcbafad97860c5990a9d88806825

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

job：https://github.com/matrixorigin/mo-nightly-regression/actions/runs/16192779084/job/45712431366

<img width="1404" height="664" alt="Image" src="https://github.com/user-attachments/assets/309f11e5-9ad1-4bff-90c6-acbd70e22bba" />

log：https://grafana.ci.matrixorigin.cn/explore?panes=%7B%229Hi%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-big-data-nightly-6f28646-20250710%5C%22%7D%20%7C%3D%20%60%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221752150604000%22,%22to%22:%221752151219000%22%7D%7D%7D&schemaVersion=1&orgId=1

### Steps to Reproduce

```Markdown
trigger big data regression
```

Ariznawlll:
<img width="943" height="872" alt="Image" src="https://github.com/user-attachments/assets/45be448c-5980-489d-b058-44cb3f55ebd0" />

LeftHandCold:
两个cn oom了

<img width="1680" height="329" alt="Image" src="https://github.com/user-attachments/assets/842b4557-6acc-4de0-8bb9-15d780867ba6" />

XuPeng-SH:
<img width="1423" height="400" alt="Image" src="https://github.com/user-attachments/assets/0ba37da5-d910-4334-873b-9e3c4c36ce2d" />



***Seems the plan is different.***

***New Plan(OOM)***
```shell
mysql> explain analyze INSERT INTO test_200w (SELECT * FROM big_data_test limit 2000000);
+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #25936, opened 2026-07-21
fix(plan): bound right dedup memory

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #22151

RIGHT DEDUP builds the target map and then inserts every incoming key while probing. That probe-side growth cannot transition into the current build-side spill engine, so large insert-select statements can retain tens of GiB and OOM a CN.

This PR:

- estimates RIGHT DEDUP's retained hash-map allocation using the runtime table's capacity and load-factor rules;
- accounts for target plus incoming keys and sums chained primary/unique-key DEDUP maps;
- falls back the entire chain to normal DEDUP when the estimate reaches the configured join spill threshold or default CN-relative stage budget;
- treats missing, non-finite, and sentinel default statistics conservatively;
- preserves the row-count semantics of small `join_spill_mem` values;
- keeps FLOAT normal DEDUP unshuffled because FLOAT hash shuffle is not supported yet, avoiding an invalid plan while retaining single-CN spill.

The default combined budget is approximately one eighth of CN memory after the file-cache reservation. Normal DEDUP hash-shuffles supported keys; FLOAT keys stay unshuffled and spill locally.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/sql/plan`
- `go build -mod=readonly ./pkg/container/hashtable`
- `go vet -mod=readonly ./pkg/container/hashtable`
- `go test -count=1 -timeout=120s ./pkg/container/hashtable`
- `../mo-tester/run.sh -n -g -p test/distributed/cases/operator/in_range_operator.sql` (368/368 passed)

Design decision: this selects the already-spillable normal DEDUP implementation for memory-unsafe plans instead of attempting to switch RIGHT DEDUP after it has streamed rows downstream. RIGHT DEDUP remains a small-input optimization; absent/default statistics take the safe fallback. FLOAT shuffle support is intentionally deferred to a separate PR.


XuPeng-SH (approved):
Deep review completed against f860ce293fb80f0b25d8b95b4a76d3300e23d905.

Verified the full closure of the change:
- RIGHT DEDUP memory accounting matches the executor hash-map type selection, load factors, growth steps, and chained-map coexistence. Unknown/non-finite/overflow estimates conservatively fall back to normal DEDUP.
- The fallback runs before child swapping and is followed by stats/shuffle recomputation, so it does not leave partially converted physical state.
- FLOAT DEDUP keys stay off the runtime hash-shuffle path, which does not support FLOAT.
- Broadcast workers still merge shared matched/capture state; shuffle workers finalize and clean their disjoint partitions independently, including Reset/reuse paths.

Local verification (CGo-transitive wrapper): full and -race tests passed for ./pkg/container/hashtable, ./pkg/sql/colexec/dedupjoin, and ./pkg/sql/plan. git diff --check also passed. No blocking issue found.

XuPeng-SH:
Follow-up: I also completed real SQL execution against the exact reviewed head `f860ce293fb80f0b25d8b95b4a76d3300e23d905`, using an isolated locally built `mo-service` and data generated by `generate_series`.

Results:
- Chained PK + UK, 4,000 target rows + 10,000 input rows, `join_spill_mem=50000`: verbose/physical plans showed two chained `RIGHT DEDUP` / `right dedup join` operators. Execution produced 14,000 rows, 14,000 distinct PKs, and 14,000 distinct UKs.
- Same shape with `join_spill_mem=25000`: the whole chain fell back to two normal `DEDUP` operators. Execution produced the same 14,000/14,000/14,000 result.
- BIGINT PK, 4,000 target rows + 400,000 input rows, forced fallback: physical plan used 10 hash-shuffle partitions and 10 independent dedup workers. Final result was 404,000 rows and 404,000 distinct PKs, confirming no partition-finalize output was lost.
- DOUBLE PK with the same 400,000-row input: plan remained non-shuffled, executed without panic, and produced 404,000 rows / 404,000 distinct PKs.
- Mixed input containing one existing PK: both RIGHT DEDUP and fallback DEDUP returned the expected duplicate-key error; both target tables remained at 4,000 rows with zero partial new rows.

One observed conservative behavior: small or not-yet-flushed target/index tables can still present `DefaultStats`, so the guard intentionally falls back to normal DEDUP until non-default table stats are available. This is a performance choice, not a correctness issue. No blocker found in the real execution pass.

XuPeng-SH:
Follow-up byte-threshold validation on exact head f860ce293f (correcting the earlier row-threshold-only experiment):

- Built real data with generate_series: target = 4,000 rows, source = 10,000 rows, with BIGINT PK plus VARCHAR UNIQUE key.
- Flushed target, hidden unique-index table, and source, then ANALYZE. EXPLAIN VERBOSE reported exact outcnt 4,000 / 4,000 / 10,000.
- For 14,000 resident keys per map, the runtime-aligned estimators give 512 KiB (int map) + 1 MiB (string map) = 1,572,864 bytes combined.
- join_spill_mem = 1,572,864 produced two RIGHT DEDUP joins.
- join_spill_mem = 1,572,863 (one byte lower) produced two normal DEDUP joins.

This confirms the planner uses the combined resident-map byte estimate and switches the whole chained dedup stage at the exact strict-greater-than memory boundary. I also reran the estimator, combined-map planner, ShouldSpill, and hashbuild memory-threshold spill tests; all passed.

Scope note: this validates the OOM-prevention planning decision without intentionally exhausting machine memory; it is not a literal OS-OOM reproduction.

Issue #24783, opened 2026-06-02
[Bug]: ROW_COUNT() not updated inside stored procedures / after CALL


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main, 4.0-dev, 3.0-dev

### Commit ID

198f1ad65

### Other Environment Information

- Follow-up to #24767 (ROW_COUNT() always returns 0).

### Actual Behavior

After #24767 is fixed for the normal foreground path, `ROW_COUNT()` is still
incorrect in two related cases:

1. **Inside stored procedures.** Stored-procedure bodies run through the PL/SQL
   interpreter -> background executor (`NullResp`). Affected rows for those
   statements are not recorded, so `ROW_COUNT()` used inside a procedure does
   not reflect the procedure's previous statement.

2. **After `CALL proc()`.** The `CALL` path does not propagate the real
   affected rows of the procedure's last statement back to the caller session,
   so `ROW_COUNT()` after a `CALL` does not return the expected value.

### Expected Behavior

`ROW_COUNT()` should follow MySQL semantics inside stored procedures and after
`CALL` (note: MySQL's exact value after `CALL` is itself underspecified and
should be confirmed against a reference server while implementing).

### Steps to Reproduce

1. Create a stored procedure that performs a DML and then `SELECT ROW_COUNT()`.
2. `CALL` it; observe `ROW_COUNT()` does not reflect the DML's affected rows.

### Additional information

Identified during review of the #24767 fix (PRs #24768 / #24769 / #24770).
The foreground path (single statement, multi-statement COM_QUERY, binary
protocol) is handled by those PRs. This issue tracks the
PL/SQL-interpreter / background-executor path, which is a separate subsystem
and carries higher regression risk, so it is split out.

heni02:
main分支已修复，bvtcase已覆盖，closed

PR #25944, opened 2026-07-21
fix: propagate ROW_COUNT() through stored procedures

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Performance
- [ ] Test and CI
- [ ] Infrastructure
- [ ] Other

issue #24783

`ROW_COUNT()` did not observe DML affected rows executed by stored procedure
interpreters, and `CALL` did not propagate the procedure's final affected-row
count back to the caller or its MySQL OK packet.

This change carries affected rows through background execution, snapshots the
last body-statement result in the SQL and Starlark interpreters, and propagates
the final value through `CALL`. It also keeps nested procedures on the shared
background transaction and adds unit, protocol, and BVT coverage.

Validation:

- `mo-cgo-test -count=1 -timeout=240s ./pkg/frontend`
- `mo-cgo-test -count=1 -timeout=300s -run '^TestRowCountOverMySQLProtocol$' ./pkg/embed`
- targeted frontend race tests
- `go vet ./pkg/frontend ./pkg/embed`
- `make -j8 build`
- `row_count.sql`: 83/83 passed


gouhongshen (changes requested):
## Codex automated review

Two correctness gaps remain in affected-row propagation.

### P1 - Preserve CALL status after result sets (pkg/frontend/result_row_stmt.go:434)
When CALL produces result sets, this branch creates ResultResponses with affectedRows=0 and emits no final OK/status response using execCtx.runResult. Therefore a procedure such as SELECT ...; UPDATE ... returns the result set but the client receives 0 instead of the final UPDATE count. MySQL requires CALL status in addition to procedure result sets: [CALL documentation](https://dev.mysql.com/doc/refman/8.4/en/call.html).

### P2 - Record failed Starlark SQL as -1 (pkg/frontend/starlark_interpreter.go:231)
backExec.Exec now stores -1 on failure, but this early return skips recordAffectedRows. Since doInterpretCall later reads interpreter.lastAffectedRows, a handled final mo.sql error leaves the previous successful count visible to CALL and ROW_COUNT().

gouhongshen (approved):
## Codex automated review

No actionable findings. The earlier protocol-final-OK and Starlark failure-state findings are addressed at this head. Focused Go tests were blocked locally by missing cgo dependency xxhash.h.

pkg/frontend/plsql_interpreter.go
```diff
@@ -505,6 +526,7 @@ func (interpreter *Interpreter) interpret(stmt tree.Statement) (SpStatus, error)
 		for _, v := range st.Variables {
 			(*interpreter.varScope)[len(*interpreter.varScope)-1][v] = value
 		}
+		interpreter.setAffectedRows(0)
```
XuPeng-SH: **[P1] Preserve ROW_COUNT across procedural-only DECLARE/local SET.** This line and the local-variable branch at line 559 force the count to zero even though neither operation executes a SQL statement that replaces ROW_COUNT(). In MySQL 8.4, `INSERT` followed by a nested `DECLARE` still exposes the INSERT count, and a procedure ending with `SET local_var = ...` returns the preceding DML count after `CALL`. A five-row INSERT followed by local SET returns 5 in MySQL but 0 here. Please remove both resets and add DML→DECLARE, DML→local SET, and final-local-SET-after-CALL tests. The `SET @user_var` path should continue recording zero because that is an executed SQL SET statement.

pkg/frontend/back_exec.go
```diff
@@ -509,11 +528,28 @@ func doComQueryInBack(
 		if err != nil {
 			return err
 		}
+		backSes.lastAffectedRows = affectedRowsForStatement(execCtx)
```
XuPeng-SH, line 536: **[P1] Do not discard result sets returned by a nested CALL.** `handleCallProcedure` puts the inner procedure's results in `execCtx.results`, but this background loop copies only its affected-row count; nothing transfers those results into `backSes.allResultSet`. Consequently an outer procedure with `SELECT 10; CALL inner(); SELECT 30`, where `inner` returns `SELECT 20; SELECT 21`, exposes only 10 and 30 instead of MySQL's ordered 10,20,21,30 result sets. Please append nested CALL results to the parent accumulator in order and add a nested multi-result BVT plus protocol test.

pkg/frontend/authenticate.go
```diff
@@ -11634,17 +11635,19 @@ func doInterpretCall(ctx context.Context, ses FeSession, call *tree.CallStmt, bg
 		}
 		defer freeStatements(stmt)
 
-		err = interpreter.ExecuteSp(stmt[0], dbName)
+		err = interpreter.ExecuteSp(stmt[0], dbName, bg)
```
XuPeng-SH, line 11646: **[P2] Seed the procedure with the caller's current affected-row state.** The newly created background executor/interpreter starts at zero, and the internal `USE`/`BEGIN` setup also records zero before the body runs. Therefore `INSERT` of six rows followed by a procedure whose first statement is `SELECT ROW_COUNT()` returns 0 here; MySQL 8.4 returns 6. The same loss occurs across a nested CALL. Please capture the caller's value before metadata/setup work and restore it into both the interpreter and backing executor immediately before SQL or Starlark body evaluation, with top-level and nested entry tests.

XuPeng-SH (changes requested):
Three correctness gaps remain in the ROW_COUNT()/CALL propagation path. I verified the expected behavior against MySQL 8.4 and traced each failure through the current head. The focused frontend tests and current non-skipped CI checks pass, but they do not cover these procedure-boundary cases. Please address the three inline findings and add the corresponding SQL/BVT and protocol coverage.

pkg/embed/cluster_test.go
```diff
@@ -228,6 +228,48 @@ func TestRowCountOverMySQLProtocol(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, int64(2), affectedRows)
+			require.NoError(t, stmt.QueryRowContext(ctx).Scan(&rowCount))
+			require.Equal(t, int64(2), rowCount)
+
+			_, err = conn.ExecContext(ctx, "create procedure caller_count() 'begin select row_count(); end'")
+			require.NoError(t, err)
+			_, err = conn.ExecContext(ctx, "insert into t values (6), (7), (8), (9), (10), (11)")
+			require.NoError(t, err)
+			rows, err := conn.QueryContext(ctx, "call caller_count()")
```
XuPeng-SH: [P1] Keep the new protocol test lint-clean

This test opens two *sql.Rows values, but neither lifetime is protected by a deferred Close, and this first result path never checks rows.Err() after iteration. The current SCA run fails here with rowserrcheck plus two sqlclosecheck errors (the two explicit closes at lines 250 and 271), so most downstream jobs are skipped. Please add a deferred close for each successful query and check rows.Err() after consuming each result set (scoping the two queries separately is also fine).

XuPeng-SH (changes requested):
Requesting changes because the new MySQL-protocol regression test does not pass repository static analysis. The current SCA job fails on one unchecked rows.Err() path and two *sql.Rows lifetimes that are not closed with defer; this failure also causes much of the downstream CI matrix to be skipped.

I re-reviewed the functional closure at c497730: the previous gaps around procedural-only statements, nested CALL result propagation, and caller ROW_COUNT seeding are fixed. Resource/transaction ownership and failure-state propagation did not reveal another blocker.

Fresh local validation on this exact HEAD:
- go build ./pkg/frontend ./pkg/embed
- go vet ./pkg/frontend ./pkg/embed
- go test -count=1 -timeout=300s ./pkg/frontend
- affected-row / nested-CALL focused tests with -race -count=10
- go test -count=1 -run TestRowCountOverMySQLProtocol ./pkg/embed
- git diff --check

All of those pass; the requested change is the confirmed repository-SCA failure described inline.

gouhongshen (approved):
## Codex automated review

No actionable findings. Reviewed the full merge-base diff, all existing reviews/issue comments/inline threads, and the resolved fixes at 8067437. Required CI, including SCA and BVT, is successful. Local focused Go tests could not run because `go` is unavailable in this environment.

XuPeng-SH (approved):
Re-reviewed latest head 8067437. The previous correctness and SCA findings are resolved:

- procedural-only statements preserve the prior ROW_COUNT, while executable SET statements record their own result
- nested CALL result sets are propagated in order
- procedure entry inherits the caller ROW_COUNT, including nested calls
- protocol result sets are closed and rows.Err is checked

Also verified failure propagation, transaction/result-set ownership, and final CALL OK affected rows. Local verification passed: full pkg/frontend tests; targeted tests x5; targeted race tests x3; TestRowCountOverMySQLProtocol; git diff --check. All current CI checks are green.

heni02 (approved):

PR #25947, opened 2026-07-21
feat(iceberg): port Iceberg connector to 4.2-dev

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23359 

feat(iceberg): port Iceberg connector to 4.2-dev


fengttt (approved):

gouhongshen (changes requested):
## Codex automated review

Found two blocking regressions: broad CALL interception breaks existing stored procedures, and access registration can leave a partially committed authorization state.

### P1 - Match only supported Iceberg built-in procedure names (pkg/frontend/iceberg_call.go:400)
This prefix check claims every unqualified procedure beginning with `iceberg_` before `doInterpretCall` performs the normal stored-procedure lookup. Thus an existing procedure such as `iceberg_backup` worked on the base branch but is now routed to `OperationForProcedure` and rejected as unsupported, even when Iceberg is disabled. Use an exact allowlist for the four built-ins so unrelated stored procedures retain prior behavior.

### P1 - Commit principal and residency registration atomically (pkg/frontend/iceberg_call.go:265)
The principal upsert executes before this residency-policy upsert without an explicit transaction. Background sessions use autocommit, so if this second statement fails, the call returns an error but the principal mapping remains committed. This has a deterministic trigger: policy validation accepts a region longer than the DDL's `varchar(128)`, causing only the second insert to fail. Under an existing cluster allow policy, runtime treats the missing account policy as optional, so the partial state can authorize the newly mapped external principal. Wrap both writes in one BEGIN/commit-or-rollback transaction and test rollback when the second write fails.

aunjgr (changes requested):
I found two independent merge blockers on the current head.

1. `pkg/frontend/iceberg_call.go:395`: `isIcebergBuiltinProcedure` claims every unqualified procedure whose name starts with `iceberg_`. That interception happens before normal stored-procedure resolution, so an existing user procedure such as `iceberg_backup` is routed into Iceberg parsing and rejected as unsupported—even with the feature disabled. Match the exact built-in procedure set, not the prefix.

2. `pkg/frontend/iceberg_call.go:250-280`: access registration performs the principal-map upsert and residency-policy upsert as two autocommit statements. If the second statement fails (for example, a value passes policy validation but exceeds a storage column limit), the call returns an error with the principal mapping already committed. Authorization can then observe a partially registered identity. Execute both mutations in one explicit transaction with rollback on either error and add a second-write-failure regression.

iamlinjunhong:
> ## Codex automated review
> Found two blocking regressions: broad CALL interception breaks existing stored procedures, and access registration can leave a partially committed authorization state.
> 
> ### P1 - Match only supported Iceberg built-in procedure names (pkg/frontend/iceberg_call.go:400)
> This prefix check claims every unqualified procedure beginning with `iceberg_` before `doInterpretCall` performs the normal stored-procedure lookup. Thus an existing procedure such as `iceberg_backup` worked on the base branch but is now routed to `OperationForProcedure` and rejected as unsupported, even when Iceberg is disabled. Use an exact allowlist for the four built-ins so unrelated stored procedures retain prior behavior.
> 
> ### P1 - Commit principal and residency registration atomically (pkg/frontend/iceberg_call.go:265)
> The principal upsert executes before this residency-policy upsert without an explicit transaction. Background sessions use autocommit, so if this second statement fails, the call returns an error but the principal mapping remains committed. This has a deterministic trigger: policy validation accepts a region longer than the DDL's `varchar(128)`, causing only the second insert to fail. Under an existing cluster allow policy, runtime treats the missing account policy as optional, so the partial state can authorize the newly mapped external principal. Wrap both writes in one BEGIN/commit-or-rollback transaction and test rollback when the second write fails.

fixed

iamlinjunhong:
> I found two independent merge blockers on the current head.
> 
> 1. `pkg/frontend/iceberg_call.go:395`: `isIcebergBuiltinProcedure` claims every unqualified procedure whose name starts with `iceberg_`. That interception happens before normal stored-procedure resolution, so an existing user procedure such as `iceberg_backup` is routed into Iceberg parsing and rejected as unsupported—even with the feature disabled. Match the exact built-in procedure set, not the prefix.
> 2. `pkg/frontend/iceberg_call.go:250-280`: access registration performs the principal-map upsert and residency-policy upsert as two autocommit statements. If the second statement fails (for example, a value passes policy validation but exceeds a storage column limit), the call returns an error with the principal mapping already committed. Authorization can then observe a partially registered identity. Execute both mutations in one explicit transaction with rollback on either error and add a second-write-failure regression.

fixed

XuPeng-SH (approved):
Reviewed exact head ad50b5e16737ef1e85063867123186986d472c91. The two prior blockers are addressed: Iceberg built-ins now use an exact allowlist so unrelated stored procedures fall through normally, and access registration writes are atomic with rollback on error and panic. I also verified protobuf regeneration preserves credential redaction, and focused plus race tests passed across frontend, pipeline, Iceberg core, SQL adapters, bootstrap, planning, and execution. No blocking issues found.

gouhongshen (approved):
## Codex automated review

Reviewed the exact head and all review/discussion context. The prior dispatch and atomicity blockers are resolved with an exact built-in allowlist and transactional registration with rollback coverage. Exact-head unit, E2E, SCA, coverage, and cluster checks pass. No actionable regressions found.

heni02 (approved):

pkg/cnservice/server.go
```diff
@@ -290,12 +298,90 @@ func NewService(
...
+			UseNativeExpireSnapshots:  true,
+		},
+	)
+	var tableCache icebergwritecore.TableCache
+	if rt := runtime.ServiceRuntime(s.cfg.UUID); rt != nil {
+		if value, ok := rt.GetGlobalVariables(icebergapi.CacheInvalidatorRuntimeKey); ok {
+			tableCache, _ = value.(icebergwritecore.TableCache)
+		}
+	}
+	cacheInvalidator := icebergwritecore.MetadataCacheInvalidator{Cache: tableCache}
```
LeftHandCold: [P1] Wire the remote cache invalidator into the production coordinator

This instance only clears `tableCache` on the CN that performs the commit. The PR implements `sqliceberg.ClusterRemoteCacheInvalidator` and registers `CmdMethod_IcebergCacheInvalidate` on every CN, but no production path constructs that sender, so `MetadataCacheInvalidator.Remote` is always nil here. `CachedTableMetadataLoader.Load` returns a cache hit without catalog revalidation, and the default TTL is five minutes. Consequently, in a healthy multi-CN cluster, CN-A can commit an append/update/delete while CN-B continues serving the pre-commit snapshot from its local cache until TTL expiry. Please populate `Remote` from the already initialized `s.moCluster` and `s.queryClient` (and add an assembly-level test that a successful commit sends invalidations to the other CNs), rather than only testing the sender in isolation.

pkg/sql/compile/compile.go
```diff
@@ -308,6 +309,7 @@ func (c *Compile) clear() {
 	c.disableDropAutoIncrement = false
 	c.keepAutoIncrement = 0
 	c.disableLock = false
+	c.icebergScanPlanner = nil
```
LeftHandCold, line 312: [P1] Release statement-local scan plans when clearing a pooled Compile

`recordIcebergScanPlan` stores the full `IcebergScanPlan` (data/delete tasks, mappings, profile, etc.) in `c.icebergScanPlans`, but `clear()` resets only the planner. `Compile` objects are returned to the reuse pool through this method, so after a large Iceberg scan the pooled object keeps the entire plan reachable across unrelated statements. With the configured limits this can retain up to a 256 MiB planning working set / one million file tasks per pooled compile and create a CN-wide memory regression under concurrency. Clear or nil `c.icebergScanPlans` here and add a reuse regression that frees and reallocates a `Compile` after recording a plan.

LeftHandCold (changes requested):
Deep-reviewed exact head `ad50b5e16737ef1e85063867123186986d472c91`, including scan/delete correctness, write/commit recovery, maintenance, parser/plan compatibility, bootstrap, security, multi-CN assembly, reuse lifecycles, and focused exact-head tests. The earlier CALL-dispatch and atomic-registration issues are fixed, but the two inline findings below are reachable merge blockers: one can return stale committed data on another healthy CN, and one retains statement-sized scan plans in pooled `Compile` objects across unrelated workloads.

pkg/sql/iceberg/cache_invalidator.go
```diff
@@ -0,0 +1,82 @@
...
+	_ = clusterservice.GetCNServiceWithoutWorkingStateWithContext(broadcastCtx, i.Cluster, clusterservice.NewSelector(), func(cn metadata.CNService) bool {
+		if cn.WorkState != metadata.WorkState_Working && cn.WorkState != metadata.WorkState_Unknown {
+			return true
+		}
+		if strings.TrimSpace(cn.QueryAddress) == "" || (self != "" && cn.ServiceID == self) {
+			return true
+		}
+		request := i.QueryClient.NewRequest(query.CmdMethod_IcebergCacheInvalidate)
+		request.IcebergCacheInvalidateRequest = payload
+		resp, err := i.QueryClient.SendMessage(broadcastCtx, cn.QueryAddress, request)
```
LeftHandCold: [P1] Isolate peer failures instead of serializing the broadcast behind one shared deadline

Every `SendMessage` runs synchronously in cluster-snapshot order under the same three-second `broadcastCtx`. If the first remote CN is still marked Working but its query endpoint is slow or unreachable, this call consumes the whole deadline; line 79 then stops iteration, so every later healthy CN receives no invalidation. The new `TestClusterRemoteCacheInvalidatorBoundsWholeBroadcast` explicitly locks in that behavior by expecting `remote-2` not to be called. Since `CachedTableMetadataLoader.Load` returns a live cache hit without catalog revalidation, those healthy peers can serve the pre-commit snapshot until the default five-minute TTL after an append/update/delete has committed. Please fan out peer sends independently (for example concurrently under the bounded overall context) and add a regression where one blocked peer does not prevent another healthy peer from being invalidated.

pkg/sql/iceberg/cache_invalidator.go
```diff
@@ -0,0 +1,82 @@
...
+		CatalogID:            req.Catalog.CatalogID,
+		Namespace:            api.NamespaceCacheKey(req.Namespace),
+		Table:                req.Table,
+		SnapshotID:           result.SnapshotID,
+		MetadataLocationHash: result.MetadataLocationHash,
+		CommitID:             result.CommitID,
+	}
+	self := i.QueryClient.ServiceID()
+	_ = clusterservice.GetCNServiceWithoutWorkingStateWithContext(broadcastCtx, i.Cluster, clusterservice.NewSelector(), func(cn metadata.CNService) bool {
+		if cn.WorkState != metadata.WorkState_Working && cn.WorkState != metadata.WorkState_Unknown {
```
LeftHandCold: [P1] Do not skip draining CNs while they can still serve existing sessions

This filter drops `WorkState_Draining`, even though the code deliberately enumerates services without the working-state filter. Draining does not mean the CN has stopped serving: proxy scaling discovers its existing tunnels and transfers them asynchronously. During that window, a session on CN-B can still issue an Iceberg SELECT. If CN-A commits while B is draining, B receives no invalidation here and its live session can read the old cached snapshot for up to the five-minute TTL. Include draining peers that still have a query address (or otherwise prove they cannot execute a query before excluding them), and cover the draining-with-live-cache case.

LeftHandCold (changes requested):
Deep-reviewed exact head `80c81f0c67f57f1e199d89d2097b7a767b4ebbf4`. The two previous blockers are fixed: production now wires the remote invalidator, and pooled `Compile` objects release retained Iceberg scan plans. The new cluster sender still has two reachable stale-read regressions, however: one slow/unreachable CN suppresses invalidation of later healthy peers, and draining CNs with live sessions are skipped entirely. Both can leave an active CN serving the cached pre-commit snapshot until the default five-minute TTL. Full tests for `pkg/sql/iceberg`, `pkg/sql/compile`, and `pkg/cnservice` pass, but the new tests do not cover these isolation cases.

LeftHandCold (approved):
Deep-reviewed exact head `366d058f99cc4b94c210132e49570f34aed449a5`. The remaining cross-CN cache-invalidation blockers are resolved: remote sends now fan out independently under one bounded deadline, so a blocked peer cannot suppress healthy peers, and draining CNs with live sessions are included. I rechecked the production query-client wait/response lifecycle, cluster membership bounds, pooled Compile cleanup, and normal business compatibility. Focused race stress, the full owning package under race, full tests for `pkg/sql/iceberg`, `pkg/sql/compile`, and `pkg/cnservice`, plus `go vet ./pkg/sql/iceberg` all pass. No remaining blocking code issue found.

aptend (approved):
Reviewed exact head `366d058f99cc4b94c210132e49570f34aed449a5`. The prior blockers are resolved: built-in CALL dispatch is exact, access registration is transactional, production cache invalidation is wired, pooled scan plans are released, peer failures are isolated, and draining CNs are included. Focused race tests, Iceberg/core integration packages, upgrade tests, and the exact-head Iceberg E2E all pass. No blocking issues found.

aunjgr (approved):
Reviewed independently as a 4.2-dev port. I compared the connector payload with the source implementation on main (378 of 398 changed source files are byte-identical; the remaining files are branch integration/generated-code seams), then audited the port-only hardening commits. The registration transaction rolls back on error and panic, protobuf logging remains redacted, compile-pool reuse releases retained Iceberg scan plans, and remote cache invalidation is post-commit, best-effort, bounded across the whole broadcast, skips self/nonworking CNs, and releases successful responses. Focused tests for these paths pass in pkg/cnservice, pkg/frontend, pkg/pb/pipeline, pkg/sql/compile, and pkg/sql/iceberg. No blocking issues found.

gouhongshen (approved):
## Codex automated review

Reviewed the exact head, diff, existing reviews, issue comments, and inline threads. No actionable regressions found. Focused test execution is locally blocked by missing MatrixOne CGo prerequisites.

Issue #25931, opened 2026-07-21
[Bug]: TestGlobalCheckpoint2 intermittently observes a pending LSN after checkpoint completion


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main / PR #25089 branch `issue-24951-main`

### Commit ID

ae752dcee9b04d199539ac7a411c84833dfe10f5

### Other Environment Information

```Markdown
- Hardware parameters: GitHub-hosted runner
- OS type: Ubuntu / x86
- Others: Matrixone CI unit-test job
```

### Actual Behavior

The Ubuntu/x86 unit-test job failed in `TestGlobalCheckpoint2`.

After `ForceCheckpoint` and `ForceGlobalCheckpoint`, the test waits for
`AllCheckpointsFinished()` and then immediately asserts that the scheduler
pending LSN count is zero. In this run, both assertions observed one pending
LSN:

```text
pkg/vm/engine/tae/db/test/db_test.go:8248
expected: 0x0
actual  : 0x1

pkg/vm/engine/tae/db/test/db_test.go:8254
expected: 0x0
actual  : 0x1
Test: TestGlobalCheckpoint2
```

CI run:
https://github.com/matrixorigin/matrixone/actions/runs/29799310841/job/88537842971

This is unrelated to PR #25089: that PR does not modify
`pkg/vm/engine/tae`, checkpoint, WAL, or scheduler code. The same test has
also failed historically in #20008 and #21721.

### Expected Behavior

`TestGlobalCheckpoint2` should deterministically wait until all
checkpoint-related asynchronous work, including pending LSN processing, has
completed before asserting the pending count.

### Steps to Reproduce

```Markdown
.agents/skills/mo-dev/scripts/mo-cgo-test -v -count=10 \
  -run '^TestGlobalCheckpoint2$' -timeout=180s \
  ./pkg/vm/engine/tae/db/test
```

The local stress run passed 10/10 at the same PR head, indicating a
low-probability timing-dependent failure.

### Additional information

The test currently waits for `AllCheckpointsFinished()`, but that condition
did not guarantee `Runtime.Scheduler.GetPenddingLSNCnt() == 0` in the failing
CI run. Consider waiting explicitly for the pending LSN count or tightening
the completion contract so the assertion cannot race asynchronous scheduler
cleanup.

jiangxinmeng1:
fixed by #25953

matrix-meow:
[...]

PR #25953, opened 2026-07-21
test(tae): wait for checkpoint completion consistently

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25931

wait for checkpoint completion consistently


gouhongshen (changes requested):
## Codex automated review

The shared waiter does not establish a completion boundary beyond the pre-existing instantaneous predicate, so the linked intermittent pending-LSN failure remains possible. Existing PR reviews, issue comments, and inline threads are empty; CI is green, but local focused testing could not run because `go` is unavailable.

### P1 - Wait for the forced checkpoint's completion boundary, not the same racy snapshot (pkg/vm/engine/tae/db/testutil/funcs.go:307)
This polls the unchanged `AllCheckpointsFinished` predicate, then re-reads it once in `require.True`. That predicate reads the WAL pending count and current incremental intent independently; it has no generation/token tied to the preceding ForceCheckpoint/ForceGlobalCheckpoint call. #25931 records the concrete failure: this predicate returned, then the immediately following pending-LSN assertion observed 1. The new helper can follow the identical interleaving (or fail its new recheck) rather than wait for the scheduler work to settle. Use a completion signal scoped to the forced checkpoint, or otherwise synchronize the scheduler/WAL state before returning; merely centralizing and reasserting the old snapshot does not fix the flake.

XuPeng-SH (approved):
Reviewed latest head 1b3432bbd4edfa3e0cea3200ab4d3cc936edbfe4. No blocking findings.

The shared waiter now waits for the WAL checkpointed LSN to reach the captured target/current watermark and turns a silent polling timeout into an explicit assertion failure. The forced ICKP/GCKP completion path and updated callers are consistent.

Local verification:
- TestGlobalCheckpoint2: 20/20 passed
- TestGlobalCheckpoint2 under -race: 3/3 passed
- TestAppend2, TestTruncate1, TestReplay5: 5/5 each passed
- Test_gcArg: 5/5 passed
- go vet passed for the changed/consumer packages

gouhongshen (approved):
## Codex automated review

No findings. The current head captures a post-force WAL target and waits for it to be checkpointed, addressing the obsolete blocking review. Existing PR reviews/comments/inline comments were inspected; Ubuntu CI passed the affected test package. Local CGo tests could not start because libmo.dylib is absent.

PR #25956, opened 2026-07-21
fix(txn): restore recovery routes and cancellation

### LINKED ISSUE #25951: TN recovery can use incomplete participant routes and block shutdown
## Bug report

TN recovery can use incomplete participant routes and can remain blocked after cancellation.

This is independently reproducible on current `main` at `3046f14d94b3f1a25839eeeb65729b64f85ac647`.

### Failure 1: recovered 2PC metadata has no live route

A recovered prepared transaction can contain only the durable participant identity (`ShardID`). The recovery coordinator currently sends `GetStatus` immediately with that incomplete shard metadata. The observed target is:

```text
<ShardID:2, ReplicaID:0, Address:"">
```

Recovery should complete each participant's current `ReplicaID` and transaction service address from the TN cluster snapshot before sending status or commit requests. If the participant is temporarily absent, recovery should wait and retry rather than send to an empty route.

### Failure 2: storage recovery ignores cancellation while publishing

`KVTxnStorage.StartRecovery` publishes recovered transactions with an unconditional channel send. With 17 records and the service's 16-entry recovery channel left full, canceling the supplied context does not terminate the producer.

```text
recovery remained blocked on a full transaction channel after cancellation
```

This also means TN shutdown cannot safely cancel a service whose `Start()` is waiting for recovery to finish.

### Expected invariant

- Prepared/committing recovery requests use a complete current TN participant route.
- A temporarily missing participant or transient cluster lookup is retried.
- Canceling recovery terminates route waits, storage producers, the recovery consumer, and the blocked `TxnService.Start()` path.
- Repeated close and partial-start cleanup remain safe.

### Scope

This issue is limited to generic transaction recovery routing and cancellation. It does not include AUTO_INCREMENT semantics, allocator changes, or 2PC replay ordering/timestamp behavior.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25951

Recovered prepared/committing transaction metadata may contain only durable participant identity. Sending recovery RPCs with that metadata currently uses an empty `ReplicaID` and address. At the same time, a missing participant can make `TxnService.Start()` wait indefinitely, while replica or store shutdown waits for `Start()` before it can close the service.

This PR establishes one recovery invariant: recovery RPCs use complete current participant routes, and TN shutdown can terminate every recovery producer, consumer, route wait, and blocked `Start()`.

| Failure | Change |
| --- | --- |
| Recovered remote participant has no RPC route | Resolve `ShardID` through the current TN cluster snapshot before `GetStatus` or `CommitTNShard`; preserve a durable `LogShardID` when the snapshot omits it. |
| Participant or cluster metadata is temporarily unavailable | Retry lookup and authoritative refresh without sending to an incomplete route. |
| A historical Prepared record names a retired participant but a later Committed record supersedes it | Fold the complete recovery stream first; resolve routes only for coordinator-owned transactions that remain Prepared/Committing. |
| Route never appears | Add a recovery cancellation context and let replica shutdown cancel it before waiting for `Start()`. |
| Store shutdown joins a replica task blocked in recovery | Deliver `cancelStart` and `CancelRecovery` to replicas before stopping the store task group; keep storage open until RPC drain completes. |
| HAKeeper never produces the initial cluster snapshot | Stop the cluster refresh task before releasing readiness waiters, so `Store.Close()` does not depend on metadata readiness. |
| Storage producer blocks on a read or a full recovery channel | Treat context-canceled MEMKV reads as normal termination and make publication select on `ctx.Done()`. |
| Multiple surviving recovery transactions share the same current TN routes | Refresh one authoritative TN snapshot per recovery pass, reuse it across transactions, and refresh/retry only when the shared snapshot lacks a required route. |

Compatibility and non-goals:

- Single-participant and non-coordinator recovery keep the existing local replay path; only coordinators with final unresolved Prepared/Committing state refresh participant routes before sending recovery RPCs.
- No wire or persisted format changes.
- No AUTO_INCREMENT behavior, allocator changes, maximum prepared timestamp repair, or replay/logtail ordering changes are included.

Validation:

- `go test ./pkg/clusterservice ./pkg/txn/service ./pkg/txn/storage/mem -count=1`
- `go test ./pkg/tnservice -count=1`
- `go test ./pkg/txn/... -count=1`
- focused recovery and store-close lifecycle tests under `-race`
- both new timing-sensitive regressions repeated 20 times
- in-flight MEMKV log-read cancellation and non-coordinator missing-route startup
- `go vet ./pkg/clusterservice ./pkg/txn/service ./pkg/tnservice ./pkg/txn/storage/mem ./pkg/vm/engine/test/testutil`
- `go test ./pkg/... -run '^$' -count=1`
- `git diff --check`

Diff size: 8 production/interface/helper files / about 260 added handwritten lines; 6 test files / about 850 added test lines; no generated code.

Co-authored-by: ULookup <CHBulookup@outlook.com>


gouhongshen (changes requested):
## Codex automated review

Two blocking recovery lifecycle regressions remain. Focused package and race tests pass but do not cover them.

### P1 - Cancellation of a MEMKV log read panics the TN (pkg/txn/storage/mem/kv_txn_storage.go:145)
`startRecovery` now passes the cancellable recovery context into `StartRecovery`. During replica shutdown, `CancelRecovery` can therefore cancel an in-flight MEMKV `logClient.Read`; the real log-service RPC future returns `ctx.Err()` on cancellation. This code panics on every read error, so shutting down a MEMKV-backed TN while its recovery read is slow/unavailable crashes the process instead of terminating recovery. The added test only cancels a full channel send, not an in-flight read. Treat context cancellation as normal producer termination.

### P1 - A non-coordinator replica now blocks startup on unrelated participant routes (pkg/txn/service/service_recovery.go:50)
Route resolution runs for every recovered prepared/committing record before `addLog`. But `end` only sends recovery RPCs when this service is the transaction coordinator (`txnMeta.TNShards[0]`). For a non-coordinator B in `[A, B, C]`, if C is absent from the snapshot, B now retries forever in route resolution even though B will never issue a recovery RPC; on `origin/main` it added the log and `end` skipped recovery work, allowing B to start. This makes B unavailable for unrelated work while another participant is missing. Resolve/wait only on the coordinator path and add a non-coordinator missing-route regression test.

VioletQwQ-0:
已处理本轮 review 中的两个 P1，修复提交为 `fa55d75e6`：

1. **MEMKV recovery read cancellation**
   - `StartRecovery` 的日志读取因 recovery context 取消而返回错误时，现在按正常 producer 终止处理，不再进入 panic。
   - 新增 `TestRecoveryCancellationStopsLogRead`，真实阻塞在 `logClient.Read` 后取消 context；修复前稳定 panic，修复后正常退出并关闭 recovery channel。

2. **非 coordinator 不再等待无关 participant route**
   - route refresh/wait 现在只在本 TN 是 transaction coordinator 时执行；非 coordinator 仍恢复本地 prepared/committing 日志，但不会等待它不会发送 RPC 的第三方 route。
   - 新增 `TestRecoveryParticipantDoesNotWaitForUnrelatedRoute`，覆盖 B 作为 `[A, B, C]` participant 且 C 缺失的场景；修复前 Start 卡住，修复后完成本地 recovery 且不发送 recovery RPC。

验证：
- `go test ./pkg/clusterservice ./pkg/txn/service ./pkg/tnservice ./pkg/txn/storage/mem -count=1`
- `go test ./pkg/txn/... -count=1`
- 两个新增测试重复 20 次
- 相关 lifecycle 测试 `-race`
- `go vet` 受影响包
- `go test ./pkg/... -run '^$' -count=1`
- `git diff --check`

同时已更新 PR 描述，明确只有 coordinator 会刷新 participant route。麻烦重新 review，谢谢。

XuPeng-SH (dismissed):
Request changes: four independently confirmed recovery lifecycle blockers remain on `dc4c0667`.

1. **P1 — store shutdown still deadlocks before recovery cancellation.** [`store.Close` calls `s.stopper.Stop()` before `r.close()`](https://github.com/matrixorigin/matrixone/blob/dc4c0667f47b51909d01ac77fc80fde80ecb3d3c/pkg/tnservice/store.go#L240-L260). The stopper waits for the replica task blocked in `TxnService.Start()`, while `CancelRecovery()` is only reached later through `r.close()`. A deterministic test remained blocked until `r.close()` was invoked manually. Cancellation must reach a published service before waiting for the store stopper, without closing storage before RPC drain.

2. **P1 — non-coordinator replicas wait forever for unrelated participant routes.** [Route resolution runs before `addLog` for every recovered Prepared/Committing record](https://github.com/matrixorigin/matrixone/blob/dc4c0667f47b51909d01ac77fc80fde80ecb3d3c/pkg/txn/service/service_recovery.go#L45-L57), while [the coordinator check happens only later in `end`](https://github.com/matrixorigin/matrixone/blob/dc4c0667f47b51909d01ac77fc80fde80ecb3d3c/pkg/txn/service/service_recovery.go#L214-L240). For participants `[A, B, C]`, B cannot start when C is absent even though B will never send a recovery RPC. Reproduced with an isolated regression test.

3. **P1 — canceling an in-flight MEMKV recovery read panics.** [Every `logClient.Read` error is still panicked](https://github.com/matrixorigin/matrixone/blob/dc4c0667f47b51909d01ac77fc80fde80ecb3d3c/pkg/txn/storage/mem/kv_txn_storage.go#L143-L147). Now that the recovery context is cancellable, shutdown can make the RPC return `context.Canceled`; the producer crashes instead of terminating normally. A cancellable fake reproduced the panic with value `context canceled`.

4. **P1 — an obsolete Prepared record can hide a later durable Committed record.** Resolution blocks per record before replay state is updated. For a stream `Prepared(missing retired participant) -> Committed`, recovery waits on the historical Prepared route forever and never consumes the terminal record. Reproduced with the real MEMKV recovery stream. Replay should first derive final transaction state, then resolve routes only for transactions still Prepared/Committing on the coordinator.

Suggested shape: complete replay/state folding first; at `end`, resolve current routes only for remaining coordinator-owned recovery work. Separately, make store shutdown deliver recovery cancellation before stopper join, and treat `ctx.Err()` from the MEMKV producer as normal termination.

Validation performed: related focused tests passed 20x; focused tests across `pkg/clusterservice`, `pkg/txn/service`, `pkg/tnservice`, and `pkg/txn/storage/mem` passed under `-race`; latest `main` merges cleanly. Those existing tests do not cover the four failing transitions above.

XuPeng-SH (changes requested):
Request changes on `fa55d75e`. This commit correctly fixes the prior non-coordinator route wait and canceled MEMKV read panic, but two independently reproduced P1 lifecycle blockers remain.

1. **P1 — production store shutdown still deadlocks before recovery cancellation.** [`store.Close` calls `s.stopper.Stop()` before `r.close()`](https://github.com/matrixorigin/matrixone/blob/fa55d75e6c9459c9aca0480ec9104d7d92b4fca6/pkg/tnservice/store.go#L240-L260). The store stopper waits for the replica task blocked in `TxnService.Start()`, while the `CancelRecovery()` that releases that task is only reached later through `r.close()`. A deterministic reproduction remained blocked until `r.close()` was invoked manually. Deliver recovery cancellation to any published service before joining the store stopper, without closing storage before RPC drain.

2. **P1 — a historical Prepared record can block a later durable Committed terminal record.** [Recovery still resolves routes per record before `addLog`](https://github.com/matrixorigin/matrixone/blob/fa55d75e6c9459c9aca0480ec9104d7d92b4fca6/pkg/txn/service/service_recovery.go#L45-L57). For a coordinator stream `Prepared(missing retired participant) -> Committed`, recovery waits forever on the obsolete Prepared route and never consumes the terminal record that removes the transaction. This remains reproducible after the coordinator-only guard in `fa55d75e`. Replay/fold the stream to final transaction state first, then resolve current routes only for coordinator-owned transactions still Prepared/Committing.

These are not covered by the current green CI or the new focused tests. The existing related tests passed 20x and under `-race`; both additional deterministic regression cases fail on the current head.

gouhongshen (changes requested):
## Codex automated review

Two P1 lifecycle regressions remain. Generated test cache was moved to Trash; worktree is clean.

### P1 - Cancel recovery before joining the store stopper (pkg/tnservice/store.go:241)
`Close` calls `s.stopper.Stop()` before it reaches the replica loop that can invoke `TxnService.CancelRecovery()`. A replica-creation stopper task calls `TxnService.Start()` synchronously; with a coordinator recovery waiting for a missing route, Start cannot return. Stopping the store only calls `r.cancelStart`, which cancels `createCtx` but not the published service's recovery context, so the stopper waits forever for that task and shutdown never reaches the later cancellation. Cancel recovery for published services before stopping the store task group, and add a regression test through `store.Close`, not only `replica.close`.

### P1 - Do not resolve routes for superseded recovery records (pkg/txn/service/service_recovery.go:49)
The recovery consumer resolves a coordinator Prepared/Committing record before `addLog`. If an old Prepared record names a retired/missing participant, `resolveRecoveryTNShards` retries indefinitely and the consumer never reads a later durable Committed record for the same transaction. On the prior behavior that terminal record would be replayed and remove the transaction; now startup is blocked despite there being no outstanding recovery work. Fold/replay the stream to final transaction state first, then resolve routes only for coordinator-owned transactions still Prepared or Committing. Add a multi-participant `Prepared(missing route) -> Committed` regression test; the existing committed-after-prepared coverage is single-participant and bypasses route resolution.

VioletQwQ-0:
已处理当前所有有效 CR；最新修复提交为 `c0d4b152d`。

针对 XuPeng-SH 的 review `4743831168` 与 gouhongshen 的 review `4744851167` 中相同的两个 P1：

1. **store shutdown 在 stopper join 前未取消 recovery**
   - `store.Close` 现在先对每个 replica 执行 `cancelStart(false)` 与 `cancelRecovery()`，再调用 `s.stopper.Stop()`。
   - 这会先释放阻塞在 `TxnService.Start()` recovery route wait 中的 replica task；RPC server drain 与 storage close 的原有先后关系保持不变。
   - 新增 `TestStoreCloseCancelsRecoveryBeforeStoppingReplicaTask`，使用真实 store stopper、MEMKV recovery 和缺失 participant route 复现生产链；修复前只能靠测试手动 `r.close(false)` 解锁，修复后 `store.Close()` 自行完成。

2. **历史 Prepared route wait 阻塞后续 Committed terminal record**
   - recovery consumer 现在先消费并折叠完整日志流。
   - 只在 `end()` 阶段，对最终仍存在、由本 TN coordinator 持有且状态为 Prepared/Committing 的事务解析当前 route。
   - 新增 `TestRecoveryCommittedTerminalRecordDoesNotWaitForObsoleteRoute`，覆盖多 participant 的 `Prepared(missing retired route) -> Committed`；修复后不会解析已被 terminal record 淘汰的 route，也不会发送 recovery RPC。

此前 gouhongshen review `4743622630` 的另外两个 P1 已由 `fa55d75e6` 处理：取消中的 MEMKV read 不再 panic，非 coordinator 不再等待无关 participant route。对应说明已在上一条回复中逐项给出。

本轮验证：

- `go test ./pkg/clusterservice ./pkg/txn/service ./pkg/txn/storage/mem -count=1`
- `go test ./pkg/tnservice -count=1`
- `go test ./pkg/txn/... -count=1`
- 两条新回归分别 `-count=20`
- 相关 recovery/store lifecycle 测试通过 `-race`
- 受影响包 `go vet`
- `go test ./pkg/... -run '^$' -count=1`
- `git diff --check`

PR 描述也已更新 failure-to-change 映射与验证范围。当前无 unresolved inline thread。麻烦 XuPeng-SH 和 gouhongshen 一起重新 review，谢谢。

pkg/tnservice/store.go
```diff
@@ -238,16 +238,20 @@ func (s *store) Start() error {
...
+	// cancellation must be delivered before joining the store stopper. Storage
+	// remains open until the RPC server drains below.
 	s.replicas.Range(func(_, value any) bool {
-		value.(*replica).cancelStart(false)
+		r := value.(*replica)
+		r.cancelStart(false)
+		r.cancelRecovery()
 		return true
 	})
+	s.stopper.Stop()
```
XuPeng-SH, line 251: **P1:** This join is still reachable with a recovery-blocked replica that the preceding `sync.Map.Range` missed. The heartbeat can register a replica after the range but before the stopper transitions out of `running`; stopper cancellation then calls only `cancelStart`, while the published service waits on its independent recovery context. `Stop()` waits forever. Synchronize replica registration with close, or make the stopper cancellation path call `cancelRecovery` too.

XuPeng-SH (changes requested):
Request changes on `c0d4b152`. The previous four recovery blockers are fixed, but one shutdown-generation hole remains.

### P1 — a replica accepted during the initial close drain can still deadlock `Store.Close`

`Close` first ranges over the current `sync.Map`, then calls `s.stopper.Stop()`. That range is not a registration barrier or a consistent snapshot. The heartbeat task can process an AddReplica after the range has passed (and `StartTNReplica` can do the same), while the stopper is still running and accepts the new replica task. If that task publishes its TxnService and blocks in recovery waiting for a missing route, stopper cancellation only runs the existing `r.cancelStart(false)` callback; it never cancels the service's independent recovery context. `Stopper.Stop` therefore waits for `TxnService.Start`, and `Close` can never reach the later replica-close pass.

I reproduced this with a phase-barrier test that holds the initial Range, inserts a late task whose Start blocks until CancelRecovery, and then releases the range. Repeated runs hit the indefinite wait; manually calling `late.cancelRecovery()` immediately releases Close. Some iterations also demonstrate why the current approach is unsafe: `sync.Map.Range` happens to observe the new entry and Close succeeds, but the API explicitly provides no snapshot guarantee.

Please close the registration boundary (for example by synchronizing Close with replica creation/marking the store closed), or propagate stopper cancellation to `r.cancelRecovery()` as well as `cancelStart`, and add the late-registration regression.

gouhongshen (changes requested):
## Codex automated review

One P1 shutdown race remains. Focused Go tests could not run because the Go executable is unavailable in this environment.

### P1 - Late replica registration can still deadlock Store.Close (pkg/tnservice/store.go:251)
The cancellation Range at lines 245-250 is not a registration barrier: StartTNReplica (including from the heartbeat task) can LoadOrStore a replica after Range has passed but before Stop changes the stopper state. If that task publishes its TxnService and blocks in recovery route resolution, Stop cancels stopperCtx, whose AfterFunc only calls cancelStart (line 340), not CancelRecovery. The task remains blocked in TxnService.Start, so Stop waits forever and Close never reaches the later replica-close pass. The added store-close test pre-registers its replica before Close and does not cover this interleaving. Synchronize replica admission with Close or propagate stopper cancellation to cancelRecovery as well.

XuPeng-SH (changes requested):
Re-review on `c0d4b152`: the prior recovery-routing, stream-folding, and cancellation blockers are fixed, but one P1 shutdown-generation hole remains.

`store.Close` ranges the current `sync.Map` and only then calls `s.stopper.Stop()`. That Range is not an admission barrier. A replica can be published after its shard key has already been visited and its stopper task can enter `TxnService.Start()` before Stop. Stopper cancellation only calls `r.cancelStart(false)`; it does not cancel the published service recovery context, so `Stop()` waits forever and Close never reaches the later replica-close pass.

I reproduced this on the exact head with a deterministic phase-barrier test: replace an already-visited shard generation, start its service blocked in recovery, release the Range, and observe Close remain blocked until `late.cancelRecovery()` is invoked manually. The same semantic failure reproduces under `-race`. The 15 focused committed route/cancel/fold/existing-replica tests all pass, but the added store-close test pre-registers its replica and does not cover this interleaving.

Please synchronize replica admission with Close or propagate stopper cancellation to `r.cancelRecovery()` as well, and add the late-registration regression.

gouhongshen (approved):
## Codex automated review

The previously reported correctness and shutdown blockers are fixed on the current head, and exact-head CI is green. One non-blocking recovery performance risk remains.

### P2 - Reuse one authoritative TN snapshot across recovered transactions (pkg/txn/service/service_recovery.go:75)
`end` invokes `resolveRecoveryTNShards` once for every surviving coordinator transaction, and each invocation calls `AuthoritativeRefresher.Refresh` before reading the snapshot. For N prepared/committing transactions, the built-in cluster therefore performs N sequential `GetClusterDetails` RPCs to HAKeeper—even when every transaction uses the same participant set and the first refresh already established a current snapshot. Recovery startup latency and control-plane load now scale with the number of in-flight transactions; with the default 10-second refresh deadline, a slow HAKeeper can amplify this substantially. Refresh once per recovery pass, reuse that snapshot to resolve all final transactions, and refresh/retry only when the shared snapshot lacks a required route.

pkg/tnservice/store.go
VioletQwQ-0: 已在 `3442ae472` 修复，并由本轮 `a4d302b6e` 保持。每个 replica creation task 的 stopper context 现在同时传播到 `cancelStart(false)` 和 `cancelRecovery()`，因此 shutdown 不再依赖初始 `sync.Map.Range` 是否观察到 late generation。`TestStoreCloseCancelsLateReplicaRecoveryFromStopper` 用 phase barrier 覆盖了同 shard generation 在 Range 后被替换、late task 阻塞于 `Start()` 的准确时序；修复前只有手动 `late.cancelRecovery()` 才能释放 `Store.Close`，修复后连续 20 次及 `-race` 均通过。

VioletQwQ-0:
已处理最新 recovery snapshot 反馈，修复提交为 `a4d302b6e`。

- `end()` 现在为一次 recovery pass 创建一个共享 route resolver，首次对 HAKeeper 做 authoritative refresh 后复用同一 TN snapshot。
- 后续 transaction 先使用共享 snapshot；只有 snapshot 缺少所需 route 时才重新 refresh/retry，原有 transient error、route-late-arrival 和 recovery cancellation 语义保持不变。
- 新增 `TestRecoveryReusesAuthoritativeSnapshotAcrossTransactions`，走真实 `doRecovery → addLog → end` 链路。旧实现两个 surviving transaction 稳定产生 2 次 refresh；修复后仅 1 次。

验证通过：新增回归 `-count=20`、相关 recovery 测试 `-race`、`pkg/clusterservice`/`pkg/txn/service`/`pkg/txn/storage/mem`/`pkg/tnservice` 全量测试、`pkg/txn/...`、相关 `go vet`、全仓 `go test ./pkg/... -run "^$"` 编译检查及 `git diff --check`。

此前 XuPeng-SH 的 late-generation shutdown thread 也已逐项回复并解决。烦请 XuPeng-SH 与 gouhongshen 基于新 head 一起重新 review，谢谢。

XuPeng-SH (approved):
Deep re-review completed on exact head a4d302b6e9. No blocking or non-blocking correctness finding remains.

I re-audited the complete diff and every previously reported edge:

- recovery first folds the complete durable stream, then resolves routes only for surviving coordinator-owned Prepared/Committing transactions;
- one authoritative TN snapshot is reused per recovery pass, with refresh/retry only when a required route is absent;
- missing/stale/transient routes, cluster readiness, retry timers, MEMKV reads and bounded channel publication all terminate through the recovery context;
- store shutdown cancels both create and recovery before stopper join, including replicas admitted after the initial Range through stopper cancellation propagation;
- the reserveStart to service publication window terminates through serviceC or startedC, and close remains idempotent with storage kept alive until RPC drain.

Q1 partial construction: service publication and startup completion both have explicit terminal signals. Q2 blocking operations: every new wait has recovery/store cancellation. Q3 partial cleanup: recovery completion is once-guarded; replica close remains the single service/storage cleanup owner.

- all tests in clusterservice, txn/service, txn/storage/mem and tnservice;
- full pkg/txn/... suite;
- all new route/fold/cancel/late-generation regressions under race, count 20;
- affected package build and vet;
- clean merge with current main c0486a2, followed by the four package suites and the same race matrix at count 10;
- PR diff-check clean.

The current-head CI jobs are still running, so this approval does not claim those unfinished jobs passed.

aptend (changes requested):
Request changes: one P1 shutdown closure gap remains on exact head a4d302b6e9.

### P1 — Store.Close still hangs when the cluster never becomes ready

Store.Close now correctly cancels replica recovery before joining the store stopper, but it then calls s.moCluster.Close(). The built-in cluster.Close() executes waitReady() before stopper.Stop(). If the initial HAKeeper GetClusterDetails calls keep failing, readyC is never closed, so shutdown blocks forever at moCluster.Close() and cannot stop the refresh task that owns the only path to readiness.

I reproduced this on the exact head with a cluster client that continuously returns a HAKeeper error and a 10 ms refresh interval: Close was still blocked after 100 ms and returned only after readyC was manually released. This leaves the PR stated shutdown invariant incomplete for the cluster-metadata-unavailable case.

Please make cluster shutdown cancel/stop the refresh task before waiting for readiness, and explicitly release readiness waiters (or model a separate closed state). Add an end-to-end regression where the initial cluster refresh never succeeds and Store.Close still terminates.

Relevant paths:
- https://github.com/matrixorigin/matrixone/blob/a4d302b6e90275310442446245da34ea074b7779/pkg/tnservice/store.go#L240-L252
- https://github.com/matrixorigin/matrixone/blob/a4d302b6e90275310442446245da34ea074b7779/pkg/clusterservice/cluster.go#L368-L371

Other validation on this head passed: affected package tests, focused recovery/lifecycle tests under -race -count=3, and go vet.

VioletQwQ-0:
Addressed the latest shutdown CR in commit `3cee83bf4`.

The failure reproduced on `a4d302b6e9`: with every initial `GetClusterDetails` call failing, `Store.Close()` remained blocked in `cluster.Close()` because it waited on `readyC` before stopping the refresh task that owned the only readiness path.

The fix now stops the cluster refresh/GC tasks first and then uses `readyOnce` to release any readiness waiters. It also no longer closes `forceRefreshC`; after the consumer has stopped, leaving this bounded channel owned by the cluster avoids a concurrent send-on-closed-channel panic and keeps repeated `Close()` safe.

Added `TestStoreCloseWhenClusterNeverBecomesReady`, which starts a real TN store with a HAKeeper client that always fails cluster refresh and verifies `Store.Close()` terminates.

Validation on `3cee83bf4`:
- regression test failed before the fix and passed `-count=20` after it;
- focused cluster/store lifecycle tests passed under `-race -count=10`;
- full tests for `pkg/clusterservice`, `pkg/txn/service`, `pkg/txn/storage/mem`, and `pkg/tnservice`;
- full `pkg/txn/...` tests;
- `go vet` for the affected packages;
- full `go test ./pkg/... -run "^$"` compile check;
- `git diff --check`.

I also re-audited Q1–Q3 for cluster shutdown: both refresh and close terminate `readyC` waiters, stopper-owned tasks are joined, and the remaining force-refresh channel is bounded to one entry.

aptend (approved):
Re-reviewed exact head `3cee83bf4d`. The previous P1 shutdown gap is fixed: cluster close now stops refresh/GC tasks before releasing readiness waiters through `readyOnce`, while the bounded force-refresh channel remains safe for concurrent/repeated close paths.

Validation passed: affected-package tests, focused lifecycle/recovery tests under `-race -count=5`, `go vet`, and `git diff --check`. No remaining findings.

PR #25962, opened 2026-07-21
fix: make lock wait timeout handling safer

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24915 

fix: make lock wait timeout handling safer


gouhongshen (changes requested):
## Codex automated review

The remote async path can exceed the newly introduced lock-wait ceiling/deadline. Focused tests were blocked by missing local CGo artifacts and headers.

### P1 - Honor the carried absolute deadline on the remote async waiter (pkg/lockservice/service_remote.go:277)
This reapplication preserves an earlier `LockWaitDeadline` but rounds its remaining time up into `LockWaitTimeout`. The async owner then discards that absolute deadline: `getAsyncLockWaitDeadline` builds a new deadline from `createAt + LockWaitTimeout` (waiter_events.go:124), and `waiterEvents.add` schedules from it. Thus a 1-second request that reaches the owner with 1ms remaining receives a fresh 1-second owner wait; the client RPC has an additional 30-second slack, so it can return near 2 seconds rather than by the original deadline. This violates the claimed single budget and makes `MaxLockWaitDuration`/caller timeout not an upper bound for remote locks. Use `LockWaitDeadline` as the async deadline (then take the minimum with the owner cap) and cover a late-arriving remote request.

aunjgr (changes requested):
The current remote async path does not enforce the absolute budget carried by this change. `applyLockWaitTimeoutCeiling` preserves `LockWaitDeadline`, but `getAsyncLockWaitDeadline` in `pkg/lockservice/waiter_events.go:119` ignores it and constructs a fresh deadline as `createAt + LockWaitTimeout`. Because the remaining duration was rounded up to whole seconds at the remote hop, a request arriving with only milliseconds left can receive nearly another full second; with multiple forwarding hops the excess repeats. That makes the advertised caller deadline/MaxLockWaitDuration not an actual upper bound. Build the async deadline from `opts.LockWaitDeadline` and take the minimum with the remote-owner cap, then add a late-arrival remote regression that asserts the original absolute deadline.

iamlinjunhong:
> ## Codex automated review
> The remote async path can exceed the newly introduced lock-wait ceiling/deadline. Focused tests were blocked by missing local CGo artifacts and headers.
> 
> ### P1 - Honor the carried absolute deadline on the remote async waiter (pkg/lockservice/service_remote.go:277)
> This reapplication preserves an earlier `LockWaitDeadline` but rounds its remaining time up into `LockWaitTimeout`. The async owner then discards that absolute deadline: `getAsyncLockWaitDeadline` builds a new deadline from `createAt + LockWaitTimeout` (waiter_events.go:124), and `waiterEvents.add` schedules from it. Thus a 1-second request that reaches the owner with 1ms remaining receives a fresh 1-second owner wait; the client RPC has an additional 30-second slack, so it can return near 2 seconds rather than by the original deadline. This violates the claimed single budget and makes `MaxLockWaitDuration`/caller timeout not an upper bound for remote locks. Use `LockWaitDeadline` as the async deadline (then take the minimum with the owner cap) and cover a late-arriving remote request.

fixed

iamlinjunhong:
> The current remote async path does not enforce the absolute budget carried by this change. `applyLockWaitTimeoutCeiling` preserves `LockWaitDeadline`, but `getAsyncLockWaitDeadline` in `pkg/lockservice/waiter_events.go:119` ignores it and constructs a fresh deadline as `createAt + LockWaitTimeout`. Because the remaining duration was rounded up to whole seconds at the remote hop, a request arriving with only milliseconds left can receive nearly another full second; with multiple forwarding hops the excess repeats. That makes the advertised caller deadline/MaxLockWaitDuration not an actual upper bound. Build the async deadline from `opts.LockWaitDeadline` and take the minimum with the remote-owner cap, then add a late-arrival remote regression that asserts the original absolute deadline.

fixed

gouhongshen (changes requested):
## Codex automated review

Commit 80634698 fixes the previously reported remote-async deadline restart, so that finding is withdrawn. However, the synchronous, binding, and forwarded paths still bypass the same absolute budget. Current CI is green; focused local tests could not start because libmo.dylib and required third-party artifacts are absent.

### P1 - Do not restart expired deadlines on synchronous local waits (pkg/lockservice/service.go:291)
When LockWaitDeadline has expired, this converts its remaining duration to a fresh one-second LockWaitTimeout while retaining the expired timestamp. The new absolute-deadline handling only initializes lockContext.lockWaitDeadline for async requests; the normal local path still waits with context.WithTimeoutCause(..., LockWaitTimeout). A contended local request arriving after its deadline therefore waits another second instead of failing immediately. Reject expired deadlines at admission and make synchronous waiters consume the absolute deadline.

### P1 - Make lock-table binding consume the effective lock deadline (pkg/lockservice/service.go:189)
The injected deadline is not passed to getLockTableWithCreate. That function can wait on an in-flight allocation channel and calls getLockTableBind through context.Background(), using an independent RPC timeout. There is also no deadline recheck afterward, so an uncontended lock can succeed after the caller's budget has expired. The remote and forwarded handlers have the same binding-before-admission gap. Propagate a deadline-aware context through allocation/bind waits and recheck before dispatch.

### P1 - Bound ForwardTo RPCs with the injected deadline (pkg/lockservice/service_forward.go:48)
Service.Lock now injects LockWaitTimeout/LockWaitDeadline before entering this mirrored-transaction path, but forwardLock sends with the original context. A deadline-less internal context reaches morpc Future.init and panics with "context deadline not set"; a longer-lived context ignores MaxLockWaitDuration. Derive the RPC context from the absolute lock deadline plus transport slack while preserving any earlier parent deadline.

### P2 - Preserve explicit zero when clearing an executor timeout (pkg/util/executor/options.go:252)
WithLockWaitTimeout(0) is documented as clearing an earlier option, but it sets hasLockWaitTimeout=false. For Options.WithTxn(existingTxn), ExtraTxnOptions are not applied and applyExecutorLockWaitTimeout returns early, after which lockWaitTimeout reuses the existing transaction's stale positive timeout. Track option presence separately from whether its value is positive and cover existing-transaction reuse.

### P2 - Honor sub-second max-lock-wait-duration values (pkg/lockservice/service.go:282)
Config accepts any positive toml.Duration, but the ceiling is rounded up before constructing the authoritative deadline. For example, max-lock-wait-duration="500ms" produces a one-second wait, twice the configured maximum. Keep the wire-compatible seconds value rounded up, but derive LockWaitDeadline from the exact configured duration, or reject sub-second configuration.

heni02 (approved):

iamlinjunhong:
> ## Codex automated review
> Commit [8063469](https://github.com/matrixorigin/matrixone/commit/80634698a510c03b7cda52deeadaa9eea46f4866) fixes the previously reported remote-async deadline restart, so that finding is withdrawn. However, the synchronous, binding, and forwarded paths still bypass the same absolute budget. Current CI is green; focused local tests could not start because libmo.dylib and required third-party artifacts are absent.
> 
> ### P1 - Do not restart expired deadlines on synchronous local waits (pkg/lockservice/service.go:291)
> When LockWaitDeadline has expired, this converts its remaining duration to a fresh one-second LockWaitTimeout while retaining the expired timestamp. The new absolute-deadline handling only initializes lockContext.lockWaitDeadline for async requests; the normal local path still waits with context.WithTimeoutCause(..., LockWaitTimeout). A contended local request arriving after its deadline therefore waits another second instead of failing immediately. Reject expired deadlines at admission and make synchronous waiters consume the absolute deadline.
> 
> ### P1 - Make lock-table binding consume the effective lock deadline (pkg/lockservice/service.go:189)
> The injected deadline is not passed to getLockTableWithCreate. That function can wait on an in-flight allocation channel and calls getLockTableBind through context.Background(), using an independent RPC timeout. There is also no deadline recheck afterward, so an uncontended lock can succeed after the caller's budget has expired. The remote and forwarded handlers have the same binding-before-admission gap. Propagate a deadline-aware context through allocation/bind waits and recheck before dispatch.
> 
> ### P1 - Bound ForwardTo RPCs with the injected deadline (pkg/lockservice/service_forward.go:48)
> Service.Lock now injects LockWaitTimeout/LockWaitDeadline before entering this mirrored-transaction path, but forwardLock sends with the original context. A deadline-less internal context reaches morpc Future.init and panics with "context deadline not set"; a longer-lived context ignores MaxLockWaitDuration. Derive the RPC context from the absolute lock deadline plus transport slack while preserving any earlier parent deadline.
> 
> ### P2 - Preserve explicit zero when clearing an executor timeout (pkg/util/executor/options.go:252)
> WithLockWaitTimeout(0) is documented as clearing an earlier option, but it sets hasLockWaitTimeout=false. For Options.WithTxn(existingTxn), ExtraTxnOptions are not applied and applyExecutorLockWaitTimeout returns early, after which lockWaitTimeout reuses the existing transaction's stale positive timeout. Track option presence separately from whether its value is positive and cover existing-transaction reuse.
> 
> ### P2 - Honor sub-second max-lock-wait-duration values (pkg/lockservice/service.go:282)
> Config accepts any positive toml.Duration, but the ceiling is rounded up before constructing the authoritative deadline. For example, max-lock-wait-duration="500ms" produces a one-second wait, twice the configured maximum. Keep the wire-compatible seconds value rounded up, but derive LockWaitDeadline from the exact configured duration, or reject sub-second configuration.

fixed

gouhongshen (changes requested):
## Codex automated review

Prior deadline-path findings are addressed, but one expiry path can still acquire a lock after its absolute deadline.

### P1 - Reject expiry before an uncontended local acquisition (pkg/lockservice/lock_table_local.go:596)
The new deadline check is reached only after a row conflict. A request can pass the service/bind checks, block on the transaction or local-table mutex until `LockWaitDeadline` expires, then acquire an otherwise-uncontended row through `doAcquireLock` and return success without reaching this branch. Check the absolute deadline before every acquisition attempt (not only on conflict), and add a regression that holds the local-table mutex past the deadline.

pkg/lockservice/service.go
```diff
@@ -171,6 +179,12 @@ func (s *service) Lock(
 	}()
 
 	s.wait()
+	// Service admission/bind work may consume the remaining budget after the
+	// entry check. Recheck before dispatch so a delayed hop cannot restart or
+	// transmit an already exhausted absolute deadline.
+	if lockWaitDeadlineExpired(options, time.Now()) {
```
XuPeng-SH: ### P1 - Make the service-readiness wait consume the lock deadline

This recheck happens only after `s.wait()` returns, but the production wait function blocks on `<-s.hakeeperConnected` and has no context/deadline input. I reproduced this with a 50 ms `LockWaitDeadline`: while the readiness gate stayed closed, `Lock` was still blocked after 250 ms and returned `ErrLockTimeout` only after the gate was manually released. During a prolonged HAKeeper outage this path remains unbounded. Make the readiness wait context-aware (using the effective lock context) so the call returns when the budget expires, rather than merely noticing expiry after readiness eventually arrives.

pkg/lockservice/service.go
```diff
@@ -181,10 +195,27 @@ func (s *service) Lock(
...
+		options.Sharding)
 	if err != nil {
 		return pb.Result{}, err
 	}
+	if err := bindCtx.Err(); err != nil {
+		return pb.Result{}, lockWaitContextError(bindCtx, err)
+	}
+	// Binding can finish concurrently with the deadline. Recheck after it
+	// returns so an uncontended local table cannot admit an expired request.
+	if lockWaitDeadlineExpired(options, time.Now()) {
```
XuPeng-SH: ### P1 - Enforce the deadline after mutex admission, including the uncontended path

This is not the last blocking edge before acquisition: `bindChangeMu.RLock()`, `txn.Lock()`, and the local-table lock are taken afterward. If one of them delays the request past the deadline, an uncontended `doAcquireLock` never calls `handleLockConflictLocked`, so it succeeds after the advertised budget. A deterministic reproduction pre-created the table, held `bindChangeMu`, started a lock with a 50 ms absolute deadline, released the mutex at 150 ms, and observed a nil error. Put the deadline check inside the final lock-admission/acquisition path (and mirror it for remote/forward handlers), so every success is linearized before expiry.

pkg/lockservice/waiter_events.go
```diff
@@ -116,12 +117,18 @@ func (c *lockContext) getLockWaitTimeoutErr() error {
...
+func getLockWaitDeadline(createAt time.Time, opts LockOptions) (time.Time, error) {
 	var (
 		deadline time.Time
 		err      error
 	)
-	if opts.LockWaitTimeout > 0 {
+	// An absolute deadline carried across an RPC is authoritative. The
+	// relative timeout is rounded to whole seconds for wire compatibility and
+	// must only be used when no earlier hop supplied an absolute budget.
+	if opts.LockWaitDeadline > 0 {
```
XuPeng-SH, line 182: ### P1 - Stop remote async waiters at an earlier request-context deadline

The MORPC server context carries the caller's remaining timeout, but this async deadline calculation only considers the option deadline and owner cap. Unlike the synchronous path, an async waiter does not block on `c.ctx`, so cancellation of that earlier server context never wakes/removes it. I reproduced this with a 100 ms caller context and a 500 ms service ceiling: the origin returned at 100 ms, but the owner still had one waiter queued. It can later acquire a lock for an already-canceled request. Include the earlier `ctx.Deadline()` in the async deadline/error model (or install a cancellation wakeup) and assert queue removal at caller cancellation.

pkg/lockservice/lock_table_remote.go
```diff
@@ -112,23 +126,22 @@ func (l *remoteLockTable) lock(
...
-	} else {
-		rpcCtx = ctx
-	}
+	// Bound the RPC by the absolute lock deadline plus transport slack so the
+	// lock-table owner has enough time to return ErrLockTimeout before the
+	// client-side RPC deadline fires.
+	// Service entry points also use this field for the safety ceiling. A zero
+	// value is possible only for direct lock-table callers and tests, where the
+	// caller context remains the fallback.
+	rpcCtx, rpcCancel := newLockRPCContext(ctx, opts.LockOptions)
```
XuPeng-SH: ### P1 - Do not run bind recovery after a terminal lock timeout

This RPC now normally returns `ErrLockWaitTimeout` from the owner at the absolute deadline, but the error path below special-cases only `ErrRemoteLockWaitTimeout`. Plain `ErrLockWaitTimeout` therefore falls through to `handleError(err, true)`, which uses `context.Background()` and can spend another `defaultRPCTimeout` (10 s) in `GetBind`. A deterministic fake client confirmed that a terminal lock timeout starts that allocator RPC. Treat both lock-timeout errors as terminal before bind recovery (or pass a still-valid bounded context), otherwise the newly enforced deadline is extended by an independent wait.

XuPeng-SH (changes requested):
## Codex deep review

Commit `6a46ff89` fixes the direct versions of the previously reported remote-async deadline restart, synchronous waiter deadline, bind RPC, ForwardTo RPC, explicit-zero, and sub-second ceiling cases. Focused current-head tests and the author's CI are green. However, four concrete unhappy paths still break the claimed single budget; each was reproduced with a temporary regression test on this exact head.

The missing invariant is that every admission/wait/recovery edge must terminate at the earliest effective deadline, and no uncontended lock may succeed after it. Please make the readiness gate context-aware, enforce the final deadline inside lock admission, propagate the earlier MORPC request context into async-owner timeout scheduling, and skip deadline-independent bind recovery for terminal lock-timeout responses.

aunjgr (approved):
Reviewed independently with an unhappy-path ownership/wait-for audit. The authoritative deadline is preserved across admission, bind allocation, forwarding, remote RPC, and async re-queue cycles; timeout completion releases the checker reference and resumes the retained lock context outside waiterEvents.mu, avoiding both self-deadlock and double cleanup. The 120s frontend fallback matches the issue's proposed 60–120s bound, while the lockservice ceiling still protects callers that omit propagation. Focused lockservice deadline/timeout regression tests passed locally (exit 0). No actionable issues found.

aptend (changes requested):
Reviewed exact head `6a46ff89a0f71ae24fb60e9f874e838f015cf074`. Requesting changes for the four unresolved P1 deadline/lifecycle gaps already documented inline:

1. the service-readiness wait is not deadline-aware;
2. mutex admission can cross the deadline and still acquire an uncontended lock;
3. remote async waiters ignore an earlier request-context deadline;
4. terminal lock timeout can still trigger an additional bind-recovery wait.

Focused race tests and affected-package tests pass, but they do not cover these reachable production windows.

iamlinjunhong:
> ## Codex automated review
> Prior deadline-path findings are addressed, but one expiry path can still acquire a lock after its absolute deadline.
> 
> ### P1 - Reject expiry before an uncontended local acquisition (pkg/lockservice/lock_table_local.go:596)
> The new deadline check is reached only after a row conflict. A request can pass the service/bind checks, block on the transaction or local-table mutex until `LockWaitDeadline` expires, then acquire an otherwise-uncontended row through `doAcquireLock` and return success without reaching this branch. Check the absolute deadline before every acquisition attempt (not only on conflict), and add a regression that holds the local-table mutex past the deadline.

fixed

iamlinjunhong:
> Reviewed exact head `6a46ff89a0f71ae24fb60e9f874e838f015cf074`. Requesting changes for the four unresolved P1 deadline/lifecycle gaps already documented inline:
> 
> 1. the service-readiness wait is not deadline-aware;
> 2. mutex admission can cross the deadline and still acquire an uncontended lock;
> 3. remote async waiters ignore an earlier request-context deadline;
> 4. terminal lock timeout can still trigger an additional bind-recovery wait.
> 
> Focused race tests and affected-package tests pass, but they do not cover these reachable production windows.

fixed

aptend (approved):
Re-reviewed the latest update at 851cde6. The previously reported deadline lifecycle gaps are closed: readiness and local admission honor the effective deadline, remote async waiters consume the earlier request deadline, and terminal lock timeouts skip bind recovery. Focused regressions, race stress, the full lockservice package, vet, and cnservice compile checks passed locally. No remaining blocking findings from my review.

XuPeng-SH (changes requested):
Deep review on exact head `851cde658d74b3b7d678ffa08987b7d50d9bd1f7`: the four blockers from the previous round are covered by the new tests and pass locally, but three P1 lifecycle gaps remain.

1. **[P1] A notified waiter that crosses the deadline before re-acquisition is never detached.** `doAcquireLock` now returns at `pkg/lockservice/lock_table_local.go:475-476`, but the error path at lines 117-127 only disables/closes `c.w`; it does not remove that waiter from the row/range lock queue, the txn blocked list, or the owner-local wait graph. I reproduced this deterministically: enqueue txn2 behind txn1, unlock txn1 so txn2 is notified, hold txn2 in `beforeAcquire` until its absolute deadline passes, then resume. Lock returns ER_LOCK_WAIT_TIMEOUT, but the empty lock entry remains with `waiters=1`. Subsequent transactions queue behind a completed waiter with no holder left to notify it. Please make the post-notification admission-error cleanup exactly-once for both row/range and sync/async paths, and add this race regression.

2. **[P1] Bind recovery can still extend a non-expired budget by `defaultRPCTimeout`.** The check at `pkg/lockservice/lock_table_remote.go:231` only rejects a deadline that is already expired. A fast transport error with a small amount of budget remaining proceeds to line 250, where `handleError()` deliberately uses `context.Background()`; its allocator GetBind can therefore run for the full default RPC timeout. The successful `resp.NewBind` branch likewise calls `maybeHandleBindChanged(ctx, resp)` with the original deadline-less caller context. Derive a recovery/refresh context from `opts.LockWaitDeadline`, use `handleErrorWithContext`, and normalize expiry back to ER_LOCK_WAIT_TIMEOUT. Please cover both fast transport failure + blocked allocator and NewBind + blocked allocator.

3. **[P1] The supported remote-local shared proxy converts timeout/cancellation into lock success.** With `EnableRemoteLocalProxy`, a second shared lock coalesced behind the first owner RPC reaches `pkg/lockservice/lock_table_proxy.go:131-134`. It ignores the result of `w.wait(ctx)` and never consumes `LockWaitDeadline`; the deferred block then calls `txn.lockAdded` and the callback with nil. A deterministic 50 ms context-timeout test returned nil. When the first RPC later completes, `sharedOps.done` invokes the same callback again. This can let execution continue without an owner lock and creates an exactly-once/data-race problem. The proxy wait must use the effective absolute budget, propagate timeout, remove the timed-out op from `sharedOps`, and ensure one callback/bookkeeping outcome. This is not merely theoretical: the config is supported and has production-path unit coverage, even though its default is false.

iamlinjunhong:
> Deep review on exact head `851cde658d74b3b7d678ffa08987b7d50d9bd1f7`: the four blockers from the previous round are covered by the new tests and pass locally, but three P1 lifecycle gaps remain.
> 
> 1. **[P1] A notified waiter that crosses the deadline before re-acquisition is never detached.** `doAcquireLock` now returns at `pkg/lockservice/lock_table_local.go:475-476`, but the error path at lines 117-127 only disables/closes `c.w`; it does not remove that waiter from the row/range lock queue, the txn blocked list, or the owner-local wait graph. I reproduced this deterministically: enqueue txn2 behind txn1, unlock txn1 so txn2 is notified, hold txn2 in `beforeAcquire` until its absolute deadline passes, then resume. Lock returns ER_LOCK_WAIT_TIMEOUT, but the empty lock entry remains with `waiters=1`. Subsequent transactions queue behind a completed waiter with no holder left to notify it. Please make the post-notification admission-error cleanup exactly-once for both row/range and sync/async paths, and add this race regression.
> 2. **[P1] Bind recovery can still extend a non-expired budget by `defaultRPCTimeout`.** The check at `pkg/lockservice/lock_table_remote.go:231` only rejects a deadline that is already expired. A fast transport error with a small amount of budget remaining proceeds to line 250, where `handleError()` deliberately uses `context.Background()`; its allocator GetBind can therefore run for the full default RPC timeout. The successful `resp.NewBind` branch likewise calls `maybeHandleBindChanged(ctx, resp)` with the original deadline-less caller context. Derive a recovery/refresh context from `opts.LockWaitDeadline`, use `handleErrorWithContext`, and normalize expiry back to ER_LOCK_WAIT_TIMEOUT. Please cover both fast transport failure + blocked allocator and NewBind + blocked allocator.
> 3. **[P1] The supported remote-local shared proxy converts timeout/cancellation into lock success.** With `EnableRemoteLocalProxy`, a second shared lock coalesced behind the first owner RPC reaches `pkg/lockservice/lock_table_proxy.go:131-134`. It ignores the result of `w.wait(ctx)` and never consumes `LockWaitDeadline`; the deferred block then calls `txn.lockAdded` and the callback with nil. A deterministic 50 ms context-timeout test returned nil. When the first RPC later completes, `sharedOps.done` invokes the same callback again. This can let execution continue without an owner lock and creates an exactly-once/data-race problem. The proxy wait must use the effective absolute budget, propagate timeout, remove the timed-out op from `sharedOps`, and ensure one callback/bookkeeping outcome. This is not merely theoretical: the config is supported and has production-path unit coverage, even though its default is false.

fixed

gouhongshen (approved):
## Codex automated review

Reviewed exact head 14d7fb1 against origin/4.2-dev, including all PR reviews, issue comments, and inline threads. The latest lifecycle fixes correctly close the previously reported waiter-detach, deadline-aware bind-recovery, and shared-proxy timeout paths. No actionable findings.

pkg/lockservice/service.go
```diff
@@ -157,6 +163,14 @@ func (s *service) Lock(
 	rows [][]byte,
 	txnID []byte,
 	options pb.LockOptions) (pb.Result, error) {
+	options = s.applyLockWaitTimeoutCeiling(options)
+	if lockWaitDeadlineExpired(options, time.Now()) {
+		return pb.Result{}, ErrLockTimeout
+	}
+	lockCtx, cancelLockCtx := newLockWaitContext(ctx, options)
```
XuPeng-SH: [P1] Keep deadline-context allocation off the uncontended lock path

`applyLockWaitTimeoutCeiling` now gives every request a deadline, so this call always creates and cancels a child context even when readiness is already satisfied and the local table bind already exists. On the same M4 host, `BenchmarkWithoutConflict/1-table` over five 200 ms samples averaged 3760 ns/op, 2931 B/op, 18 allocs/op on this head versus 3377 ns/op, 2831 B/op, 16 allocs/op on base `66dad148`—about an 11.3% latency regression plus 2 allocations and ~100 B per lock+unlock. Allocation profiling attributes the increment to `context.WithDeadlineCause`/`context.WithCancel`; because this benchmark has an earlier 1 s parent deadline, it does not even exercise the additional timer that a normal context without an earlier deadline will create. This cost is paid on the core no-wait path. Please preserve the absolute deadline/admission checks but instantiate the deadline context/timer lazily only when readiness, bind allocation, or a remote wait can actually block, and retain a benchmark that catches this regression.

XuPeng-SH (changes requested):
Requesting changes for one blocking hot-path performance regression. The correctness and lifecycle gaps raised in earlier rounds appear closed, and the focused/full lockservice race suites plus propagation tests pass on this head. However, the new service-level deadline context is now allocated for every Lock call, including uncontended locks with an existing local bind.

gouhongshen (approved):
## Codex automated review

No actionable findings.

XuPeng-SH:
Deep review on PR #25962 (`base=4.2-dev`, head `83a850b1e1`) found two merge-blocking lock-wait issues.

1. **The retry loop still drops a service-injected lock-wait deadline.** `service.Lock` now injects/bounds an absolute deadline by value (`pkg/lockservice/service.go:181-205,296-355`), and the comment there explicitly says callers that retry must propagate their own deadline. But `lockWithRetry` only refreshes `LockWaitTimeout` when `options.LockWaitDeadline > 0`; otherwise `refreshLockWaitOptions` is a no-op (`pkg/sql/colexec/lockop/lock_op.go:843-855`). The retry loop then reuses that unchanged `options` on every retry (`pkg/sql/colexec/lockop/lock_op.go:858-893`). At the same time, `ErrLockTableNotFound` is retryable (`pkg/sql/colexec/lockop/lock_op.go:1060-1063`) but not part of the bounded backend-retry set (`pkg/sql/colexec/lockop/lock_op.go:1070-1075`).

   So if the original caller entered `lockWithRetry` without a caller-owned `LockWaitDeadline`, each retry gets a fresh service ceiling and the overall statement never reaches a terminal timeout under sustained bind/allocator churn.

   **Suggested fix:** materialize one absolute deadline into `options` before the first retryable lock call when none is present, and reuse that same deadline across retries.

2. **The legacy remote fallback can restart the wait budget after local delay.** This PR makes `LockWaitDeadline` the authoritative budget, but the compatibility `LockWaitTimeout` sent over remote/forward RPCs is still left at the original full timeout unless the caller `ctx` has an earlier deadline (`pkg/lockservice/lock_table_remote.go:66-92,146-155`; `pkg/lockservice/service_forward.go:30-58`). That means local readiness / bind work can already consume most of the budget before a cross-CN hop, yet an older peer still sees the stale full `LockWaitTimeout`. The base branch behavior in `origin/4.2-dev:pkg/lockservice/waiter_events.go:119-126` derives the owner-side async wait deadline only from `LockWaitTimeout`, so a rolling-upgrade peer that ignores `LockWaitDeadline` will wait too long.

   Concrete failure: origin CN spends most of the wait budget in local readiness/bind work, then forwards to an old 4.2-dev peer; the origin times out around T, but the old owner keeps the abandoned waiter queued until roughly `T + original_timeout`, blocking later lockers behind it.

   **Suggested fix:** recompute the legacy `LockWaitTimeout` from the remaining `LockWaitDeadline - now` immediately before every cross-CN send, even when `ctx` has no earlier deadline.

iamlinjunhong:
> Deep review on PR #25962 (`base=4.2-dev`, head `83a850b1e1`) found two merge-blocking lock-wait issues.
> 
> 1. **The retry loop still drops a service-injected lock-wait deadline.** `service.Lock` now injects/bounds an absolute deadline by value (`pkg/lockservice/service.go:181-205,296-355`), and the comment there explicitly says callers that retry must propagate their own deadline. But `lockWithRetry` only refreshes `LockWaitTimeout` when `options.LockWaitDeadline > 0`; otherwise `refreshLockWaitOptions` is a no-op (`pkg/sql/colexec/lockop/lock_op.go:843-855`). The retry loop then reuses that unchanged `options` on every retry (`pkg/sql/colexec/lockop/lock_op.go:858-893`). At the same time, `ErrLockTableNotFound` is retryable (`pkg/sql/colexec/lockop/lock_op.go:1060-1063`) but not part of the bounded backend-retry set (`pkg/sql/colexec/lockop/lock_op.go:1070-1075`).
>    So if the original caller entered `lockWithRetry` without a caller-owned `LockWaitDeadline`, each retry gets a fresh service ceiling and the overall statement never reaches a terminal timeout under sustained bind/allocator churn.
>    **Suggested fix:** materialize one absolute deadline into `options` before the first retryable lock call when none is present, and reuse that same deadline across retries.
> 2. **The legacy remote fallback can restart the wait budget after local delay.** This PR makes `LockWaitDeadline` the authoritative budget, but the compatibility `LockWaitTimeout` sent over remote/forward RPCs is still left at the original full timeout unless the caller `ctx` has an earlier deadline (`pkg/lockservice/lock_table_remote.go:66-92,146-155`; `pkg/lockservice/service_forward.go:30-58`). That means local readiness / bind work can already consume most of the budget before a cross-CN hop, yet an older peer still sees the stale full `LockWaitTimeout`. The base branch behavior in `origin/4.2-dev:pkg/lockservice/waiter_events.go:119-126` derives the owner-side async wait deadline only from `LockWaitTimeout`, so a rolling-upgrade peer that ignores `LockWaitDeadline` will wait too long.
>    Concrete failure: origin CN spends most of the wait budget in local readiness/bind work, then forwards to an old 4.2-dev peer; the origin times out around T, but the old owner keeps the abandoned waiter queued until roughly `T + original_timeout`, blocking later lockers behind it.
>    **Suggested fix:** recompute the legacy `LockWaitTimeout` from the remaining `LockWaitDeadline - now` immediately before every cross-CN send, even when `ctx` has no earlier deadline.

fixed

fengttt (approved):

XuPeng-SH (changes requested):
Reviewed exact head ed89d33edf6d7b55a3b0d25ee01029aa5f22c077 from first principles, including local/remote/forward/proxy paths, retry and rolling-upgrade behavior, timeout cleanup, and the uncontended hot path.

The previous retry-budget and legacy-peer findings are fixed, and all focused deadline tests pass locally. Two merge blockers remain:

1. [P1] Shared-proxy admission can still wait forever behind a retrying unlock (pkg/lockservice/lock_table_proxy.go:79).

lockBudgetCtx cannot bound the wait for lp.mu.Lock(). unlockWithContext takes the same mutex at line 217 and keeps it held while calling the remote unlock at line 313. Ordinary Unlock supplies context.Background(), and remoteLockTable.unlockWithContext retries until success. If the owner CN is unreachable, one shared-proxy unlock can therefore hold lp.mu indefinitely; every later shared lock blocks at line 79 past LockWaitDeadline and never reaches the context check below. This recreates an unbounded production wait under an owner/network failure.

Please avoid holding the proxy mutex across the retrying RPC (using an explicit pending-transition state), or make admission context-aware without violating proxy ownership. Add a regression with a blocked contextUnlocker: a lock with a short absolute deadline must return before the unlock is released.

2. [P1] Per-row deadline polling regresses the uncontended hot path (pkg/lockservice/lock_table_local.go:534).

The new call performs ctx.Err() and time.Now() for every row immediately after doAcquireLock already performed the same admission check at line 510; handleLockConflictLocked also checks before an actual wait. With MaxLockRowCount=20000, this makes deadline enforcement linear in the number of uncontended rows and can abort a CPU-only batch mid-acquisition even though the surrounding contract says lock_wait_timeout counts time actually spent waiting.

The repository's own BenchmarkWithoutConflict/1-table on this head versus merge-base (8 x 400 ms, same machine) showed median about 3.43 us/op versus 3.14 us/op, about 9% slower, with allocations unchanged at 16/op; a separate run was about 13% slower. Please consolidate the duplicate fast-path checks or check only at a conflict/coarse batch boundary, retaining checks after genuinely blocking admission points.

Validation passed for focused tests in pkg/lockservice, pkg/sql/colexec/lockop, pkg/vm/process, pkg/sql/compile, and pkg/util/executor; current CI is green. Requesting changes because the first issue is a real timeout escape and the second violates the no-regression requirement on the core lock path.

iamlinjunhong:
> Reviewed exact head [ed89d33](https://github.com/matrixorigin/matrixone/commit/ed89d33edf6d7b55a3b0d25ee01029aa5f22c077) from first principles, including local/remote/forward/proxy paths, retry and rolling-upgrade behavior, timeout cleanup, and the uncontended hot path.
> 
> The previous retry-budget and legacy-peer findings are fixed, and all focused deadline tests pass locally. Two merge blockers remain:
> 
> 1. [P1] Shared-proxy admission can still wait forever behind a retrying unlock (pkg/lockservice/lock_table_proxy.go:79).
> 
> lockBudgetCtx cannot bound the wait for lp.mu.Lock(). unlockWithContext takes the same mutex at line 217 and keeps it held while calling the remote unlock at line 313. Ordinary Unlock supplies context.Background(), and remoteLockTable.unlockWithContext retries until success. If the owner CN is unreachable, one shared-proxy unlock can therefore hold lp.mu indefinitely; every later shared lock blocks at line 79 past LockWaitDeadline and never reaches the context check below. This recreates an unbounded production wait under an owner/network failure.
> 
> Please avoid holding the proxy mutex across the retrying RPC (using an explicit pending-transition state), or make admission context-aware without violating proxy ownership. Add a regression with a blocked contextUnlocker: a lock with a short absolute deadline must return before the unlock is released.
> 
> 2. [P1] Per-row deadline polling regresses the uncontended hot path (pkg/lockservice/lock_table_local.go:534).
> 
> The new call performs ctx.Err() and time.Now() for every row immediately after doAcquireLock already performed the same admission check at line 510; handleLockConflictLocked also checks before an actual wait. With MaxLockRowCount=20000, this makes deadline enforcement linear in the number of uncontended rows and can abort a CPU-only batch mid-acquisition even though the surrounding contract says lock_wait_timeout counts time actually spent waiting.
> 
> The repository's own BenchmarkWithoutConflict/1-table on this head versus merge-base (8 x 400 ms, same machine) showed median about 3.43 us/op versus 3.14 us/op, about 9% slower, with allocations unchanged at 16/op; a separate run was about 13% slower. Please consolidate the duplicate fast-path checks or check only at a conflict/coarse batch boundary, retaining checks after genuinely blocking admission points.
> 
> Validation passed for focused tests in pkg/lockservice, pkg/sql/colexec/lockop, pkg/vm/process, pkg/sql/compile, and pkg/util/executor; current CI is green. Requesting changes because the first issue is a real timeout escape and the second violates the no-regression requirement on the core lock path.

fixed

XuPeng-SH (changes requested):
Deep re-review of exact head `36fbd3e24bb877d128b68ad66284ee6cf6f5e7da`.

The shared-proxy correctness blocker from the previous round is closed: lock admission now waits on a per-row transition with its own budget, and the existing handoff/retry/lost-response cases plus the full proxy race suite pass. Removing the per-row deadline checks also removes the batch-size-linear polling. Two production performance blockers remain:

1. [P1] The ordinary uncontended Lock/Unlock path is still materially slower than the target branch (`pkg/lockservice/service.go:181-230`, plus the final admission check at `pkg/lockservice/lock_table_local.go:510`).

A cached local bind with empty `pb.LockOptions{}` still materializes the one-hour ceiling and performs repeated `time.Now`/deadline checks at entry, after an already-ready gate, after an already-cached bind, and again under the local-table mutex. The latest change removed the loop-level duplicate, but not this fixed hot-path cost. On the repository's own `BenchmarkWithoutConflict/1-table`, pinned to one CPU and interleaved on the same machine (2 x 5 samples, 400 ms each), this head had a median of about 3.88 us/op versus 3.45 us/op at base `66dad148`: about 12% slower, with allocations unchanged at 16/op. Please retain the checks after genuinely blocking points and the final mutex admission guard, but avoid eagerly materializing/checking the synthetic ceiling at every already-ready/cached hop. Add a benchmark threshold or focused microbenchmark so this does not regress again.

2. [P1] Every shared-proxy unlock now pays transition allocation even when no remote handoff/RPC is needed (`pkg/lockservice/lock_table_proxy.go:245`, `382-414`).

`beginUnlockTransition` is called before the code knows whether all rows are skipped local non-holders. It allocates the row-key slice, transition object/channel, installs/removes map entries, and takes two extra mutex round trips even for that pure-local path. A focused benchmark of the same non-holder unlock on the exact previous and current heads measured:

- `ed89d33`: ~104 ns/op, 80 B/op, 1 alloc/op
- `36fbd3e`: ~237 ns/op, 240 B/op, 4 allocs/op

That is about 2.27x latency and three additional allocations on the local release path of the proxy feature whose purpose is to avoid owner RPCs. Please preserve a no-allocation/all-local fast path: under the mutex, first observe any existing overlapping transition; if none exists and planning proves no remote operation is required, apply the local removal directly. Allocate/publish a transition only before releasing the mutex for an actual remote handoff, then re-plan after waiting as needed.

Validation on the exact head: all `TestProxy*` under `-race` passed, full `pkg/lockservice` passed (131.6 s), and `go vet ./pkg/lockservice` passed. CI is still running at review time.

iamlinjunhong:
> Deep re-review of exact head `36fbd3e24bb877d128b68ad66284ee6cf6f5e7da`.
> 
> The shared-proxy correctness blocker from the previous round is closed: lock admission now waits on a per-row transition with its own budget, and the existing handoff/retry/lost-response cases plus the full proxy race suite pass. Removing the per-row deadline checks also removes the batch-size-linear polling. Two production performance blockers remain:
> 
> 1. [P1] The ordinary uncontended Lock/Unlock path is still materially slower than the target branch (`pkg/lockservice/service.go:181-230`, plus the final admission check at `pkg/lockservice/lock_table_local.go:510`).
> 
> A cached local bind with empty `pb.LockOptions{}` still materializes the one-hour ceiling and performs repeated `time.Now`/deadline checks at entry, after an already-ready gate, after an already-cached bind, and again under the local-table mutex. The latest change removed the loop-level duplicate, but not this fixed hot-path cost. On the repository's own `BenchmarkWithoutConflict/1-table`, pinned to one CPU and interleaved on the same machine (2 x 5 samples, 400 ms each), this head had a median of about 3.88 us/op versus 3.45 us/op at base `66dad148`: about 12% slower, with allocations unchanged at 16/op. Please retain the checks after genuinely blocking points and the final mutex admission guard, but avoid eagerly materializing/checking the synthetic ceiling at every already-ready/cached hop. Add a benchmark threshold or focused microbenchmark so this does not regress again.
> 
> 2. [P1] Every shared-proxy unlock now pays transition allocation even when no remote handoff/RPC is needed (`pkg/lockservice/lock_table_proxy.go:245`, `382-414`).
> 
> `beginUnlockTransition` is called before the code knows whether all rows are skipped local non-holders. It allocates the row-key slice, transition object/channel, installs/removes map entries, and takes two extra mutex round trips even for that pure-local path. A focused benchmark of the same non-holder unlock on the exact previous and current heads measured:
> 
> * `ed89d33`: ~104 ns/op, 80 B/op, 1 alloc/op
> * `36fbd3e`: ~237 ns/op, 240 B/op, 4 allocs/op
> 
> That is about 2.27x latency and three additional allocations on the local release path of the proxy feature whose purpose is to avoid owner RPCs. Please preserve a no-allocation/all-local fast path: under the mutex, first observe any existing overlapping transition; if none exists and planning proves no remote operation is required, apply the local removal directly. Allocate/publish a transition only before releasing the mutex for an actual remote handoff, then re-plan after waiting as needed.
> 
> Validation on the exact head: all `TestProxy*` under `-race` passed, full `pkg/lockservice` passed (131.6 s), and `go vet ./pkg/lockservice` passed. CI is still running at review time.

fixed

XuPeng-SH (changes requested):
Deep re-review of exact head 4db604a425bd4025022962d2f2c41a4c850f54d2.

The two performance blockers from the previous round are closed: BenchmarkProxyLocalNonHolderUnlock is about 34 ns/op with 0 B/op and 0 allocs/op, and the cached uncontended service benchmark is back close to the current 4.2-dev baseline with allocations unchanged. The full pkg/lockservice race suite also passes (137 s). One correctness regression introduced by the lazy final guard remains.

[P1] A canceled caller can still acquire an uncontended local lock after the last service context check (pkg/lockservice/lock_table_local.go:501-526; pkg/lockservice/service.go:226-315).

doAcquireLock sets checkDeadline only from c.lockWaitDeadline. For a request with no explicit LockWaitTimeout/LockWaitDeadline, the synthetic ceiling remains lazy; if the local-table TryLock succeeds, checkDeadline is false and c.checkLockWaitDeadline is skipped. That skipped helper is also the final ctx.Err check. service.Lock checks ctx immediately after binding, but does not check it after bindChangeMu/txn admission, so cancellation in that interval is lost and the row is inserted with a nil result.

I reproduced this deterministically on this exact head without timing sleeps: use local.options.beforeAcquire to pause after service admission, cancel a context.WithCancel request with zero lock timeout, then resume. Lock returns nil and acquires the row instead of returning context.Canceled. Existing explicit-deadline tests do not cover this lazy/no-budget branch.

Please keep wall-clock/deadline work off the uncontended fast path, but preserve the caller-context final admission guard separately (for example, check cancellation without calling time.Now when no lock deadline is materialized), and add the deterministic regression. Requesting changes because a canceled operation must not create new lock ownership after cancellation has already happened.

iamlinjunhong:
> Deep re-review of exact head [4db604a](https://github.com/matrixorigin/matrixone/commit/4db604a425bd4025022962d2f2c41a4c850f54d2).
> 
> The two performance blockers from the previous round are closed: BenchmarkProxyLocalNonHolderUnlock is about 34 ns/op with 0 B/op and 0 allocs/op, and the cached uncontended service benchmark is back close to the current 4.2-dev baseline with allocations unchanged. The full pkg/lockservice race suite also passes (137 s). One correctness regression introduced by the lazy final guard remains.
> 
> [P1] A canceled caller can still acquire an uncontended local lock after the last service context check (pkg/lockservice/lock_table_local.go:501-526; pkg/lockservice/service.go:226-315).
> 
> doAcquireLock sets checkDeadline only from c.lockWaitDeadline. For a request with no explicit LockWaitTimeout/LockWaitDeadline, the synthetic ceiling remains lazy; if the local-table TryLock succeeds, checkDeadline is false and c.checkLockWaitDeadline is skipped. That skipped helper is also the final ctx.Err check. service.Lock checks ctx immediately after binding, but does not check it after bindChangeMu/txn admission, so cancellation in that interval is lost and the row is inserted with a nil result.
> 
> I reproduced this deterministically on this exact head without timing sleeps: use local.options.beforeAcquire to pause after service admission, cancel a context.WithCancel request with zero lock timeout, then resume. Lock returns nil and acquires the row instead of returning context.Canceled. Existing explicit-deadline tests do not cover this lazy/no-budget branch.
> 
> Please keep wall-clock/deadline work off the uncontended fast path, but preserve the caller-context final admission guard separately (for example, check cancellation without calling time.Now when no lock deadline is materialized), and add the deterministic regression. Requesting changes because a canceled operation must not create new lock ownership after cancellation has already happened.

fixed

XuPeng-SH (approved):
Deep re-reviewed exact head fdbd7d194fcdda8999041f18d1dbc319109380cc. The last correctness blocker is closed: final local admission always rejects caller cancellation without materializing or reading a synthetic wall-clock deadline, while explicit/materialized deadlines retain the absolute-deadline guard. The deterministic cancellation regression passed under -race -count=20; the full affected race matrix passed, including pkg/lockservice (134 s). BenchmarkWithoutConflict/1-table median was about 3.14 us/op versus 3.07 us/op at the current base, with allocations unchanged at 16/op.

PR #25968, opened 2026-07-21
fix(gossip): remove departed stats routes

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25941

Gossip applies set/delete notifications symmetrically to data-cache and stats
routes, but `NotifyLeave` removed only data-cache entries. Stats routes owned by
a departed member therefore remained until an exact-key overwrite or process
exit.

Add a store-local `removeByTarget` operation and invoke it for both route
stores. Each store protects its own map with its existing mutex; locks are not
nested. The removal compares the current owner while holding the lock, so a
route already overwritten from member A to member B survives a delayed leave
notification for A.

The regression test establishes routes through the production
`AddItem -> GetBroadcasts -> NotifyMsg` path and verifies departed-owner
cleanup, overwritten-owner preservation, and unrelated-owner preservation.

The confirmed issue is retained stale route metadata. The broader claim that
current main sends a stats RPC to the departed address is not established:
the broadcaster stores a reduced stats key while production lookups use the
full key, and the remote stats request is currently created without its
request body. Those separate defects are intentionally outside this PR.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/gossip`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=10 -timeout=180s -run '^TestDelegateNotifyLeaveRemovesRoutesOwnedByDepartedNode$' ./pkg/gossip`
- `go build ./pkg/gossip` (repository CGo environment)
- `go vet ./pkg/gossip` (repository CGo environment)


gouhongshen (approved):
## Codex automated review

No findings. Focused CGo tests could not start because cgo/libmo.dylib is absent from the checkout.

XuPeng-SH (approved):
Deep review of current head `3586c10008942ebb6b52bf53c6c1528ffb3df781`: approved with no merge-blocking findings.

The cleanup comparison and deletion are linearized under each store's existing mutex, so a delayed leave for target A cannot delete a route already overwritten to a different target B. The two stores are cleaned without nested locks, and the touched paths introduce no new Q1 leak/double-cleanup, Q2 wait-cycle, or Q3 unbounded accumulation.

Validation completed on this head:
- `pkg/gossip` full test suite
- `pkg/gossip` full suite with `-race`
- focused regression with `-race -count=10`
- `go build ./pkg/gossip`
- `go vet ./pkg/gossip`
- GitHub Ubuntu UT, SCA, Coverage, Compose PESSIMISTIC, and standalone BVT checks are green

Decision log: user gossip messages do not carry a member incarnation, so post-leave delayed messages or same-address generations cannot be distinguished by this patch. That is a pre-existing protocol boundary and is outside this PR's targeted symmetry fix; it does not invalidate the verified `NotifyLeave` cleanup behavior.

PR #25969, opened 2026-07-21
fix(fileservice): validate remote cache payloads

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25885

The remote-cache RPC boundary trusted request shape and response indexes.
Schema-valid requests with an empty key list or an omitted nested `CacheKey`
could panic the serving CN. A malformed or inconsistent response could use a
negative/out-of-range index, an index requested from a different target, or an
incorrect data length to panic the requester or mark invalid bytes as a hit.

Validate every request entry before allocating the I/O vector, require all
keys in one request to name the same file, and return `bad request` without
installing a response release callback on invalid input. If `ReadCache`
partially acquires cache data and then fails, release it immediately.

On the client, accept only indexes sent to that target, reject out-of-bounds
and wrong-length hits, and let the normal local-storage path handle every
invalid response. An invalid hit does not consume its index, while the first
fully valid response wins. The MORPC response is released on all received
response paths.

Normal same-version senders already generate valid shapes; nil repeated
message pointers are only possible for direct in-process callers. The network-
relevant cases are empty/missing nested request fields and invalid scalar
response indexes/lengths.

Validation:

- focused fileservice regressions: PASS
- focused fileservice regressions with `-race -count=10`: PASS
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/fileservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=240s ./pkg/fileservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=240s ./pkg/cnservice`
- `go build ./pkg/fileservice ./pkg/cnservice` (repository CGo environment)
- `go vet ./pkg/fileservice ./pkg/cnservice` (repository CGo environment)


gouhongshen (approved):
## Codex automated review

No findings. The change validates malformed request/response shapes before dereference or indexing, preserves fallback-to-local behavior for invalid remote hits, and releases partially acquired cache data on server-side read errors. Existing PR reviews, issue comments, and inline threads contained no actionable feedback; CI is green. Local CGo test execution was unavailable because the worktree lacks cgo/libmo.dylib.

XuPeng-SH (approved):
Reviewed the full diff at exact head `cb8c4cd95ffc` as an internal CN-to-CN network RPC reliability change. No actionable findings.

The server rejects empty/missing/mixed-path cache keys before dereference, transfers successful cache-data ownership to the MORPC response release callback, and releases partial cache acquisitions immediately on `ReadCache` error. The client validates response entries against the target-specific requested index set, bounds, duplicate ordering, and expected data length; invalid hits remain unfinished and flow through the existing LocalFS/S3FS fallback. Non-nil responses and server-side cache data each retain one effective release owner.

Local verification passed: focused regressions (`-count=10`), focused race (`-count=3`), full `pkg/fileservice` tests and race, `pkg/cnservice` handler tests and race, `go build`, `go vet`, and `git diff --check`. Current PR CI is green and the head merges cleanly with current `origin/main`.

PR #25972, opened 2026-07-21
fix(backup): preserve GC metadata watermarks and sizes

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25935
issue #25938

`CopyGCDir` violated two invariants while processing the same filtered GC
metadata set:

1. It iterated over `copyFiles` but detected the last entry with
   `len(metaFiles)`. If an earlier metadata file was filtered because one of
   its objects could not be copied, the last successfully retained metadata
   file did not inherit the checkpoint minimum timestamp.
2. It recorded the GC metadata file's directory-entry size for every referenced
   data object. That incorrect value was serialized into `tae_list` and summed
   into `tae_sum`.

Use the filtered slice for the last-entry decision. After each successful data
object copy, stat the destination object and record its actual size; fail the
backup if the copied object's size cannot be established instead of emitting a
false manifest.

The regression fixture writes real GC metadata and objects through MemoryFS.
It covers filtering plus watermark inheritance, the per-object `tae_list`
field, and the aggregate `tae_sum`. This PR does not claim that object contents
were corrupted; the confirmed defects are the restored watermark invariant
and manifest accounting.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/backup`
- `go build ./pkg/backup` (repository CGo environment)
- `go vet ./pkg/backup` (repository CGo environment)


gouhongshen (approved):
## Codex automated review

No defects found. Existing reviews, issue comments, and inline threads contain no actionable feedback. Focused local tests could not start because cgo/libmo.dylib is absent; current PR CI is green.

XuPeng-SH (approved):
I checked this from correctness / robustness / unhappy-path angles and I did not find a blocker.

The two functional fixes both look right:
- recording the copied object size from `dstFs.StatFile(...)` fixes the manifest accounting bug instead of reusing the `.fullgc` entry size for every referenced object;
- comparing against `len(copyFiles)-1` is the correct way to decide which retained GC metadata file should inherit the watermark when some earlier metadata files are skipped.

The new regression tests cover the two bugs being fixed.

Optional follow-up only: I would still like a few more unhappy-path unit tests in `pkg/backup/tae_gc_test.go` (for example `StatFile` failure after a successful copy, corrupt GC metadata, and the non-inheritance branch when `end >= min`), but I do not see that as a merge blocker for this patch.

PR #25981, opened 2026-07-21
add json_overlap

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24738

• 此分支实现了 JSON_OVERLAPS，并补齐其性能、安全性和回归覆盖。

  - 新增 JSON 顶层 overlap 语义：array/scalar autowrap、object 同键同值、JSON null 与 SQL NULL 区分、数值精确比较。
  - 解决超范围 Decimal：Parse256/Scale 极端指数不 panic；JSON_OVERLAPS 对无法安全使用 Decimal256 的数值走 normalized-key 比较，不展开巨大指数。
  - 优化 const 大数组：索引和 fallback numeric key 按 batch 构建复用，避免逐行重复排序；普通数值数组保持低分配。
  - 修复 parser：完整根节点后出现 malformed suffix 时释放 parse tree。
  - 补充 UT、race 覆盖、性能 benchmark，以及 JSON_OVERLAPS 的 BVT 和 result。
  - 静态检查项已修复：布尔比较改为 !Sign()。


gouhongshen (changes requested):
## Codex automated review

Found two blocking compatibility/performance issues. GitHub CI is green; focused local tests were blocked because cgo/libmo.dylib is absent.

### P1 - Accept all MySQL string storage types (pkg/sql/plan/function/func_builtin_json_overlap.go:89)
The checker accepts only JSON, CHAR, VARCHAR, and TEXT. MatrixOne's IsMySQLString(), used by sibling JSON functions, also includes BINARY, VARBINARY, and BLOB; MySQL likewise accepts these as JSON-text inputs. Consequently JSON_OVERLAPS over valid VARBINARY/BLOB JSON columns fails during function resolution instead of parsing and executing. Use the shared string-type predicate and cover these types.

### P1 - Reuse constant-array indexes even when the constant is longer (pkg/sql/plan/function/func_builtin_json_overlap.go:332)
arraysOverlapViews first moves the shorter array to the left, then only reuses left.prepared. For JSON_OVERLAPS(<large constant array>, column_with_small_arrays), the constant is moved right and its cached index is ignored, so every row scans the entire constant. An 8192-element constant against 8192 one-element rows performs roughly 67 million comparisons instead of one index build plus one logarithmic probe per row. Select the prepared constant using batch-amortized cost before the size swap and add array-valued batch coverage.

XuPeng-SH (changes requested):
I found one meaningful issue before this should merge.

`pkg/sql/plan/function/func_builtin_json_overlap.go` has a performance regression in `arraysOverlapViews()`. The function swaps the operands so that `left` is always the smaller array, but after that swap it only checks `left.prepared`:

```go
if left.document.GetElemCnt() > right.document.GetElemCnt() {
    left, right = right, left
}
...
if left.prepared != nil {
    left.prepared.ensure(left.document)
    return jsonOverlapPreparedArrayOverlaps(left.document, left.prepared, right.document)
}
return w.arraysOverlap(left.document, right.document)
```

For the common shape `json_overlaps(<large const array>, <smaller row array>)`, the const side starts with a prepared index, but the size swap moves it to `right`, so this branch is missed and the code falls back to rebuilding a transient sorted workspace for every row. That defeats the prepared-array optimization exactly for the query shape where it matters most.

Please check `right.prepared` after the size swap as well, e.g. by probing the smaller array against the prepared larger const array, and add a regression test that covers the swapped-operand case.

Non-blocking follow-ups I would still recommend:
- add typed-JSON RHS / both-sides-typed coverage for `json_overlaps`;
- add SQL-visible too-deep JSON and NULL-short-circuit cases;
- add a partial-row skip unhappy-path test where skipped rows contain malformed/deep JSON.

aunjgr (changes requested):
I found two merge-blocking issues on the current head.

1. `pkg/sql/plan/function/func_builtin_json_overlap.go:89`: the overload checker accepts JSON/CHAR/VARCHAR/TEXT but rejects BINARY, VARBINARY, and BLOB. Those are included by the shared `IsMySQLString()` contract used by the other JSON functions, so valid JSON text stored in those column types fails at function resolution. Use `input.Oid.IsMySQLString()` and add typed cases.

2. `pkg/sql/plan/function/func_builtin_json_overlap.go:332`: `arraysOverlapViews` swaps by length and then only considers `left.prepared`. For `json_overlaps(<large constant array>, small_array_column)`, the prepared constant is moved to the right and ignored, forcing a full transient sort/scan for every row. The work grows to rows × constant size despite the constant index. Preserve/probe either prepared operand after the swap and add a batch regression where the constant is the larger array.

heni02 (approved):

daviszhen:
> ## Codex automated review
> Found two blocking compatibility/performance issues. GitHub CI is green; focused local tests were blocked because cgo/libmo.dylib is absent.
> 
> ### P1 - Accept all MySQL string storage types (pkg/sql/plan/function/func_builtin_json_overlap.go:89)
> The checker accepts only JSON, CHAR, VARCHAR, and TEXT. MatrixOne's IsMySQLString(), used by sibling JSON functions, also includes BINARY, VARBINARY, and BLOB; MySQL likewise accepts these as JSON-text inputs. Consequently JSON_OVERLAPS over valid VARBINARY/BLOB JSON columns fails during function resolution instead of parsing and executing. Use the shared string-type predicate and cover these types.
> 
> ### P1 - Reuse constant-array indexes even when the constant is longer (pkg/sql/plan/function/func_builtin_json_overlap.go:332)
> arraysOverlapViews first moves the shorter array to the left, then only reuses left.prepared. For JSON_OVERLAPS(, column_with_small_arrays), the constant is moved right and its cached index is ignored, so every row scans the entire constant. An 8192-element constant against 8192 one-element rows performs roughly 67 million comparisons instead of one index build plus one logarithmic probe per row. Select the prepared constant using batch-amortized cost before the size swap and add array-valued batch coverage.

  - 类型兼容：参数检查改用 IsMySQLString()，统一支持 BINARY、VARBINARY、BLOB 等 JSON 文本类型。
  - const 数组性能：不再只检查较短数组；同时评估左右两侧 prepared index，按 batch 总成本选择复用方案，避免每行重建和排序大数组索引。
  - Decimal panic：大指数 DECIMAL 先做范围判断；无法进入 Decimal256 快速路径时使用安全的精确比较，不再解析失败后按零处理。
  - parser 资源释放：根 JSON 后存在非法 suffix 时，所有返回路径都调用 Free() 归还 parse tree。
  - 深度限制：文本 JSON 在解析阶段限制 100 层，typed JSON 在比较前校验，超深输入提前报错。
  - 回归保障：增加类型矩阵、左右 const 大数组、极端指数、malformed suffix、深度边界及 batch allocation 的 UT/BVT/benchmark。

daviszhen:
> I found one meaningful issue before this should merge.
> 
> `pkg/sql/plan/function/func_builtin_json_overlap.go` has a performance regression in `arraysOverlapViews()`. The function swaps the operands so that `left` is always the smaller array, but after that swap it only checks `left.prepared`:
> 
> ```go
> if left.document.GetElemCnt() > right.document.GetElemCnt() {
>     left, right = right, left
> }
> ...
> if left.prepared != nil {
>     left.prepared.ensure(left.document)
>     return jsonOverlapPreparedArrayOverlaps(left.document, left.prepared, right.document)
> }
> return w.arraysOverlap(left.document, right.document)
> ```
> 
> For the common shape `json_overlaps(<large const array>, <smaller row array>)`, the const side starts with a prepared index, but the size swap moves it to `right`, so this branch is missed and the code falls back to rebuilding a transient sorted workspace for every row. That defeats the prepared-array optimization exactly for the query shape where it matters most.
> 
> Please check `right.prepared` after the size swap as well, e.g. by probing the smaller array against the prepared larger const array, and add a regression test that covers the swapped-operand case.
> 
> Non-blocking follow-ups I would still recommend:
> 
> * add typed-JSON RHS / both-sides-typed coverage for `json_overlaps`;
> * add SQL-visible too-deep JSON and NULL-short-circuit cases;
> * add a partial-row skip unhappy-path test where skipped rows contain malformed/deep JSON.

  - 类型兼容：参数检查改用 IsMySQLString()，统一支持 BINARY、VARBINARY、BLOB 等 JSON 文本类型。
  - const 数组性能：不再只检查较短数组；同时评估左右两侧 prepared index，按 batch 总成本选择复用方案，避免每行重建和排序大数组索引。
  - Decimal panic：大指数 DECIMAL 先做范围判断；无法进入 Decimal256 快速路径时使用安全的精确比较，不再解析失败后按零处理。
  - parser 资源释放：根 JSON 后存在非法 suffix 时，所有返回路径都调用 Free() 归还 parse tree。
  - 深度限制：文本 JSON 在解析阶段限制 100 层，typed JSON 在比较前校验，超深输入提前报错。
  - 回归保障：增加类型矩阵、左右 const 大数组、极端指数、malformed suffix、深度边界及 batch allocation 的 UT/BVT/benchmark。

daviszhen:
> I found two merge-blocking issues on the current head.
> 
> 1. `pkg/sql/plan/function/func_builtin_json_overlap.go:89`: the overload checker accepts JSON/CHAR/VARCHAR/TEXT but rejects BINARY, VARBINARY, and BLOB. Those are included by the shared `IsMySQLString()` contract used by the other JSON functions, so valid JSON text stored in those column types fails at function resolution. Use `input.Oid.IsMySQLString()` and add typed cases.
> 2. `pkg/sql/plan/function/func_builtin_json_overlap.go:332`: `arraysOverlapViews` swaps by length and then only considers `left.prepared`. For `json_overlaps(<large constant array>, small_array_column)`, the prepared constant is moved to the right and ignored, forcing a full transient sort/scan for every row. The work grows to rows × constant size despite the constant index. Preserve/probe either prepared operand after the swap and add a batch regression where the constant is the larger array.

  - 类型兼容：参数检查改用 IsMySQLString()，统一支持 BINARY、VARBINARY、BLOB 等 JSON 文本类型。
  - const 数组性能：不再只检查较短数组；同时评估左右两侧 prepared index，按 batch 总成本选择复用方案，避免每行重建和排序大数组索引。
  - Decimal panic：大指数 DECIMAL 先做范围判断；无法进入 Decimal256 快速路径时使用安全的精确比较，不再解析失败后按零处理。
  - parser 资源释放：根 JSON 后存在非法 suffix 时，所有返回路径都调用 Free() 归还 parse tree。
  - 深度限制：文本 JSON 在解析阶段限制 100 层，typed JSON 在比较前校验，超深输入提前报错。
  - 回归保障：增加类型矩阵、左右 const 大数组、极端指数、malformed suffix、深度边界及 batch allocation 的 UT/BVT/benchmark。

gouhongshen (changes requested):
## Codex automated review

One blocking compatibility defect remains: native scalar SQL arguments are rejected by the new overload checker. The previously reported string-type and prepared-array issues appear fixed in this exact head.

### P1 - Allow implicitly castable native scalar arguments (pkg/sql/plan/function/func_builtin_json_overlap.go:89)
The checker accepts only JSON, MySQL string types, and ANY, so native scalar expressions such as INT64, BOOL, and DECIMAL are rejected before execution. For example, `JSON_OVERLAPS(1, '1')` binds as `(T_int64, T_varchar)` and fails; the PR's own `TestJSONOverlapsCheckFn` currently expects that combination to error. Existing JSON checkers in `func_builtin_json.go` use `fixedImplicitTypeCast(input, types.T_varchar)` for non-string inputs. This also conflicts with MySQL's documented scalar JSON_OVERLAPS behavior: [official documentation](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-overlaps). Add the implicit-cast path and SQL tests covering native scalar types.

aunjgr (approved):
Reviewed the current head independently. The earlier BINARY/VARBINARY/BLOB resolution gap and large-constant-array reuse issue are fixed. I audited the JSON top-level overlap rules, SQL NULL versus JSON null, malformed-suffix cleanup, text/typed-JSON depth limits, exact mixed numeric ordering (including extreme Decimal256 exponents), prepared const-array cost selection/cache lifetime, and select-list skipping. Focused unit and race tests pass for bytejson, decimal, and the function package; CI is green. I do not consider rejection of native SQL numeric/bool arguments a defect: MySQL's JSON_OVERLAPS accepts JSON documents and string/blob JSON text, while its type-resolution path rejects non-JSON native numeric types; scalar examples are scalar JSON documents such as '1'. No blocking issues found.

XuPeng-SH (approved):
Reviewed the exact current head e5765549f.

The two earlier blockers are fixed: all repository MySQL string storage types follow the shared IsMySQLString contract, and prepared indexes on either constant array survive operand-size ordering with batch-amortized cost selection.

I also rechecked the disputed native SQL scalar behavior against MySQL 8.4.10. JSON_OVERLAPS(1, "1"), BOOL, and DECIMAL arguments all fail with error 3146 because the argument must be a JSON string or JSON type. Therefore the current checker is correct to reject native numeric/bool arguments; scalar examples use JSON documents such as string JSON text or an explicit JSON value.

The full closure for top-level overlap semantics, SQL NULL versus JSON null and left-to-right validation, malformed-suffix parse-tree cleanup, text/typed JSON depth limits, exact mixed numeric comparison including extreme exponents, constant-array cache lifetime/cost selection, and selected-row skipping has no remaining blocker.

Fresh local validation: full tests for bytejson, container/types, and sql/plan/function; focused race tests; go vet; and go build all completed with exit code 0. CI is green.

daviszhen:
> ## Codex automated review
> One blocking compatibility defect remains: native scalar SQL arguments are rejected by the new overload checker. The previously reported string-type and prepared-array issues appear fixed in this exact head.
> 
> ### P1 - Allow implicitly castable native scalar arguments (pkg/sql/plan/function/func_builtin_json_overlap.go:89)
> The checker accepts only JSON, MySQL string types, and ANY, so native scalar expressions such as INT64, BOOL, and DECIMAL are rejected before execution. For example, `JSON_OVERLAPS(1, '1')` binds as `(T_int64, T_varchar)` and fails; the PR's own `TestJSONOverlapsCheckFn` currently expects that combination to error. Existing JSON checkers in `func_builtin_json.go` use `fixedImplicitTypeCast(input, types.T_varchar)` for non-string inputs. This also conflicts with MySQL's documented scalar JSON_OVERLAPS behavior: [official documentation](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-overlaps). Add the implicit-cast path and SQL tests covering native scalar types.

• 针对 reviewer comments，主要完成以下修改：

  - 修复 native scalar 参数兼容：INT/BOOL/DECIMAL/FLOAT 等可隐式转换类型统一转为 VARCHAR，再复用现有 JSON 解析流程；不修改 overlap 核心语义。
  - 修复超大 DECIMAL panic：增加安全数值比较回退路径，并强化 Decimal256.Scale/Parse256 的指数和溢出保护。
  - 优化常量大数组：数组排序索引改为 batch-scoped 缓存，每批只构建一次，避免每行重复排序和分配。
  - 修复 malformed suffix 清理：解析完成根节点后若尾部非法，确保归还 parse tree。
  - 增加 100 层 JSON 深度限制，在解析宽叶之前提前拒绝 depth-101 输入。
  - 补充 UT/BVT：覆盖 native scalar、SQL NULL、非法 JSON、超大指数、常量数组缓存、深度边界及异常清理路径。

  相关包完整测试和高风险用例重复测试均已通过。

gouhongshen (changes requested):
## Codex automated review

The earlier request to accept native SQL scalars is withdrawn as disproven. The latest change implementing it introduces one blocking compatibility defect; one non-blocking prepared-index cost issue also remains.

### P1 - Do not reinterpret native SQL scalars as JSON text (pkg/sql/plan/function/func_builtin_json_overlap.go:92)
This implicit-cast branch was added after the latest discussion, but its premise is incorrect. MySQL requires `json_doc` arguments to be JSON-typed values or strings containing JSON, as documented by the [JSON argument contract](https://dev.mysql.com/doc/refman/8.4/en/json.html) and [JSON_OVERLAPS signature](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-overlaps); native INT, BOOL, and DECIMAL arguments fail with error 3146. Here they instead become successful VARCHAR inputs. The new BVT demonstrates semantic corruption: `json_overlaps(true, '1')` returns 1 because SQL TRUE becomes text `1` and is parsed as JSON number 1, not JSON boolean true. Restore the JSON/MySQL-string/ANY-only checker and require explicit JSON or string conversion.

### P2 - Stop charging index construction after the prepared index is ready (pkg/sql/plan/function/func_builtin_json_overlap.go:373)
Every row calls `jsonOverlapPreparedArrayBatchCost`, which always includes `count*log(count)` preparation cost even after `prepared.ensure` has set `ready`. For an 8192-element constant and two heterogeneous probe rows, an 8192-element first probe builds the index, but a subsequent one-element probe rejects that already-built index because the estimate charges its 114688-comparison construction again. The fallback then probes all 8192 constant elements instead of doing one logarithmic lookup. Account for `prepared.ready` in the cost or retain the batch decision; current tests only repeat uniform probe shapes.

daviszhen:
> ## Codex automated review
> The earlier request to accept native SQL scalars is withdrawn as disproven. The latest change implementing it introduces one blocking compatibility defect; one non-blocking prepared-index cost issue also remains.
> 
> ### P1 - Do not reinterpret native SQL scalars as JSON text (pkg/sql/plan/function/func_builtin_json_overlap.go:92)
> This implicit-cast branch was added after the latest discussion, but its premise is incorrect. MySQL requires `json_doc` arguments to be JSON-typed values or strings containing JSON, as documented by the [JSON argument contract](https://dev.mysql.com/doc/refman/8.4/en/json.html) and [JSON_OVERLAPS signature](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-overlaps); native INT, BOOL, and DECIMAL arguments fail with error 3146. Here they instead become successful VARCHAR inputs. The new BVT demonstrates semantic corruption: `json_overlaps(true, '1')` returns 1 because SQL TRUE becomes text `1` and is parsed as JSON number 1, not JSON boolean true. Restore the JSON/MySQL-string/ANY-only checker and require explicit JSON or string conversion.
> 
> ### P2 - Stop charging index construction after the prepared index is ready (pkg/sql/plan/function/func_builtin_json_overlap.go:373)
> Every row calls `jsonOverlapPreparedArrayBatchCost`, which always includes `count*log(count)` preparation cost even after `prepared.ensure` has set `ready`. For an 8192-element constant and two heterogeneous probe rows, an 8192-element first probe builds the index, but a subsequent one-element probe rejects that already-built index because the estimate charges its 114688-comparison construction again. The fallback then probes all 8192 constant elements instead of doing one logarithmic lookup. Account for `prepared.ready` in the cost or retain the batch decision; current tests only repeat uniform probe shapes.

• 针对 reviewer comment 的修改：

  - 禁止 INT、BOOL、DECIMAL、FLOAT 等 native SQL scalar 被隐式转成 JSON 文本；仅接受 JSON、MySQL 字符串类型和 ANY。
  - 非法 JSON 参数改为明确提示参数位置，例如 argument 1 / argument 2，文本接近 MySQL。
  - 保留 BINARY、VARBINARY、BLOB 作为 JSON 文本输入的支持。
  - prepared constant-array 索引会比较左右候选；索引 ready 后不再重复计入构建成本。
  - 更新 UT 与 BVT .test/.result：覆盖 native scalar 拒绝、参数序号、二进制字符串类型及 prepared-index 场景。

gouhongshen (approved):
## Codex automated review

One non-blocking MySQL wire-compatibility issue remains.

### P2 - Return MySQL ER_INVALID_TYPE_FOR_JSON for rejected arguments (pkg/common/moerr/error.go:404)
The new checker correctly rejects `JSON_OVERLAPS(1, '1')`, but `ErrInvalidTypeForJSON` is mapped to `ER_UNKNOWN_ERROR`/`HY000`. The protocol therefore sends the internal error code 20316 (it falls back when the MySQL code is unknown), rather than MySQL's `ER_INVALID_TYPE_FOR_JSON` 3146 with SQLSTATE `22032`, despite the exact 3146 definition already existing in `mysql_error_define.go`. Clients that branch on MySQL error codes or SQLSTATE receive an incompatible result. Map this error to `ER_INVALID_TYPE_FOR_JSON` and add a protocol-level code/state assertion.

PR #25983, opened 2026-07-21
fix: harden MORPC decoding and deflake validation test

### LINKED ISSUE #25948: [Bug]: TestValidateService flaky — cannot find lockservice address / context deadline exceeded
## Problem

`TestValidateService` in `pkg/lockservice/rpc_test.go:617` fails intermittently:

```
ERROR: cannot find lockservice address (target=UNKNOWN, cns=[s1, s2])
ERROR: failed to ping lock service (serviceID=UNKNOWN): cannot connect to backend: none
ERROR: failed to ping lock service (serviceID=s1): context deadline exceeded
ERROR: close unix: use of closed network connection
Error: Should be false
```

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-21 | [PR #25815](https://github.com/matrixorigin/matrixone/actions/runs/29808237385/job/88563286508) | UT Test on Ubuntu/x86 |

## Root Cause

Race condition in lockservice RPC test setup — the lockservice address is not yet registered when ValidateService tries to connect, or connections are closed prematurely during the test.

## Suggested Fix

1. Add wait/retry for lockservice address registration before validation
2. Or use synchronization barriers in test setup
3. Or increase connection timeout tolerance

### LINKED ISSUE #25949: [Bug]: morpc decoder panics on truncated optional headers
## Bug description

The morpc decoder trusts optional-header flags without verifying that the complete length-prefixed frame contains the corresponding header bytes.

A peer can send a valid outer frame with body length 1 containing only one of these flags:

- `flagChecksumEnabled`: `readChecksum` reads 8 missing bytes.
- `flagHashPayload`: `readPayloadSize` reads 4 missing bytes.
- `flagStreamingMessage`: `readStreaming` reads 4 missing bytes.

All three paths panic with an index-out-of-range error instead of rejecting the malformed frame.

## Impact

A malformed or incompatible morpc peer can crash the process from the network decode path. The panic occurs after the outer goetty length codec has accepted a complete frame, so this is reachable through the normal public `Codec.Decode` entry point.

## Reproduction

Construct a frame consisting of a 4-byte length field set to 1 followed by one optional-header flag byte, then call `Codec.Decode`.

The white-box audit tested checksum, payload, and stream flags in three runs. All 9/9 cases panicked.

```bash
go test ./pkg/common/morpc \
  -run '^TestAuditCodecRejectsTruncatedOptionalHeadersWithoutPanic$' \
  -count=3 -v
```

Observed panic locations:

- `readChecksum` → `buf.Byte2Uint64`
- `readPayloadSize` → `buf.Byte2Int`
- `readStreaming` → `buf.Byte2Uint32`

## Expected behavior

The decoder should validate the remaining frame size before every optional-header read and return a decode error for truncated or inconsistent frames without panicking.

## Tested version

- Audit baseline: `main@7f9ef5de9bb9337d8db0a9b15c44cba3b2838aa2`
- GitHub `main` was checked before filing and still performs these reads without length checks.

## Suggested direction

Perform cumulative bounds checks before reading checksum, payload size, custom headers, and stream sequence. Add table-driven tests for every flag combination and all truncation lengths.

### LINKED ISSUE #25950: [Bug]: morpc receive max body size is not enforced after decompression
## Bug description

`WithCodecMaxBodySize` is enforced against `Message.ProtoSize()` on the sending side, but the receiving side only applies the outer goetty frame limit to the compressed wire size.

When compression is enabled, `baseCodec.uncompress` repeatedly doubles its destination allocation until LZ4 succeeds, then returns the result without comparing the decompressed body size with `c.maxBodySize`.

As a result, a logical message body larger than the configured receive limit is accepted whenever its compressed representation fits the outer frame limit.

## Impact

- The configured RPC body limit is not symmetric between sender and receiver.
- A highly compressible message can consume substantially more receive memory than the configured body limit allows.
- The allocation loop has no codec-level upper bound derived from `maxBodySize`.

## Reproduction

1. Encode a 64 KiB repetitive message body with compression enabled and an encoder limit above 64 KiB.
2. Decode the frame with compression enabled and `WithCodecMaxBodySize(1024)`.
3. The compressed frame is below 1 KiB, while the decompressed body is 64 KiB.

The audit was run three times. All 3/3 runs decoded the oversized logical body successfully instead of returning a size error.

```bash
go test ./pkg/common/morpc \
  -run '^TestAuditCodecRejectsOversizedDecompressedBody$' \
  -count=3 -v
```

## Expected behavior

The receiver must reject a decompressed body or payload that exceeds the configured maximum body size. Allocation growth should stop at the configured limit.

## Tested version

- Audit baseline: `main@7f9ef5de9bb9337d8db0a9b15c44cba3b2838aa2`
- GitHub `main` was checked before filing and still returns decompressed buffers without a `maxBodySize` check.

## Suggested direction

Bound decompression growth by the configured receive limit and validate the final decompressed body/payload size before unmarshalling or cloning it.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25948
Fixes #25949
Fixes #25950

- Uses the shared 10-second RPC test response timeout for the successful TestValidateService request while preserving the short timeout paths that intentionally validate failure and retry behavior.
- Validates every optional MORPC header before reading it, rejects incompatible payload/compression flags, and makes short internal ping/pong messages return decode errors instead of panicking.
- Cancels a decoded deadline context when a later decode step fails, preventing failed frames from retaining long-lived timers.
- Bounds LZ4 decompression by the configured exclusive maximum for the complete logical message. The payload receives only the remaining budget after the body, so compressed input cannot bypass the receive limit.
- Adds regression coverage for truncated header combinations, trace headers, internal messages, invalid flags, context cleanup, oversized decompressed bodies and payloads, and the aggregate body-plus-payload limit.

Design decisions:

- The maximum remains exclusive, matching the existing sender check.
- Body and detached payload share one logical-message budget.
- The wire format and successful decode path are unchanged.

Validation:

- make thirdparties
- go build ./pkg/common/morpc ./pkg/lockservice
- go vet ./pkg/common/morpc ./pkg/lockservice
- .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/common/morpc
- .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/lockservice
- git diff --check

Independent review completed with no remaining merge blockers.


XuPeng-SH (approved):
I checked this from correctness / robustness / unhappy-path angles and I did not find a blocker.

The hardening itself looks right end-to-end:
- the decode path now bounds-checks every optional field before reading it;
- the payload/body slicing check no longer subtracts an untrusted payload size from `len(data)`;
- decompression is now budgeted against `maxBodySize` so compressed input cannot bypass the logical body limit;
- decoded deadline contexts are canceled on any later decode failure;
- the trace-header off-by-one and truncated internal-message-body cases are both fixed.

The new unit coverage already hits the main crash / OOM / cleanup paths this patch is meant to close.

Optional follow-up only: I would still like a few more unhappy-path tests around an oversized uncompressed body, invalid `HeaderCodec.Decode()` consumed lengths, and one post-header failure case besides the current context-header test, but I do not see those as merge blockers here.

Issue #25930, opened 2026-07-21
[Bug]: dispatch.shuffleToAllFunc panic: index out of range on multi-CTE LEFT JOIN + RANK (jinpan invoice_acdoca_link)


heni02:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

826a3e3d294ba62c7c4da217d1445f3c568ce4d9

### Other Environment Information

```
- Hardware parameters: single-node MO (docker, host port 16001)
- OS type: Linux
- Others: jinpan_001 dataset; Fact+Dim Join suite (1000 SQLs); result join_20260721_010109
```

### Actual Behavior

All **37/37** variants of template `invoice_acdoca_link` fail immediately (`cost_ms=0`) with:

```
(20101) internal error: panic runtime error: index out of range [36] with length 1
  dispatch.shuffleToAllFunc   pkg/sql/colexec/dispatch/sendfunc.go:245
  dispatch.(*Dispatch).Call   pkg/sql/colexec/dispatch/dispatch.go:166
  ... ParallelRun / MergeRun ...
```

**Not the same as #25802** (`hashjoin.(*container).finalize` nil pointer on `bill_collection_p4`). Different stack, operator, and SQL template.

### Expected Behavior

Query returns rows (or empty set) without panicking the pipeline.

### Steps to Reproduce

1. Load jinpan_001 fact/dim + receipts tables (same schema as AP Regression).
2. Execute the SQL below (`j_batch_001` / `@q0004_invoice_acdoca_link`).
3. Observe CN panic / client `(20101)` with the stack above.

```sql
WITH agg AS (
WITH invoice_src AS (
  SELECT z.belnr, z.bukrs, z.gjahr, z.zfph, z.zfpje, z.lifnr
  FROM dwd_dcp.dwd_s4_ztfi0007 z
  WHERE COALESCE(z.zfpje, 0) <> 0
    AND z.belnr IS NOT NULL AND z.belnr <> ''
    AND z.gjahr IN ('2022', '2023', '2024', '2025')
    AND z.bukrs IN ('1000', '1100', '1200', '1300', '2000', '2100', '2200', '3000')
),
acdoca_links AS (
  SELECT DISTINCT a.ebeln, a.belnr, a.rbukrs AS bukrs, a.gjahr
  FROM invoice_src z
  INNER JOIN dwd_dcp.dwd_s4_acdoca a
[...]

PR #25987, opened 2026-07-21
fix(compile): validate distributed shuffle reuse topology

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25930

A logical `ShuffleMethod_Reuse` decision does not by itself guarantee that the
compiled probe scopes have the physical layout required by a distributed
shuffle join. A packed local scope (`Mcpu = DOP`) represents all local buckets,
but the distributed reuse path treated that single scope as one bucket. The
build shuffle could therefore emit bucket N to a dispatch with only one
receiver, causing the reported `shuffleToAllFunc` index-out-of-range panic.

This change reuses distributed probe scopes only when there is exactly one
`Mcpu=1` scope per global bucket in stage-node order. Otherwise it materializes
the normal distributed shuffle layout. Multi-CN DEDUP also falls back because
its scope normalization merges the per-bucket layout before dispatch wiring.

Regression coverage includes packed-scope fallback, valid materialized reuse,
misordered multi-CN scopes, multi-CN DEDUP normalization, and a focused BVT
using a shuffled `DISTINCT` probe with a merged `UNION ALL` build. The BVT
panics on unpatched main with `index out of range [1] with length 1` and passes
18/18 statements with this fix.

Validation:

- `make build`
- `go vet -mod=readonly ./pkg/sql/compile` with repository CGo flags
- `.agents/skills/mo-dev/scripts/mo-cgo-test -v -count=1 -timeout=240s ./pkg/sql/compile`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -v -count=1 -timeout=240s -run '^TestDistributedShuffleJoin' ./pkg/sql/compile`
- `mo-tester` BVT: `shuffle_reuse_topology.sql` (18/18)


XuPeng-SH (approved):
Deep review completed on 2aee1d3. I traced logical shuffle reuse through physical scope materialization, stage-node/bucket ordering, local/remote dispatch index wiring, hash-build/join consumption, multi-CN DEDUP normalization, and cleanup/wait boundedness. No blocking findings.

Verified locally: focused topology regressions x20; focused race; full pkg/sql/compile and full race; pkg/sql/colexec/dispatch; go build and go vet. I also clean-merged latest main a327eaa and reran the full compile package plus focused race successfully.

Ubuntu UT and SCA are green. Compose/standalone/coverage jobs were still running at review time.

gouhongshen (approved):
## Codex automated review

No actionable regressions found. The topology validation correctly prevents invalid distributed reuse and preserves valid reuse; existing review and CI evidence are consistent.

PR #25989, opened 2026-07-21
[Feature]: support native DUMP TABLE and LOAD TABLE

aptend:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25782

Adds SQL-native export and attachment of immutable table objects so large benchmark datasets used to reproduce #25782 can be reused without re-importing source data.

- Adds `DUMP TABLE ... TO ... [METADATA ONLY]` and `LOAD TABLE ... FROM ...`.
- Preserves main/index-table topology and validates schema, immutable object metadata, fixture completeness, object size/checksum, and an empty destination.
- Rejects appendable or in-memory source data, partitions, and auto-increment tables in this first version.
- Supports local and stage-backed fixtures, including provider-side CopyObject for AWS S3, COS, OSS, and MinIO with streaming fallback.
- Bounds object-copy and post-copy validation concurrency at 32.

Validation:

- **TPCH 100G LOAD completed in 9 seconds** with row-count validation in TKE.
- **TPCH 1T LOAD completed in 44 seconds** with row-count validation in TKE.
- `go test -count=1 -timeout=600s ./pkg/fileservice ./pkg/frontend ./pkg/sql/parsers/dialect/mysql ./pkg/vm/engine/disttae`
- focused frontend concurrency tests pass under `-race`
- `go build` and `go vet` pass for all modified packages.


gouhongshen (changes requested):
## Codex automated review

Five blocking security, correctness, availability, and compatibility issues remain. Existing PR discussion contains no prior technical findings. The simple PK BVT passes, but it does not cover these paths.

### P1 - Do not authorize physical dumps with SELECT alone (pkg/frontend/authenticate.go:6584)
DUMP TABLE exports immutable data objects and separate tombstone objects, not the SELECT-visible row set. A flushed data object still contains values removed by tombstones, so a SELECT-only user can dump to a writable stage, remove Tombstone entries from the editable manifest, and LOAD the remaining objects into an empty same-schema table, recovering deleted values. This operation needs a privileged physical-backup capability or must export only visible rows.

### P1 - Bound manifest collections before decoding them (pkg/frontend/table_dump.go:449)
The 64 MiB file limit does not bound decoded heap usage: such a JSON document can contain millions of empty relation/object entries, and json.Unmarshal allocates all corresponding structs and slices before tableDumpMaxObjects is checked. A user loading a controlled stage can therefore consume gigabytes of CN heap and cause OOM. Count and reject collections while streaming the JSON, before materializing them.

### P1 - Preserve foreign-key integrity during LOAD TABLE (pkg/frontend/table_dump.go:884)
DUMP TABLE permits tables with foreign keys, but LOAD TABLE attaches their objects directly through Relation.Write and never executes the normal INSERT planner's parent-row checks. A child row valid in the source can therefore be loaded when the equivalent target parent table lacks the referenced row; the schema hash still matches because it validates DDL, not parent contents. Reject externally referenced tables or validate their constraints before attaching objects.

### P1 - Do not infer copy authorization from endpoint equality (pkg/fileservice/aws_sdk_v2.go:72)
The server-side-copy capability considers two AWS stores compatible solely by SDK type and endpoint, then issues the request with the destination client's credentials. For same-endpoint source and destination buckets using separate credentials—a normal external-stage setup—the destination principal may lack source GetObject permission, so CopyObject returns 403 and DUMP/LOAD fails even though source Read plus destination Write is authorized. MinIO, Aliyun, and QCloud adapters have the same credential-domain problem.

### P1 - Populate schema hashes for hidden index relations (pkg/frontend/table_dump.go:194)
Only the main relation receives a SchemaHash; every appended index relation leaves it empty. Target references are built identically, so the LOAD comparison accepts empty against empty and never validates the hidden table's physical schema. The main hash covers logical index DDL, not plugin-generated hidden columns, so a fixture loaded by a build with a changed index layout can attach incompatible objects and corrupt index reads.

gouhongshen (changes requested):
## Codex automated review

The earlier review findings are addressed in this head, but three blocking correctness and lifecycle issues remain. Focused local tests could not run because Go is unavailable.

### P1 - Provider-side copies omit checksums (pkg/frontend/table_dump.go:317)
The provider-side copy path records only the source size and returns an empty SHA256; the batch path does the same for copied objects. LOAD consequently performs only a size check when the fixture is replaced or corrupted with same-length contents, silently attaching invalid data. Compute and persist a checksum, or verify the copied bytes before accepting the dump.

### P1 - Newly installed LOAD objects are protected as pre-existing (pkg/frontend/table_dump.go:1096)
ProtectCloneFiles classifies names as transaction-local only when they already appear in txn.writes. It is called before submitTableDumpObjects creates those writes, so objects newly copied from an external fixture are marked shared. Rollback then filters shared names out of gcFiles, leaving copied objects orphaned after transaction or statement rollback. A partial install can also return before protection and leave earlier copies behind.

### P1 - Hidden fulltext index auto-increment is not restored (pkg/frontend/table_dump.go:992)
validateTableDumpSchema checks only the main table, while LOAD later attaches every hidden index relation. Fulltext hidden tables define __mo_fake_pk_col as an auto-increment primary key, but LOAD directly writes its object metadata without resetting the target auto-increment service. Subsequent fulltext post-DML inserts allocate from the initial range and can collide with loaded hidden-index rows. Reject auto-increment on all relations or advance the hidden table's sequence from the loaded data.

XuPeng-SH (changes requested):
Deep review of head `6e8a639e3`: five merge blockers remain.

1. **P1 — Validate physical `ObjectStats` before expansion** (`pkg/frontend/table_dump.go:928-950`). The editable manifest controls block count, extent, row count, size, and zone map. Only length/name/appendable are checked before `ExpandObjectStatsToBatch`. For example, 1,000 allowed objects with `BlkCnt=65535` drive about 65 million `BlockInfo` appends, while false extents/counts can be committed without matching the object footer. Reconstruct or validate stats from the physical object and cap total blocks before allocation.

2. **P1 — Do not authorize unauthenticated physical attachment with ordinary INSERT** (`pkg/frontend/authenticate.go:6591-6594`). The manifest is editable and is not proven to be the complete output of one DUMP. Combining objects from same-schema fixtures can violate PK/unique constraints, and removing tombstones resurrects deleted rows; direct `Relation.Write` bypasses normal constraint and lock operators. Make LOAD a privileged physical-restore operation and/or authenticate the manifest, or validate logical constraints before attachment.

3. **P1 — Serialize the empty-target admission check** (`pkg/frontend/table_dump.go:1049-1102`). `Rows()==0` is a snapshot check with no lock held through submission and commit. Two LOADs, or LOAD plus INSERT, can both observe empty and commit distinct objects, producing a mixed dataset or duplicate keys. Hold a distributed exclusive table/admission lock until transaction completion and cover LOAD/LOAD plus LOAD/INSERT races.

4. **P1 — Restore hidden-table auto-increment state** (`pkg/frontend/table_dump.go:992,1016,1098`). Only the main table is checked, but fulltext hidden tables contain auto-increment `__mo_fake_pk_col`. LOAD attaches their existing values without advancing the target increment service, so later fulltext maintenance reuses loaded keys and fails or corrupts the index. Advance hidden sequences or reject every relation containing auto-increment columns.

5. **P1 — Persist checksums for provider-side copies** (`pkg/frontend/table_dump.go:301-317,394-408`). Server-side copies record only size and leave SHA256 empty, so LOAD accepts same-length fixture replacement/corruption despite the checksum guarantee. Persist and require a content checksum, or explicitly define and document the weaker integrity contract.

Verification on this exact head passed for the four modified owning packages, focused copy concurrency tests under `-race -count=3`, and `go vet`. The failures above are uncovered semantic and unhappy-path gaps, not test/build failures.

XuPeng-SH (changes requested):
Deep review of exact head `b444fe208e89`: the five blockers from my previous review are addressed, but two independent merge blockers remain.

1. **P1 — Include the physical column layout in the compatibility check** (`pkg/frontend/table_dump.go:240`). `ConstructCreateTableSQL` renders logical DDL and ignores `ColDef.Seqnum` and hidden physical columns, but LOAD attaches existing objects without rewriting them. For example, a source created as `(a INT, b INT)` and then altered with `DROP COLUMN a` has `b` at physical seqnum 1; a fresh target created as `(b INT)` has `b` at seqnum 0. Both reconstruct to the same canonical DDL and pass this hash, while target reads address the imported object with the wrong seqnum and can silently return old `a` values as `b`. The physical-object validator only checks stats, extents, counts, and zone maps; it does not compare the target column layout. Persist and compare a per-relation physical schema signature including stored/hidden columns, types, seqnums, and next-seqnum, or rewrite objects. Add an evolved-source versus fresh-target rejection test.

2. **P1 — Do not make file deletion a blocking prerequisite for transactional rollback** (`pkg/vm/engine/disttae/types.go:533`, `pkg/vm/engine/disttae/txn.go:2609`). This Delete runs while the disttae transaction mutex and the pessimistic target-table plus synthetic global LOAD locks remain owned. The two-minute `cleanupCtx` does not actually bound the S3 path: `objectStorageSemaphore.Delete` first calls the context-unaware `o.acquire()`, so exhausted slots behind hung object I/O can block here indefinitely and one failed LOAD can block every LOAD plus writes to its targets. If Delete returns an error, `Transaction.Rollback` also returns before `gcObjsByIdxRange` and `delTransaction`, leaking clone-cache state, batches, segment IDs, the compact worker, and workspace quota after the client closes the operator. Make acquisition and cleanup fail-safe, and ensure rollback/unlock plus local teardown always run; if synchronous deletion cannot complete safely, leave names protected for later GC instead of aborting the rollback control path. Cover both a blocked Delete and a Delete error.

Validation on this exact head: all four owning packages pass with `-count=1`; focused fileservice, frontend, and disttae tests pass under `-race -count=3`; `go vet` passes; `git diff --check` passes.

aptend:
Addressed the rollback cleanup blocker in 6cad9264d:

- object-storage Delete now observes context cancellation while waiting for the semaphore;
- load-file deletion no longer holds the disttae transaction mutex across file-service I/O;
- synchronous cleanup is bounded and Transaction.Rollback always runs workspace GC and delTransaction, even when Delete blocks or fails;
- added blocked-Delete, Delete-error, mutex-independence, and semaphore-cancellation coverage.

I did not add a physical-layout signature after checking the current ALTER execution contract. DROP COLUMN, ADD COLUMN, CHANGE COLUMN, and layout-changing MODIFY COLUMN resolve to AlterTable_COPY in build_alter_table.go. COPY constructs a fresh temporary table from ConstructCreateTableSQL and rewrites the rows through INSERT before replacing the original table, so the example of DROP COLUMN retaining b at seqnum 1 is not reachable. Current INPLACE operations do not remap physical column seqnums. Adding a second schema signature for an unreachable state would add manifest/versioning complexity without covering a supported layout transition. If a future INPLACE ALTER changes physical seqnums, that ALTER should either preserve the attachment contract explicitly or add compatibility metadata at that point.

Validation: full fileservice, disttae, and frontend package tests passed; fileservice and disttae passed under -race; go build, go vet, and git diff --check passed.

pkg/frontend/table_dump.go
```diff
@@ -0,0 +1,1362 @@
...
+	}
+	sum := sha256.Sum256(data)
+	return hex.EncodeToString(sum[:]), nil
+}
+
+func validateTableDumpSchema(def *plan.TableDef) error {
+	if def == nil {
+		return moerr.NewInternalErrorNoCtx("table definition is unavailable")
+	}
+	if def.Partition != nil {
```
XuPeng-SH, line 308: [P1] Reject relation kinds that this object-metadata path cannot represent

This validation still accepts temporary, external/source, view, and sequence-like relations. For a temporary table, `Database.Relation` can return `*memoryengine.Table`, but `NewImmutableTableMetaReader` immediately asserts every non-`*txnTable` relation to `*txnTableDelegate`, so an admin-visible `DUMP TABLE` can panic. External/source rows have the opposite failure mode: their data is not represented by disttae `PartitionState`, so DUMP can successfully publish an empty fixture for a table that returns rows through its external scan path. Please reject `IsTemporary` and all non-object-backed relation kinds (and enforce the same rule for LOAD) before hashing/reading, with DUMP/LOAD regression tests for these kinds.

pkg/frontend/table_dump.go
```diff
@@ -0,0 +1,1362 @@
...
+			sharedObjects = sharedObjects[:0]
+		}
+		if !manifest.MetadataOnly {
+			if err = installTableDumpObjects(ctx, dumpFS, targetFS, relationDump.Objects, func(name string) {
+				tracker.TrackLoadFiles(name)
+			}); err != nil {
+				return err
+			}
+		}
+		if err = validatePhysicalTableDumpObjects(
```
XuPeng-SH, line 1366: [P1] Bind each physical object to the target schema before attaching it

The checks above compare the target with schema hashes supplied by the manifest, and this validator only proves that the manifest `ObjectStats` are self-consistent with the referenced object. A valid object from an incompatible table can therefore be placed in a dump, have the editable schema-hash fields changed to the target hashes, pass size/SHA/stats validation, and then be committed by `submitTableDumpObjects`. Nothing compares present seqnums, `ColumnMeta.DataType()`, max seqnum, or tombstone shape with `targetRef.relation.GetTableDef(ctx)`, so later scans can interpret incompatible bytes or fail after the bad object is already catalogued. Please pass the target physical schema into validation and reject incompatible data/tombstone layouts before `rel.Write`; a manifest-substitution regression test would cover this boundary.

XuPeng-SH (changes requested):
Deep review of `6cad9264`.

The rollback-cleanup blocker from the prior round is fixed: Delete no longer holds the transaction mutex, semaphore admission observes the cleanup deadline, and workspace/TN teardown continues on cleanup failure. I also re-checked the earlier Seqnum concern; current layout-changing ALTER operations use COPY and rewrite the data, so I am not carrying that finding forward.

Requesting changes for the two remaining P1 input/data-integrity gaps called out inline. Local validation passed for all four modified packages, including focused race coverage; GitHub CI is still running on this head.

aptend:
Addressed the two P1 findings in `793c71298`.

- LOAD now compares every physical block metadata schema with the target relation during the existing metadata pass: stored seqnums/types/counts, the optional commit-TS suffix, and supported tombstone layouts including target PK type. This adds no extra object I/O. The BVT now includes a flushed delete/tombstone and passes locally (26/26).
- DUMP/LOAD now reject temporary, external/source, view, sequence, transient, and table-function/non-object-backed relations on both paths.

I also investigated the previous CI failures. The three relevant BVT jobs failed because self-handled LOAD reused a canceled prior-pipeline `proc.Ctx` for table locking. The lock path now accepts and uses the current statement request context; a focused regression test and the local BVT pass. The prior UT TAE replay timeout/`ExpectedEOB` and Coverage Kafka count mismatch were unrelated to this PR.

Validation: full frontend, lockop, fileservice, and disttae package tests; focused race tests; `go vet`; `make build`; local `table_dump_load.sql` BVT 26/26.

pkg/frontend/table_dump.go
```diff
@@ -0,0 +1,1541 @@
...
+			)
+		}
+		seqnum := uint16(col.Seqnum)
+		if _, ok := schema.dataTypes[seqnum]; ok {
+			return nil, moerr.NewInternalErrorNoCtxf(
+				"table %s has duplicate physical seqnum %d", def.Name, seqnum,
+			)
+		}
+		oid := uint8(types.T(col.Typ.Id))
+		schema.dataTypes[seqnum] = oid
```
XuPeng-SH: [P1] Bind storage-semantic type parameters, not only the OID

`tableDumpPhysicalSchema` reduces every target `plan.Type` to its one-byte OID, so this new guard still accepts objects whose values have a different logical encoding. The manifest schema hashes are editable and have no signature/object-bound authentication: a valid object from `ENUM('red','blue')` containing ordinal 2 can be substituted, both manifest hashes changed to a target `ENUM('green','red')`, and the SHA/stats/seqnum/OID checks all pass; after attachment that ordinal denotes `red`, silently changing the value. I reproduced this against this head with an object written as `T_enum` ordinal 2 and a target `plan.Type{Id: T_enum, Enumvalues: "green,red"}`; `loadPhysicalTableDumpStats` returns nil. Decimal width/scale has the same class of mismatch.

Please bind and validate the complete storage-semantic type signature (including enum/set mapping and decimal parameters), with an object-bound/authenticated source of truth for parameters that `ColumnMeta.DataType()` does not encode; otherwise reject parameterized types that cannot be proven compatible. Please add a same-OID/different-parameters manifest-substitution regression test.

XuPeng-SH (changes requested):
Deep re-review completed at `793c7129897e256cc888f9bc536da0070928a5e2`. The prior relation-kind, physical-layout, rollback-cleanup, and request-context issues are addressed, and the targeted frontend/fileservice/disttae/parser tests pass. One remaining data-integrity blocker is inline, so I’m requesting changes.

aptend:
@XuPeng-SH I disagree with treating the latest ENUM/DECIMAL finding as a blocker, and I do not think adding more type-specific checks is the right direction.

The proposed threat model assumes an attacker can rewrite the object, manifest, SHA256, ObjectStats, and schema hashes, but then expects an ENUM/DECIMAL parameter check to establish data integrity. It cannot. The same attacker can generate a valid object using the target schema and place arbitrary values in it. Checking enum labels or decimal scale blocks one contrived substitution while leaving the claimed capability—arbitrary data replacement—completely intact.

This is a physical restore path operated by an administrator against a trusted fixture. Its validation boundary is structural decodability, bounded resource use, transfer corruption, and accidental schema mismatch. It is not an authenticated import format and does not claim provenance when both manifest and objects are attacker-controlled. If untrusted fixtures are a product requirement, the correct design is a signed/authenticated manifest or a COPY/INSERT path that revalidates logical rows. That is a separate feature and cannot be approximated by an open-ended list of type special cases.

We have already accumulated too much speculative defensive code in this path by repeatedly accepting hypothetical constructions without first fixing the trust contract. Such checks provide little meaningful protection, make the implementation increasingly bloated, and impose substantial long-term maintenance cost. Rejecting ENUM/DECIMAL, or adding more parameter-specific branches, would be security theater rather than closure of an attack boundary.

For the same reason, we are reassessing earlier complexity added solely to defend against simultaneous malicious rewriting of all fixture metadata, including target physical-schema binding and full-file SHA rereads after provider-side copies. Checks with independent correctness, availability, or lifecycle value remain valid; unauthenticated provenance checks do not.

Unless the product contract is changed to require hostile-fixture support together with a real authentication mechanism, this finding should not remain a blocker.

gouhongshen (changes requested):
## Codex automated review

One blocking data-integrity gap remains.

### P1 - Validate storage-semantic type parameters before attaching objects (pkg/frontend/table_dump.go:1018)
`buildTableDumpPhysicalSchema` retains only the one-byte OID, and `ColumnMeta` likewise contains no ENUM/SET members or decimal width/scale. LOAD attaches the validated ObjectStats directly, so a valid ENUM object from a differently ordered member list (or DECIMAL with another scale) passes the physical check after its editable manifest hashes are changed to the target hashes, yet reads with changed values/scale. Admin-only execution does not make an externally supplied stage/local fixture structurally compatible. Persist and verify an object-bound complete physical schema signature, or reject parameterized types that cannot be proven compatible; add an ENUM/DECIMAL substitution regression test.

pkg/fileservice/aliyun_sdk.go
```diff
@@ -40,10 +41,29 @@ import (
...
+	src ObjectStorage,
+	srcKey string,
+	dstKey string,
+) (bool, error) {
+	s, ok := src.(*AliyunSDK)
+	if !ok || !strings.EqualFold(a.endpoint, s.endpoint) ||
+		!a.copyCredentialDomain.matches(s.copyCredentialDomain) {
+		return false, nil
+	}
+	_, err := a.bucket.CopyObjectFrom(s.bucket.BucketName, srcKey, dstKey)
```
XuPeng-SH: [P1] Propagate cancellation into Aliyun server-side copies

This implementation discards the caller's context and invokes `CopyObjectFrom` without `oss.WithContext(ctx)`. DUMP runs up to 32 copies in an `errgroup`; after the first copy error or a client cancellation, `group.Wait()` still has to wait for every in-flight Aliyun call. The S3FS caller also wraps this in `DoWithRetry` (up to 128 attempts), so a retryable OSS/network failure can keep the statement and object-storage semaphore slots occupied long after the request and transaction were canceled. `CopyObjectFrom` accepts options and the other Aliyun operations already pass `oss.WithContext(ctx)`. Please propagate this context and cover cancellation so the wait-for edge is bounded by the caller's lifetime.

pkg/frontend/table_dump.go
```diff
@@ -0,0 +1,1541 @@
...
+		return err
+	}
+	defer closeDumpFS()
+	sourceFS, err := GetObjectFSProvider(ses)
+	if err != nil {
+		return err
+	}
+
+	txn := ses.GetTxnHandler().GetTxn()
+	txn.GetWorkspace().SetCloneTxn(txn.Txn().SnapshotTS.PhysicalTime)
```
XuPeng-SH: [P1] Do not retain DUMP source names for the whole explicit transaction

`SetCloneTxn` makes `NewImmutableTableMetaReader` register every visible source object in `cloneTxnCache.sharedFiles`. `EndStatement` and statement rollback do not remove those registrations; `DeleteTxn` releases them only when the entire transaction commits or rolls back. A single DUMP is capped at 250k objects, but DUMP is explicitly allowed in an uncommitted transaction and there is no cross-statement cap, so repeated DUMPs over disjoint tables (or advancing RC snapshots) grow CN heap outside `txn.workspaceSize`/workspace quota until the transaction ends. DUMP is read-only, and clone GC only consults this set when deciding whether to delete names from transaction writes, so these read-side registrations should not outlive the copy. Please make the immutable DUMP reader avoid them or release its statement generation, with a multi-statement explicit-transaction regression test.

XuPeng-SH (changes requested):
Deep re-review completed at exact head `793c7129897e256cc888f9bc536da0070928a5e2`.

I accept the latest trusted-admin-fixture argument and resolved my prior ENUM/DECIMAL thread; that finding should not block this contract. The full-file SHA, ObjectStats, and physical-layout checks remain useful for transfer integrity, structural compatibility, and accidental mismatch.

Two independent lifecycle blockers remain inline: Aliyun provider-side copy drops cancellation, and DUMP retains source-object names across statements without a transaction-wide bound or workspace accounting. Requesting changes for those two issues.

Validation status: exact-head targeted frontend/fileservice/disttae/parser tests passed in the prior round and the head has not changed; current GitHub UT, pessimistic BVT/compose, proxy BVT, SCA, and Iceberg checks pass. The Coverage job failed in its Python gate; its uploaded patch artifact reports 81.71% (956/1170 statements).

aptend:
@gouhongshen We are explicitly not addressing the ENUM/DECIMAL substitution finding from review 4761420536. This PR treats the privileged DUMP/LOAD operator and the supplied fixture as trusted. A normal schema mismatch is already rejected by the manifest schema hashes; reproducing this case requires a privileged user to deliberately replace the physical object and rewrite the relevant manifest hashes. The same user can already load or insert arbitrary data, so authenticating an administrator-authored fixture is outside this feature contract. XuPeng-SH accepted and resolved the identical finding in the subsequent review.

To avoid false assurance and further maintenance cost, head 1d6285e96 removes the incomplete object-to-target OID/seqnum/layout comparison. File size, SHA-256, ObjectStats-versus-physical-metadata validation, bounds, and ordinary schema-hash checks remain intact. Please do not carry this threat-model finding forward.

pkg/fileservice/aliyun_sdk.go
aptend: Fixed in 1d6285e96. Aliyun CopyObject now passes oss.WithContext(ctx) into CopyObjectFrom. Added a blocking httptest regression that cancels the caller context and verifies the provider request and CopyObject return promptly; focused race coverage passes.

pkg/frontend/table_dump.go
aptend: Fixed in 1d6285e96. DUMP no longer calls SetCloneTxn, and the immutable metadata reader no longer registers read-side source names in cloneTxnCache. Added a same-transaction, multi-statement regression that exposes new object names across statements and verifies neither shared-file set grows.

XuPeng-SH (approved):
Re-reviewed the complete diff at 1d6285e968c6741bf586730e984b97870417b8c3. The two prior blockers are closed: Aliyun server-side copy now propagates cancellation, and the DUMP immutable reader no longer retains source object names across explicit-transaction statements. I also re-audited the DUMP/LOAD ownership, rollback, bounded-wait, schema/topology, copy/install, and transaction cleanup paths; focused package tests and race tests pass. No blocking issue found.

gouhongshen (changes requested):
## Codex automated review

Two blocking lifecycle failures remain in the new object-copy and rollback paths.

### P1 - Stop retrying provider copies after the request deadline (pkg/fileservice/s3_fs.go:202)
The new server-side copy path wraps `copier.CopyObject(ctx, ...)` in `DoWithRetry`, but that helper has no context check. `context.DeadlineExceeded` implements `net.Error` with `Timeout() == true`, so `IsRetryableError` accepts it and the loop sleeps/retries up to 128 times even though every subsequent provider call receives an already-expired context. `DUMP` and `LOAD` wait on these workers via errgroup, so a statement deadline can still keep the operation alive for roughly two minutes. Check `ctx.Err()` before retrying (and use context-aware backoff); the Aliyun `WithContext` fix alone does not close this wait-for path.

### P1 - Retain a cleanup path when rollback deletion fails (pkg/vm/engine/disttae/txn.go:2608)
A copied LOAD object is tracked before metadata validation/submission, but if `deleteLoadFiles` returns an error here, rollback continues to `delTransaction`, which clears `loadFiles` and deletes the clone-cache entry. No later code retains or enumerates these names. Thus a transient Delete failure after a partial external-fixture install leaves real, unreferenced target objects permanently orphaned; repeated failed LOADs accumulate storage. The new failure test only tracks a name and never creates its object, so it does not cover this outcome. Preserve failed names in a recoverable GC/retry path (or otherwise guarantee eventual deletion) before tearing down transaction state.

heni02 (approved):

aptend:
@gouhongshen Addressed both lifecycle findings in 4bf270714.\n\n1. Added a context-aware retry helper with cancellable backoff. S3FS provider copy and all provider HEAD/Exists retry paths now stop on caller cancellation instead of retrying an expired context.\n2. LOAD rollback now retries retryable Delete failures inside the existing bounded cleanup context, before transaction teardown and while the global install lock still prevents object-name reuse. The regression creates a real target object, injects two transient Delete failures, and verifies the third attempt removes the object before teardown. Cleanup remains bounded, and non-retryable failures still fail immediately.\n\nI deliberately kept recovery synchronous under the install lock: an asynchronous delete after lock release could race a later LOAD that adopts the same object name.

gouhongshen (approved):

gouhongshen (approved):
## Codex automated review

Reviewed the complete origin/main...HEAD change and all existing PR reviews, issue comments, inline threads, and replies. The earlier authorization, schema/topology, concurrency, checksum, cancellation, and rollback-cleanup findings are addressed or disproven under the documented trusted-admin fixture contract. No new blocking or non-blocking code defect was found at eb35456fdf05c64d9140a4cca72fdd15f47bd894.

PR #25990, opened 2026-07-22
scheduler: add pool-aware deterministic worker subsets

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [x] Code Refactoring

Related to #25451

This is the Phase 8 scheduler capability PR. It makes pool intent and deterministic worker-subset selection first-class across frontend, compile, engine resolution, scan/stage placement, tracing, and EXPLAIN.

Key points:

- Adds statement/session scheduling intent, strict versus legacy-compatible pool fallback, empty-worker policy, explicit worker locality, and resolved pool identity.
- Adds deterministic HRW `max-workers(N)` selection with stable membership and final execution ordering; the ingress CN may remain coordinator without being a query worker.
- Resolves tenant/label pools with truthful fallback metadata, candidate-owned `CPUTotal`, participant compatibility filtering, and strict no-cross-pool behavior.
- Removes the historical `QueryClient != nil -> CurrentCNRequired` proxy and prevents remote scope failures from silently moving execution back to the coordinator.
- Carries raw per-statement scheduling intent through multi-statement, cached-plan, prepared, retry, and EXPLAIN paths, with fail-closed parsing and prepared SQL-mode preservation.
- Closes review-found correctness and lifecycle gaps: invalid intent is rejected before control-plane discovery, route correlation handles missing/duplicate IDs, memoryengine no-match semantics are truthful, deterministic selection has a bounded fallback key, and prepared caches are bypassed without destroying reusable default topology.

Validation:

- Relevant package tests and race tests for `pkg/sql/schedule`, `pkg/vm/engine/disttae`, `pkg/vm/engine/memoryengine`, `pkg/sql/compile`, and `pkg/frontend`.
- `go build`, `go vet`, `git diff --check`, and scheduling benchmarks.
- `make dev-build` and `make dev-up` with a fresh five-service cluster.
- Real 5,000,000-row AP execution with strict `max-workers(1)`: CN1 remained ingress/coordinator, CN2 was the only query/scan/stage worker, and `sum(a)` returned `12500002500000`.


XuPeng-SH:
CI follow-up for 586dd5c22c:

- The Multi-CN proxy BVT, standalone pessimistic BVT, and Coverage failures shared one root cause: `SplitSqlByStatementWithSQLMode` split the semicolon inside `/*! ... */` away from its closing `*/`, so the new per-statement scheduling validation saw two raw fragments for one AST statement.
- The parser now extends only an executable comment’s final semicolon boundary through its closing terminator. Ordinary SQL keeps the existing hot path; executable comments use one SQL-mode-aware scanner pass.
- Local validation: full parser tests; full frontend tests through the deterministic CGo wrapper; build; vet; `make dev-build`; `make dev-up`; and the exact `ddl/mysql_ddl_1.sql` mo-tester case (46 total, 43 success, 3 ignored, 0 failed).

Fresh CI run: https://github.com/matrixorigin/matrixone/actions/runs/29888175084

XuPeng-SH:
Second first-principles and unhappy-path review completed at 59e29762bd.

Confirmed and fixed one strict-isolation gap: memoryengine historically widened sys/internal labeled requests to all CNs even under explicit strict fallback policy. Legacy privileged behavior remains unchanged; strict privileged requests now resolve exact labels only. Regression coverage includes exact/other/shared candidates and the no-exact with shared-present path.

Fresh validation: changed-package tests passed; race passed for schedule, memoryengine, compile, and frontend; vet and build passed with repository CGo settings; make dev-build and make dev-up passed; mysql_ddl_1.sql passed with CI runner flags (46 total, 43 success, 3 ignored, 0 failed); scheduling session, statement hint, prepared, EXPLAIN, invalid-hint fail-closed smoke tests passed; service logs show no panic/fatal/deadlock/data-race signatures.

Fresh CI has started for the new head.

aunjgr (approved):
Reviewed independently from candidate discovery through pool resolution, eligibility filtering, deterministic HRW subset selection, scope materialization, prepared/retry intent propagation, and remote-start failure cleanup. Strict pools do not widen through privileged compatibility fallbacks; draining/drained and unroutable workers remain observable but ineligible; required-current placement is pinned through max-worker selection; stable identity tie-breaking makes retries deterministic; and explicit prepared-statement intent cannot reuse a topology compiled under defaults. The executable-comment boundary fix also preserves raw per-statement hints without changing ordinary statement splitting. No actionable issues found.

XuPeng-SH:
Blocking CI performance regression on the current head:

- 59e29762 passed the same 2-CN LAUNCH+PROXY BVT: 98,192 commands, summed case cost 2108.540s.
- After merging main, 20e29205 and 56e016eb both timed out at 60 minutes; completed-case costs were 3204.673s and 3225.908s.
- Of 1,065 aligned cases between 56e and passing 59e, 1,036 became slower. This is systemic, not one hung case.
- Minimal case: dtype/year.sql, 477 commands, 1.849s -> 17.783s (+15.934s, about +33.4ms/command). window/window.sql: 13.135s -> 50.031s.
- The same 56e SHA passes single-CN, and #26004 alone passes Proxy BVT. This points to a post-merge interaction; MORPC/scheduler is a candidate, not yet a proven root cause.
- Exact 2-CN LAUNCH+PROXY local A/B on macOS did not reproduce the Linux CI magnitude.

Suggested next step: targeted Linux CI running only dtype/year.sql with per-statement timing or a CPU profile, then rerun full Proxy BVT after fixing the fixed per-command overhead. GitHub does not allow the PR author to submit Request changes on their own PR, so recording this as a blocking comment.

XuPeng-SH:
@aptend @gouhongshen @LeftHandCold please help to review

XuPeng-SH:
CI performance concern cleared after inspecting the latest rerun of the exact same head (56e016eb) in the exact same 2-CN LAUNCH+PROXY mode. The rerun completed successfully: summed case cost 2133.446s versus the prior healthy 2108.540s baseline (+1.2%); dtype/year.sql recovered from 17.783s to 1.900s and window/window.sql from 50.031s to 12.886s. Because no code changed and only the runner changed (westus3 worker to eastus2 worker), the earlier broad slowdown/timeout was runner-specific transient degradation, not a reproducible PR regression. No remaining blocker from this CI timeout. GitHub does not permit the PR author account to approve its own PR.

aptend (changes requested):
## Review result: Request changes

### [P1] Strict pool mode widens to every CN when the session has no label selector

A direct connection that does not pass through proxy leaves the session CN-label selector empty. Even with `query_pool_strict=1`, both pool resolvers currently interpret an empty selector as `all-compatible`:

- `pkg/vm/engine/disttae/engine.go:903` returns all candidate nodes.
- `pkg/vm/engine/memoryengine/engine.go:288` sets `allCompatible=true`.
- The result has `Fallback=false`, so the strict check in `pkg/sql/schedule/placement.go:252` accepts it.

Concrete failure: with tenant-a connected directly and two working candidates labeled for tenant-a and tenant-b respectively, a strict tenant-a request returns both candidates. I reproduced this on the PR head with a temporary regression assertion; it received `[tenant-a, tenant-b]` instead of an empty/fail-closed result.

This violates the explicit strict/no-cross-pool contract and breaks compute-pool isolation.

Suggested fix: when strict mode has no resolvable label selector, either fail closed or derive an exact selector from the authenticated tenant. Preserve the current all-compatible behavior only for legacy-compatible mode. Please add coverage for disttae, memoryengine, and the direct frontend-to-compile path.

I also checked candidate discovery → pool resolution → eligibility/HRW selection → scope materialization, prepared/retry intent propagation, and RemoteRun failure cleanup. No additional confirmed correctness or lifecycle blocker was found.

pkg/vm/engine/disttae/engine.go
```diff
@@ -871,59 +869,138 @@ func (e *Engine) DiscoverQueryCandidates(ctx context.Context) (engine.QueryCandi
...
 		ctx = context.Background()
 	}
 	if err := ctx.Err(); err != nil {
-		return nil, err
+		return engine.ResolvedQueryPool{}, err
+	}
+	if !request.FallbackPolicy.Valid() {
+		return engine.ResolvedQueryPool{}, moerr.NewInvalidInput(ctx, "invalid query pool fallback policy")
 	}
 	if len(request.CNLabel) == 0 {
```
LeftHandCold, line 903: [P1] This branch also runs when `FallbackPolicy` is strict. A normal direct connection (or proxy-disabled deployment) leaves `Session.requestLabel` empty, so the request reaches here with `CNLabel == nil`; this then returns all labeled tenant pools as `all-compatible` with `Fallback=false`, and `DecideQueryPlacement` cannot reject it. With tenant-a and tenant-b CNs, a strict tenant-a AP query can select both. Please fail closed here for strict mode or derive an exact authenticated-tenant selector, and make the same change in memoryengine.

LeftHandCold (changes requested):
Request changes: strict pool isolation is bypassed on a normal direct-CN session. Such a session has no proxy-provided CN labels, so both resolver implementations classify the request as all-compatible and return every CN with Fallback=false. The scheduler therefore accepts the result even when query_pool_strict=1, allowing a tenant-a AP query to be scheduled onto tenant-b's labeled CN. This is a concrete compute-pool isolation violation, not just a defensive concern. Please fail closed for strict requests with no resolvable selector, or derive an exact selector from the authenticated tenant; keep all-compatible only for legacy-compatible mode. Add coverage for disttae, memoryengine, and the direct frontend -> compile path.

I also reviewed candidate discovery/resolution, HRW subset selection, route materialization, prepared/retry propagation, executable-comment splitting, trace bounds, and RemoteRun failure cleanup. I found no additional confirmed blocker. Focused tests and go vet for the affected packages pass.

XuPeng-SH:
@aptend Thanks, this is a valid strict-isolation gap. Addressed in 2d759fa54.

- disttae and memoryengine now fail closed with no-match when strict policy has no CN label selector.
- Legacy-compatible requests with an empty selector retain the historical all-compatible behavior.
- Added engine regressions with tenant-a/tenant-b candidates and a direct Session -> frontend -> compile -> disttae preview regression proving that an unlabeled strict direct connection selects no cross-pool worker.

Fresh local validation after the change:
- full tests: pkg/vm/engine/disttae, pkg/vm/engine/memoryengine, pkg/frontend
- focused regressions under -race
- go vet and go build for all three packages

All completed with exit code 0. Please re-review the new head.

pkg/vm/engine/disttae/engine.go
XuPeng-SH: Fixed in 2d759fa54. Strict requests with an empty CN label selector now return no-match in both disttae and memoryengine, while legacy-compatible requests retain all-compatible behavior. Added disttae, memoryengine, and direct Session -> frontend -> compile -> disttae regressions; full affected-package tests, focused race tests, vet, and build all pass.

LeftHandCold (approved):
Re-reviewed the updated head. The previous strict-isolation blocker is closed: disttae and memoryengine now fail closed with an empty label selector under strict policy, while legacy-compatible requests retain all-compatible behavior. The direct Session -> frontend -> compile path reaches no-match and unsatisfied placement, so no cross-pool remote worker can be launched. I also rechecked resolver coverage/wrappers, placement termination, default compatibility, retry/reuse, and lifecycle/unhappy paths; no additional actionable blocker found. Focused regressions under -race, full disttae/memoryengine/frontend/schedule/compile tests, and affected-package build/vet pass locally.

aptend (approved):
Re-reviewed the updated head at 2d759fa. The previous strict-pool isolation blocker is closed: both disttae and memoryengine fail closed when strict policy has no CN label selector, while legacy-compatible behavior remains unchanged. The no-match result propagates through placement to an unsatisfied compile result before scope materialization or remote launch. Focused regressions passed under race for three iterations; full frontend, disttae, and memoryengine tests and vet also passed locally. No remaining blocking findings.

PR #25998, opened 2026-07-22
test(plan): benchmark low-cost row carrier pruning

### LINKED ISSUE #25700: [SubTask]: Plan-shape, correctness, and performance/memory tests for special-operator pruning
## Brief Description

Part of #25612. Validate the special-operator pruning work with plan-shape, correctness, and performance/memory coverage.

- Plan-shape tests comparing plans before and after pruning.
- Correctness tests: empty-input and 0 / 1 / many-row results.
- Wide-varlen performance harness and allocation / peak-memory measurements.
- BVT under `test/distributed`, with results generated and verified via mo-tester per repo conventions.

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

fixes #25700

- Adds a paired plan-shape regression test showing that consumed `SAMPLE` outputs retain both inputs while discarded outputs keep only the narrow row carrier.
- Adds a fixed-count execution memory test that runs the unchanged production sample-pool initialization path with wide `text` data and with the pruned `int8` carrier, verifies identical row counts, and checks that both retained vector memory and peak heap growth drop by more than 100x. Peak heap growth is measured once per case with controlled `runtime.MemStats` sampling outside any benchmark loop.
- Adds an execution benchmark whose loop runs only the normal `SAMPLE` production path, whose allocation metrics come from `benchmem`, and whose retained-memory metric comes from the actual result vectors.

Validation:

- `go build ./pkg/sql/plan/... ./pkg/sql/plan/explain`
- `go vet ./pkg/sql/plan/... ./pkg/sql/plan/explain`
- `mo-cgo-test -count=1 -timeout=240s ./pkg/sql/plan ./pkg/sql/plan/explain ./pkg/sql/colexec/sample`
- `mo-cgo-test -run '^$' -bench '^BenchmarkSamplePrunedWideVarlenMemory$' -benchmem -count=1 -timeout=120s ./pkg/sql/colexec/sample`

### DIFF
--- a/pkg/sql/colexec/sample/sample_test.go
+++ b/pkg/sql/colexec/sample/sample_test.go
@@ -16,6 +16,8 @@ package sample
 
 import (
 	"bytes"
+	"runtime"
+	"runtime/debug"
 	"testing"
 
 	"github.com/matrixorigin/matrixone/pkg/common/mpool"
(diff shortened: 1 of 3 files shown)


aunjgr (approved):
Reviewed independently. This is test/benchmark-only coverage. The plan-shape pair checks consumed versus discarded SAMPLE outputs, and the memory case exercises the production sample pool with controlled GC and mpool cleanup. No actionable issues found.

XuPeng-SH (approved):
Deep-reviewed the complete test-only diff and the production paths it exercises. The plan-shape pair proves the wide SAMPLE output is retained when consumed and replaced by the narrow carrier when discarded. The memory test drives the production sample pool, checks ownership/mpool cleanup, and stayed green for 100 consecutive runs. Five benchmark runs consistently measured 4,843,072 vs 64 vector-retained B/op and about 23.07 MB vs 520 B/op. Related package tests, focused race tests, vet, and CI are green. No actionable issues found.

PR #25901, opened 2026-07-20
fix(lockservice): honor context cancellation in lock waits

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

Fixes #25790
Fixes #25795

## What this PR does / why we need it:

- propagates caller context through lockservice readiness waits, bind allocation, allocator/RPC calls, remote retries, waiting-list traversal, and deadlock detection
- rejects cancellation before local lock or bind-publication linearization points while preserving must-complete transaction cleanup semantics
- makes deadlock detector shutdown cancel in-flight remote GetTxnLock retries
- serializes lock admission with rolling-restart drain transitions
- fixes proxy shared-lock cancellation and waiter notify/cancel races so callbacks and pooled waiter cleanup remain exactly once
- keeps stale-allocator bind fencing intact after bind refreshes

Validation:

- focused cancellation tests repeated with `-count=100`
- focused concurrency tests repeated under `-race -count=20`
- `pkg/lockservice` full test suite
- `pkg/lockservice` full test suite with `-race`
- `go build ./pkg/lockservice ./pkg/cnservice`
- `go vet ./pkg/lockservice ./pkg/cnservice`
- `TestDefaultCnConfig`
- independent read-only review loop converged to PASS

PR #25999, opened 2026-07-22
fix(lockservice): bound remote wait-for snapshot retries

### LINKED ISSUE #25870: [Bug]: Remote wait-for snapshot retries ignore context and can stall deadlock detection

### Branch Name

main

### Commit ID

4d78bc66670418fbdeac2b18b07b2ef5b08b1d72

```Markdown
- Hardware parameters: N/A
- OS type: all
- Others: multi-CN lockservice path
```

### Actual Behavior

`LockService.GetWaitingList(ctx, txnID)` does not honor caller cancellation when a local transaction holds locks through a remote lock table.

The synchronous path is:

```text
GetWaitingList(ctx)
  -> activeTxn.fetchWhoWaitingMe(...)
  -> remoteLockTable.getLock(...)
  -> doGetLock()                         // one RPC attempt has defaultRPCTimeout
  -> handleError(context.Background())
  -> waitRemoteRetryBackoff()
  -> retry forever
```

`remoteLockTable.getLock` has no context parameter and retries with background contexts. If the remote lock owner is unavailable while the allocator still returns the unchanged binding (or allocator refresh keeps failing), canceling or timing out the `GetWaitingList` context cannot terminate the call.

The production deadlock detector uses the same synchronous path through `service.fetchTxnWaitingList`. A persistent remote failure can therefore occupy a detector worker indefinitely; four affected checks can exhaust the current four-worker detector pool and prevent later deadlock checks from making progress.

This is pre-existing and was discovered while reviewing test-liveness changes in #25860; it is not introduced by that PR.

### Expected Behavior

- `GetWaitingList` must return when its caller context is canceled or expires.
- Remote wait-for graph lookup must have a bounded failure path.
- A failed remote snapshot must release its deadlock-detector worker instead of retrying forever.
- Temporary RPC failures should still retry while the operation's context remains live.

### Steps to Reproduce

```Markdown
1. Construct a remoteLockTable with a Client that returns a retryable error for Method_GetTxnLock.
2. Make Method_GetBind return the same lock-table binding (or continue returning an error).
3. Create a local active transaction whose lockHolders reference that remote lock table.
4. Invoke LockService.GetWaitingList with a short deadline.
5. Observe that the deadline expires but the call remains in remoteLockTable.getLock retrying GetTxnLock/GetBind.
6. Repeat through deadlock detection for four transactions and observe all detector workers remain occupied.
```

The existing `TestGetLockRemoteWithRetry` covers eventual recovery after one error, but there is no cancellation/timeout test for this path.

### Additional information

Relevant code:

- `pkg/lockservice/service_observability.go`: `GetWaitingList` accepts a context but does not pass it into `fetchWhoWaitingMe`.
- `pkg/lockservice/service.go`: `fetchTxnWaitingList` is the deadlock-detector callback and calls the same synchronous traversal.
- `pkg/lockservice/txn.go`: `fetchWhoWaitingMe` calls `lockTable.getLock` without a context or error return.
- `pkg/lockservice/lock_table_remote.go`: `getLock` uses `handleError` and `waitRemoteRetryBackoff`, both background/unbounded variants.

Suggested direction:

1. Thread a context and error result through the wait-for snapshot traversal and `lockTable.getLock`.
2. Use `handleErrorWithContext` and `waitRemoteRetryBackoffWithContext` for remote retries.
3. Give deadlock-detector snapshot work an explicit per-check deadline, since it has no external request context.
4. Add deterministic tests for cancellation during GetTxnLock, bind refresh, and retry backoff, plus an eventual-recovery case.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25870

#25901 made remote wait-for snapshot retries context-aware, so caller cancellation and detector shutdown now propagate correctly. However, deadlock-detector checks run with a service-lifetime context. While the service remains live, a permanently unavailable remote lock-table owner can still keep `remoteLockTable.getLock` retrying forever and occupy one of the four detector workers indefinitely.

This PR adds a 35-second deadline to one remote lock-key snapshot lookup. The bound covers a failed `GetTxnLock` RPC, bind refresh, retry backoff, and another RPC attempt. It deliberately does not bound an entire transaction snapshot or wait-for graph traversal: healthy checks may need to scan many lock keys and transactions, and a cumulative deadline would repeatedly restart the traversal and could permanently miss a deeper deadlock edge.

On timeout, the existing error path unwinds the deadlock check, removes the transaction from `activeCheckTxn`, and returns the worker to the detector queue. Caller deadlines and detector shutdown cancellation still take precedence over the internal bound.

The regression test fills all four detector workers with blocked remote snapshot lookups, verifies that they return with `context.DeadlineExceeded`, confirms `activeCheckTxn` cleanup, and proves that a fifth queued check is processed without invoking the deadlock-abort callback.

Validation:

- focused lockservice tests with `-count=50`
- focused race tests with `-count=10`
- full `pkg/lockservice` test suite
- full `pkg/lockservice` race suite
- `go vet ./pkg/lockservice`
- dependent `pkg/util/status` tests


gouhongshen (approved):
## Codex automated review

No findings. The timeout cleanly bounds the previously infinite remote snapshot retry chain, preserves earlier cancellation, releases detector workers and active-check state, and allows periodic rechecks. Focused tests passed 20 runs and 10 race-enabled runs; CI is green.

XuPeng-SH (approved):
Reviewed exact head 097dba698. No blocking findings. The per-lock-key deadline preserves earlier caller/shutdown cancellation, bounds the GetTxnLock + bind-refresh + backoff retry chain, releases detector workers and activeCheckTxn state on timeout, and leaves periodic rechecks intact without imposing a cumulative traversal deadline. Independently passed focused tests (20 runs), focused race tests (10 runs), full pkg/lockservice tests with and without -race, pkg/util/status tests, go vet, and go build.

PR #26004, opened 2026-07-22
fix(morpc): terminate pending work during backend and stream close

### LINKED ISSUE #25943: [Bug]: morpc Stream.Close(false) deadlocks behind a full receive channel
## Bug description

`morpc.stream.done` holds `s.mu.RLock()` while sending a response to the bounded receive channel. When that channel is full and the response context has no deadline, the send blocks indefinitely while retaining the read lock.

At the same time, `Stream.Close(false)` needs `s.mu.Lock()` before it can drain the channel and mark the stream closed. It therefore cannot make progress, producing a lock/channel deadlock:

1. The receive channel is full.
2. Another response enters `stream.done`, obtains `RLock`, and blocks on `s.c <- response`.
3. The caller invokes `Close(false)`.
4. `Close(false)` waits for the write lock, but only `Close` can drain the channel and unblock `done`.

`Close(true)` can avoid this in some cases by stopping the backend first, but the normal `Close(false)` API has no cancellation path for the blocked delivery.

## Impact

A caller that stops consuming a stream and closes it while responses are still arriving can hang permanently. This can leak the stream, block cleanup, and leave callers waiting forever during cancellation or shutdown.

## Reproduction

A white-box test with a receive channel of capacity one:

1. Deliver sequence 1 to fill the channel.
2. Deliver sequence 2 with `context.Background()` in another goroutine; it blocks while holding `RLock`.
3. Call `Close(false)` and wait for it to return.

The audit was run three times. All 3/3 runs showed `Close(false)` still blocked after 200 ms; it returned only after the test manually drained the first response.

```bash
go test ./pkg/common/morpc \
  -run '^TestAuditStreamCloseDoesNotBlockBehindFullReceiveChannel$' \
  -count=3 -v
```

## Expected behavior

`Stream.Close(false)` must promptly cancel or otherwise unblock pending response delivery, even when the receive channel is full and unread.

## Tested version

- Audit baseline: `main@7f9ef5de9bb9337d8db0a9b15c44cba3b2838aa2`
- GitHub `main` was checked again before filing and still holds `RLock` across the blocking channel send.

## Suggested direction

Do not hold the stream read lock across a potentially blocking channel send. Introduce a stream-local cancellation signal that both `done` and `Close` select on, or otherwise arrange for close to unblock pending delivery before acquiring the exclusive lock.

### LINKED ISSUE #25945: [Bug]: closing a morpc backend leaves response-waiting Futures blocked
## Bug description

Closing a `morpc.remoteBackend` does not complete futures whose requests were already written and are waiting for responses.

When the backend context is canceled, `readLoop` can exit through the `ctx.Done()` branch:

```go
case <-ctx.Done():
    normalExit = true
    rb.clean()
    return
```

`rb.clean()` only deletes entries from `rb.mu.futures`; it does not call `Future.error`. A Future that has already received its write completion therefore remains blocked in `Get()` until its caller-owned request context expires, even though the backend is definitively closed.

## Impact

- In-flight RPC calls do not fail promptly when a client/backend is closed.
- Shutdown and cancellation can remain blocked until long business deadlines expire.
- The backend loses the future from its tracking map without delivering a terminal result.

## Reproduction

Using a real local morpc connection:

1. Let the server accept a request without replying.
2. Wait until the Future reports that the request was successfully written.
3. Call `remoteBackend.Close()`.
4. Call `Future.Get()` with a 30-second request context.

The audit ran this sequence three times. In all 3/3 runs, `Get()` was still blocked after the backend and both read/write loops had stopped; it returned only after the test explicitly canceled the request context.

```bash
go test ./pkg/common/morpc \
  -run '^TestAuditCloseBackendCompletesFutureAlreadyWaitingForResponse$' \
  -count=3 -v
```

## Expected behavior

`Backend.Close()` must complete every tracked Future with a closed/canceled backend error, including requests already written and waiting for a response.

## Tested version

- Audit baseline: `main@7f9ef5de9bb9337d8db0a9b15c44cba3b2838aa2`
- GitHub `main` was checked before filing and still uses `rb.clean()` without notifying waiting futures.

## Suggested direction

Before removing tracked futures during backend shutdown, notify them with a terminal backend-closed error. Keep future-map cleanup and notification ordering safe with respect to `Future.Close` and pool reuse.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25943
Fixes #25945

- cancels a stream before waiting for its lock, so a full receive channel cannot block Stream.Close(false)
- gives each stream an independent handle, receive channel, and cancellation context instead of reusing externally visible stream objects
- stops backend I/O and closes the transport before failing all response-waiting Futures with backendClosed
- releases the backend map mutex before taking backend state locks during stream removal
- adds deterministic close, lock-order, generation-isolation, and race regressions

## Testing

- make thirdparties
- mo-cgo-test -race -count=20 focused MORPC close regressions
- mo-cgo-test ./pkg/common/morpc
- mo-cgo-test -race ./pkg/common/morpc
- go build ./pkg/common/morpc
- go vet ./pkg/common/morpc
- mo-cgo-test ./pkg/cnservice/cnclient

### DIFF
--- a/pkg/common/morpc/backend.go
+++ b/pkg/common/morpc/backend.go
@@ -198,7 +198,6 @@ type remoteBackend struct {
 	}
 
 	pool struct {
-		streams *sync.Pool
 		futures *sync.Pool
 	}
 }
@@ -229,17 +228,6 @@ func NewRemoteBackend(
 	rb.pool.futures = &sync.Pool{
 		New: func() interface{} {
 			return newFuture(rb.releaseFuture)
-		},
-	}
-	rb.pool.streams = &sync.Pool{
-		New: func() any {
-			return newStream(
-				rb,
-				make(chan Message, rb.options.streamBufferSize),
-				rb.newFuture,
-				rb.doSend,
-				rb.removeActiveStream,
-				rb.active)
 		},
 	}
 	rb.waitWriteC = make(chan struct{}, 1)
@@ -425,6 +413,8 @@ func (rb *remoteBackend) Close() {
 
 	rb.stopper.Stop()
 	rb.doClose()
+	rb.makeAllWaitingFutureFailed(backendClosed)
+	rb.clean()
 	rb.inactive()
 }
 
@@ -656,7 +646,6 @@ func (rb *remoteBackend) readLoop(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			normalExit = true
-			rb.clean()
 			return
 		default:
 			msg, err := rb.conn.Read(goetty.ReadOptions{Timeout: rb.options.readTimeout})
@@ -846,7 +835,13 @@ func (rb *remoteBackend) clean() {
 }
 
 func (rb *remoteBackend) acquireStream() *stream {
-	return rb.pool.streams.Get().(*stream)
+	return newStream(
+		rb,
+		make(chan Message, rb.options.streamBufferSize),
+		rb.newFuture,
+		rb.doSend,
+		rb.removeActiveStream,
+		rb.active)
 }
 
 func (rb *remoteBackend) cancelActiveStreams() {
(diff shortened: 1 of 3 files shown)


XuPeng-SH (approved):
Deep-reviewed exact head `78ab12de83c34e483a4624d277b50534d16ec95d`; no blocking correctness or liveness issue found.

I traced the complete backend/stream/future ownership, lock ordering, wait-for graph, and reuse generation boundary:

- `Stream.Close(false)` cancels the stream-local delivery context before acquiring the exclusive stream lock, so a response blocked on a full receive channel releases its read lock and close can drain/terminate it;
- stream unregister now happens after releasing `s.mu`, and backend removal releases `rb.mu` before taking `stateMu` through `Unlock`, removing both relevant ABBA edges;
- stream objects/channels/contexts are no longer pooled or externally reused, so an old handle cannot close or consume a later stream generation;
- backend close stops the write loop, disconnects transport, waits for the read loop, then fails response-waiting futures with `backendClosed` before deleting tracking entries; response and close cannot both terminally notify a Future;
- queued writes, stream receive channels, and tracking maps have fixed capacity or explicit removal paths; no new unbounded retention was introduced.

Q1: one effective cleanup owner remains on success, error, repeated close, and finalizer paths. Q2: every changed channel/lock wait has a cancellation or transport-close termination edge. Q3: removal of the stream pool eliminates the backend↔pooled-stream retention cycle and all changed accumulations remain bounded.

Validation:

- focused close/lock-order/generation tests under `-race -count=50`
- full `go test -race -count=1 ./pkg/common/morpc`
- `go test ./pkg/cnservice/cnclient`
- clean merge with current `origin/main`, followed by fresh MORPC race and cnclient tests
- required CI is green, including UT, SCA, CodeQL, multi-CN/proxy BVT, coverage, and Summary

PR #26005, opened 2026-07-22
fix(json): preserve JSON null in JSON_EXTRACT

### LINKED ISSUE #25960: [Bug]: JSON_EXTRACT converts JSON null to SQL NULL
### MatrixOne version

Latest `main` at `3046f14d94b3f1a25839eeeb65729b64f85ac647`.

### What happened?

`JSON_EXTRACT` converts a JSON `null` value found at an existing path into SQL `NULL`. MySQL 8.0 preserves it as a JSON null value.

This is not only a display difference: it changes the results of `IS NULL` and `COALESCE`.

### How to reproduce

```sql
CREATE TABLE json_null_test(id INT PRIMARY KEY, doc JSON);
INSERT INTO json_null_test VALUES (1, '{"a":null,"fallback":[]}');

WITH x AS (
    SELECT doc FROM json_null_test WHERE id = 1
)
SELECT JSON_EXTRACT(doc, '$.a'),
       JSON_EXTRACT(doc, '$.a') IS NULL,
       COALESCE(JSON_EXTRACT(doc, '$.a'), JSON_EXTRACT(doc, '$.fallback'))
FROM x;
```

### Actual result

```text
NULL  1  []
```

### Expected result

MySQL 8.0.45 returns:

```text
null  0  null
```

The path exists and its value is the JSON literal `null`, so it should remain distinguishable from a missing path or SQL `NULL`.

### Additional notes

The same behavior occurs without the CTE. The CTE in the reproduction confirms that the incorrect JSON-null semantics are preserved through CTE projection.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25960

`JSON_EXTRACT` previously used the JSON value's `IsNull` state to decide
whether to set the SQL null bitmap. The byte JSON query helpers represented
both a missing path and a matched JSON literal `null` with the same value, so
an existing JSON null was incorrectly converted to SQL `NULL`.

This change:

- returns explicit path-existence information from byte JSON queries;
- stops missing object keys and array indexes from producing JSON null
  placeholders;
- preserves matched JSON null values for single, multiple, and wildcard paths;
- sets the SQL null bitmap only when no path matched or an input is SQL `NULL`;
- adds unit and BVT coverage for direct queries, CTEs, `IS NULL`, `COALESCE`,
  missing paths, multiple paths, and wildcard paths.

Validation:

- `go test -count=1 -timeout=120s ./pkg/container/bytejson`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=600s ./pkg/sql/plan/function`
- modified-function coverage: 80%-100%; `jsonExtract`: 81.4%
- `go vet` and `go build` for the modified packages
- scoped `golangci-lint`: 0 issues
- `make build`
- `func_json_extract.test`: 211/211 passed


XuPeng-SH (approved):
Reviewed the JSON null/path-existence semantics and affected callers. The implementation now distinguishes a missing path from an existing JSON null, preserves array wrapping for wildcard/range/multi-path results, and handles scalar/object index/range autowrap consistently.

Verified locally: full bytejson and SQL function packages, repeated focused tests, race tests, go vet, and the external QuerySimple metadata consumer. CI is green. LGTM.

gouhongshen (approved):
## Codex automated review

No defects found. Existing PR discussion had no reviews or inline comments. Verified compatibility against MySQL 8, focused bytejson tests and vet passed; function-package tests are blocked locally by missing cgo/libmo.dylib.

heni02 (approved):

PR #26007, opened 2026-07-22
fix(frontend): handle non-string PREPARE variables

### LINKED ISSUE #25976: [Bug]: PREPARE FROM a non-string user variable panics in handlePrepareVar
### Summary

`PREPARE ... FROM @user_variable` panics when the variable contains a non-string value such as an integer or `NULL`.

### Tested version

### Reproduction

```sql
SET @p = 1;
PREPARE s FROM @p;
```

The same failure occurs with:

```sql
SET @p = NULL;
PREPARE s FROM @p;
```

### Actual result

```text
ERROR 20101 (HY000): internal error: panic interface conversion: interface {} is int64, not string
...
github.com/matrixorigin/matrixone/pkg/frontend.handlePrepareVar
    pkg/frontend/mysql_cmd_executor.go:1584
```

The integer case reproduces 3/3 times. The server remains available after the statement error.

### Expected result

Return a normal SQL error without an internal panic. MySQL 8.0.45 returns a syntax error for the integer and `NULL` cases.

String variables work normally in both systems:

```sql
SET @p = 'SELECT 1';
PREPARE s FROM @p;
EXECUTE s;
DEALLOCATE PREPARE s;
```

### Root cause

`handlePrepareVar` retrieves the user variable as `interface{}` and performs an unchecked type assertion:

```go
wrapper.Sql = p.Value.(string)
```

Non-string and nil values therefore panic instead of taking an error path.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25976

`PREPARE ... FROM @user_variable` previously asserted that every user-variable
value was a Go string. Integer and NULL values therefore panicked before the SQL
parser could return a normal statement error.

This change converts user-variable values to statement text before parsing,
matching MySQL's behavior for numeric and NULL values. Invalid text now follows
the regular syntax-error path, while string-backed prepared statements keep the
existing behavior.

Validation:

- `go test ./pkg/frontend -run 'Prepare(Var|String)' -count=1`
- `git diff --check`

Regression coverage includes NULL, boolean, array, floating-point, signed
integer boundary values, and the maximum unsigned integer. It also verifies
that failures do not populate the prepared-statement cache and that the session
can successfully prepare a valid statement afterward.

### DIFF
--- a/pkg/frontend/mysql_cmd_executor.go
+++ b/pkg/frontend/mysql_cmd_executor.go
@@ -1626,7 +1626,13 @@ func handlePrepareVar(ses *Session, execCtx *ExecCtx, st *tree.PrepareVar) (*Pre
 	if err != nil {
 		return nil, err
 	}
-	wrapper.Sql = p.Value.(string)
+	// MySQL converts numeric and NULL user variables to statement text so that
+	// the SQL parser reports the invalid statement consistently.
+	if p.Value == nil {
+		wrapper.Sql = "NULL"
+	} else {
+		wrapper.Sql = fmt.Sprint(p.Value)
+	}
 
 	return doPrepareString(ses, execCtx, wrapper)
 }
--- a/pkg/frontend/mysql_cmd_executor_test.go
+++ b/pkg/frontend/mysql_cmd_executor_test.go
@@ -1773,6 +1773,63 @@ func Test_HandlePrepareVarUsesSessionSQLMode(t *testing.T) {
 	})
 }
 
+func TestHandlePrepareVarWithNonStringValue(t *testing.T) {
+	ctx := defines.AttachAccountId(context.TODO(), catalog.System_Account)
+	setSessionAlloc("", NewLeakCheckAllocator())
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	ec := newTestExecCtx(ctx, ctrl)
+
+	runTestHandle("handlePrepareVarWithNonStringValue", t, func(ses *Session) error {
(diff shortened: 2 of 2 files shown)


gouhongshen (approved):
## Codex automated review

The change safely replaces the unchecked string assertion with conversion for the concrete user-variable value types, preserves string behavior, and routes invalid values through normal parser errors without caching a statement. Coverage exercises relevant scalar boundaries, NULL, unsupported composite values, cache state, and session recovery. Formatting checks passed; focused local tests were blocked by the absent prebuilt libmo.dylib, while the PR-head CI unit suite passed.

XuPeng-SH (approved):
Reviewed exact head `580dd416a1b3269a3116827becfd5dede6a70419`. No merge-blocking issue found. The unchecked string assertion is removed; SQL string-family, decimal, JSON, UUID, and temporal user-variable values already arrive as strings, while numeric/bool/enum/array values are converted to parser input and NULL is explicitly represented as `NULL`, producing a normal syntax error instead of an internal panic. The failure path does not register a partial prepared statement, and a valid string PREPARE still succeeds afterward.

Independent validation: focused non-string/session-SQL-mode tests 50x; focused `-race` 20x; full `pkg/frontend` tests and full-package `-race`; frontend build and vet; `git diff --check`; and a conflict-free merge tree against current `origin/main`. Current-head CI is fully green.

Issue #26006, opened 2026-07-22
[Bug]: disttae can admit a stale PartitionState after pending logtail clears


LeftHandCold:
## Summary

`disttae` can admit a stale immutable `PartitionState` after the logtail consumer has applied a table update and cleared that table's pending marker. The race was observed as a flaky `mo_ctl mergeobjects` failure in [run 29838470048 / job 88667877201](https://github.com/matrixorigin/matrixone/actions/runs/29838470048/job/88667877201): the merge saw fewer than two objects and returned `internal error: no matching objects`, while `metadata_scan` one second later saw both unmerged objects.

The failure happened on PR #25970, but that PR only changes `pkg/backup/file_index.go` and its unit test. It does not touch the `disttae`, logtail, merge, or BVT paths involved here.

## Root cause

`txnTable.getPartitionState` gets an immutable table snapshot through `PushClient.toSubscribeTable`, then calls `waitCanServeTableSnapshot`. The latter checks the already-captured snapshot first and independently reads `subEntry.pendingTo` through `canServeTableSnapshotNoWait`.

This leaves the following reachable TOCTOU window:

1. A reader captures old partition snapshot `P0`.
2. The logtail consumer applies the table update and publishes `P1`.
3. The consumer clears `pendingTo` because `P1` contains the update.
4. The reader checks `pendingTo`, sees `nil`, and admits `P0` through the no-pending fast path even though `P0.appliedTo < snapshot.Prev()`.

The ordering contract needs to be the reverse: capture the pending marker before the immutable partition snapshot while holding the same subscription generation. Then either the marker is present and an old state waits, or the marker is absent and the later snapshot necessarily includes every update already cleared from pending.

PR #25515 added the pending-table fence but did not make these two reads atomic in the required order. PR #25888 already uses the correct pending-before-snapshot ordering for the separate PK-check path; ordinary `getPartitionState` reads still use the racy sequence.

## Impact

Confirmed impact:

- flaky `mo_ctl('cn', 'mergeobjects', ...)` failures after synchronous TN flushes;
- a latest table read can transiently observe an older object set than the statement snapshot should admit.

The same mechanism is reachable by ordinary latest-state reads through `txnTable.getPartitionState`; the exact user-visible outcome depends on the query and on hitting the narrow apply/clear window.

## Reproduction / evidence

The failed BVT has the exact signature of this interleaving:

- `mo_ctl_merge.test` line 73 failed with `internal error: no matching objects`;
[...]

PR #25970, opened 2026-07-21
fix(backup): select file indexes by numeric timestamp

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #25937

## What this PR does / why we need it:

`LoadGlobalFileIndex` selected the greatest timestamp substring with ordinary
string comparison. MatrixOne timestamp components are unpadded decimal values,
so both `1-9 > 1-10` and `9-0 > 10-0` are wrong lexicographically. The stale
index then makes incremental backup recopy objects that are present only in the
newer index.

Parse the physical component as `int64` and the logical component as `uint32`,
build a `types.TS`, and compare timestamps numerically. Malformed or legacy
filenames are ignored rather than panicking or masking a valid index.

Tests cover width changes in both components, malformed and overflowing names,
and a mixed directory where a lexicographically larger malformed name must not
hide a valid index. The confirmed impact is redundant backup I/O/space; this PR
does not claim data loss.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/backup`
- `go build ./pkg/backup` (repository CGo environment)
- `go vet ./pkg/backup` (repository CGo environment)

PR #26009, opened 2026-07-22
fix(disttae): recapture table state after pending logtail apply

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26006

The latest subscribed-table read path captured an immutable `PartitionState`
before it checked the table's pending logtail marker. If the logtail consumer
published a newer state and cleared that marker between those two operations,
the reader saw `pending=false` and admitted its previously captured stale state.

This race caused the `mo_ctl_merge` BVT in job
https://github.com/matrixorigin/matrixone/actions/runs/29838470048/job/88667877201
to see fewer than two merge candidates and return `no matching objects`; one
second later, `metadata_scan` recaptured the state and saw both unmerged objects.
The job belonged to #25970, but that PR changes only backup file-index ordering
and does not touch the affected disttae/logtail path.

Capture the pending marker before the immutable partition snapshot while holding
the subscription generation. With this ordering, a cleared marker proves that
the later snapshot includes the applied update; a present marker continues to
use the existing applied-watermark wait. The no-pending fast path, historical
checkpoint fallback, cancellation, and PK-check reconnect fencing are unchanged.

Add a deterministic regression test that retains a pre-apply snapshot, publishes
a state covering the read timestamp, clears the pending marker, and verifies that
the admission path returns the refreshed state. The test failed on current main
with the stale applied watermark before the fix.

Tested with:

`go test -count=1 ./pkg/vm/engine/disttae`

`go test -race -count=1 -run '^(TestCanServeTableSnapshotBlocksPendingUpdate|TestWaitCanServeTableSnapshotWaitsForPendingUpdate|TestWaitCanServeTableSnapshotRefreshesAfterPendingApply|TestPKCheckSnapshotRejectsResetSubscriptionGeneration|TestPKCheckSnapshotRefreshesAfterPendingApply|TestPKCheckSnapshotRejectsClosedPushClientAdmission)$' ./pkg/vm/engine/disttae`

`go vet ./pkg/vm/engine/disttae`


gouhongshen (approved):
## Codex automated review

The stale-state race is correctly closed and CI is green. One non-blocking fast-path performance regression remains.

### P2 - Avoid recapturing the partition twice on every table read (pkg/vm/engine/disttae/logtail_consumer.go:398)
`getPartitionState` first calls `toSubscribeTable`, which always captures a snapshot through `isSubscribed` on success. This new call immediately discards that snapshot and repeats the subscription lookup, timestamp sampling, and `GetOrCreateLatestPart`, including a globally serialized `Engine.Lock`. Thus even the common no-pending path now acquires the global engine mutex twice per table open. Carry the pending marker alongside the initial pending-before-snapshot capture, or provide a subscription-only path so the first admission attempt needs only one capture.

pkg/vm/engine/disttae/logtail_consumer.go
```diff
@@ -389,14 +389,14 @@ func (c *PushClient) canServeTableSnapshot(
...
 	accId, dbId, tblId uint64,
-	ps *logtailreplay.PartitionState,
 	snapshot timestamp.Timestamp,
 ) (*logtailreplay.PartitionState, bool, error) {
 	ticker := time.NewTicker(time.Millisecond)
 	defer ticker.Stop()
 
 	for {
-		canServe, needWait := c.canServeTableSnapshotNoWait(dbId, tblId, ps, snapshot)
+		ps, pending := c.getSubscribedSnapshotAndPending(ctx, accId, dbId, tblId)
```
XuPeng-SH: [P1] Avoid a second global partition capture on every table read. `txnTable.getPartitionState` has just called `toSubscribeTable`; on the already-subscribed fast path that calls `isSubscribed`, which already executes `GetOrCreateLatestPart(...).Snapshot()` under the single global `Engine.Lock`. This first loop iteration unconditionally discards that snapshot and calls the same globally serialized capture again, even when there is no pending update and no wait is needed. `getPartitionState` is used by scans, merge, object-list, snapshot-scan, change-handle, and many other paths, so this affects ordinary reads across all tables, not only the rare race window.

I measured the exact old vs new fast paths with a temporary review-only benchmark that kept the ticker, subscription lookup, and pending check identical: serial admission increased from about 153 ns/op to 195 ns/op (+27%), and 10-way parallel admission from about 193 ns/op to 377 ns/op (about 2x), with identical allocations. Please make the initial subscription capture itself return the pending marker read before its snapshot, pass that coupled result into this function, and only call `getSubscribedSnapshotAndPending` after the ticker fires for a real retry.

XuPeng-SH (changes requested):
Requesting changes on exact head `1d4bd8d966108cdaf5f5bfccfc6f0f9957858779`. The stale-state correctness race is closed, generation/cleanup/wait paths are sound, and all CI plus local package/race validation pass. One blocking hot-path regression remains: the first admission attempt performs two globally serialized partition captures for every already-subscribed table read. Please preserve the pending-before-snapshot ordering while carrying that coupled result out of the subscription fast path, then recapture only after an actual wait/retry.

Q1-Q3 audit:
- Q1: immutable snapshots and ticker ownership close cleanly.
- Q2: subscription/engine lock order is consistent; waits terminate through apply or context cancellation.
- Q3: pending state is one bounded atomic marker; no queue/container growth.

Local validation:
- focused concurrency regressions under `-race -count=20`
- full `go test -race ./pkg/vm/engine/disttae`
- `go vet ./pkg/vm/engine/disttae`
- `go build ./pkg/vm/engine/disttae`

XuPeng-SH (approved):
Re-reviewed exact head `b79905f85338bc1883c05c6a916b578a70f2be8e`. No remaining actionable findings.

The previous hot-path blocker is fixed: `getPartitionState` now reuses the pending-before-snapshot capture returned by `toSubscribeTable`, so the common subscribed/no-pending path performs one partition capture; recapture occurs only after a real pending wait/retry. The stale-state race remains closed because both the initial attempt and retries validate the subscription generation and read `pendingTo` before taking the immutable snapshot. Context cancellation, historical checkpoint fallback, and PK-check reconnect fencing are preserved.

Q1-Q3: no new resource ownership, all waits terminate via apply/recapture or caller cancellation, and pending state remains one bounded atomic marker per subscription.

Validation on this exact head passed: focused concurrency regressions under `-race -count=20 -shuffle=on`, the complete disttae package both normally and under `-race`, `go build`, `go vet`, and `git diff --check`. Current-head CI has no reported failure at review time.

Issue #25922, opened 2026-07-21
[Bug] TPCH Q16 hangs indefinitely: Shuffle execution engine deadlock in waitBatchOrEnd (TPCH 100G stuck for 15+ hours)


Ariznawlll:
## Issue

TPCH Q16 (supplier count query) hangs indefinitely on TPCH 100G dataset due to a **deadlock in the Shuffle execution engine**. The query has been running for **15+ hours** (normally completes in ~3 seconds). The root cause is that Shuffle consumer goroutines are permanently blocked in `ShufflePool.waitBatchOrEnd()` because the producer pipeline exited without sending the end-of-data signal.

This causes the entire TPCH 100G benchmark to stall — Q1-Q15 complete normally, but Q16 never returns.

## Environment

| Field | Value |
|---|---|
| **CI Run** | https://github.com/matrixorigin/mo-auto-test/actions/runs/29764528161/job/88440134273 |
| **Branch** | main |
| **MO commit** | `0102f7b85bcb8d960d7cc16a14b71682a58e7910` |
| **Runner** | `mo-srv-129` (10.222.1.129, IDC self-hosted) |
| **Topology** | Single CN + Single TN (launch mode) |
| **GOMEMLIMIT** | 80000 MiB |
| **Stuck step** | `Start TPCH 100G Test` (started 2026-07-20 19:08:32 UTC, still running after 15+ hours) |

## Stuck Query (TPCH Q16)

```sql
SELECT p_brand, p_type, p_size, count(distinct ps_suppkey) as supplier_cnt
FROM partsupp, part
WHERE p_partkey = ps_partkey
    AND p_brand <> 'Brand#35'
    AND p_type not like 'ECONOMY BURNISHED%'
    AND p_size in (14, 7, 21, 24, 35, 33, 2, 20)
    AND ps_suppkey not in (
        SELECT s_suppkey FROM supplier WHERE s_comment like '%Customer%Complaints%'
    )
GROUP BY p_brand, p_type, p_size
ORDER BY supplier_cnt desc, p_brand, p_type, p_size
```

- **Query started**: 2026-07-20 19:10:19 CST (confirmed via `mo_catalog.mo_sessions`)
- **Running for**: 912+ minutes (15.2+ hours) and still not completed
- **Normal duration**: 3.0-3.2 seconds per turn (baseline from 2026-07-18 run)
- **Regression factor**: ~10000x

## Query Execution Plan

```
[...]

PR #26017, opened 2026-07-22
fix: isolate fixed-bucket shuffle backpressure

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25922

The unified shuffle implementation bounded every ready batch with one global credit pool. In fixed-bucket local shuffle, a hot bucket could consume all credits and prevent producers from publishing batches to unrelated empty buckets. Those producers then waited for global space while the corresponding consumers waited for data or shuffle completion.

This change partitions fixed-mode backpressure by bucket:

- each fixed bucket can retain at most two ready batches;
- draining a bucket wakes only writers blocked on that bucket;
- drain-all shuffle keeps its existing global ready-batch limit.

For the valid local-shuffle layout, bucket count equals DOP, so the total ready-batch bound remains `2 * DOP`.

A deterministic regression fills one bucket until it is backpressured, verifies another bucket can still accept work, verifies the matching waiter is woken, and checks ready-credit cleanup.

Validation:

- full `pkg/sql/colexec/shuffle` tests
- full shuffle package under `-race`
- focused race stress, 20 repetitions
- `pkg/sql/colexec/group` tests
- `git diff --check`


XuPeng-SH (approved):
## Codex deep review — approved

Reviewed exact head `be83e701c24ddb24f8d1c145e276d2b11a03b0ca`. No blocking finding remains.

The fixed-mode wait-for graph now terminates locally: each bucket owns an independent two-ready-batch credit pool, its matching consumer releases and closes only that bucket waiter, and abort/all-stop still close the shared ending signal. The local group/join construction paths prove `bucketNum == DOP == maxHolders`, so total retained ready batches remain bounded by `2 * DOP`; drain-all keeps the previous global limit. BatchSet partial-write failures reconcile reserved versus actual ready deltas, and final holder cleanup remains exactly once.

Validation performed on this head:

- the new regression passes, while the same test transplanted onto base `50319026a6` deterministically fails at the unrelated bucket write;
- focused fixed-bucket regression and two-worker row-preservation tests pass under `-race -count=20`;
- full `pkg/sql/colexec/shuffle` passes under `-race`;
- `pkg/sql/colexec/group` and focused compile topology tests pass;
- `go vet`, `go build`, gofmt, and `git diff --check` pass;
- current CI is green, and a merge-tree check against current `origin/main` is conflict-free.

Issue #25985, opened 2026-07-21
[Bug]: Prepared SELECT ignores a same-name temporary table created after PREPARE


Ariznawlll:
### Issue Description

A prepared `SELECT` continues to read the permanent table when a same-name temporary table is created after `PREPARE`.

Direct SQL resolves the temporary table correctly. MySQL also re-resolves the prepared statement to the temporary table. MatrixOne keeps the previously compiled permanent-table plan until the temporary table is removed.

### Environment

- Branch: `main`
- Commit: `8480da08f318a9cb998efb80062d5573da776fb7`
- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- Comparison: MySQL `8.0.45`

### Reproduction

```sql
DROP DATABASE IF EXISTS ps_temp_shadow;
CREATE DATABASE ps_temp_shadow;
USE ps_temp_shadow;

CREATE TABLE t(id INT PRIMARY KEY, v VARCHAR(20));
INSERT INTO t VALUES (1, 'permanent');

PREPARE s FROM 'SELECT id,v FROM t ORDER BY id';
EXECUTE s;

CREATE TEMPORARY TABLE t(id INT PRIMARY KEY, v VARCHAR(20));
INSERT INTO t VALUES (2, 'temporary');

SELECT id,v FROM t ORDER BY id;
EXECUTE s;

DROP TABLE t;
SELECT id,v FROM t ORDER BY id;
EXECUTE s;
```

### Actual Result

MatrixOne returns the temporary row for the direct `SELECT`, but the prepared statement still returns the permanent row:

```text
direct SELECT:  2  temporary
EXECUTE s:      1  permanent
```

The result is stable in three consecutive runs.

### Expected Result

After the same-name temporary table is created, `EXECUTE s` should resolve it in the same way as direct SQL and MySQL:

```text
EXECUTE s:      2  temporary
```

After the temporary table is dropped, execution should return to the permanent table.

### Code Analysis

`TxnCompilerContext.Resolve` correctly prefers the session temporary-table mapping. However, `initExecuteStmtParam` rebuilds a prepared plan only when a collected catalog table reports a newer version. Adding a session temporary-table alias does not change the stored permanent-table catalog entry, so the cached plan and compile continue to reference the permanent table.

Ariznawlll:
最新 main 回归结果（2026-07-22）

- 测试基线：`c0486a2446c84c6dd6bf142494fe85243877ddc5`
[...]

PR #26018, opened 2026-07-22
fix(frontend): re-resolve prepared queries after temporary table changes

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25985

Prepared statements only invalidated their cached plans when catalog table
versions changed. Creating or dropping a session temporary table changes name
resolution without changing the permanent table's catalog version, so an
`EXECUTE` could continue reading the table resolved at `PREPARE` time.

This change tracks a session-local temporary-table mapping version and stores
the version used to build each prepared plan. When the mapping changes, the
next execution rebuilds the prepared plan and cached compile. Statements still
reuse their cached compile when neither catalog metadata nor temporary-table
resolution changed.

The regression coverage verifies both transitions:

- a prepared query switches from a permanent table to a newly created
  same-name temporary table;
- after the temporary table is dropped, the query switches back to the
  permanent table.

Validation:

- `go build ./pkg/frontend`
- `go vet ./pkg/frontend`
- full `pkg/frontend` CGo test suite
- focused race-enabled frontend tests
- `make build`
- `temporary_table_basic.sql` BVT: 103/103 passed


gouhongshen (approved):
## Codex automated review

No findings. Existing PR reviews and inline comments were empty; the sole issue comment was unrelated. Focused tests and vet were blocked by missing local CGo/third-party prerequisites (libmo and headers).

XuPeng-SH (changes requested):
Blocking correctness regression: repreparation uses the execution-time default database.

The session-wide tempTableVersion invalidates every prepared statement, and initExecuteStmtParam rebuilds the plan with the current session compiler context. PrepareStmt does not retain the default database from PREPARE time, so an intervening USE can silently bind unqualified names to a different table.

Reproduced on the exact PR head:
1. Create d1.t with value 1 and d2.t with value 2.
2. USE d1; PREPARE s FROM "SELECT v FROM t".
3. USE d2; CREATE TEMPORARY TABLE x(v INT).
4. EXECUTE s returns 2 from d2.t; it must still return 1 from d1.t.

The exact base commit returns 1. MySQL 8.4 also keeps the PREPARE-time database. In the issue-specific variant, creating TEMPORARY TABLE d1.t after switching to d2 should make the prepared query read that temporary d1.t, but this PR instead reads permanent d2.t.

Please retain and use the PREPARE-time default database when rebuilding (and account for the corresponding reprepare context), and add regressions covering USE between PREPARE and both an unrelated temporary-table change and a same-name temporary table in the original database.

XuPeng-SH (approved):
Re-reviewed the latest head after the previous database-context blocker. The fix now captures the PREPARE-time default database and temporarily restores it while rebuilding after temporary-table mapping changes, then restores the EXECUTE-time context. The new BVT covers both required transitions after USE: an unrelated temporary-table change keeps binding to the original database, and a same-name temporary table in that original database shadows/unshadows correctly.

Also reviewed the temp-table mapping generation, forward/reverse map updates, failed metadata refresh state, cached compile replacement, result metadata refresh, and remap-db propagation. Verified locally with the full frontend package, focused race tests repeated 20 times, build, and vet. The previous blocking finding is resolved. LGTM.

heni02 (approved):

PR #26021, opened 2026-07-22
fix: refresh prepared DDL plans after schema changes

### LINKED ISSUE #25986: [Bug]: Prepared schema-changing statements reuse stale validation state across EXECUTE
### Issue Description

Prepared schema-changing statements reuse validation state built at `PREPARE` time instead of validating the current table definition on every `EXECUTE`.

This causes repeated `ADD COLUMN` and `DROP INDEX` operations to silently succeed after the target state is already reached. Repeated `DROP COLUMN` can either silently succeed or expose an internal hidden-index-table error, depending on whether the table has a secondary index. The equivalent direct statements and MySQL return the appropriate user-facing errors.

### Environment

- Branch: `main`
- Commit: `8480da08f318a9cb998efb80062d5573da776fb7`
- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- Comparison: MySQL `8.0.45`

### Reproduction

```sql
DROP DATABASE IF EXISTS ps_alter_repeat;
CREATE DATABASE ps_alter_repeat;
USE ps_alter_repeat;

CREATE TABLE prepared_t(id INT PRIMARY KEY);
PREPARE a FROM
  'ALTER TABLE prepared_t ADD COLUMN note VARCHAR(20) DEFAULT ''x''';

EXECUTE a;
SHOW CREATE TABLE prepared_t;

EXECUTE a;
SHOW CREATE TABLE prepared_t;

ALTER TABLE prepared_t DROP COLUMN note;
EXECUTE a;
SHOW CREATE TABLE prepared_t;
```

### Actual Result

- The first `EXECUTE a` adds `note`.
- The second `EXECUTE a` succeeds without an error although `note` already exists.
- After the column is dropped, the same prepared statement adds it again.

The result is stable in three consecutive runs. A direct second `ALTER TABLE ... ADD COLUMN note` correctly returns error 1060 in MatrixOne.

The same three-run matrix also found:

| Prepared statement executed twice | MatrixOne second execution | MySQL second execution |
| --- | --- | --- |
| `ALTER TABLE t ADD COLUMN note ...` | silently succeeds | duplicate column error |
| `ALTER TABLE t DROP COLUMN note` without another index | silently succeeds | missing column error |
| `ALTER TABLE t DROP COLUMN note` with a secondary index | internal `__mo_index_secondary_*` table error | missing column error |
| `DROP INDEX idx_v ON t` | silently succeeds | missing index error |

Repeated prepared `CREATE INDEX` and `RENAME TABLE` do return errors, so the problem is operation-dependent rather than every prepared DDL statement silently succeeding.

### Expected Result

Every execution should validate the current table definition. The second execution should return the corresponding duplicate/missing column or missing index error, matching direct SQL and MySQL, and should not expose a hidden table name.

### Code Analysis

The prepared DDL plan contains the table definition built at prepare time. `ResetPreparePlan` collects schema dependencies for query plans, while the DDL branch only examines an optional nested query. As a result, `initExecuteStmtParam` does not observe the table change and reuses the prepared DDL plan without rebuilding the current `ALTER TABLE` validation state.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25986

Prepared schema-changing statements did not record their target tables as schema dependencies. As a result, repeated `EXECUTE` calls could reuse DDL plans built from stale table definitions and skip duplicate or missing object validation.

This change collects the target table metadata for prepared `ALTER TABLE`, `CREATE INDEX`, and `DROP INDEX` plans. Existing schema-version checks can then rebuild the prepared plan whenever the table definition changes, while preserving compile-cache reuse for unchanged prepared queries.

The regression coverage verifies repeated add/drop column and create/drop index operations, including a drop-column case with a secondary index.

Validation:

- `go test -count=1 -timeout=180s ./pkg/sql/plan`
- `go test -count=1 -timeout=180s -run 'TestInitExecuteStmtParam' ./pkg/frontend`
- `mo-tester` prepared DDL BVT: 29/29 passed


gouhongshen (changes requested):
## Codex automated review

One blocking correctness regression found in prepared CLONE execution under database remapping.

### P1 - Prepared CLONE loses database remapping (pkg/frontend/mysql_cmd_executor.go:1890)
`GetDatabase()` is the actual current database, while `DefaultDatabase()` applies the saved remap for unqualified names. `remapDbInStmt` has no `CloneTable` case, so CLONE relies on `DefaultDatabase()`. This new serialization then qualifies both source and destination with the actual database and marks them explicit, causing every refreshed EXECUTE to bypass remapping and clone from/create in the wrong database or fail. Preserve the effective remapped database semantics when creating the fresh CLONE AST.

gouhongshen (approved):
## Codex automated review

Previous P1 on prepared CLONE remapping is fixed: `cloneSQL` now captures the remapped effective default database, with unit and BVT coverage. No other prior blockers exist; the follow-up merge had no conflict-resolution delta. Focused Go test could not run because `go` is unavailable in this environment.

XuPeng-SH (changes requested):
Blocking correctness issue remains on latest head `9b04ddccbbf6`.

A prepared statement rebuilt against transaction-local DDL remains cached after that transaction rolls back:

```sql
create table t(a int);
prepare s from 'select * from t';
begin;
alter table t add column b int;
execute s;
rollback;
execute s;
```

I reproduced this on the latest head with binary COM_STMT_PREPARE/EXECUTE. The final EXECUTE still reports columns `[a b]`, while the rolled-back table has only `[a]`.

`recordSessionDDL` advances `ddlVersion` after successful DDL, and EXECUTE stores the workspace-derived `PreparePlan`, `ColDefData`, cached compile, and matching generation. The full ROLLBACK path does not advance that generation, so the next transaction reuses state derived from a discarded workspace.

Please capture `workspace.GetHaveDDL()` before the transaction operator is cleared and, after a successful full rollback, advance the session DDL generation when that transaction contained DDL. This should cover explicit ROLLBACK, commit-failure/error rollback, and text/binary prepared statements retained across rollback. It should remain conditional on transaction-local DDL, rather than invalidating every rollback.

The prepared CLONE remap issue from the previous review is fixed by `6afe565744`; this request is for the remaining rollback lifecycle gap.

XuPeng-SH (changes requested):
Requesting changes at `d309dbb463` for the following production blockers:

1. **Failed rollback can preserve a prepared plan built from discarded transaction-local DDL.** `rollbackUnsafe` advances `ddlVersion` only when `TxnOperator.Rollback` succeeds. On error or panic it discards the frontend transaction without advancing the generation. DistTAE rollback can call `txn.delTransaction()` and still return `loadCleanupErr`, so the workspace may already be gone. Aborted DDL never reaches `CatalogCache`, leaving no other invalidation signal. Please advance the generation whenever `haveDDL` is true and this frontend transaction is discarded, regardless of rollback result. The current failed-rollback test incorrectly expects version 0; please also cover the subsequent prepared execution.

2. **DDL executed through `backSession` does not invalidate the upstream user session.** Both `recordSessionDDL` and `advanceDDLVersionAfterDiscardedTxnDDL` only handle a direct `*Session`. Stored-procedure SQL executes through `*backSession`, while the outer `CALL` is not catalog-changing. A procedure that drops and recreates a table can therefore leave the caller prepared plan and binary metadata reusable until asynchronous logtail delivery. Please centralize scoped generation propagation through the upstream session chain and add background or stored-procedure commit and rollback coverage.

3. **Stable prepared queries on subscription databases now execute synchronous catalog SQL on every `EXECUTE`.** `preparedSubscriptionSchemaChanged` calls `Resolve` before the cheap catalog-cache check. Production `Resolve` performs subscription account and publication catalog queries through a background executor. The benchmark uses an in-memory resolver and does not measure this cost. Please use a cheap subscription generation or high-watermark signal and resolve only when change is possible; also deduplicate collected dependencies.

4. **Historical snapshot subscription dependencies can rebuild forever.** Collection drops the scan or view snapshot, while validation resolves the current subscription with a nil snapshot. If the historical object is v1 and the current publication is v2, every execution detects a change, rebuilds against historical v1 again, and repeats. Please preserve the dependency snapshot or identity, or define an immutable-snapshot strategy, with a test proving the second unchanged execution does not rebuild.

The overall dependency plus session-generation direction is valid and materially improves #25986, but these invalidation gaps and the subscription hot-path regression need to be closed before merge. CI status is intentionally not part of this review.

heni02 (approved):

XuPeng-SH (approved):
Re-reviewed exact head `08a674c22181576cd0457434a0444c81d2e6ef50`. The previously reported rollback lifecycle, upstream background-session propagation, subscription hot-path, and historical-snapshot identity issues are addressed. Focused race tests, full affected-package tests, build/vet, and the issue #25986 embedded regression all pass.

PR #26023, opened 2026-07-22
fix(frontend): bind background exec to caller session

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25414

fix(frontend): bind background exec to caller session


XuPeng-SH (approved):
Deep review of exact head `98cd88a013ee9b58f934d51c145eb870ffcb3388`: no blocking findings.

The production change closes the existing planner/executor session-scope gap correctly. `backSession` already forwards temporary-table alias resolution to its caller session; the new binding gives the execution process the same forwarding session and uses the caller UUID for physical temporary-table names. Nested shared-transaction background executors preserve the outer session, while non-shared nested execution delegates to it, so aliases and GC identity stay in one session generation.

I traced create -> alias registration -> use -> explicit drop -> session-close cleanup -> orphan GC. Registration happens only after physical creation, drop removes the alias after physical deletion, session close snapshots and clears the maps, and GC now recognizes procedure-created tables as owned by the active caller session. The change adds no goroutine, wait, or unbounded container.

Validation:

- new binding tests: `go test -race -count=20 ./pkg/frontend -run TestBindBackExecSession`
- existing temp alias/GC tests plus new tests: `go test -race -count=5 ./pkg/frontend -run 'Test(BindBackExecSession|SessionTempTableMap|CleanOrphanTempTables|StartTempTableGC)'`
- full `pkg/frontend` test suite
- `go vet ./pkg/frontend`
- trial merge with latest main `2eb7f523e5a10f352d3e258183ed8440510388de`, followed by the focused test

All passed. The added BVT coverage exercises temporary-table use within a procedure, visibility after returning to the caller, and nested procedure propagation.

gouhongshen (approved):
## Codex automated review

No findings. The session binding, caller UUID propagation, and nested shared-executor propagation correctly preserve temporary-table ownership and GC visibility; existing reviews/comments and inline comments were assessed.

PR #26024, opened 2026-07-22
fix(frontend): bind background exec to caller session

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25414 

fix(frontend): bind background exec to caller session

### DIFF
--- a/pkg/frontend/authenticate.go
+++ b/pkg/frontend/authenticate.go
@@ -11431,7 +11431,7 @@ func doInterpretCall(ctx context.Context, ses FeSession, call *tree.CallStmt, bg
 		}
 		defer freeStatements(stmt)
 
-		err = interpreter.ExecuteSp(stmt[0], dbName)
+		err = interpreter.ExecuteSp(stmt[0], dbName, bg)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/frontend/back_exec.go
+++ b/pkg/frontend/back_exec.go
@@ -378,6 +378,7 @@ func doComQueryInBack(
 		StorageEngine: pu.StorageEngine,
 		Buf:           backSes.buf,
 	}
+	bindBackExecSession(proc, backSes)
 	proc.SetStmtProfile(&backSes.stmtProfile)
 	proc.SetResolveVariableFunc(backSes.txnCompileCtx.ResolveVariable)
 	// Frontend back-exec — session-bound resolver. backSession is a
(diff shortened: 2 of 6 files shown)


gouhongshen (changes requested):
## Codex automated review

Loaded all existing PR reviews, issue comments, and inline threads (no human review feedback existed). Static diff/gofmt checks pass. Focused frontend test could not run because this checkout lacks cgo/libmo.dylib. One blocking nested-procedure regression remains.

### P1 - Nested procedures still cannot create temporary tables (pkg/frontend/back_exec.go:523)
A nested CALL reaches `backSes.GetShareTxnBackgroundExec` (back_self_handle.go:96). Its active transaction takes `backSession.InitBackExec`'s txn branch (back_exec.go:950), which creates another back session without setting `upstream`. The new binder therefore returns at this nil-upstream guard, leaving that nested executor's process without a session; `Scope.CreateTable` then returns `session not found for temporary table`. Propagate the original client session through nested back executors (or resolve the root upstream when binding), and add a nested-CALL temporary-table regression case.

iamlinjunhong:
> ## Codex automated review
> Loaded all existing PR reviews, issue comments, and inline threads (no human review feedback existed). Static diff/gofmt checks pass. Focused frontend test could not run because this checkout lacks cgo/libmo.dylib. One blocking nested-procedure regression remains.
> 
> ### P1 - Nested procedures still cannot create temporary tables (pkg/frontend/back_exec.go:523)
> A nested CALL reaches `backSes.GetShareTxnBackgroundExec` (back_self_handle.go:96). Its active transaction takes `backSession.InitBackExec`'s txn branch (back_exec.go:950), which creates another back session without setting `upstream`. The new binder therefore returns at this nil-upstream guard, leaving that nested executor's process without a session; `Scope.CreateTable` then returns `session not found for temporary table`. Propagate the original client session through nested back executors (or resolve the root upstream when binding), and add a nested-CALL temporary-table regression case.

fixed

gouhongshen (changes requested):
## Codex automated review

The author’s new changes fix the previously reported nested-CALL path, and the new nested BVT covers it. One blocking unhappy-path regression remains: transaction rollback does not roll back the caller-session temporary-table registry.

### P1 - Rollback leaves the caller’s temporary-table registry inconsistent (pkg/frontend/back_exec.go:526)
Binding `proc.Session` makes procedure DDL mutate the caller registry through `backSession.AddTempTable`/`RemoveTempTable`. Those mutations happen immediately in `Scope.CreateTable`/`DropTable`, but a later procedure error makes `ExecuteSp` call `finishTxn`, which rolls back only the storage transaction; no path restores the registry. For example, a procedure that creates temporary table `tmp_t` and then executes a failing statement leaves `tmp_t` registered after its physical table is rolled back. A subsequent `CREATE TEMPORARY TABLE tmp_t` therefore reports “table already exists,” while queries resolve to the rolled-back physical name and fail. Dropping a pre-existing temporary table before an error has the inverse problem. Snapshot or transactionally stage the caller’s temporary-table mappings and restore them on rollback, and add failing-procedure create/drop regression cases.

heni02 (approved):

PR #26025, opened 2026-07-22
fix(pipeline): normalize retained dispatch receiver routes

### LINKED ISSUE #25816: [Bug] 30s remote dispatch watchdog can falsely fail slow receiver registration

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25816

For a nested `RemoteRun`, encoding can retain a root `Dispatch` on the caller
CN while sending its child scope to the target CN. The decoded receiver metadata,
however, still carries the compile-time target CN in `FromAddr`. When runtime
placement selects a different caller CN, the retained `Dispatch` is registered
and executed on that caller, but the receiver sends its prepare notification to
the dormant compile-time target. The active `Dispatch` never receives its
stream and waits indefinitely.

This change normalizes `FromAddr` before remote-pipeline registration, limited
to receiver UUIDs owned by a root `Dispatch` that standalone `RemoteRun`
actually retains and executes on the current caller. Ordinary remotely executed
operator trees remain unchanged.

Since #25990, a non-standalone remote scope is rejected by
`failRemoteRunBeforeStart`; it no longer falls back to local `MergeRun`.
Normalization therefore stops at rejected scopes and does not rewrite routes
owned by their root, nested operators, or descendants. Regression tests pin
that rejection boundary.

The change does not alter planning, the wire format, dispatch semantics, RPC
timeouts, or cluster topology.

Validation on latest `main` (`90e3400436`):

- focused regression, `-count=20`
- focused regression with `-race -count=3`
- full `pkg/sql/compile` package
- full `pkg/sql/compile` package with `-race`
- `go build -mod=readonly ./pkg/sql/compile`
- `go vet -mod=readonly ./pkg/sql/compile`
- `git diff --check`

The original issue reproduction was also validated on `10.222.1.50` before
this review follow-up: commit `51f629bd202b38b33fd5fc7d6b3052b0c8242db4`
was built with `make dev-build TYPECHECK=0`, the 2-CN dev cluster was recreated,
and the reproduction query completed repeatedly through both CNs with identical
expected results, no lingering query, timeout, panic, receiver error, restart,
or OOM kill.

Regression coverage includes retained-root placement, ordinary remote-tree
negative control, rejected root and nested dispatches, rejected descendants,
unrelated UUIDs, and nil input.


gouhongshen (approved):
## Codex automated review

No actionable regressions found. The normalization matches RemoteRun’s retained root-Dispatch ownership, runs before registration/goroutines, preserves ordinary remote dispatch routes, and adds focused positive/negative coverage. Existing PR discussion had no substantive reviews or inline threads. CI is green; local focused testing was unavailable because the required prebuilt libmo.dylib is absent.

XuPeng-SH (changes requested):
[P1] Normalize dispatch owners when RemoteRun falls back to local execution

The `!checkPipelineStandaloneExecutableAtRemote(s)` branch on lines 105-107 only descends into `PreScopes`, so it never records a `Dispatch` in this scope's own operator tree. This is inconsistent with `RemoteRun`: the same predicate makes it call `s.MergeRun(c)`, which executes the entire scope (including its root/operator-tree `Dispatch`) on the caller CN. `registerLocalDispatchReceivers` likewise registers that whole fallback tree on the caller. Any matching receiver therefore keeps its compile-time `FromAddr`, sends prepare-notify to the dormant target CN, and never attaches to the active dispatch on the caller; the valid query retries/waits until its context is cancelled.

I reproduced this on the exact head with a `Remote` scope targeting cn2, a root dispatch, and a child connector whose external `WaitRegister` forces the standalone check to return false. After normalization on cn1, the receiver's owner was still cn2 (expected cn1). The existing `TestGroupShuffleBucketsByCNIfNeeded` documentation also identifies this reachable local-fallback/compile-time-FromAddr mismatch as a hang. Please collect dispatch UUIDs from every operator tree that the fallback `MergeRun` retains locally (and add root plus nested-dispatch regression cases), rather than scanning only the pre-scopes.

XuPeng-SH:
@LeftHandCold need change

aunjgr (changes requested):
[P2] Do not normalize dispatches in a remote scope that is rejected before start — pkg/sql/compile/remote_dispatch_placement.go:107\n\nThis branch models behavior that no longer exists on the PR's base. Since #25990, Scope.RemoteRun handles !checkPipelineStandaloneExecutableAtRemote(s) by calling failRemoteRunBeforeStart and returning an error (scope.go:438-440); it never calls MergeRun and therefore never executes this scope's operator tree or its PreScopes on the caller. Lines 107-118 nevertheless classify every Dispatch in that rejected tree as caller-owned and rewrite matching receiver FromAddr values to the local CN. That can send a concurrently started sibling's prepare notification to a registration whose owner is being cancelled, and it contradicts this helper's stated invariant that only runtime-retained dispatches are rewritten. The two new fallback tests assert this impossible placement rather than a reachable success path.\n\nPlease make the non-standalone branch return without collecting or descending, remove/update the fallback tests and PR description, and retain the valid standalone root-Dispatch normalization. The earlier fallback review was based on the old pre-#25990 RemoteRun behavior.\n\nI otherwise traced registration ownership and wait termination: normalization runs before publication/goroutines, registrations retain exactly-once cleanup, ordinary remote Connector->Dispatch routes remain unchanged, and the focused race test (10 repetitions) plus the full pkg/sql/compile suite pass.

gouhongshen (changes requested):
## Codex automated review

One blocking correctness issue: the new fallback branch rewrites receiver routes for remote scopes that the current base rejects before execution.

### P2 - Do not normalize dispatches in a remote scope that fails standalone validation (pkg/sql/compile/remote_dispatch_placement.go:107)
`Scope.RemoteRun` now calls `failRemoteRunBeforeStart` and returns when a non-local scope fails `checkPipelineStandaloneExecutableAtRemote` (`pkg/sql/compile/scope.go:438-440`); it does not execute the tree locally. This branch nevertheless collects every dispatch UUID in that rejected tree, and the later update rewrites matching `FromAddr` values to `localAddr`. Registration occurs before `s.MergeRun` (`remoterunServer.go:281-296`), so concurrently started sibling receiver scopes can observe the rewritten address while the rejected scope is being cancelled, causing notifications to target the wrong CN. Remove this collection/descending path and retain normalization only for dispatches that runtime placement actually retains.

XuPeng-SH (approved):
Re-reviewed exact head 2f160db514e9db1ca26e7d4a7a52a7cce2612f57 after the latest correction. No blocking findings.

The previous requested change is resolved: current RemoteRun rejects non-standalone remote scopes before execution, and this head now stops normalization at that boundary. The positive path remains correctly limited to UUIDs owned by a standalone remote scope root Dispatch that RemoteRun retains on the actual caller CN. Dispatches executed in the ordinary remote operator tree, dispatches below a retained Connector, rejected roots/nested operators/descendants, and unrelated UUIDs keep their original routes.

I also traced the publication and termination paths: normalization occurs before receiver registration and goroutine start; partial registration failure cancels the scope tree and rolls registrations back; cleanup is exactly-once; notification retry terminates with the query context. The helper adds only scope/UUID-bounded temporary maps.

Local validation on this exact head:
- focused normalization regression: -count=50
- placement/registration/rollback/encoding/rejection/shuffle matrix: -race -count=10
- full ./pkg/sql/compile: normal and -race
- go vet ./pkg/sql/compile
- git diff --check

All passed. CI on this head is green.

gouhongshen (approved):
## Codex automated review

No actionable regressions found. The normalization matches RemoteRun’s retained root-Dispatch placement, stops at rejected scopes, runs before registration and goroutine startup, and preserves ordinary remote operator routes. Focused local tests could not start because libmo.dylib is absent; exact-head CI and reported focused/race suites passed.

PR #26026, opened 2026-07-22
perf(trace): avoid config allocation for retired kinds

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25472

`NonRecordingTracer.Start` still builds a `SpanConfig` and applies options for controlled span kinds whose recording pipelines have been retired. In the common production shape, the caller passes exactly one concrete `KindOption`, so that work is unnecessary.

This PR adds a narrow fast path for exactly one retired `KindOption`. It returns the normalized context and `NoopSpan` before constructing `SpanConfig`.

The general path remains unchanged for:

- zero options and non-retired kinds;
- custom `SpanStartOption` implementations;
- multiple options, including order-sensitive kind overrides;
- the alternate disabled-observability `NoopTracer` provider.

Regression coverage includes all four retired controlled kinds, nil-context normalization, custom-option side effects, and both multi-option orderings.

Validation on current `main`:

- `go test -count=1 ./pkg/util/trace ./pkg/util/trace/impl/motrace`
- focused `go test -race` and `-count=100` tests
- `go vet ./pkg/util/trace`
- compile-only checks for `pkg/frontend`, `pkg/sql/compile`, `pkg/tnservice`, `pkg/fileservice`, and `pkg/common/morpc`
- five benchmark runs: fast path `18.19-23.25 ns/op, 16 B/op, 1 alloc/op`; general-path reference `51.49-59.10 ns/op, 160 B/op, 2 alloc/op`

Independent fresh-data A/B/A on mo-55 measured:

- TPS: `+0.45%`
- CPU/query: `-0.56%`
- alloc bytes/query: `-3.86%`
- malloc objects/query: `-2.35%`
- GC cycles/query: `-3.98%`

The allocation profile removed approximately `4.018 GB / 60s` and `29.25M objects / 60s` attributed to `NonRecordingTracer.Start`, without an equivalent hotspot migration. The throughput result is intentionally treated as neutral; the demonstrated benefit is reduced allocation churn.

The remote performance run used base `3046f14d94b3f1a25839eeeb65729b64f85ac647`; the affected implementation and surrounding provider/config files were unchanged between that base and this PR's current-main base.


XuPeng-SH (approved):
Reviewed exact head 9d4f861a. No blocking issues found.

The single-KindOption fast path is semantically equivalent for retired kinds, while custom and ordered/multiple options retain the general path. Nil-context behavior and trace-context propagation remain unchanged.

Verified locally: target and dependent package tests, 100x focused race tests, go vet, formatting/diff checks, and allocation benchmarks (16 B/1 alloc fast path vs 160 B/2 alloc general path).

gouhongshen (approved):
## Codex automated review

No findings. The exact `KindOption` fast path preserves prior behavior; custom and ordered options retain the general path. Focused trace tests, vet, and allocation benchmarks passed.

PR #25883, opened 2026-07-20
feat(plan): support ALTER TABLE AUTO_INCREMENT = value

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

## Which issue(s) this PR fixes:

issue #23143

## What this PR does / why we need it:

This PR continues the unfinished work from #24531 while preserving the original commits and authorship.

It implements `ALTER TABLE ... AUTO_INCREMENT = N` across planning, execution, allocator state, COPY ALTER, schema epochs, recovery, and regression coverage.

The takeover review also closes the two remaining P1 invariants:

- COPY allocator reconciliation now carries source state by stable destination column identity, including drop/add/reorder/rename and hidden fake-PK movement, instead of mixing source and destination positional indexes;
- legacy CN writes without the epoch presence bit are treated as epoch zero, so they remain compatible before the first fenced ALTER but fail closed after the table epoch advances.

Validation:

- GitHub reports the PR mergeable against current `main`;
- `go test -count=1` passed for `pkg/sql/colexec/table_clone`, `pkg/sql/compile`, `pkg/vm/engine/tae/rpc`, and `pkg/vm/engine/tae/txn/txnimpl`;
- `go vet` and `go build` passed for the same affected packages;
- focused race tests passed 10 consecutive runs for COPY identity reconciliation and TN epoch convergence;
- changed-commit whitespace checks passed.

Issue #23143, opened 2025-11-25
ALTER TABLE AUTO_INCREMENT = value not supported


dengn:
## Description

MatrixOne does not support the `ALTER TABLE ... AUTO_INCREMENT = value` syntax, which is a standard MySQL feature for setting the next AUTO_INCREMENT value. When attempting to modify the AUTO_INCREMENT starting value, it returns an error indicating that this alter option is unsupported in inplace mode.

## Table Schema

```sql
CREATE TABLE ai_alter_set_test (
    id INT AUTO_INCREMENT PRIMARY KEY,
    data VARCHAR(50)
);
```

## SQL Statements

```sql
-- Create table and insert initial data
CREATE TABLE ai_alter_set_test (
    id INT AUTO_INCREMENT PRIMARY KEY,
    data VARCHAR(50)
);
INSERT INTO ai_alter_set_test (data) VALUES ('row1'), ('row2');

-- Attempt to set AUTO_INCREMENT starting value
ALTER TABLE ai_alter_set_test AUTO_INCREMENT = 1000;

-- Expected: Next INSERT should use id = 1000
INSERT INTO ai_alter_set_test (data) VALUES ('row1000');
```

## Error Message

```
ERROR 20301 (HY000): invalid input: unsupported alter option in inplace mode: auto_increment = 1000
```

## Expected Behavior

In MySQL, `ALTER TABLE ... AUTO_INCREMENT = value` sets the next AUTO_INCREMENT value. After executing `ALTER TABLE ai_alter_set_test AUTO_INCREMENT = 1000;`, the next `INSERT` statement should automatically use `id = 1000` (or higher if the table already contains rows with id >= 1000).

## Impact

- **Functionality**: Users cannot reset or set custom AUTO_INCREMENT starting values after table creation.
- **MySQL Compatibility**: This is a standard MySQL feature that should be supported for compatibility.
- **Use Cases**: Common scenarios include:
  - Resetting AUTO_INCREMENT after data cleanup
  - Setting custom starting values for partitioned or migrated data
  - Adjusting AUTO_INCREMENT to avoid conflicts with existing data

## Test Context

This issue was discovered during comprehensive AUTO_INCREMENT boundary testing (100 test cases). The error occurred consistently in all 10 test cases that attempted to use `ALTER TABLE ... AUTO_INCREMENT = value`.

## Related Test Cases

[...]

PR #26027, opened 2026-07-22
feat(auto-increment): add transactional epoch fence

VioletQwQ-0:
## What type of PR is this?

- [x] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23143

This is the first AUTO_INCREMENT-specific prerequisite split from the closed #25883. It does not expose `ALTER TABLE ... AUTO_INCREMENT=N` yet.

The invariant introduced here is: a CN write carries the AUTO_INCREMENT epoch observed when its values were allocated, and TN accepts the write only when that epoch matches the table epoch. Once a table enters a fenced epoch, legacy writers that cannot provide an epoch fail closed.

| Correctness requirement | Change in this PR |
| --- | --- |
| Preserve the allocator generation across the write path | Persist the epoch in `SchemaExtra` / `plan.TableDef`, propagate it through disttae workspace, row and object writes, flush/compaction, RPC, and TN relations |
| Prevent stale cached values from committing after a reset | Validate the known epoch at TN prepare/commit and require every epoch transition to be exactly `current + 1` |
| Keep the fence effective after restart | Rebuild committed DML watermarks during replay and keep prepared replay transactions fenced until commit or rollback |
| Remain compatible before the feature is enabled | Treat an unknown legacy epoch as zero only while the table epoch is zero; after the first reset, unknown writers are rejected |

Additional guards reject epoch exhaustion before mutation and keep ordinary schema ALTER operations from advancing the AUTO_INCREMENT epoch.

### Explicit non-goals

- no allocator offset/cache reset API;
- no COPY/CLONE allocator reconciliation;
- no user-facing `ALTER TABLE ... AUTO_INCREMENT=N` planner or compile entry point;
- no generic prepared/committing 2PC recovery redesign. The independent TN route/cancellation prerequisite was merged in #25956.

### Diff size

- handwritten code and tests: 23 files, `+1283/-131`;
- generated protobuf code: 2 files, `+1565/-1079`;
- total: 25 files. Proto sources and generated files are committed together.

### Validation

- `go test ./pkg/vm/engine/disttae ./pkg/vm/engine/tae/catalog ./pkg/vm/engine/tae/rpc ./pkg/vm/engine/tae/txn/txnimpl ./pkg/vm/engine/tae/db/test`
- focused `-race` coverage for epoch transitions and replay commit/rollback lifecycle
- protobuf regeneration stability check
- `git diff --check`
- `mo-pr-preflight-review`: PASS at exact head `4122a7add078c79757121a4add2d8771113c7980`
- `mo-self-review`: no open blocker

The PR is intentionally Draft while its full CI runs and while this prerequisite is reviewed independently.


XuPeng-SH (approved):
Approved at exact head a7510bf3aa3a31085f35c5044197b245ee1950da.

Correctness review:
- The dedicated epoch is persisted independently of ordinary schema version and TN enforces monotonic current+1 transitions with overflow rejection.
- CN propagation is closed across in-memory writes, object writes, deletes, soft deletes, transfer, workspace flush, and CN-object compaction. Flush grouping includes table+epoch+known, so generations cannot be coalesced accidentally.
- The live race is closed by TAE prepare ordering: DML prepared first publishes its prepare watermark before a later ALTER validates; ALTER prepared first publishes a committing schema node that later DML waits on and then rejects by epoch.
- Legacy unknown writers map to epoch zero only, so they remain compatible before the first reset and fail closed afterward.
- WAL replay obtains the real dirty-table set from the serialized TxnMemo; replayed prepared transactions fence the table until both commit and rollback terminal paths resolve them.

Unhappy-path review found no blocking leak, hang, stale-fence, or unbounded-growth path. The committed watermark is monotonic; prepared replay registrations have symmetric commit/rollback cleanup; schema waits terminate through the transaction state transition.

Fresh local validation at this exact head passed both normal and race-focused matrices for disttae, catalog, rpc, txnimpl, and TAE db tests. git diff --check also passed. Current GitHub CI is green.

Non-blocking requirements for the follow-up user-facing integration:
1. The allocator reset/cache protocol must guarantee that the epoch attached to a write is the generation that actually allocated its values; merely reading a refreshed relation epoch after allocation would be insufficient.
2. Enabling reset must be gated against old TNs that do not enforce these new protobuf fields.
3. COPY/CLONE reconciliation and end-to-end restart/multi-CN coverage should remain explicit acceptance gates, as stated in #23143.

Process nit: the PR body still says intentionally Draft, while the PR is currently Ready for review.

Issue #26013, opened 2026-07-22
[Bug]: UUID secondary-index predicates return not supported type UUID


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I checked the existing issues and found no issue for UUID secondary-index reads returning `not supported type UUID`.

### Branch / Commit

- Branch: `main`
- Commit: `506875f460ab4e13ef43071274446788358fd199`

### Summary

A normal secondary index can be created on a UUID column, but an equality or range predicate on that column returns an internal error. The same predicates return the expected rows when the secondary index is absent.

The behavior is stable across three clean reproductions.

### Reproduction

```sql
DROP DATABASE IF EXISTS uuid_secondary_index_min;
CREATE DATABASE uuid_secondary_index_min;
USE uuid_secondary_index_min;

CREATE TABLE t(
  id INT PRIMARY KEY,
  u UUID,
  KEY idx_u(u)
);

INSERT INTO t VALUES
  (1,'117a0bd5-2d0d-11ed-940f-000c29847904'),
  (2,'18b21c70-2d0d-11ed-940f-000c29847904'),
  (3,'1b50c129-2dba-11ed-940f-000c29847904');

SELECT id,u
FROM t
WHERE u='18b21c70-2d0d-11ed-940f-000c29847904';
```

Range predicates such as `u >= '18b21c70-2d0d-11ed-940f-000c29847904'` fail in the same way.

### Actual behavior

```text
ERROR 20101 (HY000): internal error: not supported type UUID
```

### Expected behavior

The equality predicate should return row `id=2`. Range predicates should return the matching UUID rows. A secondary index must not change valid UUID predicate semantics.

### Controls

- Full-table reads and `ORDER BY u` succeed with the index present.
- Dropping `idx_u` makes the equality predicate return `id=2` correctly.
- The failure is identical through a CTE because the underlying indexed predicate fails before the CTE boundary.

### White-box observation

Secondary-index lookup extracts indexed tuple values through the `serial_extract` path. UUID is handled by the serialization encoder, but the extraction type switch does not include UUID, and it falls through to `not supported type UUID`.

Ariznawlll:
## 最新 main 验证结果：通过

**测试版本**：MatrixOne `main@ebd05ba52f0d08afb4225e7b4b2e97e8030c057d`

**测试步骤**

1. 创建包含 UUID 列的表并写入多条 UUID 数据。
2. 在 UUID 列上创建普通二级索引。
[...]

PR #26028, opened 2026-07-22
fix(function): support UUID serial extraction

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26013

Normal non-unique secondary indexes serialize the indexed UUID together with the primary key. The tuple encoder and decoders already support UUID, but `serial_extract` did not dispatch `T_uuid` results, so index equality and range plans failed with `not supported type UUID` while table scans remained correct.

This change routes UUID through the existing fixed-width extraction helper. It does not change tuple encoding, index layout, or existing error behavior. The regression covers both the constant-index fast path used by index plans and the row-index deserialization path.

## Test

- `go build -mod=readonly ./pkg/sql/plan/function`
- `go vet -mod=readonly ./pkg/sql/plan/function`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -run '^TestSerialExtractUUID$' ./pkg/sql/plan/function`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=5m ./pkg/sql/plan/function`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -v -count=1 -timeout=180s -run '^TestIndexTableLookupSerialFunc$' ./pkg/sql/plan`


XuPeng-SH (approved):
Reviewed the serial_extract dispatch and the UUID tuple encode/decode closure. T_uuid is a fixed 16-byte value already decoded as types.Uuid by both UnpackNthElement and UnpackWithSchema, so routing it through the existing fixed-result helper is consistent for both constant-index and row-index paths. Existing NULL, malformed tuple, mismatch, and bounds behavior is preserved.

Verified locally: full function package, UUID race test repeated 20 times, tuple decoder parity repeated 20 times, planner dependent test repeated 10 times, plus build and vet. CI is green. LGTM.

PR #26033, opened 2026-07-22
test(tae): force scoped checkpoint before GC replay

### LINKED ISSUE #26031: [Bug]: TestAppendAndGC2 checkpoint wait chases post-fence WAL LSNs

### Branch Name

main

### Actual Behavior

`TestAppendAndGC2` failed in [run 29905103601/job 88874679247](https://github.com/matrixorigin/matrixone/actions/runs/29905103601/job/88874679247?pr=26025) after the 10s checkpoint wait:

```text
target LSN: 201, checkpointed LSN: 203, current LSN: 204, pending LSN count: 1
```

The target captured by `WaitAllCheckpointsFinished` has already been checkpointed (`203 >= 201`), but the helper also re-reads the live global watermark through `AllCheckpointsFinished`. A later WAL entry (`204`) therefore makes the predicate false and moves the finish line while the test is waiting.

This is independent of PR #26025: that PR changes only `pkg/sql/compile`, while the failing test binary has no dependency on that package.

### Expected Behavior

The helper must wait for the checkpoint fence captured at entry, rather than require global WAL quiescence after concurrently-created entries.

### Steps to Reproduce

Run `TestAppendAndGC2` under CI scheduling pressure. The assertion above is the proof state: target completion is true while the live, unrelated watermark has advanced.

### Additional information

This differs from closed #22024 (older same-name UT failure) and #19813 (data race). The regression was introduced when #25953 combined the fixed target condition with the dynamic global-quiescence predicate.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26031

`TestAppendAndGC2` used the generic global-quiescence waiter after its append
workload. That contract is not satisfiable in this test: it sets
`CheckpointCfg.MinCount = 3`, so a post-fence tail of one WAL entry is valid
background state and does not schedule another incremental checkpoint. CI
observed target LSN 201 already covered by checkpointed LSN 203, while current
LSN 204 remained pending.

The previous revision fenced only the numeric LSN but still sampled the live
checkpoint intent. That mixed generations in both directions: a newer pending
intent could extend the wait, while the nil interval between generations could
let the waiter return before the newer ICKP started. The latter happened in
`TestAppendAndGC`: the waiter returned and the following zero-pending
assertion observed 2 while the next ICKP was starting.

This revision leaves the shared waiter and checkpoint runner unchanged. Once
all append workers finish, `TestAppendAndGC2` explicitly calls
`ForceCheckpoint` with `TxnMgr.Now()` under a 10-second context.
`ForceCheckpoint` waits on the checkpoint intent that covers that target TS,
so the test gets a target-scoped completion signal without chasing or sampling
a different generation. The change is test-only and does not affect MatrixOne
runtime behavior.

Validation:

- `TestAppendAndGC|TestAppendAndGC2`: 20 iterations each
- the same two tests under `-race`: 3 iterations each
- full `./pkg/vm/engine/tae/db/test` package
- 16 representative shared-waiter callers, including checkpoint, GC, PITR,
  ISCP, snapshot, and replay tests
- `go vet ./pkg/vm/engine/tae/db/test ./pkg/vm/engine/tae/db/testutil`
- `git diff --check`


XuPeng-SH (changes requested):
Blocking finding — fence the checkpoint intent generation as well (`pkg/vm/engine/tae/db/testutil/funcs.go:323`). The predicate still re-reads the current `GetICKPIntentOnlyForTest()` through `latestCheckpointFinished`. Concrete sequence: the waiter captures T; the checkpoint covering T completes so `GetCheckpointed() >= T`; WAL T+1 arrives; `softScheduleCheckpoint` installs a new `ST_Pending` intent with `UpdateICKPIntent(..., true, false)` while its dirty data is not yet flushed. `checkpointTargetReached` is then true, but `latestCheckpointFinished` is false, so this waiter still chases post-fence work and can time out after 10 seconds even though T is durable. The existing `Test_RunnerStore3` state sequence already demonstrates finished intent -> nil -> newer pending intent. The added test only exercises the numeric helper and cannot catch this path. Please bind the lifecycle check to the checkpoint/intent generation that covered T (or another target-scoped completion signal), and add a deterministic regression with a newer pending intent installed after T completes.

XuPeng-SH (approved):
Deep-reviewed exact head `879e64bf2a962d8b0cfd983cd6d79e8964203a62`. No blocking issues remain.

The previous generation-fencing finding is resolved: after all append workers finish, the test captures a target TS and calls `ForceCheckpoint`; `ForceFlush` is bounded by the caller context, and `ForceICKP` waits on the concrete intent returned for that target, then verifies the finished incremental checkpoint covers the target. It does not re-read or chase a later live intent generation. Success and rollback paths both close the shared intent completion channel.

Verified locally: `TestAppendAndGC2` 20x, `TestAppendAndGC|TestAppendAndGC2` under race 3x, the full `db/test` and checkpoint packages, `go vet`, formatting, and diff/delivery checks. Merge after the in-progress CI finishes green.

Issue #26015, opened 2026-07-22
[Bug]: Unreferenced CTE definitions are eagerly bound and reject otherwise valid queries


Ariznawlll:
### MatrixOne version

`main` at `e6c70b68ca9c808756a61bc848bff03cfc0a50d7`.

### What happened?

MatrixOne eagerly binds every CTE declaration even when the CTE is never referenced by the statement. As a result, an otherwise valid query fails when an unused CTE contains an unresolved column.

MySQL 8.0.45 ignores the unused CTE and returns the result of the referenced CTE. Both databases reject the same unresolved column once that CTE is actually referenced.

The behavior reproduced in three consecutive runs.

### How to reproduce

```sql
WITH unused_cte AS (
    SELECT missing_column FROM mo_catalog.mo_database
), used_cte AS (
    SELECT 7 AS n
)
SELECT * FROM used_cte;
```

### Actual result

```text
ERROR 20301 (HY000): invalid input: column missing_column does not exist
```

### Expected result

The unused CTE should not be bound, and the query should return:

```text
7
```

This matches MySQL 8.0.45. As a control, selecting from `unused_cte` should still report the unresolved column.

### White-box finding

`QueryBuilder.preprocessCte` registers the CTE definitions and then unconditionally calls `bindCte(..., checkOnly=true)` for every declaration before binding the main query. That eager loop resolves columns inside definitions that are never referenced. The later table-resolution path already binds an individual CTE when it is actually referenced.

Relevant code: `pkg/sql/plan/query_builder.go`, `preprocessCte`.

LeftHandCold:
fixed

Ariznawlll:
## 最新 main 验证结果：通过

**测试版本**

- MatrixOne `main@8c9bb66898afa74f8b231e1915d99489ae8a0da7`
- 关联修复：[PR #26035](https://github.com/matrixorigin/matrixone/pull/26035)

**测试 SQL**

```sql
WITH unused_cte AS (
    SELECT missing_column FROM mo_catalog.mo_database
), used_cte AS (
    SELECT 7 AS n
)
SELECT * FROM used_cte;
```

**实际结果**

- 未引用的 `unused_cte` 不再被提前绑定。
- 查询成功返回 `7`。
- 没有再因未使用 CTE 内部的 `missing_column` 拒绝整个查询。
- 同一黑盒场景独立执行 3 轮，3/3 均返回 `7`。

[...]

Issue #26022, opened 2026-07-22
[Bug]: Nested ROLLUP duplicates the grand-total row and loses NULL semantics


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I searched open and closed issues for ROLLUP inside a CTE or derived table, duplicate grand-total rows, and outer `IS NULL` predicates. No exact duplicate was found.

### Branch Name

main

### Commit ID

`50319026a69e45c3c190cb7a09e9c5fbdb7f6857`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- Reproduced with direct SQL and with a prepared outer query.

### Actual Behavior

A direct `GROUP BY ... WITH ROLLUP` query returns one grand-total row, but reading the same ROLLUP result through a CTE returns the grand-total row twice.

Given a table with no NULL grouping keys:

```sql
CREATE TABLE t(id INT PRIMARY KEY, grp INT, amount INT);
INSERT INTO t VALUES
  (1,1,10),(2,1,30),(3,1,20),
  (4,2,40),(5,2,NULL),(6,3,50);
```

The direct query is correct:

```sql
SELECT grp,SUM(amount) total
FROM t
WHERE amount IS NOT NULL
GROUP BY grp WITH ROLLUP;
```

It returns one `NULL | 150` grand-total row.

Wrapping the same query in a CTE returns that row twice:

```sql
WITH rolled AS (
  SELECT grp,SUM(amount) total
  FROM t
  WHERE amount IS NOT NULL
  GROUP BY grp WITH ROLLUP
)
SELECT grp,total FROM rolled;
```

MatrixOne result:

```text
NULL  150
1      60
2      40
3      50
NULL  150
```

There is also inconsistent outer NULL handling for the nested ROLLUP result:

```sql
SELECT COUNT(*)
FROM (
  SELECT grp,SUM(amount) total
  FROM t
  WHERE amount IS NOT NULL
  GROUP BY grp WITH ROLLUP
) r
WHERE grp IS NULL;
```

MatrixOne returns `0`, even though the visible ROLLUP output contains a NULL grouping key.

Both behaviors reproduced 3/3 times.

### Expected Behavior

The nested ROLLUP result should contain exactly the same rows as the direct ROLLUP result. The grand-total row should appear once, and an outer `grp IS NULL` predicate should match it once.

[...]

PR #26035, opened 2026-07-22
fix(plan): bind CTEs lazily in declaration scope

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26015
issue #26022

`preprocessCte` eagerly bound every CTE declaration, then the table-resolution path bound a referenced CTE again. This rejected errors inside unused CTEs and mutated a shared ROLLUP AST twice, expanding a one-column ROLLUP into three aggregate branches with a duplicated grand total.

Removing the eager bind alone would let a CTE incorrectly capture tables introduced by the declaring query block's FROM clause. This change records a declaration-scoped binding context for each CTE: it excludes that query block's later/current FROM bindings while preserving legitimate outer correlation, visible CTE definitions, masking, recursive state, default database, snapshot, rewrite, and view lineage. CTE bodies are then bound lazily from that scope. View dependencies discovered through the detached scope are forwarded to the root query owner.

This PR fixes the duplicate-grand-total root cause of #26022. Its independent synthetic-ROLLUP-key predicate-pushdown bug is intentionally handled in a separate PR.

## Test

- Focused RED/GREEN `TestCTELazyBinding*` planner regressions
- Existing CTE and ROLLUP planner tests
- Focused `-race` run
- `git diff --check`


gouhongshen (approved):
## Codex automated review

One non-blocking planner regression found. No substantive existing PR reviews, issue comments, or inline comments were present. Focused tests were blocked because the checkout lacks required CGo artifacts.

### P2 - Preserve the single-row proof for recursive CTE `LIMIT 1` (pkg/sql/plan/query_builder.go:3353)
This changes `bindLimit` from the CTE consumer context (`ctx`) to `initCtx`. `bindLimit` records a literal `LIMIT 1` only in the context it receives, and the recursive path never propagates `initCtx.hasSingleRow` back to `ctx` (unlike the non-recursive path at line 3233). Consequently a scalar subquery reading a recursive CTE with `LIMIT 1` is flattened as a `SINGLE` join instead of the proven-single-row `LEFT` join, retaining unnecessary cardinality enforcement and the more restrictive execution path. Propagate the proof after binding the limit and add a recursive-CTE scalar-subquery plan regression test.

XuPeng-SH (approved):
Deep review completed for `e4032cc4f742191a562d59d6515870f2b43426ab`: no merge-blocking correctness, compatibility, or lifecycle issue found.

The declaration-scoped context correctly excludes bindings from the declaring query block while retaining legitimate outer correlation and visible CTE state. Repeated lazy binds no longer mutate the shared ROLLUP/CUBE AST, and view dependencies discovered through the detached declaration context are forwarded to the root query owner.

Non-blocking P2 (reproduced from the existing review): a recursive CTE with `LIMIT 1` loses its `hasSingleRow` proof because `bindLimit` now updates `initCtx` without propagating that bit to the consumer context. A review-only base/head plan test emits `LEFT` on the base and `SINGLE` on this head. Results remain correct because `LIMIT 1` caps cardinality, but this retains redundant scalar cardinality enforcement. Please propagate `initCtx.hasSingleRow` and add a plan regression in a follow-up.

Verified:
- full `./pkg/sql/plan` test suite
- focused planner race suite, 10 repetitions
- `./pkg/sql/compile` dependent package tests
- planner build and vet with the deterministic CGo environment
- review-only matrices for repeated ROLLUP + HAVING/window and recursive CTE reuse
- `git diff --check`

aunjgr (approved):
Reviewed independently from declaration scope through lazy binding, recursive/non-recursive CTE visibility, repeated AST binding, and root-owned view dependency tracking. The declaration context correctly excludes bindings introduced later by the declaring query block while retaining outer correlation and eligible prior CTEs. The per-bind ROLLUP/CUBE copy covers the parsed fields mutated by expansion. Focused race test passed: TestCTELazyBinding* in ./pkg/sql/plan. I found no actionable correctness issue.

aunjgr (commented):
Non-blocking follow-up from tracing bindLimit state propagation: changing the recursive CTE's top-level LIMIT binding from ctx to initCtx is correct for declaration-scope name resolution, but a literal LIMIT 1 now sets only initCtx.hasSingleRow. Propagating that proof to the consumer ctx after bindLimit would let scalar consumers retain the LEFT-join fast path instead of a redundant SINGLE cardinality check. This does not change query results and does not alter my approval, but a focused recursive-CTE scalar plan test would preserve the optimization.

Issue #26037, opened 2026-07-22
TAE logtail truncate timestamp has a data race


LeftHandCold:
## Bug

`logtail.Manager.truncated` is read by `GetTruncateTS` without synchronization while the logtail GC cron writes it in `GCByTS`. `types.TS` is a multi-word value, so this is both a Go data race and permits a torn timestamp read.

## Evidence

- CI failure: https://github.com/matrixorigin/matrixone/actions/runs/29905297991/job/88875200373?pr=25867
- Failing test: `TestDataBranchPickRetestCompositePKChunkProbe`
- Race reader: `pkg/vm/engine/tae/rpc/getChangedListFromDirtyTree` -> `Manager.GetTruncateTS`
- Race writer: logtail GC cron -> `Manager.GCByTS`
- The race is reproducible on current `main` with a focused concurrent `GCByTS`/`GetTruncateTS` test under `go test -race`.
- PR #25867 does not modify the logtail manager, debug RPC, or the failing DML test, so it exposed but did not introduce this bug.

## Expected behavior

The truncate timestamp must be published and read as one synchronized snapshot, while preserving monotonic GC updates.

## Proposed fix

Protect the truncate watermark comparison/update and snapshot read with a narrow RW mutex, leaving table truncation outside the critical section. Add the focused race regression test.

PR #26039, opened 2026-07-22
fix(logtail): synchronize truncate timestamp access

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26037

`logtail.Manager.truncated` was read by `GetTruncateTS` while the GC cron wrote it in `GCByTS` without synchronization. Since `types.TS` is a multi-word value, the debug changed-table-list path could observe a torn timestamp and race-enabled tests failed.

This change protects the truncate watermark snapshot and its monotonic compare/update with a narrow RW mutex. Table truncation remains outside the critical section, preserving the existing publish-before-truncate ordering without blocking readers on the table operation. A focused concurrent regression test reproduces the original race and exercises the fixed API pair.

Tested with:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -run '^TestManagerTruncateTSConcurrentAccess$' -count=10 -timeout=3m ./pkg/vm/engine/tae/logtail`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=10m ./pkg/vm/engine/tae/logtail`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=10m ./pkg/vm/engine/tae/rpc`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -run '^TestDataBranchPickRetestCompositePKChunkProbe$' -count=1 -timeout=10m ./pkg/tests/dml`
- `git diff --check`


XuPeng-SH (approved):
Reviewed exact head `1a650855a3bce0a7633abd004e1b4d8b386ecd07`. No blocking findings.

I traced every repository access to `Manager.truncated`: the GC cron is the production writer and the debug changed-table-list RPC is the concurrent reader. The new `RWMutex` makes the snapshot read and monotonic compare/update linearizable. Keeping `TxnTable.TruncateByTimeStamp` outside the critical section preserves the existing publish-before-truncate behavior, avoids holding the watermark lock across table work/logging, and introduces no lock-order, cancellation, cleanup, or lifecycle dependency. `TxnTable` also synchronizes its own structural mutation.

Regression evidence:

- On the base commit, with only the new test transplanted, `go test -race` reliably reports the original `GCByTS` write vs. `GetTruncateTS` read race.
- On this head, the focused race test passes 10 consecutive runs.
- Full `pkg/vm/engine/tae/logtail` and `pkg/vm/engine/tae/rpc` package tests pass under `-race`.
- `TestDataBranchPickRetestCompositePKChunkProbe` passes under `-race`.
- `go vet` passes for both affected/consumer packages; `git diff --check` is clean.
- Trial merge with current `main` (`c9ca2c3c5c`) is conflict-free, and the focused race test passes 10 runs in the merged tree.

The current CI Summary is green. The PR is behind `main`, but I found no correctness issue in the reviewed change.

PR #26041, opened 2026-07-22
fix(plan): preserve rollup filters on synthetic keys

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26022

Each ROLLUP grouping-set branch has a `GroupingFlag` for every output group key. A false flag means the aggregate synthesizes that key (for example the grand-total `NULL`) after reading its input. The filter-pushdown rule previously treated the output key as an ordinary group expression and rewrote an outer `grp IS NULL` predicate to the base table. With no base NULL values, that removed every input row before the aggregate could create the grand-total row.

This change makes AGG predicate pushdown aware of synthetic group-key positions. A predicate referencing any false-flag position remains above that aggregate branch, while predicates that reference only active group keys retain the existing pushdown path.

The duplicate ROLLUP expansion reported in the same issue is addressed separately by #26035.

## Test

- Focused RED/GREEN grouping-flag pushdown regressions
- Existing aggregate pushdown and ROLLUP planner tests
- Focused `-race` run
- `git diff --check`


XuPeng-SH (approved):
Reviewed the predicate-pushdown closure at exact head 771c18be362c9480fce915c8a04d67b4ef910346. Predicates that reference synthetic ROLLUP keys now remain above the corresponding AGG branch, while active-key and no-grouping-flag paths retain pushdown. Multi-key/OR cases and conservative handling of unsafe expression forms are covered. Focused tests, the full pkg/sql/plan package, focused race tests, build/vet, and CI are green. No correctness or unhappy-path blockers found.

Issue #26020, opened 2026-07-22
[Bug]: Prepared MIN/MAX/COUNT/GROUP_CONCAT lose parameter metadata


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I searched open and closed issues for `MIN(?)`, `MAX(?)`, `COUNT(?)`, `GROUP_CONCAT(?)`, aggregate parameter counts, and `Incorrect arguments to EXECUTE`. No exact duplicate was found.

### Branch Name

main

### Commit ID

`50319026a69e45c3c190cb7a09e9c5fbdb7f6857`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- SQL-level prepared statements through the mysql client.

### Actual Behavior

MatrixOne prepares constant aggregate expressions containing a parameter marker, but the prepared statement records zero parameters.

For example:

```sql
PREPARE s FROM 'SELECT MIN(?)';
SET @n = 2;
EXECUTE s USING @n;
```

The correct one-argument execution is rejected:

```text
ERROR 20301 (HY000): invalid input: Incorrect arguments to EXECUTE
```

Executing the same statement without arguments incorrectly passes the argument-count check and reaches an internal nil-vector error:

```sql
EXECUTE s;
```

```text
ERROR 20101 (HY000): internal error: panic runtime error: invalid memory address or nil pointer dereference
...
github.com/matrixorigin/matrixone/pkg/container/vector.(*Vector).Length
```

The same parameter metadata behavior was observed for `MAX(?)`, `COUNT(?)`, and `GROUP_CONCAT(?)`. `MIN(?)` reproduced 3/3 times for both the one-argument rejection and the zero-argument internal error. The session remains usable afterward.

### Expected Behavior

The parameter marker must remain in prepared statement metadata. `EXECUTE s USING @n` should return `2`, while `EXECUTE s` without an argument should return a normal argument-count error.

MySQL 8.0.45 returns `2` for the one-argument `MIN(?)` execution in 3/3 runs.

### Steps to Reproduce

```sql
SET @n = 2;

PREPARE s1 FROM 'SELECT MIN(?)';
EXECUTE s1 USING @n;
EXECUTE s1;

PREPARE s2 FROM 'SELECT COUNT(?)';
EXECUTE s2 USING @n;
EXECUTE s2;
```

### Additional Information

[...]

Issue #26019, opened 2026-07-22
[Bug]: Prepared SUM and AVG reject parameter markers as TEXT


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I searched open and closed issues for prepared aggregate parameter markers, `SUM(?)`, `AVG(?)`, and the observed TEXT type error. No exact duplicate was found.

### Branch Name

main

### Commit ID

`50319026a69e45c3c190cb7a09e9c5fbdb7f6857`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- SQL-level and MySQL binary-protocol prepared statements were checked.

### Actual Behavior

MatrixOne rejects numeric aggregate parameter markers while preparing the statement:

```sql
PREPARE s FROM 'SELECT SUM(?)';
```

Result:

```text
ERROR 20203 (HY000): invalid argument aggregate function sum, bad value [TEXT]
```

`AVG(?)` fails in the same way. The failure occurs during `PREPARE`, before any value can be supplied by `EXECUTE`.

The result reproduced 3/3 times. A binary-protocol prepared recursive CTE with `SUM(n)`, where `n` originates from a parameter marker, reaches the same error.

### Expected Behavior

The statement should be prepared successfully and the parameter type should be resolved from the value supplied at execution time, matching MySQL.

MySQL 8.0.45 returns `2` for each of three runs:

```sql
PREPARE s FROM 'SELECT SUM(?)';
SET @n = 2;
EXECUTE s USING @n;
```

### Steps to Reproduce

```sql
PREPARE s1 FROM 'SELECT SUM(?)';
SET @n = 2;
EXECUTE s1 USING @n;

PREPARE s2 FROM 'SELECT AVG(?)';
EXECUTE s2 USING @n;
```

Control statements with a direct numeric expression do work:

```sql
SELECT SUM(2);
PREPARE c1 FROM 'SELECT SUM(CAST(? AS SIGNED))';
PREPARE c2 FROM 'SELECT SUM(? + 0)';
```

### Additional Information

The parameter binder creates an untyped marker as `T_text` by default. Numeric aggregate overload resolution happens while constructing the prepared plan, before the execution value is available, so `SUM` and `AVG` reject the marker as TEXT.

LeftHandCold:
[...]

Issue #25993, opened 2026-07-22
[Bug]: Prepared window frame rejects parameter markers for PRECEDING and FOLLOWING


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Searched for prepared window-frame parameter markers and `? PRECEDING` / `? FOLLOWING`. No matching issue was found.

### Branch Name

main

### Commit ID

`c0486a2446c84c6dd6bf142494fe85243877ddc5`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- SQL-level `PREPARE` / `EXECUTE` through the mysql client

### Actual Behavior

MatrixOne supports a prepared window query when the `ROWS` frame bounds are integer literals, but rejects the equivalent MySQL-compatible parameterized form during `PREPARE`:

```text
ERROR 1105 (HY000): ... syntax error ... near " ? PRECEDING AND ? FOLLOWING) ..."
```

The literal control returns the expected running-window sums in both MatrixOne and MySQL:

```text
1  30
2  60
3  50
```

MySQL 8.0.45 returns the same result for the parameterized frame. MatrixOne reproduced the prepare-time rejection 3/3 times.

### Expected Behavior

Parameter markers should be accepted for `ROWS` frame bounds inside a prepared statement and bound at `EXECUTE` time, consistent with MySQL.

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS ps_window_param;
CREATE DATABASE ps_window_param;
USE ps_window_param;
CREATE TABLE e(id INT PRIMARY KEY,n INT);
INSERT INTO e VALUES(1,10),(2,20),(3,30);

-- Literal prepared control succeeds.
PREPARE literal_frame FROM
  'SELECT id,SUM(n) OVER (ORDER BY id ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING) FROM e ORDER BY id';
EXECUTE literal_frame;

-- Equivalent parameterized form is rejected by MatrixOne.
PREPARE parameter_frame FROM
  'SELECT id,SUM(n) OVER (ORDER BY id ROWS BETWEEN ? PRECEDING AND ? FOLLOWING) FROM e ORDER BY id';
SET @before=1,@after=1;
EXECUTE parameter_frame USING @before,@after;
```

### Additional Information

[...]

PR #26042, opened 2026-07-22
fix(plan): preserve prepared aggregate parameters

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26020
issue #25992

Prepared aggregate and GROUP BY parameters were broken at two independent planner seams. `VisitPlan` skipped `Node.AggList` and `Node.GroupBy`, so parameter discovery and reset/fill rules never reached markers stored there. In addition, binder expression-equivalence maps used normal SQL formatting, where every marker is rendered as `?`; distinct parser offsets therefore collided and could reuse the wrong bound expression.

This change visits aggregate/group expressions and adds an opt-in offset-aware key for planner-internal expression identity. Default SQL formatting remains unchanged, and the internal offsets are stripped before names enter the user-visible plan/EXPLAIN path.

The aggregate result-type propagation in #26019 and dynamic window-frame parameters in #25993 are separate design changes and are not included here.

## Test

- Focused RED/GREEN regressions for aggregate discovery, GROUP BY identity, nested/repeated markers, and display-name isolation
- Visitor and formatter unit tests
- Full `pkg/sql/parsers/tree` and `pkg/sql/plan` test packages
- Focused `-race` run
- `go build -mod=readonly` and `go vet -mod=readonly` for both changed packages
- `git diff --check`


gouhongshen (changes requested):
## Codex automated review

The aggregate/group plan traversal fix is sound, but the offset-aware key breaks the existing parameterized GROUP BY alias/ordinal path.

### P1 - Keep parameterized GROUP BY alias keys compatible with projection lookup (pkg/sql/plan/window_binder.go:44)
`windowExprAstKey` now returns an offset-bearing `semanticAstKey`, while `GroupBinder.BindExpr` still registers alias/ordinal-reused parameter expressions in `ctx.groupByAst` under plain `tree.String` (group_binder.go:94-105). Thus a legal prepared query such as `SELECT n_nationkey + ? AS x, count(*) FROM nation GROUP BY x` misses the lookup at projection_binder.go:44 and rebinds the projection against the underlying scan column. ONLY_FULL_GROUP_BY then rejects it, or MySQL-compatible mode wraps it in `any_value`, instead of projecting the grouped column. Before this PR, producer and consumer used the same formatted key. Use one parameter-aware key scheme for both sides while retaining distinct identities for independently written markers, and cover alias/ordinal cases containing a real column.

gouhongshen (approved):
## Codex automated review

No findings. The latest update consistently applies the offset-aware key to GROUP BY producers and consumers and covers alias/ordinal reuse, resolving the prior review. Focused tests could not start because local CGo artifacts are absent; static review, formatting, and diff checks were clean.

XuPeng-SH (approved):
Deep review completed on the latest head. No blocking issues found.\n\nVerified correctness: parameter identity is opt-in and does not change default SQL rendering; aggregate/GROUP BY producers and consumers use the same identity key; aliases and ordinals still reuse the grouped expression; internal parameter offsets do not leak into display names; prepared-plan discovery/reset now reaches AggList and GroupBy.\n\nVerified performance/risk: the extra formatting is bind-time only and bounded by AST size; no new goroutines, waits, external resources, or unbounded state were introduced. The latest main merge-tree is conflict-free.\n\nLocal validation passed: build and vet for pkg/sql/parsers/tree and pkg/sql/plan; full tests for both packages; focused new regression tests under -race, 10 runs.

Issue #26014, opened 2026-07-22
[Bug]: Recursive CTE result cannot be joined through two outer aliases


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I checked the existing issues and found no issue for joining two aliases of the completed result of a recursive CTE.

### Branch / Commit

- Branch: `main`
- Commit: `506875f460ab4e13ef43071274446788358fd199`

### Summary

The completed result of a recursive CTE cannot be referenced twice in the outer `FROM` clause, even when each reference has a distinct alias. MatrixOne reports that the CTE table is specified more than once.

This is different from the rule that the recursive member may reference itself only once: both references in this case are outer consumers after recursion has completed.

### Reproduction

```sql
WITH RECURSIVE seq(n) AS (
  SELECT 1
  UNION ALL
  SELECT n+1 FROM seq WHERE n<3
)
SELECT a.n AS left_n,b.n AS right_n
FROM seq a
CROSS JOIN seq b
ORDER BY a.n,b.n;
```

### Actual behavior

```text
ERROR 20301 (HY000): invalid input: table 'seq' specified more than once
```

The result is stable across three clean executions.

### Expected behavior

The outer query should return the 3 x 3 Cartesian product. MySQL 8.0.45 returns 9 rows.

The aggregate form should return:

```sql
WITH RECURSIVE seq(n) AS (
  SELECT 1
  UNION ALL
  SELECT n+1 FROM seq WHERE n<3
)
SELECT COUNT(*) AS pairs,SUM(a.n+b.n) AS checksum
FROM seq a CROSS JOIN seq b;
```

```text
pairs  checksum
9      36
```

### Controls

- A non-recursive CTE with the same two-alias CROSS JOIN returns `9 / 36` in MatrixOne.
- The recursive CTE can be consumed multiple times through independent scalar subqueries; only multiple outer table references fail.

### White-box observation

`bindRecursiveCte` installs the recursive CTE binding in the outer context under the CTE name. When the two table operands are merged, both child contexts still expose the same CTE table name, and `BindContext.mergeContexts` rejects the duplicate before the distinct outer aliases can make the references unambiguous.

LeftHandCold:
fixed

Ariznawlll:
## 最新 main 验证结果：通过

**测试版本**：MatrixOne `main@ebd05ba52f0d08afb4225e7b4b2e97e8030c057d`

**测试步骤**

1. 使用递归 CTE 生成连续数据。
[...]

PR #26044, opened 2026-07-22
fix(plan): rebind recursive CTE consumer aliases

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26014

`bindRecursiveCte` installs the final recursive result under the CTE's internal name before returning its `SINK_SCAN`. An outer explicit alias then appended a second binding with the same tag instead of replacing that internal binding. Each join child therefore still exported the original CTE name, and `mergeContexts` rejected valid self-joins such as `seq AS a JOIN seq AS b` as duplicate `seq` tables.

This change atomically rebinds a fresh recursive `SINK_SCAN` consumer under its explicit table/column aliases across every BindContext index. The old namespace is removed, the new display names are installed, and normal duplicate-alias detection remains unchanged.

## Test

- Recursive CTE self-join, explicit alias, hidden original name, duplicate alias, and column-alias regressions
- Ordinary table/subquery alias controls
- Distributed recursive CTE SQL/result expectation (`pairs=9`, `checksum=36`)
- Focused `-race` run
- `git diff --check`


XuPeng-SH (approved):
Reviewed exact head e7c1500e36dbb5d3c5cbb5e78ccfd66b0c8a46e2. The recursive SINK_SCAN consumer is rebound consistently across bindings, tag/table/column indexes, and the binding tree only after validation; the old CTE namespace is hidden while ordinary duplicate-alias behavior remains intact.

Local verification:
- focused recursive alias tests: -race -count=20
- full ./pkg/sql/plan test suite
- go vet ./pkg/sql/plan
- go build ./pkg/sql/plan

All relevant CI checks are green.

heni02 (approved):

Issue #25991, opened 2026-07-22
[Bug]: PREPARE of EXPLAIN SELECT returns an internal error


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

The existing #22765, #15389, #13885, and #14533 cover explaining an already prepared statement through `EXPLAIN ... EXECUTE`. This issue covers a different SQL shape: preparing an `EXPLAIN SELECT` statement itself.

### Branch Name

main

### Commit ID

`c0486a2446c84c6dd6bf142494fe85243877ddc5`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- SQL-level `PREPARE` / `EXECUTE` through the mysql client

### Actual Behavior

MatrixOne returns an internal error while preparing an `EXPLAIN SELECT`, even when the statement has no parameter markers:

```text
ERROR 20101 (HY000): internal error: statement: 'explain select * from src where id >= 2'
```

The parameterized form returns the same internal error:

```text
ERROR 20101 (HY000): internal error: statement: 'explain select * from src where id >= ?'
```

Direct `EXPLAIN SELECT ...` works, and MySQL 8.0.45 supports both prepared forms. The MatrixOne result reproduced 3/3 times for both literal and parameterized forms.

### Expected Behavior

`PREPARE` should accept `EXPLAIN SELECT`, and `EXECUTE` should return the explain result, consistent with MySQL. At minimum, this path should not return an internal error.

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS ps_explain;
CREATE DATABASE ps_explain;
USE ps_explain;
CREATE TABLE src(id INT PRIMARY KEY, n INT);
INSERT INTO src VALUES (1,10),(2,20),(3,30);

-- Direct control works.
EXPLAIN SELECT * FROM src WHERE id >= 2;

-- Literal prepared EXPLAIN fails at PREPARE in MatrixOne.
PREPARE pe0 FROM 'EXPLAIN SELECT * FROM src WHERE id >= 2';
EXECUTE pe0;
DEALLOCATE PREPARE pe0;

-- Parameterized prepared EXPLAIN also fails at PREPARE.
PREPARE pe1 FROM 'EXPLAIN SELECT * FROM src WHERE id >= ?';
SET @id = 2;
EXECUTE pe1 USING @id;
DEALLOCATE PREPARE pe1;
```

[...]

PR #26046, opened 2026-07-22
fix(frontend): support prepared EXPLAIN statements

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25991

PREPARE previously sent a normal ExplainStmt through the generic planner, which had no matching branch and returned an internal error. Handling only that planner branch is insufficient because EXECUTE must retain EXPLAIN semantics across schema replanning, parameter binding, protocol metadata, and authorization.

This PR:

- plans the inner query for a prepared EXPLAIN while retaining the saved ExplainStmt for execution routing;
- renders the initialized or schema-replanned prepared plan instead of rebuilding the original query, so the underlying SELECT is never executed;
- substitutes both text-protocol and binary-protocol parameters into a copied plan before rendering;
- reports the same single dynamic result column as direct EXPLAIN for COM_STMT_PREPARE;
- authorizes the inner statement rather than treating EXPLAIN as an unprivileged wrapper; and
- adds planner, execution, parameter, and wire-metadata regression tests.

## Testing

- .agents/skills/mo-dev/scripts/mo-cgo-test ./pkg/sql/plan -count=1
- .agents/skills/mo-dev/scripts/mo-cgo-test ./pkg/frontend -count=1
- targeted race tests for the new plan and frontend paths
- go vet -mod=readonly ./pkg/sql/plan ./pkg/frontend


XuPeng-SH (approved):
Reviewed exact head f3f35e6ba58ec4f8dc7f6fb9e4c668bc02bca760. The PREPARE/EXECUTE closure preserves the saved ExplainStmt while rendering the initialized or schema-replanned inner plan; parameter substitution deep-copies the reusable template, binary parameter values are copied before vector cleanup, protocol metadata matches direct EXPLAIN, and authorization is applied to the inner statement. The execution routing returns through the frontend handler without running the underlying query.

Local verification:
- focused planner/frontend regression tests: -race -count=20
- full ./pkg/sql/plan and ./pkg/frontend test suites
- go vet ./pkg/sql/plan ./pkg/frontend
- go build ./pkg/sql/plan ./pkg/frontend

All relevant CI checks are green.

PR #26049, opened 2026-07-22
test: remove obsolete scaffolding and dead converter

XuPeng-SH:
## What

- remove the unused stream SQL converter, its never-consumed field, and its fully disabled test
- delete empty, permanently skipped, comment-only, panic-swallowing, and coverage-only test scaffolding
- replace print-only checks with assertions for index SQL generation, stage paths, parser formatting, CDC JSON decoding, and shuffle panic behavior
- remove large commented-out test implementations and a commented-out moarray helper

## Why

The removed tests executed no behavior or passed regardless of the result. The stream converter was constructed and stored but never read anywhere; the live connector path writes through InternalExecutor.Exec directly.

## Verification

- CGo test wrapper across all touched packages
- go build across all touched packages
- go vet across all touched packages
- git diff --check
- pre-push multi-angle and Q1-Q3 self-review: no open blockers


gouhongshen (approved):
## Codex automated review

One non-blocking test resource-lifecycle issue. Local focused tests could not run because this environment has no Go executable; the PR’s GitHub CI checks are successful.

### P2 - Release pooled Apply arguments created by the new subtests (pkg/sql/colexec/apply/apply_test.go:38)
`NewArgument()` allocates from the `reuse` pool, but none of the three subtests calls `arg.Release()`. This increases the existing test's unreleased pooled objects from one to three. With the supported reuse checker enabled, an object collected while still marked in use causes the pool finalizer to panic; otherwise these allocations still bypass pooling. Use `defer arg.Release()` or construct a plain `Apply` for this String-only test.

XuPeng-SH:
@jiangxinmeng1 please help to review

jiangxinmeng1 (approved):
Overall, this cleanup looks good. The removed SQLConverter and
  KafkaMoConnector.converter have no consumers, and the live connector path still writes
  through InternalExecutor.Exec. The deleted tests were empty, skipped, comment-only, or
  swallowed panics, while the print-only checks are now backed by assertions. CI is
  green.

  One small non-blocking issue in pkg/sql/colexec/apply/apply_test.go: each TestString
  subtest allocates an Apply through NewArgument() but never returns it to the checker-
  enabled reuse pool. Please add defer arg.Release() immediately after allocation, or use
  &Apply{} since this test only exercises String.

  Other than that, LGTM.

gouhongshen (approved):
## Codex automated review

The new discussion only requests another reviewer and provides no technical rebuttal. The dead converter removal and scaffolding cleanup are otherwise safe. Focused local tests could not start because `cgo/libmo.dylib` is absent; head-SHA CI passed. One non-blocking lifecycle issue remains.

### P2 - Release pooled Apply arguments created by TestString (pkg/sql/colexec/apply/apply_test.go:38)
NewArgument() allocates from a reuse pool configured with WithEnableChecker, but each subtest exits without arg.Release(). TestString therefore leaves three Apply objects in use instead of the base test's one; with mo_reuse_enable_checker=true, GC can trigger the pool's missing-free panic, and otherwise the objects are never returned to the pool. Add defer arg.Release() after allocation, or use an unpooled &Apply{} for this String-only test.

PR #26050, opened 2026-07-22
chore(thirdparties): track onnxruntime binaries with git-lfs

fengttt:
## What

Add the onnxruntime shared libraries under `thirdparties/onnxruntime/` and track them with git-lfs:

- `onnxruntime.so` (linux x86_64, 22M)
- `onnxruntime_arm64.so` (linux arm64, 18M)
- `onnxruntime_arm64.dylib` (macOS arm64, 34M)

## How

- `.gitattributes`: LFS tracking scoped to `thirdparties/onnxruntime/*.so` and `*.dylib` so other `.so` files in the repo are unaffected.
- `.gitignore`: the global `*.so` rule was blocking the linux binaries; added negation exceptions for this directory.

Binaries are stored via git-lfs (pointers in git, blobs in LFS).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+thirdparties/onnxruntime/*.so filter=lfs diff=lfs merge=lfs -text
+thirdparties/onnxruntime/*.dylib filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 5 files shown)


fengttt:
Abandoning: introducing git-lfs is a repo-wide policy change that needs maintainer opt-in. Closing.

PR #26053, opened 2026-07-22
feat(function): add onnx_run builtin to evaluate ONNX models from SQL

fengttt:
## What

Adds a builtin scalar function `onnx_run` that evaluates an ONNX model over each row and returns the output as JSON.

```sql
select onnx_run(model, input, input_shape, output_shape) from T;
```

- **model** — `varbinary` model bytes, or a `datalink` to a model file in a stage.
- **input** — JSON flat array for the input tensor.
- **input_shape** / **output_shape** — JSON `'{"dim":[1,1,4],"dtype":"float32"}'`. A **NULL** `output_shape` means the output is not a plain tensor: the model's outputs (sequence/map, e.g. sklearn ZipMap probabilities) are returned as a JSON object keyed by output name.

Example:
```sql
select onnx_run(
    cast('stage://models/mnist.onnx' as datalink),
    cast(json_array(...) as varchar),
    '{"dim":[1,1,28,28],"dtype":"float32"}',
    '{"dim":[1,10],"dtype":"float32"}');
```

## How

- **`pkg/mlai/onnx`** isolates all `onnxruntime_go` usage. It is cgo-free — the onnxruntime shared library is `dlopen`'d at runtime. The environment is initialized lazily and at most once; **if the library cannot be loaded, every `onnx_run` call fails cleanly while the database still starts normally**.
- The onnxruntime shared library (**CPU, v1.26.0**) is fetched per-platform (linux x64/arm64, macOS arm64) by the `thirdparties/Makefile` `onnxruntime` target into `thirdparties/install/lib`, and resolved at runtime via `$MO_ONNXRUNTIME_LIB`, beside `mo-service`, or under `thirdparties/install/lib`. GPU is intentionally out of scope for now.
- The operator (`func_builtin_onnx.go`) caches **one onnxruntime session per expression instance** and reuses it across rows, closing it when the expression is released — the same `newOpWithFree` lifecycle as `serial`/`serial_full`.
- Supported dtypes: `int8/16/32/64`, `uint` variants, `float32/64`, `bool`, and `float16` (via `CustomDataTensor` + IEEE-754 half conversion).

## Tests

- Unit tests in `pkg/mlai/onnx` (shape/dtype parsing, half conversion, and real inference on the checked-in models).
- BVT `test/distributed/cases/function/onnx_run` — passes 13/13 — covering:
  - `sum_and_difference` (both the datalink and varbinary overloads),
  - `non_tensor_outputs` (sklearn random forest, NULL `output_shape`),
  - per-row inference over the checked-in iris dataset (session reuse + `json_array`; predictions match labels),
  - `mnist` and `mnist_float16` (bonus),
  - error paths (bad dtype, shape/length mismatch, NULL model).
- Test models and the iris dataset are checked into `test/distributed/resources/onnx/` (small files; the generating Python scripts are intentionally not included).

## Notes

- CPU-only for now; the `pkg/mlai/onnx.NewSession` seam leaves room for a later GPU (CUDA execution provider) path behind a config flag.
- ONNX model outputs are numerically deterministic for the shipped x86_64 library; a different-architecture CI runner could differ in the last digits of float outputs — worth watching if BVT runs on arm64.


fengttt:
## Review — focus: error handling & efficiency

Overall: sound architecture (lazy fail-safe init, serial-style session lifecycle, clean layer separation). Two real error-handling bugs, two worthwhile efficiency items, and one test-coverage gap that hides bug 2.

### 🔴 High — error handling

**1. Pre-allocated output tensor leaks when `Run` fails — `pkg/mlai/onnx/session.go`**

The cleanup `defer` is registered *after* the error return:

```go
outputs[0] = outTensor                            // C-allocated
if err := s.s.Run(...); err != nil {
    return nil, moerr...                          // <- outputs[0] leaks
}
defer func() { for _, o := range outputs { o.Destroy() } }()  // too late
```

Every failed inference with a declared `output_shape` leaks one C-side tensor. Fix: register the defer immediately after `outputs[0] = outTensor`, before calling `Run`.

**2. Integer input conversion silently swallows errors and truncates — `pkg/mlai/onnx/tensor.go` `convInts`/`convUints`**

```go
v, _ := strconv.ParseInt(string(num), 10, 64)   // error discarded
out[i] = T(v)                                    // unchecked narrowing
```

Two silent-wrong-answer paths for all 8 integer dtypes:
- `[1.5, 2.5]` with `"dtype":"int32"` → ParseInt fails → tensor becomes `[0, 0]`. No error, wrong inference.
- `[300]` with `"dtype":"int8"` → parses as 300, `int8(300)` wraps to `44`. No error, wrong inference.

The float path correctly propagates via `badNumber()`; the int paths should too, with a bitSize per element type.

**3. (Low) Trailing garbage after the input array is ignored — `decodeNumbers`**

`json.Decoder.Decode` stops after the first value, so `'[1,2,3,4]junk'` is accepted.

### 🟡 Medium — efficiency

**4. Triple JSON serialization per row**

`json.Marshal(result)` (session.go) → `bytejson.ParseFromByteSlice` → `bj.Marshal()` (func_builtin_onnx.go). `appendBinaryJSON` accepts only `nil/bool/int64/uint64/float64/json.Number/string/[]any/map[string]any`, so the current `[]any` (holding `float32`, `int8`, …) can't feed `CreateByteJSON` directly — the text round-trip is currently load-bearing. Fix: normalize scalars to `float64`/`int64`/`uint64`, return the value tree from `Run`, and `CreateByteJSON` it directly — one serialization pass instead of three. Meaningful for large outputs.

**5. Per-row constant re-work in the eval loop — `func_builtin_onnx.go`**

- `onnx.ParseShape` runs on every row for what is almost always a constant literal. Cache keyed by raw bytes.
- `ensureSession`'s `bytes.Equal(op.lastArg, rawArg)` is O(model size) per row — trivial for datalink URLs, ~ms per row for a 22 MB varbinary model. Checking `params[0].IsConst()` once short-circuits both the compare and the copy.

**6. (Nit) Model parsed twice per session build — `NewSession`**

`GetInputOutputInfoWithONNXData` builds a temp ORT session, then `NewDynamicAdvancedSessionWithONNXData` builds the real one. One-time per distinct model, acceptable — worth a comment.

### Test coverage

- Good: both overloads, non-tensor outputs, per-row session reuse, float16 both directions, error paths, deterministic BVT.
- **Gap**: no test exercises any integer dtype input — precisely where bug 2 lives.

### Verdict

Fix 1 and 2 before merge (both are silent-failure bugs: a C-memory leak and wrong query results), plus the integer-dtype test. 4/5 are worthwhile in-PR follow-ups.

fengttt:
Pushed 25de48f66f addressing all review findings:

- **1 (leak)**: cleanup defer now registered before `Run` — pre-allocated output tensor is destroyed on the error path.
- **2 (silent int truncation)**: integer parsing uses the declared bit width and propagates errors — `1.5` as int32 and `300` as int8 are now clean errors (new unit + BVT cases).
- **3 (trailing garbage)**: rejected.
- **4 (triple serialization)**: `Session.Run` returns a ByteJson-native value tree (float32 widened via shortest-repr so rendering is unchanged); operator uses `CreateByteJSON` directly — one encoding pass. NaN/Inf still error as before.
- **5 (per-row constant re-work)**: parsed shapes cached across rows; per-row `bytes.Equal`/copy skipped when the model vector is const.
- **6**: double-parse comment added in `NewSession`.

Verification: all unit tests pass (8/8, incl. 3 new); BVT regenerated and passes 15/15; diff of the `.result` shows only the two added error cases — all pre-existing output is byte-identical, confirming the serialization change preserves rendering exactly.

fengttt:
CI status note on the UT failures:

- Run 1 failed on `TestAppendAndGC2` (TAE replay `panic: ExpectedEOB`) and `TestReadValidatesAllRemoteShardsBeforeSending` (shardservice) — both passed on the next run.
- Run 2 failed on `TestStreamServerWithSequenceNotMatch` (`pkg/common/morpc`, hangs 40m after a stream reset) — this is known open bug #26069, pre-existing and tracked.

Each run flakes on a *different* test, all in packages this PR does not touch (this PR touches no files under `pkg/common/`, `pkg/vm/engine/tae/`, or `pkg/shardservice/`). Will rerun the failed job once the current run completes.

fengttt:
Root-caused the recurring UT failures: `TestAppendAndGC2` (2/3 failure rate here) and the shard-read flake were fixed upstream **today** by #26051 (`ff7e6f0d92`, deterministic replay/shard-read setup) and #26063 — this branch predated both. Merged upstream/main (`027535b317`) to pick them up; CI rerunning. The one remaining known flake in the area is #26069 (morpc), which passed on the last two runs.

gouhongshen (changes requested):
## Codex automated review

Two P1 resource-exhaustion paths let SQL input exhaust or crash a CN process.

### P1 - Bound declared output tensor sizes before allocating (pkg/mlai/onnx/tensor.go:81)
`ParseShape`/`NumElements` permits up to 2^62 elements, then this path preallocates based solely on caller-supplied `output_shape` before ORT validates the model output. A valid small model plus `{"dim":[4611686018427387904],"dtype":"float16"}` overflows `2*n` and panics in `make`; smaller declarations can fatal-OOM the process in this or `NewEmptyTensor`. Reject dimensions by a safe byte/result-size limit before any allocation, and add a regression test that oversized shapes return a SQL error rather than panicking.

### P1 - Enforce a size limit before reading datalink models (pkg/sql/plan/function/func_builtin_onnx.go:110)
The datalink overload passes an unbounded stage/file stream to `datalink.GetBytes`, which uses `io.ReadAll`; a user can therefore point `onnx_run` at a multi-GB object and allocate it on the Go heap outside the query mpool before model validation. The analogous `LoadFileDatalink` path stats the object/offset and rejects reads over `types.MaxBlobLen` before calling `GetBytes` (`func_unary.go:4201-4219`). Apply the same pre-read limit here (including the requested offset/size) so the datalink overload cannot bypass the varbinary limit.

heni02 (approved):

aunjgr (changes requested):
Two correctness/reliability issues need to be addressed before merge:\n\n1. [P1] Bound tensor allocations before allocating output buffers.\n\nShape.NumElements only protects arithmetic overflow, not practical memory limits. A small valid model with an attacker-controlled output shape reaches make([]byte, 2*n) or NewEmptyTensor before ONNX validates the model's actual output shape. These allocate directly on the Go heap and a sufficiently large dimension can OOM and terminate the CN. Please compute elementCount × dtypeWidth with checked arithmetic, enforce a process-appropriate byte limit, and preferably account the allocation against query memory. Datalink model loading/session memory should receive equivalent consideration.\n\n2. [P2] Mark datalink-capable onnx_run overloads volatile.\n\nNeither overload sets volatile, so calls with constant arguments may be folded during planning/PREPARE. A datalink refers to mutable external content; replacing the model after PREPARE can therefore leave EXECUTE returning inference from the old model. Because the varbinary overload also recognizes datalink URL schemes, both overloads currently need volatile: true. Alternatively, remove that coercion workaround and mark only the datalink overload volatile.

fengttt:
@aunjgr both findings were valid — addressed in f390906433:

**P1 (allocation bounds)**
- `ParseShape` now rejects declared shapes whose `elements × dtype-width` exceeds **64 MB** (`MaxTensorBytes`, mirroring `types.MaxBlobLen` — a larger result tensor couldn't be returned as a json value anyway). This is the single choke point both input and output shapes pass through, so the pre-allocation in `buildOutputTensor` can no longer be driven past it. The comparison uses the division form to stay overflow-free.
- Datalink model loading now **stats the file before reading** (same pattern as `load_file`) and refuses models over **1 GiB**, instead of `io.ReadAll`-ing an arbitrary stage file.
- Unit tests added for the cap (over-limit shapes rejected incl. an overflow-shaped dim; at-limit accepted).
- Full query-memory accounting for session/tensor allocations is a fair follow-up; the hard caps close the crash vector.

**P2 (volatile)**
Both overloads now set `volatile: true`, matching `load_file`'s precedent — and you're right that the varbinary overload needs it too because of the scheme coercion; kept the coercion (it's what makes `cast(... as datalink)` work when function resolution lands it on overload 0) and documented the linkage in the registration comment.

gouhongshen (changes requested):
## Codex automated review

The prior review findings are addressed, but four blocking resource/supply-chain issues remain.

### P1 - Make inference observe query cancellation (pkg/mlai/onnx/session.go:124)
Run calls the synchronous s.s.Run API without RunOptions and Session.Run receives no context, so proc.Ctx cancellation never reaches an in-flight model. A model containing a very large Loop or otherwise expensive graph continues occupying the executor and ONNX worker threads after KILL, timeout, or client disconnect. Use RunWithOptions and terminate the run when the query context is canceled, with synchronization that cannot race session destruction.

### P1 - Bound peak conversion memory rather than only raw tensor bytes (pkg/mlai/onnx/helpers.go:56)
MaxTensorBytes limits only elementCount*dtypeWidth. An accepted 64 MiB int8/bool tensor contains 67,108,864 elements, after which intsToAny/boolsToAny allocates a []any of that length—at least 1 GiB on 64-bit Go—before ByteJSON adds another large allocation. Input decoding has similar amplification through []json.Number. These allocations are outside the query mpool, so one valid call can still OOM the CN despite the new cap. The limit must account for the complete peak representation or the conversion must stream/encode directly with query-memory accounting.

### P1 - Apply the tensor bound to auto-allocated outputs (pkg/mlai/onnx/session.go:124)
When output_shape is NULL, outputs consists entirely of nil slots. DynamicAdvancedSession.Run therefore asks ONNX Runtime to allocate every model output with no size validation or query-memory accounting; ParseShape/MaxTensorBytes is never involved. A small model can generate an arbitrarily large tensor or sequence and exhaust the CN before valueToJSON can inspect it. NULL output_shape is required for the advertised sequence/map path, so this needs an enforceable runtime/output-memory bound rather than a post-allocation length check.

### P1 - Pin and verify the downloaded ONNX Runtime binary (thirdparties/Makefile:169)
Every clean build downloads a native shared library from the mutable matrixorigin/thirdparty main branch and installs it without checking a digest. The same MatrixOne commit can therefore silently produce different binaries, and a changed or compromised mirror artifact becomes native code loaded into every CN. Pin an immutable source revision and verify an architecture-specific SHA-256 before installing the library.

fengttt:
@gouhongshen response to the second review round (fixes in c0172fdc5b):

**P1 — pin and verify the downloaded runtime: fixed.** The Makefile now downloads from an immutable `matrixorigin/thirdparty` commit (`6553650`) instead of `main`, and verifies a per-platform SHA-256 before installing (download to temp → checksum → move; a tampered hash aborts the build with nothing installed — tested both directions). Same MatrixOne commit ⇒ same native library, byte for byte.

**P1 — apply the tensor bound to auto-allocated outputs: fixed.** `anyTensorFlat` (the NULL-`output_shape` conversion path) now applies the raw-tensor-byte cap to model-produced outputs before any conversion, since those never pass through `ParseShape` (single conservative check at `MaxTensorBytes/8` elements; regression test added). Note the C-side allocation itself happens inside onnxruntime and cannot be pre-bounded through the Go binding — the guard stops the Go-side conversion and returns a clean SQL error.

**P1 — query cancellation for in-flight inference: declined (maintainer decision).** Not necessary for the small, fast CPU models this feature targets; the plumbing (RunOptions/Terminate watcher per row) isn't warranted at this point. Can be revisited if long-running models become a real use case.

**P1 — bound peak conversion memory: declined (maintainer decision).** The raw tensor byte caps (64 MB per tensor, on declared shapes, model outputs, and the 1 GiB datalink model cap) are the intended bound; conversion-layer accounting beyond that is not needed for tensors this size.

gouhongshen (changes requested):
## Codex automated review

History assessment at `c0172fdc5b`: the runtime pin/checksum, volatility, and earlier error-handling findings are fixed. The new auto-output check limits Go-side conversion after inference, but it does not bound ONNX Runtime allocation or intermediate tensors. The author explicitly declined cancellation and conversion-memory changes as maintainer decisions; both remain blocking because the SQL API accepts arbitrary model bytes and the intended small-model usage is not enforced by a permission, model constraint, execution limit, or memory boundary.

### P1 - Enforce a memory limit during ONNX execution (pkg/mlai/onnx/session.go:124)
**Author response:** The new `anyTensorFlat` element check is intended to cap model-produced outputs before conversion; the response also acknowledges that the C-side allocation occurs inside ONNX Runtime and cannot be pre-bounded through the current Go binding.

**Why this remains blocking:** That check runs only after `DynamicAdvancedSession.Run` returns. By then ONNX Runtime has already allocated the output, and the binding has copied automatic tensor data into Go memory. Model intermediates never pass through `anyTensorFlat` at all. An arbitrary uploaded model can therefore use `Expand`, `ConstantOfShape`, or a large intermediate followed by a small reduction to exhaust CN memory before the new check executes. The intended small-model scope is not enforced by the API.

To close this finding, enforce or query-account a memory limit during ORT execution, or introduce an enforced trust/model contract that rejects graphs capable of unbounded allocation before `Run`. A post-run output check alone cannot close the allocation path.

### P1 - Propagate query cancellation into inference (pkg/sql/plan/function/func_builtin_onnx.go:221)
**Author response:** Cancellation plumbing was declined because this feature targets small, fast CPU models and a per-row `RunOptions`/termination watcher was considered unwarranted.

**Why this remains blocking:** The implemented SQL function accepts arbitrary model bytes and does not enforce a complexity, operator, duration, or privilege boundary corresponding to that intended usage. `proc.Ctx` is discarded before synchronous ORT execution, so a model containing a large `Loop` can continue consuming executor and ORT worker resources after query timeout, KILL, or client disconnect.

To close this finding, propagate cancellation through `RunWithOptions`/`RunOptions.Terminate` with lifecycle synchronization, or enforce a concrete model/execution boundary that makes long-running inference unreachable.

### P1 - Bound conversion peak rather than only raw tensor bytes (pkg/mlai/onnx/helpers.go:56)
**Author response:** Conversion-memory work was declined because the 64 MiB raw-tensor cap is the intended bound and was considered sufficient for supported tensor sizes.

**Why this remains blocking:** Raw bytes are not the peak representation. A valid 64 MiB `bool` or `int8` output can contain 67,108,864 elements; converting it to `[]any` allocates roughly 1 GiB for the interface slice alone on a 64-bit CN, before reshape structures, ByteJSON construction, serialization copies, and the final mpool-backed result. The new `MaxTensorBytes/8` check protects only the NULL-`output_shape` path; declared narrow outputs still reach this amplification path.

To close this finding, cap element count according to peak conversion/serialization cost, encode without the `[]any` expansion, or account the complete conversion against query memory.

fengttt:
UT failure on the last run was a data race in `pkg/frontend` (`Session.reset` writing through a shared `*time.Location` at session.go:2017, racing CDC log timestamp formatting) — untouched by this PR; filed as #26122 with full analysis. Branch updated to current main; CI re-running.

XuPeng-SH (changes requested):
Deep review of exact head `325242048126e1f93914bdc243956d69251ba0bf`.

The normal execution/lifecycle paths look solid, and I verified the ONNX package, its race suite, the focused SQL-function race suite, the full SQL-function package, build, and vet with the pinned ORT 1.26 runtime. However, three P1 resource-control paths remain open, so I cannot approve this SQL surface yet.

1. **Bound memory during ORT execution, not after it** (`pkg/mlai/onnx/session.go:124`, `pkg/mlai/onnx/helpers.go:135`). The NULL-shape guard runs only after `DynamicAdvancedSession.Run` returns. In the exact `onnxruntime_go` v1.31.0 dependency, auto-allocated outputs are also copied from ORT memory into Go tensors before `RunWithOptions` returns. A model can therefore allocate an oversized output, or a large intermediate followed by a small output, before this check is reachable. Since arbitrary model bytes are accepted and there is no trust/privilege/operator-complexity boundary, this remains a CN OOM path. Please enforce/account a runtime memory limit, or add an enforced trust/model contract that makes unbounded graphs unreachable.

2. **Propagate query cancellation into inference** (`pkg/sql/plan/function/func_builtin_onnx.go:221`). `proc.Ctx` is not passed to `Session.Run`; a model with a large `Loop` can keep the executor and ORT workers busy after query timeout, KILL, or disconnect. The pinned binding already exposes `RunWithOptions` plus `RunOptions.Terminate`. Please add lifecycle-safe cancellation, or enforce a concrete execution/model boundary that makes long-running inference unreachable.

3. **Cap peak conversion cost for declared narrow outputs** (`pkg/mlai/onnx/output.go:25`, `pkg/mlai/onnx/helpers.go:55`). `ParseShape` permits 64 MiB of `int8`/`bool`, i.e. 67,108,864 elements. `flatData` expands that to `[]any`, whose interface backing array alone is roughly 1 GiB on a 64-bit CN, before nested reshape slices, ByteJSON construction, marshal copying, and the mpool-backed result. The `MaxTensorBytes/8` element guard applies only to NULL `output_shape`, so the declared-output path still allows this amplification. Please cap by complete conversion/serialization peak, stream/encode without `[]any`, or account it against query memory.

The intended “small, fast CPU models” scope is reasonable, but it must be enforced in code before it can serve as the safety boundary for a generally callable SQL function.

gouhongshen (changes requested):
## Codex automated review

The clean merge since c017 adds no authored ONNX delta. Fixed: tensor cleanup, integer/trailing-JSON validation, declared-shape and datalink pre-read bounds, volatility, and runtime pin/checksum. Withdrawn: none. No previously blocking risk is demonstrably non-blocking: the author declined execution-memory, cancellation, and conversion-peak controls, and all remain reachable.

### P1 - ONNX Runtime allocations remain unbounded during Run (pkg/mlai/onnx/session.go:124)
Author response: the post-run anyTensorFlat check was presented as the auto-output bound, while acknowledging that ORT allocates before that check; the intended small, fast CPU-model scope was used to decline further controls.

Why this remains blocking: with NULL output_shape, outputs contains nil slots and line 124 calls DynamicAdvancedSession.Run. The pinned binding first lets ORT allocate output values, then copies non-string tensors into a Go slice before Run returns; only afterwards does this code reach valueToJSON/anyTensorFlat. Thus an Expand/ConstantOfShape output can exhaust native and Go memory before the new check executes. A large intermediate followed by a small result never reaches that check at all. Small arbitrary ONNX models are accepted through the varbinary overload, with no enforced model-complexity, privilege, or execution-memory boundary. Enforce/account a memory limit during ORT execution, or enforce a model trust/validation contract that rules out such allocations.

### P1 - Query cancellation never reaches ONNX execution (pkg/sql/plan/function/func_builtin_onnx.go:221)
Author response: cancellation plumbing was declined because the feature is intended for small, fast CPU models.

Why this remains blocking: line 221 invokes Session.Run without proc.Ctx; Session.Run then invokes the synchronous s.s.Run API. A query timeout, KILL, or disconnect therefore cannot interrupt a valid model containing a large Loop, leaving the executor and ORT workers occupied. The API accepts arbitrary model bytes and enforces no duration, operator, or permission boundary corresponding to the intended usage. The pinned binding exposes RunWithOptions and RunOptions.Terminate. Propagate proc.Ctx cancellation through those options, with synchronization against Session.Close/Reset, or add an enforced execution/model boundary.

### P1 - 64 MiB narrow outputs expand to multi-GiB Go allocations (pkg/mlai/onnx/helpers.go:55)
Author response: the author declined conversion-memory accounting, treating the 64 MiB raw-tensor cap as sufficient.

Why this remains blocking: ParseShape deliberately accepts 67,108,864 int8 elements at the cap. For a declared int8/bool output of that shape, tensorToNested reaches intsToAny/boolsToAny, whose make([]any, len(s)) allocates at least 1 GiB for the interface backing array on this 64-bit target before reshaping, ByteJSON construction, marshaling, and the result vector. These allocations are on the Go heap, not proc.Mp; ByteJSON then allocates its own per-element representation. A small model can Expand a scalar to this output, so the input need not be large. Cap element count by complete conversion/result cost, encode directly with query-memory accounting, or otherwise enforce a peak-memory bound.

fengttt:
@XuPeng-SH @gouhongshen all three P1 resource-control findings are now closed in code (0f073ceb23) — reversing the earlier maintainer declines after your reviews made clear the small-model scope must be enforced, not assumed:

**1. Memory bound during ORT execution.** Initialization registers a shared CPU allocator with an arena capped at **1 GiB** (`NewArenaCfg(MaxRuntimeMemoryBytes, kSameAsRequested, …)` + `CreateAndRegisterAllocator`), and every session opts in via `session.use_env_allocators`. Allocations made *inside* onnxruntime while a model runs — outputs and intermediates, including `Expand`/`ConstantOfShape` patterns — now fail the `Run` with an allocator error instead of exhausting the CN. This bounds the allocation path itself, not just post-run conversion.

**2. Query cancellation reaches inference.** `Session.Run` now takes a context and the operator passes `proc.Ctx`. Runs go through `RunWithOptions`; a watcher goroutine raises `RunOptions.Terminate` on cancellation. Lifecycle-safe: the watcher touches only the `RunOptions` and is joined before they're destroyed and before `Run` returns, so it cannot race `Close`/`Reset`. A cancelled query deterministically reports cancellation even when the run finishes first. Verified with a dedicated test under `-race` (×3).

**3. Peak conversion cost bounded.** `MaxTensorElements` (8M) now caps element count in **both** `ParseShape` (declared shapes — narrow int8/bool included) and `anyTensorFlat` (model-produced outputs), bounding the boxed `[]any` peak to ~128–256 MB. Nothing real is lost: a 67M-element result could never fit the 64 MB json blob limit.

New/updated tests: cancellation (incl. race), element-cap boundaries on both paths, and the whole suite now runs every session through the capped env allocator. All 16 package tests + operator tests green.

XuPeng-SH (changes requested):
[P1] The new element limit is per tensor, so it still does not bound one onnx_run invocation. In the NULL output_shape path, Session.Run retains every converted named output in obj, and valueToJSON does the same recursively for sequences/maps. A model with 16 int8 outputs of 8M elements each is individually legal and uses only about 128 MiB of the 1 GiB ORT arena, but intsToAny alone allocates at least 16 x 8M x 16 bytes = 2 GiB of interface storage, before boxed values and ByteJSON. One caller-supplied model can therefore still OOM a production CN.

There is an earlier unbounded Go peak too: onnxruntime_go DynamicAdvancedSession.RunWithOptions calls createGoValueFromOrtValue for every auto output before valueToJSON/anyTensorFlat, and that copies each C tensor into a Go slice. The later MaxTensorElements check cannot protect that automatic-copy peak.

Please enforce a shared aggregate per-inference output/conversion budget across all named outputs and recursively across sequence/map contents, with raw auto-output copying bounded before it is materialized in Go. Add a multi-output or sequence regression test where every tensor is individually below the per-tensor limit but the aggregate exceeds the invocation limit.

The allocator and cancellation fixes themselves are sound in the exercised paths; ONNX runtime tests and -race pass locally.

gouhongshen (changes requested):
## Codex automated review

The shared 1 GiB ORT arena fixes the prior native-execution allocation blocker, and the joined RunOptions termination path fixes in-flight cancellation. The new 8M cap fixes the earlier single-tensor conversion case, but aggregate conversion across multiple outputs or sequence elements remains blocking. No prior blocker was merely downgraded or accepted as non-blocking.

### P1 - Enforce the conversion budget across the complete result (pkg/mlai/onnx/helpers.go:134)
Author response: The author added the 8M MaxTensorElements check here and to declared shapes, arguing that it bounds boxed conversion to roughly 128–256 MB.

Why this remains blocking: The limit applies independently to each tensor. The NULL-output path converts every output and retains every resulting []any in session.go's obj map; sequences likewise retain every converted child recursively. A valid single-input model with nine int8 outputs of 8,388,608 elements passes all checks and has only 72 MiB of tensor payload, but those nine []any backing arrays alone retain 1.125 GiB on a 64-bit CN, before Go tensor copies, scalar boxing, ByteJSON construction, and Marshal's additional copy. Output count and sequence length have no aggregate bound, and the ORT arena does not account for these Go allocations, so a user-supplied model can still OOM the CN. This is distinct from the earlier single-tensor-size finding: that case is closed, while the attempted fix exposes missing whole-result accounting. Close this by threading a shared element/byte budget through all top-level and recursive outputs, including serialization, or by charging the complete output pipeline to query memory.

fengttt:
@XuPeng-SH @gouhongshen the aggregate-conversion finding is correct — the 8M cap was per tensor, and the NULL-`output_shape` path retained every named output's (and sequence child's) `[]any` with no whole-invocation bound. Fixed in f5e0218404:

**Shared per-invocation budget.** `Session.Run` now creates one `resultBudget` (8M elements — a larger total couldn't fit the 64 MB json result anyway) and threads it through the entire conversion: every tensor charges its element count, and every value node charges at least one unit, recursively through sequences and maps. Your 16×8M scenario now fails cleanly on the second output instead of allocating ~2 GiB. The declared-`output_shape` path is a single ParseShape-capped tensor, where per-tensor and aggregate coincide.

**Regression tests added** exactly as requested: (a) multiple outputs individually under the per-tensor limit whose aggregate exceeds the budget, (b) a sequence whose members are jointly over budget, and (c) a node-count bound for long sequences of tiny values.

**On the raw auto-output copies** (`createGoValueFromOrtValue` copying C→Go before any of our checks): agreed those precede the budget check, but they are byte-for-byte mirrors of output tensors that must **coexist in the 1 GiB ORT arena** — the same capped allocator every session uses — so their aggregate is transitively bounded by the arena cap. A tighter pre-copy bound isn't reachable through the binding (the copy happens inside `RunWithOptions`), and the arena already guarantees the ceiling; happy to document this invariant in the code if you'd like it explicit.

gouhongshen (changes requested):
## Codex automated review

The shared budget fixes the previously blocking aggregate `[]any` conversion across outputs and recursive results. Earlier allocation, cancellation, validation, volatility, and runtime-integrity blockers also remain fixed; none were withdrawn or accepted as non-blocking. Package tests, focused race tests, and vet pass. One distinct pre-budget sequence-materialization blocker remains.

### P1 - Apply the node budget before sequence outputs are materialized (pkg/mlai/onnx/session.go:189)
Author response: automatic C-to-Go output copies were described as byte-for-byte tensor mirrors transitively bounded by the 1 GiB ORT arena. Why this remains blocking: `resultBudget` is created only after `runWithCancel` returns. In the pinned binding, `RunWithOptions` first converts an automatic sequence by allocating `[]Value` for its full length and recursively creating a Go tensor, shape, slice, and Go-backed OrtValue for every child. A caller-controlled Loop can build a sequence containing millions of empty or tiny tensors; their payload remains small, but per-value metadata and Go objects are not byte-for-byte payload mirrors or bounded by the tensor arena. The clean budget error is therefore reached only after the dangerous allocation peak. The new test constructs an already-materialized four-element sequence, so it does not exercise this path. This is distinct from the earlier aggregate payload-copy finding: that retained `[]any` issue is fixed, while this failure is object-count amplification inside the binding before conversion begins. Enforce the sequence/value count before recursive binding materialization, use budget-aware/lazy conversion, or enforce a model/output contract making such sequences unreachable; add a model-backed long sequence regression.

fengttt:
@gouhongshen the pre-budget sequence-materialization finding is **correct, and my earlier arena argument was incomplete**: I verified `createSequenceFromOrtValue` in the pinned binding eagerly allocates `[]Value` for the sequence's full length and recursively builds Go tensor/shape/OrtValue objects per child *inside* `RunWithOptions` — per-value **metadata** is not arena-bounded payload, and `resultBudget` runs too late. I also checked the `IoBinding` route: `GetBoundOutputValues` converts through the same path, so no API through this binding avoids the eager conversion.

Fixed in d89e49d996 via the enforced model contract your review offered as the alternative closure:

**Sequence/map-output models may not contain `Loop`/`Scan`.** Sequences of unbounded length can only be produced at *runtime* by those operators; without them a sequence is built by a static node chain, so its length is bounded by the graph itself (already bounded by the model-size caps). `NewSession` now detects the ops in the raw model bytes via the NodeProto `op_type` protobuf encoding (`\x22\x04Loop` / `\x22\x04Scan` — tag-qualified, so node *names* don't match; a residual false positive fails conservatively with a clean "not supported" error). Tensor-only-output models are unaffected — their eager conversion peak is payload bytes, which the capped arena does bound, and per-output overhead is fixed by the model signature.

**Model-backed regression as requested** (`TestSequenceModelLoopRejected`): the real sklearn sequence-output model (Loop-free) keeps loading; the *same model bytes* with an appended Loop/Scan op marker — still parsed fine by ORT as an unknown field, verified empirically — are rejected at session load; a tensor-only model containing the marker is not subject to the contract.

Separately: the Compose(PROXY) BVT and Coverage failures on the previous run were a Docker Hub registry timeout (`dial tcp …:443: i/o timeout` pulling `matrixorigin/ubuntu:22.04` — the image build never started, and the coverage merge cascaded from the missing producer). Rerunning with this push.

gouhongshen (changes requested):
## Codex automated review

Earlier allocation bounds, volatility, cleanup/input validation, runtime allocator, cancellation, runtime pinning, and aggregate conversion-budget blockers are fixed. The new Loop/Scan gate does not close the remaining eager sequence-materialization risk: the advertised ZipMap model produces input-sized sequences without either operator.

### P1 - Loop/Scan-only gate leaves ZipMap sequence materialization unbounded (pkg/mlai/onnx/session.go:59)
Author response: d89e49d rejects sequence/map-output models containing Loop/Scan, assuming other sequences are statically bounded. Why this remains blocking: the checked-in sklearn model's ZipMap output is a sequence of maps whose length equals the caller's batch size, yet the model contains no Loop/Scan. ORT returns sequence lengths 1,000 and 100,000 for those batch sizes. ParseShape accepts an input shape of [2,000,000,4] (8M elements), so SQL can request millions of sequence members. The pinned binding eagerly allocates the complete []Value and recursively materializes every map and tensor inside RunWithOptions before Session.Run creates resultBudget at line 220; the 1 GiB ORT arena does not bound these Go metadata/copy allocations. This can exhaust the CN before the clean result-size error is reached. Reject input-sized sequence/map outputs or enforce a value-count limit before binding materialization, and add a large-batch regression test.

fengttt:
@gouhongshen correct again — the Loop/Scan gate's premise ("sequence length is bounded by the static graph") fails exactly for ZipMap: a *single* node emitting one member per input row, so the length is input-sized, and `ParseShape`'s 8M-element admission let SQL request millions of eagerly-materialized members. Fixed in d27bd2e5fd using the pre-materialization handle the binding does give us — the input shape, known *before* `Run`:

**Input-element bound for sequence/map-output models.** `Session` records `hasSeqOutput` at load; `Run` rejects inputs over `MaxSequenceInputElements` (**64K**) for such models. Since members ≤ rows ≤ elements, this bounds the eager-conversion peak to tens of MB (≈1 KB/member worst case), while 64K elements is far beyond classical-ML batch sizes (the sklearn BVT uses 24). The check runs **first** in `Run` — before input parsing or any ORT call — so an oversized batch fails fast with a clean error naming the limit and the reason. The Loop/Scan gate remains for the runtime-amplification case an input bound cannot cover; together: no-Loop/Scan ⇒ members ≤ rows, and rows ≤ 64K ⇒ bounded materialization.

**Large-batch regression as requested** (`TestSequenceModelLargeBatchRejected`): the real sklearn ZipMap model with a declared `[20000,4]` input (80K elements) fails with the limit error before anything is materialized (the input json is never even parsed); an in-limit batch still runs end-to-end; tensor-only models are exempt from the bound.

gouhongshen (changes requested):
## Codex automated review

The new 64K input gate fixes the prior ZipMap/input-sized sequence blocker. Earlier allocation, cancellation, validation, volatility, cleanup, runtime-integrity, and aggregate-conversion blockers remain fixed. However, the claimed Loop/Scan contract is still insufficient for static sequence construction, so eager sequence materialization remains blocking.

### P1 - Static SequenceInsert chains bypass the pre-materialization bound (pkg/mlai/onnx/session.go:205)
Author response: the Loop/Scan rejection plus the new 64K input-element guard is said to bound all sequence/map outputs before the binding eagerly materializes them. Why this remains blocking: the input guard bounds only input-driven members. ONNX permits `SequenceEmpty` followed by an arbitrary static chain of `SequenceInsert` nodes; each inserts a tensor into the prior sequence, requires neither Loop nor Scan, and can reuse a single scalar input. Such a model is admitted with a one-element input, so line 205 passes, yet its output length is proportional to graph-node count, not input size. The pinned binding then allocates `make([]Value, length)` and recursively materializes every child inside `RunWithOptions` before `resultBudget` exists. The stated model-size cap is not a safe value-count/memory bound: even a 64MB model can encode a very large chain and trigger millions of Go/C objects. Add an enforced pre-Run sequence/value-count or graph-complexity bound that covers static sequence construction (or reject sequence/map outputs), and a model-backed regression for a Loop-free SequenceEmpty/SequenceInsert chain.

XuPeng-SH (changes requested):
Deep review of exact head `d27bd2e5fd6dd7104beaee17d7d0b803a923a185`.

The earlier cleanup, input validation, native Run allocator, cancellation, pin/checksum, and post-Run aggregate-budget fixes are sound in the exercised paths. Focused `-race` tests for cancellation, ZipMap large-input rejection, and operator lifecycle pass. Three production resource-control paths remain blocking:

1. **[P1] Metadata discovery creates an uncapped temporary ORT session.**

   `NewSession` calls `GetInputOutputInfoWithONNXData` before constructing the capped `SessionOptions`. In pinned `onnxruntime_go` v1.31.0 that helper calls `createCSession(data, nil)`. Model parsing, graph optimization, constant folding/prepacking, etc. therefore run outside `session.use_env_allocators`; the later real-session allocator does not cover this peak. The datalink path additionally admits and `io.ReadAll`s up to 1 GiB before this call. Parse ModelProto metadata without a session, extend/fork the binding to accept the capped options, or otherwise create only one capped session; add a session-initialization allocation regression.

2. **[P1] The pre-Run sequence contract still admits graph-driven long sequences.**

   The 64K input cap only bounds input-driven ZipMap members. A legal small-input model can construct an arbitrarily long sequence with `SequenceEmpty` plus a static `SequenceInsert` chain, or `SequenceConstruct`, without Loop/Scan. It passes both guards. The binding then `make([]Value, length)` and recursively materializes every child inside `RunWithOptions`, before `resultBudget` exists. Model byte size is not a safe object-count bound. Enforce a parsed graph/value-count contract (including subgraphs/functions), add budget-aware/lazy binding materialization, or reject sequence/map outputs until a sound pre-materialization bound exists. Add a real static-sequence model regression.

3. **[P1] Accepted model/result limits still permit unaccounted multi-hundred-MiB Go peaks.**

   At the accepted 8M-element limit, `[]any` alone is 128 MiB. MatrixOne ByteJSON adds about 104 MiB (`5-byte entry + 8-byte numeric payload` per element), `Marshal` copies it again, and `AppendBytes` copies it into the result vector again: over 440 MiB before source tensors/boxing. `AppendBytes` does not enforce `types.MaxBlobLen`, so the comment that larger results cannot fit 64 MiB is not a guard. These Go allocations, plus the up-to-1-GiB datalink `io.ReadAll`, bypass `proc.Mp`; concurrent calls can OOM a CN. Enforce a byte/peak budget before conversion, encode once directly into query-accounted memory, and lower/account model loading.

Please address these as one resource-boundary design, then rebase onto latest `main`.

fengttt:
@XuPeng-SH all three findings addressed as one resource-boundary design (f8a1f072dc), plus merged latest main. The unified contract:

> **model ≤ 64 MB → session construction bounded; no unbounded-sequence ops → members ≤ input rows; input ≤ 64K elements → materialization bounded; results ≤ 1M elements → conversion pipeline bounded; encoded result ≤ blob cap.**

**1. Uncapped discovery session → bounded by the model cap.** Rather than eliminating the throwaway session (the binding's bytes-based helper hardcodes nil options, and the options-taking variant is path-only), the design bounds it: `NewSession` enforces **`MaxModelBytes` (64 MB)** at the package boundary, and the datalink pre-read cap is lowered **1 GiB → 64 MB**, matching the varbinary overload's inherent blob cap. Discovery parsing and graph-optimization transients are proportional to model size, so the cap bounds both — a deliberate simplicity-over-machinery trade (a session-free ModelProto metadata parser was considered and set aside as not worth the surface).

**2. Static sequence builders → rejected.** You're right that `SequenceEmpty` + a `SequenceInsert` chain (or `SequenceConstruct`) evades both prior guards. Those three ops are now rejected for sequence/map-output models alongside Loop/Scan; the flat byte scan sees subgraph and local-function nodes too. What remains allowed — `ZipMap`, `SplitToSequence` — emits at most one member per input row/element, which `MaxSequenceInputElements` bounds; verified the sklearn model uses only ZipMap. Regressions cover all three new ops via the marker-append technique.

**3. Result pipeline peak → 1M elements + explicit blob guard.** Conceded that the 8M budget permitted a >400 MB unaccounted pipeline peak ([]any + ByteJson + Marshal + vector copy) and that "can't fit 64 MB anyway" wasn't enforced anywhere. `MaxTensorElements`/`resultBudget` is now **1M** (pipeline peak ~50–80 MB per invocation; a 1M-element per-row json is already beyond any client's use), and the operator now explicitly rejects encoded results over `types.MaxBlobLen` before `AppendBytes` — covering wide-map-key width the element count can't see. Full mpool accounting of the conversion pipeline remains the acknowledged gap; the caps bound the un-accounted peak to double-digit MB.

All package tests green including `-race` ×2; boundary regressions updated to the new limits; branch is on latest main.

gouhongshen (changes requested):
## Codex automated review

Fixed: cancellation, runtime allocation, aggregate result budgeting, and encoded-result limits. Accepted as non-blocking: the now-bounded 1M-element conversion peak. Still blocking: the prior metadata-discovery concern remains unresolved by a model-byte cap alone, but is not repeated because XuPeng-SH already raised it. The previous static-sequence blocker also remains because the new byte scanner is bypassable.

### P1 - Parse NodeProto instead of scanning one protobuf encoding (pkg/mlai/onnx/session.go:154)
Author response: f8a1f072 adds Loop/Scan/Sequence* markers to `containsUnboundedSeqOp`, claiming this rejects graph-driven sequence growth, including subgraphs and local functions. Why this remains blocking: the check recognizes only the exact canonical bytes `0x22, len(op), op`. Protobuf supports larger non-minimal wire encodings that decode to the same message—for example, field-4 tag `0x22` can be encoded as the equivalent varint `0xa2 0x00`; the official documentation confirms that non-optimal varints decode identically ([Go size semantics](https://protobuf.dev/reference/go/size/)). An arbitrary uploaded ONNX model can therefore encode a SequenceInsert/Loop node without the searched marker, pass this gate with a one-element input, and recreate the previously reported eager materialization/OOM path. The regression merely appends canonical marker bytes to another model; it does not contain a real Sequence* node or test alternate legal encodings. Close this by parsing ModelProto and recursively inspecting actual NodeProto `op_type` fields, or by enforcing a different contract that makes sequence/map materialization bounded independently of serialized byte representation.

PR #26055, opened 2026-07-23
fix: prevent nested local shuffle deadlocks

### LINKED ISSUE #25922: [Bug] TPCH Q16 hangs indefinitely: Shuffle execution engine deadlock in waitBatchOrEnd (TPCH 100G stuck for 15+ hours)

XuPeng-SH:
## What type of PR is this?

- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25922

The unified Shuffle implementation added bounded backpressure to shared local Shuffle pools. Per-bucket credits fixed same-level hot-bucket starvation, but a finite fixed-bucket pool can still deadlock when more than one local Shuffle is duplicated into the same pull-based worker tree.

The observed 16-worker wait-for cycle was:

1. an outer Shuffle writer filled its target bucket and waited for that bucket's worker;
2. the target worker was blocked in an inner Shuffle waiting for every inner writer to finish;
3. the missing inner writer was the child of the blocked outer writer, so it could only finish if the outer writer resumed pulling.

This change prevents the cyclic topology instead of increasing or removing the memory bound:

- keep the first fixed-bucket local Shuffle fast path;
- detect a fixed-bucket Shuffle already present in a packed operator tree;
- materialize a second join/group Shuffle through the existing producer scopes, bounded dispatch channels, and consumer scopes;
- preserve local Shuffle reuse when no second probe Shuffle is introduced;
- cancel an active request before connection cleanup waits for migration/reset or transaction rollback, so cleanup is independent of a stuck data path.

The fallback remains bounded: its existing dispatch channel capacity is 32, and this PR does not change Shuffle ready credits or introduce an unbounded queue. It also does not introduce a new hard query timeout.

## Regression coverage

- 16-way nested local shuffle join, on both probe and build chains;
- 16-way nested local shuffle group;
- fixed Shuffle below a non-root operator to cover recursive detection;
- single-layer local group and reusable local join fast paths remain local;
- connection cleanup cancels before a blocked lifecycle operation and before rollback.

The nested-join regression failed on `origin/main` by producing one packed local scope and passes with this change by producing 16 independent bucket scopes.

## Validation

- `go test -race -count=1 -timeout=600s ./pkg/sql/compile ./pkg/frontend`
- focused new concurrency/topology tests under `-race -count=20`
- `go test -v -count=1 -timeout=600s ./pkg/sql/compile ./pkg/frontend`
- `go test -count=1 ./pkg/sql/colexec/shuffle ./pkg/sql/colexec/dispatch ./pkg/sql/colexec/merge ./pkg/sql/colexec/group`
- `go build ./pkg/sql/compile ./pkg/frontend`
- `go vet ./pkg/sql/compile ./pkg/frontend`
- `git diff --check`


XuPeng-SH:
Deep-review follow-up pushed in 02bc01ab6 and rebased by merge on the latest main in 1992a7172.

Correctness:
- close now seals lifecycle admission, cancels the active request and lifecycle operation, then waits;
- migration, migrate-from, and reset propagate RPC/lifecycle cancellation;
- reset failure restores protocol state and releases the speculative session;
- a server-wide liveness monitor detects a disconnected client during a long synchronous Handler call without consuming protocol bytes, including TLS and unread-buffer cases.

Performance:
- one ticker per MOServer, every 5 seconds;
- only requests active for at least 30 seconds are probed;
- 10,000-connection scan benchmark: 57.6-58.4 us and 0 allocs when idle; about 60.6-61.5 us and 4.4 KB with 1 percent long-running requests.

Risk controls:
- no per-query goroutine or unbounded queue;
- socket probing is nonblocking MSG_PEEK plus poll, with Linux POLLRDHUP support;
- cleanup ownership stays in Conn.Close and Routine.cleanup, while the monitor only signals cancellation;
- closing routines are excluded from later scans to avoid repeated probes and log amplification;
- unsupported platforms fail open and retain existing behavior.

Post-main-merge validation:
- full pkg/frontend tests;
- full pkg/sql/compile tests;
- targeted pkg/cnservice migration/reset handler tests;
- targeted frontend race tests;
- Linux socket-probe cross-compilation;
- git diff --check.

gouhongshen (approved):
## Codex automated review

No findings.

aptend (changes requested):
## Change requested

**[P1] Propagate reset cancellation to the actual transaction rollback.**

`Session.reset` derives a cancelable `rollbackCtx` from the lifecycle/RPC context and passes it as `ExecCtx.reqCtx`, but `TxnHandler.rollbackUnsafe` constructs the context passed to `txnOp.Rollback` from `th.txnCtx` instead. Consequently, cancellation from `migrateController.startClose` does not interrupt a reset already blocked in rollback.

Concrete failure path:

1. `ResetSession` obtains lifecycle admission and enters `TxnHandler.Rollback`.
2. The connection closes; `startClose` cancels the lifecycle operation.
3. The rollback continues because its real context is derived from `th.txnCtx`.
4. `mc.waitAndClose()` remains blocked until rollback returns or the default five-minute commit/rollback timeout expires.

I confirmed this with a deterministic blocking `TxnOperator.Rollback` reproducer under `-race -count=3`: canceling the reset operation did not release rollback; canceling the routine/transaction context did.

Please make the context passed to the actual rollback inherit the supplied operation context (or add a dedicated context-aware rollback path), and add a regression test where connection close promptly releases an in-flight reset rollback.

XuPeng-SH:
Addressed the requested change in 0b1a36a15f.

The reset path now uses a dedicated context-aware rollback entry point, so migrateController.startClose cancellation reaches the actual TxnOperator.Rollback call. Normal SQL rollback intentionally continues to inherit the transaction context; request cancellation must not skip ordinary transaction cleanup.

Added TestRoutineCloseCancelsResetRollback, which deterministically blocks inside TxnOperator.Rollback, closes the routine, and verifies both reset and waitAndClose terminate promptly.

Fresh validation after the change:
- focused regression: count=10
- focused race regression: count=10
- full pkg/frontend tests
- full pkg/frontend race tests
- full pkg/sql/compile tests
- targeted pkg/cnservice ResetSession test
- go build and go vet for all three changed production packages
- full PR self-review and git diff --check

LeftHandCold (approved):
Reviewed the complete PR at `0b1a36a15feaa7765d2304fcc3a938ff3398a478`, including the latest reset-rollback cancellation fix. The dedicated context-aware rollback path now breaks the close/reset wait cycle without changing normal statement or cleanup rollback semantics, and the new blocking regression test covers the actual storage rollback context. I found no remaining actionable correctness or regression issue.

aptend (approved):
Independent re-review completed on head `0b1a36a15feaa7765d2304fcc3a938ff3398a478` against base `56735f02fd1e045685e4072289eae99bd8621b5a`.

No merge-blocking findings remain. The review covered the complete diff, shuffle topology closure, frontend lifecycle/reset/close state transitions, socket liveness probing, and Q1-Q3 unhappy paths. Targeted and full owning-package tests, frontend race tests, build, vet, and diff checks passed.

aunjgr (approved):

PR #26061, opened 2026-07-23
fix: complete shuffle writers during reset

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25922

Each parallel shuffle holder must report that it will produce no more rows before releasing its pool ownership. Previously, `Shuffle.Call` reported writer completion only after observing child EOF. If a downstream operator stopped a pipeline earlier, successful `Reset` released the holder without incrementing the pool's stopper count. Other holders could then wait forever for a writer that had already exited.

This change makes writer completion exactly once across normal EOF and successful reset:

- normal EOF and successful cleanup share a guarded `stopWritingOnce` path;
- EOF followed by reset cannot double-count the writer;
- failed cleanup continues to abort the pool and wake waiters;
- the guard is cleared for operator reuse.

Regression tests cover successful early reset, EOF followed by reset, and failed reset.

Validation:

- focused lifecycle tests with `-race -count=20`
- full `pkg/sql/colexec/shuffle` tests with `-race`
- full `pkg/sql/colexec/group` tests
- compile-layer shuffle pool/generation tests
- `git diff --check`

### DIFF
--- a/pkg/sql/colexec/shuffle/shuffle.go
+++ b/pkg/sql/colexec/shuffle/shuffle.go
@@ -66,6 +66,7 @@ func (shuffle *Shuffle) Prepare(proc *process.Process) error {
 		return moerr.NewInternalError(proc.Ctx, "shuffle pool was aborted before prepare completed")
 	}
 	shuffle.ctr.held = true
+	shuffle.ctr.writingStopped = false
 	shuffle.ctr.ending = false
 	shuffle.ctr.runtimeFilterHandled = false
 	return nil
@@ -172,7 +173,7 @@ func (shuffle *Shuffle) Call(proc *process.Process) (vm.CallResult, error) {
 		bat := result.Batch
 		if bat == nil {
 			shuffle.ctr.ending = true
-			shuffle.ctr.shufflePool.stopWriting()
+			shuffle.stopWritingOnce()
 			result.Status = vm.ExecNext
 			result.Batch = batch.EmptyBatch
 			return result, nil
--- a/pkg/sql/colexec/shuffle/shufflepool_test.go
+++ b/pkg/sql/colexec/shuffle/shufflepool_test.go
@@ -341,6 +341,99 @@ func TestShufflePoolPeakIsReportedByExactlyOneHolder(t *testing.T) {
 	require.Equal(t, int64(0), proc.Mp().CurrNB())
 }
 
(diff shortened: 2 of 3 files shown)


XuPeng-SH (changes requested):
The `stopWritingOnce` change is valid as a complementary lifecycle hardening, but it does not provide functional closure for #25922 as claimed.

In the reported hang, the fixed-bucket workers are still blocked inside `Shuffle.Call`: an outer writer is waiting in the bounded publish/flush path, while the target worker is waiting in an inner `waitBatchOrEnd` for every inner writer. The missing inner writer is below the blocked outer writer. Since `Pipeline.Run` has not returned, `Cleanup`/`Shuffle.Reset` is not reachable, so the new successful-Reset hook cannot break that wait-for cycle. It only helps after EOF, early downstream completion, or cancellation has already caused the pipeline to unwind.

Please keep this change as the secondary Reset/EOF exactly-once defense, but add or retain the topology/cancellation fix that prevents or breaks the nested fixed-bucket shuffle cycle (for example, the materialization and cleanup-cancellation changes in #26055), with a regression that reproduces the nested join/group topology. Otherwise this PR should not claim to fix/close #25922.

gouhongshen (changes requested):
## Codex automated review

The Reset/EOF exactly-once bookkeeping is correct when cleanup is reachable, but it does not close the nested shuffle deadlock claimed by the PR. PR-head CI passed; focused local race tests could not start because cgo/libmo.dylib is absent.

### P1 - Reset is unreachable in the deadlock this PR claims to fix (pkg/sql/colexec/shuffle/types.go:119)
This stopper is emitted only from Reset, which Scope.Run invokes through deferred Pipeline.Cleanup after Pipeline.Run's vm.Exec loop returns. The existing local join/group compiler paths can recursively duplicate nested fixed-bucket shuffles into the same worker tree. An outer writer can then block on a target bucket's spaceWaiter while that target worker blocks in the inner waitBatchOrEnd waiting for an inner writer below the outer writer. No Shuffle.Call returns, so Cleanup and this line never execute, leaving the cycle intact. The added tests call stopWritingOnce/Reset directly and do not reproduce the nested join/group topology. Keep this exact-once hardening, but add the topology or cancellation fix and a regression covering the nested cycle before claiming #25922 is fixed.

PR #26063, opened 2026-07-23
test(shardservice): make remote compatibility preflight deterministic

### LINKED ISSUE #26045: [Bug]: TestReadValidatesAllRemoteShardsBeforeSending flaky — remote shard read timeout
## Problem

`TestReadValidatesAllRemoteShardsBeforeSending` in `pkg/shardservice/service_read_test.go:270` fails intermittently:

```
Error: "map[1:1]" should have 2 item(s), but has 1
Test elapsed: 60.08s
```

The test reads from 2 remote shards and expects both to respond. One shard fails to respond within the timeout, resulting in only 1 result in the map.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-22 | [PR #25815](https://github.com/matrixorigin/matrixone/actions/runs/29913023702/job/88900565336) | Coverage |

## Root Cause

Race condition or resource contention on CI runners causes one remote shard's read to exceed the 60s timeout. The shard service test creates multiple mock CN nodes and relies on inter-node communication completing within the timeout.

## Suggested Fix

1. Increase the timeout for CI environments
2. Or add retry logic for transient shard read failures
3. Or use synchronization barriers to ensure all shards are ready before the read

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26045

`TestReadValidatesAllRemoteShardsBeforeSending` was added with #25872 to protect a specific ordering contract: when a binary prepared-parameter read targets multiple remote shards, every selected target must pass commit compatibility validation before the first RPC is sent.

The observed failure was initially reported as a remote-read timeout, but no remote read had started: the counting client remained at zero `AsyncSend` calls. The roughly 60-second delay came from the test setup. It created a real three-CN shard topology through asynchronous heartbeat, allocation, rebalance, and read-cache refresh. If allocation started before all CNs were visible, replicas first accumulated on one CN; subsequent balancing could wait for the one-minute CN freeze interval. The test then chose an incompatible target from asynchronously maintained state, while `Read` could observe a different target ordering and return after the first incompatible shard, leaving only one `Adjust` observation.

#26051 improved that setup by waiting for all CNs and separating setup/read contexts. This PR narrows the test further to the contract it owns:

- install two single-replica remote shards directly in the read cache;
- make the first target compatible and the second target incompatible;
- keep the real `service.Read`, replica selection, cluster compatibility lookup, and counting RPC client;
- require both shard adjustments, the exact incompatibility error, and zero `AsyncSend` calls.

This deliberately does not increase timeouts, add retries, or change production behavior. Scheduler, heartbeat, rebalance, and real remote-handler behavior remain covered by their owning tests, including `TestBinaryReadUsesVersionedRemoteHandler`; they are not prerequisites for this preflight-ordering regression. A regression that validates and sends in the same loop will send after the first compatible shard and fail the zero-send assertion.

- focused regression: `-count=100` passed in 4.407s;
- focused race regression: `-race -count=20` passed;
- full `./pkg/shardservice` package passed in 17.895s;
- `go build ./pkg/shardservice` passed;
- `go vet ./pkg/shardservice` passed;
- `git diff --check` passed.

For comparison, current `main` with the asynchronous setup passed 100 repetitions locally but took 109.544s.


XuPeng-SH (approved):
Reviewed the complete test-only diff and the production path it exercises. The deterministic cache setup preserves the intended `service.Read` preflight-ordering contract while removing unrelated heartbeat/rebalance timing. Resource cleanup is safe, the focused race/stress and full package tests pass, and a mutation that sends after the first compatible target is correctly caught by the zero-send assertion. No blocking findings.

Issue #25908, opened 2026-07-20
[Bug]: external table filters panic when a regular column name contains account


Ariznawlll:
### Component

SQL engine / external table file filtering (`pkg/sql/colexec/external` and duplicated planner helper in `pkg/sql/plan/external.go`)

### Description

External-table filters on a normal user column whose name contains the substring `account` are incorrectly classified as file-level virtual-column filters. For example:

```sql
SELECT * FROM ext_table WHERE account_id = 'normal';
```

The filter is evaluated before the external file is opened, against a synthetic batch that only allocates vectors for the exact virtual columns `account` and `__mo_filepath`. The `account_id` vector remains nil, so expression evaluation dereferences a nil `*vector.Vector` and panics.

### Reproduction

On latest `main@dde1e26160066b1adce4c100913f725d59d1a30d`, construct an external-scan node with:

- a physical `VARCHAR` column named `account_id`;
- the appended `__mo_filepath` virtual column;
- filter expression `ext.account_id = 'normal'`;
- one input file.

Call the production `filterByAccountAndFilename` path and run the focused test three times:

```bash
go test ./pkg/sql/colexec/external \
  -run '^TestAuditFileFilterDoesNotTreatAccountIDAsAccountVirtualColumn$' \
  -count=3 -v
```

All three runs panic:

```text
runtime error: invalid memory address or nil pointer dereference
github.com/matrixorigin/matrixone/pkg/container/vector.(*Vector).IsConstNull
github.com/matrixorigin/matrixone/pkg/sql/colexec.(*ColumnExpressionExecutor).Eval
github.com/matrixorigin/matrixone/pkg/sql/colexec/external.filterByAccountAndFilename
    pkg/sql/colexec/external/external.go:410
```

### Expected behavior

Only the exact virtual account column (optionally qualified by a relation name) and the exact `__mo_filepath` column should be evaluated during file-level pruning. A physical column such as `account_id`, `accounting`, or `customer_account` must remain a row-level filter.

### Actual behavior

The physical-column filter is moved into the file-level filter list and evaluated against a nil vector, causing a statement panic.

### Root cause

Both copies of `containColname` use substring matching:

```go
return strings.Contains(col, STATEMENT_ACCOUNT) ||
       strings.Contains(col, catalog.ExternalFilePath)
```

Therefore names such as `ext.account_id` match `account`. In contrast, `makeFilepathBatch` allocates vectors only when the table column name is exactly `account` or `__mo_filepath`.

[...]

PR #26065, opened 2026-07-23
fix(external): keep mixed column filters at row level

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25908 

fix(external): keep mixed column filters at row level


XuPeng-SH (changes requested):
[P1] Do not treat an arbitrary dotted physical column as a qualified virtual column.

isFileLevelColumnName strips everything through the last dot before comparing the name. The binder stores the actual identifier in ColRef.Name, so a legal quoted physical column such as customer.account also reaches this helper as customer.account. It is therefore classified as the virtual account column. makeFilepathBatch, however, allocates vectors only for TableDef column names exactly equal to account or __mo_filepath; the customer.account vector remains nil and ColumnExpressionExecutor.Eval panics.

I reproduced this on f742ff4c06 with a production-path unit test using a physical customer.account column. The test failed 3/3 at external.go:468 with Vector.IsConstNull on a nil vector. The same suffix problem applies to a physical name ending in .__mo_filepath, and both duplicated implementations contain it.

Please classify column references from ColRef.ColPos against the exact TableDef column, or otherwise disambiguate an actual relation qualifier from a dot inside the identifier, in both pkg/sql/colexec/external and pkg/sql/plan. Add regression coverage for quoted dotted physical column names ending in .account and .__mo_filepath.

XuPeng-SH (changes requested):
A second correctness hole remains: the new classifier still treats every column named `account` as file-level, even when it is an ordinary physical column of a user-created external CSV table. `query_builder` only appends `__mo_filepath` as a virtual column; a declared `account` column is still read from the CSV. Yet `makeFilepathBatch` replaces it with `getAccountCol(filepath)`, so e.g. a physical row value `row-account` in `etl:/path-account/file.csv` is pruned by `WHERE account = 'row-account'`, and the predicate is then removed from `FilterList`. I reproduced the production `filterByAccountAndFilename` path 3/3: the one input file became an empty file list. The Hive classifier already explicitly excludes `account` for exactly this collision (`hive_partition.go`: it says a physical `account` would produce wrong results). Please scope the legacy tenant-path optimization to tables/columns with an explicit contract or metadata; for generic external tables only a proven virtual column such as `__mo_filepath` may be file-level. Fix both duplicate implementations and add an actual physical `account` regression case.

XuPeng-SH (changes requested):
The recursive classifier also needs to reject volatile/row-dependent functions, not just unsupported columns. A predicate such as `rand() < 0.5 AND 'etl:/file.csv' = __mo_filepath` references no physical column, so this implementation classifies it as file-level. It is then evaluated once per file on the synthetic batch and removed from `node.FilterList`: a kept file returns all of its rows and a rejected file returns none, whereas SQL requires `rand()` to be evaluated for each data row. A focused classifier regression expecting this predicate to stay at row level failed 3/3 on this head. This shape is especially relevant to the new recursion because the filepath column can be nested/on the RHS, where the old detector did not recognize the bare column leaf. Please inspect each `Expr_F` overload and keep the whole expression at row level when `CannotFold()`/volatility makes it unsafe to evaluate once per file; cover the same rule in both duplicated implementations.

XuPeng-SH (changes requested):
Prepared execution exposes another correctness bug in the `Expr_P` path that this classifier explicitly accepts. After successful pruning, `filterByAccountAndFilename` permanently assigns `node.FilterList = filterList2`. External scans cannot retain a prepare-time compile (`compileExternScanWithPlanNodeID` returns `ErrCantCompileForPrepare`), so subsequent EXECUTEs compile the same stored `PreparePlan` again. The first execution has already deleted the filepath predicate from that shared plan. I reproduced the production function twice on the same node with `__mo_filepath = ?`: parameter `etl:/a.csv` correctly selected A on the first call; after switching the process parameter to `etl:/b.csv`, the second call returned both A and B because `FilterList` was empty. This failed 3/3. Please do not consume predicates by mutating the reusable plan (use a compile-local node/filter copy or return consumed/leftover filters as separate state), and add a regression that executes one prepared external-table query at least twice with different filepath parameters. Audit both the planner duplicate and colexec path for the same mutation contract.

XuPeng-SH (changes requested):
File-level evaluation must also handle constant result vectors. The new recursion can classify `IF(TRUE, TRUE, 'never-selected' = __mo_filepath)` as file-level because it finds the filepath column in the unselected branch (and, with the filepath on the RHS, the old detector did not consume this shape). The expression executor correctly folds the whole predicate to a constant TRUE vector whose logical length is the two-file batch, but `MustFixedColWithTypeCheck`/`MustFixedColNoTypeCheck` exposes only the single physical value for a const vector. Both implementations iterate `for i := 0; i < len(bs); i++`, so only file 0 is retained. A production-path test with two files returned only the first file 3/3 even though the predicate is identically true. Iterate over the file/batch logical row count and read const values correctly (for example via the vector APIs that expand/access logical rows), or conservatively reject such flow-control expressions. Add true/false constant-result cases with multiple files in both paths.

gouhongshen (changes requested):
## Codex automated review

The new classifier still has blocking correctness gaps: it is not schema-aware and can move row-dependent expressions into the one-row-per-file evaluator, causing wrong results or panics.

### P1 - Do not identify virtual columns by stripping dotted names (pkg/sql/colexec/external/external.go:317)
A legal quoted physical column such as `customer.account` is reduced to `account` and classified as file-level. `makeFilepathBatch` only provides vectors for exact virtual columns, so the physical predicate is removed from row filtering and the synthetic batch has no `customer.account` vector, which can produce incorrect pruning or a panic. Match the column position/TableDef definition instead of parsing `ColRef.Name`; the duplicate implementation in `pkg/sql/plan/external.go:45` needs the same fix.

### P1 - Keep a physical `account` column at row level (pkg/sql/colexec/external/external.go:320)
An exact physical CSV column named `account` is still classified as the path-derived virtual account column. Generic external tables append only `__mo_filepath`; `makeFilepathBatch` fabricates `account` from the file path, so a predicate such as `WHERE account = 'row-value'` is evaluated against the path account and then removed from `node.FilterList`, silently returning the wrong rows/files. The classifier must distinguish the actual virtual column from a physical `account` column using schema metadata; the duplicate planner helper has the same defect.

### P1 - Reject volatile functions from file-level filters (pkg/sql/colexec/external/external.go:386)
`classifyFileLevelColumns` treats every function whose descendants have no ordinary column as supported and never checks volatility. For example, `rand() < 0.5 AND 'etl:/file.csv' = __mo_filepath` is classified as file-level, so `filterByAccountAndFilename` evaluates `rand()` once per synthetic file row and removes the whole predicate at line 484. The random predicate must remain row-level; reject volatile/time-dependent functions (and apply the same fix to the duplicate planner classifier).

### P1 - Broadcast or reject constant results across all files (pkg/sql/colexec/external/external.go:386)
The new recursion classifies `IF(TRUE, TRUE, 'never-selected' = __mo_filepath)` as file-level because it sees the virtual column in the inactive branch. Expression folding returns a constant vector whose logical length is set to the file count, but `MustFixedColWithTypeCheck[bool]` at line 476 exposes only the single physical value and the loop therefore retains only the first file. Use a logical-length-aware constant access path or conservatively keep flow-control expressions out of file-level filtering; the planner duplicate has the same issue.

heni02 (approved):

iamlinjunhong:
> ## Codex automated review
> The new classifier still has blocking correctness gaps: it is not schema-aware and can move row-dependent expressions into the one-row-per-file evaluator, causing wrong results or panics.
> 
> ### P1 - Do not identify virtual columns by stripping dotted names (pkg/sql/colexec/external/external.go:317)
> A legal quoted physical column such as `customer.account` is reduced to `account` and classified as file-level. `makeFilepathBatch` only provides vectors for exact virtual columns, so the physical predicate is removed from row filtering and the synthetic batch has no `customer.account` vector, which can produce incorrect pruning or a panic. Match the column position/TableDef definition instead of parsing `ColRef.Name`; the duplicate implementation in `pkg/sql/plan/external.go:45` needs the same fix.
> 
> ### P1 - Keep a physical `account` column at row level (pkg/sql/colexec/external/external.go:320)
> An exact physical CSV column named `account` is still classified as the path-derived virtual account column. Generic external tables append only `__mo_filepath`; `makeFilepathBatch` fabricates `account` from the file path, so a predicate such as `WHERE account = 'row-value'` is evaluated against the path account and then removed from `node.FilterList`, silently returning the wrong rows/files. The classifier must distinguish the actual virtual column from a physical `account` column using schema metadata; the duplicate planner helper has the same defect.
> 
> ### P1 - Reject volatile functions from file-level filters (pkg/sql/colexec/external/external.go:386)
> `classifyFileLevelColumns` treats every function whose descendants have no ordinary column as supported and never checks volatility. For example, `rand() < 0.5 AND 'etl:/file.csv' = __mo_filepath` is classified as file-level, so `filterByAccountAndFilename` evaluates `rand()` once per synthetic file row and removes the whole predicate at line 484. The random predicate must remain row-level; reject volatile/time-dependent functions (and apply the same fix to the duplicate planner classifier).
> 
> ### P1 - Broadcast or reject constant results across all files (pkg/sql/colexec/external/external.go:386)
> The new recursion classifies `IF(TRUE, TRUE, 'never-selected' = __mo_filepath)` as file-level because it sees the virtual column in the inactive branch. Expression folding returns a constant vector whose logical length is set to the file count, but `MustFixedColWithTypeCheck[bool]` at line 476 exposes only the single physical value and the loop therefore retains only the first file. Use a logical-length-aware constant access path or conservatively keep flow-control expressions out of file-level filtering; the planner duplicate has the same issue.

fixed

aunjgr (approved):
Reviewed independently from first principles. File-level classification is conservative, mixed predicates remain row-level, and external-scan nodes are deep-copied before pruning, so prepared-plan reuse remains safe. I found no actionable correctness defects.

XuPeng-SH (approved):
Re-reviewed current head 1d4c8fcd495d10309ac29060b1b35c61500f2479 from first principles after the requested fixes. I found no remaining blocking correctness, performance, or lifecycle issue.

Key closure checks:
- File-level columns are proven by TableDef position plus TbColToDataCol, so physical account/account_id/dotted names stay row-level.
- Mixed-column, volatile, time-dependent, unknown, and unsafe expressions conservatively remain row-level; mo_log_date keeps its intended deterministic filepath optimization.
- Constant TRUE/FALSE results are broadcast over the logical file count, including NULL handling.
- File pruning no longer mutates the reusable plan; execution uses compile-local node state, so repeated prepared EXECUTEs retain parameterized filepath predicates.
- Hive staging returns every unconsumed filepath predicate to row filtering.
- Resource/error paths release batches and executors; this change adds no goroutine/lock/channel dependency, and temporary growth is bounded by file-list/expression size.

Local validation on this exact head passed: focused regression tests; all tests for ./pkg/sql/colexec/external, ./pkg/sql/compile, and ./pkg/sql/plan; the same three packages under -race; focused race tests with -count=3; go vet; and git diff --check. CI jobs still in progress are independent of this code-review decision.

gouhongshen (changes requested):
## Codex automated review

The latest commit closes the previously reported correctness and lifecycle gaps, but introduces a blocking file-pruning regression after normal column pruning.

### P1 - Compare virtual-column identity in the correct position space (pkg/sql/colexec/external/external.go:329)
`ColRef.ColPos` is remapped into the pruned `TableDef`, while `TbColToDataCol` retains original external-file field indexes. For a table `(a,b,__mo_filepath)`, `SELECT a ... WHERE __mo_filepath LIKE ...` produces `[a,__mo_filepath]`: the filepath becomes local position 1, which collides with `b`'s original field index 1 and is falsely classified as physical. The predicate therefore remains row-level, forcing every external file to be opened and read instead of pruning the file list. The former name-based classifier did prune this case, so this is an unbounded I/O regression for common projections. The duplicate implementation in `pkg/sql/plan/external.go` has the same coordinate-space error. Use stable virtual-column metadata and add coverage using a column-pruned production-shaped node.

iamlinjunhong:
> ## Codex automated review
> The latest commit closes the previously reported correctness and lifecycle gaps, but introduces a blocking file-pruning regression after normal column pruning.
> 
> ### P1 - Compare virtual-column identity in the correct position space (pkg/sql/colexec/external/external.go:329)
> `ColRef.ColPos` is remapped into the pruned `TableDef`, while `TbColToDataCol` retains original external-file field indexes. For a table `(a,b,__mo_filepath)`, `SELECT a ... WHERE __mo_filepath LIKE ...` produces `[a,__mo_filepath]`: the filepath becomes local position 1, which collides with `b`'s original field index 1 and is falsely classified as physical. The predicate therefore remains row-level, forcing every external file to be opened and read instead of pruning the file list. The former name-based classifier did prune this case, so this is an unbounded I/O regression for common projections. The duplicate implementation in `pkg/sql/plan/external.go` has the same coordinate-space error. Use stable virtual-column metadata and add coverage using a column-pruned production-shaped node.

fixed

gouhongshen (approved):
## Codex automated review

Prior blockers for physical/dotted account columns, volatile expressions, constant-vector broadcasting, and prepared-plan mutation remain fixed. The remaining column-remap blocker is now fixed by a stable synthetic filepath ColId, preserved through pruning and covered by production-shaped tests. No topics remain blocking or require non-blocking acceptance. Focused tests could not start because this worktree lacks prebuilt CGo dependencies; diff checks passed.

volgariver6 (approved):

LeftHandCold (approved):
## Codex automated review

Re-reviewed exact head `35252ddec751e82f5a7209c6e242ed247626f3d6` against current `main` and GitHub test-merge `fe469f80572dd660a0cad66c7b2c7b099a246b67`. I found no remaining blocking correctness or normal-business regression issue.

- File-level pruning now identifies only the synthetic `__mo_filepath` column by stable schema identity; physical `account`, dotted identifiers, mixed-column predicates, volatile/real-time functions, and unsafe OR shapes remain at row level.
- Constant results are applied across the logical file count, Hive leftovers return to row filtering, and external scan compilation uses a deep-copied node so prepared-plan reuse does not consume predicates.
- The new logservice test helper uses independent raft/gossip/service ports and preserves the successful config; the TAE checkpoint wait has a bounded context.
- Q1: batches, executors, services, and contexts have matching cleanup. Q2: no new unbounded wait-for edge. Q3: temporary growth is bounded by file/test-instance count.

Validation on the test-merge tree passed: full tests for `pkg/sql/colexec/external`, `pkg/sql/plan`, `pkg/sql/compile`, and `pkg/logservice`; focused `TestCkpLeak`; focused regression tests under `-race -count=3`; build, vet, and `git diff --check`. GitHub UT, coverage, SCA, BVT, and Iceberg checks are green; remaining pending entries are merge-queue gates.

PR #26077, opened 2026-07-23
fix daemon polling after task framework reenable

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25940

fix daemon polling after task framework reenable


XuPeng-SH (approved):
No blocking findings on current head 4bf9d48c. The fix closes the disabled -> re-enabled polling transition by resetting the one-shot timer on every tick, while preserving cancellation and bounded query behavior. Ownership and unhappy paths are clean: the poll worker remains the sole timer owner, cancellation reaches the deferred Stop, and no queue or goroutine growth is introduced.

Verified locally:
- focused regression with `-race -count=100`
- full `pkg/taskservice` package with `-race`
- `go vet ./pkg/taskservice`
- clean merge with latest main

All applicable CI checks are green.

PR #26080, opened 2026-07-23
fix(colexec): honor cancellation in window pipelines

### LINKED ISSUE #26067: [Bug]: KILL QUERY / KILL CONNECTION does not cancel long-running Window+UNION query (session stays COM_QUERY for 40h+)

- [x] I have checked existing issues. Related themes exist around session/txn cleanup, but not this Window pipeline cancel failure.

## Branch Name

```
main
```

## Commit ID

```
826a3e3d294ba62c7c4da217d1445f3c568ce4d9
```

Image: `registry.cn-shanghai.aliyuncs.com/matrixorigin/matrixone:nightly-826a3e3d`

```
- Hardware: single-node MO (docker, host port 16001), rootless docker on Linux
- OS: Linux
- Dataset: jinpan_001
- Suite: cases/jinpan_001/join_adv @q1004_adv_window_frame_running
- Client: mo-regression-test run.py --sql-timeout 1200 (func_timeout + side-connection KILL)
```

## Actual Behavior

1. Start a heavy Window + `UNION ALL` + outer `LIMIT 200` query over large `dwd_s4_acdoca` (see `q1004.sql`).
2. Client hits **20-minute** SQL timeout and issues cleanup:

```text
Timeout cleanup: KILL QUERY <conn_id>
Timeout cleanup: KILL CONNECTION <conn_id>
```

3. `KILL` statements return **without error**.
4. `information_schema.processlist` still shows the **same `conn_id`** as `COM_QUERY` with the original SQL.
5. In our run, the statement stayed active for **40+ hours** (`run_sec ≈ 146700+`), CPU still busy (~200%+ on the container), memory ~13GiB — no panic/OOM in CN logs.
6. Regression runner cannot advance to the next SQL because the server query / connection never finishes after the client-side timeout.

Manual retry also fails to stop it:

```sql
KILL QUERY 2208;      -- no error
KILL 2208;            -- no error
-- processlist still: conn_id=2208, command=COM_QUERY, same query_start
```

Only **docker restart** of the MO container clears the session.

## Expected Behavior

- `KILL QUERY <id>` cancels the running statement promptly.
- `KILL CONNECTION <id>` / `KILL <id>` disconnects the session and aborts the pipeline.
- After a successful KILL, the conn_id must disappear from `processlist` (or move out of `COM_QUERY`) within a short time.
- Client timeout paths that issue KILL must be able to reclaim the CN and continue other work.

## Steps to Reproduce

1. Use nightly image `matrixone:nightly-826a3e3d` with jinpan_001 loaded (or any large table reproducing a multi-minute Window query).
2. In session A, run `q1004.sql` (Window over ~16M filtered rows, duplicated under `UNION ALL`, outer `LIMIT 200`).
3. Note `conn_id` from `SHOW PROCESSLIST` / `information_schema.processlist`.
4. In session B:

```sql
KILL QUERY <conn_id>;
-- and/or
KILL CONNECTION <conn_id>;
KILL <conn_id>;
```

5. Re-check processlist: query is still `COM_QUERY` with the same `query_start`.
6. Observe it continues for a long time (in our case >40h until container restart).

### Repro SQL

Same as join_adv `q1004` (full text in `q1004.sql`). Shape:

```sql
SELECT u.* FROM (
  SELECT * FROM ( /* window over acdoca ~16M rows */ ) a
  UNION ALL
  SELECT * FROM ( /* identical window query */ ) a
) u
LIMIT 200;
```

## Additional Context / Evidence

| Time (CST) | Event |
|------------|--------|
| 2026-07-21 18:36:20 | Query start (conn 2208) |
| 2026-07-21 18:56:20 | Client timeout; `KILL QUERY/CONNECTION 2208` logged |
| 2026-07-23 11:21+ | Same conn 2208 still `COM_QUERY`, `run_sec≈146700` |

- Plan type: `AP QUERY PLAN ON ONE CN` with nested `Window` + `Partition` under `Union All` (see companion enhancement issue on LIMIT/UNION short-circuit).
- No matching panic/fatal for this statement_id in docker logs during the hang window.
- Separate enhancement issue covers optimizer/efficiency; **this issue is specifically about cancel not working**.

## Impact

- Makes per-SQL timeouts useless for this class of query.
- Blocks entire join_adv / long-running regression suites (one uncancellable SQL freezes the worker).
- Requires container restart to recover — unacceptable for shared CI clusters.

## Suggested direction (for implementers)

- Ensure Window / Partition / merge pipeline operators periodically check cancel context.
- Make `KILL QUERY` mark the session aborted even if an operator is inside a long sort/spill.
- Add a regression test: start a long Window query, KILL it, assert processlist clears within N seconds.

jiangxinmeng1:
Poll query cancellation during long-running window frame evaluation and partition output so KILL QUERY and KILL CONNECTION can stop the pipeline promptly. Add deterministic cancellation and cleanup regressions.

Fixes #26067

## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26067

fix(colexec): honor cancellation in window pipelines


XuPeng-SH (approved):
Deep-reviewed at e6c522005348e93edb54e80a5ffa082b1bda9dc7. Cancellation polling covers the long-running window frame/value/order paths and partition output; error exits are reclaimed through the existing Reset/Free lifecycle, and the operator protocol remains intact. Verified the new cancellation tests repeatedly, both owning packages in full and under race, plus the compile integration dependency. CI is green. No blocking findings.

PR #26085, opened 2026-07-23
feat(incrservice): add transactional offset resets

VioletQwQ-0:
## What type of PR is this?

- [x] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [x] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23143

This is the next prerequisite after #26027. It does not expose `ALTER TABLE ... AUTO_INCREMENT=N` yet.

The invariant in this PR is that an AUTO_INCREMENT offset reset and its allocator-cache transition belong to the same transaction:

- the resetting transaction sees its pending offset without publishing it to other transactions;
- statement or transaction rollback retires the private reset cache;
- commit publishes a cache for the allocation epoch actually used by the write;
- stale builders and old cache generations cannot become active after reset/reload/close;
- accepted allocator actions drain during close, while a canceled queued reset cannot mutate the store.

The shared service interface now carries `autoIncrEpoch` and `TxnOperator` through both production allocation callers. `memStore` and `sqlStore` gain monotonic `SetOffset` plus DDL-constrained `ForceSetOffset`; all implementations and mocks are updated atomically.

Out of scope for this PR:

- planner/compiler support for the user-facing ALTER syntax;
- COPY/CLONE allocator reconciliation;
- rolling-upgrade enablement for mixed-version TNs;
- end-to-end BVT for the final SQL behavior.

Diff size, separated for reviewability:

- production: `+1303 / -86`
- tests: `+1832 / -8`
- mocks: `+292 / -18`

Validation:

- `go test -race ./pkg/incrservice -count=1`
- `go test ./pkg/sql/colexec/preinsert ./pkg/sql/colexec/table_clone ./pkg/frontend/test ./pkg/vm/engine/test/testutil -count=1`
- `go test ./pkg/... -run '^$'`
- `git diff --check`
- exact-head `mo-pr-preflight-review`: PASS at `891781d3a927b80b54d38e381ebee0d749c1c589`


XuPeng-SH (changes requested):
Requesting changes at exact head `acf8a235da093457d932cd372ffe3c9fdafc4c08` for two deterministic lifecycle/transactionality bugs.

[P1] Bound accepted `ForceSetOffset` execution (`pkg/incrservice/allocator.go:256`). Unlike `doAllocate` and `doUpdate`, `doForceSetOffset` forwards the caller context without adding any deadline. `allocator.close` queues a barrier behind accepted actions and waits for it before closing `done` or stopping the runner, so a store call waiting on an uncanceled caller context makes Close hang indefinitely. I reproduced this with a context-aware blocking store: Close returned only after an external manual release. Please wrap this action in a bounded context (and attach its cause), or otherwise cancel in-flight force updates during shutdown.

[P1] Keep mem-store resets transaction-private (`pkg/incrservice/store_mem.go:224`). If `txnOp` has no manually pre-created entry for this table, this branch leaves `m` pointing at `s.caches`, and line 237 changes the committed row immediately. The normal existing-table `service.SetOffset` path then calls `GetColumns(..., txnOp)`, sees no private map, reports `NoSuchTable`, and rollback cannot restore the committed offset. I reproduced this from committed offset 10: `ForceSetOffset(..., 99, txnOp)` made `s.caches` equal 99 before commit. The added tests hide this by manually calling `store.Create(..., alterTxn)` first. Please create/clone a txn-private row with symmetric commit/rollback cleanup on first transactional update, or reject this path without mutating committed state.

The generation publication, callback registration, private-cache ownership, terminal uint64 handling, and SQL-store transaction propagation otherwise look sound. Fresh local validation passed full `pkg/incrservice` race tests, 20x race stress over reset/epoch/builder/retire/close cases, and the changed preinsert/table-clone/mock packages. GitHub required CI is green. A full local compile-only sweep was attempted but the host `/tmp` filled during linking; failures were exclusively `no space left on device`, while the relevant packages compiled and passed independently.

VioletQwQ-0:
Addressed both findings in `dd4c50f72`:

1. `doForceSetOffset` now derives a cause-aware 10-second deadline while preserving caller cancellation, so an accepted store call cannot hold the allocator close barrier indefinitely. `TestForceSetOffsetUsesBoundedContext` verifies the production action path passes a bounded context.
2. Transactional `memStore.SetOffset` / `ForceSetOffset` now lazily clone the committed table row into transaction-private state and install the close callback before mutation. Rollback drops that state; commit publishes it. The service regressions no longer pre-create private rows, and `TestMemStoreForceSetOffsetCreatesTransactionPrivateState` covers both rollback and commit directly.

Validation on exact head `dd4c50f727f01655b912d92a8223220d3fa4483a`:

- `go test -race ./pkg/incrservice -count=1`
- affected caller package tests
- `go test ./pkg/... -run '^$'`
- `git diff --check`
- exact-head `mo-pr-preflight-review`: PASS

XuPeng-SH (approved):
Approved at exact head dd4c50f727f01655b912d92a8223220d3fa4483a.

Re-reviewed the full diff and the lifecycle/transaction closure. The two prior P1 findings are closed: ForceSetOffset now runs with a cause-aware 10-second bound, and memStore lazily clones committed rows into transaction-private state with symmetric commit/rollback cleanup.

Fresh local validation passed the full pkg/incrservice race suite, affected preinsert/table_clone tests, frontend and VM mock compilation, the direct pkg/vm/engine/test compile check, and git diff --check. The Ubuntu UT job's failure is infrastructure-only: the hosted runner received a shutdown signal and exited 143 while tests were running; no test assertion failure was reported. Other required coverage, SCA, and BVT checks are green.

No blocking findings.

XuPeng-SH (approved):

gouhongshen (approved):
## Codex automated review

No findings. The prior lifecycle and memStore transactionality findings are fixed on this head. The epoch/private-cache transitions, rollback/commit cleanup, allocator shutdown, callers, and boundary handling are consistent. Diff/format checks and moerr tests passed; the incrservice race suite could not run because the required local libmo.dylib artifact is absent.

Issue #154, opened 2021-06-28
(aoe): implement a second version of the table column data management and integration into the engine


XuPeng-SH:
The current implementation is completely column-oriented, the internal data segmentation, sharding, and other information are just meta-information attached to the column data. For example, for all column data in the same segment, they only share some segment-level data (meta, ...). 
```
                          Seg0           |           Seg1            |           Seg2            |
                   Blk0         Blk1     |     Blk0          Blk1    |     Blk0          Blk1    |
          Col0 ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
Table0 -- Col1 ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
          Col2 ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
```
The above design is not convenient with the current calculation layer implementation, and the following design may be more convenient for the current application layer requirements:
```
                            Blk0         Blk1                         Blk0          Blk1    
                    Col0 ----------- | ----------- |            | ----------- | ----------- |
Table0 --   Seg0 -- Col1 ----------- | ----------- | -- Seg1 -- | ----------- | ----------- | 
                    Col2 ----------- | ----------- |            | ----------- | ----------- |
```
 We have implemented the first design and we can implement the second design and keep both implementations as v1|v2 in our engine.

PR #26086, opened 2026-07-23
build: reuse native and Go caches in CI image builds

aptend:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25782 (validation infrastructure only; this PR does not close the issue)

- Adds a CI-specific multi-stage Dockerfile that separates native dependencies, Go build cache, and the final runtime image.
- Adds a Make target that consumes prebuilt CGO and third-party libraries instead of rebuilding them after every source change.
- Excludes local native outputs from the Docker context so stale host artifacts cannot overwrite cached image contents.

This is the MatrixOne half of the change. The mo-auto-test PR supplies the trusted registry/cache orchestration.

Validation:

- bash syntax and Make dry-run checks passed.
- git diff --check passed.
- TKE validation run 29981557105 built the image in 8m23s versus 19m05s in run 29938558644 (10m42s / 56% faster). COPY go.mod/go.sum took 0.0s and COPY . . took 1.9s.


.dockerignore
```diff
@@ -6,6 +6,7 @@ logs/
 # Ignore IDE files
 .vscode/
 .idea/
+.git/
```
XuPeng-SH: [P1] Preserve VCS metadata for the existing image-build paths

This is the repository-wide `.dockerignore`, so ignoring `.git/` also changes `optools/images/Dockerfile` and `Dockerfile.dev`, not only the new CI Dockerfile. The new file passes `BRANCH_NAME`/`LAST_COMMIT_ID`/`MO_VERSION`, but the existing Dockerfiles still run plain `make build`; the Makefile obtains those values from `git rev-parse`/`git describe`. On an export of this exact head without `.git` (the resulting Docker context), `VERSION_INFO` expands with empty `BranchName`, `CommitID`, and `Version`. The documented production build and `etc/launch-tae-compose` therefore produce binaries with no commit identity. This is functional, not just cosmetic: CN heartbeats publish `version.CommitID`, and shard/query routing compares commit IDs for binary compatibility, so independently built revisions can all compare equal as `""`. Please either scope this ignore to `Dockerfile.ci` or update every retained Docker build path to receive and inject trusted VCS metadata, with a check that `mo-service -version` contains the expected commit.

XuPeng-SH (changes requested):
Deep review on exact head `533ac4f3086cd52abef98d4ce755b81b46a64e21` found one compatibility blocker: the global `.git/` exclusion strips version/commit metadata from the existing Docker image paths, where empty commit IDs also defeat runtime binary-compatibility discrimination. The new native/Go cache orchestration with mo-auto-test #154 otherwise closes correctly; BuildKit checks for both Dockerfiles, the new Make target dry-run, diff checks, thread audit, and all PR CI checks are clean.

XuPeng-SH (approved):
Re-reviewed the exact latest head. The previous VCS-metadata blocker is closed: `.git` is no longer excluded repository-wide, while the trusted cache workflow scopes that exclusion to the CI path and passes explicit branch/commit/version values. The new build also fails fast on a missing commit ID and verifies the embedded commit from the built binary. Native compile/link/runtime-load closure is consistent with the Makefile contract. Both Dockerfiles pass BuildKit `--check` with no warnings, the new Make target dry-run is correct, diff/merge-tree checks against current main pass, and all PR CI checks are green. No blocking correctness, performance, or lifecycle issue found.

Issue #26047, opened 2026-07-22
[Bug]: TestISCPExecutor4 flaky — ISCP HNSW index change propagation timing


matrix-meow:
## Problem

`TestISCPExecutor4` in `pkg/vm/engine/test` fails intermittently (5.93s elapsed):

The test exercises Intra-System Change Propagation (ISCP) for HNSW vector indexes across multiple tables (hnsw_idx_0, hnsw_idx_1, hnsw_idx_2). It creates tables, inserts data, and verifies change propagation watermarks. The test fails due to timing/race conditions in the change propagation pipeline.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-22 | [PR #25815](https://github.com/matrixorigin/matrixone/actions/runs/29913023702/job/88909341158) | Coverage |

## Root Cause

ISCP executor test relies on change propagation completing within expected timeframes. Under CI runner load, the propagation pipeline (WAL → ISCP task → watermark update) may not complete in time, causing assertion failures.

## Suggested Fix

1. Increase timeout/wait tolerance for ISCP propagation in tests
2. Or add explicit synchronization barriers for test determinism
3. Or use polling with retry instead of fixed waits

jiangxinmeng1:
fixed by #26093

PR #26093, opened 2026-07-23
test: stabilize TestISCPExecutor4 recovery

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26047

stabilize TestISCPExecutor4 recovery

### DIFF
--- a/pkg/vm/engine/test/change_handle_test.go
+++ b/pkg/vm/engine/test/change_handle_test.go
@@ -2459,30 +2459,26 @@ func TestISCPExecutor4(t *testing.T) {
 		assert.NoError(t, txn.Commit(ctxWithTimeout))
 	}
 
-	appendFn := func(idx int) {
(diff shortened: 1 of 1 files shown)


XuPeng-SH (approved):
Reviewed the complete diff and the commit-TS → ISCP watermark → fault recovery closure on head dfc10dcf. No blocking findings. Using the successful append transaction CommitTS fixes the previous over-advanced clock oracle; bounded Eventually/Never checks preserve the negative assertions, and the final job-specific fault now verifies recovery of all three jobs. Local validation: TestISCPExecutor4 passed 10 consecutive runs, passed 3 consecutive runs under -race, and pkg/vm/engine/test passed go vet. Current CI is green.

PR #26094, opened 2026-07-23
fix(window): support prepared ROWS frame bounds

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25993

Prepared statements could not parse `? PRECEDING/FOLLOWING` ROWS bounds, and window planning previously folded finite bounds before execution. This change preserves typed parameter expressions, visits and renumbers all window expressions without double-processing shared partition aliases, transports parameters unchanged, and materializes immutable operator-local ROWS frames once per `Prepare` generation. Dynamic RANGE and interval bounds remain explicitly unsupported.

The runtime path rejects missing, NULL, negative, fractional, overflow, and failed-conversion bounds before processing rows. `Reset` and `Free` clear runtime frames, while cached plan expressions remain unchanged.

Tested with:

`./.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/sql/parsers/dialect/mysql ./pkg/sql/parsers/tree ./pkg/sql/plan ./pkg/sql/colexec/window ./pkg/vm/process`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -run "TestWindow(Prepare|Reset).*FrameBound" -count=1 -timeout=240s ./pkg/sql/colexec/window`

Configured `go build` and `go vet` passed for all modified packages. The distributed SQL fixture is included for CI; local `mo-tester` was unavailable.


XuPeng-SH (approved):
Deep-reviewed the exact latest head. The parser/binder/prepare visitor/runtime closure is complete: ROWS markers remain typed plan expressions, alias-shared window expressions are visited once, cached operators materialize immutable per-Prepare frames from the current parameter vector, non-cached executions deep-copy before parameter replacement, remote processes preserve parameter bytes/binary flags, and Reset/Free close the generation boundary. Invalid NULL/negative/fractional/overflow/missing values fail before row processing. Generated parser output is byte-for-byte reproducible with zero new conflicts. Fresh affected-package tests, focused race x10, full window-package race, build, vet, diff checks, and the merge-tree against current main all pass; all PR CI checks are green. No blocking correctness, performance, or lifecycle issue found.

heni02 (approved):

Issue #25996, opened 2026-07-22
[Bug]: PREPARE rejects SET statements with and without parameter markers


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Searched for prepared `SET` statements and both observed validation errors. No matching issue was found.

### Branch Name

main

### Commit ID

`c0486a2446c84c6dd6bf142494fe85243877ddc5`

### Other Environment Information

- MatrixOne: `8.0.30-MatrixOne-v1.3.0`
- MySQL comparison: `8.0.45`
- SQL-level `PREPARE` / `EXECUTE` through the mysql client

### Actual Behavior

MatrixOne supports direct `SET` statements but rejects both literal and parameterized prepared forms.

Literal form:

```sql
PREPARE p_literal FROM 'SET @x = 41 + 1';
```

returns:

```text
ERROR 20301 (HY000): invalid input: cannot prepare TCL and DCL statement
```

Parameterized form:

```sql
PREPARE p_parameter FROM 'SET @x = ?';
```

returns the contradictory error:

```text
ERROR 20301 (HY000): invalid input: only prepare statement can use ? expr
```

Both failures reproduced 3/3 times. MySQL 8.0.45 supports both prepared forms and sets the user variable to the expected values.

### Expected Behavior

`SET` should be accepted as the content of a prepared statement, with parameter markers bound at `EXECUTE` time, consistent with MySQL and MatrixOne's direct `SET` behavior.

### Steps to Reproduce

```sql
-- Direct control succeeds.
SET @x=42;
SELECT @x;

-- Literal prepared SET is rejected.
SET @x=0;
PREPARE p_literal FROM 'SET @x = 41 + 1';
EXECUTE p_literal;
SELECT @x;

-- Parameterized prepared SET is also rejected.
SET @x=0,@value=77;
PREPARE p_parameter FROM 'SET @x = ?';
EXECUTE p_parameter USING @value;
SELECT @x;
```

### Additional Information

`BuildPlan` represents `*tree.SetVar` as a DCL `SetVariables` plan. `ResetPreparePlan`, which is called for the inner statement of `PREPARE`, rejects DCL plans except a small account-management subset. In addition, the SetVar expression builder does not propagate the prepared-statement context to parameter binding, which produces the second error for `SET @x=?`.

LeftHandCold:
[...]

PR #26096, opened 2026-07-23
fix(frontend): support prepared SET statements

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25996

Prepared SET was rejected because the DCL carrier did not retain ordered parameter metadata and SET expression evaluation lost prepared context when it synthesized internal SELECT statements.

This change collects parameters and schemas from retained SET expressions plus transient scalar-subquery plans, preserves global parser ordinals across multiple assignments, explicitly propagates prepared mode through initial and retry planning, bypasses invalid empty-SQL plan-cache keys for SET-derived AST inputs, and preserves decimal result binding. Supplemental aggregate/window traversal is idempotent with the generic visitor work in #26042 and #26094.

The saved prepared AST and DCL plan remain immutable during EXECUTE. Text and binary repeated execution, multiple markers, NULL replacement, scalar/nested subqueries, aggregate/window fields, decimal expressions, cache isolation, and retry rebuilding are covered.

Tested with:

`./.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/sql/plan ./pkg/frontend`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -run "TestPreparedSet|TestSetExpression|TestPreparedSetExpression" -count=1 -timeout=240s ./pkg/frontend`

Configured `go build` and `go vet` passed for planner/frontend. The distributed SQL fixture is included for CI; local `mo-tester` was unavailable.


XuPeng-SH (approved):
Deep-reviewed exact head f154b461490c26d1465f2ae022079415ba92a59d. The prepared SET path is functionally closed across parameter discovery/order, nested scalar-subquery metadata, prepared-context propagation through initial and retry planning, AST-only cache bypass, result rebinding, and retained-plan/AST immutability. I also traced cleanup/retry ownership and did not find a blocking correctness or lifecycle issue.

Verified locally on the exact head: full pkg/sql/plan and pkg/frontend tests, focused prepared-SET/planner tests under -race, and git diff --check all pass. The two red CI jobs are not attributable to this diff: Iceberg failed during external module download with HTTP 502, and Ubuntu UT failed only in the unrelated TAE TestCkpLeak checkpoint timeout.

heni02 (approved):

PR #26097, opened 2026-07-23
fix: decouple local shuffle producers from consumers

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25922

Fixed-bucket local shuffle workers previously produced and consumed through the same pull pipeline. With bounded bucket queues, nested shuffles could form a wait cycle: a worker blocked while publishing to another bucket and therefore stopped draining the bucket needed by a sibling producer.

This change separates each local shuffle worker into an asynchronous producer and a fixed-bucket consumer. Producers use query-scoped contexts, bounded pool writes, explicit consumer closure, and deterministic producer-before-child cleanup. The VM preparation traversal now lets Shuffle prepare its child tree with the same producer process that will execute it, preventing early worker cleanup from cancelling contexts retained by prepared children such as Merge.

The compiler workarounds that rejected nested local shuffle topologies are removed because the execution layer now supports them directly.

Validation:

- `mo-cgo-test -race ./pkg/sql/colexec/shuffle ./pkg/vm/pipeline`
- critical cancellation and cleanup tests with `-race -count=20`
- `mo-cgo-test ./pkg/vm ./pkg/sql/compile ./pkg/sql/colexec/group ./pkg/sql/colexec/hashjoin`
- `git diff --check`


XuPeng-SH (approved):
Deep-reviewed the asynchronous fixed-bucket shuffle lifecycle, including producer/consumer ownership, cancellation, nested backpressure, Prepare/Reset ordering, shared-pool cleanup, terminal error propagation, and boundedness. Build, vet, targeted tests, full owning-package tests, race tests, and focused race stress passed. No blocking issues found.

PR #26099, opened 2026-07-23
test(tae): scope TestSnapshotGC checkpoint wait

### LINKED ISSUE #26098: test(tae): TestSnapshotGC waits for global checkpoint quiescence
CI flaky test synchronization bug; detailed root cause and fix plan will be added in a comment.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26098

TestSnapshotGC waited for global checkpoint quiescence after a concurrent WAL burst. That is broader than the test contract and flakes when unrelated checkpoint work remains pending.

After the test workers finish, force an incremental checkpoint through db.TxnMgr.Now() under a ten-second context. The test now waits for the checkpoint that covers its own workload; its existing GC and snapshot assertions remain unchanged. This is test-only and does not change MatrixOne runtime behavior.

Validation:

- TestSnapshotGC: 10 iterations
- TestSnapshotGC with race: 3 iterations
- TestSnapshotGC, TestAppendAndGC, TestAppendAndGC2, TestMergeGC, TestPitrMeta, and TestIscpMeta: 3 iterations
- go build ./pkg/vm/engine/tae/db
- go vet ./pkg/vm/engine/tae/db
- git diff --check


XuPeng-SH (approved):
Approved — deep-reviewed exact head `5edd778fb1f297e8f66574bbc09dfe59768a84b3`.

The synchronization scope is correct: both workload wait groups complete before `TxnMgr.Now()` captures the target; `ForceCheckpoint` flushes through that target and `ForceICKP` inherits the caller's 10-second deadline. This waits for the checkpoint covering this test's committed work without requiring unrelated global WAL/checkpoint quiescence. The timeout context is canceled on return, and any scheduled checkpoint remains runner-owned and is terminated by the existing test-engine close path.

Q1–Q3 audit found no leak, unbounded wait, or accumulation introduced by the change. Validation passed:
- `TestSnapshotGC -count=10`
- `TestSnapshotGC -race -count=3`
- Snapshot/append/merge/PITR/ISCP GC regression set `-count=3`
- `go build ./pkg/vm/engine/tae/db`
- `go vet ./pkg/vm/engine/tae/db`
- current GitHub UT, SCA, Coverage, BVT, CodeQL, and PR-valid checks are green.

Issue #24693, opened 2026-05-29
[Bug]: UT TestCCPRExecutorWithGC fails with data race in publication worker sync protection ticker


Ariznawlll:
### Reproduce link

- PR: https://github.com/matrixorigin/matrixone/pull/24636
- Failed workflow job: https://github.com/matrixorigin/matrixone/actions/runs/26616342921/job/78432810806?pr=24636
- Job: Matrixone CI / UT Test on Ubuntu/x86
- Time: 2026-05-29

### Failed UT

`TestCCPRExecutorWithGC` failed under race check.

### Error

```text
WARNING: DATA RACE
Read at 0x00c019ca0408 by goroutine 21382:
  github.com/matrixorigin/matrixone/pkg/publication.(*worker).Stop()
      pkg/publication/worker.go:513
  github.com/matrixorigin/matrixone/pkg/publication.(*PublicationTaskExecutor).Stop()
      pkg/publication/executor.go:328
  github.com/matrixorigin/matrixone/pkg/vm/engine/test.TestCCPRExecutorWithGC.deferwrap3()
      pkg/vm/engine/test/publication_test.go:4487

Previous write at 0x00c019ca0408 by goroutine 21587:
  github.com/matrixorigin/matrixone/pkg/publication.(*worker).RunSyncProtectionKeepAlive()
      pkg/publication/worker.go:566
  github.com/matrixorigin/matrixone/pkg/publication.NewWorker.gowrap3()
      pkg/publication/worker.go:346

testing.go:1617: race detected during execution of test
--- FAIL: TestCCPRExecutorWithGC (3.07s)
FAIL_UT_CASES=TestCCPRExecutorWithGC,
```

### Analysis

The race is on `worker.syncProtectionTicker`. `worker.Stop()` reads/stops `w.syncProtectionTicker` at `pkg/publication/worker.go:513`, while the keepalive goroutine assigns it with `time.NewTicker()` at `pkg/publication/worker.go:566`. There is no synchronization protecting that field.

This failure looks unrelated to PR #24636, which only adds SQL task literal/user lock test coverage. The failed code path is publication/CCPR worker shutdown.

### Existing issue check

Searched existing issues with these keywords and did not find a duplicate:

- `TestCCPRExecutorWithGC race`
- `CCPRExecutorWithGC`
- `syncProtectionTicker`
- `RunSyncProtectionKeepAlive`
- `publication worker race`

VioletQwQ-0:
[...]

PR #26100, opened 2026-07-23
fix(publication): eliminate data race on syncProtectionTicker

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24693

Takes over #24714 while preserving its original commits and authorship.

The original change moved the sync-protection ticker from shared worker state into `RunSyncProtectionKeepAlive`, so `Stop` could no longer race ticker assignment. Latest `main` independently contains the same production fix as part of the broader worker shutdown repair in #25732.

After resolving the conflict with current `main`, this PR is intentionally regression-only:

- retain current `main`'s admission, cancellation, `stopOnce`, and worker-drain semantics;
- add concurrent keepalive-start/worker-stop coverage that would expose reintroduction of the shared-ticker race.

Takeover state:

- source head: `3408a3c998c6e95faa146a88a64e51da535e3ca8`;
- conflict resolved by preserving upstream #25732 lifecycle semantics;
- final diff against current `main` contains only the concurrency regression test.

Validation:

- `go test -race -count=10 -run 'TestWorkerSyncProtection_(StopDuringKeepAliveStart|RegisterGetUnregister|UpdateTTL)$' ./pkg/publication`
- `go test ./pkg/publication -count=1`
- `go vet ./pkg/publication`
- exact-head `mo-pr-preflight-review`: PASS on `cf4fa2c469fe`


XuPeng-SH (approved):
Reviewed the full diff at cf4fa2c469fe9d1b5938e76dae330fda4080067e. No blocking issues found. The regression-only change exercises concurrent sync-protection keepalive startup and worker shutdown under the race detector, while current main keeps ticker ownership local to the keepalive goroutine.

Verified locally:
- focused sync-protection race stress: -race -count=20
- full pkg/publication test suite under -race
- go build ./pkg/publication
- go vet ./pkg/publication

PR #26101, opened 2026-07-23
test(morpc): stabilize sequence mismatch stream test

### LINKED ISSUE #26069: [Bug]: morpc TestStreamServerWithSequenceNotMatch can hang for 40 minutes after stream reset
### Existing issue check

Searched open and closed MatrixOne issues, including title/body/comments, for TestStreamServerWithSequenceNotMatch: no matching issue found.

### Environment

- Failing PR: #26048 (cleanup/remove-legacy-memory-engine), head 53b6f152e417b4883c069106904db14cc0083606
- CI job: https://github.com/matrixorigin/matrixone/actions/runs/29975365220/job/89105897144
- Package: github.com/matrixorigin/matrixone/pkg/common/morpc
- Runner: Ubuntu x86, Go 1.26.4

### Actual behavior

TestStreamServerWithSequenceNotMatch started at 2026-07-23 03:01:37 UTC. Immediately after its write request, the client scheduled a remote-connection reset; the read loop stopped and the stream receive path returned rpc stream closed.

The test then records two assertion failures at server_test.go:761 and :762 (unexpected rpc stream closed, followed by an unexpectedly nil result), but execution continues into a receive from a nil channel at server_test.go:763. The test remains blocked for about 39m43s until the package-level 40-minute timeout panics. This makes the full UT job fail after 2400s.

Relevant timeout stack:

- TestStreamServerWithSequenceNotMatch.func1 at pkg/common/morpc/server_test.go:763: chan receive (nil chan)
- testRPCServer at pkg/common/morpc/server_test.go:951
- TestStreamServerWithSequenceNotMatch at pkg/common/morpc/server_test.go:736

### Expected behavior

The sequence-mismatch/reset scenario must either complete with the expected stream result or fail immediately with a useful assertion. A failed setup/receive must not continue to block on a nil channel, and this unit test must have a bounded failure path.

### Initial scope

The failing test file is unchanged by PR #26048, so this issue is tracked independently of the legacy memory-engine cleanup. It is not yet established whether the first rpc stream closed result is an intended post-reset behavior change or a transport regression; triage should reproduce the test repeatedly and then decide whether to correct the test contract, the reset/stream lifecycle, or both.

### Suggested reproduction

Run the test repeatedly with a short timeout, for example: go test ./pkg/common/morpc -run TestStreamServerWithSequenceNotMatch -count=100 -timeout=60s. Capture the first rpc stream closed event and verify that every failure path terminates without a nil-channel wait.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26069

`TestStreamServerWithSequenceNotMatch` assumed that `Receive` would always return a channel after sending an invalid stream sequence. That assumption no longer matches the explicit MORPC terminal lifecycle: if the server-side sequence rejection resets the backend before `Receive`, the stream is already terminal and `Receive` correctly returns `rpc stream closed` with a nil channel. The test used non-fatal assertions and then received from that nil channel, hanging until the package-level 40-minute timeout.

This PR aligns the test with the stream contract without relaxing production behavior:

- obtain the receive channel before sending the invalid sequence, so the test specifically verifies that an existing receiver gets the nil terminal notification;
- use fatal `require` checks for stream setup and send prerequisites;
- bound terminal delivery with the existing 10-second test context, so a regression fails promptly instead of hanging for 40 minutes.

The invalid first sequence is still sent and still exercises server-side sequence rejection and backend reset.

Validation:

- `go test -run ^TestStreamServerWithSequenceNotMatch$ -count=100 -timeout=120s ./pkg/common/morpc`
- `go test -race -run ^TestStreamServerWithSequenceNotMatch$ -count=100 -timeout=240s ./pkg/common/morpc`
- `go test -count=1 -timeout=240s ./pkg/common/morpc`
- `go build ./pkg/common/morpc`
- `go vet ./pkg/common/morpc`


XuPeng-SH (approved):
Reviewed current head b3b5cfc63d57cedeefd17c8da3270ab197134bc5 from first principles. No blocking correctness, performance, or lifecycle issue found.

The revised ordering preserves the intended causal path: Receive registers the consumer before the invalid first stream request is sent; setting sequence=2 makes Send transmit sequence 3 while the server expects 1; server validation rejects it, cancels the session write path, the backend reset terminates active streams, and the already-obtained receive channel gets the nil terminal value. If the request handler were incorrectly reached, its echoed non-nil response would fail the assertion, so the test has not been weakened. The context-bounded select converts the former package-timeout hang into a prompt deterministic failure. Deferred stream/client/server cleanup retains one effective owner and does not race terminal notification.

Validation on the exact head passed: focused test count=100, focused -race count=100, full pkg/common/morpc tests, full package -race, go build, go vet, and diff check. The PR also merges cleanly into current main (6 commits ahead); the same full package tests, full package -race, build, and vet passed in that synthetic merge state. CI is green.

PR #26105, opened 2026-07-23
fix convert json

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24738

• 分支 0723-fix-convert 主要修改：

  - 支持 CONVERT(expr, JSON)，与 CAST(expr AS JSON) 共用 parser、binder 和 executor。
  - 支持字符串、二进制、数值、decimal、时间、BIT、YEAR、geometry 等类型转换为 typed JSON。
  - 区分文本 JSON document 与 binary/BLOB opaque payload。
  - ENUM 使用显示 label 转换为 JSON STRING，并支持投影、set operation、INSERT、UPDATE 等路径。
  - 修复 BIT 常量折叠丢失类型和位宽的问题。
  - TIMESTAMP 按 session timezone 转换，时间类型固定输出 6 位 FSP。
  - 补充 parser、binder、executor UT，以及完整的 func_convert_json BVT 和结果文件。
  - 增加 NULL、非法 JSON、unsupported type、select-list、prepared parameter 等异常路径测试。


XuPeng-SH (changes requested):
I found two correctness/compatibility blockers in the current head (`2f0f831f`).

[P1] Preserve the raw binary payload and JSON subtype instead of storing base64 as the value

`castToJSON` encodes BINARY/VARBINARY/BLOB and BIT into base64 first and then stores that text in `TpCodeBlob`. This loses two pieces of semantics:

1. `CompareByteJson` compares `TpCodeBlob.GetString()`, so it compares base64 text rather than the original bytes. The new BVT at `func_convert_json.test:73-76` documents and freezes the resulting wrong answer: `x'00' < x'd0'` is true in raw-byte order, but `"AA==" < "0A=="` is false in base64 lexical order. MySQL compares BLOB, BIT, and OPAQUE values byte-by-byte: https://dev.mysql.com/doc/refman/8.4/en/json.html#json-comparison
2. BIT is collapsed into BLOB, so `JSON_TYPE(CONVERT(bit_value, JSON))` returns `BLOB`; MySQL distinguishes BIT from BLOB. The MySQL JSON spec explicitly classifies binary types as BLOB and BIT values as BIT: https://dev.mysql.com/worklog/task/?id=7909

Please make the internal JSON representation retain raw bytes plus the subtype (at least BLOB vs BIT/OPAQUE), and base64-encode only at the text serialization / `JSON_UNQUOTE` boundary. Decoding base64 inside comparisons would restore correctness but add allocation/CPU to every comparison and would still leave subtype semantics wrong. Replace the `mo_base64_order` expected result with the MySQL-compatible raw-byte result and add JSON_TYPE/order tests for both BLOB and BIT.

[P1] SET display values are not quoted as JSON strings

The ENUM workaround only recognizes `moEnumCastIndexToValueFun` in `isEnumDisplayValueExpr`. MatrixOne represents SET as `T_uint64 + Enumvalues` and projects it through `moSetCastIndexToValueFun`, so `CONVERT(set_col, JSON)` falls through to the normal VARCHAR-to-JSON cast. A legal SET label such as `alpha` is then parsed as a JSON document and fails with `invalid input: json text alpha`, rather than producing the JSON string `"alpha"`.

I reproduced this with a focused planner test by changing the mock `nation.n_name` type to `T_uint64{Enumvalues: "alpha,beta"}`: both `CONVERT(n_name, JSON)` and `CAST(n_name AS JSON)` produce `cast(...)` instead of `json_quote(cast_index_to_set_value(...))`. MySQL classifies both ENUM and SET as JSON STRING: https://dev.mysql.com/worklog/task/?id=7909

Please generalize the provenance/rewrite path to cover SET as well as ENUM and add coverage for direct SELECT/CAST, projected/set-op values, JSON-column INSERT, and UPDATE assignment. A minimal BVT should assert `JSON_TYPE(...) = 'STRING'` and `JSON_UNQUOTE(...) = 'alpha'`.

Also, `git diff --check origin/main...HEAD` currently fails on `func_convert_json.result:166` because the empty-input error golden has trailing whitespace. Please clean that up before the next revision.

heni02 (approved):

daviszhen:
> I found two correctness/compatibility blockers in the current head (`2f0f831f`).
> 
> [P1] Preserve the raw binary payload and JSON subtype instead of storing base64 as the value
> 
> `castToJSON` encodes BINARY/VARBINARY/BLOB and BIT into base64 first and then stores that text in `TpCodeBlob`. This loses two pieces of semantics:
> 
> 1. `CompareByteJson` compares `TpCodeBlob.GetString()`, so it compares base64 text rather than the original bytes. The new BVT at `func_convert_json.test:73-76` documents and freezes the resulting wrong answer: `x'00' < x'd0'` is true in raw-byte order, but `"AA==" < "0A=="` is false in base64 lexical order. MySQL compares BLOB, BIT, and OPAQUE values byte-by-byte: https://dev.mysql.com/doc/refman/8.4/en/json.html#json-comparison
> 2. BIT is collapsed into BLOB, so `JSON_TYPE(CONVERT(bit_value, JSON))` returns `BLOB`; MySQL distinguishes BIT from BLOB. The MySQL JSON spec explicitly classifies binary types as BLOB and BIT values as BIT: https://dev.mysql.com/worklog/task/?id=7909
> 
> Please make the internal JSON representation retain raw bytes plus the subtype (at least BLOB vs BIT/OPAQUE), and base64-encode only at the text serialization / `JSON_UNQUOTE` boundary. Decoding base64 inside comparisons would restore correctness but add allocation/CPU to every comparison and would still leave subtype semantics wrong. Replace the `mo_base64_order` expected result with the MySQL-compatible raw-byte result and add JSON_TYPE/order tests for both BLOB and BIT.
> 
> [P1] SET display values are not quoted as JSON strings
> 
> The ENUM workaround only recognizes `moEnumCastIndexToValueFun` in `isEnumDisplayValueExpr`. MatrixOne represents SET as `T_uint64 + Enumvalues` and projects it through `moSetCastIndexToValueFun`, so `CONVERT(set_col, JSON)` falls through to the normal VARCHAR-to-JSON cast. A legal SET label such as `alpha` is then parsed as a JSON document and fails with `invalid input: json text alpha`, rather than producing the JSON string `"alpha"`.
> 
> I reproduced this with a focused planner test by changing the mock `nation.n_name` type to `T_uint64{Enumvalues: "alpha,beta"}`: both `CONVERT(n_name, JSON)` and `CAST(n_name AS JSON)` produce `cast(...)` instead of `json_quote(cast_index_to_set_value(...))`. MySQL classifies both ENUM and SET as JSON STRING: https://dev.mysql.com/worklog/task/?id=7909
> 
> Please generalize the provenance/rewrite path to cover SET as well as ENUM and add coverage for direct SELECT/CAST, projected/set-op values, JSON-column INSERT, and UPDATE assignment. A minimal BVT should assert `JSON_TYPE(...) = 'STRING'` and `JSON_UNQUOTE(...) = 'alpha'`.
> 
> Also, `git diff --check origin/main...HEAD` currently fails on `func_convert_json.result:166` because the empty-input error golden has trailing whitespace. Please clean that up before the next revision.

  1. BINARY/BLOB/BIT 转 JSON
      - BINARY/BLOB 保存原始字节，不再提前保存 base64 文本。
      - BIT 使用独立 JSON subtype。
      - 仅在序列化和 JSON_UNQUOTE 时进行 base64 编码。
      - 比较、JSON_CONTAINS、JSON_OVERLAPS 按 subtype 和原始字节处理。

  2. SET/ENUM 转 JSON
      - SET 与 ENUM 的显示值统一通过 JSON_QUOTE 转成 JSON STRING。
      - 覆盖直接 CAST/CONVERT、投影、集合运算、INSERT 和 UPDATE。

  3. 测试与错误信息
      - 补充 BLOB/BIT 类型、顺序及 SET 转换测试。
      - 修复空 JSON 输入错误结果中的尾随空格。
      - reviewer 指出的生产代码问题已解决；剩余问题主要是无关 result 文件改动需要清理。

XuPeng-SH (changes requested):
The latest head fixes the two blockers from my previous review (raw binary/subtype semantics and SET/ENUM display-value handling), and git diff --check is clean. Two blockers remain:

[P1] The new persisted JSON type codes are not safe for mixed-version readers (pkg/container/bytejson/types.go:147-148). EncodeJson writes the type byte verbatim, so upgraded CNs can persist 0x12/0x13 at the root or inside arrays/objects. The previous reader only recognizes string-like values through TpCodeBlob; for an unknown nested type it falls through to the array/object path and reads a document size from the short opaque payload. A concrete value such as JSON_ARRAY(CONVERT(CAST(X'01' AS BINARY), JSON)) can therefore make an older CN slice past the payload and panic during a rolling upgrade. Root values also become unserializable by the old reader. Please keep the on-disk envelope backward-readable, or gate writes on a cluster-version transition with a safe old-reader representation, and add old/new compatibility coverage.

[P2] Please restore the existing BVT assertions and remove unrelated golden regeneration. In test/distributed/cases/ddl/json_type.result:108 and :110, replacing the exact JSON primary-key and partitioning errors with [unknown result ...] means regressions in those restrictions will pass unnoticed. The PR also rewrites several existing JSON result files purely for output formatting/type-width churn. Keep only expectations whose semantics intentionally change here (for example BIT output and the empty-input message).

daviszhen:
> The latest head fixes the two blockers from my previous review (raw binary/subtype semantics and SET/ENUM display-value handling), and git diff --check is clean. Two blockers remain:
> 
> [P1] The new persisted JSON type codes are not safe for mixed-version readers (pkg/container/bytejson/types.go:147-148). EncodeJson writes the type byte verbatim, so upgraded CNs can persist 0x12/0x13 at the root or inside arrays/objects. The previous reader only recognizes string-like values through TpCodeBlob; for an unknown nested type it falls through to the array/object path and reads a document size from the short opaque payload. A concrete value such as JSON_ARRAY(CONVERT(CAST(X'01' AS BINARY), JSON)) can therefore make an older CN slice past the payload and panic during a rolling upgrade. Root values also become unserializable by the old reader. Please keep the on-disk envelope backward-readable, or gate writes on a cluster-version transition with a safe old-reader representation, and add old/new compatibility coverage.
> 
> [P2] Please restore the existing BVT assertions and remove unrelated golden regeneration. In test/distributed/cases/ddl/json_type.result:108 and :110, replacing the exact JSON primary-key and partitioning errors with [unknown result ...] means regressions in those restrictions will pass unnoticed. The PR also rewrites several existing JSON result files purely for output formatting/type-width churn. Keep only expectations whose semantics intentionally change here (for example BIT output and the empty-input message).

  - 新增 JSON 持久化兼容层：TpCodeOpaque/TpCodeBit 落盘前转换为旧 reader 可识别的 TpCodeBlob envelope，覆盖根节点、嵌套对象/数组、vector 和 MergeBuilder 路径。
  - 新 reader 可从 envelope 恢复 BIT 语义，包括 JSON_TYPE、显示、JSON_UNQUOTE 和原始字节比较。
  - 修正 BIT 常量折叠测试：区分持久化类型 BLOB 与逻辑类型 BIT。
  - 恢复 json_type.result 中 JSON 主键和分区限制的精确错误断言，移除 [unknown result ...] 及无关 golden churn。
  - 保留 BIT Base64 输出、JSON_TYPE=BIT 和空 JSON 错误文案等预期变化。
  - 仍需处理：全局 cast 矩阵意外放开 ALTER TABLE ... MODIFY ... JSON 的无关 DDL 行为变化。

XuPeng-SH (changes requested):
The latest update fixes the mixed-version reader crash from my previous review and restores the important existing goldens. I found two remaining blockers in the current head (`4b3ada3489`).

[P1] Avoid materializing the full decoded BLOB on every comparison

`StorageCompatible` now correctly persists opaque values through the legacy `TpCodeBlob` envelope, but that also means real values in vectors are base64-backed. `CompareBinaryJSON` calls `binaryJSONValue` for both operands, and `opaquePayload` does `base64.DecodeString(string(...))` for each one (`pkg/container/bytejson/bytejson.go:485-530`). A focused benchmark on this head comparing two equal 1 MiB BLOB JSON values measured 4 allocations and 4,915,201 B/op per comparison. The legal 64 MiB BLOB limit therefore permits roughly 300 MiB of transient heap allocation for one comparison; sorting, joins, JSON_CONTAINS, and JSON_OVERLAPS can repeat it many times. `BinaryJSONPayloadLen` also performs the same full decode merely to obtain a length (`:533-537`). This turns a correctness fix into an OOM/latency path on valid inputs.

Please compare legacy base64 payloads without materializing both decoded buffers (for example, fixed-size/group-wise decoding), compute decoded length without allocating, and add an allocation regression test or benchmark.

[P1] Keep expression CAST support from silently broadening ALTER TABLE admission

The new `init` mutates the global `supportedTypeCast` matrix for every JSON source type (`pkg/sql/plan/function/func_cast.go:846-868`). `checkChangeTypeCompatible` consumes that same matrix for schema changes (`pkg/sql/plan/build_alter_modify_column.go:193-202`), so this PR also changes existing `ALTER TABLE ... MODIFY nth JSON` from a rejection to a successful table rewrite and updates the old BVT to freeze that behavior (`test/distributed/cases/ddl/alter_table_modify_column.result:2014-2020`). This is not required to implement CAST/CONVERT or JSON assignment, and it broadens a high-risk DDL path for all newly added source types without migration/constraint/data-path coverage. The author update also identifies this as the remaining unrelated behavior change.

Please keep the DDL compatibility gate unchanged here (or separate an intentional ALTER-to-JSON feature with its own compatibility, constraint, populated-table, rollback, and failure-path tests).

Validation: all new compatibility tests passed under race with repeated counts (bytejson 69x, vector 73x, planner 42x); full owning-package race and vet passed; the head merges cleanly with latest main and focused race tests still pass.

daviszhen:
> The latest update fixes the mixed-version reader crash from my previous review and restores the important existing goldens. I found two remaining blockers in the current head (`4b3ada3489`).
> 
> [P1] Avoid materializing the full decoded BLOB on every comparison
> 
> `StorageCompatible` now correctly persists opaque values through the legacy `TpCodeBlob` envelope, but that also means real values in vectors are base64-backed. `CompareBinaryJSON` calls `binaryJSONValue` for both operands, and `opaquePayload` does `base64.DecodeString(string(...))` for each one (`pkg/container/bytejson/bytejson.go:485-530`). A focused benchmark on this head comparing two equal 1 MiB BLOB JSON values measured 4 allocations and 4,915,201 B/op per comparison. The legal 64 MiB BLOB limit therefore permits roughly 300 MiB of transient heap allocation for one comparison; sorting, joins, JSON_CONTAINS, and JSON_OVERLAPS can repeat it many times. `BinaryJSONPayloadLen` also performs the same full decode merely to obtain a length (`:533-537`). This turns a correctness fix into an OOM/latency path on valid inputs.
> 
> Please compare legacy base64 payloads without materializing both decoded buffers (for example, fixed-size/group-wise decoding), compute decoded length without allocating, and add an allocation regression test or benchmark.
> 
> [P1] Keep expression CAST support from silently broadening ALTER TABLE admission
> 
> The new `init` mutates the global `supportedTypeCast` matrix for every JSON source type (`pkg/sql/plan/function/func_cast.go:846-868`). `checkChangeTypeCompatible` consumes that same matrix for schema changes (`pkg/sql/plan/build_alter_modify_column.go:193-202`), so this PR also changes existing `ALTER TABLE ... MODIFY nth JSON` from a rejection to a successful table rewrite and updates the old BVT to freeze that behavior (`test/distributed/cases/ddl/alter_table_modify_column.result:2014-2020`). This is not required to implement CAST/CONVERT or JSON assignment, and it broadens a high-risk DDL path for all newly added source types without migration/constraint/data-path coverage. The author update also identifies this as the remaining unrelated behavior change.
> 
> Please keep the DDL compatibility gate unchanged here (or separate an intentional ALTER-to-JSON feature with its own compatibility, constraint, populated-table, rollback, and failure-path tests).
> 
> Validation: all new compatibility tests passed under race with repeated counts (bytejson 69x, vector 73x, planner 42x); full owning-package race and vet passed; the head merges cleanly with latest main and focused race tests still pass.

• - 大 BLOB JSON 比较改为分块 Base64 解码比较，避免每次比较完整物化 payload；长度计算同样无完整解码分配。
  - 补充 1 MiB BLOB 分配基准与回归，验证 0 B/op、0 allocs/op；兼容 legacy Base64 的 CR/LF 和 BIT 前缀。
  - 限制 ALTER TABLE ... MODIFY ... JSON：保持改动前 DDL 准入规则，阻止 INT、BINARY 等新增表达式 CAST 来源隐式进入 JSON 列重写。
  - 恢复 BVT 对 INT → JSON 的拒绝断言，并确认失败后 nth 保持 int NOT NULL。

XuPeng-SH (approved):
Re-reviewed the latest head. The previous blockers are resolved: legacy BLOB comparison/length paths now use bounded, allocation-free chunked Base64 decoding, and ALTER TABLE keeps the pre-existing JSON DDL admission contract. Focused bytejson race tests passed; the 1 MiB comparison benchmark reports 0 B/op and 0 allocs/op. No remaining blocker found.

PR #26108, opened 2026-07-23
test: reduce slow race-enabled UT workloads

XuPeng-SH:
## What changed

PR CI runs Go unit tests with `-short -race`. This PR separates correctness coverage from production-scale stress while keeping every original full-scale workload in non-short mode.

- Brute force: 1,000 deterministic vectors in short mode; original 10,000 outside short mode.
- HNSW build: capacity 1,000 in short mode; original 100,000 outside short mode. Both modes still cover rollover to exactly two indexes, concurrent Add, persistence/load, F32/F64, Contains, all keys, and recall.
- HNSW cache: removes a meaningless initial two-second sleep, replaces the blind expiry sleep with an actual eviction assertion, isolates/restores process-global state, and bounds only the short-mode concurrency stress count.
- HNSW empty-index UPSERT: uses 128 CDC entries in short mode and the original 1,000 outside short mode. It still exercises UPSERT-to-INSERT classification, multi-worker dispatch, native index insertion, persistence, and now asserts the exact insert/update counts. Process-global SQL mocks are restored on every exit path.
- Data Branch diff-as-file: bounds three scale-only inputs in short mode while retaining multi-block CSV, composite/no-PK/complex-type, snapshot divergence, output limit, and hash-diff paths. Original scales remain outside short mode.
- Data Branch composite-PK chunk probe: uses four blocks (32,768 rows) in short mode and the original 1,000,000 outside short mode. Both 8,192-key chunks, their cross-chunk ordering, and all result-count assertions are unchanged.
- TAE merge transfer: short mode still merges eight blocks and transfers all 65,536 deleted rows; original 400,000 rows remain outside short mode.
- ChangesHandle: short mode still compacts four blocks and exercises snapshot, tail, tombstone, and stale-file iteration. Assertions now derive and verify the exact data/tombstone totals instead of relying on hard-coded 20-block values.
- Gossip: polls the complete 24-service/shard state once per retry and sleeps once after a failed scan, instead of sleeping once per not-ready service. Node, shard, replica, restart, and epoch checks are unchanged.
- CN S3 delete flush: short mode uses 65,536 rows, which still produces eight blocks and a 2.003 MiB S3 tombstone with 65,535 rows. Full mode retains 524,288 rows. The test now restores the real 5 MiB production default instead of contaminating later tests with 32 MiB.
- Partition S3 LOAD: both primary-key and unique-index cases use four blocks (32,768 rows) in short mode and the original 100,000 outside short mode. Each hash partition still receives multiple blocks; affected rows, partition totals, point lookup, and index totals remain asserted.
- Object chunk reads: the multi-chunk object is now the smallest useful boundary case, 100 MiB + 1 KiB instead of 150 MiB. Sentinels on both sides of the boundary and exact byte comparisons still verify one full chunk plus one partial chunk.
- Rolling-restart lockservice tests: replace silent `ctx.Done() -> return` paths with bounded asserted waits. They verify the exact rolling-restart error, wait on the service that is actually restarted, abort the fenced transaction on `ErrLockTableBindChanged`, retry with a new transaction generation, and require eventual lock success.
- Embed cluster lifecycle: every successfully started standalone cluster registers cleanup immediately. This closes clusters even when an intermediate assertion fails, including both clusters in the multi-cluster test.

Only test files are changed; production code is untouched.

## Why

A recent Ubuntu/x86 `-short -race` run reported these individual costs: brute-force concurrency 611.66s, HNSW build 693.79s total, Data Branch 180.50s, merge transfer 90.09s, ChangesHandle 61.69s and 50.52s, gossip 59.22s, CN S3 delete 57.75s, HNSW cache 34.65s, the composite-PK chunk probe 22.83s, HNSW empty UPSERT 16.49s, partition S3 LOAD 10.19s/9.59s, object chunk read 8.82s, and three rolling-restart lockservice tests 10.18s each.

These tests mixed semantic assertions with scale/stress multipliers. Short mode now preserves the state transitions and boundary conditions needed for PR correctness checks; non-short mode preserves the original stress sizes.

The previous PR head completed the Ubuntu UT job successfully in 29m31s, down from the roughly 35-minute baseline. The third batch targets remaining entries in that run's Top 50.

Two proposed reductions were explicitly rejected during validation:

- `TestLockNeedUpgrade` needs more than 8,192 rows to trigger the real local/remote row-to-table lock upgrade; its original 20,000 rows and four independent commits remain unchanged.
- `TestDeleteAndSelect` uses an internal SQL executor that does not support `DROP DATABASE`; the attempted cleanup was removed. Its existing 24,576 rows still prove a three-block scan.

After reviewing the rest of the supplied slow-test list, no change was made to DataBranchPick, the fulltext snapshot race regression, prepared-statement second execution, partition affected-row cases, abort-all lifecycle timing, logservice bootstrap recovery, WAL recovery, PITR, ISCP, HAKeeper repair, or the Iceberg SQL matrices: their scenarios or timing are the coverage, so reducing them would weaken the tests.

## Validation

- Rebasing was converted to a protected-branch-compatible merge; the final tree is identical to the reviewed rebased tree and contains current `main` (`f74015d952`).
- `go vet` passes for all 10 affected packages.
- Final compile/link-only pass under `-short -race` passes for all 10 affected packages.
- Previous-head CI: Ubuntu UT, Linux/arm64 SCA, CodeQL, Iceberg E2E, and requested BVT jobs passed.
- Focused short race results for the third batch:
  - Rolling-restart lockservice tests: 6.68s combined; all three also pass with `-count=3` (17.55s).
  - HNSW empty UPSERT: 0.01s test time; `-count=3` package run 1.57s. Logs prove all 128 UPSERTs became inserts.
  - Composite-PK chunk probe: 19.65s package run; both 8,192-key chunks and exact result counts pass.
  - Partition S3 LOAD and S3 LOAD with unique index: 16.56s combined package run at four blocks.
  - Object chunk boundary: 3.22s package run.
  - MultiCluster cleanup/lifecycle: 35.19s package run.
- Non-short original-scale passes for the third batch:
  - HNSW empty UPSERT: 1.57s test time with all 1,000 entries.
  - Composite-PK chunk probe: 22.32s with 1,000,000 base rows.
  - Partition S3 LOAD cases: 20.24s combined with 100,000 rows each.
- Earlier focused short race results:
  - Data Branch diff-as-file: 23.63s.
  - CN S3 delete: 14.04s; log proves `CN-COMMIT-S3-Tombstone`, 2.003 MiB, 65,535 rows.
  - Gossip: 9.14s.
  - Transfer merge: 2.09s; logs prove eight blocks and 65,536 transferred rows.
  - Both ChangesHandle cases: 6.03s combined package run.
  - HNSW cache: 0.31s.
- Earlier non-short original-scale passes:
  - Data Branch diff-as-file: 15.71s.
  - CN S3 delete: 15.67s with 524,287 S3 tombstone rows.
  - HNSW cache: 6.00s.
  - Transfer merge: 1.22s.
  - ChangesHandle: 0.86s and 0.71s.
  - Brute force: 4.085s.
  - HNSW build: 100.283s with both 100,000-entry indexes and recall 1.0.


pkg/vm/engine/test/publication_test.go
```diff
@@ -1661,15 +1661,18 @@ func TestGetObjectChunk(t *testing.T) {
...
-	// Test case 2: Large file (> 100MB) - should have multiple chunks
-	// Create a file larger than 100MB (e.g., 150MB)
-	const chunkSize = 100 * 1024 * 1024       // 100MB
-	largeFileSize := int64(150 * 1024 * 1024) // 150MB
+	// Test case 2: the smallest file beyond the 100 MiB boundary must be read
+	// as one full chunk plus one partial chunk.
+	const chunkSize = 100 * 1024 * 1024
+	largeFileSize := int64(chunkSize + 1024)
 	testObjectName2 := "test_object_large"
 	largeContent := make([]byte, largeFileSize)
```
XuPeng-SH, line 1669: [P1] Preserve non-zero payload coverage across the whole object. With this initialization, every byte except the four sentinels is zero. Both the `IOEntry.Data` buffer and `ReadObjectFromEngine` result are zero-initialized, so a regression that skips reading/copying any interior range can still satisfy the full `require.Equal`; the old `byte(i % 256)` payload caught that. The sentinels validate only four offsets, not exact content across either chunk. Please keep a non-zero repeating payload while avoiding 100 MiB of race-instrumented scalar writes—for example, initialize a 256-byte seed and fill the slice with exponentially growing `copy` calls.

XuPeng-SH: Fixed in 2fea0a7e32: seed a 251-byte non-zero pattern and expand it with logarithmic bulk copies, retaining whole-object content/offset coverage without restoring the 100 MiB scalar-write cost. Verified with `TestGetObjectChunk -short -race -count=3`.

PR #26110, opened 2026-07-23
fix(frontend): close branch executor on begin failure

### LINKED ISSUE #26089: [Bug]: Data Branch background executor is not closed when BEGIN fails
## Description

The background executor created by `getBackExecutor` is not closed when its initial `BEGIN` statement returns an error.

`getBackExecutor` returns `(nil, cleanup, err)` on this path. Data Branch and Clone callers immediately return when `err != nil`, so they never register or invoke the returned cleanup function.

## Environment

- Branch: `main`

## Reproduction

Use a mocked `BackgroundExec` with these expectations:

1. `ClearExecResultSet()` succeeds.
2. `Exec(ctx, "begin")` returns an error.
3. `Close()` must be called once.

Then call:

```go
returned, cleanup, err := getBackExecutor(context.Background(), ses)
```

The focused test was repeated three times:

```text
missing call(s) to *mock_frontend.MockBackgroundExec.Close()
--- FAIL: TestAuditGetBackExecutorClosesWhenBeginFails
```

## Actual result

`getBackExecutor` returns the original error and a cleanup function, but the executor remains open because all callers return immediately on the error.

## Expected result

`getBackExecutor` should close the executor itself before returning an initialization error. A cleanup callback should only be returned when ownership has successfully transferred to the caller.

## Affected paths

The helper is shared by Data Branch create/delete/diff/merge/pick and Clone paths.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26089

When the standalone Data Branch background executor fails its initial `BEGIN`, close it immediately before returning the original error. The failed initialization now returns neither an executor nor a cleanup callback, so ownership cannot be lost between the helper and its callers.

The explicit-transaction path and the successful standalone transaction cleanup path are unchanged.

Validation:

- `go test ./pkg/frontend -run '^TestGetBackExecutorClosesWhenBeginFails$' -count=3`


XuPeng-SH (approved):
Reviewed the full diff at e7fb0fda6c48f1023d1d6e8ec17907b441738d1d.

Verified the executor ownership transfer across standalone BEGIN failure, standalone success cleanup, and shared explicit-transaction paths. The failure path now closes exactly once, returns no executor/cleanup handle, and preserves the original BEGIN error. Focused race stress (20 runs), the full frontend package tests, build, and go vet all pass; CI is green. No blocking issues found.

PR #26112, opened 2026-07-23
fix(frontend): encode data branch output names

### LINKED ISSUE #26104: [Bug]: DATA BRANCH DIFF OUTPUT FILE mishandles quoted table names containing slashes
### Summary

`DATA BRANCH DIFF ... OUTPUT FILE` uses quoted table names directly when it builds the generated file name and the returned operation hint. A legal table name containing `/` is therefore split into extra directories, while the returned hint contains an unquoted and invalid table reference.

### Version

Latest `main`:

```text
2eb7f523e5a10f352d3e258183ed8440510388de
```

### Reproduction

```sql
CREATE DATABASE g4d_output_file_names;
USE g4d_output_file_names;

CREATE TABLE `base/name`(id INT PRIMARY KEY, v VARCHAR(32));
INSERT INTO `base/name` VALUES (1, 'base'), (2, 'same');
DATA BRANCH CREATE TABLE `child/name` FROM `base/name`;
UPDATE `child/name` SET v = 'child' WHERE id = 1;
INSERT INTO `child/name` VALUES (3, 'new');

DATA BRANCH DIFF `child/name` AGAINST `base/name`
  OUTPUT FILE '/private/tmp/g4d_file_name_output';
```

The result is stable in 3 consecutive executions.

### Actual result

The returned file path is split into nested directories:

```text
/private/tmp/g4d_file_name_output/diff_child/name_base/name_20260723_111809.sql
```

The returned hint also contains an unquoted table reference:

```text
DELETE FROM g4d_output_file_names.base/name,
INSERT INTO g4d_output_file_names.base/name
```

The SQL statements inside the generated file quote the table correctly, so the behavior is limited to generated file naming and the result hint.

### Expected result

- The generated file should be created directly under the requested output directory with an encoded or sanitized flat file name.
- The returned hint should quote database and table identifiers so that legal quoted names remain unambiguous and executable.

### Code locations

- `pkg/frontend/data_branch_output.go`: `makeFileName`
- `pkg/frontend/data_branch_output.go`: `prepareFSForDiffAsFile`

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26104

Encode Data Branch DIFF table and snapshot name components before using them in generated SQL/CSV filenames. Safe ASCII and printable Unicode retain their spelling, while path separators and other unsafe bytes become uppercase `@HH` sequences so the filename remains one path segment without being reinterpreted as a CSV printf pattern.

Cap generated stems at 240 bytes. Overlong names retain a UTF-8-safe prefix plus a 128-bit digest and timestamp, keeping the final path component within common filesystem limits.

The SQL file hint now quotes the resolved database and table identifiers, including embedded backticks.

Validation:

- `go test ./pkg/frontend -run '^TestDataBranchOutput(FileNameAndHintQuotePathSeparators|FileNameSurvivesCSVFormattingAndLengthLimit|MakeFileName)$' -count=3`


XuPeng-SH (approved):
Reviewed the complete diff at 6f88ea01d8. The filename encoding closes path-separator and CSV-format reinterpretation risks, truncation preserves UTF-8/escape boundaries with a digest suffix, and the SQL hint now quotes identifiers correctly. Targeted frontend regressions passed 3 consecutive runs; CI is green. No blocking issues found.

PR #26116, opened 2026-07-23
fix(frontend): keep stream pop count on callback error

### LINKED ISSUE #26081: [Bug]: BranchHashmap ItemCount remains stale after PopByVectorsStream callback error
## Description

`BranchHashmap.PopByVectorsStream` removes matching in-memory entries before invoking its callback. If the callback returns an error, the method returns immediately without applying `removedTotal` to `shard.items`.

The removed row is no longer returned by `GetByVectors`, but `ItemCount()` still includes it. This leaves the hashmap's public count inconsistent with its actual contents after an error in a real Data Branch callback such as row decoding or batch construction.

## Reproduction

1. Create a one-shard `BranchHashmap`.
2. Insert one row through `PutByVectors` and verify `ItemCount() == 1`.
3. Call `PopByVectorsStream` for that key with a callback that returns an error.
4. Query the key with `GetByVectors` and read `ItemCount()`.

This was reproduced in 3 consecutive runs with a focused unit test.

## Actual result

- `PopByVectorsStream` returns the callback error.
- `GetByVectors` reports that the row no longer exists.
- `ItemCount()` incorrectly remains `1`.
- The returned removed count is `0`, even though the row has already been removed.

## Expected result

The bookkeeping and returned removal count should remain consistent with the entries already removed before the callback error is returned, or the removal should be rolled back.

## Code path

`pkg/frontend/databranchutils/branch_hashmap.go`, `PopByVectorsStream`: the callback error branch returns before `atomic.AddInt64(&shard.items, -removedTotal)` and before adding `removedTotal` to `totalRemoved`.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26081

Finalize each `PopByVectorsStream` shard through one accounting-and-unlock path. If a callback fails after rows have already been removed, those rows now decrement `ItemCount()` and are included in the returned removal count before the original callback error is returned.

Add regressions for both in-memory removal and an actual spilled-row tombstone, while retaining the existing success-path coverage.

Validation:

- `go test ./pkg/frontend/databranchutils -run '^TestBranchHashmapPopByVectorsStream(InMemory|Spilled)(CallbackError)?$' -count=3`
- `go test -race ./pkg/frontend/databranchutils -run '^TestBranchHashmapPopByVectorsStream(InMemory|Spilled)(CallbackError)?$' -count=3`


XuPeng-SH (changes requested):
Found one blocking unhappy-path consistency issue at head c84a977cab1319b4bcdf08ba127728ddda461f60.

[Q2] `PopByVectorsStream` still loses callback-undelivered in-memory rows, and its failure semantics depend on whether the rows are resident or spilled. `shard.mem.collect(...)` removes every matching row before the callback loop starts. With two rows for one key and `removeAll=true`, the new test returns `removed == 2` after the callback was invoked only once and failed; the second payload has been deleted without ever being presented to the callback. The test currently codifies that loss. By contrast, `collectSpillPopStream` tombstones and invokes the callback one entry at a time, so the same first-callback failure removes only the first spilled entry.

Please make removal/delivery ordering consistent across memory and spill (for example, remove and invoke per row, stopping before deleting later rows, or restore rows not delivered after an error), then add the same-key multi-row callback-error truth table for both storage tiers. The returned count, ItemCount, callback invocation count, and remaining rows should agree in both cases. This also addresses the latest #26081 comment rather than only repairing the counter after data has already been lost.

Validation on the exact head: both new callback-error tests passed under race individually and with adaptive `-count=19`; the full `pkg/frontend/databranchutils` package also passed under race. The tests confirm the accounting fix, but the in-memory expectation itself demonstrates the blocker above.

VioletQwQ-0:
Addressed the current-head callback-error consistency finding in `56a8e50f694038a16b0a4a5d2e358196746b539e`.

The in-memory path now removes and accounts for one row immediately before each callback, matching the spill path. On the first callback error, later same-key rows remain available. The regression matrix now asserts callback count, returned removal count, `ItemCount`, and the identity of the retained undelivered row for both memory and forced-spill backends.

Validation:

- focused memory/spill callback-error pair: `-race -count=20` PASS
- full `pkg/frontend/databranchutils`: `-race -count=1` PASS
- `pkg/frontend` consumer compile under `-race`: PASS
- schema v4 CR preflight: review PASS; fresh remote CI pending

XuPeng-SH (approved):
Deep review completed on exact head 56a8e50f694038a16b0a4a5d2e358196746b539e. The previous callback-error consistency blocker is resolved: in-memory and spill paths now remove/account one delivered row at a time and preserve later undelivered rows. Focused adaptive race stress, full owning-package race/non-race tests, build/vet, and a frontend consumer regression passed.

PR #26117, opened 2026-07-23
fix(frontend): reset PK segment gap statistics

### LINKED ISSUE #26084: [Bug]: DATA BRANCH PICK PK filter keeps stale gap statistics after a segment split
## Description

The adaptive PK-filter `segmentBuilder` used by DATA BRANCH PICK does not reset `gapSum` and `gapCount` when it starts a new segment. Gap decisions for the next cluster therefore keep using the previous cluster's average.

For two internally uniform numeric clusters, the first large jump correctly creates a second segment, but every regular value in the second cluster can then be split into its own single-value segment. This inflates the `engine.PKFilter` segment list and reduces the intended batching/pruning efficiency.

## Reproduction

Feed the `int64` segment builder this sorted sequence:

```text
1, 2, 3, 4, 5,
1000000, 1000100, 1000200, 1000300, 1000400
```

The first cluster has a uniform gap of 1. The second cluster has a uniform gap of 100. The jump from 5 to 1000000 should be the only adaptive split.

This was reproduced in 3 consecutive focused unit-test runs.

## Actual result

`finalize()` returns 6 segments:

- one segment for `[1, 5]`
- five single-value segments for the second cluster

After the first split, the stale average gap remains 1, so every gap of 100 is treated as another large inter-cluster gap.

## Expected result

The builder should return 2 segments, one for each internally uniform cluster. Starting a new segment should reset `gapSum` and `gapCount` so its decisions use that segment's own observations.

## Code path

`pkg/frontend/data_branch_pick.go`, `segmentBuilder.observe`: the `shouldSplit` branch calls `flushSegment` and resets `curMin`, `curMax`, and `curCount`, but does not reset `gapSum` or `gapCount`.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26084

Reset the adaptive PK segment builder's running gap sum and sample count whenever a split starts a new segment. Gap decisions in the new cluster now use only that cluster's observations instead of inheriting the previous segment's average.

Add a regression with two uniform `int64` clusters whose internal gaps differ, verifying that the builder produces exactly one ZoneMap segment per cluster with the correct bounds.

Validation:

- `go test ./pkg/frontend -run '^(TestSegmentBuilder_|TestBuildPKFilterFromVec_ProducesValidFilter$)' -count=3`
- `go test -race ./pkg/frontend -run '^TestSegmentBuilder_' -count=3`


XuPeng-SH (approved):
Reviewed the complete diff and the segment-builder lifecycle. Resetting `gapSum` and `gapCount` at the same transition that installs the new segment is the correct generation boundary: the inter-cluster gap is excluded, subsequent decisions use only the new segment's observations, and both adaptive and hard-cap splits preserve sorted/non-overlapping ZoneMap coverage. No concurrency, resource-lifecycle, compatibility, or correctness blocker found.

Fresh verification on PR head `dba7c3634`:
- focused regression/hard-cap tests: `-count=10` passed
- all segment-builder tests under `-race -count=10` passed
- `go vet ./pkg/frontend` passed
- `go build ./pkg/frontend` passed
- required CI checks are green.

PR #26119, opened 2026-07-23
fix(frontend): reject hashmap inserts after close

### LINKED ISSUE #26082: [Bug]: BranchHashmap can retain a late allocation when Close races with PutByVectors
## Description

`BranchHashmap` documents its exported APIs as safe for concurrent use, but `PutByVectors` only checks `closed` while initializing or validating its types. The method releases `metaMu` before allocation and insertion.

If `Close` runs while a `PutByVectors` call is paused in allocation, `Close` marks the hashmap closed, cleans all shards, and returns. When the allocation resumes, `PutByVectors` inserts the entry into the already-closed shard and returns success. A repeated `Close` is a no-op, so the late allocation is never released.

## Reproduction

A focused unit test uses a deterministic blocking allocator:

1. Create a `BranchHashmap` with the blocking allocator.
2. Start `PutByVectors` and pause it inside `Allocator.Allocate`.
3. Call `Close` and wait for it to return.
4. Release the allocator and wait for `PutByVectors`.
5. Call `Close` again and inspect the allocator's retained byte count.

This was reproduced in 3 consecutive runs.

## Actual result

- The first `Close` returns successfully.
- The resumed `PutByVectors` returns `nil` and inserts after cleanup.
- The second `Close` returns immediately because `closed` is already true.
- The allocation remains retained (`12` bytes in the minimal test).

## Expected result

`Close` should synchronize with in-flight exported operations, or an operation that resumes after closing begins should fail without inserting and release any allocation it obtained.

## Code path

- `PutByVectors` / `ensurePutTypes` in `pkg/frontend/databranchutils/branch_hashmap.go`
- `branchHashmap.Close` sets `closed` before closing shards, but there is no in-flight operation guard or second state check before insertion.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26082

Recheck `BranchHashmap.closed` under `metaMu.RLock` immediately before a prepared allocation is inserted into any shard. If `Close` already won the race, release every reference in the shared entry block, return the existing closed error, and insert nothing. If insertion has already reached this barrier, `Close` waits and then cleans the inserted rows.

Apply the same barrier before allocator fallback can spill, so a nil allocation resumed after `Close` cannot create new spill stores or directories that a repeated close would skip.

Add a deterministic blocking-allocator regression that pauses `PutByVectors`, completes `Close`, resumes allocation, and verifies the put fails with zero retained allocator bytes even after a repeated close.

Validation:

- `go test ./pkg/frontend/databranchutils -run '^TestBranchHashmapPut(RejectsAllocationCompletedAfterClose|DoesNotSpillAfterClose)$' -count=3`
- `go test -race ./pkg/frontend/databranchutils -run '^TestBranchHashmapPut(RejectsAllocationCompletedAfterClose|DoesNotSpillAfterClose)$' -count=3`
- `go test -race ./pkg/frontend/databranchutils -count=1`


XuPeng-SH (approved):
Deep review complete. The Close/Put linearization is sound: a close-first path rejects and releases the entire entry block, while a flush-first path holds metaMu.RLock through shard insertion so Close waits and subsequently cleans it. The spill fallback is covered by the same generation barrier. Q1/Q2/Q3 lifecycle checks found no leak, deadlock cycle, or unbounded retained state. Verified locally with focused race stress (-count=20), full databranchutils tests with and without -race, go build, go vet, and git diff --check.

PR #26129, opened 2026-07-24
fix: handle decimal constant vectors in batch casts

### LINKED ISSUE #25975: [Bug]: DECIMAL column plus integer literal in non-equi JOIN can panic
### Summary

A non-equi JOIN predicate that adds an integer literal to a `DECIMAL` column can panic in `decimal64ToDecimal128Array`. The equivalent expression with a decimal literal or an explicit decimal cast succeeds.

### Tested version

Latest `main` at `3046f14d94b3f1a25839eeeb65729b64f85ac647`.

### Reproduction

```sql
DROP DATABASE IF EXISTS decimal_join_panic;
CREATE DATABASE decimal_join_panic;
USE decimal_join_panic;

CREATE TABLE l(id INT, v DECIMAL(10,2));
CREATE TABLE r(id INT, v DECIMAL(10,2));
INSERT INTO l VALUES (1,10.00),(2,20.00);
INSERT INTO r VALUES (11,2.00),(12,15.00);

SELECT l.id, r.id
FROM l RIGHT JOIN r ON l.v <= r.v + 8
ORDER BY r.id, l.id;
```

### Actual result

```text
ERROR 20101 (HY000): internal error: panic runtime error: index out of range [1] with length 1
...
github.com/matrixorigin/matrixone/pkg/sql/plan/function.decimal64ToDecimal128Array
    pkg/sql/plan/function/func_cast.go:5058
```

The failure reproduces 3/3 times. A larger data set also reproduces the same panic with an `INNER JOIN` predicate:

```sql
SELECT l.id, r.id
FROM l JOIN r ON l.v BETWEEN r.v AND r.v + 8;
```

### Expected result

The query should return the matching rows without an internal panic. MySQL 8.0.45 returns:

```text
1  11
1  12
2  12
```

### Additional observations

Both of these equivalent predicates succeed in MatrixOne:

```sql
ON l.v <= r.v + 8.00
ON l.v <= r.v + CAST(8 AS DECIMAL(10,2))
```

The panic occurs in the fast path at `func_cast.go:5058`, where the requested batch `length` can be greater than the underlying one-element constant vector used by the implicit `DECIMAL64` to `DECIMAL128` conversion.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #25975

A non-null DECIMAL64 constant vector stores one physical value while exposing
the logical batch length. `decimal64ToDecimal128Array` indexed that physical
slice using every logical row index, which could panic when an integer literal
was promoted during a non-equi join.

This PR detects constant source vectors, converts the scalar once while
preserving sign, width, and scale semantics, and fills the preallocated
DECIMAL128 result batch with the converted value. The existing flat-vector fast
path remains unchanged.

It also adds:

- unit coverage for positive and negative constants, scale changes, and width
  narrowing;
- distributed regression coverage for RIGHT JOIN and BETWEEN non-equi join
  predicates.

Validation:

- focused CGo unit tests, including reproduction of the old out-of-range panic;
- full `pkg/sql/plan/function` test suite;
- `go build ./pkg/sql/plan/function`;
- `go vet ./pkg/sql/plan/function`;
- `make static-check`;
- `decimal.test`: 625/625 statements passed;
- `decimal64ToDecimal128Array` function coverage: 80.2%.


XuPeng-SH (approved):
Reviewed exact head `17b8409508ec9fdcdf7ed2f869e5c941a6c0074a`.

The const-vector branch fixes the physical-length/logical-length mismatch without changing the flat-vector path. Its width narrowing, scale conversion, and signed Decimal64-to-Decimal128 extension preserve the existing per-row semantics; const NULL and empty-result paths remain covered by the existing wrapper contract. The distributed RIGHT JOIN and BETWEEN regressions exercise the reported planner/execution shape.

Local verification passed with real exit codes:
- focused const-vector regression, `-count=100`;
- focused Decimal64→Decimal128 paths under `-race`;
- full `pkg/sql/plan/function` test suite;
- package build and vet;
- `git diff --check`.

No blocking findings.

heni02 (approved):

PR #26134, opened 2026-07-24
fix(lockservice): harden MORPC liveness under keeper bursts

### LINKED ISSUE #26090: [Bug]: lockservice 10s backend read timeout cancels valid long-running lock RPCs and all shared streams

Related issues #24893 and #25129 report the same surface error under memory/GC pressure, but they do not cover the timeout mismatch and connection-wide cancellation mechanism described here. #25779 covers orphan classification after a backend closes, which is a downstream but different failure mode.

## Branch Name

`main`

## Commit ID

`abd9a22d6b61a2fa2e399d14a9687591a7290416` (squash commit built by the failed check-in run; workflow head SHA: `b5bae0581216ef21eaedcaece95f7ee1753b9c17`)

- Workflow: `MO Checkin Regression On TKE`
- Run: https://github.com/matrixorigin/matrixone/actions/runs/29981343821
- Failed job: https://github.com/matrixorigin/matrixone/actions/runs/29981343821/job/89133250688
- Namespace: `mo-checkin-regression-26018`
- Source CN: `mo-checkin-regression-tp-cn-6zkn9` (`10.142.41.207`)
- Lock-table owner CN: `mo-checkin-regression-tp-cn-jgdq7` (`10.142.41.208`)
- Target CN resources: 7 CPU / 14 GiB
- Concurrent workloads: SYSBENCH mixed workload (three 30-thread groups), TPCC 50 warehouses / 50 terminals, and SSB 10 GB
- Failure time: `2026-07-23 06:22:57 UTC`

## Actual Behavior

A SYSBENCH transaction on the source CN waited for a remote row lock owned by the target CN. After approximately 10 seconds without a response, the lockservice MORPC backend read timed out:

```text
cannot connect to backend: writeto tcp4 10.142.41.207:56664->10.142.41.208:6003:
read tcp4 10.142.41.207:56664->10.142.41.208:6003: i/o timeout
```

The target CN was not down. It remained alive and detected a deadlock less than one second later. Its response then failed because the source CN had already closed the shared connection:

```text
2026-07-23 06:22:57.398 source cn-service.lock-client: read from backend failed
2026-07-23 06:22:57.398 source cn-service.lock-client: read loop stopped
2026-07-23 06:22:58.263 target cn-service.lockservice: dead lock found
2026-07-23 06:22:58.266 target: write response failed: use of closed network connection
```

The timed-out request was a row lock for SYSBENCH transaction `3a0a266f55072c7f18c4d3affba820c7`, table ID `345141` (`sbtest.sbtest10`), while executing an `INSERT INTO sbtest10 ...` statement. The request started at `06:22:47.392` and failed at `06:22:57.428`.

Because MORPC treats this backend read timeout as a connection failure, it calls `cancelActiveStreams()` and resets the backend. One slow lock request therefore failed unrelated in-flight lockservice operations sharing the connection. Prometheus recorded:

| Metric / request | Count |
| --- | ---: |
| backend read timeout | 1 |
| backend `wait_response` errors | 612 |
| remote `Lock` timeouts | 48 |
| remote `GetBind` timeouts | 14 |
| remote `GetTxnLock` timeouts | 2 |
| remote `Unlock` timeouts | 4 |

The target CN was heavily loaded (about 6.4-6.7 CPU cores, S3 reads taking 9-10.4 seconds, and about 127-129 MB/s network receive), which amplified response latency. However, there was no OOM, restart, or process loss. The direct failure mechanism was the fixed 10-second transport read timeout closing a healthy shared connection during a lock wait that the lockservice configuration permits to run much longer.

Observed lockservice durations also exceeded the transport timeout: lock acquire wait reached about 16.18 seconds and lock hold duration reached about 101.38 seconds.

## Expected Behavior

- A legitimate remote lock wait or deadlock-detection cycle longer than 10 seconds must not be classified as a dead backend while the request's own deadline is still valid.
- One slow lock request must not close the shared backend and fail unrelated `Lock`, `GetBind`, `GetTxnLock`, and `Unlock` requests.
- Transport liveness should distinguish a dead connection from a healthy connection carrying a long-running lock request.
- The effective transport timeout must be consistent with the lockservice request deadlines, or long-wait RPCs must use an isolated/request-scoped mechanism.

## Steps to Reproduce

1. Deploy MatrixOne with at least two CNs.
2. Run concurrent cross-CN SYSBENCH/TPCC workloads that create a remote row-lock wait longer than 10 seconds. A concurrent scan-heavy workload can be used to reproduce the scheduling/I/O pressure from this run.
3. Ensure the originating transaction runs on CN A while the lock table is owned by CN B.
4. Observe CN A's lock-client connection to CN B port 6003 when no response is produced for 10 seconds.
5. The MORPC backend read loop reports `i/o timeout`, closes/resets the shared backend, and cancels all active streams.
6. CN B later completes deadlock detection or the lock request, but writing the response fails because CN A already closed the connection.

### Timeout mismatch

| Layer | Current value | Meaning |
| --- | ---: | --- |
| lockservice MORPC backend read timeout | 10 seconds | Connection-level inactivity/read timeout |
| remote lock owner wait timeout | 2 minutes | Owner-side remote lock wait bound |
| maximum lock wait duration | 1 hour | Request/waiter deadline |
| remote lock RPC slack | 30 seconds | Extra RPC deadline beyond lock deadline |

Relevant code:

- [`pkg/lockservice/rpc.go`](https://github.com/matrixorigin/matrixone/blob/abd9a22d6b61a2fa2e399d14a9687591a7290416/pkg/lockservice/rpc.go#L41-L45): `defaultRPCTimeout = 10s`
- [`pkg/lockservice/rpc.go`](https://github.com/matrixorigin/matrixone/blob/abd9a22d6b61a2fa2e399d14a9687591a7290416/pkg/lockservice/rpc.go#L123-L127): configures `WithBackendReadTimeout(defaultRPCTimeout)`
- [`pkg/common/morpc/backend.go`](https://github.com/matrixorigin/matrixone/blob/abd9a22d6b61a2fa2e399d14a9687591a7290416/pkg/common/morpc/backend.go#L687-L715): read failure cancels all active streams and resets the connection
- [`pkg/lockservice/cfg.go`](https://github.com/matrixorigin/matrixone/blob/abd9a22d6b61a2fa2e399d14a9687591a7290416/pkg/lockservice/cfg.go#L24-L34): 2-minute owner timeout and 1-hour maximum lock wait
- [`pkg/lockservice/lock_table_remote.go`](https://github.com/matrixorigin/matrixone/blob/abd9a22d6b61a2fa2e399d14a9687591a7290416/pkg/lockservice/lock_table_remote.go#L45-L64): request deadline plus 30-second RPC slack

### Change history

The 10-second connection read timeout is not recent: commit `c39fda3b4` (2023-09-27, #11920) changed it from 3 seconds to 10 seconds and enabled `WithBackendReadTimeout`.

Recent lock-wait changes are all present in the failed commit but did not update the connection-level timeout:

- `7ad7f648f` (2026-06-26, #24885): added 30-second lock RPC slack.
- `826d7ee22` (2026-07-05, #25447): added the 2-minute remote lock owner wait timeout.
- `a327eaa03` (2026-07-21, #25804): added the 1-hour maximum lock wait and propagated an absolute deadline.
- `78d15d65c` (2026-07-21, #25901): made lock waits context-aware.
- `9dac43ad7` (2026-07-22, #25999): added a bounded remote wait-for snapshot query.

These changes are not proven to have introduced the regression; they expose or leave in place a pre-existing inconsistency between request-level lock deadlines and the shorter connection-level read timeout.

Potential fix directions include heartbeat-based connection liveness, per-request timeout handling that does not poison the backend, isolating long-wait lock RPCs, or aligning the connection timeout with the maximum effective request deadline.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26090

This PR closes the failure chain where periodic lock keeper fan-out, MORPC send/flush timeout semantics, and heartbeat sharing the data TCP could turn one stalled connection into hundreds of simultaneous Future failures.

### Keeper traffic

- Bounds global keeper work to 64 in-flight RPCs with completion-driven fair scheduling.
- Persists peer, bind, and refresh cursors so slow peers cannot monopolize the bounded window.
- Adds interval jitter and aggregates same-round failures.
- Uses a dedicated keeper MORPC client and physical connection.
- Applies one bounded round context across keeper RPC, discovery, cluster refresh, and bind refresh.

### MORPC timeout and ownership semantics

- Uses the earliest absolute deadline for client and server batch flushes.
- Adds explicit request ownership transfer for clients that must return when context expires before send completion.
- Makes late callbacks and pooled request/response cleanup exactly-once.

### Independent liveness and generation isolation

- Adds a ping/pong-only control client isolated from data backends, TCP, writer, read loop, breaker, and reconnect generation.
- Does not probe idle, filtered, queued-expired, or stream-only traffic.
- Tracks unmatched unary writes by request ID using monotonic time; the set is capped by backend buffer size with a sticky overflow latch.
- On a successful control probe, seals the stalled data generation as draining and creates a replacement on new demand without replaying old RPCs.
- On a failed probe, resets the current data generation and fails its Futures through the existing bounded connection-error path.
- Fences old callbacks and reset/create work from replacement generations.
- Fixes the deterministic goetty raw socket double-close during repeated replacement.

MORPC stream request and response sequences are not correlated, so stream writes are intentionally excluded from unary liveness tracking and remain governed by stream context and lifecycle semantics.

The independent investigation and full local validation evidence are recorded in #26090: https://github.com/matrixorigin/matrixone/issues/26090#issuecomment-5066965745

### Validation

- `go test -count=1 -timeout=120s ./pkg/common/morpc/...`
- `go test -count=1 -timeout=240s ./pkg/lockservice/...`
- Focused MORPC concurrency/lifecycle tests: `-race -count=20`
- Focused lockservice concurrency tests under race
- `go vet ./pkg/common/morpc/... ./pkg/lockservice/...`
- `go build ./pkg/common/morpc/... ./pkg/lockservice/...`
- Independent post-rebase Q1-Q3 review: no blocker

The branch was rebased onto current `origin/main`; `git range-diff` confirms the patch remained unchanged.


XuPeng-SH (changes requested):
The liveness/ownership closure looks solid, and the new concurrency tests pass locally (including targeted `-race`), but the keeper scheduler introduces a blocking hot-path allocation regression.

`doKeepRemoteLock` still accepts/returns `binds` as scratch, but line 172 creates `allBinds` from `cap(binds)` and then returns the original empty `binds`. Because the new implementation never appends to `binds`, its capacity stays zero, so every 1-second keeper round reallocates the full O(number of remote binds) buffer. After sorting that buffer, lines 271-279 copy every bind again into per-peer slices; the already service-ID-sorted `allBinds` could instead be partitioned into contiguous peer ranges without this second O(N) copy/map build. This changes the old amortized scratch reuse into steady allocation/GC pressure, while also adding O(N log N) sorting to a periodic path.

Please reuse the caller-owned scratch (or keep equivalent persistent scratch), build peer work from contiguous ranges of the sorted buffer instead of copying all binds again, and add an allocation/benchmark test at representative high bind cardinality so this does not regress. The scheduling/fairness behavior should remain unchanged.

aptend:
Addressed the keeper hot-path allocation regression in commit 3167be66a.

Changes:
- Reuse the caller-owned bind backing array across keeper rounds, including capacity growth handoff.
- Partition the already ServiceID-sorted binds into contiguous peer ranges; removed the per-peer bind copies, peer map, second peer sort, and physical rotation copies.
- Preserve peer and bind fairness with logical cursor indexing.
- Reuse peer scratch and clear stale descriptor/bind tails on shrink or empty rounds, so old high-water generations and LockTable pointer fields are not retained.
- Return before allocating bounded refresh/failure containers on empty rounds.

Regression coverage:
- Insufficient-capacity growth followed by backing reuse.
- Zero-allocation collection with sufficient scratch.
- Stale LockTable string/byte-slice tail cleanup and peer scratch cleanup.
- 128 slow peers across three deadline-limited rounds, covering partial start, continuation, and wrap.
- 10k binds / 100 peers benchmark: approximately 2.08 ms/op, 0 B/op, 0 allocs/op.

Validation:
- Full pkg/lockservice: PASS (105.655s)
- Focused race suite: PASS, count=10 locally; independent reviewer repeated count=20
- pkg/common/morpc full: PASS
- go vet pkg/common/morpc pkg/lockservice: PASS
- Independent Q1-Q3 correctness/performance re-review: NO BLOCKER.

aptend:
Rebased onto current origin/main (e1453242d). The performance-fix commit is now 4e4b5051f; range-diff reports both PR patches unchanged. Fresh post-rebase full tests passed: pkg/common/morpc 13.523s and pkg/lockservice 106.397s.

XuPeng-SH (approved):
Deep re-review completed on 4e4b5051f after the prior REQUEST_CHANGES. The blocking keeper allocation/copy regression is resolved.

Verified end-to-end:
- caller-owned bind backing storage is reused across rounds and stale bind/peer scratch tails are cleared;
- sorted binds are partitioned into contiguous peer ranges without per-peer bind copies;
- global keeper in-flight work is capped at 64 with completion-driven peer/bind/refresh cursor fairness;
- MORPC request ownership transfers only after enqueue, late callbacks and abandoned responses have exactly one effective cleanup path;
- independent control and keeper transports isolate their TCP/writer/read-loop/reconnect failure domains;
- successful probes drain and fence only the stalled data generation, preserve old Futures, and bound replacement growth to one extra generation; failed probes use the existing bounded reset path;
- earliest absolute batch deadlines are applied on both client and server flush paths.

Fresh local validation at the reviewed head:
- focused MORPC lifecycle/concurrency tests: PASS under -race -count=3;
- focused lockservice keeper/refresh/reset tests: PASS under -race -count=3;
- full ./pkg/common/morpc/... and ./pkg/lockservice/...: PASS;
- go vet and go build for both package trees: PASS;
- 10k binds / 100 peers benchmark (5 runs): 1.61-1.68 ms/op, 0 B/op, 0 allocs/op.

No remaining correctness, liveness, ownership, boundedness, compatibility, or performance blocker found.

aptend:
The failed Ubuntu UT was a real but pre-existing frontend race, unrelated to MORPC/lockservice: Session.reset copied a time.Location value into the new session default pointer, which points at process-global time.Local. Concurrent CDC test logging read the same Location internals.

Fixed in 14fb38e68 by inheriting the immutable Location pointer instead of mutating its value, with a deterministic pointer-identity regression test.

History: the faulty value copy was introduced by e8deb5c29 (#16828) on 2024-08-15. The permanent CDC test scanner added in 2025 made the race easier to expose.

Validation:
- focused reset tests: PASS, count=20
- focused reset/connection race tests: PASS, count=20
- full pkg/frontend under race: PASS (25.983s)
- independent Q1-Q3 review: NO BLOCKER.

XuPeng-SH (approved):

Issue #26095, opened 2026-07-23
[Bug]: Concurrent deletion of independent Data Branch objects deadlocks


Ariznawlll:
## Environment

- Branch: `main`
- Commit: `2eb7f523e5a10f352d3e258183ed8440510388de`
- Deployment: local standalone MatrixOne

## Summary

Concurrent deletion statements on independent Data Branch objects can deadlock each other. This affects plain `DROP TABLE`, `DATA BRANCH DELETE TABLE`, and `DATA BRANCH DELETE DATABASE`. These statements target independent objects, but the branch snapshot reclaim path locks the complete `mo_branch_metadata` relation after each transaction has already updated its own branch rows.

## Reproduction

Prepare four sibling branches:

```sql
DROP DATABASE IF EXISTS db_branch_drop_concurrent;
CREATE DATABASE db_branch_drop_concurrent;
CREATE TABLE db_branch_drop_concurrent.root_t(id INT PRIMARY KEY);
INSERT INTO db_branch_drop_concurrent.root_t VALUES (1);

DATA BRANCH CREATE TABLE db_branch_drop_concurrent.b1 FROM db_branch_drop_concurrent.root_t;
DATA BRANCH CREATE TABLE db_branch_drop_concurrent.b2 FROM db_branch_drop_concurrent.root_t;
DATA BRANCH CREATE TABLE db_branch_drop_concurrent.b3 FROM db_branch_drop_concurrent.root_t;
DATA BRANCH CREATE TABLE db_branch_drop_concurrent.b4 FROM db_branch_drop_concurrent.root_t;
```

Start these four statements at the same time from four independent sessions:

```sql
DROP TABLE db_branch_drop_concurrent.b1;
DROP TABLE db_branch_drop_concurrent.b2;
DROP TABLE db_branch_drop_concurrent.b3;
DROP TABLE db_branch_drop_concurrent.b4;
```

Repeat the setup and concurrent drop three times.

The same setup also reproduces when the four concurrent statements use the dedicated branch syntax:

```sql
DATA BRANCH DELETE TABLE db_branch_drop_concurrent.b1;
DATA BRANCH DELETE TABLE db_branch_drop_concurrent.b2;
DATA BRANCH DELETE TABLE db_branch_drop_concurrent.b3;
DATA BRANCH DELETE TABLE db_branch_drop_concurrent.b4;
```

The database-level entry point reproduces with two independent branch databases:

```sql
CREATE DATABASE db_branch_drop_src;
[...]

PR #26137, opened 2026-07-24
fix: avoid deadlocks in concurrent data branch deletion

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26095

Serialize concurrent Data Branch deletion by acquiring the complete
`mo_branch_metadata` coordination lock before marking any branch row deleted.
The compile and frontend deletion paths now share the same lock -> mark ->
reclaim ordering.

`DATA BRANCH DELETE TABLE` also executes its nested `DROP TABLE` through the
retry-capable background statement path, so a waiter can complete the required
RC statement retry after the coordination lock is released.

The embedded-cluster regression covers concurrent plain `DROP TABLE`,
`DATA BRANCH DELETE TABLE`, and `DATA BRANCH DELETE DATABASE` for both sys and
an authenticated non-sys tenant. Each case captures the affected table IDs
before deletion and verifies all terminal state after deletion:

- the target tables/databases are absent;
- `mo_branch_metadata.table_deleted = true` for every captured ID;
- matching `kind='branch'` / `__mo_branch_*` protect snapshots are gone.

The test also asserts the tenant session's nonzero `current_account_id()` so it
exercises the non-sys context-switch branch rather than the sys-account path.

Validation:

- `go test ./pkg/tests/issues -run '^TestIssue26095ConcurrentDataBranchDeletion$' -count=1`
- `go test -race ./pkg/tests/issues -run '^TestIssue26095ConcurrentDataBranchDeletion$' -count=1`
- `go test ./pkg/sql/compile -run '^$' -count=1`
- targeted mutation: skipping `MarkAndReclaimBranchSnapshotsCore` makes the
  regression fail with expected 4 marked metadata rows, actual 0
- `git diff --check`

Preflight: exact-head review PASS; post-push CI pending.


XuPeng-SH (approved):
Reviewed the full lock/mark/reclaim closure across frontend and compile paths. The transaction ownership and lock ordering are consistent, failure paths roll back atomically, and the concurrent sys/tenant regression passes both normally and under `-race`.

VioletQwQ-0:
Resolved the current-main conflict in merge commit c88283d1b07b5284c2560d206be0ac839599ed78 without rewriting the approved history.

The resolution preserves both sides of the semantic contract:
- PR behavior: acquire the complete branch metadata DAG lock before marking table_deleted, then reclaim snapshots in the same core.
- Current main behavior: retain the branch-participation boolean, non-branch fast path, and conditional ALTER-lineage compaction after participating plain DROP TABLE.

Exact-head local validation:
- focused lock-order test, normal and race
- pkg/sql/compile compile gate
- TestIssue26095ConcurrentDataBranchDeletion, normal and race
- sys and authenticated non-sys tenant terminal assertions for plain table, branch table, and branch database deletion
- metadata marked and protect snapshots removed for captured stable table IDs
- git diff --check on the PR diff
- preflight schema v4: PASS review=PASS, validation=PENDING

Fresh CI is pending on the new head.

PR #26138, opened 2026-07-24
fix launch proxy disconnect cancellation

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25599

fix launch proxy disconnect cancellation


gouhongshen (changes requested):
## Codex automated review

Dynamic launch still routes 6001 through the uncanceling Goetty proxy, and proxy-port ownership is inconsistent across supported launch configurations.

### P1 - Dynamic launch still exposes the uncanceling Goetty proxy on 6001 (cmd/mo-service/dynamic.go:67)
Dynamic mode never calls the new static `shouldStartBuiltinCNProxy` guard. With `-with-proxy`, it starts the configured Proxy service and then unconditionally starts `goetty.NewProxy` on 6001. The regression scenario therefore still bypasses cancellation-aware handling in dynamic-with-proxy mode, allowing a disconnected blocked write to commit.

### P2 - The built-in proxy guard breaks other launch configurations’ 6001 endpoint (cmd/mo-service/launch.go:161)
The guard uses only `*withProxy`, not the configured proxy’s listen address. Static launches using configurations such as `etc/v1/launch-with-proxy/proxy.toml` still configure the Proxy service for its default 6009, but now lose the previous built-in 6001 listener. The PR updates only one config tree, while `pkg/proxy/doc.go` documents both 6001 and 6009. Derive ownership from the actual configured endpoint or update all supported configs and documentation together.

gouhongshen (changes requested):
## Codex automated review

Prior dynamic-6001 blocker is fixed: dynamic launch now skips Goetty when the configured Proxy owns 6001. Prior static endpoint-ownership blocker is fixed: a Proxy on its default/other port preserves the legacy 6001 listener. One new compatibility blocker remains in the shared endpoint classifier.

### P1 - Accept the Proxy listener's supported URL address forms when determining 6001 ownership (cmd/mo-service/launch.go:198)
`proxyServiceOwnsPort` unconditionally passes `ListenAddress` to `net.SplitHostPort`. The production listener explicitly supports both `tcp://` and `unix://` forms in `pkg/proxy/server_listener.go`; `unix:///tmp/proxy.sock` is a valid listener with no TCP port, but `SplitHostPort` returns an error. Consequently any `-with-proxy` static launch returns at `startCluster` before CN startup, while dynamic launch starts CNs and the Proxy then returns the same error. This is a regression from the prior launcher, which started both the Unix-socket Proxy and the built-in 6001 proxy. Classify addresses using the same listener parser/contract (Unix should simply not own 6001), and add URL-form coverage.

gouhongshen (approved):
## Codex automated review

Prior dynamic-6001 and static endpoint-ownership blockers remain fixed. The prior Unix URL classifier blocker is fixed: `tcp://` is recognized and `unix://` leaves the legacy TCP listener enabled. No new defects found.

fengttt (approved):

XuPeng-SH (changes requested):
Deep-reviewed exact head `5490ae38f5e0286d8057b519b72807f60af363b5`. The new tests and both owning packages pass under `-race`, but two launch-topology blockers remain.

### [P1] Preserve the dynamic single-CN 6001 entrypoint (`cmd/mo-service/dynamic.go:81`)

`startDynamicBuiltinProxy` now reuses the static predicate `upstreamCount > 1`. Dynamic launch is different: generated CNs start at 16001, and before this PR the built-in 6001 proxy was started for every positive `service-count`. With the valid `service-count = 1` case, when the configured Proxy does not own 6001 (including `-with-proxy=false`), neither the CN nor any proxy listens on the legacy 6001 entrypoint. I reproduced this deterministically on the exact head with an injected `launchNewProxy`: `startDynamicBuiltinProxy(1, false)` returns nil without calling `Start`.

Please separate the static/dynamic admission rule (or otherwise account for the generated CN's actual frontend port), and cover dynamic `service-count=1` both with and without a real Proxy owning 6001.

### [P1] Route the v1 static launch's 6001 endpoint through the cancellation-aware Proxy (`etc/v1/launch-with-proxy/proxy.toml:32`)

This PR updates three Proxy configs to listen on 6001, but leaves `etc/v1/launch-with-proxy/proxy.toml` at the default 6009. That topology has two CNs, so the new classifier deliberately keeps Goetty on 6001. Clients using the repository-documented/local legacy 6001 path therefore still bypass `pkg/proxy` and retain the #25599 disconnect/late-write behavior. `etc/v1/docker-multi-cn-local-disk/launch.toml` reuses the same omitted config.

Please move that supported topology's Proxy listener to 6001 as well (and add a config-level ownership regression), or explicitly remove/redirect its unsafe 6001 endpoint. Leaving both listeners alive with different cancellation semantics is not behaviorally consistent.

XuPeng-SH (approved):
Deep-reviewed latest exact head `4c2b7507ae9b58f1d41bb7e606320fae6b915cee`. Both prior topology blockers are fixed: dynamic single-CN retains 6001 when no real Proxy owns it, and all four supported static/dynamic v1/current Proxy launch configs now route 6001 through pkg/proxy. I rechecked static/dynamic, one/many CN, with-proxy on/off, default/6001/6009, tcp:// and unix:// ownership, listener conflicts, proxy-enabled CN config, handler disconnect states, and latest-main integration. Focused launcher/proxy tests pass under race for 3 iterations; CI is green. The change is startup/control-plane only and adds no query hot-path cost. No blocker found.

Issue #26083, opened 2026-07-23
[Bug]: Data Branch DAG construction recursively fails on a cyclic parent relation


Ariznawlll:
## Description

`databranchutils.NewDAG` recursively calculates every node depth without tracking nodes already on the current parent walk. A cyclic parent relation such as `1 -> 2 -> 1` therefore recurses until the Go runtime terminates the process.

The snapshot-reclaim DAG already detects and stops cyclic parent walks, but the LCA-oriented `DataBranchDAG` does not apply the same guard while being built from `mo_branch_metadata`.

Tested on `main` commit `2eb7f523e5a10f352d3e258183ed8440510388de`.

## Reproduction

Run `NewDAG` with:

```go
rows := []DataBranchMetadata{
    {TableID: 1, PTableID: 2},
    {TableID: 2, PTableID: 1},
}
NewDAG(rows)
```

A focused test executes the call in an isolated child process with a bounded Go stack. The child exits with status 2 in 3 consecutive runs.

## Actual result

`NewDAG` never completes normally. `calculateDepth(1)` calls `calculateDepth(2)`, which calls `calculateDepth(1)` again until recursive stack exhaustion terminates the child process.

## Expected result

`NewDAG` should detect a node already present on the current depth-calculation path and either reject the cyclic metadata or build a safe bounded representation. Constructing the DAG must not terminate the process.

## Code path

`pkg/frontend/databranchutils/branch_dag.go`:

- `NewDAG`
- `DataBranchDAG.calculateDepth`

`PathFromAncestor` and `BranchReclaimDag` already contain bounded or visited-set walks that can serve as consistency references.

VioletQwQ-0:
Resolved by #26139, merged into main at 2026-07-24T13:22:26Z. DAG construction is now bounded for cycles and deep chains, detaches only cycle members, and preserves valid descendant edges. Focused and race tests passed, and exhaustive verification covered all 126,125 functional parent graphs with 1-6 nodes. Closing this narrow crash bug as complete.

PR #26139, opened 2026-07-24
fix: prevent recursion failure in data branch DAG cycles

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26083

Replace recursive Data Branch DAG depth calculation with a bounded iterative
parent walk. When corrupted branch metadata contains a parent cycle, all nodes
in that cycle are deterministically detached and treated as roots, while
non-cyclic descendants retain their valid parent edges.

This prevents cyclic `mo_branch_metadata` from exhausting the process stack and
also removes recursive stack growth for deep acyclic chains. The existing
`NewDAG` API and ordinary LCA behavior remain unchanged.

Adds a bounded child-process regression covering a two-node cycle, self-cycle,
non-cyclic descendant, and a 4,096-node acyclic chain.

Validation:

- `go test ./pkg/frontend/databranchutils -run '^(TestNewDAGCycle|TestDAGFunctionality)$' -count=10`
- `go test ./pkg/frontend/databranchutils -run '^TestNewDAGCycle$' -count=50`
- `go test -race ./pkg/frontend/databranchutils -run '^TestNewDAGCycle$' -count=1`
- `go test ./pkg/frontend -run '^$' -count=1`
- `git diff --check`


XuPeng-SH (approved):
Reviewed the complete DAG construction and query closure. The iterative parent walk terminates for cycles and deep chains, deterministically detaches exactly the cycle members, preserves non-cyclic descendant edges, and leaves FindLCA invariants valid. Focused cycle/deep-chain tests, race tests, and an additional randomized functional-graph invariant test passed. No correctness, performance, compatibility, or lifecycle blocker found.

XuPeng-SH (approved):
Deep re-review completed on `8723647b` against the latest `main` integration state.

Additional verification beyond the existing PR tests:

- Exhaustively enumerated all 126,125 functional parent graphs with 1-6 nodes, including every combination of roots, self-cycles, multi-node cycles, multiple disconnected cycles, and acyclic prefixes/descendants.
- For every graph, verified that exactly the cycle members are detached, every retained edge preserves its original parent, all depths satisfy `child.Depth == parent.Depth + 1`, every parent walk terminates, and `FindLCA` plus `PathFromAncestor` match independent reference implementations for every node pair.
- Merged the latest `main` in a detached worktree and reran the focused tests 10x plus race; all passed.
- Benchmarked the defensive implementation against the base: a 4,096-node chain rises from about 0.20 ms to 0.31 ms and uses one additional O(N) visited map, while a 4,096-root forest remains at the same allocation count. This is the expected bounded-memory tradeoff for cycle detection and is not material on the branch-DAG construction path. An optional root/known-parent fast path could reduce constant CPU cost but is not a merge blocker.

No correctness, liveness, compatibility, or practical performance blocker found. Approval remains valid.

PR #26148, opened 2026-07-24
fix: materialize vecuint8 in data branch

### LINKED ISSUE #26136: [Bug]: DATA BRANCH cannot materialize VECUINT8 values
## Summary

`DATA BRANCH` cannot materialize a `VECUINT8` value. The row extraction path produces the value as a display string, while the Data Branch value formatter assumes a `[]uint8` value.

## Reproduction

This was reproduced three times on `origin/main` at `3f3cdcdda64fddb93b6668a301ff38d41e73fb76` with a focused white-box test:

1. Build a `T_array_uint8` vector containing `[1, 2, 3]`.
2. Pass the cell through `extractDataBranchSQLRowValue`.
3. Pass the extracted row value and its original type to `formatValIntoString`.

The formatter terminates at the `T_array_uint8` branch because the extracted value is a `string`, not `[]uint8`.

## Expected behavior

The Data Branch row-materialization path should emit a valid SQL literal for `VECUINT8`, consistent with the other narrow vector types, so a branch operation can apply the row successfully.

## Actual behavior

The value-shape mismatch stops formatting before the row can be applied.

## Relevant flow

`extractDataBranchSQLRowValue` -> `extractRowFromVector` -> `formatValIntoString`

The dedicated frontend extraction test already documents that `VECUINT8` is represented as a display string for row-oriented consumers, while the newly added formatter branch asserts `[]uint8`.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26136

Data Branch SQL materialization previously received `VECUINT8` values as display strings, while its formatter requires the original `[]uint8` representation.

Changes:
- preserve `T_array_uint8` as `[]uint8` at the Data Branch-specific extraction boundary;
- keep the generic result-set display-string contract unchanged;
- cover non-empty, empty, and NULL values through extraction and exact SQL-literal formatting.

Validation:
- `go test ./pkg/frontend -run '^TestExtractDataBranchSQLRowValueUint8Array$' -count=3`
- `git diff --check`

Non-goals:
- no change to generic SELECT/result-set representation;
- no change to other vector or scalar formatter contracts.


XuPeng-SH (changes requested):
[P1] Preserve the const-NULL guard before accessing the uint8 array payload

The new `T_array_uint8` branch bypasses `extractRowFromVector`, which previously handled `vec.IsConstNull()` before reading the payload. The outer guard only checks `vec.GetNulls().Contains(rowIdx)`, but a const-NULL vector has no payload and represents NULL through its constant class; its null bitmap does not make every nonzero logical row index safe.

I reproduced a deterministic panic on the current head with:

```go
vec := vector.NewConstNull(types.T_array_uint8.ToType(), 3, mp)
row := make([]any, 1)
extractDataBranchSQLRowValue(context.Background(), nil, vec, 0, row, 2)
```

This reaches `vector.GetArrayAt[uint8]`, obtains an empty Varlena slice, and panics at `bs[0]` (`index out of range`). Before this PR, the generic extractor returned `nil` for the same valid vector shape.

Please make the top-level NULL check cover `vec.IsConstNull()` as well (preferably for the whole helper, which also protects the existing special-type branches), and add a regression using a const-NULL vector with length > 1 and a nonzero row index. The expected result is `row[0] == nil` and SQL literal `NULL`, with no payload access.

VioletQwQ-0:
Addressed the const-NULL review finding in `e93797df3`.

- `extractDataBranchSQLRowValue` now checks `vec.IsConstNull()` in its shared top-level NULL guard before any special-type payload access.
- Added a regression with `vector.NewConstNull(T_array_uint8, 3, mp)` at logical row index 2.
- The regression verifies extraction returns `nil` without panic and formatting produces the exact SQL literal `NULL`.

Validation:
- `go test ./pkg/frontend -run '^TestExtractDataBranchSQLRowValueUint8Array$' -count=3`
- exact-head preflight: PASS

XuPeng-SH (approved):
Re-reviewed exact head e93797df after the const-NULL fix. No remaining code blockers.

Verified the focused regression on the PR head and on a simulated merge with current main; the full pkg/frontend test suite also passes on the merged tree. The returned uint8 slice is consumed synchronously before vector release, and the generic result-set representation remains unchanged.

The current BVT failures are unrelated to this change: this stale PR head lacks optools/run_bvt_group.sh referenced by the new main workflow. Please rebase current main and rerun CI before merge.

PR #26149, opened 2026-07-24
fix: propagate data branch CSV copy errors

### LINKED ISSUE #26128: [Bug]: DATA BRANCH CSV output ignores batch copy errors
## Summary

`DATA BRANCH DIFF ... OUTPUT FILE` does not propagate a batch-copy failure before it submits the CSV conversion task. The failed copy produces a nil batch, which the worker dereferences.

## Revision

`2eb7f523e5a10f352d3e258183ed8440510388de` (main)

## Reproduction

The failure is reproducible with a focused frontend white-box test:

1. Build an off-heap text batch larger than a 1 MiB `MPool` limit.
2. Call `bat.Dup(limitedMP)`. It returns a non-nil error and a nil batch.
3. Follow the same unchecked flow used in `writeCSV`: submit the nil result to `constructByte`.

The test passed 3/3 and observed the nil-batch dereference each time.

## Actual behavior

At `pkg/frontend/data_branch_output.go`, the CSV path uses:

```go
copied, _ := bat.Dup(ses.proc.Mp())
```

The returned error is discarded. `constructByte` subsequently accesses `bat.RowCount()` and dereferences the nil value.

## Expected behavior

The batch-copy error should be returned through the existing cancellation/error path. No CSV worker should be started with a nil batch, and the statement should return a regular error after cleaning up its temporary output.

## Notes

The ordinary export path in `pkg/frontend/export.go` checks the same `bat.Dup` error before scheduling conversion work.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26128

`DATA BRANCH DIFF ... OUTPUT FILE` previously discarded `Batch.Dup` errors and could submit a nil batch to the CSV conversion worker.

Changes:
- return the original batch-copy error before mutating worker or index accounting;
- clean an untransferred copied batch and roll back accounting when worker submission fails;
- cover copy failure and submission failure with real mpool and worker-pool barriers.

Validation:
- `go test ./pkg/frontend -run '^TestSubmitCSVBatchForConversion' -count=3`
- `go test -race ./pkg/frontend -run '^TestSubmitCSVBatchForConversion' -count=1`
- `git diff --check`

Non-goals:
- no change to successful CSV formatting, ordering, or file naming;
- no change to the ordinary export path.

### DIFF
--- a/pkg/frontend/data_branch_output.go
+++ b/pkg/frontend/data_branch_output.go
@@ -1205,6 +1205,52 @@ func shouldDiffAsCSV(sqlRet executor.Result) (bool, error) {
 	return vector.GetFixedAtWithTypeCheck[uint64](sqlRet.Batches[0].Vecs[0], 0) == 0, nil
 }
 
+func submitCSVBatchForConversion(
+	ctx context.Context,
+	ses *Session,
+	tblStuff tableStuff,
+	bat *batch.Batch,
+	ep *ExportConfig,
+	workerWg *sync.WaitGroup,
+) error {
+	copied, err := bat.Dup(ses.proc.Mp())
+	if err != nil {
+		return err
+	}
+
+	idx := ep.Index.Add(1)
+	workerWg.Add(1)
+	if err = tblStuff.worker.Submit(func() {
+		defer workerWg.Done()
+		constructByte(ctx, ses, copied, idx, ep.ByteChan, ep)
+	}); err != nil {
+		workerWg.Done()
+		copied.Clean(ses.proc.Mp())
+		ep.Index.Add(-1)
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
Deep review completed; no blocking issues found on head 2fa261de. The copy failure is propagated before index/wait-group mutation, a copied batch is cleaned and the index is rolled back when conversion submission fails, and cancellation prevents the writer from waiting on an unreachable index gap. Verified the new failure/cancellation tests under the race detector for three runs, the full frontend package, go build, go vet, diff checks, and green CI. The branch is currently behind main, but main has no overlapping changes in the modified files.

PR #26152, opened 2026-07-24
[Optimizer] Push down correlated scalar aggregate keys

aunjgr:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26130

TPC-H Q17 decorrelates its scalar `AVG` subquery by grouping on the correlation key, but previously aggregated every inner `lineitem` row even when the outer `part` predicates selected only a small key set.

This change restricts the aggregate input with a SEMI join against a cloned outer base-table scan carrying deterministic, single-binding WHERE predicates. The cloned scan receives a fresh binding tag and preserves its scan snapshot. Predicate order does not affect discovery. Unfiltered, volatile, multi-table, and derived-table domains retain the existing plan.

The regression coverage verifies:

- both Q17 predicate orders produce a filtered `part` key-domain scan below the correlated `AVG`;
- the original and cloned scans use distinct binding tags;
- unfiltered and volatile domains skip the rewrite;
- the optimized query returns the expected result on a counterexample containing an excluded part;
- both high- and low-NDV EXPLAIN snapshots include the SEMI join.

Validation:

- `env GOCACHE=/tmp/mo26152-gocache .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/sql/plan`
- `env GOCACHE=/tmp/mo26152-gocache .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/sql/plan/explain`
- official `mo-tester` on `optimizer/associative.sql`: 20/20 passed
- package coverage: 76.2% (`pushdownScalarAggregateKeys`: 82.4%)
- `go build`, `go vet`, `make err-check`, and `git diff --check`


aunjgr:
Updated the branch to current `main` and replaced the ineffective original rewrite.

Root cause: the previous SEMI join copied the outer scan before WHERE predicate pushdown, so its `part` scan had no `p_brand`/`p_container` filters and still admitted every part key. It also reused the original binding tag.

The revised rewrite builds a fresh, filtered key-domain scan from the complete bound WHERE list, assigns a new binding tag, preserves scan snapshots, and skips unsafe/unhelpful unfiltered or volatile domains. The regression now checks both predicate orders, filter retention, tag ownership, fallback cases, final EXPLAIN shape, and an executed counterexample result.

Local validation passed: planner and explain packages, build, vet, SCA/forbidden-package check, 76.2% package coverage, and official `mo-tester` for `optimizer/associative.sql` (20/20).

aunjgr:
Fixed the coverage gate on current head `c3145f31f1` after merging newest `main`. The failed artifact showed 45/60 changed lines (exactly 75%, while the gate requires >75%). The added guard tests exercise nine previously uncovered changed blocks in scalar-aggregate pushdown/projection traversal, taking the artifact-based changed coverage to at least 54/60 (90%). Focused tests, full planner tests with coverage, SCA, and diff checks pass.

XuPeng-SH (approved):
Reviewed the complete optimizer rewrite against current main. Checked the SEMI-domain equivalence for duplicate and NULL keys, direct multi-key equality, outer joins and superset filters, predicate order, volatile/subquery/derived guards, scan snapshot preservation, fresh tag rebinding, and aggregate empty-input behavior. Focused planner counterexample tests pass locally and CI, coverage, SCA, and BVT are green. No P0-P3 findings.

Issue #26170, opened 2026-07-25
frontend: avoid Error logs for expected direct-connection proxy extra-info timeout


XuPeng-SH:
When `ProxyEnabled` is true, `MysqlProtocolImpl.receiveExtraInfo` waits up to `defaultSaltReadTimeout` (200ms) for the proxy extra-info header. Its own comment states that a timeout is normal for a direct connection and should simply leave the extra info unset.

However, that expected fallback is currently emitted with `ses.Error`, for example:

```
cannot get salt, maybe not use proxy: read tcp ... i/o timeout
```

In nightly regression run 30111505702 this produced widespread Error-level logs from CN, TN, proxy, and LogService direct clients, obscuring the distinct real MORPC lock-client timeout.

Expected:

- Keep direct connections supported without proxy extra info.
- Do not emit an Error for the expected timeout; use debug-level logging, a bounded warning, or a dedicated counter instead.
- Preserve Error logging for malformed extra-info or unexpected transport errors.

Relevant code: `pkg/frontend/mysql_protocol.go:receiveExtraInfo`.

PR #26176, opened 2026-07-25
fix(frontend): demote expected proxy extra-info timeout log

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26170

When proxy support is enabled, `RoutineManager.Created` probes for MatrixOne
proxy extra-info before sending the MySQL handshake. A direct MySQL client does
not send that private header and instead waits for the server handshake, so the
probe's bounded 200 ms read timeout is the expected direct-connection fallback.

PR #21090 intended to reduce routine INFO logs, but this branch was accidentally
changed from `Info` to `Error` while the other routine logs were changed to
`Debug`. As a result, normal direct connections emit
`cannot get salt, maybe not use proxy` at Error level.

Log only the timeout branch at Debug. Keep malformed extra-info and unexpected
transport/decode failures at Error, and preserve all handshake, salt, label,
connection ID, and connection classification behavior. The regression test
uses real `net.Pipe` connections to cover both the direct-client timeout and
malformed extra-info paths.

Tested with:

`CGO_CFLAGS="-I$PWD/cgo -I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/thirdparties/install/lib" GOWORK=off GOFLAGS=-mod=readonly go build ./pkg/frontend`

`CGO_CFLAGS="-I$PWD/cgo -I$PWD/thirdparties/install/include" CGO_LDFLAGS="-L$PWD/thirdparties/install/lib" GOWORK=off GOFLAGS=-mod=readonly go vet ./pkg/frontend`

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/frontend`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=240s -run '^TestMysqlProtocolReceiveExtraInfoLogLevel$' ./pkg/frontend`

`.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s -run '^TestMysqlClientProtocol_Handshake$' ./pkg/frontend`

### DIFF
--- a/pkg/frontend/mysql_protocol.go
+++ b/pkg/frontend/mysql_protocol.go
@@ -3861,7 +3861,7 @@ func (mp *MysqlProtocolImpl) receiveExtraInfo(rs *Conn) {
 	if err := i.Decode(reader); err != nil {
 		// If the error is timeout, we treat it as normal case and do not update extra info.
 		if err, ok := err.(net.Error); ok && err.Timeout() {
-			mp.ses.Error(mp.ctx, "cannot get salt, maybe not use proxy",
+			mp.ses.Debug(mp.ctx, "cannot get salt, maybe not use proxy",
 				zap.Error(err))
 		} else {
 			mp.ses.Error(mp.ctx, "failed to get extra info",
--- a/pkg/frontend/mysql_protocol_test.go
+++ b/pkg/frontend/mysql_protocol_test.go
@@ -37,7 +37,11 @@ import (
 	"github.com/smartystreets/goconvey/convey"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
 
+	molog "github.com/matrixorigin/matrixone/pkg/common/log"
 	"github.com/matrixorigin/matrixone/pkg/common/moerr"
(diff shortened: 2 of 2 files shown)


XuPeng-SH (approved):
Reviewed exact head 855e44a068. No blocking findings.

The change is correctly scoped to the existing timeout fallback: direct clients that send no private proxy ExtraInfo still follow the same bounded 200 ms path, but the expected event is now Debug instead of Error. Malformed protobuf input and non-timeout transport/decode failures remain Error. Handshake, salt, labels, connection ID, and proxy/direct classification are unchanged.

Unhappy-path closure checked: test net.Pipe endpoints have single deferred cleanup owners; the no-data read is deadline-bounded; the malformed-frame writer terminates through the buffered result channel; no new state or accumulation is introduced. Performance impact is neutral-to-positive because disabled Debug avoids normal-path Error emission.

- git diff --check: pass
- focused test under -race, -count=10: pass
- full pkg/frontend test suite: pass
- required CI including UT, SCA, proxy/standalone/compose BVT, coverage, and Iceberg: green

The branch is behind current main, but main-side commits do not touch either changed file, so I found no semantic or conflict risk from the rebase.

Issue #26174, opened 2026-07-25
[Bug]: HashBuild budget admission rejects fulltext INSERT...SELECT generate_series(3840001) in BVT


matrix-meow:
## Summary

BVT test fails on PR #25782 branch (`repro/25782-main-harness`): the HashBuild memory budget admission control rejects the `INSERT INTO t1 SELECT result, "test create big fulltext index" FROM generate_series(3840001) g` query used in fulltext tests.

## Error

```
internal error: hash build budget admission rejected: requested=18721873920 used=1323840612 cap=10737418240
```

- **requested**: 18.7 GB (hash build needs for 3.84M rows)
- **used**: 1.3 GB (current usage)
- **cap**: 10.7 GB (budget limit set by the HashBuild budget fix)

## Failed Tests (6 failures, all same pattern)

| Test File | Failed SQL | Expected | Actual |
|---|---|---|---|
| `fulltext/fulltext.sql` row 392 | `insert into t1 select result, "test create big fulltext index" from generate_series(3840001) g` | 3840001 rows | 0 (rejected) |
| `fulltext/fulltext.sql` row 394 | `select count(*) from t1 where match(b) against (...)` | 3840001 | 0 |
| `fulltext/fulltext_bm25.sql` row 389 | same INSERT | 3840001 | 0 |
| `fulltext/fulltext_bm25.sql` row 391 | same SELECT | 3840001 | 0 |
| `fulltext/gojieba.sql` row 280 | same INSERT | 3840001 | 0 |
| `fulltext/gojieba.sql` row 282 | same SELECT | 3840001 | 0 |

## CI Run

https://github.com/matrixorigin/matrixone/actions/runs/30155956368/job/89675358523

- Branch: `repro/25782-main-harness`
- Commit: `7088293ab5`
- Job: `MO CHECKIN BVT TEST`
- Step: `5 Generate MO-Tester Config and Start BVT Test`
- Summary: TOTAL 99507, SUCCESS 98366, FAILED 6, SUCCESS RATE 99%

## Root Cause

The HashBuild budget admission fix (PR #25782) sets a cap of ~10.7 GB, but the fulltext `INSERT...SELECT generate_series(3840001)` needs ~18.7 GB for the hash build. The admission controller rejects it.

The 10.7 GB cap is too tight for this legitimate large INSERT workload.

## Related

- #25782 (parent issue: HashBuild can bypass spill and trigger system OOM)

aptend:
fixed

PR #26178, opened 2026-07-25
fix(hashbuild): avoid false spill budget rejection for reused tail batches

### LINKED ISSUE #25782: [Bug]: HashBuild can bypass spill and trigger system OOM during large hash table growth

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

fixes #25782

HashBuild previously used one estimate for two materially different spill sources: the current upstream batch and a future batch retained by `CopyIntoBatches`. That conflated physical input capacity, selected-batch materialization, and marshal peak. Reused tail batches and const vectors could therefore be extrapolated into false multi-gigabyte requests and return `hash build budget admission rejected` before spill made progress.

For the failing `generate_series(3840001)` plus constant-varchar BVT shape, the old path requested 18.72 GB against a 10 GB query cap while only about 1.29 GB was in use.

This change replaces the estimator patch chain with a two-stage admission invariant:

1. Before handling an upstream shuffle batch, prove the actual direct-spill path. Direct const-varlen materialization follows `UnionInt32`: one live out-of-line payload plus per-row descriptors.
2. Before `CopyIntoBatches` may retain that batch, separately prove the largest future retained batch can be drained. The retained destination is non-const, so a later selection can copy payload per retained row and is budgeted accordingly.
3. If the retained proof is rejected, do not copy. Switch to the already-proven direct-spill path.
4. Every retained batch therefore carries a prior future-drain proof; existing retained batches can be drained to release headroom before retrying a larger direct input.

The accounting explicitly separates upstream physical input, selected materialization, row scratch, and marshal growth, uses checked arithmetic, and skips stale varlen headers in NULL rows.

Regression coverage includes:

- reused one-row tails retaining a 32K-row allocation;
- stale const logical length after shuffle;
- stale non-inline headers in NULL rows;
- MiB-scale const varlen direct materialization (one payload copy);
- the real `CopyIntoBatches` const-to-nonconst transition and future selection;
- retained-proof rejection falling back to successful direct spill;
- direct-proof rejection while retained memory is live, followed by drain and retry;
- arithmetic overflow and fail-closed admission boundaries;
- reservation, mpool, spill-disk, and spill-FD cleanup.

Validation on final rebased head:

- `go test -count=1 -timeout=180s ./pkg/sql/colexec/hashbuild`
- `go test -count=1 -timeout=180s ./pkg/sql/colexec/hashjoin`
- focused `go test -race` over the new materialization and state-transition tests
- `go build ./pkg/sql/colexec/hashbuild`
- `go vet ./pkg/sql/colexec/hashbuild`
- local CI-equivalent diff coverage: 82.02% (requested floor: 80%)
- prior TKE BVT run https://github.com/matrixorigin/ci-test/actions/runs/30166184404: the three previously failing inserts succeeded in 1.355s, 1.431s, and 1.425s, with zero hash-build budget admission rejections. A fresh CI/TKE validation is required for this final refactored head.

### DIFF
--- a/pkg/sql/colexec/hashbuild/build.go
+++ b/pkg/sql/colexec/hashbuild/build.go
@@ -263,18 +263,34 @@ func (hashBuild *HashBuild) build(proc *process.Process, analyzer process.Analyz
 		// same upstream batch a second time when it is spilled directly.
 		ctr.hashmapBuilder.InputBatchRowCount += result.Batch.RowCount()
 		if hashBuild.IsShuffle {
-			if err := ctr.ensureSpillScratchReservation(result.Batch, analyzer); err != nil {
-				// A larger ingress batch may require growing the emergency lease.
-				// If retained copies are consuming the missing headroom, drain them
-				// under the already-admitted lease, then retry for the current batch.
+			// First prove that the current upstream batch can always be spilled
+			// directly. This uses its actual materialization semantics and never
+			// projects a hypothetical retained batch.
+			if err := ctr.ensureDirectSpillScratchReservation(result.Batch, analyzer); err != nil {
+				// Existing retained batches were admitted with a future-drain
+				// proof. Drain them under that lease, then retry the direct proof
+				// after their source reservations have been released.
 				if spillMode || !errors.Is(err, process.ErrHashBuildBudgetAdmission) || len(ctr.hashmapBuilder.Batches.Buf) == 0 {
 					return err
 				}
 				if err := startSpill(); err != nil {
 					return err
 				}
-				if err := ctr.ensureSpillScratchReservation(result.Batch, analyzer); err != nil {
+				if err := ctr.ensureDirectSpillScratchReservation(result.Batch, analyzer); err != nil {
 					return err
+				}
+			}
+			if !spillMode {
+				// A batch may become retained only after its future spill scratch
+				// is admitted. If that proof does not fit, do not copy it: switch
+				// to the already-proven direct-spill path.
+				if err := ctr.ensureRetainedSpillScratchReservation(result.Batch, analyzer); err != nil {
+					if !errors.Is(err, process.ErrHashBuildBudgetAdmission) {
+						return err
+					}
+					if err := startSpill(); err != nil {
+						return err
+					}
 				}
 			}
 		}
--- a/pkg/sql/colexec/hashbuild/build_test.go
+++ b/pkg/sql/colexec/hashbuild/build_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/common/mpool"
 	"github.com/matrixorigin/matrixone/pkg/container/batch"
 	"github.com/matrixorigin/matrixone/pkg/container/types"
+	"github.com/matrixorigin/matrixone/pkg/container/vector"
 	"github.com/matrixorigin/matrixone/pkg/pb/plan"
 	"github.com/matrixorigin/matrixone/pkg/sql/colexec/merge"
 	plan2 "github.com/matrixorigin/matrixone/pkg/sql/plan"
@@ -665,6 +666,68 @@ func TestHashBuildIsShuffle(t *testing.T) {
 	tc.proc.Free()
 }
 
+func TestShuffleHashBuildFallsBackToDirectSpillWhenRetainedProofRejects(t *testing.T) {
+	tc := newTestCase(t, []bool{false}, []types.Type{types.T_varchar.ToType()}, []*plan.Expr{newExpr(0, types.T_varchar.ToType())})
+	tc.arg.IsShuffle = true
+	tc.arg.ShuffleIdx = 0
+	tc.arg.SpillThreshold = 1 << 30
+	tc.arg.RuntimeFilterSpec = &plan.RuntimeFilterSpec{Tag: tc.arg.JoinMapTag + 3500}
+	tc.arg.SetChildren([]vm.Operator{tc.marg})
+	require.NoError(t, tc.marg.Prepare(tc.proc))
+	require.NoError(t, tc.arg.Prepare(tc.proc))
+
+	const capBytes = uint64(8 << 20)
+	budget := process.MustNewHashBuildBudget(capBytes, capBytes)
+	generation, err := budget.OpenGeneration(1)
+	require.NoError(t, err)
+	tc.arg.ctr.hashmapBuilder.setBudget(generation)
+
+	payload := make([]byte, 1<<20)
+	for i := range payload {
+		payload[i] = 'x'
+	}
+	build := batch.NewWithSize(1)
+	build.Vecs[0], err = vector.NewConstBytes(types.T_varchar.ToType(), payload, 1, tc.proc.Mp())
+	require.NoError(t, err)
+	build.SetRowCount(1)
+
+	directNeed, err := spillBudgetBytes(build)
+	require.NoError(t, err)
+	require.Less(t, directNeed, capBytes)
+	retainedNeed, err := spillRetainedBudgetBytes(build)
+	require.NoError(t, err)
+	require.Greater(t, retainedNeed, capBytes)
+
+	tc.proc.Reg.MergeReceivers[0].Ch2 <- process.NewPipelineSignalToDirectly(build, nil, tc.proc.Mp())
+	tc.proc.Reg.MergeReceivers[0].Ch2 <- process.NewPipelineSignalToDirectly(nil, nil, tc.proc.Mp())
+	_, err = vm.Exec(tc.arg, tc.proc)
+	require.NoError(t, err)
+
+	result, err := message.ReceiveJoinMapResult(tc.arg.JoinMapTag, true, tc.arg.ShuffleIdx, tc.proc.GetMessageBoard(), tc.proc.Ctx)
+	require.NoError(t, err)
+	require.True(t, result.IsSuccess())
+	jm := result.JoinMap()
+	require.NotNil(t, jm)
+	require.True(t, jm.IsSpilled(), "failed future-retained proof must choose direct spill")
+	require.Equal(t, int64(1), jm.GetRowCount())
+	for _, file := range jm.TakeSpillBuildFiles() {
+		if file != nil {
+			require.NoError(t, file.Close())
+		}
+	}
+	require.Empty(t, tc.arg.ctr.hashmapBuilder.Batches.Buf)
+	require.Zero(t, generation.Used())
+	require.Zero(t, generation.SpillDiskUsed())
+	require.Zero(t, generation.SpillFDUsed())
+
+	tc.arg.Reset(tc.proc, false, nil)
+	tc.arg.Free(tc.proc, false, nil)
+	tc.marg.Reset(tc.proc, false, nil)
+	generation.Close()
+	tc.proc.Free()
+	require.Zero(t, tc.proc.Mp().CurrNB())
+}
+
 func TestShuffleHashBuildSpillsExpressionKey(t *testing.T) {
 	bindProc := testutil.NewProcessWithMPool(t, "", mpool.MustNewZero())
 	col := newExpr(0, types.T_int32.ToType())
@@ -857,7 +920,10 @@ func TestShuffleHashBuildDrainsRetainedBatchBeforeGrowingScratch(t *testing.T) {
 	forcedGrowReject := false
 	aggregate.SetAggregateCapProvider(func() (uint64, error) {
 		providerCalls++
-		if providerCalls == 3 {
+		// First ingress reserves direct scratch, grows it for future retained
+		// drain, then reserves its copy. Reject the fourth admission: growing
+		// direct scratch for the larger second ingress while that copy is live.
+		if providerCalls == 4 {
 			forcedGrowReject = true
 			return generation.Used(), nil
 		}
@@ -865,7 +931,7 @@ func TestShuffleHashBuildDrainsRetainedBatchBeforeGrowingScratch(t *testing.T) {
 	})
 
 	first := newBatch(tc.types, tc.proc, 8192)
-	second := newBatch(tc.types, tc.proc, 16384)
+	second := newBatch(tc.types, tc.proc, 65536)
 	tc.proc.Reg.MergeReceivers[0].Ch2 <- process.NewPipelineSignalToDirectly(first, nil, tc.proc.Mp())
 	tc.proc.Reg.MergeReceivers[0].Ch2 <- process.NewPipelineSignalToDirectly(second, nil, tc.proc.Mp())
 	tc.proc.Reg.MergeReceivers[0].Ch2 <- process.NewPipelineSignalToDirectly(nil, nil, tc.proc.Mp())
@@ -879,7 +945,7 @@ func TestShuffleHashBuildDrainsRetainedBatchBeforeGrowingScratch(t *testing.T) {
 	require.True(t, result.IsSuccess())
 	jm := result.JoinMap()
 	require.True(t, jm.IsSpilled())
-	require.Equal(t, int64(24576), jm.GetRowCount())
+	require.Equal(t, int64(73728), jm.GetRowCount())
 	for _, file := range jm.TakeSpillBuildFiles() {
 		if file != nil {
 			require.NoError(t, file.Close())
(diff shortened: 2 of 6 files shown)


XuPeng-SH (changes requested):
I still see one merge-blocking spill-accounting issue on the current head:

**Null varlen tail batches can still be mischarged as multi-row payload and abort spill.**
In `pkg/sql/colexec/hashbuild/spill.go:188-224`, `spillProjectedSourceBytes` skips only `vec.IsConstNull()`, then sums out-of-line varlen payload from every row in `values`. That misses the reused non-const tail-batch shape where the only published row is NULL but its slot still retains a stale non-inline varlena header from an earlier batch. We already have an explicit guard for this exact vector shape in `pkg/container/vector/vector.go:2784-2798` (and `unionbatch_fastpath_test.go:329-360`), where null rows are excluded because reused varlen vectors can keep stale large headers in null slots.

With the new spill estimator, a one-row NULL tail can still be projected as `DefaultBatchSize * staleLen`, so `ensureSpillScratchReservation` in `pkg/sql/colexec/hashbuild/build.go:265-279` can return `hash build budget admission rejected` before spill makes progress (or after the retry path rechecks the same batch). Please skip non-const null rows when summing external varlen payload, mirroring the existing vector pre-grow logic, and add a regression test for a reused one-row NULL varchar tail with a planted stale large header.

aptend:
Addressed the stale-null varlena accounting issue in 50a797fd4. `spillProjectedSourceBytes` now excludes non-const NULL rows when summing out-of-line payload, matching `Vector.Union` semantics. Added a regression test with a one-row NULL varchar tail carrying a planted 1 MiB stale non-inline header; the legacy projection exceeds 10 GiB while the corrected emergency estimate remains below 16 MiB. Added nil/invalid/const-null boundary coverage as well. HashBuild/HashJoin full tests, focused race, build, and vet all pass after rebasing onto the latest remote branch.

XuPeng-SH (commented):
I took another pass and found one more independent blocker on the current head:

**Large const varlen values are still budgeted as if their out-of-line payload were copied once per row.**
`spillSourceBytes` / `spillProjectedSourceBytes` still multiply a large const varlen payload by `rows` / `targetRows` (`pkg/sql/colexec/hashbuild/spill.go:94-105`, `:216-224`). But the actual const-varlen materialization path does not duplicate the area per row: `Union` calls `BuildVarlenaFromVarlena` once and then `fillSlice` fans out the descriptor (`pkg/container/vector/vector.go:2758-2769`, `5341-5348`), `BuildVarlenaNoInline` appends the bytes once (`vector.go:5296-5313`), and marshal writes `area` once (`vector.go:773-777`). So a shuffled/tail batch with a large const varchar/blob can still be overcharged by orders of magnitude and falsely fail spill admission.

The new regression at `pkg/sql/colexec/hashbuild/spill_test.go:888-930` uses a ~30-byte const string, so it never exercises the out-of-line path. Please charge one live out-of-line payload copy for const varlen vectors and add a regression with a MiB-scale const payload.

aptend:
Implemented the source-level two-stage admission design on fd3e5dbdb (rebased onto current main).

The large-const observation is correct for direct `UnionInt32`: the direct estimator now charges one out-of-line payload plus per-row descriptors, and a MiB-scale regression verifies that against the actual materialized vector.

The retained path intentionally differs. `CopyIntoBatches` materializes a non-const destination; a later `UnionInt32` can copy the payload once per retained row. HashBuild now proves that future retained drain separately before copying. If it cannot be admitted, the batch is never retained and is spilled through the already-proven direct path. A test executes the real const → `CopyIntoBatches` → non-const → selection closure, and an operator test verifies retained-proof rejection succeeds via direct spill without leaking budget, mpool, disk, or FD resources.

Local CI-equivalent diff coverage is 82.02%; HashBuild/HashJoin full tests, focused race, build, and vet pass.

XuPeng-SH (commented):
Latest-head update: the null-slot guard and the direct-spill const-varlen path look fixed now, but I still see one remaining blocker.

**Exact-8192-row const varlen batches are still budgeted as if their out-of-line payload were copied once per row.**
`spillRetainedBudgetBytes` always models the retained batch with `spillRetainedMaterialization` (`pkg/sql/colexec/hashbuild/spill.go:215-251`), and that mode repeats the full live varlen payload for every output row (`spill.go:128-143`). That matches the compact `UnionBatch` retain path, but it is not what the exact-full-batch retain path does: `CopyIntoBatches` uses `src.Dup()` when `RowCount()==DefaultBatchSize` (`pkg/sql/colexec/join_util.go:76-88`), and `Vector.Dup` preserves the const class while copying only one const varlen payload (`pkg/container/vector/vector.go:1023-1044`).

So for a large const varchar/blob batch of exactly 8192 rows, the retained admission proof can still overcharge by orders of magnitude and `build()` will switch into spill mode on that inflated estimate before `copyBuildBatch` ever runs (`pkg/sql/colexec/hashbuild/build.go:283-294`). The current const regression (`spill_test.go:876-939`) uses an inline ~30-byte string, so it never exercises the out-of-line payload path. Please special-case the exact-full-batch `Dup()` closure (or otherwise budget retained const varlen by its real retained representation) and add a MiB-scale const regression for that path.

aptend:
I traced and executed the exact-8192-row closure. The proposed special case would under-budget the real retained spill path.

`CopyIntoBatches` does call `Batch.Dup`, but `Batch.Dup` delegates to `Batch.Clone` → `Batch.CloneTo` → `Vector.UnionBatch`; it does **not** call `Vector.Dup` (`batch.go:580-613`). The destination is therefore non-const. For const varlen input, that `UnionBatch` initially stores one shared payload, but a later spill sees a non-const source and `UnionInt32` copies the payload once per selected retained row. If all identical keys land in one bucket, an exact 8192-row batch with a 1 MiB value can require an 8 GiB selected area.

Commit 0f55971b5 adds executable closure coverage:
- exact 8192-row, 1 MiB const input becomes non-const after the real `CopyIntoBatches` call;
- direct spill remains below 16 MiB while retained future-drain proof correctly exceeds 10 GiB;
- a 4 KiB variant executes the full retained `UnionInt32` selection and materializes exactly `8192 * 4 KiB`; the retained estimate covers the actual allocation.

So the current retained accounting is necessary for OOM safety. Special-casing this path as const-preserving would reintroduce the failure class. Full HashBuild tests and focused race pass.

XuPeng-SH (commented):
Latest-head re-review from scratch: one correction and one current blocker.

**Correction on my previous exact-8192 const-retained comment:** that point was based on the wrong closure. `CopyIntoBatches` does call `Batch.Dup`, but `Batch.Dup` goes through `Batch.Clone -> CloneTo -> Vector.UnionBatch` (`pkg/container/batch/batch.go:580-613`), so the retained batch is not preserved as a const vector via `Vector.Dup`. Please disregard that earlier exact-8192 const-retained concern from me.

**Current blocker:** `startSpill()` drains retained batches while the current ingress batch is still live, but the drain path is only admitted against the retained-batch scratch proof. The overlap state is under-budgeted.
- `spillBudgetBytes` for a direct spill explicitly includes the live upstream batch via `uint64(bat.Allocated())` (`pkg/sql/colexec/hashbuild/spill.go:183-195`).
- But `startSpill()` drains retained copies with `spillBatchBounded(..., sourceAlreadyCharged=true)` (`pkg/sql/colexec/hashbuild/build.go:213-245`), and that path subtracts the retained source from the scratch proof (`spill.go:197-212`, `468-507`).
- `startSpill()` is called from `build()` before `result.Batch` is spilled or released (`build.go:266-293`), so the current ingress batch is still resident while the retained queue is being drained.

That means the real recovery peak is **current ingress source bytes + retained-drain scratch**, but the code only proves either the current direct spill path or the retained-drain path in isolation. A budget that is sufficient for the retained drain alone can therefore pass admission and still exceed the query/CN budget during recovery.

Please make the spill transition reserve/prove that overlap state explicitly while draining retained batches, and add a regression that exercises `startSpill()` with one live ingress batch plus a retained queue under a cap that separates the isolated proofs from the combined peak.

XuPeng-SH (approved):
Re-evaluated the overlap-accounting finding against its production impact and the scope of this PR. The gap is real, but it is limited to the spill-transition overlap window, has no known result/data-correctness impact, and should not expand or block the #26174 fix in this PR.

Tracked separately in #26186 with the required overlap proof and regression boundaries. Superseding my earlier merge-blocking position: this PR is okay to merge first.

Issue #26175, opened 2026-07-25
[Bug] panic: assignment to entry in nil map in sanitizeNonFiniteFloatValue (TPCC warehouse=1000)


Ariznawlll:
### Summary

TPCC warehouse=1000 测试期间，CN 节点触发 `panic: assignment to entry in nil map`，发生在 `pkg/frontend/mysql_cmd_executor.go` 的 `sanitizeNonFiniteFloatValue` 函数中。

### CI

- **Run**: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/30143464491
- **Namespace**: `mo-main-commit-36f8b76c3-20260725`
- **MO commit**: `36f8b76c398e31b4f763fc9b32af5591ee0e1060` (main)
- **Pod**: `nightly-regression-dis-tp-cn-qfszt`
- **时间**: 2026-07-25 06:59:59 UTC

### Panic Stack Trace

```
panic: assignment to entry in nil map

runtime.mapassign
    /usr/local/go/src/internal/runtime/maps/runtime.go:208
reflect.mapassign0
    /usr/local/go/src/runtime/map.go:254
reflect.mapassign
    /usr/local/go/src/reflect/value.go:3746
reflect.Value.SetMapIndex
    /usr/local/go/src/reflect/map.go:439
github.com/matrixorigin/matrixone/pkg/frontend.sanitizeNonFiniteFloatValue
    /go/src/github.com/matrixorigin/matrixone/pkg/frontend/mysql_cmd_executor.go:5402   ← panic here
github.com/matrixorigin/matrixone/pkg/frontend.sanitizeNonFiniteFloatValue
    /go/src/github.com/matrixorigin/matrixone/pkg/frontend/mysql_cmd_executor.go:5384   ← Struct recurse
github.com/matrixorigin/matrixone/pkg/frontend.sanitizeNonFiniteFloatValue
    /go/src/github.com/matrixorigin/matrixone/pkg/frontend/mysql_cmd_executor.go:5381   ← Ptr recurse
github.com/matrixorigin/matrixone/pkg/frontend.sanitizeNonFiniteFloatValue
    /go/src/github.com/matrixorigin/matrixone/pkg/frontend/mysql_cmd_executor.go:5388   ← Slice recurse
    ... (deeply nested recursive traversal)
```

调用路径：`Struct → Ptr → Struct → Ptr → Slice → Struct → ... → Map → SetMapIndex → panic`

### 问题分析

[...]

PR #26180, opened 2026-07-26
fix(tae): align late aobject visibility with parent txn

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

N/A. This regression was isolated from the `TestDelete3` failure after #25534. It is unrelated to #26175.

### Root cause

A flush/merge transaction can collect deletes again from `flushTableTailEntry.PrepareCommit`, after its prepare/commit timestamp has already been assigned. If that late transfer rolls over the current tombstone appender, `prepareApplyANode` creates one or more appendable tombstone objects.

Since #25534, appendable objects are committed catalog entries outside the transaction. The active append path is correct because the object create timestamp precedes the append transaction commit. In the late-transfer path, however, each object received a fresh timestamp later than the parent flush timestamp. A snapshot could therefore see the rewritten data while seeing none or only a prefix of its transferred tombstones.

The deterministic reproducer on unmodified main returned visible rows after the delete. The strengthened concurrent-snapshot mutation test returns 22 rows without this fix and 0 with it.

### Fix

This PR keeps appendable objects outside transactions. It does not add object-create transaction entries or object-create WAL records.

When an appendable object is created while its append transaction is already `Preparing`, its committed catalog entry uses the parent prepare timestamp. The object remains only a carrier: its appended rows are still owned by the parent append MVCC node. Correctness scan paths that cross that timestamp wait on the parent's rewritten data/append MVCC before snapshotting transfer tombstones. WAL replay reconstructs a missing appendable object from the append command at the same timestamp.

The normal `Active` path remains on `AllocateAndPublishCommitTS`, preserving #25534's atomic timestamp allocation/catalog publication contract. The `Preparing` branch is a deliberate, narrow exception for parent-owned late-transfer rows; independently committed object state must continue to use the normal publication path.

### Design and performance

- aobjects remain outside transactions and may survive creator rollback as empty catalog shells;
- an empty shell without append/delete state remains intentionally non-durable, as in #25534;
- rollback removes the parent-owned append MVCC rows, so transferred deletes cannot become visible after parent abort;
- replay reconstructs committed carriers from append WAL without a format or upgrade change;
- multiple carriers may share the parent timestamp; object name remains the catalog ordering tie-breaker;
- the active path adds one transaction-state read per aobject creation, not per row or per commit;
- the rare preparing path avoids the per-object global timestamp allocation;
- no new I/O, WAL record, storage scan, allocation, or global publication fence is introduced;
- tombstone-only incremental dedup remains a best-effort fallback rather than gaining a strict global catalog fence.

## Coverage

The tests force multiple tombstone-object rollovers and cover:

- snapshots immediately before, at, and after every relevant visibility boundary;
- a reader snapshot explicitly greater than the parent prepare timestamp, fixed while the parent is `Preparing` and before any late transfer object exists;
- cleanup that always releases the deterministic prepare barrier and waits for the commit goroutine, including assertion failures;
- rollback after late transfer object creation;
- restart/replay after both commit and rollback;
- the existing outside-transaction aobject lifecycle through the full txnimpl suite.

Validation on the final diff:

- enhanced concurrent-snapshot test: fails on unmodified main with 22 visible rows, passes with this fix;
- focused regression set: `-count=10`;
- `TestDelete3` plus the new tests: `-race -count=20`;
- full `pkg/vm/engine/tae/db/test` suite;
- full `pkg/vm/engine/tae/...` suite with `-p=1`;
- `go vet` for both modified packages;
- `go build` for `pkg/vm/engine/tae/txn/txnimpl`;
- `git diff --check`.

### DIFF
--- a/pkg/vm/engine/tae/db/test/db_test.go
+++ b/pkg/vm/engine/tae/db/test/db_test.go
@@ -3816,6 +3816,211 @@ func TestDelete3(t *testing.T) {
 	t.Log(tae.Catalog.SimplePPString(common.PPL1))
 }
 
+type flushPrepareBarrier struct {
+	reached chan struct{}
+	release chan struct{}
+}
+
+func (entry *flushPrepareBarrier) PrepareCommit() error {
+	close(entry.reached)
+	<-entry.release
+	return nil
+}
+
+func (*flushPrepareBarrier) PrepareRollback() error { return nil }
+func (*flushPrepareBarrier) ApplyCommit(string) error {
+	return nil
+}
+func (*flushPrepareBarrier) ApplyRollback() error { return nil }
+func (*flushPrepareBarrier) MakeCommand(uint32) (txnif.TxnCmd, error) {
+	return txnbase.NewTxnStateCmd(
+		"flush-prepare-barrier",
+		txnif.TxnStateCommitted,
+		types.TS{},
+	), nil
+}
+
+type txnEntryLogger interface {
+	LogTxnEntry(txnif.TxnEntry, []*common.ID, []*common.ID) error
+}
+
+func newFlushTransferScenario(
+	t *testing.T, ctx context.Context, beforeFlush txnif.TxnEntry,
+) (*testutil.TestEngine, txnif.AsyncTxn, map[types.Objectid]struct{}) {
+	opts := config.WithLongScanAndCKPOpts(nil)
+	tae := testutil.NewTestEngine(ctx, ModuleName, t, opts)
+	t.Cleanup(func() { require.NoError(t, tae.Close()) })
+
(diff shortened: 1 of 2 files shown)


jiangxinmeng1 (approved):
> LGTM. The fix correctly aligns late-created appendable tombstone objects with the parent transaction’s prepare timestamp while
  > preserving the normal timestamp publication path for active transactions. The commit, rollback, snapshot visibility, and WAL replay
  > paths are well covered.
  >
  > I also ran the three new regression tests five times; all passed. Ubuntu UT and BVT checks are green. Coverage and ARM SCA are
  > still pending.

PR #26183, opened 2026-07-26
fix(iscp): retry drain while executor initializes

### LINKED ISSUE #26181: [Bug]: ISCP task runner is published before executor readiness, causing clone failure in Multi-CN + Proxy

### Branch Name

`xp/ci-race-coverage-artifacts`

### Commit ID

`bb6cf549be44c5162255f74c87dc03fb9b6fa6a7`

- Hardware parameters: GitHub Actions Linux CI runner (Compose BVT job)
- OS type: Linux
- Deployment: Docker Compose, 2 CN + TN + LogService + real Proxy; SQL traffic enters Proxy on port 6001 and is routed to either CN.
- Failing workflow/job: [Coverage Integration run 30182890014](https://github.com/matrixorigin/matrixone/actions/runs/30182890014/job/89743130108)

### Actual Behavior

A clone operation involving an asynchronous index can fail immediately after a Multi-CN cluster becomes SQL-ready.

The failing statement is:

```sql
create database test03_new clone test03 {snapshot = 'sp03'};
```

It fails with:

```text
internal error: cannot confirm ISCP consumer quiescence on CN dd1dccb4-4d3c-41f8-b482-5251dc7a41bf for tableID=272617 jobName=index_idx01 jobID=1
```

The remaining failures in `test/distributed/cases/pessimistic_transaction/clone/clone.sql` are cascading failures because `test03_new` was never created.

The failure is not caused by coverage collection, image-cache loading, or the BVT partitioning:

- the shared Compose image was loaded successfully;
- Docker Compose started successfully;
- the BVT ran for about 15 minutes before reporting 9 failures, all from this clone case;
- coverage data was still generated after teardown.

### Expected Behavior

Once a CN is exposed as the ISCP task runner, ISCP operations routed to that CN must be serviceable. A valid clone statement must not fail merely because the assigned CN has not yet completed ISCP executor startup.

The operation may wait for a bounded startup interval, but must remain fail-closed: clone must not proceed until consumer quiescence has been confirmed.

### Steps to Reproduce

1. Start MatrixOne with Docker Compose using two CN services and Proxy enabled; connect client traffic through Proxy.
2. Wait only until SQL is reachable through Proxy (for example, a simple `create table` succeeds).
3. Run `test/distributed/cases/pessimistic_transaction/clone/clone.sql`, or create a database containing an asynchronous index and execute a clone from a snapshot immediately after startup.
4. Observe that the clone may route to a CN whose `task_runner` assignment is visible but whose local ISCP executor runtime has not been registered.

#### Evidence and root-cause analysis

The lifecycle currently has a readiness gap:

```text
taskservice.startDaemonTask:
  persist task_runner=<CN> and status=RUNNING
    -> invoke ISCPTaskExecutorFactory
      -> repair/rebuild ISCP state
        -> RegisterExecutorRuntime(<CN>)
```

`GetTaskRunner` reads `mo_task.sys_daemon_task.task_runner` and `drainIndexCdcTaskConsumer` sends `ISCPDrainConsumer` to that CN. `handleISCPDrainConsumer` then immediately fails if `iscp.GetExecutorRuntime(s.cfg.UUID)` is absent.

In the failing logs:

- the clone request was routed to CN `dd1dccb4-4d3c-41f8-b482-5251dc7a41bf`;
- that CN logged the missing-runtime error at 01:42:57;
- another CN initialized and started an ISCP executor only about 1.7 seconds later;
- the request therefore observed task-runner assignment before a usable executor, and the task-runner target may change while startup/recovery is in progress.

#### Proposed fix

Do not use daemon-task assignment as the readiness signal.

1. In `pkg/cnservice/server_query.go`, let `handleISCPDrainConsumer` wait for a short, context-bounded interval for a local, running executor rather than immediately returning an internal error.
2. Return a distinct, retryable `ISCP executor not ready` result/error when that interval expires.
3. In `pkg/sql/compile/iscp_util.go`, on only that retryable condition, use bounded backoff and **re-read `task_runner` plus its query address** before retrying. Do not retry arbitrary drain errors.
4. Preserve fail-closed behavior: only proceed after `CancelAndDrainJobConsumer` succeeds.

Avoid changing generic `taskservice.startDaemonTask` semantics globally: it is shared by every daemon task and its RUNNING/task_runner state is also part of lease/failover handling.

#### Acceptance criteria

- A delayed ISCP executor registration after task assignment no longer makes clone fail.
- If task-runner ownership changes during startup, the caller refreshes the target and succeeds once a running executor is available.
- A permanently unavailable executor fails after a bounded timeout with an actionable readiness error.
- Non-readiness drain errors are returned without retry.
- Add handler and compile-layer unit tests plus a 2-CN + Proxy regression test that forces the assignment-before-registration window.

XuPeng-SH:
## Summary

Fixes #26181.

- wait briefly for a late ISCP executor runtime only for the initial drain request
- return the existing retryable CN-transition error if it remains unavailable
- retry only that error with a bounded readiness window and fresh task-runner/address discovery
- revalidate task ownership after every successful drain; if ownership moves, fence every observed runner before proceeding
- renew and roll back all acquired fences, including best-effort cleanup of an uncertain remote target after a lost/error response
- preserve legitimate long drains by starting a fresh readiness generation after each successful drain
- force the assignment-before-readiness path in the existing async-index clone BVT, including when Proxy routes SQL to the task-runner CN

## Validation

- focused handler/readiness race stress: 10 iterations
- focused runner-handoff/cleanup/timeout race stress: 10 iterations
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=600s ./pkg/cnservice ./pkg/sql/compile`
- `GOWORK=off go build -mod=readonly ./pkg/cnservice ./pkg/sql/compile`
- `GOWORK=off go vet -mod=readonly ./pkg/cnservice ./pkg/sql/compile`
- formal UT/SCA and deployment BVT workflows are triggered by this PR update


jiangxinmeng1 (changes requested):
> [P2] Prevent one runner from exhausting the deadline for all fence operations
  >
  > renewISCPDrainTargetFences and removeISCPDrainTargetFences process targets
  > sequentially using the same context. After a runner handoff, an unreachable
  > earlier runner may consume the entire deadline, causing later reachable runners
  > to receive an already-canceled context. Their fences may then expire during a
  > long DDL or remain after rollback.
  >
  > Please execute per-target operations concurrently under the overall deadline,
  > or use independently bounded contexts. A test where the first target blocks
  > while the second remains reachable would cover this case.

XuPeng-SH:
@jiangxinmeng1 Fixed in `d4d41bda3d`.

- renew/remove operations now start concurrently for every fenced runner under the existing overall deadline
- errors are collected per target and joined after all operations finish
- added deterministic renew and remove regressions where runner A blocks but reachable runner B must complete before A is released
- updated existing multi-runner tests to require exactly one cleanup per target without assuming concurrent completion order

Validation: focused `-race -count=50`, full `pkg/sql/compile` race, build, vet, and `make err-check` all pass.

ck89119:
Deep review of `d4d41bda3d42f3b94b9d39274be227198a6f7bb6`: I found four code issues plus the current SCA blocker. I do not think this head is ready to merge.

1. **[P1] Re-observing the same CN UUID is not an ownership/fence linearization point.**

   `fencedRunners` is keyed only by CN UUID and the loop exits as soon as a fresh read returns an already-seen UUID. The fence, however, lives in a specific in-memory `ISCPTaskExecutor` generation. After fencing `CN-A/generation-1`, the executor can restart on the same CN with an empty fence map; the task row still returns `CN-A`, so the loop exits without fencing generation 2. There is also an unavoidable window after the final owner read where the task can move from A to B before the DDL transaction terminates. The lease only renews the saved executor pointer/query address and never follows the new owner.

   This needs a durable job fence or an ownership epoch/token coupled to task assignment and checked by every executor generation when registering a consumer. Seeing the same UUID twice is not sufficient.

2. **[P1] Earlier runner fences can expire before the lease is started.**

   Successful targets are only appended to `fencedTargets`; `startISCPJobFenceLease` is called after the entire owner-convergence loop. The actual drain intentionally uses the statement context and may legitimately take longer than the readiness deadline. For example, A is fenced, ownership moves to B, and draining B takes longer than the 30-minute fence TTL. A's fence expires while waiting for B, allowing its consumer to restart. If ownership cycles back to A, the UUID set still causes the loop to exit even though A is no longer fenced.

   Renewal must start as soon as the first target is fenced and safely incorporate later targets. Please add a short-TTL A/B handoff test with a long second drain.

3. **[P1] `errors.Join` destroys the primary `moerr` identity on failed drain cleanup.**

   `cleanupAfterFailure` always returns `errors.Join(cause, cleanupErr)`. Even when cleanup succeeds and `cleanupErr` is nil, this returns a `joinError`. `moerr.IsMoErrCode`, commit-result-unknown handling, and frontend whole-transaction rollback classification require the error itself to be `*moerr.Error`; they do not unwrap joined errors. A first remote drain failure such as `ErrBackendClosed` is already wrapped because the uncertain target is included, so ALTER/DROP INDEX in an explicit transaction can take statement-only rollback instead of the required whole-transaction rollback.

   Return `cause` unchanged when cleanup succeeds. If cleanup also fails, preserve the primary `moerr` identity or update the relevant classifiers to unwrap. Tests should assert `moerr.IsMoErrCode`, not only `ErrorContains`.

4. **[P2] Rollback can synchronously block for up to 30 minutes.**

   The rollback cost-event callback creates a background context with `iscp.DefaultRollbackFenceTTL` and waits for all target operations via `wg.Wait`. The generic query client has no shorter default backend read timeout, so one connected but unresponsive old runner can hold the transaction terminal path for the full 30 minutes. Fence TTL and RPC cleanup timeout should be separate; use a short bounded cleanup deadline, log cleanup failures, and rely on TTL for fail-closed expiry. A blackholed-target rollback latency test would cover this.

5. **Current CI blocker:** `Matrixone CI / SCA Test on Linux/arm64` fails at `pkg/sql/compile/iscp_util.go:523` with `copyloopvar`: the `i := i` copy can be deleted under Go 1.22+.

Independent validation on this exact head:

- focused `pkg/sql/compile` drain/target tests: `-race -count=10` passed
- focused `pkg/cnservice` drain-handler tests: `-race -count=10` passed
- `git diff --check` passed

The previous sequential multi-target renewal/removal deadline issue is fixed by the concurrent operation helper in this head; the findings above are additional closure gaps.

XuPeng-SH:
Addressed the actionable fence-lifecycle findings in `94f936adb4`, with the scope kept to the production failure path in #26181.

1. Executor generation replacement: a CN-scoped in-process fence registry now outlives an `ISCPTaskExecutor` generation. `RegisterExecutorRuntime` inherits active fences before publishing the replacement executor, and renew is an idempotent upsert. Initial drain installs this fence before waiting for executor readiness. The normal consumer-admission hot path still reads only the executor-local map; it does not take the process-global registry lock.
2. Early expiry: lease renewal starts when the first target installs a fence. A synchronized target set incorporates later handoff targets. Added a short-TTL A/B test where B drain cannot finish until A has renewed.
3. Error identity: failed-drain cleanup is best effort and returns the original cause unchanged. Added an `ErrBackendClosed` assertion using `moerr.IsMoErrCode`, including a simultaneous cleanup failure.
4. Rollback latency: cleanup RPCs use a separate 5-second overall deadline and run concurrently. Failures are logged and TTL remains the fail-closed fallback. Added a context-aware blackhole test plus a lease Stop test proving cancellation and join of an in-flight renew, so cleanup cannot be followed by a stale renew.
5. SCA: the Go 1.22 loop copy was already removed in `7f8af5d0e8`; targeted golangci-lint now reports 0 issues.

I deliberately did not add a durable catalog fence or task ownership epoch in this bugfix. A full CN process loss still loses its in-memory registry until the next renewal arrives, and eliminating the final cross-process handoff window completely requires a taskservice/executor protocol change. That is materially larger than the observed assignment-before-executor-readiness bug and should be designed separately rather than hidden inside this PR.

Validation on the new head:
- full `./pkg/iscp ./pkg/cnservice ./pkg/sql/compile` with `-race`: pass
- focused lifecycle tests with `-race -count=10`: pass
- build and vet for all three packages: pass
- targeted golangci-lint: 0 issues
- `make err-check`: pass
- `git diff --check`: pass

XuPeng-SH:
> Deep review of `d4d41bda3d42f3b94b9d39274be227198a6f7bb6`: I found four code issues plus the current SCA blocker. I do not think this head is ready to merge.
> 
> 1. **[P1] Re-observing the same CN UUID is not an ownership/fence linearization point.**
>    `fencedRunners` is keyed only by CN UUID and the loop exits as soon as a fresh read returns an already-seen UUID. The fence, however, lives in a specific in-memory `ISCPTaskExecutor` generation. After fencing `CN-A/generation-1`, the executor can restart on the same CN with an empty fence map; the task row still returns `CN-A`, so the loop exits without fencing generation 2. There is also an unavoidable window after the final owner read where the task can move from A to B before the DDL transaction terminates. The lease only renews the saved executor pointer/query address and never follows the new owner.
>    This needs a durable job fence or an ownership epoch/token coupled to task assignment and checked by every executor generation when registering a consumer. Seeing the same UUID twice is not sufficient.
> 2. **[P1] Earlier runner fences can expire before the lease is started.**
>    Successful targets are only appended to `fencedTargets`; `startISCPJobFenceLease` is called after the entire owner-convergence loop. The actual drain intentionally uses the statement context and may legitimately take longer than the readiness deadline. For example, A is fenced, ownership moves to B, and draining B takes longer than the 30-minute fence TTL. A's fence expires while waiting for B, allowing its consumer to restart. If ownership cycles back to A, the UUID set still causes the loop to exit even though A is no longer fenced.
>    Renewal must start as soon as the first target is fenced and safely incorporate later targets. Please add a short-TTL A/B handoff test with a long second drain.
> 3. **[P1] `errors.Join` destroys the primary `moerr` identity on failed drain cleanup.**
>    `cleanupAfterFailure` always returns `errors.Join(cause, cleanupErr)`. Even when cleanup succeeds and `cleanupErr` is nil, this returns a `joinError`. `moerr.IsMoErrCode`, commit-result-unknown handling, and frontend whole-transaction rollback classification require the error itself to be `*moerr.Error`; they do not unwrap joined errors. A first remote drain failure such as `ErrBackendClosed` is already wrapped because the uncertain target is included, so ALTER/DROP INDEX in an explicit transaction can take statement-only rollback instead of the required whole-transaction rollback.
>    Return `cause` unchanged when cleanup succeeds. If cleanup also fails, preserve the primary `moerr` identity or update the relevant classifiers to unwrap. Tests should assert `moerr.IsMoErrCode`, not only `ErrorContains`.
> 4. **[P2] Rollback can synchronously block for up to 30 minutes.**
>    The rollback cost-event callback creates a background context with `iscp.DefaultRollbackFenceTTL` and waits for all target operations via `wg.Wait`. The generic query client has no shorter default backend read timeout, so one connected but unresponsive old runner can hold the transaction terminal path for the full 30 minutes. Fence TTL and RPC cleanup timeout should be separate; use a short bounded cleanup deadline, log cleanup failures, and rely on TTL for fail-closed expiry. A blackholed-target rollback latency test would cover this.
> 5. **Current CI blocker:** `Matrixone CI / SCA Test on Linux/arm64` fails at `pkg/sql/compile/iscp_util.go:523` with `copyloopvar`: the `i := i` copy can be deleted under Go 1.22+.
> 
> Independent validation on this exact head:
> 
> * focused `pkg/sql/compile` drain/target tests: `-race -count=10` passed
> * focused `pkg/cnservice` drain-handler tests: `-race -count=10` passed
> * `git diff --check` passed
> 
> The previous sequential multi-target renewal/removal deadline issue is fixed by the concurrent operation helper in this head; the findings above are additional closure gaps.

@ck89119 please review again

XuPeng-SH:
Follow-up safety correction in `ff0a3faf2b` (rebased onto current `origin/main` `3128259d4b`): the previous head made remote renewal an upsert. That could let a delayed renew processed after rollback/remove recreate a fence and block the consumer until TTL. Renewal is now strictly conditional and removal is terminal.

The in-process dual-copy state is also linearized under one lock order:
- renew requires an existing CN fence and, when an executor is published, an existing executor-local fence; it updates both together
- remove deletes the CN fence and the currently published executor copy together
- a stale executor pointer cannot recreate state after removal
- publishing/re-publishing an executor clears and rebuilds its local fence map from the CN registry, so an old object cannot reintroduce removed state
- the executor-not-yet-published path can still renew the already-installed pending CN fence

New deterministic regressions cover renew-before-initial-drain rejection, active renewal, remove-then-late-renew rejection, stale executor renewal, old-executor re-publication, and pending-executor renewal.

Post-rebase validation:
- full `pkg/iscp`, `pkg/cnservice`, and `pkg/sql/compile` with `-race`: pass
- focused ordering/generation regressions with `-race -count=50`: pass
- build and vet: pass
- targeted golangci-lint: 0 issues
- `make err-check` and `git diff --check`: pass

XuPeng-SH (commented):
Latest-head update from another deep pass: I do not see new blockers beyond the remaining ownership-linearization gap already under discussion, but I also do not see that gap closed on the current head.

The convergence condition is still `if _, ok := fencedRunners[runnerCN]; ok { break }` after a fresh runner read (`pkg/sql/compile/iscp_util.go:359-363`). That proves only that the current CN UUID was seen/fenced before; it does **not** prove that no different runner owned the daemon task in between. The new tests cover `A -> B -> B` handoff and same-owner revalidation after a long drain (`pkg/sql/compile/iscp_util_extra_test.go:560-639`, `924-997`), but they still do not cover `A -> B -> A`. A transient `B` owner can still admit a consumer because fences are CN-scoped and an already-running consumer is only canceled on a later failed lease check, not immediately on ownership change (`pkg/iscp/executor.go:516-523`, `pkg/iscp/iteration.go:517-539`).

So my latest-head conclusion is: the renew/remove/generation fixes look good, but the drain loop still lacks a true ownership linearization point / terminal quiescence proof.

Issue #26185, opened 2026-07-26
observability: establish bounded structured logging contract and pilot adoption


XuPeng-SH:
## Goal

Establish one small, reusable logging contract for new and migrated MatrixOne diagnostics, then validate it in a few representative high-value control/data paths. This work improves incident evidence and log safety; it does **not** claim to fix the INSERT DUPLICATE memory root cause in #26172.

## This PR scope

- `logutil.Event`: a stable event name/message contract with bounded per-event output and occurrence/suppression evidence.
- Lazy field builders: no formatting, hashing, serialization, or scans until level and budget admission.
- Standard non-retained correlation fields for sensitive/unbounded strings and errors.
- Pilot use in ObjectIO invariant handling, CCPR retry handling, and Frontend CDC control-plane transitions.
- Documentation and tests that make the policy actionable for developers and operators.

## Non-goals

- No repository-wide mechanical log rewrite.
- Legacy `logutil.Info/Warn/Error` remains supported during gradual migration.
- No new root-cause claim for #26172 without profile evidence.

## Follow-ups

- [ ] Migrate retry/queue/connection loops chosen from incident data, one component at a time.
- [ ] Extend service-local `MOLogger` to the same event contract where a component logger is available.
- [ ] Define retention/redaction exceptions with Operations/SRE before raw diagnostic artifacts are retained.
- [ ] Use the #26172 profile timeline to choose the next logging points only if they improve allocation/latency causality.

Related: #26172

XuPeng-SH:
Pilot implementation is ready for review in #26187. The PR is intentionally limited to the shared contract plus ObjectIO, CCPR, and Frontend CDC exemplars; the follow-up checklist in this issue remains the migration boundary.

XuPeng-SH:
## Historical branch disposition

The earlier local `codex/observability-log-hardening` branch is preserved; it is **not** discarded and will not be force-squashed into the pilot PR. Inventory against current `main`: **248 commits, 443 files, +16,765/-9,249 lines**. It was built on an older main and includes a different logging-foundation implementation, so it must be extracted/revalidated against the current `logutil.Event` contract rather than cherry-picked wholesale.

### Measured module buckets

| Batch | Changed files | Handling |
| --- | ---: | --- |
| TAE | 117 | Separate storage-engine review and test matrix; do not mix with foundation work. |
| SQL/index/ISCP | 95 | Separate execution/index-plugin review; preserve GPU/CPU gates. |
[...]

PR #26187, opened 2026-07-26
feat(logutil): add bounded structured event contract

### LINKED ISSUE #26172: [Bug] sysbench INSERT DUPLICATE 高并发路径内存分配（1000 并发 heap +16G 导致连接超时）
## 背景

sysbench 1000w `simple_insert_duplicate` 测试在 nightly regression 中频繁失败，main 和 4.2-dev 均受影响。

**CI 现象：**
- Main (07-25): https://github.com/matrixorigin/mo-nightly-regression/actions/runs/30143464491/job/89660348510
- 4.2-dev (07-24): https://github.com/matrixorigin/mo-nightly-regression/actions/runs/30111505702/job/89660467002

**测试结果：**
- 总操作 675,249，错误 5~10（均为 `Communications link failure` / `i/o timeout`）
- 成功率 99.999%，但因 `EXP_RATE=1.0` 判定 FAILED

## 问题

3 CN（14 cores / 55G each）跑 1000 并发 INSERT DUPLICATE 时，单 CN 资源在临界点：

### 内存分析（cn-qsp54，07-25 failing run）

| 指标 | 测试前 | 峰值 | 说明 |
|---|---|---|---|
| RSS | 35G | **45G** | 接近 55G limit |
| Go heap_alloc | 4G | **20G** | **+16G spike** |
| Go heap_sys | 36G | 39G | 预分配虚拟内存 |
| Go heap_idle | 17G | 34G | 已归还 OS 的保留内存 |
| go_sys | - | 40G | Go runtime 总占用 |

### CPU 分析

| CN | avg | peak / 14 cores |
|---|---|---|
| qsp54 | 5.66 (40%) | **12.43 (89%)** |
| qfszt | 5.49 (39%) | **13.23 (94%)** |
| 89qsg | 4.25 (30%) | **10.65 (76%)** |

### GC 不是瓶颈

- GC pause p75: 0.8ms
- GC time rate: avg 0.7ms/s, peak 2.4ms/s

### 事务吞吐

- 单 CN peak: **3,205 txn/s**（DUPLICATE 冲突导致大量短事务快速回滚）

### 对比通过 vs 失败

| | 07-23 通过 | 07-25 失败 |
|---|---|---|
| heap_alloc peak | ~16G | **~20G** |
| RSS peak | 41G | **45G** |
| CPU peak | 92% | **94%** |

差异仅 4G / 2%，说明系统运行在悬崖边上。

## 核心问题

**INSERT DUPLICATE 路径在 1000 并发下 heap 从 4G 飙到 20G（+16G），3 分钟内分配了 16G 的活跃堆内存。** 可能的分配热点：

1. **Duplicate key 检测** — 1000 并发同时做 duplicate 检测，大量临时对象
2. **事务元数据** — 每个短事务的锁结构、undo log 等 per-txn 开销
3. **Batch 处理** — 并发 batch 同时物化驻留内存
4. **冲突回滚** — duplicate 冲突回滚路径的内存分配

## 建议

1. **Profile 分配热点** — 用 malloc profile 定位 INSERT DUPLICATE 路径的具体分配热点函数
2. **buffer pool / 预分配** — duplicate 检测用预分配 buffer，减少频繁 alloc
3. **并发 batch 限流** — 限制单 CN 上同时处理的 batch 数量
4. **per-txn 内存优化** — 减少短事务的元数据开销
5. **GOMEMLIMIT** — 考虑设置 GOMEMLIMIT 让 Go 更积极回收，控制 RSS 上界

### Main

| 日期 | 结果 | 备注 |
|---|---|---|
| 07-25 | ❌ | insert_duplicate 超时 |
| 07-23 | ✅ | |
| 07-21 | ✅ | |
| 07-20 | ✅ | |
| 07-19 | ✅ | |
| 07-18 | ✅ | |

### 4.2-dev

| 日期 | 结果 | 备注 |
|---|---|---|
| 07-24 | ❌ | insert_duplicate 超时 |
| 07-23 | ✅ | |
| 07-22 | ❌ | insert_duplicate 超时 |
| 07-21 | ✅ | |
| 07-20 | ✅ | |
| 07-18 | ❌ ×2 | CN busy / 超时 |
| 07-17 | ❌ | CN busy |

4.2-dev 上更频繁，可能与分支代码差异有关。

XuPeng-SH:
## Summary

This PR deliberately contains both the common logging contract and a closed first operational adoption batch; it is not an API-only pilot.

### Contract

- Adds `logutil.Event`: stable event/message pair, bounded event budget, and `occurrence` / `suppressed` evidence.
- Makes global Event, `RateLimitedLogger`, and service-local `MOLogger` use the same bounded `EventRateLimiter`; `With`, `Named`, `WithContext`, and module loggers share one budget.
- Adds lazy field builders plus non-retained string/error correlation fields. Disabled lazy paths allocate zero heap objects; suppressed paths do not build fields.
- Legacy `MOLogger.Debug` and direct `Log` retain their original unsampled behavior. Only explicit Event APIs use the event budget.

### Closed first operational batch

- ObjectIO destination-vector invariant.
- CCPR retry lifecycle: all retry call sites carry a stable operation; retry, exhaustion, lease-owner mismatch, lease-stop, task submit, and iteration-state failures share the event/field contract.
- CDC restart lifecycle: request persistence leaves catalog state unchanged; local restart waits for replacement readiness then CASes `RestartRequested + observed runner` to daemon `Running`. Stale/unassigned restart work goes through a heartbeat-fenced start claim that also requires `RestartRequested`, so newer pause/cancel/drop requests cannot be overwritten. Catalog startup state is CAS-guarded: local restart uses its exact pre-restart state; fresh takeover permits only `running`, `failed`, or `paused`, never pausing/deleted state.
- Publication executor startup records the original failure as a lazy fingerprinted event, then returns a stable generic error so taskservice cannot re-log or persist dynamic raw error text.
- Connection-close transitions and proxy CN health, backend connect/disconnect/dial/handshake, and rebalance queue pressure.

## Scope

Legacy global APIs remain compatible. The historical 443-file candidate branch is tracked and decomposed in #26185; TAE/DistTAE and SQL/index/ISCP remain independent follow-up reviews rather than being hidden in this PR. This does not claim to fix #26172 without profile evidence.

## Verification

Rebased onto `origin/main` at `fef2b645d4`; head is `af4fa56467`.

- `GOWORK=off go vet -mod=readonly ./pkg/taskservice`
- `GOWORK=off go build -mod=readonly ./pkg/cdc ./pkg/frontend ./pkg/taskservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=180s ./pkg/frontend` selected CDC restart, callback-fence, catalog-transition tests
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=180s ./pkg/taskservice` selected local/empty/stale-owner restart and control-request-CAS tests
- `git diff --check`

A direct local `go vet` over frontend is currently blocked before package analysis by the environment missing `cppjieba/Jieba.hpp`; the CGo build/test wrapper above is the project-supported local test path.


XuPeng-SH (commented):
Deep review on the latest head, with emphasis on abstraction quality, logging engineering quality, and performance. I still see three merge-blocking issues in the current pilot:

1. **The abstraction is internally inconsistent: `MOLogger.Debug` now collapses every legacy debug call into one synthetic event population.**
   The new contract says one event name identifies exactly one operational outcome and one rate-limit population (`pkg/logutil/event.go:22-25`). But `pkg/common/log/logger.go:128-129,216-218` rewrites every legacy debug call to `Event{Name: "legacy-debug", Message: msg}`. Because the limiter is intentionally shared across derived loggers (`pkg/common/log/logger_test.go:175-194`), unrelated debug streams on the same service now compete for one `legacy-debug` budget. Frontend sessions all share the service logger (`pkg/common/runtime/runtime.go:44-53,91-100`, `pkg/frontend/session.go:2262-2287`), so one noisy path can suppress unrelated debug diagnostics service-wide. This is both an abstraction break and an operational/performance regression.
   **Fix:** keep legacy `Debug` / debug `Log` behavior unchanged, and reserve the bounded event contract for explicit `*Event*` call sites with one stable event constant per real outcome.

2. **The “Frontend CDC restart” pilot is not functionally closed.**
   `handleRestartCdc` emits fingerprinted `frontend.cdc.restart.*` events (`pkg/frontend/cdc_handle.go:74-90`), but the same restart flow immediately falls back to legacy shared logs that still retain raw `task-name`, raw `service`, and raw `zap.Error(err)` through `handleUpdateCDCTaskRequest -> doUpdateCDCTask -> onPreUpdateCDCTasks` (`cdc_handle.go:182-299`, `312-445`). The doc explicitly lists “Frontend CDC restart” in the initial adoption batch and says shared logs must not retain raw identifiers or untrusted error text by default (`docs/observability/logging.md:69-89`). So the pilot does not actually satisfy the contract it claims to adopt.
   **Fix:** either migrate the full restart closure to bounded fingerprinted events, or remove CDC restart from this pilot/docs until that closure is complete.

3. **Publication retry unhappy paths still re-log raw IDs and raw errors after the new structured retry events.**
   `retryPublication` fingerprints retry/exhaustion errors (`pkg/publication/executor.go:1390-1414`), but callers and lease helpers still emit raw `cnUUID`, `runner`, `taskID`, and raw `zap.Error(err)` (`pkg/publication/util.go:38-45,91-95`, `pkg/publication/worker.go:534-552`). The lease-loss path (`ok=false, err=nil`) bypasses the new retry events entirely and only writes the raw mismatch log. That means the pilot does not actually give a bounded logging contract on the unhappy path, and it also double-logs the same failure with both structured event output and legacy raw output.
   **Fix:** convert the post-retry caller logs and the `ok=false, err=nil` lease-loss path to dedicated event-based fingerprinted logs, and avoid re-logging raw IDs / raw `err` after `retryPublication` has already emitted the structured event.

Current conclusion: the idea is promising, but on the latest head the abstraction boundary is still leaky and the pilot migrations are not yet closed enough for the contract being introduced.

XuPeng-SH:
Addressed the review findings in 210970de5c:

- Restored legacy `MOLogger.Debug` and direct Debug `Log` behavior; only explicit Event APIs are budgeted. Added a regression test proving legacy Debug does not create limiter state.
- Completed the CDC restart chain (request → task-service → lookup → status update → terminal outcome) and the CCPR retry/lease chain (all retry callers, retry/exhaustion, owner mismatch, executor stop, submit/state-update failures). Dynamic identifiers and errors are emitted only as lazy fingerprint fields.
- Replaced the two combined slice-capacity computations flagged by CodeQL, preserving caller-slice isolation without overflow-prone arithmetic.

Rebased onto c8d29e9008. Race/build/vet and targeted CGo CDC, CCPR, and proxy tests are listed in the PR body.

XuPeng-SH (commented):
Latest-head update from another deep pass: I still do not think this pilot is ready to merge. The current blockers on the latest head are:

1. **`MOLogger.Debug` / debug `Log` still collapse all legacy debug streams into one synthetic `legacy-debug` event population.** `pkg/common/log/logger.go:128-129,216-218` is unchanged here, so the abstraction still violates the contract in `pkg/logutil/event.go:22-25` (one event name = one operational outcome / one rate-limit population). Because the limiter is shared across derived loggers, unrelated debug paths on the same service still compete for one budget.

2. **`frontend.cdc.restart.completed` is emitted before the restart actually happens.** `handleRestartCdc` logs `...completed` on any nil return from `handleUpdateCDCTaskRequest` (`pkg/frontend/cdc_handle.go:74-90`), but `doUpdateCDCTask` still returns `nil` when `TaskService` is absent (`cdc_handle.go:246-252`). More generally, this handler only means the request/update path returned; the real restart is asynchronous and happens later in taskservice / executor code. So the event currently overclaims terminal success.

3. **The CDC restart pilot is still not functionally closed.** The entrypoint is fingerprinted now, but the downstream restart closure still emits raw `task-id` / `task-name` / runner IDs and raw `zap.Error(err)` in taskservice restart handling (`pkg/taskservice/daemon_task.go:186-305,784-790`) and task-state updates (`pkg/taskservice/mysql_task_storage.go:1910-1930`). Since the docs still list Frontend CDC restart in the initial adoption batch (`docs/observability/logging.md:87-89`), this is still a contract breach, not just a deferred migration.

4. **The CCPR/publication retry unhappy path is still not closed either.** The retry helper emits fingerprinted events, but the same path still re-logs raw IDs and raw `err` afterward (`pkg/publication/util.go:38-45,91-95`, `pkg/publication/worker.go:546-552`). The docs still include CCPR retries in the initial adoption batch (`logging.md:87-89`), so the pilot still double-logs and still leaks raw identifiers/error text on the unhappy path.

So my latest-head conclusion is: the bounded structured event contract is still leaky in both abstraction and pilot closure, and the new `...restart.completed` event currently overstates success semantics.

XuPeng-SH:
Adopted the valid latest-head CDC findings in 8902203a80:

- `frontend.cdc.restart.completed` was removed. Frontend now records only request acceptance/persistence; it never claims asynchronous restart completion.
- The closure now continues through `pkg/taskservice`: `RestartRequested` persistence, enqueue/no-candidate, query/status/runner branches, active-routine invocation, and the actual `ActiveRoutine.Restart()` success/failure are stable budgeted events with lazy fingerprint fields for task name/ID, account, runners, and errors.
- Targeted CGo frontend and taskservice restart tests pass, including direct failure branches and successful local/empty-runner restart paths.

The Debug and CCPR findings quoted in this review were against the older head: legacy Debug was restored with a regression test, and CCPR retry/lease callers already use fingerprinted events in the preceding update. They remain verified in this head.

XuPeng-SH (commented):
Latest-head update from another deep pass: the earlier `legacy-debug` abstraction issue now looks fixed (`pkg/common/log/logger.go:128-130,216-218` goes back to raw debug logging), but I still see three current blockers on the latest head:

1. **CDC restart still reports success/completion before the replacement generation is actually running.** `CDCTaskExecutor.Restart` logs `cdc.frontend.task.restart_success` on defer before the real restart work runs in the async goroutine (`pkg/frontend/cdc_exector.go:459-505`), and `restartTask.Handle` treats that immediate `nil` return as `taskservice.cdc.restart.completed` (`pkg/taskservice/daemon_task.go:305-320`). So a later `startFunc` failure still leaves a false-positive lifecycle record.

2. **The CDC restart pilot is still not functionally closed.** The task row is flipped to `Running` before async restart admission actually succeeds (`pkg/taskservice/daemon_task.go:273-277`), so a later async start failure has no retry path. At the same time, `restartTasks()` still includes lagged foreign-runner candidates (`daemon_task.go:782-795`), but `Handle()` rejects any non-empty foreign `TaskRunner` (`daemon_task.go:251-256`), and the generic task runner still re-logs the returned raw error via `failed to handle task` (`daemon_task.go:894-895`).

3. **Publication startup retry exhaustion still escapes the bounded structured contract.** The retry helper fingerprints retry/exhaustion inside `retryPublication` (`pkg/publication/executor.go:1407-1431`), but the exhausted error still escapes unchanged from executor startup (`executor.go:160-161`) into daemon-task handling (`pkg/taskservice/daemon_task.go:105-107`), which then logs raw `zap.Error(err)` and persists `err.Error()` in task details (`daemon_task.go:971-978`). Since the docs still include “CCPR retry and lease control” in the initial adoption batch (`docs/observability/logging.md:87-89`), this unhappy path is still outside the bounded contract.

So my latest-head conclusion is: one abstraction issue is fixed, but the CDC restart success semantics, the CDC restart closure, and the publication startup failure escape path still keep this pilot from being contract-clean.

XuPeng-SH:
Adopted the latest valid review findings in 0662ea80e4 (rebased onto fef2b645d4):

- CDC no longer emits success from request acceptance or goroutine launch. `CDCTaskExecutor.Restart` allocates a generation-scoped, buffered readiness waiter and returns only when the replacement reports `Running` or startup failure. The waiter is bounded; timeout fences the generation so a late Start cannot publish `Running`.
- Taskservice now calls `Restart` before persisting `TaskStatus_Running`. A failed replacement keeps the row `RestartRequested`; a focused regression asserts that retryability. Returned handler errors are stable generic errors, while the original cause is captured once as a lazy fingerprinted event, preventing the generic daemon runner from re-logging/persisting raw dynamic text.
- Restart dispatch is now consistent with the handler: only local and unassigned requests are selected. Stale foreign ownership is deliberately excluded because local restart is not a valid ownership transfer; #26185 records it as the explicit follow-up.
- Publication executor startup records its real error in `publication.executor.start-failed` with lazy fingerprint fields and returns a stable generic error to the task runner, closing the raw retry-exhaustion escape path.

Rebase-post verification passed: logutil/common-log race tests; frontend restart race test; taskservice local/empty-runner and failure paths; and publication retry/startup tests. The PR body has the exact commands and the local standalone-vet environment limitation.

XuPeng-SH (commented):
Latest-head update from another deep pass: the previous `legacy-debug` concern and the publication startup raw-error escape path now look fixed on this head, but I still see two CDC correctness blockers.

1. **Restart requests on a dead runner are now orphaned.** `restartTasks()` now only polls `RestartRequested` tasks whose `TaskRunner` is local or empty (`pkg/taskservice/daemon_task.go:786-798`). But the restart request persistence path still updates only `TaskStatus` and keeps the existing foreign `TaskRunner` (`pkg/taskservice/mysql_task_storage.go:1924-1942`). So if runner A owns the task, the user issues `RESTART CDC`, and A dies before handling it, every other runner will ignore that row forever. I understand the comment that stale foreign ownership transfer is a follow-up, but as the code stands the persisted restart request can be stranded indefinitely.

2. **The request path still publishes `running` before the replacement generation reaches its ready point.** `onPreUpdateCDCTasks()` still chooses `cdc.CDCState_Running` for restart and persists it synchronously via `PrepareUpdateTask(...)` (`pkg/frontend/cdc_handle.go:477-519`). The executor does not publish readiness until after `TransitionStartSuccess` / `finishRestartWaiter(..., nil)` (`pkg/frontend/cdc_exector.go:388-406`, `547-580`). So even on this head, a restart can be visible as `running` in catalog-facing state before the replacement generation is actually admitted.

My latest-head conclusion is therefore: the logging-contract closure is much better now, but these two CDC restart correctness gaps still keep the PR from being done.

gouhongshen (changes requested):
## Codex automated review

Loaded all existing reviews, issue comments, and inline comments; excluded the already-reported CDC concerns. One distinct restart control-plane race remains.

### P1 - Do not overwrite a concurrent CDC control request after restart readiness (pkg/taskservice/daemon_task.go:304)
This now waits for ActiveRoutine.Restart() to finish before persisting Running, but writes the task snapshot read at line 221 with no status/runner condition and ignores the affected-row count. During that wait (up to four seconds), DROP CDC can validly persist CancelRequested; this stale write then restores Running, so the cancel handler no longer sees the request and the CDC task can continue after a successful drop. Previously Running was persisted before Restart(), so a later control request won. Make this update a compare-and-set from RestartRequested (and ownership as appropriate), and treat zero affected rows as a superseded control request rather than overwriting it. No test covers this interleaving.

XuPeng-SH:
Adopted both latest CDC blockers in 726ef60a2d:

1. **Dead runner no longer strands a restart request.** `restartTasks()` again includes heartbeat-stale foreign owners. Dispatch now separates the semantics: an already-local task uses in-process `Restart`; unassigned or stale-foreign work has no local active routine, so it enters `newStartTask` and is atomically claimed by the existing `startDaemonTask` heartbeat condition. `TestRestartDaemonTaskTakesOverStaleForeignRunner` proves the row reaches `Running` with the new runner ID.

2. **Catalog `running` is no longer written at request time.** `onPreUpdateCDCTasks` now records the restart request but deliberately leaves `mo_cdc_task.state` unchanged. The replacement Start transitions the catalog only after its ready point. The restart waiter carries the prior catalog state (`running`, `failed`, or `paused`) so the ready/error transition remains optimistic and does not overwrite a concurrent pause. `Test_updateCdcTask_restart` now asserts that no catalog state update occurs in the request transaction, and `TestCDCTaskUpdateErrMsgUsesRestartCatalogState` covers failed/paused recovery.

Also fixed the reported deterministic lock cycle: `callbackMu` now ends immediately after state/generation fencing and waiter installation, before cleanup or readiness wait. The replacement can therefore execute `handleNewTablesForGeneration`'s `RLock`; `TestRestartReleasesCallbackFenceBeforeWaitingForReplacement` and the original permanent-table-error restart test cover this.

Fresh `-race` CGo frontend/taskservice regressions and taskservice vet pass; exact commands are in the PR body.

gouhongshen (changes requested):
## Codex automated review

Previously blocking topics: legacy Debug behavior, CDC readiness-before-Running, raw publication failure escape, dead-runner discovery, premature catalog `running`, and the callback lock cycle are fixed. The prior local post-readiness control-request overwrite remains unmodified (not duplicated below per follow-up rules). The new stale-owner handoff adds the two distinct P1 failures below. Focused CGo tests could not run because `go` is absent from the environment.

### P1 - Carry the prior catalog state through stale-owner restart takeover (pkg/frontend/cdc_exector.go:194)
The new stale-foreign path dispatches `RestartRequested` work via `newStartTask` (`pkg/taskservice/daemon_task.go:641-649`), creating a fresh `CDCTaskExecutor` rather than calling `Restart()`. That executor has generation 0 and no `restartCatalogState` entry, so this helper defaults its optimistic catalog transition to `running`. The request path now deliberately leaves the catalog unchanged. Thus a paused or failed CDC whose former owner has died is started successfully by the new runner, but `Start` attempts `... WHERE state = 'running'`; the zero-row error is only logged as `clearErrMsgErr` and startup continues. The daemon task becomes Running while `mo_cdc_task` remains paused/failed. This is reachable because restart accepts those states, and the new local-restart test explicitly recognizes both as valid prior states. Preserve/carry the catalog state for takeover startup (or read and CAS it at admission) and make the handoff test cover paused and failed catalog rows.

### P1 - Do not let a stale restart claim overwrite a newer control request (pkg/taskservice/daemon_task.go:977)
This is distinct from the previously reported local `restartTask` post-readiness race: the new delta routes a stale foreign `RestartRequested` row into the generic start path. After it is selected at `dispatchTaskHandle` lines 636-649, a concurrent DROP/CANCEL can persist `CancelRequested` while retaining the stale heartbeat. `startDaemonTask` then writes its old snapshot as `Running` and claims ownership with only `WithLastHeartbeat`; it does not require `RestartRequested` (or otherwise compare the control state). The write succeeds and the newer cancel is lost, so the new executor can run after a successful drop. Claim this restart handoff with a CAS on both stale ownership and `RestartRequested`, treating zero affected rows as a superseded request; add an interleaving regression test.

XuPeng-SH:
Adopted the latest automated review findings in af4fa56467:

- **Local restart post-readiness CAS:** after `ActiveRoutine.Restart()` returns ready, `restartTask.Handle` now updates the daemon row only when it is still `RestartRequested` and still owned by the runner observed in its initial read. A zero-row update is a superseded control request and returns without emitting completion. `TestRestartTaskDoesNotOverwriteSupersedingControlRequest` interleaves `CancelRequested` during Restart and proves it survives.

- **Stale-owner claim CAS:** stale/unassigned restart work is marked `restartClaim`; `startDaemonTask` then requires both the stale-heartbeat predicate and `TaskStatus_RestartRequested`. Thus a concurrent DROP/CANCEL cannot be overwritten by the generic start path. `TestRestartStartClaimDoesNotOverwriteSupersedingControlRequest` proves the claim returns false and leaves `CancelRequested` intact.

- **Catalog state during fresh takeover:** a fresh executor has no in-memory restart waiter, so startup now uses a constrained catalog CAS over the only valid pre-start states: `running`, `failed`, and `paused`. Pause/delete states do not match and cannot be overwritten. This restores failed/paused catalog state to `running` after a successful stale-owner replacement; `TestCDCTaskStartupUpdateAllowsActiveCatalogStates` covers the generated guarded transition.

Fresh build, taskservice vet, and CGo `-race` frontend/taskservice regressions passed before the push.

XuPeng-SH:
已采纳最新两条 P1，并已推送 `af4fa56467`。

- 本地重启：`Restart()` ready 返回后持久化 `Running` 改为 CAS：`status=RestartRequested && task_runner=read-time runner`。若受影响行数为 0，表示 DROP/CANCEL 等后续控制请求已经线性化，记录 `restart-request-superseded` 后不反写。新增 `TestRestartTaskDoesNotOverwriteSupersedingControlRequest`。
- stale foreign handoff：专用 restart-start 路径在原有过期 heartbeat claim 上追加 `status=RestartRequested` CAS；零行不启动本地 executor。新增 `TestRestartStartClaimDoesNotOverwriteSupersedingControlRequest`。
- 同时补齐 handoff 的 catalog state：新 runner 启动成功时只在 `running|failed|paused` 状态集内 CAS 更新为 running，避免 paused/failed 留在旧状态；有 generation 的本地 restart 仍精确按它捕获的原状态更新。

已通过 `git diff --check`、taskservice vet，以及 frontend/taskservice 的针对性 CGo `-race` 回归测试。当前 head 的 CI 已重新触发，等待远端结果。

gouhongshen (changes requested):
## Codex automated review

Earlier blockers around legacy Debug, CDC/publication logging closure, restart readiness, dead-runner discovery, lock ordering, and daemon CAS races are fixed. The prior paused/failed takeover mismatch is addressed, but the new active-state CAS leaves a distinct concurrent-pause overwrite race.

### P1 - Fresh stale-owner startup can overwrite a newer completed pause (pkg/frontend/cdc_exector.go:1119)
Author response: the takeover fix intentionally permits catalog states running, failed, and paused while excluding pausing/deleted states. Why this remains blocking: that predicate is not tied to the restart admission or control generation. A stale-owner restart claims the daemon row and starts a fresh executor; if a concurrent PAUSE completes while startup is delayed, the catalog becomes paused, but this UPDATE still matches `state IN (..., 'paused')` and changes it back to running. The daemon task can remain Paused while the catalog says Running. This is distinct from the prior finding about failing to transition a pre-existing paused/failed row. Guard startup with an admission generation/token or serialize/check the newer control request; a bare active-state predicate cannot distinguish an old paused state from a newer pause.

XuPeng-SH:
已采纳最新 P1，提交 `726805675e`。

此前为让 stale-owner takeover 恢复 paused/failed catalog 行，采用了 `running|failed|paused` 状态集合 CAS；该集合确实无法区分“重启前已暂停”和“重启 admission 后新完成的 PAUSE”。现改为显式、持久化的 admission fencing：

1. RESTART 的同一请求事务将 `mo_cdc_task.state` 标为 `restarting`，它表示已被接纳、但 replacement 尚未 ready，绝不提前声称 `running`。
2. local restart waiter 及 stale/unassigned fresh takeover 都只允许从精确的 `restarting` 转为 `running`/`failed`。fresh 路径通过 `taskservice.WithRestartAdmission` 显式传递已成功的 restart claim；普通 CREATE 启动仍只接受 `running`，不能替一个后到的 restart 完成状态发布。
3. 若 PAUSE 在 admission 后完成，它先变为 `pausing` 再为 `paused`；旧 startup 的 `WHERE state = restarting` 受影响行数为 0，保留 pause 的最终状态。DROP 删除行同样不会被反写。

新增 `TestCDCTaskRestartStartupDoesNotOverwriteCompletedPause`、精确 admission-state SQL 测试，以及 stale restart context 传递测试。`-race` 的 frontend/taskservice focused regressions、build、taskservice vet 均通过。

gouhongshen (approved):
## Codex automated review

The prior P1 is fixed: restart admission is durably fenced as `restarting`, and local/fresh startup can publish only from that exact state, preserving a completed PAUSE or DROP. Earlier blockers (legacy Debug behavior, CDC readiness/ownership/CAS/lock issues, and publication error escape) remain fixed; no blocking or non-blocking findings remain. Focused tests could not run locally because `go` is unavailable.

XuPeng-SH:
Blocking notes from my unhappy-path review (GitHub won't let me use **Request changes** on my own PR, so I'm leaving the blocking findings here instead).

1. **The new local-restart timeout path is not generation-safe.** `Restart()` can now return after the 4s timeout while the spawned `Start()` goroutine is still unwinding (`pkg/frontend/cdc_exector.go:579-604`). When that late `Start()` eventually fails, its deferred cleanup still mutates shared lifecycle state (`detector.UnRegister(taskId)`, `stateMachine.SetFailed`, `activeRoutine.Close*`, and the startup err-msg path in `pkg/frontend/cdc_exector.go:274-302`) without re-checking that it still owns the current generation. On the taskservice side, a failed local restart returns before the daemon-task row is moved away from `RestartRequested` (`pkg/taskservice/daemon_task.go:318-339`), so another poll can enqueue a fresh restart while the previous `Start()` goroutine is still unwinding. Because `TableDetector` ownership is keyed only by task id (`pkg/cdc/table_scanner.go:329-377`), the stale generation can still tear down or mark failed a newer attempt.

2. **A timed-out local restart can leave `mo_cdc_task.state` stuck at `restarting` and drop the actual startup error.** The waiter metadata is removed when `Restart()` returns (`pkg/frontend/cdc_exector.go:540-542`), but the timed-out `Start()` goroutine can fail later. At that point `updateErrMsgForStartup()` no longer has either `hasRestartCatalogState` or `restartAdmission`, so it falls back to `currentState == running` (`pkg/frontend/cdc_exector.go:300-302,1094-1107`). That CAS cannot move the catalog row from `restarting` to `failed`, so the row stays in `restarting` and the late startup error text is never persisted.

Suggested fix direction: make timeout-path cleanup generation-owned, and keep the `Restarting` catalog admission context alive until that specific `Start()` goroutine exits.

XuPeng-SH:
已在 d04d7f5152 修复并 rebase 到当前 PR head。

1. 每个 Start 现在有不可变 generation、cancel 和 done；Restart 先 cancel/drain 旧 attempt，未退出则直接失败且不启动 replacement。旧 attempt 的 deferred cleanup 仅清理其持有资源，不能再写 state/metrics/catalog/waiter。
2. timeout 立即以 restarting -> failed CAS 落库；重试以 failed -> restarting CAS 重新 admission。若同一 timeout attempt 随后返回非取消类真实错误，只在 timeout fence 仍当前且 catalog 仍为 failed 时精化 err_msg，后续 retry/pause/cancel 不会被旧 goroutine 覆盖。
3. activeRoutine 的替换、读取和关闭已受保护；Resume 同样等待旧 Start 退出，Pause/Cancel 会 fence 已排队的 Resume，避免异步复活。

验证：相关 restart/timeout/resume 生命周期用例在 -race 下连续 20 次通过。

XuPeng-SH:
Latest head re-review (`4f7de39f2d`) note: the earlier blocker comments on this PR look fixed, but I still see two merge-blocking CDC restart issues.

1. **A local CDC restart can succeed and still be reported as failed because the handler context expires first.** `restartTask.Handle` gives the entire restart flow one 5s `handleCtx` (`pkg/taskservice/daemon_task.go:247-249`), but `CDCTaskExecutor.Restart()` now has two independent waits, each bounded by `restartTimeout() = 4s` (`pkg/frontend/cdc_exector.go:252-256`): first waiting for the old start attempt to drain (`pkg/frontend/cdc_exector.go:725-748`), then waiting for the replacement generation to reach ready (`pkg/frontend/cdc_exector.go:823-845`). Start attempts stay installed until the old run actually exits (`pkg/frontend/cdc_exector.go:584-590,661-668,802-808`). So a valid restart can legitimately take >5s and still return `nil`, after which the final `UpdateDaemonTask(...Running...)` is executed on an already-expired context (`pkg/taskservice/daemon_task.go:334-345`).

   Concrete failure: old generation takes ~4s to unwind through `oldAttempt.done`, replacement startup reaches ready ~2s later, `(*ar).Restart()` returns success after ~6s, but the daemon-task status write uses the original 5s context and fails. The row stays `RestartRequested`, and the poller will enqueue it again.

   **Suggested fix:** do not reuse the initial 5s handler context for the post-restart status update; give that write its own fresh bounded context, or raise the outer budget above the full two-phase restart protocol.

2. **Restart can falsely time out when completion and the timeout fire together.** Both restart waits use `select` between completion and `time.After(timeout)` (`pkg/frontend/cdc_exector.go:738-748,823-845`). In Go, if both cases are ready together, either branch may win. So an on-time drain or an on-time replacement startup can still be treated as a timeout.

   Concrete failure: `oldAttempt.done` closes at the same instant as the 4s timer, and the first `select` still takes the timeout arm, marking the restart failed even though the old generation already exited. The same race exists for the replacement `ready` signal at `pkg/frontend/cdc_exector.go:823-845`: the code can cancel the fresh attempt, bump `callbackGeneration`, and persist a failed restart even though the replacement already reached its ready point.

   **Suggested fix:** use an explicit timer and re-check completion before declaring timeout, so completion deterministically wins over a simultaneous deadline.

jiangxinmeng1 (changes requested):
> Thanks for addressing the previous restart-generation and timeout issues. I
  > found one remaining blocking problem.
  >
  > P1 — Do not acknowledge restart readiness before the catalog transition
  > succeeds
  > pkg/frontend/cdc_exector.go:568
  >
  > finishRestartWaiter(restartGeneration, nil) wakes Restart() before
  > updateErrMsgForStartup() performs the required restarting -> running catalog
  > transition. Moreover, clearErrMsgErr is ignored when restartReady is true.
  >
  > If that SQL update fails—for example, because of a transient database error—the
  > daemon task is no longer RestartRequested, subsequent polling will not retry
  > the transition, leaving SHOW CDC inconsistent indefinitely even though the
  > executor is running. The fresh takeover path can hit the same inconsistency
  > because it also ignores this error after claiming the daemon task.
  >
  > The exact restarting -> running catalog transition is part of restart
  > admission, not merely best-effort diagnostic cleanup. Please signal readiness
  > only after it succeeds, or split the required state transition from the
  > optional error-message cleanup. If the state transition fails, the startup
  > result should be propagated so the daemon task is not acknowledged as running.

XuPeng-SH:
Adopted the latest two issue-comment blockers and the requested-changes review in `6800803c8c`, rebased onto current `origin/main@f5c41e81cf`.

Root-cause reflection: the restart protocol had three different success boundaries—in-memory Running, catalog Running, and daemon-task Running—but treated the first one as readiness. It also reused a single five-second handler context around two independently bounded restart phases, and used an unfenced `select` at the deadline. That made individually reasonable local choices inconsistent as one end-to-end protocol.

The completion order is now explicit and one-way:

`old attempt drained -> replacement locally Running -> catalog restarting-to-running CAS committed -> restart waiter success -> daemon RestartRequested-to-Running CAS`

Changes:

1. Both restart waits use explicit timers and re-check completion after the timer fires. If completion and timeout are simultaneously ready, completion deterministically wins.
2. The final daemon status CAS has its own fresh bounded context, so a valid two-phase restart that takes more than the initial five-second query budget can still publish Running.
3. Local restart readiness is emitted only after the exact catalog transition succeeds. A catalog error is propagated through Start/Restart and the daemon row is not advanced.
4. Fresh takeover uses the same required catalog boundary. On startup failure its `Running + current runner` claim is CAS-released to unassigned `RestartRequested`; a newer PAUSE/CANCEL makes that release a zero-row no-op. The persisted task error remains stable while the original cause is retained in bounded fingerprinted events.
5. Full-closure audit found and fixed two adjacent retry gaps: a fresh retry now idempotently reopens `failed -> restarting`, and an early replacement failure before detector registration now moves an already-Starting restart generation to Failed instead of making every later Restart transition invalid.

New deterministic regressions cover simultaneous completion/deadline, catalog failure before readiness, fresh update context, retry-claim release and superseding control requests, fresh failed-catalog re-admission, and early-start failure. Fresh post-rebase evidence: focused frontend/taskservice race stress x20; complete frontend/taskservice tests; complete frontend/taskservice race tests; build; vet; diff-check; and range-diff verification.

XuPeng-SH (commented):
BLOCKING — would request changes, but GitHub does not allow the PR author to submit that review state.

P1: keep the restart timeout path independent from the startup operation it is stopping (pkg/frontend/cdc_exector.go:1447-1452).

Both restart timeout branches synchronously call recordRestartTimeout, which uses context.Background() and the same internalExecutor as Start. internalExecutor.ExecWithStatus acquires ie.Lock() before it creates its session timeout. If the old/startup attempt failed to drain because it is stalled in a catalog Query/Exec while holding that lock, this timeout path queues behind that exact operation and cannot return after the advertised 4 seconds. The only lower bound is the default 24-hour session timeout; meanwhile taskRunner.handleTask is serial, so every daemon control operation on the runner can be frozen. The late-error refinement at lines 1465-1477 has the same ownership dependency.

Please make timeout persistence unable to wait behind the timed-out attempt (for example, separate executor/ownership or asynchronous bounded best-effort persistence) and add a deterministic regression where the old executor holds the catalog path past cancellation while Restart is required to return within its bound.

All other reviewed closures passed. Local race stress, affected-package tests, build, and vet passed on head 74256427dde5e57d97d54c3279dc82e7e14ae880.

gouhongshen (approved):
## Codex automated review

Previous blockers remain fixed: restart readiness now waits for catalog transition, deadline ties prefer completion, and daemon status uses a fresh context. The merge has no conflict-resolution delta. The author’s newly documented timeout-persistence lock stall remains blocking on this head, but is already recorded and is intentionally not duplicated. Focused tests could not run because `go` is unavailable.

jiangxinmeng1 (changes requested):
> Requesting changes due to one remaining P1 issue.
  >
  > The CDC restart timeout path is not actually bounded. Both timeout branches
  > synchronously call recordRestartTimeout, which uses context.Background() and
  > the same internal executor as the startup attempt being stopped.
  > internalExecutor.ExecWithStatus acquires the executor lock before creating its
  > session timeout.
  >
  > If the timed-out startup attempt is blocked in a catalog query or execution
  > while holding that lock, recordRestartTimeout will wait behind it.
  > Consequently, Restart() may block far beyond its advertised four-second timeout
  > —potentially until the session timeout—and the serial task-runner handler can
  > stop processing all daemon control operations.
  >
  > Please make timeout persistence independent of the timed-out operation, for
  > example by using separate executor ownership or bounded asynchronous best-
  > effort persistence. Please also add a deterministic regression test where the
  > old startup attempt holds the catalog path after cancellation and verify that
  > Restart() still returns within its timeout.
  >
  > Apart from this issue, the previously reported restart-readiness, deadline-tie,
  > and daemon-status-context problems appear to be fixed, and I found no
  > additional blockers in the reviewed changes.

pkg/frontend/cdc_exector.go
```diff
@@ -424,18 +770,88 @@ func (exec *CDCTaskExecutor) Resume() error {
...
+		}
+
+		if _, timedOut := waitForCDCCompletion(oldAttempt.done, timeout); timedOut {
+			drainTimeoutErr := moerr.NewInternalErrorNoCtx("CDC restart startup timed out while waiting for previous start to exit")
+			// The replacement has not been launched, so the stale attempt is
+			// the only owner of the shared resources. Fence its later failure
+			// publication and persist this terminal admission result.
+			exec.callbackGeneration.Add(1)
+			_ = exec.stateMachine.SetFailed(drainTimeoutErr.Error())
+			exec.recordRestartTimeout(drainTimeoutErr)
```
LeftHandCold: [P1] 这里的 timeout path 仍会同步阻塞在旧 Start 正在占用的同一个 internal executor 上，因此 4s admission bound 实际上不成立。recordRestartTimeout 使用 context.Background() 调 updateCatalogStateAndErrMsg；生产 internalExecutor.Query/ExecWithStatus 都先 ie.Lock()，再创建 session timeout。若前一个 Start 正是卡在 catalog Query/Exec 并持有这把锁，走到这里后 Restart() 会无限等锁，taskRunner 的控制处理也随之冻结。

我用与生产实现相同的串行 fake 做了确定性复现：旧 Query 持锁，restartStartupTimeout=25ms，Restart() 到 150ms 仍未返回，只有释放旧 Query 后才返回 timeout error。请让 timeout 返回不依赖这个被旧 generation 占用的 executor（仅给这里再套 context 不能解决，因为锁等待发生在 executor timeout 之前），并补一个 serialized-executor 回归测试。line 923 的 startup-timeout 分支有同样问题。

LeftHandCold (changes requested):
Requesting changes for one reproduced P1 in the CDC restart timeout path. The logging contract and the other ObjectIO/publication/proxy/taskservice adoption paths looked sound in this round, and their targeted/race tests pass. However, Restart timeout return can synchronously wait on the same serialized internal executor held by the timed-out Start, so the advertised bound and taskRunner liveness are not preserved.

pkg/frontend/cdc_exector.go
XuPeng-SH: 已在 `1b2b1a7a60` 系统性修复并验证：

- 两个 timeout 分支都改为异步 best-effort 持久化，使用 factory 创建的独立 internal executor，并给写入设置 restart timeout；`Restart()` 的 timeout 返回不再等待被旧 Start 占用的 executor。
- 增加单任务 restart generation 栅栏：并发 Restart 快速失败；前一次 timeout 已返回后的 retry 会有界等待其 `restarting -> failed` CAS 完成，避免旧代写覆盖新代 `running`。
- late startup error 仅输出带 fingerprint 的事件，不再执行无 generation token 的 `failed -> failed` catalog 更新，避免跨代 ABA 覆盖。
- 新增 production-like serialized-executor 确定性测试，同时覆盖 previous-start drain 与 replacement-startup timeout；并覆盖并发 Restart、顺序 retry、late error 跨代隔离。

验证通过：完整 `./pkg/frontend`、完整 `./pkg/frontend -race`、定向 30 轮、定向 race 10 轮、build、vet。

gouhongshen (approved):
## Codex automated review

The only new review surface is the timeout-persistence fix after 7425642. The prior blockers around restart readiness, catalog fencing, ownership/CAS, and logging closure remain resolved. The timeout blocker is fixed: persistence uses an independent executor, bounded context, serialized retry fencing, and generation-safe late-error handling. No new findings.

fengttt (approved):

XuPeng-SH (commented):
Blocking issue: the new Resume/Restart replacement starts are detached from task-runner shutdown.

The initial CDC start is still runner-owned: `CDCTaskExecutorFactory` attaches the executor to the daemon task and calls `exec.Start(ctx)` with the task-runner context (`pkg/frontend/cdc_exector.go:76-114`). But the new replacement generations introduced here use `context.Background()` instead:

- `Resume()` creates `newCDCStartAttempt(context.Background(), generation)` at `pkg/frontend/cdc_exector.go:755`
- `Restart()` creates `newCDCStartAttempt(context.Background(), generation)` at `pkg/frontend/cdc_exector.go:910`

Those attempts then run `startFunc` in standalone goroutines. On shutdown, `taskRunner.Stop()` only stops the stopper and closes channels (`pkg/taskservice/task_runner.go:301-310`), and CN shutdown only calls `task.holder.Close()` + `task.runner.Stop()` (`pkg/cnservice/server_task.go:261-273`). There is no shutdown pass that walks daemon tasks and calls `ActiveRoutine.Cancel()`.

That matters because CDC readers derive their lifetime from the Start root context (`pkg/cdc/table_change_stream.go:340-342`), and cleanup ownership stays with that generation until `cleanup()` finishes (`pkg/cdc/table_change_stream.go:382`, `580-650`). So after a task has been resumed or restarted once, stopping the task runner/CN no longer cancels the new generation. The resumed/restarted CDC goroutines can outlive runner shutdown and keep touching watermark/sinker state after taskservice teardown.

Suggested fix: keep replacement starts under a runner-owned context (or persist an executor-owned root ctx/cancel that task-runner shutdown cancels), instead of spawning them from detached `context.Background()` roots.

I checked the previous pause->resume wait concern again and did not keep it; this shutdown-ownership gap is the blocking one that survived full closure review.

XuPeng-SH:
已合理采纳最新 lifecycle blocker，修复已推送到 2a360e5503。

根因确认：initial Start 由 task-runner context 持有，但 Resume/Restart replacement 使用 context.Background()，导致 taskRunner/CN shutdown 不能终止 replacement generation；而 ActiveRoutine 又没有统一的 shutdown traversal，因此确实可能残留 CDC reader/sinker goroutine。

修复闭环：
- factory 在发布 ActiveRoutine 前绑定 executor lifecycle；runner context 结束会取消所有后续 replacement Start。
- lifecycle 只继承 cancellation，不继承 fresh-takeover 的 RestartAdmission 等 attempt-specific value，保持原 catalog CAS 行为。
- Resume 等待旧 Start cleanup 时也监听 lifecycle cancellation，shutdown 不再留下 wait goroutine；Restart replacement 同样使用该 lifecycle root。
- terminal Cancel、attach 失败和 initial Start 失败都会停止/回收 lifecycle callback，避免失败重试产生无界 callback 累积。
- 新增 Resume running、Resume waiting-old-generation、Restart startup、RestartAdmission isolation 四类确定性回归。

验证通过：focused lifecycle -race 重复 20 次；完整 pkg/frontend 普通测试与 -race；go build；go vet；git diff --check。

jiangxinmeng1 (changes requested):
> [P1] Fence the restart generation across catalog publication
  >
  > The generation check occurs before completeStartupCatalogTransition, but that
  > call can block while executing the restarting -> running catalog CAS.
  >
  > A concrete race is:
  >
  > 1. The replacement passes the generation check and begins the catalog update.
  > 2. The four-second timer fires while that update is still in flight.
  > 3. Restart() cancels and fences the attempt, changes the in-memory state to
  >    Failed, and starts the asynchronous restarting -> failed update.
  >
  > 4. The original update then commits running and returns successfully.
  >
  > There is no generation check after catalog publication, so the stale Start
  > still records running metrics and returns without the error cleanup that
  > unregisters the table detector. Meanwhile, the timeout update conflicts with
  > the now-running catalog row. The caller therefore receives a timeout while the
  > catalog says running, the replacement workers have been canceled, and the stale
  > detector registration can prevent subsequent retries.
  >
  > The current timeout tests block before catalog publication; they do not cover
  > an in-flight restarting -> running update.
  >
  > Please establish a generation-aware linearization point between timeout fencing
  > and catalog publication, and add a regression test whose ExecWithStatus blocks
  > during restarting -> running, then completes after Restart() has timed out.

XuPeng-SH (commented):
Blocking P1: the latest lifecycle fix propagates cancellation to replacement generations, but taskRunner.Stop still does not join them.

After Pause, the original TaskExecutor invocation registered with the stopper has returned. Resume (cdc_exector.go:802) and Restart (cdc_exector.go:983) launch replacement startFunc calls as standalone goroutines. bindLifecycleContext wires the stopper context through context.AfterFunc (line 234), but context cancellation/AfterFunc does not wait for those goroutines or for Start cleanup. taskRunner.Stop only calls stopper.Stop and then closes its channels (task_runner.go:301-310), so once the original executor has returned it has no WaitGroup ownership of a resumed/restarted generation.

Production consequence: CN/task-runner shutdown can return while a replacement CDC Start is still unwinding readers, sinkers, watermarks, detector registration, or catalog work. The new tests prove that startFunc observes ctx.Done(), but they do not hold cleanup after cancellation and prove that taskRunner.Stop waits for completion.

Please make the replacement generation runner-owned through completion (not only cancellation-bound), and add a deterministic shutdown test: pause the initial tracked Start, resume/restart, block replacement cleanup after ctx.Done(), call taskRunner.Stop, verify Stop remains blocked, then release cleanup and verify Stop returns.

The focused lifecycle/logging race tests, build/vet, taskservice tests, and current CI are green; they do not exercise this join invariant. GitHub does not permit the PR author to request changes on their own PR.

LeftHandCold (approved):
Reviewed exact head 2a360e55. No blocking correctness or normal-business regression issues found. Rechecked the CDC restart/resume lifecycle end to end: replacement generations are now bound to the task-runner stopper lifetime without inheriting attempt-specific admission values; runner shutdown cancels both waiters and replacement starts; catalog and daemon-task transitions remain generation/CAS fenced. Also reviewed bounded event state, legacy logging compatibility, and the proxy/publication/objectio/frontend adoption paths. Local package tests passed for logutil, common/log, objectio, proxy, publication, taskservice, and frontend; targeted race tests passed for event limiting and CDC/taskservice lifecycle paths; git diff --check passed.

LeftHandCold (changes requested):
Revising my just-submitted approval after reconciling the two newest exact-head review threads. Both remaining issues are production-reachable blockers.

P1 — Fence restart timeout against an in-flight restarting-to-running catalog publication (pkg/frontend/cdc_exector.go around completeStartupCatalogTransition). The generation check happens before the catalog CAS. If ExecWithStatus is still in flight when the four-second Restart timeout fires, the timeout path cancels/fences the attempt, sets the state machine to Failed, and starts restarting-to-failed persistence; the older CAS can then commit running and return nil. There is no post-publication generation check, so Start records running metrics and exits successfully without error cleanup, while the timeout persistence loses its CAS. The caller sees a timeout but the catalog says running, the replacement routine has been canceled, and detector registration may remain. Add a generation-aware linearization point and a deterministic test that blocks the running CAS past the timeout.

P1 — Runner shutdown must join replacement generations, not only cancel them (pkg/frontend/cdc_exector.go Resume/Restart lifecycle binding and pkg/taskservice/task_runner.go Stop). After Pause, the original stopper-tracked TaskExecutor invocation has returned. Resume/Restart start standalone goroutines; context.AfterFunc propagates cancellation but does not register or wait for their cleanup. taskRunner.Stop can therefore return while replacement Start is still unwinding readers, sinkers, watermark/catalog work, and detector state, after which CN shutdown proceeds to close RPC/I/O and other dependencies. Make replacement generations runner-owned through completion and add a shutdown test that blocks replacement cleanup after ctx.Done and verifies Stop waits until cleanup is released.

The logging limiter and other adoption paths passed review and targeted/race tests; these two CDC lifecycle linearization gaps are the remaining blockers.

jiangxinmeng1 (changes requested):
> [P1] Join timeout-persistence work during task-runner shutdown
  >
  > recordRestartTimeoutAsync still launches its catalog repair with a raw
  > goroutine and context.Background(). The new task-runner scheduler owns the
  > Resume/Restart replacement generation, but it does not own this persistence
  > goroutine.
  >
  > A deterministic failure sequence is:
  >
  > 1. Block the timeout catalog executor.
  > 2. Let Restart() time out and enter recordRestartTimeoutAsync.
  > 3. Call taskRunner.Stop().
  > 4. The stopper joins the replacement Start, but returns while the catalog-
  >    persistence goroutine is still blocked.
  >
  > 5. CN shutdown can then continue closing services while that goroutine
  >    subsequently accesses the internal executor and catalog state.
  >
  > The persistence context should remain independent of the request timeout, but
  > the task runner must still cancel or join it during shutdown. Please schedule
  > it through the captured lifecycle scheduler, or otherwise give the executor an
  > explicit shutdown/join owner. Add a test that blocks persistence, calls
  > taskRunner.Stop, and verifies that shutdown does not return until the work is
  > canceled or released.
  >
  > Relevant code: cdc_exector.go lines 1633–1685
  >
  (https://github.com/matrixorigin/matrixone/blob/05a79c6e8236fa4a096c99b5a0b962faef55d761/pkg/frontend/cdc_exector.go#L1633-L1685).

  > [P2] Cancel completed start-attempt contexts
  >
  > Each newCDCStartAttempt creates a context.WithCancel child of the long-lived
  > executor lifecycle context. finishStartAttempt removes the active pointer and
  > closes

LeftHandCold (approved):
Reviewed exact head 533d150d after the rebase. The two remaining CDC blockers are closed: replacement generations are admitted into the task-runner stopper and therefore canceled and joined by Stop; restart completion and timeout now race on a single ownership token, with exact-state catalog reconciliation for a late running publication. I traced scheduler rejection, catalog CAS failure, timeout persistence, retry, pause/cancel supersession, detector cleanup, and generation reuse paths; no blocking correctness or normal-business regression issue remains. The nine previously reviewed commits are semantically unchanged by range-diff. The new ISCP synctest change is test-only and its race stress passed. Fresh full race tests passed for frontend, taskservice, and iscp; focused new regressions passed under race stress; affected-package tests, go build, go vet, and git diff --check passed.

jiangxinmeng1 (changes requested):
> P1 — Do not overwrite a superseding control request when a restart claim fails
  >
  > When startDaemonTask returns an error, startTask.Handle calls
  > setDaemonTaskError even for a restart claim. That helper writes the original
  > RestartRequested task snapshot without status or runner conditions.
  >
  > A concurrent PAUSE/CANCEL can therefore commit after the failed claim, then be
  > overwritten when the error-recording update succeeds. This bypasses the CAS
  > protection used by the other restart failure paths.
  >
  > Please either avoid setDaemonTaskError for a pre-claim restart failure or
  > condition the error update on the observed RestartRequested status and runner.
  > A regression test should make the claim update fail, commit CancelRequested,
  > and verify the subsequent error path preserves it.

jiangxinmeng1 (approved):
> Reviewed the latest head 47892456.
  >
  > The previously reported restart-claim race is fixed correctly. A failed restart
  > claim no longer writes the stale RestartRequested snapshot back, so a newer
  > PAUSE/CANCEL request is preserved. The new deterministic regression test covers
  > that exact sequence.
  >
  > I also rechecked the CDC restart lifecycle, generation fencing, timeout cleanup
  > ownership, catalog/daemon CAS boundaries, and the logging adoption paths. I
  > found no remaining merge-blocking correctness or lifecycle issues.
  >
  > The remaining logging API details are non-blocking and can be handled as
  > follow-up improvements.
  >
  > Approved.

PR #26193, opened 2026-07-27
fix(frontend): reject reversed pick snapshot ranges

### LINKED ISSUE #26078: [Bug]: DATA BRANCH PICK silently accepts a reversed BETWEEN SNAPSHOT range
### Describe the bug

`DATA BRANCH PICK ... BETWEEN SNAPSHOT` accepts a reversed snapshot range and silently performs no work. A statement with the later snapshot as the start and the earlier snapshot as the end succeeds without an error, making an invalid range indistinguishable from a valid range with no changes.

### Steps to reproduce

```sql
drop database if exists g4d_pick_time;
drop snapshot if exists g4d_pick_early;
drop snapshot if exists g4d_pick_late;
create database g4d_pick_time;
use g4d_pick_time;

create table dst(id int primary key, v int);
insert into dst values(1,10);
data branch create table src from dst;

create snapshot g4d_pick_early for account sys;
insert into src values(2,20);
create snapshot g4d_pick_late for account sys;

data branch pick src into dst
  between snapshot g4d_pick_late and g4d_pick_early;

select * from dst order by id;
```

### Actual behavior

The `PICK` statement succeeds without an error and `dst` remains unchanged:

```text
1  10
```

The result reproduced in 3/3 runs. The corresponding forward range (`g4d_pick_early` to `g4d_pick_late`) correctly adds row `(2,20)`.

### Expected behavior

The statement should reject a range whose start snapshot is later than its end snapshot and return a clear invalid-range error.

### Code context

`resolveBetweenSnapshots` resolves both timestamps but does not compare them. `constructChangeHandle` intersects the target range with `[betweenFrom.Next(), betweenTo]`; a reversed range therefore becomes an empty range and is silently treated as a successful no-op.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26078

`DATA BRANCH PICK ... BETWEEN SNAPSHOT` resolved both snapshot timestamps but did not validate their order. A reversed range reached the collect-range intersection as an empty interval and incorrectly succeeded as a no-op.

This change rejects only ranges whose start timestamp is later than the end timestamp. Forward and equal timestamp ranges remain valid. The error identifies both snapshot names.

Coverage includes physical-time order, logical-time order, equal timestamps, the full snapshot resolution path, and an end-to-end BVT case using two real snapshots.

Validation:

- `go test ./pkg/frontend -run 'Test(ValidateBetweenSnapshotRange|ResolveBetweenSnapshotsReversedRange)$' -count=1 -coverprofile=/tmp/pr26193-frontend.cover`
- `go vet ./pkg/frontend`
- `make build`
- `pick_10.sql` BVT: 111/111 statements passed
- `git diff --check`


XuPeng-SH (approved):
Reviewed the exact head `7b9c3421e7e4`. The range check is placed after both snapshots resolve and before producer goroutines, change handles, or destination writes. `types.TS.GT` compares physical time and then logical time, so forward and equal ranges remain accepted while both physical- and logical-time reversals are rejected. The error propagates through the existing transaction defer, preserving rollback/cleanup behavior.

The BVT is repeatable after interruption: database state is reset at file start, account-level snapshots are removed both before and after the case, and the assertion verifies the concrete invalid-input error.

Validation on this head:
- both new tests independently passed `-race -count=100` (30s adaptive budget, measured test elapsed below timer resolution)
- full `pkg/frontend` suite passed with `-race -count=1`
- local standalone: reversed range rejected 3/3 with error 20301 and destination unchanged; equal range remained a valid no-op; forward range applied the row
- full build, `go vet ./pkg/frontend`, and `git diff --check` passed
- current `main` merge-tree is conflict-free

No blocking findings.

heni02 (approved):

PR #26196, opened 2026-07-27
test(frontend): cover branch snapshot identifier bytes

### LINKED ISSUE #26073: [Bug]: Data Branch snapshot metadata changes backslashes in quoted object names
## Description

`DATA BRANCH CREATE TABLE` stores a branch-protection snapshot whose `database_name` and `table_name` no longer match the source object when a legal quoted identifier contains a backslash.

The source catalog preserves the backslash byte (`5C`), while the snapshot insert treats it as part of a SQL string escape. For example, `g4d\db` becomes `g4ddb`, and `src\t` becomes `src<TAB>` (`09`) in `mo_snapshots`.

## Version

- Branch: `main`

## Reproduction

```sql
drop database if exists `g4d\db`;
create database `g4d\db`;
use `g4d\db`;

create table `src\t`(id int primary key);
insert into `src\t` values (1);
data branch create table `dst\t` from `src\t`;

select reldatabase, hex(reldatabase), relname, hex(relname), rel_id
from mo_catalog.mo_tables
where reldatabase like 'g4d%';

select sname, database_name, hex(database_name),
       table_name, hex(table_name), obj_id, kind
from mo_catalog.mo_snapshots
where kind = 'branch' and database_name like 'g4d%';
```

This was reproduced 3 out of 3 runs.

## Actual result

The object catalog contains:

```text
database: g4d\db   hex=6734645C6462
table:    src\t    hex=7372635C74
```

The matching branch snapshot contains:

```text
database_name: g4ddb    hex=6734646462
table_name:    src<TAB> hex=73726309
```

## Expected result

The branch snapshot should preserve the exact source database and table names stored in the catalog.

## Code path

`createBranchProtectSnapshot` interpolates `receipt.srcDb` and `receipt.srcTbl` directly into single-quoted SQL literals. Backslashes are therefore interpreted again when the generated `INSERT` is parsed.

The values should be encoded with the same SQL-literal helper used by other Data Branch generated statements.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26073

#26218 already fixed branch-snapshot SQL literal escaping on `main` and added
parser-based unit coverage for quotes, backslashes, and injection-like input.
After rebasing onto that implementation, this PR contains only the
end-to-end BVT coverage that is not present in #26218.

The BVT creates database and table identifiers containing backslashes, creates
a Data Branch table, and verifies:

- the exact database-name bytes stored in `mo_catalog.mo_snapshots`;
- the exact table-name bytes stored in `mo_catalog.mo_snapshots`;
- the branch snapshot `obj_id` references the source table.

No production code or unit tests are changed.

### Validation

- `make build`
- `protect_11.sql` against rebased commit `e88eb003b`: 13/13 statements passed
- `git diff --check`


XuPeng-SH (changes requested):
The production-code and unit-test portions of this PR have been superseded by #26218 (commit f2807845), which is already on `main`. That implementation covers the same Data Branch snapshot string fields and includes a stronger parser-based round-trip test for quotes, backslashes, and injection-like input. This PR is also currently conflicting with `main`, so the previously green checks do not validate the final merge result.

Please rebase onto the latest `main`, resolve the conflict by dropping the duplicate `data_branch_snapshot.go` changes and the exact-generated-string unit test, and preserve the stronger test already on `main`. The new `protect_11.sql` / `.result` BVT remains valuable because it verifies the catalog bytes and parent object ID end to end; keeping this PR as a BVT-only regression change would be appropriate. Please also link/close #26073 accordingly.

gouhongshen:
Addressed the requested change. I rebased onto current `main` (including #26218), dropped the superseded production-code helper and exact-string unit test, and retained only `protect_11.sql/.result`. The BVT passes 13/13 against rebased commit `e88eb003b`; the final diff contains only those two BVT files. `Fixes #26073` remains in the updated PR description.

XuPeng-SH (approved):
Re-reviewed the exact updated head `e88eb003bbf9`. The previous requested changes are fully addressed: the superseded production/helper and unit-test changes were dropped, leaving only the end-to-end BVT, and the PR now links `Fixes #26073`.

The case locates the branch snapshot by the destination table ID, asserts the exact database/table bytes via `HEX`, and verifies `obj_id` still points to the source table. Its setup/cleanup is repeatable and the initial `DROP DATABASE IF EXISTS` recovers a prior interrupted run.

Validation on this head:
- full build and `go vet ./pkg/frontend`
- `protect_11.sql` on a local standalone instance three consecutive times, always yielding database `70726F746563745C6462`, table `7372635C74`, and parent match `1`
- related production regression test with `-race -count=100`
- full `pkg/frontend` suite with `-race -count=1`
- `git diff --check`

No blocking findings.

heni02 (approved):

PR #26197, opened 2026-07-27
fix: reconcile auto-increment state for copy and clone

VioletQwQ-0:
issue #23143

## What type of PR is this?

- [x] bug
- [ ] feature
- [ ] improvement
- [ ] documentation

This is the next serial prerequisite in the #25883 split. It makes COPY/CLONE allocator reconciliation independently correct before the user-facing `ALTER TABLE ... AUTO_INCREMENT=N` entry is enabled.

Invariant:

- a visible destination allocator is reset to `max(requested offset, copied-row maximum, source allocator offset)`;
- fresh CLONE maps allocator state by cloned column name because planner IDs are in a new coordinate space;
- ALTER COPY maps retained columns by stable `ColId`, preserving rename/reorder and preventing a dropped-and-readded same-name column from inheriting old state;
- hidden fake-PK allocators remain independent from visible AUTO_INCREMENT requests;
- failed/cancelled ALTER COPY statements discard transaction-private resets, and all internal SQL results are closed.

## Changes

- collect source allocator offsets and visible row maxima at the selected snapshot/account;
- translate source catalog indexes to destination identities before execution;
- replace fake value insertion with transaction-aware `SetOffset`;
- reconcile temporary ALTER COPY tables after copied rows become visible in the DDL transaction;
- validate destination signed/unsigned integer bounds before resetting;
- add regression coverage for fresh ID collisions, rename/reorder, drop/add name reuse, hidden allocators, three-way bounds, overflow, cancellation, rollback cleanup, quoted identifiers, and error-result cleanup;
- add an issue-linked distributed case that snapshots the real source allocator offset and proves COPY ALTER preserves its reserved range.

## Non-goals

- no `ALTER TABLE ... AUTO_INCREMENT=N` SQL entry;
- no mixed-version enablement gate;
- no change to the epoch attached to allocated writes;
- no generic transaction/recovery changes.

## Testing

- `go test ./pkg/incrservice ./pkg/sql/colexec/table_clone ./pkg/sql/compile -count=1`
- focused `-race` runs for the new incrservice, table_clone, and compile tests
- `go test ./pkg/... -run '^$'`
- dedicated `issue_23143_copy_auto_increment` distributed regression added; fresh CI execution is pending for the current head
- `git diff --check`
- `mo-self-review`: PASS
- exact-head `mo-pr-preflight-review`: `PASS review=PASS validation=PENDING` for `6b1bb91fa0c292ab056adab3a8ba34935125e9fc`

## Diff size

- production code: 5 files, `+416/-99`
- tests: 7 files, `+689/-4`
- generated code: none

## Dependency

Depends on the merged epoch fence (#26027) and transactional allocator reset (#26085). The next split item will start from `main` only after this PR merges.


XuPeng-SH:
Deep review on PR #26197 (`base=main`, head `a064f6efc6`) found one merge-blocking clone/auto-increment gap.

**Clone still leaves fulltext hidden-table AUTO_INCREMENT stale.** In `TableClone.Call`, index hidden tables are cloned first (`pkg/sql/colexec/table_clone/table_clone.go:449-460`), then `updateDstAutoIncrColumns()` reconciles AUTO_INCREMENT state only through `tc.dstMasterRel` (`pkg/sql/colexec/table_clone/table_clone.go:469-515`). On the metadata side, `constructTableClone` only captures offsets/max values from `clonePlan.SrcTableDef` / `clonePlan.SrcTableDef.TblId` (`pkg/sql/compile/operator.go:2360-2467`), so there is no per-hidden-table capture or publish path.

That breaks fulltext hidden tables specifically: their schema defines the hidden `__mo_fake_pk_col` as `AutoIncr` (`pkg/fulltext/plugin/plan/schema.go:200-216`), and later fulltext maintenance inserts `NULL` for that column with `HasAutoCol: true` (`pkg/sql/plan/build_dml_util.go:4602-4660`), which means post-clone writes depend on that hidden-table allocator being correct.

Concrete failure: clone or restore a table that already has fulltext rows, then `INSERT` or `UPDATE` the cloned table. The fulltext write path allocates fake PKs from the reset destination hidden-table allocator instead of from the cloned max, so it can collide with existing hidden-table primary keys and fail index maintenance deterministically.

**Suggested fix:** reconcile AUTO_INCREMENT state for every cloned hidden table that owns an allocator, not just the master relation.

VioletQwQ-0:
Confirmed the `vector_index_08` result change. Before the COPY ALTER, the source allocator has already reserved through offset `20000`, so preserving that allocator state makes the next generated IDs `20001` and `20002`. The old `9777`/`9778` expectations rebuilt from the copied row maximum and could reuse values that were still reserved in an old CN cache.

I also added a dedicated #23143 prerequisite BVT. It snapshots the actual `mo_increment_columns.offset` before COPY ALTER and checks that the next generated ID is `saved offset + 1`, so the test does not depend on the current allocation batch size. This PR still does not enable the final `ALTER TABLE ... AUTO_INCREMENT=N` syntax; that user-facing case stays in the final integration PR.

Local validation on `c2b237f6a`:

- `go test -count=1 ./pkg/incrservice ./pkg/sql/colexec/table_clone ./pkg/sql/compile`
- `git diff --check`
- `mo-self-review`: PASS
- exact-head preflight: `PASS review=PASS validation=PENDING`

Fresh CI will execute the new distributed case.

gouhongshen (approved):
## Codex automated review

No new independent findings. The existing unaddressed fulltext hidden-table allocator issue remains and is intentionally not duplicated. Focused Go tests could not run because the environment lacks `go`.

VioletQwQ-0:
Fixed in eb32404f7.

The clone metadata now captures AUTO_INCREMENT offset and copied max for each hidden index relation, and TableClone reconciles every destination hidden relation using the same runtime key as the clone map. This covers both ordinary keys such as `ftidx.` and partition keys such as `p0.ftidx.`, so each hidden allocator advances independently to `max(copied max, source offset)`.

I also extended the existing fulltext clone BVT with a post-clone INSERT. That INSERT drives the real fulltext maintenance path and would fail if the copied hidden fake-PK allocator were still reset.

Validation:
- `go test ./pkg/sql/colexec/table_clone ./pkg/sql/compile -count=1`
- `go vet ./pkg/sql/colexec/table_clone ./pkg/sql/compile`
- `git diff --check`
- exact-head preflight: `PASS review=PASS` (CI/BVT validation is running on the pushed head)

VioletQwQ-0:
Fixed the current-head CI failures in f5d4b8077:

- preallocated the clone-index mapping result to satisfy the SCA prealloc check;
- replaced the Pessimistic BVT dependency on immediate mo_increment_columns visibility with a public-SQL regression: allocate ID 1001, delete it, run COPY ALTER, then verify the next generated ID does not reuse 1001;
- Coverage was only skipped because required producer jobs failed, so it needs no independent code change.

Local validation: affected package tests, focused race tests, go vet, go build, git diff --check, full mo-self-review, and exact-head preflight all pass. Fresh CI is now running.

VioletQwQ-0:
Follow-up CI fix in 177edcff8:

The Pessimistic BVT confirmed that both issue #23143 invariants hold, but MatrixOne renders these comparison expressions as numeric int8 values (`1`), while the expected result used textual `true`. I changed only the four expected cells from `true` to `1`, matching the exact actual output from run 30254408830 / job 89939569325.

The Coverage failure was cascading from the failed BVT producer and needs no independent change. Affected package tests, `git diff --check`, full self-review, and exact-head preflight pass. Fresh CI is running.

VioletQwQ-0:
Resolved the current merge conflict in 202a5473e by normally merging upstream/main at a55820b1f.

The only content conflict was in pkg/sql/compile/alter_test.go, where this PR added the AUTO_INCREMENT reset-cleanup test and upstream added the IVF INCLUDE-column affected-index test at the same location. The resolution keeps both complete tests unchanged.

Validation passed: affected package tests, focused race tests for both conflict-adjacent paths, git diff checks, index-plugin domain guards, full self-review, and exact-head preflight. The PR remains Ready; fresh CI is running.

heni02 (approved):

VioletQwQ-0:
The Ubuntu UT failure on `202a5473e` was a 60-minute timeout during the heavy-race phase, not a test failure: the raw job log contains no failed package, panic, assertion, or generated `ut-report`. The previous PR head completed the same UT in about 25 minutes, and recent upstream PRs on the same runner label completed it in about 29–32 minutes. I classified this as runner/infrastructure performance rather than a code defect.

The branch was four commits behind `main`, so I rebased it onto `c6d7a1168`. The only conflict was two adjacent independent tests in `pkg/sql/compile/alter_test.go`; both complete test functions and their assertions were retained.

Validation on new head `6b1bb91fa`:
- `go test ./pkg/incrservice ./pkg/sql/colexec/table_clone ./pkg/sql/compile -count=1`
- focused `-race` tests for the ALTER COPY cleanup, upstream IVF INCLUDE behavior, stable-identity reconciliation, and clone allocator bounds
- complete `mo-self-review`: PASS
- exact-head preflight: `PASS review=PASS validation=PENDING`

Fresh CI is now running on the rebased head.

XuPeng-SH (approved):
Deep-reviewed exact head `6b1bb91fa0`.

Verified visible allocator reconciliation (`max(requested, copied max, source offset)`), stable ColId/name mapping, hidden fake-PK isolation, fulltext/partition hidden-index allocator mapping, rollback cleanup, result lifecycle, integer bounds, snapshot/account handling, and DDL-only performance impact. The previously reported fulltext hidden-table gap is fixed on this head.

Local affected-package tests pass, and exact-head CI/BVT/SCA/coverage are green. No blocking production issue found.

PR #26199, opened 2026-07-27
fix: honor group_concat_max_len

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25370

fix: honor group_concat_max_len


heni02 (approved):

XuPeng-SH (changes requested):
Blocking correctness issue: `group_concat_max_len` is resolved only while the operator tree is constructed, so a cache-eligible TP prepared statement captures the value at PREPARE time. `createPrepareStmt` stores that compiled pipeline, later EXECUTEs only call `Compile.Reset`, and `Group.Reset` recreates its aggregate from the unchanged `Aggs.ExtraConfig`.

Repro:
```sql
set session group_concat_max_len = 1024;
prepare s from 'select group_concat(s order by s separator "") from t';
set session group_concat_max_len = 5;
execute s;
```
For a small `t` containing `aa`, `bb`, `cc`, this continues using 1024 and returns `aabbcc` instead of `aabbc`; changing the limit in the opposite direction leaves the result incorrectly truncated. Fresh statements and the new BVT do not exercise prepared compile reuse.

Please refresh this dynamic session value for every execution (or invalidate/rebuild the cached compile when it changes) and add a prepared-statement regression test.

XuPeng-SH (approved):
Re-reviewed head 40b143e5. The previous blocker is fixed: cached prepared executions now refresh the group_concat configuration in Compile.Reset across Group/MergeGroup/Window while preserving the separator, and the regression BVT covers changing group_concat_max_len between EXECUTEs. I also rechecked scope traversal, remote serialization, and Reset/Free/error-path lifecycle. No blocking issues found. Targeted CGo tests pass for aggexec, compile, and frontend.

PR #26200, opened 2026-07-27
fix(frontend): quote LCA primary key references

### LINKED ISSUE #26071: [Bug]: DATA BRANCH DIFF fails when LCA primary-key columns require quoting
### Describe the bug

`DATA BRANCH DIFF` fails for sibling branches when a primary-key column name requires identifier quoting.

The failure occurs while reconstructing rows deleted on one branch from the common ancestor. The generated join query quotes projected columns and table names, but inserts primary-key names into the `VALUES` alias list and join predicates without quoting.

### Steps to reproduce

Single-column primary key using a reserved word:

```sql
drop database if exists g4d_lca_names;
create database g4d_lca_names;
use g4d_lca_names;

create table base_single(`select` int primary key, payload varchar(20));
insert into base_single values (1,'one'),(2,'two');
data branch create table left_single from base_single;
data branch create table right_single from base_single;

delete from left_single where `select`=1;
update right_single set payload='two-right' where `select`=2;

data branch diff left_single against right_single;
```

Composite primary key using a reserved word and punctuation:

```sql
create table base_composite(
  `left` int,
  `a-b` varchar(10),
  payload varchar(20),
  primary key(`left`,`a-b`)
);
insert into base_composite values (1,'x','one'),(2,'y','two');
data branch create table left_composite from base_composite;
data branch create table right_composite from base_composite;

delete from left_composite where `left`=1 and `a-b`='x';
update right_composite set payload='two-right' where `left`=2 and `a-b`='y';

data branch diff left_composite against right_composite;
```

The same generated query is shared by MERGE and PICK when applying a deletion from a sibling branch:

```sql
create table apply_base(`select` int primary key, payload varchar(20));
insert into apply_base values (1,'one'),(2,'two');
data branch create table apply_src from apply_base;
data branch create table apply_dst from apply_base;
delete from apply_src where `select`=1;

data branch merge apply_src into apply_dst;
-- or, on a freshly recreated pair:
data branch pick apply_src into apply_dst keys(1);
```

### Actual behavior

The single-key case fails near this generated fragment:

```text
... as pks(__idx_,select) on lca.select = cast(pks.select as INT) ...
```

The composite-key case fails near:

```text
... as pks(__idx_,left,a-b) on lca.left = ... AND lca.a-b = ...
```

DIFF, MERGE, and PICK return a SQL syntax error instead of completing. MERGE and PICK leave the destination table unchanged after the failure.

### Expected behavior

Every generated reference to a user-defined primary-key column should use identifier quoting. Both DIFF commands should return their rows normally, regardless of valid primary-key names.

### Reproducibility

Reproduced in 3/3 runs for both the single-column and composite-primary-key DIFF cases, and in 3/3 runs for the MERGE and PICK variants.

### Code context

`handleDelsOnLCA` in `pkg/frontend/data_branch_hashdiff.go` builds the LCA lookup query. It uses raw `pkNames` in:

- the `VALUES` alias list;
- `lca.<pk>` expressions;
- `pks.<pk>` expressions.

The projected LCA columns in the same query already use `quoteIdentifierForSQL`, but these primary-key references do not.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26071

`handleDelsOnLCA` quoted the LCA database, table, and projected columns, but inserted raw primary-key names into the generated VALUES alias and join predicates. Reserved words or punctuation in a primary-key column therefore produced invalid SQL during DATA BRANCH DIFF, MERGE, and PICK.

This change quotes the LCA primary-key identifiers once and reuses the same ordered list for the VALUES alias, LCA references, and probe references. It preserves the existing key order and cast behavior.

Coverage includes:
- a unit test that captures the generated join SQL for a reserved-word primary key;
- BVT cases for single and composite quoted primary keys across DIFF, MERGE, and PICK.

Validation:
- `make build`
- `go vet ./pkg/frontend`
- `go test ./pkg/frontend -run "^TestHandleDelsOnLCA_SQLPaths$" -count=1`
- `mo-tester ... -i branch_lca_quoted_pk -r 100` (32/32 passed)


XuPeng-SH (approved):
Reviewed the LCA delete-probe SQL path.

The fix now quotes primary-key identifiers consistently in the VALUES alias and join predicates, while preserving key order and the existing cast behavior. The added unit/BVT coverage matches the reachable failure modes for reserved-word and punctuation PK names, and I did not find a high-confidence blocking correctness issue in this diff.

PR #26203, opened 2026-07-27
fix(embed): preserve base cluster startup errors

### LINKED ISSUE #25929: [Bug]: TestDataBranchDiffAsFile panics in RunBaseClusterTests cleanup

### Branch Name

main (observed in PR #25872 CI)

### Commit ID

08858301789ff9928f7c5dd1372c61c4fd93e211

- Hardware parameters: GitHub Actions runner `amd64-mo-guangzhou-2xlarge16`
- OS type: Ubuntu/Linux x86_64
- Go version: 1.26.4
- CI run: https://github.com/matrixorigin/matrixone/actions/runs/29798346785/job/88534254610

### Actual Behavior

`pkg/tests/dml.TestDataBranchDiffAsFile` panicked immediately with a nil pointer dereference in the embedded cluster
test cleanup path:

```text
github.com/matrixorigin/matrixone/pkg/embed.RunBaseClusterTests.func2()
    pkg/embed/testing.go:85
github.com/matrixorigin/matrixone/pkg/embed.RunBaseClusterTests()
    pkg/embed/testing.go:113
github.com/matrixorigin/matrixone/pkg/tests/dml.TestDataBranchDiffAsFile()
    pkg/tests/dml/dml_test.go:116
```

### Expected Behavior

`TestDataBranchDiffAsFile` should either complete successfully or report the cluster startup/test error without
panicking during cleanup.

### Steps to Reproduce

```text
go test ./pkg/tests/dml -run TestDataBranchDiffAsFile -count=1
```

The failure may depend on an earlier embedded-cluster initialization failure in the CI environment.

### Additional information

Observed in MatrixOne ALL CI run `29798346785`. The panic appears to dereference partially initialized cleanup state.
The failure is unrelated to the prepared binary parameter changes in PR #25872; none of that PR's files touch
`pkg/embed` or `pkg/tests/dml`.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #25929

`RunBaseClusterTests` stored the base cluster globally but kept its initialization error in a call-local variable. If the first `NewCluster` or `Start` attempt failed, `sync.Once` was permanently consumed. A later call then observed a fresh nil error and dereferenced the still-nil global cluster, replacing the original startup failure with an unrelated nil-pointer panic.

This change:
- stores the one-time cluster result and initialization error together;
- reports the same original startup error on every later call without retrying initialization or invoking the test callback;
- requires a test reporter (`t` or `b`) so startup errors are fatal and cannot be discarded by callers;
- migrates every existing `RunBaseClusterTests` test and benchmark caller to the non-ignorable API;
- separates cluster startup and per-test session preparation so the lifecycle is explicit.

The focused regression test invokes a failing initializer twice and verifies that initialization runs once, both calls report the original error, and the test callback is never invoked.

Validation:
- `make build`
- `go vet ./pkg/embed ./pkg/tests/ddl ./pkg/tests/issues ./pkg/tests/dml ./pkg/tests/txnexecutor`
- compile all five affected packages with `go test ... -run "^$" -count=1`
- `go test -v ./pkg/embed -run "^TestOnceClusterReportsInitializationError$" -count=1`
- `go test -v ./pkg/embed -run "^TestBaseClusterOnlyStartOnce$" -count=1`
- `go test -v ./pkg/tests/dml -run "^TestDataBranchDiffAsFile$" -count=1`
- post-merge caller validation: `go test ./pkg/tests/issues -run '^$' -count=1` and `go vet ./pkg/tests/issues`
- after the latest `main` merge, migrated the newly added `issue_25986_test.go` caller to the required test-reporter API


XuPeng-SH (changes requested):
Blocking finding:

[P1] Make all RunBaseClusterTests callers observe the cached startup error. onceCluster.run now returns the stored initialization error before invoking the callback, but this PR updates only TestDataBranchDiffAsFile. Of 45 RunBaseClusterTests occurrences in the tree (including the helper definition), only 3 calls are wrapped with require.NoError; roughly 41 test/benchmark call sites still discard the return value. If the shared cluster fails to initialize, those callers will skip their entire test callback and still report success. Before this change a later call at least panicked, so the suite could not go green without running its assertions. Please update every caller to fail on the returned error, or change the helper API to report the error through testing.TB/another non-ignorable mechanism.

The focused regression test, build, vet, DML compile check, diff check, and current CI all pass on 6c5531e9; this issue concerns the failure path that normal green CI does not exercise.

gouhongshen:
Addressed in `78524340d`. The finding is valid: caching the startup error while leaving callers free to discard it could turn an initialization failure into a false PASS.

`RunBaseClusterTests` now requires a test reporter and calls `Fatalf` with the cached original error before any test callback can run. All 44 existing test/benchmark call sites now pass `t` or `b`, so there are no remaining callers using the ignorable-error API. The regression test now verifies both the first and repeated failure reports, one initializer invocation, and zero callback invocations.

Verified with all five affected packages compiled, vet across those packages, `make build`, the real cluster reuse test, and `TestDataBranchDiffAsFile`.

XuPeng-SH (approved):
Re-reviewed exact head `438a71bbfbc79450b3ede0c456c5617f5c83ac72`. The previous blocker is resolved: the helper now requires a `testReporter`, all current call sites pass `t`/`b`, and a cached base-cluster initialization error calls `Fatalf` before the test callback can run. I also checked the `sync.Once` state/error publication and the outer deferred mutex unlock on the fatal path.

Validation on the exact head:
- `TestOnceClusterReportsInitializationError`: enumerated, terminal time below timer resolution, `-race -count=100` passed
- `TestBaseClusterOnlyStartOnce`: T=16.26s, N=1, post-measurement race run passed
- `TestDataBranchDiffAsFile`: T=26.59s, N=1, post-measurement race run passed
- full owning packages under race: `pkg/embed`, `pkg/tests/dml`, `pkg/tests/ddl`, `pkg/tests/issues`, `pkg/tests/txnexecutor`
- current-main merge-tree is conflict-free

No blocking findings.

PR #26204, opened 2026-07-27
fix: support ordered GROUP_CONCAT arguments

### LINKED ISSUE #24359: [Bug]: GROUP_CONCAT with ORDER BY panics with index out of range
### Branch

3.0-dev (git_version 5c1a345)

### Actual Behavior

`GROUP_CONCAT` with multiple arguments and `ORDER BY` triggers a panic:

```
ERROR 20101 (HY000): internal error: panic runtime error: index out of range [1] with length 1:
runtime.goPanicIndex
    /home/github/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.4.linux-amd64/src/runtime/panic.go:115
github.com/matrixorigin/matrixone/pkg/sql/colexec/aggexec.(*groupConcatExec).Fill
```

### Reproduction

```sql
SELECT GROUP_CONCAT(attname, ':', att_length ORDER BY attnum)
FROM mo_catalog.mo_columns
WHERE att_database='mo_catalog' AND att_relname='mo_columns' LIMIT 1;
```

### Root Cause

`(*groupConcatExec).Fill` in `pkg/sql/colexec/aggexec` accesses an index beyond the slice length when the function has multiple expression arguments combined with an ORDER BY clause.

### Expected Behavior

Should return the concatenated string of column values separated by the specified separator, ordered by `attnum`.

### Impact

Any query using `GROUP_CONCAT` with multiple arguments + ORDER BY will crash the CN with a panic. This is a server-crash-level bug.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #24359

`GROUP_CONCAT` previously collected every aggregate's `ORDER BY` keys into one
global sort below the aggregation node. Besides the reported multi-argument
index panic, one input order cannot satisfy multiple `GROUP_CONCAT` aggregates
with independent ordering requirements.

This change binds ordering metadata to each `GROUP_CONCAT`, stores its ordering
expressions as hidden aggregate arguments, and sorts each group inside the
aggregate executor through the shared multi-vector sorting implementation.
It also preserves explicit null placement, makes ordered `DISTINCT` sort before
deduplication, and keeps legacy separator-only configuration compatible.

The ordered payload is covered across merge and intermediate serialization
paths. The return type continues to depend only on concatenated arguments, not
on hidden ordering-key types.

Validation:

- CGo-backed tests for `pkg/sort`, `pkg/sql/colexec/order`,
  `pkg/sql/colexec/aggexec`, `pkg/sql/colexec/group`, and `pkg/sql/plan`
- `go build` and `go vet` for affected packages and `pkg/sql/compile`
- `make build`
- original issue SQL against a clean local MatrixOne service
- `mo-tester` BVT: 158/158 statements passed
- changed non-test Go statement coverage: 89.8%


XuPeng-SH (approved):
Deep-reviewed the latest head across binding/config encoding, per-aggregate ordering, DISTINCT selection, partial/remote serialization, H0 and grouped spill, fan-in merge, NULL placement, prepared reuse, and cancellation. Focused sort/aggexec/group tests passed under race. I also measured the multi-key spill heap comparator directly (0 B/op, 0 allocs/op), so it does not reintroduce per-comparison GC pressure. No blocker found.

heni02 (approved):

PR #26206, opened 2026-07-27
fix(bootstrap): skip tenants at upgrade target

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25939 

fix(bootstrap): skip tenants at upgrade target


LeftHandCold (approved):
Reviewed exact head c683a10f. The >= boundary matches the tenant-upgrade invariant. I traced the full async closure: tenant versions are read FOR UPDATE; tenants already at the target still mark their batch ready; the next loop reconciles ReadyTenant to TotalTenant after all tasks are ready; lower versions still execute and higher versions retain the prior skip behavior. No blocking correctness or normal-business regression issues found. Full pkg/bootstrap tests, focused race stress x20, go build, go vet, and git diff --check passed.

XuPeng-SH (approved):
Deep-reviewed exact head `c683a10f6b9c467f6c655c567eb49b4561ae8930`.

The `>=` boundary matches the tenant-version invariant. I traced the full transactional closure: tenant rows are selected `FOR UPDATE`; an at-target tenant skips both the migration handler and version write; the batch is still marked ready; and the next iteration reconciles `ReadyTenant` to `TotalTenant` once no unready task remains. Lower versions still execute normally and higher versions retain the existing skip behavior. No Q1 resource-ownership, Q2 wait-termination, or Q3 boundedness regression found.

Validation passed: the new regression plus both directly affected reconciliation tests each passed `-race -count=100` (terminal T below timer resolution, B=30s, N=100); full `./pkg/bootstrap` passed under `-race`; `go build`, `go vet`, `git diff --check`, latest-main merge-tree, and online CI all passed.

Issue #26107, opened 2026-07-23
[Bug]: Embedded Iceberg credential UT loses logservice connection during CREATE ACCOUNT


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

`issue-25986-main` (CI merge commit for PR #26021)

### Commit ID

`2dced0f074c3737bf9e5a667a287ddceb75c3601`

### Other Environment Information

- Hardware parameters: GitHub Actions Ubuntu/x86 runner
- OS type: Ubuntu/x86
- Others: MatrixOne ALL CI, UT job 89234679631

### Actual Behavior

`TestIcebergSQLEngineEmbeddedCredentialSecurityAndRedaction` failed while creating its test account because the embedded CN lost its connection to logservice/Hakeeper:

```text
Error 1105 (HY000): writeto tcp4 127.0.0.1:43412->127.0.0.1:10980:
read tcp4 127.0.0.1:43412->127.0.0.1:10980: read: connection reset by peer
--- FAIL: TestIcebergSQLEngineEmbeddedCredentialSecurityAndRedaction (1.06s)
```

The preceding logs show Hakeeper connection shutdown and ID allocation failure, followed by `create account` failing. The PR under test changes prepared-statement schema invalidation and does not modify the Iceberg test or logservice lifecycle involved in this failure.

### Expected Behavior

The embedded service should remain available for the full test lifecycle, and the account setup should succeed.

### Steps to Reproduce

1. Open MatrixOne ALL CI run https://github.com/matrixorigin/matrixone/actions/runs/30015600156/job/89234679631.
2. Inspect the failed UT summary.
3. Observe `TestIcebergSQLEngineEmbeddedCredentialSecurityAndRedaction` failing during `create account` after the local logservice connection is reset.

### Additional information

Failing run: https://github.com/matrixorigin/matrixone/actions/runs/30015600156/job/89234679631

The failure is infrastructure/service-lifecycle shaped rather than an assertion mismatch in the credential-redaction behavior.

XuPeng-SH:
The failure recurred on current code in PR #26402's Ubuntu/x86 UT job: https://github.com/matrixorigin/matrixone/actions/runs/30446934445/job/90559727222?pr=26402

Two serialized `pkg/embed` tests failed at the same setup helper after exactly 10.01s:

- `TestIcebergSQLEngineEmbeddedImportNativePinsSnapshot`
[...]

PR #26207, opened 2026-07-27
fix(embed): retry transient Iceberg account creation

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26107 

fix(embed): retry transient Iceberg account creation


XuPeng-SH (approved):
Deep review complete; no blocking issues found.

Key checks:
- The linked connection-reset failure is classified by morpc.IsConnectionError, while non-connection SQL errors still fail immediately.
- CREATE ACCOUNT IF NOT EXISTS makes retry safe when the first attempt commits but its response is lost.
- The 10s context bounds both ExecContext and the 100ms backoff; timer/cancellation paths do not leak or wait indefinitely.
- Moving only the allocation-count assertion behind !race preserves functional JSON-overlaps coverage under race.
- The exact head merges cleanly with current main; GitHub CI is green.

Local verification on 05b2cbf4b85e58bd177c348f37fcd09b533cb94c:
- embed retry helper: -race -count=10
- JSON allocation regression: -count=3
- JSON overlaps functional tests: -race
- go vet ./pkg/embed ./pkg/sql/plan/function

PR #26211, opened 2026-07-27
fix(frontend): enforce global prepared statement limit

### LINKED ISSUE #25982: [Bug]: max_prepared_stmt_count does not limit prepared statements
### Summary

Changing the global `max_prepared_stmt_count` variable does not change the number of prepared statements that MatrixOne allows. The variable reports the configured value, but session-level prepared statement creation continues to use a separate hard-coded limit.

### Tested version

`main` at `8480da08f318a9cb998efb80062d5573da776fb7`.

### Reproduction

As an administrator:

```sql
SET GLOBAL max_prepared_stmt_count = 1;
SELECT @@global.max_prepared_stmt_count;
```

Open a new connection and run:

```sql
PREPARE limit_a FROM 'SELECT 1';
PREPARE limit_b FROM 'SELECT 2';
EXECUTE limit_b;
```

Restore the original value after the test:

```sql
SET GLOBAL max_prepared_stmt_count = 16382;
```

### Actual result

Both prepared statements are created and `EXECUTE limit_b` returns `2`. The behavior reproduces 3/3 times even though `@@global.max_prepared_stmt_count` reports `1`.

### Expected result

The second `PREPARE` should fail after the configured global limit is reached. MySQL 8.0.45 returns error 1461 (`Can't create more than max_prepared_stmt_count statements`).

### Root cause direction

The system variable is defined as a dynamic global variable with a default of 16382 in `pkg/frontend/variables.go`, but prepared statement creation checks `MaxPrepareNumberInOneSession` in `Session.SetPrepareStmt`. That atomic value is initialized independently to 100000 in `pkg/frontend/session.go` and has no update path tied to `max_prepared_stmt_count`.

As a result, the visible system variable and the enforced limit are independent.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25982

`max_prepared_stmt_count` was exposed as a dynamic global system variable, but
prepared statement admission used an unrelated hard-coded limit. As a result,
changing the variable did not affect runtime behavior.

This change:

- makes each session observe the dynamic global variable when admitting a new
  prepared statement;
- retains the existing CN startup setting as a hard upper bound, so deployments
  cannot silently exceed their configured limit;
- returns MySQL error 1461 with SQLSTATE `42000` when the effective per-session
  limit is reached;
- closes newly built prepared statements when admission fails;
- adds unit, dynamic-limit, CN-config compatibility, and BVT coverage.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s
  ./pkg/common/moerr`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s
  ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=600s
  ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=50 -timeout=300s
  -run 'TestPreparedStmtLimit|TestGetMaxPrepareStmtCount' ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s
  ./pkg/cnservice`
- `make build`
- focused BVT with a fresh `mo-service`: 16/16 statements passed
- `git diff --check`

`make static-check` reaches two existing `unused` findings in
`pkg/common/malloc/sane_allocator_cache.go`; that file is unchanged from
`main`.


XuPeng-SH:
Deep review on PR #26211 (`base=main`, head `554017e91a`) found two merge-blocking prepared-statement quota issues.

1. **The new “account-global” prepared-statement cap is still only enforced inside one CN process.** The admission path is entirely local: `SetPrepareStmt()` takes `ses.gSysVars` / cached `SystemVariables` and calls `reservePrepareStmt()` (`pkg/frontend/session.go:1245-1278`; `pkg/frontend/variables.go:1063-1085`). `GSysVarsMgr` itself only holds a local Go map (`pkg/frontend/variables.go:991-1010,1020-1023`), and the live count is the in-memory field `SystemVariables.preparedStmtCount`. There is no shared catalog mutation, lock service, or RPC on this path.

   Concrete failure: set `max_prepared_stmt_count = 2`, connect the same account to CN-A and CN-B, prepare two statements on each CN. All four prepares succeed because each CN sees its own local count `0 -> 2`. That violates the advertised global/account-wide limit immediately.

   **Suggested fix:** move admission accounting to shared cluster state, or explicitly scope the feature back to per-CN / per-session semantics instead of exposing it as a global account limit.

2. **The existing CN startup config `max_prepared_stmt_count` is silently ignored for normal authenticated sessions.** The CN TOML setting still only seeds `MaxPrepareNumberInOneSession` (`pkg/cnservice/types.go:289-290,431-438`), but authenticated login now always calls `InitSystemVariables()` (`pkg/frontend/mysql_protocol.go:1487-1491`; `pkg/frontend/session.go:293-301`), and `SetPrepareStmt()` prefers `ses.gSysVars` whenever it exists (`pkg/frontend/session.go:1248-1255,1269-1278`). `GlobalSysVarsMgr.Get()` builds that sysvar map from `getGlobalSysVars()` defaults plus a tiny TOML overlay (`pkg/frontend/variables.go:991-999`), and `useTomlConfigOverOtherConfigs()` only copies version fields. The sysvar default for `max_prepared_stmt_count` remains `16382` (`pkg/frontend/session.go:1838-1842`; `pkg/frontend/variables.go:2470-2476`).

   Concrete failure: start a CN with `max_prepared_stmt_count = 100` in TOML, do not run `SET GLOBAL`, log in as a normal user, and prepare 101 statements. The 101st now succeeds because the session enforces 16382 from sysvar defaults instead of 100 from CN config.

   **Suggested fix:** seed `max_prepared_stmt_count` from CN config when building account global sysvars, or keep `MaxPrepareNumberInOneSession` as an effective fallback for authenticated sessions so existing deployments retain their configured limit.

ck89119:
@XuPeng-SH Thanks for the detailed review. Both blockers are addressed in
`4377c8c4542eb8a40c20cf07534e5a26a967a7c5`.

1. I removed the CN-local account-wide counter and explicitly scoped admission
   enforcement back to each session. `SetPrepareStmt` now compares the
   session's prepared-statement count directly against the effective limit, so
   the implementation no longer claims cluster-wide/account-wide accounting
   without distributed state.
2. The effective limit is now
   `min(@@global.max_prepared_stmt_count, CN startup max_prepared_stmt_count)`.
   This keeps the global variable dynamically usable for lowering the limit
   while preserving the existing CN configuration as a hard upper bound for
   authenticated sessions.

I also updated the unit tests, BVT, and PR description to match the per-session
semantics. Validation after the change passed for `pkg/common/moerr`, full
`pkg/frontend`, full `pkg/frontend -race`, focused race stress (`-count=50`),
`pkg/cnservice`, `make build`, and the fresh-service BVT (16/16).

heni02 (approved):

XuPeng-SH (approved):
Re-review after commit 4377c8c454: the two previous blockers are resolved. Admission is now explicitly per session, while the effective limit is `min(@@global.max_prepared_stmt_count, CN startup max_prepared_stmt_count)`, preserving the existing deployment cap. Dynamic lowering/zero, replacement, deallocation, rejection cleanup, session cleanup, MySQL 1461/SQLSTATE 42000, locking, and boundedness are consistent.

I found no remaining merge blocker. The PR also merges cleanly with the latest main in a local temporary merge; focused race stress (50x), the full frontend package, and the moerr regression passed on that merged state. Approved.

Issue #25473, opened 2026-07-06
[Bug]: LOAD DATA parquet decimal scale conversion shifts decimal point


Ariznawlll:
### Description

`LOAD DATA INFILE ... format='parquet'` handles Parquet `DECIMAL` values incorrectly when loading into a MatrixOne `DECIMAL` column with a smaller scale.

When the source Parquet column is `decimal128(10,3)` and the target MatrixOne column is `decimal(10,2)`, the value is scaled incorrectly by one decimal digit instead of being rounded/truncated to the target scale.

For example:

- Parquet value `1.235` becomes `12.35` in MO `decimal(10,2)`.
- Parquet value `-1.235` becomes `-12.35`.
- Parquet value `12.345` becomes `123.45`.

The same Parquet file loads correctly into MO `decimal(10,3)` and into `varchar`, so the source file itself is valid. The problem appears when converting a Parquet decimal to a target MO decimal with smaller scale.

### Version

Reproduced on:

```sql
select git_version();
-- 2a008285e
```

### Steps to Reproduce

Generate a Parquet file with a `decimal128(10,3)` column:

```python
from decimal import Decimal
import pyarrow as pa
import pyarrow.parquet as pq

path = '/tmp/decimal_scale.parquet'
vals = [Decimal('1.235'), Decimal('-1.235'), Decimal('12.345'), Decimal('0.001')]
pq.write_table(
    pa.Table.from_arrays(
        [pa.array(vals, pa.decimal128(10, 3))],
        names=['dec_col'],
    ),
    path,
)
```

Run SQL:

```sql
drop database if exists parquet_decimal_scale_check;
create database parquet_decimal_scale_check;
use parquet_decimal_scale_check;

create table t_dec3(dec_col decimal(10,3));
load data infile {'filepath'='/tmp/decimal_scale.parquet', 'format'='parquet'} into table t_dec3;
select * from t_dec3 order by dec_col;

create table t_dec2(dec_col decimal(10,2));
load data infile {'filepath'='/tmp/decimal_scale.parquet', 'format'='parquet'} into table t_dec2;
select * from t_dec2 order by dec_col;

create table t_v(dec_col varchar(32));
load data infile {'filepath'='/tmp/decimal_scale.parquet', 'format'='parquet'} into table t_v;
select * from t_v order by dec_col;
```

### Actual Behavior

Loading into the same scale is correct:

```text
t_dec3:
-1.235
0.001
1.235
12.345
```

[...]

PR #26212, opened 2026-07-27
fix(external): rescale parquet decimals on load

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25473 

fix(external): rescale parquet decimals on load


XuPeng-SH (approved):
Reviewed the decimal rescale path end-to-end.

The new DECIMAL logical-type branch now resizes scale through the value path for both plain and dictionary pages, preserves the existing raw fast-path when source and target scales already match, and adds targeted coverage for scale up/down, overflow, and unsupported physical types. I did not find a high-confidence blocking correctness issue in this diff.

XuPeng-SH (changes requested):
I found one blocking correctness gap in the new decimal-dispatch logic.

- Same-scale DECIMAL loads still bypass target precision enforcement.
  `canUseRawDecimalMapping` returns true for any Parquet DECIMAL whose source scale already matches the target scale, so those values stay on the raw `decodeDecimal{64,128,256}Values` path. That path only checks storage fit, not the destination column's declared DECIMAL precision. A Parquet `DECIMAL(10,2)` value like `12345.67` loaded into a target `DECIMAL(4,2)` column would still be accepted instead of being rejected.

This is now wired directly into the new dispatcher (`canUseRawDecimalMapping`), so I think the safety decision should also consider target precision, not only scale. If precision cannot be proven safe, the logical DECIMAL path should go through the new precision-aware `parquetDecimalValueToDecimal64/128/256` helpers even when the scales already match.

PR #26217, opened 2026-07-27
fix flow control

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25317

调整了 IF、CASE、COALESCE、IFNULL、GREATEST/LEAST 的类型与视图元数据推导：

  - 字符串/数值分支保留精确 VARCHAR 宽度。
  - signed/unsigned 整数分支提升为足够精度的 DECIMAL，避免 UINT64 溢出。
  - DATE/DATETIME CASE 双向提升均正确标记 nullable。
  - binary/character 混合 CASE、IF 返回 VARBINARY，UTF-8 宽度按字节保留；同型 BINARY 不变。
  - IFNULL decimal literal 保留实际精度，如 9.5 -> DECIMAL(2,1)。
  - COALESCE、GREATEST/LEAST 的 nullable 元数据按目标行为修正。
  - 补充 plan/function UT 与 case_when BVT，并清理结果文件尾随空白。


XuPeng-SH (changes requested):
[P1] `binaryStringCommonType` under-allocates valid utf8mb4 values. MatrixOne VARCHAR width is a character count (`makePlan2StringConstExprWithType` uses `RuneCountInString`), while VARBINARY width is bytes. Multiplying by 3 at `pkg/sql/plan/function/operatorSet.go:103` is insufficient for 4-byte code points. On exact head `7284b3d6f0cc266b2fb7efcd4c3b599f8512f8ab`, `CREATE VIEW v AS SELECT CASE WHEN 0 THEN _binary 0x61 ELSE "😀" END AS c` records `VARBINARY(3)`, while `SELECT HEX(c) FROM v` returns `F09F9880` (4 bytes). The view therefore advertises a maximum shorter than values it emits, which can break clients/schema consumers that trust column length. Please derive an actual or safe utf8mb4 byte width (up to 4 bytes per character, capped appropriately), and add 4-byte-character regressions for both CASE and IF.

daviszhen:
> [P1] `binaryStringCommonType` under-allocates valid utf8mb4 values. MatrixOne VARCHAR width is a character count (`makePlan2StringConstExprWithType` uses `RuneCountInString`), while VARBINARY width is bytes. Multiplying by 3 at `pkg/sql/plan/function/operatorSet.go:103` is insufficient for 4-byte code points. On exact head `7284b3d6f0cc266b2fb7efcd4c3b599f8512f8ab`, `CREATE VIEW v AS SELECT CASE WHEN 0 THEN _binary 0x61 ELSE "😀" END AS c` records `VARBINARY(3)`, while `SELECT HEX(c) FROM v` returns `F09F9880` (4 bytes). The view therefore advertises a maximum shorter than values it emits, which can break clients/schema consumers that trust column length. Please derive an actual or safe utf8mb4 byte width (up to 4 bytes per character, capped appropriately), and add 4-byte-character regressions for both CASE and IF.



> [P1] `binaryStringCommonType` under-allocates valid utf8mb4 values. MatrixOne VARCHAR width is a character count (`makePlan2StringConstExprWithType` uses `RuneCountInString`), while VARBINARY width is bytes. Multiplying by 3 at `pkg/sql/plan/function/operatorSet.go:103` is insufficient for 4-byte code points. On exact head `7284b3d6f0cc266b2fb7efcd4c3b599f8512f8ab`, `CREATE VIEW v AS SELECT CASE WHEN 0 THEN _binary 0x61 ELSE "😀" END AS c` records `VARBINARY(3)`, while `SELECT HEX(c) FROM v` returns `F09F9880` (4 bytes). The view therefore advertises a maximum shorter than values it emits, which can break clients/schema consumers that trust column length. Please derive an actual or safe utf8mb4 byte width (up to 4 bytes per character, capped appropriately), and add 4-byte-character regressions for both CASE and IF.

 reviewer comment，修改如下：

  - binaryStringCommonType 将 VARCHAR/CHAR 的字符宽度按 utf8mb4 最大 4 字节转换为 VARBINARY 字节宽度，不再使用错误的 *3。
  - 结果宽度限制在 MaxVarBinaryLen，并处理未知宽度，避免乘法溢出。
  - CASE 和 IF 统一复用该逻辑。
  - 增加 😀 四字节字符回归测试，确认两者均得到 VARBINARY(4)。
  - 更新现有 BVT 中受影响的视图元数据结果。

XuPeng-SH (changes requested):
Two blocking metadata regressions remain on latest head `e8e9e0cdb75126adcafa644f80e85a8d3d11bf3b`.

1. `controlFlowStringWidth` under-allocates DECIMAL values when a control-flow expression mixes strings and numerics. It returns `typ.Width`, but DECIMAL precision excludes the decimal point and sign. A real binder plan gives `VARCHAR(3)` for each of `IF(1, "x", 1.23)`, `COALESCE("x", 1.23)`, and the equivalent CASE, although the emitted value `1.23` is 4 bytes. `IF(1, "x", CAST(-123.45 AS DECIMAL(5,2)))` advertises `VARCHAR(5)` although `-123.45` needs 7 bytes. This recreates the same production contract violation as the previous utf8mb4 issue: views and protocol metadata can advertise a maximum shorter than values they emit. Please derive decimal display width from precision/scale plus decimal separator and possible sign, cap safely, and cover positive/negative literals and DECIMAL columns across IF, CASE, and COALESCE.

2. `binaryStringCommonType` always returns VARBINARY whenever any binary type is present, and `caseCheck` invokes it before the same-type path. Consequently `CASE WHEN 1 THEN CAST("a" AS BINARY(4)) ELSE CAST("b" AS BINARY(4)) END` now plans as `VARBINARY(4)`, while the equivalent IF remains `BINARY(4)`. This contradicts the stated same-type BINARY contract and makes equivalent control-flow forms expose different schema types. Preserve BINARY when every value branch has the same BINARY type/width; use VARBINARY only for genuinely mixed binary/character or binary-width semantics that require it.

The previous utf8mb4 four-byte-width blocker is fixed. CI is green, and clean local tests for `pkg/sql/plan` and `pkg/sql/plan/function` pass.

daviszhen:
> Two blocking metadata regressions remain on latest head `e8e9e0cdb75126adcafa644f80e85a8d3d11bf3b`.
> 
> 1. `controlFlowStringWidth` under-allocates DECIMAL values when a control-flow expression mixes strings and numerics. It returns `typ.Width`, but DECIMAL precision excludes the decimal point and sign. A real binder plan gives `VARCHAR(3)` for each of `IF(1, "x", 1.23)`, `COALESCE("x", 1.23)`, and the equivalent CASE, although the emitted value `1.23` is 4 bytes. `IF(1, "x", CAST(-123.45 AS DECIMAL(5,2)))` advertises `VARCHAR(5)` although `-123.45` needs 7 bytes. This recreates the same production contract violation as the previous utf8mb4 issue: views and protocol metadata can advertise a maximum shorter than values they emit. Please derive decimal display width from precision/scale plus decimal separator and possible sign, cap safely, and cover positive/negative literals and DECIMAL columns across IF, CASE, and COALESCE.
> 2. `binaryStringCommonType` always returns VARBINARY whenever any binary type is present, and `caseCheck` invokes it before the same-type path. Consequently `CASE WHEN 1 THEN CAST("a" AS BINARY(4)) ELSE CAST("b" AS BINARY(4)) END` now plans as `VARBINARY(4)`, while the equivalent IF remains `BINARY(4)`. This contradicts the stated same-type BINARY contract and makes equivalent control-flow forms expose different schema types. Preserve BINARY when every value branch has the same BINARY type/width; use VARBINARY only for genuinely mixed binary/character or binary-width semantics that require it.
> 
> The previous utf8mb4 four-byte-width blocker is fixed. CI is green, and clean local tests for `pkg/sql/plan` and `pkg/sql/plan/function` pass.

 - 修复 DECIMAL 与字符串混合时控制流结果的 VARCHAR 元数据宽度：按 DECIMAL 精度、scale、小数点、可选负号及纯小数前导 0 计算，避免实际值超过声明宽度。
  - 修复 CASE 同型 BINARY(n) 被降级为 VARBINARY(n)：所有分支均为同宽 BINARY 时保留 BINARY；混合字符、VARBINARY 或不同宽度时仍使用 VARBINARY。
  - 补充 IF、CASE、COALESCE 的正/负 DECIMAL literal、DECIMAL 列回归 UT/BVT，以及同型和不同宽度 BINARY 的类型协商覆盖。

heni02 (approved):

XuPeng-SH (changes requested):
One production metadata blocker remains on exact head 1c291b1b80b06ad5c008b3f80785fceb710fa2c4. The two previous blockers are fixed and the focused planner tests pass, but the new binary common-type path does not treat a NULL (`T_any`) value branch as neutral. `binaryStringCommonType` returns false as soon as it sees T_any, so CASE falls through to the old generic BLOB candidate:

- `CASE WHEN TRUE THEN CAST("a" AS BINARY(4)) ELSE NULL END` -> `BLOB(4)`
- reversed branch order -> `BLOB(4)`
- `CASE WHEN TRUE THEN CAST("a" AS VARBINARY(4)) ELSE NULL END` -> `BLOB(4)`
- adding a NULL arm between two identical BINARY(4) arms also -> `BLOB(4)`

The equivalent `IF(TRUE, CAST("a" AS BINARY(4)), NULL)` remains `BINARY(4)`. This violates the PR’s same-type BINARY contract and makes equivalent control-flow forms expose different schema/protocol types; a nullable value branch should affect nullability, not replace the non-NULL branch type with BLOB. Please make T_any neutral when reconciling binary value branches (while still requiring at least one concrete binary branch), preserve BINARY only when all concrete branches are identical fixed BINARY, and add CASE regressions for leading/trailing/middle NULL with BINARY and VARBINARY.

Non-blocking cleanup: `git diff --check` still reports six trailing-whitespace lines in `case_when.result`, despite the PR description saying result whitespace was cleaned.

daviszhen:
> One production metadata blocker remains on exact head [1c291b1](https://github.com/matrixorigin/matrixone/commit/1c291b1b80b06ad5c008b3f80785fceb710fa2c4). The two previous blockers are fixed and the focused planner tests pass, but the new binary common-type path does not treat a NULL (`T_any`) value branch as neutral. `binaryStringCommonType` returns false as soon as it sees T_any, so CASE falls through to the old generic BLOB candidate:
> 
> * `CASE WHEN TRUE THEN CAST("a" AS BINARY(4)) ELSE NULL END` -> `BLOB(4)`
> * reversed branch order -> `BLOB(4)`
> * `CASE WHEN TRUE THEN CAST("a" AS VARBINARY(4)) ELSE NULL END` -> `BLOB(4)`
> * adding a NULL arm between two identical BINARY(4) arms also -> `BLOB(4)`
> 
> The equivalent `IF(TRUE, CAST("a" AS BINARY(4)), NULL)` remains `BINARY(4)`. This violates the PR’s same-type BINARY contract and makes equivalent control-flow forms expose different schema/protocol types; a nullable value branch should affect nullability, not replace the non-NULL branch type with BLOB. Please make T_any neutral when reconciling binary value branches (while still requiring at least one concrete binary branch), preserve BINARY only when all concrete branches are identical fixed BINARY, and add CASE regressions for leading/trailing/middle NULL with BINARY and VARBINARY.
> 
> Non-blocking cleanup: `git diff --check` still reports six trailing-whitespace lines in `case_when.result`, despite the PR description saying result whitespace was cleaned.

  - binaryStringCommonType 将 NULL 分支的 T_any 视为中性，不再导致 CASE 二进制表达式退化为 BLOB。
  - 保留所有非 NULL 分支均为同宽 BINARY(n) 时的 BINARY(n) 类型；出现 VARBINARY、字符类型或不同固定宽度时归并为 VARBINARY。
  - 补充 helper UT，覆盖 BINARY/VARBINARY 的 leading、trailing、middle NULL 分支，以及全 NULL 反例。
  - 补充 binder UT，验证 CASE 返回类型、宽度与可空性。
  - 补充 CASE BVT，通过 view 的 DESC 验证 BINARY(4) / VARBINARY(4) 和 YES 可空性。

XuPeng-SH (changes requested):
[P1] A NULL arm still bypasses the new safe signed/unsigned promotion in CASE. `caseCheck` includes NULL literals as `T_any` in `source`, but `signedUnsignedIntegerCommonType` returns false for every non-integer instead of treating `T_any` as neutral. For example:

```sql
SELECT CASE
  WHEN TRUE THEN CAST(18446744073709551615 AS UNSIGNED)
  WHEN FALSE THEN NULL
  ELSE CAST(-1 AS SIGNED)
END;
```

The value types are `[uint64, any, int64]`, so the helper is skipped. The generic matcher then chooses INT64 (cost 2) ahead of DECIMAL (cost 3). The selected UINT64_MAX branch is converted by `uint64ToOthers`/`numericToNumeric` to INT64 as `-1` (the existing explicit-cast test codifies this wrap), and the expression exposes INT64 metadata instead of the intended safe DECIMAL(21,0). Thus adding an unreachable NULL arm changes both the result type and a valid selected value.

Please make `T_any` neutral in `signedUnsignedIntegerCommonType` while still requiring at least one concrete signed and one concrete unsigned input, and add unit/planner/BVT regressions with leading, trailing, and middle NULL arms, including UINT64_MAX and a negative signed branch.

daviszhen:
> [P1] A NULL arm still bypasses the new safe signed/unsigned promotion in CASE. `caseCheck` includes NULL literals as `T_any` in `source`, but `signedUnsignedIntegerCommonType` returns false for every non-integer instead of treating `T_any` as neutral. For example:
> 
> ```sql
> SELECT CASE
>   WHEN TRUE THEN CAST(18446744073709551615 AS UNSIGNED)
>   WHEN FALSE THEN NULL
>   ELSE CAST(-1 AS SIGNED)
> END;
> ```
> 
> The value types are `[uint64, any, int64]`, so the helper is skipped. The generic matcher then chooses INT64 (cost 2) ahead of DECIMAL (cost 3). The selected UINT64_MAX branch is converted by `uint64ToOthers`/`numericToNumeric` to INT64 as `-1` (the existing explicit-cast test codifies this wrap), and the expression exposes INT64 metadata instead of the intended safe DECIMAL(21,0). Thus adding an unreachable NULL arm changes both the result type and a valid selected value.
> 
> Please make `T_any` neutral in `signedUnsignedIntegerCommonType` while still requiring at least one concrete signed and one concrete unsigned input, and add unit/planner/BVT regressions with leading, trailing, and middle NULL arms, including UINT64_MAX and a negative signed branch.

  - 将 CASE 中的 NULL/T_any 视为类型推导中性值，不再阻断 signed/unsigned 的安全提升。
  - 仍要求至少一个具体 signed 与一个具体 unsigned 分支，统一推导为 DECIMAL(21,0)。
  - 补充 unit、planner 和 BVT，覆盖 NULL 位于 leading/middle/trailing 的场景，并验证 UINT64_MAX 与负数分支结果正确。

XuPeng-SH (changes requested):
Blocking: `iffCheck` still bypasses `binaryStringCommonType` when both branches have the same binary OID. At lines 603-615, `IF(TRUE, CAST("a" AS BINARY(4)), CAST("b" AS BINARY(8)))` takes the legacy shortcut and is typed as `BINARY(8)`, while the equivalent `CASE` reaches the new helper and is typed as `VARBINARY(8)`. MySQL 8.4 reports `VARBINARY(8)` for both expressions. I reproduced the mismatch with a planner regression test: the CASE subtest passes and the IF subtest gets type id 64 (`BINARY`) instead of 65 (`VARBINARY`). Please route different-width fixed-binary IF branches through the common-type helper and add binder/BVT coverage for this case. The new NULL + signed/unsigned tests pass.

daviszhen:
> Blocking: `iffCheck` still bypasses `binaryStringCommonType` when both branches have the same binary OID. At lines 603-615, `IF(TRUE, CAST("a" AS BINARY(4)), CAST("b" AS BINARY(8)))` takes the legacy shortcut and is typed as `BINARY(8)`, while the equivalent `CASE` reaches the new helper and is typed as `VARBINARY(8)`. MySQL 8.4 reports `VARBINARY(8)` for both expressions. I reproduced the mismatch with a planner regression test: the CASE subtest passes and the IF subtest gets type id 64 (`BINARY`) instead of 65 (`VARBINARY`). Please route different-width fixed-binary IF branches through the common-type helper and add binder/BVT coverage for this case. The new NULL + signed/unsigned tests pass.

  - 修复 CASE(UINT64, NULL, INT64)：NULL/T_any 现在是中性分支，不会阻断 signed/unsigned 的安全公共类型推导。
  - 保持必须同时存在具体 signed 与 unsigned 分支；仅 NULL、或 NULL + 单一数值域 不会错误提升。
  - 混合分支统一提升为 DECIMAL(21,0)，避免 UINT64_MAX 被转换成 INT64 后溢出为 -1。
  - 增加 leading/middle/trailing NULL 的 checker、planner binder 与 BVT 覆盖，验证 UINT64_MAX 和 -1 的结果与元数据。

  另：不同宽度 BINARY(4) / BINARY(8) 的 IF 已与 CASE 一致推导为 VARBINARY(8)。

XuPeng-SH (approved):
Reviewed exact head 624a39611f14. The latest update fixes the remaining IF/CASE different-width BINARY reconciliation issue; the prior NULL, signed/unsigned, decimal-width, and utf8mb4 metadata blockers are covered. Local build, vet, and full tests for pkg/sql/plan/function and pkg/sql/plan pass, and current CI is green.

PR #26219, opened 2026-07-27
fix: preserve user tables with internal-looking names

### LINKED ISSUE #26121: [Bug]: Database Data Branch and CLONE filter persistent user tables by internal-looking names
### Describe the bug

Database-level `CLONE` and Data Branch database operations incorrectly ignore
queryable user tables when their names match internal-looking names or
prefixes. The create paths silently omit the tables, while
`DATA BRANCH DELETE DATABASE` can skip the same tables during validation and
delete the database that contains them. Confirmed examples are:

- an ordinary persistent `relkind='r'` table named `__mo_tmp_user_data`;
- an ordinary persistent `relkind='r'` table named `__mo_account_lock`;
- a user-created, queryable table named `mo_increment_columns`.

The tables remain queryable and can be copied correctly with the table-level
Data Branch statement. The database-level create statements nevertheless
return success without creating them in the destination.

### Steps to reproduce

```sql
drop database if exists g4d_tmp_prefix_branch;
drop database if exists g4d_tmp_prefix_clone;
drop database if exists g4d_tmp_prefix_src;

create database g4d_tmp_prefix_src;
use g4d_tmp_prefix_src;
create table `__mo_tmp_user_data` (
  id int primary key,
  note varchar(30)
);
create table `__mo_account_lock` (
  id int primary key,
  note varchar(30)
);
create table `mo_increment_columns` (
  id int primary key,
  note varchar(30)
);
insert into `__mo_tmp_user_data` values (1, 'persistent-user-row');
insert into `__mo_account_lock` values (2, 'ordinary-lock-name');
insert into `mo_increment_columns` values (3, 'ordinary-auto-name');

select relname, relkind
from mo_catalog.mo_tables
where account_id = 0
  and reldatabase = 'g4d_tmp_prefix_src'
  and relname in (
    '__mo_tmp_user_data',
    '__mo_account_lock',
    'mo_increment_columns'
  );

data branch create table direct_branch
  from `__mo_tmp_user_data`;
data branch create table direct_lock
  from `__mo_account_lock`;
data branch create table direct_auto
  from `mo_increment_columns`;
select * from direct_branch;
select * from direct_lock;
select * from direct_auto;

data branch create database g4d_tmp_prefix_branch
  from g4d_tmp_prefix_src;
create database g4d_tmp_prefix_clone
  clone g4d_tmp_prefix_src;

select reldatabase, relname, relkind
from mo_catalog.mo_tables
where account_id = 0
  and reldatabase in (
    'g4d_tmp_prefix_branch',
    'g4d_tmp_prefix_clone'
  )
  and relname in (
    '__mo_tmp_user_data',
    '__mo_account_lock',
    'mo_increment_columns'
  );

-- Deletion-validation comparison.
data branch create database g4d_delete_normal
  from g4d_tmp_prefix_src;
create table g4d_delete_normal.ordinary_keep(
  id int primary key,
  note varchar(30)
);
insert into g4d_delete_normal.ordinary_keep values (10, 'must-remain');
data branch delete database g4d_delete_normal;

data branch create database g4d_delete_filtered
  from g4d_tmp_prefix_src;
create table g4d_delete_filtered.`__mo_tmp_user_keep`(
  id int primary key,
  note varchar(30)
);
insert into g4d_delete_filtered.`__mo_tmp_user_keep`
  values (11, 'must-remain');
data branch delete database g4d_delete_filtered;

select count(*)
from mo_catalog.mo_database
where datname = 'g4d_delete_filtered';
```

### Actual behavior

The first two source tables are normal persistent relations:

```text
__mo_tmp_user_data  r
__mo_account_lock   r
```

All three table-level branches contain the expected rows:

```text
1  persistent-user-row
2  ordinary-lock-name
3  ordinary-auto-name
```

Both database-level statements return success, but the destination metadata
query returns no row for any of the three tables.

The Data Branch database path was reproduced with three independent
destinations. All three silently contained only the other source tables.

For deletion, the ordinary `ordinary_keep` table correctly prevents
`DATA BRANCH DELETE DATABASE`, and both the database and row remain. Replacing
that table with the normal persistent relation `__mo_tmp_user_keep`
(`relkind='r'`) changes the result: the delete statement succeeds and the
database count becomes zero. The prefix case was reproduced with three
independent branch databases. The exact names `__mo_account_lock` and
`mo_increment_columns` are ignored by the same validation path.

### Expected behavior

A database-level copy should not omit a queryable user-created table solely
because its legal name resembles an internal object name. It should distinguish
internal relations using object metadata rather than only the user-visible
name.

At minimum, a database-level statement must not report a complete successful
copy while silently dropping a source base table. The delete statement must
also reject a database that contains any ordinary non-branch table, regardless
of that table's legal name.

### Code context

Both database-level paths use `collectCloneDatabaseSource`, which calls
`getTableInfos`. `buildTableInfoListWhereClause` excludes every matching name:

```go
relname not like '__mo_tmp\\_%' escape '\\'
```

The same filter also excludes the exact names `mo_increment_columns` and
`__mo_account_lock`. These predicates do not verify that a relation is the
corresponding internal object. Consequently, queryable user tables are removed
from `srcTblInfos` before cloning begins.

`branchDeleteDatabaseTableIDsSQL` reuses
`buildTableInfoListWhereClause`. As a result, the same user tables are removed
from the set passed to `validateActiveBranchChildTableIDs`. The validator never
sees them, so database deletion proceeds even though they are ordinary
non-branch tables.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26121

Database-level CLONE and Data Branch operations filtered source tables by internal-looking names. Legal persistent user tables could therefore be silently omitted from copies, and DATA BRANCH DELETE DATABASE could skip them during validation.

This change identifies actual index, temporary, partition, and sequence relations by `relkind`. The two exact special-table exclusions remain scoped to `mo_catalog`, so snapshot/restore behavior for real system tables is preserved.

Regression coverage exercises all three reported name shapes through Data Branch database creation, database CLONE, and destructive Data Branch database deletion validation. The test failed on the old implementation and passes after the fix.

Validation:

- `go test ./pkg/tests/issues -run '^TestIssue26121DatabaseOperationsKeepOrdinaryInternalLookingTables$' -count=1`
- schema v4 exact-head preflight: `PASS review=PASS`


XuPeng-SH (approved):
Reviewed the complete relkind/index-plugin filtering closure across database CLONE, Data Branch create/delete validation, snapshot restore, FK ordering, and hidden index-table reconstruction. No blocking findings. Focused SQL-shape tests, the issue-level embedded regression, and the full frontend package pass on a local merge with latest main. Please update the branch to latest main and rerun CI before merging.

Issue #25952, opened 2026-07-21
[Bug]: TestDedupSnapshot1 intermittently observes a pending WAL entry in Ubuntu/x86 UT


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

issue-25752-main

### Commit ID

cdbdb261309652b76f35ef906db190ac5796d83e

### Other Environment Information

- Hardware parameters: GitHub-hosted Ubuntu/x86 runner
- OS type: Ubuntu 22.04
- Others: MatrixOne ALL CI, UT_PARALLEL=6, UT_TIMEOUT=40

### Actual Behavior

The Ubuntu/x86 unit-test job failed in pkg/vm/engine/tae/db/test.TestDedupSnapshot1.

After waiting up to 10 seconds for pending WAL entries to drain, the assertion at pkg/vm/engine/tae/db/test/db_test.go:9242 observed one pending entry:

    Error Trace: pkg/vm/engine/tae/db/test/db_test.go:9242
    Error:       Not equal:
                 expected: 0x0
                 actual  : 0x1
    Test:        TestDedupSnapshot1
    --- FAIL: TestDedupSnapshot1 (21.55s)

The failure was observed while validating PR #25861. The PR changes planner numeric assignment context and does not touch the failing TAE package, so this appears unrelated to that PR and should be tracked independently.

### Expected Behavior

TestDedupSnapshot1 should deterministically observe zero pending WAL entries after WaitExpect and pass in the Ubuntu/x86 UT job.

### Steps to Reproduce

1. Check out commit cdbdb261309652b76f35ef906db190ac5796d83e.
2. Run the CI unit-test command:

       make clean
       make config
       make ut UT_PARALLEL=6 UT_TIMEOUT=40

3. Observe TestDedupSnapshot1 failing because tae.Wal.GetPenddingCnt() returns 1 instead of 0.

CI occurrence:

- Workflow run: https://github.com/matrixorigin/matrixone/actions/runs/29807628095
- Failed job: https://github.com/matrixorigin/matrixone/actions/runs/29807628095/job/88561413722

### Additional information

The test calls testutils.WaitExpect(10000, ...) before asserting the pending WAL count, but the condition was still false when the assertion ran. This may indicate timing sensitivity or incomplete synchronization around WAL draining.

jiangxinmeng1:
fixed by #26221

PR #26221, opened 2026-07-27
test: stabilize dedup snapshot checkpoint wait

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25952

## Summary

  Stabilize `TestDedupSnapshot1` while preserving coverage of the asynchronous
  background flush and checkpoint path.

  ## Root Cause

  The test waited only 10 seconds for `Wal.GetPenddingCnt()` to become zero. In the
  reported CI failure, `writer.Sync` alone took 20.39 seconds, so the incremental
  checkpoint could not advance the checkpointed WAL LSN before the assertion timed
  out.

  The previous condition also observed only the WAL pending count and did not
  ensure that the background ICKP had fully finished or covered the transaction
  under test.

  ## Changes

  - Wait up to 30 seconds for the real background checkpoint flow without calling
  `ForceCheckpoint`.
  - Capture the first transaction's WAL LSN and require the checkpointed LSN to
  cover it.
  - Require `AllCheckpointsFinished()` so the ICKP intent is fully completed.
  - Remove the test-specific global flush fault injection and its excessive
  scheduling logs.
  - Check both append and commit results for the incremental dedup transaction.
  - Roll back the transaction if append fails.
  - Scan the relation after commit and verify that all 20 rows are present.

  ## Validation

  - `TestDedupSnapshot1 -count=100`: passed
  - `TestDedupSnapshot1 -race -count=100`: passed, no data races
  - Full `pkg/vm/engine/tae/db/test` package: passed
  - `go build ./pkg/vm/engine/tae/db`: passed
  - `go vet ./pkg/vm/engine/tae/db/test`: passed
  - Pre-push self-review: no remaining blockers


XuPeng-SH (approved):
Deep-reviewed at head 4e0c34ac56960370f886783ccf8410b7e5c872e5. The new wait condition matches the existing WAL/checkpoint completion contract and covers the target transaction through a finished ICKP. The removed fault injection only enabled verbose flush logging. Commit/append outcomes and final row visibility are now asserted. Local validation passed 20 repeated runs plus 10 race-enabled runs. No blocking issue found.

PR #26225, opened 2026-07-27
Optimize TN ObjectList ordering and traversal

### LINKED ISSUE #26223: [Enhancement]: Optimize TN ObjectList ordering and bounded traversal
## Background

TN `catalog.ObjectList` currently orders all object entries by the maximum of `CreatedAt` and `DeletedAt`. Appendable and non-appendable objects, serving create entries, create entries with a drop counterpart, and drop entries are interleaved by timestamp.

This makes consumers scan unrelated object categories and rely on implicit global timestamp ordering. It also prevents safe group-bounded seeks and early termination in hot traversal paths.

## Proposal

Order `ObjectList` entries in six contiguous groups:

1. appendable create-only entries
2. appendable create entries with a drop counterpart
3. appendable drop entries
4. non-appendable create-only entries
5. non-appendable create entries with a drop counterpart
6. non-appendable drop entries

Within each group, order entries by their entry commit timestamp (`CreatedAt` for create entries and `DeletedAt` for drop entries), then by object name as a deterministic tie-breaker.

## Scope

- Update `ObjectEntry.Less` and ObjectList ID indexing for the new composite key.
- Preserve BTree ordering across create-to-drop, rollback, commit, WAL replay, and checkpoint replay transitions.
- Update consumers that rely on the old global timestamp order.
- Use group-aware seek helpers and stop traversal at group or timestamp boundaries.
- Avoid unnecessary row-tree copies and scans for non-appendable object-list replay.
- Add ordering, seek-boundary, transition, replay, and incremental traversal tests.

## Acceptance criteria

- Object entries are returned in the six-group order above.
- Entries inside each group are ordered by commit timestamp and object name.
- ID lookup still returns the latest object version correctly.
- Drop, rollback, and replay preserve immutable BTree snapshot ordering.
- Logtail, checkpoint flusher, soft-delete collection, and incremental dedup/delete traversal remain correct.
- Traversals with safe bounds exit at the group/time boundary.
- Relevant package tests, race tests, `go vet`, and end-to-end logtail regression pass.

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26223

 - Reorder TN ObjectList entries into six contiguous groups: appendable/non-
    appendable, each split into create-only, create-with-drop, and drop entries.

  - Sort entries within each group by commit timestamp and object name.
  - Preserve BTree correctness across drop, rollback, commit, WAL replay, and
    checkpoint replay transitions.

  - Optimize logtail, checkpoint, soft-delete, and incremental deduplication
    traversals with group-aware seeks and early termination.

  - Avoid unnecessary row-tree scans during CN object-list replay.
  - Add ordering, transition, replay, and traversal-boundary tests, including race
    coverage.


XuPeng-SH:
Deep review on PR #26225 (`base=main`, head `3ef981866a`) found one merge-blocking soft-delete regression.

**Backup-replayed deleted objects are now skipped by soft-delete collection.** `GetSoftdeleteObjects()` only traverses `ObjectListGroupAppendableDrop` and `ObjectListGroupNonAppendableDrop` (`pkg/vm/engine/tae/catalog/table.go:200-214`). But the backup replay path still reconstructs some committed deleted objects by setting `CreateNode` / `DeleteNode` directly, without linking `prevVersion` / `nextVersion` (`pkg/vm/engine/tae/catalog/catalogreplay.go:796-817`). `ObjectListGroup()` does not look at `DeletedAt`; it classifies strictly by version links (`pkg/vm/engine/tae/catalog/object.go:131-136,473-484`). So this replayed deleted-object shape is classified as a create-group entry and never returned by `GetSoftdeleteObjects()`.

Before this PR, the soft-delete collector filtered by `DeletedAt` directly, so it still picked these replayed deleted objects up. After this PR, `txn/txnimpl/table.go` depends on `GetSoftdeleteObjects()` for delete transfer (`pkg/vm/engine/tae/txn/txnimpl/table.go:351-355`). Skipping the replayed deleted object means its delete state is not transferred/processed after backup recovery.

Concrete failure: replay a backup containing a committed deleted object in this backup-only shape, then run the tombstone-transfer path. The object is absent from `GetSoftdeleteObjects()`, so previously deleted rows are missed after recovery.

**Suggested fix:** either replay these backup deleted objects as proper D-entries with linked versions, or make soft-delete grouping/collection fall back to `DeletedAt` / `DeleteNode` for this replayed shape.

XuPeng-SH (changes requested):
see latest comment

XuPeng-SH (approved):
Re-reviewed the latest head. The backup checkpoint replay path now reconstructs linked C/D entries, preserves soft-delete grouping, and covers idempotent replay. I also verified the six-group state transitions and bounded logtail/checkpoint/dedup traversals; targeted/full affected-package tests, race tests, and go vet pass locally.

gouhongshen (approved):
## Codex automated review

No blocking or non-blocking defects found. The prior backup-replay finding is addressed at the current head, and the reordered state transitions and bounded traversals preserve prior behavior. Focused tests could not run in this checkout because the required prebuilt cgo/libmo.dylib is absent.

Issue #26202, opened 2026-07-27
[Bug] Stability (2734005 / 66dad1483): TPCH q15 hangs forever in RemoteRun receiveMessage + PipelineSignalReceiver; found leak txn unique=1 on live session


Ariznawlll:
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-stb-4-2-dev-2734005-202607231128` |
| Namespace | `mo-stb-4-2-dev-2734005-202607231128` |
| Image | **`matrixone:commit-66dad1483`** (rolled onto this soak ~2026-07-24 14:16 CST) |
| Endpoint | `10.222.6.255:6001` |
| Artifact root | `/data1/stability-test/29976993561_2734005` |
| Workloads | TPCC, sysbench-1, TPCH, qvec, fulltext, issue-24446, issue-10313 |
| Hang window | **TPCH q15 started 2026-07-26 15:05:28 CST**; still `COM_QUERY` at analysis **2026-07-27 ~11:25 CST** (~20h+) |

**Scope:** On a healthy soak (no CN roll at hang time), **TPCH q15** never returns. Session stays alive in `SHOW PROCESSLIST`. Leak-check periodically emits **`found leak txn`** for the same `txn-id` → quality gate **FAIL** (unique=1). Goroutine dump shows the statement stuck in **CN↔CN RemoteRun receive** and local **`PipelineSignalReceiver`**, not LockService bind.

**Out of scope / not this issue:**
- Lock bind / `20702` / `failed to keep lock table bind` — **0** after the Jul-24 image roll settled (only roll-window noise).
- `wait-active` / `createTxnOpUnsafe` / panic / `+Inf` / OOM — **0** in the hang window.
- TPCC consistency — **PASS** throughout.
- #24766-class disconnect-without-rollback orphans (here the mysql client and session are **still connected**).

**Related:** #26125 (roll-time hangs; aptend: nested shuffle / remote dispatch paths), #25816 (`PipelineSignalReceiver`), #26025 / #26055 / #26097 (candidate fixes — need confirm whether **`66dad1483` includes them**; hang still reproduces on this image).

## Summary

| Signal | Value | Interpretation |
|--------|-------|----------------|
| Hung SQL | TPCH **q15** (CTE `q15_revenue0` + join + nested `max`) | Not a slow query — **20h+** with no progress |
| `txn-id` | `18dd0759dd8a4d5618c52552667de939` | Same id in processlist + leak logs |
[...]

Issue #26125, opened 2026-07-24
[Bug] Stability (3ebdb25 / 90e3400): After CN roll, some statements hang forever on `lock table bind changed` (no `Exec.Run` return); `found leak txn` flags live sessions


heni02:
## Environment

| Field | Value |
|-------|--------|
| Cluster | `mo-stb-main-3ebdb25-202607101119` |
| Namespace | `mo-stb-main-3ebdb25-202607101119` |
| Image (incident) | **`matrixone:commit-90e3400`** (full-stack roll inside the window) |
| Endpoint | `10.222.6.254:6001` |
| Artifact root | `/data1/stability-test/29066558663_3ebdb25` |
| Workloads | TPCC, sysbench, TPCH, fulltext/vector (`ivf_dml_scenarios`) |
| Analysis window | **2026-07-23 10:00 CST → 2026-07-24 ~10:40 CST** (~24h; roll is inside the window) |

**Scope:** After a CN / full-stack roll, a **small subset** of statements already on the execution path **never return** during a LockService **`lock table bind changed`** storm (stuck after `Exec.Build` completes or after `ranges`), while the client blocks until `query_result_timeout` / `wait_timeout` (~24h). Peer statements on the **same table** correctly fail after a **10s retry budget**. Leak-check reports those still-`COM_QUERY` aged txns as **`found leak txn`** (unique=2) → release gate **FAIL**.

**Out of scope / not this issue:**
- #24766-class “disconnect without Rollback” orphan txns (here the session is **still alive**; processlist shows the same `session_id` / `txn_id`).
- Durable consistency corruption (consistency **PASS** in this window).
- `wait-active` / `createTxnOpUnsafe` / panic / `+Inf` (all **0** in this window).

## Summary

| Signal | Value | Interpretation |
|--------|-------|----------------|
| Roll time | **07-23 11:38–11:42 CST** | log/DN/proxy/CN restart steps; CN `qr22s` → `swzgh` |
| Hung #1 | TPCH **q10** `txn=54298e79…e6ef` | After `Exec.Build` @11:40:15 → **23h+ with no `Exec.Run` completion** |
| Hung #2 | vector **UPDATE** `test_vectors` `txn=442f9f6c…cf8a` | Silent after `ranges` @11:41:22; tbl-id **318606** |
| Same-second peer fate | 11:41:32 **`lock retry budget exhausted`** / `lock table bind changed` | **Same table** other INSERT/UPDATE **fail correctly** |
[...]

PR #26228, opened 2026-07-27
[4.2-dev] fix(pipeline): wake receivers from terminal edge state

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #26202 and #26125.

This PR targets `4.2-dev` for #26202. The same logical change needs to be forward-ported to `main` for the #26125 reproduction.

A fatal `PipelineEdge` terminal is recorded and closes `Done()` before all required terminal signals are enqueued into `Ch2`. When `Ch2` is backpressured, the typed terminal signal can remain partially or completely undelivered. `PipelineSignalReceiver` previously waited only on its user context and `Ch2`, so it could remain blocked after the edge was already terminal. That leaves `Merge`, `RemoteRun`, and ultimately `Compile.runOnce` waiting indefinitely.

This change:

- listens to each edge `Done()` channel in both the optimized 1–8 input selects and the `reflect.Select` path;
- drains buffered `Ch2` signals before synthesizing a missing terminal from the durable edge state;
- synchronizes and rechecks `Ch2` to preserve ordering across the fatal-state/enqueue race;
- makes fatal terminal enqueue non-blocking while holding the terminal-state mutex, avoiding a control-path wait cycle behind the data channel;
- adds regressions for buffered ordering, undelivered fatal terminals, shared edges, connector fallback cleanup, and reflect-select removal/reindexing across the 9-to-8 input transition.

Normal `End` delivery and the legacy nil-batch compatibility path remain unchanged.

## Validation

- `go build ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/merge ./pkg/sql/colexec/dispatch`
- `go vet ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/merge ./pkg/sql/colexec/dispatch`
- focused new regressions with `-race -count=20`
- `go test -race -count=1 -timeout=600s ./pkg/vm/process ./pkg/sql/colexec/connector`
- `go test -count=1 -timeout=600s ./pkg/vm/process ./pkg/sql/colexec/connector`
- `go test -count=1 -timeout=600s ./pkg/sql/colexec/merge ./pkg/sql/colexec/dispatch`
- `git diff --check`

### DIFF
--- a/pkg/sql/colexec/connector/types_test.go
+++ b/pkg/sql/colexec/connector/types_test.go
@@ -179,6 +179,48 @@ func TestConnectorResetFallsBackToAbortWhenEndSignalCannotBeDelivered(t *testing
 	require.Same(t, process.ErrPipelineEndSignalDeliveryFailed, info)
 }
 
+func TestConnectorResetUndeliveredFallbackAbortWakesReceiver(t *testing.T) {
+	oldSignalSendTimeout := process.PipelineSignalSendTimeout
+	process.PipelineSignalSendTimeout = 10 * time.Millisecond
+	t.Cleanup(func() {
+		process.PipelineSignalSendTimeout = oldSignalSendTimeout
+	})
+
+	reg := process.NewPipelineEdge(1, 0)
+	reg.Ch2 <- process.NewPipelineSignalToDirectly(batch.EmptyBatch, nil, nil)
+
+	conn := &Connector{Reg: reg}
+	conn.Reset(nil, false, nil)
+
+	receiverCtx, cancelReceiver := context.WithCancel(context.Background())
+	defer cancelReceiver()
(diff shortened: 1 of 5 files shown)


XuPeng-SH (approved):
Deep-reviewed the terminal-edge wakeup protocol on 9c7d4b93. Done wakeups preserve queued Ch2 ordering, synchronize with in-flight fatal senders, and keep shared-edge and reflect-select bookkeeping correct. Focused race tests (20x), affected/dependent package tests, build, and vet all pass locally. The remaining Coverage check failure is an artifact-collection issue (missing compose BVT profile); the latest SCA, UT, and BVT checks pass.

PR #26230, opened 2026-07-27
fix: encode clone object keys losslessly

### LINKED ISSUE #26118: [Bug]: Database CLONE and DATA BRANCH mishandle object identifiers containing #
## Summary

`DATA BRANCH CREATE DATABASE` mishandles database objects when a database, foreign-key table, or dependent view name contains the legal identifier character `#`.

For foreign-key table names containing `#`, the statement reports success, but none of the affected tables or rows are copied. For a source database name containing `#`, the statement fails while resolving the foreign key and rolls the destination back. Dependent view names containing `#` either make the operation fail or produce a successful destination that is missing all views. The ordinary database `CLONE` path has the same table-name behavior because both paths share the database source collector.

## Environment

- Branch: `main`
- MatrixOne endpoint: local standalone server

## Reproduction

```sql
DROP DATABASE IF EXISTS g4d_hash_fk_src;
DROP DATABASE IF EXISTS g4d_hash_fk_dst;
CREATE DATABASE g4d_hash_fk_src;
USE g4d_hash_fk_src;

CREATE TABLE `parent#p` (
    id INT PRIMARY KEY,
    note VARCHAR(32)
);
CREATE TABLE `child#c` (
    id INT PRIMARY KEY,
    parent_id INT,
    CONSTRAINT fk_hash_parent
        FOREIGN KEY (parent_id) REFERENCES `parent#p`(id)
);

INSERT INTO `parent#p` VALUES (1, 'one'), (2, 'two');
INSERT INTO `child#c` VALUES (10, 1), (20, 2);

DATA BRANCH CREATE DATABASE g4d_hash_fk_dst FROM g4d_hash_fk_src;
SHOW FULL TABLES FROM g4d_hash_fk_dst;
```

The source objects and join are valid:

```text
child#c   BASE TABLE
parent#p  BASE TABLE

10  one
20  two
```

## Actual result

`DATA BRANCH CREATE DATABASE` returns success, while `SHOW FULL TABLES FROM g4d_hash_fk_dst` returns no rows.

The result reproduced in 3 out of 3 independent destination databases. The source database remained readable after every attempt.

The shared ordinary path behaves the same way:

```sql
CREATE DATABASE g4d_hash_fk_clone_dst CLONE g4d_hash_fk_src;
SHOW FULL TABLES FROM g4d_hash_fk_clone_dst;
```

It also succeeds with an empty destination.

A source database name containing `#` reproduces the same key-decoding defect through a different later path:

```sql
CREATE DATABASE `g4d#hash_fk_src`;
USE `g4d#hash_fk_src`;
CREATE TABLE parent_p (id INT PRIMARY KEY);
CREATE TABLE child_c (
    id INT PRIMARY KEY,
    parent_id INT,
    FOREIGN KEY (parent_id) REFERENCES parent_p(id)
);
DATA BRANCH CREATE DATABASE g4d_hash_fk_db_dst FROM `g4d#hash_fk_src`;
```

This variant failed in 3 out of 3 attempts with:

```text
internal error: can't find fkTable from fk ...child_c.(parent_id)
```

The failed destination database was absent after each attempt, so rollback remained atomic.

Dependent views are affected by the same key representation:

```sql
CREATE DATABASE g4d_hash_view_src;
USE g4d_hash_view_src;
CREATE TABLE base_t (id INT PRIMARY KEY, v INT);
CREATE VIEW `view#1` AS SELECT id, v FROM base_t;
CREATE VIEW `view#2` AS SELECT id, v + 1 AS v2 FROM `view#1`;

DATA BRANCH CREATE DATABASE g4d_hash_view_dst FROM g4d_hash_view_src;
SHOW FULL TABLES FROM g4d_hash_view_dst;
```

Across four attempts, every result was incorrect:

- two attempts returned success but copied only `base_t`, silently omitting both views;
- two attempts failed while restoring `view#1` and rolled the destination back.

A single view whose name contains `#` but does not depend on another view copied successfully. The defect appears when the encoded object keys participate in dependency ordering.

## Expected result

Both foreign-key-related tables, their data, and the foreign-key definition should be present in the destination database. If this identifier form cannot be supported, the statement should return an explicit error and leave no successful-but-empty result.

## Code analysis

Database cloning represents an object key as:

```go
const KeySep = "#"

func genKey(dbName, tblName string) string {
    return fmt.Sprintf("%s%s%s", dbName, KeySep, tblName)
}
```

The reverse operation uses an unrestricted split and keeps only the first two components:

```go
func splitKey(key string) (string, string) {
    parts := strings.Split(key, KeySep)
    if len(parts) >= 2 {
        return parts[0], parts[1]
    }
    return parts[0], ""
}
```

For `g4d_hash_fk_src#parent#p`, this resolves the table name as `parent` instead of `parent#p`. `getTableInfoMap` therefore cannot populate the expected table information. Later, `handleCloneDatabaseWithSource` skips these tables in the ordinary loop because they are foreign-key members, but the foreign-key loop has no valid table information to clone.

For a key such as `g4d#hash_fk_src#parent_p`, the decoded database and table are `g4d` and `hash_fk_src`, so foreign-key restoration cannot find the referenced table. The same lossy decoding is used while rewriting and ordering dependent views.

Relevant locations:

- `pkg/frontend/util.go`: `genKey`, `splitKey`
- `pkg/frontend/snapshot.go`: `getFkDeps`, `getTableInfoMap`, `fkTablesTopoSort`
- `pkg/frontend/clone.go`: `handleCloneDatabaseWithSource`

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26118

Database clone dependency maps encoded object identity as `database#relation` and decoded it with an unrestricted split. A legal `#` in either identifier truncated the key, causing foreign-key tables and dependent views to fail or be silently omitted.

This change uses a versioned byte-length representation only when an identifier contains `#`. Ordinary and legacy keys keep their existing `database#relation` bytes and `SplitN` behavior, including literal backslashes. The encoded form is reversible when either or both components contain arbitrary `#` sequences and UTF-8 text.

Regression coverage verifies:

- codec round trips for `#` on either side, both sides, adjacent/boundary runs, backslashes, UTF-8, and empty components;
- database Data Branch and ordinary database CLONE both preserve FK tables, rows, FK metadata, and two dependent views when database/table/view names contain `#`.

Validation:

- `go test ./pkg/frontend -run '^TestObjectKeyRoundTrip$' -count=1`
- `go test ./pkg/tests/issues -run '^TestIssue26118DatabaseCopiesPreserveHashIdentifiers$' -count=1`
- schema v4 exact-head preflight: `PASS review=PASS`


XuPeng-SH (changes requested):
Blocking finding — `pkg/tests/issues/issue_26118_test.go:46`

`prefixDB = "\x001"` is a Go hex escape, so this creates a database whose name contains a NUL byte; it is not a literal `\x001` name and it no longer exercises the current `##` codec prefix. On exact head `9982e178`, the deferred `DROP DATABASE` hits `panic delete table failed`, while `execSQLMaybe` discards the error. Since `RunBaseClusterTests` reuses one global cluster, the database remains behind: `-count=2` makes iteration 1 pass and iteration 2 fail with `database \x001 already exists`.

Please use a printable name that covers the intended boundary (for example a name beginning with `##` for the current prefix, or a raw string if the intent is a literal backslash) and ensure the created database is actually removable / cleanup failure is visible. The production codec and clone producer/consumer paths otherwise looked closed in this review.

VioletQwQ-0:
@XuPeng-SH Addressed the formal review finding in 3c0a54a69f: the regression now exercises a printable database name beginning with ##, and every setup/deferred cleanup statement requires success. The old test reproduces the shared-cluster leak with -count=2; the repaired test passes with -count=2. Also passed the targeted codec/frontend/planner tests and the full pkg/tests/issues package. Please re-review when convenient.

XuPeng-SH (approved):
Re-reviewed on 3c0a54a6. The previous blocker is resolved: the regression now uses a printable ##-prefixed database, setup/cleanup errors are surfaced, and the exact embedded-cluster test passes with -count=2 without leaking state. I also rechecked codec unambiguity, UTF-8 byte lengths, legacy ordinary keys/backslashes, and producer/consumer consistency across clone, FK, snapshot/PITR, dependency sorting, and view privilege parsing. Full objectkey/frontend/plan tests, build, vet, and all current CI checks pass.

PR #26236, opened 2026-07-27
fix(parser): reject NULLS FIRST/LAST in MySQL syntax

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25332 #26243

fix(parser): reject NULLS FIRST/LAST in MySQL syntax


XuPeng-SH (changes requested):
[P2] Please update the remaining user-facing `NULLS LAST` example before rejecting the syntax.

`clients/python/guides/json_type_guide.md:785` still tells users to run:

```sql
ORDER BY specifications NULLS LAST;
```

After this PR, that documented query deterministically returns the new parser error. The PR already rewrites the BVT callers, so the same compatibility migration needs to include this guide (for example, `ORDER BY specifications IS NULL, specifications`) to keep the repository user-facing SQL valid.

I otherwise verified that `mysql_sql.go` regenerates cleanly from `mysql_sql.y`, the MySQL parser package tests pass, and the JSON_OVERLAPS tests pass in both normal and `-race` builds.

XuPeng-SH (changes requested):
Reviewed exact head 611669f1359f32eee2aad7ddcc0b8a3e8766b741 against current main c8e443bf86.

The previous documentation finding is fixed, the parser generated file reproduces cleanly from mysql_sql.y, and the focused MySQL parser plus JSON_OVERLAPS normal/race tests pass on the PR head. One current-main merge blocker remains:

[P1] Drop the duplicate !race JSON_OVERLAPS allocation test after updating from main.

Current main already contains pkg/sql/plan/function/func_builtin_json_overlap_alloc_test.go from commit 251dcfefcd, with the same TestJSONOverlapsAccessorAllocationsDoNotScaleWithRows function. This PR independently adds pkg/sql/plan/function/func_json_overlap_alloc_test.go with that function. The files have different names, so Git auto-merges them without a text conflict and retains both.

I reproduced the resulting target-branch build failure by merging current origin/main into this exact head and running the focused package test:

pkg/sql/plan/function/func_json_overlap_alloc_test.go:31:6: TestJSONOverlapsAccessorAllocationsDoNotScaleWithRows redeclared in this block
pkg/sql/plan/function/func_builtin_json_overlap_alloc_test.go:31:6: other declaration of TestJSONOverlapsAccessorAllocationsDoNotScaleWithRows

Please rebase/update from main and drop the now-redundant f7b6c01d65 test move (or otherwise retain only main's existing allocation-test file), then rerun CI on the updated merge result. The parser fix itself has no other actionable finding from this review.

XuPeng-SH (changes requested):
Reviewed exact head `32559c268c211444833adb3de7e6bd7cd5e886ee` against current `main@159742db6672ff5dadb617880b831da026f0a3f4` and its auto-merge result.

[P2] Make the startup-failure cleanup actually close partially started services.

The new `t.Cleanup(c.Close)` is registered before `c.Start()`, but it cannot provide the guarantee stated by this commit. `cluster.Close()` iterates services in reverse and returns on the first `operator.Close()` error. In the HAKeeper-timeout path being fixed, LOG has started, TN fails while waiting for HAKeeper, and CN has never started. Cleanup therefore reaches the stopped CN first, gets `service already stopped`, returns immediately, and leaves the started LOG plus the partially initialized TN stopper/goroutines alive. It also adds a second cleanup assertion failure that obscures the original timeout and can contaminate `TestPreparedParams` in the same test binary.

Please make embedded-cluster startup roll back/close every service that was started or partially initialized, continuing past never-started services, and cover the LOG-started/TN-timeout/CN-not-started sequence with a focused failure-path test. Merely ignoring the `Close` error would retain the leak. If startup rollback is intentionally out of scope, register cleanup only after a successful `Start` and remove the claim that startup failures are cleaned, but that leaves the cleanup requirement from #26299 unresolved.

The parser portion otherwise looks correct: the BVT/doc rewrites preserve MySQL default NULL ordering, the merged generated parser reproduces byte-for-byte from `mysql_sql.y`, and no live repository SQL still uses the rejected syntax. Verification on the current-main merge result passed: parser focused race x100, parser package race, txnexecutor focused race x2, txnexecutor package race, and embed package race.

heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed latest exact head `5fe9b097af9f5ae6ba532f85fc5e569b18837816`. The previous documentation, duplicate allocation-test, and unrelated HAKeeper changes are all resolved/reverted. The grammar now rejects NULLS FIRST/LAST consistently in top-level and window ORDER BY while preserving valid MySQL ordering; BVT rewrites are semantically equivalent for ASC/DESC NULL placement, and no live user-facing SQL still advertises the rejected syntax. Generated mysql_sql.go matches mysql_sql.y, focused parser race passed for 20 iterations, and latest-main merge is clean. No production-path performance change. No blocker found.

Issue #26242, opened 2026-07-27
[Bug]: TestCkpLeak waits for global checkpoint quiescence after WAL burst


LeftHandCold:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main / PR #26096 branch `fix/issue-25996-prepared-set`

### Commit ID

f154b461490c26d1465f2ae022079415ba92a59d

### Other Environment Information

```Markdown
- Hardware parameters: 8-core / 16 GiB GitHub runner
- OS type: Ubuntu / x86
- Others: Matrixone CI unit-test job
```

### Actual Behavior

`TestCkpLeak` failed after its 10-second checkpoint wait in [run 30251258817/job 89929582475](https://github.com/matrixorigin/matrixone/actions/runs/30251258817/job/89929582475?pr=26096):

```text
target LSN: 201, checkpointed LSN: 1, current LSN: 203, pending LSN count: 202
```

The first incremental checkpoint completed at LSN 1. A second incremental intent was created with `flushChecked=false`, but no later ICKP executed before timeout. Normal table-tail flush tasks completed while that intent remained below the two-minute stalled-checkpoint recovery threshold.

This is independent of PR #26096: its diff changes only frontend, planner, and prepared-statement tests, with no TAE/checkpoint/WAL/catalog/logtail changes.

### Expected Behavior

`TestCkpLeak` should wait for a checkpoint covering the concurrent append workload it created, without requiring global background WAL/checkpoint quiescence.

### Steps to Reproduce

```Markdown
.agents/skills/mo-dev/scripts/mo-cgo-test -count=20 -run TestCkpLeak -timeout=180s ./pkg/vm/engine/tae/db/test
```

The same checkpoint code passed 20/20 locally, confirming a timing-dependent failure. The linked CI log captures the failing interleaving.

### Additional information

Root cause: #25953 replaced the old best-effort wait/early return in `TestCkpLeak` with `WaitAllCheckpointsFinished`. After a concurrent WAL burst this helper waits for global background quiescence, although the test only needs a checkpoint covering its own writes. A pending background ICKP can still be waiting for asynchronous table-tail flush work, while the production stalled-intent recovery path intentionally activates only after two minutes; the test asserts after ten seconds.

This is the same synchronization-pattern class fixed for other TAE tests by #26033 and #26099. Older `TestCkpLeak` issues #21171 and #21722 covered different assertions.

[...]

PR #26244, opened 2026-07-27
test(tae): scope TestCkpLeak checkpoint wait

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [ ] Performance
- [ ] Refactoring
- [ ] Dependency Upgrade
- [x] Test and CI

issue #26242

`TestCkpLeak` waited for global checkpoint quiescence after 200 concurrent appends. A new incremental checkpoint intent can stay pending after table-tail flush completes, so the 10-second helper can observe `target LSN: 201, checkpointed LSN: 1, current LSN: 203` even though the checkpoint worker is healthy. The production stalled-intent recovery threshold is two minutes, which does not fit this test contract.

Fence the completed workload at `db.TxnMgr.Now()` and use a bounded `ForceCheckpoint` to flush, schedule, and wait for a checkpoint covering that fixed timestamp. This follows the synchronization pattern already used by related TAE GC tests and changes test code only.

Validation:

- `mo-cgo-test -count=20 -run ^TestCkpLeak$ -timeout=180s ./pkg/vm/engine/tae/db/test`
- `mo-cgo-test -race -count=3 -run ^TestCkpLeak$ -timeout=240s ./pkg/vm/engine/tae/db/test`
- `mo-cgo-test -count=3 -run ^(TestCkpLeak|TestSnapshotGC|TestAppendAndGC2)$ -timeout=240s ./pkg/vm/engine/tae/db/test`
- `go vet -mod=readonly ./pkg/vm/engine/tae/db/test`
- `go build -mod=readonly ./pkg/vm/engine/tae/db/...`

### DIFF
--- a/pkg/vm/engine/tae/db/test/db_test.go
+++ b/pkg/vm/engine/tae/db/test/db_test.go
@@ -8848,9 +8848,9 @@ func TestCkpLeak(t *testing.T) {
 	}
 	wg.Wait()
 	ckpCtx, cancel := context.WithTimeout(ctx, testutil.TestCheckpointTimeout)
-	defer cancel()
-	require.NoError(t, db.ForceCheckpoint(ckpCtx, db.TxnMgr.Now()))
-	testutil.WaitAllCheckpointsFinished(t, db)
+	err = db.ForceCheckpoint(ckpCtx, db.TxnMgr.Now())
+	cancel()
+	require.NoError(t, err)
 	t.Log(tae.Catalog.SimplePPString(common.PPL1))
 	checkLeak := func() bool {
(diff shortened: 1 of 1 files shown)


XuPeng-SH (approved):
Deep-reviewed the rebased head. It now keeps testutil.TestCheckpointTimeout and removes only the global WaitAllCheckpointsFinished wait. ForceCheckpoint flushes and waits for an incremental checkpoint covering the fixed post-workload timestamp, so the restart/leak assertions retain their intent without chasing later background checkpoint activity. Focused TestCkpLeak passes locally; no production code is changed and no blocking issue was found.

Issue #26237, opened 2026-07-27
[Bug]: TestCheckpointTableIDBatch2 flakes on checkpoint timeout in CI


aptend:
### Is there an existing issue for the same bug?

No issue was found by searching for `TestCheckpointTableIDBatch2`, `ICKP context deadline exceeded`, and checkpoint table-ID batch failures.

### Branch and CI evidence

Observed in the Ubuntu/x86 UT job of PR #26231:

- Run: https://github.com/matrixorigin/matrixone/actions/runs/30246533106
- Job: https://github.com/matrixorigin/matrixone/actions/runs/30246533106/job/89914775889
- Runner: `amd64-mo-shanghai-8c16g-92mgx-runner-dw44g`
- Package: `pkg/vm/engine/tae/db/test`
- Test: `TestCheckpointTableIDBatch2`

The PR only changes `pkg/sql/colexec/hashbuild`; this TAE checkpoint test is unrelated to its diff. The same run's coverage UT job passed.

### Actual behavior

The forced incremental checkpoint did not finish within its fixed deadline:

```text
ICKP-Schedule-Force-End ... cost=3.998993037s error="context deadline exceeded"
ICKP-Control-Force-End ... total-cost=5.00062386s error="context deadline exceeded"
engine.go:166: Received unexpected error: context deadline exceeded
```

The follow-up table-ID assertion then observed incomplete progress:

```text
expected: 100006
actual:   100001
```

The test failed after 7.44 seconds.

### Expected behavior

`TestCheckpointTableIDBatch2` should deterministically wait for the checkpoint/table-ID synchronization it triggers, or fail with evidence of a real checkpoint defect rather than depending on a runner-sensitive fixed timeout.

### Suggested investigation

1. Re-run this test repeatedly under the 8c/16GiB CI runner profile.
2. Inspect why `sync-table-id` consumes the complete checkpoint deadline under runner load.
3. Replace timing-only completion assumptions with an explicit completion condition where possible; otherwise calibrate the test timeout and retain diagnostics for the blocked phase.

### Additional information

This report tracks the flaky TAE UT separately so unrelated PRs are not blocked without an owner or evidence.

PR #26245, opened 2026-07-27
test(tae): relax forced checkpoint timeout

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26237

`TestEngine.ForceCheckpoint` previously imposed a fixed 5-second deadline across both force flush and incremental checkpoint execution. Under CI runner load, `TestCheckpointTableIDBatch2` can spend part of that budget flushing and then time out while synchronizing its 100,000-row table-ID batch.

This PR raises the test helper deadline to two minutes, matching the checkpoint runner safety net, while retaining an explicit caller deadline because `ForceFlush` relies on the caller context for cancellation.


XuPeng-SH (approved):
Reviewed the complete one-file diff and the failure evidence in #26237. The explicit caller deadline remains necessary because ForceFlush consumes the caller context, and two minutes aligns with ForceICKP's existing default safety net while preserving a hard bound. The target test waits on actual checkpoint intent completion rather than a timing sleep. Local validation passed: build, vet, 10 repeated target runs, and the target race test. No blocking findings.

PR #26248, opened 2026-07-27
add date extract

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24504

• 此分支主要完成：

  - 为 DAYOFMONTH、DAYNAME、MONTHNAME、QUARTER、WEEKOFYEAR 增加 VARCHAR 参数支持。
  - 新增统一的字符串日期解析执行器，非法日期返回 NULL，并正确处理 NULL、零日期及选择掩码。
  - 修复 FunctionSelectList.AllNull 路径，避免被跳过的行仍被解析并产生非 NULL 结果。
  - WEEKOFYEAR 统一使用 WEEK(date, 3) 的 ISO 周语义，修复 2005-01-01 应返回 53 等跨年边界问题。
  - 补充函数解析、执行器、AllNull 和跨年边界 UT。
  - 为五个函数补充 BVT，结果已与本地 MatrixOne 和 MySQL 行为核对。


XuPeng-SH (changes requested):
Blocking: the new VARCHAR DAYOFMONTH path loses MySQL incomplete-date semantics.

`dateStringToFixedWithNullOnError` calls `types.ParseDateCast` before the function-specific extractor. `ParseDateCast` accepts the exact all-zero special case, but otherwise delegates to `ValidDate`, which requires `day > 0`. As a result, `DAYOFMONTH('2001-11-00')` returns NULL on this branch. MySQL explicitly defines incomplete-date extractors to work and documents this exact expression as returning 0: https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html

This is also why the current tests miss it: they cover `0000-00-00` and a normal VARCHAR date, but no partially-zero date. Please preserve the raw date components (or otherwise use function-specific parsing) so DAYOFMONTH can return 0 while strict/name functions such as WEEKOFYEAR and DAYNAME still return NULL, then add resolver/executor and BVT coverage for `DAYOFMONTH('2001-11-00')` and audit the other newly added extractors against the same documented incomplete-date matrix.

I verified the rest of the changed execution paths: every added/modified focused test passes under `-race -count=100`; the full `pkg/sql/plan/function` package passes under race; build and vet pass; current-main merge-tree is clean.

heni02 (approved):

daviszhen:
> Blocking: the new VARCHAR DAYOFMONTH path loses MySQL incomplete-date semantics.
> 
> `dateStringToFixedWithNullOnError` calls `types.ParseDateCast` before the function-specific extractor. `ParseDateCast` accepts the exact all-zero special case, but otherwise delegates to `ValidDate`, which requires `day > 0`. As a result, `DAYOFMONTH('2001-11-00')` returns NULL on this branch. MySQL explicitly defines incomplete-date extractors to work and documents this exact expression as returning 0: https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html
> 
> This is also why the current tests miss it: they cover `0000-00-00` and a normal VARCHAR date, but no partially-zero date. Please preserve the raw date components (or otherwise use function-specific parsing) so DAYOFMONTH can return 0 while strict/name functions such as WEEKOFYEAR and DAYNAME still return NULL, then add resolver/executor and BVT coverage for `DAYOFMONTH('2001-11-00')` and audit the other newly added extractors against the same documented incomplete-date matrix.
> 
> I verified the rest of the changed execution paths: every added/modified focused test passes under `-race -count=100`; the full `pkg/sql/plan/function` package passes under race; build and vet pass; current-main merge-tree is clean.

  - 新增不完整日期的解析结构，保留原始 year/month/day，避免先转为 types.Date 时丢失零日期分量。
  - DAYOFMONTH('2001-11-00') 现在返回 0。
  - 按 MySQL 语义处理其他函数：QUARTER -> 4、MONTHNAME -> November，WEEKOFYEAR 和 DAYNAME -> NULL。
  - 增加对应的函数解析、执行单测和五组 BVT case。
  - 同时修正 WEEKOFYEAR 的 ISO week 年边界结果。

XuPeng-SH (changes requested):
Re-reviewed exact head e0916fbd08db9a79741370aa6d3f9b68dc30d6de and prospectively merged it cleanly onto latest main ca4a91c5aa.

The exact previously reported 2001-11-00 case is now correct for all five functions, and the new focused test passes under race for 44 repetitions; the full pkg/sql/plan/function race suite, vet, and diff checks also pass. However, the fix only closes one spelling of the incomplete-date input.

[P1] Preserve the existing VARCHAR date format family for incomplete dates (pkg/sql/plan/function/func_unary.go:862-895).

parseDateExtractParts first delegates to types.ParseDateCast, whose documented parser accepts dashed variable-width dates, compact YYYYMMDD, and date strings with a time suffix. Once a zero component makes ParseDateCast fail ValidDate, the fallback accepts only len(value) == 10 with dashes at fixed offsets. Consequently these MySQL-equivalent forms all become NULL on this head:

- 2001-11-00 12:34:56
- 20011100
- 2001-11-0

I verified on MySQL 8.0.44 that each form returns DAYOFMONTH=0, QUARTER=4, MONTHNAME=November, with WEEKOFYEAR and DAYNAME NULL. A deterministic current-head unit probe also confirmed parseDateExtractParts returns ok=false for all three. The same issue affects zero-component datetime forms such as 0000-00-00 12:34:56.

Please parse incomplete raw components with the same accepted grammar/normalization as the complete-date path, rather than replacing it with a fixed-width side grammar, and extend the resolver/executor or parser table plus BVT to cover at least datetime-suffixed, compact, and variable-width incomplete inputs. Requesting changes because the new overloads otherwise change results solely based on an equivalent input representation.

daviszhen:
> Re-reviewed exact head [e0916fb](https://github.com/matrixorigin/matrixone/commit/e0916fbd08db9a79741370aa6d3f9b68dc30d6de) and prospectively merged it cleanly onto latest main [ca4a91c](https://github.com/matrixorigin/matrixone/commit/ca4a91c5aaf50ebeceb71bd798eab4863adcd91b).
> 
> The exact previously reported 2001-11-00 case is now correct for all five functions, and the new focused test passes under race for 44 repetitions; the full pkg/sql/plan/function race suite, vet, and diff checks also pass. However, the fix only closes one spelling of the incomplete-date input.
> 
> [P1] Preserve the existing VARCHAR date format family for incomplete dates (pkg/sql/plan/function/func_unary.go:862-895).
> 
> parseDateExtractParts first delegates to types.ParseDateCast, whose documented parser accepts dashed variable-width dates, compact YYYYMMDD, and date strings with a time suffix. Once a zero component makes ParseDateCast fail ValidDate, the fallback accepts only len(value) == 10 with dashes at fixed offsets. Consequently these MySQL-equivalent forms all become NULL on this head:
> 
> * 2001-11-00 12:34:56
> * 20011100
> * 2001-11-0
> 
> I verified on MySQL 8.0.44 that each form returns DAYOFMONTH=0, QUARTER=4, MONTHNAME=November, with WEEKOFYEAR and DAYNAME NULL. A deterministic current-head unit probe also confirmed parseDateExtractParts returns ok=false for all three. The same issue affects zero-component datetime forms such as 0000-00-00 12:34:56.
> 
> Please parse incomplete raw components with the same accepted grammar/normalization as the complete-date path, rather than replacing it with a fixed-width side grammar, and extend the resolver/executor or parser table plus BVT to cover at least datetime-suffixed, compact, and variable-width incomplete inputs. Requesting changes because the new overloads otherwise change results solely based on an equivalent input representation.

  - 将 ParseDateCast 的字符串语法解析抽取为 ParseDateCastComponents，统一支持 datetime 后缀、紧凑 YYYYMMDD 和变宽日期格式。                                                                                                                                                                      
  - 五个 VARCHAR 日期函数改为读取原始年月日分量，正确处理不完整日期：                                                                                                                                                                                                                            
      - DAYOFMONTH 返回 0                                                                                                                                                                                                                                                                        
      - QUARTER 返回 4                                                                                                                                                                                                                                                                           
      - MONTHNAME 返回 November                                                                                                                                                                                                                                                                  
      - DAYNAME、WEEKOFYEAR 返回 NULL                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                 
  - 普通 DATE cast 仍执行 ValidDate，原有严格校验行为不变。                                                                                                                                                                                                                                      
  - 增加 parser/function UT，并为五个函数补充 BVT 及对应结果，覆盖 reviewer 指出的三种格式和零分量 datetime。

XuPeng-SH (changes requested):
Re-reviewed exact head e8b7d62e81de85a58eed29b3b0df3e5d5712f4d9 against latest main 522a157c0afd8c8c9e5102997b234c68928a3d38. CI is green and the prospective merge is clean.

Requesting changes for two independent production issues.

[P1] Make the new VARCHAR temporal parser complete and range-safe instead of returning plausible values for malformed input.

The shared component parser currently drifts in both directions from the canonical datetime grammar:

- A normal ISO datetime such as `2024-01-01T12:34:56` is rejected, so all five new VARCHAR overloads return NULL. MySQL 8.4.10 accepts it, and MatrixOne already accepts this form in `types.ParseDatetime`.
- Out-of-range time suffixes such as `2024-01-01 24:00:00` and `2024-01-01 23:60:00` are accepted because the state machine counts time-field digits but never records or validates their values. All five functions then return normal results, whereas MySQL returns NULL.
- `4294967297-01-01` is accumulated as int64 and narrowed to int32 before validation, wrapping to year 1. The probe on this head produced a valid `0001-01-01`, so DAYNAME/WEEKOFYEAR and the component extractors return plausible but false values; MySQL returns NULL.
- The month state has no rejection branch for unexpected characters. For example, `2024-0x-01` is accepted as an incomplete date and produces DAYOFMONTH=1 and QUARTER=0, while MySQL returns NULL for every extractor.

These are one parser-contract problem, not isolated literals: accept the supported complete datetime grammar, validate separators and time ranges, and bounds-check raw components before narrowing. Please add category tests for ISO separators, malformed separators, time boundaries, and oversized components.

[P2] Remove the measurable regression added to the existing ParseDateCast hot path.

`ParseDateCast` now trims and calls `isZeroDatetimeString`, then `ParseDateCastComponents` repeats both operations. `isZeroDatetimeString` includes a full `IndexByte` scan, so every existing date cast and external-ingestion caller pays duplicated preprocessing even though the PR only needs raw components for the new overloads.

A same-host, zero-allocation benchmark against latest main, 500 ms x 5, gave these median times:

- dashed date: 23.70 -> 29.06 ns/op (+22.6%)
- compact date: 14.79 -> 17.92 ns/op (+21.2%)
- datetime with fraction: 53.53 -> 62.29 ns/op (+16.4%)

Please structure the shared parser around an internal already-trimmed path or return enough metadata for ParseDateCast to avoid redoing the scan. Keep a regression benchmark or otherwise verify parity for existing ParseDateCast callers.

Other verification on this exact head passed: full `pkg/container/types` and `pkg/sql/plan/function` suites under race; the new focused tests for 100 repetitions under race; vet and diff checks. I also exhaustively compared `Date.Week(3)` with Go ISOWeek over every valid date from year 1 through 9999, and that WEEKOFYEAR replacement is correct.

daviszhen:
> Re-reviewed exact head [e8b7d62](https://github.com/matrixorigin/matrixone/commit/e8b7d62e81de85a58eed29b3b0df3e5d5712f4d9) against latest main [522a157](https://github.com/matrixorigin/matrixone/commit/522a157c0afd8c8c9e5102997b234c68928a3d38). CI is green and the prospective merge is clean.
> 
> Requesting changes for two independent production issues.
> 
> [P1] Make the new VARCHAR temporal parser complete and range-safe instead of returning plausible values for malformed input.
> 
> The shared component parser currently drifts in both directions from the canonical datetime grammar:
> 
> * A normal ISO datetime such as `2024-01-01T12:34:56` is rejected, so all five new VARCHAR overloads return NULL. MySQL 8.4.10 accepts it, and MatrixOne already accepts this form in `types.ParseDatetime`.
> * Out-of-range time suffixes such as `2024-01-01 24:00:00` and `2024-01-01 23:60:00` are accepted because the state machine counts time-field digits but never records or validates their values. All five functions then return normal results, whereas MySQL returns NULL.
> * `4294967297-01-01` is accumulated as int64 and narrowed to int32 before validation, wrapping to year 1. The probe on this head produced a valid `0001-01-01`, so DAYNAME/WEEKOFYEAR and the component extractors return plausible but false values; MySQL returns NULL.
> * The month state has no rejection branch for unexpected characters. For example, `2024-0x-01` is accepted as an incomplete date and produces DAYOFMONTH=1 and QUARTER=0, while MySQL returns NULL for every extractor.
> 
> These are one parser-contract problem, not isolated literals: accept the supported complete datetime grammar, validate separators and time ranges, and bounds-check raw components before narrowing. Please add category tests for ISO separators, malformed separators, time boundaries, and oversized components.
> 
> [P2] Remove the measurable regression added to the existing ParseDateCast hot path.
> 
> `ParseDateCast` now trims and calls `isZeroDatetimeString`, then `ParseDateCastComponents` repeats both operations. `isZeroDatetimeString` includes a full `IndexByte` scan, so every existing date cast and external-ingestion caller pays duplicated preprocessing even though the PR only needs raw components for the new overloads.
> 
> A same-host, zero-allocation benchmark against latest main, 500 ms x 5, gave these median times:
> 
> * dashed date: 23.70 -> 29.06 ns/op (+22.6%)
> * compact date: 14.79 -> 17.92 ns/op (+21.2%)
> * datetime with fraction: 53.53 -> 62.29 ns/op (+16.4%)
> 
> Please structure the shared parser around an internal already-trimmed path or return enough metadata for ParseDateCast to avoid redoing the scan. Keep a regression benchmark or otherwise verify parity for existing ParseDateCast callers.
> 
> Other verification on this exact head passed: full `pkg/container/types` and `pkg/sql/plan/function` suites under race; the new focused tests for 100 repetitions under race; vet and diff checks. I also exhaustively compared `Date.Week(3)` with Go ISOWeek over every valid date from year 1 through 9999, and that WEEKOFYEAR replacement is correct.

  - 重构 VARCHAR 日期组件解析，支持 T 分隔的 ISO datetime，同时统一完整日期与零日期/不完整日期的校验路径。
  - 修复非法输入被误解析的问题：拒绝非法分隔符、无效时分秒、悬空 T、超大年份及越界月份/日期，相关五个日期提取函数均返回 NULL。
  - 保留并优化 ParseDateCast 快路径，避免重复 trim/扫描和重复解析，消除审查指出的热路径性能回退。
  - 补充类型层与函数层 UT，包括 ISO 格式、非法时间、非法分隔符、超大年份等场景，并增加 ParseDateCast benchmark。
  - 新增 BVT func_datetime_varchar_date_validation，覆盖 DAYOFMONTH、DAYNAME、MONTHNAME、QUARTER、WEEKOFYEAR 对合法 ISO 输入和各类非法输入的行为。

  验证已通过：git diff --check、相关 UT 的 race 重复执行，以及 pkg/container/types、pkg/sql/plan/function 的 race 测试。

XuPeng-SH (changes requested):
Re-reviewed exact head `858ea17e1c12b780adc1677d952c0b31b2d5a24e`. The previous parser-safety and ParseDateCast performance blockers are fixed: ISO-T, range checks, oversized years, and malformed separators now follow the intended paths, and same-host 0-allocation benchmarks show the common dashed/compact/datetime-with-fraction paths are about 46%/24%/55% faster than current main.

Two independent MySQL-compatibility issues remain:

1. [P1] Implement MySQL year-0 calendar semantics instead of treating every year-0 value as an incomplete date and validating it as year 2000. On MySQL 8.4.10, `0000-01-01` is a valid date: DAYOFMONTH/DAYNAME/MONTHNAME/QUARTER/WEEKOFYEAR return `1 / Sunday / January / 1 / 52`. This head returns NULL for DAYNAME and WEEKOFYEAR because `types.ValidDate` rejects year 0. Conversely, MySQL rejects `0000-02-29` for all five functions because year 0 is not a leap year, while `validationYear = 2000` makes this head return `29 / February / 1` from the component extractors. MatrixOne already has the MySQL-specific year-0 rule in `calcDaysInYear`; please validate and calculate this category consistently and add both valid and invalid year-0 regressions.

2. [P2] Accept the nonstandard date delimiters that MySQL still accepts. MySQL 8.4.10 returns `15 / Monday / January / 1 / 3` for both `2024/01/15` and `2024:01:15` (with deprecation warnings). The shared state machine accepts only `-`, so all five new overloads return NULL. This is a real migration compatibility break for existing MySQL data even though those delimiters are deprecated. Please handle the delimiter grammar as a category rather than adding literal-specific exceptions, and cover date plus datetime forms.

Both categories were reproduced against MySQL 8.4.10 and with direct current-head parser probes. The PR-owned focused tests pass, confirming these are coverage gaps rather than environment failures.

daviszhen:
> Re-reviewed exact head `858ea17e1c12b780adc1677d952c0b31b2d5a24e`. The previous parser-safety and ParseDateCast performance blockers are fixed: ISO-T, range checks, oversized years, and malformed separators now follow the intended paths, and same-host 0-allocation benchmarks show the common dashed/compact/datetime-with-fraction paths are about 46%/24%/55% faster than current main.
> 
> Two independent MySQL-compatibility issues remain:
> 
> 1. [P1] Implement MySQL year-0 calendar semantics instead of treating every year-0 value as an incomplete date and validating it as year 2000. On MySQL 8.4.10, `0000-01-01` is a valid date: DAYOFMONTH/DAYNAME/MONTHNAME/QUARTER/WEEKOFYEAR return `1 / Sunday / January / 1 / 52`. This head returns NULL for DAYNAME and WEEKOFYEAR because `types.ValidDate` rejects year 0. Conversely, MySQL rejects `0000-02-29` for all five functions because year 0 is not a leap year, while `validationYear = 2000` makes this head return `29 / February / 1` from the component extractors. MatrixOne already has the MySQL-specific year-0 rule in `calcDaysInYear`; please validate and calculate this category consistently and add both valid and invalid year-0 regressions.
> 2. [P2] Accept the nonstandard date delimiters that MySQL still accepts. MySQL 8.4.10 returns `15 / Monday / January / 1 / 3` for both `2024/01/15` and `2024:01:15` (with deprecation warnings). The shared state machine accepts only `-`, so all five new overloads return NULL. This is a real migration compatibility break for existing MySQL data even though those delimiters are deprecated. Please handle the delimiter grammar as a category rather than adding literal-specific exceptions, and cover date plus datetime forms.
> 
> Both categories were reproduced against MySQL 8.4.10 and with direct current-head parser probes. The PR-owned focused tests pass, confirming these are coverage gaps rather than environment failures.

  - 新增 year-0 日历校验：允许有效的 0000-01-01，并按 MySQL 规则将 year 0 视为非闰年，拒绝 0000-02-29。
  - 为 DAYNAME、WEEKOFYEAR 增加基于原始年月日的 year-0 计算，避免错误依赖不支持 year 0 的内部 Date 表示。
  - 日期字符串解析改为支持 MySQL 兼容的标点分隔符，包括 /、:、. 等，以及相应 datetime 输入。
  - 保留普通 DATE 类型的年份下界，year 0 仅用于 VARCHAR 日期提取函数兼容路径。
  - 补充类型层和函数层 UT，覆盖有效/无效 year 0、日期与 datetime 分隔符形式。
  - 扩展 func_datetime_varchar_date_validation BVT 及结果文件，覆盖五个新增日期提取函数。

XuPeng-SH (changes requested):
P1: relaxed temporal parsing is still only partially compatible and now codifies a valid MySQL date-context value as invalid. In parseDateCastComponents, the `(s[i] != : || yearLen == 4)` gate makes `10:11:12` fail, and the new unit test explicitly expects that error. MySQL date-context rules interpret it as `2010-11-12`; delimited `YY-MM-DD`/`YY:MM:DD` values use the documented 00-69 => 2000-2069 and 70-99 => 1970-1999 conversion. The same root issue exists in the datetime suffix: `2024/01/15 12*34*56` is valid relaxed MySQL syntax, but this parser accepts only `:` between time fields. These inputs therefore return NULL from DAYNAME/WEEKOFYEAR/etc. even though MySQL returns a value. Please implement the delimiter category coherently rather than special-casing four-digit-year colon: normalize two-digit years at the shared parser boundary, accept punctuation delimiters between time fields while retaining range/full-consumption validation, and add boundary/oracle tests (for example `10:11:12`, years 69/70, and a datetime with `*` time delimiters). The year-0 calendar fix and standard hot paths otherwise look sound; focused tests passed and ParseDateCast remains 0-allocation.

daviszhen:
> P1: relaxed temporal parsing is still only partially compatible and now codifies a valid MySQL date-context value as invalid. In parseDateCastComponents, the `(s[i] != : || yearLen == 4)` gate makes `10:11:12` fail, and the new unit test explicitly expects that error. MySQL date-context rules interpret it as `2010-11-12`; delimited `YY-MM-DD`/`YY:MM:DD` values use the documented 00-69 => 2000-2069 and 70-99 => 1970-1999 conversion. The same root issue exists in the datetime suffix: `2024/01/15 12*34*56` is valid relaxed MySQL syntax, but this parser accepts only `:` between time fields. These inputs therefore return NULL from DAYNAME/WEEKOFYEAR/etc. even though MySQL returns a value. Please implement the delimiter category coherently rather than special-casing four-digit-year colon: normalize two-digit years at the shared parser boundary, accept punctuation delimiters between time fields while retaining range/full-consumption validation, and add boundary/oracle tests (for example `10:11:12`, years 69/70, and a datetime with `*` time delimiters). The year-0 calendar fix and standard hot paths otherwise look sound; focused tests passed and ParseDateCast remains 0-allocation.

  - 移除对 : 两位年份的错误限制，支持 10:11:12。
  - 在共享解析器中加入 MySQL 两位年份映射：
      - 00-69 -> 2000-2069
      - 70-99 -> 1970-1999

  - 时分秒分隔符从仅支持 : 扩展为统一支持标点符号，例如 12*34*56。
  - 保留原有时间范围、日期范围和完整输入校验。
  - 增加 YY-MM-DD、YY:MM:DD、69/70 边界以及带 * 时间分隔符的单测。
  - 在 func_datetime_varchar_date_validation 中补充对应 BVT 和结果文件。

XuPeng-SH (changes requested):
The two-digit-year fix is still incomplete for compact MySQL date forms. `parseDateCastComponents` rejects every all-digit input shorter than 7 bytes before applying `normalizeDateCastYear`, so the new VARCHAR overloads return NULL for `101112`, `691231`, `700101`, and incomplete `101100`.

MySQL 8.4 interprets these as `2010-11-12`, `2069-12-31`, `1970-01-01`, and `2010-11-00`, respectively. The five functions added by this PR consequently return the corresponding day/month/quarter values (and NULL only for the calendar-dependent functions on the incomplete date). My focused test against this head confirmed `ParseDateCastComponents` returns `invalid argument parsedate` for all four inputs.

Please extend the two-digit-year normalization to compact `YYMMDD` forms as well, including the 69/70 boundary and incomplete-date cases, and add unit/BVT coverage. MySQL also accepts compact `YYMMDDHHMMSS` (for example `101112123456`) as the same date, so that form should be covered if the intent is to preserve the accepted VARCHAR temporal grammar.

XuPeng-SH (approved):
Re-reviewed the current head in the context of the full discussion and the author's latest response. The response implements the previously requested categories: two-digit delimited years (including the 69/70 window), relaxed punctuation in time fields, year-0 calendar handling, parser range/full-consumption checks, and focused UT/BVT coverage. Full `pkg/container/types` and `pkg/sql/plan/function` tests pass locally, as do the relevant focused tests under `-race -count=10`.

I am treating compact `YYMMDD` / `YYMMDDHHMMSS` parsing as a non-blocking follow-up rather than expanding this PR beyond issue #24504 after the author addressed the requested cases. MySQL accepts those forms, while the shared parser currently rejects them, so retaining that compatibility gap as a follow-up would still be useful.

Issue #25114, opened 2026-06-24
[Bug] S3 LOAD DATA 10M wiki vectors causes CN OOMKilled (41GB) during IVF pre/post filter test


Ariznawlll:
## Summary

IVF PRE POST FILTER VECTOR RECALL TEST failed during data import phase: S3 LOAD DATA of 10M wiki vectors caused CN pod to be OOM-killed at ~41GB memory.

## Reproduction

**Run**: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/28037252650/job/83066708254
**Namespace**: `mo-branch-commit-d3355da06-20260623`
**Branch**: `commit-d3355da06` (4.0-dev)

## Error

```
21:32:15  [run_wiki] 2/3  import (S3 LOAD DATA)
          LOAD DATA S3: bucket=mo-load-guangzhou-1308875761, filepath=vector/wiki_all_10m/historical_file_blocks_wiki_all_10m.csv
21:33:32  pymysql.err.OperationalError: (2013, 'Lost connection to MySQL server during query')
```

## Root Cause Analysis

CN pod `nightly-regression-dis-tp-cn-sjrhk` memory grew steadily during LOAD DATA and was OOM-killed:

```
Memory timeline (CN tp-cn-sjrhk):
39.05 → 39.36 → 39.75 → 40.25 → 40.71 → 41.13 GB  (steady increase)
→ 0 GB     (OOM killed at 21:33:33)
→ 0.44 GB  (container restart)
```

After restart, the CN logged massive `dedup load did not complete` errors and `cannot get salt` connection failures.

## Cluster Memory Pressure at Time of Failure

| Pod | Memory |
|---|---|
| tp-cn-sjrhk | **41 GB** (OOM killed) |
| tp-cn-qlsg4 | 34-37 GB |
| tp-cn-whsbf | 31-36 GB |
| dn-0 | 27-32 GB (OOM alerts with available-mem ~300MB) |

Total cluster memory: **~140 GB**

## DN OOM Alerts

DN also reported memory pressure during the same window:
```
MergeExecutorEvent-PauseDueToOOMAlert, available-mem: 297MB ~ 470MB
```

## Expected Behavior

S3 LOAD DATA of 10M rows should complete without CN OOM. Memory should be bounded during bulk import.

## Environment

- MO version: 4.0-dev (commit d3355da06)
- Cluster: 3 CN + 1 DN on TKE
[...]

PR #26252, opened 2026-07-27
fix: bound memory during parallel vector load

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25114 #26272

fix: bound memory during parallel vector load

### DIFF
--- a/pkg/fileservice/local_fs.go
+++ b/pkg/fileservice/local_fs.go
@@ -380,9 +380,14 @@ func (l *LocalFS) write(ctx context.Context, vector IOVector) (bytesWritten int,
 		size = int64(last.Offset + last.Size)
 	}
 
+	nativePath, tempDir, err := l.writePaths(nativePath)
+	if err != nil {
+		return 0, err
+	}
+
 	// write
 	f, err := os.CreateTemp(
-		l.rootPath,
+		tempDir,
 		".tmp.*",
 	)
 	if err != nil {
@@ -1120,8 +1125,13 @@ func (l *LocalFS) NewWriter(ctx context.Context, filePath string) (io.WriteClose
 		return nil, moerr.NewFileAlreadyExistsNoCtx(path.File)
 	}
 
+	nativePath, tempDir, err := l.writePaths(nativePath)
+	if err != nil {
+		return nil, err
+	}
+
 	f, err := os.CreateTemp(
-		l.rootPath,
+		tempDir,
 		".tmp.*",
 	)
 	if err != nil {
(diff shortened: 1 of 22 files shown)


jiangxinmeng1:
The Ubuntu/x86 UT failure is not introduced by this PR. All five failing packages share a pre-existing LocalFS cross-filesystem rename failure (`invalid cross-device link`) triggered by the CI `/tmp` → `/mnt` workspace layout. Tracking the root cause in #26272: https://github.com/matrixorigin/matrixone/issues/26272

jiangxinmeng1:
Fix for #26272 has been pushed in `8f585a4539`: LocalFS temp files are now created on the target filesystem when the root path is empty, preventing `EXDEV` in the Ubuntu UT workspace. Validation: `go build ./pkg/fileservice`, `go vet ./pkg/fileservice`, all `TestLocalFS*`, and all downstream `TestWriteTransferPage_*` passed.

XuPeng-SH (approved):
Reviewed exact head 70f9e1d766beeb1d7d0fd2264891ed94d8112409 against 4.2-dev@46dfc97cf284c51ce12095aa75618b1ef0410453. No blocking findings.

I traced the full memory-bound closure: parallel vector CSV decoding now avoids the varchar-plus-binary overlap; LOAD write parallelism and merge receiver buffering are capped; connector backpressure occurs before upstream batch production and terminates on context/edge completion; hashbuild transfers retained batches into spill ownership one at a time, cleans the current/remaining batches on error, and bounds per-bucket rows by observed row width. Reset and deferred cleanup close the reuse/error paths.

The compile delta is intentionally adapted to the older 4.2 join architecture rather than being patch-identical to main. SINK_SCAN-dependent shuffle receivers stay coordinator-local, while remote dispatch sources are attached to an executable scope tree; ordinary joins retain the existing 4.2 path. Late irregular-index steps rerun shuffle/runtime-filter planning after remap, with the existing RuntimeFilterBuildList guard preventing duplicate filters. The final unused-helper removal is consistent with that adaptation.

I also reviewed the added #26272 fix. With an empty LocalFS root, Write/NewWriter now create temp files beside the absolute destination, preserving same-filesystem atomic rename; non-empty-root semantics are unchanged. Creation, error cleanup, close/rename ownership, directory sync, and fd-cache boundedness were checked. I applied the exact 8f585a4 patch in an isolated worktree: the cross-filesystem regression passed under -race x10; full pkg/fileservice -race, vet, and build passed. GitHub CI is green on this head.

PR #26258, opened 2026-07-27
fix(frontend): deallocate stale statement before re-PREPARE

### LINKED ISSUE #25977: [Bug]: Failed re-PREPARE keeps the previous statement instead of deallocating it
### Summary

When a prepared statement name is reused and the new `PREPARE` fails, MatrixOne keeps the previous statement under that name. A later `EXECUTE` therefore runs stale SQL.

MySQL deallocates the previous statement before attempting the new prepare; after the new prepare fails, no statement remains under that name.

### Tested version

### Reproduction

Run the following in one session:

```sql
USE join_main_audit;
PREPARE s FROM 'SELECT 1';

SET @bad = 'SELECT * FROM prepare_missing_table';
PREPARE s FROM @bad;

EXECUTE s;
```

### Actual result

The second `PREPARE` correctly reports that `prepare_missing_table` does not exist, but `EXECUTE s` still runs the old statement and returns:

```text
1
```

This reproduces 3/3 times.

### Expected result

The previous statement should be deallocated before the replacement is prepared. Since the replacement fails, `EXECUTE s` should report that the prepared statement does not exist.

MySQL 8.0.45 returns:

```text
ERROR 1243 (HY000): Unknown prepared statement handler (s) given to EXECUTE
```

### Root cause

The frontend builds the replacement object before calling `Session.SetPrepareStmt`. The old object is only closed inside `SetPrepareStmt`, so any parse, authorization, or planning error before that call leaves the old statement in `ses.prepareStmts`.

This also creates a stale-statement risk: a client may assume the failed replacement invalidated the old statement, while MatrixOne continues to execute the previous SQL.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25977

MySQL deallocates an existing prepared statement before attempting to prepare
another statement with the same name. MatrixOne previously kept the old
statement until the replacement was successfully built, so a parse,
authorization, or planning failure left stale SQL executable under that name.

This change removes and closes the existing statement at the frontend PREPARE
dispatch boundary before building the replacement. It applies consistently to
direct statements, string literals, and user-variable PREPARE forms. If the new
PREPARE fails, a later EXECUTE now reports that the prepared statement does not
exist.

Regression coverage verifies planning failures, inner SQL parse failures,
missing user variables, resource cleanup, and successful reuse after a failed
replacement.

Validation:

- `go vet ./pkg/frontend/...`
- `make build`
- full `pkg/frontend` unit tests
- focused race test for failed PREPARE replacement
- `prepare_reprepare.test`: 15/15 passed


heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed exact head 98d1d8b55b43a1ef3fe280f114d13bd6c7d34978.

The replacement boundary is correct: once the outer PREPARE parses to a named AST, the old map owner is closed before transaction admission, planning, and authorization, so every failed replacement path leaves no stale executable statement. Case-normalized Set/Get/Remove closes the mixed-case gap and preserves quota accounting. I also checked the interaction with latest main's prepared-AST ownership changes; the removed object and the incoming wrapper own distinct ASTs, so there is no double-free edge.

Verified against MySQL 8.0.44 for inner parse failure, missing user variable, mixed-case replacement, and outer PREPARE syntax failure (the latter correctly retains the old statement because no replacement AST exists). Prospectively merged cleanly onto latest main 2dd3f83d9e330852daa789253178463f3173c202. Focused tests passed normally and under -race for 33 iterations; full pkg/frontend -race, pkg/common/malloc -race, go vet, and diff checks passed (the only diff-check whitespace findings are pre-existing latest-main enum result formatting). CI is green.

PR #26259, opened 2026-07-27
fix(frontend): reject unknown prepared deallocation

### LINKED ISSUE #25978: [Bug]: DEALLOCATE PREPARE silently succeeds for an unknown statement
### Summary

`DEALLOCATE PREPARE` silently succeeds when the named prepared statement does not exist or has already been deallocated.

### Tested version

### Reproduction

```sql
DEALLOCATE PREPARE does_not_exist;
```

Also:

```sql
PREPARE s FROM 'SELECT 1';
DEALLOCATE PREPARE s;
DEALLOCATE PREPARE s;
```

### Actual result

Both statements complete successfully without an error. The missing-name case reproduces 3/3 times.

### Expected result

Return an unknown prepared statement error. MySQL 8.0.45 returns:

```text
ERROR 1243 (HY000): Unknown prepared statement handler (...) given to DEALLOCATE PREPARE
```

### Root cause

`doDeallocate` builds a deallocate plan and then calls `Session.RemovePrepareStmt`. `RemovePrepareStmt` checks whether an entry exists only to close it, but unconditionally deletes the map key and returns no status. As a result, `doDeallocate` cannot distinguish a successful removal from a missing statement.

The binary `COM_STMT_CLOSE` path already performs an explicit `GetPrepareStmt` check before deallocation, so textual and binary prepared-statement cleanup currently have different missing-name behavior.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25978

Text-protocol `DEALLOCATE PREPARE` previously succeeded silently when the named
prepared statement did not exist or had already been deallocated. The session
removal API unconditionally deleted the map key and did not report whether an
entry was present.

This change:

- makes prepared-statement removal atomically report whether the statement
  existed;
- returns MySQL-compatible error 1243 (`HY000`) for an unknown statement
  handler;
- keeps internal cleanup paths idempotent when they intentionally ignore the
  removal result;
- adds unit coverage and a text-protocol BVT for missing and repeated
  deallocation.

Validation:

- `go test -count=1 -timeout=120s ./pkg/common/moerr/...`
- `go build ./pkg/common/moerr/... ./pkg/frontend/...`
- `go vet ./pkg/common/moerr/... ./pkg/frontend/...`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=240s -run 'TestRemovePrepareStmt|Test_HandleDeallocate' ./pkg/frontend`
- `golangci-lint run -c .golangci.yml ./pkg/common/moerr ./pkg/frontend`
- `mo-tester` BVT: `deallocate_unknown.sql` (6/6)

The repository-wide `make static-check` reaches two existing `unused` findings
in `pkg/common/malloc/sane_allocator_cache.go`; the same findings reproduce on
the clean base commit.


heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed exact head `c935282ec65b18d27def1a84f85afd86e4a7f656`. The text-protocol path now atomically distinguishes present vs. missing prepared statements and returns MySQL 1243/HY000, while internal cleanup remains idempotent by intentionally ignoring the boolean. I separately traced `COM_STMT_CLOSE`: its lookup/rewrite and ROW_COUNT handling remain on the protocol path, so this change does not conflate SQL `DEALLOCATE PREPARE` with binary close semantics. The `create_role` BVT adjustment is also correct: cleanup was previously issued after switching out of the session that owned the statements; it is now performed in the owning session, and `stmt4` is correctly omitted because its prepare failed authorization.

Exact-head validation:
- `TestNew_MyErrorCode`, `TestRemovePrepareStmt`, `Test_HandleDeallocate`, and `Test_ExecRequestPrepareCommandMissingStmt` each passed independently with `-race -count=100`
- full `pkg/common/moerr/...` and `pkg/frontend` passed under race
- exact-head standalone text-protocol regression passed 3 consecutive runs: two 1243/HY000 errors per run and subsequent statements continued normally
- CI logs confirm the modified `result_count.sql` and `create_role.sql` were actually executed successfully
- CI Summary/UT/SCA/BVT/Iceberg/coverage passed; current-main merge-tree is conflict-free

No blocking findings.

Issue #26249, opened 2026-07-27
[Bug]: proxy: hard-coded 3s MigrateConnTo deadline rejects valid session transfers


LeftHandCold:
## Branch and commit

- Discovery branch: `main`
- Discovery commit: `3009816821b5db6a734e5aec378e8dcd77a61ec6`

## Code location

- [`pkg/proxy/conn_migration.go:121`](https://github.com/matrixorigin/matrixone/blob/3009816821b5db6a734e5aec378e8dcd77a61ec6/pkg/proxy/conn_migration.go#L121)
- Symbol: `(*clientConn).migrateConnToContext`

## Actual behavior

A session transfer has an outer, end-to-end 10-second deadline, but the final `MigrateConnTo` RPC is independently capped at 3 seconds. The target CN performs database restoration and sequentially replays every prepared statement under this RPC context. A valid migration that needs between 3 and 10 seconds therefore fails even though the transfer's owning context is still valid.

The failure is externally visible as:

```text
context deadline exceeded
internal error: migrateConnTo
```

The Proxy closes the candidate backend connection, excludes that CN for the current routing attempt, and retries another CN. If every available CN needs more than 3 seconds for cold replay, the client connection cannot be migrated.

## Expected behavior

The `MigrateConnTo` step should honor the already bounded end-to-end transfer context. It must remain cancellable and bounded, while avoiding a shorter nested deadline that rejects otherwise valid transfers.

## Trigger path

1. A CN drain, crash, or chaos kill causes Proxy to migrate an existing session.
2. The session contains enough prepared statements, or the target CN is cold and prepared-statement planning triggers table subscription/checkpoint/FileService reads.
3. Target-side replay exceeds the nested 3-second RPC deadline but remains within the outer 10-second transfer budget.
4. `MigrateConnTo` returns `context deadline exceeded`; Proxy closes the target connection and retries/fails migration.

### Confirmed code facts

- [`tunnel.transfer`](https://github.com/matrixorigin/matrixone/blob/3009816821b5db6a734e5aec378e8dcd77a61ec6/pkg/proxy/tunnel.go#L742) bounds the complete transfer by `defaultTransferTimeout` (10 seconds).
- [`BuildConnWithServer`](https://github.com/matrixorigin/matrixone/blob/3009816821b5db6a734e5aec378e8dcd77a61ec6/pkg/proxy/client_conn.go#L465) also preserves a 10-second bound for transfer construction.
[...]

PR #26261, opened 2026-07-27
fix(proxy): honor transfer deadline for connection migration

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26249

The end-to-end Proxy connection transfer is already bounded by
`defaultTransferTimeout` (10 seconds), but the final `MigrateConnTo` RPC
derived a second hard-coded 3-second deadline. MORPC propagated that shorter
deadline to the target CN, so database restoration and sequential
prepared-statement replay could be canceled while the owning transfer context
was still valid.

Use `defaultTransferTimeout` for the child RPC context. An earlier parent
deadline or cancellation still wins, while deadline-free internal callers
remain bounded and compatible with MORPC's deadline requirement.
`MigrateConnFrom` keeps its existing timeout.

The regression test uses a real QueryService/MORPC handler and verifies that
the target-side deadline is not earlier than the transfer deadline. It also
covers Proxy-side parent cancellation while the target handler is blocked.

Tested with:

`TMPDIR=/tmp .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/proxy`

`TMPDIR=/tmp .agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=300s ./pkg/proxy`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=20 -timeout=240s -run 'TestMigrateConnTo(UsesTransferDeadline|PropagatesCancellation)$' ./pkg/proxy`

`TMPDIR=/tmp .agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=180s ./pkg/queryservice/... ./pkg/common/morpc/...`


XuPeng-SH (approved):
Deep-reviewed exact head `8370a13da4`.

Verified parent deadline precedence, deadline-free caller bounding, MORPC deadline propagation to the target handler, cancellation behavior, error-cause semantics, and transfer resource bounds. The change does not extend the existing 10-second end-to-end transfer budget and does not affect the normal query path.

Local full `pkg/proxy` tests pass; the two new MORPC regression tests also pass under `-race -count=20`. Exact-head CI/BVT/SCA/coverage are green. No blocking production issue found.

PR #26263, opened 2026-07-27
fix(tae): align stalled checkpoint flush recovery

### LINKED ISSUE #26260: [Bug]: TestAppend2 can strand checkpoint progress after concurrent append

### Branch Name

main (observed in PR #25312 CI; the relevant TAE test/checkpoint code is unchanged on main)

### Commit ID

115abf7933

- Hardware parameters: GitHub Actions Ubuntu/x86 runner
- OS type: Ubuntu/x86
- Others: `go test -race`; MatrixOne CI / UT Test on Ubuntu/x86

### Actual Behavior

`pkg/vm/engine/tae/db/test/TestAppend2` intermittently fails after its 300 concurrent rows have been appended and verified. The WAL checkpoint wait reaches its existing 10-second deadline with checkpoint progress stranded near the initial checkpoint:

```text
target LSN: 101, checkpointed LSN: 1, current LSN: 102, pending LSN count: 101
```

The failing trace shows a background flush merging 290 of the 300 appended rows, followed by repeated `flusher.table.tail.empty.fire` events and no subsequent incremental-checkpoint execution. A previous run of the same PR failed in the same state (`target=102`, `checkpointed=1`, `pending=102`), while later runs passed quickly, indicating a timing-dependent flush/checkpoint scheduling gap.

### Expected Behavior

After all concurrent append transactions have completed, background flush/checkpoint processing should deterministically make progress through the captured WAL target. The test should retain its concurrent-append, full-row verification, checkpoint-completion, and duplicate-append failure coverage without increasing its timeout or runtime.

### Steps to Reproduce

1. Run `TestAppend2` with the race detector as part of the concurrent Ubuntu/x86 UT workload.
2. Allow the 30 append tasks (300 total rows) to commit concurrently.
3. Observe the background flush/checkpoint trace after the row-scan verification.
4. Intermittently, one flush processes only part of the append tail and later scans report an empty tail, while the WAL checkpoint remains at LSN 1 until the existing 10-second assertion fails.

### Additional information

Failing runs:

- https://github.com/matrixorigin/matrixone/actions/runs/30277581556/job/90015469245?pr=25312
- https://github.com/matrixorigin/matrixone/actions/runs/30163396630/job/89692255700?pr=25312

Related older issues #21716 and #21830 addressed catalog-GC interference by changing the test options. This failure occurs after that fix and has a different trace.

Acceptance constraints for the fix:

- fix the underlying synchronization/progress problem rather than matching one observed timing;
- do not increase test timeout or normal execution time;
- do not remove or weaken the test's original coverage intent.

XuPeng-SH:
## What type of PR is this?

- [x] BUG
- [ ] Improvement
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26260.

`TestAppend2` exposed a real checkpoint-progress gap rather than a slow-runner problem.

The checkpoint completion predicate waits for every active appendable object at or before the checkpoint end. The stalled-checkpoint flush path, however, selected objects using the previous checkpoint as a creation-time lower bound. Since appendable-object creation moved outside append transactions, an object can be created before the previous checkpoint and receive committed appends afterward. In that state the checkpoint correctly keeps waiting, but the recovery flush can never select the blocking object.

This change aligns the recovery side with the completion predicate:

- checkpoint-bounded and forced flushes scan all active appendable objects up to their request end;
- normal cron flushes retain the existing previous-checkpoint lower bound, preserving their steady-state scan cost;
- stalled-checkpoint recovery now follows the configured flush cadence (two flush intervals), capped at the historical two-minute production threshold. The default production configuration therefore remains unchanged, while intentionally fast test configurations recover promptly.

The existing `TestAppend2` timeout, concurrent append workload, full-row scan, checkpoint completion assertion, and duplicate-append failure assertion are unchanged.

## Validation

- focused checkpoint recovery tests: pass (`0.039s`)
- `go test -race ./pkg/vm/engine/tae/db/checkpoint`: pass (`1.152s`)
- `TestAppend2 -race -count=20`: pass; `4.352s` after vs `4.533s` clean-main baseline
- `go test ./pkg/vm/engine/tae/db/test`: pass (`53.004s`)
- `go test -race ./pkg/vm/engine/tae/db/test`: pass (`68.188s`)
- `go vet ./pkg/vm/engine/tae/db/checkpoint`: pass
- `git diff --check`: pass

No timeout was increased and no coverage assertion was removed or weakened.

### DIFF
--- a/pkg/vm/engine/tae/db/checkpoint/flusher.go
+++ b/pkg/vm/engine/tae/db/checkpoint/flusher.go
@@ -42,9 +42,7 @@ import (
 
 var ErrFlusherStopped = moerr.NewInternalErrorNoCtx("flusher stopped")
 
-const (
-	stalledCheckpointFlushAge = checkpointIntentOldAge
-)
+const maxStalledCheckpointFlushAge = checkpointIntentOldAge
 
 type FlushCfg struct {
 	ForceFlushTimeout       time.Duration
@@ -102,6 +100,17 @@ func (mode flushScheduleMode) bypassFlushReady() bool {
 
 func (mode flushScheduleMode) resetFlushDeadline() bool {
 	return mode != flushScheduleCheckpointBounded
+}
+
+func (mode flushScheduleMode) objectScanLowerBound(lastCkp types.TS) types.TS {
+	// A checkpoint-bounded or forced flush must consider the same object set
+	// as IsTableTailFlushed: every active appendable object at or before the
+	// request end. Such an object can receive writes after it was created, so
+	// its creation timestamp cannot be used to exclude it from recovery.
+	if mode.bypassFlushReady() {
+		return types.TS{}
+	}
+	return lastCkp
 }
 
 type FlushRequest struct {
@@ -332,6 +341,7 @@ type flushImpl struct {
 	objMemSizeList []tableAndSize
 
 	stalledCheckpointPickBounded bool
+	stalledCheckpointFlushAge    time.Duration
 
 	// Log throttling for collectTableMemUsage
 	logThrottleMu   sync.Mutex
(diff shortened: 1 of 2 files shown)


LeftHandCold (approved):
Deep-reviewed the checkpoint completion and stalled-flush closure on this exact head. The recovery scan now covers the same active appendable-object set that can block IsTableTailFlushed, while normal cron scanning and the default production threshold remain unchanged. I also checked intent generation/age handling, ObjectList zero-TS seek ordering, PrepareCompact/scoped-task retry behavior, bounded queue/object growth, and the merge tree. No blocking correctness, liveness, ownership, or compatibility issue found. Focused checkpoint tests, checkpoint race tests, TestAppend2 race stress x20, full db/test, full db/test race, vet, build, and diff-check passed locally.

PR #26251, opened 2026-07-27
ci: remove obsolete merge MOC update workflow

XuPeng-SH:
## Summary

- Remove the obsolete post-merge MOC coverage-status workflow. The current coverage pipeline is self-contained and no longer creates the legacy MOC record this workflow expects.
- Stabilize three user-level-lock cleanup tests by asserting durable terminal state instead of transient asynchronous-worker ownership/progress, and always reset shared test state on early assertion exit.
- Add the read-only `mo-pr-radar` skill for deterministic PR review queues, conflict/CI/approval filters, request-changes freshness, and per-PR blocker explanations. Its review model uses each reviewer's latest opinionated review, so a later comment cannot erase an approval/change request and a later approval clears historical change requests.
- Strengthen `mo-self-review` with exact affected-test selection, adaptive per-test race budgets, deterministic flake checks, and one complete owning-package race run. Zero/under-resolution JSON test durations safely use the adaptive ceiling.

## Validation

- Confirmed the removed workflow runs only after a merged PR closes, is not a required check on main or 4.2-dev, and has no repository/Mergify references.
- PR radar: 22 offline unit tests, Bash syntax checks, both skill validators, and live GitHub regressions: #25089 remains `CHANGES_REQUESTED` after a later comment; approved #26233 is absent from the stale-change queue.
- Adaptive race measurements from `go test -json`, with `B=30s`: release-cleanup `T=0.07s, N=100`; timed-out cleanup `T=1.06s, N=28`; successful-probe cleanup `T=0.05s, N=100`.
- Final focused race stress passed for all three exact tests (`-count=100/28/100`); matching non-race stress also passed.
- Full `pkg/sql/plan/function` owning package passed with `-race -count=1`.
- `go build` and `go vet` passed for `pkg/sql/plan/function`.
- `git diff --check` passed after merging the latest main.

Fixes #26262.
Related to #26253.

PR #26268, opened 2026-07-27
fix(frontend): stop routine manager worker cleanly

### LINKED ISSUE #26267: [Bug]: frontend RoutineManager worker races with shared test configuration

### Branch Name

main

### Commit ID

12032c315b749eac555beef44bdde2e06676d2b2

- Hardware parameters: `amd64-mo-shanghai-8c16g`
- OS type: Linux/x86_64
- Others: Go race UT, runner `amd64-mo-shanghai-8c16g-92mgx-runner-5lm8f`

### Actual Behavior

`pkg/frontend` fails under `-race` with a cross-test data race. A `RoutineManager` worker created by `TestDoDropStage` reads `pu.SV.CleanKillQueueInterval` in `cleanKillQueue`, while `TestRoutineManagerKillAndCleanKillQueue` writes the same shared configuration. The race is reported against `TestRoutineManagerKillRoutineConnections`, which is only the test running when the race detector reports the earlier conflict.

Failing job: https://github.com/matrixorigin/matrixone/actions/runs/30292685275/job/90066019234?pr=26224

The worker can outlive the test because `NewRoutineManager` does not use the ParameterUnit supplied through its context, and cancellation does not wait for the worker to exit.

### Expected Behavior

`RoutineManager` workers use an immutable configuration snapshot, stop promptly and completely on cancellation, and frontend tests do not mutate shared configuration observed by background workers.

### Steps to Reproduce

1. Check out main after `12032c315b749eac555beef44bdde2e06676d2b2`.
2. Run `pkg/frontend` with the Go race detector.
3. Observe a race between `routine_manager.go:457` and `routine_manager_test.go:479`; the exact occurrence depends on goroutine scheduling.

Focused command:

```bash
.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=3 -timeout=180s \
  -run "^(TestDoDropStage|TestRoutineManager(KillAndCleanKillQueue|KillRoutineConnections))$" \
  ./pkg/frontend
```

### Additional information

This is not an OOM, network failure, or test timeout. It is a genuine race caused by mutable global test configuration combined with an incompletely-owned background worker lifecycle.

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26267

A frontend race UT exposed two coupled problems: `RoutineManager` background workers read mutable shared configuration, and cancellation did not wait for a worker to exit. A test modifying the global `CleanKillQueueInterval` could therefore race with a worker leaked by an earlier test.

This change:

- captures the worker and kill-queue intervals from the constructor ParameterUnit;
- honors the ParameterUnit carried by the constructor context, with the service-level PU as fallback;
- makes the periodic wait cancellation-aware and joins the worker during cancellation;
- removes shared PU mutation and a 2-second sleep from the related tests;
- adds a deterministic, zero-sleep test for configuration snapshot and worker shutdown.

The immediate-first-run and delay-after-work scheduling semantics are preserved. No long-running stress test is added.

Validation:

- focused related tests: PASS
- focused `-race -count=3`: PASS
- full `pkg/frontend`: PASS (14.6s)
- full `pkg/frontend -race`: PASS (21.4s)
- `go build ./pkg/frontend`: PASS
- `go vet ./pkg/frontend`: PASS

### DIFF
--- a/pkg/frontend/authenticate_test.go
+++ b/pkg/frontend/authenticate_test.go
@@ -10081,7 +10081,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10136,7 +10136,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10191,7 +10191,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10242,7 +10242,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10294,7 +10294,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10342,7 +10342,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10397,7 +10397,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10442,7 +10442,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
@@ -10491,7 +10491,7 @@ func Test_doAlterAccount(t *testing.T) {
 		pu.SV.KillRountinesInterval = 0
 		ctx := context.WithValue(context.TODO(), config.ParameterUnitKey, pu)
 
-		rm, _ := NewRoutineManager(ctx, "")
+		rm := newTestRoutineManager(t, ctx)
 		ses.rm = rm
 
 		//no result set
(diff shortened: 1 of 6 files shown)


XuPeng-SH:
Follow-up for the UT failure in https://github.com/matrixorigin/matrixone/actions/runs/30296615658/job/90079361000?pr=26251:

`TestSessionCloseRetainsCleanupWhenRetainedCapIsFull` filled the detached backlog but left `backlogStarted=false`. The production handoff then legitimately started the backlog consumer, freed capacity, transferred cleanup ownership, and bypassed the retained fallback that the test expected. The failure was therefore a scheduling-dependent test setup, not lost production cleanup ownership.

The added test fix now:

- freezes the synthetic backlog consumer while exercising the retained fallback;
- restores the blocked unlock path on every exit;
- removes only synthetic admission pressure after the fallback is observed;
- verifies the terminal invariant: the old synthetic lock and holder state are cleared, then a contender can acquire the lock.

Validation on latest main (including #26251):

- focused test `-race -count=3`: PASS, about 70ms per run
- full `pkg/sql/plan/function -race`: PASS
- full `pkg/frontend -race`: PASS
- full non-race tests, build, and vet for both packages: PASS

No sleep-based phase synchronization or long stress loop was added.

XuPeng-SH (commented):
Blocking issue found in the new parameter-unit hardening path.

`NewRoutineManager` now tries to return `"invalid parameter unit"` when the frontend parameters are missing, but the fallback path still panics before that guard runs:

- `routine_manager.go:595-601` falls back to `getPu(service)` when the context does not carry `config.ParameterUnitKey`.
- `server.go:583-584` still implements `getPu(service)` as `getServerLevelVars(service).Pu.Load().(*config.ParameterUnit)`.
- `InitServerLevelVars("")` only creates an empty `ServerLevelVariables`; it does not store a PU first.

So `NewRoutineManager(context.Background(), "")` still panics on the actual missing-PU path instead of returning the new explicit error. The added test only covers `pu.SV == nil`, so this fallback case remains untested.

Suggested fix: make the service-level PU lookup nil-safe (including the uninitialized-service case), then let `NewRoutineManager` return the intended error for all missing/invalid-PU inputs.

I did not find another blocking issue in the worker shutdown ordering itself.

XuPeng-SH:
Addressed the latest parameter-unit review on top of current main.

- Added a nil-safe service-level PU lookup for callers that explicitly handle missing initialization, while preserving the strict getPu contract for existing production callers.
- NewRoutineManager now reaches its explicit invalid-parameter error for both an unknown service and an initialized service with no PU.
- Reused the same lookup in the existing Iceberg preflight path instead of maintaining duplicate atomic-load logic.
- Added deterministic coverage for both missing-service states.

Fresh local validation after the final merge/edit:
- all affected focused tests: race count=100 PASS
- pkg/frontend full race: PASS (23.0s)
- pkg/sql/plan/function full race: PASS (35.2s)
- both full non-race packages: PASS
- build and vet for both packages: PASS

XuPeng-SH (commented):
Fresh pass on the latest diff: the original missing-PU constructor panic is fixed, but the new ctx-supplied ParameterUnit path still does not close the full runtime loop.

`NewRoutineManager` can now succeed by reading `config.ParameterUnitKey` from `ctx`, but it never persists that resolved `ParameterUnit` into the service-level slot or onto `RoutineManager` itself. Later runtime paths still read config exclusively through `getPu(rm.service)`, for example:

- `routine_manager.go:getConnID()`
- `routine_manager.go:Created()` (`NewMysqlClientProtocol(...)`, `NewRoutine(...)`, proxy flag checks)

So a concrete failure remains: build `ctx := context.WithValue(context.Background(), config.ParameterUnitKey, pu)`, call `NewRoutineManager(ctx, service)` without setting a server-level PU for that service, then exercise the first connection path. Construction succeeds, but the first `Created()` / `getConnID()` path still panics on `getPu(rm.service)`.

Suggested fix: either persist the resolved `ParameterUnit` into the service-level slot during construction, or store it on `RoutineManager` and stop re-reading it via the global service map on later paths.

I did not find another blocking issue in the worker-shutdown or retained-cleanup test changes.

XuPeng-SH:
Addressed the latest ParameterUnit runtime-closure review in 8491cc15cc.

The constructor now treats the service registry as the canonical publication point and keeps the resolved PU on RoutineManager:

- context-only construction completes all failure-prone initialization before atomically publishing the PU;
- an existing context/service PU mismatch fails without overwriting the registered service state;
- failed TLS initialization leaves the service unpublished, so a corrected initialization can retry;
- getConnID and Created use the RoutineManager snapshot, while Created uses the same service ID for NewSession, closing the downstream global-lookup path;
- affected tests use unique service IDs and check constructor errors, so repeated runs do not inherit package-global registry state.

Fresh validation after the final semantic edit:

- 16 explicitly listed focused tests measured individually under race; T=0–0.10s, B=30s, therefore N=100 for each;
- all 16 focused tests: race count=100 PASS;
- full pkg/frontend race/non-race: PASS (23.7s / 13.6s);
- full pkg/sql/plan/function race/non-race: PASS (33.7s / 14.0s);
- build, vet, err-check, and diff-check: PASS.

Reflection: the previous fix stopped at constructor validation and did not trace the resolved PU through Created -> NewSession -> service-level lookup. Its regression tests also stopped before the first real connection. This update uses the end-to-end state transition (resolve -> initialize -> publish -> consume -> close), including failed publication/retry and repeated-process test isolation, instead of another local guard.

XuPeng-SH (commented):
Self-review of latest head 8491cc15cc: no blocking correctness or performance issue found. Checked RoutineManager worker lifecycle, cancellation/join ordering, ParameterUnit resolve-initialize-publish-consume flow, concurrent/mismatched publication, failed TLS retry, connection creation, zero/negative interval behavior, and saturated retained-cleanup test isolation. Fresh local build/vet, focused repeated race tests, and full pkg/frontend + pkg/sql/plan/function race/non-race tests all passed. GitHub does not permit the PR author to approve their own PR.

Issue #26270, opened 2026-07-27
[Bug]: stale global sysvar refresh can overwrite a committed SET GLOBAL value


XuPeng-SH:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Related but not a duplicate:

- #24837 reported stale `SET GLOBAL` readback in multi-CN and was closed by #24844.
- #24844 refreshes the snapshot for `SHOW GLOBAL VARIABLES`, but does not prevent a stale snapshot from overwriting a newer `SET GLOBAL` update.

## Branch Name

main / CI run for PR #26251

## Commit ID

`d074e33040f83d924ba0f8dd728ef79a42ec56ec`

## Other Environment Information

- Topology: TKE, proxy endpoint with two CN services
- Failing job: https://github.com/matrixorigin/matrixone/actions/runs/30299453491/job/90090904734
- Failing BVT: `test/distributed/cases/zz_accesscontrol/password_management/password_reuse.sql`
- Local confirmation: two independent CN processes with direct SQL connections; the relevant global-sysvar paths are unchanged between the locally tested binary commit and the failing CI head.

## Actual Behavior

The BVT executes successfully until:

```sql
SET GLOBAL password_history = default;
SET GLOBAL password_reuse_interval = default;
SELECT @@global.password_history;
```

`password_history` has default value `0`, but the immediate read returns its old value `5`:

```text
expected: 0
actual:   5
```

The SQL statement succeeds; there is no disconnect, timeout, or SQL error.

A local black-box stress reproduction confirms a logical stale-snapshot lost update on a **single CN**:

- no concurrent global-sysvar refresh: 5,000 iterations of `SET 5 -> SET DEFAULT -> SELECT` had no mismatch;
- one concurrent connection repeatedly executing `SHOW GLOBAL VARIABLES LIKE 'password_history'`: mismatch at iteration 55;
- twelve concurrent refreshers: mismatch at iteration 3; three subsequent runs each mismatched at iteration 1.

At mismatch time:

```text
SELECT @@global.password_history  => 5
catalog variable_value             => 0
```

So the committed catalog value is correct, but the in-memory cache has been reverted to the old value.

A separate two-CN check also confirms that `SET GLOBAL` has no cross-CN cache invalidation: after CN1 commits `0`, a pre-existing CN2 session still reads `5` until it performs a global-sysvar refresh. This widens the impact, but the above same-CN race alone reproduces the CI symptom.

## Expected Behavior

[...]

PR #26271, opened 2026-07-28
fix(frontend): prevent stale global sysvar refresh

XuPeng-SH:
## Summary

Prevent a catalog snapshot captured before `SET GLOBAL` from replacing the committing CN's newer in-memory global-system-variable value.

- Record the account cache's local mutation generation before a refresh reads the catalog.
- Publish the full snapshot only if no local mutation advanced that generation while the read was in flight.
- Preserve the existing cache object when an entry was created or replaced during the read.
- Keep the generation specific to successful local mutations; refreshes do not invalidate each other.

## Root cause

`GlobalSysVarsMgr.Get` read the catalog outside the cache lock, then replaced the whole cache map. A concurrent `SET GLOBAL` could commit and update one cache entry in the gap, after which the older full snapshot restored the old value.

## Scope

This is Option A for the same-CN lost-update race described in #26270. It deliberately does not claim cross-CN cache propagation; that needs a durable revision plus invalidation/reconciliation design.

## Validation

- `go test -count=1 -timeout=10m ./pkg/frontend`
- `go test -race -count=1 -timeout=15m ./pkg/frontend`
- Each new exact regression test: `go test -race -count=100 -run '^Test…$' ./pkg/frontend`
- `go vet ./pkg/frontend`
- `go build ./pkg/frontend`

## Regression coverage

- Deterministically blocks a stale refresh, commits `SET GLOBAL password_history = 0`, releases the refresh, and verifies the cache remains `0`.
- Ensures refresh publication does not advance the local mutation generation.


XuPeng-SH (commented):
Approve-equivalent deep review on exact head bdb3f8352bf816766568f9458abc070ebb0addad: no merge-blocking correctness, performance, compatibility, or lifecycle issue found.

The same-CN publication invariant closes correctly. A refresh captures the current account object and its local mutation generation before catalog I/O. Final publication rechecks object identity under the manager lock and generation under the account-entry lock. Therefore a concurrent successful SET GLOBAL either publishes after the refresh or advances the generation and makes the older snapshot ineligible; the stale snapshot cannot be the final cache state.

I also traced entry creation/replacement, concurrent refreshes, catalog-error paths, session initialization, SHOW GLOBAL VARIABLES refresh, and SET GLOBAL commit-before-cache ordering. The lock graph is one-way from GlobalSysVarsMgr to SystemVariables with no reverse edge, no lock is held across catalog I/O, and the change adds no goroutine, resource owner, wait cycle, or unbounded container. The stated cross-CN invalidation gap remains intentionally outside Option A and is not obscured by this patch.

Local validation on this exact head passed:
- both exact generation regressions were enumerated and passed with race, count 100;
- full pkg/frontend tests passed in normal and race modes;
- pkg/frontend build and vet passed;
- diff check is clean.

GitHub does not allow the PR author to approve their own pull request, so this COMMENT is the strongest available review action. On code, this is an approve.

XuPeng-SH (commented):
Deep re-review on unchanged exact head bdb3f8352bf816766568f9458abc070ebb0addad, prospectively merged with latest main 3653b5e8d8c263936a55d926f36f3e0399c72fad: no merge-blocking correctness, performance, compatibility, or lifecycle issue found.

The target invariant is sound across the relevant interleavings. A refresh snapshots both the current account object identity and its local mutation generation before catalog I/O. Publication then rechecks identity under the manager lock and generation under the account lock. A successful same-CN SET GLOBAL either publishes after an older refresh or advances the generation and makes that refresh ineligible, so the older catalog snapshot cannot become the final local cache state.

This pass independently checked:
- SET commit-before-cache ordering, catalog failures, equal-value mutations, first cache creation, object replacement, and overlapping refreshes;
- lock ordering and contention: manager -> account is one-way, no account -> manager edge exists, and no lock is held across catalog I/O;
- allocation/lifecycle effects: no new goroutine, resource owner, retry loop, or unbounded state; the extra uint64 and uncontended lock operations are not material beside the catalog refresh;
- scope boundaries: cross-CN convergence and the pre-existing broader SET-vs-SET durable/cache ordering are not solved here, but neither is worsened or hidden by this focused Option A fix.

Fresh validation:
- both exact generation regressions passed under race for 100 repetitions;
- full pkg/frontend passed in normal and race modes;
- pkg/frontend build, vet, and diff checks passed;
- the prospective latest-main merge is clean, builds, and both regressions again passed under race for 100 repetitions;
- the latest CI attempt is green (the rollup still lists superseded cancelled/failed jobs from the earlier run).

GitHub does not permit an author to approve their own PR, so this COMMENT is the strongest available action. On code, this remains an approve.

Issue #26265, opened 2026-07-27
[Bug]: LOAD DATA parquet same-scale DECIMAL path bypasses target precision validation


XuPeng-SH:
### Description

This is a follow-up to #25473 and merged PR #26212.

PR #26212 fixed Parquet DECIMAL rescaling when the source and target scales differ, but the new dispatch still has a gap when the source Parquet DECIMAL scale already matches the target MatrixOne DECIMAL scale.

In that same-scale case, logical DECIMAL columns still go through the raw `decodeDecimal{64,128,256}Values` path via `canUseRawDecimalMapping`. That path only checks whether the value fits the storage width (DECIMAL64/128/256), not whether it fits the target column's declared `DECIMAL(precision,scale)`.

As a result, schema-invalid values can be loaded silently into narrower target precisions.

For example, a Parquet `DECIMAL(10,2)` value `12345.67` can still be accepted into a MatrixOne target column `DECIMAL(4,2)`, even though `DECIMAL(4,2)` should reject anything outside `[-99.99, 99.99]`.

### Version

Observed from the merged logic in PR #26212 (`81890b81a1ef479e6452ae81a0067113bdad33e8`).

### Steps to Reproduce

Generate a Parquet file with a logical DECIMAL column whose scale already matches the target, but whose precision exceeds the target column:

```python
from decimal import Decimal
import pyarrow as pa
import pyarrow.parquet as pq

path = '/tmp/decimal_same_scale_precision.parquet'
vals = [Decimal('12345.67')]
pq.write_table(
    pa.Table.from_arrays(
        [pa.array(vals, pa.decimal128(10, 2))],
        names=['dec_col'],
    ),
    path,
)
```

Run SQL:

```sql
drop database if exists parquet_decimal_same_scale_precision;
create database parquet_decimal_same_scale_precision;
use parquet_decimal_same_scale_precision;

create table t_bad(dec_col decimal(4,2));
load data infile {'filepath'='/tmp/decimal_same_scale_precision.parquet', 'format'='parquet'} into table t_bad;
select * from t_bad;
```

### Actual Behavior

The load can succeed and store `12345.67` in `DECIMAL(4,2)`.

### Expected Behavior

The load should be rejected with a DECIMAL overflow / precision error, because the target column is `DECIMAL(4,2)`.

### Root Cause

The merged dispatch uses:

- `canUseRawDecimalMapping(...)` to keep same-scale logical DECIMALs on the raw fast path
[...]

PR #26274, opened 2026-07-28
fix(external): validate parquet decimal target precision

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26265 

fix(external): validate parquet decimal target precision

### DIFF
--- a/pkg/sql/colexec/external/parquet.go
+++ b/pkg/sql/colexec/external/parquet.go
@@ -2849,8 +2849,9 @@ func isParquetDecimalCastSource(st parquet.Type) bool {
 }
 
 func canUseRawDecimalMapping(st parquet.Type, precision, scale int32) bool {
-	if isDecimalLogicalType(st.LogicalType()) {
-		return parquetDecimalScale(st) == scale
+	if lt := st.LogicalType(); isDecimalLogicalType(lt) {
+		return parquetDecimalScale(st) == scale &&
+			(precision == 0 || lt.Decimal.Precision <= precision)
 	}
 	return useRawIntegerDecimalMapping(st, precision, scale)
 }
--- a/pkg/sql/colexec/external/parquet_test.go
+++ b/pkg/sql/colexec/external/parquet_test.go
@@ -91,6 +91,61 @@ func TestParquetDecimalMappingRegression(t *testing.T) {
 	}, got)
 }
 
+func TestParquetDecimalSameScalePrecisionOverflow(t *testing.T) {
+	proc := testutil.NewProc(t)
+	ctx := context.Background()
+	encodings := []struct {
+		name       string
+		dictionary bool
+	}{
+		{name: "plain"},
+		{name: "dictionary", dictionary: true},
+	}
+	targets := []struct {
+		name            string
+		oid             types.T
+		targetPrecision int32
+		sourcePrecision int
+		byteWidth       int
+	}{
+		{name: "decimal64", oid: types.T_decimal64, targetPrecision: 4, sourcePrecision: 5, byteWidth: 8},
+		{name: "decimal128", oid: types.T_decimal128, targetPrecision: 19, sourcePrecision: 20, byteWidth: 16},
+		{name: "decimal256", oid: types.T_decimal256, targetPrecision: 39, sourcePrecision: 40, byteWidth: 32},
+	}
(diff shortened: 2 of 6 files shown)


XuPeng-SH (changes requested):
Deep-reviewed exact head `0509fcdab2091f941163b691a1cea66ba099a798`. The Parquet change itself is sound: same-scale logical DECIMAL stays on the raw path only when source precision proves its value domain fits the target, while narrower targets use the precision-aware conversion for plain/dictionary DECIMAL64/128/256. One real concurrency blocker remains in the accompanying error-preservation logic.

### P1 — A canceled sibling can still win before the real execution error

`preferPrimaryScopeError` only lets a substantive candidate replace `ErrPipelineEndSignalDeliveryFailed` (`pkg/sql/compile/compile.go:469-487`). In `runOnce`, results arrive in scheduler order. If a sibling observes the query cancellation first, `errToThrowOut` becomes `context.Canceled` or `ErrQueryInterrupted`; when the producer later publishes the actual Parquet DECIMAL overflow, the helper keeps the cancellation because current is no longer the synthetic delivery error (`compile.go:620-637`). This is reachable: dispatch/sendfunc/insert and VM paths directly return `context.Canceled`/`ctx.Err()` after another scope calls `Proc.Cancel`. The client can therefore still receive cancellation rather than the precision error this PR is trying to preserve.

Add the missing matrices `current=context.Canceled, candidate=executionErr` and `current=ErrQueryInterrupted, candidate=executionErr`; both currently fail. Avoid solving this with another static ranking that could overwrite a genuine client cancellation. Preserve/inspect the `CancelCauseFunc` cause (or carry causal scope/result metadata) so internally canceled siblings resolve to the error that triggered cancellation, while an externally canceled query still returns its external cause. Apply the same principle to nested MergeRun collection.

The current tests cover only cleanup-fallback-first ordering, so green CI does not exercise this production race.

iamlinjunhong:
> Deep-reviewed exact head `0509fcdab2091f941163b691a1cea66ba099a798`. The Parquet change itself is sound: same-scale logical DECIMAL stays on the raw path only when source precision proves its value domain fits the target, while narrower targets use the precision-aware conversion for plain/dictionary DECIMAL64/128/256. One real concurrency blocker remains in the accompanying error-preservation logic.
> 
> ### P1 — A canceled sibling can still win before the real execution error
> `preferPrimaryScopeError` only lets a substantive candidate replace `ErrPipelineEndSignalDeliveryFailed` (`pkg/sql/compile/compile.go:469-487`). In `runOnce`, results arrive in scheduler order. If a sibling observes the query cancellation first, `errToThrowOut` becomes `context.Canceled` or `ErrQueryInterrupted`; when the producer later publishes the actual Parquet DECIMAL overflow, the helper keeps the cancellation because current is no longer the synthetic delivery error (`compile.go:620-637`). This is reachable: dispatch/sendfunc/insert and VM paths directly return `context.Canceled`/`ctx.Err()` after another scope calls `Proc.Cancel`. The client can therefore still receive cancellation rather than the precision error this PR is trying to preserve.
> 
> Add the missing matrices `current=context.Canceled, candidate=executionErr` and `current=ErrQueryInterrupted, candidate=executionErr`; both currently fail. Avoid solving this with another static ranking that could overwrite a genuine client cancellation. Preserve/inspect the `CancelCauseFunc` cause (or carry causal scope/result metadata) so internally canceled siblings resolve to the error that triggered cancellation, while an externally canceled query still returns its external cause. Apply the same principle to nested MergeRun collection.
> 
> The current tests cover only cleanup-fallback-first ordering, so green CI does not exercise this production race.

fixed

XuPeng-SH (approved):
Deep-reviewed exact head b4a51193bbcf7ba9909710fdd0f2bc1fcb350cb5. The prior cancellation/error-causality blocker is fixed for both top-level and nested MergeRun paths while preserving genuine external cancellation. The Parquet DECIMAL precision routing is sound for plain and dictionary DECIMAL64/128/256. Focused race tests and full race tests for ./pkg/sql/compile ./pkg/sql/colexec/external passed.

Issue #26264, opened 2026-07-27
[Bug] JSONLine external-table projection fails beyond the second column


Ariznawlll:
## Description

JSONLine external tables with object-shaped records cannot read projected columns from the third declared column onward. The first two columns can be queried normally, but reading the third or later column, or using `COUNT(*)`, returns an internal `index out of range` error.

The same behavior occurs for a writable external table after inserting rows and for a read-only external table created over the emitted files.

## Version

Reproduced three times on `origin/main` at `cd8e175349932ccf4be8d388080a0c3310e538ac`.

## Steps to reproduce

Use a local stage directory and create a writable JSONLine external table:

```sql
CREATE STAGE jsonline_audit_stage URL = 'file:///tmp/jsonline-audit';

CREATE EXTERNAL TABLE ext_rows (
  a INT,
  b VARCHAR(20),
  c TEXT,
  d DECIMAL(10,2),
  e JSON,
  f BOOL
)
INFILE {
  'filepath' = 'stage://jsonline_audit_stage/rows_*.jsonl',
  'format' = 'jsonline',
  'jsondata' = 'object',
  'write_file_pattern' = 'stage://jsonline_audit_stage/rows_%U.jsonl'
};

INSERT INTO ext_rows VALUES
  (1, 'one', 'alpha', 1.25, '{"n":1}', true),
  (2, 'two', 'beta', 2.50, '{"n":2}', false);

-- Both queries complete normally.
SELECT a, b FROM ext_rows ORDER BY a;

-- Each of the following fails.
SELECT c FROM ext_rows ORDER BY a;
SELECT d FROM ext_rows ORDER BY a;
SELECT e FROM ext_rows ORDER BY a;
SELECT f FROM ext_rows ORDER BY a;
SELECT COUNT(*) FROM ext_rows;
```

To check the read-only form, recreate `ext_rows` without `write_file_pattern`, pointing its `filepath` at the same `rows_*.jsonl` files. The projected-column and aggregate queries fail in the same way.

## Actual behavior

`SELECT a, b` returns the two expected rows. Queries that access `c` or a later column, and `COUNT(*)`, return error 20101 with an internal `index out of range` stack rooted in `pkg/sql/colexec/external.getFieldFromLine`.

## Expected behavior

All declared JSONLine columns should be independently selectable, and aggregates should complete normally for both writable and read-only external tables.

## Impact

[...]

PR #26276, opened 2026-07-28
fix(external): preserve JSONLine projection field positions

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26264 

fix(external): preserve JSONLine projection field positions


heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed at head 834c34774e468d87bfef8aa9b8f0f0b6a84cc236.

The fix preserves the planner/executor coordinate contract: projected JSON object values are placed at their original physical field indexes, while __mo_filepath remains synthetic and is excluded from the JSON lookup. Invalid metadata is rejected before allocation, and the 4096-column bound prevents malformed field indexes from becoming an unbounded allocation path. Missing keys and missing column definitions remain explicit errors.

Validation performed locally on the exact head:
- each of the 5 new projection/error tests under -race, first measured individually and then repeated independently with -count=16
- full ./pkg/sql/colexec/external package under -race -count=1

All passed. CI is green, including UT, coverage, BVT, SCA, and Iceberg E2E. No blocking issue found.

PR #26277, opened 2026-07-28
fix(parser): support MySQL OJ and STRAIGHT_JOIN USING

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25378

fix(parser): support MySQL OJ and STRAIGHT_JOIN USING


XuPeng-SH (approved):
Reviewed exact head `8a13ec392c3972b2361c536e6cb11dbe77a74a01` against base `c8e443bf8610b2af4333ad943ba2075485b16cf8`.

The grammar change matches MySQL's `escaped_table_reference` / `join_specification` shape: the OJ wrapper is case-insensitively validated and lowers to the underlying join AST, while `STRAIGHT_JOIN ... USING` reaches the existing `UsingJoinCond` planner path. Existing `STRAIGHT_JOIN ... ON` behavior is preserved. I also checked nested/parenthesized OJ, comma composition, multi-column and chained USING, and malformed escape/outer-join cases.

The ~20k-line generated-parser churn is reproducible: forcing the repository goyacc generation produced the exact committed `mysql_sql.go` hash with zero conflicts. Local gates passed: full MySQL parser package, focused parser regression under `-race`, focused plan regression under `-race`; the PR CI matrix is green.

Non-blocking follow-up: MatrixOne already lowers `JOIN_TYPE_STRAIGHT` to ordinary `Node_INNER`, so the optimizer may reorder it instead of enforcing MySQL's left-before-right semantics. That predates this parser-only change (the existing ON form has the same behavior), so it is not a blocker for this PR.

XuPeng-SH:
@Mergifyio refresh

PR #26281, opened 2026-07-28
fix: distribute sink scan shuffle hashbuild

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25114

  - Keep the in-process SINK_SCAN producer on its owning CN.
  - Distribute shuffle hash-build receivers across all query worker CNs instead of
    concentrating them on the coordinator.

  - Ensure the local SINK_SCAN scope is attached to exactly one local receiver tree
    and is never serialized to a remote CN.

  - Preserve mixed local/remote shuffle dispatch and standalone remote execution.


XuPeng-SH (changes requested):
Blocking: the added SINK_SCAN owner is not carried into the later per-CN grouping stage.

`shuffleJoinStageNodes` now correctly appends the SINK_SCAN owner when it is absent from `queryWorkerStageNodes()`. However, every affected INSERT/MULTI_UPDATE path later calls `groupShuffleBucketsByCNIfNeeded`, which recomputes only `queryWorkerStageNodes()` and does not know about that appended owner.

I reproduced both deterministic failure modes with an end-to-end compileShuffleJoin -> groupShuffleBucketsByCNIfNeeded test where the local SINK_SCAN owner is outside `c.cnList`:

- With one scheduled remote worker and DOP=2, compile creates 4 buckets across 2 CNs, but grouping returns all 4 unchanged because it sees `len(stageNodes) <= 1`. The remote DOP buckets therefore remain separate trees even though their dispatch targets sibling local receivers, so they are not standalone remote units.
- With two scheduled remote workers and DOP=2, compile creates 6 buckets across 3 CNs, but grouping returns only 2 trees: `mergeScopesByStageNodes` iterates the two scheduled workers and silently drops every bucket on the appended SINK_SCAN-owner CN.

Please carry the actual receiver stage-node set into grouping (or derive/preserve all unique execution nodes from `ss`) and add closure tests for an owner outside one-worker and multi-worker scheduled sets. The existing tests separately verify owner appending and grouping only when the owner is already in `cnList`, so they miss this composition boundary.

The PRs own focused tests pass under `-race -count=100`; the full `pkg/sql/compile` package passes under race; build, vet, diff check, current-main merge-tree, and online CI pass.

XuPeng-SH (approved):
Re-reviewed exact head `c94c5c1c3f555e84f70a962384ab0c31f8187ff1` against latest `main` (`1b6d4308036f69e0486420fb175ab3de3a5c4ae2`). Approved.

The prior blocker is closed:

- grouping now derives the unique execution-node set from the actual shuffle buckets, using the same node-equivalence rule as `mergeScopesByStageNodes`;
- an external SINK_SCAN owner is therefore preserved for both one-worker and multi-worker scheduled sets;
- the new composition tests cover the prior "left ungrouped" and "silently dropped" failure modes, including DOP fan-out and standalone remote-tree invariants.

I also checked the full compile topology and unhappy paths: the production SINK_SCAN owner is materialized before distribution, receiver/source ownership remains reachable through per-CN grouping, all buckets are retained exactly once, and the added planning work is bounded by the receiver count. The CDC test's phase barrier is cancellation-released and makes the assertion deterministic without introducing a goroutine/resource leak.

Local validation on the clean merge with current main:

- each of the four changed tests under `-race -count=100`;
- full `pkg/sql/compile` and `pkg/cdc` under `-race`;
- `go build` and `go vet` for both packages;
- merge and diff checks.

One non-blocking scope note: this PR makes hash-build memory distributable and is a sound mitigation for #25114, but it does not by itself prove the original 10M-vector workload's peak-memory target. The issue should remain open until that workload is measured; the PR currently does not auto-close it, which is appropriate.

Issue #21819, opened 2025-05-12
[Tech Request]: add debug tool


jiangxinmeng1:
### Is there an existing issue for the same tech request?

- [x] I have checked the existing issues.

### Does this tech request not affect user experience?

- [x] This tech request doesn't affect user experience.

### What would you like to be added ?

```Markdown
copy data from certain table, and apply it in another mo instance
```

PR #26282, opened 2026-07-28
refactor: remove legacy mo_ctl table dump path

aptend:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [x] Code Refactoring

issue #21819

related issue #25782

The debug path introduced by #21821 is superseded by the native SQL `DUMP TABLE` and `LOAD TABLE` support added in #25989 for #25782, so the old debug-only `mo_ctl` command and data-application path are no longer needed.

This PR removes:

- the `dump-table` and `apply-table-data` inspect commands and implementations
- the `txn.debug-mode` configuration and TAE option used only by that path
- dedicated dump-table ObjectIO names, writer type, and fault injection
- tests that exercised the removed implementation

For rolling-upgrade safety, it retains only a cleanup-only `dumpTable` TmpFileService registration and legacy directory decoder. New code cannot create these files; upgraded TNs continue reclaiming table-sized dumps left by an older version.

## Operational compatibility

The native SQL replacement operates on immutable/flushed objects and rejects unsupported schemas including partitioned, auto-increment, foreign-key/referenced, temporary, external, source, view, sequence, and transient tables. Existing debug users must flush live data first and cannot migrate unsupported schemas through native `DUMP TABLE` / `LOAD TABLE`.

## Testing

- `go build ./pkg/vm/engine/tae/rpc`
- `go vet ./pkg/vm/engine/tae/rpc`
- full `go test` for `pkg/vm/engine/tae/rpc`
- focused race test for legacy upgrade GC
- dependent `pkg/txn/storage/tae` test
- existing CI coverage for affected objectio, tnservice, TAE options/RPC, and native frontend DUMP/LOAD paths


XuPeng-SH (changes requested):
The removal is otherwise internally consistent: no live references remain, `WriterDumpTable` was the final enum value, affected packages pass race/vet, and the native DUMP/LOAD path still passes its focused race suite. One upgrade blocker remains.

[P1] Keep a transition GC for dump data created before the upgrade

The deleted `dump_table.go` did more than expose the commands: its package `init` registered the `dumpTable` TmpFileService app and `GCDumpTableFiles`. `dump-table` writes full table object copies below `<data-dir>/tmp/dumpTable/<table-id>_<creation-time>_<snapshot-ts>`, and the command itself was not gated by `debug-mode` (only apply was). Those copies can therefore be table-sized.

After this deletion, an upgraded TN still discovers the directory in `TmpFileService.init`, but `pkg/fileservice/tmp_fs.go:214-233` looks it up in the global `appConfigs`, logs `fs.tmp.init.not.found`, and skips it. Periodic GC iterates only registered `apps` (`:139-176`). Consequently every pre-upgrade dump, including one created by an old TN immediately before a rolling restart, becomes permanently orphaned and can fill the TN local volume.

Please retain a cleanup-only legacy `AppConfig`/directory decoder for at least the old maximum retention or one compatibility window, while removing the inspect commands, apply implementation, writer type, and debug flag. Add an upgrade-shaped test that pre-populates a legacy `dumpTable` directory, initializes TmpFileService with the new code, advances/injects expiry, and verifies the old files and directory are reclaimed.

Non-blocking documentation note: this is an operational API removal rather than only a code refactor. Native DUMP/LOAD requires immutable/flushed objects and rejects partitions, auto-increment tables, and foreign keys, while the removed TAE diagnostic path handled live appendable data directly. The PR/release note should state that scope difference so existing debug users know to flush and whether their schema is unsupported.

Validation: full race passed for objectio, tnservice, TAE options/RPC, and focused native frontend DUMP/LOAD tests; the TAE db/test package compiles; vet passed for all affected/alternate-path packages. The head merges cleanly with latest main and RPC/TN race still passes in the prospective merge.

aptend:
Addressed the upgrade blocker in commit `465862b97`:

- retained a cleanup-only registration for the legacy `tmp/dumpTable` application
- retained only the legacy directory decoder and 24-hour GC behavior; no command, writer, apply path, debug flag, or fault injection was restored
- added an upgrade-shaped test that pre-populates an expired legacy dump directory before `TmpFileService` initialization and verifies the periodic GC removes both its object and directory
- updated the PR description with the native DUMP/LOAD operational limitations

Validation completed:

- focused legacy GC test and focused `-race` test
- full `pkg/vm/engine/tae/rpc` test suite
- `go build` and `go vet` for the RPC package
- dependent `pkg/txn/storage/tae` test

XuPeng-SH (approved):
Deep-reviewed exact head `465862b97e2e5af36e3793e81cc9d489a8c2aca6`.

The previous upgrade blocker is resolved: the cleanup-only `dumpTable` app registration remains reachable in production, the decoder matches the legacy writer's directory format and actual 24-hour GC behavior, and the upgrade-shaped test creates legacy data before TmpFileService initialization. Error and timeout paths retry instead of reporting a partial cleanup as complete.

I found no remaining production correctness, compatibility, or performance blocker. On installations without a legacy `dumpTable` directory, this adds only an init-time registry entry; no app is instantiated and no periodic legacy scan is added.

Validated on this exact head:
- focused legacy upgrade GC under `-race`, 20 repetitions
- full `pkg/vm/engine/tae/rpc` tests
- dependent `pkg/txn/storage/tae` tests
- compile-only checks for `pkg/objectio`, `pkg/tnservice`, and TAE options
- `go vet` across all affected production packages
- `git diff --check`

CI is still running and the branch is currently behind `main`; neither revealed a code-review blocker in the reviewed diff.

PR #26284, opened 2026-07-28
fix: distribute sink scan shuffle hashbuild

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25114

Keep the in-process SINK_SCAN producer on its owning CN.

Distribute shuffle hash-build receivers across all query worker CNs instead of
concentrating them on the coordinator.

Ensure the local SINK_SCAN scope is attached to exactly one local receiver tree
and is never serialized to a remote CN.

Preserve mixed local/remote shuffle dispatch and standalone remote execution.

### DIFF
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -5467,7 +5467,7 @@ func scopeTreeHasCrossCNDispatch(s *Scope) bool {
 // doSetRootOperator) appends a connector *on top of* that existing root using AppendChild
 // semantics, so the caller's operator is preserved as the connector's child, not replaced.
 func (c *Compile) groupShuffleBucketsByCNIfNeeded(ss []*Scope) []*Scope {
-	stageNodes := c.queryWorkerStageNodes()
+	stageNodes := shuffleBucketStageNodes(ss)
 	if len(stageNodes) <= 1 || len(ss) <= len(stageNodes) {
 		return ss
 	}
@@ -5482,6 +5482,26 @@ func (c *Compile) groupShuffleBucketsByCNIfNeeded(ss []*Scope) []*Scope {
 		return ss
 	}
 	return c.mergeScopesByStageNodes(ss, stageNodes)
+}
+
+// shuffleBucketStageNodes derives the grouping boundary from the receiver
+// scopes themselves. The receiver set may include a local SINK_SCAN owner that
+// is intentionally absent from the scheduled query-worker set.
+func shuffleBucketStageNodes(ss []*Scope) engine.Nodes {
+	stageNodes := make(engine.Nodes, 0, len(ss))
+	for _, scope := range ss {
+		found := false
+		for _, node := range stageNodes {
+			if sameExecutionNode(node, scope.NodeInfo) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			stageNodes = append(stageNodes, scope.NodeInfo)
+		}
+	}
+	return stageNodes
 }
 
 func (c *Compile) mergeScopesByStageNodes(ss []*Scope, stageNodes engine.Nodes) []*Scope {
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -5249,11 +5249,12 @@ func (c *Compile) queryWorkerStageNodes() engine.Nodes {
 	return c.materializeScheduledWorkers(decision.Workers)
 }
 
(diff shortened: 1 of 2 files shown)


XuPeng-SH (changes requested):
Blocking: the SINK_SCAN owner added to the receiver stage is not preserved by the later per-CN grouping.

`shuffleJoinStageNodes` now appends the SINK_SCAN owner when it is absent from `queryWorkerStageNodes()` (`compile.go:5258-5272`). However, every affected INSERT/MULTI_UPDATE path later calls `groupShuffleBucketsByCNIfNeeded`, which independently recomputes only `queryWorkerStageNodes()` (`compile.go:5469-5484`). `mergeScopesByStageNodes` then iterates only that scheduled-worker list (`compile.go:5487-5503`).

I reproduced this deterministically on the exact head through the real `compileJoin` path with DOP=2 and a SINK_SCAN owner outside `c.cnList`:

- With one scheduled worker, `compileJoin` creates 4 buckets over 2 CNs, but grouping returns all 4 unchanged because `len(stageNodes) <= 1`. The remote send trees are still non-standalone: their dispatch local receivers point to sibling bucket trees, which is the wrong-CN/hang condition this grouping helper is intended to prevent.
- With two scheduled workers, `compileJoin` creates 6 buckets over 3 CNs, but grouping returns only 2 trees. All buckets owned by the SINK_SCAN CN are silently dropped because that owner is not in the recomputed worker list.

Please carry the actual join stage-node set into grouping, or derive and preserve all unique execution nodes from `ss`, and add closure tests for an owner outside both one- and two-worker scheduled sets through `compileJoin` plus grouping.

The new `TestCompileShuffleJoinDistributesSinkScanHashbuild` does not exercise this boundary: it calls `compileShuffleJoin` directly, which uses `queryWorkerStageNodes()` and bypasses `shuffleJoinStageNodes` (`compile.go:3701-3703`), and its owner is already in `cnList`.

Validation on this exact head: all five directly affected tests passed independently with `-race -count=100`; the full `pkg/sql/compile` race suite, build, and vet also passed. The missing composition case is why current CI remains green.

XuPeng-SH (approved):
Re-reviewed exact head 3cb335d08ba2591782ae842d61c215e8bb81bc97 against exact base 4.2-dev@1c48a13971bdfb63ae13fe53c3485e8decc1e079.

The previous blocker is resolved. `groupShuffleBucketsByCNIfNeeded` now derives the grouping boundary from the actual receiver buckets instead of recomputing only the scheduled query-worker set. Therefore an external SINK_SCAN owner survives both cases that previously failed: 1 worker now groups 4 buckets into 2 complete CN trees, and 2 workers group 6 buckets into 3 without dropping the owner's 2 buckets. The new `compileJoin` composition test covers both cases and verifies that every remote group is standalone-executable and exactly one owner tree retains the in-process SINK_SCAN.

The full topology is consistent: stage nodes are deduplicated once; probe/build shuffle operators use the resulting `len(stageNodes) * dop` bucket count; the SINK_SCAN source attaches to the receiver on its owner CN; each bucket keeps a unique ShuffleIdx; and later per-CN merging partitions the same bucket set with the same execution-node identity relation. Ordinary shuffle and single-CN paths keep their existing gating. This is compile-time topology work only and adds no per-row execution overhead; hashbuild state is distributed instead of concentrated on the coordinator.

Local verification:
- all seven directly affected topology tests passed independently under -race for 31-35 repetitions;
- the complete pkg/sql/compile package passed under -race;
- vet, build-only test, diff checks, and the current 4.2-dev merge tree passed;
- all current substantive GitHub checks are successful.

Non-blocking follow-ups:
- `shuffleJoinStageNodes` materializes an empty owner address, but the original source scope remains empty and `attachShuffleDispatchSource` falls back to receiver 0. Production `compileSinkScanNode` assigns the current CN address, so this is not currently reachable there; if the empty-address branch is intended as a supported contract, normalize the source or make the owner receiver first and add that case.
- This is a plausible second-stage #25114 improvement on top of merged #26252, which already bounds vector LOAD buffers/hashbuild spill. Keep #25114 open until the original 10M wiki-vector nightly reproduction confirms both successful import and the expected per-CN memory distribution; topology UTs alone do not prove the 41 GB incident is closed.

PR #26288, opened 2026-07-28
fix: align explicit integer string casts with MySQL

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25216

This PR aligns explicit string-to-integer casts with MySQL integer-prefix
semantics.

For example:

```sql
SELECT CAST('7e0' AS SIGNED),
       CAST('7e+2' AS SIGNED),
       CAST('1.5e0' AS SIGNED),
       CAST('7e2' AS UNSIGNED);
```

now returns `7, 7, 1, 7`. Explicit integer casts consume only the leading
decimal integer prefix; they do not consume the exponent or fractional suffix.
Inputs whose fractional form has no integer digits, such as `.5e0`, `-.5e0`,
and `+.5e+2`, convert to zero.

The change is restricted to the explicit `CAST` path. Assignment casts,
implicit comparison/arithmetic conversion, prepared statements, joins,
indices, temporal arithmetic, and warning behavior are unchanged. Existing
MatrixOne binary/octal/hexadecimal string extensions continue to use their
strict radix parser.

Validation:

- `make build`
- full `pkg/sql/plan/function` CGo unit tests
- focused race test repeated 100 times
- full `pkg/sql/plan/function` race test
- `mo-tester`: `func_cast.test` 243/243
- focused `golangci-lint`: 0 issues
- `make static-check` reached the full lint stage but is currently blocked by
  two pre-existing unused methods on `main` in
  `pkg/common/malloc/sane_allocator_cache.go`


XuPeng-SH (approved):
Deep-reviewed exact head 8fc8f08b89c23b3fe66201525805183f3956245f and prospectively merged it cleanly onto latest main ca4a91c5aa.

The new scanner is correctly limited to explicit string-to-integer casts; normal/assignment/implicit paths remain unchanged, while MatrixOne binary/octal/hex string extensions still take the strict radix parser. I cross-checked the expanded edge matrix against a disposable MySQL 8.0.44 server: exponent and fractional suffixes consume only the leading integer prefix, leading fractional forms become zero, signed-to-unsigned modulo behavior matches, and overflow behavior remains consistent. Empty/non-numeric input behavior is unchanged by this PR.

Validation on the merged tree passed: the exact new test under race for 44 repetitions, the full pkg/sql/plan/function race suite, go vet for the owning package, and diff checks. Current substantive CI is green. No blocking findings.

heni02 (approved):

Issue #26246, opened 2026-07-27
[Bug] Backup ignores caller cancellation while copying object files


Ariznawlll:
## Summary

Object-file copying during a backup no longer observes the context passed to `BackupData`. Once copying has started, cancellation or a deadline on the caller does not make the backup return until the active file operation itself finishes.

## Reproduction

1. Start a backup that reaches `parallelCopyData` with a file service whose `Read(ctx, ...)` waits for its context to finish.
2. Cancel the context originally passed to the backup after `Read` has started.
3. The copy remains waiting; it returns only after the test manually releases the file service.

The behavior reproduced 3/3 times on `8c9bb66898afa74f8b231e1915d99489ae8a0da7`.

## Actual behavior

`execBackup` receives the caller context, but `parallelCopyData` initializes each job with `context.Background()` and invokes `CopyFileWithRetry(context.Background(), ...)`. The worker wait is unbounded, so a blocked copy keeps the backup request active after the caller has cancelled it.

## Expected behavior

The copy jobs and retry loop should derive from the backup caller context. A cancelled or expired backup request should return promptly and stop waiting for outstanding copy work.

## Impact

A backup operation can remain active longer than its caller allows when a storage read or write does not finish.

LeftHandCold:
已在当前 `main`（`6004becb56adc7f4d22dbfb9554bef8d7091e25d`）确认是生产可达的真实问题。

根因是 `BackupData` 虽然收到 caller context，但 `parallelCopyData` 创建 job 和调用 `CopyFileWithRetry` 时都换成了 `context.Background()`；同时旧的 retry helper 也不能在 backoff 中响应取消。另一个并发边界是单槽 `errC` 会让多个失败 worker 阻塞，函数提前 return/close 后存在生命周期风险。

修复 PR：#26289

修复会贯穿 caller context、用 cancel cause 保留首个根因并取消兄弟任务、有界接单且等待所有已提交任务退出，同时改用 context-aware retry。新增 blocked read、retry backoff、兄弟任务取消的确定性回归；`pkg/backup` 完整 non-race/race、focused `-race -count=100`、build/vet 和 `pkg/datasync.TestCopyFile` 均通过。

LeftHandCold:
fixed

Ariznawlll:
### 测试版本

- 分支：`main`
- commit：`6b3ad84c9d01d453c2bcdd0f440637350e3aa8c6`
- 测试日期：2026-08-02
[...]

PR #26289, opened 2026-07-28
fix(backup): honor cancellation during object copy

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26246

Current `main` already contains #26290's bounded admission, first-error sibling
cancellation, joining of scheduled jobs, and missing-object failure semantics.
The remaining bug is that backup object-copy jobs derive their context from
`context.Background()`, and retry backoff uses the non-context-aware retry
helper. Consequently, canceling `BackupData` cannot stop a blocked file
operation or interrupt a retry delay.

This PR is intentionally limited to that remaining defect:

- pass the `execBackup` caller context into `parallelCopyData`;
- derive the existing cancel-cause context from the caller while preserving
  #26290's error and scheduling behavior;
- use `DoWithRetryContext` so caller or sibling cancellation interrupts retry
  backoff;
- add deterministic regressions for cancellation during a blocked read and
  during retry backoff.

Verification:

- focused cancellation regressions fail to compile on current `main` because
  `parallelCopyData` has no caller-context path, and pass after the fix;
- focused cancellation/concurrency tests: `-race -count=100`;
- `pkg/backup`: full tests and full race tests;
- `pkg/backup`: `go build` and `go vet`;
- dependent `pkg/datasync.TestCopyFile`.


XuPeng-SH (approved):
Deep-reviewed exact head fb2c31f8a89fa501767a18edb788878f67b7fcbf against current main 42ba9fb1a2cf74b45190b4d93b497fc4a1a8e199.

The cancellation and ownership model is sound. The caller context now reaches both the scheduled job and every source/destination file operation; context-aware retry makes backoff interruptible. Admission is bounded by parallelism, a permanent error establishes the cancel cause before siblings unwind, and the caller joins every successfully submitted job before reading results or stopping the scheduler. This closes the old undersized errC blocked-send / post-return worker lifetime hazards while preserving the first causal error.

Local verification:
- caller cancellation, sibling cancellation, and retry-backoff cancellation passed individually under -race for 43/43/44 repetitions;
- existing backup CopyFileWithRetry and dependent datasync TestCopyFile passed under -race for 43/42 repetitions;
- the complete pkg/backup package passed under -race;
- vet, datasync regressions, diff checks, and a prospective merge with latest main passed;
- all current substantive GitHub CI checks are successful.

Non-blocking coordination: use `Fixes #26246` so the issue closes automatically. PR #26290 changes the same scheduler block and test filename; whichever PR lands second must rebase and preserve both contracts: missing checkpoint objects are fatal, and caller cancellation remains end-to-end with joined workers. Also, the timeout cleanup in TestParallelCopyDataCancelsSiblingAfterCopyFailure would be more robust with a second bounded select instead of an unbounded `<-result`.

Issue #26286, opened 2026-07-28
[Bug] Concurrent MinIO backup copies can exhaust the shared per-host HTTP pool and hang CN


LeftHandCold:
## Summary

A backup from an object-store SharedFs to a MinIO/S3 backup target can deadlock when object-copy parallelism exceeds the shared HTTP transport's per-host connection limit. In the reported production configuration, backup parallelism was 150 while the default `MaxConnsPerHost` is 100; the CN became unresponsive after the pool was exhausted.

This is reproducible on current `main` at `6004becb56adc7f4d22dbfb9554bef8d7091e25d`.

## Root cause

`pkg/backup/tae.go:CopyFile` currently does:

1. `srcFs.Read` opens a streaming source GET.
2. `defer reader.Close()` retains that response/connection until `CopyFile` returns.
3. `dstFs.Write` is called with the source reader.
4. `S3FS.Write` performs a destination `Exists` HEAD before it consumes the reader.

The source and dynamically-created backup destination use the same process-wide HTTP round tripper when `max-conns-per-host` is not configured. With 100 or more workers against the same MinIO host, 100 source GET responses can hold all 100 connections. Every worker then waits for a destination HEAD/PUT connection. No worker can consume its source reader, reach EOF, close it, or release a connection.

Wait-for cycle:

```text
worker -> destination HEAD/PUT -> shared per-host permit
shared per-host permits -> source GET readers
source GET readers -> worker's destination Write completion
```

The `defer` is not a missing-close leak: it provides exactly-once cleanup on success and error. The bug is that the reader lifetime crosses a second request that depends on the same bounded connection pool.

## Deterministic reproduction

A focused test models a shared two-connection pool and starts two `CopyFile` calls:

- both source reads acquire and retain one connection;
- both destination writes require another connection before consuming the reader;
- both copies block indefinitely until the test aborts the pool.

The same topology exists with MinIO at 150 backup workers and the default 100 per-host connections.

## Impact

- backup can stop making progress;
- the backup shares the CN-wide HTTP transport, so unrelated object-store requests queue behind the same exhausted pool;
- the CN can appear hung until the backup is canceled or the process is restarted.

## Proposed fix

[...]

PR #26291, opened 2026-07-28
fix(backup): avoid MinIO copy connection deadlock

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26286

At backup parallelism 150, the default shared HTTP transport can deadlock at its
100 per-host connection limit: every worker retains a source GET response while
`S3FS.Write` waits for a destination HEAD/PUT connection before consuming that
reader.

This change breaks both sides of that wait-for cycle:

- compatible object stores use the existing provider-side `ObjectCopier`
  before opening a source response in CN, then hash the copied destination;
  the backup SHA-256 manifest contract is preserved and describes the bytes
  actually stored in the backup;
- destination checksum reads retry as a separate post-copy phase, so a
  transient GET error cannot repeat a successful provider copy and fail on the
  now-existing target;
- dynamically-created S3 backup targets get a dedicated HTTP transport at the
  normal default capacity, isolating the streaming fallback for incompatible
  endpoints or credentials from the CN's process-wide SharedFs pool;
- explicit `max-conns-per-host` settings remain unchanged, and an unsupported
  provider-copy fallback still reads the source exactly once.

The concurrency regression models a saturated shared pool and deterministically
reproduces the old cycle. It also checks that provider copy starts without a CN
source reader, validates the provider copy count and checksum, and covers the
single-read fallback. A retry regression verifies a transient destination hash
error does not repeat provider copy. A fileservice test verifies default pool
isolation and explicit-setting preservation.

Verification:

- deterministic connection-cycle regression red on current main, green after
  the fix;
- `pkg/backup` and `pkg/fileservice` full tests and full race tests;
- all focused race regressions stressed separately with `-count=100`;
- `go build` and `go vet` for both modified packages;
- dependent `pkg/datasync.TestCopyFile`.


XuPeng-SH (approved):
Deep-reviewed exact head `a017fcd1887057e62c1b201fada2a86fccd67401`. The fix closes the reported wait-for cycle on both paths: compatible stores perform provider-side copy before any CN source reader exists, while the fallback uses an independently constructed destination transport. The post-copy checksum hashes the stored destination bytes, retries independently, and the typed error prevents the outer retry from repeating a successful provider copy. I also traced the `S3FS`/`subPathFS`/`FileServices` capability and path resolution, reader ownership on read/write/close failures, cancellation, and the fallback single-read path; no blocking Q1-Q3 lifecycle findings.

Exact-head validation:
- all four new regressions enumerated; each passed independently with `-race -count=100`
- full `pkg/backup` and `pkg/fileservice` passed under race
- dependent `pkg/datasync.TestCopyFile` passed with `-race -count=100`
- CI Summary/UT/SCA/BVT/coverage passed
- current-main merge-tree is conflict-free

No blocking findings.

PR #26292, opened 2026-07-28
fix: support cyclic foreign keys in database branch

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26111

`DATA BRANCH CREATE DATABASE` currently rejects valid cyclic foreign-key graphs before cloning any object. This change:

- preserves topological ordering for acyclic schemas and uses deterministic forward-reference ordering only when a cycle is detected;
- scopes `foreign_key_checks=0` to that cyclic clone statement, including background compilation, and restores the original value on success or error;
- rewrites same-source-database FK references to the destination for both same-account and cross-account clones, then lets existing CREATE TABLE forward-reference handling backfill target table IDs;
- adds production-authentic regression coverage for live and named-snapshot sources, same-account and cross-account target-local FK metadata/enforcement, repeated-clone failure behavior, and session restoration.

Local validation on exact head `008b42c741792bd6b35782b1f3f34074e89e70b6`:

- `./pkg/frontend`: PASS
- `./pkg/sql/plan`: PASS
- `TestIssue26111DataBranchDatabaseWithCyclicForeignKeys`: PASS
- existing acyclic `TestIssue26118DatabaseCopiesPreserveHashIdentifiers`: PASS
- affected-package `go vet`: PASS
- targeted mutation falsification: the cross-account snapshot case fails when the destination-schema rewrite is removed
- schema v4 preflight: `review=PASS`, fresh remote CI pending

## QA handoff

- QA required: yes
- Reason: the fix changes persisted foreign-key metadata and runtime constraint behavior in Data Branch database cloning.
- Required scenarios: validate live and snapshot database branches from a valid mutual-FK schema, including a cross-account snapshot branch; verify cloned data, target-local `mo_foreign_keys` references, invalid-write rejection, and that a failed/repeated branch statement leaves `foreign_key_checks` enabled without altering the existing target.


XuPeng-SH (approved):
Deep-reviewed exact head b1690f383b588e9c8222894b07c52cfbfdb99fc5.

The change preserves the acyclic topological path and scopes foreign_key_checks=0 only to cyclic database clones, with restoration on success and failure. Forward references retain the source TableDef until the target appears, same-source FK database names are rewritten to the destination for same- and cross-account clones, and subscription aliases are resolved to the publisher database before that comparison. Mixed graphs remain safe under disabled checks, and unresolved references are backfilled by the existing create path. No steady-state production overhead is added outside database-clone setup. Focused frontend tests for cycle ordering and background-session variable inheritance pass under the deterministic CGo setup.

Issue #26113, opened 2026-07-23
[Bug]: CLONE and DATA BRANCH cannot use a table created earlier in the same transaction


Ariznawlll:
## Summary

`CREATE TABLE ... CLONE` and `DATA BRANCH CREATE TABLE` cannot use a table
created earlier in the same explicit transaction as their source, even though
ordinary queries in that transaction can read the new table.

## Tested version

`main` at `2eb7f523e5a10f352d3e258183ed8440510388de`

## Reproduction

```sql
DROP DATABASE IF EXISTS g4d_tx_repro;
CREATE DATABASE g4d_tx_repro;
USE g4d_tx_repro;

CREATE TABLE src(id INT PRIMARY KEY, v VARCHAR(20));
INSERT INTO src VALUES (1, 'a');

BEGIN;
DATA BRANCH CREATE TABLE c1 FROM src;

-- This succeeds and returns 1, so c1 is visible in the transaction.
SELECT COUNT(*) FROM c1;

-- This fails although c1 is visible above.
DATA BRANCH CREATE TABLE c2 FROM g4d_tx_repro.c1;
ROLLBACK;
```

The same problem is visible in the underlying clone path:

```sql
BEGIN;
CREATE TABLE c1 CLONE src;
SELECT COUNT(*) FROM c1;
CREATE TABLE c2 CLONE g4d_tx_repro.c1;
ROLLBACK;
```

Both variants were repeated three times with fresh databases.

## Actual result

The first table is readable, but the second create fails:

```text
c1_visible
1
ERROR 1064 (HY000): SQL parser error: table g4d_tx_repro-c1 does not exist
```

The rollback removes the first table and leaves no branch metadata from the
transaction.

## Expected result

The second clone/branch create should resolve the table created earlier in the
same transaction, consistently with ordinary query visibility. Both tables
should commit or roll back together.

## Code analysis

`dataBranchCreateTable` delegates to `handleCloneTable`. Without an explicit
snapshot, `handleCloneTable` creates an `MO_TS` hint using
`tryToIncreaseTxnPhysicalTS` and calls `BackgroundExec.ExecRestore`.

The restore source lookup reads the catalog at that timestamp and does not
include the transaction's newly created source table, while an ordinary query
through the shared transaction can already read it. The source-resolution path
therefore has different read-your-own-DDL semantics from the surrounding
transaction.

## Additional checks

- Two independent `DATA BRANCH CREATE TABLE` statements in one transaction
[...]

PR #26293, opened 2026-07-28
fix: preserve transaction visibility for table clone

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26113

Clone restore currently injects a generated `MO_TS` even inside an explicit transaction. That forces source resolution through a cloned snapshot transaction, which cannot see a table created earlier by the shared transaction.

This change:

- keeps the generated timestamp for branch metadata/protect-snapshot bookkeeping, but does not inject that generated hint into restore SQL under explicit `BEGIN`;
- uses the same shared-transaction view for `COPY GRANTS` unless the user explicitly requested a snapshot;
- interprets a same-commit parent edge as having no parent-owned history before the parent creation commit;
- resolves direct and intermediate ancestors by stable table ID only when `edge TS < creation CTS <= endpoint snapshot`;
- resolves a dropped intermediate at the bounded lineage endpoint rather than the current snapshot;
- clamps LCA row probes to the LCA creation commit so post-commit DIFF never reads before the table exists.

User-provided snapshot hints remain intact for both data and grants, ordinary autocommit clones retain the existing generated-hint path, and historical ancestors continue to resolve at their original clone snapshot.

Regression coverage uses a real embedded cluster and one MySQL connection to verify:

- ordinary autocommit and explicit named-snapshot clones remain distinct after the source changes;
- explicit named-snapshot `COPY GRANTS` preserves snapshot data and privileges;
- raw CLONE, `COPY GRANTS`, and three-level Data Branch chains read their own uncommitted DDL;
- commit preserves exact `src -> b1 -> b2 -> b3` lineage and matching protect snapshots;
- direct DIFF and deep `b3 against src` DIFF after dropping intermediate `b1` return the expected counts;
- rollback removes tables, copied role privileges, `mo_branch_metadata`, and branch protect snapshots atomically.

Local validation on exact head `e4e1154df21a59df41e35e0b90d366f5f11f6578`:

- focused issue regression under `-race`: PASS (`T=20.739s`, `B=30s`, `N=1`)
- full `./pkg/tests/issues -race`: PASS (367.233s)
- exact `UnionBatch` regressions under `-race -count=100`: PASS
- full `./pkg/container/vector -race`: PASS (3.071s)
- full `./pkg/frontend -race`: PASS (27.134s)
- production package build and affected `go vet`: PASS
- BVT authored: `test/distributed/cases/git4data/clone/clone_can_rollback.sql` and `.result`; exact-head CI execution pending
- `git diff --check`: PASS
- rebase proof: all ten commits are patch-identical by `git range-diff`; the only upstream change in a touched production file is non-overlapping
- schema v4 preflight: `PASS review=PASS validation=PENDING`, diff hash `718fee756f387d451f79c27514863104d891973cb3ca27296c92c5349f2995bd`

## QA handoff

- QA required: yes
- Reason: this changes explicit-transaction visibility and the interpretation of persisted Data Branch timestamps during later DIFF.
- Required scenarios: on a deployed environment, use one connection and explicit `BEGIN` to build multi-level `CREATE TABLE ... CLONE`, `COPY GRANTS`, and `DATA BRANCH CREATE TABLE` chains; verify read-your-own-DDL, explicit snapshot data/grant preservation, commit lineage/protect snapshots, direct and deep post-commit DIFF after dropping an intermediate ancestor, and rollback cleanup of tables, privileges, metadata, and branch snapshots. Retain an autocommit control.


XuPeng-SH (changes requested):
Deep review found two transaction-visibility blockers:

[P1] Do not use the current snapshot as the only fallback for an uncommitted intermediate branch. `buildSideCollectRange` falls back from the pre-commit clone timestamp to `getRelationById(..., nil)`. If that intermediate table was later dropped, the current lookup cannot resolve its ID even though the historical lineage is still valid. I reproduced this on this exact head by extending `TestIssue26113CloneCreatedTableInSameTransaction`: after the existing b1 -> b2 -> b3 chain commits, `drop table b1; data branch diff b3 against src output count` fails with `internal error: can not find table by id ...` instead of returning 0. Please resolve the fallback at a stable snapshot at or after the table creation CTS and no later than the lineage endpoint, and cover the dropped-intermediate case.

[P1] `CREATE TABLE ... CLONE ... COPY GRANTS` is still not transaction-visible for a source table created earlier in the same explicit transaction. The restore SQL correctly stopped injecting the generated snapshot hint, but `copyGrantsSnapshotTS` still defaults to `snapshotTS`; `copyTablePrivileges` then uses snapshot-hinted source table-ID and role-privilege queries. That snapshot predates the uncommitted source clone, so the valid c1 -> c2 COPY GRANTS chain cannot see c1. Please use the shared transaction view when no user snapshot was requested, while retaining user-specified snapshot semantics, and add the same-transaction COPY GRANTS regression.

Validation: the original embedded issue_26113 path passed up to the added dropped-intermediate assertion, and `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/frontend` passed. Temporary review-only changes were removed.

VioletQwQ-0:
Addressed both P1 findings on exact head `78d74b3fc19e1beb476aee67caa8448f8ba5c1fa`:

1. Dropped intermediate lineage: `buildSideCollectRange` now resolves a same-transaction intermediate by immutable table ID at the lineage endpoint snapshot and accepts it only when `cloneTS < creationCTS <= endpointCTS`. The regression commits `b1 -> b2 -> b3`, drops `b1`, then verifies `DIFF b3 AGAINST src` returns 0.
2. Same-transaction `COPY GRANTS`: generated timestamps now use the shared transaction view for both restore and privilege lookup, while an explicitly requested snapshot still governs both. The regression verifies a `cg1 -> cg2 COPY GRANTS` chain propagates `SELECT ... WITH GRANT OPTION`; a distinguishable named-snapshot control reads the older one-row state and copied grant; rollback removes all created tables and privilege rows.

Final-head validation:

- focused regression: `-race -count=1` PASS and `-count=3` PASS
- full `./pkg/tests/issues -race`: PASS
- full `./pkg/frontend -race`: PASS
- production build, affected vet, and `git diff --check`: PASS
- schema v4 CR-update preflight: `review=PASS`; fresh remote CI is pending

The branch also incorporates current `main` (including the adjacent cyclic-FK clone change) and the full affected validation was rerun afterward. Please re-review.

XuPeng-SH (changes requested):
[P1] A descendant branch remains unusable when an intermediate branch is created and dropped in the same transaction. Repro: begin; create d1 from src; create d2 from d1; drop d1; commit; then diff d2 against src. The diff fails with can not find table by id because both the clone snapshot and endpoint CTS cannot resolve d1 when creation/drop share one commit. Please resolve this zero-history lineage case without requiring the intermediate relation to be visible at endpoint CTS, and add this transaction-level regression case.

VioletQwQ-0:
Addressed the current-head zero-history lineage CR in bb3e85457.

The failure was broader than relation lookup: simply skipping the same-CTS create/drop intermediate made unchanged DIFF pass but silently lost writes materialized into the descendant. The fix now requires definite stable-ID misses at both historical probes, then falls back to symmetric whole-endpoint history comparison so inherited rows cancel and intermediate writes remain visible. Other lookup errors still propagate.

Regression coverage now creates/clones/drops the intermediate in one transaction and verifies: deleted metadata plus live descendant lineage/protection, unchanged DIFF=0 in both directions, and a pre-clone inserted row DIFF=1 in both directions.

Validation on exact head: focused embedded regression x3 PASS; full pkg/frontend PASS; Data Branch frontend race PASS; go vet PASS; git diff --check PASS; schema-v4 preflight PASS (review=PASS, validation=PENDING). The embedded race attempt is separately blocked before DIFF by the repository-existing vector.Vector.UnionBatch checkptr failure during clone materialization.

VioletQwQ-0:
Addressed the current review blocker and the exact Ubuntu UT failure on head `126ffb9b2ccc047a1d2e39e6ff852319e77f8b05`.

- Rebased onto current `main` and integrated zero-history ancestry with the new schema-lineage validation path.
- A same-transaction parent with no committed catalog generation now reuses the already-materialized child schema only after the same strict two-miss/CTS proof; unrelated lookup errors still propagate.
- Fixed the `UnionBatch` varlena fast-path checkptr abort seen in the exact CI stack by replacing the one-past unsafe pointer walk with typed slice indexing.
- `TestIssue26113CloneCreatedTableInSameTransaction` passed twice with `-race`; full `pkg/frontend` and `pkg/tests/issues` packages passed with `-race`; focused vector fast-path tests passed normally and with `-race`; vet/compile/diff-check passed.
- Exact-head preflight: `PASS review=PASS validation=PENDING`, diff hash `19a6a5802539c166923e666060de9e249f8aaf8d23880f2e417e75e6b3b489be`.

Please re-review the new head.

XuPeng-SH (changes requested):
[P1] Handle a zero-history table when it is the LCA itself

The new fallback covers a created-and-dropped table only while walking one side of a path. It does not cover the same zero-history node when it is the LCA of two surviving sibling branches: `diffOnBase` resolves a non-endpoint LCA with `getRelationById` at the pre-commit clone timestamp and returns immediately on the missing relation, before `resolveSameTransactionDataBranchRelation` / `hasZeroHistory` can run.

I reproduced this on this exact head with an authenticated embedded cluster:

```sql
begin;
data branch create table issue_26113.f1 from issue_26113.src;
insert into issue_26113.f1 values (4, shared txn change);
data branch create table issue_26113.f2 from issue_26113.f1;
data branch create table issue_26113.f3 from issue_26113.f1;
drop table issue_26113.f1;
commit;
data branch diff issue_26113.f2 against issue_26113.f3 output count;
```

Both endpoints have identical inherited data, so the expected count is 0. Instead the last statement fails with `internal error: can not find table by id <f1-id>`. This is the same transaction-visibility class addressed by the PR, but in a branched DAG rather than the currently tested linear chain. Please make zero-history resolution apply systemically to LCA resolution as well and add a sibling-LCA regression test.

VioletQwQ-0:
Addressed the latest zero-history LCA review finding in fbedd960a204dd11d7cbd11305b6f408d7192cc2.

- Reproduced the exact sibling-LCA failure on the previous head before changing production code.
- Require endpoint-bounded stable-ID misses before classifying the LCA as zero-history.
- Preserve the original DAG for schema/range derivation, then compare whole endpoint histories without dereferencing the absent LCA relation.
- Added bidirectional identical-sibling assertions and bidirectional divergent-sibling assertions, plus deleted-LCA/live-child metadata checks.

Exact-head local validation:
- go test ./pkg/frontend -run Test.*DataBranch -count=1
- go test ./pkg/tests/issues -run ^TestIssue26113CloneCreatedTableInSameTransaction$ -count=3
- go test -race ./pkg/tests/issues -run ^TestIssue26113CloneCreatedTableInSameTransaction$ -count=2
- focused UnionBatch normal/race tests
- git diff --check
- preflight schema v4: PASS review=PASS, validation=PENDING

@XuPeng-SH please re-review the current head.

XuPeng-SH (approved):
Re-reviewed the updated head and the author's responses. The previous zero-history LCA blocker is now closed: classification requires stable-ID misses at the historical probe and both endpoint creation timestamps, schema/range construction handles the absent ancestor without swallowing unrelated errors, and the symmetric whole-endpoint comparison preserves inherited and divergent rows. The full authenticated issue regression (including sibling LCA, dropped intermediates, COPY GRANTS, commit/rollback) passes locally, as do the focused UnionBatch fast-path tests under -race.

XuPeng-SH (approved):
Re-reviewed the updated head, including the author’s response to the zero-history LCA blocker. The prior nine commits are patch-equivalent across the rebase; the new LCA fallback is guarded by missing stable-ID probes and endpoint CTS bounds, preserves schema/range derivation, and compares endpoint histories symmetrically. The embedded issue regression (including sibling LCA, linear lineage, COPY GRANTS, and rollback cases) and focused frontend Data Branch suite pass locally. No remaining blocking findings.

Issue #26131, opened 2026-07-24
[Optimizer] TPCH Q15 executes the same multi-reference CTE twice


XuPeng-SH:
## Problem

The canonical TPCH Q15 shape defines `q15_revenue0` once but references it both in the supplier join and in the scalar `MAX(total_revenue)` subquery. MatrixOne inlines and fully executes the CTE twice, including two full scans and two identical grouped aggregates over `lineitem`.

## Reproduction

Environment: local standalone MatrixOne at commit `9c2e84ad5c`, TPCH SF10.

```sql
with q15_revenue0 as (
  select l_suppkey as supplier_no,
         sum(l_extendedprice * (1 - l_discount)) as total_revenue
  from lineitem
  where l_shipdate >= date '1995-12-01'
    and l_shipdate < date '1995-12-01' + interval '3' month
  group by l_suppkey
)
select s_suppkey, s_name, s_address, s_phone, total_revenue
from supplier, q15_revenue0
where s_suppkey = supplier_no
  and total_revenue = (select max(total_revenue) from q15_revenue0)
order by s_suppkey;
```

`EXPLAIN ANALYZE` contains two independent copies of the same subtree. Each copy:

- scans all 59,986,052 `lineitem` rows / 7,323 blocks
- applies the same three-month date filter and outputs 2,265,119 rows
- groups by `l_suppkey` and outputs the same 100,000 groups
- consumes about 1.34 GiB input and 83.81 MiB scan memory

Measured cumulative operator data:

- first scan: 670 ms; grouped aggregate: 1,143 ms; aggregate memory: 71.34 MiB
- second scan: 678 ms; grouped aggregate: 558 ms; aggregate memory: 71.34 MiB

The second copy exists only to calculate `MAX(total_revenue)` from the same CTE result.

## Expected behavior

For a deterministic CTE referenced multiple times, the optimizer should cost alternatives and reuse/materialize the result when that is cheaper than recomputation. Q15 should calculate the 100,000-row revenue relation once and feed both consumers from it.

Possible implementations include shared CTE materialization, common-subplan reuse, or a Q15-equivalent rewrite that computes the maximum from the already-produced revenue stream.

Please add a plan-shape regression test that verifies the base `lineitem` aggregate is not duplicated for this multi-reference CTE.

VioletQwQ-0:
[...]

PR #26295, opened 2026-07-28
perf(plan): reuse eligible multi-reference CTEs

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26131

Reuse only deterministic, profitable, fixed-width, bounded multi-reference CTEs through one local materialized producer and independent full-drain readers. Unsupported or uncertain shapes remain inline.

The materialized source keeps the first 64 MiB in the query MPool and spills later batches to an anonymous query-scoped local file, so underestimated planner statistics cannot turn a formerly successful inline query into a runtime-limit error. Each reader owns its decoded spill batch until it advances or releases; the final producer/reader release reclaims the shared batches, per-reader batches, and spill file.

The branch is composed onto current `main`. Its superseded pipeline-terminal commit was dropped because #26250 now provides the durable terminal-state implementation; only the materialized Dispatch reset branch is composed onto that upstream behavior.

Validation on exact head `aa815b68971e29daff1ea86587ec4cac1e7b4eef`:

- planner/materialized/merge/compile focused package tests
- Source and Merge race tests
- embedded TPCH Q15-shaped execution with a real grouped producer, join consumer, scalar `MAX` consumer, two `Sink Scan` nodes, expected result, EOF, `rows.Err()` and `Close()` checks
- targeted mutation: restoring the old 64 MiB hard error makes the overflow regression fail
- affected-package vet and `git diff --check`
- schema v4 preflight: `PASS review=PASS validation=PENDING`, diff hash `8cc8631ac675171d883041af41d62493a9a0b52bfbec949dfd3bf99990f1c75c`

Prior-head CI classification:

- SCA `copyloopvar` was PR-caused and is fixed on this head.
- Ubuntu UT failed in unrelated TAE `TestCkpLeak` with `panic: ExpectedEOB`; this PR's planner tests passed in that run.
- Fresh exact-head CI is pending.

Performance status:

- Current-head SF10 A/B/A is PENDING because both available hosts failed the idle/ownership gate (local user mo-service; mo-55 long-running MatrixOne container).
- A superseded-base strict A/B/A showed about 49.9% median improvement and the expected one-producer/two-consumer plan, but that result is supporting evidence only and is not claimed as current-head proof.


XuPeng-SH (changes requested):
Deep-reviewed exact head aa815b68971e29daff1ea86587ec4cac1e7b4eef. The materialized Source producer/reader ownership, cancellation, prepared-reuse generation, and spill cleanup paths are otherwise coherent. One P1 semantic blocker remains.

[P1] SAMPLE is treated as deterministic and can be shared (pkg/sql/plan/cte_reuse.go, cteSubtreeIsDeterministic).

The node-type denylist excludes volatile functions but not plan.Node_SAMPLE, whose executor uses random selection. I reproduced this on the exact head with an expensive fixed-width producer:

    with c as (
      select sample(total, 2 rows) as n from (
        select l_suppkey, sum(l_extendedprice) as total
        from lineitem group by l_suppkey
      ) grouped
    ) select a.n from c a join c b on a.n = b.n;

Before this optimization the plan has two independent SAMPLE nodes. On this head the cost guard wins and rewrites it to one materialized SINK plus two SINK_SCAN consumers. The two independent random samples therefore become the same sample; joins, intersections, aggregates, and returned rows can change.

Please classify Node_SAMPLE as nondeterministic and add this profitable shape to TestCTEMultiReferenceReuseGuards (asserting zero reachable SINK_SCAN). Focused CTE race tests and the materialized/dispatch/merge/compile race packages passed; the full plan race package was killed by the local resource limit, while its focused CTE race suite passed.

VioletQwQ-0:
Superseded by #26331, which is now the canonical PR for #26131. The successor removes the disk-spill experiment, keeps bounded query-MPool materialization, and includes the reviewer-reported profitable SAMPLE SQL as a production planner guard. The guard fails on this reviewed head with two reachable SINK_SCAN nodes and passes on #26331 with zero, preserving independent sampling semantics. Closing this duplicate; the branch is retained.

PR #26297, opened 2026-07-28
fix(sql): support DROP TEMPORARY TABLE

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25404 

fix(sql): support DROP TEMPORARY TABLE


heni02 (approved):

XuPeng-SH (changes requested):
Requesting changes for two lifecycle-closure blockers around persistent side effects.

[P1] Classify the resolved target, not the spelling of DROP — `pkg/frontend/authenticate.go:11760`, `pkg/frontend/connector.go:241`

The new guards only check `DropTable.Temporary`, but MatrixOne already resolves an ordinary `DROP TABLE t` to the session temporary table when `t` shadows a permanent table. I confirmed this with a temporary focused planner test: parsing `drop table nation` against a temporary `nation` produces a `DropTable` plan whose `TableDef.IsTemporary` is true. The executor therefore drops the temporary object, while the AST flag remains false. Before execution `doRevokePrivilegeImplicitly` can consequently revoke ownership from the same-named permanent table, and after execution `handleDropDynamicTable` can cancel the daemon task for a same-named permanent dynamic table. The two supported ways to drop the same temporary target now have different and potentially destructive side effects.

Please derive/capture the actual per-target classification from the resolved plan or session mapping and use it consistently for these hooks. This must be captured before execution because the temporary alias is removed by `dropTableSingle`, and multi-table DROP needs to filter each target independently. Add shadowing coverage for both `DROP TABLE` and `DROP TEMPORARY TABLE`, plus a mixed multi-target case.

[P1] Skip implicit ownership grant when creating a temporary table — `pkg/frontend/status_stmt.go:93-112`, `pkg/frontend/authenticate.go:11726-11750`

The new revoke guard states that temporary tables have no persistent ownership metadata, but the create side still unconditionally calls `doGrantPrivilegeImplicitly` for every `CreateTable`. For a temporary table it builds `grant ownership on table db.<logical-alias>` even though the engine table has a generated physical name. For a non-admin session, if a permanent table is shadowed this resolves the grant to that permanent table; without a permanent table the post-create grant can fail after the temporary object was already created. The new BVT runs as an admin and cannot exercise this path. Please make create/revoke share the same temporary-target contract and add a non-admin shadowing test proving the permanent owner is unchanged.

The parser generation is reproducible and the core planner/executor fall-through protection is otherwise sound.

Validation on exact head `9360cfcc8ba15308292151dd7356b467d6c72739`:
- all new focused tests pass normally and under race (34x frontend/compile, 37x plan, 58x parser)
- full `frontend`, `sql/compile`, `sql/plan`, parser tree, and MySQL parser packages pass under race
- `go vet` and `git diff --check` pass
- regenerating `mysql_sql.go` produces no diff
- prospective merge with latest `origin/main` (`1b6d4308036f69e0486420fb175ab3de3a5c4ae2`) is clean; merged focused race tests pass

iamlinjunhong:
> The new guards only check `DropTable.Temporary`, but MatrixOne already resolves an ordinary `DROP TABLE t` to the session temporary table when `t` shadows a permanent table. I confirmed this with a temporary focused planner test: parsing `drop table nation` against a temporary `nation` produces a `DropTable` plan whose `TableDef.IsTemporary` is true. The executor therefore drops the temporary object, while the AST flag remains false. Before execution `doRevokePrivilegeImplicitly` can consequently revoke ownership from the same-named permanent table, and after execution `handleDropDynamicTable` can cancel the daemon task for a same-named permanent dynamic table. The two supported ways to drop the same temporary target now have different and potentially destructive side effects.

fixed

XuPeng-SH (approved):
Deep-reviewed exact head c1219fcee02382719d7d0fe103b545358e699b90. The two prior P1 issues are closed: pre-execution per-target classification now prevents temporary aliases from triggering persistent ownership/dynamic-task side effects, including mixed DROP lists, and CREATE TEMPORARY TABLE skips persistent ownership grants. Planner/executor guards prevent DROP TEMPORARY or stale plans from falling through to same-named persistent tables. Affected package tests passed for ./pkg/frontend ./pkg/sql/compile ./pkg/sql/plan ./pkg/sql/parsers/tree ./pkg/sql/parsers/dialect/mysql.

PR #26298, opened 2026-07-28
fix(sysview): hide temporary tables from information schema

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25405

fix(sysview): hide temporary tables from information schema


XuPeng-SH (changes requested):
Requesting changes for two metadata-contract blockers.

[P1] Do not classify every legal `__mo_tmp_` name as an internal temporary object — `pkg/util/sysview/predefined.go:311`

The new predicate hides every row whose name starts with `__mo_tmp_`, but that namespace is not reserved by CREATE TABLE. I verified with a temporary focused planner test that `create table __mo_tmp_customer (id int)` binds successfully as a non-temporary table. After this view upgrade that permanent table disappears from `information_schema.tables`. The new BVT only checks `xxmo_tmp_visible`, which catches wildcard underscore mistakes but does not exercise the actual prefix collision.

Please establish one enforceable invariant before relying on the prefix: either reject/reserve the internal namespace for non-temporary CREATE/rename/clone/restore paths with a clear compatibility strategy for existing tables, or persist/use an explicit temporary marker (or another classifier that cannot hide arbitrary user tables). Add an exact-prefix permanent-table test. This should be shared with the other prefix consumers rather than defining a view-only approximation.

[P1] Hide the same physical objects from the rest of information_schema metadata — `pkg/util/sysview/predefined.go:202-205`, `:402-406`, `:528-537`

Changing only `information_schema.TABLES` leaves the underlying temporary base/index rows visible through `information_schema.COLUMNS`, `STATISTICS`, and `TABLE_CONSTRAINTS`, whose definitions still join `mo_tables`/`mo_columns`/`mo_indexes` without a temporary-object predicate. The PR BVT now creates an indexed temporary table, so these leaks are deterministic, but it queries only TABLES. Users and schema tools can still discover the same `__mo_tmp_<session>...` names and internal index metadata through those views.

Please apply the same canonical classifier to every metadata view that can expose temporary base/index objects, add the corresponding versioned MODIFY_VIEW entries, and extend the BVT to assert no temporary names appear in TABLES/COLUMNS/STATISTICS/TABLE_CONSTRAINTS while ordinary permanent metadata remains visible. PARTITIONS does not need special handling while temporary partitions remain unsupported.

The TABLES predicate itself, the v4.0.5 offset-based upgrade path, clone-safe DDL generation, and latest-main merge are otherwise sound.

Validation on exact head `2d3aae27d16e3ba4e5ef188343181a70721f1d5f`:
- all new focused bootstrap/sysview tests pass normally
- focused tests pass under race (67x bootstrap, 68x sysview)
- full v4.0.5 bootstrap, sysview, and sql/plan packages pass under race
- `go vet` and `git diff --check` pass
- prospective merge with latest `origin/main` (`1b6d4308036f69e0486420fb175ab3de3a5c4ae2`) is clean; merged focused race tests pass

iamlinjunhong:
> Requesting changes for two metadata-contract blockers.
> 
> [P1] Do not classify every legal `__mo_tmp_` name as an internal temporary object — `pkg/util/sysview/predefined.go:311`
> 
> The new predicate hides every row whose name starts with `__mo_tmp_`, but that namespace is not reserved by CREATE TABLE. I verified with a temporary focused planner test that `create table __mo_tmp_customer (id int)` binds successfully as a non-temporary table. After this view upgrade that permanent table disappears from `information_schema.tables`. The new BVT only checks `xxmo_tmp_visible`, which catches wildcard underscore mistakes but does not exercise the actual prefix collision.
> 
> Please establish one enforceable invariant before relying on the prefix: either reject/reserve the internal namespace for non-temporary CREATE/rename/clone/restore paths with a clear compatibility strategy for existing tables, or persist/use an explicit temporary marker (or another classifier that cannot hide arbitrary user tables). Add an exact-prefix permanent-table test. This should be shared with the other prefix consumers rather than defining a view-only approximation.
> 
> [P1] Hide the same physical objects from the rest of information_schema metadata — `pkg/util/sysview/predefined.go:202-205`, `:402-406`, `:528-537`
> 
> Changing only `information_schema.TABLES` leaves the underlying temporary base/index rows visible through `information_schema.COLUMNS`, `STATISTICS`, and `TABLE_CONSTRAINTS`, whose definitions still join `mo_tables`/`mo_columns`/`mo_indexes` without a temporary-object predicate. The PR BVT now creates an indexed temporary table, so these leaks are deterministic, but it queries only TABLES. Users and schema tools can still discover the same `__mo_tmp_<session>...` names and internal index metadata through those views.
> 
> Please apply the same canonical classifier to every metadata view that can expose temporary base/index objects, add the corresponding versioned MODIFY_VIEW entries, and extend the BVT to assert no temporary names appear in TABLES/COLUMNS/STATISTICS/TABLE_CONSTRAINTS while ordinary permanent metadata remains visible. PARTITIONS does not need special handling while temporary partitions remain unsupported.
> 
> The TABLES predicate itself, the v4.0.5 offset-based upgrade path, clone-safe DDL generation, and latest-main merge are otherwise sound.
> 
> Validation on exact head `2d3aae27d16e3ba4e5ef188343181a70721f1d5f`:
> 
> * all new focused bootstrap/sysview tests pass normally
> * focused tests pass under race (67x bootstrap, 68x sysview)
> * full v4.0.5 bootstrap, sysview, and sql/plan packages pass under race
> * `go vet` and `git diff --check` pass
> * prospective merge with latest `origin/main` (`1b6d4308036f69e0486420fb175ab3de3a5c4ae2`) is clean; merged focused race tests pass

fixed

XuPeng-SH (changes requested):
The two blockers from my previous review are still present on the current head. The branch remains at `2d3aae27d16e3ba4e5ef188343181a70721f1d5f`; the “fixed” reply was not followed by a new commit.

[P1] The view still classifies temporary objects solely by the broad `__mo_tmp_` prefix. That prefix is accepted for ordinary CREATE TABLE names, so a permanent table such as `__mo_tmp_customer` is silently removed from `information_schema.tables`. The underlying cause is visible in the create path: the ordinary `relkind` property is built before `TableDef.TableType` is changed to `temporary_table`, so the durable catalog marker remains `r`. Please establish a catalog-level invariant—preferably persist a reliable temporary-object marker and consume it from metadata views—or formally reserve and enforce the namespace across create/rename/clone/restore with a compatibility plan. Add the exact-prefix permanent-table regression; `xxmo_tmp_visible` only validates LIKE escaping.

[P1] Only `information_schema.TABLES` is changed. `COLUMNS`, `STATISTICS`, and `TABLE_CONSTRAINTS` still join the same temporary base/index rows without a temporary-object predicate, so schema tooling still sees the internal physical names and index metadata. Use one canonical classifier for all affected metadata views, add the required versioned MODIFY_VIEW entries, and assert the indexed temporary table is absent from each view while permanent metadata remains visible.

Validation on the exact head:
- bootstrap v4.0.5 and sysview packages pass under race
- six affected planner/view/show/create paths pass under race
- full planner package passes without race
- go vet and git diff --check pass
- CI is green, but it does not exercise these contract failures

The PR is also behind current main; rebase after fixing the blockers.

gouhongshen (changes requested):
## Codex automated review

Earlier blockers are fixed, but pre-upgrade temporary tables can reappear and be cloned during asynchronous upgrade.

### P1 - Preserve the legacy temporary-table exclusion during the asynchronous upgrade (pkg/frontend/snapshot.go:1971)
Before this PR, CREATE TEMPORARY TABLE left the persisted relkind as ordinary ('r'). Existing v4.0.4 temporary rows therefore have __mo_tmp_ physical names but no temporary_table marker. The v4.0.5 async tenant upgrade installs marker-only view filters, and this query likewise drops the prior name filter, so live legacy temporary rows become visible and can be included in data-branch/snapshot restore. Retain a compatibility classifier until such sessions cannot exist, or explicitly drain/migrate them; add an upgrade regression.

iamlinjunhong:
> ## Codex automated review
> Earlier blockers are fixed, but pre-upgrade temporary tables can reappear and be cloned during asynchronous upgrade.
> 
> ### P1 - Preserve the legacy temporary-table exclusion during the asynchronous upgrade (pkg/frontend/snapshot.go:1971)
> Before this PR, CREATE TEMPORARY TABLE left the persisted relkind as ordinary ('r'). Existing v4.0.4 temporary rows therefore have __mo_tmp_ physical names but no temporary_table marker. The v4.0.5 async tenant upgrade installs marker-only view filters, and this query likewise drops the prior name filter, so live legacy temporary rows become visible and can be included in data-branch/snapshot restore. Retain a compatibility classifier until such sessions cannot exist, or explicitly drain/migrate them; add an upgrade regression.

fixed

gouhongshen (changes requested):
## Codex automated review

The prior metadata-contract blockers are fixed. The legacy-upgrade blocker is only partially fixed: non-canonical SQL formatting can still expose legacy temporary tables and copy them during snapshot/clone enumeration.

### P1 - Legacy temporary tables with non-canonical whitespace bypass the compatibility filter (pkg/catalog/temp_table.go:50)
Author response: the classifier recognizes legacy rows via `lower(ltrim(rel_createsql)) LIKE 'create temporary table%'`. Why this remains blocking: `rel_createsql` stores raw request SQL, while `ltrim` removes only spaces and the pattern requires exactly one space between keywords. Legal input such as `CREATE TEMPORARY\nTABLE t (...)` or `CREATE TEMPORARY  TABLE ...` is not recognized; the base row remains relkind `r`, so the fallback also skips it. The upgraded information_schema views expose it, and snapshot/data-branch enumeration can copy it. Normalize or parse legacy CREATE SQL and add a formatting regression test.

gouhongshen (changes requested):
## Codex automated review

The original prefix-collision and incomplete metadata-view blockers are fixed. The prior whitespace examples are also fixed, but the legacy classifier remains blocking because it classifies request-wide SQL rather than the statement that created each table. Focused tests could not run because cgo/libmo.dylib is absent.

### P1 - Classify the creating statement, not the entire COM_QUERY (pkg/catalog/temp_table.go:51)
Author response: a7311f64 replaces the prior ltrim/LIKE check with a whitespace-tolerant regexp and tests newlines, tabs, and repeated spaces; those exact cases are fixed. Why this remains blocking: rel_createsql comes from ctx.GetRootSql(), while TxnCompilerContext.GetRootSql() falls back to ses.GetSql(), which doComQuery sets to the entire multi-statement input. For the supported request `CREATE TEMPORARY TABLE t(...); CREATE TABLE p(...)`, permanent table p therefore has relkind `r` but rel_createsql beginning with `CREATE TEMPORARY TABLE`, so this predicate hides p from SHOW/information_schema and excludes it from snapshot/data-branch enumeration. Reversing the statements makes an old-version temporary row miss the anchored regexp; legal comments between keywords do likewise. Persist or classify statement-specific canonical DDL with a migration/drain contract for legacy rows, and cover both multi-statement orders and comments.

gouhongshen (changes requested):
## Codex automated review

The earlier broad-prefix collision and incomplete information_schema coverage blockers are fixed: all four metadata views now use the shared predicate and are upgraded. The prior legacy SQL-format and request-wide-statement blocker is also addressed for normal names by parsing statements and persisting canonical per-statement DDL. One distinct legacy-name impersonation blocker remains.

### P1 - Do not hide a permanent legacy table that mimics a temporary physical name (pkg/sql/plan/function/func_mo_is_legacy_temporary_table.go:103)
Author response: `cdd5f3cc7` replaces the request-prefix regexp with parsed statement/alias matching, and newly created tables persist statement-specific canonical DDL. Why this remains blocking: pre-upgrade rows still retain request-wide `rel_createsql` and lack a durable temporary marker. Before upgrading, a legal request such as `CREATE TEMPORARY TABLE t (...); CREATE TABLE __mo_tmp_123e4567e89b12d3a456426614174000_db_t (...)` creates a permanent `relkind = 'r'` row whose stored SQL contains the temporary statement (the old create path used `GetRootSql`). The new function decodes that permanent name to logical alias `t` and returns true on the first statement at line 103, so the shared predicate removes the permanent table from SHOW/information_schema and snapshot/data-branch enumeration. This is distinct from the earlier broad-prefix finding: it requires the exact generated-name layout plus a matching legacy multi-statement request, and is introduced by the new statement-association classifier. Close it with a non-spoofable durable classifier/migration or an enforced namespace contract that also safely handles existing permanent rows; add this legacy regression.

gouhongshen (approved):
## Codex automated review

All previous blockers are resolved: durable relkind classification fixes broad-prefix collisions; all four metadata views and upgrade paths are covered; statement-specific DDL fixes formatting/request-wide ambiguity; and the latest change preserves temp-shaped permanent tables using explicit CREATE precedence and rename metadata. Focused tests could not run because cgo/libmo.dylib is absent.

XuPeng-SH (changes requested):
Reviewed updated exact head `b7d40a627fc99b8ab87cde9b20d160df92e8894e`. The previous metadata-view, legacy-format, multi-statement, and permanent-name collision blockers are fixed. One direct DML regression from the new durable marker remains.

### P1 — `INSERT IGNORE` into every newly created temporary table now fails

`MarkTableDefTemporary` changes the resolved `TableDef.TableType` from ordinary `r` to `temporary_table`. The modern insert binder at `pkg/sql/plan/bind_insert.go:1287-1292` treats every non-ordinary/non-index table type as a vector/text index table when the statement is not ODKU. Plain INSERT happens to survive because `build.go:83-86` permits its legacy fallback, but INSERT IGNORE is encoded as a one-element nil `OnDuplicateUpdate`; it is neither ODKU nor eligible for that fallback. The user therefore receives `unsupported DML: insert into vector/text index table` for a normal temporary table.

I reproduced this deterministically by marking the existing planner mock table with the new catalog helper, setting its resolved session `IsTemporary` bit, and planning `insert ignore into nation values (...)`. The focused new relkind, canonical-DDL, and legacy-classifier tests pass, while this probe fails exactly at the guard above.

Please treat `SystemTemporaryTable` / `IsTemporary` as a normal user DML target in the modern insert routing (without opening the internal vector/index table types), and add planner plus BVT coverage for plain INSERT, INSERT IGNORE, and ODKU against a temporary table with the persisted marker.

iamlinjunhong:
> Reviewed updated exact head `b7d40a627fc99b8ab87cde9b20d160df92e8894e`. The previous metadata-view, legacy-format, multi-statement, and permanent-name collision blockers are fixed. One direct DML regression from the new durable marker remains.
> 
> ### P1 — `INSERT IGNORE` into every newly created temporary table now fails
> `MarkTableDefTemporary` changes the resolved `TableDef.TableType` from ordinary `r` to `temporary_table`. The modern insert binder at `pkg/sql/plan/bind_insert.go:1287-1292` treats every non-ordinary/non-index table type as a vector/text index table when the statement is not ODKU. Plain INSERT happens to survive because `build.go:83-86` permits its legacy fallback, but INSERT IGNORE is encoded as a one-element nil `OnDuplicateUpdate`; it is neither ODKU nor eligible for that fallback. The user therefore receives `unsupported DML: insert into vector/text index table` for a normal temporary table.
> 
> I reproduced this deterministically by marking the existing planner mock table with the new catalog helper, setting its resolved session `IsTemporary` bit, and planning `insert ignore into nation values (...)`. The focused new relkind, canonical-DDL, and legacy-classifier tests pass, while this probe fails exactly at the guard above.
> 
> Please treat `SystemTemporaryTable` / `IsTemporary` as a normal user DML target in the modern insert routing (without opening the internal vector/index table types), and add planner plus BVT coverage for plain INSERT, INSERT IGNORE, and ODKU against a temporary table with the persisted marker.

fixed

XuPeng-SH (approved):
Re-reviewed at f1ae36d5ff26c1480ed7bacba3537ea047137a62. The prior temporary-table INSERT blocker is closed: marked/session-resolved temporary tables now stay on the ordinary DML path, while internal index table types remain rejected. Planner coverage includes plain INSERT, INSERT IGNORE, and ODKU, with matching distributed regression coverage. I also rechecked the durable marker/legacy catalog classification and upgrade-view changes. Relevant catalog, planner/function, sysview, bootstrap, frontend, and disttae test packages pass locally.

heni02 (approved):

PR #26301, opened 2026-07-28
fix: escape FK dep SQL queries to handle quoted identifiers

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26144

`getFkDeps`, `getFkDepsInPitrRestore`, and `getFkDepsWithTS` interpolated database and table names directly into SQL string literals. A legal backtick-quoted identifier containing a quote or backslash could therefore make CLONE, snapshot restore, or PITR restore issue malformed internal SQL and abort.

Changes:
- Encode every FK dependency filter with `quoteSQLStringLiteral`.
- Exercise the production topology callers for ordinary CLONE/snapshot, PITR, and cross-account `ExecRestore` paths.
- Assert the exact single SQL statement and the parent-before-child dependency order for quote/backslash-bearing names.

Validation:
- Targeted mutation of the old interpolation made all three regression paths fail; restoring the fix made them pass.
- `TestDataBranchAuditFkDepsEscapesQuotedNames` passed.
- Focused race stress passed with `T=0.03s`, `B=30s`, `N=100`.
- `pkg/frontend` passed full race and non-race tests, build, vet, and `git diff --check`.
- Preflight: `PASS review=PASS validation=PENDING` on `383968e0d23cd8b899b9f9a2be73ed79c5a9224c`.

QA Decision
- required: yes
- reason: user-visible CLONE, snapshot restore, PITR, catalog, and cross-account behavior is affected
- production_entrypoint: FK dependency discovery used by CLONE and restore topology
- automated_terminal_coverage: all three topology callers assert the emitted SQL and dependency order
- remaining_validation: deployed SQL/BVT coverage for complete CLONE, snapshot restore, and PITR operations with quoted identifiers


XuPeng-SH (approved):
Deep-reviewed at head 383968e0d23cd8b899b9f9a2be73ed79c5a9224c. All three FK-dependency query paths now cross the SQL string-literal boundary through the established escaping helper, without changing semantic identifier handling. Full pkg/frontend tests, the targeted quoted-name regression, go vet, and diff checks passed locally.

PR #26302, opened 2026-07-28
fix(lockservice): avoid false deadlocks on converging waits

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26195

The distributed deadlock detector used a breadth-first traversal and treated a transaction that had already been processed on another branch as a cycle. Acyclic wait graphs with converging branches could therefore abort a transaction with a false `deadlock detected` error.

Use iterative three-color DFS so only an edge to a transaction on the active DFS stack is a cycle. Fully explored transactions are safely pruned, each transaction is fetched at most once, and the closing edge retains its `WaiterAddress` for remote victim aborts. Cycle logging now contains the exact cycle without the acyclic entry path.

The regression tests cover both acyclic branch reconvergence and a real cycle spanning two discovery branches.

Tested with:

`./.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s ./pkg/lockservice`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=240s ./pkg/lockservice`

Six focused deadlock tests also passed individual adaptive `-race -count=100` stress runs.


XuPeng-SH (approved):
Approved on exact head `dcdbd2d78ec58b713df6ae6bbf866a69de5b88a9`.

The old global processed-node check demonstrably misclassified DAG branch reconvergence as a cycle. The iterative three-color DFS fixes that distinction, keeps the closing edge and its `WaiterAddress` for victim abort, prunes only fully explored nodes, and preserves cancellation/close behavior. I also compared the implementation against an independent three-color DFS oracle over 10,000 deterministic random graphs (including self-loops, duplicate edges, reconvergence, and cross-branch cycles) with no mismatch.

Validation: both new tests passed under `-race` 68x each; the complete `pkg/lockservice` package passed under `-race`; vet is clean; latest `origin/main` (`159742db6672ff5dadb617880b831da026f0a3f4`) merges cleanly and the focused race tests pass on the prospective merge.

Scope note: this fixes a real detector defect, but the archived #26195 run has no wait-for graph, so it does not prove this reconvergence shape was the unique cause of that specific TPCC 20701. The PR appropriately says `issue #26195` rather than auto-closing it.

Issue #26280, opened 2026-07-28
[Bug] Fulltext post-DML cannot build a valid statement for quoted string primary keys


Ariznawlll:
## Summary

Fulltext post-DML generation does not preserve a character primary key that contains a single quote. The generated maintenance statement is not a valid representation of the original key.

## Reproduction

At commit `c8e443bf8610b2af4333ad943ba2075485b16cf8`, use a `varchar` primary key with the value `a'b` and construct the fulltext delete maintenance statement.

The package-level test repeats this case three times:

```text
go test ./pkg/sql/colexec/postdml -run TestAuditFullTextStringKeyPreservesQuotedLiteral -count=3 -v
```

Each run produces:

```sql
DELETE FROM `d`.`idx` WHERE doc_id IN ('a'b')
```

The value needs to be represented as one quoted literal, for example:

```sql
DELETE FROM `d`.`idx` WHERE doc_id IN ('a''b')
```

## Expected behavior

Post-DML generation should encode character primary-key values so the generated fulltext maintenance statement denotes the exact same key as the source row.

## Actual behavior

`runPostDml` wraps character values with a leading and trailing quote but does not encode quotes already present in the value.

## Impact

The source-row change and the follow-up fulltext maintenance no longer refer to the same primary key. This can leave fulltext results out of sync with table rows for affected keys.

## Location

`pkg/sql/colexec/postdml/postdml.go`, `runPostDml`.

VioletQwQ-0:
## QA handoff

Merged fix: #26306, merge commit `82c164a83ae83af98e85a0bfd8fdab94327bd72f` (final PR head `f2edf361e2a5a35a909d2cc724b31289c298acd4`).
Current `main` at handoff: `41332062a6740978f269088a92ee362a888c7bd7`.

QA Decision
- required: yes
- reason: the change affects user-visible fulltext index maintenance, data consistency, and session SQL-mode behavior
- production_entrypoint: fulltext-indexed INSERT/UPDATE/DELETE operations whose maintenance SQL is generated by PostDml and executed by Compile
- automated_terminal_coverage: merged regression tests parse the generated DELETE and INSERT with the production MySQL scanner and verify that the original key is preserved under the default SQL mode and `NO_BACKSLASH_ESCAPES`, including resolver and remote-session mode sources
- remaining_validation: run deployed fulltext DML in both SQL modes and verify that the indexed row can be inserted, updated, deleted, and queried without a syntax error or stale/wrong index entry

[...]

PR #26306, opened 2026-07-28
fix(postdml): escape character primary keys in fulltext maintenance SQL

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26280

`runPostDml` generated fulltext DELETE and INSERT statements from primary-key values. Plain wrapping failed for quotes, while always doubling backslashes produced the wrong key when `NO_BACKSLASH_ESCAPES` was active.

Changes:
- Preserve the existing `sqlquote.String` behavior in the default SQL mode.
- Under `NO_BACKSLASH_ESCAPES`, leave backslashes unchanged and double only single quotes.
- Resolve `sql_mode` from the session resolver and fall back to captured `SessionInfo.SqlMode` for remote processes.
- Validate both generated DELETE and INSERT statements with the production MySQL scanner under the exact SQL mode.

Validation:
- Targeted mutation back to unconditional `sqlquote.String` made both resolver and remote-captured `NO_BACKSLASH_ESCAPES` cases fail; restoring the mode-aware fix passed.
- Default and `NO_BACKSLASH_ESCAPES` modes pass for both resolver and captured mode sources.
- DELETE asserts exactly one key literal; DELETE and INSERT both decode to the original key.
- Focused race stress passed with `T=0.01s`, `B=30s`, `N=100`.
- `pkg/sql/colexec/postdml` passed full race and non-race tests, build, vet, and `git diff --check` after the conflict-free rebase.
- Preflight: `PASS review=PASS validation=PENDING` on `563b0805c1cca7442f7da5909c106d3a825effe7`.

QA Decision
- required: yes
- reason: fulltext index data consistency and session SQL-mode behavior are user-visible
- production_entrypoint: PostDml-generated fulltext maintenance DELETE and INSERT executed by Compile
- automated_terminal_coverage: generated statements are decoded by the production scanner and assert the original primary key in every mode/source/operation combination
- remaining_validation: deployed fulltext DML verification under default and `NO_BACKSLASH_ESCAPES` sessions


XuPeng-SH (approved):
Deep-reviewed exact head `30e1d5feabd7a90dfd5e08183c5ed0683485ca06`; no blocking findings. Character/binary key literals now round-trip through the production scanner as one value in both default and NO_BACKSLASH_ESCAPES modes. Session resolution is cached once per batch; resolver errors and remote execution correctly fall back to the captured SQL mode. Both DELETE and INSERT maintenance statements use the same encoded key, without changing numeric-key behavior. Focused mode/source/key stress passed under `-race -count=50`; full postdml, sqlquote, and MySQL parser packages passed under race.

Issue #26216, opened 2026-07-27
[Bug]: Data Branch copied view with boolean predicate returns an internal runtime error


Ariznawlll:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

main

### Commit ID

7cb78aa5f88e3b98163f6b3bffb5f71586c0cbbf

### Actual Behavior

A view whose predicate is a boolean column can be queried in its source database, but querying the copied view after `DATA BRANCH CREATE DATABASE` returns an internal runtime error instead of rows.

The failing query consistently reaches:

```text
pkg/sql/plan.calcBlockSelectivityUsingShuffleRange
pkg/sql/plan/stats.go:2488
```

### Steps to Reproduce

```sql
DROP DATABASE IF EXISTS view_bool_branch_dst;
DROP DATABASE IF EXISTS view_bool_branch_src;
CREATE DATABASE view_bool_branch_src;
USE view_bool_branch_src;

CREATE TABLE t(
  id INT PRIMARY KEY,
  amount DECIMAL(10,2),
  enabled BOOL NOT NULL
);
INSERT INTO t VALUES (1, 10.00, true), (2, 20.00, false);
CREATE VIEW v AS SELECT id, amount FROM t WHERE enabled;

SELECT * FROM v ORDER BY id;
-- expected and returned in the source database: (1, 10.00)

DATA BRANCH CREATE DATABASE view_bool_branch_dst FROM view_bool_branch_src;
SELECT * FROM view_bool_branch_dst.v ORDER BY id;
```

### Expected Behavior

The copied view should return the same single row as the source view. Its boolean predicate should remain usable after the database copy.

### Reproduction Evidence

The final query failed in 3/3 independent databases created from the same latest-main binary.

Control cases in the same matrix were successful in 3/3 runs:

- `CREATE VIEW v AS SELECT id, amount FROM t`;
- `CREATE VIEW v AS SELECT id, amount FROM t WHERE id > 0`.

The failure is therefore specific to the bare boolean predicate in a copied view, rather than database copy or a general view query.

### Additional Information

`estimateFilterBlockSelectivity` sends the predicate to `calcBlockSelectivityUsingShuffleRange`. That helper reads function metadata from `expr.GetF()` without handling the bare-column predicate form. The boolean control path reaches this form after the copied-view rewrite.

VioletQwQ-0:
## QA handoff

[...]

PR #26312, opened 2026-07-28
fix: handle bare boolean filters in block selectivity

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26216

Copied views can expose a bare boolean column as a zonemappable filter. The
block-selectivity estimator treated every such filter as a function expression
and dereferenced a missing function, causing an internal error when querying
the copied view.

This change only applies function-specific selectivity handling when the
expression is actually a function. Bare columns continue through the existing
generic overlap estimate. Regression coverage includes both the estimator's
expression-shape boundary and a real embedded Data Branch database copy whose
source and copied views are queried through SQL.

Validation:

- `go build ./pkg/sql/plan`
- `go vet ./pkg/sql/plan ./pkg/tests/issues`
- full `pkg/sql/plan` tests, normal and `-race`
- focused estimator test, `-race -count=100`
- embedded copied-view regression

QA handoff required: yes. Please verify the reported Data Branch copied-view
scenario with a bare boolean predicate in a deployed test environment before
closing the issue.

### DIFF
--- a/pkg/sql/plan/stats.go
+++ b/pkg/sql/plan/stats.go
@@ -2485,9 +2485,11 @@ func getOverlap(s *pb.StatsInfo, colname string) float64 {
 
 func calcBlockSelectivityUsingShuffleRange(s *pb.StatsInfo, colname string, expr *plan.Expr) float64 {
 	sel := expr.Selectivity
-	switch expr.GetF().Func.ObjName {
-	case "isnull", "is_null", "prefix_eq", "prefix_in", "prefix_between", "prefix_in_range": //special handle
-		return sel
+	if fn := expr.GetF(); fn != nil {
+		switch fn.Func.ObjName {
+		case "isnull", "is_null", "prefix_eq", "prefix_in", "prefix_between", "prefix_in_range": //special handle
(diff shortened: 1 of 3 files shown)


XuPeng-SH (approved):
Deep review completed on exact head 567da1539f1151eb50aa9a04261ed2b9b8589839. The crash was an expression-shape violation: zonemappable filters may be bare Col expressions, while the estimator unconditionally dereferenced GetF(). The nil/type guard is placed narrowly around function-specific exceptions; bare columns still use the existing overlap/dynamic-parameter/sort-order estimate, so there is no new hot-path work or selectivity policy change. I also checked constant/non-zonemappable, nested function, special prefix/is-null, missing stats, and ordinary comparison paths. Focused planner regression passes, the embedded branch-copy test validates source and copied-view results plus rows.Err(), CI is green, and the PR is mergeable. No blocking correctness, performance, or unhappy-path issue found.

Issue #26299, opened 2026-07-28
[Bug]: Test_TxnExecutorExec flakes on embedded HAKeeper startup timeout under CI load


jiangxinmeng1:
### What happened?

The Ubuntu/x86 unit-test job intermittently fails before `Test_TxnExecutorExec` reaches its SQL assertion because the embedded cluster cannot observe HAKeeper reaching the running state within the fixed two-minute startup timeout.

Failing job: https://github.com/matrixorigin/matrixone/actions/runs/30326473317/job/90172863760

Failure:

```text
txn_executor_test.go:31: Received unexpected error:
    context deadline exceeded
    internal error: waitHAKeeperRunningLocked
--- FAIL: Test_TxnExecutorExec (124.17s)
```

The preceding log shows Dragonboat scheduler delay and repeated heartbeat/proposal timeouts, including batches of 8 and 16 LocalTick messages.

### Why this appears flaky rather than PR-specific

The failure occurs in `c.Start()` before any SQL is executed. The failing PR has no changes under `pkg/tests/txnexecutor`, `pkg/embed`, or `pkg/logservice`. The exact test passed locally on both the PR head (14.22s) and a clean `upstream/main` worktree at `6004becb5` (13.36s). The same PR head also passed the coverage UT job.

### Expected behavior

`Test_TxnExecutorExec` should not fail solely because a heavily loaded CI runner delays the embedded single-node HAKeeper election/startup. The test should also reuse or reliably clean up embedded cluster resources.

### Possible direction

Reuse the package base cluster for this test, or otherwise make embedded-cluster startup and cleanup resilient to transient scheduler starvation without weakening the SQL assertion.

PR #26316, opened 2026-07-28
Agent/fix 26299 hakeeper startup cleanup

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26299

Agent/fix 26299 hakeeper startup cleanup


XuPeng-SH (changes requested):
Deep-reviewed latest head `eac719010d2ac6f79f0563aecf7991c6f0289a68` across initial startup, partial service construction, reverse-order close, repeated close, dynamic CN rollback, file-service cleanup, and the CI timeout change. The timeout is test-only and does not affect production or the successful startup path, but two real cleanup/state blockers remain.

### P1 — Rollback still cannot close services whose Start/Bootstrap partially failed

The concrete service is assigned to `op.reset.svc` only after all startup work succeeds: LOG at `pkg/embed/operator.go:226-245`, TN at `:260-273`, and CN at `:286-302`. If `s.Start()` returns after starting an earlier component (CN starts query/frontend/RPC stages sequentially), or LOG `BootstrapHAKeeper` fails after the service is running, `operator.Start` returns with `reset.svc == nil`. `cluster.Start` then calls the new rollback, but `operator.Close` has no service handle and only stops the outer stopper/file service. The partially started service can retain goroutines/listeners and its own resources. Store ownership immediately after construction and make rollback close it on every later error. The new `startFn` test pre-populates `reset.svc`, so it cannot catch this production path.

### P1 — A close error leaves a failed dynamic CN in the active cluster topology

`rollbackNewServicesLocked` truncates `services`, `files`, and restores `options.cn` only when every close returned nil (`pkg/embed/cluster.go:260-267`). CN Close legitimately aggregates RPC/query/gossip errors. On such an error, `StartNewCNService` returns failure but the cluster remains `started` while `GetCNService`/`ForeachServices` expose the failed, partly closed CN and subsequent additions use the inflated CN count/config sequence. Logical rollback must happen regardless of cleanup errors; retain any still-owned cleanup handles separately if retries are required. Add an unhappy-path test with a new CN whose Close returns an error.

The PR is also now genuinely conflicting with latest main: #26203 changed `RunBaseClusterTests` to take `t`, while this branch changes its old return-value form. Please rebase/merge and preserve the main-side startup-error propagation when resolving it.

XuPeng-SH (changes requested):
One production-relevant cleanup blocker remains on latest head `61a11da4a61d254917157fc8c971177127dd3561`.

`waitClusterConditionLocked` acquires a managed HAKeeper client from `waitHAKeeperReadyLocked`, but closes it only after `waitFunc(client)` succeeds. If `waitHAKeeperRunningLocked` or `waitAnyShardReadyLocked` returns an error, the function returns immediately and the client is never closed. The exact #26299 failure takes this branch: the client is successfully created, the HAKeeper-running wait reaches its deadline, and startup rollback can only see the operator stopper/file service because this client remains a local variable. `NewCNHAKeeperClient` owns an underlying managed MORPC client, so repeated failed starts/tests leak client registration and transport resources even though the new cluster rollback reports cleanup complete.

Please establish client ownership immediately after acquisition and close it on every exit, while preserving the primary wait error if Close also fails. Add an injected/fake-client regression that verifies Close is called exactly once on both wait success and wait failure; the existing tests call the wait functions directly and cannot detect this leak.

The two prior blockers are fixed in this head: constructed services are now retained before `Start`, and failed dynamic CNs are removed from active topology even when cleanup returns an error. CI is green, and the focused cleanup tests pass under `-race` locally.

XuPeng-SH (approved):
Re-reviewed exact head e2acf3e49d44a18ba625c045e1701c82b0c7bebe after the prior CHANGES_REQUESTED. The remaining HAKeeper-client leak is fixed: ownership is established immediately after acquisition, Close runs exactly once on both wait success and wait failure, a wait error remains the primary return value if Close also fails, and the success-path close behavior remains unchanged. The earlier partial-service and dynamic-CN rollback blockers remain fixed after the main merges. The longer timeout is test-only, with no production startup latency change. Focused deterministic CGo tests for client cleanup, timeout selection, startup rollback, and operator cleanup pass.

PR #26318, opened 2026-07-28
fix(hashbuild): avoid false budget admission rejection

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26192

HashBuild budget accounting used whole-payload multipliers that double-counted memory already covered by retained batches and auxiliary reservations. Wide LOAD DATA input could therefore be rejected even though one decoded/runtime-filter payload fit the hard query budget.

This PR:

- replaces duplicated whole-payload multipliers with one payload plus explicit allocator, metadata, and framing slack;
- accounts transient old/new allocation overlap when reusable vectors, spill arrays, unique keys, and marshal buffers grow;
- reuses the existing spill high-water lease and grows only the uncovered transient difference, preventing duplicate admission rejection;
- falls back to releasing the old reader batch before decoding when a reuse peak cannot be admitted;
- adds tight-budget, large/small batch, spill replacement, cleanup, and corrupted-record regression coverage.

Validation:

- hashbuild and spillutil package tests
- hashbuild and spillutil package tests with race detector
- focused transient-overlap tests with race detector repeated 100 times
- go vet for both packages
- independent blocker reviews converged with no remaining correctness blocker


pkg/sql/colexec/hashbuild/budget.go
```diff
@@ -231,6 +231,18 @@ func (hb *HashmapBuilder) projectedBatchCopyBytes(src *batch.Batch) (uint64, err
 		if tail > projected {
 			projected = tail
 		}
+		// Vector growth allocates its replacement before releasing the retained
+		// tail. For large buffers mpool follows Go's 1.25x growth policy; the
+		// existing tail reservation covers the old allocation, while this new
+		// reservation must cover the complete replacement capacity.
+		growthSlack := projected / 4
```
XuPeng-SH: `max(tail, source) * 1.25` is not an upper bound for the replacement allocated by `UnionBatch`. The grow target is `oldLen + appendedLen`; when old and appended varlen areas are comparable, `mpool.GrowCapacity` applies the 1.25x step repeatedly before it reaches that target.

Concrete repro: copy a 4096-row batch containing one 1 KiB varchar column, then copy a second identical batch into the partial tail. On this head, `projectedBatchCopyBytes` reserves 6,098,944 bytes, while the measured mpool peak rises from 4,292,608 to 14,663,680 (10,371,072 additional bytes), and `copyBuildBatch` returns `ErrHashBuildBudgetInvalid` even with a 128 MiB budget. The exact same test passes on base `522a157c`, whose old estimate reserves 19,038,208 bytes.

Please derive the partial-tail replacement bound from each destination vector's old capacity and actual required `old + append` size (using `mpool.GrowCapacity` or an equivalent exact upper bound), and add this two-half-batch varlen case as a regression test.

XuPeng-SH (changes requested):
Requesting changes for one merge blocker: the new partial-tail estimate under-admits the real replacement peak and rejects a valid wide-varlen append that succeeds on the base commit. The inline comment includes a deterministic reproduction and measured allocation evidence. The remaining production/test diff passed my lifecycle/error-path review, both owning-package test suites (including full `-race`), `go vet`, and the current GitHub CI.

pkg/sql/colexec/hashbuild/budget.go
aptend: Fixed in 82d84f224. Partial-tail admission now derives each destination vector data/area replacement with mpool.GrowCapacity(oldCap, old+append), tracks transient replacement peak separately from the retained delta, and accounts for any remainder copied after the tail. Added the exact 2 x 4096-row, 1 KiB varchar regression; it failed with ErrHashBuildBudgetInvalid before the change and now passes. Also covered fixed/const-varlen, no-growth, remainder, and invalid-input paths. Full hashbuild/spillutil tests and race tests pass; changed-code coverage is 80.77% using the CI parser.

XuPeng-SH (approved):
Deep re-review complete. The previous partial-tail blocker is fixed at 82d84f224: admission now follows the exact per-vector mpool.GrowCapacity replacement sequence, preserves earlier retained deltas, and covers remainder allocation. The deterministic 2 x 4096-row / 1 KiB varchar regression now passes. I also re-audited reservation ownership, spill reader/writer cleanup, recursive scatter source charging, and bounded serialization/reuse peaks. Full affected packages pass under -race on the exact head and after a clean merge with current main; all 17 new boundary tests pass with -race -count=50; go vet and diff check pass. No remaining blocker.

PR #26319, opened 2026-07-28
test: stabilize authenticated issue cluster tests

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26320

The authenticated data-branch regression tests each started and stopped a full embedded cluster. Meanwhile, `pkg/tests/issues` kept its separate 3-CN base cluster alive and ran concurrently with other heavy race packages. On an 8-core CI runner this caused both embedded clusters to miss Dragonboat ticks and lose HAKeeper connections ([failing job](https://github.com/matrixorigin/matrixone/actions/runs/30338984176/job/90210230582)).

This change:

- introduces one serialized, package-scoped authenticated cluster suite with explicit shutdown in `TestMain`;
- reuses that cluster across the issue 26087, 26114, 26120, and 26127 tests instead of starting one cluster per test;
- runs `pkg/tests/issues` in the exclusive race-test group rather than concurrently with the other heavy packages;
- does not add retries or relax any timeout or assertion.

Local race results:

- the three authenticated tests pass and decrease from 54.30s to 26.25s on the same machine;
- the complete `pkg/tests/issues` package, including the newly merged 26120 and 26127 tests, passes under `-race` in 72.39s;
- shell syntax, package-scope resolution, and `git diff --check` pass.


XuPeng-SH (approved):
Deep-reviewed at head d6c6966092d227a169853ec5dc2132ebce16e592. The process-scoped authenticated cluster has a single initialization/cleanup owner, serializes test bodies safely (including FailNow/panic unwinding), preserves assertion strength, and the CI package partition neither duplicates nor omits pkg/tests/issues. The three focused tests, focused race run, full issues package (-short), go vet, script syntax, and diff checks passed locally.

PR #26324, opened 2026-07-28
fix(pipeline): preserve terminal cause during spool abort

### LINKED ISSUE #26322: [Bug]: pipeline cleanup fallback masks the original execution error

## Branch Name

main

## Commit ID

a777e80be4b3763592e279501bf93d11c0f1713b

## Actual Behavior

In multi-input pipeline cleanup, an operator can first record the real execution error on a shared `PipelineEdge`, while another Connector sharing that edge fails to deliver its normal End because `Ch2` is already full. `Connector.Reset` then aborts its spool with `ErrPipelineEndSignalDeliveryFailed`. A buffered `GetFromSpool` signal observes that synthetic cleanup error and can return it before the durable original edge error, masking the user-visible error.

Observed in PR #26318 compose BVT job 90243593095. The dedup join correctly produced the expected duplicate-entry error for value 1000000, but the client received `internal error: pipeline end signal delivery failed`.

## Expected Behavior

If a shared `PipelineEdge` already has a durable terminal error, Connector/Dispatch cleanup must preserve that original error when aborting queued spool work. `ErrPipelineEndSignalDeliveryFailed` should only be used when no real terminal cause exists.

## Steps to Reproduce

1. Use a shared `PipelineEdge` with a full `Ch2` containing a `GetFromSpool` signal.
2. Record a real fatal error on the edge while the channel remains full.
3. Reset another successful Connector sharing that edge so its End delivery fails.
4. Consume the buffered spool signal.
5. Observe `ErrPipelineEndSignalDeliveryFailed` instead of the original edge error.

## Additional information

Related to #26250. That change correctly made fatal terminal wakeup non-blocking and observable through `PipelineEdge.Done()`, but exposed the existing fallback-error precedence problem. The fix should preserve #26250 liveness behavior and only change cleanup error selection.

CI run: https://github.com/matrixorigin/matrixone/actions/runs/30349542976/job/90243593095?pr=26318

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #26322

When normal End delivery fails on a shared `PipelineEdge` that already records a real terminal error, Connector or Dispatch cleanup currently aborts its spool with `ErrPipelineEndSignalDeliveryFailed`. Buffered spool signals can then surface that synthetic cleanup error before the durable execution error.

This change resolves the spool abort cause from receiver edges first and only falls back to `ErrPipelineEndSignalDeliveryFailed` when no more specific terminal error exists. It keeps the terminal wakeup and bounded cleanup behavior from #26250 unchanged.

Validation:

- focused Connector regression with `-race -count=20`
- `go test -race -count=1 ./pkg/vm/process ./pkg/sql/colexec/connector ./pkg/sql/colexec/dispatch`
- `go test -count=1 ./pkg/vm/pipeline`
- `go build` and `go vet` for all changed packages plus `pkg/vm/pipeline`


XuPeng-SH (approved):
Deep-reviewed the cleanup lifecycle and failure precedence at the latest head. The change preserves a durable receiver-edge cause before aborting queued spool work, retains the synthetic delivery error only as fallback, keeps first-fatal-wins semantics, adds no normal data-path cost, and does not extend the bounded send budget. Race tests for process, connector, and dispatch pass. No blocker found.

PR #26326, opened 2026-07-28
[Enhancement]: gate auto-increment resets on TN epoch-fence support

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #23143

This is the mixed-version prerequisite in the split #25883 plan. It does not add the final ALTER TABLE ... AUTO_INCREMENT=N SQL entry.

Allocator reset must remain disabled unless the transaction can prove that its target TN understands and enforces the AUTO_INCREMENT epoch fence. The final design uses two layers:

1. Early capability check. TN heartbeat capability is persisted by HAKeeper and captured in the transaction TN snapshot. COPY ALTER and TABLE CLONE reject reset before SetOffset unless every captured TN reports support.
2. Authoritative commit gate. A reset transaction and every DML transaction that writes a nonzero known allocator epoch are marked for a V9-only terminal commit method. V8 and legacy TNs reject that method before invoking the commit handler. This closes stale-heartbeat and check-to-commit downgrade windows. V6 remains assigned to ordered aggregate pipeline configuration; V7 remains assigned to structured CHECK constraints; V8 remains assigned to the runtime-filter key contract.

The transaction-wide requirement is monotonic across statement rollback and snapshot application, but is cleared for a restarted transaction generation. Row, file, skip-transfer, flush, and compaction paths preserve the allocator generation that actually allocated the values. Commit transport failures retain normal unknown-result semantics and never fall back to the legacy Commit method.

### Mixed-version behavior

| Topology | Result |
| --- | --- |
| new CN / old or V8 TN | reset or fenced DML commit fails closed before commit handling |
| new CN / mixed TNs | reset remains disabled; stale capability is still caught by the terminal method |
| new CN / all V9 TNs | reset and fenced DML commit proceed |
| old CN / new TN | existing TN epoch validation remains authoritative |
| local dispatch | guarded commit fails closed on the unsupported local path |
| restart | persisted capability and transaction epoch state are restored without leaking the previous transaction generation |

All protobuf changes are additive. Legacy missing capability fields decode as false.

### Validation

- `git range-diff` proves all seven functional commits are unchanged after merging `upstream/main@4a747e44586f09c25bb3b77a8067639c7e8ac346`.
- The only merge conflict was the MORPC version allocation: upstream retains V7/V8 and the AUTO_INCREMENT guarded commit now requires V9. Production dispatch tests prove V8 rejects before handler invocation and V9 invokes exactly once.
- Focused V8-reject/V9-accept tests passed under `-race -count=100`. Full `pkg/txn/client`, `pkg/txn/rpc`, `pkg/vm/engine/disttae`, `pkg/sql/compile`, and `pkg/incrservice` owning-package race tests passed.
- `make pb` regenerated cleanly with only the intended source/generated V9 contract comment change; `pkg/pb/txn` and `git diff --check` pass.
- BVT: N/A — ordinary BVT uses homogeneous current binaries and cannot create the old-TN/new-CN or MORPC-version-skew topology this prerequisite guards. Equivalent deterministic coverage uses the production reset, transaction, network/local dispatch, rollback/restart, and unknown-result paths.
- Exact-head `mo-self-review` and `mo-pr-preflight-review`: `PASS review=PASS validation=PENDING` at `f1c4374ff080ab8dad143a681bc0898ad5033bbf`; fresh GitHub CI will run after push.

### QA

QA is required. Final acceptance belongs to the later SQL integration stage and must cover old/new CN/TN, restart, rolling upgrade, and multi-CN.


gouhongshen (approved):
## Codex automated review

No defects found. Existing PR reviews, issue comments, and inline threads contained no human findings. Pure protobuf tests passed; CGo-dependent tests could not run because this worktree lacks cgo/libmo.dylib.

XuPeng-SH (approved):
Deep-reviewed exact head f44821a365d4528f14d7e676dd45d239182890f9.

The mixed-version contract is closed at both layers: legacy/missing TN capability fails the reset before allocator mutation, while epoch-bearing transactions monotonically select the V6-only terminal commit method and never fall back after transport/unknown-result failures. New TNs advertise and persist capability through heartbeat/HAKeeper/cluster snapshots; old peers decode missing fields as false. The server rejects the guarded method before commit handling on V5/unknown-method binaries, local dispatch fails closed, and ordinary legacy/known-zero writes retain rolling-upgrade compatibility before the first reset. Transaction snapshots preserve the requirement monotonically, while restart creates a clean generation. Row, object, skip-transfer, flush/compaction entry paths are covered by the late precommit gate.

No steady-state overhead is added to non-epoch writes or RPC count. Epoch-bearing writes add only constant-time capability/transaction marking; commit remains a single terminal request. Focused deterministic CGo tests passed for txn/client lifecycle and unknown outcomes, rpc version/network gates, disttae target snapshots/write paths, COPY ALTER, and TABLE CLONE.

aptend (approved):
Reviewed exact head `c70040799b9754c6543b45a441d6048e411d1174`.

The mixed-version safety closure is complete: TN capability is propagated through heartbeat/HAKeeper into the transaction TN snapshot; allocator resets fail closed before `SetOffset`; epoch-bearing writes monotonically select the V7-only terminal commit method; V6/legacy TNs reject it before commit handling; snapshot application preserves the requirement while restart clears the prior generation; and transport failures retain unknown-result semantics without falling back to legacy Commit. Protobuf changes are additive and missing capability fields decode safely as false.

Focused race stress (`-count=100`), full owning-package race tests, affected-package vet/build, and current GitHub CI passed. Approved.

XuPeng-SH:
@VioletQwQ-0 conflict

iamlinjunhong (approved):

PR #26327, opened 2026-07-28
perf(fulltext): bound filter-only boolean AND candidates

VioletQwQ-0:
## What type of PR is this?

- [x] Improvement

issue #26307

This PR adds the first, deliberately narrow optimization for the classic FULLTEXT workload in #26307.

- The planner marks only a bounded, filter-only, single-index query through query-local table-function parameters.
- Runtime enables the fast path only for TF-IDF Boolean mode patterns parsed as strict `+term +term ...` AND queries.
- The table function pushes `LIMIT + OFFSET` into its internal SQL, skips `runCountStar`, and avoids the general per-document aggregation maps and fixed byte pool.
- The fast path emits tied scores for the existing Sort and retains at most the bounded candidate count.
- Visible scores, BM25, unbounded or very large limits, async indexes, explicit sort, residual filters, complex Boolean syntax, phrase/prefix/wildcard, and other unsupported shapes keep the existing path.

The current strict-AND JOIN pattern assigns the same TF-IDF score to every matching document. Therefore this optimization preserves the existing unordered Top-K semantics; it intentionally does not promise stable document IDs among ties.

The fast path is capped at 1,048,576 candidates. This PR addresses the bounded #26307 workload only and does not claim to solve the general high-cardinality memory boundary tracked by #25638.

### Validation at exact head `918c3fc01ecef36859aadb8add9a325814924d54`

- `mo-cgo-test -count=1 -timeout=300s ./pkg/fulltext ./pkg/sql/colexec/table_function ./pkg/sql/plan`
- `go vet ./pkg/fulltext ./pkg/sql/colexec/table_function ./pkg/sql/plan`
- Final stacked head additionally passed focused race tests (100 repetitions), full owning-package race tests, `go build`, `go vet`, and `make build`.
- Linux `GOAMD64=v3 GOEXPERIMENT=simd` binary built on host 129; SHA256: `34001480973d73d17c9d254e883b9f9f7668a674b2d5759ecae5f9b26c5ef9aa`.

### Performance gate

Pending; no performance result is claimed in this draft.

Two 230w/C10 baseline attempts were invalidated before warm-up because an unrelated self-hosted GitHub Runner took ownership of host 129 and executed `kill -9` against the benchmark MO. The conflicting workflow was `matrixorigin/mo-auto-test` run `30356127038`; another run started ten minutes later and repeated the same setup behavior. Per the benchmark protocol, these are contaminated host attempts rather than product failures.

The PR will remain draft until a reserved 129 window passes the five-minute idle gate and the per-PR threshold (QPS or CPU/query improvement at least 5%, errors zero, P99 regression no more than 5%).

### QA

QA is required because this changes fulltext execution behavior. Please cover the existing fulltext/gojieba/pushdown/membership suites and bounded strict-AND fallback boundaries.

### DIFF
--- a/pkg/fulltext/fulltext.go
+++ b/pkg/fulltext/fulltext.go
@@ -106,6 +106,29 @@ func (s *SearchAccum) PatternAnyPlus() bool {
 	return s.AnyPlus
 }
 
+// IsPureBooleanAnd reports whether the parsed pattern is the strict Boolean
+// conjunction produced for "+term +term ...".  In the classic fulltext
+// implementation a JOIN pattern has one result slot for the whole
+// conjunction, so every matching document receives the same TF-IDF score.
+// Keep this deliberately narrow: prefix, phrase, group, negative and weighted
+// forms must retain the general scoring path.
+func (s *SearchAccum) IsPureBooleanAnd() bool {
+	if s.Mode != int64(tree.FULLTEXT_BOOLEAN) || len(s.Pattern) != 1 {
+		return false
+	}
+
+	root := s.Pattern[0]
+	if root.Operator != JOIN || len(root.Children) < 2 {
+		return false
+	}
+	for _, child := range root.Children {
+		if child.Operator != PLUS || len(child.Children) != 1 || child.Children[0].Operator != TEXT {
+			return false
+		}
+	}
+	return true
+}
+
 func hasPatternAnyPlus(ps []*Pattern) bool {
 	for _, p := range ps {
 		if p.Operator == PLUS || p.Operator == JOIN {
--- a/pkg/fulltext/fulltext_test.go
+++ b/pkg/fulltext/fulltext_test.go
@@ -659,6 +659,29 @@ func TestFullText2(t *testing.T) {
 
 }
 
+func TestSearchAccumIsPureBooleanAnd(t *testing.T) {
+	tests := []struct {
+		name    string
+		pattern string
+		want    bool
+	}{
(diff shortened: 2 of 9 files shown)


VioletQwQ-0:
Closing this draft after completing a controlled A/B/A evaluation for the 2.3M-document, C10 strict Boolean-AND workload.

Measured throughput was:

- A1: 1.6378 QPS
- B: 1.7188 QPS
- A2: 1.6954 QPS

The earlier comparison against A1 suggested a 4.95% improvement, but the baseline itself drifted by 3.51% from A1 to A2. Against the mean of A1 and A2, the net QPS improvement is only 3.14%, and B is only 1.39% faster than the adjacent A2 run. This does not pass the predefined 5% independent-effect gate.

Profile and EXPLAIN analysis also show that this change optimizes non-dominant work: it skips `runCountStar`, bounds table-function result state, and avoids the general per-document aggregation maps, but it does not remove the dominant posting-side `Group`, `HashBuild`, `HashJoin`, hash-table resize, scan, and spill work. Because those operators are blocking, the outer LIMIT does not materially reduce the millions of posting rows processed before the final candidates are produced.

The implementation is not being closed because of a correctness failure. Its bounded-result and cancellation ideas may be reused as part of a dedicated rarest-term-driven posting-intersection path, where LIMIT can terminate retrieval early. As a standalone performance optimization for #26307, however, the measured effect is too small and not stable enough to justify merging.

PR #26330, opened 2026-07-28
fix(clone): preserve foreign key metadata across alter copy

### LINKED ISSUE #26328: [Bug]: CREATE DATABASE CLONE fails after ALTER TABLE on a foreign-key child table

## Branch Name

`main`

## Commit ID

`1ca928aefd`

- Reproduced on the latest `main` commit above.
- Also reproduced on MatrixOne Cloud `8.0.30-MatrixOne-v4.1.4`.
- The failure is independent of table data and deployment environment.

## Actual Behavior

`CREATE DATABASE ... CLONE ...` fails when the source database contains a foreign-key child table that has been altered after creation:

```text
ERROR 20101 (HY000): internal error: can't find fkTable from fk clone_fk_dst.custom_operator.(source_file_id) {...}
```

A schema created directly in its final form clones successfully. The equivalent schema produced by creating the foreign key first and then running `ALTER TABLE ... ADD COLUMN` fails consistently.

## Expected Behavior

Database clone should reconstruct the same schema and data regardless of whether the current schema was created directly or reached through valid `ALTER TABLE` operations.

## Steps to Reproduce

```sql
DROP DATABASE IF EXISTS clone_fk_src;
DROP DATABASE IF EXISTS clone_fk_dst;

CREATE DATABASE clone_fk_src;
USE clone_fk_src;

CREATE TABLE file (
    file_id VARCHAR(36) NOT NULL,
    PRIMARY KEY (file_id)
);

CREATE TABLE custom_operator (
    id BIGINT NOT NULL AUTO_INCREMENT,
    source_file_id VARCHAR(36) DEFAULT NULL,
    PRIMARY KEY (id),
    CONSTRAINT fk_custom_operator_source_file
        FOREIGN KEY (source_file_id)
        REFERENCES file (file_id)
);

ALTER TABLE custom_operator
    ADD COLUMN kind VARCHAR(32) NOT NULL DEFAULT 'code';

CREATE DATABASE clone_fk_dst CLONE clone_fk_src;
```

The last statement fails with `can't find fkTable from fk clone_fk_dst.custom_operator.(source_file_id)`.

Control case: define `kind` in the original `CREATE TABLE custom_operator` statement and omit the `ALTER TABLE`; cloning succeeds.

### Verified observations

- The issue reproduces without inserting any rows.
- Cloning the already cloned/upgraded database fails in the same way.
- A real upgraded schema has `custom_operator.rel_version = 4` and fails.
- A freshly created equivalent schema has `custom_operator.rel_version = 0` and succeeds.
- The error is raised in `pkg/sql/plan/build_show_util.go` when `ctx.ResolveById(fk.ForeignTbl, snapshot)` returns no parent table definition while reconstructing the child table DDL.

### Suggested repair direction

The clone path must resolve the child table definition, its `ForeignTbl` reference, and the referenced parent table from one consistent catalog snapshot/version.

Please avoid resolving the parent from the current catalog as a fallback, because that could silently combine schema versions. A clean fix should either:

1. make snapshot-aware relation-ID resolution correctly handle altered table versions; or
2. have database clone build and use a snapshot-scoped table-ID-to-definition map for all source tables when reconstructing foreign keys.

A regression test should cover:

- parent table + foreign-key child table;
- `ALTER TABLE` on the child after FK creation;
- `CREATE DATABASE ... CLONE ...`;
- cloned FK points to the cloned parent and remains enforceable.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Closes #26328

COPY-mode `ALTER TABLE` replaces a relation with a new physical table. The old implementation could lose or split foreign-key metadata while doing that:

- `mo_catalog.mo_foreign_keys` rows owned by the source child were deleted before the replacement inherited them;
- catalog-only forward references created under `FOREIGN_KEY_CHECKS=0` were absent from `TableDef`, so reconstructing the temporary table could not recreate them;
- self references could leave multiple `RefChildTableDef` values whose add/remove/replace paths disagreed;
- databases and historical snapshots/PITR corrupted by an older binary retained the FK in `SHOW CREATE TABLE` but lacked the catalog dependency used by clone and restore ordering.

This PR:

- applies ALTER FK mutations to the source catalog rows, transfers those rows around the source drop, and restores the original logical table name after the physical replacement;
- keeps parent-side names stable during COPY so self references never point at the temporary relation;
- centralizes reverse-reference add/remove/restore/replace into one canonical mutation that emits exactly one `RefChildTableDef`, deduplicates IDs, preserves the self-reference sentinel `0`, and is also used by `TRUNCATE TABLE`;
- derives dependency edges from historical source table DDL in addition to `mo_foreign_keys` for database clone, snapshot restore, PITR restore, and dropped-account timestamp restore; table DDL repairs resolved historical constraints while the catalog remains necessary for unresolved forward references;
- enumerates ordinary snapshot source databases by `MO_TS` under the source account instead of resolving the snapshot name there. Snapshot metadata belongs to the system account, so name resolution under a dropped source account incorrectly reported `mo_snapshots` missing;
- adds unit and BVT coverage for repeated COPY ALTERs, altered parents/children, sibling and multi-level FKs, composite/multiple-parent FKs, self references plus external children, FK recreation, both forward-reference paths, snapshot clone, clone-after-clone, table/column renames, missing historical catalog rows, canonical TRUNCATE replacement, and timestamp-based source enumeration.

The generated `clone_fk_after_alter` result was manually audited: all 17 recorded errors are intentional FK enforcement checks; there are no unexpected errors.

## Validation

- affected Go packages (`./pkg/frontend ./pkg/sql/compile ./pkg/sql/plan`) pass with the repository CGo test environment;
- `clone_fk_after_alter`: 240/240;
- `snapshot_restore_to_table_fk_2`: 484/484;
- `restore_dropped_account_to_new`: 20/20;
- `restore_dropped_account_with_account_sp`: 34/34;
- `restore_database_using_cluster_sp`: 557/557;
- release-to-target compatibility check from the initial fix:
  - binary at `1ca928aefd` reproduced the old corruption (`mo_foreign_keys` count changed from 1 to 0 after COPY ALTER while `SHOW CREATE TABLE` retained the FK);
  - after restarting the same data directory with this PR, cloning both the live database and a snapshot created by the old binary succeeded;
  - invalid child inserts into both clones were rejected by the restored FK.
- post-main-merge FK topo-sort signature sync: focused `pkg/frontend` tests pass.


gouhongshen:
I did a destructive review of the FK metadata paths. I do not think this is merge-ready yet; there are two compatibility/data-integrity blockers and one invariant bug.

### 1. Existing corrupted deployments and historical snapshots are not repaired

Issue #26328 confirms that already-deployed v4.1.4 databases can have missing `mo_foreign_keys` rows. This PR changes only future COPY ALTER writes. There is no upgrade/reconciliation step, so upgrading an already-affected database and cloning it immediately will still fail.

Historical snapshots/PITR are a harder form of the same problem: the restore paths query `mo_foreign_keys` at the requested timestamp (`snapshot.go:getFkDeps`, `snapshot_restore_with_ts.go:getFkDepsWithTS`, and `pitr.go:getFkDepsInPitrRestore`). A current catalog write cannot repair a pre-fix timestamp. The new snapshot BVT takes the snapshot *after* running ALTER on the fixed code, so it does not cover this compatibility case.

This needs an explicit source-of-truth/reconciliation design, plus a release-to-target compatibility test that creates the corruption with the old release and clones/restores it with the new binary. Requiring users to ALTER every affected table again would not be an acceptable repair.

### 2. COPY ALTER still silently loses catalog-only forward references

The engine explicitly supports forward references with `FOREIGN_KEY_CHECKS=0`. When the parent is absent, `getForeignKeyData` sets `ForwardRefer=true`; `buildCreateTable` writes the relationship to `mo_foreign_keys` but deliberately does **not** add it to `TableDef.Fkeys`.

COPY ALTER reconstructs the temporary table from `TableDef`, so such an FK produces no temporary-table FK row. Dropping the source deletes the only catalog row, and the new copy-table rename statements update zero rows. Creating the parent later can no longer discover and attach the constraint, silently removing data-integrity enforcement.

A minimal regression case is:

```sql
set foreign_key_checks = 0;
create database d;
create table d.child (
  id int,
  parent_id int,
  constraint fk_forward foreign key (parent_id) references d.parent(id)
);
alter table d.child add column payload int;
create table d.parent (id int primary key);
set foreign_key_checks = 1;
-- the FK must still be present and enforceable
```

The same case should be tested after dropping an existing parent under `FOREIGN_KEY_CHECKS=0`, ALTERing the child, and recreating the parent.

### 3. `RefChildTableDef` is not canonicalized

A self-referencing temporary copy already gets a `RefChildTableDef{Tables:[0]}` from `CreateTable`. `restoreNewTableRefChildTbls` then blindly appends another `RefChildTableDef`. The new `reconcileParentRefChildTableID` updates only the first definition and returns; `removeChildTblIdFromParentTable` also stops at the first, while disttae table-def reconstruction lets the last definition win.

Concrete sequence: self-referencing parent + external child → ALTER parent → DROP child. The parent can retain the deleted child ID and remain incorrectly blocked as “referenced”.

Please centralize add/remove/restore/replace behind one canonical mutation that emits exactly one `RefChildTableDef`, deduplicates table IDs, and handles the `0` sentinel consistently.

There is also a root-cause evidence gap around the reverse-reference change: on the normal path, temporary-table CREATE already calls `AddChildTblIdToParentTable` with the new physical ID, and the internal source DROP uses `IgnoreForeignKey`, so the old cleanup should remove the old ID while leaving the new one. The new unit tests prove the helper in isolation, but do not reproduce a production path where the replacement ID is actually absent. If another path/cache behavior causes that absence, it should be captured by an integration regression test.

Verification on this head:

- `pkg/sql/compile` and `pkg/sql/plan` tests pass through the repository CGo wrapper.
- `git diff --check` passes.
- At review time, the GitHub upgrade compatibility jobs are skipped and the main UT/BVT jobs are still pending.

gouhongshen:
Addressed all three blockers in `d7a7d02513`:

1. **Existing data and historical snapshots:** clone now unions dependencies parsed from the source `SHOW CREATE TABLE` DDL with `mo_foreign_keys`. The DDL is the durable source for resolved constraints; the catalog remains the source for unresolved forward references. I reproduced the corruption with pre-fix commit `1ca928aefd` (catalog count 1 → 0 after ALTER), created a snapshot with that binary, restarted the same data directory on this PR, and verified both live-database clone and old-snapshot clone. Both cloned FKs reject invalid inserts. An automated topology test covers the missing-catalog-row case.
2. **Catalog-only forward references:** COPY ALTER now applies FK ALTER actions to the source rows, transfers those child-owned rows across the source drop, then restores the logical source name. BVT covers both an initially missing parent and drop-parent → ALTER child → recreate-parent; both rows survive and enforcement resumes.
3. **`RefChildTableDef` invariant:** add/remove/restore/replace now use one canonical mutation, emit exactly one definition, deduplicate IDs, and preserve sentinel `0`. Unit tests cover multiple definitions/duplicates for every mutation; BVT covers self-referencing parent + external child → ALTER parent → DROP child/parent.

The root-cause gap was also corrected: the implementation no longer assumes the temporary CREATE omitted the replacement reverse ID. It preserves source-owned catalog metadata explicitly and canonicalizes whatever reverse-reference state CREATE produced.

Validation: affected Go packages pass; `clone_fk_after_alter` is 240/240, with all 17 recorded errors manually confirmed as intentional FK enforcement checks.

gouhongshen:
第二轮破坏性 review（head `d7a7d0251`）：仍有 2 个 blocker，暂不能 approve。

1. **[compatibility / data loss] 历史缺失 `mo_foreign_keys` 的恢复只接到了 database clone，普通 snapshot/PITR restore 仍未修复。**

   `doRestoreSnapshot` 仍调用 `fkTablesTopoSort(..., nil)`（`pkg/frontend/snapshot.go:665`），PITR 的 `fkTablesTopoSortInPitrRestore` 也只读 `mo_foreign_keys`（`pkg/frontend/pitr.go:1675`）；dropped-account 的 TS restore 同样只读 catalog。这里和 `fkTablesTopoSort` 中“restore and clone reconcile both representations”的注释不一致。

   对旧版本 COPY ALTER 已经丢掉 catalog row、但 `SHOW CREATE TABLE` 仍保留 FK 的快照：topo 结果为空，`fkTableMap` 为空，表会按 `relname` 顺序进入普通恢复。若 `child` 排在 `parent` 前，`recreateTable` 创建 child 失败后会把 `no such table` 显式改成 `nil` 并返回（`snapshot.go:1553-1558`），后续没有重试；PITR 有完全相同的吞错逻辑（`pitr.go:1493-1498`）。最终 restore 可以成功返回，但 child 表和数据消失。TS restore 则直接失败。

   建议把“schema DDL + catalog”的依赖收集做成所有 clone/snapshot/PITR source 共用的单一入口，并增加从旧损坏 snapshot/PITR 恢复的回归测试（不只测试 clone old snapshot）。

2. **`RefChildTableDef` 的 mutation 还没有真正收口。**

   `TRUNCATE TABLE` 仍通过 `updateRefChildTableConstraintIDs` 直接逐项改 `RefChildTableDef.Tables`（`pkg/sql/compile/ddl.go:3231,3266-3280`），没有走本次新增的 canonical helper。对旧版本遗留的多个 `RefChildTableDef` / 重复 table ID，truncate child 后仍保留多个 definition 和重复的新 ID；而 TableDef materialization 遇到多个 definition 时是后者覆盖前者，行为仍依赖 definition 顺序。

   建议让 truncate 也复用同一个 canonical replace primitive，并补一个“多个 definition + 重复 ID -> truncate child -> 恰好一个 canonical definition”的测试，避免兼容性修复只覆盖部分写入口。

本地验证：`git diff --check` 通过；`./pkg/frontend ./pkg/sql/compile ./pkg/sql/plan -count=1` 全部通过。上述两个问题是调用链/状态兼容性缺口，现有 UT 没有覆盖。

gouhongshen:
两个 blocker 已在 `66b8a2b0b1` 收敛修复：

1. **所有历史恢复入口统一读取 schema DDL + catalog 依赖。** 新的 `collectRestoreSourceTableInfos`/`topoSortRestoreFkDeps` 被普通 snapshot restore、PITR restore、dropped-account TS restore 和 database clone 共用。新增 UT 直接构造历史 `mo_foreign_keys` 为空、但 child `CREATE TABLE` 仍含 FK 的状态，验证 PITR 与 TS restore 均按 parent → child 排序；另有 UT 验证 database/table/account 三种 source 收集范围。
2. **TRUNCATE 已改用 canonical replace primitive。** 删除了 `updateRefChildTableConstraintIDs`；现在多个 `RefChildTableDef`、重复 table ID 和 sentinel `0` 会收敛为恰好一个 definition，并把旧 child ID 替换为新 ID。新增对应回归 UT。ALTER 对历史缺失反向 ID 的 repair 语义保留在同一 owner 层的 `reconcileRefChildTableID`，TRUNCATE 不会凭空新增缺失引用。

验证：受影响的 `frontend/sql/compile/sql/plan` 测试全部通过；`clone_fk_after_alter` 240/240；`snapshot_restore_to_table_fk_2` 484/484。`snapshot_restore_to_table_fk` 为 206/211，5 个失败均发生在进入本次 FK 恢复逻辑前的 account snapshot 创建/查找阶段，并在原样重跑中稳定复现，未将其误报为通过。

我也尝试用旧提交 `1ca928aefd` 重新制造普通 snapshot 损坏数据；该旧二进制在本地会于 child COPY ALTER 提前报 `no such table .z_parent`，因此没有伪造端到端成功证据。缺失历史 catalog row 的恢复语义由上述精确 UT 覆盖，PR 描述已同步更新。

gouhongshen:
The repeated standalone pessimistic BVT failure is fixed in `b6008394ad`.

Root cause: the new ordinary snapshot source-schema collection called `SHOW DATABASES {snapshot=...}` under the source account context. Snapshot metadata is system-account-owned; after the source account was dropped, resolving the snapshot name from that tenant failed with `table "mo_snapshots" does not exist`. The restore had not failed in FK cleanup or schema parsing itself.

Fix: enumerate the source account at the already-resolved snapshot timestamp (`SHOW DATABASES {MO_TS=...}`), while retaining the source account context for historical catalog visibility. No fallback was added.

Local verification on a clean single-node instance built from the exact head:
- reproduced before fix: `restore_dropped_account_to_new` 18/20 with the identical `mo_snapshots` error;
- after fix: `restore_dropped_account_to_new` 20/20;
- `restore_dropped_account_with_account_sp` 34/34;
- `restore_database_using_cluster_sp` 557/557;
- focused frontend unit tests pass and `git diff --check` is clean.

PR description has been updated with the root cause and validation evidence.

gouhongshen:
第三轮破坏性 review（head `b6008394ad`）：**从代码审查角度 APPROVE，无剩余 blocker。**

已复验：

- 历史 FK 依赖现在统一由 schema DDL + `mo_foreign_keys` 合并，覆盖 database clone、普通 snapshot、PITR、cluster/dropped-account TS restore；`MO_TS` 枚举也避免了在已删除 source account 下解析 system-owned snapshot metadata。
- COPY ALTER 仍正确转移 source-owned catalog rows，catalog-only forward reference 不再丢失。
- `RefChildTableDef` 的 add/remove/restore/replace/TRUNCATE 已收敛到 canonical primitives；TRUNCATE 不会凭空补缺失引用，ALTER 的 repair 语义独立保留。
- 本地 `./pkg/frontend ./pkg/sql/compile ./pkg/sql/plan -count=1` 通过，`git diff --check` 通过；当前 GitHub 12 项检查通过，无失败或 pending。

我已尝试提交正式 `APPROVE` review，但当前 GitHub 身份 `gouhongshen` 同时是 PR 作者，平台拒绝作者批准自己的 PR（`Review Can not approve your own pull request`）。因此本评论记录技术审查结论；GitHub 的正式 approval 仍需非作者 reviewer 提交。

heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed exact head 127bd41caef3140c7d7a68a0906bb2a14db50e7c. The COPY ALTER FK catalog-transfer/canonical reverse-reference changes and historical restore/clone dependency reconstruction are internally consistent, including self references and unresolved forward references. Focused CGo tests passed for ./pkg/frontend ./pkg/sql/compile ./pkg/sql/plan.

PR #26331, opened 2026-07-28
perf(plan): reuse eligible multi-reference CTEs

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26131

Supersedes #26295. This is the canonical PR for the CTE reuse fix.

Multi-reference CTEs are currently bound and executed independently. TPCH Q15 therefore scans and aggregates `lineitem` twice. Reusing the existing streaming `SINK -> SINK_SCAN` fan-out directly is unsafe for this shape because dependent consumers can stop a bounded lock-step spool from advancing.

This change:

- conservatively identifies deterministic, non-recursive, non-correlated, full-drain, profitable multi-reference CTEs;
- keeps single-reference, volatile/real-time, sampled, recursive, correlated, short-circuiting, variable-width, poorly estimated, oversized, or unprofitable CTEs inline;
- emits one local producer and independent materialized `SINK_SCAN` readers;
- keeps retained query-MPool data below 64 MiB and 4096 batches, then spills sequentially to an anonymous query-scoped local file;
- returns a caller-owned batch to every reader, so destructive downstream operators cannot mutate source storage or another reader;
- preserves current streaming behavior for ordinary and single-consumer sink scans.

The planner estimate gate remains 32 MiB and fixed-width only. Runtime underestimation no longer changes a query that succeeded through the inline path into a hard 64 MiB error. There is no SQL syntax, protobuf, public API, or persistence-format change.

### Review follow-ups

- The exact profitable `SAMPLE` SQL from review is covered by `TestCTEMultiReferenceReuseGuards`; `Node_SAMPLE` is rejected as nondeterministic and produces zero shared `SINK_SCAN` nodes.
- The shared-batch ownership finding is covered through the production `Source -> Merge -> MergeOrder` spill path. One reader retains a batch while the other destructively spills and cleans its independently owned copy.
- Underestimated fixed-width output and many tiny batches are covered through bounded-memory spill tests, independent reader cursors, exact MPool restoration, and race runs.

### Validation

Exact base `82c164a83ae83af98e85a0bfd8fdab94327bd72f`, exact head `3164294f7555fdfbf2c5c532ca1b147fd3cbcb03`:

- focused planner, ownership, spill, compile, and embedded Q15 tests: PASS
- CR ownership/spill tests under `-race -count=20`: PASS
- full affected package tests (`materialized`, `dispatch`, `merge`, `compile`, `plan`): PASS
- affected package `go build`, `go vet`, and `git diff --check`: PASS
- schema-v4 preflight: `PASS review=PASS validation=PENDING` (remote CI and renewed review pending)
- strict local SF10 A/B/A: PASS
  - result: supplier `33987`, revenue `2296670.4443`
  - cardinalities: scan input `59,986,052`, scan output `2,265,119`, grouped output `100,000`
  - plan: one patched `lineitem` scan/group producer and two `Sink Scan` readers; clean has two scan/group subtrees
  - clean medians: A1 `1.45s`, A2 `1.44s` (`0.69%` drift)
  - patched median: `0.75s` versus combined clean median `1.44s` (`47.92%` faster)
  - materialized producer output: `1.95 MiB`

### QA decision

QA required: yes. This changes optimizer selection and an execution fan-out path for analytical SQL. The exact-head production terminal is covered by the embedded Q15 test and local SF10 A/B/A; remaining QA is fresh CI/nightly coverage and environment-level regression verification before closing #26131.

Non-goals: Window rewrites and #26067.


XuPeng-SH (changes requested):
I found two production correctness blockers in the new materialized execution path (reviewed head 7f04b1b0febd3dd79cf1d9b210ce07015322779f):

1. `Source.Next` returns the exact same owned `*batch.Batch` to every reader (`source.go:155-158`), but the downstream operator contract is not read-only. A concrete allowed shape is a SORT consumer: `cteConsumersFullyDrain` explicitly accepts SORT, while merge-order spill paths call `input.Batch.Clean(proc.Mp())` (`mergeorder/order.go:534-585`). One reader can therefore free the source-owned vectors while another reader is concurrently reading them; the source later cleans them again. This can cause corruption/use-after-free/double-free under normal spill pressure. Each consumer needs independent ownership (or an explicit immutable/ref-counted batch contract enforced end to end), with an integration test that drives one materialized CTE reader through the spill cleanup path while another still reads.

2. Eligibility only checks an estimate (`Outcnt * Rowsize <= 32 MiB`), but runtime `Append` turns actual size >64 MiB into a query error. Cardinality is not a hard upper bound, so stale/underestimated stats can make an SQL that previously succeeded via inline CTE expansion now fail solely because this optimization was selected. An optimization must preserve success semantics: spill/disable materialization using a hard bound/fallback safely, rather than exposing estimate error as a new fatal query limit. Please cover a deliberately underestimated fixed-width producer whose actual result crosses the runtime threshold.

Both issues are independent of the Q15 happy path and can occur in production; the current unit tests only read values directly and do not exercise destructive downstream ownership or estimate miss.

VioletQwQ-0:
Addressed both findings from the CHANGES_REQUESTED review on `7f04b1b0f` and revalidated on exact head `3164294f755`:

1. Every materialized reader now receives a caller-owned batch (`Dup` for retained batches, unmarshal-owned batches for spill). The regression uses the production `Source -> Merge -> MergeOrder` spill path: one reader retains data while the other destructively spills/cleans its copy, then both readers finish and the query MPool returns to baseline.
2. A runtime underestimate no longer turns an otherwise successful inline query into the former 64 MiB hard error. Retained query-MPool storage is bounded at 64 MiB and 4096 batches; later fixed-width batches spill sequentially to an anonymous query-scoped local file. Reader state stays O(reader count), terminal order remains buffered-data-before-error, and the last owner closes all retained batches and the spill file.

The reviewer-provided profitable `SAMPLE` SQL remains an explicit planner guard and produces zero shared `SINK_SCAN` nodes.

Exact-head validation:

- CR ownership/spill guards: `-race -count=20` PASS
- full affected package tests/build/vet and `git diff --check`: PASS
- embedded production Q15 test: PASS
- schema-v4 preflight: `PASS review=PASS validation=PENDING` (remote CI/re-review pending)
- strict SF10 A/B/A: correct result and cardinalities; clean medians `1.45s`/`1.44s` (0.69% drift), patched `0.75s` (47.92% faster); one patched scan/group producer and two readers

The PR remains Ready. Re-requesting review from XuPeng-SH now.

XuPeng-SH (approved):
Re-reviewed exact head `3164294f7555fdfbf2c5c532ca1b147fd3cbcb03`. Both previous correctness blockers are closed: every materialized reader receives caller-owned batches, including through destructive MergeOrder spill cleanup, and underestimated output now spills to a query-scoped anonymous file instead of failing at the former 64 MiB runtime limit.

I re-audited producer/reader ownership, data-before-error ordering, cancellation, last-owner cleanup, compile-owner safety cleanup, prepared-execution generations, planner determinism/full-drain guards, and local-only source placement. No new blocker found. Local validation passed: ownership/spill suites under `-race -count=20`, full `pkg/sql/plan` and `pkg/sql/compile` tests, the embedded Q15 production terminal, affected-package build/vet, and `git diff --check`. The exact head also merges cleanly with current main and CI is green.

PR #26332, opened 2026-07-28
fix: reject malformed batch encodings

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26325

Rejects truncated and malformed batch/vector binary encodings before any unsafe slice operation. It preserves zero-copy decoding while correctly handling reused vector slots and mpool-owned storage; callers that reuse batches now supply their existing mpool.

## Tests

- `go test -race -count=100 -run '^TestBatchUnmarshalWithAnyMpRejectsTruncatedData$' ./pkg/container/batch`
- `go test -race ./pkg/container/batch ./pkg/container/vector`
- `go test ./pkg/sql/colexec/multi_update ./pkg/sql/colexec/mergedelete ./pkg/vm/engine/disttae`
- `go build` and `go vet` for changed packages


gouhongshen (changes requested):
## Codex automated review

Loaded all PR discussion/review evidence: no submitted reviews or inline threads exist. One blocking malformed-input case remains; focused tests could not run because this checkout lacks cgo/libmo.dylib.

### P1 - Reject undersized fixed-vector payloads (pkg/container/vector/vector.go:829)
The bounds check only proves `dataLen` fits inside the vector frame; it never verifies that the bytes can represent the decoded class/type/length. A flat int64 vector claiming length 2 with `dataLen` 8 and otherwise valid framing returns nil here, and a batch can also claim rowCount 2. Later fixed-vector access converts `v.data` to two int64s; `util.UnsafeSliceCastToLength` detects the 8<16 mismatch and panics. This is a non-truncated malformed encoding that still terminates the caller, contrary to the PR’s reject-malformed-input goal. Validate the data length against the decoded vector metadata before assigning `v.data`, and add this undersized-payload case to the regression test.

gouhongshen (changes requested):
## Codex automated review

Previously blocking undersized fixed-vector payload is fixed: the new check rejects it and adds coverage. A distinct malformed-Type metadata bypass remains. Focused test execution was unavailable because `cgo/libmo.dylib` is absent locally.

### P1 - Reject fixed vectors whose serialized Type.Size understates the element width (pkg/container/vector/vector.go:870)
The new check trusts `decodedType.TypeSize()`, but `DecodeType` preserves the untrusted serialized `Type.Size` field. An otherwise well-framed flat int64 vector can claim `Type.Size=1`, `length=2`, and retain its 8-byte payload. Lines 875-878 then accept it because 8 >= 2. Later `GetFixedAtNoTypeCheck[int64]` converts the payload to two int64 values; `UnsafeSliceCastToLength` requires 16 bytes and panics. Validate the decoded type metadata against the OID’s canonical fixed width (and reject inconsistent/unknown types) before accepting the payload, with a regression test for this case.

gouhongshen (approved):
## Codex automated review

The earlier undersized-payload blocker was already fixed; this update also closes the forged Type.Size bypass by validating the OID and canonical element width, with targeted regression coverage. No blocking issues remain. Focused tests could not execute because this checkout lacks cgo/libmo.dylib.

XuPeng-SH (changes requested):
Deep review of exact head 106f6b3a8d7d1a24fcdf2c3e0337526b0091b73a found that malformed encodings can still be accepted and panic later. The new checks validate framing, but not the decoded cross-field invariants:

1. Batch row count is never checked against vector logical lengths. Starting from a valid one-row/int64 batch, changing only the first 8-byte batch rowCount to 2 leaves the vector frame internally valid, so UnmarshalBinaryWithAnyMp returns nil. Any consumer iterating `Batch.RowCount()` and reading vector row 1 then indexes past the one-element vector and panics. The current undersized-payload test changes both rowCount and vector length, so it misses this simpler bypass.

2. Null bitmap metadata is only checked as `declared data bytes <= nsp frame`. A bitmap can declare `len=64` with zero bitmap data; `ReadNoCopy` accepts it, and a later `Contains(0)` passes the declared-length check then indexes `data[0]`, panicking. Forged count/length can also silently change null semantics. Validate that bitmap storage covers its declared logical length (and that bitmap metadata is consistent with the vector), not only that its byte slice fits the outer frame.

3. Varlen descriptors are accepted without validating non-inline offset+length against `area`. A valid varchar frame with a forged Varlena offset unmarshals successfully, then `GetBytesAt`/comparison slices outside `area` and panics.

These are production-real corruption/untrusted-payload paths and the same class as #26325. Please establish the full post-decode safety invariants (including FLAT/CONSTANT class validity and constant-null representation) and add mutation tests that call representative accessors after successful decode. Fixing only the next observed byte boundary will keep producing one-case-at-a-time regressions.

volgariver6 (approved):

XuPeng-SH (changes requested):
Deep review of exact head dc24db14baad found two blocking failures, both reproduced locally against this head:

1. The prior batch/vector cross-field blocker is still open. Commit 88ff23f removed the row-count check wholesale to preserve an empty-vector encoding, but a non-empty short FLAT vector is still unsafe: mutate only a valid one-row/int64 batch rowCount from 1 to 2, and UnmarshalBinaryWithAnyMp returns nil with Batch.RowCount()==2 and Vecs[0].Length()==1. A normal rowCount-driven GetFixedAt(..., 1) then panics out of bounds. Preserve any genuinely required zero-length special case separately, but reject short non-empty vectors (and add the exact rowCount-only mutation/accessor regression).

2. Reuse with a changed vector count is broken at batch.go:211-231. The decoder assigns the incoming rowCount, then bat.Clean(mp) clears it. A valid one-row batch decoded into a reused target while changing from one vector to two returns nil with RowCount()==0, so downstream silently drops all rows. Through the public UnmarshalBinary path, the same branch can call Clean(nil) on an owned vector and panic before the later mp==nil guard. Cleanup/ownership validation must happen before publishing decoded metadata, and the parsed rowCount must survive reallocation; cover both mp-aware reuse and owned/no-mpool behavior.

The new null-bitmap representation and live-varlena area checks do close the other two findings from the previous review. The two reproductions above passed as bug-demonstration tests on the exact head; latest-main merge-tree is conflict-free.

pkg/container/vector/vector.go
```diff
@@ -794,55 +795,183 @@ func (v *Vector) MarshalBinaryWithBuffer(buf *bytes.Buffer) error {
...
+		}
+	} else if uint64(len(data)) != uint64(length)*uint64(typeSize) {
+		return moerr.NewInvalidInputNoCtx("invalid vector data size")
+	}
+	if typ.IsVarlen() {
+		values := types.DecodeSlice[types.Varlena](data)
+		for i := range values {
+			// Null varlen slots may retain stale offset/length metadata. The
+			// payload is never dereferenced, so only validate live values.
+			if nsp.Contains(uint64(i)) || values[i].IsSmall() {
```
XuPeng-SH: [P1] Validate array element alignment before treating inline or area descriptors as safe. For a one-row T_array_float64 vector, changing only the inline Varlena length byte from 8 to 1 leaves the frame valid here; UnmarshalBinary returns nil, then GetArrayAt[float64] panics because DecodeSlice requires the byte length to be a multiple of 8. The same issue applies to area-backed values and the other array element widths. Please validate every live array payload against its element width and cover successful decode followed by an accessor.

pkg/container/vector/vector.go
```diff
@@ -794,55 +795,183 @@ func (v *Vector) MarshalBinaryWithBuffer(buf *bytes.Buffer) error {
...
+		types.T_bool,
+		types.T_int8, types.T_int16, types.T_int32, types.T_int64,
+		types.T_uint8, types.T_uint16, types.T_uint32, types.T_uint64,
+		types.T_float32, types.T_float64,
+		types.T_decimal64, types.T_decimal128, types.T_decimal256,
+		types.T_date, types.T_time, types.T_datetime, types.T_timestamp, types.T_interval, types.T_year,
+		types.T_char, types.T_varchar, types.T_json, types.T_uuid,
+		types.T_binary, types.T_varbinary, types.T_enum, types.T_geometry, types.T_geometry32,
+		types.T_blob, types.T_text, types.T_datalink,
+		types.T_TS, types.T_Rowid, types.T_Blockid, types.T_tuple,
```
XuPeng-SH: [P1] Do not admit T_interval and T_tuple into a path that calls Type.IsVarlen. Both are listed here and TypeLen returns 0, but T.FixedLength has no case for either; a zero-length vector encoding with either OID therefore makes UnmarshalBinary panic at line 897 instead of returning an error. Remove or reject plan-only types here, or branch without calling IsVarlen, and add a no-panic decode test for every whitelisted OID.

XuPeng-SH (changes requested):
Re-reviewed exact head 0b95ac48035af0030f2333e41b4450bf8bfd6472. The previous reuse blockers are closed: rowCount now publishes after cleanup, changed vector counts preserve it, and owned vectors without an mpool are rejected. The new short-vector test also demonstrates that Batch.RowCount is intentionally independent from individual vector lengths, so I am not carrying forward my earlier cross-field length request.

Two vector-decoding blockers remain, both reproduced locally on this exact head:

1. Array Varlena payload lengths are not checked against element width. A frame can unmarshal successfully and then panic on its first array accessor.
2. canonicalVectorTypeSize admits T_interval and T_tuple, but validateVectorBinary immediately calls Type.IsVarlen; those OIDs panic in T.FixedLength instead of returning an error.

All changed package tests pass, but focused mutation tests reproduce both panic paths. Please reject these encodings during UnmarshalBinary and add accessor/no-panic regressions.

XuPeng-SH (changes requested):
[P1] Preserve distinct decoded columns when the reuse batch contains aliased vectors

`UnmarshalBinaryWithAnyMp` reuses each existing slot independently at `pkg/container/batch/batch.go:260-266`, but valid `Batch` values may contain the same `*Vector` in multiple slots (`SelectColumns` can create this directly, and `ReplaceVector` supports replacing all aliases). On the second aliased slot, `Free` resets the same object that already received the first decoded column, and `UnmarshalBinary` overwrites it with the second column. The call returns nil while both slots now point at the second column.

I reproduced this at the exact head with a two-column source `[1]`, `[2]` and a same-length target whose two slots share one owned int64 vector: after successful unmarshal, `target.Vecs[0] == target.Vecs[1]` and both read `2`. Please break repeated receiver aliases (while freeing the old owned vector exactly once) or decode into distinct receivers, and add a regression covering this reuse shape.

The previous array-payload-width and interval/tuple panic blockers are closed. Full batch/vector tests and focused race tests pass aside from this added reproducer.

XuPeng-SH (approved):
Deep re-reviewed exact head `419c07821cee94cbbb86c85873885708ef397fd5`: no blocking findings. The new receiver-identity tracking closes the prior aliasing blocker: each existing `*Vector` is reused at most once, later aliased slots receive distinct vectors, and the original owned receiver is released/reused exactly once. Both mpool-owned and borrowed/no-mpool alias shapes are covered.

Fresh local evidence: the two new alias regressions pass under `-race -count=10`; `go build` and `go vet` pass for `pkg/container/batch` and `pkg/container/vector`; final diff is clean, latest CI is green, and the PR remains MERGEABLE.

Issue #26311, opened 2026-07-28
[Bug]: TAE Merge PrepareCommit swallows phase-2 tombstone transfer errors


LeftHandCold:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

1b6d4308036f69e0486420fb175ab3de3a5c4ae2

### Other Environment Information

```Markdown
- Discovery method: static code review of the existing TAE Merge transaction path
- Discovery date: 2026-07-28
- No customer data or private logs are included
```

### Actual Behavior

`mergeObjectsEntry.PrepareCommit` runs the second phase of concurrent-delete transfer through `collectDelsAndTransfer`. When that call returns an error, the current code returns `nil` instead of propagating the error:

https://github.com/matrixorigin/matrixone/blob/1b6d4308036f69e0486420fb175ab3de3a5c4ae2/pkg/vm/engine/tae/tables/txnentries/mergeobjects.go#L460-L465

The swallowed error can come from:

- `TombstoneRangeScanByObject`;
- a missing or invalid old-row to new-row transfer mapping;
- looking up a created target Object;
- `DeleteByPhyAddrKeys`.

Because `PrepareCommit` reports success, the Merge transaction can continue committing the new Objects and dropping the source Objects even though phase-2 tombstones were not completely transferred.

The analogous flush path propagates the same class of error:

https://github.com/matrixorigin/matrixone/blob/1b6d4308036f69e0486420fb175ab3de3a5c4ae2/pkg/vm/engine/tae/tables/txnentries/flushTableTail.go#L294-L297

### Expected Behavior

Any error while collecting or transferring tombstones during Merge PrepareCommit must abort the Merge transaction. Source Objects must not be retired unless every relevant concurrent delete has either been transferred to the created Objects or the transaction fails closed.

### Steps to Reproduce

```Markdown
1. Start a Merge transaction that rewrites source Objects into one or more created Objects.
2. Commit a DELETE against a source Object after the Merge's first delete-collection phase and before its PrepareCommit phase.
3. Inject or trigger an error in the second collectDelsAndTransfer call, for example a TombstoneRangeScanByObject error, invalid transfer mapping, target Object lookup error, or DeleteByPhyAddrKeys error.
4. Observe that mergeObjectsEntry.PrepareCommit returns nil.
5. Verify that the Merge is allowed to continue instead of aborting.
[...]

PR #26333, opened 2026-07-28
fix: propagate merge tombstone transfer errors

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26311

Propagates merge phase-2 tombstone-transfer errors through transaction prepare and preserves the first prepare error when phase 2 has appended additional transaction entries. This makes the transaction manager roll back instead of committing a partially transferred merge.

## Tests

- `go test -race -count=14 -run '^(TestRollbackMergeInQueue|TestRollbackMergeOnTransferError|TestRollbackMergeAfterTransferredDeleteError)$' ./pkg/vm/engine/tae/db/test`
- `go test -race ./pkg/vm/engine/tae/db/test ./pkg/vm/engine/tae/tables/txnentries`
- `go build` and `go vet` for touched TAE packages


XuPeng-SH (approved):
Deep-reviewed at head 90d16303d2fd0be6c3d2ac80ec94835010d98dee. Phase-2 transfer errors now remain terminal even when transfer appended txn entries: txnTable returns the first prepare error before the second pass can overwrite it, so TxnManager enters rollback and covers both the merge entry and appended tombstone entries. Transfer-table, page/delete-map, metadata, and object cleanup ownership remains closed. The three focused regressions passed under -race for 14 iterations; full race tests for all four affected packages, go vet, and diff checks also passed.

Issue #26344, opened 2026-07-29
Support rollback holes and persisted abort metadata in appendable objects


jiangxinmeng1:
## Background

`AppendMVCCHandle.GetVisibleRowLocked` already computes rollback holes, but `memoryNode.Scan`, `FillBlockTombstones`, and `CollectObjectTombstoneInRange` discard them. Rollback also removes the corresponding `AppendNode`. Once append data is applied before commit, this leaves physical rows and PK-index entries without an MVCC owner.

The object format reserves `SEQNUM_ABORT`, but abort metadata is not supported end to end: object reads panic on it and appendable block reads load only `commitTS`.

## Expected behavior

- Keep rolled-back append ranges in the MVCC chain and mark them aborted.
- Consume rollback holes in memory scans and tombstone collection.
- Persist an abort column alongside commit timestamps for new appendable objects.
- Filter aborted rows in objectio/blockio snapshot reads, persisted dedup and PK checks, tombstone scans, merge/flush input, logtail, WAL replay, publication, dump/restore, and custom object readers.
- Read legacy commit-only objects with a missing abort column interpreted as `false`.

## Design requirement

Introduce a shared special-column layout resolver based on object metadata. Consumers must not infer `commitTS` or `abort` from the last column or total column count.

Supported layouts must include:

- `rowid -> commitTS -> abort`
- `commitTS -> rowid -> abort`
- legacy `commitTS` only

## Acceptance criteria

- Rolled-back physical rows retain an aborted `AppendNode` owner and are invisible to scans and dedup.
- New appendable objects persist `commitTS + abort`; legacy objects remain readable.
- Aborted data and tombstone rows do not reach merge output, logtail publication, snapshot reads, or restored dump-table data.
- WAL V1 remains readable and the new version preserves aborted state.
- Unit, DB flush/replay, and race tests cover new and legacy formats.

PR #26345, opened 2026-07-29
Codex/abort column

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26344

This PR adds end-to-end support for rollback holes and persisted abort metadata
  in appendable objects.

  - Keeps rolled-back AppendNode ranges in the MVCC chain and marks them aborted,
    preserving ownership of already-applied physical rows and PK-index entries.

  - Makes memory scans and tombstone collection paths consume rollback holes.
  - Adds a unified special-column layout resolver instead of inferring commitTS
    from the last column or column count.

  - Persists and reads the abort column for new appendable objects.
  - Filters aborted rows across snapshot reads, deduplication, tombstone
    processing, merge/flush, logtail, publication, and custom object readers.

  - Extends WAL serialization to preserve the aborted state while retaining V1
    compatibility.

  - Supports legacy objects containing only commitTS; a missing abort column is
    interpreted as false.

  - Supports both valid special-column layouts:
      - rowid → commitTS → abort
      - commitTS → rowid → abort


gouhongshen (changes requested):
## Codex automated review

The new persisted abort metadata is not stripped by the appendable-object CDC reader, causing a storage-only boolean vector to leak into returned data batches. No prior substantive review findings existed to suppress.

### P1 - Do not expose the persisted abort column through CDC reads (pkg/vm/engine/tae/tables/mnode.go:180)
This change appends the abort vector and seqnum to persisted appendable objects, producing a layout such as `[user..., rowid, commitTS, abort]`. The persisted CDC path loads all physical vectors via `prefetchObjects`/`ioutil.LoadOneBlock`, then `AObjectHandle.getNextAObject` calls `updateDataBatch`. That function removes only Rowid vectors, moves the TS vector to the end, and retains every other type in its default branch. Therefore the new bool is returned as `[user..., abort, commitTS]`; `AObjectHandle.next` clones every vector and does not remove it later. This breaks the existing CDC batch schema by exposing hidden storage metadata. Strip the abort vector using its special-column metadata before normalizing the batch, and add a persisted appendable-object CDC regression test.

fengttt (approved):

gouhongshen (changes requested):
## Codex automated review

The sole prior blocker—the CDC reader exposing the persisted abort column—is fixed by layout-aware normalization and a regression test. No prior blocker remains; three newly identified correctness and availability regressions are blocking.

### P1 - Filter aborts in cache-backed block reads (pkg/vm/engine/tae/blockio/read.go:1044)
Abort filtering is added only to `readBlockData`, but ordinary reads bypass it. Full scans use `BlockDataReadInner` → `LoadColumnsDataInto`, while cached point/range lookups use `ReadDataByFilter` → `LoadColumnDataBySearch`; both helpers load and filter only `SEQNUM_COMMITTS`. Selected-row materialization then passes nil visibility. Consequently an `abort=true` row persisted by this PR is returned by normal scans and point reads. The optimized paths arrived from main, but this regression exists only after introducing persisted abort rows. Load and filter `SEQNUM_ABORT` in both cache-backed paths, treating legacy const-null as false, and cover full scans and cached searches.

### P1 - Do not treat aborted persisted tombstones as deletes (pkg/vm/engine/tae/tables/functions.go:471)
`containsABlkFuncFactory` nulls the candidate data rowid at line 470 before checking whether the matched tombstone row is aborted. Thus even a single rolled-back tombstone in a persisted appendable object makes `findDeletes` classify the original row as deleted. A later insert of the same primary key can then pass dedup and create duplicate data. Only null the rowid after selecting a non-aborted matching tombstone; if multiple physical matches exist, continue searching for a live one.

### P1 - Handle multiple tombstone offsets left by rollback retries (pkg/vm/engine/tae/index/indexwrapper/mutindex.go:249)
The new `ErrNotFound` handling occurs after `Contains` asserts that an ART key has exactly one offset. `PrepareRollback` now retains the aborted append node and index entry; `GetDuplicatedRows` skips that owner, allowing a retry delete of the same rowid, and `BatchUpsert` appends the retry's offset to the same ART key. Subsequent `findDeletes` calls `Contains`, receives two offsets, and panics before either owner is checked. Iterate all offsets, skip aborted owners, and mark the key deleted only when a non-aborted owner remains.

gouhongshen (changes requested):
## Codex automated review

The prior CDC blocker remains fixed. Commit 63c9a898 fixes all three blockers from the previous review: cache-backed reads now filter aborted rows, persisted tombstone matching skips aborted duplicates, and MutIndex handles rollback-retry offsets. One distinct mixed-version compatibility blocker remains. Focused tests could not start because the controlled CGo prerequisites were absent; both diff ranges pass `git diff --check`.

### P1 - Gate the new object layout from mixed-version readers (pkg/vm/engine/tae/tables/mnode.go:182)
This unconditionally adds `SEQNUM_ABORT` to newly persisted appendable objects without an object-format or cluster-capability gate. During a rolling upgrade, an old CN can therefore read an object flushed by a new TN. The base-version CDC reader loads every physical column with `LoadOneBlock`; its `updateDataBatch` retains every non-Rowid/non-TS vector, exposing the abort bool as user data, while `updateTombstoneBatch` ignores that bool and can publish an aborted tombstone as live. The base `BlockDataReadBackup` also decodes the final vector as commitTS, which is now the bool abort vector. This is distinct from the prior CDC finding: the current-head reader is fixed, but old readers remain reachable. Delay emitting the new layout until reader support is enforced, or introduce a versioned/capability-gated format that old readers safely interpret.

gouhongshen (changes requested):
## Codex automated review

The prior CDC, cache-read visibility, tombstone matching, and MutIndex blockers remain fixed. The mixed-version blocker is also fixed by the v9 capability gate: pre-v9 deployments compact abort holes and retain the legacy layout. Two newly identified row-coordinate regressions remain blocking. Both diff ranges pass `git diff --check`; focused tests could not start because `cgo/libmo.dylib` is absent.

### P1 - Remap tombstones after compacting abort holes (pkg/vm/engine/tae/blockio/read.go:400)
`visibleRows` compacts the loaded batch before tombstones are fetched, but `BackupDeltaLocDataSource.GetTombstones` builds its bitmap from each rowid's original physical row offset. Line 415 then applies those original offsets to the compacted batch. For original rows `[0 live, 1 aborted, 2 live+tombstoned, 3 live]`, the first shrink produces `[0,2,3]`; tombstone mask `{2}` then removes original row 3 while deleted row 2 survives. `ReWriteCheckpointAndBlockFromKey` reaches this path, so backup rewriting can resurrect deleted data and discard live data. Combine both masks before one shrink or remap tombstone offsets through `visibleRows`, with an interior-abort regression test.

### P1 - Preserve original row offsets through publication filtering (pkg/publication/filter_object_batch.go:442)
The new abort condition permits an arbitrary interior row to be compacted, but `createObjectFromBatch` subsequently builds `rowOffsetMap` from post-compaction indexes and labels them as original offsets. For original rows `0=live(key20), 1=aborted, 2=live(key10)`, sorting after compaction produces map `{1:0, 0:1}` instead of `{2:0, 0:1}`. A tombstone for upstream offset 2 therefore misses the map; `rewriteTombstoneRowidsBatch` changes its object ID but leaves offset 2, which is outside or points at the wrong row in the two-row downstream object. This independently corrupts CCPR delete rewriting. Carry original-offset provenance through filtering and sorting, and cover an aborted interior row plus a tombstone targeting a later live row.

gouhongshen (approved):
## Codex automated review

The two blockers from the previous review are fixed: backup reads now combine abort and tombstone masks in original physical-row coordinates, and publication filtering preserves original offsets through compaction and sorting. Earlier blockers involving CDC propagation, cache-backed reads, aborted tombstone matching, duplicate-key retries, and mixed-version object layout remain fixed. The intervening merge only resolves the MORPC version collision by assigning abort-column support to v10; no upstream-only change is attributed to this PR. No current finding remains. `git diff --check` passes; focused tests, build, and vet could not complete because the checkout lacks `cgo/libmo.dylib` and required third-party C headers.

pkg/vm/engine/tae/tables/mnode.go
```diff
@@ -158,27 +158,51 @@ func (node *memoryNode) getDataWindowOnWriteSchema(
...
+		// Old readers interpret every non-rowid/non-TS physical column as user
+		// data. Keep their commitTS-only layout and remove rollback holes before
+		// the batch can be flushed or published through logtail.
+		aborts := vector.MustFixedColWithTypeCheck[bool](abort.GetDownstreamVector())
+		for row, aborted := range aborts {
+			if aborted {
+				inner.Delete(row)
+			}
+		}
+		inner.Compact()
```
LeftHandCold: **P1 — Preserve physical row offsets in the pre-v10 layout.** When the protocol is still v9, deleting and compacting rollback holes here shifts every later live row within the same appendable object. For example, `[row 0 live, row 1 aborted, row 2 live]` is flushed as two physical rows, so original row 2 becomes on-disk offset 1. Any tombstone already committed for `(objectID, offset 2)` is not remapped by this snapshot-flush path, while ordinary persisted reads reconstruct RowIDs from the new on-disk offsets in `buildRowidColumn`; that tombstone therefore misses and the deleted row can become visible again. This is the supported rolling-upgrade branch (`PersistedAObjectAbortSupported() == false`), not a malformed-input case. Please keep the legacy layout without changing physical row coordinates (or transfer/remap all affected tombstones), and add a regression with an interior abort hole plus a tombstone targeting a later live row across a v9 flush.

LeftHandCold (changes requested):
The latest commit correctly restores the leading-RowID contract for persisted change batches, and the previous abort-filtering fixes remain intact. One production-reachable mixed-version data-correctness blocker remains in the v9 compatibility path.

XuPeng-SH (changes requested):
Deep review of exact head `30f350d169e8173cd9ced765214627d0b4c2991d`.

The latest commit correctly restores the leading-RowID contract for persisted change batches, and the earlier abort-column fixes remain intact. One production-reachable data-correctness blocker remains:

**P1 — Preserve physical RowID coordinates in the pre-v10 compatibility layout** (`pkg/vm/engine/tae/tables/mnode.go:186`)

When `PersistedAObjectAbortSupported()` is false, `getDataWindowOnWriteSchema` removes abort holes with `inner.Compact()`. This changes the physical offset of every later live row. The snapshot-flush path then writes that compacted batch for the same appendable object and updates its location/stats, while normal persisted reads reconstruct RowIDs from the new physical positions in `buildRowidColumn`. Existing tombstones still reference the original physical offsets; the merge transfer map cannot repair reads of the old aobject required by historical snapshots.

Concrete counterexample: original rows `[offset 0 live, offset 1 aborted, offset 2 live+tombstoned]` are flushed in v9 as physical rows `[0, 1]`. Original row 2 is now reconstructed as offset 1, but its tombstone remains `(objectID, offset 2)`, so the deleted row can become visible again.

Please keep physical coordinates stable in the compatibility path, or comprehensively remap every coordinate consumer before publishing the new location. Add an end-to-end v9 regression with an interior abort hole, a tombstone for a later live row, a table-tail snapshot flush, and a historical read proving that row remains invisible.

XuPeng-SH (changes requested):
The v9 RowID-coordinate blocker is fixed: rollback holes now retain their physical offsets and are encoded with `UncommitTS`, and the new historical-snapshot regression exercises the later tombstone correctly. One independent persisted-abort visibility path remains blocking.

**P1 — Filter aborted rows when backup readers use no snapshot cutoff** (`pkg/vm/engine/tae/blockio/read.go:375`)

`BlockDataReadBackup` resolves and filters `SEQNUM_COMMITTS`/`SEQNUM_ABORT` only inside `if !ts.IsEmpty()`. Its empty-TS branch applies tombstones alone. This PR makes rollback rows persistable, so an empty cutoff must still exclude rows that can never be visible: `abort=true` in the v10 layout and `commitTS=UncommitTS` in the new v9 compatibility layout.

This is production-reachable, not a helper-only edge: `SnapshotMeta.GetSnapshot` calls this function with `types.TS{}` at `snapshot.go:1087`, and PITR/ISCP use `specialTableInfo.processObjects`, which does the same at `snapshot.go:418`. Those calls also request only selected user columns, so the current load does not include either special vector. A rolled-back append to these system tables can therefore be processed as live snapshot/PITR/ISCP metadata, affecting GC and checkpoint retention decisions.

I reproduced the leak on exact head `04a43fd7cae2` by extending the existing backup-read regression's persisted rows `[0 live, 1 abort, 2 tombstoned, 3 live]`: with a non-empty cutoff it returns `[0,3]`, but the same call with `types.TS{}` returns `[0,1,3]`. The PR's focused blockio/memory-node/v9 end-to-end tests pass without that assertion.

Please make abort visibility unconditional while making only the ordinary commit-time cutoff optional. For selected-column reads, load the needed special columns without exposing them in the returned schema (or pass a real cutoff through all of these callers). Cover empty-cutoff reads for both v10 `abort=true` and v9 `UncommitTS`, plus at least one SnapshotMeta consumer regression.

LeftHandCold (changes requested):
The latest main merge drops upstream fixes and would regress unrelated production paths. fa10ccb3 has 607a9263 (the current PR base) as its second parent, but its tree is byte-for-byte identical to the first parent a718fc1a. As a result, the PR now carries 105 changed files and effectively reverts all four commits added on main since 8fc0655a: #26641 CDC resume recovery, #26497 session-reset request fencing, #26635 ONLY_FULL_GROUP_BY enforcement, and #26700 snapshot-restore role-privilege ID rebinding. Concrete examples are the deletion of pkg/sql/plan/mysql_full_group_by.go and pkg/frontend/snapshot_catalog_restore.go. These are normal-business regressions unrelated to abort metadata. Please rebase or redo the main merge while retaining the upstream side, then verify the PR diff is limited to the rollback-hole change. The a718fc1a backup-read change itself addresses the previous empty-TS issue for both v9 UncommitTS and v10 abort-column layouts.

XuPeng-SH (changes requested):
Deep review of exact head `fa10ccb333052e672164ef215766325a6c17f7d5`.

The `a718fc1a` change fixes the previous empty-cutoff blocker: backup reads now filter both v10 `abort=true` rows and v9 `UncommitTS` rows for full and selected-column reads, and the new focused and SnapshotMeta regressions pass. One independent P1 blocker remains in the final merge commit.

**P1 — Redo the main merge without discarding the upstream tree**

`fa10ccb3` records current base `607a9263` as its second parent, but its tree `c0dfd7d4` is byte-for-byte identical to first parent `a718fc1a`. Therefore the merge accepted none of the upstream side. Relative to the current base, this PR now changes 105 files (+3358/-4680) and reverts all four commits added on main since common ancestor `8fc0655a`: #26700 snapshot-restore role privilege ID rebinding, #26635 ONLY_FULL_GROUP_BY enforcement, #26497 session-reset request fencing, and #26641 CDC resume recovery.

Concrete evidence includes deleting `pkg/frontend/snapshot_catalog_restore.go`, `pkg/sql/plan/mysql_full_group_by.go`, `pkg/tests/issues/issue_26640_test.go`, and their regression coverage. The full PR `diff --check` also fails on old `cte.sql` whitespace restored by the same merge error. Merging this head would regress unrelated production behavior even though branch CI is green because the corresponding tests were reverted too.

Please rebase or recreate the merge while retaining the upstream side, then verify that the PR diff is limited to the rollback-hole/abort-column feature.

pkg/vm/engine/tae/logtail/tools.go
```diff
@@ -147,24 +155,47 @@ func TombstoneChangeToLogtailBatch(src *containers.BatchWithVersion) *containers
...
 
+func filterAbortedLogtailRows(src *containers.BatchWithVersion) {
+	abortPos := -1
+	for i, seqnum := range src.Seqnums {
+		if seqnum == objectio.SEQNUM_ABORT {
+			abortPos = i
+			break
+		}
+	}
+	if abortPos != -1 && !src.Vecs[abortPos].IsConstNull() {
```
LeftHandCold, line 190: **P1 — Filter the v9 rollback sentinel before emitting logtail.** When `PersistedAObjectAbortSupported()` is false, `getDataWindowOnWriteSchema` deliberately omits `SEQNUM_ABORT` and rewrites every rollback hole to `commitTS=txnif.UncommitTS` while retaining its physical row. This helper is shared by `TableLogtailRespBuilder.visitObjData`, so this branch finds no abort vector and emits the rollback row in both data and tombstone logtail. CN then inserts it into `PartitionState` with `Time=MaxTs`; `PKExistInMemBetween` checks only `Time >= from`, so normal PK rechecks can keep reporting a change for a key that only belonged to an aborted append. The row also survives appendable-object GC because that path removes only entries with `Time <= DeleteTime`. Please resolve the commit-TS vector by seqnum and also filter `UncommitTS` (at least when the abort column is absent), with a v9 logtail-to-PartitionState regression.

pkg/vm/engine/tae/tables/functions.go
```diff
@@ -439,27 +462,36 @@ func containsABlkFuncFactory[T types.FixedSizeT](comp func(T, T) int) func(args
...
-				if commitTS.GT(&startTS) {
-					ts, err := delsFn(v1, commitTS)
-					if err != nil {
-						return err
+				first := row
+				for first > 0 && comp(vs[first-1], v1) == 0 {
+					first--
+				}
+				for row = first; row < len(vs) && comp(vs[row], v1) == 0; row++ {
+					if isAborted(abortVec, row) {
```
LeftHandCold, line 481: **P1 — Treat the v9 `UncommitTS` encoding as aborted in persisted tombstone matching.** `LoadPersistedMVCC` synthesizes a const-null abort vector for the v9 layout, so `isAborted` returns false for a rollback hole even though its commit TS is `txnif.UncommitTS`. `findDeletes` reaches this closure through `aobject.Contains`; the match is marked deleted and, because MaxTs is later than the reader start, `delsFn` is invoked. An aborted tombstone has no committed transfer-delete mapping, so this returns `ErrTxnWWConflict`. Retrying does not help because the same persisted rollback row remains, which can indefinitely reject an insert whose only matching delete was rolled back. Please skip a match when either the abort bit is true or the legacy commit TS is `UncommitTS`, and cover a v9-flushed aborted tombstone in the persisted `Contains` regression.

LeftHandCold (changes requested):
Deep review of exact head `3e95f76d23cb7dbbf2cc3fa8e084748742f46cc3`. The previous bad-merge blocker is fixed: the PR is again limited to the rollback-hole feature and retains the upstream fixes. Earlier blockers around CDC, cache reads, abort-aware tombstones, duplicate offsets, RowID preservation, publication rewriting, and empty-cutoff backup reads also remain fixed. Two production-reachable v9 compatibility blockers remain. Both occur because v9 deliberately encodes a rollback hole as `commitTS=UncommitTS` without an abort column, while these consumers recognize only the abort column. `git diff --check` passes. I did not launch broad BVT/UT jobs, per the requested code-review-only/resource-conscious scope; the repository CI is still running.

LeftHandCold (approved):
Re-reviewed current head a89842070830fc5e8413145dc39581d438673a6c.

The latest commit closes both previously requested correctness changes:
- logtail generation now filters both v10 abort=true rows and v9 commitTS=UncommitTS rollback sentinels before publishing data/tombstone batches;
- persisted appendable-block duplicate checks and tombstone Contains now skip the v9 rollback sentinel before matching/nulling candidates, for both fixed and varlen paths.

I also re-checked the remaining commitTS/abort consumers and their boundary semantics. Cache-backed reads and change/range paths use finite snapshot/range upper bounds, so the MaxTS sentinel is excluded there; the unbounded logtail and persisted-dedup cases are now explicitly covered. Batch/vector ownership and cleanup paths remain balanced, and the added row scans are bounded by batch size.

No further blocking production issue found. git diff --check is clean. I did not run broad local UT/BVT to avoid duplicating the resource-heavy CI; several CI jobs are still in progress.

XuPeng-SH (approved):
Deep review of exact head `3afb38a24b86992fd73c6c26b291fee9f04f38eb`.

The previous blockers are closed. The bad main merge is corrected; the v9 path preserves physical RowID coordinates and represents rollback holes with `commitTS=UncommitTS`; empty-cutoff backup reads filter both the v9 sentinel and v10 abort metadata; and the latest change closes the remaining unbounded logtail plus persisted dedup/tombstone-Contains gaps.

I re-audited the same-shape commitTS consumers. Cache-backed reads, CDC/change collection, publication, snapshot, and backup range paths use finite snapshot/range bounds, so the v9 MaxTS sentinel stays invisible there. The consumers that can use an unbounded upper bound now reject the sentinel explicitly. Hidden-column layout resolution supports legacy commitTS-only objects and both RowID placements without exposing abort metadata.

Lifecycle audit also closes: rollback retains a single MVCC owner marked aborted; reader waits release the MVCC lock before waiting on txn completion; retained holes are bounded by the appendable object and reclaimed through its flush/object lifecycle; vector/cache release paths are balanced.

Validation: all 19 focused regression tests across objectio, blockio, logtail replay, publication, TAE tables/index/update/DB packages pass under the controlled CGo environment; `git diff --check` is clean; required CI is green. No blocking issue remains.

PR #26347, opened 2026-07-29
fix: align rowids for nullable index updates

### LINKED ISSUE #26334: [Bug]: nullable index UPDATE keeps Rowid from filtered NULL row
## Description

For an UPDATE input batch with a nullable index key, `preinsertunique` and the
single-column path of `preinsertsecondaryindex` compact the index key and
primary-key vectors but copy the complete Rowid vector.

This affects:

- nullable single-column unique indexes;
- nullable composite unique indexes;
- nullable single-column secondary indexes.

The composite non-unique secondary-index path is a control case: it retains
NULL-containing rows and keeps all vectors aligned.

The affected paths produce vectors with different lengths and associate the
surviving index entry with the Rowid of the filtered NULL row.

## Environment

- Branch: `main`
- Commit: `cad24c348791e5f05e517e0a49111d965ddb48c5`

## Minimal case

Use a two-row UPDATE batch:

| row | primary key | first index column | second index column | Rowid |
| --- | --- | --- | --- | --- |
| 1 | 10 | NULL | 100 | rowid-1 |
| 2 | 20 | 200 | 201 | rowid-2 |

For the single-column case, only the first index column is configured. For the
composite case, both index columns are configured.

## Expected result

For a unique index, the NULL-containing row is filtered and every output
vector remains aligned:

- index key: one serialized entry for `(200, 201)`
- primary key: `[20]`
- Rowid: `[rowid-2]`
- all vector lengths: `1`

## Actual result

The output batch reports one row, but its vectors are:

- index key: one serialized entry (length 1)
- primary key: `[20]` (length 1)
- Rowid: `[rowid-1, rowid-2]` (length 2)

The first visible Rowid is therefore `rowid-1`, which belongs to the filtered
NULL row rather than the surviving index entry.

The single-column tests reproduced the result 3/3 times in both affected
packages. The composite unique-index test also reproduced it 3/3 times:

```text
go test ./pkg/sql/colexec/preinsertunique \
  -run TestAuditCompositeNullableUpdateKeepsAlignedRowID \
  -count=3 -v

expected Rowid vector length: 1
actual Rowid vector length:   2
```

The composite non-unique secondary-index control passed 3/3 and retained two
aligned rows:

```text
go test ./pkg/sql/colexec/preinsertsecondaryindex \
  -run TestAuditCompositeNullableUpdateKeepsAllRowsAligned \
  -count=3 -v
```

## Code path

The affected operators follow the same pattern:

- `CompactSingleIndexCol` or `SerialWithCompacted` returns the bitmap for
  nullable index rows.
- `CompactPrimaryCol` applies that bitmap to the primary-key vector.
- `UnionBatch` copies the entire Rowid input and does not apply the bitmap.

Locations:

- `pkg/sql/colexec/preinsertunique/preinsertunique.go`
- `pkg/sql/colexec/preinsertsecondaryindex/preinsertsecondaryindex.go`

## Impact

Nullable index maintenance during UPDATE can construct an internally
inconsistent batch and pair a surviving index key/primary key with the Rowid
of a different row. Downstream index maintenance can therefore act on the
wrong row.

XuPeng-SH:
## What changed

- Compact update RowIDs with the same NULL bitmap already used for unique-index keys and primary keys.
- Apply the same alignment rule to single-column secondary-index preinsert; keep the composite secondary-index behavior unchanged.
- Preserve the existing `UnionBatch` fast path when no indexed value is NULL.
- Add white-box coverage for single/composite indexes and mixed/all/zero surviving rows.
- Reuse the existing multi-table `update_index` BVT data to reproduce the SQL-visible failure without adding redundant fixtures.

## Root cause

During an UPDATE of a nullable unique index, the index key and primary-key vectors discard rows whose indexed value becomes NULL, but the RowID vector was copied without the same compaction. The duplicate-check pipeline could therefore compare a surviving key with the RowID of a different row and reject a legal self-update as a duplicate.

## User impact

Fixes false `Duplicate entry` errors for batch updates where one indexed row becomes NULL while other indexed rows retain their existing keys.

## Performance and risk

- The common no-NULL path still uses the original bulk `UnionBatch`; the added branch only checks whether the bitmap is empty.
- Local 8K-RowID microbenchmark, 5 runs: original bulk path 2.296 µs/op vs guarded no-NULL path 2.300 µs/op, both 0 alloc/op (difference within measurement noise).
- A 25%-NULL batch takes about 20.6 µs per 8K rows, 35 B/op and 1 alloc/op; this work is required only when compaction is needed, and the selection buffer is pooled.
- No schema, wire, storage, public API, concurrency, or lifecycle changes.
- Pooled selections are returned on both success and error paths.

## Validation

- `go build` for the modified Go packages
- `go vet` for the modified Go packages
- deterministic CGo test wrapper over `preinsertunique`, `preinsertsecondaryindex`, and `sql/util`
- targeted race tests, 20 repetitions
- `make ARCHSIMD=0 build-with-prebuilt-native`
- real-server BVT: `update_index.test` passes 148/148
- regression oracle: clean latest `main` fails the reproducer with `Duplicate entry` (143/148 after cascading failures), while this change passes 148/148

Fixes #26334

### DIFF
--- a/pkg/sql/colexec/preinsertsecondaryindex/preinsertsecondaryindex.go
+++ b/pkg/sql/colexec/preinsertsecondaryindex/preinsertsecondaryindex.go
@@ -128,7 +128,14 @@ func (preInsertSecIdx *PreInsertSecIdx) Call(proc *process.Process) (vm.CallResu
 
 	if isUpdate {
 		rowIdInBat := len(inputBat.Vecs) - 1
-		if err = preInsertSecIdx.ctr.buf.Vecs[rowIdColPos].UnionBatch(inputBat.Vecs[rowIdInBat], 0, inputBat.Vecs[rowIdInBat].Length(), nil, proc.Mp()); err != nil {
+		if bitMap.IsEmpty() {
+			err = preInsertSecIdx.ctr.buf.Vecs[rowIdColPos].UnionBatch(
+				inputBat.Vecs[rowIdInBat], 0, inputBat.Vecs[rowIdInBat].Length(), nil, proc.Mp())
+		} else {
+			err = util.CompactRowIdCol(
+				inputBat.Vecs[rowIdInBat], preInsertSecIdx.ctr.buf.Vecs[rowIdColPos], bitMap, proc)
+		}
+		if err != nil {
 			return result, err
 		}
 	}
--- a/pkg/sql/colexec/preinsertsecondaryindex/preinsertsecondaryindex_test.go
+++ b/pkg/sql/colexec/preinsertsecondaryindex/preinsertsecondaryindex_test.go
@@ -161,6 +161,94 @@ func TestPreInsertSecondarySingleVarcharUsesSizedUkType(t *testing.T) {
 	require.Equal(t, 1, arg.ctr.buf.Vecs[indexColPos].Length())
 }
 
+func TestPreInsertSecondaryUpdateKeepsRowsAligned(t *testing.T) {
+	testCases := []struct {
+		name         string
+		indexColumns [][]int64
+		nullRows     [][]bool
+		expectedRows []int
+	}{
+		{
+			name:         "single column filters nulls",
+			indexColumns: [][]int64{{100, 200, 300, 400}},
+			nullRows:     [][]bool{{true, false, true, false}},
+			expectedRows: []int{1, 3},
+		},
+		{
+			name: "composite key retains nulls",
+			indexColumns: [][]int64{
+				{100, 200, 300, 400},
+				{101, 201, 301, 401},
+			},
+			nullRows: [][]bool{
+				{true, false, false, false},
+				{false, false, true, false},
+			},
+			expectedRows: []int{0, 1, 2, 3},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			proc := testutil.NewProc(t)
+			pks := []int64{10, 20, 30, 40}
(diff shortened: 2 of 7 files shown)


ck89119 (approved):
Reviewed the complete diff at 285c39cb0fa34d3ed08893610a41fae830284777.

The nullable-index UPDATE paths now apply the same compaction bitmap to the index key, primary key, and Rowid vectors. I checked single-column and composite unique indexes, the single-column secondary-index path, the composite secondary-index control path, empty/all-survive/mixed selections, operator buffer reuse, selection-vector ownership, and error cleanup. No actionable correctness or lifecycle issues found.

Local validation passed: focused race stress (100 repetitions per new regression test), full race runs for both owning operator packages, pkg/sql/util tests, build, vet, and diff checks. The latest applicable CI run is green.

XuPeng-SH (commented):
Deep-reviewed exact head `031e2b711e1f98a830032b0c000e48614339bf69`; recommendation: APPROVE, with no blocking findings. GitHub prevents this authenticated account from approving because `XuPeng-SH` is also the PR author. The nullable-row bitmap is now applied consistently to index key, primary key, and Rowid vectors; empty/all-filtered and composite non-unique control paths preserve vector lengths and row identity. Selection-slice pooling is returned on errors, and operator buffer reuse remains safe. Focused update-alignment tests passed under `-race -count=20`; full preinsertunique, preinsertsecondaryindex, and sql/util packages passed under race.

PR #26348, opened 2026-07-29
fix(parser): preserve uint64 auto-increment values

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26314

The MySQL lexer represents integral literals up to `MaxInt64` as `int64` and
larger values as `uint64`. The table-level `AUTO_INCREMENT` grammar action
previously asserted every integral literal as `int64`, causing an internal
parser panic for valid unsigned values.

This change accepts both lexer representations and preserves the value as
`uint64`. It also adds parser and formatting coverage for `MaxInt64`,
`MaxInt64 + 1`, `MaxUint64 - 1`, and `MaxUint64`.

Validation:

- Focused parser test with `-race -count=100`
- Full MySQL parser package with `-race -count=1`
- Full MySQL parser package tests with 74.0% package coverage
- New conversion helper coverage: 75.0%
- `go test ./pkg/sql/parsers/...`
- `go build ./pkg/sql/parsers/dialect/mysql`
- `go vet ./pkg/sql/parsers/dialect/mysql`


XuPeng-SH (approved):
Deep-reviewed exact head `6dcb925d2b94b3ce471abb32ad5445bb221821eb`; no blocking findings. The grammar action now handles both integral representations emitted by the lexer without truncating unsigned AUTO_INCREMENT values, the generated parser stays in sync with the grammar source, and AST formatting preserves all uint64 boundaries. Focused boundary tests passed under `-race -count=20`; the full MySQL parser package also passed under race.

Issue #21795, opened 2025-05-06
[Bug]: drop relation takes too much in-use memory


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

latest

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

<img width="1791" alt="Image" src="https://github.com/user-attachments/assets/828dbdcf-3512-4c0c-8832-fc5600b172bc" />

### Steps to Reproduce

```Markdown
package main

import (
	"database/sql"
	"fmt"
	"log"
	"sync"
	"time"

	_ "github.com/go-sql-driver/mysql"
)

func create_account(db *sql.DB, account string) {
	_, err := db.Exec(fmt.Sprintf("CREATE ACCOUNT %s admin_name = 'test_account' identified by '111';", account))
	if err != nil {
		log.Fatal(err)
	}
}

func drop_account(db *sql.DB, account string) {
	_, err := db.Exec(fmt.Sprintf("DROP ACCOUNT IF EXISTS %s;", account))
	if err != nil {
		log.Fatal(err)
	}
}

func do_checkpoint(db *sql.DB) {
	start := time.Now()
	_, err := db.Exec("SELECT mo_ctl('dn','checkpoint','');")
	fmt.Printf("checkpoint cost: %v:%v\n", err, time.Since(start))
	if err != nil {
		log.Fatal(err)
	}
}

func main() {
	db, err := sql.Open("mysql", "dump:111@tcp(127.0.0.1:6001)/mo_catalog")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	drop_account(db, "acc02")

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		var cnt int
		timeout := time.After(10 * time.Minute)
		for {
			select {
			case <-timeout:
				fmt.Println("checkpoint timeout", time.Now().Format("2006-01-02 15:04:05"))
				wg.Done()
			case <-time.After(2 * time.Second):
				do_checkpoint(db)
				cnt++
				fmt.Printf("checkpoint count: %d, now=%s\n", cnt, time.Now().Format("2006-01-02 15:04:05"))
			}
		}
	}()

	create_drop := func(account string) {
		var cnt int
		timeout := time.After(10 * time.Minute)
		ticker := time.NewTicker(3 * time.Second)
		for {
			select {
[...]

PR #26350, opened 2026-07-29
fix(catalog): avoid cloning schema when dropping relation

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #21795

Dropping a relation currently uses the generic update clone path. For table entries, that path serializes and deserializes the complete Schema and TombstoneSchema. Since dropped MVCC history remains in the catalog until GC, frequent account or table create/drop workloads retain unnecessary schema copies and increase TN memory usage.

This PR adds a delete-specific MVCC clone path that shares immutable table schemas with the previous MVCC version. It preserves copy-on-write when an internal same-transaction ALTER follows DROP, and makes Schema extra serialization read-only so shared schemas are safe under concurrent serialization. WAL, checkpoint, and replay formats remain unchanged.

In the local 45-second reproduction workload, DROP-path in-use memory decreased from 378.15 MB to 9 MB (97.6%), total heap growth decreased by 18.6%, normalized heap growth per cycle decreased by 22.5%, and RSS decreased by 11.1% without reducing throughput.

Validation includes catalog package tests, catalog race tests, repeated focused race tests, txnimpl tests, TAE restart/replay tests, build, vet, and an end-to-end account create/drop heap-profile comparison.


XuPeng-SH (approved):
Deep-reviewed exact head `3ebdfc4e59b00e35c30d99f727bb6850925f63bc`; no blocking findings. DROP versions receive a distinct MVCC payload while sharing only immutable table/tombstone schemas; rollback and commit generations therefore remain isolated. Same-transaction ALTER restores copy-on-write before mutation, and schema-extra serialization now uses a local value copy instead of mutating shared state. Non-table BaseNode implementations retain their prior cloning semantics, and persistence/replay formats are unchanged. Focused ownership/serialization regressions passed under `-race -count=20`; full catalog and txnimpl packages passed under race.

XuPeng-SH:
@aptend invlid commit message

PR #26354, opened 2026-07-29
test(frontend): stabilize CDC restart timeout race

### LINKED ISSUE #26353: [Bug]: TestRestartTimeoutWinsInFlightCatalogPublication flaky — CDC restart timeout race
## Problem

`TestRestartTimeoutWinsInFlightCatalogPublication` in `pkg/frontend/cdc_start_cleanup_test.go:813` fails intermittently:

```
cdc_start_cleanup_test.go:813: replacement did not enter restarting-to-running publication
ERROR: late-startup-error-after-timeout (moerr-code: 20101, occurrence: 3)
```

The test verifies that when a CDC executor restart timeout wins over an in-flight catalog publication, the replacement enters the restarting-to-running state. Under CI load, the replacement fails to transition within the expected timeframe (occurrence count reaches 3 before state transition).

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-29 | [PR #26217](https://github.com/matrixorigin/matrixone/actions/runs/30420593417/job/90476439890) | UT Test on Ubuntu/x86 |

## Suggested Fix

1. Increase the restart timeout tolerance for CI environments
2. Or add retry logic for the state transition assertion
3. Or use explicit synchronization barriers instead of timing-based waits

XuPeng-SH:
## What
- add a test-only timeout signal for the replacement-startup wait
- wait until restarting-to-running catalog publication is in flight before triggering timeout
- release the injected timeout from test cleanup on early assertion failures

## Why
TestRestartTimeoutWinsInFlightCatalogPublication used a 25ms real timer both as the timeout oracle and as the scheduling budget for the replacement goroutine. Under CI load, timeout could win before catalog publication started, so the test failed at its phase barrier instead of exercising the intended race.

Production behavior is unchanged: when no test signal is configured, Restart continues to use the existing real timer and four-second default.

Closes #26353

## Testing
- focused regression, count=100
- focused race regression, count=100
- full pkg/frontend test suite
- full pkg/frontend race test suite
- go build and go vet for pkg/frontend with GOWORK=off and module readonly mode
- pkg/bootstrap dependent package tests


LeftHandCold (approved):
Deep review completed on exact head 89e5c282a0ae83bcac13b6ca9032a00d411c93c5. No merge-blocking correctness, production-regression, concurrency, lifecycle, hang, or unbounded-growth issue found. The injected timeout signal is test-only and nil in production, so the existing real timer and four-second startup bound remain unchanged. The test now deterministically enters the restarting-to-running catalog publication before triggering timeout; completion-vs-timeout still linearizes through the existing restartOwner CAS, and cleanup releases both the injected timeout wait and the blocked catalog executor. Focused 100-round and race 100-round regressions, full frontend/bootstrap tests, full frontend race, build/vet, diff-check, and current CI pass.

PR #26357, opened 2026-07-29
fix(plan): preserve ROLLUP source order keys

### LINKED ISSUE #26269: [Bug] ROLLUP ORDER BY GROUPING fails with wrapped selected grouping keys
## Summary

`WITH ROLLUP` fails when grouping columns are wrapped only for display in the select list while `ORDER BY` directly uses `GROUPING()` and the original grouping columns.

## Version

MatrixOne `main` at `4f82ab5476a798bb1c61903abf77a29a2f6c4fc8`.

## Reproduction

```sql
CREATE TABLE rollup_audit (
  region varchar(8),
  product varchar(8),
  amount int
);

INSERT INTO rollup_audit VALUES
  ('east', 'a', 10),
  ('east', 'b', 20),
  ('west', 'a', 30),
  (NULL, 'a', 5);

SELECT
  COALESCE(region, '<null>'),
  COALESCE(product, '<null>'),
  SUM(amount),
  GROUPING(region),
  GROUPING(product)
FROM rollup_audit
GROUP BY region, product WITH ROLLUP
ORDER BY GROUPING(region), region, GROUPING(product), product;
```

## Actual behavior

MatrixOne returns:

```text
ERROR 20301 (HY000): invalid input: column region does not exist
```

The result reproduces in three independent executions.

## Expected behavior

The query should return the base groups, subtotals, and grand total. Local MySQL accepts the same SQL and returns eight rows. The `GROUPING()` values distinguish the source `NULL` group from the rollup rows.

## Scope

The base ROLLUP query succeeds. The query also succeeds when either the display wrappers or the `ORDER BY` clause is removed. The failure requires their combination.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26269

ROLLUP, CUBE, and GROUPING SETS are expanded into an internal UNION ALL. ORDER BY expressions containing GROUPING() were already materialized as hidden branch projections, but ordinary source grouping columns were left to bind above the union. When the visible select list wrapped those columns for display, the original source columns were no longer available and planning failed with `column ... does not exist`.

This change:

- materializes source-dependent non-DISTINCT grouping-set order keys while branch bindings are still available;
- reuses visible projections when an ORDER BY expression matches a selected expression, including after star expansion, so expressions are not evaluated twice;
- preserves ORDER BY alias precedence, DISTINCT restrictions, positional ordering, and the final output schema;
- adds planner coverage for ROLLUP, CUBE, GROUPING SETS, aliases, visible-expression reuse, hidden source keys, and DISTINCT, plus a distributed ROLLUP regression case with source NULL values.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=20m ./pkg/sql/plan`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=20m ./pkg/sql/plan`
- `go build -mod=readonly ./pkg/sql/plan/...`
- `go vet -mod=readonly ./pkg/sql/plan/...`
- `make static-check`
- rollup BVT: 162 passed, 0 failed, 2 ignored
- `git diff --check`


XuPeng-SH (approved):
Deep-reviewed exact head 857ae9842c56dbb9586447119a7f228e58ec2c4a. The grouping-set ORDER BY fix preserves source-bound hidden keys, visible projection reuse, alias precedence, DISTINCT restrictions, star-expanded result width, and final output trimming. I found no blocking correctness or lifecycle issue. The full pkg/sql/plan test suite passes locally; required CI is green, and the head merges cleanly with the latest main.

heni02 (approved):

PR #26358, opened 2026-07-29
fix(udf): support DROP FUNCTION IF EXISTS

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25416 

fix(udf): support DROP FUNCTION IF EXISTS


heni02 (approved):

XuPeng-SH (changes requested):
Deep-reviewed exact head 43cecf8387e7d4fd7aa95e27fb7c0e41205f0a98.

Blocking correctness issue: the new `IfExists` handling only runs when the lookup returns zero rows for the function name. If another overload with the same name exists but the requested signature does not, `execResultArrayHasData` is true, the loop finds no match, and the function returns the nil `err` and commits regardless of `df.IfExists`.

Reproducer at the `doDropFunction` level:
1. Catalog contains `db.testfunc(varchar)`.
2. Execute `DROP FUNCTION db.testfunc(int)` (without `IF EXISTS`).
3. Expected: `ErrDropNonExistsFunction`; actual: nil, transaction commits, and no catalog row is deleted.

This is a production-visible false success: deployment/cleanup automation can believe it removed the requested overload when nothing happened. It also means the new flag is not the semantic discriminator for all exact-target misses, while the added test only covers the simpler zero-same-name-row case. Please track whether an exact signature matched, and after exhausting overloads return nil only for `IF EXISTS`; otherwise return `NewNoUDFNoCtx`. Add tests for both with/without `IF EXISTS` when a different same-name overload exists.

XuPeng-SH (approved):
Re-reviewed exact head 8e17becceb6cba1b1ff97f4a4e8f6dcea0656412. The previous blocker is fixed: deletion success is now gated by an exact signature match, while a same-name/different-overload miss commits only with IF EXISTS and otherwise returns ErrDropNonExistsFunction and rolls back. The new tests assert both transaction outcomes and that no wrong overload is deleted. Parser/AST pooling and formatting are sound, and I independently regenerated mysql_sql.go with the repository goyacc and verified the generated body matches with no parser conflicts. Focused frontend/parser tests pass under race x50, full affected packages and go vet pass, git diff --check is clean, the head merges cleanly with current main, and the latest CI checks are green. No blocking issue remains.

PR #26360, opened 2026-07-29
fix(plan): preserve fulltext rewrites in nested queries

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24822 

fix(plan): preserve fulltext rewrites in nested queries


XuPeng-SH (approved):
Deep review completed on exact head 615d04b6e848477d6223ff97945b4ef85133654a. The change preserves the PROJECT boundary only until the fulltext placeholder/index rewrite runs, after which the existing second projection-removal pass can optimize it away. Filter-only, projected-score, nested CTE/derived-table, outer-join, window, and INSERT SELECT shapes are covered, and reachable-plan assertions verify both that a fulltext index scan exists and no unreplaced fulltext_match remains. I found no correctness, unhappy-path, or production performance blocker. CI is green and the PR is mergeable.

PR #26361, opened 2026-07-29
fix: refresh view metadata after alter column

### LINKED ISSUE #26227: [Bug]: View metadata stays stale after source column type changes
## Summary

After an `ALTER TABLE ... MODIFY COLUMN` changes the type of a source column, a dependent View executes with the new type but retains its original type metadata in `DESC` and `information_schema.columns`.

This leaves metadata consumers with a schema that no longer describes the View result.

## Environment

- MatrixOne commit: `7cb78aa5f88e3b98163f6b3bffb5f71586c0cbbf`
- Server version: `8.0.30-MatrixOne-v1.3.0`
- Reference engine: local MySQL `8.0.45`

## Reproduction

```sql
CREATE DATABASE view_alter_column_matrix;
USE view_alter_column_matrix;

CREATE TABLE source_t (
  id INT PRIMARY KEY,
  code VARCHAR(5) NOT NULL UNIQUE,
  qty INT NOT NULL DEFAULT 1,
  price DECIMAL(10,2) NOT NULL
);
INSERT INTO source_t VALUES (1, 'short', 2, 1.25);

CREATE VIEW v_source_t AS
  SELECT id, code, qty, price, qty * price AS total FROM source_t;

ALTER TABLE source_t MODIFY COLUMN code VARCHAR(60) NOT NULL;
ALTER TABLE source_t MODIFY COLUMN qty BIGINT NOT NULL DEFAULT 1;
ALTER TABLE source_t MODIFY COLUMN price DECIMAL(20,5) NOT NULL;

INSERT INTO source_t VALUES
  (2, 'this-code-is-longer-than-five-characters', 5000000000, 12345.67891);

SELECT id, code, qty, price, total FROM v_source_t ORDER BY id;
DESC v_source_t;
CREATE TABLE ctas_view AS SELECT id, code, qty, price FROM v_source_t;
DESC ctas_view;
```

## Actual result

The View query correctly returns the long string, `BIGINT` value, and five-decimal value. `ctas_view` also has the updated source definitions:

```text
code  VARCHAR(60)
qty   BIGINT(64)
price DECIMAL(20,5)
```

However, `DESC v_source_t` and `information_schema.columns` retain the definitions captured at View creation:

```text
code  VARCHAR(5)
qty   INT(32)
price DECIMAL(10,2)
total DECIMAL(38,2)
```

The mismatch reproduced in three fresh database runs. The same SQL on MySQL refreshes View metadata to `VARCHAR(60)`, `BIGINT`, `DECIMAL(20,5)`, and the corresponding new expression type.

## Expected result

View metadata should match the columns produced by the current View definition after compatible source-column type changes.

## Revised implementation scope

This issue owns the lifecycle that refreshes persisted View column metadata after dependency DDL. It does not own output-column provenance rules.

View creation persists generated `TableDef` columns, while query-time `bindView` rebinds the stored SQL against current dependencies. `DESC` and `information_schema.columns` read the persisted catalog definition, so query execution or CTAS using the new source types is not sufficient evidence that the View metadata was refreshed.

After #26226 and #26232 define the authoritative View output metadata contract, dependency refresh must rebind the current View definition and regenerate its complete column definitions through that same path. It must not patch selected type fields or maintain a second metadata derivation implementation.

The refreshed catalog update must preserve View identity and persisted definition attributes and must not use DROP/CREATE semantics. A refresh failure must not leave a partially rewritten definition or silently present stale metadata as current; invalid, temporarily unbindable, retryable, and terminal failures require explicit handling.

## Acceptance criteria

- Compatible source-column changes refresh persisted type, width, scale, nullability, default metadata, and derived-expression result types as produced by the authoritative View schema generator.
- `DESC`, `information_schema.columns`, direct View queries, and CTAS observe a mutually consistent current schema.
- Direct Views, aliases, expressions, and transitive View-of-View dependencies are covered.
- Relevant ALTER execution paths, including metadata-only/INPLACE and COPY-style replacements, cannot bypass refresh.
- Catalog replacement preserves View identity, SQL text, SQL mode, default database, security attributes, ownership, and privileges.
- Tenant, subscription, snapshot, and missing or temporarily unavailable dependency contexts are resolved under the correct account and database semantics.
- Any deferred recovery is bounded, fair across more than one page of pending Views, concurrency-safe, and cannot let an older scan generation move shared progress backward.
- Public SQL-path tests verify both successful refresh and controlled failure/recovery behavior; focused tests cover catalog replacement, pagination, concurrency, and dependency-context boundaries.

## Dependencies

Implementation order: #26226 -> #26232 -> #26227. Prepared-plan invalidation after View replacement or deletion is a separate lifecycle and is not part of this issue.
## Duplicate check

Checked #7358 (closed expression-length issue) and #24436 (open direct CTAS schema tracker). Neither covers stale View metadata after source `ALTER TABLE`.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #26227

`ALTER TABLE ... MODIFY COLUMN` updates the source table schema, but persisted view column metadata remained unchanged. View queries were rebound against the current source schema, while `DESC`, `information_schema.columns`, and other catalog consumers continued to report the stale types stored in `mo_columns`.

This change refreshes ordinary tenant view metadata after ALTER COPY column changes. It rebinds saved view definitions in stable database/name order, updates only views whose visible column metadata changed, and replaces the catalog definition in place so the view table ID and stored SQL are preserved. The refresh covers direct, nested, and cross-database views. Invalid historical views remain non-blocking for SQL definition or binding errors, while transaction, cancellation, internal execution, and catalog write failures are propagated.

The regression test covers VARCHAR width, INT-to-BIGINT, DECIMAL precision/scale, derived expression propagation, nested and cross-database views, an unrelated invalid view, `DESC`, `information_schema.columns`, CTAS, and view ID preservation.

Validation:

- `make build`
- `make static-check`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/sql/compile`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=600s ./pkg/sql/compile`
- Changed-test race stress with `-race -count=100`
- mo-tester BVT: 28/28 passed for `view_alter_column_metadata.sql`


heni02 (approved):

XuPeng-SH (changes requested):
Deep review of exact head `89a17a9460c00acda9f62fe07af9e9aad56a27df`. The identity/generation checks, snapshot/subscription handling, bounded retry cursor, and ALTER/Catalog tests are generally sound, but two reachable recovery paths still block merge.

### P1 — Do not report CREATE FUNCTION failure after the function transaction has committed

`InitFunction` owns its background transaction and commits it in the deferred `finishTxn` before returning (`authenticate.go:11020-11023`). `handleCreateFunction` then opens a separate background executor and directly returns the pending-view retry error (`mysql_cmd_executor.go:2403-2410`). `retryPendingViewMetadata` intentionally propagates catalog, transaction, cancellation, and internal execution errors, so any such failure now makes the client receive a failed CREATE even though the UDF row is already committed. Retrying the statement can then fail with “function already exists”.

Please either include recovery in the same transaction as function creation, or make a post-commit recovery failure best-effort/observable without changing the already-committed statement result. Add a regression where UDF creation succeeds and the subsequent catalog refresh fails.

### P1 — Recreating a missing view never retries dependents that were marked pending

The recovery hooks cover CREATE TABLE, CREATE FUNCTION, CREATE SNAPSHOT, and ALTER PUBLICATION, but `Scope.CreateView` returns immediately after creating the view (`ddl.go:2532-2623`) and no generic retry hook exists. A concrete path is: create `v1` on `t`, create `v2` on `v1`, drop `v1`, alter `t` so rebinding `v2` fails and marks it pending, then recreate `v1`. The dependency is valid again, but `v2` keeps its stale visible-column metadata and pending flag until an unrelated retry-triggering DDL happens.

Please trigger a bounded pending refresh for CREATE/REPLACE VIEW, preferably keyed by the recreated view identity/name as CREATE TABLE already does, and cover the nested-view recovery path in BVT.

Validation: focused view-metadata tests passed under race (`-count=3`) across plan/compile/frontend; full `pkg/sql/compile`, build, vet, diff checks, and current remote CI are green. These are lifecycle/coverage gaps rather than test failures.

gouhongshen (changes requested):
## Codex automated review

Two additional blocking defects remain. The existing unaddressed review already covers CREATE FUNCTION post-commit errors and missing CREATE VIEW recovery, so those are not duplicated here.

### P1 - Do not reject ALTER TABLE after 1,024 dependent views (pkg/sql/compile/alter.go:2472)
Scope.AlterView increments dependentViews and calls this guard before checking whether the view's visible columns or dependency metadata changed. The 1,025th confirmed view therefore returns ErrInvalidInput, which bubbles out and rolls back the source ALTER. A valid COPY alter of an unprojected column—or an alter of a published table with sufficient subscriber fan-out—now fails solely because of catalog size, whereas it succeeded before this PR and no public view-count limit exists. The refresh needs scalable processing without converting this internal bound into a semantic DDL rejection.

### P1 - Retry pending views when CREATE PUBLICATION restores subscriptions (pkg/frontend/publication_subscription.go:635)
The retry is added only to ALTER PUBLICATION, but createPublication also restores existing mo_subs rows from Deleted to Normal and then returns without retrying. Reproduce by dropping a publication, altering its source column so a subscriber view is marked metadata_refresh_pending while rebinding cannot resolve the subscription, and recreating the same publication for that subscriber. The subscription becomes usable again, but the view retains stale metadata and remains pending until an unrelated retry-triggering DDL occurs. CREATE PUBLICATION must perform the bounded recovery in its transaction as well.

gouhongshen (changes requested):
## Codex automated review

CREATE FUNCTION post-commit handling, CREATE VIEW recovery, and CREATE PUBLICATION recovery are fixed. The 1,024-view guard was removed, but the same catalog-size regression remains at 16,384 candidates. Subscription-database recreation also remains an uncovered recovery path. Diff checks passed; focused tests could not link because this workspace lacks the required CGo library and headers.

### P1 - Do not retain a catalog-size failure at 16,384 candidates (pkg/sql/compile/alter.go:2103)
Author response: f2ff144 removes the 1,024-dependent-view guard and adds a 1,025-call unit test. Why this remains blocking: refreshViewMetadataAfterAlter still increments examinedCandidates before dependency confirmation and returns ErrInvalidInput on candidate 16,385. The paged query includes both exact dependencies and legacy substring candidates, so 16,385 dependent views—or false legacy candidates—still roll back an otherwise valid COPY ALTER solely because of catalog size. No enforced or documented view-count limit excludes this case. Close this by traversing the supported catalog size without converting an internal processing bound into a statement error.

### P1 - Retry pending views when a subscription database is recreated (pkg/sql/compile/ddl.go:110)
This is distinct from the earlier CREATE PUBLICATION finding: the publication can remain present and Normal throughout. Dropping a subscriber database sets its mo_subs.sub_name to NULL; if the publisher then alters a referenced table, the surviving cross-database view is found through its persisted publisher dependency but cannot rebind, so it is marked metadata_refresh_pending. CREATE DATABASE ... FROM ... PUBLICATION restores sub_name in createSubscription, then CreateDatabase immediately returns after creating the database without invoking either pending-view retry path. The view becomes queryable again but DESC/information_schema retain stale metadata until unrelated DDL triggers recovery. Run bounded recovery as part of subscription-database creation and cover this sequence end-to-end.

gouhongshen (changes requested):
## Codex automated review

The follow-up is one non-merge commit. Prior blockers for CREATE FUNCTION, CREATE/REPLACE VIEW, CREATE PUBLICATION, both catalog-size limits, and ordinary subscription-database recreation are fixed; none remains downgraded or unresolved. Two distinct recovery gaps remain. Diff checks pass, but focused tests could not start because cgo/libmo.dylib is absent.

### P1 - Handle account-level publications before opening a publisher database (pkg/sql/compile/alter.go:2402)
This is distinct from the earlier ordinary subscription-database recovery finding. CREATE PUBLICATION explicitly accepts DATABASE * and persists `pub_database='*'`; subscription validation does not reject it. During CREATE DATABASE ... FROM ... PUBLICATION, this new helper passes that wildcard to `c.e.Database`, whose disttae implementation performs an exact database-name lookup. Since `*` is not a database, every account-level subscription creation now fails and rolls back after the new recovery step. Enumerate the account-level publication’s actual database/table coverage, or enforce a clear rejection during pre-mutation validation if such subscriptions are unsupported, and add regression coverage.

### P1 - Retry pending dependents after ALTER VIEW repairs their source (pkg/sql/compile/ddl.go:526)
This is distinct from the earlier CREATE/REPLACE VIEW finding: normal ALTER VIEW uses this separate path and returns immediately after replacing the definition. For example, create an intermediate view and a downstream view over tables `live` and `missing`; drop `missing`, then alter `live` so both views are marked pending. `ALTER VIEW intermediate AS SELECT ... FROM live` makes the downstream view queryable again, but no targeted pending refresh runs, so its persisted columns and pending flag remain stale until unrelated DDL. RENAME TABLE has the same gap because non-COPY alters bypass refresh. Route dependency-restoring ALTER VIEW and rename-target transitions through the shared bounded recovery mechanism and cover these public paths.

gouhongshen (changes requested):
## Codex automated review

Account-level publication, ALTER VIEW, CREATE FUNCTION/VIEW/PUBLICATION, subscription recreation, and catalog-size blockers are fixed. Single-target RENAME works, but supported chained RENAME statements still violate pending-view recovery. `git diff --check` passed; focused tests could not start because `cgo/libmo.dylib` is absent.

### P1 - Defer recovery until a chained RENAME reaches its final mapping (pkg/sql/compile/alter.go:1905)
Author response: 20e181d invokes pending-view recovery after each rename subplan, and the new BVT covers a single rename target. Why this remains blocking: `Scope.RenameTable` executes chained renames sequentially. For a pending view over `missing`, `RENAME TABLE replacement TO missing, missing TO moved` refreshes the view after the first leg, persists the transient replacement type, and clears its pending flag. The second leg removes `missing`, but recovery targets `moved`; a later `CREATE TABLE missing` also scans only pending views, so query results use the recreated schema while `DESC`/information_schema retain the transient type. Chained rename is an established supported path (`rename_table_chained.sql`). Defer recovery until the entire statement's final name mapping is installed, and add this intermediate-target counterexample.

gouhongshen (changes requested):
## Codex automated review

The chained-RENAME blocker is fixed by 90324df, which defers recovery until the final mapping and covers the reported counterexample. Prior blockers involving CREATE FUNCTION/VIEW/PUBLICATION, ALTER VIEW, account-level publications, and catalog-size limits are fixed. Subscription-database recreation remains blocking only for pre-upgrade views lacking persisted dependencies. Focused tests could not start because local CGo artifacts are absent; both diff checks passed.

### P1 - Recover pre-upgrade subscriber views after subscription recreation (pkg/sql/compile/alter.go:2763)
This is the unresolved upgrade-compatibility portion of my earlier subscription-recreation finding. Author response: 66c920 added `refreshPendingViewMetadataAfterSubscriptionCreate` and BVT coverage for a newly created view carrying publisher dependencies. Why this remains blocking: historical `ViewData` has no `Dependencies`. After its subscription database is dropped, a publisher ALTER can select that legacy subscriber view through the still-Normal `mo_subs` row, fail rebinding, and mark it pending without adding dependencies. On subscription recreation, the new helper calls this `onlyPending` query using the publisher account, but the dependency-less branch requires `account_id = sourceAccountID`; the view belongs to the subscriber, while every other branch requires dependency JSON it does not have. Its persisted columns therefore remain stale despite the dependency becoming usable again. Close this by including legacy pending views scoped to the restored subscriber/subscription identity, or by enforcing a migration that guarantees dependencies, and add an upgrade-shaped regression with dependency-less `ViewData`.

gouhongshen (changes requested):
## Codex automated review

7c5095be fixes the previously reported case when subscription recreation itself restores an existing source. The legacy upgrade-compatibility topic remains blocking for a later source recreation; all earlier CREATE FUNCTION/VIEW/PUBLICATION, ALTER VIEW/RENAME, account-publication, and catalog-size blockers remain fixed. Diff checks and package loading passed; focused tests could not start because cgo/libmo.dylib is absent.

### P1 - Recover legacy subscriber views when the published table is recreated later (pkg/sql/compile/ddl.go:2208)
Author response: 7c5095be passes a subscriber-scoped legacy candidate for every source enumerated during subscription-database creation. This fixes the reported sequence when the published table exists at that moment.

Why this remains blocking: With a publication exposing all tables, a dependency-less pre-upgrade subscriber view can be marked pending while its subscription is absent. If the publisher then drops that source, recreating the subscription succeeds but cannot enumerate the absent table, so no legacy candidate is generated. When the publisher later recreates the table, this CREATE TABLE hook calls refreshViewMetadataAfterAlter with the publisher account and onlyPending=true, but without a legacy subscriber candidate. The dependency-less query branch consequently requires the view's account_id to equal the publisher's; the subscriber-owned view cannot match. The wildcard publication makes the new table immediately usable, so the view query rebinds successfully while DESC/information_schema and metadata_refresh_pending remain stale. This is distinct from the fixed ordering because CREATE TABLE—not subscription creation—is the event that restores the dependency. Close this by making published source recreation discover dependency-less pending views in live subscriber accounts, or by enforcing a migration contract that guarantees persisted dependencies, and cover this event order.

gouhongshen (changes requested):
## Codex automated review

The latest merge has no authored conflict-resolution delta. 9f88b3a fixes ordinary later-source recovery, and prior CREATE FUNCTION/VIEW/PUBLICATION, ALTER VIEW/RENAME, account-publication, and catalog-size blockers remain fixed. Legacy subscription recovery is still blocking for supported quoted table names containing spaces.

### P1 - Preserve spaces when matching explicitly published legacy sources (pkg/sql/compile/alter.go:2778)
Author response: 9f88b3a replaces the subscription-recreation-only candidate with live `mo_subs` discovery so later source recreation can also recover legacy pending views.

Why this remains blocking: the new membership predicate removes every space from `pub_tables` before comparing it with the unchanged source name. Publication creation stores table object names verbatim and merely joins them with commas (`genPubTablesStr`), while quoted identifiers containing spaces are not rejected. For an explicitly published table `odd name`, `pub_tables` is `odd name`, so this evaluates like `find_in_set('odd name', 'oddname')` and cannot select the dependency-less subscriber view. A reachable sequence is: create a pre-upgrade view over `subdb`.`odd name`, drop the subscription database, ALTER the published table so the view becomes pending, then recreate the subscription database. Queries can bind again, but the view remains pending and `DESC`/information_schema retain stale metadata. The other query branches exclude this subscriber-owned dependency-less view. Close this with publication membership matching that preserves identifier contents, or an enforced contract excluding such identifiers, and cover the quoted-name recovery sequence.

gouhongshen (changes requested):
## Codex automated review

3d26418 fixes the exact internal-space predicate, and earlier CREATE FUNCTION/VIEW/PUBLICATION, subscription, ALTER VIEW/RENAME, account-publication, and hard catalog-limit blockers remain fixed. Edge-whitespace compatibility and fan-out scalability remain blocking. Diff checks and parser tests passed; compile tests could not start because libmo.dylib is absent.

### P1 - Preserve edge whitespace while enumerating published tables (pkg/sql/compile/alter.go:2534)
Author response: 3d26418 correctly stops removing spaces in FIND_IN_SET and covers `odd name`. Why this remains blocking: subscription recovery independently applies strings.TrimSpace to every persisted pub_tables token. genPubTablesStr stores quoted object names verbatim with comma separators and no padding, while the parser explicitly supports names such as `table11 `. CREATE DATABASE ... FROM ... PUBLICATION therefore looks up `table11`, either rolling back with no-such-table or targeting a distinct table and leaving pending metadata stale. Preserve tokens exactly, or enforce rejection when creating the publication, and add an end-to-end edge-whitespace case.

### P1 - Avoid one additional catalog query per refreshed leaf view (pkg/sql/compile/alter.go:2310)
Every confirmed view is enqueued as another source. With N flat leaf views directly depending on the altered table, the initial scan already refreshes all N, but the queue then executes N additional loadViewMetadataRefreshPage queries even though none has dependents. processedViews is checked only after each query returns. At the explicitly supported 16,385-view fan-out this means more than 16,385 sequential internal catalog queries; nested graphs additionally return quadratic candidate rows. This is distinct from the fixed hard candidate limit: the statement no longer rejects the catalog size, but can become unavailable from unbounded redundant traversal. Bound traversal by actual downstream dependency work and add a query-count regression for flat fan-out.

gouhongshen (changes requested):
## Codex automated review

84bb841 fixes trailing-whitespace publication enumeration and avoids redundant traversal for views already carrying structured dependencies. The prior fan-out blocker remains for dependency-less pre-upgrade views. Earlier CREATE FUNCTION/VIEW/PUBLICATION, ALTER VIEW/RENAME, subscription recovery, account-publication correctness, internal-space matching, and hard catalog-limit blockers remain fixed; none is merely downgraded. The intervening merge has no conflict-resolution delta. Two additional blocking compatibility/recovery gaps remain.

### P1 - Bound recovery must not strand the 1,025th repaired view (pkg/frontend/view_metadata_retry.go:43)
This function loads and processes exactly one page, while the query is limited to 1,024 rows. The cursor only chooses a page for a future invocation; there is no worker or guaranteed subsequent trigger. For example, make 1,025 views pending by dropping their shared snapshot and altering another referenced table, then recreate that snapshot. CREATE SNAPSHOT repairs at most 1,024 views and commits successfully, leaving at least one view pending with stale DESC/information_schema metadata indefinitely unless unrelated DDL occurs. Ensure the dependency-restoring event deterministically processes or durably schedules every affected view, and cover the 1,025-row boundary.

### P1 - Match legacy definitions using identifier-aware encoding (pkg/sql/compile/alter.go:2746)
Dependency-less pre-upgrade views are selected by searching viewdef for the raw sourceTable string. That string is not necessarily present in persisted ViewData.Stmt: MatrixOne supports embedded backticks, so table `src``table` has the raw name src`table but its stored CREATE VIEW SQL necessarily contains src``table; JSON-special identifier characters are escaped again by json.Marshal. The raw instr predicate therefore misses the view, and ALTER COLUMN leaves its catalog metadata stale. This is distinct from the fixed whitespace findings: the failure is SQL/JSON identifier encoding in the ordinary legacy-candidate path. Parse or migrate legacy definitions, or construct an encoding-aware match, and add a dependency-less quoted-identifier regression.

### P1 - Avoid per-leaf traversal for dependency-less legacy fan-out (pkg/sql/compile/alter.go:2306)
Author response: 84bb841 skips enqueueing a refreshed view when its pre-refresh ViewData already contains structured dependencies, and adds a query-count test for that case. Why this remains blocking: pre-upgrade views have no Dependencies, and this condition examines the stale in-memory definition loaded before the successful rebind. Consequently every dependency-less leaf is still appended to sources; each then executes another loadViewMetadataRefreshPage query before processedViews can be consulted. A supported 16,385-view legacy fan-out therefore still performs over 16,385 redundant sequential catalog queries and can make ALTER TABLE unavailable during upgrade. Close this by bounding or batching legacy downstream discovery without querying once per confirmed leaf, and add a dependency-less fan-out query-count regression.

gouhongshen (changes requested):
## Codex automated review

46fb8a fixes the prior 1,025th-row retry pagination defect and batches dependency-less legacy fan-out. Its raw/backtick candidates also fix the previously reported embedded-backtick case, but ANSI_QUOTES legacy discovery remains incomplete. Earlier blockers involving post-commit UDF handling, CREATE VIEW/PUBLICATION, subscriptions, account publications, pending ALTER VIEW recovery, chained renames, whitespace, and catalog limits remain fixed for their reported cases. A distinct ordinary ALTER VIEW dependency-closure defect remains: already-pending dependents recover, but healthy dependents are excluded. Both diff checks and package loading passed; focused tests could not start because cgo/libmo.dylib is absent.

### P1 - Support ANSI-quoted identifiers in legacy dependency discovery (pkg/sql/compile/alter.go:2983)
Author response: 46fb8a adds raw-name and JSON-encoded backtick-spelling candidates, closing the previously reported embedded-backtick case. Why this remains blocking: historical ViewData preserves the original root SQL and SQLMode, and ANSI_QUOTES is supported. For a dependency-less legacy view over a table named `src"table`, the stored statement spells the identifier as `"src""table"`, with those quotes escaped again in viewdef JSON. Neither candidate produced here—the raw name nor its backtick-quoted spelling—occurs in that representation, so ALTER TABLE never selects the view for refresh. Queries rebind against the current source type while DESC/information_schema retain stale columns. This is a distinct SQL-mode spelling case, and no enforced identifier restriction excludes it. Parse or migrate legacy definitions using their stored SQLMode, or generate complete mode-aware candidates, and add an ANSI_QUOTES embedded-quote upgrade regression.

### P1 - Maintain downstream dependency closure after ALTER VIEW (pkg/sql/compile/ddl.go:540)
Author response: the post-ALTER hook calls pending-only recovery, fixing the earlier case where a downstream view had already been marked pending while a missing source view was repaired. Why this remains blocking: `refreshPendingViewMetadataForRelation` passes `onlyPending=true`, so every healthy downstream view is excluded. For example, create INT tables t1/t2, v1 over t1, and v2 over v1; then redefine v1 over t2. v2 retains the transitive dependency closure {v1,t1}, not t2. A later ALTER TABLE t2 changing the column to BIGINT refreshes v1 only; because v1 already had structured dependencies, the traversal stops at alter.go:2344 and never discovers v2. Queries through v2 bind the BIGINT result, while its persisted columns remain INT. This is distinct from the resolved missing-source case because v2 is healthy and non-pending when the dependency changes. Propagate successful view redefinitions to all downstream views, or otherwise enforce maintenance of their dependency closure, and cover this reparenting sequence.

gouhongshen (changes requested):
## Codex automated review

Prior CREATE FUNCTION/VIEW/PUBLICATION, subscription/account-publication, rename, catalog-limit/fan-out, whitespace, and embedded-backtick blockers remain fixed. The ANSI_QUOTES legacy-discovery and healthy ALTER VIEW dependency-closure blockers remain unresolved, while ee6b3e1 reintroduces the previously fixed >1,024 pending-recovery gap. Nothing was withdrawn or accepted as non-blocking. The merge resolution is import-only; diff checks and package loading passed, but focused tests could not start because libmo.dylib is absent.

### P1 - Do not require another DDL to process the next pending page (pkg/frontend/view_metadata_retry.go:44)
Author response: 46fb8a4 previously closed the 1,025-row recovery finding by traversing every page. ee6b3e1 replaces that behavior with a process-local cursor, and its test now expects the second page only after a second invocation. Why this remains blocking: this function loads one LIMIT 1024 page and returns, while dependency-restoring paths such as CREATE SNAPSHOT call it only once and no worker or durable continuation exists. Recreating a snapshot needed by 1,025 pending views therefore refreshes only 1,024; the last view remains pending with stale DESC/information_schema metadata until an unrelated retry-triggering DDL happens—and progression is not even guaranteed across CNs because the cursor is process-local. Close this by completing or durably scheduling all affected pages from the restoring event, with a one-event 1,025-row regression.

### P1 - Discover ANSI-quoted legacy dependencies using their stored SQL mode (pkg/sql/compile/alter.go:2990)
Author response: 46fb8a4 added raw-name and JSON-encoded backtick candidates, fixing the previously reported embedded-backtick spelling. Why this remains blocking: origin/main already persisted ViewData.SQLMode but did not persist Dependencies, so upgrade views can retain ANSI_QUOTES SQL. For a table named `src"table`, such a view stores the identifier as `"src""table"` inside its JSON statement. Neither the raw `src"table` candidate nor the backtick-quoted candidate generated here occurs in that representation, so ALTER TABLE does not select the view for refresh. Queries rebind to the new type while DESC/information_schema remain stale. Close this by parsing or migrating dependency-less definitions using their stored SQLMode, or by providing complete mode-aware matching and regression coverage.

### P1 - Refresh healthy downstream views after ALTER VIEW reparenting (pkg/sql/compile/ddl.go:540)
Author response: the post-ALTER hook invokes pending-view recovery and covers dependents already marked pending by a missing source. Why this remains blocking: refreshPendingViewMetadataForRelation passes onlyPending=true, excluding healthy downstream views. If v2 depends on v1/t1 and `ALTER VIEW v1 AS SELECT ... FROM t2` preserves the current type, v2 remains healthy but retains the old transitive dependency closure. A later type-changing ALTER of t2 refreshes v1 only; v2 does not mention t2, and traversal stops at structured-dependency v1, leaving v2's persisted columns stale although queries bind the new type. Close this by propagating successful view redefinitions through all downstream views and updating their dependency closure, with a healthy reparenting regression.

gouhongshen (changes requested):
## Codex automated review

45112785 fixes the prior >1,024 pending-page gap, ANSI_QUOTES legacy discovery, and healthy ALTER VIEW dependency closure. Earlier blockers for post-commit CREATE FUNCTION, pending CREATE/REPLACE VIEW/PUBLICATION recovery, subscriptions, renames, fan-out limits, whitespace, and backticks remain fixed. The upstream merge has no conflict-resolution delta. Two distinct correctness gaps remain.

### P1 - Refresh healthy dependents after CREATE OR REPLACE VIEW (pkg/sql/compile/ddl.go:2650)
This hook always passes onlyPending=true, including after qry.GetReplace() drops and recreates an existing view. For example, create INT table t1, BIGINT table t2, v1 over t1, and healthy v2 over v1; then CREATE OR REPLACE VIEW v1 AS SELECT a FROM t2. v2 is excluded because it is not pending, so queries through v2 bind BIGINT while DESC/information_schema retain INT and its persisted dependency closure still references the old v1/t1. This is distinct from the earlier CREATE/REPLACE finding, which covered already-pending dependents after a missing view was recreated. Refresh all downstream views when replacing an existing view and cover the healthy replacement case.

### P1 - Use system-account routing when rebinding tenant views over cluster tables (pkg/sql/compile/sql_executor_context.go:550)
The normal frontend resolver switches non-system tenants to the system account for readable cluster/shared system tables, and this PR records those dependencies with AccountID 0. The internal resolver used by runViewMetadataRefreshSQL only switches for snapshots and subscriptions, then opens the relation under the tenant account. Thus a tenant can create a view over a sys-owned mo_catalog cluster table, but a sys ALTER COLUMN finds that view and fails to rebind it because the internal lookup cannot see the cluster relation; the error is treated as skippable, the ALTER commits, and the view remains pending with stale DESC/information_schema metadata. Share the normal cluster/system-table account-routing contract with the internal resolver and add a sys-source/tenant-view regression.

gouhongshen (changes requested):
## Codex automated review

The two blockers from 45112785 are fixed for their reported cases: existing CREATE OR REPLACE VIEW now refreshes healthy dependents, and structured live system/cluster dependencies now use recorded account routing. All earlier blockers remain fixed; none was withdrawn, accepted as non-blocking, or remains unchanged. Three distinct correctness gaps remain. The merge has no conflict-resolution delta; diff checks and package builds passed, while focused tests could not link because cgo/libmo.dylib is absent.

### P1 - Refresh healthy dependents when a dropped relation is recreated (pkg/sql/compile/ddl.go:2216)
Author response: the new createViewRefreshOnlyPending helper performs a full refresh when CREATE OR REPLACE replaces an existing view, fixing the prior blocker. Why this remains blocking: DROP TABLE/VIEW only deletes the relation and never marks its dependents pending. CREATE TABLE still hard-codes onlyPending=true here, and the helper also returns true when the recreated view did not exist. Thus `CREATE TABLE t(a INT); CREATE VIEW v AS SELECT a FROM t; DROP TABLE t; CREATE TABLE t(a BIGINT)` leaves v healthy and excluded from refresh: queries bind BIGINT, but DESC/information_schema remain INT. The same gap affects a dropped intermediate view recreated with a changed schema. Mark dependents pending on DROP or refresh healthy name-matched dependents on recreation, and cover recreation with an immediately changed type rather than relying on a subsequent ALTER.

### P1 - Apply system-account routing to snapshot cluster dependencies (pkg/sql/compile/sql_executor_context.go:553)
Author response: the latest commit routes non-snapshot dependencies through the account recorded in ViewData, fixing the prior live cluster-source case. Why this remains blocking: this condition explicitly bypasses that routing for every snapshot, and relationAccountID also rejects Snapshot dependencies. Normal frontend resolution applies cluster/shared-system routing after snapshot setup, so tenant snapshot reads of such relations are supported. A tenant view combining a live tenant table with a cluster relation at an account snapshot therefore creates successfully, but altering the live table makes the internal rebind open the snapshot relation under the tenant account. The resulting no-such-table error is skippable, so ALTER commits while the view becomes pending with stale columns. Moreover, makeViewDependency overwrites the cluster relation's system AccountID with the snapshot tenant ID. Preserve and apply the storage account independently of the snapshot's logical tenant, and add a mixed live-plus-snapshot cluster regression.

### P1 - Discover dependency-less tenant views over system cluster tables (pkg/sql/compile/alter.go:2930)
Author response: recorded AccountID routing fixes newly created views that carry structured Dependencies. Why this remains blocking: pre-upgrade ViewData has no Dependencies, and this discovery branch only admits views owned by the source account or subscribers of its publications. A tenant-owned legacy view directly reading a sys-owned cluster table satisfies neither condition, so a sys ALTER COLUMN never even attempts to rebind it. Queries subsequently bind the new type, while DESC/information_schema retain the old metadata. This is distinct from the fixed resolver defect because the failure occurs during candidate discovery. Include and confirm dependency-less tenant cluster-view candidates, or enforce a migration contract that backfills their dependencies, with an upgrade-shaped regression.

XuPeng-SH (changes requested):
Deep review completed against 2e6f72bc50bd. The latest changes close the reported CREATE OR REPLACE and live structured system-source cases, but three reachable recovery paths still violate the metadata invariant.

1. Healthy dependents are skipped when a dropped relation is recreated with a new schema. DropTable and DROP VIEW delete the relation without marking dependents pending. The CREATE TABLE hook then calls refreshViewMetadataAfterAlter with onlyPending=true, and a newly recreated view also uses the pending-only mode. For CREATE TABLE t(a INT); CREATE VIEW v AS SELECT a FROM t; DROP TABLE t; CREATE TABLE t(a BIGINT), v remains non-pending and is excluded: query binding sees BIGINT while DESC/information_schema keep INT. Please either mark dependents pending at DROP or refresh all name-matched dependents on recreation, with table and intermediate-view regressions.

2. Snapshot cluster dependencies lose and cannot use the physical system account. makeViewDependency first assigns AccountID=0 for a cluster/shared-system relation, then the snapshot block overwrites it with snapshot.Tenant.TenantID. compilerContext.getRelation also deliberately bypasses relationAccountID whenever the snapshot is valid, and relationAccountID rejects snapshot dependencies. A tenant view that combines a live tenant source with a sys-owned cluster relation at an account snapshot can therefore be selected by the live ALTER but fail internal rebinding under the tenant account, leaving pending stale metadata. Keep logical snapshot tenant and physical relation account as separate identities and test the mixed live-plus-snapshot path.

3. Dependency-less pre-upgrade tenant views over system cluster tables are never candidates. In the legacy branch, loadViewMetadataRefreshPage admits only views owned by the source account or subscriber accounts. A tenant-owned legacy view directly over a sys-owned cluster table is neither account 0 nor a publication subscriber, so a sys ALTER never attempts refresh. This is a discovery gap distinct from the resolver fix for new structured dependencies. Add an upgrade-shaped legacy candidate path or a guaranteed migration/backfill contract.

Validation on the exact head: diff check, package selection, build, vet, and the newly added focused compile tests all pass. Current CI is green; the blockers are uncovered public-path state transitions.

gouhongshen (changes requested):
## Codex automated review

d4693ed fixes the reported ordinary table/view recreation cases, recorded snapshot physical-account routing, and direct legacy tenant-cluster discovery. However, recreation still skips system databases, dependency-less snapshot/cluster views remain unresolved, and the widened system-account discovery corrupts ordinary same-named tenant metadata. Healthy subscription reparenting is also still unhandled. Earlier blockers involving post-commit UDF handling, view/publication recovery, renames, SQL modes, pagination, fan-out, and quoting remain fixed; none is downgraded or withdrawn. The merge has no conflict-resolution delta. Focused tests could not run because libmo.dylib and required third-party headers are absent; diff checks passed.

### P1 - Limit cross-tenant legacy discovery to genuinely shared relations (pkg/sql/compile/alter.go:2869)
Author response: d4693ed broadens dependency-less discovery to every tenant when sourceAccountID is 0, fixing legacy views over cluster/shared system tables. Why this remains blocking: account 0 also owns ordinary tenant-isolated tables. If sys and a tenant both have app.t, a dependency-less tenant view over its own app.t becomes a candidate when sys alters app.t. The new resolver fallback then deliberately opens app.t under account 0, so rebinding confirms the wrong source and overwrites the tenant view's persisted columns/dependencies with sys metadata. No namespace or permission contract forbids identical database/table names across accounts. Widen discovery and source routing only when the actual source is a cluster/shared relation, with an ordinary sys/tenant same-name control.

### P1 - Run recreation recovery for cluster tables in system databases (pkg/sql/compile/ddl.go:2198)
Author response: d4693ed changes CREATE TABLE/VIEW to refresh healthy dependents and adds ordinary recreation regressions. Why this remains blocking: the CREATE TABLE hook is still skipped for every IsFkBannedDatabase, including mo_catalog and other supported system databases. Sys can create a cluster table in mo_catalog and tenants can read/view it; after DROP and recreation with a different type, the dependent view was never marked pending and this guard prevents any refresh. Queries bind the recreated type while DESC/information_schema retain the old type. Decouple view recovery from the foreign-key database restriction and cover a recreated system-database cluster table.

### P1 - Refresh healthy views when a subscription is reparented (pkg/sql/compile/alter.go:2643)
Author response: prior changes recover views already marked pending when a subscription database is recreated. Why this remains blocking: this call still passes onlyPending=true. A supported sequence is to create subdb from publication p1, create healthy view v over subdb.t, drop subdb, then recreate subdb from p2 whose t has a different type. v was never marked pending, so it is excluded: normal queries bind p2's type while persisted metadata and dependency closure still describe p1. This is distinct from the fixed pending-recovery case. Reparenting must refresh all dependents of the subscription identity, with an immediate DESC assertion before any later source ALTER.

### P1 - Route dependency-less snapshot cluster relations during refresh (pkg/sql/compile/view_metadata_resolver.go:66)
Author response: d4693ed preserves physical account IDs for recorded snapshot dependencies and lets relationAccountID use them, fixing structured views. Why this remains blocking: pre-upgrade ViewData has no Dependencies. For a legacy view combining a live tenant table with a cluster table at an account snapshot, altering the live table selects the view, but this resolver can route only recorded dependencies or the altered live source. The secondary snapshot cluster relation is therefore opened under the snapshot tenant instead of account 0, unlike normal frontend resolution. Rebinding fails as skippable, leaving pending stale metadata although queries remain valid. This differs from direct cluster-source discovery because the cluster relation is an unrecorded secondary dependency. Share the normal cluster/shared routing contract or enforce dependency backfill, and add an upgrade-shaped mixed-source regression.

PR #26362, opened 2026-07-29
fix: preserve temporary index table classification

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25406 

fix: preserve temporary index table classification


XuPeng-SH (approved):
Deep-reviewed exact head e7917bcaadf252d701b91359c27e4650f6a4c139. The temporary-name classifier keeps normal index-prefix behavior, validates the generated index UUID for temporary physical names, and avoids database-name false positives by using the last marker. The affected unique/secondary index consumers now receive the intended classification without changing final user-table output. No blocking issue found. Full tests pass for pkg/catalog, pkg/sql/colexec/multi_update, and pkg/sql/util; required CI is green and the head merges cleanly with latest main.

PR #26363, opened 2026-07-29
fix(plan): validate derived table aliases

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25380 

fix(plan): validate derived table aliases


XuPeng-SH (approved):
Deep review completed at 49ebc9554844d5552f3fadfab1edd101650d3a8c. The alias check correctly unwraps parenthesized derived SELECT/VALUES forms, preserves ordinary parenthesized joins, returns MySQL-compatible 1248/42000 and 1353/HY000 errors, and updates the internal CTAS/FK/index-writer SQL generators that relied on alias-less derived tables. I also verified the relevant plan, ISCP, and moerr packages locally.

Non-blocking follow-up: MySQL also rejects duplicate final derived-column names (for example, `select 1 from (select 1, 2) as d(a, a)` and duplicate implicit headings) with 1060/42S21, while MatrixOne still accepts them when those columns are not referenced. That behavior predates this patch and is outside the issue's explicit missing-alias/count-mismatch cases, so it does not block this fix.

heni02 (approved):

Issue #26060, opened 2026-07-23
[Bug]: DATA BRANCH statements retain one worker pool after every execution


Ariznawlll:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

`2eb7f523e5a10f352d3e258183ed8440510388de`

### Description

Every `DATA BRANCH DIFF`, `MERGE`, or `PICK` statement creates a new `ants.Pool` in `getTableStuff`, but the pool is never released when the statement finishes.

`ants.NewPool` starts one purge goroutine and one ticktock goroutine for the lifetime of the pool. Idle business workers expire, but these two maintenance goroutines remain until `Pool.Release()` is called. Since the Data Branch statement path does not call `Release`, repeated statements retain one pool and two goroutines per execution.

Relevant code:

```go
// pkg/frontend/data_branch.go
defer func() {
    if err == nil {
        tblStuff.worker, err = ants.NewPool(runtime.NumCPU())
    }
}()
```

The returned `tblStuff.worker` is used by the hash-diff and output paths, but production code contains no `tblStuff.worker.Release()` or equivalent cleanup. Existing calls to `worker.Release()` are only in tests that allocate their own pools.

### Steps to Reproduce

Start a standalone MatrixOne with the debug HTTP endpoint enabled, prepare a simple Data Branch table pair, and record the `ants` pool maintenance goroutine counts before and after each statement:

```sql
CREATE DATABASE g4d_pool_lifecycle;
USE g4d_pool_lifecycle;
CREATE TABLE base(id INT PRIMARY KEY, v VARCHAR(20));
INSERT INTO base VALUES (1, 'a');
DATA BRANCH CREATE TABLE child FROM base;
INSERT INTO child VALUES (2, 'b');

DATA BRANCH DIFF child AGAINST base;
DATA BRANCH DIFF child AGAINST base;
DATA BRANCH DIFF child AGAINST base;
```

Inspect `/debug/pprof/goroutine?debug=1` after each statement and count these stacks:

```text
github.com/panjf2000/ants/v2.(*poolCommon).purgeStaleWorkers
github.com/panjf2000/ants/v2.(*poolCommon).ticktock
```

Observed in three consecutive runs:

```text
before:                         19 purge / 19 ticktock
after DATA BRANCH DIFF #1:      20 purge
after DATA BRANCH DIFF #2:      21 purge
after DATA BRANCH DIFF #3:      22 purge
[...]

PR #26369, opened 2026-07-29
fix data branch worker pool leak

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26060

`getTableStuff` created a per-statement ants worker pool, but the production Data Branch DIFF, MERGE, and PICK path did not originally release it. Each statement therefore retained the pool's purge and ticktock maintenance goroutines.

The first fix registered pool release after the normal producer/consumer shutdown path. Review then identified two earlier exits: invalid DIFF projections and PICK statements on tables without a primary key returned before release was registered.

This PR now:

- validates DIFF projections and PICK primary-key requirements before allocating the pool;
- creates the pool in a single preparation function;
- registers `Release` immediately after successful preparation;
- preserves the existing shutdown order so producers and consumers finish before release;
- covers both rejected-statement paths and the valid allocation/release path.

Validation:

- `go test ./pkg/frontend -run '^(TestDataBranch|TestHashDiff)' -count=1`
- `go test -race ./pkg/frontend -run '^(TestDataBranch|TestHashDiff)' -count=1`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s -run 'TestDataBranch|TestHashDiff|TestPrepareDataBranchWorker' ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=300s -run 'TestPrepareDataBranchWorkerValidatesBeforeAllocation|TestDataBranchOutputValidateProjectedColumns' ./pkg/frontend`


XuPeng-SH (changes requested):
Reviewed exact head `741e77803a6b86a9ba261a85876ab6830d2896cb`. The normal producer/consumer shutdown order is safe, but the release is still registered too late to close every pool created by `getTableStuff`.

### P1 — Early validation errors still leak the per-statement ants pool

`getTableStuff` allocates the pool in its success defer. In the DIFF branch, `validateProjectedColumns` can then fail at `pkg/frontend/data_branch.go:752-754` (for example, an unknown projected column or unsupported file projection) before `defer tblStuff.worker.Release()` is registered at line 784. The PICK branch has the same gap at lines 776-780 when the source has only the fake primary key. Each rejected statement therefore retains the exact purge/ticktock goroutines this PR is intended to eliminate.

Register ownership immediately after every successful `getTableStuff` return, before any subsequent validation, or move pool construction/ownership into a helper whose cleanup cannot be skipped. Add regressions for the invalid DIFF projection and no-PK PICK exits that assert the created pool is closed.

Validation: the focused `TestDataBranch|TestHashDiff` frontend suite passes with the deterministic CGo wrapper, but its projection tests call the validator directly and do not exercise the agency-level pool lifetime on these error paths.

gouhongshen:
Fixed in `22c687e4a`. Pool construction now occurs in `prepareDataBranchWorker` only after DIFF projection and PICK primary-key validation succeeds, and `diffMergeAgency` registers `Release` immediately after successful preparation. The two rejected-statement regressions assert that no worker is allocated; the valid path asserts the worker remains releasable. Focused normal and race tests pass.

XuPeng-SH (approved):
Re-reviewed exact head `22c687e4ac456a444658b637b670a4306151a740` after reading the author's response. The response matches the implementation: `prepareDataBranchWorker` now performs invalid DIFF projection and no-primary-key PICK validation before allocating the ants pool, and `diffMergeAgency` registers `Release` immediately after successful preparation. `getTableStuff` no longer owns allocation, and it has no other production call sites that expect a worker. This closes both early-return leaks from the previous review while preserving the producer/consumer shutdown order.

The added regressions verify no allocation on both rejected statements and a releasable worker on the valid path. Complete frontend tests pass locally, as do the DataBranch/HashDiff focused suite and the ownership/projection tests under `-race -count=100`; exact-head CI is green and the PR is mergeable.

gouhongshen:
@Mergifyio refresh

Issue #26365, opened 2026-07-29
[Bug]: TestISCPExecutor8 intermittently times out waiting for the initial watermark


ck89119:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

## Branch Name

main

## Commit ID

47638ee49cab44411b8d2aac0e0128ba68c6ea53

## Other Environment Information

- Hardware parameters: GitHub-hosted Ubuntu/x86 runner
- OS type: Ubuntu/x86
- Others: MatrixOne ALL CI, run 30428810277, job 90501108335

## Actual Behavior

`TestISCPExecutor8` intermittently fails while waiting for the initial ISCP
watermark to reach the timestamp captured immediately after job registration:

```text
pkg/vm/engine/test/change_handle_test.go:3065
Error: Should be true
Test: TestISCPExecutor8
```

The assertion is evaluated after `testutils.WaitExpect(4000, ...)`. In the
failed run, the target database/table creation and the first ISCP iteration
started only after the assertion had already failed, indicating that the
fixed four-second wait can expire before asynchronous initialization finishes
on a loaded CI runner.

## Expected Behavior

`TestISCPExecutor8` should deterministically wait for the observable ISCP
watermark transition or fail with diagnostics that identify why progress
stopped, instead of depending on a fixed wall-clock delay.

## Steps to Reproduce

```text
.agents/skills/mo-dev/scripts/mo-cgo-test \
  -v -count=1 -timeout=10m \
  -run '^TestISCPExecutor8$' \
  ./pkg/vm/engine/test
```

The failure is intermittent. The CI run above reproduced it once. Follow-up
same-configuration runs passed 3/3 on the PR head and 3/3 on clean `mo/main`.

## Additional information

- Failing CI run:
  https://github.com/matrixorigin/matrixone/actions/runs/30428810277/job/90501108335
- The originating PR only changes grouping-set planner code and tests.
  `pkg/sql/plan` passed in the same CI run.
- There are no changes under the ISCP/engine test path between the PR base and
  the clean `mo/main` commit used for the control runs.
- This issue tracks the independent intermittent UT failure; it is not
  attributed to the planner PR.

ck89119:
Verified fixed on main at current head e457c2398fd776c5ecd53b4f612d2f5077c5114b, which contains the #26371 merge commit 83a090b5d129a50284b603e51f0f668ebd629f20.

[...]

PR #26371, opened 2026-07-29
test: stabilize ISCP watermark wait

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26365

`TestISCPExecutor8` used two fixed four-second waits for asynchronous ISCP
watermark progress. On a loaded CI runner, initialization could finish after
that deadline, and the subsequent bare boolean assertions provided no state
diagnostics.

This change:

- waits up to ten seconds for the observable watermark condition, matching the
  stabilization pattern used by adjacent ISCP tests;
- reports the account, table, job, target watermark, last observed watermark,
  and whether the job was found when the wait times out;
- checks executor startup and relevant transaction errors immediately instead
  of continuing to a misleading watermark failure.

Validation:

- `go build ./pkg/vm/engine/test`
- `go vet ./pkg/vm/engine/test`
- `TestISCPExecutor8`, 100 repeated runs
- `TestISCPExecutor8` with the race detector, 20 repeated runs
- all `TestISCPExecutor[0-9]+` tests
- full `pkg/vm/engine/test` package with the race detector
- `git diff --check`


XuPeng-SH (approved):
Deep-reviewed exact head 42513ba744a9d75b4348ebb765f68ea866f71967. This is test-only and preserves both semantic checkpoints (watermark after insert and after delete). The replacement still polls every 10ms and exits immediately on success; only the failure budget grows from 4s to 10s. GetWatermark is protected by TableEntry RLock, and previously ignored write-commit/executor-start/register errors now fail at their source. Local validation: TestISCPExecutor8 passed once, then 10 consecutive runs (13.46s total), and passed under -race. No production performance impact or test-intent weakening found.

Issue #26337, opened 2026-07-29
[Subtask]: Inventory and instrument legacy UPDATE planner fallbacks


XuPeng-SH:
## Parent issue

#26336

## Problem

`bindAndOptimizeUpdateQuery` currently treats almost every `ErrUnsupportedDML` returned by `bindUpdate` as permission to call `buildTableUpdate`; only the Iceberg message is dispatched separately. Free-form errors from deeper binder calls can therefore expand the legacy surface accidentally.

Directly verified route sources on current `main` include:

- more than one updated target table;
- any target with `Fkeys` or `RefChildTbls` (the UPDATE resolver currently ignores `foreign_key_checks` for routing);
- affected irregular-index columns or PK updates on an irregular-index table;
- PK/UNIQUE updates on `mo_pubs` / `mo_subs`;
- the `ENUM/SET + AutoIncr` compatibility branch, whose legal-schema reachability must be proven;
- writable external-table, unsupported-table-form, and empty-table-name sentinels, which are not necessarily supported legacy workloads;
- any propagated `ErrUnsupportedDML` not declared at the UPDATE entry point.

Before migration, planner selection must become an explicit, measurable decision rather than an error-text side effect.

## Required work

- Build a call-graph-backed inventory of every `ErrUnsupportedDML` that can escape `bindUpdate`, including errors propagated by shared bind helpers.
- Give every intended temporary legacy route a stable machine-readable reason.
- Classify non-legacy outcomes separately: named specialized planner, deliberate rejection, and unexpected binder error.
- During migration, allow legacy routing only for an explicit allowlist of classified reasons; an unknown reason must not silently call `buildTableUpdate`.
- Emit counters and structured log fields for selected planner and route reason without SQL text, literals, or sensitive values.
- Add executable plan fixtures for every classified outcome.
- Define the regression/soak window and evidence required by #26343.

## Acceptance criteria

- [ ] Every call from `bindAndOptimizeUpdateQuery` to `buildTableUpdate` is selected by a typed/structured reason, never by matching arbitrary error text.
- [ ] Iceberg remains a named specialized route rather than a legacy-fallback reason.
- [ ] Writable external UPDATE and invalid table forms are classified as rejection/error routes unless a test proves intentional supported behavior.
- [ ] The `ENUM/SET + AutoIncr` branch has a legal SQL reproduction or is recorded as unreachable/dead.
[...]

PR #26373, opened 2026-07-29
refactor(plan): make UPDATE planner routing explicit

ck89119:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26337

`bindAndOptimizeUpdateQuery` previously treated any propagated `ErrUnsupportedDML` as permission to invoke the legacy UPDATE planner. That allowed new binder errors to expand the legacy surface accidentally.

This change:

- introduces typed UPDATE planner routes with stable machine-readable reasons;
- allows `buildTableUpdate` only for explicitly classified legacy reasons;
- keeps Iceberg as a named specialized route;
- rejects writable external tables and unsupported table forms explicitly;
- makes unknown `ErrUnsupportedDML` values visible without falling back;
- records planner, reason, and result through `mo_sql_update_planner_route_total` and structured logs without SQL text or values;
- adds executable binder/plan fixtures for all known route classes and the unknown-route guard;
- confirms that `SET + AUTO_INCREMENT` is legal while `ENUM + AUTO_INCREMENT` is rejected.

Legacy UPDATE use can be monitored with:

```promql
sum by (planner, reason, result) (
  rate(mo_sql_update_planner_route_total{planner="legacy"}[5m])
)
```

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -p=2 -count=1 -timeout=600s ./pkg/sql/plan`
- `go test -p=2 -count=1 -timeout=300s ./pkg/util/metric/v2`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -p=2 -race -count=1 -timeout=900s ./pkg/sql/plan`
- `go test -p=2 -race -count=1 -timeout=600s ./pkg/util/metric/v2`
- `make static-check`
- `git diff mo/main...HEAD --check`


XuPeng-SH (approved):
Deep review complete. Explicit UPDATE planner routing is closed over the current UnsupportedDML sources, unknown reasons cannot expand the legacy allowlist, Iceberg remains specialized, and telemetry contains only bounded route fields. Existing multi-table/JOIN behavior remains routed to the legacy planner. Full pkg/sql/plan and pkg/util/metric/v2 tests pass; required CI is green and the exact head merges cleanly with current main.

PR #26378, opened 2026-07-29
fix(hakeeper): retain bootstrap commands until heartbeat acknowledgement

### LINKED ISSUE #26370: [Bug]: TestCalculateObjectCount flaky — waitHAKeeperRunningLocked timeout
## Problem

`TestCalculateObjectCount` fails with HAKeeper wait timeout:

```
internal error: waitHAKeeperRunningLocked
Elapsed: 127.58s
```

The test waits for HAKeeper to enter running state but the lock acquisition times out under CI load.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-29 | [PR #26329](https://github.com/matrixorigin/matrixone/actions/runs/30431705561/job/90510216588) | UT Test on Ubuntu/x86 |

Also in the same run: `TestMultiClusterCanWork` failed (33s, related HAKeeper issue, see closed #18381/#19018).

## Suggested Fix

1. Increase HAKeeper wait timeout for CI environments
2. Or add retry logic for HAKeeper state transitions

XuPeng-SH:
## What

- retain bootstrap `StartReplica` commands until the target Log/TN heartbeat reports the exact shard and replica
- preserve the existing one-shot delivery behavior for ordinary schedule commands
- add a deterministic state-machine regression test for a committed heartbeat proposal whose response is lost

## Root cause

A heartbeat proposal can commit after its caller times out. The HAKeeper state machine previously deleted the command while applying that proposal, but the timed-out service never received or executed the returned command. Bootstrap then remained in `HAKeeperBootstrapCommandsReceived` until the outer wait expired.

This fixes the delivery ambiguity rather than extending the CI timeout. It adds no sleeps, polling loops, goroutines, protocol changes, or unbounded state.

Fixes #26370

## Tests

- `TestBootstrapReplicaCommandsRetriedUntilHeartbeatAcknowledges`: `-race -count=100`
- `TestGetCommandBatch`: `-race -count=100`
- `./pkg/hakeeper`: normal and race
- `TestCalculateObjectCount`: passed in 12.32s
- `go build ./pkg/hakeeper`
- `go vet ./pkg/hakeeper`


LeftHandCold (approved):
Deep review completed on exact head 016b982e9725d32ed14e86354333cd5821f22131. No merge-blocking correctness, lifecycle, compatibility, or normal-business regression issue found. Retention is restricted to bootstrapping StartReplica commands for Log/TN services; ordinary schedule commands remain one-shot. Heartbeat state is applied before acknowledgement is evaluated, and the exact target shard/replica must be reported before removal. Duplicate delivery is bounded by heartbeat acknowledgement and follows the existing idempotent Log/TN startup paths; after transition to Running, the normal checkers also reschedule replicas missing from store heartbeats. Focused and race stress (100 rounds), full pkg/hakeeper normal/race, full pkg/logservice, bootstrap dependency normal/race, TN startup normal/race, TestCalculateObjectCount, build, vet, and diff checks pass. Completed CI checks are green; remaining checks are still running.

PR #26380, opened 2026-07-29
fix(mergeblock): handle legacy indexes and append failures

### LINKED ISSUE #26323: [Bug]: mergeblock fails on a valid non-zero legacy table index
## Description

`MergeBlock` recognizes the legacy block-info batch format containing `BlockMeta_TableIdx_Insert`, but it initializes an output batch only for table index `0`. A valid legacy batch for a unique-index table uses index `1` (and higher indices for additional unique-index tables). Processing such a batch dereferences a missing map entry.

The legacy format explicitly defines:

- `0`: the main table
- `1`: the first unique-index table
- `2`: the second unique-index table

## Environment

- Branch: `main`
- Commit: `a777e80be4b3763592e279501bf93d11c0f1713b`
- Package: `pkg/sql/colexec/mergeblock`

## Reproduction

1. Build a valid block-info batch with these attributes:

   ```text
   %!%mo__meta_tbl_index
   %!%mo__meta_loc
   __mo_object_stats
   ```

2. Put one valid encoded `BlockInfo` and `ObjectStats` row in the batch.
3. Set the legacy table-index value to `1`.
4. Call `MergeBlock.Split`.

The focused unit case was executed three times on the commit above.

## Actual result

All three runs terminate with:

```text
runtime error: invalid memory address or nil pointer dereference
pkg/sql/colexec/mergeblock/types.go:264
```

## Expected result

A valid legacy unique-index batch should be processed according to its table index. If that format cannot be supported, the operator should return an explicit error instead of terminating through a nil dereference.

## Code analysis

`GetMetaLocBat` detects the legacy table-index column but creates only:

```go
mergeBlock.container.mp[0] = bat
```

`Split` then selects an output batch using the input value:

```go
dest = mergeBlock.container.mp[int(tblIdx[i])].Vecs[0]
```

For the valid legacy value `1`, `mp[1]` is absent.

## Impact

During mixed-version processing, a legacy CN batch containing unique-index block metadata can terminate the write pipeline instead of being handled or rejected cleanly.

### LINKED ISSUE #26355: [Bug]: mergeblock silently drops block metadata on mpool allocation failure

- [x] I have checked the existing open and closed issues.

### Branch Name

main

### Commit ID

e371ea420c6c1be233c8172a4a48fa5da08125ba

- Hardware parameters: local unit test with a 1 MiB capped mpool
- OS type: macOS arm64
- Go version: go1.26.4

### Actual Behavior

`MergeBlock.Split` ignores the errors returned by three `vector.AppendBytes` calls:

- `pkg/sql/colexec/mergeblock/types.go:224`
- `pkg/sql/colexec/mergeblock/types.go:227`
- `pkg/sql/colexec/mergeblock/types.go:268`

For the block-info path, `affectedRows` is incremented before the append. If the append fails because the process mpool is exhausted:

1. `Split` returns `nil`.
2. The destination metadata batch contains zero rows.
3. `affectedRows` still includes the dropped block's rows.
4. `MergeBlock.Call` skips `source.Write` because `RowCount() == 0`, so the failure is not propagated to the pipeline.
5. A later batch can succeed, leaving the final affected-row count polluted by the silently dropped batch.

Observed with one block containing 15 rows:

```text
first Split while mpool is exhausted:
  error         = nil
  output rows   = 0
  affectedRows  = 15

second Split after releasing memory pressure:
  error         = nil
  output rows   = 1
  affectedRows  = 30
```

This can silently omit block metadata from the transaction workspace while reporting a successful operation and an inflated affected-row count.

### Expected Behavior

An allocation failure from `vector.AppendBytes` should be returned from `Split` so the pipeline fails instead of continuing with an incomplete metadata batch. `affectedRows` should only be updated after the corresponding metadata/data has been accepted successfully, or it should be rolled back on failure.

### Steps to Reproduce

Add a unit test in `pkg/sql/colexec/mergeblock/mergeblock_test.go` that:

1. Creates a process with `mpool.NewMPool(..., 1<<20, mpool.NoFixed)`.
2. Builds a one-row input batch containing `catalog.BlockMeta_BlockInfo` and `catalog.ObjectMeta_ObjectStats`. The encoded block info should report 15 rows.
3. Allocates `procMP.Cap() - procMP.CurrNB() - 1` bytes from the process mpool, leaving insufficient room for the destination `AppendBytes`.
4. Calls `arg.Split(proc, input, nil)`.
5. Observe that the call returns no error, `arg.container.mp[0].RowCount()` is 0, and `arg.container.affectedRows` is 15.
6. Free the filler allocation and call `Split` with another equivalent batch.
7. Observe that the destination now contains one row while `affectedRows` is 30.

Verification command:

```bash
go test ./pkg/sql/colexec/mergeblock \
  -run '^(TestMergeBlock|TestMergeBlockSplitSilentlyDropsBlockInfoOnAppendOOM)$' \
  -count=1 -v
```

The existing baseline test and the diagnostic reproducer both pass; the diagnostic test asserts the current silent-drop behavior described above.

### Additional information

Relevant control flow:

- `vector.AppendBytes` returns allocation errors from `BuildVarlenaFromByteSlice` / vector extension.
- `MergeBlock.Split` currently discards those return values.
- `MergeBlock.Call` writes the generated metadata batch only when its row count is greater than zero.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Closes #26323
Closes #26355

- Reject legacy unique-index table targets before mutating row state and return `ErrRetryForCNRollingRestart` so the whole transaction can be safely retried instead of panicking.
- Propagate allocation failures from all three `vector.AppendBytes` calls in the MergeBlock block-info and object-stats paths.
- Publish `affectedRows` only after the complete split succeeds, preventing failed batches from polluting the reported count.
- Add deterministic regression coverage for positive and negative legacy index encodings and for block-info/object-stats mpool exhaustion.

Design decision: MergeBlock owns only the main-table relation and Call writes only `mp[0]`/`mp2[0]`. The legacy non-zero index path is therefore failed with the rolling-restart retry error instead of adding incomplete unique-index routing solely for the mixed-version window.

Validation:

- `go build ./pkg/sql/colexec/mergeblock`
- `go vet ./pkg/sql/colexec/mergeblock`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=120s -v ./pkg/sql/colexec/mergeblock`
- focused race stress for both new tests: `-race -count=100`
- full owning package: `-race -count=1`
- dependent regression: `TestConstructAggregateConfigIncludesGroupConcatMaxLen` in `pkg/sql/compile`


XuPeng-SH (approved):
Deep-reviewed exact head 9c77fce1aecfe70fdd520deae7b4bba47b567711.

The legacy index handling is fail-closed before row-state mutation: main-table encodings 0/-1 remain supported, while both positive and negative encodings for non-zero unique-index targets return the whole-transaction rolling-restart error instead of dereferencing or misrouting through the main relation. This matches MergeBlock ownership and Call, which only write mp[0]/mp2[0].

All three AppendBytes failures now propagate. affectedRows is accumulated locally and published only after block-info/object-stats/data splitting completes, so an OOM cannot silently omit metadata while reporting success. Partial destination vectors/batches remain operator-owned and are released by Reset/Free on the failed pipeline; a subsequent successful batch starts from cleaned metadata state.

The validation scan is limited to the legacy table-index format; current-format batches add no scan. Normal-path changes are constant-time error checks/local accumulation, with no extra I/O, copies, or RPCs. Full deterministic CGo tests for pkg/sql/colexec/mergeblock and the dependent compile regression pass. One GitHub Ubuntu UT job is still reported in progress; all other reported checks are terminal/pass or skipped.

Issue #26374, opened 2026-07-29
[Bug]: Rolling restart CN lock tests flaky — TestReLockInRollingRestartCN and related


matrix-meow:
## Problem

Multiple rolling restart CN lock tests fail intermittently:

| Test | Elapsed |
|---|---|
| TestClusterCanStartNewCNServices | — |
| TestReLockInRollingRestartCN | 1.43s |
| TestPreTxnLockInRollingRestartCN | 1.62s |
| TestOldTxnLockInRollingRestartCN | 7.65s |
| TestIssue5543 | 6.76s |

All tests involve restarting CN nodes and verifying lock re-acquisition. Under CI load, the CN restart/lock handoff timing causes failures.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-29 | [PR #26366](https://github.com/matrixorigin/matrixone/actions/runs/30432883515/job/90513927654) | UT Test on Ubuntu/x86 |

## Related Closed Issues

- #22121 — TestReLockInRollingRestartCN failed (closed)
- #16739 — TestReLockInRollingRestartCN (closed)
- #16492 — TestReLockInRollingRestartCN (closed)

## Suggested Fix

1. Increase CN restart/lock handoff timeouts for CI
2. Or add retry logic for transient restart failures
3. Or use synchronization barriers instead of timing-based waits

daviszhen:
fixed

PR #26381, opened 2026-07-29
fix ut

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26374

  - 锁服务测试：将忙等改为最长 10 秒、10ms 间隔的状态轮询，明确等待 LockWaiting 与锁表引用的创建/清理完成。
  - CN 启动测试：将一次性 SQL 就绪检查改为最长 30 秒的重试检查，每次请求限时 1 秒。
  - 不涉及生产逻辑修改，仅调整测试稳定性。


XuPeng-SH (changes requested):
[P1] The changes do not cover the failure modes recorded in #26374. (1) TestClusterCanStartNewCNServices failed by panicking inside the CN bootstrap goroutine before validCNCanWork runs, so retrying the later SQL readiness check cannot prevent it. (2) TestReLockInRollingRestartCN and TestPreTxnLockInRollingRestartCN failed after the service reached the requested state, when the following one-shot remote Lock returned backend connection closed. State polling does not retry those operations; please use the existing rolling-restart retry contract/helper for the transient RPC errors and add deterministic coverage. Also prefer a fatal readiness assertion so a timed-out prerequisite does not let the test continue.

daviszhen:
> [P1] The changes do not cover the failure modes recorded in #26374. (1) TestClusterCanStartNewCNServices failed by panicking inside the CN bootstrap goroutine before validCNCanWork runs, so retrying the later SQL readiness check cannot prevent it. (2) TestReLockInRollingRestartCN and TestPreTxnLockInRollingRestartCN failed after the service reached the requested state, when the following one-shot remote Lock returned backend connection closed. State polling does not retry those operations; please use the existing rolling-restart retry contract/helper for the transient RPC errors and add deterministic coverage. Also prefer a fatal readiness assertion so a timed-out prerequisite does not let the test continue.

  - 将 TestReLockInRollingRestartCN 与 TestPreTxnLockInRollingRestartCN 的一次性远端 Lock 改为使用既有滚动重启重试 helper。
  - 覆盖 ErrBackendClosed、ErrBackendCannotConnect、bind 变更等短暂错误；ReLock 在 bind 变更后释放旧事务并更新事务 ID/快照。
  - 新增确定性单测，验证 transient RPC error 会重试至成功。
  - 将 CN SQL readiness 检查改为 require.EventuallyWithT，超时立即终止测试，避免前置就绪失败后继续执行。

XuPeng-SH (changes requested):
[P1] The lockservice half of the previous review is now fixed, but the `TestClusterCanStartNewCNServices` failure from #26374 is still not addressed. In the cited failing run, the process panics in `cnservice.handleBootstrapErr -> (*service).bootstrap -> initEngine -> NewService -> embed.(*operator).Start`, from the CN-start goroutine created by `cluster.doStartLocked`. This happens while `c.Start()` is waiting and before any of the `validCNCanWork` calls at `cluster_test.go:93-95` can run.

This PR only retries the later SQL probe inside `validCNCanWork`; the latest update changes that retry from `assert.EventuallyWithT` to `require.EventuallyWithT`. Making the later timeout fatal is correct cleanup, but neither change can catch or prevent the earlier bootstrap panic caused by the transient HAKeeper connection reset. A passing local/CI invocation is not evidence of closure for this intermittent path because no code on the failing stack changed.

Please fix or synchronize the concurrent CN bootstrap/start path that produces the transient panic (or otherwise make this test use a startup contract that can retry/return that failure safely), and add deterministic fault/scheduling coverage for the pre-readiness failure. The rolling-restart Lock retries and their new focused test look good.

daviszhen:
> [P1] The lockservice half of the previous review is now fixed, but the `TestClusterCanStartNewCNServices` failure from #26374 is still not addressed. In the cited failing run, the process panics in `cnservice.handleBootstrapErr -> (*service).bootstrap -> initEngine -> NewService -> embed.(*operator).Start`, from the CN-start goroutine created by `cluster.doStartLocked`. This happens while `c.Start()` is waiting and before any of the `validCNCanWork` calls at `cluster_test.go:93-95` can run.
> 
> This PR only retries the later SQL probe inside `validCNCanWork`; the latest update changes that retry from `assert.EventuallyWithT` to `require.EventuallyWithT`. Making the later timeout fatal is correct cleanup, but neither change can catch or prevent the earlier bootstrap panic caused by the transient HAKeeper connection reset. A passing local/CI invocation is not evidence of closure for this intermittent path because no code on the failing stack changed.
> 
> Please fix or synchronize the concurrent CN bootstrap/start path that produces the transient panic (or otherwise make this test use a startup contract that can retry/return that failure safely), and add deterministic fault/scheduling coverage for the pre-readiness failure. The rolling-restart Lock retries and their new focused test look good.

• - 将 CN bootstrap 的连接错误处理从“立即 panic”改为“在既有 5 分钟 bootstrap deadline 内重试”。
  - 重试范围覆盖 HAKeeper 常见连接失败：connection reset、backend closed、backend cannot connect。
  - 保留非连接类 bootstrap 错误的原有 fail-fast 行为。
  - 新增确定性 UT：每类连接错误首次失败、后续成功；非连接错误验证不重试。
  - 保留并验证 lockservice 的滚动重启重试修复：后续 Lock 遇到短暂 RPC 断连会重试，readiness 超时会直接失败。
  - 修正 TestIssue5543：子 goroutine 通过 error channel 向主测试 goroutine 回传错误。

XuPeng-SH (changes requested):
[P1] `bootstrapWithRetry` cannot safely retry the whole `bootstrap.Service.Bootstrap` state machine after bootstrap ownership has been acquired. `Bootstrap` first calls `locker.Get`, and the CN locker implements that with `AllocateIDByKeyWithBatch(..., 1)`: only ID 1 owns initialization. If that first owner then gets a connection error from `execBootstrap`, the new wrapper starts `Bootstrap` again; the second call gets ID 2, takes the non-owner wait branch, and no owner remains to complete bootstrap. It waits until the five-minute context deadline and `handleBootstrapErr` still panics. The same unknown-outcome risk exists if the allocation RPC commits ID 1 but its response is lost.

I reproduced the post-lock case with the real `bootstrap.Service` and its stateful `memLocker`: return `ErrConnectionReset` from the first init SQL after `Get` succeeds, then retry. The second call never re-enters initialization, the keyed ID advances to 2, and it returns `context.DeadlineExceeded`. `TestBootstrapWithRetry` misses this because its stateless callback simply returns success on attempt two and bypasses the ownership state machine.

Please put the retry at a layer/phase that preserves an already acquired bootstrap privilege (and handles the allocation unknown-outcome case), with a deterministic test injecting the connection failure after ownership acquisition. Also update the PR description: this head now changes production CN bootstrap behavior, contrary to the current “test-only” statement.

daviszhen:
> [P1] `bootstrapWithRetry` cannot safely retry the whole `bootstrap.Service.Bootstrap` state machine after bootstrap ownership has been acquired. `Bootstrap` first calls `locker.Get`, and the CN locker implements that with `AllocateIDByKeyWithBatch(..., 1)`: only ID 1 owns initialization. If that first owner then gets a connection error from `execBootstrap`, the new wrapper starts `Bootstrap` again; the second call gets ID 2, takes the non-owner wait branch, and no owner remains to complete bootstrap. It waits until the five-minute context deadline and `handleBootstrapErr` still panics. The same unknown-outcome risk exists if the allocation RPC commits ID 1 but its response is lost.
> 
> I reproduced the post-lock case with the real `bootstrap.Service` and its stateful `memLocker`: return `ErrConnectionReset` from the first init SQL after `Get` succeeds, then retry. The second call never re-enters initialization, the keyed ID advances to 2, and it returns `context.DeadlineExceeded`. `TestBootstrapWithRetry` misses this because its stateless callback simply returns success on attempt two and bypasses the ownership state machine.
> 
> Please put the retry at a layer/phase that preserves an already acquired bootstrap privilege (and handles the allocation unknown-outcome case), with a deterministic test injecting the connection failure after ownership acquisition. Also update the PR description: this head now changes production CN bootstrap behavior, contrary to the current “test-only” statement.

  - 将 bootstrap 初始化重试收敛到已获得 owner 权限后的 execBootstrap 阶段，避免重试整个 Bootstrap 状态机导致再次分配锁 ID。
  - 为 CNAllocateID 增加 RequestID，CN 使用稳定的 cfg.UUID 作为该值。
  - HAKeeper RSM 持久化 (bootstrap key, RequestID) -> firstID，响应丢失后的重传、HAKeeper snapshot 恢复或 CN 重启都会返回原始 ID，而不是分配 ID 2。
  - CN locker 强制使用幂等分配接口；不支持时 fail-fast，不会回退到可能导致无限等待的旧接口。
  - 新增 UT 覆盖：owner 初始化连接错误重试、分配响应丢失重传、snapshot 后幂等性、CN UUID 传递和不支持客户端拒绝。

XuPeng-SH (changes requested):
Deep re-review of exact head 7b964acf0bd515c921c85e8b7cdc63dd108d675c. The idempotent HAKeeper allocation closes the prior lock-ownership hole, but bootstrap connection-error closure is still incomplete.

[P1] Make bootstrap retries phase-aware across the initial check and an unknown transaction commit — `pkg/bootstrap/service.go:186-190`, `pkg/bootstrap/service.go:223-237`

Two retryable windows still reach `handleBootstrapErr` or can make the owner panic:

1. `Bootstrap` returns any connection error from the initial `checkAlreadyBootstrapped` directly, before the new idempotent lock path is reached. A transient SQL connection reset in this pre-owner check therefore still reproduces the original pre-readiness failure class.

2. `execBootstrapWithRetry` blindly starts the full bootstrap transaction again after every connection error. If `ExecTxn` committed the initialization but the commit response was lost, the next attempt does not re-check bootstrap state and reruns non-idempotent initialization SQL (`create database`, version insert, etc.); that can fail with a non-connection duplicate-object error and still panic. This is the SQL-side equivalent of the allocation unknown-outcome case the new request ID correctly handles.

`TestBootstrapRetriesOwnerInitializationWithoutReacquiringLock` only injects the error from the first init SQL before a successful commit, so it cannot exercise either case. Please retry the safe pre-owner check, and after an owner transaction connection error reconcile committed bootstrap state before deciding whether to rerun, with deterministic tests for the initial-check reset and a committed-but-response-lost `ExecTxn`. The rolling-restart lock retry changes and idempotent HAKeeper request handling otherwise look sound.

daviszhen:
> Deep re-review of exact head [7b964ac](https://github.com/matrixorigin/matrixone/commit/7b964acf0bd515c921c85e8b7cdc63dd108d675c). The idempotent HAKeeper allocation closes the prior lock-ownership hole, but bootstrap connection-error closure is still incomplete.
> 
> [P1] Make bootstrap retries phase-aware across the initial check and an unknown transaction commit — `pkg/bootstrap/service.go:186-190`, `pkg/bootstrap/service.go:223-237`
> 
> Two retryable windows still reach `handleBootstrapErr` or can make the owner panic:
> 
> 1. `Bootstrap` returns any connection error from the initial `checkAlreadyBootstrapped` directly, before the new idempotent lock path is reached. A transient SQL connection reset in this pre-owner check therefore still reproduces the original pre-readiness failure class.
> 2. `execBootstrapWithRetry` blindly starts the full bootstrap transaction again after every connection error. If `ExecTxn` committed the initialization but the commit response was lost, the next attempt does not re-check bootstrap state and reruns non-idempotent initialization SQL (`create database`, version insert, etc.); that can fail with a non-connection duplicate-object error and still panic. This is the SQL-side equivalent of the allocation unknown-outcome case the new request ID correctly handles.
> 
> `TestBootstrapRetriesOwnerInitializationWithoutReacquiringLock` only injects the error from the first init SQL before a successful commit, so it cannot exercise either case. Please retry the safe pre-owner check, and after an owner transaction connection error reconcile committed bootstrap state before deciding whether to rerun, with deterministic tests for the initial-check reset and a committed-but-response-lost `ExecTxn`. The rolling-restart lock retry changes and idempotent HAKeeper request handling otherwise look sound.

  - 初始 bootstrap 状态检查遇到连接错误或 ErrTxnUnknown 时自动重试。
  - bootstrap 重试改为阶段感知：
      - 提交前失败可安全重试，且不重新获取 owner lock。
      - 提交结果未知时只轮询 bootstrap 状态，不重放非幂等初始化 SQL。

  - 对账过程支持延迟可见、临时连接错误和 context 超时退出。
  - 增加确定性单测，覆盖初始检查失败、提交响应丢失、延迟可见、禁止 SQL 重放及超时退出。

XuPeng-SH (changes requested):
Deep re-review of exact head `fd0ea3a24a344a60fa29559455633ebec4a71caf`. The phase-aware uncertain-commit reconciliation closes the two conceptual gaps from the previous review, and the new focused tests pass, but the retry classifier still does not receive the error shape produced by the real SQL executor.

[P1] Preserve retryable error classification through the SQL executor rollback wrapper — `pkg/bootstrap/service.go:250-251`

`checkAlreadyBootstrapped` calls `SQLExecutor.Exec`, and pre-commit initialization runs inside `SQLExecutor.ExecTxn`. In the production implementation, a transaction-body error takes `pkg/sql/compile/sql_executor.go:179-186` and `txnExecutor.rollback` returns `errors.Join(originalErr, Rollback(...))` even when rollback succeeds (`sql_executor.go:670-674`). The new `isBootstrapRetryableError` then calls classifiers whose MatrixOne-code paths only accept a top-level `*moerr.Error`; they do not traverse that joined wrapper.

On this exact head I exercised `isBootstrapRetryableError(errors.Join(err, nil))`: connection reset happens to pass via `morpc` string matching, but `ErrBackendClosed`, `ErrBackendCannotConnect`, and `ErrTxnUnknown` all return false. Consequently those errors from the initial state check or an init SQL before commit still return from `Bootstrap` and reach the CN bootstrap panic instead of the new retry path. `TestBootstrapRetriesInitialStateCheckBeforeLock` uses `NewMemExecutor`, and the owner retry test also returns raw errors, so both bypass the production wrapper and cannot catch this.

Please make the retry classification unwrap/traverse joined errors (or fix the shared classifiers) and add deterministic coverage using the production-shaped joined error for the initial-check and pre-commit paths. The uncertain-commit no-replay logic and its delayed-visibility/context tests otherwise look sound.

Non-blocking: the PR description still says this only changes tests, although the head now changes production bootstrap, HAKeeper allocation, and protocol state.

daviszhen:
> Deep re-review of exact head `fd0ea3a24a344a60fa29559455633ebec4a71caf`. The phase-aware uncertain-commit reconciliation closes the two conceptual gaps from the previous review, and the new focused tests pass, but the retry classifier still does not receive the error shape produced by the real SQL executor.
> 
> [P1] Preserve retryable error classification through the SQL executor rollback wrapper — `pkg/bootstrap/service.go:250-251`
> 
> `checkAlreadyBootstrapped` calls `SQLExecutor.Exec`, and pre-commit initialization runs inside `SQLExecutor.ExecTxn`. In the production implementation, a transaction-body error takes `pkg/sql/compile/sql_executor.go:179-186` and `txnExecutor.rollback` returns `errors.Join(originalErr, Rollback(...))` even when rollback succeeds (`sql_executor.go:670-674`). The new `isBootstrapRetryableError` then calls classifiers whose MatrixOne-code paths only accept a top-level `*moerr.Error`; they do not traverse that joined wrapper.
> 
> On this exact head I exercised `isBootstrapRetryableError(errors.Join(err, nil))`: connection reset happens to pass via `morpc` string matching, but `ErrBackendClosed`, `ErrBackendCannotConnect`, and `ErrTxnUnknown` all return false. Consequently those errors from the initial state check or an init SQL before commit still return from `Bootstrap` and reach the CN bootstrap panic instead of the new retry path. `TestBootstrapRetriesInitialStateCheckBeforeLock` uses `NewMemExecutor`, and the owner retry test also returns raw errors, so both bypass the production wrapper and cannot catch this.
> 
> Please make the retry classification unwrap/traverse joined errors (or fix the shared classifiers) and add deterministic coverage using the production-shaped joined error for the initial-check and pre-commit paths. The uncertain-commit no-replay logic and its delayed-visibility/context tests otherwise look sound.
> 
> Non-blocking: the PR description still says this only changes tests, although the head now changes production bootstrap, HAKeeper allocation, and protocol state.

  - 修复 bootstrap 对 errors.Join(bodyErr, rollbackErr) 的识别：递归解包 joined/wrapped error，并以首个 body error 决定是否重试。
  - 避免 rollback 的临时连接错误掩盖不可重试的初始化 SQL 错误，防止错误重放 bootstrap。
  - 补充锁前 checkAlreadyBootstrapped 的 joined-error 重试覆盖。
  - 补充 owner 初始化阶段的 joined-error 重试覆盖，确认不重复获取 bootstrap lock。
  - 增加反向回归：不可重试 body error 即使 joined 了可重试 rollback error，也会直接返回而不重试。

XuPeng-SH (approved):
Deep re-review of exact head `a5e608784b7531f4a08f1315ae013a2fd9436322`: the previous blocker is resolved; no remaining blocking findings.

`isBootstrapRetryableError` now traverses production-shaped single/joined wrappers and deliberately classifies only the transaction-body child, so backend-closed/backend-cannot-connect/txn-unknown errors survive SQLExecutor rollback wrapping without allowing a retryable rollback error to turn a non-retryable bootstrap SQL failure into a replay. The phase-aware ownership and uncertain-commit safeguards from the prior head remain intact.

Fresh local evidence: all six bootstrap retry/reconciliation regressions pass under `-race -count=5`; `go build`, `go vet`, and `git diff --check` pass across the affected bootstrap/CN/embed/HAKeeper/lockservice/logservice packages. Latest CI is green and the PR is MERGEABLE.

PR #26382, opened 2026-07-29
feat(sql): persist and enforce CHECK constraints for INSERT

### LINKED ISSUE #26368: [Subtask]: Persist and enforce CHECK constraints for INSERT
## Parent Issue

Parent issue: #24918

## Detail of Subtask

Establish the shared CHECK-constraint foundation required by #24950:

- Bind `CREATE TABLE` CHECK expressions into structured `TableDef` metadata.
- Persist and restore CHECK metadata across catalog round trips, including
  transaction-local relations.
- Enforce CHECK constraints for `INSERT` after defaults, generated columns,
  and casts have produced the final row.
- Apply SQL CHECK semantics: `NULL` passes and `FALSE` fails.
- Return a constraint-violation error and ensure failed rows are not written.

This task intentionally excludes `REPLACE`, which remains in #24950. It also
excludes `UPDATE`, ALTER/RENAME compatibility, IGNORE warning behavior,
`SHOW WARNINGS`, and unrelated MySQL constraint-name compatibility.

## Describe implementation you've considered

- Add versioned, safely skippable structured storage for CHECK metadata instead
  of extending an unframed legacy constraint stream.
- Restore the structured metadata for both catalog-backed and
  transaction-local relations.
- Add one shared planner helper that validates the final projected row and can
  later be reused by `REPLACE`.
- Gate any new plan or execution representation that an older CN cannot safely
  interpret during rolling upgrades.
- Cover `INSERT ... VALUES`, `INSERT ... SELECT`, `NULL`/`FALSE` behavior,
  catalog re-resolution, transaction-local tables, and no-write-on-failure.

## Additional information

This issue blocks #24950. The previous combined implementation in #25070 was
closed because it accumulated unrelated compatibility work and diverged
substantially from the current `main` branch.

The clean baseline used for the redesign is
`f6dab28046d70412cec132f0068840896852101c`.

ck89119:
## What type of PR is this?

- [x] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [x] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26368

This PR establishes the structured CHECK-constraint prerequisite needed by
subsequent DML work:

- persists `plan.CheckDef` values in the length-delimited `SchemaExtra`
  protobuf and restores them through catalog and transaction table paths;
- binds and validates column-level and table-level CHECK expressions during
  `CREATE TABLE`;
- rejects unsupported `NOT ENFORCED` constraints and expressions that depend
  on other columns, variables, parameters, AUTO_INCREMENT columns, session
  state, or non-deterministic functions;
- enforces CHECK constraints for regular `INSERT` and filters violating rows
  for `INSERT IGNORE`;
- returns the CHECK-specific constraint violation error for regular failures;
- protocol-gates CHECK creation and enforcement until all CNs support version
  6, preventing rolling-upgrade enforcement gaps.

Validation:

- full tests for `pkg/sql/plan/function`, `pkg/sql/plan`, `pkg/vm/engine`, and
  `pkg/vm/engine/disttae/cache`;
- MySQL parser and tree package tests;
- focused race tests repeated 100 times;
- full race tests for all four affected packages;
- `make build`;
- `make static-check` (`golangci-lint`: 0 issues);
- `git diff --check`.

A distributed INSERT BVT case is included. Local execution was not counted as
validation because another worktree owned the standard service ports and the
attempted isolated launch detected pre-existing local cluster state.


gouhongshen (approved):
## Codex automated review

No actionable defects found. Existing reviews, issue comments, and inline comments were inspected; none contained prior findings. The implementation correctly persists CHECK metadata, protocol-gates mixed-version clusters, and validates final INSERT row images after casts/defaults/generated columns, including ODKU and NULL semantics. Protobuf packages pass. Broader focused tests could not run because the checkout lacks the required prebuilt cgo/libmo.dylib; no weakened fallback was used.

XuPeng-SH (changes requested):
Blocking issue at exact head b8a0990cc0871969d0d562594e7e47a129532809:

- `pkg/sql/plan/build_ddl.go:1513` and `:1641` reject CHECK constraints for `stmt.Param` and `stmt.IcebergParam`, but omit `stmt.MongoDBParam`. As a result both column-level and table-level CHECK constraints are accepted on MongoDB external tables. I reproduced this with `CREATE EXTERNAL TABLE tpch.mongo_check (v BIGINT CHECK (v > 0)) ENGINE=MONGODB WITH ("connection"='source', "database"='telemetry', "collection"='samples')`: planning succeeds and the resulting TableDef contains one CheckDef. This leaves inconsistent metadata: external tables have no INSERT enforcement path, and `constructCheckDefs` deliberately hides all checks for `SystemExternalRel`, so SHOW CREATE also drops the accepted constraint. Please include `stmt.MongoDBParam != nil` in both external-table guards and add column/table-level regression coverage.

The core CHECK persistence/enforcement path and focused/full owning-package race tests otherwise passed, and exact-head required CI is green.

aptend (changes requested):
[P1] Preserve legacy CHECK definitions in CREATE TABLE LIKE

`buildCreateTable` always calls `constructCreateTableSQL(..., false)`, which strips both structured and legacy CHECK definitions, and `preserveChecksForCreateLike` returns immediately when `src.Checks` is empty. Tables created before this PR have CHECK text only in `Createsql` because `SchemaExtra` did not persist `CheckDef`; cloning one after upgrade therefore silently removes the constraint from both the clone metadata and its CREATE SQL.

Reproduction: resolve a source table whose `Createsql` contains `CONSTRAINT legacy_positive CHECK (a > 0)` and whose `Checks` is nil, then build `CREATE TABLE clone_t LIKE source_t`; the resulting table has no CHECK metadata and its stored CREATE SQL no longer contains `legacy_positive`.

Please retain/extract the legacy top-level definitions for this path (and apply the protocol gate if they are rebound into structured checks), with a test simulating a pre-upgrade source table.

aptend (approved):
Re-reviewed the complete diff at 09c47fd6d2. No blocking findings. The legacy CREATE TABLE LIKE CHECK-loss issue is fixed with conservative all-parser-mode recovery, ambiguity rejection, protocol gating, and regression coverage. Focused race tests and current CI pass.

XuPeng-SH (approved):
Deep-reviewed exact head 09c47fd6d2379524f4c925dc74249602435620e0. No blocking issues found. Verified CHECK metadata persistence/restore and deep-copy ownership, CREATE/LIKE recovery across SQL modes, protocol gating, INSERT/INSERT IGNORE/ODKU/LOAD enforcement placement, NULL/error semantics, and compatibility paths. Focused race tests passed at count=100, all owning packages passed full race and non-race tests, and modified/dependent packages passed build, vet, and diff checks.

XuPeng-SH:
@ck89119 conflict

heni02 (approved):

PR #26383, opened 2026-07-29
fix: enforce parent foreign key actions for REPLACE

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24951

Make REPLACE apply parent-side foreign key semantics to the actual conflicting old rows. The plan now orders and materializes parent/unique-key locks before FK actions, supports RESTRICT/NO ACTION/CASCADE/SET NULL including recursive and self-referencing cases, preserves prepared-plan sensitivity to foreign_key_checks, and carries lock metadata through compile and remote execution.

It also adds planner, compile, lock operator, prepared-plan, BVT, and pessimistic concurrency coverage.


XuPeng-SH (approved):
Reviewed exact head edb179f71b74. I checked parent/child and unique-index lock ordering, shared/exclusive mode propagation through local/remote/deep-copy paths, NULL key handling, old-row ownership across RESTRICT/CASCADE/SET NULL (including self-reference), and prepared-plan sensitivity to foreign_key_checks. Local full package tests for pkg/sql/plan, pkg/sql/colexec/lockop, and pkg/sql/compile pass; the focused frontend prepared-cache test, build, and vet also pass. Current CI is green.

heni02 (approved):

PR #26385, opened 2026-07-29
fix(frontend): support bool and bit literal pick keys

### LINKED ISSUE #26074: [Bug]: DATA BRANCH PICK literal KEYS rejects BOOL and BIT primary keys
## Description

`DATA BRANCH PICK ... KEYS(...)` rejects literal keys when the primary key type is `BOOL` or `BIT`, even though the same operation succeeds when the key is provided by a typed subquery.

The literal path and subquery path therefore support different primary-key type sets.

## Environment

- Branch: `main`
- Version: `8.0.30-MatrixOne-v1.3.0`

## Steps to reproduce

```sql
DROP DATABASE IF EXISTS g4d_pick_bool_bit;
CREATE DATABASE g4d_pick_bool_bit;
USE g4d_pick_bool_bit;

CREATE TABLE bool_base(k BOOL PRIMARY KEY, v INT);
INSERT INTO bool_base VALUES(FALSE, 10), (TRUE, 20);
DATA BRANCH CREATE TABLE bool_src FROM bool_base;
DATA BRANCH CREATE TABLE bool_dst FROM bool_base;
UPDATE bool_src SET v = v + 100;

DATA BRANCH PICK bool_src INTO bool_dst KEYS(1);
DATA BRANCH PICK bool_src INTO bool_dst KEYS(TRUE);

-- Control: the same BOOL key succeeds through the typed subquery path.
DATA BRANCH PICK bool_src INTO bool_dst
  KEYS(SELECT k FROM bool_src ORDER BY k DESC LIMIT 1);

CREATE TABLE bit_base(k BIT(8) PRIMARY KEY, v INT);
INSERT INTO bit_base VALUES(b'00000001', 10), (b'00000010', 20);
DATA BRANCH CREATE TABLE bit_src FROM bit_base;
DATA BRANCH CREATE TABLE bit_dst FROM bit_base;
UPDATE bit_src SET v = v + 100;

DATA BRANCH PICK bit_src INTO bit_dst KEYS(1);

-- Control: the same BIT key succeeds through the typed subquery path.
DATA BRANCH PICK bit_src INTO bit_dst
  KEYS(SELECT k FROM bit_src ORDER BY k LIMIT 1);
```

## Actual result

The literal cases fail consistently:

```text
invalid input: unsupported PK type for engine filter: BOOL
invalid input: unsupported PK type for engine filter: BIT
```

The corresponding typed-subquery cases succeed and update the expected destination rows.

The result reproduced in 3 out of 3 runs.

## Expected result

Literal `KEYS(...)` should accept values compatible with `BOOL` and `BIT` primary keys, consistently with the typed-subquery form.

## Code observation

`appendNumericStringToVec` in `pkg/frontend/data_branch_pick.go` does not handle `BOOL` or `BIT`, while `formatPickKeyVectorValueAsString` supports these types for the subquery path.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes https://github.com/matrixorigin/matrixone/issues/26074

`DATA BRANCH PICK ... KEYS(...)` materializes literal primary-key values through `appendNumericStringToVec`, but that path omitted BOOL and BIT even though typed subqueries already support both physical vector types.

This change:

- parses BOOL literals with the existing `types.ParseBool` semantics, including `TRUE` and numeric boolean values;
- parses BIT literals using the column width so out-of-range values are rejected before entering the filter vector;
- adds regression coverage for BOOL literals, BIT(8) boundaries, and the BOOL AST path used by `KEYS(TRUE)`.

Tests:

- `go test ./pkg/frontend -run '^(TestAppendNumericStringToVec|TestAppendExprToVec)' -count=1`
- `go test -race ./pkg/frontend -run '^(TestAppendNumericStringToVec|TestAppendExprToVec)' -count=1`


XuPeng-SH (approved):
Deep-reviewed exact head 960ed9c8651c21a0e9a7a07f593388283b8a0270. The literal path now materializes BOOL and width-bounded BIT values into the same physical vector types already supported by the typed-subquery path. I checked scalar/composite callers, sorting, hashmap insertion, ZoneMap construction, parse/range errors, and cleanup on failure; no ownership or lifecycle regression found. Focused tests pass under race x20, the full frontend package and go vet pass, git diff --check is clean, the head merges cleanly with current main, and the latest CI checks are green. No blocking issue found.

Issue #26123, opened 2026-07-24
[Bug]: Database CLONE and DATA BRANCH misresolve foreign-key table names containing LIKE metacharacters


Ariznawlll:
### Describe the bug

Database-level `CREATE DATABASE ... CLONE` and
`DATA BRANCH CREATE DATABASE` fail when a foreign-key table name contains SQL
`LIKE` pattern characters.

For example, when the source contains both `a0b` and the foreign-key table
`a_b`, both database-copy statements try to create `a0b` twice and return
`table a0b already exists`. The destination database is rolled back.

A foreign-key table named `child\fk` also fails because the backslash is
interpreted again by the `LIKE` matcher. In that case the statements cannot
resolve the child table while rebuilding the foreign key.

Tested at commit `2eb7f523e5a10f352d3e258183ed8440510388de`.

### Steps to reproduce

```sql
drop database if exists g4d_fk_like_src;
drop database if exists g4d_fk_like_dst;
drop database if exists g4d_fk_like_clone;

create database g4d_fk_like_src;
create table g4d_fk_like_src.parent_t(
  id int primary key,
  note varchar(40)
);
create table g4d_fk_like_src.a0b(
  id int primary key,
  note varchar(40)
);
create table g4d_fk_like_src.`a_b`(
  id int primary key,
  parent_id int,
  note varchar(40),
  constraint fk_a_b_parent foreign key(parent_id)
    references g4d_fk_like_src.parent_t(id)
);

insert into g4d_fk_like_src.parent_t values (1, 'parent');
insert into g4d_fk_like_src.a0b values (10, 'similar-name');
insert into g4d_fk_like_src.`a_b` values (20, 1, 'fk-child');

select relname
from mo_catalog.mo_tables
where reldatabase = 'g4d_fk_like_src'
  and relname like 'a_b'
order by relname;

data branch create database g4d_fk_like_dst
  from g4d_fk_like_src;

create database g4d_fk_like_clone
  clone g4d_fk_like_src;
```

Backslash variant:

```sql
drop database if exists g4d_fk_slash_src;
drop database if exists g4d_fk_slash_dst;

create database g4d_fk_slash_src;
create table g4d_fk_slash_src.parent_t(
  id int primary key,
  note varchar(40)
);
create table g4d_fk_slash_src.`child\fk`(
  id int primary key,
  parent_id int,
  note varchar(40),
  constraint fk_child_slash_parent foreign key(parent_id)
[...]

PR #26386, opened 2026-07-29
fix(frontend): match foreign key table names exactly

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26123

Database CLONE and Data Branch pass an exact foreign-key table name into catalog lookup, but the query used LIKE. Underscore and percent could select a different sorted relation, while backslash could suppress the intended match.

This change uses equality for nonempty table-name lookups. Database-wide enumeration remains unchanged.

The embedded regression isolates all six combinations:

- Data Branch and CLONE
- underscore, percent, and backslash identifiers

Each case asserts the exact destination relation, copied row, target-local foreign-key catalog state, and invalid-write rejection. The newly merged cyclic-FK regression also passes with this change.

Exact-head validation:

- SQL builder test: race, 100 repetitions
- six production-authentic embedded cases: race
- cyclic-FK embedded regression: race
- complete pkg/frontend and pkg/tests/issues: race
- go vet
- git diff --check
- schema v4 preflight: review=PASS

## QA handoff

- QA required: yes
- Reason: this changes user-visible CLONE/Data Branch behavior and persisted foreign-key metadata.
- Required scenarios: run database CLONE and Data Branch with FK child names containing underscore, percent, and backslash; verify exact table names and rows, target-local mo_foreign_keys rows, and rejection of a child insert with a missing parent.

### DIFF
--- a/pkg/frontend/snapshot.go
+++ b/pkg/frontend/snapshot.go
@@ -2023,7 +2023,7 @@ func buildTableInfoListWhereClause(dbName string, tblName string, accountId uint
 		)
 	}
 	if len(tblName) > 0 {
-		whereClause += fmt.Sprintf(" and relname like %s", quoteSQLStringLiteral(tblName))
+		whereClause += fmt.Sprintf(" and relname = %s", quoteSQLStringLiteral(tblName))
 	}
 	whereClause += fmt.Sprintf(" and relkind != %s", quoteSQLStringLiteral(catalog.SystemSequenceRel))
 	return whereClause
--- a/pkg/frontend/snapshot_test.go
+++ b/pkg/frontend/snapshot_test.go
@@ -441,12 +441,19 @@ func TestRestoreExternalTableDefensiveCloneGuards(t *testing.T) {
 }
 
 func TestBuildTableInfoListSQLEscapesLiterals(t *testing.T) {
-	sql := buildTableInfoListSQL("db'name", "tbl'name", 0, uint32(sysAccountID))
-	if !strings.Contains(sql, "reldatabase = 'db''name'") {
-		t.Fatalf("database name was not escaped in SQL: %s", sql)
-	}
-	if !strings.Contains(sql, "relname like 'tbl''name'") {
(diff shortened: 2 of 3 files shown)


VioletQwQ-0:
Handled the current-head Ubuntu UT failure and strengthened the exact-name regression on head `2b4e90bf6dedb8980a3b087f89d6f869468d4cd9`.

- Classified the failed `TestISCPExecutor3` assertion as unrelated upstream watermark timing; rebased onto current `main`, which includes #26371, and the exact test now passes with `-race`.
- Strengthened #26123 coverage so populated LIKE-collision tables are copied independently and source/target `mo_foreign_keys.table_name` must match exactly.
- The six Data Branch/CLONE × underscore/percent/backslash scenarios pass with `-race`; full `pkg/frontend` and `pkg/tests/issues` packages pass with `-race`; vet/compile/diff-check pass.
- The old Coverage failure belonged to the cancelled run and is superseded by successful same-head run 30447015180; it was not rerun independently.
- Exact-head preflight: `PASS review=PASS validation=PENDING`, diff hash `1f7ab65597475496014e77e524ea795c6c43d403fa34a94343561be3419b8a65`.

Please review the new head.

XuPeng-SH (approved):
Deep-reviewed on exact head `2b4e90bf6dedb8980a3b087f89d6f869468d4cd9`. The non-empty table-name argument is an exact identifier across clone, Data Branch, snapshot restore, and PITR callers; database-wide enumeration still passes an empty name and is unchanged. Replacing `LIKE` with equality fixes `_`, `%`, and backslash interpretation without adding queries, allocations, locking, or a broader scan. Verified the SQL builder under `-race` (20 repetitions), all six authenticated embedded Data Branch/CLONE regressions under `-race`, and the existing cyclic-FK Data Branch regression under `-race`.

PR #26387, opened 2026-07-29
test(motrace): restore shared tracer state

### LINKED ISSUE #26239: [Bug]: motrace unit tests retain tracer state across repeated runs
## Summary

The `pkg/util/trace/impl/motrace` unit-test package is not repeatable when the Go test runner executes it multiple times in one process. Tests change the shared tracer-provider configuration and do not restore it before the next iteration.

## Reproduction

At commit `2d0adac64ea6c72028c5443813428f84fbcd1575`:

```bash
go test ./pkg/util/trace/impl/motrace -count=3
```

The first iteration completes, while later iterations consistently fail. Representative failures include:

* `TestAggregator` receives repeated SQL text instead of the expected single value.
* `TestReportZap` receives an empty extra field instead of the expected structured fields.
* `TestStatementInfo_Report_EndStatement` observes fewer exported statements than expected.

Running the affected targets in fresh processes with `-count=1` succeeds.

## Root cause

Tests mutate the process-wide tracer provider directly (for example, statement-merge and running-statement options) and lifecycle tests change the shared provider state. Those mutations remain in place for the next test iteration.

## Expected behavior

The package test suite should reset its shared tracer configuration and lifecycle state between iterations, allowing `go test -count=3` to pass.

## Actual behavior

Later iterations inherit state from earlier ones and report deterministic assertion failures.

VioletQwQ-0:
## What type of PR is this?

- [x] Test and CI

Fixes #26239

Several motrace tests mutate fields on the package-global tracer provider, and
`Test_initExport` shuts that provider down. Repeated package runs therefore
inherit statement-merge, log-size, running-statement, or disabled-provider state
from an earlier test iteration.

Changes:

- restore every mutated tracer field with `t.Cleanup` on the exact provider
  instance that owns the field;
- run each `Test_initExport` case against a fresh no-op provider and restore the
  original provider afterward.

Validation:

- each of the 7 affected tests passed separately with `-race -count=100`;
- complete `pkg/util/trace/impl/motrace` race run passed;
- complete package `-count=3` reproduction passed;
- package build, vet, and `git diff --check` passed;
- preflight manifest: `PASS review=PASS validation=PENDING` on
  `0cd686216338f75397286bf59dbb4c039a9b6100`.

QA Decision

- required: no
- reason: this changes test isolation only; no production code or user-visible behavior changes
- production_entrypoint: repeated `go test` execution of the motrace package
- automated_terminal_coverage: affected tests and the complete package pass repeatedly in one process and under the race detector
- remaining_validation: fresh PR CI


XuPeng-SH (approved):
Deep-reviewed exact head 0cd686216338f75397286bf59dbb4c039a9b6100. Each mutated tracer field is restored on the exact provider instance that owns it. Test_initExport now uses a fresh no-op provider generation per subtest, shuts down only that generation, and restores the original provider afterward, so a closed/disabled provider cannot leak into later iterations. Cleanup ordering and provider lifecycle are sound; no production code changes. All affected targets pass under race x100, the complete package passes under race and with count=3, go vet and git diff --check pass, the head merges cleanly with current main, and the latest CI checks are green. No blocking issue found.

gouhongshen (approved):
## Codex automated review

No regressions found. The changed tests restore each captured shared-provider field and isolate shutdown to a temporary provider; no existing reviewer findings require action.

PR #26388, opened 2026-07-29
test(metric): clean up shared package state

### LINKED ISSUE #26238: [Bug]: Metric unit tests retain package state across repeated runs
## Summary

Metric unit tests are not repeatable when Go runs the same test binary more than once in a process. Package-level state modified by an earlier iteration remains visible to the next iteration.

## Reproduction

At commit `2d0adac64ea6c72028c5443813428f84fbcd1575`:

```bash
go test ./pkg/util/metric -run '^TestConnectionCounter$' -count=3 -v
go test ./pkg/util/metric/stats -run '^TestRegister$' -count=3 -v
```

Both commands fail consistently after their first iteration:

* `TestConnectionCounter` expects `1`, then observes `2` and `3`.
* `TestRegister` panics with `Duplicate Family Name MockServiceStats1` on the second iteration.

Launching each target in a fresh test process with `-count=1` succeeds three times.

## Root cause

`ConnectionCounter` returns a label instance from the package-level `ConnFactory`, while the test only calls `Inc` and does not restore the label value. `TestRegister` writes to the package-level `DefaultRegistry` and does not unregister `MockServiceStats1`.

## Expected behavior

The unit tests should be independent across repeated in-process runs so that `go test -count=3` completes successfully.

## Actual behavior

State from a prior iteration changes the next iteration's result, causing an assertion failure and a duplicate-family panic.

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes #26238

Metric tests reuse a package-global labeled gauge and the default stats registry. Repeated execution therefore accumulated the gauge value and later panicked on duplicate family registration.

This change resets the owned gauge before and after the assertion and unregisters every test-owned stats family with failure-safe cleanup. Production metric behavior is unchanged.

Exact-head validation:

- affected metric tests: race, 100 repetitions
- complete pkg/util/metric and pkg/util/metric/stats: race
- go vet
- git diff --check
- schema v4 preflight: review=PASS

## QA handoff

- QA required: no
- Reason: this is test-only package-state cleanup; repeated race execution and both complete owning packages are the terminal workflow.


XuPeng-SH (approved):
Deep-reviewed exact head 614d343d8d805e543a5d257fcef9eea7efc5badb. The change restores both package-global states at the owning test boundary: the shared labeled gauge is reset before and after mutation, and both test-owned registry families are unregistered through failure-safe cleanup registered before mutation. Unregister is locked and idempotent; no production behavior changes. Both repeated targets pass under race x100, both complete owning packages pass under race, go vet and git diff --check pass, the head merges cleanly with current main, and the latest CI checks are green. No blocking issue found.

PR #26393, opened 2026-07-29
fix(dedupjoin): preserve parallel finalize failures

### LINKED ISSUE #26375: [Bug]: dedupjoin finalize swallows context cancellation when a worker result is missing
## Description

The `dedupjoin` finalize merger silently converts context cancellation into a successful end-of-stream when one parallel worker result is missing.

When `receiveWorkerMsg` returns `nil` because `proc.Ctx` is canceled, `container.finalize` returns `nil`. `DedupJoin.Call` then advances to `End` and returns `ExecStop` with no error.

This can make a canceled/failed parallel pipeline look successful. Because finalize is responsible for emitting unmatched build rows, the same path can also silently omit output.

## Environment

- MatrixOne `main`: `f6dab28046d70412cec132f0068840896852101c`
- macOS arm64
- Go 1.26.x

## Reproduction

Add the following diagnostic test to `pkg/sql/colexec/dedupjoin/join_test.go`:

```go
func TestDedupFinalizeMissingWorkerHonorsCancellation(t *testing.T) {
    proc := testutil.NewProcessWithMPool(t, "", mpool.MustNewZero())
    arg := &DedupJoin{
        NumCPU:            2,
        IsMerger:          true,
        Channel:           make(chan *WorkerJoinMsg),
        OnDuplicateAction: plan.Node_FAIL,
    }
    arg.ctr.state = Finalize
    arg.ctr.matched = &bitmap.Bitmap{}
    arg.ctr.matched.InitWithSize(1)

    ctx, cancel := context.WithCancel(proc.Ctx)
    proc.Ctx = ctx
    cancel()

    defer func() {
        arg.Free(proc, true, context.Canceled)
        proc.Free()
    }()

    res, err := arg.Call(proc)
    if !errors.Is(err, context.Canceled) {
        t.Fatalf("cancellation was swallowed: status=%v err=%v", res.Status, err)
    }
}
```

Run:

```text
go test ./pkg/sql/colexec/dedupjoin \
  -run '^TestDedupFinalizeMissingWorkerHonorsCancellation$' \
  -count=1 -v
```

The failure reproduced twice with `-count=1`.

## Actual Result

```text
missing worker cancellation was swallowed: status=0 err=<nil>
```

`Call` returns `ExecStop` and no error.

## Expected Result

`Call` should return `context.Canceled` (or the original terminal worker/pipeline error). It must not report successful end-of-stream when finalize could not collect all worker results.

## White-box Analysis

`pkg/sql/colexec/dedupjoin/join.go`:

```go
for cnt := 1; cnt < int(ap.NumCPU); cnt++ {
    msg := receiveWorkerMsg(proc.Ctx, ap.Channel)
    if msg == nil {
        return nil
    }
    // merge worker state
}
```

`receiveWorkerMsg` returns `nil` for both a closed channel and context cancellation. The caller does not inspect `context.Cause(proc.Ctx)` and therefore treats cancellation as successful finalize completion.

After `finalize` returns `nil`, `Call` sees no buffered output, transitions to `End`, and returns `ExecStop` with no error.

## Impact

- A canceled or failed parallel dedup pipeline can be reported as successful.
- Unmatched build rows that should be emitted during finalize can be silently omitted.
- The original terminal cause is lost, preventing correct retry and incident classification.
- A missing worker and a genuinely empty/closed worker channel are indistinguishable at the caller.

## Additional Information

This issue was found during colexec unhappy-path testing. It does not involve security-sensitive behavior.

XuPeng-SH:
## What changed

- preserve context cancellation and explicit worker failures during parallel `dedupjoin` finalization instead of reporting a successful end-of-stream
- distinguish normal early aborts from errors and prevent incomplete unmatched-build output from escaping
- make worker publication non-blocking and linearize it with merger shutdown so captured vectors always have exactly one owner
- drain transferred capture state on every terminal path and retain local ownership when publication is rejected
- keep the mailbox reusable across spill buckets and prepared executions with an all-participant Reset generation barrier
- scope shared mailboxes to one parallel operator-duplication generation instead of mutating the reusable template

## Why

The old `nil` message protocol represented context cancellation, channel close, and normal abort with the same value. The merger consequently swallowed cancellation and could silently omit finalize output. It also made failure cleanup and capture ownership ambiguous.

The new protocol carries an explicit abort/error status. Sender publication never waits, merger receive remains context-cancellable, and abort/error paths stop immediately without waiting for missing workers.

## Tests

- full `pkg/sql/colexec/dedupjoin` package
- full `pkg/sql/colexec/dedupjoin` package under `-race`
- focused cancellation, ownership, generation-reuse, and data-correctness tests under `-race -count=100`
- `pkg/sql/compile` mailbox-generation duplication test
- `go vet` for `pkg/sql/colexec/dedupjoin` and `pkg/sql/compile`

All new unhappy-path tests complete in milliseconds; no sleeps, retries, external services, or full-suite stress loops are used.

Fixes #26375


XuPeng-SH:
Self-review completed on exact head `e18795a763729432e1a47ee3bbbe7f34895670b3`.

Found and fixed a production concurrency blocker in the original head: parallel finalize workers shared one mailbox but had no per-spill-bucket round barrier, so a fast worker could publish bucket N+1 before a slow worker published N and the merger could combine bitmaps from different buckets. The fix adds a mailbox-owned round completion channel: publication transfers ownership once, every worker waits for the merger to collect the complete round, cancellation unblocks the wait, and reset creates a fresh round. This is protocol-level synchronization, not a worker-count/test special case.

Unhappy paths covered: delayed worker, cancellation after publication, repeated Reset/Free, and abort draining. No steady-state row-processing cost was added; the extra channel synchronization happens only once per spill bucket during finalize. `pkg/sql/colexec/dedupjoin` and `pkg/sql/compile` full tests passed, plus 20x targeted `-race` stress. CI is running on this head.

XuPeng-SH:
Follow-up test-only head `e039e9518bb9061e6b387f51392c24f5b92433cc` strengthens the same invariant with 1 merger + 2 workers across two spill buckets: the fast worker is proven blocked until the slow worker publishes, both remain blocked until the complete round is acknowledged, then both may advance. This passed 50x under `-race`; production code is unchanged from the reviewed barrier fix.

XuPeng-SH:
Coverage fix pushed on exact head `6d90f3ec88b744f26802329867c63e7f32536634`.

The missing coverage was concentrated in real worker-publication boundaries, so the added tests exercise protocol contracts rather than line-specific branches: missing mailbox, cancellation before ownership transfer, merger already stopped, bounded mailbox unexpectedly full, and nil finalize status. These verify fail-fast behavior, no publication after cancellation, stopped-generation ownership retention, and rejection of invalid protocol messages.

Validation after the test edit: targeted tests 20x, targeted `-race` 20x, full dedupjoin tests, package build/vet, and dependent compile regression all pass. Merging the new local profile with the failed CI profile by max block hit projects modified-statement coverage at 83.23% overall and 79.17% for `join.go`, above the 75% gate with margin. New CI is running on this head.

gouhongshen (approved):
## Codex automated review

No findings. The explicit mailbox protocol preserves cancellation and worker errors, prevents partial finalize output after abort, linearizes capture ownership with shutdown, bounds waits through cancellation/stop, separates spill rounds, and isolates prepared-execution generations. Local tests and vet could not run because the checkout lacks required CGo libraries and headers; the worktree remains clean.

LeftHandCold (approved):
Deep-reviewed exact head 6d90f3ec. No merge-blocking correctness, concurrency/lifecycle, compatibility, or normal-business regression issue was found in this diff.

The mailbox protocol has a clear ownership linearization point under the stop/send lock; all capture states retain exactly one cleanup owner. Worker waits terminate through round completion, cancellation, or merger stop, and the all-participant Reset barrier prevents a stopped generation from reopening while an old participant remains active. Dynamic error/cancellation paths stop without emitting partial unmatched-build output.

Verified all 20 added/modified focused tests individually under adaptive race stress (T below timer resolution, B=30s, N=100), full dedupjoin and compile packages under normal and race builds, plus build, vet, and diff checks. Current GitHub checks are green.

PR #26395, opened 2026-07-29
perf(tae): cache object list seek pivots

XuPeng-SH:
## What changed

- prebuild immutable min/max seek pivots for all six ObjectList groups at `txnif.UncommitTS`
- reuse those pivots in forward and reverse group seeks while preserving the existing dynamic timestamp and out-of-range group paths
- add fresh-vs-cached key equivalence coverage, concurrent shared-key coverage, allocation regression assertions, and focused benchmarks

## Why

The grouped ObjectList traversal added by #26225 made `WaitUntilCommitted` perform six group seeks and visible-object iteration initialize four group seeks. Each seek rebuilt a full `ObjectEntry`, `ObjectStats`, and `ObjectName`, including dummy version entries.

The performance bisect in [mo-auto-test run 30431812058](https://github.com/matrixorigin/mo-auto-test/actions/runs/30431812058) measured read-write TPS dropping from 2,396 to 1,262 at commit `29f4ada`. A local allocation profile attributed 93.6% of `WaitUntilCommitted` allocation volume to the `makeObjectListKey` call chain.

## Impact

Single-object local benchmarks on the same host:

| Benchmark | Before | After |
| --- | --- | --- |
| WaitUntilCommitted | ~1.92 µs, 6,288 B/op, 41 allocs/op | ~116 ns, 112 B/op, 1 alloc/op |
| VisibleCommitted | ~1.67 µs, 4,296 B/op, 34 allocs/op | ~232 ns, 448 B/op, 4 allocs/op |

The cached pivots are built by the existing constructor and remain immutable. The change does not alter ObjectList ordering, BTree contents, MVCC visibility, or iterator merge order.

## Validation

- `go build ./pkg/vm/engine/tae/catalog`
- `go vet ./pkg/vm/engine/tae/catalog`
- CGo-aware tests for:
  - `./pkg/vm/engine/tae/catalog`
  - `./pkg/vm/engine/tae/txn/txnimpl`
  - `./pkg/vm/engine/tae/logtail`
  - `./pkg/vm/engine/tae/db/checkpoint`
  - `./pkg/vm/engine/disttae/logtailreplay`
- focused race stress, `count=100`, for cached/fresh equivalence and concurrent shared-key use
- complete catalog package with `-race -count=1`
- allocation regression test with `count=100`
- allocation benchmarks with `-benchmem -count=6`

## Design notes

This intentionally does not change the comparator, manually encode ObjectStats, or alter the four-way visible iterator merge. Dynamic timestamp keys continue to use the existing construction path.


LeftHandCold (approved):
Deep-reviewed exact head 126d30fc. I found no merge-blocking correctness, concurrency/lifecycle, compatibility, or normal-business regression issue.

The cached pivots are fixed, package-private immutable seek keys. The B-tree comparator only reads the relevant group/commit-TS/object-name fields and the seek path neither stores nor mutates the supplied key. Dynamic timestamps and out-of-range groups retain the existing fresh-key path; iterator/COW/ownership and wait semantics are unchanged.

Verified with adaptive per-test race stress (100 iterations for each affected functional test), full catalog tests under normal and race builds, dependent-package tests, allocation checks/benchmarks, build/vet, and diff checks.

PR #26400, opened 2026-07-29
test: stabilize user lock cleanup cap accounting

### LINKED ISSUE #26397: [Bug]: user-level lock retained cleanup cap test races worker and deadlocks on assertion

### Branch Name

main

### Commit ID

`52ea9f874f492c368a20efef16b46ab8ced2de37` (base of the failing PR; the relevant files are unchanged on current main `8694a8d26dd8ef6e05ab0bdc133e4611c7430e1e`)

- Hardware parameters: GitHub Actions Ubuntu/x86 runner, 16 GiB RAM
- OS type: Ubuntu/x86
- Others: Go 1.26.4, coverage UT

### Actual Behavior

`Matrixone UT Coverage / UT Coverage on Ubuntu/x86` fails in
`TestActiveUserLevelLockOwnersBoundRetainedCloseCleanupGrowth`:

```text
func_unary_test.go:9627:
Error: "map[]" should have 1 item(s), but has 0
panic: test timed out after 10m0s
```

The test calls `retainUserLevelLockCloseCleanup`, then asserts the transient
length of `userLevelLocks.retainedCloseCleanups`. The retained-cleanup worker
can consume that entry before the assertion, so the map is legitimately empty.

The assertion is also executed while holding `userLevelLocks`. When
`require.Len` calls `FailNow`, `userLevelLocks.Unlock()` is skipped. The deferred
`resetUserLevelLocksForTest` then blocks on the same mutex until the package-wide
10-minute timeout.

Runner diagnostics show 13 GiB available memory and 45 GiB free disk, with no
OOM signal.

### Expected Behavior

The test should deterministically verify that active-owner cleanup accounting is
bounded without depending on a transient asynchronous map entry. Assertions
that can terminate a test must run after releasing global mutexes, so an
assertion failure cannot deadlock test cleanup.

### Steps to Reproduce

1. Run the coverage UT workflow on main or a PR whose user-level-lock files are
   unchanged from main.
2. Execute `pkg/sql/plan/function` tests under coverage instrumentation and
   scheduling pressure.
3. Observe the map-length assertion fail after the retained-cleanup worker
   consumes the entry, followed by a package timeout in
   `resetUserLevelLocksForTest`.

Failing job:
https://github.com/matrixorigin/matrixone/actions/runs/30441011147/job/90540178827

### Additional information

Related but distinct closed issues:

- #26262 stabilized another retained-cleanup lifecycle test.
- #26279 covered ownership transfer in
  `TestSessionCloseRetainsCleanupWhenRetainedCapIsFull`.

The current failure is in a different cap-growth test and additionally exposes
the held-mutex assertion deadlock.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

Fixes https://github.com/matrixorigin/matrixone/issues/26397

`TestActiveUserLevelLockOwnersBoundRetainedCloseCleanupGrowth` verifies retained
cleanup admission accounting, but it started the asynchronous cleanup worker
while asserting transient global-map state. The worker could consume the entry
before the assertion.

The failed assertion also ran while holding `userLevelLocks`. Because
`require.Len` terminates the test goroutine, the mutex remained locked and the
deferred reset blocked until the package-wide ten-minute timeout.

This change:

- keeps the retained cleanup worker stopped while this accounting-only test
  observes capacity invariants;
- snapshots protected state under the mutex and performs every assertion after
  unlocking;
- restores the worker state before leaving the test.

Production user-level-lock behavior is unchanged.

Validation:

- focused test: `-count=100`
- focused race test: `-race -count=30`
- focused coverage test: `-cover -count=100`
- full `pkg/sql/plan/function` test package


XuPeng-SH (approved):
Deep-reviewed exact head e809af95f322b1024b9bfebabad4ce7e6cf15545. The accounting-only test now suppresses worker startup while inspecting transient retained state, snapshots every userLevelLocks invariant under the mutex, and performs fail-fast assertions only after unlocking. I checked all failure points: runUserLevelLockTest defers the generation-aware reset, the synthetic started-without-done state is handled safely, and no cleanup goroutine crosses into the next test generation. The focused test passes count=200, race x100, and coverage x100; the full function package and go vet pass, git diff --check is clean, the head merges cleanly with current main, and the latest CI checks are green. No blocking issue found.

PR #26401, opened 2026-07-29
fix: release mergerecursive cached batches in Free

### LINKED ISSUE #26394: [Bug]: mergerecursive Free retains copied recursive batches
## Summary

`MergeRecursive.Free` does not clean or clear the batches copied into its
internal cache.  If execution ends on a path that calls `Free` without a
preceding `Reset`, the operator retains those batches and their references.

## Reproduction

The following temporary package-level diagnostic test uses a normal recursive
input batch and invokes the standard operator lifecycle up to the first emitted
batch:

```go
source := colexec.MakeMockBatchs(proc.Mp())
source.SetLast()
arg.AppendChild(colexec.NewMockOperator().WithBatchs([]*batch.Batch{source}))
require.NoError(t, arg.Prepare(proc))

result, err := arg.Call(proc)
require.NoError(t, err)
require.NotNil(t, result.Batch)
require.NotEmpty(t, arg.ctr.freeBats)

arg.Free(proc, false, nil)
require.Nil(t, arg.ctr.freeBats)
require.Nil(t, arg.ctr.bats)
require.Nil(t, arg.ctr.buf)
```

Add the test temporarily to `pkg/sql/colexec/mergerecursive/mergerecursive_test.go`
and run it twice with:

```bash
go test ./pkg/sql/colexec/mergerecursive \
  -run TestAuditMergeRecursiveFreeReleasesCachedBatches -count=1
```

Both runs fail at `require.Nil(t, arg.ctr.freeBats)`.  After `Free`, the cache
still contains the copied batch.

## Actual result

`pkg/sql/colexec/mergerecursive/types.go:84-85` implements `Free` as an empty
method, while `Call` copies recursive batches into `ctr.freeBats`
(`mergerecursive.go:68-84`).  The slice, the pending-output queue, and the
current buffer remain reachable after `Free`.

## Expected result

`Free` should clean the cached batches and clear `freeBats`, `bats`, `buf`, and
the execution markers, matching the resource ownership cleanup already done by
`Reset`.  It must be safe when `Reset` was not reached first.

## Impact

An interrupted recursive query can retain all copied recursive batches until
the operator object is eventually discarded or reused.  With large recursive
results this causes unnecessary memory retention and can also preserve stale
operator state across lifecycle boundaries.

## Environment

- MatrixOne baseline: `f6dab28046d70412cec132f0068840896852101c`
- Package: `pkg/sql/colexec/mergerecursive`
- Go: 1.26.4

XuPeng-SH:
## What changed

- centralize `MergeRecursive` owned-batch cleanup in an idempotent helper
- invoke the same cleanup from both `Reset` and `Free`
- clear the complete execution container after releasing its uniquely owned cached batches
- add a regression test for direct `Free` after an interrupted execution, including repeated cleanup

## Root cause

`MergeRecursive.Call` duplicates recursive input batches and retains them in `ctr.freeBats`. `Reset` cleaned those owned batches, but `Free` was empty. A lifecycle path that terminated with `Free` before `Reset` therefore retained the copied vectors and the `freeBats`/`bats`/`buf` references.

The cleanup only walks `freeBats`, which is the ownership set; `bats` and `buf` are aliases and are cleared without being cleaned again, avoiding double-free.

## Impact

Interrupted recursive queries now release copied batches on terminal cleanup. The query hot path is unchanged; normal `Reset` cleanup has the same work as before, and a following `Free` sees an empty container.

## Validation

- `go build ./pkg/sql/colexec/mergerecursive`
- `go vet ./pkg/sql/colexec/mergerecursive`
- full `pkg/sql/colexec/mergerecursive` tests
- full package tests under `-race`
- dependent `pkg/sql/compile` operator conversion test

Closes #26394


XuPeng-SH:
Self-review completed on exact head `9acdcb3f85d7841b78833d65a21ba7338f413a4c`.

- **Ownership/risk:** `freeBats` is the only ownership set; `bats` and `buf` are aliases, so cleanup visits each copied batch exactly once and then zeroes the container. `Reset -> Free`, direct `Free`, and repeated `Free` are idempotent; no double-free path was found.
- **Unhappy path:** the regression now stops with one copied batch already emitted, another still pending, `last=true`, and `i=2`, then calls `Free` without `Reset`. It verifies both cached vectors and every execution-state reference/marker are cleared. This ran 100 consecutive times and under `-race`.
- **Performance:** `Call` is unchanged. Normal `Reset` performs the same O(n) owned-batch cleanup as before; the following `Free` only observes an empty container. No SQL execution hot-path work, allocation, or lock was added.
- **Scope/overfitting:** adjacent empty-`Free` operators were audited; none owns this copied-batch cache. `MergeCTE` already releases its cached batch data and has different reuse semantics, so it was intentionally not changed.

Fresh validation after rebasing main: package tests, 100x targeted regression, `-race`, `go build`, `go vet`, and dependent `pkg/sql/compile` operator conversion test all passed.

gouhongshen (approved):
## Codex automated review

No findings. `freeBats` is the sole owner of duplicated batches; `bats` and `buf` are aliases. The shared cleanup preserves prior Reset behavior, makes Free idempotent, and clears all execution state without changing the hot path. The regression test covers direct and repeated Free. Independent tests were blocked by missing local CGo artifacts, but the author reports package and race validation on this exact head.

Issue #23182, opened 2025-12-01
[Feature Request]: Improve mo-tester Usability


aptend:
### Is there an existing issue for the same feature request?

- [x] I have checked the existing issues.

### Is your feature request related to a problem?

```Markdown

```

### Describe the feature you'd like

1. **Pattern matching only works for SQL errors**: `@pattern` flag fails for result sets because `StmtResult.regularMatch()` only checks `errorMessage` (null for result sets). Pattern stored in `orginalRSText` is ignored.

2. **Serial execution only**: Tests run sequentially in a simple loop, no parallelization support. Cannot leverage multi-core systems.

3. **Poor multi-line output handling**: `ResultParser.convertToRSSet()` uses simple `split()` which breaks when column values contain separator characters. No escape handling.

### Documentation, Adoption, Use Case, Migration Strategy

```Markdown

```

PR #26402, opened 2026-07-29
fix(frontend): preserve view definitions during database clone

### LINKED ISSUE #26070: [Bug]: Database branch/clone corrupts or silently omits views
### Describe the bug

`DATA BRANCH CREATE DATABASE` changes view semantics when the source database name appears anywhere in the view definition outside a qualified object reference.

During database cloning, the implementation rewrites the stored `CREATE VIEW` text with a global string replacement. As a result, string constants, embedded text, and quoted column aliases that happen to contain the source database name are also changed to the destination database name.

### Steps to reproduce

```sql
drop database if exists g4d_replace_src;
drop database if exists g4d_replace_dst;
create database g4d_replace_src;

create view g4d_replace_src.v_text as
select
  'g4d_replace_src' as exact_text,
  'prefix_g4d_replace_src_suffix' as embedded_text,
  1 as `g4d_replace_src`;

select * from g4d_replace_src.v_text;

data branch create database g4d_replace_dst from g4d_replace_src;

select * from g4d_replace_dst.v_text;
show create view g4d_replace_dst.v_text;
```

The same global replacement also changes references to a different database whose name merely starts with the source database name:

```sql
create database g4d_rsrc;
create database g4d_rsrc_aux;
create database g4d_rdst_aux;
create table g4d_rsrc_aux.ext(id int primary key, marker varchar(20));
create table g4d_rdst_aux.ext(id int primary key, marker varchar(20));
insert into g4d_rsrc_aux.ext values (1,'SOURCE_AUX');
insert into g4d_rdst_aux.ext values (1,'TARGET_AUX');

create view g4d_rsrc.v_external as
select id, marker from g4d_rsrc_aux.ext;

data branch create database g4d_rdst from g4d_rsrc;
select * from g4d_rdst.v_external;
```

### Actual behavior

The source view returns:

```text
g4d_replace_src | prefix_g4d_replace_src_suffix | g4d_replace_src=1
```

The cloned view returns:

```text
g4d_replace_dst | prefix_g4d_replace_dst_suffix | g4d_replace_dst=1
```

`SHOW CREATE VIEW` confirms that the constants and alias were modified in the stored definition.

In the external-database case, the source view reads `g4d_rsrc_aux.ext` and returns `SOURCE_AUX`, while the cloned view is rewritten to read `g4d_rdst_aux.ext` and returns `TARGET_AUX`.

### Expected behavior

Database cloning should rewrite only parsed database qualifiers that refer to the cloned source database. Constants, partial text matches, aliases, comments, and unrelated identifiers must retain their original values.

The cloned view should return the same values and expose the same result column names as the source view.

### Reproducibility

Reproduced in 3/3 runs for both the constant/alias case and the external-database-prefix case.

### Code context

`rewriteCloneViewInfos` in `pkg/frontend/clone.go` applies `strings.ReplaceAll(info.createSql, srcDBName, dstDBName)` to the complete stored view definition. This operation has no SQL token or AST awareness, so every textual occurrence is rewritten.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

fixes #26070

Database clone and DATA BRANCH CREATE DATABASE rewrote stored CREATE VIEW SQL with strings.ReplaceAll. That changed string literals and aliases containing the source database name, and could rewrite table-name substrings into nonexistent dependencies.

This PR:
- parses each stored CREATE VIEW with its original SQL mode and rewrites only AST database qualifiers;
- uses the shared remap_db traversal, including qualified column references such as db.table.column;
- keeps literals, aliases, and table-name substrings unchanged;
- preserves the original CREATE VIEW text byte-for-byte when no database qualifier needs remapping, avoiding formatting-only clone regressions;
- adds unit coverage plus branch and regular-clone regression coverage.

Verification:
- `go test ./pkg/frontend -run 'Test_rewriteCloneCreateSQL|TestApplyRemapDb' -count=1`
- `git4data/branch/edge/branch_unhappy_path.sql`: 88/88 passed locally
- `pessimistic_transaction/clone/clone.sql`: both CI-failing SHOW CREATE VIEW assertions pass locally; the only remaining local mismatch is the pre-existing #23182 system_metrics placeholder


XuPeng-SH (approved):
Reviewed exact head dc42c5578688. The AST-based rewrite correctly limits database remapping to qualified identifiers while preserving unchanged CREATE VIEW SQL. Targeted frontend tests (rewriteCloneViewInfos and ApplyRemapDb) pass locally with the deterministic CGo test wrapper.

PR #26403, opened 2026-07-29
fix: respect scheduler CPU limits for RPC workers

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Related to #26299 and #26316.

The Ubuntu/x86 race job for #26316 failed while `pkg/tests/shard` was starting
its embedded cluster. The first test timed out in `waitAnyShardReadyLocked`; a
later nil dereference was secondary because the shared fixture's `sync.Once`
had already completed without publishing a cluster.

The runner had an 8-CPU scheduling limit but exposed 112 host CPUs through
`runtime.NumCPU()`. Each embedded TN therefore created 896 transaction RPC
workers (`8 * 112`). With the existing heavy-package parallelism of 3, several
race-instrumented clusters created thousands of unnecessary workers and
starved HAKeeper startup.

This fixes the resource model rather than changing test scheduling:

- system initialization never raises a scheduler limit already established by
  the Go runtime or `GOMAXPROCS`;
- the default MORPC server-worker count uses `runtime.GOMAXPROCS(0)`, which is
  the capacity available to Go scheduling, instead of host CPU visibility;
- the existing minimum of 100 workers and multiplier of 8 remain unchanged;
- explicit `server-workers` configuration remains authoritative.

There is no package serialization, test-package allowlist, retry, timeout
extension, assertion weakening, or reduction of `HEAVY_RACE_PARALLEL`.
Bare-metal defaults remain unchanged when `GOMAXPROCS == runtime.NumCPU()`.

## Validation

- `go build ./pkg/common/morpc ./pkg/common/system`
- `go vet ./pkg/common/morpc ./pkg/common/system`
- focused race stress, each separately with a 30-second budget:
  - `TestAdjustConfig`: terminal duration below timer resolution, `count=100`
  - `TestDefaultServerWorkers`: terminal duration below timer resolution,
    `count=100`
  - `TestEffectiveGoMaxProcs`: terminal duration below timer resolution,
    `count=100`
- complete owning packages:
  - `mo-cgo-test -race -count=1 ./pkg/common/morpc`
  - `mo-cgo-test -race -count=1 ./pkg/common/system`
- controlled startup with `GOMAXPROCS=8` preserved 8 rather than raising it to
  the 24 CPUs visible on the local host
- concurrently, with `GOMAXPROCS=8` and `-race`:
  - `TestPartitionBasedShardCanBeCreated` in `pkg/tests/shard`
  - `Test_TxnExecutorExec` in `pkg/tests/txnexecutor`
- `mo-cgo-test -race -count=1 ./pkg/txn/rpc`

### DIFF
--- a/pkg/cnservice/server_query_test.go
+++ b/pkg/cnservice/server_query_test.go
@@ -283,7 +283,7 @@ func Test_service_handleGoMaxProcs(t *testing.T) {
 				resp: &query.Response{},
 			},
 			wantErr: nil,
-			want:    &query.Response{GoMaxProcsResponse: query.GoMaxProcsResponse{MaxProcs: int32(goruntime.NumCPU())}},
+			want:    &query.Response{GoMaxProcsResponse: query.GoMaxProcsResponse{MaxProcs: int32(goruntime.GOMAXPROCS(0))}},
 		},
 	}
 	for _, tt := range tests {
--- a/pkg/common/morpc/cfg.go
+++ b/pkg/common/morpc/cfg.go
@@ -15,7 +15,6 @@
 package morpc
 
 import (
-	"math"
 	"runtime"
 	"time"
 
@@ -24,6 +23,11 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/common/moerr"
 	"github.com/matrixorigin/matrixone/pkg/util/toml"
 	"go.uber.org/zap"
+)
+
+const (
+	defaultMinServerWorkers    = 100
+	defaultServerWorkersPerCPU = 8
 )
 
 var (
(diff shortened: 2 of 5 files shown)


XuPeng-SH:
Deep self-review updated on exact head `dcec4ac8afd53cb840cf5b785bde661bf6b1bc45`.

The production change is appropriate: default MORPC worker sizing now uses the effective runtime CPU budget (`GOMAXPROCS`) rather than host `NumCPU`, while preserving explicit config, the minimum of 100 workers, and the existing 8x multiplier. System initialization no longer raises an already-lower scheduler/container limit. This changes startup sizing only; there is no request hot-path cost. In the motivating 8-CPU limit on a 112-CPU host, the default falls from 896 to the existing minimum 100 workers.

Coverage CI exposed an adjacent stale test that equated a `GOMAXPROCS(0)` query with `NumCPU`; CI correctly observed 7 vs 8. The assertion now checks current `GOMAXPROCS`, preserving the API intent without environment-specific values. morpc/system package tests and race tests passed; the cnservice regression passed 20x and 5x under `-race`. CI restarted on the new head.

Issue #26338, opened 2026-07-29
[Subtask]: Move child-side foreign-key UPDATE validation to MULTI_UPDATE


XuPeng-SH:
## Parent issue

#26336

## Problem

The modern UPDATE resolver currently rejects a target as soon as its table definition contains any `Fkeys` or `RefChildTbls`. Because `ResolveUpdateTables` calls `ResolveTables(..., respectFKCheck=false)`, this routing also ignores the session's `foreign_key_checks` value.

Planner selection should instead depend on the columns actually updated:

- updating no FK child column needs no child-to-parent lookup;
- updating a child FK column needs validation of the new key;
- updating no referenced parent-key column needs no parent-side action;
- updating a referenced parent-key column belongs to #26339.

A table can be both child and parent, so these decisions must be made per constraint, not per table.

## Required work

- Keep unrelated-column updates on FK-bearing tables in the modern `MULTI_UPDATE` path.
- Detect which child FK constraints are affected by the final updated row image.
- Validate changed child keys against the corresponding parent key.
- Preserve single/composite FK NULL semantics, including partial-NULL keys.
- Apply the defined `foreign_key_checks=0/1` semantics consistently to routing and validation; do not preserve the current metadata-only routing accident.
- Reuse modern base-table, regular-index, lock, retry, and rollback behavior.
- Leave referenced-parent-key actions to #26339.

## Acceptance criteria

- [ ] An unrelated-column UPDATE produces `MULTI_UPDATE` even when the target has child FKs, referenced children, or both.
- [ ] Changing a child FK to an existing parent succeeds; changing it to a missing parent fails before mutation.
- [ ] Unchanged child FK columns do not cause redundant parent probes.
- [ ] Single and composite FKs cover all-non-NULL, all-NULL, partial-NULL, nullable, and NOT NULL cases.
- [ ] `foreign_key_checks=0` and `=1` have explicit plan and behavior tests matching the intended SQL contract.
- [ ] The real-server shape from #26334 (nullable UNIQUE plus unrelated FK metadata) uses `MULTI_UPDATE`, contains no `PreInsert UniqueKey`, and leaves correct hidden-index contents.
- [ ] #26334 remains independently fixed so the legacy path is safe during rollout.
- [ ] Optimistic/pessimistic conflict, retry, cancellation, and rollback cases remain atomic.
- [ ] INSERT/ODKU FK behavior is unchanged.

## Test matrix

[...]

Issue #26339, opened 2026-07-29
[Subtask]: Move parent-side FK referential actions to MULTI_UPDATE


XuPeng-SH:
## Parent issue

#26336

## Problem

When an UPDATE changes a referenced parent key, the legacy planner recursively builds child checks/updates through `buildUpdatePlans` / `buildDeletePlans`. This is the largest semantic dependency preventing removal of `buildTableUpdate`.

The modern planner needs an explicit referential-action graph. Merely having `RefChildTbls` is not enough to invoke it: parent-side work is required only for constraints whose referenced columns can change in the final row image.

## Scope

- RESTRICT and NO ACTION.
- CASCADE.
- SET NULL.
- Current SET DEFAULT disposition: the legacy code groups it with NO ACTION/RESTRICT; preserve that compatibility or reject it explicitly rather than silently inventing mutation semantics.
- Self-referencing constraints.
- Multiple constraints, multiple child tables, and multi-level actions.
- Tables that are both parent and child.

## Required work

- Discover affected constraints from updated referenced columns.
- Build a finite action DAG with explicit cycle/duplicate-row handling rather than uncontrolled recursive expansion.
- Use final parent keys and correct old/new child row images.
- Fold all base-table and regular hidden-index writes into one atomic statement plan.
- Define deterministic table/row lock ordering and retry behavior.
- Honor the defined `foreign_key_checks` semantics consistently with #26338.

## Acceptance criteria

- [ ] Updating a non-referenced parent column stays on the modern path without child probes/actions.
- [ ] RESTRICT and NO ACTION reject a referenced-key change when matching children remain.
- [ ] CASCADE updates every affected child row and regular index atomically.
- [ ] SET NULL validates child nullability before mutation and maintains indexes atomically.
- [ ] SET DEFAULT has an explicit compatibility test and documented disposition.
- [ ] Composite keys preserve column order and NULL semantics.
- [ ] Self-reference, cycles, and deep graphs terminate deterministically with a defined error where the graph is unsupported.
- [ ] Diamond/multiple FK paths do not double-update a row or corrupt affected-row accounting.
- [ ] Lock acquisition order is documented and covered by concurrent conflict/deadlock/retry tests.
- [ ] Any failure or cancellation rolls back parent, child, and hidden-index changes together.
- [ ] Supported cases produce modern plans with no `buildTableUpdate` route.

## Test matrix

[...]

PR #26407, opened 2026-07-29
fix(plan): validate child keys in modern update

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26338

The modern UPDATE planner previously rejected every FK-bearing target based only
on table metadata, even when the updated columns were unrelated to any foreign
key. This forced those statements onto the legacy planner and did not apply
`foreign_key_checks` consistently during route selection.

This PR:

- defers foreign-key route selection until the final updated columns are known;
- keeps unrelated updates on FK-bearing tables on the modern `MULTI_UPDATE`
  path;
- validates affected child keys with per-constraint parent MARK joins;
- preserves single and composite MATCH SIMPLE NULL semantics;
- allows unchanged orphan keys while rejecting newly invalid keys;
- routes referenced-parent-key updates to the legacy planner for the parent-side
  work tracked by #26339;
- applies `foreign_key_checks=0/1` consistently to routing and validation;
- adds planner and real-server regression coverage for regular indexes,
  generated columns, UPDATE FROM deduplication, composite keys, and
  self-referencing foreign keys.

Validation:

- seven focused planner tests passed with `-race -count=100`;
- the complete `pkg/sql/plan` package passed with `-race -count=1`;
- `go build`, `go vet`, `make build`, formatting, and diff checks passed;
- `update_modern_fk.sql` passed 46/46 cases;
- the affected self-referencing FK suites passed 60/60, 101/101, and 81/81
  cases.


XuPeng-SH (changes requested):
Requesting changes for two issues found at head `c32538b011875a25514a9b9c647ef9a465b93614`:

1. Child FK validation can run before AUTO_INCREMENT materializes the final updated value, allowing an unchecked generated value to be written.
2. Self-referencing FK UPDATE failures regress from the typed 1452/SQLSTATE 23000 error to a generic internal error.

The intended affected-column routing, MATCH SIMPLE NULL behavior, unchanged-orphan handling, and parent-key legacy routing otherwise look sound. Full `pkg/sql/plan` tests and focused race tests passed; the two cases above need dedicated regression coverage.

pkg/sql/plan/bind_update.go
```diff
@@ -418,6 +421,18 @@ func (builder *QueryBuilder) bindUpdate(stmt *tree.Update, bindCtx *BindContext)
 		}
 	}
 
+	lastNodeID, selectNodeTag, selectNode, err = builder.appendUpdateForeignKeyChecks(
```
XuPeng-SH, line 452: The FK check here runs before the `PRE_INSERT` node added below for updated AUTO_INCREMENT columns. If an AUTO_INCREMENT child-FK assignment evaluates to NULL at runtime (for example via an expression or prepared parameter), the MARK check treats NULL as valid under MATCH SIMPLE, then `PRE_INSERT` materializes a new non-NULL auto-increment value that is never checked against the parent table. This can commit an orphan key. Please validate the post-`PRE_INSERT` final row image, or route this schema/update combination to the legacy planner, and add a runtime-NULL regression case.

pkg/sql/plan/bind_update_fk.go
```diff
@@ -0,0 +1,309 @@
...
+		affectedFks := affectedUpdateChildFks(tableDef, alias, newColName2Idx)
+		if len(affectedFks) == 0 {
+			continue
+		}
+
+		fkTableDef := *tableDef
+		fkTableDef.Fkeys = make([]*plan.ForeignKeyDef, len(affectedFks))
+		for j, fk := range affectedFks {
+			fkTableDef.Fkeys[j] = DeepCopyFkey(fk)
+			if fkTableDef.Fkeys[j].ForeignTbl == 0 {
```
XuPeng-SH, line 62: Rewriting the self-reference sentinel (`ForeignTbl == 0`) to the current table ID sends self-referencing UPDATE checks through generic `assert()`. That function returns `moerr.NewInternalError`, whereas the previous DetectSql path returned `ErrFKNoReferencedRow2` (MySQL 1452 / SQLSTATE 23000). The changed `.result` files confirm this user-visible error-code regression (`internal error:`). Please preserve the typed FK error, either with a typed data-flow assertion or a reliable execution-layer translation.

heni02 (approved):

XuPeng-SH (changes requested):
[P1] Mark modern FK-sensitive UPDATE plans for rebuild/cache invalidation across `foreign_key_checks` changes.

`appendUpdateForeignKeyChecks` makes the plan shape depend on the session value at lines 32-38: with checks off it returns before adding the child MARK joins/asserts, while with checks on it adds them (and parent-key routing depends on the same value). Neither branch sets `Query.HasForeignKeyAction`.

The prepared-statement contract in `pkg/frontend/computation_wrapper.go` only rebuilds such a plan on EXECUTE when `HasForeignKeyAction` is true, and both prepared compile caching and generic plan caching use that flag as well. Consequently, PREPARE with `foreign_key_checks=0`, then switch to 1 and EXECUTE, can reuse the cached no-check plan and write an invalid child key. The inverse transition can retain enforcement after checks are disabled. The new `ordinary child update keeps prepare cacheable` test currently codifies the unsafe false flag.

Please mark the query FK-sensitive before the disabled early return whenever the UPDATE can depend on FK metadata, and add prepared-statement regressions for both 0→1 and 1→0 transitions. Please also make the AUTO_INCREMENT child-key routing decision observe the disabled setting if consistent `foreign_key_checks=0` routing is intended.

Deep re-review of exact head `9f2fa1182cee2392c7f279cbe02b9b1e7e7adf21`: the prior AUTO_INCREMENT final-value validation and typed 1452/23000 error blockers are resolved. Focused planner tests passed under race (`-count=5`); `go build`, `go vet`, and `git diff --check` also passed. CI is still running.

XuPeng-SH (approved):
Re-reviewed exact head `25d7f042acb4a66e5d664b12e524633880ed7af5`. The prior blockers are resolved: FK-sensitive UPDATE plans are marked independently of the current `foreign_key_checks` value, both prepared/generic cache paths invalidate correctly across 0↔1, and disabled checks retain the intended modern AUTO_INCREMENT route. Child-key row-image validation, MATCH SIMPLE NULL handling, unchanged-orphan behavior, self-reference, parent-key fallback, and typed 1452/23000 errors remain sound. Focused planner/frontend tests passed under race; the full `pkg/sql/plan` package and exact-head CI pass.

PR #26410, opened 2026-07-29
perf(tae): restore ObjectList hot paths after grouped ordering

XuPeng-SH:
## What changed

- keep the six-group ObjectList B-tree for group/range traversal and add a C-entry-only copy-on-write B-tree ordered globally by `CreatedAt` and object name
- publish both indexes through one atomic snapshot so readers cannot observe mismatched generations
- restore visible-object scans to one reverse iterator instead of four group iterators plus a four-way merge
- build dynamic seek pivots from only comparator-relevant fields and recycle them after synchronous B-tree seeks
- cover create, drop, rollback, replay timestamp updates, create timestamp updates, direct D-entry replay, and GC deletion in both indexes

This is a follow-up to #26395 and the grouped ObjectList ordering introduced by #26225.

## Why

The grouped ordering fixed TN ObjectList range traversal, but it also moved two costs onto transaction hot paths:

1. every visible-object scan initialized four B-tree iterators and merged four group heads for every returned object;
2. incremental PK dedup performed five dynamically timestamped group seeks, and every seek constructed a full `ObjectEntry`/`ObjectStats`/printable object name.

In a 30-second local CPU profile before this change, `makeObjectListKey` accounted for 4.41s cumulative CPU and `SeekObjectListGroup` for 3.32s in the `PrePrepareDedup -> foreachIncrementalObject` path. After compact pooled pivots, `makeObjectListKey` disappeared from the profile and `SeekObjectListGroup` fell to about 0.15s cumulative.

The original performance regression is visible in [mo-auto-test run 30467293159](https://github.com/matrixorigin/mo-auto-test/actions/runs/30467293159/job/90630158871): the parent was reported at 2,360 TPS while `29f4ada` was reported at 1,262 TPS.

## Microbenchmarks

Apple M4, six runs per case:

| Benchmark | #26395 baseline | This PR |
| --- | ---: | ---: |
| VisibleCommitted | ~200-202 ns/op, 448 B/op, 4 allocs/op | 43.4-43.9 ns/op, 112 B/op, 1 alloc/op |
| Five dynamic group seeks | ~1,080-1,131 ns/op, 4,704 B/op, 33 allocs/op | 242.8-245.8 ns/op, 112 B/op, 1 alloc/op |

A clean end-to-end `oltp_write_only.lua` TPS claim is intentionally omitted. The current local host had unrelated load, which caused large time drift between exact-commit runs; those samples were stopped and treated as invalid. The microbenchmark allocation counts and CPU-profile call-path removal are deterministic, while the macro A/B still needs an idle-host rerun.

## Correctness and risk

- the primary six-group index and all existing group seek semantics remain unchanged
- the secondary index stores pointers to C entries, not duplicate object data; it adds one B-tree index entry per C object and extra COW work on catalog mutations
- object mutations update both trees while holding the existing ObjectList write lock, then publish one paired atomic snapshot
- visible iteration order remains exactly descending `CreatedAt`, then descending object name
- B-tree seek is synchronous and does not retain or mutate its pivot; every pooled dynamic pivot is returned with `defer`
- no on-disk, wire, configuration, or catalog metadata format changes

## Validation

After rebasing onto `0ebba1c0c9`:

- focused behavioral race tests: 13 exact tests, each `-race -count=100` (`T=0s` at test-event resolution, 30s adaptive budget, capped at 100)
- allocation regression tests: `-count=100`
- full `./pkg/vm/engine/tae/catalog`: normal and `-race -count=1`
- full `./pkg/vm/engine/tae/txn/txnimpl`: normal and `-race -count=1`
- `go vet` for both owning packages
- `make build-with-prebuilt-native`
- `git diff --check` and `gofmt -d`


aptend (approved):
Reviewed the full change and independently validated the hot-path improvement against the parent commit. The latest head is a merge from main; the PR patch-id is unchanged from the reviewed 92cef85f revision. The paired atomic snapshot, ObjectList mutation lifecycle, seek-key ownership, visibility ordering, rollback/replay/GC paths, focused race stress, full owning-package tests, and allocation behavior all look correct. No blocking issues found.

PR #26412, opened 2026-07-29
fix join spill ownership and lifecycle

### LINKED ISSUE #26408: [Bug]: parallel broadcast dedupjoin/hashjoin spill consumers lose bucket ownership or wait forever

## Branch and commit

- Discovery branch: `fix/issue-26375-dedupjoin-cancellation`
- Discovery commit: `6d90f3ec88b744f26802329867c63e7f32536634`
- The ownership behavior is also present in PR #26393's base `8694a8d26dd8ef6e05ab0bdc133e4611c7430e1e`; this is not introduced by that PR.

## Code location

- [`pkg/sql/colexec/dedupjoin/join.go:307-348`](https://github.com/matrixorigin/matrixone/blob/6d90f3ec88b744f26802329867c63e7f32536634/pkg/sql/colexec/dedupjoin/join.go#L307-L348), symbol: `(*DedupJoin).build`
- [`pkg/vm/message/joinMapMsg.go:378-383`](https://github.com/matrixorigin/matrixone/blob/6d90f3ec88b744f26802329867c63e7f32536634/pkg/vm/message/joinMapMsg.go#L378-L383), symbol: `(*JoinMap).TakeSpillBuildFiles`
- [`pkg/sql/compile/scope.go:958-966`](https://github.com/matrixorigin/matrixone/blob/6d90f3ec88b744f26802329867c63e7f32536634/pkg/sql/compile/scope.go#L958-L966), symbol: `newParallelScope`
- [`pkg/sql/compile/operator.go:2142`](https://github.com/matrixorigin/matrixone/blob/6d90f3ec88b744f26802329867c63e7f32536634/pkg/sql/compile/operator.go#L2142), shared JoinMap ref count

- Hardware parameters: code-path review; no hardware dependency identified
- OS type: Darwin/arm64 review environment
- Others: discovered while reviewing PR #26393

## Actual Behavior

A non-shuffle parallel dedupjoin creates `Mcpu` operator copies, and HashBuild publishes one broadcast `JoinMap` pointer with ref count `Mcpu`. When that map is spilled, every dedupjoin clone calls the destructive, unsynchronized `TakeSpillBuildFiles()` on the same object.

Only one consumer is expected to receive the spill-file slice (and concurrent callers can race on the slice field). Other consumers initialize a `SpillEngine` with zero build buckets but still consume their own probe input. `ScatterProbeTable` creates `probeFiles` with `len(e.buckets)`, so an empty-engine consumer retains none of its probe partition. Only the file-owning worker can have later spill rounds, while parallel finalization expects statuses from all workers per round.

Depending on which clone takes the files, the observable result can be incomplete dedup output, a later-round wait for participants that already exited, or termination that drops the file owner's remaining buckets.

## Expected Behavior

Parallel spill execution must preserve every probe partition and have a consistent participant set for every spill round. Spill files/budget must have one explicit owner: either all probe partitions are routed through one spill consumer, or each worker receives an owned bucket representation that matches its round participation.

## Steps to Reproduce

1. Execute a non-shuffle dedupjoin plan with `Mcpu > 1`.
2. Set/drive the hash-build spill threshold low enough that HashBuild publishes a spilled broadcast `JoinMap`.
3. Let the parallel dedupjoin clones concurrently enter `(*DedupJoin).build`.
4. Observe that they all receive the same JoinMap pointer and call `TakeSpillBuildFiles`; only one engine owns buckets, while the others have an empty bucket queue and discard their probe partitions.
5. Continue into a later spill bucket and observe incomplete output or a finalize wait/early stop caused by the missing workers.

### Confirmed code facts

- `newParallelScope` creates exactly `Mcpu` dedupjoin clones.
- `constructHashBuild` sets `JoinMapRefCnt = mcpu`, and broadcast receivers return the same `JoinMap` pointer.
- `TakeSpillBuildFiles` moves the slice out by reading the field and assigning nil without synchronization or cloning.
- Every spilled dedupjoin clone executes this Take and initializes its own SpillEngine.
- `ScatterProbeTable` retains probe files only for existing `e.buckets`.
- No parallel + actual spilled JoinMap integration test exists in `pkg/sql/colexec/dedupjoin`; the current round test manually gives every worker the same synthetic bucket sequence.

### Inference

Under an actual parallel spilled build, consumer bucket sequences are not isomorphic. Probe rows processed by zero-bucket engines cannot contribute to output, and a file-owning merger can wait for workers that already exhausted their empty engines.

## Impact

Parallel INSERT/REPLACE/UPDATE paths using dedupjoin under hash-build spill can return incomplete results, silently omit work, or fail to terminate until cancellation.

## Unverified assumptions

- A complete SQL-level reproduction and the minimum row count/threshold have not yet been constructed.
- The exact owner winner is scheduler-dependent; validation should force ownership order deterministically.

## Additional information

- Duplicate searches: `"TakeSpillBuildFiles" dedupjoin`; `parallel dedup join spill wrong result hang` across open and closed issues.
- Reproduction status: code-path proof; no end-to-end SQL reproduction yet.
- Suggested validation: add a deterministic 2- or 3-worker integration test using one actual spilled broadcast JoinMap, distinct probe partitions, and at least two buckets; assert full output, bounded completion, and race-clean single ownership.

## Follow-up audit: HashJoin and RightDedupJoin

Follow-up audit commit: `0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c`.

### HashJoin: confirmed affected

The same ownership bug is reachable through normal non-shuffle broadcast HashJoin plans:

- [`pkg/sql/compile/operator.go:214-241`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/compile/operator.go#L214-L241) creates `Mcpu` HashJoin copies for the broadcast probe side and designates copy 0 as merger.
- [`pkg/sql/compile/operator.go:2072-2151`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/compile/operator.go#L2072-L2151) publishes one broadcast `JoinMap` with `JoinMapRefCnt = mcpu`; all copies therefore receive the same pointer.
- [`pkg/sql/colexec/hashjoin/join.go:328-390`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/colexec/hashjoin/join.go#L328-L390) makes every spilled HashJoin copy call `TakeSpillBuildFiles()` and `TakeSpillBudget()`, create a private `SpillEngine`, and scatter its own probe partition.
- [`pkg/sql/colexec/spillutil/join_spill.go:1575-1662`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/colexec/spillutil/join_spill.go#L1575-L1662) allocates probe files from `len(e.buckets)`. A non-owner engine has zero buckets, consumes all child probe batches, and retains none of them.

For inner/left/semi/anti/single/mark joins, every non-owner probe partition is silently discarded, so only the worker that won the destructive Take can contribute rows.

For right/full/right-semi/right-anti/right-single joins, the shared bitmap finalization at [`pkg/sql/colexec/hashjoin/join.go:801-847`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/colexec/hashjoin/join.go#L801-L847) adds two failure modes:

- If the spill-file owner is the merger, it can wait for per-round bitmaps from zero-bucket workers that already exited; cancellation is the only bounded escape.
- If the owner is not the merger, the merger can finish before the owner completes all buckets, dropping remaining buckets and unmatched-build output; a late bitmap may remain buffered after the merger exits.

`TakeSpillBuildFiles()` and `TakeSpillBudget()` are also unsynchronized. A racy interleaving can let multiple workers observe the same spill files or budget. Shared `*os.File` offsets and independent engine cleanup then make duplicated ownership invalid even if more than one caller receives a non-empty value.

### RightDedupJoin: currently protected by topology, but fragile

RightDedupJoin contains the same destructive Take pattern in [`pkg/sql/colexec/rightdedupjoin/join.go:192-262`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/colexec/rightdedupjoin/join.go#L192-L262), but the audited normal SQL paths currently avoid shared ownership:

- The broadcast planner forces RightDedupJoin probe scopes to `Mcpu = 1` at [`pkg/sql/compile/compile.go:4324-4334`](https://github.com/matrixorigin/matrixone/blob/0ebba1c0c96bbcf0ff4091236e10df7b97e24a0c/pkg/sql/compile/compile.go#L4324-L4334), so the broadcast HashBuild map has ref count 1.
- In shuffle mode, each operator copy has a distinct `ShuffleIdx`, and each shuffle HashBuild publishes a map with ref count 1.

Therefore RightDedupJoin is not confirmed as a currently reachable production bug. However, a manually constructed non-shuffle parallel topology, or a future removal of the `Mcpu = 1` restriction, would expose the same loss. This topology invariant should be covered by regression tests.

### Additional test gaps

- HashJoin's actual spill/re-spill integration coverage uses `IsShuffle = true` with one consumer; it does not exercise a shared broadcast `JoinMap`.
- The zero-bucket HashJoin test only asserts that an empty engine returns no batch; it does not assert probe-row conservation.
- RightDedupJoin spill coverage also uses shuffle mode and ref count 1.
- There is no non-shuffle broadcast + actual spilled `JoinMap` + `Mcpu > 1` integration test for HashJoin.

Suggested deterministic coverage:

1. Use 2 or 3 HashJoin workers sharing one actual spilled broadcast `JoinMap`, with distinct probe partitions and at least two buckets.
2. Force both ownership orders: merger owns the files, and a non-merger owns them.
3. Cover an inner-family join and a right/full-family join; assert complete rows, bounded completion, and race-clean ownership.
4. Add compile-topology tests that pin broadcast RightDedupJoin to `Mcpu/refcount = 1` and shuffle RightDedupJoin to distinct refcount-1 maps.

A mutex around `TakeSpillBuildFiles()` alone is insufficient: it would remove the data race but preserve the data-loss topology. The fix needs either one spill coordinator that receives all probe partitions, or explicit independent/partitioned bucket ownership with probe conservation and per-bucket finalization coordination.

XuPeng-SH:
## What changed

- Replace split spill files / FD tokens / budget references on `JoinMap` with one atomic move-only `SpillBuildPayload`, and migrate hashjoin, dedupjoin, and rightdedupjoin consumers.
- Reject invalid or shared spill publication and make `Take` / `FreeMemory` cleanup race-safe and exactly-once.
- Scope bitmap channels, DedupJoin mailboxes, generated parallel scopes, and borrowed spill file services to one execution generation.
- Propagate previously ignored join output failures; restore Dedup UPDATE vector ownership on error; put RightDedup probe-only hashmaps under the shared build budget.
- Add cancellation checks at hash-build, spill record, physical I/O, rebuild, iterator, probe-read, and final bucket-flush boundaries.
- Cache SpillEngine file-service resolution, skip semantically unused build-only buckets, preserve explicit remote spill thresholds, and bound spill FD admission against runtime `RLIMIT_NOFILE`.
- Make reduced spill fanout correct: validate 1..32 power-of-two fanout before consuming probe input and partition probes using the build payload's actual fanout.
- Fix DedupJoin multi-bucket finalize state so a worker reset between rounds publishes an abort for the next round instead of leaving the merger blocked forever.
- Treat a 1..15 byte trailing frame header as corruption, and release an empty decoded build's builder before handing its probe file to the caller.
- Validate compiler-real DedupJoin UPDATE map shapes, legacy payload budget identity, and pre-cancel FD ownership with end-to-end operator and live-resource tests.

## Why

Issue #26408 exposed a split-ownership protocol shared by hashjoin, dedupjoin, and rightdedupjoin. The normal broadcast topology currently cannot produce that payload because non-shuffle HashBuild never spills, so the reported shared-broadcast failure is latent rather than normally reachable. The production-visible design gap is broader: broadcast joins have no spill fallback, while the existing spill path also had ownership, generation reuse, error propagation, cancellation, memory-budget, file-descriptor, file-corruption, and multi-round lifecycle weaknesses.

This PR closes the unsafe cross-join protocol and the concrete reachable unhappy paths without pretending to implement a broadcast spill scheduler or redesign the on-disk format in the same patch.

## First-principles closure

- **Q1 — cleanup:** every file, FD token, disk token, memory reservation, decoded batch, capture vector, generated scope, and result vector has one terminal owner on success, error, cancellation, Reset, and publication loss.
- **Q2 — termination:** waits observe context or an explicit terminal mailbox/channel status; physical I/O is checked immediately before entry; DedupJoin publishes one status per worker per finalize round.
- **Q3 — boundedness:** memory/disk/FD admission is finite, recursive passes and queue depth are bounded, invalid fanout fails before child consumption, and cancellation polls remain at batch/record/I/O boundaries rather than per row.

## Real and unhappy-path validation

- A canceled initial spill writes to a real physical file and proves file size, spill bytes, and spill rows all remain zero.
- A real one-bucket build payload scatters 1,000 probe rows and proves exact row conservation; 0/3/33 buckets fail before reading child input.
- A real two-record spill stream cancels at the second frame boundary and proves the second payload is never decoded and all batch/token memory is released.
- A real valid record plus a partial trailing header returns `io.ErrUnexpectedEOF`, clears the decoded destination and reservation, and returns budget usage to zero.
- A two-round DedupJoin mailbox reproduces normal worker early-stop between spill buckets and proves Reset publishes the nil-error abort required to terminate the merger.
- An empty physical build file uses a constant-key executor that owns mpool memory at Prepare time; both keep-probe and skip-probe branches return `CurrNB` to baseline and zero memory/disk/FD ledgers.
- A real `HashBuild -> JoinMap -> DedupJoin UPDATE` pipeline covers compact and duplicate `GroupSels` maps, matched and unmatched-finalize paths, emitted value/NULL semantics, vector-owner restoration, and terminal mpool=0.
- Legacy payload transfer proves the consumer receives the exact process budget generation and that the moved FD is released once; pre-cancel reader/writer/scatter/probe/re-spill tests use live tracked FDs to prove retain-on-cancel and release-on-cleanup.
- Real capped-budget hashjoin initial spill plus re-spill passes 10/10 with row conservation; local guardrail time was 0.57–0.80 s total, depending on concurrent load.
- A checked-in 4,096-row embedded-cluster SQL regression generates a real optimizer range-shuffle plan on one runtime bucket, executes resident and constrained-spill variants through the MySQL protocol, requires positive `SpillRows` / `SpillSize` plus depth-1 spill and depth-2 re-spill metrics, and proves the skewed equal-key batch neither panics nor loses rows.

## Spill regression test plan

BVT remains the minimum SQL correctness and plan smoke layer; this PR does not add a large spill dataset to BVT. Spill correctness is covered by a parameterized regression job in which data size is an input, not a correctness oracle.

### Harness and scale contract

- Generate deterministic physical build/probe tables from a seed and a configurable scale `N`. CI, nightly, and soak jobs may use different `N`, DOP, and CN counts; expected results are derived from the generated keys and are never hard-coded to a particular row count.
- The checked-in presubmit SQL case defaults to `N=4096` and accepts only `1024 <= N <= 65536`. The upper bound protects CI hosts from accidental OOM; larger nightly/soak scales belong in a dedicated external harness rather than BVT or this presubmit test.
- Run each semantic case twice against the same data snapshot: once with a resident-memory configuration as the reference, and once with a constrained spill configuration.
- Require `EXPLAIN` to contain the intended join and shuffle topology. A spill run passes only when `EXPLAIN ANALYZE` reports `SpillRows > 0` and `SpillSize > 0`; a query that happens not to spill is unexercised, not passed.
- Trigger spill primarily by constraining the admitted working set relative to the generated data. Increasing `N` is optional, so the regression can run at any practical data scale.
- For re-spill cases, require an explicit spill-depth/round metric greater than one when available; until then, require `SpillRows` to exceed the computed single-pass build-plus-probe row total. The comparison uses generated row counts, not a fixed constant.

### Required coverage

1. **HashJoin semantics:** INNER, LEFT, RIGHT, FULL, SEMI, ANTI, and MARK joins; uniform and highly skewed keys; matched, unmatched, duplicate, and NULL keys. Cover signed and unsigned integers, strings, expression keys, and supported composite keys.
2. **Range-shuffle regression:** generate a real join plan containing the Shuffle operator and execute highly skewed integer batches. With one runtime bucket, verify constant/equal/ordered batches do not panic or lose rows and all rows reach bucket 0. With multiple buckets and no effective sampled boundary, verify the min/max fallback distributes every row. Exercise signed, unsigned, NULL, expression, and zonemap consumers so an empty boundary is never treated as an enabled sampled range.
3. **Initial spill and recursive re-spill:** exercise build spill, probe scatter, bucket rebuild, multiple spill rounds, and the maximum-depth path. Verify successful cases against the resident reference and verify the depth limit returns a controlled error with complete cleanup.
4. **DML and dedup operators:** run `INSERT ... ON DUPLICATE KEY UPDATE` for primary-key-only and multiple-UNIQUE targets, including mixed matched/new rows and UPDATE expressions that reference nullable or absent source columns. Cover `INSERT IGNORE`, REPLACE, and RightDedup where their execution topology supports spill. Compare final table contents, affected rows, PK/UK cardinalities, NULL results, and duplicate rejection with the resident run.
5. **Lifecycle and unhappy paths:** cancel before spill-file creation, during scatter/write/read/rebuild, and between buckets; repeat prepared executions and operator reset/reuse; inject truncated/corrupt frames and constrained disk/FD budgets. Assert bounded completion, no partial output, no double cleanup, zero live spill files, and memory/disk/FD ledgers returning to baseline.
6. **Topology matrix:** cover one and multiple shuffle buckets, reduced fanout, one CN, and multi-CN where the test environment supports it. Broadcast topology remains an explicit capability test: until it has a scheduler path for spill, it must stay resident or fail fast rather than silently claiming spill coverage.

### Correctness oracles and execution tiers

- The resident and spill runs must have identical row counts, stable-key checksums, grouped multiplicities, NULL/unmatched-row counts, and DML final tables. These checks detect loss and duplication without depending on result order.
- The SQL-level range-shuffle case is the final regression for the empty-boundary panic; focused signed/unsigned/zonemap tests enforce the shared boundary contract below it.
- Presubmit runs a small deterministic scale and the focused contract tests. Nightly varies scale, seed, skew, DOP, and topology. Soak adds repeated execution, cancellation, and resource-pressure cases.
- Exit criteria are: the intended plan was generated, spill/re-spill was observed rather than inferred, output matches the resident oracle, and all resources are reclaimed without panic or hang.
## Test matrix

- Full tests and full `-race` runs pass for `vm/message`, `vm/process`, `hashbuild`, `spillutil`, `hashjoin`, `dedupjoin`, and `rightdedupjoin`.
- `TestIssue26408JoinSpillSQLRegression` passes at the default 4,096-row scale, passes two consecutive focused `-race` runs, and its full owning `pkg/tests/issues` package passes `-race`; the test-only CGo compile and `go vet -mod=readonly` checks are clean.
- Fourteen representative changed contracts were enumerated with `go test -list`, measured once under `-race -json`, then repeated with `N=clamp(floor(30/T),1,100)`: the RLIMIT test ran 26 times from a 1.14 s sample; the other 13 ran 100 times each.
- Affected compile topology, generation, mailbox/channel, and remote-threshold tests pass together under `-race`.
- Full compile `-race` has exactly three Parquet empty-file fanout failures. The same three tests fail with the same assertions on a clean latest `origin/main` worktree at `2e68774061`; they are not introduced by this PR.
- The official current-head CI coverage merge passes at **291/376 changed blocks = 77.39%** (`approved=true`, 75% threshold); the local affected-package-only profile was 285/375 = 76.00% before CI added cross-package UT/BVT profiles.
- `go vet` and `go build` pass for all eight affected packages. Final CGo-wrapper compile checks, `make err-check`, and `git diff --check` are clean.

## Performance evidence

For 8,192 build rows / 9,216 probe rows with initial spill and re-spill:

- Shared file-service resolution reduces the hot path from roughly one resolution per create (~2,079) to two per engine.
- The two-pass bucket classifier is about 8.14 µs per 8,192 rows with zero allocations versus about 105.5 µs for the legacy bucket-by-bucket row scan (~12.9x).
- End-to-end remains roughly 57–80 ms/run under local load and about 85.7 MB allocated/run. File creation is still the dominant unresolved cost: about 2.1k creates/engine, with create sampled at roughly 38% CPU and scatter append/copy accounting for about 65.8 MB.
- New cancellation checks are per batch/record/I/O boundary; fanout validation is O(1); no new admission or polling was added to row-at-a-time hash/probe loops.

## Risks and follow-ups

- Broadcast HashBuild still has no shared spill scheduler/rebuild task and remains fail-fast for multi-consumer spill.
- Expression executors can retain a high-capacity result vector after a later small batch replaces/releases the corresponding budget charge; executor and reservation lifetime need one owner.
- Anonymous create-and-unlink partition files force queued buckets to retain open FDs. Named/reopenable partition handles, active-window FD accounting, adaptive fanout, and skew-aware terminal fallback require a separate design.
- Spill frames have structural checks but no version/checksum; already-started `os.File.Read`, `os.File.Write`, and unmarshal calls cannot be interrupted by context.
- Resident and spill equality still need one typed-key canonicalization layer; known risk cases include DOUBLE `+0/-0`, NaN, JSON numeric forms, scaled FLOAT, and long VECF32 keys.
- `join_spill_mem` still overloads values `<=100000` as rows and larger values as bytes; explicit versioned units and migration are needed.
- Removing the old exported split spill fields/APIs is a source-compatibility change; all in-repository callers are migrated.

Closes #26408.


XuPeng-SH:
Final coverage + self-review closure (`683196b64a`):

- Added a real `HashBuild -> JoinMap -> DedupJoin UPDATE` integration fixture using the compiler-reachable `HashOnPK=false` configuration. Data drives both compact `HashOnUnique` and duplicate `GroupSels` representations. It covers matched unique, matched duplicate, and unmatched duplicate-finalize folding; asserts output values/NULL semantics and terminal mpool=0.
- Added legacy spill-payload transfer validation: the consumer receives the exact `Process` budget generation, the moved FD is closed once, and `JoinMap.Free` cannot release it again.
- Added deterministic pre-cancel boundary validation across read, physical write, scatter, probe read, and re-spill. Real tracked FDs/release counters prove cancellation returns before I/O or ownership transfer; explicit cleanup then releases each resource exactly once.
- Ran the repository's current CI `parse_coverage.py` against the complete PR diff and local affected-package profiles: **285/375 changed blocks = 76.00%**, above the strict 75% gate. This is a conservative affected-package result; CI also merges cross-package UT/BVT profiles.
- Final gates: full `dedupjoin` + `spillutil` tests pass; full race passes; focused lifecycle tests pass under race `-count=25` (an earlier equivalent run used `-count=50`); compile-only CGo-wrapper checks, `make err-check`, and `git diff --check` pass.
- A final first-principles ownership/reachability review found no blocker after requiring real resources, exact budget identity, guarded cleanup, and compiler-real topology.

XuPeng-SH:
Current-head CI follow-up:

- Previous head `683196b64a` passed UT, UT coverage, Compose BVT, Standalone pessimistic BVT, CodeQL, Iceberg E2E, and the official coverage merge (`78.4%` changed-code coverage; threshold `75%`).
- SCA found one actionable govet/nilness diagnostic in `hashbuild/spill.go`: the new fail-fast control flow proved `firstErr == nil`, making a legacy nested nil check tautological.
- Fixed without changing behavior in `989b875fec`: assign the same typed error directly, then preserve identical buffer cleanup.
- Local revalidation: deterministic CGo hashbuild tests, default vet, MO molint vet, golangci-lint govet (`0 issues`), err-check, and diff-check all pass.

A fresh full CI run is now in progress for `989b875fec`; earlier green jobs are not being treated as current-head evidence.

gouhongshen (approved):
## Codex automated review

No actionable findings. Existing PR reviews and inline comments contain no reviewer findings. I traced ownership transfer, cancellation, mailbox, generation-reuse, range-shuffle, budget/FD, and error paths. Local package tests could not start because required CGo artifacts/headers are absent; no state was changed.

aptend (approved):
Reviewed the latest changes, including spill payload ownership, cancellation/error cleanup, multi-round lifecycle, reuse-generation isolation, and the SQL regression coverage. No blocking issues found.

Issue #26091, opened 2026-07-23
[Bug]: TestPartitionBasedShardCanBeCreated flakes after HAKeeper connection reset


ck89119:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

Issue #24618 tracks a different failure signature in the same test
(`transaction ... has been committed or aborted`). This report tracks a
HAKeeper/logservice connection reset before the test database can be created.

## Branch Name

main

## Commit ID

1ac64a870

## Other Environment Information

- Hardware parameters: GitHub Actions hosted runner
- OS type: Ubuntu/x86
- Go version: 1.26.4
- Workflow: MatrixOne ALL CI, UT Test on Ubuntu/x86

## Actual Behavior

`TestPartitionBasedShardCanBeCreated` in `pkg/tests/shard` failed while creating
its test database:

```text
logservice/store.go:813 propose failed: invalid deadline
hakeeper_client.go:464 failed to allocate id:
writeto tcp4 127.0.0.1:54666->127.0.0.1:16020:
read tcp4 127.0.0.1:54666->127.0.0.1:16020:
read: connection reset by peer

--- FAIL: TestPartitionBasedShardCanBeCreated (41.08s)
```

The embedded cluster also reported that the system was busy and CN heartbeats
exceeded their deadlines immediately before the logservice connection reset.

CI run:
https://github.com/matrixorigin/matrixone/actions/runs/29985797682/job/89137225057

This was observed on PR #26021. That PR does not change `pkg/tests/shard`,
`pkg/shardservice`, or the embedded test-cluster setup. The focused test passed
once on the PR head and five consecutive times on a clean `main` worktree at
`1ac64a870`, which indicates an intermittent infrastructure or embedded-cluster
lifecycle failure rather than a deterministic assertion failure.

## Expected Behavior

The embedded HAKeeper/logservice connection should remain available throughout
test setup, and `TestPartitionBasedShardCanBeCreated` should not fail because
the service connection is reset.

## Steps to Reproduce

```text
make ut
```

Focused diagnostic command:

```text
.agents/skills/mo-dev/scripts/mo-cgo-test \
  -v -count=5 -timeout=300s \
  -run '^TestPartitionBasedShardCanBeCreated$' \
  ./pkg/tests/shard
```

The failure is intermittent and has currently only been observed in the full
Ubuntu/x86 CI UT workload.

## Additional information

Relevant sequence from the CI log:

```text
[...]

PR #26414, opened 2026-07-30
fix bootstrap retry after transient disconnect

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26091

Embedded-cluster bootstrap runs all system initialization statements in one transaction. Under heavy CI load, a temporary HAKeeper/TN connection reset can abort that transaction. The bootstrap owner currently returns the connection error immediately, and CN startup converts it into a process-wide panic.

This PR makes the bootstrap owner recover only from errors classified by `morpc.IsConnectionError`:

- wait for the connection to recover within the existing five-minute bootstrap context;
- check durable bootstrap state before retrying, because a lost commit response can hide a successful commit;
- retry the complete all-or-nothing bootstrap transaction only when the state is definitively incomplete;
- continue returning non-connection errors immediately.

The bootstrap lock ownership and global timeout behavior are unchanged.

Validation:

- `go test ./pkg/bootstrap -count=1`
- `go test -race ./pkg/bootstrap -run 'TestBootstrap(RetriesConnectionError|ChecksStateBeforeRetry|DoesNotRetryNonConnectionError)' -count=1`
- `go test ./pkg/embed -run '^TestBasicCluster$' -count=1` in three independent processes

### DIFF
--- a/pkg/bootstrap/service.go
+++ b/pkg/bootstrap/service.go
@@ -27,6 +27,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/bootstrap/versions"
 	"github.com/matrixorigin/matrixone/pkg/catalog"
 	"github.com/matrixorigin/matrixone/pkg/common/log"
+	"github.com/matrixorigin/matrixone/pkg/common/morpc"
 	"github.com/matrixorigin/matrixone/pkg/common/stopper"
 	"github.com/matrixorigin/matrixone/pkg/container/vector"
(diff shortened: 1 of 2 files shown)


gouhongshen:
Closing because this work is already being handled elsewhere.

PR #26415, opened 2026-07-30
test(tae): restore checkpoint replay coverage

jiangxinmeng1:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26308

  - Restored and enhanced TestReplay7.
  - Used a table-creation helper that persists system catalog entries.
  - Eliminated the race between background checkpointing and manual compact/merge
    operations.

  - Waited for the observable checkpoint GC watermark before restarting and
    verifying recovery.

  - Verified that the initial Close() succeeds and that all 151 rows remain
    readable after restart.


XuPeng-SH (changes requested):
The TAE replay fix and its focused recovery tests look coherent, and TestReplay7 plus the GCed-table/database and replay-store lifecycle tests pass on this head. However, this PR is currently blocked by unrelated scope: commit 7ded9b58d33e (`fix(proxy): quarantine disconnected CN backends`) adds 324 lines of production/test changes across 12 proxy files, while the PR title, description, linked issue #26308, and validation discuss only TAE checkpoint/WAL replay. Please remove that proxy commit from this branch and submit it separately (or point to the intended separate PR), then this TAE change can be reviewed and merged on its own risk/ownership boundary.

gouhongshen (changes requested):
## Codex automated review

The TAE replay change is consistent with existing `OkExpectedEOB` checkpoint-replay handling. An existing review already requests removal of the unrelated proxy commit, so that scope issue is not repeated below. The proxy code itself has two blocking availability regressions. `git diff --check` passed; focused tests could not start locally because `cgo/libmo.dylib` is absent.

### P1 - Do not treat every per-session backend close as a CN outage (pkg/proxy/handler.go:403)
Every `codeServerDisconnect` now calls `ReportBackendDisconnect`, which immediately opens the CN-wide breaker. However, a server-side close only proves that one session ended: CN intentionally closes individual sessions for `wait_timeout` and `KILL CONNECTION`. Those legitimate EOF/reset paths therefore quarantine a healthy CN for the five-second cooldown; in a single-CN deployment new sessions return `ErrAllCNServersBusy`. The tunnel error contains no evidence that the CN itself cannot accept connections, so a session-local close must not directly trip the CN-wide breaker.

### P1 - Use the recorded endpoint before applying the pipe-direction fallback (pkg/proxy/tunnel.go:500)
The new `pipeIOError` endpoint is consulted only when `isConnEndErr` recognizes the underlying error. That whitelist excludes common socket failures such as `syscall.EPIPE`. If a client disconnects while the server-to-client pipe is writing and the write returns EPIPE, this function leaves the error unclassified; `reportPipeError` then applies the pipe's default `codeServerDisconnect`, and the handler quarantines the healthy CN. Endpoint-tagged I/O failures must not fall through to a direction-based code that identifies the opposite peer.

XuPeng-SH (approved):
Deep-reviewed `45381577be60131500b2d7c57bd2883a21f70f02` after the unrelated proxy changes were removed.

The replay change only tolerates catalog misses reported as `OkExpectedEOB`; unexpected errors still fail fast, while existing tables retain the full replayed prepared-DML register/resolve lifecycle. The restored checkpoint-GC recovery test now synchronizes on an observable watermark and verifies close/reopen plus all 151 rows.

`git diff --check`, build, vet, focused tests, adaptive race stress (`TestReplay7` x30; the three replay-store tests x100 each), and full race runs for both owning packages passed locally.

Issue #25441, opened 2026-07-03
[Bug]: recursive CTE silently truncates growing string column instead of reporting Data too long


Ariznawlll:
## Summary

On latest `main` (`7e70bee98`), a recursive CTE with a string anchor and a wider recursive expression does not follow MySQL behavior. MySQL rejects the query with `Data too long for column`, while MatrixOne silently returns truncated / non-growing string values.

This was found during CTE / recursive CTE compatibility exploration against MySQL 8.0.45.

## Environment

- MatrixOne: latest `main` at `7e70bee98`
- MySQL oracle: MySQL 8.0.45

## Reproduce

```sql
with recursive r(n, s) as (
  select 1, 'a'
  union all
  select n + 1, concat(s, 'b') from r where n < 4
)
select * from r order by n;
```

## Expected behavior

MySQL 8.0.45 reports an error because the anchor column infers a narrow string type and the recursive expression grows beyond that width:

```text
ERROR 1406 (22001): Data too long for column 's' at row 1
```

MySQL users need to widen the anchor explicitly, for example:

```sql
with recursive r(n, s) as (
  select 1, cast('a' as char(100))
  union all
  select n + 1, concat(s, 'b') from r where n < 4
)
select * from r order by n;
```

## Actual behavior on MatrixOne

MatrixOne does not report an error, but the recursive string value also does not grow:

```text
n  s
1  a
2  a
3  a
4  a
```

So MO silently produces truncated / incorrect recursive results instead of matching MySQL's type-width behavior.

## Additional coverage

Other CTE paths checked in the same pass matched MySQL semantically:

- non-recursive CTE with join/aggregate
- CTE column aliases and decimal expression type
- `UNION` / `UNION ALL` inside CTE
- NULL and mixed type inference in non-recursive CTE
- recursive tree walk with explicit `cast(id as char(100))`
- recursive numeric sequence and aggregate
- CTE with `INSERT INTO ... WITH c AS (...) SELECT ...`
- CTE with `UPDATE` / `DELETE`
- nested CTE scope

Unhappy-path error codes/messages may differ, but this issue is specifically about a successful wrong result where MySQL errors.

hiSandog:
[...]

PR #26417, opened 2026-07-30
fix recursive cte

daviszhen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25441

  - 递归成员超过锚点 CHAR/VARCHAR 宽度时改为报错，不再静默截断。
  - 显式扩大锚点宽度后，字符串可正常递增长。
  - 新增计划单测和 distributed BVT，覆盖超宽失败及正常递归结果。


XuPeng-SH (changes requested):
Reviewed exact head `12d7b15292af6bf11129258e7fdede1b27d9b428`. The strict-mode corruption is stopped, but the chosen cast hard-codes strict behavior and produces the wrong error class.

### P1 — Recursive string width must honor `sql_mode`

`query_builder.go:4021-4030` always uses `cast_strict` for CHAR/VARCHAR anchors. Official MySQL 8.4.10 gives error 1406 under its default `STRICT_TRANS_TABLES`, but with `SET SESSION sql_mode = ""` the same recursive CTE succeeds and returns four truncated `a` values. This head errors in both modes because `cast_strict` is intentionally unconditional. The new BVT only covers default strict mode, so it misses the non-strict regression.

The existing `cast_assign` execution contract already matches both sides: strict mode rejects over-width CHAR/VARCHAR with 1406, non-strict mode truncates, and prepared execution resolves runtime `sql_mode`. Using it would also avoid the current golden result exposing `internal error: Can\x27t cast...`, which is not MySQLs `Data too long for column` error.

Please route recursive CHAR/VARCHAR coercion through the sql-mode-aware assignment cast and add BVT/unit coverage for strict and non-strict sessions (including prepared execution if the plan is reused across a mode change).

Validation: official MySQL 8.4.10 oracle used for both modes. All focused `TestRecursiveCte*` planner tests pass with the deterministic CGo wrapper, but the new test currently asserts the unconditional `cast_strict` choice rather than the compatibility contract.

daviszhen:
> Reviewed exact head `12d7b15292af6bf11129258e7fdede1b27d9b428`. The strict-mode corruption is stopped, but the chosen cast hard-codes strict behavior and produces the wrong error class.
> 
> ### P1 — Recursive string width must honor `sql_mode`
> `query_builder.go:4021-4030` always uses `cast_strict` for CHAR/VARCHAR anchors. Official MySQL 8.4.10 gives error 1406 under its default `STRICT_TRANS_TABLES`, but with `SET SESSION sql_mode = ""` the same recursive CTE succeeds and returns four truncated `a` values. This head errors in both modes because `cast_strict` is intentionally unconditional. The new BVT only covers default strict mode, so it misses the non-strict regression.
> 
> The existing `cast_assign` execution contract already matches both sides: strict mode rejects over-width CHAR/VARCHAR with 1406, non-strict mode truncates, and prepared execution resolves runtime `sql_mode`. Using it would also avoid the current golden result exposing `internal error: Can\x27t cast...`, which is not MySQLs `Data too long for column` error.
> 
> Please route recursive CHAR/VARCHAR coercion through the sql-mode-aware assignment cast and add BVT/unit coverage for strict and non-strict sessions (including prepared execution if the plan is reused across a mode change).
> 
> Validation: official MySQL 8.4.10 oracle used for both modes. All focused `TestRecursiveCte*` planner tests pass with the deterministic CGo wrapper, but the new test currently asserts the unconditional `cast_strict` choice rather than the compatibility contract.

  - 递归 CTE 的 CHAR/VARCHAR 类型转换改用 SQL mode 感知的 assignment cast。
  - 严格模式下，字符串超出锚点宽度返回 MySQL 兼容的 1406 错误。
  - 非严格模式下，超长字符串按锚点宽度截断。
  - Prepared Statement 执行时动态读取 sql_mode，支持模式切换。
  - 增加 MORPC 版本兼容处理，避免旧 CN 接收不支持的 cast_assign。
  - 补充 strict、non-strict、prepared execution 和显式扩宽锚点的 UT/BVT 覆盖。

XuPeng-SH (approved):
Re-reviewed exact head `e074375220c30aab8f1c6b0954112be932bf2b91`. The previous sql_mode blocker is closed.

Recursive CHAR/VARCHAR coercion now uses the existing protocol-gated assignment-cast path: fully upgraded clusters emit volatile `cast_assign`, so strict mode rejects over-width values while non-strict mode truncates and prepared execution observes mode changes; mixed-version clusters retain the old-node-compatible strict cast. Non-string recursive columns keep their existing generic cast behavior. The BVT covers strict, non-strict, prepared-mode switching, and an explicitly widened anchor.

Validation passed: full `pkg/sql/plan` tests, focused recursive tests under race x20, package build, `go vet`, and `git diff --check`.

heni02 (approved):

PR #26418, opened 2026-07-30
Revert "perf(plan): reuse eligible multi-reference CTEs (#26331)"

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Reverts #26331.

issue #26131 remains open because this rollback restores the previous inline multi-reference CTE behavior.

This reverts squash commit `538a45b37d7162182734c943c59bbbd723da3508` from #26331 at the author request.

The revert removes the multi-reference CTE reuse planner, query-scoped materialized source and spill path, and the associated compile/Dispatch/Merge wiring as one complete unit. Multi-reference CTEs return to independent inline planning, and sink scans use the existing streaming pipeline path.

The revert is synchronized with current `upstream/main` (`2c9608dd45960126ff3a91fc3bbb0addd53d816d`). Later overlapping fixes #26357 and #26393 remain present.

There is no SQL syntax, protobuf, public API, or persistence-format change.

### Validation

Exact head `fb1b16f27b9665203197678f385815755b143b06`:

- previous head `d693b810af44d4d9f0ae22ad3b18d68e78e6dc06`: all major functional CI jobs passed; Matrixone Utils Coverage alone failed because modified-line coverage was exactly 75%, while the gate requires strictly greater than 75%
- added `TestDispatchResetSkipsInvalidRemoteReceiversAndNotifiesHealthyPeer`, covering both nil receiver and nil error-channel guards while proving that a later healthy receiver still receives the original terminal error
- focused coverprofile: `pkg/sql/colexec/dispatch/types.go:227.33,234.14` changed from count 0 in CI to count 1 locally
- exact-head remote Coverage: PASS with 4/4 modified lines covered (`pr_coverage=1.0`, threshold `>0.75`)
- affected planner/compile/dispatch/merge package tests: PASS
- complete affected owning packages under `-race -count=1`: PASS
- focused dispatch terminal and compile sink-scan tests under `-race -count=20`: PASS
- affected package build and vet: PASS
- `git diff --check`: PASS
- exact reverse/overlap audit: only later #26357 and #26393 edits remain relative to the #26331 parent
- schema-v4 preflight: `PASS review=PASS validation=PASS`

BVT: the previous exact-head Proxy multi-CN and pessimistic BVT jobs passed. This follow-up adds package-internal unit coverage only and does not change product behavior.

### QA decision

QA required: yes. This rolls back optimizer selection and the materialized fan-out execution path. CI and environment-level analytical-query regression coverage should complete before merge.


XuPeng-SH (approved):
Deep review completed at fb1b16f27b9665203197678f385815755b143b06. I verified this as a complete rollback rather than only reviewing the final diff:

- the nine product files untouched after #26331 match the pre-#26331 parent byte-for-byte;
- the surviving query_builder.go and operator.go deltas have identical stable patch IDs to the later #26357 and #26393 fixes, respectively;
- all materialized-source planner/compile/operator symbols, package files, and the reopened #26131 regression test are removed with no stale references;
- the added dispatch cleanup test covers the retained nil receiver/error-channel guard without changing rollback behavior.

Local normal tests passed for dispatch, merge, and plan. The only full compile-package failures were the three known Parquet fanout tests already reproducible on clean main; rollback-focused dispatch/compile/CTE tests pass under -race. CI is green and the exact head is mergeable.

Issue #24816, opened 2026-06-03
[Bug] Window aggregate with implicit RANGE frame over a string ORDER BY key errors out


dengn:
## Summary
A window aggregate with an implicit frame over a VARCHAR `ORDER BY` key errors out instead of using a ROWS frame / supporting string ordering.

## Environment
- **Reported on:** MatrixOne Cloud free tier, `8.0.30-MatrixOne-v3.0.11`
- **Retested on:** local Docker `matrixorigin/matrixone` **main branch** build, Git commit `0bfda60` (built 2026-05-30), version string `8.0.30-MatrixOne-v`

## Test case
```sql
WITH m AS (SELECT '2026-01' ym, 100 rev UNION ALL SELECT '2026-02', 200)
SELECT ym, SUM(rev) OVER (ORDER BY ym) cum FROM m;            -- fails
-- workaround: explicit ROWS frame
SELECT ym, SUM(rev) OVER (ORDER BY ym ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) FROM m;  -- works
```

## Actual result
```
ERROR 1064 (HY000): Window '<unnamed window>' with RANGE frame requires ORDER BY expression of numeric or temporal type
```

## Expected
A running aggregate ordered by a string column should work (default to ROWS, as MySQL does for these cases).

## Version status
- v3.0.11 (Cloud): reproduces
- main / 4.0-dev (commit 0bfda60): reproduces

slyunnn:
回归测试结论：通过：对应复现用例结果符合预期。

slyunnn:
### Extended verification for #24816 (2026-08-11)

  Follow-up to the string ORDER BY window-frame report: verified string-key RANGE frame behavior after #26419.

  ### Verification environment

  - MatrixOne branch: `main`
  - MatrixOne commit: `d01c85974f9540f8baa2e768ca86fea952941fe2`
  - Fix under verification: `41332062a6740978f269088a92ee362a888c7bd7`
    `fix(window): support string keys in range frames (#26419)`
  - MOTR branch: `test/issue-24816-string-window-range`
  - MOTR commit: `7ba88825f87f2f34d3e233d6ab18c7532f9351fc`

  ### What was covered

  - Original implicit RANGE frame over a VARCHAR key.
  - Repeated string ORDER BY peers.
  - PARTITION BY with a string ORDER BY key.
  - Explicit RANGE versus explicit ROWS semantics.
[...]

PR #26419, opened 2026-07-30
fix(window): support string keys in range frames

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24816 

fix(window): support string keys in range frames


heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed exact head `8e61b3687eb30a450ce9bddd19274a6a95566fc4`. The binder now relaxes only non-offset RANGE frames, while `RANGE N PRECEDING/FOLLOWING` remains restricted to one numeric/temporal key. Runtime peer lookup correctly reuses the full sorted ORDER BY boundary set, so string duplicates, multiple keys, partitions, DESC, NULL peers, CURRENT ROW, and unbounded endpoints preserve RANGE semantics.

I independently checked the issue query against MySQL 8.4.10: duplicate `2026-02` rows both receive the peer-inclusive cumulative value, matching this PR. No operator lifecycle or memory-ownership change is introduced. Focused plan/window tests and the full `pkg/sql/colexec/window` package passed locally with the deterministic CGo wrapper; CI is green.

PR #26420, opened 2026-07-30
test(plan): cover vector score scalar subquery regression

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23157

test(plan): cover vector score scalar subquery regression


XuPeng-SH (approved):
Reviewed exact head 0cdeaf7c6408. The planner regression reproduces the nested scalar-subquery shape from #23157, asserts that no executable AGG/FILTER/PROJECT/ORDER BY expression retains Expr_Sub, and the BVT validates the actual vector-score aggregate/HAVING result. The focused planner test passes locally and current CI is green.

Issue #25415, opened 2026-07-03
[Bug]: SQL UDF table reads fail when called with per-row column arguments


Ariznawlll:
### Environment

- Branch: latest `main`
- MO git version: `51656deb6`
- MySQL oracle: `8.0.45`

### Reproduce

```sql
drop database if exists mysql_compat_model15_r3_min;
create database mysql_compat_model15_r3_min;
use mysql_compat_model15_r3_min;

create table t_base (
  id int primary key,
  v int
);
insert into t_base values (1,5),(2,15),(3,25),(4,null);

create function f_lookup(p_id int)
returns int
language sql as 'select v from t_base where id = $1';

select f_lookup(2) as lookup_2;
select id, f_lookup(id) as lookup_v from t_base order by id;
select id from t_base where f_lookup(id) >= 15 order by id;
```

### MySQL result

Using equivalent MySQL standard stored-function syntax:

```text
lookup_2
15

id  lookup_v
1   5
2   15
3   25
4   NULL

id
2
3
```

### MO result

Single literal argument works:

```text
lookup_2
15
```

But per-row calls fail:

```text
ERROR 20101 (HY000): internal error: scalar subquery returns more than 1 row
ERROR 20101 (HY000): internal error: scalar subquery returns more than 1 row
```

### Expected

A SQL UDF that reads a table and filters by its argument should be evaluated per input row when called with a column argument, matching MySQL stored-function behavior.

slyunnn:
## Regression and extended verification for #25415 (2026-08-12)

  Follow-up to `fix(plan): bind SQL UDF column arguments per row` (#26421).

  ### Verification environment

  - MatrixOne main: `9a7c98b8f3aa07fad24b411d54c7a9f6cb3a8731`
    - `fix(plan): enforce index metadata invariants safely (#26982)`
  - Fix commit: `f2476f2f6aea8649c3ad23a8dee54a2970dc4923`
    - `fix(plan): bind SQL UDF column arguments per row (#26421)`
  - MOTR branch: `test/issue-25415-sql-udf-column-args`
  - MOTR commit: `d63bef24f67949f557c2156f7ce4de400fd1ff7b`
  - Topology: local single-node MatrixOne service.

  ### What was covered

  - Original regression:
    - SQL UDF reads a table and receives an outer-table column argument per row.
    - Result: OK. Each row returns its matching lookup value; no scalar-subquery multi-row error occurred.

  - Happy path:
    - Literal UDF argument.
[...]

PR #26421, opened 2026-07-30
fix(plan): bind SQL UDF column arguments per row

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25415 

fix(plan): bind SQL UDF column arguments per row


heni02 (approved):

pkg/sql/plan/base_binder.go
```diff
@@ -2649,6 +2654,221 @@ func bindFuncExprImplUdf(b *baseBinder, name string, udf *function.Udf, args []t
...
+
+	var callID uint64
+	if b.builder != nil {
+		b.builder.nextSQLUdfCallID++
+		callID = b.builder.nextSQLUdfCallID
+	}
+
+	markers := make(map[string]*plan.Expr, len(args))
+	markerForOrdinal := func(ordinal int) string {
+		name := fmt.Sprintf("__mo_sql_udf_%d_arg_%d", callID, ordinal)
```
XuPeng-SH: The generated name is not actually reserved by the SQL grammar or catalog, so it can capture a user-authored identifier. For example, expanding `select __mo_sql_udf_1_arg_1, $1` for the first call makes the second expression the same identifier as the first; because `baseBindExpr` checks `sqlUdfArgs` before normal column binding, both expressions resolve to the call argument. A focused test that parses the rewritten SQL and asserts the first identifier is not captured fails, which means a valid UDF body can silently return the caller value instead of its own column. Please make each marker provably absent from the original body (or track the generated AST nodes independently of identifier text) and add a collision regression test.

XuPeng-SH (changes requested):
The new bound-expression approach fixes the reported per-row lookup and the supplied tests pass, but the textual marker identity can collide with a legal identifier already present in the UDF body. That collision is resolved before normal column binding and silently changes query semantics. Please address the inline finding and add the regression case.

iamlinjunhong:
> The new bound-expression approach fixes the reported per-row lookup and the supplied tests pass, but the textual marker identity can collide with a legal identifier already present in the UDF body. That collision is resolved before normal column binding and silently changes query semantics. Please address the inline finding and add the regression case.

fixed

pkg/sql/plan/base_binder.go
```diff
@@ -2649,6 +2654,235 @@ func bindFuncExprImplUdf(b *baseBinder, name string, udf *function.Udf, args []t
...
+		switch sql[i] {
+		case '\'', '"', '`':
+			quote := sql[i]
+			result.WriteByte(quote)
+			i++
+			for i < len(sql) {
+				ch := sql[i]
+				result.WriteByte(ch)
+				i++
+				if ch == '\\' && quote != '`' && !noBackslashEscapes && i < len(sql) {
```
XuPeng-SH: Honor ANSI_QUOTES while scanning double-quoted identifiers. Under stored mode `ANSI_QUOTES` (and the `ANSI` composite mode), `"` closes an identifier after the backslash; it is not a string escape. This condition consumes that closing quote and treats the later live parameter as quoted text. Repro: body `select 1 from (select 1) as "a\" where $1 = 2` builds at the base commit, but this head leaves `$1` unresolved. Derive the quote behavior from the parser SQL-mode flags and cover this case.

pkg/sql/plan/base_binder.go
```diff
@@ -2649,6 +2654,235 @@ func bindFuncExprImplUdf(b *baseBinder, name string, udf *function.Udf, args []t
...
+						break
+					}
+					i++
+				}
+				continue
+			}
+			result.WriteByte(sql[i])
+			i++
+		case '/':
+			if i+1 < len(sql) && sql[i+1] == '*' {
```
XuPeng-SH: Do not skip live parameters inside executable comments. MatrixOne parses the contents of `/*! ... */` as SQL, but this branch treats every block as an ordinary comment. With body `/*! $1 + */ 1`, the base commit builds `f(2)` while this head leaves `$1` as a column reference and fails. Distinguish executable comments and rewrite parameters in their SQL payload.

pkg/sql/plan/base_binder.go
```diff
@@ -2649,6 +2654,235 @@ func bindFuncExprImplUdf(b *baseBinder, name string, udf *function.Udf, args []t
...
+						i++
+						break
+					}
+					i++
+				}
+				continue
+			}
+			result.WriteByte(sql[i])
+			i++
+		case '/':
```
XuPeng-SH: Recognize the lexer-supported `//` line comment before scanning quotes. The real scanner ignores the rest of that line, while this helper walks it as SQL; an unmatched quote in the comment then swallows a parameter on the next line. Body `select 1 // quote-start\n + $1` (with an actual single quote after `//`) builds at the base commit, but this head skips `$1` and fails. Mirror the lexer line-comment rule and add a multiline regression.

XuPeng-SH (changes requested):
The previous marker-collision blocker is fixed. The remaining issue is that the new handwritten marker scanner does not match the MatrixOne lexer for several supported lexical forms, so valid existing SQL UDF bodies can leave live $n parameters unresolved. I reproduced all three cases against this exact head; the same end-to-end BuildPlan cases pass at the base commit and fail here with `invalid input: column $1 does not exist`. The PR package tests, build, and vet pass otherwise. Please align marker substitution with the real lexer semantics and add focused regressions.

XuPeng-SH:
@iamlinjunhong CI failed

XuPeng-SH (approved):
Re-reviewed exact head 8a1e4198799f. The prior marker-collision and lexer-mismatch blockers are resolved: substitution now uses the MatrixOne MySQL scanner with the stored SQL mode, including ANSI/ANSI_QUOTES, executable comments, and // comments, while body-unique markers preserve normal identifier binding. Focused end-to-end, collision, scanner, correlation traversal, scalar-subquery argument, and race tests pass. No remaining actionable issues found.

PR #26423, opened 2026-07-30
fix: enforce MySQL DML target restrictions

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25382 

fix: enforce MySQL DML target restrictions

### DIFF
--- a/pkg/common/moerr/error.go
+++ b/pkg/common/moerr/error.go
@@ -97,6 +97,8 @@ const (
 	ErrViewWrongList        uint16 = 20318
 	ErrWrongArguments       uint16 = 20319
 	ErrDerivedMustHaveAlias uint16 = 20320
+	ErrWrongUsage           uint16 = 20321
+	ErrUpdateTableUsed      uint16 = 20322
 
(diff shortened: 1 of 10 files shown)


XuPeng-SH (approved):
Deep-reviewed at exact head d98e9ed674622e69d2a4ecfa9ede7f06a626b3a5.

Verified the MySQL 1221/1093 error mapping, multi-table UPDATE shape detection through joins/parentheses/aliases, target identity matching, traversal of UPDATE/DELETE expressions and nested SELECT/UNION/JOIN/HAVING/ORDER/LIMIT paths, legal self-join behavior, and the intentional derived-table/CTE materialization boundary. Both modern and legacy planner entry paths perform the validation.

Local exact-head verification passed: full ./pkg/sql/plan tests, ./pkg/common/moerr tests, go vet for both touched packages, and git diff --check. GitHub CI is green and the head remained unchanged.

Issue #26229, opened 2026-07-27
[Feature Request]: Support MongoDB as a native external table source


iamlinjunhong:
- [x] I have checked the existing issues.

## Problem

Today MongoDB data is pulled and aggregated by an external Java/Python worker, then written to MatrixOne through JDBC/MySQL protocol:

```text
MongoDB -> Java/Python worker -> MatrixOne CN -> MO storage
```

This adds another process, another network/protocol hop, and prevents MongoDB reads from participating directly in the MatrixOne execution pipeline. We want users to query MongoDB as an external table and let normal MO operators perform filter, time-window aggregation, group aggregation, join, and `INSERT/REPLACE ... SELECT`.

## Proposed feature

Add a read-only MongoDB external table engine. Users should not need to deploy a JDBC driver, Python worker, or sidecar. The MongoDB official Go Driver v2 is linked inside the CN process, so the data path is directly:

```text
MongoDB socket -> MO CN MongoScan -> standard batch.Batch -> MO operators/storage
```

Proposed SQL surface (syntax is illustrative):

```sql
CREATE MONGODB CONNECTION nesr_mongo WITH (
  'hosts' = 'mongo-1:27017,mongo-2:27017',
  'credential_secret_ref' = 'secret://nesr/mongodb-reader',
  'tls' = 'required',
  'read_concern' = 'majority'
);

CREATE EXTERNAL TABLE mongo_sensor_readings (
  mongo_id VARBINARY(12) MONGODB_PATH '_id',
  pump VARCHAR(20) MONGODB_PATH 'pump',
  crew VARCHAR(10) MONGODB_PATH 'crew',
  ts DATETIME(3) MONGODB_PATH 'ts',
  value DOUBLE MONGODB_PATH 'value'
) ENGINE = MONGODB WITH (
  'connection' = 'nesr_mongo',
  'database' = 'nesr_source',
  'collection' = 'sensor_readings'
);
```

## Simple implementation design

1. **SQL/catalog**
   - Add `ENGINE=MONGODB`, column-level `MONGODB_PATH`, and a versioned MongoDB connection/table mapping.
   - Store credential references only. URI credentials must never enter `rel_createsql`, plan, pipeline protobuf, logs, trace, or `EXPLAIN`.

2. **Logical plan**
   - Reuse `plan.Node_EXTERNAL_SCAN` with a new `ExternType_MONGODB_TB` and `plan.MongoScan` metadata.
[...]

PR #26424, opened 2026-07-30
feat: add MongoDB external table scan

iamlinjunhong:
Add tenant-scoped MongoDB connection DDL and catalog metadata, secure driver-backed scanning, conversion and predicate semantics, downstream aggregation support, and complete ALTER/DROP/account lifecycle handling across CNs. Include one connector E2E CI job, regression coverage, and customer-neutral operational templates.

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26229 

feat: add MongoDB external table scan


XuPeng-SH (changes requested):
Blocking correctness issue: the external-table DDL accepts `NOT NULL`, but `ParseTableMappingSpec` drops `planned.Typ.NotNullable` when it builds `ColumnMapping` (mapping.go:137-144), and the runtime converter unconditionally appends NULL for missing/BSON null/undefined values and for `try_null` conversion failures (converter.go:138-143, 153-164). I reproduced this end to end at the component boundary with `CREATE EXTERNAL TABLE t (v BIGINT NOT NULL MONGODB_CONVERT "try_null") ...`: mapping succeeds with the plan column marked `NotNullable=true`, then an empty BSON document produces a NULL in the output vector. This violates a planner-visible schema invariant and can make nullability-based rewrites/join reasoning unsound. Please either reject NOT NULL (and other unenforceable constraints) on MongoDB external columns, or carry nullability through catalog/plan metadata and make the converter fail rather than emit NULL for a non-nullable mapping. Add coverage for missing, explicit BSON null, and try_null conversion failure on a declared NOT NULL column.

fengttt (approved):

fengttt (commented):
Wait, why does mongo db has anything to do with frontend?

iamlinjunhong:
> Wait, why does mongo db has anything to do with frontend?

MongoDB data access has nothing to do with the frontend package. The frontend integration exists only because MatrixOne’s frontend is the server-side SQL session and DDL execution layer, which owns tenant-aware connection DDL, catalog mutations, and post-commit cache retirement. MongoDB scans themselves run in mongoscan and pkg/sql/mongodb.

XuPeng-SH (approved):
Deep re-reviewed exact head `b7ec571875852b6d691a7543ef8dd9f422b88af8`: no blocking findings. The previous NOT NULL correctness blocker is closed end to end: declared nullability is captured from the planned TableDef, persisted in catalog mapping JSON, preserved through protobuf plan conversion, and enforced by the runtime converter for missing paths, BSON null/undefined, invalid traversal, and `try_null` conversion failures. Error rollback restores every vector and the batch row count, so a rejected document cannot leave a partial row.

Fresh local evidence: converter/catalog NOT NULL tests pass under `-race -count=10`, the planner DDL round-trip test passes, and `go build` plus `go vet` pass for `pkg/sql/mongodb`, `pkg/sql/colexec/mongoscan`, and `pkg/sql/plan`. Final diff is clean, latest CI is green, and the PR remains MERGEABLE.

heni02 (approved):

XuPeng-SH:
@iamlinjunhong conflict

iamlinjunhong:
> @iamlinjunhong conflict

fixed

Issue #26425, opened 2026-07-30
[Bug]: ALTER TABLE ADD COLUMN panics on a wide table with a composite primary key


gouhongshen:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

## Branch Name

`v4.1.4`; the same failure was also observed on `v4.1.2`.

## Commit ID

- `v4.1.4`: `a185c23b375956afe098320baf8df39000a43936`
- Docker image: `registry.cn-shanghai.aliyuncs.com/matrixorigin/matrixone:4.1.4`
- Image digest: `sha256:16ef37311b43c6882ed8242bf38b1e281d853d4f4f6163830fe10773c9fe011d`

## Other Environment Information

- Hardware parameters: Apple Silicon / arm64
- OS type: macOS 26.5.2 (Darwin 25.5.0)
- `SELECT VERSION()`: `8.0.30-MatrixOne-v4.1.4`
- Standalone quickstart MatrixOne in Docker

## Actual Behavior

Adding a nullable column to the table below causes an internal MatrixOne panic:

```text
ERROR 20101 (HY000): internal error: panic runtime error: index out of range [41] with length 41
runtime.panicBounds64
    /usr/local/go/src/runtime/panic.go:236
runtime.panicBounds
    /usr/local/go/src/runtime/asm_arm64.s:1770
github.com/matrixorigin/matrixone/pkg/sql/colexec/preinsert.(*PreInsert).constructColBuf
    /go/src/github.com/matrixorigin/matrixone/pkg/sql/colexec/preinsert/preinsert.go:146
github.com/matrixorigin/matrixone/pkg/sql/colexec/preinsert.(*PreInsert).Call
    /go/src/github.com/matrixorigin/matrixone/pkg/sql/colexec/preinsert/preinsert.go:242
```

The table has 40 user-visible columns and a composite primary key. MatrixOne exposes an additional hidden composite-key column named `__mo_cpkey_col`. The panic index is consistent with the hidden column being included in the pre-insert attribute count while the input batch contains one fewer vector.

A minimal three-column table with the same two-column primary key does **not** reproduce the issue. The complete schema below reproduces it consistently with one row.

## Expected Behavior

`ALTER TABLE ... ADD COLUMN` should complete successfully and preserve existing rows, or return a normal SQL error. It must not panic inside `preinsert.constructColBuf`.

## Steps to Reproduce

Run on MatrixOne v4.1.4:

```sql
[...]

PR #26427, opened 2026-07-30
fix(plan): avoid treating checkpoint columns as CHECK constraints

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26425

`extractTopLevelCheckDefs` used prefix matching for the `CHECK` and `CONSTRAINT` keywords. As a result, ordinary columns such as `checkpoint_version` and `checkpoint_json` were extracted from the stored CREATE TABLE SQL as table-level CHECK definitions.

During ALTER TABLE copy-table reconstruction, those false definitions were appended again, creating duplicate columns. With a composite primary key, the malformed reconstructed table later panicked in `PreInsert.constructColBuf`; the table width and hidden composite-key column were symptoms, not the root cause.

This PR uses the existing identifier-boundary-aware `hasKeywordAt` helper for both keywords, so only complete SQL keywords match. Its identifier predicate now matches the MySQL lexer by treating `$` as an unquoted identifier character, preventing accepted names such as `check$point` from being classified as CHECK definitions.

It adds:

- A focused plan-layer unit test for `checkpoint_*` and `check$point` columns alongside a real table-level CHECK constraint.
- A DDL BVT with both the independently minimized reproducer and the complete #26425 reproducer. The minimized case also exercises `check$point` through the real CREATE, INSERT, ALTER, SELECT, and SHOW CREATE paths. The complete case retains all 40 columns, eight indexes, the composite primary key, table-level CHECK, original INSERT, and ALTER TABLE ADD COLUMN statement.

Validation:

- `go test -race ./pkg/sql/plan -run TestExtractTopLevelCheckDefsIgnoresCheckPrefixedColumns -count=100`
- `./run.sh -p <matrixone>/test/distributed/cases/ -i alter_table_check_prefixed_column -r 100` (18/18 statements passed)
- Reproduced #26425 on an unmodified current `main` build.
- Rebuilt MatrixOne with this fix and ran the full issue schema: ALTER succeeded, the existing row was preserved, the new column appeared once, both `checkpoint_*` columns appeared once, and the real CHECK constraint remained present.


XuPeng-SH (changes requested):
[P1] The author's root-cause description is correct—this must distinguish a complete `CHECK` keyword from a column-name prefix—but the reused boundary helper still does not match MatrixOne's actual identifier grammar. `pkg/sql/parsers/dialect/mysql/scanner.go` includes `$` in `isLetter`, so `check$point` is a valid unquoted column identifier. `isIdentChar` in `build_show_util.go` omits `$`, causing `hasKeywordAt("CHECK$POINT INT", "CHECK", 0)` to return true.

I reproduced this through the real parser on exact head `1778952161d0e9884d44a46a332aebc1fcc64777`: `CREATE TABLE t (id INT, check$point INT, CONSTRAINT chk_id CHECK (id > 0))` binds successfully, but `extractTopLevelCheckDefs` returns both `check$point INT` and the real constraint. The ALTER copy path can therefore still duplicate an ordinary column and reach the same malformed reconstruction class fixed for `checkpoint_*`.

Please make the keyword-boundary predicate consistent with the lexer (at least include `$`) and add a regression using an accepted `$` identifier, ideally through the ALTER path as well as the focused extractor test. The issue's exact `checkpoint_*` reproducer and the existing CHECK constraint behavior otherwise look fixed; full plan tests and the focused tests under `-race -count=100` pass.

gouhongshen:
Fixed in `e3edf34a0`. The review is correct: MatrixOne MySQL lexer treats `$` as an unquoted identifier character, while `isIdentChar` did not. The boundary predicate now includes `$`. I added `check$point` to the focused extractor regression and to the BVT real CREATE/INSERT/ALTER/SELECT/SHOW CREATE path. Verified with full `pkg/sql/plan`, the focused test under `-race -count=100`, and BVT 18/18.

heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed updated exact head `db0ea88dec4a7335edf53a7e4b5be16912326d55` and the author response. The requested `$` identifier-boundary fix is correctly aligned with the MatrixOne MySQL lexer (`isLetter` accepts `_` and `$`), and both the focused extractor regression and real ALTER/SHOW CREATE BVT now cover `check$point` while preserving the genuine table-level CHECK.

I also rechecked the surrounding quote/comment/top-level splitting path. No remaining correctness blocker found. Local full `./pkg/sql/plan` passed with the deterministic CGo wrapper (`3.304s`), in addition to green CI.

Issue #23104, opened 2025-11-24
Feature Not Supported: LIKE ESCAPE syntax not supported


dengn:
## Description

MatrixOne does not support the `LIKE ... ESCAPE` syntax for specifying custom escape characters in LIKE pattern matching. When attempting to use the ESCAPE clause, MatrixOne returns a SQL syntax error.

## Error Message

```
ERROR 1064 (HY000): SQL parser error: You have an error in your SQL syntax; check the manual that corresponds to your MatrixOne server version for the right syntax to use. syntax error near "ESCAPE"
```

## Affected Cases

This issue affects **1 test case** from our boundary testing on the `industry_radar_test` database.

### Example 1: LIKE with ESCAPE

```sql
SELECT * FROM company WHERE full_name LIKE '%\%' ESCAPE '' LIMIT 10;
```

**Error**: `ERROR 1064 (HY000): SQL parser error: syntax error near "ESCAPE"`

## Expected Behavior

In MySQL, the `LIKE ... ESCAPE` syntax is fully supported. It allows you to specify a custom escape character for pattern matching, which is useful when you need to match literal special characters.

Example MySQL syntax:
```sql
-- Match strings containing literal '%'
SELECT * FROM table WHERE column LIKE '%\%' ESCAPE '\';

-- Match strings containing literal '_'
SELECT * FROM table WHERE column LIKE '%\_' ESCAPE '\';

-- Custom escape character
SELECT * FROM table WHERE column LIKE '%#%' ESCAPE '#';
```

## Impact

- **MySQL Compatibility**: LIKE ESCAPE is a standard MySQL feature
- **Functionality**: Users cannot match literal special characters in LIKE patterns
- **Migration**: Applications using LIKE ESCAPE will fail when migrating to MatrixOne
- **Workaround**: Users must use alternative methods (e.g., REGEXP) which may have different semantics

## Use Cases

LIKE ESCAPE is commonly used for:
1. **Matching Literal Special Characters**: Matching strings that contain literal '%' or '_' characters
2. **Custom Escape Sequences**: Using custom escape characters for specific pattern matching needs
3. **Data Validation**: Validating data that may contain special characters

## Workaround

As a temporary workaround, users can:
1. Use REGEXP instead of LIKE (if the pattern is simple enough)
2. Use string functions to escape special characters before matching
3. Use application-level pattern processing

However, these workarounds:
- May have different semantics
- May be less efficient
- May not be equivalent in all cases

## Suggested Fix

[...]

PR #26429, opened 2026-07-30
fix: support explicit LIKE ESCAPE semantics

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #23104

MatrixOne parses `LIKE ... ESCAPE` but previously drops the ESCAPE expression during binding, so custom escapes and prepared ESCAPE parameters do not affect execution.

This change:

- threads ESCAPE through LIKE, NOT LIKE, ILIKE, and NOT ILIKE binding;
- adds three-argument LIKE/ILIKE overloads while preserving the existing two-argument path;
- accepts empty or one Unicode escape character, treats NULL ESCAPE as disabled escaping without suppressing NULL propagation from the value or pattern, and rejects nonconstant or multi-character ESCAPE values with `Incorrect arguments to ESCAPE`;
- rejects empty ESCAPE under execution-time `NO_BACKSLASH_ESCAPES`, using the live resolver with serialized `SessionInfo` fallback for remote execution, while NULL ESCAPE bypasses this explicit-empty validation;
- recognizes an ILIKE escape before case folding;
- skips the two-argument prefix optimization for explicit ESCAPE predicates;
- preserves the third ESCAPE argument in EXPLAIN for direct and NOT-wrapped LIKE/ILIKE expressions.

The execution-time constant requirement and NULL ESCAPE behavior follow MySQL 8.4.10. Prepared parameters are accepted; row-dependent column escapes are rejected.

## Validation

- old implementation reproduction: binder regression failed because LIKE retained only two arguments;
- review reproduction on exact head `a72938624`: the new NULL ESCAPE SQL-mode regression failed with `expected true, got NULL`;
- focused binder, matcher, vector evaluator, SQL-mode, NULL ESCAPE, EXPLAIN, and invalid-arity tests: PASS;
- seven exact focused tests, each independently run with `-race -count=100`: PASS;
- full `pkg/sql/plan`, `pkg/sql/plan/function`, and `pkg/sql/plan/explain` with `-race -count=1`: PASS;
- `go build` and `go vet` for all three owning packages: PASS;
- `git diff --check`: PASS;
- CR Learning Gate preflight: `PASS review=PASS validation=PENDING` on rebased head `c05ddd0b9ea5bfaf7d40021111cc50378480eb33`, diff hash `70b4859a3f1be8b732257edd45cfbb89cddea8dca9a8d8ebfcfe2caaa8850954`.

BVT coverage is added to `test/distributed/cases/operator/like_operator.sql`, including direct and prepared NULL ESCAPE under default and `NO_BACKSLASH_ESCAPES` modes plus the existing prepared statement whose `sql_mode` changes before execution. The superseded head passed this canonical case `190/190` in PROXY BVT; that job later failed on an unrelated `pitr_basic.sql` `DROP ACCOUNT` 300-second timeout, and Coverage failed only because its producer failed. The branch was cleanly rebased onto current `upstream/main`; fresh exact-head CI remains pending. The PR remains Ready for Review.

## QA handoff

- QA required: yes — user-visible MySQL compatibility behavior.
- Production entrypoint: SQL LIKE/NOT LIKE/ILIKE/NOT ILIKE with explicit ESCAPE, including prepared ESCAPE parameters and `NO_BACKSLASH_ESCAPES`.
- Automated terminal coverage: canonical operator BVT plus binder, vector execution, SQL-mode, NULL ESCAPE, and EXPLAIN unit tests.
- Remaining validation: exact-head CI/BVT, then QA comparison against MySQL for custom, empty, Unicode, NULL, prepared, SQL-mode, and invalid ESCAPE cases.
- Closure policy: after merge, assign to QA and move the issue to testing; close only after QA verifies the deployed build.


heni02 (approved):

XuPeng-SH (changes requested):
Deep-reviewed updated exact head `a72938624a8b4f527006f641d27b772e9a726ae1`, including the latest SQL-mode and EXPLAIN fixes. Custom, empty, Unicode, prepared, NOT, and ILIKE plumbing are otherwise coherent, but one user-visible MySQL compatibility blocker remains.

### P1 — `ESCAPE NULL` is not SQL NULL propagation in MySQL

`likeFnWithEscape` marks every output row NULL when the third argument is a constant NULL, and the new BVT locks that behavior in. Official MySQL 8.4.10 instead treats a NULL ESCAPE value as no escape character:

- `'axb' LIKE 'a_b' ESCAPE NULL` returns `1`, with `ISNULL(...) = 0`.
- `'a_b' LIKE 'a!_b' ESCAPE NULL` returns `0`, with `ISNULL(...) = 0`.
- A prepared ESCAPE parameter set to NULL has the same non-NULL behavior.
- Under `NO_BACKSLASH_ESCAPES`, `ESCAPE NULL` still succeeds as no escape, while explicit `ESCAPE ''` correctly raises error 1210.

So NULL cannot simply share the explicit-empty validation path either. Please normalize a NULL ESCAPE to disabled escaping without applying the `NO_BACKSLASH_ESCAPES` empty-string rejection, adjust the function strict/null behavior as needed, and replace the current NULL BVT/UT expectations.

Validation: official MySQL 8.4.10 container under both default and `NO_BACKSLASH_ESCAPES` modes. A direct execution-function probe on this exact head failed with `expected true, but get NULL`. The PR's focused matcher/SQL-mode/EXPLAIN tests pass locally, confirming the remaining gap is absent from the current oracle matrix.

pkg/sql/plan/function/func_builtin_regexp.go
```diff
@@ -88,6 +99,65 @@ func (op *opBuiltInRegexp) iLikeFn(parameters []*vector.Vector, result vector.Fu
...
+	proc *process.Process,
+	length int,
+	selectList *FunctionSelectList,
+	caseInsensitive bool,
+) error {
+	rs := vector.MustFunctionResult[bool](result)
+	if !parameters[2].IsConst() {
+		return moerr.NewInvalidInputNoCtx("Incorrect arguments to ESCAPE")
+	}
+	if parameters[2].IsConstNull() {
```
XuPeng-SH: **P1:** This early NULL propagation contradicts MySQL ESCAPE semantics. MySQL 8.4.10 returns `1` (not NULL) for `'axb' LIKE 'a_b' ESCAPE NULL`; a prepared NULL ESCAPE behaves the same, including under `NO_BACKSLASH_ESCAPES`. NULL means disabled escaping here, but unlike explicit `ESCAPE ''`, it must not trigger the SQL-mode error. The current BVT expectation of NULL should be changed as part of the fix.

VioletQwQ-0: Fixed in 9270b34282cc534f974d374dd43572ef8035dfd4. NULL ESCAPE now disables escaping and bypasses explicit-empty SQL-mode validation, while NULL value or pattern operands still propagate NULL. I added LIKE/ILIKE unit coverage plus direct and prepared BVT cases under default and NO_BACKSLASH_ESCAPES. The focused tests (including race count 100), all three owning-package race suites, build, vet, diff check, and exact-head CR preflight pass; BVT remains pending exact-head CI.

XuPeng-SH (approved):
Re-reviewed the exact latest head. The previous ESCAPE NULL blocker is closed: NULL disables escaping without suppressing NULL propagation from value/pattern or triggering explicit-empty SQL-mode validation. I also checked NOT LIKE/NOT ILIKE binding, ILIKE case-fold ordering, explicit-escape optimizer bypass, explain output, prepared parameters, and remote/session SQL-mode fallback. Full pkg/sql/plan/function, pkg/sql/plan/explain, and pkg/sql/plan tests pass locally; current CI is green.

Issue #26120, opened 2026-07-23
[Bug]: DATA BRANCH from a snapshot links children to a recreated same-name table


Ariznawlll:
### Describe the bug

When `DATA BRANCH CREATE TABLE` reads a named snapshot after the source table has been dropped and recreated with the same name, the new branch contains the correct historical data but its ancestry metadata points to the current same-name table rather than the table object stored in the snapshot.

The matching branch-protection snapshot also records the current table ID with a timestamp that predates that table's creation. Subsequent `DATA BRANCH DIFF` operations fail because the ancestry resolver looks for the current table ID at the historical timestamp.

Tested at commit `2eb7f523e5a10f352d3e258183ed8440510388de`.

### Steps to reproduce

```sql
drop snapshot if exists g4d_parent_identity_sp;
drop database if exists g4d_parent_identity;
create database g4d_parent_identity;
use g4d_parent_identity;

create table parent_t (
  id int primary key,
  val varchar(20)
);
insert into parent_t values (1, 'snapshot-row');
create snapshot g4d_parent_identity_sp
  for table g4d_parent_identity parent_t;

set @snapshot_parent_id = (
  select rel_id
  from mo_catalog.mo_tables {snapshot='g4d_parent_identity_sp'}
  where account_id = 0
    and reldatabase = 'g4d_parent_identity'
    and relname = 'parent_t'
);

drop table parent_t;
create table parent_t (
  id int primary key,
  val varchar(20)
);
insert into parent_t values (2, 'current-row');

set @current_parent_id = (
  select rel_id
  from mo_catalog.mo_tables
  where account_id = 0
    and reldatabase = 'g4d_parent_identity'
    and relname = 'parent_t'
);

data branch create table child_t
  from parent_t{snapshot='g4d_parent_identity_sp'};

select * from child_t;

select bm.table_id, bm.p_table_id, bm.clone_ts
from mo_catalog.mo_branch_metadata bm
join mo_catalog.mo_tables mt on mt.rel_id = bm.table_id
where mt.reldatabase = 'g4d_parent_identity'
  and mt.relname = 'child_t';

select sname, obj_id, ts
from mo_catalog.mo_snapshots
where kind = 'branch'
  and database_name = 'g4d_parent_identity'
  and table_name = 'parent_t';

data branch diff child_t against parent_t;
```

### Actual behavior

The historical and current table IDs are different:

```text
snapshot parent ID = 287580
current parent ID  = 287581
```

[...]

Issue #26114, opened 2026-07-23
[Bug]: DATA BRANCH TO ACCOUNT ignores the target account branch quota


Ariznawlll:
## Description

`DATA BRANCH CREATE TABLE ... TO ACCOUNT` and `DATA BRANCH CREATE DATABASE ... TO ACCOUNT` check the branch quota of the initiating account instead of the target account.

When the target account has the BRANCH feature disabled with quota `0`, `sys` can still create branch tables in that account. The created branch metadata rows are attributed to `sys`, so the target account quota and active branch count are both bypassed.

## Environment

- Branch: `main`
- Commit: `2eb7f523e5a10f352d3e258183ed8440510388de`

## Reproduction

1. Create a target account and set its branch quota to `0`:

```sql
CREATE ACCOUNT g4d_cross_quota ADMIN_NAME = 'admin' IDENTIFIED BY '111';

SET @target_id = (
  SELECT account_id
  FROM mo_catalog.mo_account
  WHERE account_name = 'g4d_cross_quota'
);

SELECT mo_feature_limit_upsert(@target_id, 'branch', '', 0);
```

2. In the target account, create the destination database:

```sql
CREATE DATABASE g4d_cross_quota_dst;
```

3. In `sys`, create the source table and snapshot:

```sql
CREATE DATABASE g4d_cross_quota_src;
CREATE TABLE g4d_cross_quota_src.base(
  id INT PRIMARY KEY,
  v VARCHAR(20)
);
INSERT INTO g4d_cross_quota_src.base VALUES (1, 'one');
CREATE SNAPSHOT g4d_cross_quota_sp
  FOR TABLE g4d_cross_quota_src base;
```

4. From `sys`, create three branch tables in the target account:

```sql
DATA BRANCH CREATE TABLE g4d_cross_quota_dst.b1
  FROM g4d_cross_quota_src.base{snapshot='g4d_cross_quota_sp'}
  TO ACCOUNT g4d_cross_quota;

DATA BRANCH CREATE TABLE g4d_cross_quota_dst.b2
  FROM g4d_cross_quota_src.base{snapshot='g4d_cross_quota_sp'}
  TO ACCOUNT g4d_cross_quota;

DATA BRANCH CREATE TABLE g4d_cross_quota_dst.b3
  FROM g4d_cross_quota_src.base{snapshot='g4d_cross_quota_sp'}
  TO ACCOUNT g4d_cross_quota;
```

5. Query the target account:

```sql
[...]

PR #26430, opened 2026-07-30
test: add Data Branch issue BVT coverage

Ariznawlll:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26071, #26073, #26074, #26078, #26081, #26114, #26118, #26120, #26121, #26127, #26128, #26136, #26205

Add executable BVT `.sql/.result` regression cases for fixed Data Branch and CLONE issues. The cases cover quoted identifiers, snapshots, BOOL/BIT primary keys, bulk DML, cross-account quota, CSV output errors, vector values, and deep lineage.

### DIFF
--- a/test/distributed/cases/git4data/branch/edge/issue_26071_quoted_primary_keys.result
+++ b/test/distributed/cases/git4data/branch/edge/issue_26071_quoted_primary_keys.result
@@ -1,0 +1,90 @@
+drop database if exists bvt_issue_26071;
+create database bvt_issue_26071;
+use bvt_issue_26071;
+create table control_base(id int primary key, value int);
+insert into control_base values (1, 10), (2, 20);
+data branch create table control_src from control_base;
+data branch create table control_dst from control_base;
+update control_src set value = 200 where id = 2;
+data branch pick control_src into control_dst keys(2);
+create table quoted_base(`order` int primary key, value int);
+insert into quoted_base values (1, 10), (2, 20);
+data branch create table quoted_src from quoted_base;
+data branch create table quoted_left from quoted_base;
+data branch create table quoted_pick from quoted_base;
+data branch create table quoted_merge from quoted_base;
+delete from quoted_src where `order` = 1;
+update quoted_src set value = 200 where `order` = 2;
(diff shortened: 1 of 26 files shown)


XuPeng-SH (changes requested):
Blocking finding on exact head 14890204c9253728542e2aa0e55cb594ebb316f1:

[P1] The #26081 BVT records two assertion-query failures as the expected baseline instead of validating the PICK/MERGE result. In test/distributed/cases/git4data/branch/edge/issue_26081_hashmap_bulk.sql:40-45 and :53-55, SUM(boolean-expression) is unsupported; the paired result therefore contains "invalid argument aggregate function sum, bad value [BOOL]" at lines 43 and 55. Because there is no intentional-error directive and these are the only selected/unselected and per-change-type checks, the BVT can stay green while those behavioral assertions never execute.

Please rewrite the expressions to produce integers, for example SUM(CASE WHEN condition THEN 1 ELSE 0 END), regenerate the result, and verify the expected counts for selected updates/deletes/inserts, unselected controls, and merged change classes.

I reviewed all 13 SQL/result pairs and checked all other recorded errors against explicit @regex expectations. CI is green, but it currently confirms the erroneous baseline above, so it does not close this finding.

XuPeng-SH (approved):
Deep re-review complete on exact head `9c7528b89cc9a2eeeac1791e87fe7b539fa211ba`; no blocking finding.

Mergify rebased the already-reviewed change onto current `main` (`37694c49f2f450f7aa7bb5560279d7ab7007c05d`). All 26 PR-owned SQL/result files are byte-identical to the reviewed tree; the only tree delta is eight unrelated files brought in from `main`, with no conflict overlap. The exact-head PR diff passes `git diff --check`.

The prior blocker remains fixed: both unsupported `SUM(BOOL)` assertions are integer `CASE` aggregates, and the regenerated result proves PICK selected/unselected controls plus MERGE update/delete/insert counts instead of recording query errors. The arithmetic closes against the 50k base and the 10k update/delete/insert transformations.

PR #26438, opened 2026-07-30
fix(hashbuild): align batch copy budget with allocation

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26413

HashBuild budget admission estimated copied batches from source allocation plus empirical multipliers. That model diverged from `CopyIntoBatches`: large external-table batches are split into 8192-row destinations, varlen payloads can be copied differently for whole, const, and shared-area vectors, and allocator capacity is rounded per destination. The mismatch could either admit an allocation that later exceeded its reservation or reject memory that actually fit.

This change:

- caps each new destination batch at `DefaultBatchSize`;
- projects destination data and varlen-area capacity using the same segmentation and `GrowCapacity` behavior as the copy path;
- accounts for partial-tail replacement peak and retained delta without empirical payload multipliers;
- adds regression coverage for 50K-row external batches, const/shared-area varlen vectors, exact multi-segment capacity, partial tails, and false-admission boundaries.

Verification:

- issue #26413 real Parquet self-join succeeds with `part_id` 1 and 7 returning 50,000 rows each;
- focused changed tests pass under `-race -count=100`;
- full `colexec` and `hashbuild` packages pass under race;
- `colexec`, `hashbuild`, and `spillutil` package tests pass;
- owning packages pass `go build` and `go vet`.


XuPeng-SH (approved):
Deep-reviewed exact head `08259098d46ab0991f797cfa3345d285ed9a2356`. The projection now follows the actual copy path: every new destination is capped at 8192 rows, fixed/header and varlen-area capacities use `mpool.GrowCapacity`, whole-source/shared-area and const vectors match their respective `UnionBatch` branches, and partial-tail admission covers replacement peak plus retained deltas before subsequent segments. Reservation cleanup/reconciliation and operator reuse remain balanced.

I also probed a reused const-varlen area against actual allocation; the copy path materializes only the live const payload as projected, so no hidden under-bound was found. Changed tests passed under race, the full `pkg/sql/colexec/hashbuild` package passed, and the dependent `TestBatches` path passed with the deterministic CGo wrapper. CI is green.

PR #26439, opened 2026-07-30
fix: preserve quoted foreign key names in database copies

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26144

- Preserve the exact bytes of foreign-key catalog values when MatrixOne builds internal SQL, including backslashes and apostrophes.
- Use the scanner-compatible literal contract consistently for FK catalog insert, lookup, delete, rename, transfer, and reference-check paths.
- Add an authenticated embedded regression that exercises both Data Branch and CLONE, checks persisted catalog bytes, proves FK enforcement, and cleans up every database.
- Add a dedicated distributed BVT case; keeping it separate avoids the mo-tester command-boundary ambiguity exposed by the earlier current-head CI run.

Local exact-head validation (`1168157ca44595184c5fce5e46af913711169be7`):

- focused planner literal tests: `-race -count=100`
- focused authenticated Data Branch/CLONE regression: `-race -count=1`
- full `pkg/sql/plan` and `pkg/tests/issues`: `-race -count=1`
- clean-cache compile-only build for both packages
- `go vet` for both packages
- BVT command/result structure checks and `git diff --check`
- preflight: `PASS review=PASS validation=PENDING`, diff hash `ecc445174e4df7b42fbefd29775bc4b8045a56a962db5482be60cbfc25947ae8`

GitHub CI for the rebased head remains pending. The issue is QA-gated and should stay open for testing after merge.


XuPeng-SH (approved):
Reviewed at 1168157ca44595184c5fce5e46af913711169be7. The foreign-key catalog SQL now matches the MatrixOne scanner contract for backslashes and apostrophes, and every mutation/query helper uses the same literal encoder. I verified the internal executor parses these statements in the default mode, then checked clone/branch catalog byte preservation and FK behavior. The planner package and the focused issue #26144 embedded-cluster test pass locally.

Issue #26441, opened 2026-07-30
[Bug]: TestIssue26114LegacyCrossAccountMetadataCountsTowardTargetQuota loses HAKeeper connection


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

c32538b011875a25514a9b9c647ef9a465b93614

### Other Environment Information

- Hardware parameters: GitHub Actions Ubuntu/x86 runner
- OS type: Linux/amd64
- Others: MatrixOne ALL CI, UT job 90771135520

### Actual Behavior

`TestIssue26114LegacyCrossAccountMetadataCountsTowardTargetQuota` failed while
creating the target account:

```text
Error 1105 (HY000): writeto tcp4 127.0.0.1:50668->127.0.0.1:15220:
read tcp4 127.0.0.1:50668->127.0.0.1:15220: read: connection reset by peer
```

Immediately before the assertion failure, the embedded cluster logged
LogService proposal timeouts, failed TN/CN heartbeats, and a reset HAKeeper
backend connection to `127.0.0.1:15220`.

The failure was observed in PR #26407, whose diff only changes UPDATE/foreign
key planner code, planner tests, and foreign-key BVT cases. It does not change
this test or the account, embedded-cluster, HAKeeper, LogService, or MORPC
paths.

### Expected Behavior

The authenticated embedded cluster should remain available, and the account
creation in the test should complete without losing the HAKeeper/LogService
connection.

### Steps to Reproduce

1. Run the `pkg/tests/issues` UT package in the MatrixOne CI environment.
2. Execute
   `TestIssue26114LegacyCrossAccountMetadataCountsTowardTargetQuota`.
3. Observe whether the embedded cluster resets the connection to port 15220
   during `CREATE ACCOUNT`.

### Additional information

- Workflow run:
  https://github.com/matrixorigin/matrixone/actions/runs/30511041712
- Failed job:
  https://github.com/matrixorigin/matrixone/actions/runs/30511041712/job/90771135520
- Test duration before failure: 3.59 seconds
- Failure timestamp: 2026-07-30 03:56:09 UTC
- Test source: `pkg/tests/issues/issue_26114_test.go`

LeftHandCold:
https://github.com/matrixorigin/matrixone/actions/runs/34025410243/job/101510885837?pr=28280

LeftHandCold:
[...]

PR #26448, opened 2026-07-30
test: stabilize authenticated cluster heartbeats

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26441

The authenticated issue tests previously started a second persistent embedded
LogService/TN cluster in the same UT process. Under race-enabled CI scheduling
pressure, the two clusters competed for Dragonboat tick/proposal processing,
which could starve HAKeeper heartbeats and reset a foreground SQL connection.

This PR routes authenticated issue tests through the existing
`RunBaseClusterTests` cluster instead of creating a second cluster. Because
that base cluster is shared by all `RunBaseClusterTests` callers, the test-only
configuration changes apply to every base-cluster consumer:

- authentication remains enabled on the shared CN services;
- CN and TN HAKeeper heartbeat RPC deadlines are 30 seconds;
- HAKeeper CN and TN store expiration windows are 60 seconds, keeping the
  heartbeat deadline strictly below the liveness window;
- the Iceberg tenant test uses the tenant's `accountadmin` role now that the
  shared cluster enforces authentication.

`TestAuthenticatedTestsReuseBaseCluster` verifies cluster identity and the
effective CN, TN, and LogService timeout configuration. Production defaults,
SQL behavior, assertions, and retry semantics are unchanged.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -race ./pkg/tests/issues -run
  '^TestAuthenticatedTestsReuseBaseCluster$' -count=1`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race ./pkg/tests/issues -run
  '^TestIssue26114LegacyCrossAccountMetadataCountsTowardTargetQuota$' -count=10`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race ./pkg/embed -run
  '^TestIcebergSQLEngineEmbeddedWriteDMLAndMaintenanceSQL$' -count=1`
- `make build`
- `make static-check`
- `git diff --check`


XuPeng-SH (changes requested):
Deep review of exact head `f652ffc5cc8418ed85029ec601a47cc63d9fea35`.

[P1] The current head no longer applies the heartbeat-timeout fix described by the PR — `pkg/embed/testing.go:67`

The first patch in this branch set both CN and TN HAKeeper heartbeat deadlines to 30 seconds and tested those values, but `91f4392b` removed that configuration and its coverage when the authenticated tests were switched to `RunBaseClusterTests`. The final `startBasicCluster` calls only `WithCNCount(3)` and `WithTesting()`; it does not call the `WithHAKeeperHeartbeatTimeout` option now available on main. Consequently `cluster.createServiceOperators` performs no override, and CN/TN validation retains the same 3-second defaults implicated by #26441. `TestAuthenticatedTestsReuseBaseCluster` asserts only pointer identity and `SkipCheckUser`; it cannot detect this regression.

Reusing one cluster may reduce scheduling pressure, and the focused issue test passed once locally under `-race`, but that is not deterministic closure for an intermittent heartbeat deadline failure. Please apply the intended timeout to the shared base cluster (preferably through `WithHAKeeperHeartbeatTimeout(30*time.Second)`) and assert the effective CN and TN service configs, or otherwise add deterministic evidence that the original timeout path is eliminated. Please also update the PR description to match the final strategy if cluster reuse is retained.

Local checks on this head: the reuse test and `TestIssue26114LegacyCrossAccountMetadataCountsTowardTargetQuota` pass under `-race -count=1`; embed build, embed/issues vet, and `git diff --check` pass. CI was still running at submission time.

XuPeng-SH (approved):
Re-reviewed the updated PR at a5d23de54752df126509e8b10cc68035c15f9ab8. The previous blocker is resolved: the shared embedded cluster now explicitly sets a 30s HAKeeper heartbeat and 60s CN/TN store timeouts, and the regression test inspects every CN/TN/LOG service and asserts the required ordering rather than only pointer reuse/auth state. The focused real-cluster test and full pkg/embed package passed under -race, and exact-head required CI is green.

PR #26450, opened 2026-07-30
fix(join): canonicalize DOUBLE signed zero across spill modes

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

Related to #26432. This is an incremental contract fix; it intentionally does **not** close the whole issue.

Spill is not a separate equality domain. A join key that is equal under SQL must remain equivalent in every physical phase: resident hashmap lookup, initial spill partitioning, probe partitioning, and recursive re-spill.

This PR fixes one concrete violation of that invariant: SQL `DOUBLE +0 = -0` was true in the scalar evaluator, while raw IEEE-754 bytes sent the two values to different hashmap keys or spill partitions.

The implementation defines signed-zero canonicalization once in `keycodec` and uses the same typed contract in all CN-local phases:

- resident `IntHashMap` keys;
- resident composite `StrHashMap` keys;
- initial build-side spill partitioning;
- probe-side partitioning and recursive re-spill.

Resident hashmaps dispatch once per vector and retain specialized row loops. Initial spill and re-spill call the same vector-level partition-hash implementation. Common non-`DOUBLE` loops are unchanged; there is no per-row interface dispatch or allocation.

Only the ephemeral hashmap/partition key is canonicalized. Input vectors, result values, and serialized spill batches retain their original values and format, so this introduces no spill-file or wire-format change.

## Contract status

Each pending case first executes the real scalar `=` evaluator as the SQL oracle, then skips before asserting the known-incomplete physical behavior. The PR does not bless a current mismatch as expected output.

| Type / values | Scalar SQL `=` | Resident | Initial spill | Re-spill | Status |
| --- | --- | --- | --- | --- | --- |
| `DOUBLE +0 / -0` | `TRUE` | match | match | match | **fixed and active** |
| scaled `FLOAT32(scale=2) 1.234 / 1.23` | `TRUE` | pending | pending | pending | skipped with #26432 reason |
| `JSON 1 / 1.0` | `TRUE` | pending | pending | pending | skipped with #26432 reason |
| `VECF32 +0 / -0` | `TRUE` | pending | pending | pending | skipped with #26432 reason |
| same-payload `DOUBLE NaN / NaN` | `FALSE` | pending | pending | pending | skipped with #26432 reason |
| `DOUBLE NULL / NULL` | `NULL` | no match | no match | no match | **active and correct** |

`FLOAT64` is not currently eligible for the planner's remote-shuffle key path. This PR establishes the local resident/spill/re-spill invariant; remote typed-key equivalence and the remaining types stay tracked by #26432.

## Regression design

These are deterministic operator-level tests, not large-data BVTs or embedded-cluster SQL. A low spill threshold selects the physical mode with 257 build rows plus one probe row, so runtime and memory remain small while the same batches traverse real join operators and spill files.

| Operator | Exact oracle checked in resident, initial-spill, and re-spill modes |
| --- | --- |
| HashJoin | One `-0` probe must return exactly the 129 distinct payloads belonging to the skewed `+0/-0` build equivalence class—no loss, duplicate, wrong build row, or NULL payload. |
| DedupJoin | The exact 257-row build multiset is preserved; only the signed-zero row receives probe capture `42`, with exact capture NULL states. |
| RightDedupJoin | The signed-zero pair returns `ErrDuplicateEntry` for `contract_key`, with zero output before and at the error. |

The fixtures also prove that:

- the signed-zero target starts in the second hashmap chunk (`start > 0`);
- resident mode does not spill;
- initial-spill mode increments depth 1 but not depth 2;
- re-spill mode increments both depths, and the target bucket itself meets the re-spill threshold;
- operator cleanup returns memory-budget usage, spill-disk usage, open spill FDs, and mpool bytes to zero.

A public hashmap-iterator contract matrix covers both `IntHashMap` and composite `StrHashMap`: flat vectors with a non-zero start, nullable vectors with and without actual NULLs, NULL filtering, NULL-as-key, constants, and constant NULLs. It asserts `zValues`, live group IDs, and `GroupCount` across insert and find operations.

## Scope and performance

Microbenchmark setup: latest `main` versus this branch, identical harness, `GOMAXPROCS=1`, five samples/median, 8192 spill rows and 256 resident hashmap rows.

| Path | latest `main` ns/row | this PR ns/row | delta |
| --- | ---: | ---: | ---: |
| spill `INT64` | 4.824 | 4.861 | +0.8% |
| spill `DOUBLE` | 4.788 | 2.660 | -44.4% |
| resident IntHashMap `INT64` | 2.879 | 2.882 | +0.1% |
| resident IntHashMap `DOUBLE` | 2.941 | 2.814 | -4.3% |
| resident StrHashMap composite `INT64` | 7.869 | 7.942 | +0.9% |
| resident StrHashMap composite `DOUBLE` | 7.887 | 6.903 | -12.5% |

The common integer paths show no material regression. The `DOUBLE` specializations are faster than the previous raw-byte path while implementing SQL-equivalent hashing.

## Test plan

All commands use the repository's deterministic CGo/libmo wrapper and were rerun after rebasing onto `origin/main` at `38ce3a774a`.

Focused race stress, all pass with `-race -count=100`:

- `TestHashJoinKeyEqualityContract`
- `TestDedupJoinDoubleSignedZeroContract`
- `TestRightDedupJoinDoubleSignedZeroContract`
- `TestHashMapsFloat64SignedZeroContract`

Complete non-race and `-race -count=1` suites both pass for all owning packages:

- `pkg/common/hashmap`
- `pkg/sql/colexec/hashbuild`
- `pkg/sql/colexec/spillutil`
- `pkg/sql/colexec/hashjoin`
- `pkg/sql/colexec/dedupjoin`
- `pkg/sql/colexec/rightdedupjoin`

`go vet` and `go build -mod=readonly` pass for the same six packages. The new hashmap state matrix directly exercises the `FLOAT64` const, nullable, NULL, non-zero-start, and composite-key branches that the prior CI coverage review identified.

### DIFF
--- a/pkg/common/hashmap/inthashmap.go
+++ b/pkg/common/hashmap/inthashmap.go
@@ -17,8 +17,10 @@ package hashmap
 import (
 	"bytes"
 	"io"
+	"math"
 	"unsafe"
 
+	"github.com/matrixorigin/matrixone/pkg/common/hashmap/keycodec"
 	"github.com/matrixorigin/matrixone/pkg/common/mpool"
 	"github.com/matrixorigin/matrixone/pkg/container/hashtable"
 	"github.com/matrixorigin/matrixone/pkg/container/types"
@@ -99,6 +101,9 @@ func (itr *intHashMapIterator) encodeHashKeys(vecs []*vector.Vector, start, coun
 			fillKeys[uint32](itr, vec, 4, start, count)
 		case 8:
 			fillKeys[uint64](itr, vec, 8, start, count)
+			if vec.GetType().Oid == types.T_float64 {
+				normalizeFloat64HashKeys(itr, vec, start, count)
+			}
 		default:
 			if !vec.IsConst() && vec.GetArea() == nil {
 				fillVarlenaKey(itr, vec, start, count)
@@ -106,6 +111,21 @@ func (itr *intHashMapIterator) encodeHashKeys(vecs []*vector.Vector, start, coun
 				fillStrKey(itr, vec, start, count)
 			}
 		}
+	}
+}
+
+func normalizeFloat64HashKeys(itr *intHashMapIterator, vec *vector.Vector, start, n int) {
+	if vec.IsConstNull() {
+		return
+	}
+	nsp := vec.GetNulls()
+	for i := 0; i < n; i++ {
+		if nsp.Contains(uint64(i + start)) {
+			continue
+		}
+		valueOffset := itr.keyOffs[i] - uint32(types.T_float64.TypeLen())
+		ptr := (*uint64)(unsafe.Add(unsafe.Pointer(&itr.keys[i]), valueOffset))
+		*ptr = keycodec.CanonicalFloat64Bits(math.Float64frombits(*ptr))
 	}
 }
 
--- a/pkg/common/hashmap/inthashmap.go
+++ b/pkg/common/hashmap/inthashmap.go
@@ -17,7 +17,6 @@ package hashmap
 import (
 	"bytes"
 	"io"
-	"math"
 	"unsafe"
 
 	"github.com/matrixorigin/matrixone/pkg/common/hashmap/keycodec"
@@ -100,9 +99,10 @@ func (itr *intHashMapIterator) encodeHashKeys(vecs []*vector.Vector, start, coun
 		case 4:
 			fillKeys[uint32](itr, vec, 4, start, count)
 		case 8:
-			fillKeys[uint64](itr, vec, 8, start, count)
 			if vec.GetType().Oid == types.T_float64 {
-				normalizeFloat64HashKeys(itr, vec, start, count)
+				fillFloat64Keys(itr, vec, start, count)
+			} else {
+				fillKeys[uint64](itr, vec, 8, start, count)
 			}
 		default:
 			if !vec.IsConst() && vec.GetArea() == nil {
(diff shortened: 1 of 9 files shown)


aptend (approved):
Reviewed the signed-zero canonicalization across resident, initial-spill, and re-spill paths. Focused contract tests passed with -race -count=100, and all six owning packages passed with -race -count=1.

VioletQwQ-0 (commented):
Reviewed the current exact head d234f5a060d5615e382483623f03592dcc7fc534 from multiple angles. I did not find a production-correctness blocker in the DOUBLE signed-zero canonicalization across resident, initial-spill, and re-spill paths.

The only remaining validation blocker is the required Matrixone Utils CI / Coverage check: PR coverage is 61.48%, below the 75% threshold. Please add focused coverage for the new FLOAT64 specialized branches, especially const, const-null, nullable/mixed-NULL, non-zero start, and const/short-vector hashing with a non-zero seed, then re-request review after the check is green.

VioletQwQ-0 (approved):

XuPeng-SH (commented):
Deep review of exact head `73a335ae27d590f0b7a0820f2877273593cf4a03`: no blocking findings.

Signed-zero canonicalization is defined once in the typed key codec and is consistently consumed by resident IntHashMap/composite StrHashMap, initial spill partitioning, and recursive re-spill. It preserves NULL behavior, non-zero starts, const vectors, original payload values, and spill-file format. The intentionally incomplete equality cases remain explicitly skipped and tracked by #26432 rather than being blessed as expected behavior.

Fresh local evidence: hashmap, HashJoin, DedupJoin, and RightDedupJoin signed-zero contracts all pass under `-race -count=3` across resident/initial-spill/re-spill modes, including cleanup counters; `go build` and `go vet` pass for all six affected packages. Final diff is clean, latest CI is green, and the PR remains MERGEABLE.

Issue #26445, opened 2026-07-30
[Bug]: TAE Merge: failures before LogTxnEntry leak transfer slab and runtime transfer state


LeftHandCold:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

This is distinct from:

- #26376, which tracks an allocation error being converted into a TN process
  panic before Merge output is published.
- #26377, which tracks the deadline and work-budget gaps in phase-2 tombstone
  scanning after the Merge transaction entry has been registered.

This issue tracks the ownership gap between `NewMergeObjectsEntry` and
`LogTxnEntry`.

## Branch Name

`main`

## Commit ID

`66142a6f7dee80a9cdbcaea1b035b812889e3745`

## Other Environment Information

- Reproduced with the real TAE test engine on macOS.
- The reproduction used the existing `FJ_TransferErrorAfterTransfer` fault
  point at a production-reachable error boundary.
- The same code exists on the discovery branch
  `agent/issue-24853-tiered-lifecycle-design` at
  `f6de5a90928e4f79a2bc204664d099b174548f0d`.

## Actual Behavior

`HandleMergeEntryInTxn` constructs a `mergeObjectsEntry` before registering it
with `txn.LogTxnEntry`.

During construction, `NewMergeObjectsEntry` may already:

- transfer concurrent tombstones;
- add entries to `Runtime.TransferDelsMap`;
- prepare and install runtime transfer pages;
- retain ownership of the off-heap `TransferTable` slab.

If phase 1, transfer-page preparation, or `LogTxnEntry` fails, the entry has not
been registered. Transaction rollback therefore does not call the entry's
`PrepareRollback`, which is currently the only path that releases the transfer
slab and removes the runtime transfer state.

A targeted real-TAE fault-injection test confirmed:

```text
merge snapshot starts
-> a concurrent DELETE commits
-> merge phase 1 transfers the DELETE and writes TransferDelsMap
-> FJ_TransferErrorAfterTransfer returns an error
-> NewMergeObjectsEntry returns nil/error before LogTxnEntry
-> the merge transaction rolls back
-> TransferDelsMap state for the created Object remains
```

The same error path also misses `TransferTable.Release()`. The slab is backed
by the off-heap mpool and has no GC finalizer; losing the registered-entry
cleanup path can therefore retain a 4 MiB or 16 MiB quantized slab allocation.

The normal Merge task error handler deletes created Object files, so this report
does **not** claim that this specific path leaks published Object files.

[...]

PR #26451, opened 2026-07-30
fix: release merge transfer state before registration

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26445

`DoMergeAndWrite` transfers ownership of its off-heap `TransferTable` slab to
the merge task before `HandleMergeEntryInTxn` constructs and registers the
transaction entry. A committed large delete can be recorded after the task's
initial big-delete check but before `NewMergeObjectsEntry`; that constructor
then rejects the merge. Since no transaction entry was registered, the normal
`PrepareRollback` path never released the slab. The same ownership gap also
left transfer pages and `TransferDelsMap` entries behind for later constructor
failures.

This change gives transfer state one explicit owner at every boundary:

- `HandleMergeEntryInTxn` releases a table when it fails before construction.
- `NewMergeObjectsEntry` rolls transfer state back on every construction
  error, including the production big-delete rejection.
- A failed `LogTxnEntry` rolls back the unregistered entry's transfer state.
- Registered entries retain the existing commit/rollback lifecycle; rollback
  now reuses the same idempotent transfer-state cleanup. Output-object file
  deletion remains owned by the existing merge-task error path, so it is not
  duplicated.

The regression executes a real merge with a committed delete, fails after
phase-1 transfer, and asserts both that `TransferDelsMap` is empty and that
the transfer-slab mpool allocation returns to its baseline.

Verified with:

- focused regression: `TestRollbackMergeBeforeRegistrationCleansTransferState`
- adjacent merge rollback/transfer tests:
  `TestRollbackMergeInQueue`, `TestRollbackMergeOnTransferError`,
  `TestRollbackMergeAfterTransferredDeleteError`,
  `TestRollbackMergeBeforeRegistrationCleansTransferState`,
  `TestTransferInMerge`, and `TestTransferInMerge2`
- full `./pkg/vm/engine/tae/db/test` package
- `go test -race -count=100` for the focused regression and race tests for
  `./pkg/vm/engine/tae/tables/txnentries` and
  `./pkg/vm/engine/tae/tables/jobs`
- `go build` and `go vet` for the changed implementation packages

### DIFF
--- a/pkg/vm/engine/tae/db/test/db_test.go
+++ b/pkg/vm/engine/tae/db/test/db_test.go
@@ -12168,6 +12168,77 @@ func TestRollbackMergeAfterTransferredDeleteError(t *testing.T) {
 	require.Equal(t, 3, rel.GetMeta().(*catalog.TableEntry).ObjectCnt(false) /*Aobj(created + deleted), Nobj(rollbacked)*/)
 }
 
+func transferSlabCurrBytes(t *testing.T) int64 {
+	t.Helper()
+	var reports []map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal([]byte(mpool.ReportMemUsage("transfer-slab")), &reports))
+	for _, report := range reports {
+		raw, ok := report["transfer-slab"]
+		if !ok || string(raw) == `""` {
+			continue
+		}
+		var stats struct {
+			CurrBytes int64 `json:"currBytes"`
+		}
+		require.NoError(t, json.Unmarshal(raw, &stats))
+		return stats.CurrBytes
+	}
+	return 0
+}
+
+func TestRollbackMergeBeforeRegistrationCleansTransferState(t *testing.T) {
+	ctx := context.Background()
+
+	opts := config.WithLongScanAndCKPOpts(nil)
+	tae := testutil.NewTestEngine(ctx, ModuleName, t, opts)
+	defer tae.Close()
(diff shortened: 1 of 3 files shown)


XuPeng-SH (approved):
Deep review completed on exact head `58f4e1cf141d3595b9572d49b5462f0aa854e185`.

The ownership transfer is closed at each boundary: `HandleMergeEntryInTxn` owns the slab before construction; `NewMergeObjectsEntry` owns and rolls back all transfer state once the entry exists, including constructor errors; successful `LogTxnEntry` transfers lifecycle ownership to the registered transaction entry. Registration failures occur before the entry append and invoke the same idempotent cleanup. Output-object deletion remains with the task/registered rollback path, so it is not duplicated.

| Audit | Result |
|---|---|
| Q1 cleanup on error | Closed for pre-constructor, constructor/phase-1, registration, and registered rollback failures |
| Q2 wait termination | No new wait edge; cleanup only releases the slab and removes runtime map/page references |
| Q3 accumulation | Repeated failures no longer retain slab allocations or `TransferDelsMap`/active page entries; physical page close remains on the existing bounded TTL queue |

Validation passed: the focused real-TAE regression, adjacent merge rollback/transfer tests, focused race x10, txnentries/jobs tests, package builds, `go vet`, and `git diff --check`.

PR #26455, opened 2026-07-30
fix: align expression memory with hash build budget

### LINKED ISSUE #26433: [Design]: align ExpressionExecutor retained capacity with HashBuild budget lifetime
## Context

Follow-up from #26408 and #26412. #26412 closes spill-payload ownership and several concrete cleanup leaks, but expression executors can retain reusable vectors/capacity outside the lifetime currently represented by HashBuild memory reservations.

## Problem

Budget accounting must describe memory that remains physically retained, not only the logical rows or batches currently being processed. If an `ExpressionExecutor` keeps vector capacity across calls, resets, or execution generations while the corresponding HashBuild reservation is released or recomputed independently, admission can undercount memory and later growth can exceed the intended query envelope.

The inverse error is also possible: charging retained capacity twice after reuse can reject work unnecessarily.

## Proposed direction

- Give retained executor buffers an explicit owner and generation.
- Couple reservation acquisition, resize, reuse, reset, and release to the actual retained capacity lifetime.
- Define transfer rules when executor results/vectors are borrowed or moved into another operator.
- Expose retained versus active bytes in diagnostics without double counting shared memory.

## Acceptance criteria

- Capped-budget tests cover initial allocation, growth, reuse, Reset, Free, cancellation, expression failure, and repeated execution generations.
- Accounted bytes track retained vector capacity within a documented bounded delta.
- Cleanup returns both mpool usage and budget reservations to baseline exactly once.
- Reuse does not double-charge existing capacity and cannot consume unreserved growth.
- Tests include HashJoin, DedupJoin, and RightDedupJoin expression-key paths.

Related: #26408, #26412, and the broader resource envelope in #25782.

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26433

Expression executors retain reusable vector capacity across batches and resets. The previous HashBuild accounting replaced expression reservations independently from that physical lifetime, which could either undercount retained memory or double-charge reused capacity.

This PR:

- introduces an explicit per-root expression memory lease tied to one HashBuild budget generation;
- admits constant executor allocations before construction and reconciles them to retained capacity;
- reserves only uncovered allocate-copy-free overlap when an executor grows, while retaining the admitted high-water mark;
- accounts for IFF/CASE/COALESCE selected-result and selected-parameter scratch;
- applies the same ownership model to initial spill, probe scatter, recursive re-spill, HashJoin, DedupJoin, and RightDedupJoin;
- releases failed builder executors before replacement re-spill admission;
- adds retained-versus-borrowed ownership accounting and lifecycle regression coverage.

Design boundary: this lease covers expressions participating in the shared HashBuild/spill working set. Resident probe expressions remain under normal process/mpool accounting; this is not a general query-memory budget.

Validation:

- related six packages pass normal and race test suites;
- all 21 new or directly modified behavioral tests pass race stress with count=100;
- go build and go vet pass for all owning packages;
- hashbuild package coverage: 81.4%;
- final independent review converged with no findings.


XuPeng-SH (changes requested):
[P1] Include the narrow 16-bit array payload width in the expression peak bound

`expressionTypePeak` in `pkg/sql/colexec/hashbuild/hashmap.go` treats every non-allowlisted varlena result as at most `MaxVarcharLen` (65,535 bytes per row). `T_array_bf16` and `T_array_float16` are missing from the array cases, but each permits 65,535 elements at 2 bytes per element, i.e. 131,070 payload bytes per row.

I reproduced the resulting under-admission on this exact head with `iff(bool_col, vecbf16_col1, vecbf16_col2)`, two rows, and legal 65,535-element arrays in both branches. `expressionVectorPeak` and the lease reserve 196,670 bytes, while `ExpressionMemoryLease.Retained()` is 270,384 bytes after `Eval` returns nil; the budget generation still reports only 196,670 bytes used. This lets HashBuild/spill expression memory exceed the hard budget this PR is intended to enforce.

Please derive array payload bounds from element width (or otherwise include all narrow array types) and add a max-width `vecbf16`/`vecf16` regression that asserts retained capacity never exceeds the admitted reservation.

XuPeng-SH (approved):
Deep-reviewed current head 5c5280b625cafa7b1b07015ae0767bc9f8e6a204.

The previous P1 is resolved: array result bounds are now derived from the element width, including BF16/F16, and the max-width regressions verify that retained expression memory stays within the admitted reservation. I also checked constructor admission, high-water/replacement overlap, error cleanup, spill/re-spill ownership, Reset/Free generation boundaries, and the three join variants. No further blocking issues found.

Validation: all six affected packages pass normal tests and race tests; all 26 directly relevant tests pass individually with -race -count=100; go build and go vet pass; CI is green.

Issue #26456, opened 2026-07-30
bug: ALTER TABLE panic — pitr_length 类型不匹配（DDL tinyint unsigned vs 代码 int64）


Ariznawlll:
## 问题

PR #25882 引入的 `loadAlterDataBranchHistoricalSourcesWithQuery` 中，`pitr_length` 列的类型读取与表 DDL 定义不一致，导致 `ALTER TABLE` 时 panic。

## CI

- **Run**: https://github.com/matrixorigin/mo-nightly-regression/actions/runs/30472882209/job/90822782976
- **Workflow**: MO Main Nightly Regression On TKE — MOTR Regression Test Suite
- **MO Commit**: `2e68774061`
- **影响**: 7 个测试用例同时 panic（所有涉及 ALTER TABLE 的测试）

## Panic

```
panic unsafe slice cast: from length 8 is not enough for length 16

github.com/matrixorigin/matrixone/pkg/common/util.UnsafeSliceCastToLength
    pkg/common/util/unsafe.go:89
github.com/matrixorigin/matrixone/pkg/container/vector.MustFixedColNoTypeCheck[int64]
    pkg/container/vector/tools.go:51
github.com/matrixorigin/matrixone/pkg/sql/compile.loadAlterDataBranchHistoricalSourcesWithQuery.func2
    pkg/sql/compile/alter.go:535                        ← panic here
...
github.com/matrixorigin/matrixone/pkg/sql/compile.(*Scope).dropTableSingle
    pkg/sql/compile/ddl.go:3681
github.com/matrixorigin/matrixone/pkg/sql/compile.(*Scope).DropTable
    pkg/sql/compile/ddl.go:3335
```

## 根因

**类型不匹配**：

| 位置 | 类型 | 每元素大小 |
|---|---|---|
| **DDL** `predefined.go:154` | `pitr_length tinyint unsigned` | **1 byte** |
| **代码** `alter.go:535` | `MustFixedColNoTypeCheck[int64](cols[5])` | **8 bytes** |
| **Mock** `mock.go:556` | `T_int64` | 8 bytes（也与 DDL 不一致） |

代码用 `int64`（8 字节）对 `cols[5]` 做 unsafe slice cast，但实际 `mo_pitr.pitr_length` 是 `tinyint unsigned`（1 字节）。当 mo_pitr 有活跃 PITR 任务时，查询返回的向量按 1 字节存储，cast 时字节数不够直接 panic。

## 影响范围

[...]

PR #26457, opened 2026-07-30
fix: read PITR length using catalog type

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26456

`mo_catalog.mo_pitr.pitr_length` is persisted as `tinyint unsigned`, but the compile-layer DATA BRANCH lineage loader reinterpreted its result vector as `int64`. With multiple active PITR rows this requested more bytes than the `T_uint8` vector contained and panicked during lineage compaction.

This PR keeps the catalog DDL and the supported `1..100` range unchanged. It:

- reads `pitr_length` as checked `uint8` before the lossless conversion to `int`;
- aligns the planner catalog mock and compile test fixtures with the production `T_uint8` representation;
- covers direct loading, active/expired lineage retention, the background executor path, and delete-error propagation;
- adds a canonical DATA BRANCH + database PITR + COPY ALTER BVT case.

### Validation

- The new direct loader test reproduces the original panic before the production fix: `unsafe slice cast: from length 8 is not enough for length 16` at `pkg/sql/compile/alter.go:535`.
- After rebasing onto `upstream/main` at `11eaec8833e79b3b178ebeb9b6b2bb43fd6b679e`, `range-diff` showed all three PR patches are exactly equivalent.
- Focused compile tests passed with `-race -count=20`; full `./pkg/sql/compile` and `./pkg/sql/plan` tests passed with `-race -count=1` through the repository CGo wrapper.
- `git diff --check` passed.
- Superseded-head run `30538612665` passed SCA, Coverage, PROXY BVT, and PESSIMISTIC BVT. Ubuntu UT failed across unrelated cluster/lockservice tests because HAKeeper/TN startup timed out and a remote backend closed; the branch was rebased again instead of rerunning that superseded head.
- Exact-head preflight: `PASS review=PASS validation=PENDING` on `ea4e498bf086caed8a8feb1d984b5a1408df9540` (`diff_hash=67549bacba9eb7898120fa7d90b2b33a0d61acc11736ffb44d2bfed3f3b9a41a`).
- Local BVT execution is pending because `mo-tester` is unavailable on this machine; the SQL/result pair is included for exact-head CI.

### QA

QA required: yes

After merge, please validate PITR + DATA BRANCH + COPY ALTER in TKE or an equivalent environment and confirm that the original MOTR panic failure cluster no longer appears. Issue #26456 should remain open until that QA gate passes.

### DIFF
--- a/pkg/sql/compile/alter.go
+++ b/pkg/sql/compile/alter.go
@@ -532,7 +532,7 @@ func loadAlterDataBranchHistoricalSourcesWithQuery(
 	err = appendAlterDataBranchHistoricalSources(
 		res,
 		func(i int, cols []*vector.Vector) (int64, error) {
-			lengths := vector.MustFixedColNoTypeCheck[int64](cols[5])
+			lengths := vector.MustFixedColWithTypeCheck[uint8](cols[5])
 			units := executor.GetStringRows(cols[6])
 			return databranchutils.PitrRetentionLowerBound(now, int(lengths[i]), units[i])
 		},
--- a/pkg/sql/compile/alter_test.go
+++ b/pkg/sql/compile/alter_test.go
@@ -1175,6 +1175,55 @@ func newAlterCopyFixedResult[T any](t *testing.T, mp *mpool.MPool, typ types.Typ
 	return memRes.GetResult()
 }
 
+func TestLoadAlterDataBranchHistoricalSourcesUsesPitrCatalogType(t *testing.T) {
+	now := time.Date(2026, time.July, 17, 12, 0, 0, 0, time.UTC)
+	ctrl := gomock.NewController(t)
+	c := newAlterCopyPrecheckCompile(t, ctrl, &alterCopyInsertSpyExecutor{})
+	mp := c.proc.Mp()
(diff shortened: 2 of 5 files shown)


XuPeng-SH (approved):
Deep review of exact head `ea4e498bf086caed8a8feb1d984b5a1408df9540`: no blocking findings.

The compile-layer lineage loader now decodes `mo_pitr.pitr_length` using the catalog's actual `T_uint8` representation with a checked vector accessor, then performs the existing lossless conversion to `int`. The planner mock and all direct/background compaction fixtures use the same physical type; coverage includes multiple rows, the maximum supported retention value, active/expired retention, and delete-error propagation.

Fresh local evidence: the PITR loader and compaction regressions pass under `-race -count=10`; `go build`, `go vet`, and `git diff --check` pass for compile/plan. Latest CI is green and the PR is MERGEABLE.

heni02 (approved):

Issue #26372, opened 2026-07-29
[Bug]: SHOW WARNINGS returns accumulated session errors instead of diagnostics for the latest statement


Ariznawlll:
## Description

`SHOW WARNINGS` (and `SHOW ERRORS`, which shares the same execution path) returns errors accumulated over the lifetime of the current session instead of diagnostics produced by the most recent non-diagnostic statement.

As a result, a later DML failure is reported together with unrelated errors from earlier statements. This makes it difficult for clients, automated retry logic, and operators to identify the actual cause of the current failure.

The MySQL documentation states that `SHOW WARNINGS` displays conditions resulting from the most recent non-diagnostic statement in the current session:
https://dev.mysql.com/doc/refman/8.0/en/show-warnings.html

## Environment

- MatrixOne `main`: `f6dab28046d70412cec132f0068840896852101c`
- Local single-CN deployment
- macOS arm64
- MySQL Client 8.3.0

## Reproduction

```sql
create database if not exists unhappy_show_warnings;
use unhappy_show_warnings;

create table t(id int primary key, uk int unique);
insert into t values (1, 10), (2, 20);

-- First explicit failure: missing table
select * from explicit_history_marker_20260729;

-- Most recent non-diagnostic statement: unique-key conflict
update t set uk = 20 where id = 1;

show warnings;
select 1 as health_after_show_warnings;
```

The batch client must continue after errors (for example, `mysql --force`) so that `SHOW WARNINGS` is executed after both failures.

## Actual Result

`SHOW WARNINGS` returns the latest unique-key conflict, the earlier missing-table error, and an even earlier connection-probe parser error sent by the MySQL client:

```text
Level  Code  Message
Error  1062  Duplicate entry '20' for key 'uk'
Error  1064  SQL parser error: table "explicit_history_marker_20260729" does not exist
Error  1064  ... syntax error ... near " $$"
```

The connection remains usable and `health_after_show_warnings` returns `1`.

## Expected Result

`SHOW WARNINGS` should return only the conditions produced by the most recent non-diagnostic statement, which is the failed `UPDATE`:

```text
Error  1062  Duplicate entry '20' for key 'uk'
```

Errors produced by earlier statements must not be included in the current diagnostics area.

## White-box Analysis

The current implementation maintains an accumulating error array on each session:

- `pkg/frontend/mysql_protocol.go`: `sendErrPacket` calls `ses.GetErrInfo().push(...)` for every error.
[...]

PR #26460, opened 2026-07-30
fix(frontend): scope diagnostics to latest statement

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26372

- resets the session diagnostics area at each top-level non-diagnostic statement boundary, including parse and protocol paths that return before normal execution
- preserves diagnostics across `SHOW WARNINGS` / `SHOW ERRORS` and internal, migration, or derived statements
- returns a locked deep snapshot to diagnostics consumers instead of exposing mutable session slices
- adds focused unit coverage and an unskipped BVT for missing-table error -> unique-key `UPDATE` error -> repeated `SHOW WARNINGS` -> successful-statement clearing

Validation:

- seven affected tests independently passed `-race -count=100`
- `go test -race -count=1 ./pkg/frontend`
- `go build ./pkg/frontend/...`
- `go vet ./pkg/frontend/...`
- exact-head single-node MySQL protocol scenario passed
- preflight: `PASS review=PASS validation=PENDING`

QA is required after merge because this changes user-visible MySQL diagnostics behavior.

### DIFF
--- a/pkg/frontend/mysql_cmd_executor.go
+++ b/pkg/frontend/mysql_cmd_executor.go
@@ -1106,7 +1106,7 @@ func doShowErrors(ses *Session, execCtx *ExecCtx) error {
 	mrs.AddColumn(CodeCol)
 	mrs.AddColumn(MsgCol)
 
-	info := ses.GetErrInfo()
+	info := ses.diagnosticsSnapshot()
 
 	for i := info.length() - 1; i >= 0; i-- {
 		row := make([]interface{}, 3)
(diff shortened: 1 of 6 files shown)


XuPeng-SH (approved):
Deep review of exact head `fea1db7a0d64a7636e75b44df65e818b979eccb0`: no blocking findings.

The diagnostics area is reset at each top-level non-diagnostic statement boundary, including parse/prepare/sidecar/rewrite/execute paths that can return before normal execution, while SHOW WARNINGS/ERRORS and internal/migration/derived statements preserve the preceding client diagnostics. Error recording and reads now share the session lock, and consumers receive a deep snapshot rather than mutable backing slices. Multi-statement behavior and repeated diagnostics queries remain coherent.

Fresh local evidence: all affected lifecycle/protocol regressions pass under `-race -count=10`; `go build ./pkg/frontend/...`, `go vet ./pkg/frontend/...`, and `git diff --check` pass. Latest CI is green and the PR is MERGEABLE.

Issue #26449, opened 2026-07-30
[Bug]: TestLockWithBindNotFound / TestUnlockWithBindNotFound flaky — lockservice bind lookup race


matrix-meow:
## Problem

`TestLockWithBindNotFound` (0.54s) and `TestUnlockWithBindNotFound` (6.83s) in `pkg/lockservice` fail intermittently. Both tests verify lock/unlock behavior when the lock table bind is not found. Under CI load, timing races cause failures.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-30 | [PR #26304](https://github.com/matrixorigin/matrixone/actions/runs/30519549544/job/90796776499) | UT Test on Ubuntu/x86 |

## Related Closed Issues

- #17414 — TestLockWithBindNotFound and TestUnlockWithBindNotFound failed (closed)
- #11105 — TestUnlockWithBindNotFound failed (closed)
- #10962 — TestUnlockWithBindNotFound and TestGetLockWithBindNotFound fail (closed)

## Suggested Fix

1. Increase bind lookup retry timeout for CI environments
2. Or add retry logic for transient bind-not-found conditions
3. Or use synchronization barriers instead of timing-based waits

PR #26461, opened 2026-07-30
test(lockservice): stabilize bind-not-found cases

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26449

Stabilizes `TestLockWithBindNotFound` and `TestUnlockWithBindNotFound` by:

- deterministically injecting the first `ErrBackendClosed` observed in CI so the regression no longer depends on scheduler timing
- retrying only transient backend-closed and connection errors within the existing bounded test context
- closing RPC clients before their remote servers so reconnect teardown cannot race goroutine leak detection

### DIFF
--- a/pkg/lockservice/service_remote_test.go
+++ b/pkg/lockservice/service_remote_test.go
@@ -1540,8 +1540,25 @@ func runBindChangedTests(
 		func(alloc *lockTableAllocator, s []*service) {
 			l1 := s[0]
 			l2 := s[1]
+			l3 := s[2]
 			ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
 			defer cancel()
+			// l2 owns RPC backends to l1 and l3. Close the client side before
+			// those servers; otherwise a backend reset can race leak detection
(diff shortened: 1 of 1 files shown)


XuPeng-SH (approved):
Deep review of exact head `8fa301f2b5c8cb8684462dd1f97d53fb98b66e18`: no blocking findings. The test now deterministically injects the observed first `Method_Lock` backend-close failure, retries only the two transient backend readiness errors within the existing context budget, and closes the owning RPC client before the remote servers. The original bind-not-found assertions remain intact rather than being weakened.

Fresh local evidence: `TestLockWithBindNotFound`, `TestUnlockWithBindNotFound`, and `TestGetLockWithBindNotFound` pass under `-race -count=5`; each run logs the injected `backend connection closed` path. `go build` and `go vet` pass for `pkg/lockservice`; final diff is clean, latest CI is green, and the PR remains MERGEABLE.

PR #26109, opened 2026-07-23
refactor: define canonical compute-group execution and harden ingress-owned scheduling

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [x] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

## Which issue(s) this PR fixes

Related to #25451. This does **not** close the canonical compute-group implementation gate.

## Outcome and scope

Rebased on `main` at `0b195e4f5a23c06a40552a8db3e7fdf182979a3a`.
Latest repair: `6a96463a66`.

This PR separates a **draft** canonical compute-group RFC/acceptance matrix from
independently useful legacy ingress-scheduling and LOCAL-upload correctness
fixes. It does not implement compute-group identity, authorization, admission,
or the future distributed drain protocol. The rejected workload-policy
DDL/catalog/RPC/cache and raw-label product model remain removed. RFC owner
approval and Milestone-B implementation/validation remain separate gates.

## Implementation

- Writable AP workspaces retain ingress participation. LOCAL input is
  ingress-only and is allocated only after successful compile/placement.
- General vector-index scan IR from current main retains the ingress/ordinal
  invariant; no resurrection of the removed IVF-only function-scan IR.
- Exact service ID wins over aliases; local routes use the real ingress
  address. Authoritative empty/excluded/draining pools fail closed rather than
  manufacturing local membership.
- Interrupted LOCAL upload now terminates at the **network protocol boundary**:
  cancel interrupts both pipe and network I/O; cleanup joins the watcher and
  producer before freeing buffers or reusing session/ExecCtx state.
- Removed unbounded discard-to-EOF and consecutive-error retries. Packet
  retries cannot reconstruct a partially consumed frame, and successful packets
  previously reset the error budget indefinitely.
- UDF file import shares the same repaired lifecycle. Storage write errors and
  panics cancel the producer before joining; its duplicate watcher is removed.
- Compatibility tradeoff: a failed/canceled upload before protocol EOF closes
  its client connection; clients must reconnect. Successful EOF retains normal
  connection reuse. No successful-upload size cap or per-row cache is added.

## RFC correction

DrainRequested publication is not a distributed claim barrier. The authority
must collect generation/incarnation-fenced receiver seals, serialized against
local/remote claim admission, before publishing effective Draining. A stale
Ready cache and valid lease cannot bypass an acknowledged seal. Missing acks
produce bounded pending/failure responses and retain DrainRequested, not false
[...]

PR #26472, opened 2026-07-30
fix(join): enforce typed-key equality across spill and runtime filters

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [x] Code Refactoring

Related to #26432. This is an incremental typed-key/spill contract PR and intentionally does **not** close the entire issue.

## Why

Spill is not a second equality implementation. If scalar SQL says two join keys are equal, resident join, initial spill, probe partitioning, recursive re-spill, and any pre-join runtime filter must preserve that match.

Two correctness gaps violated that rule:

1. Scaled `FLOAT32` scalar comparison normalizes each operand to its declared scale, but resident hash keys and spill partition keys previously used raw IEEE-754 bytes. SQL-equal values could land in different hash groups/buckets.
2. An exact runtime filter is optional but may run before the join in scans, zonemaps, and Bloom filters. Those consumers compare encoded payloads, not the final join equality. A stale or under-specified plan could therefore silently remove a row that the join itself would match. The concrete stale-plan shape is:

   ```
   probe:   DECIMAL(...,2), raw 100  = 1.00
   spec:    DECIMAL(...,3)
   payload: DECIMAL(...,3), raw 1000 = 1.000
   ```

   SQL `=` is true, while old raw `IN` compares `100 != 1000`.

### One typed-key contract for resident and spill

- `Float32ScaleNormalizer` is shared by scalar comparison and key encoding.
- Each operand is normalized with its **own** declared scale, so mixed-scale equality is symmetric.
- `Float32Codec` supplies the canonical resident/spill key and closes `+0/-0`.
- The same codec is used by `IntHashMap`, composite `StrHashMap`, initial build spill, probe partitioning, and recursive re-spill.
- Only ephemeral hash/partition keys are canonicalized. Input vectors and spill-file payloads retain their original values and format.
- Type dispatch and scale setup happen once per vector/comparison; there is no per-row allocation or interface dispatch.

### Exact runtime filters are versioned, selective, and fail open

The planner and all producers now share an explicit contract:

```
probe declared type
        ↕ SQL equality
declared build type
        ↕ physical payload
actual materialized build type
```

- `RuntimeFilterKeyEncoding` is an explicit versioned marker. Default, legacy, unknown, contradictory, or incomplete metadata publishes `PASS`.
- `RuntimeFilterSpec` carries the probe type, producer expression, and ordered component probe types for tuple encodings.
- HashBuild, IndexBuild, and FuzzyFilter validate the complete triangle before collecting or publishing an exact payload.
- Decimal scale is part of the contract. The real scale-2 probe / scale-3 spec+payload stale-plan case now fails open.
- Metadata-independent raw types such as integer and string retain the legacy expression as well as the v1 contract. They continue to produce exact filters on protocol v6 during rolling upgrade.
- Metadata-dependent decimal, signed-zero-closed float, serialized tuple, and FuzzyFilter contracts require protocol v8. An old producer sees no usable legacy expression and publishes `PASS`.
- Safe `FLOAT64` and unscaled `FLOAT32` exact filters remain enabled. If zero exists, the payload contains both physical `+0` and `-0`, so raw Bloom consumers cannot reject a SQL-equal zero.
- Scaled `FLOAT32` remains `PASS` for exact raw filters: its rounded SQL equivalence class cannot be represented by the current raw `IN`/zonemap/Bloom payload without false negatives.
- Planner statistics, dependencies, placement, and index rewrites are changed only after a filter is proven materializable. A filter that must be `PASS` no longer creates optimistic side effects.
- Cached/cross-version plans remain defensive at execution; an optional filter failure never turns into wrong results.

This is selective conservatism, not a blanket FLOAT runtime-filter disable.

### Spill/runtime-filter memory and unhappy paths

- Runtime-filter payloads use one shared budgeted marshaller and reserve actual wire/capacity costs before allocation.
- HashBuild accounts selected key slots only, actual vector capacity, varlen area copied by `UnionBatch`, reallocation overlap, tuple packer classes, scratch memory, and message-owned wire payloads.
- A whole-vector varlen copy now budgets the complete source area, including dead/stale area, before destination allocation.
- Optional filter collection or serialization failure abandons the optional state and publishes `PASS`; mandatory hash-map construction continues.
- Terminal success/error paths publish exactly one terminal message and release retained optional keys.
- Dedup's destructive retained-batch rewrite marks recovery unsafe before mutation. Admission failure re-spills only while retained rows are still replayable; otherwise it stops safely instead of duplicating or losing rows.
- Capacity checks return immediately when no vector growth is needed, avoiding repeated `chunks × retained-batches` scans.

### NaN scope is deliberately narrow

This PR does **not** define NaN join equality and does not redesign every zonemap predicate.

- Float min/max construction skips NULL and NaN.
- All-NaN/all-NULL input leaves the zonemap uninitialized.
- Legacy float zonemaps already poisoned by a NaN bound fail open only in the runtime-`IN` pruning chain (`AnyGEByValue`, `AnyLEByValue`, `SubVecIn`, `AnyIn`).
- Float `AnyIn`/`SubVecIn` uses the same total-order comparator as sorting.

This is the minimum compatibility fix needed to prevent a poisoned legacy bound from silently hiding an ordinary matching value.

## Contract status

Every active equality case uses the real scalar `=` evaluator as the oracle.

| Type / values | SQL equality | Resident | Initial spill | Re-spill | Exact runtime filter |
| --- | --- | --- | --- | --- | --- |
| `DOUBLE +0 / -0` | `TRUE` | match | match | match | enabled with zero closure on v8 |
| unscaled `FLOAT32 +0 / -0` | `TRUE` | match | match | match | enabled with zero closure on v8 |
| scaled/mixed-scale `FLOAT32` equal after declared-scale normalization | `TRUE` | exact match | exact match | exact match | `PASS` |
| nearby scaled `FLOAT32` values that normalize differently | `FALSE` | no match | no match | no match | `PASS` |
| equal-scale `DECIMAL` | `TRUE` | match | match | match | enabled on v8 |
| stale/mismatched-scale `DECIMAL` contract | may be `TRUE` | join remains authoritative | join remains authoritative | join remains authoritative | `PASS` |
| integer/string raw keys | normal SQL equality | unchanged | unchanged | unchanged | remains enabled on v6/v7/v8 |
| `NULL / NULL` | `NULL` | no match | no match | no match | NULL removed / conservative |
| `JSON 1 / 1.0` | `TRUE` | pending #26432 | pending | pending | pending |
| `VECF32 +0 / -0` | `TRUE` | pending #26432 | pending | pending | pending |
| `NaN / NaN` join semantics | `FALSE` | pending #26432 | pending | pending | only legacy runtime-IN zonemap poisoning addressed |

The regular planner still restricts remote shuffle join keys to its existing type set. This PR does not silently enable remote FLOAT shuffle.

## Test plan

The regression is small, deterministic operator execution. It does not require large BVT data or an embedded-cluster SQL fixture.

### Resident / initial spill / re-spill

- Use the real scalar evaluator as the oracle.
- Run the same mixed-scale `FLOAT32` inputs through all three physical modes.
- Low memory thresholds force spill with 257 build rows and one probe row.
- The true case returns exactly 129 distinct payload rows; the near-but-distinct case returns zero.
- The test proves the expected spill depths/buckets, row conservation, and zero retained mpool/spill-disk/open-FD state after cleanup.
- Shared HashJoin/HashBuild/spill paths are exercised directly; existing HashJoin, DedupJoin, and RightDedupJoin spill suites continue to cover the three operators without adding large fixtures.

### Runtime-filter correctness and rollout

- Real planner serialization into real HashBuild for integer, string, float-zero, decimal, and tuple contracts.
- Real scale-2 probe / scale-3 spec+payload regression proves scalar `=` is true, old raw `IN` is false, and the new producer publishes `PASS`.
- Legacy/default/unknown markers, forged function layouts, mismatched component order/type, missing payload, and old/new protocol directions all fail open.
- Safe integer/string v6 plans still publish exact filters.
- Safe float v8 plans publish both zero bit patterns; scaled float publishes `PASS`.
- IndexBuild and FuzzyFilter cover the same validation, allocation-failure, Reset, and terminal-message contracts.

### Budget, liveness, and recovery

- Whole-vector varlen dead-area admission is rejected before destination allocation.
- Wire payload, allocator size class, scratch, retained capacities, and reallocation overlap return to zero.
- Unsafe post-Dedup rewrite cannot re-spill; safe pre-rewrite admission can re-spill and conserves all rows.
- Terminal and cancellation paths cannot hang consumers or publish duplicate terminal messages.

### NaN compatibility

- `[NaN, 7]` and `[7, NaN]` build `[7, 7]`.
- all-NaN remains uninitialized.
- legacy NaN-poisoned bounds fail open for runtime `IN`.
- ordinary zonemaps still prune precisely with sorted lists containing NaN and ordinary keys.

### Local validation after merging latest main

All tests use `.agents/skills/mo-dev/scripts/mo-cgo-test`.

- Full related matrix passed for hashmap/keycodec/malloc/types/vector/zonemap/readutil; runtimefilter/hashbuild/indexbuild/fuzzyfilter/spillutil; HashJoin/DedupJoin/RightDedupJoin; plan/function/explain/compile/frontend.
- Targeted `-race` passed for runtimefilter, HashBuild, IndexBuild, FuzzyFilter, and spillutil.
- Eleven high-risk stale-plan/budget/re-spill/NaN/protocol regressions passed with `-count=10`.
- Proto bindings were regenerated from the merged definitions. Main's `gap_fill_mode=75` is preserved and the not-yet-released `fuzzy_build_side` moved to 76.
- Existing optimizer BVT changes are plan-golden updates only; no large-data BVT was added.

## Performance

No measurable non-spill regression was found.

Alternating precompiled A/B runs used one core, five paired one-second samples, and compared the same workloads before/after the runtime-filter/budget changes:

| Non-spill path | baseline | candidate | paired median |
| --- | ---: | ---: | ---: |
| cached integer build | 4,954 ns/op | 4,921 ns/op | `-0.02%` |
| cached string build | 8,996 ns/op | 9,106 ns/op | `+0.88%` |
| cold integer build | 10,164 ns/op | 10,110 ns/op | `-0.03%` |
| cold string build | 24,232 ns/op | 24,366 ns/op | `+0.38%` |

`bytes/op` and `allocs/op` were exactly unchanged:

- cached: `560 B/op`, `8 allocs/op`
- cold integer: `10,376 B/op`, `24 allocs/op`
- cold string: `24,152 B/op`, `539 allocs/op`

Float zonemap update at 10k rows measured `+0.53%` for FLOAT32 and `-1.27%` for FLOAT64, with unchanged allocation; both are within observed noise.

Why the hot path remains stable:

- integer/string row loops do not execute float normalization;
- contract checks run once per planned/built filter, not per probe row;
- safe float zero closure scans the already-unique optional payload once;
- unsafe/legacy filters stop before collecting, sorting, or serializing optional keys;
- no-growth retained-vector checks return before scanning batches.

The intentional tradeoff is pruning, not join execution: scaled `FLOAT32` or an unprovable stale contract may scan more rows because the optional exact filter publishes `PASS`. Correct matches are preserved, and safe float/integer/string filters remain enabled.


gouhongshen (approved):
## Codex automated review

No findings. The canonical FLOAT32 contract is consistently applied to resident Int/Str hashmaps and spill partition hashing, uses each operand’s declared scale, and matches the updated scalar comparison semantics. Runtime-filter changes fail open where raw encodings are unsound and preserve signed-zero closure for safe float plans. PR discussion contained no reviews or inline threads. `git diff --check` passed; focused CGo tests could not run because the checkout lacks `cgo/libmo.dylib`.

aptend (approved):
Re-reviewed the complete diff at ad3b393b. No blocking issues found; focused race tests and current CI checks pass.

XuPeng-SH:
@aptend @LeftHandCold @gouhongshen @aunjgr 帮我用最好的模型deep review, 改动有点虚：性能，正确性, unhappy等

gouhongshen (approved):
## Codex automated review

Prior review had no blocking findings at ad3b393b; the FLOAT key/spill and safe-float filter work remains accepted with no reintroduced blocker. The new versioned contract fails stale/legacy metadata open. Merge a9 preserves upstream tag 75 and assigns FuzzyBuildSide tag 76; 771 has no conflict-resolution delta. Focused CGo tests could not run because cgo/libmo.dylib is absent.

XuPeng-SH:
CI follow-up after run 30609110472:

- UT: `TestFloatRuntimeFilterUsesOnlySoundEncoding` now owns/restores protocol v7 explicitly; this removes its previous order/global-state dependency.
- SCA: removed the three reported ineffective assignments; targeted `ineffassign` is clean.
- Proxy BVT: the small `REPLACE` failures exposed a pre-existing PR #26412 HashBuildBudget check-then-open race inherited through main, not a runtime-filter causal regression. `GetHashBuildBudget` now clamps the sampled query cap and constructs the generation atomically under the shared aggregate lock. Explicit public oversized-cap configuration remains strict.
- Regression: process/plan/hashbuild/spillutil passed `-count=3`; the broader related package matrix passed normally; focused process/plan/hashbuild/spillutil race passed. No large-data BVT was added. The new Proxy BVT run is the real SQL confirmation for the original failures.

Commits are separated for review attribution: `c535a79cda` (PR-local UT/SCA) and `2a90b449cd` (inherited spill-budget race).

aptend:
### Compose PROXY BVT failure：`invalid hash build budget`

结论：这是本 PR 引入的 HashBuild/runtime-filter 预算记账回归，不是 CI runner、OOM 或真实的 budget admission rejection。

失败证据：

- [Compose PROXY BVT job](https://github.com/matrixorigin/matrixone/actions/runs/30609110472/job/91087856049) 在 `2026-07-31 06:23:49 UTC` 首次失败：
  ```sql
  replace into t_replace_dup_uk_existing
  values (1, 1, 10), (2, 1, 20), (3, 2, 30);
  ```
  实际返回 `internal error: invalid hash build budget`。
- 后面另外 5 条 REPLACE 也返回同一错误；对应的 6 条 SELECT 因 REPLACE 事务回滚而得到旧数据，所以报告中共有 12 个失败。
- `replace.test` 最终为 `TOTAL: 471, SUCCESS: 459, FAILED: 12`，全局成功率 99% 低于要求的 100%，job 因此以 exit code 1 结束。
- CN 日志中没有 OOM、panic 或进程崩溃信号。这里返回的是预算内部不变量错误 `ErrHashBuildBudgetInvalid`，不是内存额度不足时的 `ErrHashBuildBudgetAdmission`。

触发逻辑如下：

1. 这些失败用例都会进入 REPLACE 的 `DedupBuildKeepLast` 路径。存在重复/扇出时，`keepDiscardedRowsForDelete` 会压缩存活行，同时追加只用于删除旧行的 delete-only 行。
2. 随后进行 canonical HashMap 重建。此时 [`InputBatchRowCount` 被设为有效参与 HashMap 的行数](https://github.com/matrixorigin/matrixone/blob/771c2bd5c23e2748f1acacf710965958271c00c5/pkg/sql/colexec/hashbuild/hashmap.go#L974-L986)，即 `totalRowCount - DelRows.Count()`；辅助预算也按这个 active row count 重算。
3. `buildAuxBytes` 的预算公式包含 [`rows * 64`](https://github.com/matrixorigin/matrixone/blob/771c2bd5c23e2748f1acacf710965958271c00c5/pkg/sql/colexec/hashbuild/budget.go#L494-L526)，所以此时 reservation 是按 active rows 建立的。
4. 递归重建完成后，代码又把 [`InputBatchRowCount` 恢复为包含 delete-only 行的物理总行数](https://github.com/matrixorigin/matrixone/blob/771c2bd5c23e2748f1acacf710965958271c00c5/pkg/sql/colexec/hashbuild/hashmap.go#L986-L990)。
5. 接着 `handleRuntimeFilter` 的 defer 无条件调用 [`releaseOptionalRuntimeFilterKeys`](https://github.com/matrixorigin/matrixone/blob/771c2bd5c23e2748f1acacf710965958271c00c5/pkg/sql/colexec/hashbuild/build.go#L700-L718)。该函数再次用当前的 `InputBatchRowCount` 计算 mandatory auxiliary budget。
6. 这次使用的是更大的 physical total row count，而现有 reservation 是按 active row count 计算的，于是命中：
   ```go
   if required > hb.auxReservation.Size() {
       return process.ErrHashBuildBudgetInvalid
   }
   ```
   [对应检查](https://github.com/matrixorigin/matrixone/blob/771c2bd5c23e2748f1acacf710965958271c00c5/pkg/sql/colexec/hashbuild/budget.go#L570-L590)
7. 这个清理阶段错误最终覆盖为查询错误，REPLACE 回滚，产生本次 BVT 失败。

建议修复：

- 明确区分 active hash rows 与包含 delete-only rows 的 physical retained rows，确保 canonical rebuild、aux reservation 和 terminal cleanup 始终使用同一种行数语义；不要在 reservation 建立后仅切换 `InputBatchRowCount` 的含义。
- 增加一个覆盖 `DedupBuildKeepLast + delete-only rows + budget enabled + RuntimeFilterSpec` 的测试，并通过完整的 `handleRuntimeFilter` 路径验证：查询不报错、JoinMap 正常发布、budget/mpool 最终归零。

aptend (changes requested):
Deep re-review at 3abc35c7. Requesting changes for two code-level blockers:

[P1] Preserve the aux-budget contract after restoring the REPLACE row count

In pkg/sql/colexec/hashbuild/hashmap.go, the DedupBuildKeepLast rewrite sizes the auxiliary reservation while InputBatchRowCount is the survivor count, then restores InputBatchRowCount to totalRowCount without resizing the reservation. handleRuntimeFilter later calls releaseOptionalRuntimeFilterKeys, which recomputes the mandatory projection from the larger count and returns ErrHashBuildBudgetInvalid. A minimal budgeted dedup test with one discarded/delete-only row reproduces this deterministically. Please decouple the map-accounting row count from the published total, or reconcile the reservation when restoring the total count.

[P2] Keep ENUM secondary/unique index joins eligible

SupportsExactRawRuntimeFilter omits types.T_enum even though enum equality is direct ordinal equality, the serial encoder supports it, and secondary/unique enum indexes are valid. Exact pair construction therefore fails and applyIndicesForJoins skips every non-forced enum index rewrite. A planner reproducer leaves the original base scan in place. Please add enum to the raw exact-key allowlist and cover direct and serialized index joins.

XuPeng-SH:
Correction to my previous CI follow-up: aptend’s diagnosis is correct. The shared-CN cap-race attribution was wrong, and commits 2a90b449cd/3db10fc5d9 are fully reverted in the resulting code.

Fixed at 2739d1bed7:

- P1: HashBuild now keeps physical retained rows (published JoinMap row count) separate from active hashmap rows (auxiliary budget projection). The complete DedupBuildKeepLast + delete-only row + runtime-filter path now publishes 3 retained rows / 2 groups, marks the delete row, and releases budget/mpool to zero without ErrHashBuildBudgetInvalid.
- P2: ENUM is admitted only after closing the complete consumer contract: raw exact-key allowlist plus executable ordinal IN(enum, enum). Planner tests cover both direct unique-index RAW_V1 and serialized secondary-index SERIAL_FULL_V1 rewrites.

Fresh evidence after the semantic edits:
- focused tests: count=10, all pass
- focused race tests: count=5, all pass
- full related packages: keycodec, plan/function, plan, hashbuild, process, spillutil, dedupjoin, rightdedupjoin, compile, frontend, all pass
- targeted go vet and ineffassign: pass

CI is now rerunning. The Proxy BVT is the final real-SQL confirmation for the original REPLACE failures; I am not claiming that result before the new run completes.

@aptend please re-review 2739d1bed7.

gouhongshen (changes requested):
## Codex automated review

The DedupBuildKeepLast budget blocker is fixed by separating physical retained rows from active hashmap rows, with focused lifecycle coverage. ENUM index eligibility is restored, but its implementation introduces a new rolling-upgrade blocker. Previously approved typed-FLOAT/spill and versioned-filter behavior remains accepted.

### P1 - Preserve existing IN overload wire IDs and gate ENUM from old consumers (pkg/sql/plan/function/list_operator.go:724)
This inserts the ENUM overload in the middle of `IN.Overloads`. Encoded function IDs use the overload's slice index (`GetFunctionByName` records `check.idx`, and `GetFunctionById` indexes `Overloads[oIndex]`), not the declared `overloadId` field. Consequently every existing overload after this insertion changes its wire ID: for example, CHAR moves from index 14 to 15, while an older CN interprets index 15 as DATE. A plan produced by a new CN and executed by an old CN during rolling upgrade can therefore select an executor for the wrong physical type and panic or filter incorrectly. Additionally, ENUM is classified as `LegacyExactRawProducerSafe`, so protocol-v6 plans can send ENUM runtime filters to old consumers which have no ENUM `IN` overload at all. This is distinct from the earlier ENUM eligibility finding: eligibility is now restored, but the chosen implementation breaks cross-version execution. Append the new overload without shifting existing indices, and require the versioned rollout gate for ENUM rather than advertising it as legacy-consumer-safe.

gouhongshen (commented):
Re-review at `2739d1bed7`: there is no new commit after the previous request, and the rolling-upgrade blocker remains in the current head.

- `types.T_enum` is still inserted before the existing VARCHAR overload in `IN.Overloads` (`pkg/sql/plan/function/list_operator.go:724`). Function IDs encode `check.idx`, and decoding indexes `Overloads[oIndex]`, so all following overload wire IDs remain shifted relative to an older CN.
- `LegacyExactRawProducerSafe(types.T_enum)` still returns true, so protocol-v6 plans may advertise an ENUM exact filter to an older consumer which has no ENUM `IN` overload.

The new ENUM execution and planner tests prove same-version behavior only; they do not close either cross-version contract. The existing `REQUEST_CHANGES` therefore stands. Please push the compatibility fix before requesting another review.

XuPeng-SH:
Addressed the rolling-upgrade blocker at 39feef49dd.

The review is correct: IN function IDs encode the overload slice index, not overload.overloadId. My previous change inserted ENUM before existing entries and shifted every later wire ID. I also treated LegacyExactRawProducerSafe as only a payload property and missed consumer executor availability.

The fix closes both planner and execution boundaries:

- moved ENUM IN to the end of the overload slice; every pre-existing IN wire index remains unchanged
- added a complete pre-existing type-to-wire-index contract table, with ENUM assigned only the new appended index
- classified ENUM RAW as non-legacy-safe, so protocol v6 does not plan it and v7 hides legacy Expr from old producers
- additionally gated non-legacy-safe RAW in the execution contract; a cached v7 ENUM/decimal plan fails open again after the deployment gate is lowered to v6, while int/string legacy-safe RAW stays enabled

Fresh verification after the final semantic edit:

- all focused compatibility/ENUM/contract tests passed count=10
- adaptive race stress used T=0-0.01s, B=30s, N=100 for each exact test; all passed
- complete race packages passed: keycodec, plan/function, plan, runtimefilter; dependent hashbuild and compile also passed
- targeted build, vet, and ineffassign passed

The previous head’s full CI, including Compose PROXY BVT, completed green. CI for 39feef49dd is now the remaining delivery check.

@gouhongshen @aptend please re-review 39feef49dd.

pkg/sql/plan/function/list_operator.go
```diff
@@ -721,6 +721,16 @@ var supportedOperators = []FuncNew{
 					return newOpOperatorFixedIn[types.Decimal256]().operatorIn
 				},
 			},
+			{
```
LeftHandCold: [P1] Please append this overload instead of inserting it here. Encoded function IDs use `check.idx` as the low 32 bits (`GetFunctionByName`), and consumers decode by indexing `Overloads[oIndex]` (`GetFunctionById`); the declared `overloadId: 101` does not stabilize the wire ID. This insertion shifts every existing IN overload at indices 13-53. During a normal mixed-version rolling upgrade, an old VARCHAR plan at index 13 is decoded as ENUM by a new CN, while a new CHAR plan at index 15 is decoded as DATE by an old CN. Remote expression execution uses the encoded ID without rebinding by name, so this can select the wrong executor and panic or return incorrect rows. Please preserve all existing indices and add a cross-version ID regression test.

pkg/common/hashmap/keycodec/keycodec.go
```diff
@@ -25,6 +25,127 @@ import (
...
+	switch oid {
+	case types.T_bool,
+		types.T_bit,
+		types.T_uint8, types.T_uint16, types.T_uint32, types.T_uint64,
+		types.T_int8, types.T_int16, types.T_int32, types.T_int64,
+		types.T_decimal64, types.T_decimal128, types.T_decimal256,
+		types.T_char, types.T_varchar, types.T_blob, types.T_text,
+		types.T_binary, types.T_varbinary,
+		types.T_date, types.T_time, types.T_datetime, types.T_timestamp,
+		types.T_uuid, types.T_year, types.T_enum:
```
LeftHandCold, line 59: [P1] Adding ENUM to this allowlist also makes `LegacyExactRawProducerSafe(T_enum)` return true through the default branch. Consequently `exactRuntimeFilterPlanEncoding` permits an ENUM exact filter below protocol v7 and retains the legacy `Expr`, even though an old CN has no ENUM IN overload. The old consumer then cannot bind the ENUM predicate safely. Please keep ENUM raw support for v7, but exclude it from the legacy-safe set so this path fails open until the versioned rollout gate is enabled; cover the v6 case explicitly.

LeftHandCold (changes requested):
Requesting changes for the two reachable rolling-upgrade compatibility regressions called out inline. Both must be fixed before approval.

gouhongshen (approved):
## Codex automated review

Previously accepted typed-FLOAT/spill and versioned-filter behavior remains accepted. The Dedup auxiliary-budget regression and ENUM index eligibility are fixed. The remaining rolling-upgrade blocker is now fixed: ENUM is appended without shifting existing IN wire IDs, v6 planning suppresses ENUM exact filters, and execution fails cached v7 ENUM/decimal contracts open after gate downgrade. No new findings. Focused tests could not run locally because cgo/libmo.dylib is absent; diff checks passed.

XuPeng-SH:
Compatibility scope correction at a6234a61a5:

- This is not long-term support for an old release family. The latest released v4.1.4 uses protocol v4; v5 and v6 entered main on July 28/29, and this PR introduces the next rollout gate. The intended scope is one controlled rolling-upgrade/rollback window.
- Append-only `IN` overload indices remain required: the current released/main executors decode the existing slice indices, and preserving them has no steady-state cost.
- I removed the extra execution-side gate I had just added for metadata-dependent RAW/ENUM. My prior “cached v7 plan” justification was too broad: normal plan-cache and prepared-plan entries already rebuild when the protocol version changes. The remaining gap is only an in-flight query across a rollback boundary; the deployment contract now explicitly requires draining v7 work before lowering the gate/reintroducing old participants.
- The planner gate remains, because an old ENUM consumer has no `IN(enum, enum)` executor. Normal v7 behavior and the full type/scale/payload triangle validation are unchanged. RAW execution also loses one service-runtime lookup.

Fresh checks: focused compatibility/type tests count=10 passed; exact-contract race count=100 passed; full runtimefilter/hashbuild/function/plan/compile packages passed; vet and ineffassign passed under the repository module contract.

There is also an allocation-order dependency with open PR #26109, which independently reserves v7. Whichever PR merges second must rebase and take the next protocol version; this PR should not pretend an unmerged capability already exists.

aptend (changes requested):
Re-review at a6234a61a5. The earlier REPLACE-budget and ENUM compatibility findings are fixed, but one code-level blocker remains.

[P1] Keep process generation cap sampling and construction atomic

GetHashBuildBudget reads aggregate.AggregateCap outside the generation-open lock, then passes that sampled cap to the strict OpenGenerationWithSpillCaps path. Another statement can lower the shared aggregate cap through refreshAggregateCap or UpdateAggregateCap between those operations. The first statement then reaches OpenGenerationWithCap with cap greater than the new aggregateCap and an ordinary hash-build Prepare fails with ErrHashBuildBudgetInvalid.

A deterministic interleaving on the current head reproduces this as internal error: invalid hash build budget: statement A samples 100, statement B refreshes the aggregate to 40, and statement A tries to open 100. Commit 3db10fc5d9 had already closed this race with a process-only openGeneration helper that clamped and constructed under one lock, but 2739d1bed7 removed both that helper and its regression test.

Please restore a single-lock clamp-and-construct path for process-owned generations while keeping the public explicit-cap API strict, and restore the deterministic stale-sample regression test.

gouhongshen (approved):
Deep re-review at `a6234a61a5`: no in-scope blocking findings.

The latest compatibility narrowing is sound under the stated deployment contract: v6 planning suppresses ENUM/decimal/new-closure exact filters; existing `IN` overload wire indices remain unchanged and ENUM is append-only; v7 non-legacy contracts omit the legacy `Expr`, so an older producer fails open. Session plan-cache and prepared-plan paths rebuild on v7↔v6 changes, and rollback explicitly requires draining in-flight v7 work before lowering the gate/reintroducing old participants. Removing the RAW execution-time lookup therefore does not create a supported mixed-version misexecution path.

I also rechecked the earlier REPLACE/Dedup accounting path, exact-filter type/scale/payload validation, signed-zero handling, spill/hash consistency, terminal publication, message payload release, FD/disk-token cleanup, budget-admission fallback, and IndexBuild/FuzzyFilter failure terminals. The prior REPLACE budget regression and ENUM consumer/wire-ID regressions remain fixed.

The `GetHashBuildBudget` cap-sampling TOCTOU raised in another latest-head review is real, but it exists unchanged in the PR base: `pkg/vm/process/hashbuild_budget.go` has zero net diff, and base HashBuild already calls that path. The only new IndexBuild/FuzzyFilter callers fail open with PASS on this error. It should be fixed separately, not reintroduced as unrelated scope here.

`git diff --check` passes. Local Go tests could not compile because this checkout lacks the external C headers (`usearch.h`, `xxhash.h`, `roaring.h`). The previous semantic head completed full MatrixOne ALL CI including PROXY BVT successfully; the current-head long-running UT/SCA/BVT jobs have no reported failure at review time. CI and the required rebase remain merge gates.

XuPeng-SH:
Further compatibility/performance simplification at ba13bafb8b:

- Removed the deprecated `RUNTIME_FILTER_KEY_LEGACY_FLOAT_ZERO_CLOSED=1` reservation. It only represented an earlier, unmerged commit of this PR and never existed in main or a release, so retaining it would have created branch-history protocol debt. The new enum is now compactly assigned RAW=1, FLOAT_ZERO_CLOSED=2, SERIAL=3, SERIAL_FULL=4. The large generated `plan.pb.go` line delta is the compressed descriptor changing, not handwritten complexity.
- Removed the execution-side protocol lookup from HashBuild, IndexBuild, and FuzzyFilter exact-contract validation. Execution now validates only the explicit encoding plus the probe/build/payload triangle. No protocol `sync.Map` lookup remains per filter construction.
- Compatibility `Expr` duplication is now limited to the pre-v7 rolling window for legacy-safe RAW types. After v7 activation, plans carry only `BuildExpr`; steady-state plans no longer pay permanent duplicate expression bytes. Plan/prepared caches already rebuild on the gate transition.
- The v4-to-next-release planner gate and append-only pre-existing `IN` overload IDs remain; those protect a real released/current consumer contract.

Validation after the semantic/protobuf edit: focused count=10 passed; focused race count=100 passed; full runtimefilter/hashbuild/indexbuild/fuzzyfilter/plan/compile/pb packages passed; vet, ineffassign, and diff checks passed.

Compatibility cost after this change is confined to planning: one local protocol `sync.Map.Load` per candidate exact filter and temporary dual metadata only while the gate is below v7. There is no per-row, per-batch, or steady-state execution compatibility branch.

heni02 (approved):

aptend (approved):
Deep re-review at c6ef8f2f0d. No in-scope blocking findings.

The earlier Dedup/REPLACE budget, ENUM eligibility, and ENUM rolling-upgrade/wire-ID blockers remain fixed. The latest planner-only rollout change is sound under the documented deployment contract: v6 plans retain only legacy-safe raw filters, v7 plans carry the typed BuildExpr/ProbeType/component contract, stale or incomplete execution metadata fails open, and transformed plans expose no unsafe legacy Expr.

I also rechecked typed FLOAT equality across scalar comparison, resident hashmaps, and spill; exact-filter signed-zero/decimal/enum/tuple validation; Dedup destructive-rewrite recovery; query-wide budget ownership and optional-filter fallback; IndexBuild/FuzzyFilter terminals; message/JoinMap resource release; planner placement/stat side effects; and legacy NaN zonemap handling.

The process hash-budget stale-cap race from my prior review exists unchanged in the PR base and is not introduced by this diff, so it should not block this PR.

Fresh local tests pass for the affected hashmap/keycodec/zonemap/runtimefilter/hashbuild/spill/indexbuild/fuzzyfilter/plan/function/compile packages, including race tests for hashbuild and runtimefilter.

pkg/sql/plan/function/list_operator.go
XuPeng-SH: Addressed in 39feef49dd and revalidated at latest head 88c114bea3. The ENUM overload is appended after every pre-existing IN overload, so existing wire indices are unchanged. TestInOverloadWireIDsRemainAppendOnly locks the complete old type-to-index mapping and requires ENUM to use only the new appended index. Please re-review the latest head.

pkg/common/hashmap/keycodec/keycodec.go
XuPeng-SH: Addressed in 39feef49dd and revalidated at latest head 88c114bea3. LegacyExactRawProducerSafe now explicitly excludes T_enum: protocol v6 suppresses the ENUM exact filter, while v7 uses the versioned BuildExpr/ProbeType contract without exposing an unsafe legacy Expr. The rollout tests cover both v6 rejection and v7 admission. Please re-review the latest head.

XuPeng-SH:
@aptend @LeftHandCold @gouhongshen @heni02 @aunjgr please re-review latest head 88c114bea3.

This follow-up closes the remaining review and deep-review gaps:

- restores the process-only single-lock clamp-and-construct path in GetHashBuildBudget, eliminating the stale aggregate-cap TOCTOU while keeping explicit public oversized-cap APIs strict
- makes Closed, Invalid, CeilingMissing, cancellation, and accounting failures fatal; only typed capacity Admission and explicitly marked optional allocation failures may spill/fail open
- applies the same error-fidelity policy to HashBuild, IndexBuild, FuzzyFilter, BucketReader, and SpillEngine, with bounded ExtraStats/global budget metrics
- makes runtime-filter terminal publication idempotent across repeated Reset and reopens the gate only at the next Prepare generation
- removes the per-row normalizer branch from unscaled FLOAT32 hashing; the benchmark remains zero-allocation
- adds no compatibility for temporary spill files/state; rollout handling remains confined to the real cross-CN planner/wire contracts

Fresh validation on the final semantic diff:

- full non-race tests: hashmap, keycodec, runtimefilter, process, hashbuild, spillutil, hashjoin, indexbuild, fuzzyfilter
- full race tests for all owning packages
- exact TOCTOU, fatal-first, Admission/Closed, optional fallback, and repeated-Reset tests stressed independently with race count=100
- focused IN overload wire-ID and v6/v7 rollout tests pass
- build, vet, and git diff --check pass
- final first-principles/Q1-Q3 self-review found no open blocker

XuPeng-SH:
Real MatrixOne spill validation completed on head `88c114bea33276c85e2c91ffee3b6214b8546a5d` using an isolated clean checkout/data directory on 10.222.1.128.

### Real workload

- 5,000,000 probe rows (including 50 NULL keys) and 4,000,000 unique build rows.
- DOP=8; forced spill through the existing `join_spill_mem=1000` session control.
- The non-spill and spill inner joins returned the exact same result:
  - rows: `3,999,960`
  - probe payload sum: `23,999,760,000,000`
  - build payload sum: `55,999,440,000,000`
- Additional semantics passed:
  - LEFT JOIN: `5,000,000`
  - SEMI/IN: `3,999,960`
  - NOT IN: `999,990`
  - MARK three-valued result: true `3,999,960`, false `999,990`, NULL `50`
- Ten alternating executions were byte-identical.

### Spill and cleanup evidence

- `EXPLAIN ANALYZE`: Join spill rows `12,750,000`, spill size `322.65 MiB`, join memory `3.56 MiB` (non-spill join memory: `202.88 MiB`).
- Live `/proc/<pid>/fd` monitoring observed up to 176 anonymous deleted `join_*` spill files; the count returned to zero after the query.
- For the measured query, Prometheus spill-disk reservation and release bytes matched exactly: `404,343,616` each.
- After the full validation run, cumulative disk reservation/release bytes still matched exactly (`7,407,116,818` each), as did FD reservation/release bytes (`321,248` each); no spill FD remained open.
- Actual server-side cancellation was exercised during an active DOP=1 spill: after 32 spill FDs were open, `KILL QUERY` propagated `context canceled`, removed the query from processlist, and closed all spill FDs in 99 ms. The service remained healthy.

A service-wide low-`RLIMIT_NOFILE` experiment was rejected as a causal HashBuild test because LogService/Dragonboat consumes the limit first; HashBuild FD-cap admission remains isolated in the focused unit tests instead.

### Performance observation

Warm-cache, single-host, alternating runs of the same DOP=8 join:

- no spill: median `0.15 s` (0.19, 0.16, 0.14, 0.15, 0.15)
- forced spill every ~1,000 build rows: median `1.36 s` (1.37, 1.34, 1.40, 1.36, 1.36)

This intentionally extreme spill setting is about 9.1x slower, while reducing join memory from 202.88 MiB to 3.56 MiB. Given that spill is rare and temporary in the controlled rollout environment, this PR keeps compatibility narrowly scoped and does not add broader fallback machinery.

No new correctness/resource-lifecycle defect was found in this real-service pass. @aunjgr @LeftHandCold @gouhongshen @aptend @heni02 please re-review the latest head.

aptend (approved):
Re-reviewed the complete diff at 88c114bea3. No blocking findings. The earlier budget, ENUM compatibility, and lifecycle issues are fixed; focused package tests and full race runs for the latest affected packages pass.

LeftHandCold (approved):
Deep re-review completed at `88c114bea3`, including #26432 and the real-service spill/cancellation evidence in the latest author comment. APPROVED.

The latest `c6ef8f2f..88c114be` delta closes the remaining failure-path gaps without reopening earlier blockers:

- budget recovery is fatal-first: Closed/Invalid/CeilingMissing/cancellation/accounting errors no longer masquerade as capacity admission; only typed Admission and explicitly marked optional allocation failures may spill or publish PASS;
- HashBuild, IndexBuild, and FuzzyFilter now publish one terminal runtime-filter message per execution generation and release optional vectors/reservations on every terminal path;
- the process generation cap is clamped and constructed atomically under the aggregate lock;
- unscaled FLOAT32 keeps the same signed-zero canonicalization without adding per-row normalization overhead.

I also re-traced resident IntHashMap/StrHashMap, initial spill, probe partitioning, and recursive re-spill. They use the same per-operand FLOAT32 scale contract; spill payloads retain the original Type.Scale and are canonicalized again when rebuilt. Integer/string hot paths remain byte-based.

The v6/v7 rollout boundary is conservative: v6 keeps only legacy-safe raw filters; decimal, ENUM, tuple, scaled FLOAT32, stale metadata, and contradictory contracts fail open. ENUM's IN overload is append-only and ENUM remains excluded from the v6 legacy producer set. New protobuf fields are append-only with non-conflicting tags, and planner stats/ForceOneCN side effects occur only after a complete exact-filter pair is proven materializable.

The real-service run adds useful end-to-end evidence for result equivalence across forced spill, LEFT/SEMI/NOT IN/MARK semantics, exact disk/FD reservation release, and cancellation cleanup. JSON/VECF32/NaN join semantics remain explicitly outside this incremental PR and #26432 is not being closed, so they are not blockers for this scope.

Current UT, coverage, SCA, BVT, Iceberg, and MongoDB checks are green. I did not run additional broad local tests.

XuPeng-SH:
Pushed `f85fc5a8a8` after the requested first-principles 16 GiB / 100 GB real-service harness.

### What the extreme harness found and fixed

The old head failed the exact 100 GB spill workload after 365s at the shared default disk ceiling:

```
requested=325544 used=103079102168 cap=103079215104
```

Under a 16 GiB cgroup, the default spill disk ceiling is 96 GiB (the usable-memory cap times 8). An explicit positive `processLimitationSpillSize` was still clamped by that default, so an operator could lower the cap but could not deliberately raise it.

`f85fc5a8a8` fixes that contract at the ownership boundary: an explicit positive process spill limit may monotonically raise the shared CN ledger before opening a generation. Zero keeps the bounded default; a lower per-query value still narrows only that generation. The update and generation open are serialized by the existing aggregate mutex. No automatic disk probing, compatibility mode, background work, or steady-state hash-row overhead was added.

### Extreme A/B evidence on `github@10.222.1.128`

Environment:

- Ubuntu 24.04 Docker container, hard `--memory 16g --memory-swap 16g`, 16 CPUs
- exact binary from `f85fc5a8a8` (SHA-256 `6ce558b430c2ed7a18be1b8338c4ea022dc513e522ee8e44c4105306d23ebeef`)
- DOP=1
- build table: 4,000,000 rows × 25,000-byte payload = exactly 100,000,000,000 payload bytes
- probe table: 5,000,000 rows

No-spill control: the same join/cgroup/DOP with a functionally equivalent non-shuffle predicate and an even smaller 80 GB retained build projection was OOM-killed (`exit=137`, `OOMKilled=true`) after 171s. Therefore the full 100 GB payload cannot complete resident in this container.

Forced real spill, exact full 100 GB payload:

```sql
set max_dop=1;
set join_spill_mem=1000;
select count(*), sum(length(b.payload))
from t_probe p
left join t_build_100gb b on b.c1 = p.c1;
```

Result after 784s:

```
5000000  99999000000
```

This exactly matches the independently derived expected result: 5,000,000 output rows, 3,999,960 matched rows × 25,000 bytes.

Resource/metric closure:

- spill disk reserve/release: `301,366,898,424 / 301,366,898,424` bytes
- spill FD reserve/release: `67,648 / 67,648`
- peak join spill FDs: 32; final: 0
- recursive spill depths observed: 1, 2, 3 (1 / 32 / 1024 events)
- spill disk/FD rejects: 0
- observed build working set was about 1.8–2.0 GiB; container remained healthy (`OOMKilled=false`) and SQL health checks passed
- the run crossed the old 96 GiB boundary before completing, proving the new explicit ceiling was exercised rather than bypassed

### Verification

- both focused budget tests: `-race -count=100` PASS
- full `./pkg/vm/process`: PASS
- full `./pkg/vm/process -race`: PASS
- deterministic CGo build and vet: PASS
- added a concurrent max-cap raise test plus lower-cap/default/closed-state coverage

This is intentionally narrow for the stated deployment reality: spill is rare and the upgrade window is short/controlled, so the default remains conservative and there is no new compatibility machinery. The fix preserves the original error when the explicitly provisioned disk budget is genuinely exhausted.

All previously raised review comments remain addressed. @aunjgr @LeftHandCold @gouhongshen @aptend @heni02 please review the new commit again, especially the explicit-cap ownership contract and the real 100 GB spill evidence.

XuPeng-SH:
Follow-up from the exact 100 GB no-spill control: `b2d557758a` fixes the admission-error user experience now, rather than deferring it.

The previous client error was:

```
internal error: hash build budget admission rejected
```

That was not reasonable: it classified an expected finite-resource rejection as an internal fault, discarded requested/used/limit, and gave the operator no action.

The fix keeps typed admission errors internal while spill recovery is still possible. Only after HashBuild has exhausted recovery paths does it publish a client-facing resource error. The error carries the actual resource kind (memory, spill disk, or spill FD), requested/used/limit, and resource-specific action. It reuses the existing resource-exhaustion/OOM wire code, so no new cross-version error code or compatibility mechanism was added. The same detailed message is serialized to remote consumers and used by the terminal operator diagnostic; existing fixed-cardinality reject metrics are unchanged.

A fresh exact-binary run on the 16 GiB container used:

- commit: `b2d557758a`
- binary SHA-256: `d8e3f757a3dc048f2bf2ba0bbabfcc1e5fc79bf9514428f63d72907f3dc7be56`
- exact build payload: `100,000,000,000` bytes
- DOP=1, spill threshold set to 1 TiB, non-spill-eligible join predicate

After 40s the client now receives:

```
ERROR 3015 (HY000): error: resource exhausted: hash build memory budget exceeded
(requested=205389824, used=12730105856, limit=12884901888);
make the join spill-eligible, reduce query memory, or increase processLimitationSize
```

The 16 GiB container remained running (`OOMKilled=false`), SQL health passed, final join spill FDs were 0, working set returned to 1.5 GiB, and the CN memory reject metric increased by exactly one. This is the correct production behavior: the exact 100 GB no-spill workload cannot complete, but the budget guard prevents a physical OOM. The earlier, smaller 80 GB control demonstrated that bypassing this guard does result in Docker OOM kill; deliberately disabling the guard solely to reproduce another kernel OOM would weaken the harness rather than validate the production contract.

Verification for this delta:

- focused memory/disk/FD error-fidelity and terminal-publication tests: `-race -count=100` PASS
- full `pkg/common/moerr`, `pkg/vm/process`, and `pkg/sql/colexec/hashbuild`: PASS
- full process/hashbuild race suites: PASS
- build and vet: PASS
- remote error serialization round trip: covered and PASS

@aunjgr @LeftHandCold @gouhongshen @aptend @heni02 please re-review the latest head `b2d557758a`, including the resource-specific error contract and the exact 100 GB safe-rejection evidence.

XuPeng-SH:
Pushed `a6e96b63df` to remove a non-spill performance defect found during the first-principles audit.

Root cause: budgeted `copyBuildBatch` called `batchesAllocated` both before and after every retained copy. Each call scanned all retained physical batches, so a resident build with N batches accumulated O(N^2) accounting work even when spill never started.

Fix: snapshot only the old tail and compute the post-copy allocation delta from that tail plus this copy's appended suffix. This preserves `CopyIntoBatches` semantics for:

- empty builders and multiple new destination batches;
- full-tail append;
- partial-tail growth followed by a remainder batch;
- the full-ingress case that swaps a new 8192-row batch around an existing partial tail;
- invalid/lost-tail state, which still fails closed.

Local same-process accounting comparison (AMD Ryzen 9 7900X, `-cpu=1`, 3 runs):

- 1 retained batch: old full scan ~1.29 ns; incremental ~3.35 ns;
- 1,024 retained batches: old 469-566 ns; incremental ~3.35 ns;
- 16,384 retained batches: old 6.1-7.6 us; incremental ~3.35 ns.

The old production path performed two full scans per copy. The new delta remains constant with historical batch count and allocates 0 B/op in this comparison. The temporary comparison benchmark was not committed; the real append/grow/swap correctness cases are permanent unit tests.

Non-spill resident hashmap A/B against the PR base (8 runs, 1024 rows, medians) still shows no ordinary INT/STRING core regression: cached INT -0.56%, cached STR -1.64%, cold INT -2.61%, cold STR -0.15%; bytes/op and allocs/op are identical. Negative deltas favor the new head and should be treated as noise, not claimed as an improvement.

- `mo-cgo-test -count=1 ./pkg/sql/colexec/hashbuild` PASS;
- focused `-race -count=100` PASS independently for `TestBatchCopyAllocatedDeltaTracksOnlyChangedSuffix`, `TestBatchCopyAllocatedDeltaRejectsLostTail`, `TestCopyBuildBatchBudgetsPartialTailGrowth`, and `TestCopyBuildBatchBudgetsPartialTailWithRemainder`;
- full `mo-cgo-test -race -count=1 ./pkg/sql/colexec/hashbuild` PASS;
- `git diff --check` PASS.

No-spill is not literally zero-cost: capacity projection and budget admission still run, shuffle builds also prove spill scratch readiness, and FLOAT32 keys intentionally canonicalize SQL signed-zero/scale semantics. However, no spill file I/O occurs unless spill starts, and the unintended history-dependent O(N^2) resident-path cost is now removed.

Please re-review the latest head `a6e96b63df`.

LeftHandCold (approved):
Deep re-review completed at a6e96b63df. No blocking findings.

The f85fc5a8..a6e96b63 delta preserves typed admission through all recovery paths and converts only terminal HashBuild capacity rejection to the existing ErrOOM wire code. Closed, Invalid, cancellation, accounting faults, and ordinary errors remain fatal; IndexBuild and FuzzyFilter optional budgets still fail open as PASS. Memory, spill-disk, and spill-FD diagnostics propagate consistently to the local return and JoinMap consumers.

The retained-copy accounting now inspects only the old tail plus the appended suffix. It remains equivalent to the prior full before/after scan for empty, full, and partial tails, partial growth with remainder, and the full-ingress tail swap. Reservation ownership still terminates through spill drain, JoinMap memory release, or Reset/Free, and every new error path releases its token. The cumulative accounting work is linear, with no new wait, lock, or unbounded-growth edge.

Previously reviewed typed FLOAT equality, resident/spill consistency, runtime-filter fail-open behavior, ENUM overload wire IDs, and v6/v7 rollout guards remain fixed. git diff --check passes. I did not run additional broad local tests due the stated resource constraint; current-head CI has no reported failure, with UT, SCA, and two BVT jobs still pending. The branch is behind current main, so updating the base remains a merge gate and any semantic head change requires re-review.

XuPeng-SH:
Final first-principles follow-up at f934a5ba89414fbf1e1e64e837eeeb7b9f099278.

### Design / scope

- Spill is a rare cold path and its files are ephemeral. This update adds no compatibility layer for temporary spill files or in-memory spill state. Existing compatibility remains only at real cross-CN planner/wire boundaries.
- The resident HashBuild now decides to spill before retaining the threshold-crossing batch. It does not preflight or reserve speculative scatter scratch on the normal non-spill path.
- Rebuild recovery headroom is admitted only after a spill has already occurred. A bounded one-batch scatter floor is held across map-build commit so shared-CN contention cannot create a release/reacquire TOCTOU; it is released idempotently on success and every failure path.
- Error fidelity is unchanged and explicit: only typed capacity Admission may enter spill/re-spill recovery. Closed, Invalid, cancellation, lifecycle, and accounting failures remain terminal. A best-effort scratch-floor admission miss is counted but does not turn an otherwise valid resident rebuild into an internal failure.
- Recursion is bounded by 32 buckets, 3 spill passes, and a 1024-entry queue. Skew/no-progress and exhausted recovery headroom return controlled errors rather than growing without bound.

### Correctness and unhappy paths

- A deterministic 12 MiB-cap regression proved the previous intermediate head 8f5ea4bfcd failed while rebuilding three full wide records: requested=5144844 used=7831552 cap=12582912. The final head re-spills and succeeds.
- The regression checks the exact ownership partition: retained, pending, and unread are disjoint and their union contains every unique key exactly once. It also verifies result rows, memory reservation, disk tokens, spill FDs, and mpool all return to zero.
- Added coverage includes pre-copy threshold recovery, retained-copy admission recovery, map admission recovery, Dedup safe/unsafe paths, partial-tail accounting, builder full/partial MemSize, scratch-floor best-effort rejection, fatal lifecycle-error preservation, and idempotent cleanup.
- New metrics distinguish scratch floor reserve/reject/grow, pre-copy/copy/map re-spill attempts, and floor bytes. Terminal memory errors include resource, requested/used/limit, and actionable operator guidance.

### Non-spill cost

- The f934 rebuild-floor change performs zero work on a normal resident join because it is confined to SpillEngine.RebuildHashmap.
- The PR-wide hard-budget path still has small O(1) accounting and synchronization cost; it is not claimed to be literally free. Serialized alternating A/B measurements found no material regression: INT64 +0.28 percent (noise), VARCHAR -2.37 percent, and 2 fewer allocations/op.

### Real MatrixOne harness on 10.222.1.55

- Exact binary SHA256: 2078a59d480c07458d80da259a6326ccf642e5bca8f886ac1af55e8c1e957dc6, built from f934a5ba89.
- Isolated Docker container: 4 GiB memory, no swap, 4 CPUs. Build side was 1.6 million rows x 4000-byte payload = 6.4 GB logical data; probe side was 500,000 rows.
- Automatic mode, join_spill_mem=0: PASS in 48 s. Exact oracle was 500000 / 500000 / 0 / 125000250000 / 2000000000. Peak live deleted spill state was 32 FDs and 6,452,002,816 bytes; it returned to zero. Container stayed healthy, OOM=false.
- Forced 32 MiB threshold: PASS in 55 s with the identical oracle. Peak sampled spill state was 32 FDs and 6,394,105,408 bytes; it returned to zero. Container stayed healthy, OOM=false.
- Automatic run metrics observed spill depth 1 once and re-spill depth 2 thirty-two times. Cumulative disk reserve/release matched at 12,931,003,530 bytes and FD reserve/release matched at 2,112.
- No-spill negative control used a non-shuffle expression join and a 1 TiB spill threshold. EXPLAIN contained no shuffle. It created zero spill files and was rejected safely in 1 s by the 214,748,364-byte hard memory budget, with requested=33,357,824 and used=199,753,728. The returned ResourceExhausted message tells the operator to reduce build width/concurrency, increase processLimitationSize, or lower join_spill_mem for an eligible shuffle join. The service remained healthy and oom_kill stayed zero. We did not bypass the product guard merely to manufacture a kernel OOM.

### Validation / review state

- Final spillutil package passes with race; fresh hashbuild, hashjoin, dedupjoin, and rightdedupjoin packages pass; targeted go vet and make build pass; diff checks are clean.
- Both existing inline review threads are resolved. CI has restarted for this new semantic head and is not being claimed green yet.

@aunjgr @gouhongshen @heni02 @aptend @LeftHandCold please re-review the complete final head.

XuPeng-SH:
Conflict resolution pushed at 19dbe7fc65bf3baeafa7b8d7de7ac03fb5404bbd, merging current main ecc389d4206b310784595ef4e1c131c97908ab4a.

The real semantic conflict was protocol allocation, not just generated-code text: main now owns v7 for structured CHECK constraints, while this PR had previously allocated v7 to the exact runtime-filter contract. The resolution preserves main v7, assigns the exact runtime-filter contract to v8, and makes MORPCLatestVersion v8. All exact-filter gates/tests moved to v8; CHECK remains gated at v7.

I also strengthened the boundary regression: a CHECK-capable v7 service must still suppress metadata-dependent exact runtime filters, while v8 enables them; lowering v8 back to v7 suppresses them again and invalidates cached/prepared plans.

Both conflicting generated files were regenerated from the cleanly merged proto sources. The result simultaneously preserves upstream CheckDef.origin_sql and removed-field reservations for SOURCE_SCAN, stream_scan, and SHOW_CONNECTORS, plus this PR's FuzzyFilter runtime-filter field, RuntimeFilterSpec tags 7-10, and FuzzyBuildSide tag 76. No removed source/stream operator symbol remains.

Per request, I did not rerun the heavy spill harness, race suite, or full packages. The spill admission/re-spill/cleanup implementation is unchanged from f934a5ba89. Lightweight targeted checks passed:

- protocol constants and plan/pipeline protobuf package compilation
- v7-disabled / v8-enabled exact runtime-filter planner cases
- upstream CHECK-v7 contract
- frontend session-plan and prepared-plan v7-to-v8 / v8-to-v7 invalidation
- FuzzyFilter and HashBuild protobuf roundtrips
- append-only IN overload and function wire IDs
- diff checks, conflict-marker scan, and generated-code determinism

GitHub now reports the PR MERGEABLE; it is BLOCKED only by the restarted checks/reviews, not by conflicts.

@aunjgr @gouhongshen @heni02 @aptend @LeftHandCold please re-review the conflict-resolution head.

LeftHandCold (approved):
Deep re-review completed at 19dbe7fc65. APPROVED; no production-reachable blocking findings.

The latest spill changes preserve row conservation and ownership across the resident-to-spill transition and recursive rebuild: the threshold-crossing input is spilled before retention, retained/pending/unread batches are consumed exactly once, and map-admission recovery is allowed only while HashmapBuilder proves that Dedup has not destructively rewritten its input. Scratch, memory, spill-disk, FD, expression, file, and JoinMap ownership all terminate on success, cancellation, error, Reset/Free, and re-spill paths; recursion and queue growth remain bounded. The resident path adds only constant-time threshold/accounting work and does not pre-admit spill scratch.

The merge with current main also resolves the real protocol collision correctly: structured CHECK remains on MORPC v7, the versioned exact runtime-filter contract advances to v8, all production planner gates use v8, and v7 rollback fails open. Protobuf tags are append-only and preserve both main and PR fields/removals; plan/prepared-plan cache invalidation covers v7 to v8 and v8 to v7.

Focused low-concurrency tests passed for the three new HashBuild threshold/lazy-admission cases, eight rebuild/re-spill headroom, conservation, lifecycle, and Dedup safety cases, and the v7/v8 runtime-filter planner boundary. git diff --check is clean. Current-head CI has no reported failure; UT, coverage, SCA, BVT, and connector jobs are still running and remain merge gates.

PR #26473, opened 2026-07-30
fix: stabilize lockservice RPC lifecycle under CI load

### LINKED ISSUE #26467: [Bug]: TestCallLockOpWithConflict / TestRemoteWithBindChanged / TestIssue26095 flaky — lockservice operation race
## Problem

Multiple lockservice operation tests fail intermittently under CI load:

| Test | Elapsed |
|---|---|
| TestCallLockOpWithConflict | 0.53s |
| TestCallLockOpWithConflictWithRefreshNotEnabled | 6.08s |
| TestRemoteWithBindChanged | — |
| TestIssue26095ConcurrentDataBranchDeletion | — |

All involve lock operations with conflicts, bind changes, or concurrent data branch deletion. Under CI load, timing races cause failures.

## Recent Occurrences

| Date | PR/Run | Job |
|---|---|---|
| 2026-07-30 | [PR #26257](https://github.com/matrixorigin/matrixone/actions/runs/30538881499/job/90858683808) | UT Test on Ubuntu/x86 |

## Suggested Fix

1. Increase lock operation timeout tolerance for CI environments
2. Or add retry logic for transient lock conflicts in tests
3. Or use explicit synchronization barriers instead of timing-based waits

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26467 #26475 

The failures grouped in #26467 are not one lock-conflict race and are not test-only. A loaded CI host exposes four production contract violations:

1. **Local scheduling was treated as peer failure.** MORPC charged time spent waiting for a process-wide create worker to the backend factory timeout and could poison a healthy peer's breaker.
2. **Endpoint observations were treated as service identity.** A stale endpoint could return one negative validation result and disable a newer/live lockservice bind.
3. **Cleanup ownership was implicit.** Backend, table, callback, and response cleanup could block while holding publication locks, race with replacement generations, or be performed twice.
4. **Shutdown was not a protocol.** Close paths did not consistently seal admission, cancel admitted work, join it, drain queues, and then release resources. Under adverse timing this caused late sends, long waits, leaks, or deadlocks.

#### MORPC scheduling, failure evidence, and lifecycle

- Separate the local backend-create queue budget from the peer factory budget. Queue-full/queue-timeout are local transient errors and do not update the peer breaker.
- Replace synchronous queue-full fallback with a fixed four-worker process-wide factory pool, bounding concurrent DNS/dial work under overload.
- Track exact create state and generation through queued -> started -> completed, share one factory budget across retries, and reject late results from invalidated generations.
- Preserve definitive `BackendCannotConnect` evidence while keeping timeout/closed/local-congestion failures transient.
- Pin a manager generation to each client, support safe manager replacement/reconfiguration, use one metrics ticker, and gate workers again after dequeue so shutdown cannot start new factory work.
- Make `CloseBackend` invalidate and detach under the client mutex, then physically close outside it. Replacement publication is no longer blocked by a slow backend close, while client close still joins the detached cleanup exactly once.

#### Lockservice recovery and shutdown

- Isolate normal, control, active-transaction, validation, and keeper traffic so transaction saturation cannot starve recovery/control paths.
- Require two independent negative validation observations: reset the validation backend and refresh authoritative discovery before disabling a bind. Missing reset capability fails closed instead of accepting one stale negative.
- Commit negative validation only if the bind-set identity and keepalive generation still match the snapshot, so stale validation cannot overwrite newer positive evidence.
- Add idempotent/sticky close ownership and best-effort `errors.Join` cleanup for clients, server, allocator, and service resources.
- Implement service/server admission gates: seal new work, cancel admitted work, wait for callbacks/operations, drain queued responses, then close owned tables and transports.
- Propagate service cancellation through remote/forward lock work and transfer async callback finalization with an explicit atomic ownership state.
- Run unknown-commit callbacks outside resolver tasks so a re-entrant `Close` cannot wait on its own stopper task.
- Detach lock tables and waiter/store edges under locks, publish the new holder version, and perform potentially blocking close/notification outside locks. This removes holder/table -> bounded event queue -> worker -> holder/table cycles.

### Test-only changes

- Add deterministic, channel-synchronized regression tests for queue admission, create-generation invalidation, manager stop/reconfiguration, close ownership, callback re-entry, service cancellation, stale validation evidence, and full waiter queues.
- Complete the bind-change recovery protocol in test doubles instead of relying on timing.
- Reuse the package-owned authenticated embedded cluster in Issue26095 instead of starting another persistent cluster, reducing CI port/resource contention.

The tests expose the interleavings; the correctness and performance fixes are in production code.

### Performance impact

- Concurrent factory work is hard-bounded at four workers; there is no unbounded synchronous dial fallback.
- Slow physical backend/table close no longer serializes publication under global/holder locks.
- Independent recovery transports trade a small, bounded connection footprint for isolation from transaction-traffic head-of-line blocking.
- Transport cleanup is concurrent where resources are independent, avoiding additive close latency.

### Validation on merged latest `main`

- `mo-cgo-test -race -count=1 ./pkg/common/morpc` — PASS
- `mo-cgo-test -race -count=1 ./pkg/lockservice` — PASS
- original lock-conflict tests, `-race -count=100` — PASS
- `TestRemoteWithBindChanged`, `-race -count=100` — PASS
- `TestIssue26095ConcurrentDataBranchDeletion`, `-race -count=1` — PASS
- `go vet` with the repository CGo environment for MORPC, lockservice, lockop, and issue tests — PASS

### Known bounded follow-up

`BackendFactory.Create` does not accept a context today. Cancellation can therefore invalidate its result but cannot interrupt a factory call that has already started. The fixed worker pool bounds the blast radius; a contextual factory API (and, if measurements justify it, an explicit control-priority lane) should be handled as a separate interface change.

Unknown-commit completion callbacks are detached from the resolver stopper task so that callback re-entry into `Close` cannot self-deadlock. MatrixOne's current callbacks are short admission-release operations; the public callback contract should explicitly require non-blocking completion, or move to a bounded dispatcher, before accepting arbitrary external callbacks.


gouhongshen (approved):
## Codex automated review

No actionable defects found. Existing reviews, PR comments, and inline threads contained no relevant findings. Focused native tests could not run locally because cgo/libmo.dylib is absent.

aptend (changes requested):
Request changes: local backend-create queue congestion leaks half-open circuit-breaker probe admissions.

The new queue-full/queue-timeout paths return without `RecordFailure` or `RecordSuccess`, but `Allow()` increments `halfOpenRequests`, and each logical operation invokes it more than once (the operation-level handle plus `getBackendForOperation`). With `HalfOpenMaxRequests=3` and a saturated create queue, the first `Ping` returns `ErrBackendCreateQueueFull` after consuming two probes; the second consumes the final probe and then returns `ErrCircuitHalfOpen`; every subsequent request is rejected permanently because no request can reach a terminal breaker update.

I reproduced this deterministically under `-race`. The new queue-congestion test covers only the closed state and does not exercise half-open recovery.

Please make breaker admission single-use per logical request/attempt, or add a neutral completion path that releases half-open admission, and add a half-open + queue-full/queue-timeout regression test.

gouhongshen (approved):
## Codex automated review

The prior blocking half-open admission leak is fixed: each RPC attempt owns one permit, and queue-full/queue-timeout paths neutrally release it. Generation protection prevents stale completions from affecting newer breaker state. Focused race stress and vet passed; no new blocking findings.

aptend (approved):
Re-reviewed the updated diff at 7c752d1c. The half-open circuit-breaker admission leak is fixed with single-use, generation-bound permits and neutral release paths. Focused race stress, full MORPC/lockservice race suites, dependent tests, build, and vet passed. No blocking issues found.

iamlinjunhong (approved):

PR #26476, opened 2026-07-31
perf(disttae): index tombstone object ranges

gouhongshen:
## What type of PR is this?

- [x] Improvement

Related to #26465

After #26470 removed the per-object allocation, large table scans still test
every visible tombstone object for every data block. The QA workload in #26465
has system tables with roughly 700 visible tombstone objects and scans roughly
132 data blocks per statement, resulting in about 92,000 object-level Rowid
zone-map checks per scan.

This PR builds a per-data-source interval index from each tombstone object's
Rowid zone-map block range:

- ranges are sorted by minimum block ID and carry a cumulative maximum;
- each data block queries only intervals that can contain its block ID;
- candidate objects are restored to the original visible-object order before
  the existing tombstone reader is called, preserving read and error order;
- objects without an initialized Rowid zone map remain unconditional
  candidates because they cannot be safely excluded;
- scans below 32 blocks keep the existing allocation-free linear iterator to
  avoid paying the index construction cost for small reads;
- the index and candidate scratch space are released when the data source is
  closed.

The existing object/block zone-map checks remain in place. This changes only
candidate selection, not tombstone visibility or deletion semantics.

Validation:

- `go test ./pkg/vm/engine/disttae -count=1`
- `go test -race ./pkg/vm/engine/disttae -run '^TestTombstoneObjectIndex' -count=1`
- exhaustive equivalence test over 701 objects and 1,100 block IDs
- QA-shape benchmark (700 tombstone objects, 132 data blocks, 5 iterations):
  - linear scan: about 501 us/op
  - index build: about 72 us/op, 57,344 B/op, 1 alloc/op
  - indexed block queries: about 11.8 us/op, 0 B/op
  - build plus queries is about 6x faster than the linear scan

This is intentionally separate from the cgroup pressure handling in #26471.
It reduces the work generated by the workload but does not claim to solve all
sources of CN memory pressure by itself.


XuPeng-SH (approved):
Reviewed exact head 4bb4fc7792e5. The interval query is equivalent to the existing object-level RowidPrefixEq check, retains unindexed objects conservatively, restores visible-object order before I/O, and is scoped to the fixed data-source snapshot/lifetime. Focused equivalence, integration, iterator, and race tests pass. No actionable issues found.

PR #26477, opened 2026-07-31
fix(external): align parquet timestamp and float conversions

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24914

fix(external): align parquet timestamp and float conversions


XuPeng-SH (approved):
Reviewed the complete diff at fdf54dba4a27d080fc06fc19ac725dbc145f89bd. The parquet DATETIME conversion now consistently distinguishes UTC-adjusted timestamps from wall-clock timestamps across units and dictionary/non-dictionary paths, and the float32 overflow guard covers both direct DOUBLE and general numeric conversions. Focused regression tests passed under -race with repetition, the full external package passed under -race, and exact-head CI is green.

Issue #22736, opened 2025-10-31
[Bug]: Self-Referencing Foreign Key DELETE Causes Deadlock


XuPeng-SH:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main,3.0-dev

### Commit ID

latest

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

When deleting a row from a table with self-referencing foreign key constraint (`ON DELETE SET NULL`), the DELETE operation **hangs indefinitely** and never completes.

The DELETE operation **hangs indefinitely**:

```
mysql> DELETE FROM t WHERE id = 2;
^C^C^C  -- Ctrl-C doesn't work
        -- Need to kill the connection
```

### Expected Behavior

The DELETE should:
1. Delete the row with `id = 2`
2. Set `parent_id = NULL` for row with `id = 3` (because it references deleted row)
3. Complete successfully

### Steps to Reproduce

```Markdown
DROP DATABASE IF EXISTS test_fk_self_minimal;
CREATE DATABASE test_fk_self_minimal;
USE test_fk_self_minimal;

-- Create self-referencing table
CREATE TABLE t (
    id INT PRIMARY KEY,
    parent_id INT,
    FOREIGN KEY (parent_id) REFERENCES t(id) ON DELETE SET NULL
);

-- Insert hierarchy
INSERT INTO t VALUES (1, NULL);  -- Top level
INSERT INTO t VALUES (2, 1);     -- Child of 1
INSERT INTO t VALUES (3, 2);     -- Child of 2

-- THIS OPERATION HANGS
DELETE FROM t WHERE id = 2;
```

slyunnn:
回归验证结论：通过：自引用外键 ON DELETE SET NULL 删除不再卡住，BVT 11/11。

slyunnn:
## Extended verification for #22736 (2026-08-11)

  Follow-up to the self-referencing foreign-key `ON DELETE SET NULL` hang.

  ### Verification environment

  - MatrixOne main: `d01c85974f9540f8baa2e768ca86fea952941fe2`
    - `fix(hashjoin): broadcast const vectors in spill paths (#26938)`
  - Fix commit: `0b8b53dff3869dcd5e3ab52992852c9ea6545026`
    - `fix: avoid self-referencing SET NULL delete hang (#26479)`
[...]

PR #26479, opened 2026-07-31
fix: avoid self-referencing SET NULL delete hang

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #22736 

fix: avoid self-referencing SET NULL delete hang

### DIFF
--- a/pkg/sql/plan/build_dml_util.go
+++ b/pkg/sql/plan/build_dml_util.go
@@ -335,6 +335,36 @@ func getStepByNodeId(builder *QueryBuilder, nodeId int32) int {
 	return -1
 }
 
+// collectStepDependencyNodes returns every node needed to produce sourceStep,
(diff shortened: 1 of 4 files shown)


XuPeng-SH (approved):
Reviewed the complete diff at 03f81d81365947ea1b3c879586183bb0079f735a. The dependency-closure exclusion prevents the merged delete source from being rewritten to depend on its own sink, removing the self-referencing ON DELETE SET NULL wait cycle while preserving rewrites for independent consumers. The traversal is cycle-safe and query-bounded. The focused regression passed 20 times under -race, the full plan package passed under -race, and exact-head required CI is green.

heni02 (approved):

Issue #24997, opened 2026-06-16
correlated columns in SCALAR subquery deeper than 1 level not yet implemented (error 20102) — v4.0.0-rc2


dengn:
## Version

`8.0.30-OmniFabric-v4.0.0-rc2` (output of `SELECT version()`)

## Description

A scalar subquery that contains a nested scalar subquery whose correlation references an outer table alias **more than one level up** fails to plan with:

```
ERROR 20102 (HY000): correlated columns in SCALAR subquery deeper than 1 level will be supported in future version is not yet implemented
```

This is a common "value at the most-recent timestamp per entity" pattern and runs on MySQL 8 / PostgreSQL.

## Reproduction

```sql
CREATE TABLE j_dim  (id INT);
CREATE TABLE j_fact (id INT, dim_id INT, ts INT, val DOUBLE);

SELECT a.id,
       (SELECT MAX(x.val)
          FROM j_fact x
         WHERE x.ts = (SELECT MAX(y.ts)
                         FROM j_fact y
                        WHERE y.dim_id = a.id)   -- correlates to outer `a`, two levels up
       ) AS latest
FROM j_dim a;
```

## Actual

```
ERROR 20102 (HY000): correlated columns in SCALAR subquery deeper than 1 level will be supported in future version is not yet implemented
```

## Expected

The statement plans and runs, resolving the correlation to the outermost `a` from the doubly-nested scalar subquery.

## Impact

Hit by a natural-language-to-SQL assistant. Questions like "give me the exact value at the latest timestamp for entity X" produce this nested-correlated scalar shape, which fails instead of running.

slyunnn:
回归测试结论：通过：BVT 15/15。

slyunnn:
### Extended verification for #24997 (2026-08-10)

  Follow-up to the nested correlated scalar-subquery report: verified the fix from #26480 on the latest main branch.

  ### Verification environment

  - MatrixOne branch: `main`
  - MatrixOne commit: `fa0ea95aa77b59a2947159baab31e7215d040e93`
    `fix(plan): support subqueries in left and right join ON clauses (#26763)`
  - Fix under verification: `ce384091488175e08410f0795379d217112421e3`
    `fix(plan): support nested correlated scalar aggregates (#26480)`
  - MOTR branch: `test/issue-24997-nested-correlated-scalar`
  - MOTR commit: `8134ac43484bb8a4c20ef6cd7e6515214faa9132`

  ### What was covered

[...]

PR #26480, opened 2026-07-31
fix(plan): support nested correlated scalar aggregates

iamlinjunhong:
## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #24997 

fix(plan): support nested correlated scalar aggregates


heni02 (approved):

XuPeng-SH (changes requested):
Blocking correctness issue in pkg/sql/plan/flatten_subquery.go: the new deepScalarAggregate guard assumes every aggregate except count/starcount is NULL-on-empty. That is not true for approx_count and approx_count_distinct (their executor has emptyNull=false and flushes an empty sketch as 0). pullupThroughAgg turns the correlated inner expression into a GROUP BY key, so an outer key with no matching inner rows loses the aggregate row; the LEFT JOIN then exposes NULL instead of the original 0. For example, if the middle table has a row with ts=0, `WHERE x.ts = (SELECT approx_count(y.ts) FROM y WHERE y.id = outer.id)` should match when that outer id has no y rows, but this rewrite cannot match it. Please gate this on the aggregate empty-input contract (or conservatively whitelist only NULL-on-empty aggregates) and add a no-matching-key regression test.

iamlinjunhong:
> Blocking correctness issue in pkg/sql/plan/flatten_subquery.go: the new deepScalarAggregate guard assumes every aggregate except count/starcount is NULL-on-empty. That is not true for approx_count and approx_count_distinct (their executor has emptyNull=false and flushes an empty sketch as 0). pullupThroughAgg turns the correlated inner expression into a GROUP BY key, so an outer key with no matching inner rows loses the aggregate row; the LEFT JOIN then exposes NULL instead of the original 0. For example, if the middle table has a row with ts=0, `WHERE x.ts = (SELECT approx_count(y.ts) FROM y WHERE y.id = outer.id)` should match when that outer id has no y rows, but this rewrite cannot match it. Please gate this on the aggregate empty-input contract (or conservatively whitelist only NULL-on-empty aggregates) and add a no-matching-key regression test.

fixed

XuPeng-SH (changes requested):
Blocking correctness issue in `pkg/sql/plan/flatten_subquery.go` around `deepScalarAggregate` / `allAggregatesReturnNullOnEmpty`:

Checking only the inner aggregate's empty-input value is not sufficient. `pullupThroughAgg` adds the deep correlation key to `GROUP BY`; when an outer key has no matching inner rows, the rewritten subtree has no group, so the deep filter removes the enclosing input rows. That is not equivalent to producing a NULL scalar value when the caller subsequently handles that NULL.

I reproduced this on the current head `e119c10bab75` with `j_dim = (1),(2),(3)` and five non-empty `j_fact` rows, none having `dim_id = 3`:

```sql
SELECT a.id,
       (SELECT COUNT(COALESCE(
                   (SELECT MAX(y.ts)
                      FROM j_fact y
                     WHERE y.dim_id = a.id),
                   0))
          FROM j_fact x) AS actual
  FROM j_dim a
 ORDER BY a.id;
```

For `a.id = 3`, SQL semantics are: the whitelisted inner `MAX` returns NULL, `COALESCE` turns it into 0 for each of the five `x` rows, and `COUNT` returns 5. The PR-built server returns 0 because the rewrite removes all five `x` rows. A scalar projection such as `COALESCE(MAX(...), 0)` has the same problem (expected 900 in the added fixture, observed NULL).

Please keep this shape on the NYI path unless the rewrite can synthesize the missing per-outer-key aggregate row, or constrain eligibility using the complete consuming-expression semantics rather than only `subCtx.aggregates`. Please add a no-matching-key regression with `COALESCE` around the nested `MAX`.

XuPeng-SH (changes requested):
Blocking: the new guard validates that the aggregate itself returns NULL on empty and that the consumer outside the deep scalar subquery rejects NULL, but it does not validate the scalar subquery's own result projection.

Using the BVT data added by this PR, this query should return 900 for id=3:

```sql
SELECT a.id,
       (SELECT MAX(x.val)
          FROM j_fact x
         WHERE x.ts = (
               SELECT COALESCE(MAX(y.ts), 0)
                 FROM j_fact y
                WHERE y.dim_id = a.id)) AS latest
  FROM j_dim a
 WHERE a.id = 3;
```

The current head accepts it. The generated plan groups the inner aggregate by y.dim_id, projects COALESCE(MAX(...), 0), and then LEFT JOINs it. For the missing id=3 group there is no right-hand row, so the LEFT JOIN supplies NULL for the entire projected column; COALESCE never executes. The enclosing x.ts = NULL filter drops the ts=0 row and the result becomes NULL instead of 900.

Please either require the complete inner scalar result expression to be proven NULL-propagating from its aggregate(s), or keep this shape on the NYI path, and add this missing-key case to the BVT/unit coverage.

iamlinjunhong:
> Blocking: the new guard validates that the aggregate itself returns NULL on empty and that the consumer outside the deep scalar subquery rejects NULL, but it does not validate the scalar subquery's own result projection.
> 
> Using the BVT data added by this PR, this query should return 900 for id=3:
> 
> ```sql
> SELECT a.id,
>        (SELECT MAX(x.val)
>           FROM j_fact x
>          WHERE x.ts = (
>                SELECT COALESCE(MAX(y.ts), 0)
>                  FROM j_fact y
>                 WHERE y.dim_id = a.id)) AS latest
>   FROM j_dim a
>  WHERE a.id = 3;
> ```
> 
> The current head accepts it. The generated plan groups the inner aggregate by y.dim_id, projects COALESCE(MAX(...), 0), and then LEFT JOINs it. For the missing id=3 group there is no right-hand row, so the LEFT JOIN supplies NULL for the entire projected column; COALESCE never executes. The enclosing x.ts = NULL filter drops the ts=0 row and the result becomes NULL instead of 900.
> 
> Please either require the complete inner scalar result expression to be proven NULL-propagating from its aggregate(s), or keep this shape on the NYI path, and add this missing-key case to the BVT/unit coverage.

fixed

XuPeng-SH (changes requested):
Deep-reviewed updated exact head `873419c34d7ceab3a25c194c7e1a9503e8314b7c`. The previous aggregate/projection/consumer NULL-safety blockers are fixed conservatively, but one structural correctness blocker remains.

### P1 — LIMIT/OFFSET become global after per-key decorrelation

The new `deepScalarAggregate` eligibility checks the aggregate empty-input contract and NULL propagation, but not row-order/row-limit wrappers on the scalar aggregate plan. `pullupThroughAgg` adds the deep correlation key to GROUP BY, turning the original one-row implicit aggregate into multiple key groups. A `LIMIT` or `OFFSET` that originally ran independently inside each correlated scalar invocation then remains on the grouped plan and truncates keys globally.

I reproduced this through SQL on a server built from this exact head using the PR fixture:

```
shape                       id=1   id=2   id=3
no LIMIT                    220    300    NULL
inner aggregate LIMIT 1     220    NULL   NULL
inner LIMIT 1 OFFSET 1      NULL   300    NULL
```

`LIMIT 1` must be equivalent to no limit because an implicit scalar aggregate already returns exactly one row per invocation. `LIMIT 1 OFFSET 1` must return no scalar row for every outer id. The observed result shows that limit/offset are applied once across the decorrelated correlation-key groups.

Please make deep-scalar eligibility validate the complete plan topology from the scalar root to AGG. Keep LIMIT/OFFSET (and any other wrapper whose semantics are per correlated invocation but global after grouping) on the NYI path unless it is explicitly rewritten per key. Add public regressions for both `LIMIT 1` equivalence and `LIMIT 1 OFFSET 1`, with multiple matching and missing outer keys.

Fresh exact-head service build, `go list`, build, vet, focused new tests, and the full `pkg/sql/plan` suite pass; the blocker is an uncovered reachable plan shape.

iamlinjunhong:
> Deep-reviewed updated exact head `873419c34d7ceab3a25c194c7e1a9503e8314b7c`. The previous aggregate/projection/consumer NULL-safety blockers are fixed conservatively, but one structural correctness blocker remains.
> 
> ### P1 — LIMIT/OFFSET become global after per-key decorrelation
> The new `deepScalarAggregate` eligibility checks the aggregate empty-input contract and NULL propagation, but not row-order/row-limit wrappers on the scalar aggregate plan. `pullupThroughAgg` adds the deep correlation key to GROUP BY, turning the original one-row implicit aggregate into multiple key groups. A `LIMIT` or `OFFSET` that originally ran independently inside each correlated scalar invocation then remains on the grouped plan and truncates keys globally.
> 
> I reproduced this through SQL on a server built from this exact head using the PR fixture:
> 
> ```
> shape                       id=1   id=2   id=3
> no LIMIT                    220    300    NULL
> inner aggregate LIMIT 1     220    NULL   NULL
> inner LIMIT 1 OFFSET 1      NULL   300    NULL
> ```
> 
> `LIMIT 1` must be equivalent to no limit because an implicit scalar aggregate already returns exactly one row per invocation. `LIMIT 1 OFFSET 1` must return no scalar row for every outer id. The observed result shows that limit/offset are applied once across the decorrelated correlation-key groups.
> 
> Please make deep-scalar eligibility validate the complete plan topology from the scalar root to AGG. Keep LIMIT/OFFSET (and any other wrapper whose semantics are per correlated invocation but global after grouping) on the NYI path unless it is explicitly rewritten per key. Add public regressions for both `LIMIT 1` equivalence and `LIMIT 1 OFFSET 1`, with multiple matching and missing outer keys.
> 
> Fresh exact-head service build, `go list`, build, vet, focused new tests, and the full `pkg/sql/plan` suite pass; the blocker is an uncovered reachable plan shape.

fixed

XuPeng-SH (approved):
Deep-reviewed updated exact head `25aca5e219f4df1b5efa69c6983380d118a7ddfa`. The previous LIMIT/OFFSET blocker is closed by a conservative topology proof: only `PROJECT* -> AGG(aggregateTag)` is eligible, LIMIT/OFFSET/RANK are rejected on every accepted node, and SORT/DISTINCT/FILTER/other wrappers remain on the NYI path.

Exact-head public SQL verification covered:

- the supported direct aggregate and CAST-projection shapes, both returning `1 -> 220`, `2 -> 300`, `3 -> NULL`;
- the earlier missing-group hazards (`APPROX_COUNT`, inner `COALESCE`) remaining NYI;
- LIMIT 1, LIMIT 1 OFFSET 1, DISTINCT, and SORT wrapper controls all remaining NYI.

`go list`, package build, vet, focused new tests, the full `pkg/sql/plan` suite, and an exact-head service build all pass. I found no remaining correctness blocker.

PR #26488, opened 2026-07-31
fix: count binary char length in bytes

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25295

`CHAR_LENGTH()` previously had overloads only for text string types. Binary
arguments were therefore implicitly cast to text and counted as UTF-8 code
points. For example, `CHAR_LENGTH(_binary '你好')` returned `2` instead of the
MySQL-compatible byte length `6`.

This change:

- adds direct `BINARY`, `VARBINARY`, and `BLOB` overloads that count raw bytes;
- keeps `CHAR`, `VARCHAR`, and `TEXT` on the existing UTF-8 character-counting
  path;
- registers `CHARACTER_LENGTH()` as an alias of `CHAR_LENGTH()`; and
- adds unit and BVT coverage for binary literals, binary columns, invalid UTF-8,
  empty values, and `NULL`.

Validation:

- focused function tests;
- `TestLengthBinary` and `Test_GetFunctionByName` with `-race -count=100`;
- full `pkg/sql/plan/function` race tests;
- `make build`;
- focused `go vet` and `golangci-lint` (`0 issues`);
- `func_string_char_length.test` BVT (`83/83` passed); and
- `git diff --check`.


XuPeng-SH (approved):
Reviewed the complete diff at 2ee5ed13097bc12533fe2eb91d40c404d3315a72. The shared LENGTH_UTF8 overloads preserve UTF-8 character counting for text while defining BINARY/VARBINARY/BLOB as raw byte length, including empty, NULL, and invalid-UTF8 inputs. Focused tests passed repeatedly under -race, the full function package passed under -race, and the exact-head check run is green.

heni02 (approved):

PR #26492, opened 2026-07-31
fix(frontend): avoid LCA probe alias collisions

### LINKED ISSUE #26266: [Bug]: DATA BRANCH LCA probe fails when a primary key is named __idx_
## Description

`DATA BRANCH DIFF` fails when a user-defined primary-key column is named `__idx_`.

`__idx_` is a legal user column name: the `CREATE TABLE` statement succeeds. The failure occurs later in `handleDelsOnLCA`, which also uses `__idx_` as an internal ordinal column in the generated `VALUES` alias. The derived table therefore has two columns with the same name, and references such as `pks.__idx_` are ambiguous.

This is not introduced by #26200. That PR correctly quotes user-defined primary-key identifiers, but quoting does not prevent the internal-name collision.

## Environment

- Branch: `main` plus PR #26200
- PR head: `a888fedef8dcf748086dfab3b58ab0fbc0f007ca`
- Latest `main` tested: `81890b81a1ef479e6452ae81a0067113bdad33e8`
- Deployment: local single-node MatrixOne

## Steps to reproduce

```sql
drop database if exists br_lca_idx_collision;
create database br_lca_idx_collision;
use br_lca_idx_collision;

create table base_t(
  `__idx_` int primary key,
  payload varchar(20)
);
insert into base_t values (1, 'one'), (2, 'two');

data branch create table left_t from base_t;
data branch create table right_t from base_t;

delete from left_t where `__idx_` = 1;
update right_t set payload = 'two-right' where `__idx_` = 2;

data branch diff left_t against right_t;
```

## Actual behavior

The table and branches are created successfully, but the final DIFF fails deterministically:

```text
ERROR 20301 (HY000): invalid input: ambiguous column reference 'pks.__idx_'
```

The LCA probe builds a derived-table alias shaped like:

```sql
... right join (values row(0, 1))
    as pks(__idx_, `__idx_`)
    on lca.`__idx_` = cast(pks.`__idx_` as INT)
    order by pks.__idx_
```

Both the internal ordinal and the user primary key are named `__idx_`.

## Expected behavior

DATA BRANCH operations must work for every legal user-defined primary-key identifier. Internal generated-column names must not collide with user schema names.

## Affected paths

`handleDelsOnLCA` in `pkg/frontend/data_branch_hashdiff.go` generates this LCA lookup query. The path is shared by DIFF and by deletion handling used by MERGE and PICK.

## Suggested acceptance criteria

- DIFF succeeds when the PK is named `__idx_`.
- MERGE and PICK deletion paths succeed for the same schema.
- Existing reserved-word, punctuation, composite-PK, and embedded-backtick identifier coverage continues to pass.
- The internal ordinal remains unambiguous without weakening identifier quoting added by #26200.

## Related

- #26071
- #26200

gouhongshen:
## What type of PR is this?

- [x] BUG
- [ ] Improvement
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26266.

`handleDelsOnLCA` reused user-defined primary-key names as column aliases for its generated `VALUES` table while reserving `__idx_` for the input ordinal. A legal primary key named `__idx_` therefore produced duplicate derived-table column names and ambiguous references.

This change gives the ordinal and each generated PK value a distinct internal alias based only on its position. User-defined identifiers are now used only for the quoted LCA-side column references, so they cannot collide with the generated `pks` layout.

The existing quoted-PK BVT now covers `__idx_` through DIFF, MERGE, and PICK deletion paths, and the frontend unit test verifies the generated SQL aliases directly.

## Validation

- `go vet ./pkg/frontend`
- `mo-tester ... -i branch_lca_quoted_pk -r 100` (53/53 statements passed)
- `git diff --check`


XuPeng-SH (approved):
Reviewed exact head `9f0ef4140dde598896e157fd916749a857c47a97`. The generated VALUES aliases are now positional and mutually distinct, while user-defined PK identifiers remain confined to correctly quoted LCA references. The closure is covered for real/composite keys and DIFF/MERGE/PICK deletion paths. `TestHandleDelsOnLCA_SQLPaths` passed under `-race -count=20`, the full `pkg/frontend` package passed under race, `git diff --check` is clean, and exact-head CI is green.

PR #26495, opened 2026-07-31
fix: resolve MongoDB external scan follow-ups

### LINKED ISSUE #26485: [Bug]: MongoDB external scan follow-ups after #26424
## Summary

PR #26424 introduced MongoDB external table scans and was merged as
`cf6dca31107769ad7855ebb45b1a6b48298284d6`. A post-merge first-principles
review found several independently reproducible production issues across
authorization, temporal correctness, memory admission, snapshot/PITR
consistency, and `max_by` complexity. The connector's required CI passed, but
the cases below are outside the current test matrix.

This is an umbrella follow-up issue. Each checked item should be closed only
after its regression test and production fix land.

## Confirmed production issues

- [ ] **Authorization boundary can be bypassed through generic external-table
  metadata.**

  `ParseCreateSQLEnvelope` searches the entire `rel_createsql` string for
  `MO_MONGODB:`. Generic external-table options, including user-controlled
  filepath values, are serialized into the same field. Authentication checks
  account-admin only when the original AST has `MongoDBParam != nil`, while
  compile and query planning classify the persisted metadata by the loose
  marker. A normal `CREATE TABLE` holder can therefore cause a generic
  external table to use an administrator-created MongoDB connection.

  References:
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/mongodb/envelope.go#L58-L69
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/frontend/mysql_cmd_executor.go#L3533-L3591
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/plan/build_ddl.go#L1129-L1163
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/compile/ddl.go#L2163-L2193

- [ ] **Temporal predicate pushdown produces false negatives for
  `DATETIME/TIMESTAMP(0..2)`.**

  Pushdown eligibility ignores the mapped column scale. For example, BSON
  `10.100s` normalizes to `DATETIME(0) = 10s`, so the MO residual accepts it,
  but the pushed Mongo equality is `{ts: 10.000s}` and discards the document
  before the residual runs.

  Reproduction result: `source=10100 candidate=10000`.

  Reference:
  https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/mongodb/plan_predicate.go#L133-L168

- [ ] **BSON DateTime conversion accepts invalid values, can overflow, and
  does not honor TIMESTAMP scale.**

  Conversion performs millisecond-to-microsecond arithmetic before validating
  the MO temporal domain. Reproductions:

  - year 10000 is accepted as an invalid DATETIME;
  - BSON DateTime `math.MaxInt64` wraps to `1969-12-31 23:59:59`;
  - BSON `.100` remains fractional inside `TIMESTAMP(0)`.

  Reference:
  https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/mongodb/converter.go#L307-L320

- [ ] **Raw BSON batch limits do not bound decoded/vector memory.**

  `MaxBatchBytes` counts only raw BSON. The converter separately copies a
  varlen value into every mapped output vector, and multiple columns may map
  the same BSON path. A diagnostic test expanded one 256 KiB value mapped to
  64 columns into 16,778,752 bytes of vector allocation. With up to 4096
  columns, a legal near-16 MiB document has a tens-of-GiB theoretical peak.

  References:
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/colexec/mongoscan/mongoscan.go#L177-L234
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/mongodb/converter.go#L119-L180
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/mongodb/mapping.go#L105-L152

- [ ] **Snapshot/PITR restore can create orphan MongoDB mappings.**

  Bulk restore skips external tables, but `mo_mongodb_tables` is not in the
  mo_catalog restore skip policy. Restore can therefore clone a historical
  mapping row whose external table and table ID were not restored. The
  orphan dependency then prevents `DROP MONGODB CONNECTION`, leaving a
  successful restore with unmanageable catalog state. PITR follows the same
  path.

  References:
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/frontend/snapshot.go#L92-L135
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/frontend/snapshot.go#L1262-L1277
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/frontend/snapshot.go#L1324-L1344
  - https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/frontend/pitr.go#L1442-L1457

- [ ] **`max_by` varlen compaction can regress to
  `O(winners * chunk_groups)`.**

  Every winner runs compaction checks for all state vectors. Once varlen area
  capacity exceeds 1 MiB, each check scans the whole state vector, up to 8192
  groups, even when compaction is not needed. The existing compaction test has
  one group and does not exercise the multi-group complexity.

  Reference:
  https://github.com/matrixorigin/matrixone/blob/cf6dca31107769ad7855ebb45b1a6b48298284d6/pkg/sql/colexec/aggexec/maxby.go#L231-L289

## Secondary hardening

- [ ] Add per-key singleflight to cold client acquisition to avoid duplicate
  validation, TLS/SCRAM, Connect, and Ping work.
- [ ] Move best-effort post-commit client retirement to a bounded asynchronous
  queue. Local cleanup can wait up to 10 seconds and remote fanout up to 3
  seconds; cluster restore repeats this serially per account.
- [ ] Remove the MongoDB E2E `free_port` TOCTOU. Both selected ports are
  released immediately, and the MO port is not bound until after a complete
  build. An adjacent-allocation experiment reused the same port twice in
  20,000 pairs.

## Required fix properties

1. Use a typed, trusted MongoDB catalog discriminator; reject marker injection
   from generic external metadata and add a non-admin E2E test.
2. Disable temporal comparison/IN pushdown below millisecond scale unless a
   proven preimage-range translation is implemented.
3. Validate temporal range before arithmetic; apply strict/try-null semantics
   consistently and normalize TIMESTAMP to its declared scale.
4. Enforce a decoded/vector byte budget during conversion.
5. Define an explicit restore policy: either skip historical mapping rows or
   recreate external tables and regenerate mappings with restored table IDs.
6. Track max_by live/stale varlen bytes incrementally and benchmark many groups
   with repeated winners.

## Verification already performed

- Diagnostic red tests reproduced the five authorization/correctness/memory
  failures above; the temporary files were removed afterward.
- Existing `pkg/sql/mongodb`, `pkg/sql/colexec/mongoscan`, and targeted
  `max_by` tests pass under `-race`.
- PR required CI, including MongoDB E2E, UT, SCA, and required BVT, passed.
  Upgrade compatibility jobs were skipped by the workflow.

Related: #26424

iamlinjunhong:
Harden catalog typing, temporal conversion, and predicate pushdown. Bound scan and max_by memory work, share client dials, retire clients asynchronously, and align restore plus local E2E lifecycle behavior.

Fixes #26485

## What type of PR is this?

- [ ] API-change
- [ ] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26485 

fix: resolve MongoDB external scan follow-ups


fengttt (commented):
Does this need a bvt test?   All these fixes seems do not have a real test case so how do we know this PR actually fixed anything at all?

iamlinjunhong:
> * #### [MongoDB Connector E2E / MongoDB E2E Local (pull_request)](/matrixorigin/matrixone/actions/runs/30620852695/job/91124802742?pr=26495)Successful in 7m

ut and ci [MongoDB Connector E2E / MongoDB E2E Local (pull_request)](https://github.com/matrixorigin/matrixone/actions/runs/30620852695/job/91124802742?pr=26495)Successful in 7m

heni02 (approved):

XuPeng-SH (approved):
Deep-reviewed exact head ae43f4a90c3468e6b3385a9575df28ba1e6c3b0c. No blocking findings.

Verified the fix as a set of system invariants rather than isolated cases:
- MongoDB classification now requires planner-owned durable typing for v2 plus an anchored envelope; generic external metadata cannot inject the marker, while anchored v1 remains a constrained compatibility path.
- Low-precision temporal mappings remain residual-only, BSON DateTime conversion rejects out-of-range years before MO arithmetic, and DATETIME/TIMESTAMP scale is applied consistently.
- Decoded vector growth is admitted incrementally against the batch budget, with row-level rollback and bounded one-document deferral.
- Per-key cold acquisition is singleflighted; ALTER/DROP generations, queue saturation, cancellation, shutdown, and remote fanout preserve bounded ownership and wakeup paths.
- Snapshot/PITR/TS restore share the MongoDB mapping skip policy, preventing orphan table-ID mappings.
- max_by uses incremental live/stale varlena accounting and amortized compaction instead of rescanning every group for every winner; merge, unmarshal, flush, OOM-preflight, and many-group cases are covered.

Local validation:
- go test: mongodb, mongoscan, aggexec, plan, compile, frontend, cnservice, test/mongodb
- go test -race: mongodb, mongoscan
- go vet: all changed owning packages
- max_by many-group repeated-winner benchmark: 36-38 us/op, 0 Go allocs/op on Apple M4
- git diff --check clean

MongoDB E2E Local is green on this exact head. Broader CI jobs are still running and were not awaited.

Issue #20022, opened 2024-11-13
[Bug]: config proxy cache connection mo hung


heni02:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Branch Name

2.0-dev

### Commit ID

3ecc49ab9

### Other Environment Information

```Markdown
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

测试场景：
tke proxy配置conn-cache-enabled = true ,测试工具配置为短连接，并发100点查测试一会儿后，客户端无法登陆mo
<img width="882" alt="企业微信截图_3ad5133a-58d3-4741-abdd-ee71c549462e" src="https://github.com/user-attachments/assets/71cc2170-1dcd-4140-be7a-d7fa1c42af05">
tke yaml proxy配置：
<img width="446" alt="企业微信截图_8d1f07b7-74e3-494f-847b-9bd6cd915b34" src="https://github.com/user-attachments/assets/9e4268e3-6d85-40f4-8c03-3032179e2f2a">

mo-load测试工具配置为短连接：（具体使用见复现步骤）
<img width="470" alt="企业微信截图_d791f333-476b-4229-a2c2-a05e06e72cc6" src="https://github.com/user-attachments/assets/ef2c017a-ca8a-4d6c-8a6c-235558052f68">

mo log：
https://grafana.ci.matrixorigin.cn/explore?panes=%7B%22pGq%22:%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mo-branch-nightly-2e5ddb165-20241109%5C%22%7D%20%7C%3D%20%60%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221731404462204%22,%22to%22:%221731411550749%22%7D%7D%7D&schemaVersion=1&orgId=1

### Steps to Reproduce

```Markdown
1.create database sysbench_db;
2.使用mo-sysbench工具加载sysbench100w点查数据
sysbench  --mysql-host=127.0.0.1 --mysql-port=6001 --mysql-user=dump --mysql-password=111 oltp_point_select.lua --mysql-db=sysbench_db --tables=10 --table_size=1000000 --threads=100 --time=30 --report-interval=10    prepare
[...]

PR #26497, opened 2026-07-31
fix(frontend): fence session reset from active requests

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #20022

Proxy can observe a terminal MySQL response and start `ResetSession` while the
original CN `RoutineManager.Handler` is still completing response accounting
and defers. The old lifecycle controller serialized reset/migration/close, but
did not serialize ordinary requests, so reset could close the old Session,
process, and transaction handler while Handler still owned them.

This change:

- adds a lightweight request lease to the existing lifecycle controller;
- makes `ResetSession` fail fast with the existing busy error while a request
  is active, so Proxy discards that backend and reconnects;
- lets migration wait with its caller context, while routine close keeps its
  existing cancel-without-request-wait behavior;
- rejects a new request before it reads Session if lifecycle work already owns
  the routine;
- makes `ExecRequest` panic recovery safely handle a nil transaction handler.

Proxy cache/generation/in-flight production logic is unchanged.

### Regression evidence

- Base: `4959b7cdd853923f60b168d9ef69e242a1cadc37`
  - `TestRoutineManagerResetSessionRejectsRequestAfterResponseWrite` failed:
    reset returned `nil` after the terminal response write while Handler was
    deliberately blocked before returning.
- PR head: `af17db6b3db4b19cf5b4a604395574ce21a4a602`
  - the same deterministic interleaving returns busy, preserves the old
    Session/proc/TxnHandler and SessionManager entry, then succeeds once after
    Handler exits and serves the next `COM_PING`;
  - the race regression passed 100 consecutive iterations;
  - lifecycle-first, migration wait/cancel, close-no-wait, replacement reuse,
    rejected Proxy cache entry, and nil-TxnHandler recovery paths pass.

### Validation

- `.agents/skills/mo-dev/scripts/mo-cgo-test ./pkg/frontend -count=1`: PASS
- `.agents/skills/mo-dev/scripts/mo-cgo-test ./pkg/proxy -count=1`: PASS
- complete `pkg/frontend` and `pkg/proxy` with `-race -count=1`: PASS
- focused request/reset interleaving with `-race -count=100`: PASS
- `go build -mod=readonly ./pkg/frontend ./pkg/proxy`: PASS
- `go vet -mod=readonly ./pkg/frontend ./pkg/proxy`: PASS
- `git diff --check`: PASS
- schema-v4 preflight: `PASS review=PASS validation=PENDING`
- current exact-head required CI: PASS, including Ubuntu UT, SCA, UT coverage,
  Proxy Compose BVT, pessimistic launch BVT, and coverage.

### Local deployment evidence

The local topology was launched with the real Proxy path
(`mo-service -with-proxy`), 2 CNs, and true new client connections.

- direct-CN control, 100 workers / 2 s / 2 min: 5,987 queries, 0 failures;
- Proxy cache disabled, 100 workers / 2 s / 2 min: 5,988 queries and one
  isolated 5 s context deadline; all Proxy/CN probes passed;
- Proxy cache enabled after warm-up, 100 workers / 1 s / 60 s: 5,995 queries,
  0 failures, all 36 endpoint probes passed, and backend sessions converged
  from 119 to 17/18 rather than growing with total connections;
- no CN/Proxy panic, pod/process restart, `wait active`, or
  `createTxnOpUnsafe` was observed in these valid runs.

A malformed first result was observed in two cold-start runs immediately after
ports opened. A later diagnostic run without s2c batching passed 5,984 queries,
but it ran after the cluster was fully warm and therefore does not prove a
buffering defect. This cold-start observation is tracked as unresolved
acceptance evidence and is not used to justify an unrelated Proxy code change.

The complete 10-minute local soak is still pending: the initial 100
connections/s setup exhausts the macOS loopback ephemeral-port range, while
the later 50 connections/s rerun was interrupted by the local disk safety
threshold. TKE validation remains pending.

### QA / BVT

- QA required: yes
- BVT: N/A — regular SQL BVT cannot enable Proxy connection cache or create
  independent true short connections. Deterministic Frontend/Proxy concurrency
  tests provide the in-repo regression coverage.
- Deployment validation pending:
  - local 2 CN + Proxy, warmed and readiness-gated, 100 concurrent true short
    connections for 10 minutes at a host-safe connection rate;
  - TKE 3 CN + 2 Proxy, 1000 vusers for 10 minutes;
  - TKE 300 terminals for 70 minutes, with independent Proxy/CN login probes,
    no panic/restart/continuous login failure, and bounded backend sessions.
- After merge, keep #20022 open under `phase/testing` for `Ariznawlll` to run
  the deployment matrix; link `matrixorigin/mo-nightly-regression#835` for the
  recurring short-connection coverage.

### Current-head CR follow-up (2026-08-05)

Head `e70c5a45ca7617ab668d09150184814f98218078` also routes both rolling-upgrade
`MigrateConnFrom` compatibility actions through the Routine lifecycle admission
before reading or mutating Session user-lock state. Skip now checks the current
admitted Session after a request/reset owner exits; Enable mutates only that
admitted Session generation.

- Parent-head reproduction: request-first/reset-first × Skip/Enable all failed
  against `1bc7d0021fb1fbcbd00ce4af3b703711a90d9e38`.
- Focused and adjacent migration tests: PASS.
- Focused race stress: request-first `T=0.21s`, `B=30s`, `N=100` PASS;
  reset-first `T=0.22s`, `B=30s`, `N=100` PASS.
- `pkg/frontend`: list/build/vet/test and full-package race PASS.
- Direct consumer `pkg/cnservice`: list/build/vet/test PASS.
- `git diff --check`: PASS.
- Schema-v4 preflight: `PASS review=PASS validation=PENDING` at exact head
  `e70c5a45ca7617ab668d09150184814f98218078`.
- BVT: N/A — these legacy QueryService `MigrateConnFrom` action values are
  selected by old Proxy versions and cannot be invoked through SQL BVT. The
  deterministic `RoutineManager.MigrateConnectionFromWithContext` interleaving
  tests exercise the production RPC entry and assert the final lock/Session
  state. New exact-head CI is pending.


XuPeng-SH (changes requested):
Blocking: the new request/lifecycle fence is bypassed by the legacy MigrateConnFrom actions.

In `RoutineManager.MigrateConnectionFromWithContext`, `MigrateConnFromSkipUserLevelLockRelease` reads the current session user-lock state before `routine.migrateConnectionFromWithContext` acquires `mc`. A reachable ordering is:

1. A SQL request owns `requestInProgress` but has not acquired its user lock yet.
2. The legacy migration RPC checks `UserLevelLocksForMigration` and observes no locks, then blocks in `beginOperationWithContext`.
3. The SQL request completes `GET_LOCK` and releases the request lease.
4. Migration resumes, sets `userLevelLocksMigrated = true`, and succeeds without rechecking the lock state.

That defeats the safety check and can let a connection migrate with a held lock; close then takes the migrated-lock discard path instead of releasing it. The `MigrateConnFromEnableUserLevelLockRelease` branch is also completely outside lifecycle admission and can mutate an old session while reset/close owns it. These compatibility actions remain reachable during rolling upgrades.

Please move both action-specific session reads/mutations behind the same lifecycle admission (with the check performed after admission), and add request-first/reset-first interleaving tests for these two actions.

VioletQwQ-0:
@XuPeng-SH Addressed on `e70c5a45ca7617ab668d09150184814f98218078`.

Both legacy `MigrateConnFrom` actions now enter the same Routine lifecycle admission before any Session access:

- `MigrateConnFromSkipUserLevelLockRelease` re-reads user-lock state from the current admitted Session and sets `userLevelLocksMigrated` only after that check succeeds.
- `MigrateConnFromEnableUserLevelLockRelease` clears the flag only on the current admitted Session generation.

Regression coverage calls the production `RoutineManager.MigrateConnectionFromWithContext` entry and covers request-first/reset-first interleavings for both actions. The four cases failed on parent head `1bc7d0021fb1fbcbd00ce4af3b703711a90d9e38` and pass with this repair. The Skip cases add a lock or publish a locked replacement while migration is queued; the Enable cases assert that the old/current Session is untouched until admission.

Validation:

- focused and adjacent migration tests: PASS
- request-first race stress: `T=0.21s`, `B=30s`, `N=100`: PASS
- reset-first race stress: `T=0.22s`, `B=30s`, `N=100`: PASS
- full `pkg/frontend` normal + race, build, and vet: PASS
- direct consumer `pkg/cnservice` test, build, and vet: PASS
- `git diff --check`: PASS
- schema-v4 preflight: `PASS review=PASS validation=PENDING` on exact head `e70c5a45ca7617ab668d09150184814f98218078`
- BVT: N/A — old-Proxy QueryService action selection is not reachable through SQL BVT; the deterministic production-RPC-entry tests are the equivalent regression coverage

The PR remains Ready for Review. New exact-head CI is pending.

XuPeng-SH (approved):
The previous lifecycle-fence blocker is resolved on `e70c5a45ca7617ab668d09150184814f98218078`.

Both legacy `MigrateConnFrom` actions now acquire the same request/lifecycle admission before reading or mutating the session. In particular, the skip action rechecks user-level locks on the admitted current session, and the enable action only changes that same session generation. The new request-first and reset-first tests cover both actions and would expose the stale/pre-admission access from the prior head.

I also rechecked the surrounding ownership paths: reset remains fail-fast while a request owns the session; migration waits with caller cancellation; close seals admission and cancels active work while preserving its existing non-waiting request-close behavior. I found no remaining correctness or lifecycle blocker in this diff.

Local exact-head validation passed:

- all newly added focused tests
- the same focused set under `-race`
- full `pkg/frontend`
- direct consumer `pkg/cnservice`
- compile-only checks and `go vet` for both packages
- `git diff --check`

Approve.

PR #26499, opened 2026-07-31
fix(parser): reject out-of-range branch diff limits

### LINKED ISSUE #26038: [Bug]: DATA BRANCH DIFF OUTPUT LIMIT returns an internal panic for uint64 literals
### Summary

`DATA BRANCH DIFF ... OUTPUT LIMIT` returns an internal panic error when the integer literal is larger than `math.MaxInt64` but still fits in `uint64`.

### Version

- Branch: `main`
- Commit: `e15943929885c2c5c83daed90f0f8caeb9a1f0d0`

### Minimal reproduction

```sql
DATA BRANCH DIFF missing_a AGAINST missing_b
  OUTPUT LIMIT 9223372036854775808;
```

No tables are required because the failure occurs while parsing the statement.

### Actual behavior

```text
ERROR 20101 (HY000): internal error: panic interface conversion:
interface {} is uint64, not int64

runtime.panicdottypeE
...
mysql_sql.y:9380
```

The behavior reproduced in 3 out of 3 runs. The service remained available after each statement.

Control results:

- `OUTPUT LIMIT 9223372036854775807` parses and executes normally.
- A literal larger than `uint64` is rejected as a regular SQL parser error.

### Expected behavior

An out-of-range limit should be rejected as a normal syntax or invalid-input error without an internal panic and stack details.

### Code observation

The `diff_output_opt` grammar action uses `x := $3.(int64)`. Integer tokens in the unsigned range are represented as `uint64`, so the direct type assertion fails.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26038

`DATA BRANCH DIFF ... OUTPUT LIMIT` stores its limit as `int64`, but the MySQL lexer represents integer literals above `math.MaxInt64` as `uint64`. The grammar action directly asserted the token as `int64`, so values in the uint64-only range caused an internal panic instead of a SQL parser error.

This PR:

- converts the token through the existing checked `util.GetInt64` path and reports `OUTPUT LIMIT is out of range` when the literal cannot fit in `int64`;
- regenerates the MySQL parser;
- adds focused parser coverage for `MaxInt64`, `MaxInt64 + 1`, `MaxUint64`, and `MaxUint64 + 1`;
- adds a BVT case proving `MaxInt64` reaches semantic validation while uint64-only values return normal parser errors without panic details.

Validation:

- pre-fix focused test reproduced `panic: interface conversion: interface {} is uint64, not int64`;
- `make` in `pkg/sql/parsers/dialect/mysql` (parser regenerated with zero conflicts);
- `go build ./pkg/sql/parsers/dialect/mysql`;
- `go vet ./pkg/sql/parsers/dialect/mysql`;
- focused parser regression test;
- focused race stress: measured `T=0s`, budget `B=30s`, therefore `N=100`; passed with `-race -count=100`;
- full `pkg/sql/parsers/dialect/mysql` package passed with `-race -count=1`;
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=240s ./pkg/sql/parsers/...`;
- BVT `branch_output_limit_range`: 3/3 passed, 100% success; generated results were manually checked against the intended parser semantics.

Residual risk: none known. The AST and execution contract remain `int64`; this change rejects values outside that existing contract rather than widening or wrapping them.


XuPeng-SH (approved):
Deep-reviewed current head cac089ce8479ffcc3287e2780cce4c7a15cda7c9. Checked the lexer int64/uint64 boundary, parser error path, BVT expectations, and forced parser regeneration; the generated file is reproducible. The boundary test passed race runs 1+100; the full parser package race suite, build, and vet passed. No blocking issues found.

PR #26501, opened 2026-07-31
fix(cte): support UNION DISTINCT in recursive CTE

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25955

Recursive CTE binding previously treated only `UNION ALL` as recursive. A plain
`UNION` or explicit `UNION DISTINCT` was bound as a non-recursive CTE and
incorrectly rejected when its recursive member referenced the CTE.

This change:

- recognizes `UNION` and `UNION DISTINCT` recursive members;
- propagates the distinct mode through the plan to `MergeCTE`;
- keeps a query-scoped full-row seen set across the anchor and every recursive
  iteration, so duplicate rows are not fed into later iterations and recursion
  terminates at the correct fixpoint;
- clears the seen set on `Reset` and `Free` while leaving the `UNION ALL` path
  unchanged;
- rejects mixed `UNION ALL` and `UNION DISTINCT` recursive-member chains whose
  left-associative semantics cannot be represented by the current execution
  topology;
- adds planner, operator, and distributed regression coverage, including NULL,
  multi-column, cross-batch, reset, fixpoint termination, and `UNION ALL`
  control cases.

Validation:

- `make pb`
- focused planner and `MergeCTE` unit tests
- full unit tests for `pkg/sql/plan`, `pkg/sql/compile`, and
  `pkg/sql/colexec/mergecte`
- focused regression tests with `-race -count=100`
- `go vet` for all modified Go packages
- `make`
- distributed BVT: 5/5 passed
- modified-function coverage is at least 75%
- `git diff --check`

The full-package race build was also attempted, but the local machine exhausted
disk space while compiling dependencies (`no space left on device`); no test
failure or race report was produced.


XuPeng-SH (approved):
Deep review completed on eec58f8058babea45e67322b73712f7d1a98974f. Binder mode detection, mixed-mode rejection, protobuf/deep-copy propagation, MergeCTE full-row seen-set semantics (including NULLs and cross-batch duplicates), empty-frontier fixpoint behavior, and Reset/Free ownership are consistent. Local build/vet and non-race plan/compile/pipeline/operator tests passed; all 4 changed Go tests passed individually under race with computed repeat counts (7–8), followed by full mergecte and sql/plan race tests. CI reruns supersede the earlier cancelled/failed attempts and are successful.

XuPeng-SH:
@ck89119 conflict

heni02 (approved):

XuPeng-SH:
@ck89119 conflict

PR #26502, opened 2026-07-31
fix(logservice): make snapshot initialization idempotent

### LINKED ISSUE #25889: [Bug]: Repeated snapshot initialization duplicates entries and fills the export quota
## Summary

`snapshotManager.Init` appends every on-disk exported snapshot to the existing in-memory record each time it is called. It does not clear the previous view or deduplicate an equal snapshot index.

As a result, retrying replica initialization makes one physical snapshot appear many times in memory. After 20 retries, one snapshot is counted as 20 snapshots, which reaches the default `MaxExportedSnapshot` quota.

## Affected version

Latest `main` tested at `862d9d30c2cd7a55995fd0bae0951af0809f5752`.

## Production path

Both `store.startReplica` and `store.startNonVotingReplica` call `snapshotMgr.Init(shardID, replicaID)` before `NodeHost.StartReplica`.

A repeated/retried bootstrap command therefore appends the same snapshots before Dragonboat can return `ErrShardAlreadyExist`. The inflated count is then used by `shouldDoExport`:

```go
if l.snapshotMgr.Count(shardID, replicaID) >= l.cfg.MaxExportedSnapshot {
    return false
}
```

This can prematurely stop snapshot export and delay the export/import path used for log truncation and WAL compaction.

## Root cause

`snapshotManager.Init` sorts the directory indexes and calls `addLocked` for each one, but it never replaces the existing in-memory record. `snapshotRecord.add` also permits an index equal to the current last index.

## Expected behavior

Initialization should be idempotent for a `(shardID, replicaID)`: the in-memory view should reflect the unique on-disk snapshot directories, regardless of how many times initialization is retried.

No matching open issue was found with searches for snapshot initialization idempotency, duplicate snapshot counts, bootstrap retry, and `MaxExportedSnapshot` quota exhaustion.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #25889.

### Root cause

`snapshotManager.Init` loaded sorted on-disk snapshot directories by appending them to the existing in-memory `snapshotRecord`. Retrying replica startup therefore counted the same physical snapshot repeatedly and could fill `MaxExportedSnapshot`, suppressing later exports.

There was a second interleaving in the export path: Dragonboat can finish writing a snapshot directory before `snapshotManager.Add` runs. If a retrying replica start calls `Init` in that window, `Init` loads the new index and the delayed `Add` appended it again because `snapshotRecord.add` accepted an index equal to the current latest index.

### Changes

- Rebuild each `(shardID, replicaID)` record from the current validated disk view and replace the old record atomically under the snapshot-manager lock.
- Return replaced `snapshotItem` objects to their pool after they become inaccessible.
- Treat adding the already-recorded latest index as an idempotent success, preserving the successful export flow without duplicating quota entries.
- Add focused coverage for 20 repeated initializations, the `Init` then delayed equal-index `Add` ordering, changed disk contents, sorted reconstruction, and an empty disk view.

No public API, configuration default, or on-disk format changes.

### Validation

Pre-fix, `TestInitSnapshotMgrReplacesInMemoryView` failed deterministically: the second initialization increased a two-directory view to three items, and a later disk refresh retained a removed directory and grew the view to five items.

Post-fix on `main@2902d56d2`:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -v -count=1 -timeout=120s -run '^TestInitSnapshotMgr(ReplacesInMemoryView)?$' ./pkg/logservice`
- Race test selection was enumerated with `-list` and contained `TestAddSnapshot`, `TestInitSnapshotMgr`, and `TestInitSnapshotMgrReplacesInMemoryView`.
- Each selected test measured `T=0s` in `go test -race -json` output. With a 30-second stress budget, the adaptive rule selected `N=100`; all three exact-test `-race -count=100` runs passed.
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=10m -skip '^TestNewServiceClosesStoreOnReplicaStartFailure$' ./pkg/logservice` (PASS, 128.778s)
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=15m -skip '^TestNewServiceClosesStoreOnReplicaStartFailure$' ./pkg/logservice` (PASS, 133.148s)
- CGo-configured `go build -mod=readonly ./pkg/logservice`
- CGo-configured `go vet -mod=readonly ./pkg/logservice`

### Residual risks / test baseline

The complete owning-package command without `-skip` has one unrelated current-`main` failure: `TestNewServiceClosesStoreOnReplicaStartFailure` expects duplicate metadata to reach Dragonboat and return `ErrShardAlreadyExist`, while the newer zombie-replica membership check skips both starts. The identical focused command fails on clean `HEAD` with no changes applied. This PR does not modify that unrelated service-start test or behavior.

No residual correctness risk was identified in the changed snapshot-manager paths. Initialization is serialized with snapshot mutations so a concurrent completed export cannot be lost during record replacement; equal-index `Add` reconciles the directory-written-before-Add ordering.


XuPeng-SH (approved):
Deep review completed on head 6961ff16d217e947618f43b92216c944f4559181. No blocking findings.

The fix closes both duplication paths: repeated Init now atomically replaces the per-replica disk-derived view, and a delayed Add of the already-loaded latest index is idempotent. All manager readers/mutators are serialized by the same RWMutex; snapshotItem pointers do not escape the lock, so releasing the replaced record back to the pool is safe.

Q1 failure: Stat/List failures leave the published record untouched. Q2 success: the new record is published once and the inaccessible old items are released once. Q3 concurrency: both directory-before-Add and directory-after-Init-list interleavings converge without duplicate or lost entries.

Performance impact is confined to low-frequency replica initialization: O(number of on-disk snapshots) rebuild/release, with no truncation hot-path regression; duplicate Add is cheaper.

Validation performed locally:
- focused snapshot regression under race, count=50
- forced Init/list vs export/Add interleaving under race, count=100
- full ./pkg/logservice under race, PASS (131.7s)
- clean diff/check and latest-main merge compatibility

The red Ubuntu UT job contains broad unrelated lease/checkpoint/embedded-cluster failures; the owning package passes independently.

Issue #26489, opened 2026-07-31
[Bug]: Test_TxnExecutorExec panics after HAKeeper connection reset during CN bootstrap


ck89119:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Branch Name

main

### Commit ID

a5d23de54752df126509e8b10cc68035c15f9ab8

### Other Environment Information

- Hardware parameters: GitHub Actions Ubuntu/x86 runner
- OS type: Linux/amd64
- Others: MatrixOne ALL CI, UT job 91069918726

### Actual Behavior

`Test_TxnExecutorExec` intermittently panics while starting its private embedded
single-CN cluster:

```text
panic: writeto tcp4 127.0.0.1:43814->127.0.0.1:18020:
read tcp4 127.0.0.1:43814->127.0.0.1:18020:
read: connection reset by peer
```

The panic originates from `cnservice.handleBootstrapErr` during
`embed.(*operator).startCNServiceLocked`. Before the reset, LogService reports
Dragonboat proposal/tick timeouts and delayed `LocalTick` batches, followed by
failed CN/TN heartbeats.

This is a recurrence after #26299 and #26316, but with a different terminal
failure. The earlier issue failed by returning a HAKeeper startup timeout;
the current failure panics from CN bootstrap after the HAKeeper backend
connection is reset.

The failure is unrelated to PR #26448. `Test_TxnExecutorExec` creates its own
cluster with `embed.NewCluster`, while that PR changes the shared
`RunBaseClusterTests` configuration and authenticated issue-test routing. The
PR does not modify the CN bootstrap or `Test_TxnExecutorExec` paths.

### Expected Behavior

Transient HAKeeper or LogService unavailability during embedded-cluster startup
should return a start error and clean up partial services. It should not panic
the UT process.

### Steps to Reproduce

```text
1. Run the Ubuntu/x86 MatrixOne UT job under CI load.
2. Execute Test_TxnExecutorExec in pkg/tests/txnexecutor.
3. Observe CN bootstrap while Dragonboat tick/proposal processing is delayed.
```

### Additional information

- Workflow run:
  https://github.com/matrixorigin/matrixone/actions/runs/30603151585
- Failed job:
  https://github.com/matrixorigin/matrixone/actions/runs/30603151585/job/91069918726
- Failed case: `Test_TxnExecutorExec`
- Previous tracker: #26299
- Previous fix: #26316

PR #26503, opened 2026-07-31
fix(cnservice): return bootstrap failures

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #26489

CN bootstrap previously panicked for every failure except `context.Canceled`.
A transient HAKeeper connection reset therefore terminated the test process.

Returning the error safely requires complete lifecycle ownership. Bootstrap
must run after the embed operator owns the CN; constructor-opened SQL listeners
and bootstrap-created runtime services must be closed on rollback; and trace
consumers must remain alive until every producer has stopped.

This change:

- returns bootstrap failures while preserving their context causes;
- moves bootstrap to `Service.Start()` after operator ownership transfer;
- closes SQL listeners even when `MOServer.Start()` was never called;
- makes CN own bootstrap, auto-increment, and transaction-trace services, with
  generation-safe runtime cleanup;
- closes pipeline admission atomically with handler registration, including
  requests already read by a goetty session when shutdown starts;
- registers a server-owned cancellation context for every admitted pipeline and
  cancels all of them before waiting during shutdown;
- gives fragmented message caches an independent cross-handler lifetime and
  releases every transferred fragment context on assembly, timeout, or session close;
- cancels rejected and other unowned MORPC requests exactly once;
- stops frontend, tasks, pipeline RPC, and transaction clients before closing
  trace; and
- waits for all canceled pipeline handlers, then closes the auto-increment
  producer before the trace consumer.

Validation:

- real bootstrap rollback, SQL listener reuse, concurrent trace shutdown, and
  real RPC admission-race tests passed 100 times
- a real TCP/MORPC fragmented pipeline survives two cache cleanup scans and
  assembles all fragments
- cache delete, request timeout, and session shutdown release transferred
  fragment contexts exactly once
- focused lifecycle tests pass normally and under the race detector
- full `pkg/common/morpc` and `pkg/cnservice` tests passed
- full `pkg/frontend` and runtime tests, plus embedded rollback tests, passed
- build and vet for affected packages passed
- targeted golangci-lint passed with 0 issues
- focused coverage is 94.6% for `handleRequest`, 88.9% for fragment handling,
  and 85.7% for cancelable cache creation


XuPeng-SH (changes requested):
Deep review on `53de8f3c39f20da8d23f0549e6e808dc530bcbe9` found two lifecycle blockers.\n\n1. **Bootstrap is retired while frontend consumers are still live.** `service.Close` invokes `closeBootstrapService` before `stopFrontend`, and `closeBootstrapService` sets `s.bootstrapService = nil` before calling `Close`. At the same time active/new frontend sessions call `GetFinalVersion`, `CheckTenantUpgrade`, and `UpgradeTenant` through `BaseService`, all of which dereference `s.bootstrapService` without synchronization. A concrete shutdown interleaving is: Close stores nil -> an authentication/upgrade request enters before the SQL listener/session is stopped -> `GetFinalVersion` invokes a method on the nil interface and panics (and the field read/write is itself a race). Simply moving the step may still be insufficient if frontend stop closes sessions without joining their handlers; the bootstrap generation must remain stable until every consumer is drained, or access/retirement must be synchronized. Please add a deterministic barrier-based close-vs-upgrade/auth regression.\n\n2. **The new returned Start error is not cleanup-safe for all CN owners.** The embedded operator now transfers ownership before `Start`, but `pkg/tests/service.(*cnService).Start` leaves status as `ServiceInitialized` on error while its `Close` only closes `ServiceStarted`. Therefore the exact bootstrap reset now returned by this PR leaves the constructor-opened SQL listener plus bootstrap-created incr/trace services and clients alive; a later cluster Close is a no-op for that CN. `cmd/mo-service.startCNService` also panics on `s.Start()` without first closing the constructed service. Please close the ownership loop for every direct `cnservice.NewService` caller (or make failed Start transactionally roll itself back), and test failure -> cleanup -> SQL address reuse/runtime globals removed.\n\nThe MORPC cache callback-outside-lock change and pipeline `closing + WaitGroup` admission boundary look correct; focused CN/MORPC lifecycle tests passed under race locally. While adding the missing regressions, please replace the 2.2s sleep in `TestFragmentedPipelineCacheSurvivesHandlerRelease` with an observable scan/barrier: elapsed time does not prove that either cleanup scan actually ran under CI scheduling.

XuPeng-SH (changes requested):
One lifecycle blocker remains at `service.Start` / `service.Close`:

`Start` rolls a failed attempt back through `s.Close()`, which permanently consumes `closeOnce`, but `Start` has no state/admission guard. A caller that retries the now-returned bootstrap error creates and publishes a new auto-increment service and txn-trace service; when that second attempt fails, the deferred `Close` is a no-op, so the new goroutines/runtime globals remain live and every later `Close` is also unable to reclaim them. The `Service` interface does not document a single-call-only contract or reject a start on a closed instance. The same missing state boundary also permits teardown and publication to overlap if `Start` and `Close` race.

I reproduced the sequential case on exact head `ac5a2b391f2a1c1653c066d0e4cb69735858b5a6` with a real `Start -> bootstrap error -> Start -> bootstrap error` counterexample: `runtime.AutoIncrementService` was still present after the second return (the test failed in 0.05s). Please add an explicit one-shot lifecycle state gate (or make retries transactionally safe), and keep this counterexample as a deterministic regression test.

The previous listener/global cleanup, bootstrap-consumer synchronization, fragmented-cache ownership, cancellation-outside-lock, and trace-producer ordering issues are otherwise closed. The new targeted CN/MORPC/frontend/runtime tests pass locally under `-race`, and all current CI checks are green.

XuPeng-SH (approved):
Deep re-review completed on af1b2f809955c01eed0abc7ab1103dac87ca3ec4. The previous lifecycle blockers are resolved and I found no new merge blocker.

The one-shot Start/Close state gate now makes failed startup cleanup terminal and rejects reuse after close; bootstrap consumers remain protected until retirement; pipeline admission, cancellation, handler draining, trace shutdown, and fragmented-message cancel ownership form a complete shutdown closure. Runtime cleanup is generation-safe through CompareAndDelete, and retained handler/cache state is bounded and reclaimed on completion, timeout, or session close.

Fresh local validation passed: adaptive focused race stress for all changed lifecycle transitions, Test_TxnExecutorExec under race with count=2, full race runs for cnservice, morpc, frontend, runtime, and txnexecutor, plus build and vet for the production packages.

PR #26504, opened 2026-07-31
fix(frontend): preserve non-finite data branch floats

### LINKED ISSUE #26072: [Bug]: Data Branch SQL materialization cannot preserve NaN and infinity float values
### Describe the bug

Data Branch operations that materialize rows through generated SQL cannot write valid `DOUBLE` values representing NaN or positive/negative infinity.

Plain `DATA BRANCH DIFF` returns these values correctly, but `DIFF ... OUTPUT AS`, MERGE, and PICK convert them to bare `NaN`, `+Inf`, or `-Inf` tokens. Those tokens are parsed as column expressions rather than floating-point values.

### Steps to reproduce

```sql
drop database if exists g4d_float_branch;
create database g4d_float_branch;
use g4d_float_branch;

create table base_t(id int primary key, f double);
insert into base_t values (1,1.5);
data branch create table src_t from base_t;
data branch create table dst_t from base_t;

insert into src_t values
  (2,cast('NaN' as double)),
  (3,cast('Inf' as double)),
  (4,cast('-Inf' as double));

-- This returns all three rows correctly.
data branch diff src_t against dst_t;

-- These fail when the values are materialized through generated SQL.
data branch diff src_t against dst_t output as diff_out;
data branch merge src_t into dst_t;
```

PICK fails in the same way on a fresh source/destination pair:

```sql
insert into pick_src values (2,cast('NaN' as double));
data branch pick pick_src into pick_dst keys(2);
```

### Actual behavior

- Plain DIFF returns `NaN`, `+Infinity`, and `-Infinity`.
- OUTPUT AS fails with an error such as `ambiguous column reference 'Inf'`; the incomplete result table is cleaned up.
- MERGE fails with the same class of error and leaves the destination unchanged.
- PICK fails on `NaN` and leaves the destination unchanged.

### Expected behavior

Every Data Branch operation should preserve all floating-point values accepted by MatrixOne. Generated SQL should use a representation that round-trips NaN and both infinity values, such as an explicit cast from a quoted value.

### Reproducibility

Reproduced in 3/3 runs for OUTPUT AS, MERGE, and PICK. Plain DIFF succeeded in all runs.

### Code context

`formatValIntoString` in `pkg/frontend/data_branch_helpers.go` formats `FLOAT` and `DOUBLE` with `strconv.AppendFloat`. Finite values are valid numeric literals, but the special outputs `NaN`, `+Inf`, and `-Inf` are appended without quoting or an explicit cast. The same formatter feeds OUTPUT AS and the MERGE/PICK SQL appenders.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26072

### Root cause

Data Branch converts rows back into SQL for MERGE, PICK, and portable output. The shared float formatter emitted bare `NaN`, `+Inf`, and `-Inf` tokens, which the parser treated as expressions rather than FLOAT/DOUBLE values.

Generated SQL and conflict matching also lost MatrixOne's bit-preserving FLOAT/DOUBLE storage identity:

- Scalar equality cannot match NaN, collapses signed zero, and cannot distinguish NaN payloads.
- Textual NaN casts canonicalize payload bits, while `cast(-0 as double)` produces positive zero.
- Applying a same-key update as DELETE plus INSERT can remove or collide with a bit-distinct peer.
- PICK initially bypassed the exact update path, and UPDATE alone could not restore a source update when the destination independently deleted the row.
- Simple FLOAT/DOUBLE primary-key duplicate checks and fuzzy-filter collision recovery still consumed scalar keys, so a direct exact-key restore could fail or expose serialized bytes in duplicate diagnostics.
- Hash-diff conflict sorting and equality also consumed raw scalar FLOAT/DOUBLE keys, collapsing crossed independent changes on signed zero or distinct NaN payloads into false conflicts.

### Changes

- Format non-finite FLOAT/DOUBLE values as typed casts and NaN payloads/negative zero as bit-preserving `bit_cast(unhex(...))` literals.
- Use `serial(...)` as the shared exact identity for FLOAT/DOUBLE key components in LCA probes, staged deletes, portable predicates, MERGE, and PICK. Non-float key behavior is unchanged.
- Preserve update pairing. Same-key FLOAT/DOUBLE updates use an exact-key UPDATE; genuine inserts remain direct INSERTs.
- Apply PICK conflict policy before materialization. FAIL errors, SKIP preserves the destination state, and ACCEPT restores a proven-missing source row with one direct INSERT.
- Make simple FLOAT/DOUBLE primary-key DEDUP, constraint, intra-batch, and fuzzy-filter checks compare serialized identities. Fuzzy-filter allocation now follows the projected identity type, background collision SQL groups on `serial(pk)`, and duplicate errors decode the original FLOAT/DOUBLE value.
- Validate staged key metadata before emitting SQL and share the exact-key logic across MERGE, PICK, and portable output.
- Give simple FLOAT/DOUBLE hash-diff conflict matching one shared bit-preserving IEEE key order for both sorting and equality; composite serialized keys retain their existing exact byte order.

### Tests

Verification on head `12db1fb854`, based on `c58cd292d3`:

- Full controlled CGo tests for `pkg/frontend`, `pkg/sql/plan`, `pkg/sql/colexec`, and `pkg/sql/compile` — pass.
- Controlled `go vet` for those four packages — pass.
- Focused exact-key tests cover FLOAT/DOUBLE signed zero, distinct NaN payloads, null/const vectors, exact ordering, and companion-column permutation.
- Coverage-mode package tests confirm malformed serialized-key errors, compound/hidden fuzzy-check SQL modes, projected-type fallback, and the background exact-float duplicate decode path are exercised.
- `make build` — pass.
- `git diff --check` — pass.
- Public `branch_float_special_values.sql` regression on this exact head — 256/256 statements pass.
- On the equivalent production head before the test-only coverage commits: `alter_table_change_column.sql` 923/923 and `alter_table_modify_column.sql` 764/764 statements pass.

The public Data Branch regression covers FLOAT, DOUBLE, and composite keys; finite values, infinities, signed zero, and distinct NaN payloads; MERGE, PICK, portable SQL, and no-PK predicates; FAIL/SKIP/ACCEPT when source update conflicts with destination delete; and crossed independent updates/deletes on bit-distinct signed-zero and NaN keys. Exact `hex(serial(...))` assertions verify both changed and untouched keys.

### CI follow-up

The failed proxy BVT on head `ad38897235` was caused by this PR: exact serialized FLOAT/DOUBLE DEDUP identities leaked into duplicate-entry messages in ALTER TABLE cases. This head decodes those identities for diagnostics and closes the same identity contract in fuzzy-filter collision recovery. That run's Coverage failure only propagated the producer failure.

The later Coverage-only failures were also PR-caused. Head `0570eb78b8` reported 178/240 (74.17%); after the first test-only follow-up, CI's merged profile on `e91247f83a` reported 175/240 (72.92%) because the background fuzzy-check path remained uncovered. The matching executor/result boundary test adds 14 covered statements, taking the observed 175/240 baseline above the strict 75% gate.

Unrelated Ubuntu/x86 failures remain tracked in #26583 and #26632 through the single machine-readable blocker comment; this PR contains no workaround for them.

### Residual risk

#26432 still tracks the engine-wide resident/spill typed-key inconsistency. This PR does not change that general engine contract. Data Branch and simple FLOAT/DOUBLE primary-key constraint paths avoid raw float identity matching by using exact bit/serialized encodings. Non-float paths are unchanged.

### DIFF
--- a/pkg/frontend/data_branch_hashdiff.go
+++ b/pkg/frontend/data_branch_hashdiff.go
@@ -1723,10 +1723,11 @@ func findDeleteAndUpdateBat(
 							return err2
 						}
 						if err2 = send(batchWithKind{
-							name:  tblName,
-							side:  side,
-							batch: updateBat,
-							kind:  diffInsert,
+							name:       tblName,
+							side:       side,
+							batch:      updateBat,
+							kind:       diffInsert,
+							fromUpdate: tblStuff.def.pkKind != fakeKind,
 						}); err2 != nil {
 							return err2
 						}
@@ -2094,6 +2095,7 @@ func diffDataHelper(
 			tarBat        *batch.Batch
 			baseBat       *batch.Batch
 			baseDeleteBat *batch.Batch
+			tarUpdateBat  *batch.Batch
 			tarTuple      types.Tuple
 			baseTuple     types.Tuple
 			checkRet      databranchutils.GetResult
@@ -2101,6 +2103,11 @@ func diffDataHelper(
 
 		tarBat = tblStuff.retPool.acquireRetBatch(tblStuff, false)
 		baseBat = tblStuff.retPool.acquireRetBatch(tblStuff, false)
+		defer func() {
+			if tarUpdateBat != nil {
+				tblStuff.retPool.releaseRetBatch(tarUpdateBat, false)
+			}
+		}()
 
 		if err2 = cursor.ForEach(func(key []byte, row []byte) error {
 			select {
@@ -2176,10 +2183,13 @@ func diffDataHelper(
 							if baseDeleteBat == nil {
 								baseDeleteBat = tblStuff.retPool.acquireRetBatch(tblStuff, false)
 							}
+							if tarUpdateBat == nil {
+								tarUpdateBat = tblStuff.retPool.acquireRetBatch(tblStuff, false)
+							}
 							if err2 = appendTupleToBat(ses, baseDeleteBat, baseTuple, tblStuff); err2 != nil {
 								return err2
 							}
-							if err2 = appendTupleToBat(ses, tarBat, tarTuple, tblStuff); err2 != nil {
+							if err2 = appendTupleToBat(ses, tarUpdateBat, tarTuple, tblStuff); err2 != nil {
 								return err2
 							}
 						} else {
@@ -2206,11 +2216,28 @@ func diffDataHelper(
 
 		if baseDeleteBat != nil {
 			if stop, err3 := emitBatch(emit, batchWithKind{
-				batch: baseDeleteBat,
-				kind:  diffDelete,
-				name:  tblStuff.baseRel.GetTableName(),
-				side:  diffSideBase,
+				batch:      baseDeleteBat,
+				kind:       diffDelete,
+				name:       tblStuff.baseRel.GetTableName(),
+				side:       diffSideBase,
+				fromUpdate: true,
 			}, false, tblStuff.retPool); err3 != nil {
+				return err3
+			} else if stop {
+				return nil
+			}
+		}
+
+		if tarUpdateBat != nil {
+			stop, err3 := emitBatch(emit, batchWithKind{
+				batch:      tarUpdateBat,
+				kind:       diffInsert,
+				name:       tblStuff.tarRel.GetTableName(),
+				side:       diffSideTarget,
+				fromUpdate: true,
+			}, false, tblStuff.retPool)
+			tarUpdateBat = nil
+			if err3 != nil {
 				return err3
 			} else if stop {
 				return nil
--- a/pkg/frontend/data_branch_hashdiff.go
+++ b/pkg/frontend/data_branch_hashdiff.go
@@ -51,6 +51,30 @@ type lcaProbeLayout struct {
 	enumValues  []string
 }
 
+func sortDataBranchBatchByPrimaryKey(bat *batch.Batch, pkColIdx int, mp *mpool.MPool) error {
+	pkVec := bat.Vecs[pkColIdx]
+	if !isDataBranchFloatType(*pkVec.GetType()) {
+		return mergeutil.SortColumnsByIndex(bat.Vecs, pkColIdx, mp)
+	}
+
+	identityVec := vector.NewVec(types.T_uint64.ToType())
+	defer identityVec.Free(mp)
+	for row := range pkVec.Length() {
+		identity, isNull, err := dataBranchFloatPKIdentityAt(pkVec, row)
+		if err != nil {
+			return err
+		}
+		if err = vector.AppendFixed(identityVec, identity, isNull, mp); err != nil {
+			return err
+		}
+	}
+
+	cols := make([]*vector.Vector, 0, len(bat.Vecs)+1)
+	cols = append(cols, bat.Vecs...)
+	cols = append(cols, identityVec)
+	return mergeutil.SortColumnsByIndex(cols, len(cols)-1, mp)
+}
+
 func (layout lcaProbeLayout) columnNameForTargetIndex(targetIdx int) (string, bool) {
 	for i, candidateIdx := range layout.targetIdxes {
 		if candidateIdx == targetIdx {
@@ -1062,8 +1086,8 @@ func hashDiffIfHasLCA(
 
 	handleBaseDeleteAndUpdates := func(wrapped batchWithKind) error {
 		wrapped.side = diffSideBase
-		if err2 := mergeutil.SortColumnsByIndex(
-			wrapped.batch.Vecs, tblStuff.def.pkColIdx, ses.proc.Mp(),
+		if err2 := sortDataBranchBatchByPrimaryKey(
+			wrapped.batch, tblStuff.def.pkColIdx, ses.proc.Mp(),
 		); err2 != nil {
 			return err2
 		}
@@ -1133,8 +1157,8 @@ func hashDiffIfHasLCA(
 			return nil
 		}
 
-		if err2 = mergeutil.SortColumnsByIndex(
-			wrapped.batch.Vecs, tblStuff.def.pkColIdx, ses.proc.Mp(),
+		if err2 = sortDataBranchBatchByPrimaryKey(
+			wrapped.batch, tblStuff.def.pkColIdx, ses.proc.Mp(),
 		); err2 != nil {
 			return err2
 		}
@@ -1159,7 +1183,7 @@ func hashDiffIfHasLCA(
 
 			i, j := 0, 0
 			for i < tarVec.Length() && j < baseVec.Length() {
-				if cmp, err3 = compareSingleValInVector(
+				if cmp, err3 = compareDataBranchPrimaryKeyInVectors(
 					ctx, ses, i, j, tarVec, baseVec,
 				); err3 != nil {
 					return
--- a/pkg/frontend/data_branch_hashdiff.go
+++ b/pkg/frontend/data_branch_hashdiff.go
@@ -289,8 +289,8 @@ func handleDelsOnLCA(
 
 				valsBuf.WriteString(fmt.Sprintf("row(%d,", i))
 				for j := range tuple {
-					if err = formatValIntoString(
-						ses, tuple[j], colTypes[expandedPKColIdxes[j]], valsBuf,
+					if err = formatValIntoStringWithFloatCast(
+						ses, tuple[j], colTypes[expandedPKColIdxes[j]], valsBuf, true,
 					); err != nil {
 						return nil, err
 					}
@@ -320,7 +320,7 @@ func handleDelsOnLCA(
 				valsBuf.WriteString(fmt.Sprintf("row(%d,", i))
 				b := tBat.Vecs[0].GetRawBytesAt(i)
 				val := types.DecodeValue(b, tBat.Vecs[0].GetType().Oid)
-				if err = formatValIntoString(ses, val, pkType, valsBuf); err != nil {
+				if err = formatValIntoStringWithFloatCast(ses, val, pkType, valsBuf, true); err != nil {
 					return nil, err
 				}
 				valsBuf.WriteString(")")
@@ -357,12 +357,14 @@ func handleDelsOnLCA(
 		)
 
 		for i := range quotedPKNames {
-			sqlBuf.WriteString(fmt.Sprintf("lca.%s = ", quotedPKNames[i]))
+			left := fmt.Sprintf("lca.%s", quotedPKNames[i])
+			right := fmt.Sprintf("pks.%s", quotedPKValueAliases[i])
 			if castType, ok := lcaProbeJoinCastType(colTypes[expandedPKColIdxes[i]]); ok {
-				sqlBuf.WriteString(fmt.Sprintf("cast(pks.%s as %s)", quotedPKValueAliases[i], castType))
-			} else {
-				sqlBuf.WriteString(fmt.Sprintf("pks.%s", quotedPKValueAliases[i]))
+				right = fmt.Sprintf("cast(%s as %s)", right, castType)
 			}
+			sqlBuf.WriteString(dataBranchSQLKeyEqual(
+				left, right, colTypes[expandedPKColIdxes[i]],
+			))
 			if i != len(quotedPKNames)-1 {
 				sqlBuf.WriteString(" AND ")
 			}
--- a/pkg/frontend/data_branch_hashdiff.go
+++ b/pkg/frontend/data_branch_hashdiff.go
@@ -1055,8 +1055,9 @@ func hashDiffIfHasLCA(
 		wg        sync.WaitGroup
 		atomicErr atomic.Value
 
-		baseDeleteBatches []batchWithKind
-		baseUpdateBatches []batchWithKind
+		baseDeleteBatches  []batchWithKind
+		baseUpdateBatches  []batchWithKind
+		restoreMissingKeys = make(map[string]struct{})
 	)
 
 	handleBaseDeleteAndUpdates := func(wrapped batchWithKind) error {
@@ -1078,6 +1079,50 @@ func hashDiffIfHasLCA(
 	handleTarDeleteAndUpdates := func(wrapped batchWithKind) (err2 error) {
 		wrapped.side = diffSideTarget
 		var pickConflictBat *batch.Batch
+		if wrapped.kind == diffInsert && wrapped.fromUpdate && len(restoreMissingKeys) > 0 {
+			var keep []int64
+			restoreBat := tblStuff.retPool.acquireRetBatch(tblStuff, false)
+			for rowIdx := range wrapped.batch.RowCount() {
+				key, keyErr := extractPKAsString(ses, tblStuff, wrapped.batch, rowIdx)
+				if keyErr != nil {
+					tblStuff.retPool.releaseRetBatch(restoreBat, false)
+					return keyErr
+				}
+				if _, restore := restoreMissingKeys[key]; !restore {
+					keep = append(keep, int64(rowIdx))
+					continue
+				}
+				if err2 = restoreBat.UnionOne(wrapped.batch, int64(rowIdx), ses.proc.Mp()); err2 != nil {
+					tblStuff.retPool.releaseRetBatch(restoreBat, false)
+					return err2
+				}
+				delete(restoreMissingKeys, key)
+			}
+			if restoreBat.Vecs[0].Length() > 0 {
+				restoreBat.SetRowCount(restoreBat.Vecs[0].Length())
+				if stop, e := emitBatch(emit, batchWithKind{
+					batch:          restoreBat,
+					kind:           diffInsert,
+					name:           wrapped.name,
+					side:           wrapped.side,
+					fromUpdate:     true,
+					restoreMissing: true,
+				}, false, tblStuff.retPool); e != nil {
+					tblStuff.retPool.releaseRetBatch(wrapped.batch, false)
+					return e
+				} else if stop {
+					tblStuff.retPool.releaseRetBatch(wrapped.batch, false)
+					return nil
+				}
+			} else {
+				tblStuff.retPool.releaseRetBatch(restoreBat, false)
+			}
+			if len(keep) == 0 {
+				tblStuff.retPool.releaseRetBatch(wrapped.batch, false)
+				return nil
+			}
+			wrapped.batch.Shrink(keep, true)
+		}
 		if len(baseUpdateBatches) == 0 && len(baseDeleteBatches) == 0 {
 			// no need to check conflict
 			if stop, e := emitBatch(emit, wrapped, false, tblStuff.retPool); e != nil {
@@ -1141,6 +1186,15 @@ func hashDiffIfHasLCA(
 						i++
 						j++
 					} else if copt.conflictOpt.Opt == tree.CONFLICT_ACCEPT {
+						if tarWrapped.kind == diffDelete && tarWrapped.fromUpdate &&
+							baseWrapped.kind == diffDelete && !baseWrapped.fromUpdate {
+							key, keyErr := extractPKAsString(ses, tblStuff, tarWrapped.batch, i)
+							if keyErr != nil {
+								err3 = keyErr
+								return
+							}
+							restoreMissingKeys[key] = struct{}{}
+						}
 						if tarWrapped.kind == diffDelete &&
 							baseWrapped.kind == diffDelete &&
 							!tarWrapped.fromUpdate && !baseWrapped.fromUpdate {
@@ -1256,11 +1310,6 @@ func hashDiffIfHasLCA(
 			return false
 		})
 
-		if wrapped.batch.RowCount() == 0 {
-			tblStuff.retPool.releaseRetBatch(wrapped.batch, false)
-			return
-		}
-
 		if pickConflictBat != nil {
 			if stop, e := emitBatch(emit, batchWithKind{
 				batch: pickConflictBat,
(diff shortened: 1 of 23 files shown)


gouhongshen:
Unrelated CI blockers in the current Ubuntu UT run are tracked in #26583 (checkpoint rollback leaves an unpublished object) and #26632 (shared hash-build budget recovery cancels its query). Evidence from this PR's failed job has been added to both issues.

XuPeng-SH (changes requested):
Blocking correctness gap on head 0d7238e150c7: portable SQL for a table without an explicit PK still cannot apply a delete/update whose old row contains NaN. In portable mode, buildDataBranchApplyLayout selects full-row deletion for fake-PK tables, and writeDeleteRowSQLFull emits `column = <literal>`. This patch turns the literal into `cast('NaN' as float/double)`, but MatrixOne float equality executes `a == b`, so NaN never matches itself. The generated script can therefore execute successfully while silently leaving the old row behind; an update can then also insert the new row. The added no-PK BVT exercises online MERGE through the hidden fake-PK path, not the portable full-row path. Please make full-row predicates NaN-aware and add a regression that generates and applies portable SQL for no-PK NaN deletion/update, covering FLOAT and DOUBLE while retaining NULL and infinity behavior. Focused formatter/output tests pass, but they do not cover this semantic round trip.

gouhongshen:
Fixed as a valid in-scope correctness bug in 3cb39c7fed. Portable full-row deletes now emit self-inequality for old FLOAT/DOUBLE NaN values, while NULL keeps IS NULL and finite/infinity values keep typed equality. The generator unit test covers both float widths plus NULL and infinities. The BVT now generates an actual no-PK OUTPUT FILE script and applies its exact delete/update predicate forms; 37/37 statements passed, and the generated file was inspected to confirm both NaN columns use self-inequality and only the expected keep/updated rows remain. The PR description is updated with the new root cause, change, and verification.

heni02 (approved):

XuPeng-SH (changes requested):
Deep-reviewed exact head 3cb39c7fedcc235a3c7383035773e5c11b3a3026. Requesting changes for one remaining same-shape correctness gap.

Blocking: real FLOAT/DOUBLE primary keys still depend on ordinary equality for NaN.

The new full-row predicate closes the fake-PK portable path, but buildDataBranchApplyLayout routes real primary keys through the staged-key path. The apply SQL uses DELETE ... WHERE pk IN (SELECT ...), while the LCA probe joins with lca.pk = pks.pk. MatrixOne permits FLOAT/DOUBLE primary keys, but scalar float equality follows IEEE semantics (`NaN = NaN` is false). I confirmed this with a live DOUBLE-PK case: the NaN key is accepted and scalar equality is false; the staged IN currently matches only because the hash-key path does not yet implement SQL NaN equality consistently. #26432 explicitly tracks resident/spill/re-spill typed-key semantic inconsistency, and its hashjoin contract tests still skip NaN. Data Branch correctness therefore depends on the join implementation and data size, and will regress when that underlying equality bug is fixed.

Please make both LCA lookup and real-key delete matching explicitly NaN-aware for every FLOAT/DOUBLE component, including composite keys, rather than relying on hash-key encoding. Add public-path regressions for online MERGE and portable SQL with real FLOAT and DOUBLE primary keys containing NaN, with finite/+Inf/-Inf controls; cover composite keys and the relevant join/spill boundary where practical.

The previously reported no-PK full-row gap is fixed. Fresh build, vet, focused regressions, and the full pkg/frontend test suite pass on this head.

gouhongshen:
XuPeng-SH's review is valid and in scope. Fixed on head 85d18d9c6b.\n\nThe LCA probe now gives every FLOAT/DOUBLE VALUES cell an explicit type (including finite cells), preventing mixed VALUES inference from converting NaN to its integer bit pattern. Both the LCA join and staged real-key delete use one shared per-component identity predicate: compare NaN classification, then compare values after normalizing only NaN to zero. Composite non-float components retain ordinary equality, and non-float-only staged deletes retain the existing IN path. This removes raw NaN from the equality/hash input, so correctness no longer depends on #26432's resident/spill hash behavior.\n\nRegression coverage now exercises online MERGE and portable SQL for a FLOAT PK and a composite FLOAT/DOUBLE/int PK, with NaN in each float component plus finite and +/-Inf controls. The generated result was inspected after generation; 66/66 BVT statements pass. Focused tests, the full pkg/frontend suite, vet, and make build also pass. The PR description has been updated with the added root cause, changes, tests, and residual risk.

gouhongshen:
Rebased the reviewed fix onto current main (820ee12144); the equivalent head is now 374faa73b0. The only conflict was a test fixture touched by #26588, resolved by retaining both destination-column metadata and FLOAT/DOUBLE key types. On the rebased head, the full pkg/frontend suite (19.782s), vet, diff checks, make build, and a fresh 66/66 public BVT run pass. The PR description is updated to the rebased head.

XuPeng-SH (changes requested):
Deep-reviewed exact head 374faa73b0884c7948103a8b2ac739694f163d26. Requesting changes for a remaining storage-identity correctness gap that causes silent data loss.

Blocking: `dataBranchSQLKeyEqual` collapses distinct legal FLOAT/DOUBLE primary keys.

MatrixOne primary-key identity is bit-preserving, while the new predicate is based on scalar float equality plus one NaN bucket. Through public SQL, a DOUBLE primary key table accepts and commits all four distinct keys below when inserted in separate statements:

- `bit_cast(unhex('000000000000f87f') as double)` — NaN payload 0
- `bit_cast(unhex('010000000000f87f') as double)` — NaN payload 1
- positive zero
- `bit_cast(unhex('0000000000000080') as double)` — negative zero

`serial(k)` confirms four different packed keys. However, the generated predicate classifies both NaN payloads as identical, and ordinary equality classifies positive and negative zero as identical. An exact LCA-shaped probe for only payload 1 and negative zero therefore returns both NaN rows and both zero rows. The staged-delete predicate with only payload 1 deletes both legal NaN-key rows.

I then reproduced the full public Data Branch path on this head. The base and destination contained those four rows. The source updated only NaN payload 1 and negative zero, selected by exact `serial(k)`. `DATA BRANCH MERGE` returned success, but the destination silently dropped the untouched NaN payload 0 and positive-zero rows; only two rows remained. The updated negative-zero key was also materialized as positive zero. This is not only a matcher issue: `formatValIntoStringWithFloatCast` emits `cast(-0 as double)`, which MatrixOne evaluates as positive zero, and textual NaN casts canonicalize payload bits.

Please define Data Branch key identity from the exact encoded/binary key representation and preserve FLOAT/DOUBLE bit identity through generated LCA values, staged deletes, and replacement inserts. Add public regressions using `bit_cast` plus `serial` for distinct NaN payloads and signed zero, for FLOAT and DOUBLE and a composite-key control. Canonical NaN, infinity, and ordinary finite controls should remain.

Fresh focused tests, full `pkg/frontend`, vet, and the exact-head build pass; the existing tests do not exercise multiple SQL-visible bit-distinct float keys.

gouhongshen:
XuPeng-SH’s storage-identity review is valid and in scope. Fixed in 8e852a0008.

Data Branch now compares FLOAT/DOUBLE key components through exact `serial(...)` encodings and materializes NaN payloads/negative zero with `bit_cast(unhex(...))`. Expanded update batches retain their update identity: same-key changes use exact-key `UPDATE` statements so key bits never pass through a delete/reinsert cycle, while genuine new float-key rows use separate direct `INSERT ... VALUES` statements. Non-float key paths remain unchanged.

Public regressions cover FLOAT, DOUBLE, and composite FLOAT/DOUBLE/int primary keys with two NaN payloads, signed zero, finite, and infinity controls. Online MERGE and portable SQL retain the untouched payload/+0 rows and update only payload 1/-0; `hex(serial(...))` verifies every exact key. The inspected generated script uses the exact predicates/literals. Focused tests, the full `pkg/frontend` suite (17.984s), vet, `make build`, `git diff --check`, and the 110/110 BVT pass. The PR description is updated.

XuPeng-SH (changes requested):
Deep-reviewed updated exact head `8e852a000875b8811b6b6bfee06d837ae2d6e350`. The exact storage-identity fix closes MERGE and portable-output paths, but one original-issue path still bypasses it.

### P1 — FLOAT/DOUBLE primary-key updates still fail through DATA BRANCH PICK

`mergeDiffs` and portable output call `appendBatchRowsAsSQLValues`, which recognizes `fromUpdate` and emits the new exact-key `UPDATE`. `pickMergeDiffs`, however, calls `appendPickedBatchRows`; that function ignores `fromUpdate` and still queues the pair as staged DELETE plus replacement INSERT. This is precisely the delete/reinsert cycle that the PR correctly identifies as unsafe for FLOAT/DOUBLE keys.

I reproduced it through public SQL on a service built from this exact head. The base/source/destination contain four legal, bit-distinct DOUBLE primary keys: two NaN payloads and both signed zeros. The source changes only NaN payload 1 and negative zero, then:

```
DATA BRANCH PICK src INTO dst
KEYS(SELECT k FROM src WHERE note IN ('nan1_updated', 'negzero_updated'))
WHEN CONFLICT ACCEPT;
```

returns:

```
ERROR 1062: Duplicate entry '0' for key 'k'
```

The destination transaction rolls back and retains all four original rows. Ordinary MERGE on the same representations is covered by the PR, but PICK is explicitly part of #26072 and remains broken for a real FLOAT/DOUBLE key update.

Please route PICK same-key updates through exact-key `UPDATE` while preserving its FAIL/SKIP/ACCEPT conflict semantics. Add public PICK regressions with bit-distinct NaN payloads and signed zero, plus a composite float-key control; verify both updated and untouched encodings with `serial(...)`.

gouhongshen:
XuPeng-SH's PICK review is valid and in scope. Fixed on rebased head `74cef706ec`.

PICK now uses the same shared exact FLOAT/DOUBLE update classification and row application as MERGE/portable output, but only after its existing FAIL/SKIP/ACCEPT decision. Marked delete rows still drive conflict handling and are then suppressed; accepted marked inserts execute an exact-key `UPDATE` through `serial(...)`, so the key never enters the unsafe delete/reinsert cycle. Non-float paths are unchanged.

Public regressions cover FLOAT, DOUBLE, and composite FLOAT/DOUBLE/int PICK updates with two NaN payloads and both signed zeros. The untouched payload-0/+0 rows remain, only payload-1/-0 rows change, and `hex(serial(...))` verifies every stored key. On the rebased head: full `pkg/frontend` 21.159s, vet, `make build`, diff checks, and the inspected 119/119 BVT pass. The PR description is updated.

XuPeng-SH (changes requested):
Deep-reviewed exact head `74cef706ec05471643afb24102edc3c034648e48`. The previous PICK case with an unchanged destination is fixed, but the UPDATE-only application path introduces a silent-loss case when the destination no longer contains the source key.

### P1 — ACCEPT loses a FLOAT/DOUBLE source update when the destination deleted the row

`dataBranchExactFloatKeyUpdateBatch` classifies both halves of every same-key update on a FLOAT/DOUBLE-containing PK as an exact update. The consumer then suppresses the DELETE half, and `appendOrExecuteDataBranchApplyRow` turns the INSERT half into only:

```sql
UPDATE dst SET ... WHERE serial(pk) = serial(source_pk) LIMIT 1
```

If the destination branch independently deleted that row, the UPDATE affects zero rows. There is no INSERT fallback, so `WHEN CONFLICT ACCEPT` reports success but does not materialize the accepted source row.

I reproduced this through public SQL on a service built from this exact head:

```sql
CREATE TABLE base_t(k DOUBLE PRIMARY KEY, note VARCHAR(32));
INSERT INTO base_t VALUES (1.5, 'base');
DATA BRANCH CREATE TABLE src FROM base_t;
DATA BRANCH CREATE TABLE dst FROM base_t;
UPDATE src SET note = 'source_updated' WHERE k = 1.5;
DELETE FROM dst WHERE k = 1.5;

DATA BRANCH PICK src INTO dst
  KEYS(SELECT k FROM src WHERE note = 'source_updated')
  WHEN CONFLICT ACCEPT;
SELECT COUNT(*) FROM dst; -- 0, expected 1
```

The FAIL control correctly reports this pair as a conflict, proving ACCEPT is taking the wrong resolution rather than seeing a no-op. The same witness with an INT PK restores the source row as expected. I also reproduced the zero-row result for:

- FLOAT through PICK;
- DOUBLE through MERGE;
- a composite `(FLOAT, INT)` PK through MERGE.

The portable path shares the same UPDATE-only helper, so its generated script has the same missing-row state unless it is handled separately.

Please preserve exact float-key identity while closing the complete conflict state: after ACCEPT, the source row must exist whether the destination retained, modified, or deleted the ancestor row. Add public PICK and MERGE regressions for destination-delete versus source-update across FLOAT, DOUBLE, and a composite float key, with FAIL/SKIP controls; also round-trip the portable script and retain the existing NaN-payload/signed-zero identity assertions.

Fresh exact-head evidence: `go build` and `go vet` for `pkg/frontend` pass; the full `pkg/frontend` suite passes; the focused float-key regressions pass under `-race -count=3`. Those tests currently cover only the destination-retained branch of this state machine.

gouhongshen:
XuPeng-SH's destination-delete review is valid and in scope. Fixed on rebased head `ae748dc651`.\n\nExact FLOAT/DOUBLE-key materialization is now policy-aware at the shared apply boundary: retained rows always use exact `serial(...)` UPDATE; source-winning ACCEPT and portable output additionally issue an exact-key conditional INSERT when the destination independently deleted the row; SKIP deliberately performs no insertion and preserves that deletion. PICK keeps its FAIL/SKIP/ACCEPT decision ahead of the shared apply path.\n\nPublic regressions cover FLOAT PICK FAIL/SKIP/ACCEPT, DOUBLE MERGE FAIL/SKIP/ACCEPT, composite FLOAT/int MERGE ACCEPT, and a portable DOUBLE round trip. The prior NaN-payload/signed-zero assertions remain. On the rebased head, full `pkg/frontend` (21.141s), vet, `make build`, diff checks, and the inspected 171/171 BVT pass. The PR description is updated.

XuPeng-SH (changes requested):
Blocking correctness issue in pkg/frontend/data_branch_output.go:2055: the new restore path uses INSERT ... SELECT, even though lines 232-236 document that INSERT ... SELECT collapses bit-distinct FLOAT/DOUBLE keys. Reproduced on exact head ae748dc6515f278741e50c734a76696d98c5c616: independently insert +0 and -0 into a DOUBLE PK table; update the exact -0 row on the source branch; delete exact -0 on the destination while +0 remains; then MERGE ... WHEN CONFLICT ACCEPT fails with Error 1062: Duplicate entry -0. PICK ... WHEN CONFLICT ACCEPT fails identically. The serial-based NOT EXISTS sees -0 as missing, but the INSERT ... SELECT path then compares with scalar float semantics and collides with +0. Portable output emits the same statement. Please restore missing exact-key updates through a path that preserves independent single-row INSERT semantics, and add combined missing-key/bit-distinct-peer regressions (+0/-0 and distinct NaN payloads) for MERGE, PICK, and portable output.

gouhongshen:
XuPeng-SH's review is valid and in scope. Fixed on rebased head `ad38897235`.

The review correctly identified that `INSERT ... SELECT` cannot preserve an independently missing FLOAT/DOUBLE key beside a bit-distinct peer. Self-checking found a deeper constraint: replacing it with `INSERT ... VALUES` alone still failed because the primary-key DEDUP plan joined raw floats, so +0/-0 and distinct NaN payloads were still collapsed.

The clean fix closes both boundaries:

- The shared diff state machine now marks only source-update/destination-delete conflicts as missing. Retained rows use exact-`serial(...)` UPDATE; source-winning ACCEPT and portable output restore the proven-missing row with one direct `INSERT ... VALUES`; SKIP preserves the deletion. The apply layer no longer guesses existence.
- Simple FLOAT/DOUBLE primary-key duplicate checks now compare `serial(...)` encodings in DEDUP joins, constraint probes, and intra-batch grouping. Non-float keys are unchanged.

Public regressions combine a missing key with its untouched bit-distinct peer for both +0/-0 and two NaN payloads across MERGE, PICK, and portable SQL. All four exact `hex(serial(k))` identities remain and only the selected source row changes.

Verification on base `acfb8ef77c`: full `pkg/frontend` (20.226s), full `pkg/sql/plan` (10.393s), focused race test (3 runs), vet, `make build`, diff checks, and the inspected 212/212 BVT pass. The PR description is updated.

gouhongshen:
The proxy BVT failure on ad38897235 was valid and PR-caused. The exact FLOAT/DOUBLE identity expression was correctly serialized for DEDUP, but duplicate diagnostics printed the binary key, producing the ALTER TABLE expectation mismatch; Coverage only propagated that producer failure. Fixed on rebased head 0570eb78b8 by decoding exact keys for errors and closing the same projected-type/serial(pk) contract in fuzzy-filter collision recovery. Rebased onto current main bb011f2376 with no conflicts. Controlled tests/vet, focused race x3, make build, diff checks, DDL BVT 923/923 + 764/764, and Data Branch BVT 212/212 pass. The PR description is rewritten and current.

gouhongshen:
The Coverage failure on `0570eb78b8` is valid and PR-caused. All producer jobs passed, but changed-code coverage was 178/240 (74.17%), below the strict 75% gate. Fixed on rebased head `e91247f83a` with focused tests for malformed serialized-key decoding, compound/hidden fuzzy-check SQL modes, and projected identity-type fallback; runtime behavior is unchanged. On the final base, full frontend/plan/colexec/compile tests, vet, `make build`, coverage-mode package tests, and diff checks pass. The PR description is updated.

gouhongshen:
The Coverage failure on `e91247f83a` is valid and PR-caused. All producer jobs passed; CI's merged profile reported 175/240 (72.92%). The earlier tests brought `dedup_key.go` and `operator.go` to 100%, but the real background fuzzy-check executor/result path remained uncovered. Fixed in `105457f602` with a boundary test that verifies decoded `-0` duplicate diagnostics and malformed serialized identity errors, including result cleanup and runtime executor restoration. A matching coverage-mode run adds 14 covered statements (enough to take 175/240 above the strict 75% gate); the focused test, full `pkg/sql/compile`, vet, and diff checks pass. The PR description is updated.

XuPeng-SH (changes requested):
The previous `INSERT ... SELECT` blocker is fixed, but the hash-diff conflict join still collapses bit-distinct FLOAT/DOUBLE primary keys.

`handleTarDeleteAndUpdates` sorts each side by the raw PK vector (`mergeutil.SortColumnsByIndex`) and then treats `compareSingleValInVector(...) == 0` as the same key. Both paths use scalar float ordering/comparison: `+0` and `-0` compare equal, and distinct NaN payloads compare equal. A minimal white-box witness on this head expects non-zero for `compareSingleValInVector(+0, -0)` but gets `0`.

This leaves an uncovered cross-change shape. For example, start with both `+0` and `-0` as DOUBLE PKs; update `-0` on the source while independently deleting `+0` on the destination, then MERGE/PICK with conflict handling. These are different storage keys, but the merge join pairs them as a conflict. Under ACCEPT, lines 1188-1196 can mark the source `-0` update as `restoreMissing`; the new direct INSERT then collides with the still-existing `-0` row. FAIL/SKIP can likewise reject or skip a non-conflict. The same shape exists for two NaN payloads.

Please make hash-diff sorting and equality share one exact, bit-preserving key order (including composite keys), and add crossed-key regressions for `+0/-0` and distinct NaN payloads across MERGE/PICK and FAIL/SKIP/ACCEPT. The current tests delete the same exact key that the source updates while merely retaining a bit-distinct peer, so they do not exercise this pairing bug.

gouhongshen:
XuPeng-SH’s hash-diff review is valid and in scope. Fixed in `12db1fb854`.

The conflict merge join now uses one shared exact key identity for both ordering and equality. Simple FLOAT/DOUBLE PKs are ordered by their raw IEEE bits, so signed zero and distinct NaN payloads remain separate storage keys; non-float keys retain the existing comparator, and composite PKs retain their existing serialized-byte order. Both sides are sorted through the same helper before the merge join, eliminating the former scalar-order mismatch.

Regression coverage now crosses independent changes: source updates -0 while destination deletes +0, and source updates NaN payload 1 while destination deletes payload 0. MERGE and PICK cover FAIL/SKIP/ACCEPT, with a composite control and exact `hex(serial(...))` assertions. Focused tests, full `pkg/frontend`, controlled CGo vet, `make build`, diff checks, and the public 256/256 BVT pass. The PR description is updated.

XuPeng-SH (approved):
Deep-reviewed the updated exact head. The previous hash-diff blocker is closed.

The conflict matcher now derives a raw IEEE-bit identity for a simple FLOAT/DOUBLE PK, sorts both sides by that `uint64` identity, and uses the identical comparison in the merge join. This distinguishes `+0` from `-0` and distinct NaN payloads while leaving non-float keys unchanged; composite keys continue to use their existing exact serialized-byte identity. The temporary identity vector is batch-scoped and freed after sorting, so the extra O(n), 8-byte/row allocation is limited to simple float-key conflict batches.

The new public regressions cover the previously missing crossed-change shape: source updates `-0`/NaN payload 1 while destination deletes `+0`/payload 0, across MERGE and PICK with FAIL/SKIP/ACCEPT, plus a composite control and exact `serial(...)` assertions. This closes the false-conflict and erroneous restore/duplicate path without weakening real conflict handling.

Fresh exact-head validation passed: diff check; `go list`, `go build`, `go vet`, and the full `pkg/frontend` suite; focused exact-identity/sort tests under `-race -count=3`. CI is green.

XuPeng-SH (approved):
Revalidated the Mergify queue merge head `6f72fe538464`. Its first parent is the reviewed/fixed head `12db1fb854`; its second parent is current main `2190897330`. The Data Branch, colexec, compile, plan, and float-special-value test patch is byte-identical to the approved parent. The only upstream delta from the first parent is `pkg/logservice/service_bootstrap_test.go`, and the merge result is exactly the current-main version—there is no conflict rewrite in this PR surface.

The exact float identity/sort tests pass on this merge head, and `git diff --check` is clean. Approval stands.

PR #26505, opened 2026-07-31
fix(frontend): surface data branch file finalization errors

### LINKED ISSUE #26054: [Bug]: DATA BRANCH DIFF OUTPUT FILE can report success after a late write failure
## Description

`DATA BRANCH DIFF ... OUTPUT FILE` can return a successful result even when the target file service reports an error after it has consumed all streamed data.

This affects both file-output implementations:

- the non-`MutableFileService` path implemented by `newSingleWriteAppender`;
- the `MutableFileService` path, whose `release` function discards `mut.Close()` errors.

## Code path

On current `main` (`2eb7f523e5a10f352d3e258183ed8440510388de`):

1. `newSingleWriteAppender` starts `targetFS.Write` in a worker and stores its final error in `done`.
2. If the file service reads the complete pipe and only then returns an error, all producer-side `writeFile` calls have already succeeded.
3. `release` receives the final error but only invokes `onError`; its signature has no error result, so the caller cannot observe the failure.
4. `satisfyDiffOutputOpt` sets `succeeded = true` and adds the success result row before the deferred `release` runs.

The mutable path has the same ordering problem: every `mut.Append` may succeed, while the deferred `release` executes `_ = mut.Close()`. `Mutator.Close()` performs the final `Sync` and close operations and can return an error, but that result is discarded after the statement has already been marked successful.

Relevant locations:

- `pkg/frontend/data_branch_output.go:899-905`
- `pkg/frontend/data_branch_output.go:990-992`
- `pkg/frontend/data_branch_output.go:1679-1689`
- `pkg/frontend/data_branch_output.go:1701-1753`

## Reproduction

A package-level test used a `FileService` wrapper whose `Write` method:

1. fully reads every `ReaderForWrite` entry;
2. returns a deterministic final write error afterward.

Observed in 3/3 runs:

- `writeFile("BEGIN;\n")` returned `nil`;
- `writeFile("COMMIT;\n")` returned `nil`;
- `release()` observed the final error only through the cleanup callback;
- no error value was available to propagate to the SQL statement.

Test command:

```text
go test ./pkg/frontend -run '^TestAuditDataBranchOutputLateWriteFailureIsNotReturned$' -count=3 -v
```

## Expected behavior

The final file-service write result should be awaited before the statement is marked successful. A final write error should be returned to the SQL caller, and the partial output should be cleaned up.

## Actual behavior

The final write/close error is discarded by `release()`. The non-mutable path runs its cleanup callback, while the mutable path does not receive the close error at all; in either case the SQL statement can still report a successful file path/result row.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26054

### Root cause

`DATA BRANCH DIFF ... OUTPUT FILE` treated successful producer-side appends as a completed file write. In the streaming path, `FileService.Write` ran in a worker and its final result was consumed only by a no-result `release` callback. A file service could therefore consume the entire pipe, let every `writeFile` call succeed, and then return an error that the SQL statement could not observe. The mutable path similarly discarded the final `Mutator.Close` error. In both cases the success row was added before deferred finalization ran.

### Changes

- Make the common file-output finalizer return an error.
- Await and return the worker's final `FileService.Write` result in the non-mutable path.
- Return the final `Mutator.Close` result in the mutable path.
- Explicitly finalize the output before adding the success result row, and retain deferred finalization for early-error paths.
- Join an early statement error with any concurrent finalization error instead of discarding either one.
- Finalize the writer before deleting partial output so cleanup cannot race an active stream.
- Add focused tests for a file service that consumes the complete stream before failing, SQL-level error/result-row/cleanup behavior, mutable close failure propagation, success, immediate write failure, and setup failures.

No public API, SQL syntax, output format, or configuration behavior changes.

### Tests

Pre-fix reproduction on `main` used a package-level file service wrapper that fully consumed each `ReaderForWrite` before returning `mock late write failure`. In 3/3 runs, both `BEGIN` and `COMMIT` writes returned nil, and the no-result release callback was the only observer of the failure.

Post-fix exhaustive verification on `main@84f0d1116`:

- CGo-configured `go build -mod=readonly ./pkg/frontend`
- CGo-configured `go vet -mod=readonly ./pkg/frontend`
- Focused output-file regression set, `-count=3` (9 exact tests)
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=20m ./pkg/frontend` (PASS, 24.471s)
- Race selection was enumerated and contained all 7 new/modified behavioral tests.
- Race measurements were `T=0s` to `0.04s`; with `B=30s`, the adaptive rule selected `N=100` for every test.
- Each selected exact test passed independently with `-race -count=100`.
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=20m ./pkg/frontend` (PASS, 37.818s)

Rebase verification on `main@0b8b53dff`:

- Focused output-file finalization set, `-count=3` (8 exact tests)
- `.agents/skills/mo-dev/scripts/mo-cgo-test ./pkg/frontend -count=1` (PASS, 33.817s)
- CGo-configured `go build -mod=readonly ./pkg/frontend`
- CGo-configured `go vet -mod=readonly ./pkg/frontend`
- `git range-diff` confirmed the rebase preserved the patch exactly.

BVT is not appropriate for this regression because the required boundary is a deterministic post-consumption `FileService.Write` or `Mutator.Close` fault, which the SQL BVT harness cannot inject. The package-level tests cover both file-service implementations and the SQL result-row boundary directly.

### CI follow-up

Two attempt-1 Ubuntu/x86 UT failures were proven unrelated to this PR; in both jobs, `pkg/frontend` passed and the failures were in packages outside this two-file diff. The first failure exposed ISCP, embedded-cluster heartbeat, and lockservice signatures: the previously unreported ISCP cases were filed as #26517, while existing #26374 and #26441 / #26448 cover the other signatures. After rebasing, a second failure exposed current-main embedded-cluster heartbeat/reset cases and was filed as #26522. Per CI policy, the failed jobs for the current-head workflow were rerun exactly once; attempt 2 passed.

### Residual risks

No known correctness gap remains in final write/close error propagation. Partial-output deletion retains the existing best-effort cleanup contract: this change guarantees cleanup is invoked after writer termination, but a backend `Delete` failure remains secondary to the original output failure and is not returned separately.


gouhongshen:
Investigated the failed Ubuntu/x86 UT job from run 30617011418. The PR changes only `pkg/frontend/data_branch_output.go` and its tests, and `pkg/frontend` passed in the job. Failures were in `pkg/vm/engine/test`, `pkg/embed`, and `pkg/lockservice`: ISCP timing assertions are tracked in #26517; the lockservice rolling-restart signature is covered by #26374; the embedded heartbeat starvation signature is covered by #26441 and its merged fix #26448. No unrelated change was added to this PR. The branch has been rebased onto current `main`, and focused tests, full frontend tests, CGo-configured build/vet, and patch-equivalence checks pass.

gouhongshen:
Investigated the second Ubuntu/x86 UT failure (job 91130900423). It is again outside this PR: the failed packages were `pkg/vm/engine/test`, `pkg/embed`, `pkg/tests/ddl`, and `pkg/tests/features`; `pkg/frontend` passed. The ISCP watermark/teardown-panic recurrence was added to #26517. The current-main recurrence of HAKeeper/LogService heartbeat resets across PITR, multi-cluster, Iceberg credential, and table-features tests is tracked in #26522. No PR code was changed.

XuPeng-SH (approved):
Approved exact head `d7a02bfaa1a976c996766a6862a7d236e31826a6`.

No blocking findings. I reviewed the complete finalization closure for both mutable and streaming file-service paths: final write/close errors reach the SQL caller before the success row, early errors are joined with finalizer errors, finalization remains single-owner, and partial-output cleanup runs only after the active writer has terminated.

Local verification passed: all 9 new/modified exact tests were enumerated; each measured `T=0s` with `B=30s` and passed independently under `-race -count=100`; full `pkg/frontend` passed under race and non-race; build, vet, and diff checks passed. Current CI is green and GitHub reports the PR mergeable.

Issue #26487, opened 2026-07-31
[Bug]: logservice shutdown races detached HAKeeper SyncRead with Dragonboat destruction


XuPeng-SH:
## Problem

`pkg/embed.TestBaseClusterCanWorkWithNewCluster` can race during logservice shutdown: Dragonboat destroys a state machine while a detached HAKeeper task-scheduling goroutine is still executing `SyncRead`.

Observed in [run 30603896298 / job 91072145422](https://github.com/matrixorigin/matrixone/actions/runs/30603896298/job/91072145422?pr=26473) at exact head `8410ce20f36f321a3ea8da32770d459feba34503`:

```text
WARNING: DATA RACE
Write:
  dragonboat/internal/rsm.(*OffloadedStatus).SetDestroyed
  dragonboat/internal/rsm.(*NativeSM).Close
  dragonboat.(*node).destroy

Previous read:
  dragonboat/internal/rsm.(*NativeSM).Lookup
  dragonboat.(*NodeHost).SyncRead
  logservice.(*store).read                    store.go:705
  logservice.(*store).getCheckerState        store_hakeeper_check.go:483
  logservice.(*store).tickerForTaskSchedule  store.go:1174
```

The test fails with `race detected during execution of test`.

## Relationship to PR #26473

This is not a regression introduced by #26473:

- #26473 does not modify `pkg/embed`, `pkg/logservice`, or the Dragonboat dependency.
- The race stack contains no lockservice or MORPC frame.
- The same CI run passed both `pkg/lockservice` and `pkg/logservice`; the race is exposed by embed cluster teardown.
- On the exact PR head, the test passed `-race -count=3` in 67.51s.
- On pre-PR main `cf6dca31107769ad7855ebb45b1a6b48298284d6`, it passed `-race -count=3` in 67.37s.
- Existing #20604 has the same test name but a different root cause (port already in use), so it is not a duplicate.

## Root cause

`store.ticker` starts task scheduling with an untracked goroutine:

```go
go l.tickerForTaskSchedule(ctx, l.cfg.HAKeeperCheckInterval.Duration)
```

`store.close` calls:

```go
l.tickerStopper.Stop()
l.stopper.Stop()
l.nh.Close()
```

[...]

PR #26507, opened 2026-07-31
fix: join HAKeeper task scheduler before log store close

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26487

`store.ticker` launched the HAKeeper task scheduler as a raw goroutine. During logservice shutdown, `tickerStopper.Stop()` joined the parent HAKeeper ticker but not that detached child, so `NodeHost.Close()` could destroy Dragonboat state while the child was still executing `SyncRead`.

This change registers the task scheduler with the existing `tickerStopper`. Scheduling remains independent from the HAKeeper tick and health-check loop, while shutdown now cancels and joins both loops before destroying the NodeHost.

The regression test blocks the scheduler in checker-state lookup, starts `store.close`, proves the ticker stopper has entered its terminal admission state, and verifies close cannot complete until the scheduler exits.

No SQL behavior, wire format, persisted data, configuration, or scheduling policy is changed.

Tested with:

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=100 -timeout=10m -run '^TestStoreCloseWaitsForTaskScheduleTicker$' ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=52 -timeout=10m -run '^TestTickerForTaskSchedule$' ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=30m ./pkg/logservice`

`.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=10 -timeout=30m -run '^TestBaseClusterCanWorkWithNewCluster$' ./pkg/embed`

`go vet ./pkg/logservice`

`go build ./pkg/logservice`


XuPeng-SH (approved):
Deep-reviewed current head 7d3e68dd60190be15a2fc2ea59e3c51dfb29e842. Stopper admission and Stop linearization now own and join the task-scheduler ticker before NodeHost.Close, closing the shutdown race without coupling it to the HAKeeper tick loop. The close regression passed 100 race runs, TestTickerForTaskSchedule passed 44 race runs, and the full logservice race suite, build, and vet passed. No blocking issues found.

Issue #25959, opened 2026-07-21
[Bug]: COALESCE inside an empty correlated aggregate subquery returns NULL


Ariznawlll:
### MatrixOne version

Latest `main` at `3046f14d94b3f1a25839eeeb65729b64f85ac647`.

### What happened?

When a correlated scalar aggregate subquery has no matching input rows, `COALESCE` or `IFNULL` inside the subquery is not evaluated. MatrixOne returns `NULL` instead of the requested fallback value.

The same aggregate works when the subquery is uncorrelated, and moving `COALESCE` outside the scalar subquery also works. This indicates that the correlated-subquery rewrite drops the aggregate result row for an empty input.

### How to reproduce

```sql
CREATE TABLE parent(id INT PRIMARY KEY);
CREATE TABLE child(parent_id INT, v INT);
INSERT INTO parent VALUES (1), (2);
INSERT INTO child VALUES (1, 10);

SELECT p.id,
       (SELECT COALESCE(SUM(c.v), 0)
        FROM child c
        WHERE c.parent_id = p.id) AS total
FROM parent p
ORDER BY p.id;
```

### Actual result

```text
1  10
2  NULL
```

### Expected result

```text
1  10
2  0
```

MySQL 8.0.45 returns the expected result.

### Scope

- The issue is reproducible with `SUM`, `AVG`, `MIN`, and `MAX` wrapped by `COALESCE`.
- `COUNT(*)` returns `0` correctly.
- Wrapping the correlated source in a CTE has the same incorrect result:

```sql
SELECT p.id,
       (WITH x AS (SELECT v FROM child c WHERE c.parent_id = p.id)
        SELECT COALESCE(SUM(v), 0) FROM x)
FROM parent p;
```

- Moving the fallback outside the subquery is a working control:

```sql
SELECT p.id,
       COALESCE((SELECT SUM(c.v) FROM child c WHERE c.parent_id = p.id), 0)
FROM parent p;
```

VioletQwQ-0:
QA handoff

PR #26510 merged with final head `55203b022b21cc79c73b64e2b65c032e82c41725` as merge commit `d6800b98855b871459c3c4bad0ce73aa7321610f`. The merge is present in current main `d34a6c896f5f0c07f378861704520db548f404b8`.

Automated evidence:
- canonical `scalar_correlated_projection` BVT passed 40/40 on the PR implementation head;
[...]

PR #26510, opened 2026-07-31
fix(plan): preserve empty correlated aggregate projections

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #25959

Correlated scalar aggregates are decorrelated by grouping the inner rows on the pulled-up correlation key and joining the result back with a `LEFT JOIN`. When an outer key has no inner group, an expression evaluated below that join never runs, so examples such as `COALESCE(SUM(v), 0)`, `COUNT(*) + 1`, `CASE`, and the JSON empty-array fallback produce the wrong value.

For the supported direct `AGG` and `PROJECT -> AGG` shapes, this change:

- preserves and validates the scalar final expression before mutating the plan;
- exposes raw aggregate outputs plus existing pulled-up correlation keys on the join right side;
- defines the canonical empty-input contract for every current aggregate ID in `aggexec` and restores typed `NULL`, zero, or all-bits-set results after null extension;
- evaluates `COALESCE`, `IFNULL`, `CASE`, arithmetic, or a multi-aggregate final expression above the join;
- conservatively rejects aggregates with non-trivial unsupported empty state, explicit grouping, `HAVING`, row-order/row-limit wrappers, deep correlation, and non-aggregate inner-column references without partial plan mutation.

The compatibility oracle for this PR is MySQL 8.0.45, as recorded in #25959. Prisma `relationLoadStrategy: join` compatibility remains follow-up #24737 scope.

This PR does not modify executor execution behavior, the parser, protobuf, public API, persisted formats, `LATERAL`, `Node_APPLY`, `mo-auto-test`, or Node/npm dependencies.

## Validation

- Exact base: `d29b4758f88f0894feeca15f31412446c17cb9d3`
- Exact head: `eb983b2a748c8069845b4924680d89df591f3de8`
- Canonical BVT: `test/distributed/cases/subquery/scalar_correlated_projection.sql`
  - historical base: exit `1`, `31/39`, 8 target semantic failures
  - current head: exit `0`, `40/40`
  - includes the original #25959 CTE shape and a metamorphic missing-key comparison for `BIT_AND`, `BIT_OR`, `BIT_XOR`, `APPROX_COUNT_DISTINCT`, and `SUM`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/sql/colexec/aggexec ./pkg/sql/plan`: PASS
- Focused contract, typed-boundary, eligible/rejected-shape, output-mapping, and residual-`Expr_Corr` tests: PASS
- `GOWORK=off go build -mod=readonly ./pkg/sql/plan/... ./pkg/sql/colexec/aggexec/...`: PASS
- `GOWORK=off go vet ./pkg/sql/plan/... ./pkg/sql/colexec/aggexec/...`: PASS
- `make build`: PASS
- `git diff --check`: PASS
- `mo-self-review`: PASS, no unresolved findings
- `mo-pr-preflight-review`: `PASS review=PASS validation=PENDING`, diff hash `20cdcad04d6ecfa637ca9c41ab10f0b6296dfc2250e423c2251685b3f4f57ca9`
- New exact-head GitHub CI: PENDING

## BVT

BVT: required

The canonical `subquery/scalar_correlated_projection` case covers empty/present/all-NULL inner inputs, repeated parent keys, SUM/AVG/MIN/MAX, raw SUM, COUNT variants, BIT_AND/OR/XOR, APPROX_COUNT_DISTINCT, arithmetic, multiple aggregates, CASE, JSON fallback, a forwarding CTE, the issue-authentic CTE-internal correlation, and explicit GROUP BY/HAVING controls.

## QA

QA required: yes

After merge, verify the original #25959 query and the JSON empty-array fallback on an identified MatrixOne build/environment. Keep #25959 open until the tested version/environment and terminal PASS are recorded.

This is the prerequisite for #24737. Real Prisma 5.22/current `relationLoadStrategy: join` QA remains PENDING and belongs to the follow-up PR/issue lifecycle; it is not claimed by this PR.


XuPeng-SH (changes requested):
Deep-reviewed exact head `fcefaa8a6a5aff7c1d8b0150b7aeefa4904c92da`. Requesting changes for one remaining same-shape correctness gap.

### P1 — Empty-input reconstruction is incomplete and aggregate-name-specific

`prepareCorrelatedScalarAggregatePostJoinProjection` hard-codes only `sum/avg/min/max/json_arrayagg/count/starcount`. For every other aggregate it falls back to the legacy grouped-correlated plan. A missing correlation key then has no right-hand group, so the LEFT JOIN produces NULL even when the aggregate contract requires a non-NULL empty value.

I reproduced this through SQL on a server built from this exact head. The independent empty-input control returned:

```
BIT_AND(empty) = 18446744073709551615
BIT_OR(empty) = 0
BIT_XOR(empty) = 0
APPROX_COUNT_DISTINCT(empty) = 0
```

For the equivalent correlated scalar subqueries whose outer key has no matching child row, this head returned NULL for all four. `COUNT(empty)=0` and `SUM(empty)=NULL` were correct controls. Thus the rewrite still violates the core invariant: an implicit scalar aggregate over a correlated empty input must produce the same scalar value as that aggregate over the equivalent uncorrelated empty input.

Please drive post-join reconstruction from a complete canonical aggregate empty-input contract rather than a local name whitelist (or conservatively keep unsupported shapes from producing a wrong plan). Apply it to every aggregate referenced by the final scalar projection, including mixed expressions. Add a public metamorphic regression comparing correlated missing-key results with equivalent uncorrelated empty-input results across both NULL-on-empty and neutral-on-empty aggregate families; include BIT_AND/OR/XOR and APPROX_COUNT_DISTINCT.

Fresh `go list`, build, vet, focused new regressions, full `pkg/sql/plan` tests, and the exact-head service build pass; the blocker is semantic coverage, not test instability.

VioletQwQ-0:
Addressed the new empty-input aggregate finding on exact head `eb983b2a748c8069845b4924680d89df591f3de8`.

- Added a canonical aggregate-ID empty-result contract in `aggexec` for all current aggregate families (`NULL`, typed zero, all bits set, or conservatively unsupported).
- Applied it to both direct `AGG` and `PROJECT -> AGG` correlated scalar shapes, including mixed final expressions.
- Added typed boundary coverage for `uint64`, `BINARY`/`VARBINARY` width-preserving identities, unknown IDs, and conservative unsupported direct/project paths.
- Added the requested public metamorphic BVT comparing a correlated missing key with an equivalent uncorrelated empty input for `BIT_AND`, `BIT_OR`, `BIT_XOR`, `APPROX_COUNT_DISTINCT`, and `SUM`.

- canonical BVT: `40/40`, 100%
- full `pkg/sql/colexec/aggexec` and `pkg/sql/plan` CGo package tests: PASS
- package build and vet: PASS
- `make build`: PASS
- `git diff --check`: PASS
- self-review and exact-head preflight: `PASS review=PASS validation=PENDING`

The prior CI cancellations were superseded by run `30784170502`, whose required jobs passed. New exact-head CI is now pending.

heni02 (approved):

XuPeng-SH (approved):
Re-reviewed exact head `eb983b2a748c`. The previous blocker is closed: empty-input reconstruction is now driven by a complete current aggregate-ID contract in `aggexec`, with NULL/zero/all-bits-set/unsupported outcomes, and is applied to both direct AGG and PROJECT→AGG shapes. BIT_AND/OR/XOR, APPROX_COUNT_DISTINCT, mixed expressions, typed uint64/binary identities, and conservative unsupported paths are covered; the public regression compares correlated missing-key results with equivalent uncorrelated empty inputs.

Fresh exact-head evidence: full `pkg/sql/colexec/aggexec` and `pkg/sql/plan` tests pass, controlled build/vet pass, diff check passes, and GitHub CI is green. No remaining blocker found.

Issue #11297, opened 2023-08-18
[Bug]: cte recursive: multi recursive layers, insufficient memory error reported


heni02:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Environment

```Markdown
- Version or commit-id (e.g. v0.1.0 or 8b23a93):
- Hardware parameters:
- OS type:
- Others:
```

### Actual Behavior

mo limit 100 iterations，mem 16GB mac execute sql reported  :Invalid alloc size 1147486208，about 10GB
<img width="548" alt="企业微信截图_0a65cf72-5009-4fb5-ac79-051c425becdc" src="https://github.com/matrixorigin/matrixone/assets/113406637/3f5f56b4-5ce6-4cae-b21f-e54022b21721">

mysql:
<img width="1403" alt="企业微信截图_a47a46d1-161d-43cd-a489-db7c67a0e74e" src="https://github.com/matrixorigin/matrixone/assets/113406637/4a7ac550-0c1f-49ed-bd6c-b749fca8420b">

### Steps to Reproduce

```Markdown
with recursive  cte_ab_8(productID,price) as(select p_id,price from product  union all select c.productID,p.price from product p join cte_ab_8 c on p.p_id = c.productID)select * from cte_ab_8;
```

iamlinjunhong:
split in 1.1.0

heni02:
retest on 128 server， it is reported same error ：internal error: Invalid alloc size 1147486208
<img width="1340" alt="企业微信截图_cb2411c2-88ca-413c-baf5-519bd82a78e8" src="https://github.com/matrixorigin/matrixone/assets/113406637/c6719b7b-c19a-4782-82d4-da92d15178fa">

so it is two questions：
1.Invalid alloc size 1147486208, about 1GB is small？
2.which priority is higher for memory limit and limit 100 limit errors? mysql returns limit errors quickly during execution

iamlinjunhong:
无进度，在看其他问题

iamlinjunhong:
join 的结果是个超过 1G 的batch，导致在 dup 的时候出错

iamlinjunhong:
把join probe 出来的大 batch，切分成小 batch 后由于递归计算出的中结果太大（几十上百亿行数据），mo 跑一段时间会 oom，需要做 spill。mysql 在 129 上还未能跑出结果

iamlinjunhong:
mysql 在 128上报错
[...]

PR #26511, opened 2026-07-31
fix: add recursive CTE memory quota

VioletQwQ-0:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

issue #11297

Adds a statement-scoped, per-CN memory circuit breaker for batches retained by recursive CTE operators.

- Adds the dynamic GLOBAL/SESSION variable `cte_max_memory_bytes` (default 1 GiB, `0` disables it).
- Shares one overflow-safe budget across recursive CTE owners under the same query `BaseProcess`.
- Admits logical payload before copying and reconciles successful cache ownership to `Batch.Allocated()`.
- Preserves `MergeCTE` cache reuse while safely re-admitting it across statement generations; `MergeRecursive` continues full cleanup on Reset/Free.
- Replaces Last markers without retaining old data backing and returns error 20825 with `ExecStop` on quota failure.
- Keeps `cte_max_recursion_depth` behavior unchanged.

## Testing

- CGo package tests: `process`, `cteaccount`, `mergecte`, `mergerecursive`, `compile`, and `pipeline`.
- Race detector: all four owning packages; 13 focused lifecycle/concurrency tests separately with `-count=100`.
- Targeted `go vet`, `make build`, and `git diff --check`.
- Isolated CN protocol run: 500-level convergent recursion passed; the #11297 query returned error 20825 at a 16 MiB limit; a depth limit of 3 still returned error 20819; a following `SELECT 1` succeeded.
- Added distributed BVT `recursive_cte_memory.sql/.result`.

Standard mo-tester execution is pending because this host has no Java runtime. The requested 8 GiB Docker base/head comparison is also pending because the Docker daemon is unavailable; the unbounded base query was not run on the shared host.

## QA

QA required: yes.

Please verify on an 8 GiB node that the default 1 GiB quota returns the documented error while the CN remains alive, and cover the default value, `0` escape hatch, and recursion-depth compatibility.


XuPeng-SH (approved):
Deep review completed on 7a8df4de13f9c71ba48f06b45840812403b2d013. The statement-generation budget, shared reservations, cache replacement rollback/discard paths, Reset/Free ownership, coordinator-CN session-variable resolution, and error propagation are internally consistent. Local build/vet and non-race package tests passed; all 15 changed Go tests passed individually under race with computed repeat counts (5–10), followed by full owning-package race tests. The 8 GiB CN survival check remains an explicit QA item.

Issue #26500, opened 2026-07-31
perf(vector): separate checked decoding from hot zero-copy UnmarshalBinary


XuPeng-SH:
## Problem

[#26332](https://github.com/matrixorigin/matrixone/pull/26332) correctly fixes [#26325](https://github.com/matrixorigin/matrixone/issues/26325): a truncated `Batch.UnmarshalBinary` payload must return an error rather than be accepted or panic.

However, it also adds a full `validateVectorBinary` pass to every zero-copy `Vector.UnmarshalBinary` call. For varlen vectors this scans every element, including `Varlena` bounds and array-payload alignment. That changes a hot decode from O(1) metadata binding to O(rows) work.

## Reproduction and evidence

Exact adjacent-commit A/B on the same CI runner, fresh `tpcc10.10` data for each side:

| commit | tpmTOTAL | tpmC | result |
| --- | ---: | ---: | --- |
| `b2aec782` (parent) | 31,191.64 | 13,999.86 | 0 transaction errors; consistency check passed |
| `b609292` (#26332) | 23,777.67 | 10,714.92 | 0 transaction errors; consistency check passed |

This is a 23.77% tpmTOTAL and 23.46% tpmC decrease in one ordered pair. The shared runner makes the absolute percentage noisy, but the causal commit is unambiguous because the commits are adjacent.

CPU profiles from the same runs show the added work directly:

- `b609292`: `vector.validateVectorBinary` is 140.98 CPU seconds / 9.26% flat / 15.13% cumulative.
- `b2aec782`: the function does not exist.

The earlier `cf6dca` vs `411017` comparison independently showed the same hotspot, including `Varlena.OffsetLen` and `Varlena.IsSmall` under this validator.

## Why this is a design issue

The original issue requires safe framing and bounds handling: no unchecked slices, no acceptance of truncated metadata, correct cleanup on failure. It does not require every consumer of a self-produced canonical vector encoding to rescan all values.

`UnmarshalBinary` is used by object IO, plan/runtime-filter decoding, batch decode, checkpoint tooling, and other paths. The API currently has no explicit trust/provenance contract, so the expensive semantic validation runs for every decode. It defeats the purpose of the existing zero-copy representation on high-frequency object-read paths.

[...]

PR #26513, opened 2026-07-31
fix: avoid repeated validation and full cache copies on object reads

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26500
issue #26521

PR #26332 correctly made the default `Vector.UnmarshalBinary` reject malformed input, but the full null-bitmap and varlen validation added an O(rows) semantic scan to every object-column decode. The first version of this PR removed that repeated scan on validated cache hits, but preserving cache immutability by duplicating every marked varlen Vector still copied the complete block before filtering or selecting a few rows. That retained a foreground performance regression for TPCC compound-PK reads and sysbench `random_points`.

This PR keeps the checked decoder as the default for wire, disk, RPC, or otherwise unvalidated bytes. ObjectIO grants a private trusted capability only after a V2 varlen column passes the full checked decoder after decompression and before cache admission. The cache representation is sealed: public `Bytes`, generic `DecodeCached`, and length-changing `Slice` consumers receive independent storage, and mutable Vectors never escape with aliases to the marked backing. V1 and all unmarked data stay on the checked path.

The foreground read path now avoids the remaining full-block copy:

- ObjectIO can execute an immutable search description while the FileService cache entry is pinned. It supports the existing varchar EQ/IN, prefix, less/greater, between, and open-range semantics used by hidden compound primary keys. Search values and bounds are cloned at construction; no callback or borrowed Vector crosses the ObjectIO boundary.
- Sorted `PREFIX_IN` retains the existing merge/gallop algorithm and allocates by hits/needles rather than by block rows. Unsorted search scans rows and short-circuits on the first matching needle.
- Fake-PK, membership-filter/BF, unsupported-type, malformed-filter, and mismatched-input cases retain the legacy owned-vector path or fail open. This preserves the existing residual-filter and conservative conflict-check contracts.
- Selected block reads copy only selected rows directly into caller-owned output Vectors. Normal full scans make the one output copy they inherently require, without an intermediate full `Dup`. Vector TopN and other unsupported paths retain the legacy behavior.
- `PKPersistedBetween` searches varchar PKs and checks only matched commit timestamps while entries are pinned. CN/TN tombstone checks use the same exact scoped search, with tombstone PK search always unsorted because tombstones are ordered by rowid. Commit timestamps retain the `(from, to]` contract; unavailable/null/malformed data remains conservative.
- Appendable visibility, rowid generation, tombstone masks, cache leases, mpool ownership, and error cleanup are covered explicitly. Constant commit timestamps are treated as logical broadcasts; constant-null timestamps fail closed.

The default checked API contract is documented in `Vector.UnmarshalBinary`: trusted decoding is restricted to internally validated immutable bytes and is not a general replacement for checked decoding.

No on-disk or wire format changes are made. `UnmarshalBinaryWithCopy`, reader-based decoding, and V1 decoding are not granted the trusted capability. The ObjectTool short-header path preserves its existing `decode object column` error context. The PR also removes the fixed map allocation added to reused narrow Batch decodes; batches with at most 64 columns use a bounded prefix scan, while wider batches retain the map path.

## Performance evidence

The included 8192-row decoder microbenchmarks isolate the validation scan:

| case | checked | trusted |
| --- | ---: | ---: |
| inline varchar | 7.24-7.28 us/op | 17.29-17.41 ns/op |
| null-heavy int64 | 86.85-87.34 ns/op | 20.70-20.81 ns/op |
| float32 array | 8.43-8.49 us/op | 18.07-18.19 ns/op |

The earlier `shanghai50` A/B on main `475b1d19a4` versus PR head `2befe844c5` established the causal validator hotspot: main spent 28.93 CPU seconds in `vector.validateVectorBinary` (10.27% flat, 16.54% cumulative), while it disappeared from the trusted profile. That run predates the final scoped materialization change and is root-cause evidence, not a final throughput claim.

For #26521, the static production path is closed: sysbench `random_points` rewrites secondary-index `k IN (...)` to a varchar compound-PK `PREFIX_IN`; the cache entry is now searched in scope and only the matching `c`/`pad` rows are materialized. A regression test with 8192 rows and 10 sparse needles enforces less than 4 KiB allocated per search, preventing a block-sized temporary marks allocation. A fresh end-to-end sysbench/TPCC throughput A/B was not run for this final head, so throughput recovery remains subject to performance-environment verification.

## Verification

Tested after rebasing onto main `26ed429b60`:

- Full non-race and full `-race -count=1` tests for ObjectIO, ObjectIO/ioutil, readutil, blockio, disttae, vector, batch, FileService, FIFO cache, and ObjectTool.
- Post-rebase full tests for all five newly modified packages.
- Post-rebase focused `-race -count=5` coverage for immutable predicate parity, bounded `PREFIX_IN` allocation, constant/null handling, selected/full materialization, appendable visibility, malformed/fake/BF fallback, PK range checks, and CN/TN varchar tombstones.
- `go vet` for all ten owning/dependent packages.
- `go build ./cmd/mo-service`.
- `git diff --check`.


pkg/objectio/constructors.go
```diff
@@ -65,12 +93,94 @@ func constructorFactory(size int64, algo uint8) CacheConstructor {
...
+// its full validation.
+func DecodeCached(data fscache.Data) (any, error) {
+	if data == nil {
+		return nil, moerr.NewInvalidInputNoCtx("nil object cache data")
+	}
+	return decode(data.Bytes(), isValidatedVectorCacheData(data))
+}
+
+func decode(buf []byte, trusted bool) (any, error) {
+	if len(buf) < IOEntryHeaderSize {
```
LeftHandCold, line 225: [P2] Preserve the decode context for short headers

This new early return changes a truncated object column from the contextual `decode object column: ...` error to a bare `unexpected EOF`. That is a concrete downstream regression: the current Ubuntu coverage run fails `TestDecodeCommitTSVectorReturnsErrorForTruncatedHeader` with exactly `Error "unexpected EOF" does not contain "decode object column"`.

Please preserve/wrap the context on the normal error path (for example at `decodeObjectColumn`), or deliberately update the downstream contract and its test if the raw error is now intended. As submitted, this head cannot pass the existing UT.

pkg/container/vector/vector.go
```diff
@@ -795,6 +795,22 @@ func (v *Vector) MarshalBinaryWithBuffer(buf *bytes.Buffer) error {
 }
 
 func (v *Vector) UnmarshalBinary(data []byte) error {
```
LeftHandCold, line 806: [P2] Document the checked decoder's safety contract

#26500 explicitly requires the default/checked API to have a documented safety contract. The new trusted API is documented well, but the exported `UnmarshalBinary` entry point still has no comment explaining that it performs the full null-bitmap/varlen semantic validation and is the required choice for external or otherwise unvalidated bytes. Since safe caller selection is central to this split, please document that contract here.

LeftHandCold (commented):
Blocking review comments (GitHub does not permit the currently authenticated author account to formally REQUEST_CHANGES on its own PR).

I reviewed this against #26500 and the regression introduced by #26332. The core design looks sound: V2 cache bytes are fully checked before local cache admission, the private marker cannot be forged outside objectio, the trusted bind retains constant-time structural checks, and V1/unmarked data stays on the checked path. I did not find a data-correctness or cache-ownership blocker.

Please address these concrete acceptance gaps before approval:

1. This head deterministically fails the Ubuntu coverage UT in pkg/tools/objecttool because the new short-header return bypasses the existing decode-context contract (inline comment).
2. #26500 explicitly requires a fresh tpcc10.10 alternating-order A/B and a CPU profile showing validateVectorBinary is gone from the trusted object-read hot path without weakening boundary validation. The PR currently contains microbenchmarks only; please attach that end-to-end evidence.
3. The issue also explicitly requires the default/checked API safety contract to be documented (inline comment).

This review is about behavior and the issue's performance acceptance, not formatting or style.

pkg/objectio/constructors.go
LeftHandCold: Fixed in 61ba79558b. decodeObjectColumn now restores the existing `decode object column` context only for the new short-header error return, without broadening or changing ObjectIO production error semantics. I reproduced the exact Ubuntu Coverage failure before the fix; `TestDecodeCommitTSVectorReturnsErrorForTruncatedHeader` now passes under coverage. Full objecttool coverage is 87.3%, and the affected-package race suite passes.

pkg/container/vector/vector.go
LeftHandCold: Fixed in 61ba79558b. `Vector.UnmarshalBinary` now documents that it performs full null-bitmap and varlena/array payload validation and is the required API for wire, disk, RPC, or otherwise unvalidated bytes. The trusted API remains restricted to ObjectIO-owned, previously validated cache data.

LeftHandCold:
Review follow-up on head `61ba79558b` (rebased onto current main `60e36bef64`):

- Restored the objecttool short-header error contract; the exact Ubuntu Coverage regression now passes.
- Documented the checked `Vector.UnmarshalBinary` safety/caller contract.
- Re-ran build, vet, complete affected-package UT, objecttool coverage (87.3%), and complete affected-package race tests; all passed.
- I am deliberately not presenting a local Apple-laptop TPCC result as merge evidence. The TPCC10.10 alternating A/B run and CPU profile requested by #26500 remain appropriate for MatrixOne’s dedicated benchmark environment; the PR body now states this explicitly. The included microbenchmarks demonstrate removal of the rescanning mechanism, but are not labeled as end-to-end throughput proof.

Both inline review threads have been addressed and resolved.

XuPeng-SH (changes requested):
Requesting changes on head `61ba79558b75`.

Two implementation blockers remain:

1. **The validation capability does not actually seal its backing bytes.** `validatedVectorCacheData` embeds `fscache.Data`, whose public `Bytes() []byte` exposes the marked encoding for mutation. I reproduced this locally: validate a V2 out-of-line varchar, overwrite its Varlena offset through `data.Bytes()`, then `DecodeCached(data)` still sees the marker and returns nil because `UnmarshalBinaryTrusted` skips the Varlena scan; `GetStringAt(0)` then panics on the forged offset. `Slice` has the same alias problem because `fileservice.Bytes.Slice` mutates the underlying object while existing marked aliases survive. This violates #26500's requirement that the trusted path cannot accidentally consume external/unvalidated bytes. Please make the trusted owner actually immutable/revocable rather than relying on a mutable interface plus convention.

2. **Remote-cache hits never transition to the validated capability, so the hot-path regression remains after local admission.** `RemoteCache.Read` installs plain `NewBytes` (`pkg/fileservice/remote_cache.go:147-153`). Before the remote read, `S3FS.Read` has already registered deferred `memCache.Update` (`pkg/fileservice/s3_fs.go:670-686`); when the remote hit completes the vector, the return runs that defer and admits the same unmarked data locally. Every later memory-cache hit therefore executes the full O(rows) V2 validation again. Please validate/mark once before local admission and add a remote -> memory-cache-hit regression test proving subsequent binds take the trusted path.

Separately, #26500 explicitly requires fresh alternating-order `tpcc10.10` A/B runs plus a CPU profile showing `validateVectorBinary` is absent from the trusted object-read hot path. The PR explicitly does not provide that evidence. The direct decoder microbenchmarks are useful, but do not establish the end-to-end acceptance criterion.

Local verification otherwise passed: diff check; build/vet; full non-race and race tests for the five owning packages; the nine added/modified tests repeated individually under race (11-17x); dependent blockio/containers tests; and both new benchmarks.

LeftHandCold:
Addressed the blocking review on new head `2befe844c5` (rebased onto main `475b1d19a4`).

1. The trusted capability now actually seals its backing bytes: it no longer embeds mutable `fscache.Data`; public `Bytes()` returns a clone, length-changing `Slice` returns an independent unmarked copy, and ObjectIO alone has private zero-copy access. The exact forged-Varlena mutation reproducer now remains valid and does not panic.
2. Remote hits now use the decompressed cached-representation size, copy into independently owned/capacity-guarded storage, validate/seal before deferred memory admission, and release rejected inputs. The new remote -> memory integration test proves one validation, one RPC, and a marked second memory hit.
3. Full FileService/FIFO, vector, batch, ObjectIO, ObjectIO/ioutil, and ObjectTool UT/vet/race/coverage gates pass. Added/modified regressions were also stressed separately under race (100x; ObjectWriter 50x).
4. Dedicated-host evidence is now in the PR body. In the completed fresh-data A1/B1 pair both sides had 0 transaction errors and passed consistency. Main profile: `validateVectorBinary` 28.93s / 10.27% flat / 16.54% cumulative. PR profile: the function is absent from the top 100 and the ObjectIO decode hot path. The host had enough throughput drift that I am not overclaiming the single-pair tpm difference; a reverse B2 run was intentionally stopped early and is not reported.

The prior inline threads are already resolved; this review was submitted as a review body and has no resolvable thread. Please re-review the new head when convenient.

XuPeng-SH (changes requested):
Request changes on head `2befe844c59df8d5802728da7639c271900ea887`.

The validation marker still does not seal the bytes consumed by the trusted decoder. `DecodeCached` obtains the marker-private backing and `UnmarshalBinaryTrusted` binds the returned `Vector` directly to it (`pkg/objectio/constructors.go:201-223`). That Vector exposes mutable aliases through `GetData`/`GetArea` (and mutating vector APIs).

Reproduction: validate a V2 out-of-line varchar cache entry; decode it once; overwrite the Varlena offset through the first decoded Vector’s `GetData()`; then decode the same still-marked cache entry again. A normal `DecodeCached(fileservice.NewBytes(cacheData.Bytes()))` rejects the mutated offset, but the marked entry takes the trusted path successfully and `GetStringAt(0)` panics. Thus a cache entry can remain marked after its semantic validation is invalidated, so later hits skip the check that should reject it.

Please make the trusted backing genuinely immutable to all decoded-vector consumers, or revoke/drop the validation capability before exposing a writable alias, and add this cross-hit regression. This is a code-correctness/cache-lifecycle blocker.

LeftHandCold:
Addressed the cache-marker aliasing blocker on head `b25e9299bb`. The marker no longer exposes its sealed backing to returned vectors: generic `DecodeCached` binds an independent snapshot, while the hot `LoadColumnsData` path only parses the backing synchronously and duplicates into its existing mpool before returning. Fixed-width vectors remain unmarked/checked-zero-copy so the isolation fix does not turn their O(1) decode into an O(bytes) copy. Added the exact cross-hit regression covering mutation through both `GetData` and `GetArea`; later marked hits still decode the original value. Rebased onto main `5b9eeb54ec`; focused race stress and the complete seven-package race matrix pass. Per request, I did not rerun `shanghai50`.

LeftHandCold:
Updated the PR to head `d1db54628f` after rebasing onto current main `bd890eee0e`.

SCA was failing on a real control-flow contradiction in `BlockDataReadInner`: the selected-row fast path has already returned whenever `orderByLimit` is nil or ordered, so the later `if orderByLimit != nil` was tautologically true. The guard is now removed, while the same TopN call and error return remain. This is an equivalent control-flow cleanup, not a behavioral-path expansion.

The cache-search/materialization fix remains unchanged: cache backing stays confined to synchronous pinned ObjectIO operations, selected reads only materialize matching rows, and fake PK/BF/unsupported/null/malformed inputs keep their conservative legacy fallbacks. The static `random_points` path from #26521 is still covered (`k IN (...)` -> varchar compound-PK `PREFIX_IN` -> selected `c`/`pad` materialization).

Verification on this head:

- CI-version `golangci-lint v2.6.2` with Go 1.26.4: `./pkg/vm/engine/tae/blockio/...` reports 0 issues;
- `TestHandleOrderByLimitOnSelectRows` under race: measured below timer resolution, then passed `-count=100` using the 30-second adaptive-budget cap;
- complete `pkg/vm/engine/tae/blockio` race test, ordinary test, `go vet`, and build all passed;
- `git diff --check` passed.

No fresh end-to-end sysbench/TPCC A/B result is claimed for this head.

XuPeng-SH (changes requested):
Request changes on head `d1db54628fc0`.

The two previous implementation blockers are resolved. The validated cache representation now seals its backing from `Bytes`, `Slice`, and decoded-Vector mutation; remote hits are copied into independent storage, validated before memory admission, and retain/release ownership is closed. I found no new correctness, concurrency, fallback, or resource-lifecycle blocker in the latest code.

One acceptance blocker remains: #26500 explicitly requires a fresh `tpcc10.10` run with fresh data and alternating main/PR order, plus a CPU profile demonstrating that `validateVectorBinary` is absent from the trusted object-read hot path without weakening boundary validation. The PR body states that no fresh end-to-end sysbench/TPCC throughput A/B was run for this final head. The cited profile is from `2befe844c5`, before the final scoped-materialization/cache-isolation changes, so it is causal evidence but not final-head acceptance evidence. Given the 30–40% end-to-end regression recorded in #26521, decoder microbenchmarks and static path closure do not establish recovered throughput.

Please attach final-head A/B evidence against base `bd890eee0e83` and the corresponding profile, or obtain an explicit waiver of that acceptance criterion from the performance owner. If that criterion is waived, I consider the code ready to approve.

Verification on this exact head: affected-package tests, focused disttae PK/tombstone tests, targeted race tests, build, vet, diff check, and current GitHub CI all pass.

LeftHandCold:
@XuPeng-SH Thanks for the final review. I agree with the code assessment: there are no remaining correctness, concurrency, fallback, or resource-lifecycle blockers on `d1db54628fc0`.

Per the PR author’s explicit direction, the fresh final-head `tpcc10.10` A/B and CPU-profile acceptance item is waived for this PR. We will not claim a final end-to-end throughput number from this head; the older `2befe844c5` profile remains root-cause evidence only.

The merge decision is therefore based on the code-level closure already verified on the exact head: the trusted boundary remains checked and sealed, the hot ObjectIO path does not revalidate/copy whole cached varlen blocks before filtering, conservative fallbacks are preserved, and affected-package/race/vet/build/current CI checks pass.

No code update is needed for this review. Please re-review/approve `d1db54628fc0` under this explicit acceptance waiver.

fengttt (approved):

XuPeng-SH (approved):

PR #26518, opened 2026-07-31
fix(plan): maintain irregular indexes in modern update

### LINKED ISSUE #26341: [Subtask]: Move synchronous irregular-index UPDATE maintenance to the modern planner
## Parent issue

#26336

## Problem

The modern UPDATE binder currently falls back when:

- any column covered by an irregular index is updated; or
- the primary key changes on any table carrying an irregular index.

The legacy path performs inline delete+insert maintenance for synchronous FULLTEXT and IVF-FLAT indexes. It rejects PK updates when such a synchronous index exists because hidden rows are keyed by the old source PK (#25617). Async FULLTEXT/IVF-FLAT and always-async HNSW/CAGRA/IVF-PQ instead rely on ISCP CDC.

The migration must preserve this distinction. It must not introduce an SQL-layer list of algorithm names.

## Required work

- Use plugin metadata (`SyncDescriptor`) plus the index's `async` parameter to select inline versus CDC maintenance.
- Use the existing plugin rewrite contract (`UpdateColumnRewriteHook`) for indexed/INCLUDE-column impact, extending an algorithm-agnostic plugin hook only if the modern DML plan needs more information.
- For synchronous indexes, delete old hidden entries using the old PK/key identity and insert entries from the final updated row image.
- Skip inline work for unrelated columns.
- Preserve CDC-only behavior for async indexes without double maintenance.
- Keep PK updates on tables with synchronous irregular indexes explicitly rejected before mutation until a proven safe rewrite contract exists.
- Verify async PK-update behavior against CDC identity semantics before removing its current broad fallback.
- Keep base-row and synchronous hidden-table mutations in one atomic statement plan.

## Acceptance criteria

- [ ] Updating a synchronous FULLTEXT or IVF-FLAT indexed/INCLUDE column uses the modern planner and produces correct old-entry deletion plus final-entry insertion.
- [ ] Updating an unrelated column produces no irregular-index rewrite.
- [ ] Async FULLTEXT/IVF-FLAT and always-async HNSW/CAGRA/IVF-PQ use CDC only and receive no duplicate inline writes.
- [ ] Synchronous-index PK updates fail before locks/mutations with a stable compatibility error unless safe support is implemented and tested.
- [ ] Async-index PK updates have an explicit tested disposition based on CDC behavior.
- [ ] Classification and rewrite decisions are plugin-driven; no new per-algorithm switch/if chain appears in SQL/catalog layers.
- [ ] Real, composite, and fake source PKs are covered.
- [ ] Indexed parts and INCLUDE columns are covered through plugin contracts.
- [ ] Failure, cancellation, retry, and rollback leave the base and synchronous hidden tables consistent.
- [ ] Supported cases no longer route to `buildTableUpdate`.

## Test matrix

- Synchronous and async FULLTEXT.
- Synchronous and async IVF-FLAT.
- Always-async HNSW controls; CAGRA/IVF-PQ controls under the appropriate GPU build/test coverage.
- Indexed part, INCLUDE column, and unrelated-column updates.
- Real, composite, and fake PKs; PK update accepted/rejected cases.
- Optimistic/pessimistic conflict, retry, cancellation, commit, and rollback.
- CPU-runnable unit coverage for generic plan/plugin decisions; GPU-tagged coverage where an algorithm is GPU-only.

## Dependencies

- Parent: #26336
- Uses #26337 telemetry.
- Must follow the index-plugin registry contract and avoid new algorithm-specific dispatch.

ck89119:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26341

The modern UPDATE planner previously routed affected FULLTEXT and vector indexes back to the legacy planner because it could not maintain their hidden tables. This prevented the UPDATE planner migration from completing and made index behavior depend on planner fallback.

This change keeps plugin-supported irregular indexes on the modern planner path:

- classifies synchronous and asynchronous indexes through plugin metadata and update-column hooks;
- materializes the final UPDATE row image once and shares it with the base-table mutation and synchronous hidden-index delete/insert plans;
- keeps asynchronous indexes CDC-only;
- rejects primary-key updates when a synchronous irregular index would otherwise retain hidden rows keyed by the old primary key;
- covers indexed columns, included columns, implicit `ON UPDATE` columns, route selection, and legacy handling for algorithms not yet migrated to the plugin contract.

Validation:

- `make -j8`
- `make static-check`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 ./pkg/sql/plan`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 ./pkg/sql/plan`
- focused planner tests with `-race -count=100`
- `vector_ivfflat_include_dml_sync.sql` (25/25)
- `fulltext_update_pk.sql` (21/21)
- `fulltext.sql` (255/255)
- manual transaction commit/rollback checks for synchronous FULLTEXT maintenance


heni02 (approved):

XuPeng-SH (approved):
Approved exact head `a8d1aa716b5ad0a459037c317278a184e6debf94`.

No blocking findings. Reviewed the modern UPDATE path through the shared final-row image, base MULTI_UPDATE, and synchronous hidden-index delete/insert maintenance. Async/AlwaysAsync indexes remain CDC-only, synchronous primary-key updates reject before mutation, and the plugin-dispatch guard is clean.

Local verification passed: focused CPU race tests (planner tests x100; cancel/retry integration x1), full plan/issues race and non-race suites, build, vet, and diff checks. The GPU-tag classification test could not link locally because the available libmo artifact is CPU-only; the PR changes no CGo/GPU runtime code, and its registration and planner path were statically checked. All reported CI checks are green.

PR #26520, opened 2026-07-31
refactor: remove stream feature

### LINKED ISSUE #26516: [Refactoring]: Remove all stream-related code and tests
### Why do you want to refactor this code?

The `stream` feature (Kafka adapter + stream connector) is no longer needed and should be removed entirely to reduce maintenance burden, shrink the codebase, and eliminate dead dependencies.

The scope includes:

- `pkg/stream/adapter/kafka` — Kafka adapter (`adapter.go`, `adapter_test.go`, test protobuf files)
- `pkg/stream/connector` — stream connector (`connector.go`, `decoder*.go`, `options.go`, `error.go` and their tests)
- All dependent code that imports `pkg/stream`:
  - `pkg/sql/plan/build_ddl.go`
  - `pkg/sql/compile/compile.go`
  - `pkg/sql/colexec/source/source.go`
  - `pkg/cnservice/server_task.go` (streaming connector task)
  - `pkg/frontend/connector.go`
- All stream-related tests, including `*_test.go` files under `pkg/stream` and any other tests covering the stream feature

### Describe the solution you'd like

1. Delete the `pkg/stream` directory entirely (both `adapter/kafka` and `connector` subpackages, including test files and generated protobuf).
2. Remove all imports/usages of `pkg/stream` in the dependent files listed above and clean up the corresponding dead code paths (e.g. the streaming connector task in `pkg/cnservice/server_task.go`, stream table def handling in `pkg/frontend/connector.go`, and Kafka-related SQL DDL/compile/source code).
3. Delete all stream-related tests and any test resources (e.g. `pkg/stream/adapter/kafka/test/`).
4. Verify the repository still builds (`go build ./...`) and the test suite passes (`go test ./...`) after removal.

### Describe alternatives you've considered

- Keeping the code but marking it deprecated: rejected, since the feature is unused and leaving it in place keeps paying compile-time and maintenance costs.
- Feature-flagging the stream code behind a build tag: rejected, as it adds complexity without benefit for a feature slated for removal.

### Additional information

- Stream code is only referenced from the 5 files listed above; no external consumers outside the repo are known.
- Removing the stream feature will also drop its transitive dependencies (e.g. Kafka client libraries) if no other package uses them.

XuPeng-SH:
## Summary

Closes #26516.

Removes the unused Kafka stream feature closure: source and connector SQL/AST, planner and executor paths, source operator, dynamic-table task integration, Kafka sink executor, feature-only tests, and exclusive dependencies.

The removal preserves the boundaries that still matter:

- `engine.StreamConfigsDef` remains because it is the persisted carrier for generic table properties.
- `catalog.SystemSourceRel` remains as a catalog tombstone. Historical source objects fail closed on reads and remain droppable; they are never reinterpreted as ordinary tables.
- Former plan, pipeline, task, and VM wire numbers and names are reserved, and generated bindings come from the updated schemas.
- Dead parser/frontend identifiers and misleading connector naming in CDC code are removed.

## Legacy daemon-task closure

Historical Kafka sink daemon tasks persisted executor code 4. Removing its executor alone leaves active rows permanently undispatchable and repeatedly logged.

This PR closes the lifecycle without restoring Kafka business code:

- the v4.0.6 cluster upgrade moves only known non-terminal code-4 states to `CancelRequested`; terminal and unknown future states are untouched;
- every new-CN daemon poll also performs a bounded reconciliation of reserved code 4 and those same known non-terminal states, closing late inserts and stale-snapshot overwrites from old CNs during the supported rolling-upgrade window;
- reconciliation preserves the observed owner and uses executor + status + runner CAS, so it requests cancellation without stealing a live old owner or overwriting a newer generation;
- SQL NULL and empty-string runners are treated as the same unassigned owner, while every non-empty runner remains a strict ownership fence;
- cancellation is a generic task-runner control-plane transition; no code-4 executor is retained;
- the queued cancel handler carries only a stable task ID and resolves the current generation at handling time;
- a fresh foreign owner is never retired by a stale handler, and a freshly claimed local task remains `CancelRequested` until its local task and `ActiveRoutine` are published;
- an orphaned/stale task, or a task whose executor has been removed, can still reach `Canceled` without a business executor;
- status and runner predicates fence the terminal write, so a concurrent state or ownership change wins without invoking the stale routine;
- nullable historical Details payloads can be updated without panic, and legacy field-10 payloads remain unknown fields through terminal serialization;
- CDC enters the cancelable `Starting` state before `Attach`; cancellation before the factory calls `Start` fences startup, releases its wait, and does not decrement the not-yet-incremented running gauge;
- daemon storage honors executor conditions consistently in MySQL and memory implementations.

Daemon admission now publishes the claimed `Running` snapshot locally. Heartbeat, error, and lifecycle paths no longer carry the pre-claim dispatcher snapshot. Admission failures do not write an unowned error, and executor errors update only the still-owned `Running` generation.

`Details.Type()` is total: valid CDC, ISCP, and Publication details map correctly; absent or unrecognized details map to `Unknown`. Creation APIs reject unknown details, and `AddCDCTask` rejects non-CDC details with an error rather than panicking. Diagnostic task-name extraction is nil-safe.

## Performance

There is no Kafka code or dependency in the runtime, and no tombstone executor is registered. Ordinary SQL execution, operators, and non-daemon task execution are unchanged.

The durable rolling-upgrade fence adds one control-plane metadata SELECT per enabled CN daemon poll (10 seconds by default), filtered by executor code 4 plus known non-terminal statuses and capped at 16 rows. An empty result allocates no executor or handler and performs no update. Only matching legacy rows incur per-row CAS updates. This bounded control-plane cost is intentional: a one-time migration alone cannot close old-CN late writes. Cancellation readiness checks run only for `CancelRequested` daemon tasks; cancel dispatch also avoids its former map lookup and temporary `daemonTask` allocation. Admission adds no I/O.

## Test plan and results

- Parser and protobuf generation run twice with `-mod=readonly`; the second run is clean.
- On the latest merged main, complete non-race and race suites pass for taskservice, frontend, plan, and process; the MySQL parser and v4.0.6 upgrade suites pass.
- Every added or modified taskservice lifecycle regression passes with `-race -count=100`; the CDC attach-before-start regression also passes with `-race -count=100`.
- A mixed-version sequence covers migration to `CancelRequested`, an old CN overwriting it with a fresh `Running` snapshot, a late unassigned code-4 insert, periodic reconciliation, live-owner deferral, stale-owner takeover, and final `Canceled` convergence.
- Mock-MySQL SQL assertions cover the real `task_runner IS NULL` representation and strict non-empty owner fence; nullable historical Details update is also covered.
- Deterministic lifecycle tests cover storage claim → local task publication → `ActiveRoutine` publication, fresh foreign owner → stale orphan, concurrent state/owner supersession, status-update failure, and stale executor-error writeback.
- The migrated-task regression uses a real TaskRunner with no executor registered for code 4, a fresh persisted owner/heartbeat, and historical field-10 wire data; it reaches `Canceled` without sleeps and preserves the unknown payload.
- The normal CDC factory regression verifies that pre-`Start` cancellation fences startup without leaking a runnable generation.
- Daemon task creation rejects nil/unknown details and a non-CDC payload on the CDC-only API; valid Publication details no longer panic.
- Affected-package build/vet checks and `git diff --check` pass.
- `go list -deps -test ./...` confirms no Kafka, jhump/protoreflect, protocompile, or `pkg/stream` dependency remains.
- Existing parser rejection and legacy source fail-closed tests remain in the affected package suites; no large-data BVT is added.


pkg/taskservice/daemon_task.go
```diff
@@ -595,12 +595,22 @@ func (t *cancelTask) Handle(ctx context.Context) error {
...
 	tk.TaskStatus = task.TaskStatus_Canceled
-	tk.EndAt = time.Now()
-	_, err = t.runner.service.UpdateDaemonTask(handleCtx, []task.DaemonTask{tk})
+	tk.UpdateAt = now
+	tk.EndAt = now
+	updated, err := t.runner.service.UpdateDaemonTask(
+		handleCtx,
+		[]task.DaemonTask{tk},
+		WithTaskStatusCond(task.TaskStatus_CancelRequested),
+		WithTaskRunnerCond(EQ, tk.TaskRunner),
```
LeftHandCold, line 647: [blocking] 这里对未分配 runner 的 CAS 在真实 MySQL 上永远匹配不到。daemon task 初次 INSERT 不写 `task_runner`，表列又可 NULL，所以迁移前尚未 claim 的 code-4 Created 行是 SQL NULL；查询用 `sql.NullString` 后 `tk.TaskRunner` 变成 `""`，但这个 condition 最终生成 `task_runner=''`，而 NULL 不等于空串。于是迁移把它改成 CancelRequested 后，cancel poll 能通过 `last_heartbeat IS NULL` 找到它，却每次 update 都 affected=0 并从 611-612 静默返回，永远到不了 Canceled。这个回归也影响任何在 claim 前被 cancel 的正常 daemon task。请让 empty owner condition 对 NULL-aware（同时保持非空 owner 的严格 fence），并补一条覆盖真实 SQL NULL 语义的 MySQL/mock-SQL 测试；当前 mem test 把 NULL 表示成空串，验证不到这一点。

pkg/bootstrap/versions/v4_0_6/cluster_upgrade_list.go
```diff
@@ -13,6 +13,62 @@
...
+var clusterUpgEntries = []versions.UpgradeEntry{
+	retireKafkaSinkDaemonTasks,
+}
+
+var retireKafkaSinkDaemonTasks = versions.UpgradeEntry{
+	Schema:    catalog.MOTaskDB,
+	TableName: catalog.MOSysDaemonTask,
+	UpgType:   versions.MODIFY_METADATA,
+	UpgSql: fmt.Sprintf(
+		"update %s.%s set task_status = %d, update_at = current_timestamp() where %s",
```
LeftHandCold, line 39: [blocking] 这个一次性 UPDATE 在 mixed-version CN 升级窗口里不能形成闭包。旧 binary 可以在迁移前已经 poll 到 Created，随后用没有 status fence 的旧 `startDaemonTask` 把刚改成 CancelRequested 的行覆写回 Running；旧 CN 也仍保留 CREATE CONNECTOR，并会在迁移提交后直接插入新的 executor=4 Created 行，代码里没有 cluster-version/retirement gate。entry 完成后不会再 sweep，而新 runner 没有 code-4 executor，只会持续 dispatch 失败，因此这些晚到/被覆写的行会永久留在非终态。请增加持续的 fail-closed reconciler（按 reserved code 4 + 已知非终态做带状态 CAS 的周期性 retirement）或等效 durable fence，并补 mixed-version 状态序列测试；若这里依赖运维上先彻底 drain 所有旧 CN，需要先给出可由代码保证的升级约束。

LeftHandCold (changes requested):
深度复核后需要先修两个 legacy-task retirement 闭包问题：真实 MySQL 的 NULL runner 会让未 claim 任务永久卡在 CancelRequested；一次性 upgrade UPDATE 也无法覆盖 mixed-version CN 的旧 handler 覆写和 late writer。stream 删除闭包、普通表路径、protobuf/VM wire 保留本身未发现其他 blocker。

pkg/taskservice/daemon_task.go
XuPeng-SH: 已在 ec32ed4c26 修复。空 owner 的 SQL 条件现在同时匹配 task_runner 为空字符串或 SQL NULL；非空 owner 仍保持严格等值 CAS。新增 mock-SQL 回归直接断言 NULL-aware predicate，并补了历史 NULL Details 更新不 panic 的存储边界。合并最新 main 后，taskservice 完整 non-race/race 与目标用例 race 100 次均通过。

pkg/bootstrap/versions/v4_0_6/cluster_upgrade_list.go
XuPeng-SH: 已在 ec32ed4c26 增加持续 fail-closed 对账。每轮 daemon poll 按 reserved code 4 + 六个已知非终态查询，limit 16；写入 CancelRequested 时带 executor + observed status + observed runner CAS，因此能收敛旧 CN 的晚插入/旧快照覆写，又不会抢占新 owner 或改写未知/终态。新增混合版本状态序列覆盖 migration 后 Running 覆写、late Created 插入、fresh owner deferral、stale takeover 和最终 Canceled。最新 main 已合并于 5873312b0f，交叉包 non-race/race 均通过。

aptend (approved):
Reviewed the stream-feature removal and the follow-up lifecycle/rolling-upgrade fixes. The NULL-aware owner fence and continuous retirement reconciler close the previously identified gaps; focused non-race/race tests passed locally.

pkg/taskservice/daemon_task.go
```diff
@@ -770,24 +826,59 @@ func (r *taskRunner) dispatchTaskHandle(ctx context.Context) {
...
+		retired.TaskStatus = task.TaskStatus_CancelRequested
+		retired.UpdateAt = retiredAt
+
+		updateCtx, cancel := context.WithTimeout(ctx, r.options.fetchTimeout)
+		updated, err := r.service.UpdateDaemonTask(
+			updateCtx,
+			[]task.DaemonTask{retired},
+			WithTaskExecutorCond(EQ, retiredKafkaSinkTaskCode),
+			WithTaskStatusCond(observed.TaskStatus),
+			WithTaskRunnerCond(EQ, observed.TaskRunner),
```
LeftHandCold, line 874: [blocking] This is not a lease CAS. A live old CN can renew `last_heartbeat` after `observed` is read; `HeartbeatDaemonTask` changes only that field, so executor/status/runner still match and this full-row `UpdateDaemonTask` overwrites the fresh heartbeat with the stale snapshot while writing `CancelRequested`. A new CN can then treat the owner as stale and write `Canceled` without having its `ActiveRoutine`. Once the row is `Canceled`, the old CN no longer selects it in `cancelTasks`, so it never closes the Kafka connector's `cancelC` and that routine can keep consuming/inserting until the old CN exits. `cancelTask` has the same read/check/CAS gap because its terminal write also omits heartbeat. Please preserve liveness in reconciliation and atomically require a foreign lease to remain expired at terminalization (for example, a partial status write plus a heartbeat cutoff/generation fence), and add a regression that renews heartbeat between the read and update.

LeftHandCold (changes requested):
深度复核后还剩 1 个实际可达的 mixed-version lifecycle correctness blocker：旧 CN 的 heartbeat renewal 没有进入 retirement/terminalization CAS，新 CN 可能把仍在运行的旧 Kafka task 写成 Canceled，却无法停止其本地 routine。此前的 NULL runner 与 one-shot migration 缺口已修复；stream 删除、wire reservation、catalog tombstone、Details/CDC admission 等路径未发现其他 blocker。

pkg/taskservice/daemon_task.go
XuPeng-SH: 已在 5531d56e06f 按 lease 状态机修复，不再用 stale snapshot 整行回写：

1. 新增 status-only storage transition，只写 task_status/update_at/end_at，保留 TaskRunner、LastHeartbeat 和 payload；reconciler 仍用 executor + observed status + observed runner CAS，因此 concurrent heartbeat 不会被覆盖。
2. cancelTask 代替 foreign/unassigned owner 终态化时，使用与读取判定相同的 heartbeat cutoff，并将 status + runner + last_heartbeat<=cutoff 放进同一条原子 UPDATE；续租发生在写入前时 affected=0，任务保持 CancelRequested。
3. memory storage 的 heartbeat 也改为只写 LastHeartbeat，与 MySQL 契约一致，避免测试后端继续整行覆盖状态。

新增两个确定性 race 回归，均在 read 与 update 之间注入续租：
- reconciler 仍写入 CancelRequested，但保留新 heartbeat；
- foreign terminalizer 被 lease CAS 拒绝，不能写入 Canceled。

验证：
- 6 个新增/修改行为测试各自 `-race -count=100` 通过；
- `pkg/taskservice` 完整 non-race 与 `-race -count=1` 通过；
- taskservice/cnservice/frontend/mock package build、vet 通过。

LeftHandCold (approved):
复核当前 c5e34a5：上轮 heartbeat/lease blocker 已关闭。reconciler 现在使用 status-only CAS，不再覆盖并发续租；foreign/unassigned terminalization 将 observed status、runner 与 last_heartbeat <= cutoff 放进同一条原子更新，写前续租会使 CAS 失败；memory heartbeat 也与生产 MySQL 的 heartbeat-only 语义一致。精确增量及合并 main 未发现新的 stream/wire/catalog 或普通 daemon task 回归。定向 lifecycle/storage 用例通过，git diff --check 干净。失租后 owner revival 属于既有 daemon at-least-once lease 边界，并非本次增量引入或放大的回归，不作为本 PR blocker。

LeftHandCold:
Post-merge 深度复核当前 head `9905d4a`，未发现新的、生产可达的 correctness 或正常业务回归 blocker。

重点重新确认：
- legacy code-4 daemon task 的 migration + continuous reconciliation 覆盖 mixed-version late writer；status-only CAS 不覆盖 owner/heartbeat，foreign/unassigned terminalization 将 status、runner、heartbeat cutoff 放在同一原子更新中，旧 handler 不会误杀新 generation。
- CDC 在 Attach 前进入 Starting，Cancel 可封住尚未进入 Start 的窗口；Start attempt generation、runner lifecycle、detector/routine cleanup 的竞态闭包完整。
- 历史 source catalog 对 SELECT/DML/TRUNCATE fail-closed，同时仍可用普通 `DROP TABLE` 清理；移除的 plan/pipeline/task/VM 编号和字段均正确保留，旧 payload 不会被误解释成其他操作。
- `c5e34a5..9905d4a` 只合并 main；冲突生成物已复核：MySQL parser 强制重新生成后无 diff，plan descriptor 同时保留 stream tombstone 与 main 新增的 check-constraint 字段。

`git diff --check` 干净。按资源约束没有重复跑大范围本地测试；GitHub 的核心 UT/SCA 已通过。PR 已经 merged，GitHub 不能再补交新的 Approve/Changes，因此以此 comment 记录本轮结论。

PR #26527, opened 2026-07-31
fix(tae): synchronize transfer page paths

### LINKED ISSUE #26523: [Bug]: TransferHashPage path races with TTL cleanup in TestClearPersistTransferTable

- [x] I searched open and closed `matrixorigin/matrixone` issues for `TestClearPersistTransferTable`, `TransferHashPage`, `SetPath`, and `ClearPersistTable`; no matching issue was found.

### Branch Name

main

### Commit ID

0b8b53dff3869dcd5e3ab52992852c9ea6545026

- GitHub Actions workflow: `MatrixOne ALL CI`
- Job: `Matrixone CI / UT Test on Ubuntu/x86`
- Runner: Ubuntu x86, Go race detector enabled
- PR head tested: `1fd0b38f5c31b681acca920cd983b05779095b08`

### Actual Behavior

`TestClearPersistTransferTable` fails under the race detector because `TransferHashPage.path` is accessed concurrently without synchronization.

The detected write is:

```text
github.com/matrixorigin/matrixone/pkg/vm/engine/tae/model.(*TransferHashPage).SetPath()
    pkg/vm/engine/tae/model/pages.go:351
github.com/matrixorigin/matrixone/pkg/vm/engine/tae/db/test.TestClearPersistTransferTable.func1()
    pkg/vm/engine/tae/db/test/db_test.go:11690
```

The concurrent read is:

```text
github.com/matrixorigin/matrixone/pkg/vm/engine/tae/model.(*TransferHashPage).ClearPersistTable()
    pkg/vm/engine/tae/model/pages.go:404
github.com/matrixorigin/matrixone/pkg/vm/engine/tae/model.(*TransferHashPage).Close()
github.com/matrixorigin/matrixone/pkg/vm/engine/tae/model.(*TransferTable).executeTTL()
```

The read runs from the transfer-table TTL cron goroutine while the test goroutine assigns the persisted path. The race detector terminates the test with `race detected during execution of test`.

Failing job: https://github.com/matrixorigin/matrixone/actions/runs/30623452935/job/91133128589

This was observed while validating PR #26504, whose diff only changes Data Branch float SQL formatting and its regression tests; it does not touch TAE transfer pages or TTL lifecycle code.

### Expected Behavior

Publication and consumption of a persisted transfer-page path must have a defined synchronization/ownership contract. `SetPath`, loading, TTL cleanup, close, and string/logging paths should not concurrently read and write `TransferHashPage.path` without synchronization, and the focused test should pass under `-race`.

### Steps to Reproduce

1. Check out commit `0b8b53dff3869dcd5e3ab52992852c9ea6545026` or the tested PR head.
2. Run the focused TAE DB test with the race detector, using the repository's required CGo test environment:

```text
go test -race ./pkg/vm/engine/tae/db/test -run '^TestClearPersistTransferTable$' -count=1
```

3. If the timing does not trigger on the first attempt, repeat the focused test or run the full Ubuntu/x86 UT workflow.
4. Observe the race between `TransferHashPage.SetPath` and `TransferHashPage.ClearPersistTable` through the TTL cron job.

### Additional information

The relevant field is a composite `model.Path`. A correct fix should cover all concurrent readers of `page.path`, not only the two race-detector frames shown above, and should preserve the transfer-page TTL and persisted-file cleanup semantics.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26523

`TransferHashPage.path` was a composite value written by `SetPath` while loading, logging, and TTL cleanup read it without synchronization. The page TTL goroutine could therefore race path publication and observe a mixed path.

Path publication and terminal cleanup also form one resource-lifecycle state machine. A page may close before its persisted path is published, and one `WriteTransferPage` batch stores several pages in the same physical file. Page-local deletion is therefore insufficient: a late publisher can leak a file, while deleting for one closed page can invalidate still-live sibling pages.

This PR:

- synchronizes all page path publication, reads, and terminal cleanup;
- introduces one persisted-file cohort per successful write batch, shared by every page published from that file;
- seals the cohort after publication and deletes the physical file exactly once only after the last live page releases it;
- lets already-closed pages reject publication without deleting a file still owned by live siblings;
- uses the same cohort publication path for normal write success and `FileAlreadyExists` recovery;
- keeps file-service deletion outside page/cohort locks;
- adds deterministic single-page and concurrent lifecycle tests, plus two-page tests covering partial/all closure for both write-success branches.

BVT is not applicable because this is an in-process Go concurrency and file-ownership bug with no SQL-visible semantic change.

Validation after the final code change and merge from `main` (using the repository CGo include/library environment):

- focused lifecycle tests under `-race -count=100`
- `go test -race ./pkg/vm/engine/tae/db/test -run '^TestClearPersistTransferTable$' -count=14`
- `go test -race ./pkg/vm/engine/tae/model -count=1`
- `go test -race ./pkg/vm/engine/tae/db/test -count=1`
- `go test ./pkg/vm/engine/tae/model ./pkg/vm/engine/tae/db/test -count=1`
- `go vet ./pkg/vm/engine/tae/model ./pkg/vm/engine/tae/db/test`

API, on-disk format, TTL timing, and transfer-file naming are unchanged. Cohort ownership may retain a shared file until its last page reaches terminal cleanup, which is required to keep live sibling pages loadable. File-service deletion errors retain the existing logged-error and temporary-file GC behavior.


XuPeng-SH (changes requested):
Requesting changes on exact head `0e048ab27de41b9a9954c80cc414cec50b52936c`.

Blocking lifecycle gap: `SetPath` and terminal cleanup are atomic individually, but they are not linearized as one state machine. If TTL removes the page and `Close -> ClearPersistTable` loads nil at `pkg/vm/engine/tae/model/pages.go:411`, then `SetPath` stores the persisted path at `pages.go:350`, the page already has refcount 0 and is absent from `TransferTable`; no later page cleanup can delete that file. It remains until the coarse temporary-file sweep or restart (the sweep interval is one hour), violating the page's disk-TTL cleanup contract and allowing orphaned files to accumulate.

This is a valid ordering of the concurrency exercised by `TestTransferPageConcurrentPathAccess`, but that test's `Delete` is a no-op and it never asserts exactly-once cleanup. I also forced the cleanup-before-publication ordering locally under `-race`: after terminal cleanup, publishing the path left the delete count at zero and another TTL pass could no longer find the page.

Please linearize path publication with terminal state: either cleanup sees and purges every published path, or a late `SetPath` detects terminal state and immediately owns cleanup. If the intended contract is instead `SetPath` before `AddPage`/Close, enforce that at every caller and update the concurrent test accordingly. Please add a deterministic cleanup-before-publication test that asserts the file is deleted exactly once.

gouhongshen:
Fixed in `a908b7021`. The lifecycle gap is valid: the previous atomic pointer removed the data race but did not linearize publication with terminal cleanup.

`SetPath`, synchronized reads, and `ClearPersistTable` now share one page-local state mutex. Cleanup marks the state terminal and claims any published path; if cleanup wins first, the late publisher does not store the path and immediately owns deletion. File-service deletion runs after unlocking. Repeated cleanup is a no-op.

Added deterministic tests for both cleanup-before-publication and publication-before-cleanup, each asserting exactly one delete, and updated the concurrent test to assert exactly-once cleanup across the competing ordering. Focused race stress passed at `count=100`; the original `TestClearPersistTransferTable` passed under race at `count=14`; full model and db/test packages passed with and without race.

XuPeng-SH (changes requested):
Reviewed exact head `a908b7021f7c43dc9a5c8e8991de1958628dc750`. The original single-page cleanup-before-publication leak is fixed, but the new late-publication cleanup has a blocking shared-ownership bug.

`WriteTransferPage` stores every page in one `ioVector.FilePath` and then publishes that same `Path.Name` to each page (`pages.go:499-506`, with the same logic at 516-523). If one page has already reached terminal cleanup, `SetPath` immediately deletes that whole file at `pages.go:357-360`; the loop then publishes the same now-deleted file name to any still-live sibling pages. Those pages retain a non-empty path but cannot load their transfer map after memory eviction.

I reproduced this deterministically under `-race` with two pages from `makeTestPages(2)`: close page 0 before `WriteTransferPage`, leave page 1 live, then run the write. Delete receives `test-transfer-page`, while page 1 subsequently stores `Path{Name: "test-transfer-page", ...}`. The diagnostic test was removed after confirmation. The current tests only construct one page, so they cannot detect this file-level ownership failure.

Please coordinate late cleanup at the write/file cohort rather than inside one page. For example, let `SetPath` report whether the page accepted publication and have `WriteTransferPage` delete the shared file once only if no live page accepted it; otherwise retain it for the live pages. Alternatively, enforce and test the invariant that all paths are published before pages can reach terminal cleanup. Add a mixed two-page regression: one closed before publication, one live, and assert the live page never points at a deleted shared file. Apply the same handling to the normal-success and FileAlreadyExists branches.

gouhongshen:
Valid and fixed in 084774a17. The physical file is now owned by one shared publication cohort rather than by any individual page: publication is sealed only after every page has accepted or rejected its path, and the file is deleted exactly once after sealing when the last live page releases it. This also covers a page closing after acceptance but before later siblings are published.

Both normal write success and FileAlreadyExists use the same cohort helper. TestWriteTransferPage_ClosedPagePublication deterministically covers mixed and all-closed two-page batches on both branches; focused lifecycle tests passed with -race -count=100, the original regression passed with -race -count=14, and the full model and TAE DB test packages passed with and without -race. Scoped go vet also passed.

XuPeng-SH (approved):
Approved exact head 084774a178dad2a557bdb91ff9f87949fd2cbcaa after full lifecycle re-review.

The two prior blockers are closed. Path publication and terminal cleanup are linearized per page, while the shared physical file is owned by one sealed cohort: publication cannot delete the file mid-batch, partial closure preserves live siblings, and the last live page triggers exactly one deletion after sealing. File-service I/O stays outside page/cohort locks. Q1-Q3 review found no remaining leak, wait-cycle, or unbounded-growth defect.

Local validation on the exact head:
- all four focused lifecycle tests passed under race independently at count=100;
- full pkg/vm/engine/tae/model passed under race;
- the original TestClearPersistTransferTable regression passed under race;
- scoped go vet passed.

The added mutexes are confined to path publication, disk fallback, and terminal cleanup; the in-memory Transfer hot path is unchanged. UT/SCA CI were still running when this review was submitted, so merge remains gated on CI.

Issue #26459, opened 2026-07-30
[Design]: replace estimator-driven HashBuild admission with allocation-accounted memory ownership


aptend:
## Context

PR #25837 introduced finite hierarchical HashBuild admission and bounded spill to stop the real OOM in #25782. That safety direction is necessary, but the implementation still uses operation-level estimates as hard admission facts. Subsequent regressions show that this model is not converging: each new data shape exposes another mismatch between a logical estimate and the allocator's physical behavior.

This issue is the umbrella for replacing estimator-driven hard rejection with one allocation-accounted ownership model. It is not a request to raise caps, exempt workloads, or add another type/function-specific multiplier.

## Confirmed history

| Incident | Symptom | Confirmed mismatch | Status |
| --- | --- | --- | --- |
| #25782 / PR #25837 | HashBuild bypassed spill and caused node OOM | hash table growth was not admitted at allocation boundaries | initial bounded-budget system merged |
| #26174 / PR #26178 | fulltext INSERT requested 18.72 GB with about 1.29 GB used | current ingress, retained tail, const materialization, and future drain were conflated | fixed locally |
| #26192 / PRs #26231 and #26318 | LOAD DATA rejected | runtime-filter 3x and whole-payload multipliers double-counted retained owners | fixed locally |
| #26413 / PR #26438 | Hive external-table self-join rejected | 50K-row ingress was estimated differently from CopyIntoBatches' real 8192-row segmentation and allocator rounding | fixed locally and real SQL verified |
| #26454 | string-expression join rejected immediately | TEXT maximum 64 MiB was multiplied by 8192 rows and recursively summed with CAST/CONCAT nodes, producing exactly 551,368,048,640 bytes while used memory was about 1.48 GB and observed peak was about 6-7 GB | open |
| #26186 | spill transition could under-account ingress overlap | isolated estimates missed a simultaneously-live ownership state | closed follow-up |
| #26433 / PR #26455 | retained ExpressionExecutor capacity could be under-counted or double-charged across reuse | reservation lifetime did not follow physical retained capacity lifetime | in progress |

Related spill ownership work: #26408 and PR #26412.

The false rejections and under-accounting are opposite outcomes of the same design problem. Making an estimate more conservative prevents one OOM path but creates false rejection; relaxing it restores workloads but can lose a real allocation peak.

## Root design problem

The current implementation mixes three separate concerns:

1. **Ownership ledger**: which live allocation is charged, who owns it, and when it is released.
[...]

Issue #26454, opened 2026-07-30
[Bug]: hash build budget admission rejected


heni02:
## Is there an existing issue for the same bug?

- [x] I have checked the existing issues.
  - Related false-rejection fixes: #26178, #26231 — this case still rejects after those.
  - Not the same as shuffle panic #25930 / window split #25813 / hashjoin nil #25802.

## Branch Name

```
main
```

## Commit ID

```
2e6877406157470e453245ef287d35ff81fc0c84
```

Image: `registry.cn-shanghai.aliyuncs.com/matrixorigin/matrixone:nightly-2e687740`

## Other Environment Information

```
- Hardware: single-node MO (docker, host port 16001), rootless docker on Linux
- Host memory: ~251 GiB; container Memory limit unset (0)
- OS: Linux
- Dataset: jinpan_001 (local LOAD DATA INFILE)
- Suite: cases/jinpan_001/join (1000 SQLs), template srm_contract_attach
- processLimitationSize default: 42949672960 (40 GiB) from pkg/config/configuration.go
```

## Actual Behavior

All executed variants of template `srm_contract_attach` fail immediately (`cost_ms=0`) with:

```
(20101) internal error: hash build budget admission rejected
```

Operator diagnostic (CN log `hashbuild/types.go` summary), repeated almost identically every hit:

```
hash_build budget admission rejected:
  requested = 551368048640   (~513 GiB / ~551 GB)
  used      ≈ 1589595280     (~1.48 GiB)
  cap       = 42949672960    (40 GiB = default processLimitationSize)

extra_stats:
  QueryHashBudgetCapBytes   = 42949672960
  QueryHashBudgetPeakBytes  ≈ 6–7e9
  QueryHashBudgetRejects    = 1
  QueryHashBudgetReserves   ≈ 570
```

Client sees only the short error string; detailed `requested/used/cap` is in CN operator diagnostics.

### Regression evidence

| Build | Template hits | Result |
| --- | --- | --- |
| `nightly-826a3e3d` (before hash-build budget hard path in this form) | 31/31 | **success**, cost_ms ≈ 0.7–1.3s, returns rows |
| `nightly-2e687740` (current) | 13/13 so far (of 31) | **failed**, cost_ms=0, same admission error |

So this is a **behavior regression**: same SQL + same data used to finish in ~1s.

## Expected Behavior

1. Query succeeds (as on `826a3e3d`), returning up to `LIMIT` rows; **or**
[...]

PR #26531, opened 2026-07-31
executor: replace HashBuild estimates with allocation-owned admission

aptend:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [x] Improvement
- [x] Documentation
- [x] Feature
- [x] Test and CI
- [x] Code Refactoring

issue #26459

Replaces estimator-driven HashBuild hard admission with one allocation-owned production path:

- MPool admits the physical capacity it is about to allocate and records immutable account, owner, and site provenance; physical Free is the only release owner.
- Removes the activation gate, estimated-memory reservation, compatibility ledger, and manual partial expression ledger. Every MPool-backed vector owned by a HashBuild/join expression tree now allocates through the same attempt account; borrowed vectors retain source ownership, and opaque library Go heap remains an explicit boundary instead of being estimated.
- Covers retained HashBuild/join storage: hash tables, copied build batches, JoinMap state, retained keys, join state, runtime-filter payloads, Product results, spill encode/decode/rebuild buffers, and HashJoin/LoopJoin/DedupJoin/RightDedupJoin result/finalize batches.
- Result and expression storage is rejected by the same hard capacity contract and is released before prepared execution clears its generation (`Reset -> ClearAllocationAccount`).
- ProductL2 remains one explicit static boundary: producer-owned JoinMap storage stays accounted until Free, while ProductL2 CPU/GPU index and scratch storage stays wholly outside the controlled domain until both native allocators expose the same physical capacity contract.
- Gives every local/remote/retry attempt one account generation, attaches runtime scan/load clones before Prepare, drains transferred objects, clears owners in reverse order, and validates terminal zero exactly once.
- Keeps memory, spill-disk, and spill-FD pressure disjoint; memory capacity can make bounded progress through spill, reclaim, smaller unpublished input, or optional runtime-filter degradation.
- Makes JoinMap, bitmap mailbox, pSpool, and spill-file transfers single-owner and cancellation-safe.
- Preserves GROUPING sentinel semantics through copy/hash/equality/order and Sample late/alternating grouping domains.
- Removes obsolete estimator, expression scratch, compatibility, and duplicate ownership code.

Validation was completed locally after rebasing onto main `49f9e33cae48`; candidate head is `258010da171c`:

- complete affected package matrix: pass;
- selected lifecycle, expression, join-result, transfer, ProductL2 boundary, spill, Sample, and compile packages under `-race -p=2`: pass;
- focused expression/join-result/spill tests under `-race -count=10`: pass;
- `go vet -mod=readonly` and `go build -mod=readonly` for every modified production package: pass;
- `git diff --check` and `gofmt`: clean;
- three independent correctness/ownership/performance review loops: no blocker, major, or minor finding after closing physical-capacity wire identity, fatal-first pressure classification, Packer allocation shape, and prepared `Reset -> ClearAllocationAccount` lifecycle.

Local benchmark evidence shows zero Go allocations for physical admission/release, bounded vector/pSpool reuse, fixed iterator count for mixed GROUPING batches, and spill serialization improving from 67,751 ns/op / 2,904 B/op / 516 allocs/op to 62,709 ns/op / 344 B/op / 4 allocs/op.

Design, implementation plan, local validation, and raw benchmark evidence:

- `docs/rfcs/00000000_allocation_accounted_memory_admission.md`
- `docs/design/allocation_accounted_memory_admission_impl.md`
- `docs/design/evidence/26459_local_validation.md`
- `docs/design/evidence/26459_allocation_accounting_bench.txt`

A distributed TPCH 100G validation run (`30690122244`) exposed one lifecycle boundary bug rather than OOM: Q7 remote fragments sharing a CN also shared a message board, while the first fragment completion closed that board before its siblings finished. The fix gives each physical remote execution a UUID and expected-fragment count per CN; registration and completion are counted, and only the final fragment closes/drains the board and finalizes every allocation account exactly once. A four-fragment lost-final-response test covers the failure edge. Full `pkg/sql/compile` normal/race tests, vector/batch normal/race tests, affected vet, repeated alias decode benchmarks, and an independent correctness re-review all pass after the fix.

### Latest-main spill-recovery integration

Integrated the spill-recovery changes from `d29b4758f8` (PR #26585); the branch is current through `820ee12144` and preserved its recovery-liveness invariant without restoring a second estimator ledger. Each shuffle HashBuild worker now pre-admits an isolated recovery-capacity class; physical MPool allocations borrow that headroom and remain the only exact allocation owner, so recovery capacity is not double-charged. The floor covers expression replacement, partition/materialization, marshal, and serial Packer headroom, and is released at the build terminal boundary. Vector varlena-area proofs and the serial encoded-size/Packer contracts from #26585 are retained.

Post-merge local validation at `0552adba24`: affected and dependent package tests pass; `mpool`, `process`, `hashbuild`, and `hashjoin` pass under `-race`; focused recovery transfer/isolation tests pass under `-race -count=100`; and the forced-shuffle SQL blackbox (`TestHashBuildSharedBudgetRecoverySQL`, 28 MiB cap, DOP 8, `serial_full`, one million rows) spills successfully, returns the exact result, and admits a fresh statement afterward. The retained varlena projection benchmark remains allocation-free (~0.24 µs per 32-column 8K-row batch).


XuPeng-SH (changes requested):
Blocking: retained join result batches bypass the new allocation-account admission.

HashJoin.SetAllocationAccount stores only the account. resetResultBat creates ordinary off-heap vectors, and every later Union/append allocates through the MPool without an AllocationAccountSelection. I reproduced this at the PR head with a 1-byte account: appending 4,096 int64 values to ctr.resBat succeeds, the MPool retains more than 1 byte, and account.Snapshot().Used remains zero.

This is a family-wide closure gap:
- HashJoin: resetResultBat
- LoopJoin: resetResultBat and finalize
- DedupJoin: rbat plus the multi-batch finalize buffers (its selection currently covers capture state only)
- RightDedupJoin: resultBatch

These batches are container-owned and reused/retained across calls, and the design explicitly includes “join matched/capture/result state” in the controlled domain. As written, wide join output can consume retained MPool memory outside the hard query/CN budget, so admission can report room while the operator keeps allocating.

Please give every owned join result/finalize batch a result allocation selection (with cleanup/clear invariants), and add tests proving both capacity rejection and terminal zero for these paths.

pkg/sql/colexec/hashbuild/expression_memory.go
```diff
@@ -15,473 +15,25 @@
...
-	slots     []expressionMemoryLeaseSlot
-	released  bool
-}
-
-// NewBudgetedExpressionExecutors admits the mpool-backed capacity owned by
-// constant children before constructing them. The returned lease adopts those
-// reservations, so construction and later evaluation have one continuous
-// budget lifetime.
-func NewBudgetedExpressionExecutors(
+// NewExpressionExecutors constructs expression trees used by HashBuild and
```
XuPeng-SH, line 32: [blocking] These vectors are not temporary in the ownership sense used by the new invariant. `FunctionExpressionExecutor` keeps `resultVector`, child result vectors, selected-parameter vectors, selected results, and IFF/CASE caches across `Eval` calls and releases their MPool backing only in `Free`.

I reproduced this at this exact head with the existing issue-#26454 `CONCAT(CAST(col), "-", CAST(col))` expression: initialize the spill expression executor under a 1-byte account, evaluate 10,000 rows, and the MPool retains 737,280 additional bytes while both `account.Snapshot().Used` and `generation.Used()` remain zero. The test passes under `-race`; only freeing the executor returns the MPool to baseline. The same constructor is used by normal HashBuild and re-spill, so this is not limited to a short-lived runtime-filter helper.

Charging the later destination copy does not cover the simultaneously-live reusable source cache and allows the real resident set to exceed the hard query/CN cap. It also contradicts the Vector/Expression integration required by #26459 and the retained-capacity lifetime described by #26433. The committed `TestSpillExpressionTemporaryIsOutsideRetainedAccount` currently codifies the bypass instead of testing the safety invariant.

Please propagate an allocation selection to executor-owned MPool vectors/FunctionResult storage. Opaque library Go-heap scratch can remain a separately bounded domain. Add exact-cap/one-byte-short plus reuse/Free/cancellation coverage for normal build, spill, and re-spill.

docs/design/evidence/26459_local_validation.md
```diff
@@ -0,0 +1,95 @@
...
+- optional runtime-filter degradation;
+- grouping-aware copy, hash, equality, and ordering;
+- late and alternating Sample grouping domains across row, percent, and merge
+  modes;
+- Product cleanup and account terminal zero;
+- accounted JoinMap release after an unaccounted ProductL2 consumer frees it.
+
+## Performance evidence
+
+Performance validation is local and allocation-focused. Benchmarks record
```
XuPeng-SH, line 127: [validation blocker] This evidence is not sufficient for a PR that says it fixes #26459. The issue acceptance criteria explicitly require durable workload-level regressions for #26174, #26192, #26413, and #26454; a no-OOM run for the original #25782 high-cardinality case; and TPCH spill/non-spill performance validation.

This PR changes no workload SQL/BVT fixture, and the evidence explicitly excludes end-to-end/remote validation. The resident HashBuild benchmark is also PR-head-only rather than a same-machine base-vs-head A/B. Allocator microbenchmarks can show fixed admission cost and allocation shape, but they do not exercise scheduler DOP, the shared CN cap, operator-result lifetime, spill transition, or disk/FD behavior. They therefore cannot establish either the no-OOM guarantee or absence of query-level regression across this 150-file change.

Before approval, please add reproducible commands and results for the listed real queries plus #25782, and controlled base/head TPCH runs that force both resident and spill modes. The regression should assert the behavioral outcome (no theoretical rejection, no cap escape/OOM, correct rows, terminal zero), not only that package unit tests pass.

pkg/sql/colexec/hashbuild/expression_memory.go
aptend: Fixed in f834cce37c. The HashBuild/join expression constructor now creates one immutable AllocationAccountSelection and propagates it through the entire executor tree: nested FunctionResult storage, fixed/variable values, list results, selected-parameter/result buffers, and IFF/CASE/null caches all allocate through the attempt account. Borrowed input and serialized-plan vectors keep source ownership; only opaque library Go heap remains outside the stated domain. The obsolete manual/estimated expression ledger and its bypass test were removed. Coverage now includes exact-cap/one-byte-short rejection, result-vector transfer/reuse and terminal Free, normal HashBuild, spill evaluation, and function-expression recursive re-spill. Focused cases passed under -race -count=10, and the complete affected package/race matrices pass.

aptend:
Addressed the retained join-result review in f834cce37c. HashJoin, LoopJoin, DedupJoin (including multi-batch finalize), and RightDedupJoin now construct result vectors with dedicated account selections, reject growth at capacity, and release the generation-bound batches during prepared Reset before ClearAllocationAccount. Tests cover positive accounting/reuse, 1-byte capacity rejection, multi-batch finalize, terminal zero, and the production Reset -> Clear lifecycle. An independent cleanup/ownership re-review found no blocker/major and no double-free edge; the full affected package matrix, selected -race -p=2 matrix, focused -race -count=10 tests, vet, and build all pass locally.

fengttt (approved):

aptend:
Final review closure pushed at 258010da17 (rebased onto main 49f9e33cae): physical MPool capacity now retains a dedicated MO error code and real AllocAccounted owner/site context across pipeline wire encoding; lifecycle/invariant failures dominate joined memory admission; runtime-filter serialization uses account-owned fixed Packer storage without growing the Packer object (48 B/op, same as main); and the scaled FLOAT32/GROUPING test proves both initial spill and recursive respill across 8,192 rows. Test-only allocation checkpoint/fault branches were removed. The full affected CGo package matrix, focused race suites, vet, and make static-check pass locally. Three independent final reviews report 0 blocker / 0 major / 0 minor. Re-review was requested through the GitHub reviewer API.

pkg/sql/compile/remote_allocation_statement_group.go
```diff
@@ -0,0 +1,476 @@
...
+		if abort {
+			cancels = activeRemoteAllocationStatementCancelsLocked(group)
+		}
+		if !group.expired && group.registered < group.expected && group.timer == nil {
+			group.timer = time.AfterFunc(
+				remoteAllocationStatementRegistrationTimeout,
+				func() { expireRemoteAllocationStatementGroup(group) },
+			)
+		}
+		complete := group.expired && group.finished == group.registered ||
```
XuPeng-SH, line 336: [blocking] cause != nil sets expired, and this condition treats an incomplete group as terminal as soon as all currently registered fragments finish. With expected=2, if fragment A is the only registered fragment and fails, A immediately finalizes; CloseAndDrain removes remoteExecutionID from MessageCenter and releaseRemoteAllocationStatementGroup deletes the group. Fragment B from the same physical execution may already be in flight but not registered. Its newCompile then creates a fresh MessageBoard for the same execution ID, and acquireRemoteAllocationStatementParticipant creates a second expected=2 group. B can run after the abort, or form a second partial group, and its allocation/MPool is excluded from the first terminal aggregate. The cancellation list contains only registered participants, so it cannot close this race.

Do not finalize/remove an aborted incomplete generation in a way that permits the same (remoteExecutionID, CN) generation to reopen. Keep an execution-keyed tombstone/group until expected arrivals are rejected/drained (or bounded expiry), and add a regression where A fails before B registers, B arrives late with the same ID, and a retry with a different ID remains independent.

XuPeng-SH (changes requested):
I re-reviewed the current head and rechecked all prior comments. The retained expression tree and the HashJoin/LoopJoin/DedupJoin/RightDedupJoin result batches now carry allocation selections and have terminal cleanup; those two previous code blockers are closed.

Two blockers remain:

1. An aborted incomplete remote group can be reopened by a late fragment from the same physical execution (inline). This breaks the single statement-boundary owner and can execute/account the late fragment in a second partial group.
2. The validation blocker is still open. docs/design/evidence/26459_local_validation.md explicitly excludes remote auto-test; the PR adds no durable workload regressions for #26174/#26192/#26413/#26454, and it still lacks the #25782 high-cardinality no-OOM run and a controlled base/head TPCH resident+spill A/B required by #26459. Local allocator/package tests cannot cover distributed fragment lifecycle or query-level no-OOM/performance.

Please retain an aborted-generation tombstone keyed by execution ID and CN so late same-generation fragments cannot recreate a board/group, test that race plus retry isolation, and close the issue acceptance matrix with durable regressions and workload evidence.

gouhongshen (changes requested):
## Codex automated review

The incremental selection-coalescing change is consistent, but a prior blocking bitmap regression remains at 8ce077f8. Existing remote-lifecycle and validation blockers are omitted because they are already raised; expression and join-result blockers were addressed.

### P1 - Ordered remap panics on sparse accounted bitmaps (pkg/common/bitmap/bitmap.go:600)
Accounted vectors grow null/grouping storage only when an appended row carries that bit. A valid vector with NULL at row 0 followed by 129 non-NULL rows therefore has one bitmap word despite length 130. Duplication preserves bitmap length-based capacity, then OFFSET calls Batch.Shrink; FilterInPlaceOrdered dispatches here for the 129-row selection, requiring three words and panicking because cap is one. Ensure remapping admits/grows output bitmap capacity and add sparse-capacity regression coverage.

pkg/sql/compile/remote_allocation_statement_group.go
aptend: Fixed in 2aefd528ea. Remote allocation groups are now indexed by the physical `(remoteExecutionID, CN)` key as well as by MessageBoard. If an incomplete group aborts, a bounded key-only tombstone is installed while holding the registry lock before the active group is released, so a late fragment carrying a newly created board cannot reopen the same generation. A distinct retry execution ID remains independent. `TestRemoteAllocationStatementGroupRejectsLateFragmentAfterAbort` covers A failing before B registers, late B rejection with the same key/new board, and successful retry with a different key; the focused test passes under `-race -count=100`, and the full `pkg/sql/compile` suite passes under `-race`.

docs/design/evidence/26459_local_validation.md
aptend: TKE workload validation is currently running in https://github.com/matrixorigin/mo-auto-test/actions/runs/30753023018. We are still resolving the observed TPCH performance regression, so the durable base/head resident+spill evidence and issue acceptance matrix are not being claimed complete yet; they will be updated after the performance path is stabilized and the TKE runs finish.

aptend:
Addressed the sparse accounted-bitmap P1 in 1129ec8a27. `Vector.Dup` now sizes every non-empty external null/grouping bitmap for the vector’s complete row domain instead of the highest set bit. The regression constructs row 0 with the bit plus 129 rows without it, duplicates the vector, and performs the ordered 129-row shrink without panic; both null and grouping dimensions are covered. The focused regression passes at `-count=100`, and the full `pkg/container/vector` suite plus vet pass locally.

gouhongshen (changes requested):
## Codex automated review

The earlier expression-tree and join-result blockers are fixed. The remote late-fragment blocker is resolved by the execution/CN-keyed tombstone and is withdrawn. The original short sparse-bitmap case is fixed, but its replacement sizing rule introduces a distinct blocking panic for valid stale-long bitmaps. The separate workload-validation blocker remains open per the author and is omitted because another reviewer already raised it. Focused tests could not run because this worktree lacks cgo/libmo.dylib.

### P1 - Preserve stale bitmap extent when duplicating accounted vectors (pkg/container/vector/vector.go:1677)
Author response: The author states that `Dup` now admits every non-empty external bitmap for the complete vector row domain, with coverage for `bitmap.Len() < v.length`. That reported sparse case is fixed. Why this remains blocking: the inverse state is also valid—`SetLength` changes only `v.length`, and the existing `TestUnionBatchFastPathStaleBitmapBits` explicitly documents bits beyond the shortened length as normal reuse state. This change removes the previous bitmap-length sizing and admits only `v.length`. For an accounted vector preextended to N rows, with a null/grouping bit near N-1 and then `SetLength(1)`, `Dup` allocates one-row external bitmap capacity before `Bitmap.InitWith` copies the N-row source; `InitWith` panics when the source data exceeds that capacity. This is a distinct stale-extent root cause, not a repeat of the short sparse case. Size for the checked maximum of vector length and bitmap length, or deliberately trim the copied bitmap to the row domain, and cover stale null/grouping extents—including the constant-grouping branch.

XuPeng-SH (changes requested):
Request changes.

The earlier expression-tree, retained join-result, and remote late-fragment ownership blockers have responses, but this head is still not ready to approve.

1. Vector.Dup must preserve valid stale bitmap extent. A vector may be preextended with a null/grouping bit near the old high watermark and then shortened with SetLength(1). If Dup allocates external bitmap capacity from only the current vector length, Bitmap.InitWith can copy the longer valid source bitmap into too-small storage and panic. Admit the checked maximum of the vector row domain and each source bitmap extent, or explicitly trim semantics before copying, for both null and grouping, including the constant-grouping path. Add the stale-long null/grouping regression; the sparse-short case alone is insufficient.

2. Workload acceptance is incomplete. The PR itself says the controlled base/head TPCH resident+spill A/B evidence and issue acceptance matrix are not yet complete while a TPCH regression is being resolved. This change replaces a production memory-admission path across 205 files; package/race tests do not establish distributed no-OOM, spill correctness, or the non-spill performance contract. Please add durable regressions/workload evidence and update the stated acceptance matrix before approval.

aptend:
Addressed the stale-long bitmap blocker in d13b9103c8. `Vector.Dup` now reserves external null/grouping bitmap storage for `max(vector row domain, source bitmap extent)` before `InitWith`, including constant grouping and zero-count stale extents. The sparse-short, stale-long flat/constant, and empty-stale regressions pass; both focused tests pass under `-race -count=100`, and the full `pkg/container/vector` package passes under `-race`, build, and vet. The completed TPCH100G run averaged 95.029s across four measured turns versus 98.252s on the cited recent main run (3.3% faster), with all Q1-Q22 correct and no OOM/query failure. Fresh head validation for TPCH100G + TPCH1T is running at https://github.com/matrixorigin/mo-auto-test/actions/runs/30758186183. Re-review has been requested through the GitHub review-request API.

gouhongshen (approved):
## Codex automated review

The d13b9103c delta fixes the prior stale-long Vector.Dup panic by admitting max(vector length, source bitmap extent) for null/grouping storage, including constant and empty-stale cases; that blocker is withdrawn. The sparse-short bitmap issue remains fixed. Earlier expression-tree and join-result blockers are fixed, and the remote late-fragment blocker remains withdrawn. The workload-validation blocker remains unresolved in another reviewer’s thread and is not duplicated here. Focused tests/build/vet could not run because local CGo artifacts and headers are absent; formatting and diff checks pass.

pkg/common/mpool/accounted_buffer.go
```diff
@@ -0,0 +1,242 @@
...
+			b.site,
+		)
+		if err != nil {
+			return err
+		}
+		b.data = data[:oldLength]
+		return nil
+	}
+
+	data, err := b.mp.Grow(b.data, int(capacity), true)
```
XuPeng-SH: [P1/blocking] `capacity` has already been computed by `GrowCapacity(cap(b.data), required)`, but `MPool.Grow` unconditionally calls `reAllocWithDetailK(..., bufferMore=true)`, which runs `GrowCapacity` again using this already-grown value as the new requirement. This inflates both the replacement allocation and its admission peak. A focused counterexample with old cap 10240 and required 10241 gives a correct next cap of 16384, so a 26624-byte account exactly covers old+replacement; this path instead requests a 24576-byte replacement and fails with `used=10240 requested=24576 limit=26624`. The same double-growth shape is present at `pkg/common/hashmap/strhashmap.go:220`. Pass the original required size to `Grow` or use an exact-cap growth API, and add exact-cap/one-byte-short regressions for both callers.

pkg/sql/compile/remote_allocation_statement_group.go
```diff
@@ -0,0 +1,529 @@
...
+// completed. Bound the orphan lifetime by the same five-minute interval used
+// by blocking MessageBoard receives. The timer is canceled as soon as every
+// expected fragment has registered, so it never limits a fully dispatched
+// statement's execution time.
+var remoteAllocationStatementRegistrationTimeout = 5 * time.Minute
+
+// A late RPC can carry a new MessageBoard, so keep a bounded record of an
+// incomplete execution after its active group has been released. The record
+// contains no statement resources.
+var remoteAllocationStatementTombstoneTimeout = 5 * time.Minute
```
XuPeng-SH: [P1/blocking] Five minutes is not a valid upper bound for late delivery: `newMessageSenderOnClient` gives a stream without a caller deadline the 24-hour `MaxRpcTime`. When this timer deletes the key, `acquireRemoteAllocationStatementParticipant` has no remaining generation state and accepts the same `(remoteExecutionID, CN)` with a fresh `MessageBoard`, recreating the aborted group. I reproduced that exact transition by firing the tombstone cleanup and retrying the same key; registration returned nil. The committed test only retries while the tombstone is still live. Preserve the no-reopen invariant for the full possible in-flight delivery lifetime, or use a generation registry/acknowledgement that proves no old fragment can arrive, and test expiry plus same-key rejection and different-key retry isolation.

XuPeng-SH (changes requested):
Re-reviewed the complete PR at `d13b9103c8` and read all existing discussions and author responses. The retained expression-tree and join-result ownership blockers are fixed. The sparse-short and stale-long `Vector.Dup` panics are also fixed at this head; their focused tests pass locally.

Three blockers remain:

1. Accounted buffer growth applies `GrowCapacity` twice, producing false hard-cap rejections (inline). The same shape exists in the accounted string-hash iterator.
2. The remote late-fragment fix only rejects the old generation for five minutes (inline). A pipeline RPC can remain alive for the 24-hour `MaxRpcTime`, so the same execution can still reopen after tombstone expiry.
3. Workload acceptance remains incomplete. The latest TPCH100G+1T run is still in progress, with the 1T step and result collection unfinished. The committed evidence predates the final semantic fixes, and the PR still has no durable workload-level acceptance matrix for #26174/#26192/#26413/#26454 plus the original #25782 high-cardinality case. Package tests cannot establish distributed no-OOM, terminal-zero, spill correctness, or the base/head resident+spill performance contract.

Local validation: `git diff --check`, build, vet, and the complete tests for `mpool`, `hashmap`, `vector`, `hashbuild`, `spillutil`, and `compile` pass both normally and under `-race`; the focused bitmap/account/remote tests also pass under `-race -count=100`. Two counterexample regressions fail on this head: the exact single-replacement admission boundary is rejected as `used=10240 requested=24576 limit=26624`, and deleting the expired tombstone allows the same execution key to register successfully. Please fix both invariants and close the workload acceptance evidence before approval.

pkg/common/mpool/accounted_buffer.go
aptend: Fixed in 078d05ae61. Existing accounted buffers now pass the original logical requirement to MPool.Grow, so GrowCapacity is applied exactly once by the allocator. The accounted string-hash iterator uses the same rule. Exact old+replacement and one-byte-short regressions cover old cap 10240 / required 10241 for both callers, including rollback, peak, terminal zero, and live-metadata zero. Both owning package suites pass normally and under -race; the focused regressions pass under -race -count=100.

pkg/sql/compile/remote_allocation_statement_group.go
aptend: Fixed in 078d05ae61. The aborted-generation key-only tombstone now lives for MaxRpcTime (24h), which is the maximum lifetime of a pipeline sender without a caller deadline; it cannot expire while an old fragment may still arrive. The expiry regression deterministically expires an incomplete registration, finalizes it, rejects the same execution/CN key with a new MessageBoard, and accepts a distinct retry key. The compile suite passes normally and under -race; the two late-fragment regressions pass under -race -count=100.

pkg/sql/colexec/hashbuild/expression_memory.go
aptend: This blocker was fixed in f834cce37c and the later re-review explicitly confirmed it closed. The complete expression tree now uses the attempt allocation selection, with exact-cap, one-byte-short, reuse/Free, spill, and recursive re-spill coverage. Resolving this superseded thread.

pkg/sql/compile/remote_allocation_statement_group.go
aptend: The original no-reopen blocker was fixed in 2aefd528ea; 078d05ae61 closes the follow-up lifetime gap by retaining the key-only tombstone for MaxRpcTime and adds deterministic expiry/same-key rejection coverage. Normal/race compile suites and focused -race -count=100 tests pass. Resolving this superseded thread.

docs/design/evidence/26459_local_validation.md
aptend: Updated the committed evidence in 078d05ae61. Current semantic head d13b9103c8 completed TKE TPCH100G+1T run 30758186183 with all Q1-Q22 correct and no query failure: 100G averaged 96.298s versus cited main 98.252s (-1.99%); 1T averaged 1042.472s versus main 1027.439s (+1.46%). The document now includes the incident-by-incident durable-regression/workload matrix and explicitly identifies the external-data/current-head gaps rather than claiming them complete. I am leaving this thread open because the original private #25782 harness and unavailable external datasets were not rerun at the final head.

XuPeng-SH (changes requested):
Blocking issue on the current head: optools/images/Dockerfile changes the native stage from make -C thirdparties to init usearch xxhash croaring, so it no longer produces the platform ONNX Runtime library. This stage is not non-ONNX-only: .github/workflows/native-dependency-check.yml builds target native from this Dockerfile and explicitly verifies thirdparties/install/lib/onnxruntime.so on amd64 and onnxruntime_arm64.so on arm64. Both Native build jobs on run 30795881950 fail at that exact test. Please preserve the native-stage output contract, or update every consumer and its validation coherently; skipping the required artifact in this shared stage leaves the branch image incomplete.

aptend:
The temporary ONNX-runtime skip used only for remote acceptance validation has been moved off this PR to `aptend/matrixone:validation/26459-skip-onnx` (commit `95baebc538e0889995f8d8560d8fd3043234b648`).

It is reverted from this PR by `57811ed7fa7cc5ed9234f92a9167c8187abaacdf`, so the shared native-stage dependency contract (including ONNX Runtime) is restored here. The new acceptance runs are being triggered exclusively against the validation branch.

XuPeng-SH (approved):
Re-reviewed exact head `57811ed7fa7c`. The new commit exactly reverts the Dockerfile change that omitted ONNX Runtime; its tree is byte-identical to reviewed head `cd36bbbe776e`, where the HashBuild allocation-owned admission, remote tombstone lifetime, spill-recovery reservation, ownership/lifecycle paths, and focused package tests were already checked. The shared native stage now again satisfies `native-dependency-check.yml`. No remaining code blocker found. CI for this head is still running.

docs/design/evidence/26459_local_validation.md
aptend: Closing this thread based on the maintainer validation decision. The remaining external-data and original high-cardinality workload matrix is too expensive to require as a blocking pre-merge gate. The committed local/race/property coverage and completed TPCH100G/TPCH1T evidence remain the pre-merge evidence; the remaining workload coverage will run through the existing daily/nightly regression flows, whose lag is acceptable for this change. Any regression found there will be tracked and fixed as a follow-up rather than holding this PR open.

iamlinjunhong (changes requested):
[P1] 24 小时 tombstone 注册表缺少硬空间上限
不完整的远端执行结束时，[installRemoteAllocationStatementTombstoneLocked](https://github.com/matrixorigin/matrixone/blob/42ac0d4dbe541279b73d48bb652820091eb1f681/pkg/sql/compile/remote_allocation_statement_group.go#L414-L429) 会把唯一的 executionID@CN key 和一个 timer 放入进程级全局 map：
remoteAllocationStatementGroups.tombstones[key] = tombstone
tombstone.timer = time.AfterFunc(
    remoteAllocationStatementTombstoneTimeout,
    ...
)
该 tombstone 固定保留 [MaxRpcTime = 24h](https://github.com/matrixorigin/matrixone/blob/42ac0d4dbe541279b73d48bb652820091eb1f681/pkg/sql/compile/remoterunClient.go#L45-L47)，但注册表没有数量、字节或 admission 上限。
触发路径是：
远端计划预计多个 fragment；
dispatch 只成功一部分，或某个 fragment 提前失败；
group 被终止并安装 tombstone；
每次重试使用新的 execution ID，旧条目无法复用；
在网络异常或持续 partial-dispatch 下，CN 保留约 失败速率 × 24h 个 map entry、字符串和 runtime timer。

aptend:
Addressed the tombstone-registry P1 in bf244d6b5c.

The fix reserves bounded generation capacity at first remote-group admission: len(byKey) + len(tombstones) < 4096. Existing fragments of an admitted generation reuse that reservation. A partial dispatch can therefore transfer its already-reserved generation into a 24-hour tombstone, but retries cannot create more retained keys or timers than the hard limit. At capacity, only new generations fail closed; no tombstone is evicted, so a late fragment cannot reopen an aborted execution.

The regression fills the registry to the exact boundary, proves an existing sibling is still admitted, a new execution is rejected, and the active partial execution transitions to the 4096th tombstone. Focused tests passed under race with count 100; the full pkg/sql/compile package passed under race, and vet passed.

iamlinjunhong (approved):

PR #26532, opened 2026-07-31
fix(fileservice): honor context while waiting for object storage slots

### LINKED ISSUE #26273: [Bug] Object storage write waits after its context is canceled
## Summary

`objectStorageSemaphore` does not observe the caller context while waiting for a slot in `Write`.

`CopyObject` and `Delete` use `acquireContext`, but `Write` uses the unconditional `acquire` path. The same waiting pattern is also used by `Exists`, `Read`, `Stat`, and `WriteMultipartParallel`.

## Reproduction

On `main` commit `c8e443bf8610b2af4333ad943ba2075485b16cf8`:

1. Create an `objectStorageSemaphore` with capacity 1 and occupy its only slot.
2. Create and cancel a context.
3. Call `Write` with that canceled context.

The call remains waiting for the semaphore slot instead of returning `context.Canceled`. The isolated test repeated this result 3/3 times with a 100 ms upper bound per run.

## Expected behavior

Once the caller context has ended, a queued file operation should return that context result without waiting for another operation to release the slot, and it should not call the upstream storage.

## Actual behavior

The queued `Write` call remains blocked until the slot is released. Therefore an operation that has already been canceled can still wait behind a long-running storage operation.

## Suggested direction

Use the context-aware acquisition path consistently for operations that accept a context, while retaining the existing release behavior after a slot is acquired.

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26273

`objectStorageSemaphore` limited concurrency with a blocking channel send. Although every object-storage operation receives a caller context, `Exists`, `List`, `Read`, `Stat`, `Write`, and `WriteMultipartParallel` waited for a semaphore slot through the unconditional `acquire` path. Once all slots were occupied, cancellation could not terminate that local wait, so an already-canceled operation remained blocked until another storage operation released a slot and could then reach the upstream storage.

This change:

- uses the existing context-aware acquisition path for every context-bearing object-storage operation;
- returns the caller context error without invoking upstream storage when cancellation wins while queued;
- propagates acquisition failure through the lazy `List` iterator;
- preserves the existing release ownership after successful acquisition, including `Read` holding its slot until EOF, read error, or `Close`;
- removes the now-unused unconditional acquisition helper.

The regression test occupies the only slot, cancels the caller context, and verifies all six previously affected operations return `context.Canceled` without acquiring the semaphore. The same test fails deterministically on the pre-fix `main` implementation because each operation remains blocked until the test releases the slot.

Validation:

- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=100 -timeout=120s -run '^TestObjectStorageSemaphoreOperationsObserveContextWhileWaiting$' ./pkg/fileservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=300s ./pkg/fileservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s ./pkg/fileservice`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -timeout=300s -coverprofile=/tmp/issue26273.cover ./pkg/fileservice`
- `go build -mod=readonly ./pkg/fileservice`
- `go vet -mod=readonly ./pkg/fileservice`

All newly added context-acquisition branches are covered by the regression test.


XuPeng-SH (approved):
Deep review passed.

The change closes the cancellation contract consistently across every context-bearing object-storage operation that still used the unconditional semaphore path. Acquisition failure owns no slot; every successful acquisition retains exactly one release path, including Read's existing EOF/error/Close once-release contract and List's lazy iterator lifetime. Cancellation now terminates the local queue wait independently of the operation holding the slot. No unbounded resource, API/format compatibility, or material performance concern was found.

Validation on head 16b91d80b408050a7687fd3ed3002e64c8625e52: focused semaphore tests passed under -race -count=100, the full pkg/fileservice package passed under -race, and build/vet passed.

PR #26533, opened 2026-07-31
fix(compile): prevent distributed delete dispatch hang

### LINKED ISSUE #26526: [Bug] PITR Backup Restore Test checkpoint dump restore 挂起 12 小时无输出
## 环境

| Field | Value |
|-------|-------|
| 分支 | 4.2-dev |
| Run | https://github.com/matrixorigin/mo-nightly-regression/actions/runs/30579495362/job/91056897644 |
| Job | CHECKPOINT DUMP S3 WITH PITR DATA |
| Commit | `6f85cb6d3c` |

## 问题

`CHECKPOINT DUMP S3 WITH PITR DATA` job 在 restore 阶段挂起 **11 小时 44 分钟**，无任何输出，最终手动 cancel。

## 时间线

```
02:26:05  verify_s3_checkpoint_dump_restore.sh 开始
02:26:12  Wait and list checkpoint databases
02:26:32  List checkpoint tables（38 张表，含 100M/1B 行大表）
02:26:46  Dump whole database to S3
03:10:09  Dump 完成（~43 分钟，含 42GB 数据）
03:10:09  Download and rewrite restore.sql from S3
03:10:10  Restore dump into target account ckps3pitr:test_account 开始
          ↓↓↓ 挂起 11 小时 44 分钟，无任何输出 ↓↓↓
14:54:03  手动 cancel
```

## 关键发现

1. **restore_target.log 为 0 字节** — restore 命令启动后完全没有输出
2. **Dump 阶段正常** — 38 张表（含 `big_data_test` 的 100M/1B 行表，总计 42GB）成功导出到 S3
3. **Orphan processes**: cancel 时清理到 `pid 387 (bash)` 和 `pid 681 (mysql)`，说明 mysql 客户端仍在运行
4. **挂起原因**: 可能是 `restore.sql` 中的 `LOAD DATA` 语句从 S3 加载大表数据时 MO 服务端 hang 住

## 附件

- Artifact: `checkpoint-dump-s3-pitr-result` (ID: 8797229815)
- 包含 databases.txt、tables.txt、restore_target.log（空）、database_dump.log

XuPeng-SH:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26526

### Root cause

`newDeleteMergeScope` created one receiver scope per delete bucket and attached only source `i` beneath receiver `i`. However, every source `Dispatch.LocalRegs` targeted all same-CN delete receivers. Each receiver tree therefore owned only one of the local channels used by its source dispatch and was not independently executable on its planned CN.

`RemoteRun` silently converted such a remote tree to local `MergeRun` on the coordinator. The dispatch UUID was then registered and executed on the coordinator while `RemoteReceivRegInfo.FromAddr` still named the planned worker CN. Receivers notified a CN that would never publish that UUID and waited indefinitely in `GetProcByUuid`.

The generated topology reproduces the incident exactly: two distributed-delete steps, three CNs with DOP 16, 64 misplaced remote source dispatches, 32 remote UUID routes per source, 2048 coordinator-side waits, and 1024 lookups on each remote CN.

### Incident correlation (sanitized)

The request-changes review correctly noted that the original issue discussion did not publish enough evidence to connect the nightly hang to this path. Rechecking the real CI Loki data and archived raw goroutine snapshots closes that gap:

- At `03:10:10.363181 UTC`, the target restore connection completed authentication and received the OK packet. At `03:10:10.567937`, the planner logged the sanitized query shape `LOAD DATA URL (...) ... PARALLEL=true`.
- At `03:10:10.590517`–`03:10:10.601298`, statement `019fb626-984b-7eda-bb4f-aac9e7d65773`, trace `c8489d33-8079-91e3-8f0b-56321c81ac0d` emitted 128 `remoterunClient.go:155` fallback messages: 64 unique pipeline pointers, each checked twice, explicitly converted to local execution because its dispatch referenced another local pipeline tree.
- In the `03:13:18` coordinator snapshot, the authenticated frontend goroutine is inside `executeStatusStmt -> Compile.Run -> runOnce`. The same compile execution contains 2 `MergeDelete.Call` stacks and 96 `Deletion.remoteDelete` stacks; every remote-delete stack passes through `Scope.RemoteRun -> Scope.MergeRun`. All 64 fallback pipeline pointers from Loki match those raw remote-delete stacks. The remaining 32 are the coordinator-local sources: `2 steps * 3 CNs * DOP 16 = 96`, of which `2 steps * 2 remote CNs * DOP 16 = 64` were misplaced.
- The wait-for graph is exact and stable from the `03:13` snapshots to the `14:50` snapshots: 2048 coordinator `sendNotifyMessageWithFactoryAndWait` waits and 1024 `GetProcByUuid` waits on each of the two remote CNs. The same stacks remained blocked for about 700 minutes.
- The earlier "zero Exec.Run" observation does not mean no statement executed. `status_stmt.go` calls `runner.Run(0)` first and emits `time of Exec.Run` only after it returns; the archived frontend stack shows that call still blocked. The cited salt timeout belongs to a different connection, was already occurring before restore started, and does not appear on the target restore session.

This directly ties the reported restore hang to the distributed `compileDelete -> newDeleteMergeScope -> RemoteRun -> MergeRun` fallback. A literal `DELETE` statement is not required: the physical `Deletion` and `MergeDelete` operators are present in the `LOAD DATA` execution stacks.

### Fix

- Group distributed-delete receiver buckets into one `RemoteRun` envelope per execution CN. Every envelope now owns all same-CN `LocalRegs` referenced by its source dispatches while preserving all delete buckets, UUIDs, and remote routes.
- Validate standalone execution against the complete operator tree that will actually be encoded, including dispatches or connectors wrapped by DML operators.
- Reject any remaining non-standalone remote tree before RPC or operator start, cancel the query context, and do not pre-register phantom dispatch owners on the coordinator.
- Update the `RemoteRun` guide to document the fail-fast contract.

The success path fixes the compiler topology. The runtime checks are a liveness backstop so a future invalid topology returns an error instead of changing placement and hanging.

### Validation

Based on latest `origin/4.2-dev` at `890270f4e6`.

- Focused topology and fail-fast regressions, `count=50`
- Focused race regressions, `count=3`
- Full `pkg/sql/compile` test package
- Full `pkg/sql/compile` test package with race detector
- `go build -mod=readonly ./pkg/sql/compile`
- `go vet -mod=readonly ./pkg/sql/compile`
- `git diff --check`

Regression coverage verifies per-CN standalone execution for multi-CN and single-remote-CN DOP, all `IBucket`/`Nbucket` values, UUID-to-receiver route bijection, the observed 2048/1024/1024 topology, pre-registration ownership, nested operators, direct fail-fast cancellation, and bounded `MergeRun` sibling shutdown.

### DIFF
--- a/pkg/sql/compile/Remote_Run_DEV_GUIDE.md
+++ b/pkg/sql/compile/Remote_Run_DEV_GUIDE.md
@@ -49,7 +49,8 @@ func (s *Scope) RemoteRun(c *Compile) error {
     
     // Check if pipeline can be executed standalone at remote
     if !checkPipelineStandaloneExecutableAtRemote(s) {
-        return s.MergeRun(c)  // Fallback to local execution
+        return s.failRemoteRunBeforeStart(c,
+            moerr.NewInternalErrorNoCtx("remote pipeline is not standalone executable"))
     }
     
     // Create pipeline and send to remote node
@@ -76,7 +77,9 @@ func (s *Scope) RemoteRun(c *Compile) error {
 - Check if same address (execute locally via `MergeRun`)
 - Check if any operator cannot be executed remotely via `holdAnyCannotRemoteOperator()`
 - Check if pipeline can be executed standalone at remote via `checkPipelineStandaloneExecutableAtRemote`
-- If not standalone executable, fallback to local execution (`MergeRun`)
+- If not standalone executable, fail before start, cancel sibling pipelines, and clean up
+- Never silently execute a remote-selected tree on the coordinator: dispatch/connector
+  receiver locality is part of the compiled topology, and changing it can hang the query
 - Create pipeline and call `remoteRun` to send
 - Use `CleanRootOperator` for cleanup (not general cleanup)
 
--- a/pkg/sql/compile/compile.go
+++ b/pkg/sql/compile/compile.go
@@ -5170,6 +5170,13 @@ func (c *Compile) newDeleteMergeScope(arg *deletion.Deletion, ss []*Scope, node 
 		arg.Nbucket = uint32(len(rs))
 		rs[i].setRootOperator(dupOperator(arg, 0, len(rs)))
 	}
+	// Every source dispatch targets all delete receivers on its CN through
+	// LocalRegs. Put those receivers in one RemoteRun tree so the encoded tree
+	// owns every local channel its dispatches reference.
+	stageNodes := shuffleBucketStageNodes(rs)
+	if len(rs) > len(stageNodes) {
+		rs = c.mergeScopesByStageNodes(rs, stageNodes)
+	}
 	return c.newMergeScope(rs)
 }
 
@@ -5423,10 +5430,8 @@ func (c *Compile) mergeShuffleScopesIfNeeded(ss []*Scope, force bool) []*Scope {
 // dispatch operator that sends to remote (cross-CN) receivers. A non-empty RemoteRegs is
 // the signature of a cross-CN shuffle dispatch (see constructDispatchLocalAndRemote).
 //
-// Precondition: it only inspects each scope's RootOp, assuming a shuffle dispatch is always
-// the root operator of its scope (constructDispatch results are attached via setRootOperator
-// with IsEnd=true, so nothing is appended on top of them). A dispatch nested as a child of
-// another operator would be missed -- which does not happen for shuffle dispatches today.
+// It inspects every operator in every scope because DML operators can wrap a shuffle dispatch
+// before this gate decides whether the receiver buckets need a per-CN transport envelope.
 //
 // Scope of the check (intentionally narrower than checkPipelineStandaloneExecutableAtRemote,
 // which rejects out-of-tree dispatch *and* out-of-tree connector targets): this gate only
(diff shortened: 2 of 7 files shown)


LeftHandCold (dismissed):
Requesting changes on issue targeting; I did not find a code-side blocker in the topology or lifecycle changes.

The published evidence for #26526 does not currently establish that this code path was reached:

- The issue investigation says the target account never authenticated and all three CNs had zero Exec.Run during 03:10–14:54: https://github.com/matrixorigin/matrixone/issues/26526#issuecomment-5144452191
- The exact nightly restore script starts a mysql client after CREATE ACCOUNT, and its generated restore contains CREATE DATABASE, CREATE TABLE, and LOAD DATA statements, not a DELETE workload: https://github.com/matrixorigin/mo-nightly-regression/blob/6f85cb6d3cc7b35377522f4ba7f05d839c4e39d0/script/checkpoint_dump/verify_s3_checkpoint_dump_restore.sh#L827-L840
- This patch changes the large plan.Node_DELETE path in newDeleteMergeScope. If the published authentication and zero-execution evidence is correct, that path was unreachable in the reported incident.

The structural regression does demonstrate a real latent distributed-delete topology defect, but it does not yet connect that defect to #26526. Merging this as Fixes #26526 would close the reported Proxy/authentication incident without showing it is fixed. Please either link/create the correct issue for this distributed-delete defect and stop auto-closing #26526, or add concrete evidence from that run (affected SQL/statement or trace ID, query plan, and/or goroutine stack) proving the two distributed-delete steps ran, and reconcile it with the existing zero-Exec/authentication diagnosis.

For the code itself, I checked bucket/UUID/FromAddr preservation, per-CN ownership, full encoded-tree validation, pre-registration, cancellation/cleanup, bounded waits, and compile-time cost. Seventeen focused topology, encoding, registration, and unhappy-path tests pass at head 93bd29cb.

XuPeng-SH:
Thanks — this was a valid evidence gap in the original issue discussion. I re-queried the real CI Loki data and correlated it with the archived raw goroutine snapshots. I kept `Fixes #26526` and added a sanitized, reproducible incident-correlation section to the PR body because the new evidence directly reaches this code path:

- `03:10:10.363181 UTC`: the target restore connection completed authentication and received OK.
- `03:10:10.567937`: the planner logged the sanitized shape `LOAD DATA URL (...) ... PARALLEL=true`.
- `03:10:10.590517`–`.601298`: statement `019fb626-984b-7eda-bb4f-aac9e7d65773`, trace `c8489d33-8079-91e3-8f0b-56321c81ac0d` emitted 128 fallback logs: 64 unique pipeline pointers, each checked twice, explicitly converted locally because its dispatch referenced another local pipeline tree.
- The `03:13:18` raw coordinator dump shows the authenticated frontend blocked in `executeStatusStmt -> Compile.Run -> runOnce`, plus 2 `MergeDelete.Call` and 96 `Deletion.remoteDelete` stacks in the same compile execution. Every remote-delete stack contains `Scope.RemoteRun -> Scope.MergeRun`, and all 64 Loki fallback pointers match those raw stacks. The arithmetic closes as `2 steps * 3 CNs * DOP 16 = 96` delete sources, with `2 steps * 2 remote CNs * DOP 16 = 64` misplaced sources.
- The same snapshots show 2048 coordinator notify/receive waits and 1024 `GetProcByUuid` waits on each remote CN. Counts and stacks are unchanged at `14:50`, after about 700 minutes.

This also reconciles the two conflicting observations:

- `time of Exec.Run` is logged only after `runner.Run(0)` returns; the raw frontend stack shows that call still blocked, so zero completion-timing logs are expected and do not mean zero execution.
- The cited salt timeout belongs to another connection, was already recurring before restore started, and is absent from the target restore session.

I also posted the sanitized correction on #26526. No code change was needed for this review because your code-side review found no blocker; the requested change was addressed by completing the incident evidence chain. Please re-review when convenient.

LeftHandCold (approved):
Re-reviewed the code at head 93bd29cb. I found no production-reachable blocker in the per-CN distributed-delete grouping, bucket and UUID routing preservation, encoded-tree standalone validation, dispatch pre-registration, cancellation/cleanup paths, or compile-time performance. Seventeen focused topology, encoding, registration, and unhappy-path tests passed with bounded local concurrency. My earlier issue-attribution concern is non-blocking for this code review and should not gate this patch.

PR #26534, opened 2026-07-31
fix(plan): infer prepared SUM and AVG parameters

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #26019

Prepared SUM and AVG bound a bare parameter marker as TEXT before aggregate overload resolution, so SQL and binary PREPARE failed before an execution value could be supplied.

This change applies the existing prepared numeric-inference contract to SUM and AVG in direct and window aggregates. It also propagates aggregate numeric intent through derived tables and ordinary or recursive CTE projections before those sources are bound. Compatible targets from multiple consumers are merged to a safe common numeric type; existing assignment-conflict behavior is unchanged.

The parameter keeps the existing dynamic T_any / MYSQL_TYPE_NULL transport metadata and is cast to the inferred computation type in the plan. Ordinary VARCHAR aggregate inputs and explicit character casts remain rejected, while explicit numeric casts keep their declared type. Repeated fill preserves the canonical prepared plan.

Tested with:

- full pkg/sql/plan and pkg/frontend tests
- full pkg/sql/plan and pkg/frontend tests under race
- focused race stress, count 100, for planner and binary metadata regressions
- configured go build and go vet for both packages
- git diff --check
- added distributed SQL fixture for direct, window, recursive, repeated numeric-string, NULL, and invalid-string execution; local mo-tester was unavailable


XuPeng-SH (approved):
Deep-reviewed exact head dc83a3b393fb518bcec126edac35a701d1d96d25. The fix establishes a static numeric contract for SUM/AVG markers, keeps transport metadata dynamic, and propagates only parameter-bearing numeric intent through derived/CTE projections. Ordinary VARCHAR aggregate inputs and explicit character casts remain rejected; explicit numeric casts, recursive CTE common types, window aggregates, repeated fill, and multiple compatible consumers preserve their contracts. I found no blocking correctness, unhappy-path, state-ownership, or material performance issue. Local verification: focused planner and binary-protocol tests under race count=100; full pkg/sql/plan and pkg/frontend under race; scoped go vet; diff check; latest CI green.

PR #26537, opened 2026-07-31
fix(hakeeper): preserve CN load across batch scheduling

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25833

HAKeeper rebuilt a filtered CN pool for every Created task, but that pool reset the Running-task frequencies. A successful allocation then updated only the temporary pool, so the next task in the same scheduling round saw the same stale load snapshot and could select the same CN repeatedly.

This PR:

- preserves source frequencies when filtering a CN pool;
- selects from the per-task eligible view but records a successful allocation in the shared round pool;
- leaves failed allocations out of the load count;
- adds regressions for existing load, batch scheduling, label/resource filtering, and allocation failure.

The change does not alter label/resource eligibility, expired-CN handling, task persistence, timeout behavior, or tie-breaking.

Tested with:

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=100 -run '<each focused regression>' ./pkg/hakeeper/task`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 ./pkg/hakeeper/task`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -run '^TestTaskSchedulerCan(ScheduleTasksToCNs|ReScheduleExpiredTasks)$' ./pkg/logservice`

`go build -mod=readonly ./pkg/hakeeper/task`

`go vet -mod=readonly ./pkg/hakeeper/task`

### DIFF
--- a/pkg/hakeeper/task/cn_selector.go
+++ b/pkg/hakeeper/task/cn_selector.go
@@ -61,9 +61,11 @@ func (p *cnPool) selectCNs(rules ...rule) *cnPool {
 	newPool := newCNPool()
 	for _, cn := range p.sortedCN {
 		if matchAllRules(cn.info, rules...) {
-			heap.Push(newPool, cn)
+			newPool.freq[cn.uuid] = p.getFreq(cn.uuid)
+			newPool.sortedCN = append(newPool.sortedCN, cn)
 		}
 	}
+	heap.Init(newPool)
 	return newPool
 }
 
--- a/pkg/hakeeper/task/cn_selector_test.go
+++ b/pkg/hakeeper/task/cn_selector_test.go
@@ -192,3 +192,27 @@ func TestContains(t *testing.T) {
 	assert.True(t, contains([]string{"a"}, "a"))
 	assert.False(t, contains([]string{"a"}, "b"))
 }
+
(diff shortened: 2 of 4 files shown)


XuPeng-SH (approved):
Approved exact head 37bcb7473cf96a3471fcc988ed9d8f886a63edf8 after full scheduling-closure review.

The fix preserves the running-task frequency in each eligible candidate view and records only successful allocations in the shared per-round pool, so later Created tasks observe earlier reservations. Label/resource/expiry filtering remains unchanged. The new selectCNs construction also replaces repeated heap insertion with a linear copy plus heap.Init.

Reviewed ordinary, filtered, existing-load, allocation-failure, expired-CN, and dependent logservice paths. No remaining correctness, lifecycle, unbounded-growth, or material performance blocker found.

Local validation on the exact head: four focused regressions each passed under race at count=100; the full pkg/hakeeper/task package passed under race; dependent task-scheduler logservice regressions passed under race; scoped go vet passed.

PR #26538, opened 2026-07-31
fix(gossip): disable unsupported push-pull route snapshots

LeftHandCold:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [x] Test and CI
- [ ] Code Refactoring

issue #25942

The gossip delegate had a route-snapshot codec that was not part of MatrixOne's effective memberlist protocol: join exchanges called `LocalState(true)`, which already returned nil, while periodic exchanges called `MergeRemoteState(..., false)`, which already ignored the snapshot. Honest nodes therefore never consumed this state, but malformed join state could still reach the decoder and trigger a short-buffer panic or an allocation controlled by its inner length header.

This PR makes that existing protocol decision explicit:

- `LocalState` returns nil for both push/pull modes;
- `MergeRemoteState` ignores remote user state without parsing or allocating;
- the dead binary/LZ4 snapshot codec and state builders are removed;
- the live incremental `GetBroadcasts` / `NotifyMsg` / `NotifyLeave` route path is unchanged.

The effective old/new rolling-upgrade behavior is unchanged. This PR is scoped to MatrixOne's delegate state and does not change the pinned memberlist transport/framing layer.

Tested with:

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=100 -run '^TestDelegatePushPullStateDisabled$' ./pkg/gossip`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=45 -run '^TestNodeGossip$' ./pkg/gossip`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 ./pkg/gossip`

`./.agents/skills/mo-dev/scripts/mo-cgo-test -count=1 -run '^$' ./pkg/cnservice`

`go build -mod=readonly ./pkg/gossip`

`go vet -mod=readonly ./pkg/gossip`

### DIFF
--- a/pkg/gossip/delegate.go
+++ b/pkg/gossip/delegate.go
@@ -15,7 +15,6 @@
 package gossip
 
 import (
-	"encoding/binary"
 	"fmt"
 
 	"github.com/hashicorp/memberlist"
@@ -23,12 +22,7 @@ import (
 	"github.com/matrixorigin/matrixone/pkg/pb/query"
 	"github.com/matrixorigin/matrixone/pkg/pb/statsinfo"
 	"github.com/matrixorigin/matrixone/pkg/queryservice/client"
-	"github.com/pierrec/lz4/v4"
 	"go.uber.org/zap"
-)
-
-var (
-	binaryEnc = binary.BigEndian
 )
 
 type Module interface {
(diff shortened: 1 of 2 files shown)


XuPeng-SH (approved):
Approved exact head 9abe377d6ceda02ed0fe9258c5c0e9cd63a6be63 after tracing the complete memberlist push/pull and live route-propagation paths.

MatrixOne sets PushPullInterval to zero. Join exchanges propagate join=true to both LocalState calls, and the old LocalState(true) already returned nil; non-join state was generated but ignored by MergeRemoteState(..., false). The removed snapshot codec therefore had no honest producer/consumer closure. Rolling old/new behavior remains unchanged, while arbitrary delegate state can no longer reach the short-buffer decoder or its inner length-controlled allocation. Incremental GetBroadcasts/NotifyMsg and NotifyLeave routing are untouched.

Q1-Q3 review found no remaining blocker; this also removes periodic serialization/compression/locking work. Local validation: the disabled-state regression passed under race at count=100; TestNodeGossip passed under race at adaptive count=50; the full pkg/gossip package passed under race; pkg/cnservice compiled through go test; scoped go vet passed.

XuPeng-SH (approved):
Revalidated exact head 8e5a2ccdc3004067c2e0d78142dc66965d676219 after the automatic main merge. The only old-head-to-new-head delta is main commit 4fe0c66d718 (TAE transfer-page synchronization); pkg/gossip and pkg/cnservice are byte-identical to the previously reviewed head, and the net PR diff against the updated base is unchanged. Re-ran disabled-state race count=100, full pkg/gossip race, pkg/cnservice compile, and gossip vet: all pass. No failed CI checks; refreshed jobs are still running.

PR #26540, opened 2026-07-31
fix(frontend): honor zero data branch diff limits

### LINKED ISSUE #26034: [Bug]: DATA BRANCH DIFF OUTPUT LIMIT 0 returns one row
### Summary

`DATA BRANCH DIFF ... OUTPUT LIMIT 0` returns one diff row instead of an empty result.

### Version

- Branch: `main`
- Commit: `e15943929885c2c5c83daed90f0f8caeb9a1f0d0`

### Reproduction

```sql
DROP DATABASE IF EXISTS g4d_limit;
CREATE DATABASE g4d_limit;
USE g4d_limit;

CREATE TABLE base(id INT PRIMARY KEY, v INT);
INSERT INTO base VALUES (1,10),(2,20),(3,30);
DATA BRANCH CREATE TABLE child FROM base;
UPDATE child SET v = v + 1;
INSERT INTO child VALUES (4,40);

DATA BRANCH DIFF child AGAINST base OUTPUT LIMIT 0;
```

The result contains one row, for example:

```text
diff child against base  flag    id  v
child                    UPDATE  1   11
```

The behavior reproduced in 3 out of 3 clean runs. `OUTPUT LIMIT 1` and `OUTPUT LIMIT 2` returned one and two rows respectively, and a negative limit was rejected by the parser.

### Expected behavior

`OUTPUT LIMIT 0` should return zero rows.

### Actual behavior

`OUTPUT LIMIT 0` returns one row.

### Code observation

In `satisfyDiffOutputOpt`, the current row is appended before the limit condition is checked. For a zero limit, the first row is therefore added before the producer is stopped.

gouhongshen:
## What type of PR is this?

- [ ] API-change
- [x] BUG
- [ ] Improvement
- [ ] Documentation
- [ ] Feature
- [ ] Test and CI
- [ ] Code Refactoring

Fixes #26034

`DATA BRANCH DIFF ... OUTPUT LIMIT 0` appended the first diff row before checking whether the output limit had already been reached. As a result, a valid zero limit returned one row.

This PR:

- centralizes the row-limit predicate and evaluates it before consuming rows, stopping producers immediately when the limit is zero while retaining the existing channel-drain and batch-recycle protocol;
- adds a focused unit test for limits 0, 1, the exact row count, and a limit above the row count, including producer-stop behavior;
- adds BVT coverage for zero limits with and without column projection, plus adjacent and upper boundaries;
- makes the `LIMIT 1` BVT setup deterministic by testing it while exactly one row differs. This avoids asserting which row an unordered multi-row diff produces first while preserving the row-count boundary coverage.

CI follow-up:

- The first PESSIMISTIC BVT run failed because the new `LIMIT 1` case expected row `id=1`, while the unordered diff validly returned `id=2`. The test setup now has exactly one differing row at that boundary.
- The Coverage job in that workflow was a dependent failure (`coverage merge skipped because at least one required producer was skipped or failed`), not an independent product failure.

Validation:

- `GOWORK=off go build -mod=readonly ./pkg/frontend`
- `GOWORK=off go vet -mod=readonly ./pkg/frontend`
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -json -count=1 -timeout=120s -run '^TestDataBranchOutputLimitBoundaries$' ./pkg/frontend`
- focused race stress: `T=0.01s`, `B=30s`, `N=100`; all 100 runs passed
- `.agents/skills/mo-dev/scripts/mo-cgo-test -race -count=1 -timeout=300s ./pkg/frontend`
- BVT `branch_output_limit_zero`: 15/15 statements passed at 100% in each of 5 consecutive runs; generated output was manually checked for 0/1/4/5 row semantics

Residual risk: none known. Parsing and positive-limit semantics are unchanged; negative limits remain parser-rejected. Diff row order remains unspecified, and the regression test does not impose an ordering contract.

### DIFF
--- a/pkg/frontend/data_branch_output.go
+++ b/pkg/frontend/data_branch_output.go
@@ -806,6 +806,10 @@ func satisfyDiffOutputOpt(
 		var (
 			rows = make([][]any, 0, 100)
 		)
+		limitReached := func() bool {
+			return stmt.OutputOpt != nil && stmt.OutputOpt.Limit != nil &&
+				int64(len(rows)) >= *stmt.OutputOpt.Limit
+		}
 
 		// Resolve column projection (nil means show all visible columns).
 		displayIdxes, resolveErr := resolveProjectedIdxes(stmt.Columns, tblStuff)
@@ -834,6 +838,12 @@ func satisfyDiffOutputOpt(
 			}
 			// Sparse row: indexed by colIdx+2, so size must accommodate the largest index.
 			rowSize = slices.Max(extractIdxes) + 3
+		}
+
+		if limitReached() {
+			// The limit is already satisfied before consuming any rows (LIMIT 0).
+			hitLimit = true
+			stop()
 		}
 
 		for wrapped := range retCh {
@@ -869,8 +879,7 @@ func satisfyDiffOutputOpt(
 				}
 
 				rows = append(rows, row)
-				if stmt.OutputOpt != nil && stmt.OutputOpt.Limit != nil &&
-					int64(len(rows)) >= *stmt.OutputOpt.Limit {
+				if limitReached() {
 					// hit limit, cancel producers but keep draining the channel
 					hitLimit = true
 					stop()
--- a/pkg/frontend/data_branch_output_test.go
+++ b/pkg/frontend/data_branch_output_test.go
@@ -746,6 +746,61 @@ func TestDataBranchOutputBuildOutputSchema(t *testing.T) {
 	})
 }
 
+func TestDataBranchOutputLimitBoundaries(t *testing.T) {
+	tests := []struct {
+		name      string
+		limit     int64
+		wantRows  uint64
+		wantStops int
+	}{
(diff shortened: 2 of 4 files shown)


XuPeng-SH (approved):
Deep-reviewed exact head 7ee509137922722402c7738f7c4bb1c0d6753fd3. The boundary predicate is now evaluated before row consumption, so LIMIT 0 is correct without changing positive-limit semantics. Lifecycle closure is intact: stop is sync.Once-guarded, producers observe stopCh both before and during send, the consumer keeps draining retCh, every queued batch is recycled, the producer closes out before retCh is closed, and both goroutines are waited. The BVT also avoids imposing an order on unordered diff output. No blocking correctness, leak/hang/double-close, compatibility, or performance issue found. Local verification: focused boundary test under race count=100, full pkg/frontend under race, scoped go vet, diff check, latest CI green.

